PR #48, opened 2018-02-28
Refactor HID, fix issues

gdkchan:
This is a small rewrite of the HID class, the motivations for this, and issues this fixes are:

- Fix a bug where the button would alternate between pressed/released state even if the user held the key.
- Removed the use of Marshaling - this is, afaik, slow. Instead we store offsets/sizes as constant values, data is only written when needed.
- Removed some structs that where not being used.
- Moved everything related to HID to the Ryujinx.Core.Input namespace, and all files to the respective folder.

This is just the first part. There are still some things that needs to be addressed, like multiple hid shared memory mappings, making names used on the enums consistent with what is used on the rest of the code, and so on.

This has both joycon and touchscreen support, but I was only able to test joycon input because there are no homebrew using the touch screen currently afaik.

fyi @AcK77 @emmauss


Ryujinx.Core/Hid/Hid.cs
```diff
@@ -0,0 +1,218 @@
...
+            long BaseControllerOffset = HidControllersOffset + 8 * HidControllerSize;
+
+            HidControllerType Type =
+                HidControllerType.ControllerType_Handheld |
+                HidControllerType.ControllerType_JoyconPair;
+
+            HidControllerColorDesc ColorDesc =
+                HidControllerColorDesc.ColorDesc_ColorsNonexistent;
+
+            WriteInt32(BaseControllerOffset + 0x0,  (int)Type);
```
VPeruS: Maybe Write could adjust BaseControllerOffset for next write?

emmauss:
while we are on it, what do we do about the touchscreen resolution?

emmauss:
looks good.

gdkchan:
For the touch screen problem -- we can just divide the mouse coodinates by the window size and multiply by 1280/720.

emmauss:
same thing I did in the other pr, mathwise, just in a different way. there may be very minute difference due to the order, but we are still truncating them in the end, so at worse it will be off by 1.

Ryujinx.Core/Hid/Hid.cs
gdkchan: this would be possible if WriteInt32 used ref on the Position and add the offset automatically. I don't feel like we gain a lot doing that tbh.

gdkchan:
I think I already addressed all issues, so let's get this merged.

Ryujinx.Core/Hid/Hid.cs
VPeruS: Yeap, it just remove by hand calculation

PR #50, opened 2018-03-02
Improve CPU initial translation speeds

gdkchan:
This enables background translation of subroutines on the CPU.
First, it trasnlates a slow version, only considering basic blocks. This runs slow but is fast to translate.
In the background, it translates a faster version of the same code. When the translation finishes, it swaps the slow version with the fast version.
This should improve startup speed.

A few things still needs to be done, like removing cold methods to avoid memory leaking. It can also be improved by not recompiling a faster version of methods that only runs once, or rarely runs.

### DIFF
--- a/ChocolArm64/ATranslatedSub.cs
+++ b/ChocolArm64/ATranslatedSub.cs
@@ -31,7 +31,7 @@ namespace ChocolArm64
 
         private bool NeedsReJit;
 
-        private int MinCallCountForReJit = 5000;
+        private int MinCallCountForReJit = 250;
 
         public ATranslatedSub(DynamicMethod Method, List<ARegister> Params, HashSet<long> Callees)
         {
--- a/ChocolArm64/ATranslatedSub.cs
+++ b/ChocolArm64/ATranslatedSub.cs
@@ -14,8 +14,6 @@ namespace ChocolArm64
 
         private AA64Subroutine ExecDelegate;
 
-        private bool HasDelegate;
-
         public static int StateArgIdx  { get; private set; }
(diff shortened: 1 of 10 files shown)


gdkchan:
After some testing from me and other devs, I decided to make a few changes to the initial implementation:

- It doesn't do "background translation" anymore. It doesn't seem to be really worth it, since .net JIT still need to run when the method is called for the frist time, and this still takes a significant amount of time (resulting in lags). It's probably still a bit better than doing it on the same thread, but I don't feel that the added code complexity is worth it.
- Translated subroutines are now separated as 3 types, "SubBlock", "Tier0" and "Tier1". SubBlock is used for "basic blocks" that are not the entry point of a subroutine. Tier0 is used for the entry point of subroutines. Tier1 is used for fully translated subroutines.
-- TranslatedSubs of the type SubBlock should never be re-jited.
-- TranslatedSubs of the type Tier0 are re-jited only if the subroutine is "hot", that is, only if it is called a lot of times. Currently this is triggered by the hardcoded value of 5000 calls minimum.
-- TranslatedSubs of the type Tier1 are re-jited if the code can be improved, like for example, if a subroutine that it calls has now been jited aswell, and it can just call said subroutine directly, instead of returning to the dispatcher.

Theres still some room for improvement, and some TODOs. However, hopefully this is good enough as a initial implementation of the feature.

Ghabry:
Wow this made the easyrpg-player.nro startup superfast for me.
Went down on my slow system from ~10 minutes to 5 seconds 👍  awesome.

Issue #55, opened 2018-03-13
System.NotImplementedException: Tid 1 ServiceHid 124


Cpasjuste:
Hi,

Since this commit : https://github.com/switchbrew/libnx/commit/e7a4eba71c2aada52182a977c2068667b16b9f2a , Ryujinx is not working anymore with libnx. There's probably an "easy" fix, but i'm not sure i should take a closer look myself as i'm not used to it : 

` 
00000000ms | INFO  > Loading as homebrew.
 00000066ms | INFO  > Image base at 0x0000000008000000.
 00002269ms | INFO  > HID shared memory successfully mapped to 0x000000000b6d7000!

Unhandled Exception: System.NotImplementedException: Tid 1 ServiceHid 124
   at Ryujinx.Core.OsHle.Ipc.IpcHandler.IpcCall(Switch Ns, Process Process, AMemory Memory, HSession Session, IpcMessage Request, Int32 ThreadId, Int64 CmdPtr, Int32 HndId) in /home/cpasjuste/dev/switch/Ryujinx/Ryujinx.Core/OsHle/Ipc/IpcHandler.cs:line 119
   at Ryujinx.Core.OsHle.Svc.SvcHandler.SendSyncRequest(AThreadState ThreadState, Boolean UserBuffer) in /home/cpasjuste/dev/switch/Ryujinx/Ryujinx.Core/OsHle/Svc/SvcSystem.cs:line 129
   at Ryujinx.Core.OsHle.Svc.SvcHandler.SvcCall(Object sender, AInstExceptionEventArgs e) in /home/cpasjuste/dev/switch/Ryujinx/Ryujinx.Core/OsHle/Svc/SvcHandler.cs:line 86
   at Sub0000000008cd42e8(AThreadState , AMemory )
   at ChocolArm64.ATranslator.ExecuteSubroutine(AThread Thread, Int64 Position) in /home/cpasjuste/dev/switch/Ryujinx/ChocolArm64/ATranslator.cs:line 65
   at ChocolArm64.AThread.<Execute>b__18_0() in /home/cpasjuste/dev/switch/Ryujinx/ChocolArm64/AThread.cs:line 45
   at System.Threading.Thread.ThreadMain_ThreadStart()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
`

gdkchan:
[...]

PR #57, opened 2018-03-13
stub some hid service functions

emmauss:
for #55 , allows tetris to boot

### DIFF
--- a/Ryujinx.Core/OsHle/Services/Hid/ServiceHid.cs
+++ b/Ryujinx.Core/OsHle/Services/Hid/ServiceHid.cs
@@ -1,5 +1,6 @@
 using Ryujinx.Core.OsHle.Ipc;
 using System.Collections.Generic;
+using Ryujinx.Core.Input;
 
 using static Ryujinx.Core.OsHle.IpcServices.ObjHelper;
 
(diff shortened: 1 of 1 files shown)


Ryujinx.Core/OsHle/Services/Hid/ServiceHid.cs
```diff
@@ -87,6 +90,40 @@ public long GetNpadJoyHoldType(ServiceCtx Context)
             return 0;
         }
 
+        public long SetNpadJoyAssignmentModeSingleByDefault(ServiceCtx Context)
+        {
+            long Unknown = Context.RequestData.ReadInt32();
```
gdkchan: Those values are not unknown, they are Id of type HidControllerID, see libnx implementation on hid.c. (more occurrences below).

emmauss: okay, i will check on it. I wonder why switchbrew was not updated.

emmauss: okay, i will check on it. I wonder why switchbrew was not updated.

emmauss:
updated. I do not know what the unknowns in the final method are.

gdkchan:
the controller id value is a 32 bits int, not a 64 bits long, also would be nice if you could use the HidControllerID type inside Ryujinx.Core.Hid (you just need to cast the int to HidControllerID, and use HidControllerID on the variable type too).

gdkchan:
lgtm, thanks.

PR #61, opened 2018-03-21
Rename some fsp calls to current names

emmauss:


gdkchan:
Ah, those so the old names weren't the official names. Looks good, thanks!

PR #62, opened 2018-03-21
Add Frint Instructions and Tests

MS-DOS1999:
All tests pass except NaN tests


gdkchan:
Good work. It seems that the current FRINTI implementation is exactly the same of the FRINTX instruction. According to the manual the difference is that FRINTX should raise a inexact exception if the result is not equal to the source operand. There are other cases where the difference between two instructions is some exception being raised aswell and it being ignored, I'm a bit worried that this could cause problems in the future. So if you could implement the exception behaviour (on FRINTX) would be nice, however it's fine to do that later too since I didn't implement it either.

About the NaN test, can you skip it by default like we did for FMAX/FMIN?

MS-DOS1999:
Sorry for 3 commits that failed. I am currently forced to edit the code with my phone so I couldn't check by compiling before commit x)

gdkchan:
lgtm, thanks!

PR #71, opened 2018-04-05
Show calling information in logs

emmauss:
This add the calling information to the logs, and also moves the actual printing of the log to one method. calling information is show for all log levels. ~~Only the relative path in the source code structure is shown~~ Only the calling method is shown in logs, although the log entries contains full information about the log.

Example log = https://pastebin.com/Aqthk5kM

### DIFF
--- a/Ryujinx.Core/Logging.cs
+++ b/Ryujinx.Core/Logging.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace Ryujinx.Core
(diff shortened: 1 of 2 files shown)


emmauss:
The repeated lines is just the message set by `SvcOutputDebugString` which included itself.

gdkchan:
I'd like to hear the opinion of other devs aswell. Imo this is not necessary, those not familiar with the code can just do a quick search to find the function name. It bloats the log with all those paths. Usually the important information is the values being logged.

emmauss:
i cut down on the amount of information that is shown in the logs, to only the calling method.

gdkchan:
Looks like `CallingFile` is not used anymore after your last update. Can you remove it since its not being used? `CallingLine` also doesn't seem to be used. I'm fine with including the line number on the log, if you want. But if you don't, then please remove it aswell.

emmauss:
remove the callingfile and added line number to the logs

emmauss:
fixed

gdkchan:
lgtm, thanks!

PR #72, opened 2018-04-05
Implement Frsqrte_S

MerryMage:


ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    a = (a << 1) + 1;
+                }
+                else
+                {
+                    a = (a | 1) << 1;
+                }
+
+                ulong b = 256;
+                while (a * (b+1) * (b+1) < (1ul << 28))
```
gdkchan: nit: Can you change to `(b + 1)` for consistency?

gdkchan:
Thanks for implementing this! Would be nice to have the vector variant implemented aswell, since some games are currently requiring it.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    scaled <<= 1;
+                    x_exp--;
+                }
+                scaled <<= 1;
+            }
+
+            if (((ulong)x_exp & 1) == 1)
+            {
+                scaled >>= 45;
```
gdkchan: From what I understood, according to the manual this should be `>>= 44` no?

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -0,0 +1,103 @@
...
+                scaled &= 0xFF;
+                scaled |= 0x80;
```
gdkchan: Shouldn't this be |= 0x100?

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+            else
+            {
+                scaled >>= 44;
+                scaled &= 0xFF;
+                scaled |= 0x100;
```
gdkchan: Shouldn't this be |= 0x200?

ChocolArm64/Instruction/ASoftFloat.cs
MerryMage: I have the `if` statement switched round compared to the manual.

ChocolArm64/Instruction/ASoftFloat.cs
MerryMage: No, this and the below are correct.

gdkchan:
My confusion came from the fact that the meriac documentation and the official manual are actually different. The official docs works with floating point numbers while meriac uses fixed point.

Code from ARM manual:
```
if exp<0> == '0' then
    scaled = '0' : '01111111110' : fraction<51:44> : Zeros(44);
else
    scaled = '0' : '01111111101' : fraction<51:44> : Zeros(44);

double recip_sqrt_estimate(double a)
{
    int q0, q1, s;
    double r;
    if (a < 0.5) /* range 0.25 <= a < 0.5 */
    {
        q0 = (int)(a * 512.0); /* a in units of 1/512 rounded down */
        r = 1.0 / sqrt(((double)q0 + 0.5) / 512.0); /* reciprocal root r */
    }
    else /* range 0.5 <= a < 1.0 */
    {
        q1 = (int)(a * 256.0); /* a in units of 1/256 rounded down */
        r = 1.0 / sqrt(((double)q1 + 0.5) / 256.0); /* reciprocal root r */
    }
    s = (int)(256.0 * r + 0.5); /* r in units of 1/256 rounded to nearest */
    return (double)s / 256.0;
}
```

gdkchan:
lgtm, thanks for the contribution!

PR #74, opened 2018-04-08
Add FMUL (scalar, by element) instruction; add FRECPE, FRECPS (scalar & vector) instructions. Add 5 simple tests.

LDj3SNuD:


ChocolArm64/AOpCodeTable.cs
```diff
@@ -225,12 +225,17 @@ static AOpCodeTable()
             Set("000111110x0xxxxx1xxxxxxxxxxxxxxx", AInstEmit.Fmsub_S,       typeof(AOpCodeSimdReg));
             Set("000111100x1xxxxx000010xxxxxxxxxx", AInstEmit.Fmul_S,        typeof(AOpCodeSimdReg));
             Set("0>1011100<1xxxxx110111xxxxxxxxxx", AInstEmit.Fmul_V,        typeof(AOpCodeSimdReg));
+            Set("010111111<<xxxxx1001x0xxxxxxxxxx", AInstEmit.Fmul_Se,       typeof(AOpCodeSimdRegElemF));
```
gdkchan: The names are organized in alphabetical order so this should be below Fmul_S (applies to the method inside AInstEmitSimdArithmetic.cs aswell).

gdkchan:
lgtm, thanks!

@MerryMage was working on a soft float implementation that should be more accurate to what the instruction does on the arm64 cpu. When her implementation is ready we will replace it. The tests will most likely need to be updated aswell since the instruction computes an estimate and the result is probably slightly different.

PR #78, opened 2018-04-12
[GPU] Fix some of the current framebuffer issues

gdkchan:
The goal here is fixing the frame buffer related issues on the few games that actually renders something.

The two things that are left to do is re-adding support for drawing the raw frame buffer (since homebrew needs that) and drawing the final image to a temp texture and then to the screen when the render method is called.


gdkchan:
Well, theres still a bunch of issues left, and a bunch of workarounds still needed here and there unfortunately... But at least, should be better than it was before so it should be good to merge.

PR #79, opened 2018-04-13
Add logclass, made changes to logging calls

emmauss:
This adds log classes, and the option to filter logs based on them in config.
Filtering by a main log class will show all subclass logs, no matter if they are enabled or not. Like, filtering by `Kernel` will also show all `KernelScheduler`, `KernelIpc` and `KernelSvc` logs

### DIFF
--- a/Ryujinx.Core/Config.cs
+++ b/Ryujinx.Core/Config.cs
@@ -9,15 +9,17 @@ namespace Ryujinx.Core
 {
     public static class Config
     {
-        public static bool EnableMemoryChecks   { get; private set; }
-        public static bool LoggingEnableInfo    { get; private set; }
-        public static bool LoggingEnableTrace   { get; private set; }
-        public static bool LoggingEnableDebug   { get; private set; }
-        public static bool LoggingEnableWarn    { get; private set; }
-        public static bool LoggingEnableError   { get; private set; }
-        public static bool LoggingEnableFatal   { get; private set; }
-        public static bool LoggingEnableIpc     { get; private set; }
-        public static bool LoggingEnableLogFile { get; private set; }
+        public static bool   EnableMemoryChecks     { get; private set; }
+        public static bool   LoggingEnableInfo      { get; private set; }
+        public static bool   LoggingEnableTrace     { get; private set; }
+        public static bool   LoggingEnableDebug     { get; private set; }
+        public static bool   LoggingEnableWarn      { get; private set; }
+        public static bool   LoggingEnableError     { get; private set; }
+        public static bool   LoggingEnableFatal     { get; private set; }
+        public static bool   LoggingEnableIpc       { get; private set; }
+        public static bool   LoggingEnableLogFile   { get; private set; }
+        public static bool   LoggingEnableFilter    { get; private set; }
+        public static bool[] LoggingFilteredClasses { get; private set; }
 
         public static JoyCon FakeJoyCon { get; private set; }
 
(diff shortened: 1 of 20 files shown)


Ryujinx.Core/Config.cs
```diff
@@ -27,15 +29,32 @@ public static void Read()
...
+            LoggingEnableFatal     = Convert.ToBoolean(Parser.Value("Logging_Enable_Fatal"));
+            LoggingEnableIpc       = Convert.ToBoolean(Parser.Value("Logging_Enable_Ipc"));
+            LoggingEnableLogFile   = Convert.ToBoolean(Parser.Value("Logging_Enable_LogFile"));
+            LoggingEnableFilter    = Convert.ToBoolean(Parser.Value("Logging_Enable_Filter"));
+            LoggingFilteredClasses = new bool[(int)LogClass.Count];
+
+            string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes",string.Empty).Split(',');
+            foreach(string LogClass in FilteredLogClasses)
+            {
+                if(Enum.TryParse(typeof(LogClass),LogClass,out object parsedLogClass))
```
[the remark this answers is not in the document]
emmauss: it was a typo. I was checking if the value in the config is a valid logclass.

Ryujinx.Core/Config.cs
```diff
@@ -27,15 +29,32 @@ public static void Read()
...
+            LoggingFilteredClasses = new bool[(int)LogClass.Count];
+
+            string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes", string.Empty).Split(',');
+            foreach (string LogClass in FilteredLogClasses)
+            {
+                if (Enum.TryParse(typeof(LogClass), LogClass, out object ParsedLogClass))
+                {
+                    foreach (LogClass EnumItemName in Enum.GetValues(typeof(LogClass)))
+                    {
+                        if (EnumItemName.ToString().Contains(ParsedLogClass.ToString()))
```
gdkchan: Can't you remove the `if (Enum.TryParse(typeof(LogClass), LogClass, out object ParsedLogClass))` and use `LogClass.Trim()` instead of `ParsedLogClass.ToString()` here?

emmauss: so as long as the value partially/fully matches values in the enumuration, it will enable that value?

emmauss:
allowed partial matching of filters

Ryujinx/Ryujinx.conf
```diff
@@ -22,6 +22,12 @@ Logging_Enable_Fatal = true
 #Enable print Ipc logs
 Logging_Enable_Ipc = false
 
+#Enable log filter
+Logging_Enable_Filter = true
+
+#Filtered log classes, seperated by ',', eg. `Logging_Filtered_Classes = Loader,ServiceFS`
+Logging_Filtered_Classes = s
```
gdkchan: What is the `s` here?

emmauss: forgot it was there.

Ryujinx.Core/Config.cs
```diff
@@ -27,15 +29,30 @@ public static void Read()
...
+            LoggingEnableFatal     = Convert.ToBoolean(Parser.Value("Logging_Enable_Fatal"));
+            LoggingEnableIpc       = Convert.ToBoolean(Parser.Value("Logging_Enable_Ipc"));
+            LoggingEnableLogFile   = Convert.ToBoolean(Parser.Value("Logging_Enable_LogFile"));
+            LoggingEnableFilter    = Convert.ToBoolean(Parser.Value("Logging_Enable_Filter"));
+            LoggingFilteredClasses = new bool[(int)LogClass.Count];
+
+            string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes", string.Empty).Split(',');
+            foreach (string LogClass in FilteredLogClasses)
+            {
+                if(!string.IsNullOrEmpty(LogClass))
```
gdkchan: I missed that. This if is a bit confusing, can you add the curly braces so its clear that this is a pre-condition for the foreach?

PR #87, opened 2018-04-17
Update IAudioDeviceService.cs

AcK77:
Stubs:
- QueryAudioDeviceSystemEvent
- GetActiveChannelCount

### DIFF
--- a/Ryujinx.Core/OsHle/Services/Aud/IAudioDeviceService.cs
+++ b/Ryujinx.Core/OsHle/Services/Aud/IAudioDeviceService.cs
@@ -13,6 +13,8 @@ namespace Ryujinx.Core.OsHle.Services.Aud
         public override IReadOnlyDictionary<int, ServiceProcessRequest> Commands => m_Commands;
 
         private KEvent SystemEvent;
+        //TODO: We shouldn't be signaling this here.
+        SystemEvent.Handle.Set();
 
         public IAudioDeviceService()
         {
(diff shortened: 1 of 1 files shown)


Ryujinx.Core/OsHle/Services/Aud/IAudioDeviceService.cs
```diff
@@ -45,6 +52,9 @@ public long ListAudioDeviceName(ServiceCtx Context)
                 Position += Buffer.Length;
             }
 
+            //TODO: We shouldn't be signaling this here.
+            SystemEvent.Handle.Set();
```
gdkchan: It's weird that this is signaled here. Can't you signal right after `SystemEvent = new KEvent();`? Feels less hacky this way at least :P

Ryujinx.Core/OsHle/Services/Aud/IAudioDeviceService.cs
```diff
@@ -11,13 +12,21 @@ class IAudioDeviceService : IpcService
 
         public override IReadOnlyDictionary<int, ServiceProcessRequest> Commands => m_Commands;
 
+        private KEvent SystemEvent;
+        //TODO: We shouldn't be signaling this here.
+        SystemEvent.Handle.Set();
```
gdkchan: I meant below `SystemEvent = new KEvent();`, not the variable declaration. I think this code doesn't compile as this call outside of a function is not valid code.

AcK77:
Sorry for the mistake xD!

gdkchan:
lgtm, thanks!

PR #88, opened 2018-04-18
Add ABS (scalar & vector), ADD (scalar), NEG (scalar) instructions.

LDj3SNuD:
Abs is Int64.MinValue compliant.

(All instructions have been successfully tested.)


gdkchan:
lgtm, thanks!

I think that `x  = x >= 0 ? x : -x` is probably better than using Math.Abs (which I used for some instructions) for performance. I will change the other instructions later to be consistent.

PR #96, opened 2018-04-21
Fix Addp_S in AOpCodeTable. Add 5 Tests: ADDP (scalar), ADDP (vector), ADDV.

LDj3SNuD:
Update Tester as well.


gdkchan:
lgtm, thanks! Looks like there are a few issues of this error, where I used the variable-size encoding on instructions where it isn't valid.

PR #99, opened 2018-04-21
Some implementations

AcK77:
- ICommonStateGetter
  * GetBootMode
- ISelfController
  * SetHandlesRequestToDisplay
- IServiceGetterInterface
- ISystemUpdateInterface
- IVulnerabilityManagerInterface
- IPrepoService
- ISettingsServer
  * GetLanguageCode
- ISystemSettingsServer
  * GetFirmwareVersion2
- IHOSBinderDriver
  * TransactParcelAuto

### DIFF
--- a/Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
+++ b/Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
@@ -34,9 +34,17 @@ namespace Ryujinx.Core.OsHle.Services.Set
         {
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
+                { 0, GetLanguageCode },
                 { 1, GetAvailableLanguageCodes     },
                 { 3, GetAvailableLanguageCodeCount }
             };
+        }
+
+        public static long GetLanguageCode(ServiceCtx Context)
+        {
+            Context.ResponseData.Write(1L); //en-US
+
+            return 0;
         }
 
         public static long GetAvailableLanguageCodes(ServiceCtx Context)
--- a/Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
+++ b/Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
@@ -1,5 +1,7 @@
 using Ryujinx.Core.OsHle.Ipc;
+using System;
 using System.Collections.Generic;
+using System.IO;
 
 namespace Ryujinx.Core.OsHle.Services.Set
 {
@@ -34,7 +36,7 @@ namespace Ryujinx.Core.OsHle.Services.Set
         {
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
-                { 0, GetLanguageCode },
+                { 0, GetLanguageCode               },
                 { 1, GetAvailableLanguageCodes     },
                 { 3, GetAvailableLanguageCodeCount }
             };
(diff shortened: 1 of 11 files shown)


Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
```diff
@@ -34,11 +34,19 @@ public ISettingsServer()
         {
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
+                { 0, GetLanguageCode },
```
gdkchan: nit: Alignement of the closing `}`.

Ryujinx.Core/OsHle/Services/Set/ISettingsServer.cs
```diff
@@ -34,11 +34,19 @@ public ISettingsServer()
...
                 { 1, GetAvailableLanguageCodes     },
                 { 3, GetAvailableLanguageCodeCount }
             };
         }
 
+        public static long GetLanguageCode(ServiceCtx Context)
+        {
+            Context.ResponseData.Write(1L); //en-US
```
gdkchan: This is a string encoded on a u64, so its "en-US" and not "1". See for example `GetAvailableLanguageCodes`.
I think that:
```
string LanguageCode = LanguageCodes[Index];

foreach (char Chr in LanguageCode)
{
	Context.Memory.WriteByte(Position++, (byte)Chr);
}

for (int Offs = 0; Offs < (8 - LanguageCode.Length); Offs++)
{
	Context.Memory.WriteByte(Position++, 0);
}
```
Should be extracted on its own method, and used by both functions.

Ryujinx.Core/OsHle/Services/Set/ISystemSettingsServer.cs
```diff
@@ -14,11 +17,46 @@ public ISystemSettingsServer()
...
+        {
+            long ReplyPos  = Context.Request.RecvListBuff[0].Position;
+            long ReplySize = Context.Request.RecvListBuff[0].Size;
+
+            //http://switchbrew.org/index.php?title=System_Version_Title
+            using (MemoryStream MS = new MemoryStream(0x100))
+            {
+                BinaryWriter Writer = new BinaryWriter(MS);
+
+                Writer.Write((byte)0x3); //Major FW Version
```
gdkchan: IMO, instead of adding those comments at the right, it's better to add constants on the top of the class with each value and descriptive names, and then just write the constant here. For example `private byte FwVersionMajor = 3` and then `Writer.Write(FwVersionMajor);`

Ryujinx.Core/OsHle/Services/Set/ISystemSettingsServer.cs
```diff
@@ -14,11 +17,46 @@ public ISystemSettingsServer()
...
+                Writer.Write((byte)0x0); //Minor FW Version
+                Writer.Write((byte)0x0); //Micro FW Version
+                Writer.Write((byte)0x0); //Unknown/Build?
+
+                Writer.Write(0x0A); //Revision Number
+
+                Writer.Write(Encoding.ASCII.GetBytes("NX"), 0, 0x02); //Platform String "NX"
```
gdkchan: Specifying the size of the string and the initial position is not necessary if you're going to write the entire string. `Writer.Write(Encoding.ASCII.GetBytes("NX"));` should be enough. Note that with the comment above in consideration, it should be something like `Writer.Write(Encoding.ASCII.GetBytes(PlatformString));`.

gdkchan:
lgtm, thanks!

PR #107, opened 2018-04-25
 Update ISystemSettingsServer.cs

AcK77:
Implement `GetSettingsItemValue`.
Added `NxSettings.cs` who is generated automatically from a Switch 3.0 config file (Tid: `0100000000000818`).


Ryujinx.Core/OsHle/Services/Set/ISystemSettingsServer.cs
```diff
@@ -83,5 +85,55 @@ public static long SetColorSetId(ServiceCtx Context)
...
+                    SettingBuffer[0] = BoolValue ? (byte)1 : (byte)0;
+                }
+                else
+                {
+                    throw new NotImplementedException(NxSetting.GetType().Name);
+                }
+
+                AMemoryHelper.WriteBytes(Context.Memory, ReplyPos, SettingBuffer);
+
+                Context.Ns.Log.PrintInfo(Logging.LogClass.ServiceSet, $"{AskedSetting} set value: {NxSetting} as {NxSetting.GetType()}");
```
gdkchan: IMO this should be PrintDebug, and not PrintInfo.

Ryujinx.Core/OsHle/Services/Set/ISystemSettingsServer.cs
```diff
@@ -83,5 +85,55 @@ public static long SetColorSetId(ServiceCtx Context)
...
+
+            NxSettings.Settings.TryGetValue(AskedSetting, out object NxSetting);
+
+            if (NxSetting != null)
+            {
+                byte[] SettingBuffer = new byte[ReplySize];
+
+                if (NxSetting is string StringValue)
+                {
+                    //Ain't be able to test, so we need to know what the size of ReplySize is...
```
gdkchan: `Ain't be able to test` doesn't sound correct.

Ryujinx.Core/OsHle/Services/Set/ISystemSettingsServer.cs
```diff
@@ -83,5 +85,55 @@ public static long SetColorSetId(ServiceCtx Context)
...
+                    //Array.Resize(ref SettingBuffer, StringValue.Length);
+                    SettingBuffer = Encoding.ASCII.GetBytes(StringValue);
```
gdkchan: imo just check `if (StringBuffer.Length > ReplySize)` and print a error in this case (and maybe don't write the value since it would overwrite memory).

gdkchan:
lgtm, thanks!

PR #110, opened 2018-04-29
Add Sqxtn_S, Sqxtn_V, Uqxtn_S, Uqxtn_V instructions and Tests (6).

LDj3SNuD:
Update Tester as well.
Nits.

### DIFF
--- a/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
+++ b/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
@@ -205,6 +205,79 @@ namespace ChocolArm64.Instruction
             }
         }
 
+        private static void EmitQxtn(AILEmitterCtx Context, bool Signed, bool Scalar)
+        {
+            AOpCodeSimd Op = (AOpCodeSimd)Context.CurrOp;
+
+            int Elems = (!Scalar ? 8 >> Op.Size : 1);
+            int ESize = 8 << Op.Size;
+
+            int TMaxValue = (Signed ? (1 << (ESize - 1)) - 1 : (int)((1L << ESize) - 1L));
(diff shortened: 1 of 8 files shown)


gdkchan:
Thanks for implementing this! The implementation seems correct, the only problem is that it seems to use EmitLdint/Stint to load/store temp values. However the Ldint/Stint methods are used to read/write to the general purpose registers, so this instruction is modifying the values of the GPRs (X0/X1/X2).

You can use Ldinttmp/Stinttmp to load/store to a temp variable, however the emitter only exposes one temp. I'd recommend writing it in a way that only uses 1 or no temps.

Also, more descriptive names to the labels (like LblEnd or LblExit at the end of the function) facilitates the understanding.

LDj3SNuD:
Updated.
(Ldinttmp/Stinttmp do not exist, it would be useful to be able to use them.)

gdkchan:
Sorry, the correct (re: Ldinttmp/Stinttmp) is EmitLdtmp/EmitSttmp.

ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
```diff
@@ -205,6 +205,79 @@ private static void EmitHighNarrow(AILEmitterCtx Context, Action Emit, bool Roun
...
+                Context.Emit(OpCodes.Pop);
+
+                Context.EmitLdc_I4(TMinValue);
+                SetQCFlag();
+
+                Context.MarkLabel(LblGeEnd);
+
+                if (Scalar)
+                {
+	                EmitVectorZeroLower(Context, Op.Rd);
```
gdkchan: nit: Seems to have one extra identation level.

LDj3SNuD:
ReUpdated & ReTested.

gdkchan:
lgtm, thanks!

PR #115, opened 2018-04-30
A fix for some games.

Cyuubi:
Apparently the audio rendering buffer size has something to do with games booting.


gdkchan:
From where the value 0x800 came from? The current value is already a hack, we should calculate it based on the struct that the game sends. I don't feel theres much value in just bumping the value.

Cyuubi:
I sorta just made it up, some games boot further with 0x800.

emmauss:
try not to make hacks. find out why it works that way.

gdkchan:
Closing in favor of #116 .

PR #116, opened 2018-05-01
A REAL fix for some games.

AcK77:
Apparently the audio rendering buffer size has something to do with games booting so we reversed it with help of @gdkchan.


gdkchan:
lgtm, thanks! We can improve it more when we figure out what each of those values are and give better names to the variables.

PR #118, opened 2018-05-04
Implement ISystemDisplayService::GetDisplayMode & Add IManagerDisplayService::SetLayerVisibility stub

Thog:
Also add some missing PrintStub calls

### DIFF
--- a/Ryujinx.Core/OsHle/Services/Vi/ISystemDisplayService.cs
+++ b/Ryujinx.Core/OsHle/Services/Vi/ISystemDisplayService.cs
@@ -1,5 +1,6 @@
 using Ryujinx.Core.OsHle.Ipc;
 using System.Collections.Generic;
+using Ryujinx.Core.Logging;
 
 namespace Ryujinx.Core.OsHle.Services.Vi
 {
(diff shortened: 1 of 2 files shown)


Ryujinx.Core/OsHle/Services/Vi/ISystemDisplayService.cs
```diff
@@ -14,17 +15,29 @@ public ISystemDisplayService()
...
         {
+            Context.Ns.Log.PrintStub(LogClass.ServiceVi, "Stubbed.");
+            return 0;
+        }
+
+        public static long GetDisplayMode(ServiceCtx Context)
+        {
+            Context.ResponseData.Write(1280);
+            Context.ResponseData.Write(720);
+            Context.ResponseData.Write(16.0 / 9.0);
```
gdkchan: Note: This should write a double, are you sure it is correct? afaik GetDisplayMode is supposed to write 16 bytes of data. Also it's, imo, better to use constant for the width/height of the screen, the aspect ratio can be calculated as Width / Height.

Ryujinx.Core/OsHle/Services/Vi/ISystemDisplayService.cs
```diff
@@ -1,5 +1,6 @@
 using Ryujinx.Core.OsHle.Ipc;
 using System.Collections.Generic;
+using Ryujinx.Core.Logging;
```
gdkchan: nit: Usings are sorted in alphabetical order, so `using Ryujinx.Core.Logging;` should go before `using Ryujinx.Core.OsHle.Ipc;`.

gdkchan:
lgtm, thanks!

PR #119, opened 2018-05-05
Add a C++ demangler

Thog:
This is a simple C++ demangler (only supporting name demangling) that will
probably be enough for any stacktrace cases.

### DIFF
--- a/Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
+++ b/Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
@@ -1,13 +1,16 @@
 using System;
 
-namespace Ryujinx.Core.OsHle.Diagnostics {
-    static class Demangler {
+namespace Ryujinx.Core.OsHle.Diagnostics
+{
+    static class Demangle
+    {
         public static string ReadName(string mangled)
         {
             string result = null;
             string charCountTemp = null;
             int charCount = 0;
-            foreach(var chr in mangled)
+
+            foreach (var chr in mangled)
             {
                 if (charCount == 0)
                 {
--- a/Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
+++ b/Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
@@ -9,27 +9,27 @@ namespace Ryujinx.Core.OsHle.Diagnostics
     public static class Demangle
     {
         /*
-          <builtin-type> ::= v	# void
-		 ::= w	# wchar_t
-		 ::= b	# bool
-		 ::= c	# char
-		 ::= a	# signed char
-		 ::= h	# unsigned char
-		 ::= s	# short
-		 ::= t	# unsigned short
-		 ::= i	# int
-		 ::= j	# unsigned int
-		 ::= l	# long
-		 ::= m	# unsigned long
-		 ::= x	# long long, __int64
-		 ::= y	# unsigned long long, __int64
-		 ::= n	# __int128
-		 ::= o	# unsigned __int128
-		 ::= f	# float
-		 ::= d	# double
-		 ::= e	# long double, __float80
-		 ::= g	# __float128
-		 ::= z	# ellipsis
+          <builtin-type> ::= v    # void
+         ::= w    # wchar_t
+         ::= b    # bool
+         ::= c    # char
+         ::= a    # signed char
+         ::= h    # unsigned char
+         ::= s    # short
+         ::= t    # unsigned short
+         ::= i    # int
+         ::= j    # unsigned int
+         ::= l    # long
+         ::= m    # unsigned long
+         ::= x    # long long, __int64
+         ::= y    # unsigned long long, __int64
+         ::= n    # __int128
+         ::= o    # unsigned __int128
+         ::= f    # float
+         ::= d    # double
+         ::= e    # long double, __float80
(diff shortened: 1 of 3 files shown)


gdkchan:
Thanks for working on this. Can you move the Demangle method to another class, since it doesn't really belong to process? A static class called `Demangler` should be fine.

Thog:
Should be good now ^^

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+
+namespace Ryujinx.Core.OsHle.Diagnostics {
```
gdkchan: Please place `{` on a new line to be consistent with the rest of the project.

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,57 @@
...
+    static class Demangler {
+        public static string ReadName(string mangled)
```
gdkchan: imo `Demangle` would be a better name as it is not reading the name but demangling it.

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
Thog: oops sorry... bad habit ^^'

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            string result = null;
+            string charCountTemp = null;
+            int charCount = 0;
+            foreach(var chr in mangled)
```
gdkchan: nit: `foreach(` -> `foreach (`.

gdkchan:
lgtm, thanks!

Does this omit any information that would be otherwise available on the mangled name? Like the arguments of the function.

Thog:
@gdkchan Yeah this code can only parse simple [nested name](http://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.nested-name) (it doesn't support templates and ignores CV-qualifiers).

Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
```diff
@@ -0,0 +1,60 @@
+using System;
+
+namespace Ryujinx.Core.OsHle.Diagnostics
+{
+    static class Demangle
+    {
+        public static string ReadName(string mangled)
```
emmauss: nit: follow the naming scheme of the project, all identifiers are in PascalCase, not camelCase. change below too

Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
```diff
@@ -0,0 +1,383 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace Ryujinx.Core.OsHle.Diagnostics
+{
+    public static class Demangle
```
gdkchan: This shouldn't be a public class since we don't need to access the demangler from outside of this project, so you can just remove the `public` keyword. Also, imo, `Demangler` is better than `Demangle`.

Ryujinx.Core/OsHle/Diagnostics/Demangle.cs
```diff
@@ -0,0 +1,383 @@
...
+                    {
+                        pos  += tempPos;
+                        res = res + "::" + type[type.Count - 1];
+                    }
+                }
+            }
+            return res;
+        }
+
+        public static List<string> ReadName(string mangled, List<string> compressionData, out int pos)
```
gdkchan: I believe this should be private as its not used outside of this class.

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,383 @@
...
+            { "c", "char" },
+            { "a", "signed char" },
+            { "h", "unsigned char" },
+            { "s", "short" },
+            { "t", "unsigned short" },
+            { "i", "int" },
+            { "j", "unsigned int" },
+            { "l", "long" },
+            { "m", "unsigned long" },
+            { "x", "long long, __int64" },
```
gdkchan: The comma here doesn't make sense, the name needs to be either `long long` or `__int64`.

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,383 @@
...
+            { "y", "unsigned long long, __int64" },
```
gdkchan: Same comment as above.

Ryujinx.Core/OsHle/Diagnostics/Demangler.cs
```diff
@@ -0,0 +1,383 @@
...
+            { "n", "__int128" },
+            { "o", "unsigned __int128" },
+            { "f", "float" },
+            { "d", "double" },
+            { "e", "long double, __float80" },
```
gdkchan: Same comment as above.

gdkchan:
lgtm, thanks.

PR #120, opened 2018-05-05
NvServices refactoring

gdkchan:
The main change is that the ioctls were moved from the main nvservices class and were split into their respective classes. Also separated per folder, and added structs for the arguments passed/returned from the ioctls (should be better than reading/writing it directly).


gdkchan:
This should be good to merge now...

AcK77:
Amazing!!

gdkchan:
Quite a bunch of ioctls are still stubbed, and event/syncpt stuff is incomplete aswell, but this can be improved later since I don't want to invest too much time into nvservices for now.

PR #121, opened 2018-05-10
Add intrinsics support

gdkchan:
Now that .net core 2.1 is pretty much out we can use the new intrinsics feature and take advantage of the SSE/AVX instructions directly. Only a few instructions are currently implemented using those intrinsic methods. More to come in the future.


NekokoTime:
Update AInstEmitSimdArithmetic.cs #123

PR #123, opened 2018-05-10
Update AInstEmitSimdArithmetic.cs

NekokoTime:
class Program
    {
        unsafe static void Main(string[] args)
        {
            Vector128<short> test = Sse2.SetVector128(0, 0, 0, 0, 0, 0, 0, -1);

            Console.WriteLine(ExtractShortLow(test));
            Console.WriteLine(ExtractShortLow2(test));
        }

        private static int ExtractShortLow(Vector128<short> Vector)
        {
            return Sse2.Extract(Vector, 0);
        }

        private static int ExtractShortLow2(Vector128<short> Vector)
        {
            short tmp = Sse2.Extract(Vector, 0);
            return tmp;
        }
    }
}


gdkchan:
This is just undoing my fixes and pasting my bug repro code on a place that doesn't make any sense. Whats even the point of this?

PR #132, opened 2018-06-04
Texture/Vertex/Index data cache

gdkchan:
The goal here is avoiding unnecessary texture uploads by detecting writes to the texture data region. This is done by using the windows GetWriteWatch API function. On other OSes it still sends the texture for every frame as I don't have to test it on other OSes and still need to investigate similar APIs for linux and mac.

I also made an attempt to implement it by calculating a hash over the entire texture data and comparing with the texture at the cache. This approach proved to be even slower because hashing the entire texture takes more time than just reading + sending it to the gpu.

With the current approach there are small perf gains on some games and no perf regressions. The same approach can be used with vertex/index data aswell.


gdkchan:
Guess it should be good enough. It will cache texture, vertex and index data. Every time it needs to send any of this type of data to the GPU, it will check if the data is available on some cache, and if it is still valid. If it isn't, then it will upload and again, and the cache will be updated. If it is, then it just binds the data that is already on the GPU (saving us from reading memory + uploading it to the GPU).

Additionaly, to avoid running out of memory, old data is automatically cleaned. Currently this is hardcoded with a 5 minutes limit. If the data hasn't been used in the last 5 minutes, then it will be deleted over time (older data is deleted first). This is triggered only when new data is sent.

There is also a range map for the pages that are accessed through the GPU Virtual Memory Manager. All address ranges that are accessed and uploaded to the GPU are stored on a per-page basis, since the GetWriteWatch API can only keep track of entire pages that have been modified, we still need to manage the parts of the page that are modified ourselves. This also performs automatically cleanup of old pages for the same reasons, however it works a bit differently. The limit is 1 minute instead of 5, and it also has an upper limit for the maximum number of ranges that can be kept on the list. And instead of deleting stuff over time, it deletes all at once. This was implemented using a Dictionary and a LinkedList to provide both fast access times for any address (key), and the LinkedList is used to allow getting the older pages with fast, constant access time instead of the linear access time we would otherwise have when iterating over the Dictionary or a List, at the expense of slightly increased memory usage.

PR #134, opened 2018-06-06
Faster soft implementation of smulh and umulh

riperiperi:
Was doing a quick profile and noticed this took up quite a bit of time. BigInteger is general purpose for variable size and does a lot of heap allocation, so just using fixed integer math is much faster here.

When multiplying 100000 sets of integers from a prebuilt array, completes about 14x faster on average than using BigInteger, and gets the same result. Also passed `Ryujinx.Tests.Cpu.CpuTestMul`.

### DIFF
--- a/ChocolArm64/Instruction/ASoftFallback.cs
+++ b/ChocolArm64/Instruction/ASoftFallback.cs
@@ -153,12 +153,30 @@ namespace ChocolArm64.Instruction
 
         public static long SMulHi128(long LHS, long RHS)
         {
-            return (long)(BigInteger.Multiply(LHS, RHS) >> 64);
+            bool LSign = (LHS < 0);
+            bool RSign = (RHS < 0);
+
+            long Result = (long)UMulHi128((ulong)(LSign ? -LHS : LHS), (ulong)(RSign ? -RHS : RHS));
+
+            if (LSign != RSign)
+                return (-Result) - 1; //for negative results, hi 64-bits start at 0xFFF... and count back
+            return Result;
         }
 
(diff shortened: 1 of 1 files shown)


ChocolArm64/Instruction/ASoftFallback.cs
```diff
@@ -153,12 +153,30 @@ private static ulong ReverseBytes(ulong Value, RevSize Size)
 
         public static long SMulHi128(long LHS, long RHS)
         {
-            return (long)(BigInteger.Multiply(LHS, RHS) >> 64);
+            bool LSign = (LHS < 0);
+            bool RSign = (RHS < 0);
+
+            long Result = (long)UMulHi128((ulong)(LSign ? -LHS : LHS), (ulong)(RSign ? -RHS : RHS));
```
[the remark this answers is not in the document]
riperiperi: That's a deliberate attempt at reusing the sign for the abs and the negative check later.

ChocolArm64/Instruction/ASoftFallback.cs
```diff
@@ -153,12 +153,30 @@ private static ulong ReverseBytes(ulong Value, RevSize Size)
...
+
+            if (LSign != RSign && LHS != 0 && RHS != 0)
+                return (-Result) - 1; //for negative results, hi 64-bits start at 0xFFF... and count back
```
MerryMage: return ~Result;

riperiperi: Addressed. Thanks.

gdkchan:
lgtm, thanks!

PR #135, opened 2018-06-07
Add UncheckedReadBytes in AMemory, for MemoryHelper when memory checks disabled

riperiperi:
Before this change, services reading bytes from memory would always run the memory checks for each byte being copied, which would get rather extreme for large reads. This was particularly bad when reading GPU constant buffers from memory, but might easily crop up in other HLE functions. 

This change implements a method to quickly copy out bytes with Marshal.Copy() if memory checking is disabled. Right now this is triggered by the same config setting as enabling/disabling memory checks for the JIT compiler. 

With this, the smulh change and the texture cache merged, Odyssey's menus now run at about 10 fps guest/2fps host, making it a lot quicker to get to the menu and test things there.

### DIFF
--- a/ChocolArm64/Memory/AMemory.cs
+++ b/ChocolArm64/Memory/AMemory.cs
@@ -301,6 +301,13 @@ namespace ChocolArm64.Memory
             return *((ulong*)(RamPtr + (uint)Position));
         }
 
+        public byte[] ReadByteArrayUnchecked(long Position, long Length)
+        {
+            byte[] Result = new byte[Length];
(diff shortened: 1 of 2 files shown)


gdkchan:
I don't think that doing an unchecked read is necessary on this case. The checked function can be optimized to do only 1 check per page. This means that be can copy chuncks of 4kb (page size) doing just a single check, and still have the benefits of memory checking (detecting invalid read/writes and catching errors early). It would be better if you implement something like that IMO.

riperiperi:
Never even crossed my mind. I’ll have a look into it today.

PR #136, opened 2018-06-07
Add ranged memory reads and checks.

riperiperi:
See #135. Greatly speeds up ReadBytes in MemoryHelper while still doing memory checks.

Memory permissions and access is checked one page at a time. The function checks if it's mapped and the permissions at the same time, and returns the first invalid address it finds (at the base of the page table entry). If you want to check without permission just pass 0 as the permission flag.

### DIFF
--- a/ChocolArm64/Memory/AMemory.cs
+++ b/ChocolArm64/Memory/AMemory.cs
@@ -301,6 +301,15 @@ namespace ChocolArm64.Memory
             return *((ulong*)(RamPtr + (uint)Position));
         }
 
+        public byte[] ReadBytes(long Position, long Size)
+        {
+            EnsureRangeIsValid(Position, Size, AMemoryPerm.Read);
+
+            byte[] Result = new byte[Size];
+            Marshal.Copy((IntPtr)(RamPtr + (uint)Position), Result, 0, (int)Size);
+            return Result;
+        }
+
         public Vector128<float> ReadVector8Unchecked(long Position)
         {
             if (Sse2.IsSupported)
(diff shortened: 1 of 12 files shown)


gdkchan:
Thanks for working on this. With this change, looks like all that the ReadBytes method does now is calling ReadByteArray on AMemory, so the method doesn't seems to be really necessary. I would suggest either:
- Removing AMemoryHelper.ReadBytes and replacing all its uses directly with ReadByteArray.
- Moving all the code from ReadByteArray to AMemoryHelper.ReadBytes and removing ReadByteArray (this one is probably easier, althrough you would also need to make the address/perm checking methods public). The `EnsureRangeIsValid` and `IsValidRange` methods doesn't seems to be necessary. You can just call `EnsureAccessIsValid` for each page. For example:
```
long EndPos = Position + Size;

while (Position < EndPos)
{
    EnsureAccessIsValid(Position, AMemoryPerm.Read);

    Position += AMemoryMgr.PageSize;
}
```
And then doing the copy after that (if any of the pages is not mapped or you don't have the permissions to read it, then it will throw).

riperiperi:
Makes sense. The reason things are as they are now is because I wanted to mimic the existing structure for memory checks. I'll go with your first approach, since it seems rather messy to be doing the pointer resolution and copy in AMemoryHelper when all other methods do that in AMemory.

riperiperi:
Fixed and rebased. Renamed the function in AMemory to ReadBytes and replaced the AMemoryHelpers calls, and did the simpler validity check.

gdkchan:
lgtm, thanks! It would be nice if WriteBytes received the same treatment but this can be saved for later.

PR #141, opened 2018-06-10
[WIP] Experimental basic XInput support.

Cyuubi:
Title says it all, this is experimental and probably doesn't work on Linux. Both left and right joysticks aren't implemented (though they will be soon) and eventually I'll add support for changing the keybinds in config.


zoltx23:
I will see if I can test the inputs in my Xbox one controller tomorrow. Maybe also experiment with the Joystick and Trigger inputs as well.

Cyuubi:
Joystick and triggers aren't added.

zoltx23:
Well, I knew that much. I meant more so try and implement them if I can. Since it seems you are using another library for Xinput.

gdkchan:
The way you handle button presses looks wrong. Based on the name I'm pretty sure that the button values are suppsoed to be used as flags, but instead you use it with a switch. The end result is that if you press multiple buttons at once on the controller, only one will be detected.

Cyuubi:
Added more buttons and fixed multiple inputs at the same time.

Cyuubi:
Added XInput config

Luffykun007:
@Cyuubi Having this error now, soon after the emulator close.

capturar

Cyuubi:
Fixed, my bad.

gdkchan:
Was this tested on linux? I don't think you can use XInput there as DirectX is windows only.

Cyuubi:
Yeah, if someone wants to test this on Linux (to see if it crashes), please do!

DrHacknik:
I am assuming Ryujinx works in Solus. So I'll try~

ReinUsesLisp:
On Arch Linux (no extra packages were installed):
```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'SharpDX.XInput.Controller' threw an exception. ---> System.DllNotFoundException: Unable to load shared library 'kernel32' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libkernel32: cannot open shared object file: No such file or directory
   at SharpDX.XInput.Controller.LoadLibrary(String lpFileName)
   at SharpDX.XInput.Controller..cctor()
   at SharpDX.XInput.Controller..ctor(UserIndex userIndex)
   at Ryujinx.GLScreen..ctor(Switch Ns, IGalRenderer Renderer) in Ryujinx-Cyuubi/Ryujinx/Ui/GLScreen.cs:line 25
   at Ryujinx.Program.Main(String[] args) in Ryujinx-Cyuubi/Ryujinx/Ui/Program.cs:line 63
```
Tried installing wine (in case it magically brought the dependency) without luck.

I think that conditionally initializing XInput on Windows might be a solution.

Cyuubi:
Closing because #196 is merged and better than this.

PR #142, opened 2018-06-10
Implement ListAudioOutsAuto & OpenAudioOutAuto

greggameplayer:
This function is used in some games like 3D Minigolf.


Ryujinx.Core/OsHle/Services/Aud/IAudioOutManager.cs
```diff
@@ -111,5 +112,31 @@ public long OpenAudioOut(ServiceCtx Context)
...
+            }
+            else
+            {
+                Context.Ns.Log.PrintError(LogClass.ServiceAudio, $"Output buffer size {Size} too small!");
+            }
+
+            Context.ResponseData.Write(NameCount);
+
+            return 0;      
+
```
gdkchan: Whats up with all those unneeded empty lines?

greggameplayer:
@gdkchan I've corrected this mistake

AcK77:
What's differencies between ListAudioOuts and ListAudioOutsAuto ?
Because if there is 2 different functions, I don't think do the same behavior is great...

greggameplayer:
@AcK77 same behavior just the buffer size is different which is about 0x22 instead of 6

greggameplayer:
like OpenAudioOutAuto which takes buffer size 0x21 instead of 5

AcK77:
```2 - ListAudioOuts_ex() -> (u32, buffer<unknown, 0x22, 0>)
verified to be the same as ListAudioOuts, just with different buffer types```

If we trust SwIPC and Switchbrew it's apparenlty the only one difference... So It's seems good to me.

AcK77:
`[1] OpenAudioOut(u64, u64, pid, KObject process, buffer<unknown, 5, 0> name_in) -> (u32 sample_rate, u32 channel_count, u32 pcm_format, u32, object<nn::audio::detail::IAudioOut>, buffer<unknown, 6, 0> name_out)`

`[3] OpenAudioOut_ex(u64, u64, pid, KObject, buffer<unknown, 0x21, 0>) -> (u128, IUnknown, buffer<unknown, 0x22, 0>)`

You haven't change the returned buffer type... Then there is too much Unknown in OpenAudioOut_ex (OpenAudioOutAuto).

AcK77:
And at least... Have you tried to build your code before PR it?
Because : 
`OsHle\Services\Aud\IAudioOutManager.cs(147,17): error CS1503: Argument 2: cannot convert from '(long Position, long Size)' to 'long'`

greggameplayer:
@AcK77 how can I solve that ?

greggameplayer:
@AcK77 mistake corrected

AcK77:
The build problem is fixed but : 
`[3] OpenAudioOut_ex(u64, u64, pid, KObject, buffer<unknown, 0x21, 0>) -> (u128, IUnknown, buffer<unknown, 0x22, 0>)`

The output buffer is wrong... And we are not sure about the function too... At least, You can do 2 generics methods, and use them with the correct buffers as args.

greggameplayer:
@AcK77 what's the method to read u128 ?

greggameplayer:
I will recreate a PR for that

PR #144, opened 2018-06-11
Implement GetImageBaseSize, LoadImage on IProfile

Lordmau5:


Ryujinx.HLE/OsHle/Services/Acc/IProfile.cs
```diff
@@ -10,11 +10,15 @@ class IProfile : IpcService
 
         public override IReadOnlyDictionary<int, ServiceProcessRequest> Commands => m_Commands;
 
+        private byte[] profile_image = { };
```
AcK77: I don't think it's necessary to store the image in a bytes array... At least, use a PascalCase.

Ryujinx.HLE/OsHle/Services/Acc/IProfile.cs
```diff
@@ -32,5 +36,26 @@ public long GetBase(ServiceCtx Context)
 
             return 0;
         }
+
+        public long GetImageSize(ServiceCtx Context)
```
AcK77: I think it's better here to determine the real size of the IProfile image, iirc it's 128x128x3bytes... Or just load a real image before and then use the size of it.

Ryujinx.HLE/OsHle/Services/Acc/IProfile.cs
```diff
@@ -32,5 +36,26 @@ public long GetBase(ServiceCtx Context)
...
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAcc, "Stubbed.");
+
+            Context.ResponseData.Write(profile_image.Length);
+
+            return 0;
+        }
+
+        public long LoadImage(ServiceCtx Context)
```
AcK77: You have to use a real JPG file here and write the bytes buffer of it in the type-0x22 buffer.

AcK77:
One month without any update. We can close this PR. If someone want to update it. You feel free to...

PR #145, opened 2018-06-11
Implement ConstantColor for 0xC001 (G80)

Lordmau5:


gdkchan:
It would be nice if you added all the values on the list I sent you, adding a "Legacy" or "Old" suffix on the name (as those values seems to be used on older nvidia gpus.

Also, instead of adding two returns:
```
case GalBlendFactor.ConstantColor: return BlendingFactorDest.ConstantColor;
case GalBlendFactor.ConstantColorG80: return BlendingFactorDest.ConstantColor;
```
it's, IMO, better to use just one like this:
```
case GalBlendFactor.ConstantColor:
case GalBlendFactor.ConstantColorG80:
    return BlendingFactorDest.ConstantColor;
```

greggameplayer:
@Lordmau5 please close this in favor of #207

PR #150, opened 2018-06-11
Implement CreateManagedDisplayLayer on ISelfController

Lordmau5:


Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -105,6 +106,15 @@ public long SetOutOfFocusSuspendingEnabled(ServiceCtx Context)
             return 0;
         }
 
+        public long CreateManagedDisplayLayer(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
+
+            Context.ResponseData.Write(1L);
```
AcK77: `CreateManagedDisplayLayer - Returns an output u64 LayerId which is then used by the user-process with Display_services#OpenLayer.`

So if I understand well, you have to return the same LayerId who is set in `IApplicationDisplayService.cs#L97`. You probably need to "manage" the DisplayLayer... But we don't really handle all those DisplayLayer so it's good for now...

AcK77:
One month without any update. We can close this PR. If someone want to update it. You feel free to...

PR #138, opened 2018-06-08
Implement IApplicationCreator & IApplicationAccessor & IAppletAccessor

greggameplayer:

PR #156, opened 2018-06-12
Implement IApplicationCreator & IApplicationAccessor & IAppletAccessor

greggameplayer:
Rebase of #138


AcK77:
Sorry but stubbed methods are not allowed anymore since you don't have a good reasons (Like stub for JoyCon Vibration who can't be emulate with the keyboard)... So do you plan to implement most of those methods ? If not... You can close the PR. Thanks.

PR #157, opened 2018-06-12
[PCTL] Add CreateServiceWithoutInitialize and Initialize

Cyuubi:


Ryujinx.HLE/OsHle/Services/Pctl/IParentalControlService.cs
```diff
@@ -9,12 +10,24 @@ class IParentalControlService : IpcService
 
         public override IReadOnlyDictionary<int, ServiceProcessRequest> Commands => m_Commands;
 
+        public bool Initialized = false;
+
         public IParentalControlService()
```
AcK77: Here it's better to use `IParentalControlService(bool NeedInitialize = true)` and then use this bool later to know if Initialize is needed or not.

Ryujinx.HLE/OsHle/Services/Pctl/IParentalControlServiceFactory.cs
```diff
@@ -13,15 +13,27 @@ public IParentalControlServiceFactory()
...
             MakeObject(Context, new IParentalControlService());
 
             return 0;
         }
+
+        public long CreateServiceWithoutInitialize(ServiceCtx Context)
+        {
+            IParentalControlService Service = new IParentalControlService();
+
+            Service.Initialized = true;
```
AcK77: The method is `CreateServiceWithoutInitialize` so you don't have to set that here!

Ryujinx.HLE/OsHle/Services/Pctl/IParentalControlServiceFactory.cs
```diff
@@ -13,15 +13,27 @@ public IParentalControlServiceFactory()
...
+
+            MakeObject(Context, Service);
```
AcK77: You can use `MakeObject(Context, new IParentalControlService(false));` here now.

Ryujinx.HLE/OsHle/Services/Pctl/IParentalControlService.cs
```diff
@@ -9,12 +10,24 @@ class IParentalControlService : IpcService
...
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
-                //...
+                { 1, Initialize }
             };
         }
+
+        public long Initialize(ServiceCtx Context)
+        {
+            if (!Initialized)
```
AcK77: You can check if the Initialize is needed here now.
Plus, the Initialize method should doing something on the real sysmodule... But we don't wan't to handle the Parental Control in an emulator I think... So stubs the Initialize should be fine.

AcK77:
Thanks!

PR #158, opened 2018-06-13
Partial GPU DMA support

gdkchan:
This is used to copy data on the GPU mapped memory. It's usually used to copy textures, and converting linear to block linear (swizzled) for example. Some games seems to use it for that purpose, one example of such game is Has Been Heroes. Maybe Pokkén Tournament aswell?


gdkchan:
I don't plan to do more work on it for now, its good enough to support the games that are currently requiring it. So I think its good to merge.

PR #166, opened 2018-06-15
Port bcat:u from yuzu

RyuDanuer:
We can port it because it has no conflicts,this will take DOOM further


gdkchan:
I don't see how this is related to the bcat service.

PR #168, opened 2018-06-16
SPIR-V Intermediate Shading Language support

ReinUsesLisp:
SPIR-V Binary format support for Ryujinx.Graphics shaders. There are some changes in Ryujinx.ShaderTools to be able to output GLSL or SPIR-V.

Before explaining technical issues about this PR, I'd like to explain why a SPIR-V decoder is useful.
Since we're decoding from Maxwell's binary shaders, an assembly-like language is easier to translate to than a C-like language like GLSL, mainly because it has branches and conditional branches as part of it's opcodes.
It'd allow us to have a single shader decompiler for Vulkan and Modern OpenGL (4.6).
This commits should not do a performance boost in Ryujinx (since drivers should be good at compiling GLSL).

When ARB_gl_spirv on GNU/Linux or Windows is available it will unconditionally use SPIR-V. OS X is masked. If it's chosen, it will print it on the console (otherwise it says GLSL is used).

It was tested on NVIDIA's driver on Windows and Linux. Any test on other OpenGL 4.6 drivers is appreciated.

https://github.com/ReinUsesLisp/Ryujinx/blob/spirv/Ryujinx.Graphics/Gal/OpenGL/OpenGLRenderer.cs#L37
There I'm queueing an Initialization command. If it's not executed **before** calling shader decompilation, an exception will be thrown since it didn't decide which shading language to use. Maybe that needs some improvements.

https://github.com/ReinUsesLisp/Ryujinx/blob/spirv/Ryujinx.Graphics/Gal/OpenGL/OpenGLExtension.cs
That file exists for the sole reason that glSpecializeShader is not parsed in OpenTK.NETCore

There should not be any graphical changes (if there are, it's a bug or maybe something was fixed).

Please highlight anything that can be improved (specially in C# usage).


emmauss:
does glSpecializeShader exist in opentk main branch?

ReinUsesLisp:
It is. It's already part of OpenTK nuget.

ReinUsesLisp:
SPIR-V shaders crash on some devices: Works on NVIDIA Pascal driver but crashes on AMD GCN 1.0.
Due to this, there's a key added to Ryujinx.conf named "Enable_SPIRV" which tries to use SPIR-V when ARB_gl_spirv is available.

ReinUsesLisp:
Added UBOs for GLSL and SPIR-V.

GLSL searches for free bindings, so it should support any constant buffers index.

SPIR-V has to declare bindings in at compile time (con can't change it with glUniformBlockBinding) and Khronos solution is a joke:
```
24. What about UniformBlockBinding and ShaderStorageBlockBinding?

    RESOLVED. Because it is not possible to obtain the uniform block index
    or storage block index from an unnamed block, the binding value remains
    as specified in the shader by the layout qualifier or decoration.
    If you are feeling lucky you could just guess and remap used index values
    [0, #active blocks), but you won't really know what is going
    where, so just treat it as an immutable binding, similar to the atomic
    counter buffer binding point. Really.
```

For now, SPIR-V is locked to the first 12 cbufs per stage (I haven't found a game that uses high indexes, but it's a bug none the less).

12 is a magic number chosen because of OpenGL [ARB](http://www.opengl.org/registry/specs/ARB/uniform_buffer_object.txt) max buffers per stage.

ReinUsesLisp:
I'm closing this PR for now because it makes OGLShader bigger than it should be, and it's not useful for now.
It might be useful for a theorical Vulkan support, but OpenGL drivers' implementation of GL_ARB_gl_spirv is buggy and unsafe (it doesn't even report errors on some drivers, the program just crashes).
I'm going to keep the branch mainly because I don't want to write a SPIR-V assembler again.

PR #170, opened 2018-06-17
Add  support for Nro Asset Parsing

emmauss:
This adds parsing for the nro asset section and nacp archive. I am pushing this for later work on game list.

### DIFF
--- a/Ryujinx.HLE/Loaders/Archives/ControlArchive.cs
+++ b/Ryujinx.HLE/Loaders/Archives/ControlArchive.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+
+namespace Ryujinx.HLE.Loaders.Archives
+{
+    public class ControlArchive
+    {
+        public LanguageEntry[] LanguageEntries { get; set; }
+        public long            ApplicationTitleID { get; set; }
+        public long            BaseTitleID { get; set; }
+        public long            ProductCode { get; set; }
+        public string          ApplicationVersion { get; set; }
+
+        public ControlArchive(Stream Input)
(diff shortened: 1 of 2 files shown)


emmauss:
I am planning on using this for the gui. Showing icon and title in a game list for a homebrew is mostly cosmetic, and so is true for an actual cart game. It wont even be necessary if a gui does not exist in the first place. It is purely there just to look nice and tell the player what is being selected. That is one of the primary aim of a user interface, to look nice and still do what is is meant to do. Also if only the cart games have icons, and like 20 or so nro's are blank with names, like LovePotion which is just an engine, and not the game itself, it is not going to look nice.

gdkchan:
I don't have anything against showing icons/names for homebrew. My main problem with that is that NRO is supposed to be dynamic link libraries, that's how the format is used by official games. Officially, it also doesn't have any of this (like the ASSET section), and it is not meant to be executed directly like this. So this kind of thing is really only used by homebrew. This will most likely be problematic when we need to support actual NROs (used by the official games) in the future as it will try to parse a bunch of stuff that doesn't exist, and we will need to have two separate paths for this format, one for homebrew and other for official games.

emmauss:
I didn't get it at first. so I should treat it as an extension of nro, that is only for homebrew, and therefore is only relevant for anything that will handle hmbrew, like the launcher.  so a subclass extending it only in the gui project can be used.

gdkchan:
Yes I think this is a reasonable way to handle it. I still think they use NRO for that instead of NSOs that would be more suited for this kind of thing. But anyway, as the asset section is at the "end" of the NRO anyway, it should be easy to parse separately.

emmauss:
okay, I will close this and push it along with the gui.

PR #171, opened 2018-06-17
Add Cmeq_S, Cmge_S, Cmgt_S, Cmhi_S, Cmhs_S, Cmle_S, Cmlt_S (Reg, Zero) & Cmtst_S compare instructions. Add 22 compare tests (Scalar, Vector). Add Eor_V, Not_V tests.

LDj3SNuD:
Update Tester as well.
Nits.


ChocolArm64/Instruction/AInstEmitSimdCmp.cs
```diff
@@ -362,7 +369,47 @@ private static void EmitVectorCmp(AILEmitterCtx Context, OpCode ILOp)
                 Context.MarkLabel(LblEnd);
             }
 
-            if (Op.RegisterSize == ARegisterSize.SIMD64)
+            if ((Op.RegisterSize == ARegisterSize.SIMD64) || Scalar)
```
gdkchan: This will only zero out the upper 64-bits of the destination registers. I believe that scalar values < 64-bits will still contain the old value on the bits that are not set and not cleared.

LDj3SNuD: For all these scalars we have "if size != '11' then ReservedValue();".

gdkchan:
lgtm, thanks!

PR #172, opened 2018-06-18
Fix some thread sync issues

gdkchan:
By doing some stress test I was able to find more (yes, again...) thread synchronization issues. The below issues are fixed:
- CondVarWait mutex unlock + add to wait list was not atomic. This caused races in some games.
- CondVarSignal would sometimes write the updated mutex value to an invalid address.
- The ARM synchronization instructions (like LDAXR/STLXR) were mostly broken and races were possible even when using it correctly.
- Ordering of the arm load atomic instruction were wrong aswell (it tried to acquire the address after the load rather than before).
- TryToRun logic was broken, it would allow a thread to run on 2 different cores.

This PR should fix all the above issues.

Additionaly, during my testing, I also found a bug on libnx mutex, and also PRed a fix for that:
https://github.com/switchbrew/libnx/pull/120

Unfortunately games like Splatoon 2 still deadlocks with this, but now it is at least deterministic (it aways freezes at the same point now). SMO no longer deadlocks for me either, and I can consistently hit the start option. Games that requires SetThreadCoreMask such as VVVVVV also works with this PR. Other games with synchronization related issues seems to also have improved.

I still want to do more investigation through, like the splatoon issue or some other regressions I may have missed. Until then this is WIP. I may also put the SetThreadCoreMask hack back in the worst case to avoid regressions.

### DIFF
--- a/ChocolArm64/AThread.cs
+++ b/ChocolArm64/AThread.cs
@@ -46,7 +46,7 @@ namespace ChocolArm64
             {
                 Translator.ExecuteSubroutine(this, EntryPoint);
 
-                Memory.RemoveMonitor(ThreadId);
+                Memory.RemoveMonitor(ThreadState);
 
                 WorkFinished?.Invoke(this, EventArgs.Empty);
             });
--- a/ChocolArm64/Instruction/AInstEmitMemoryEx.cs
+++ b/ChocolArm64/Instruction/AInstEmitMemoryEx.cs
@@ -48,6 +48,16 @@ namespace ChocolArm64.Instruction
         {
             AOpCodeMemEx Op = (AOpCodeMemEx)Context.CurrOp;
 
+            if (AccType.HasFlag(AccessType.Ordered))
+            {
+                EmitBarrier(Context);
+            }
+
+            if (AccType.HasFlag(AccessType.Exclusive))
+            {
+                EmitMemoryCall(Context, nameof(AMemory.SetExclusive), Op.Rn);
+            }
+
             Context.EmitLdarg(ATranslatedSub.MemoryArgIdx);
             Context.EmitLdint(Op.Rn);
 
@@ -66,16 +76,6 @@ namespace ChocolArm64.Instruction
                 EmitReadZxCall(Context, Op.Size);
 
                 Context.EmitStintzr(Op.Rt2);
-            }
-
-            if (AccType.HasFlag(AccessType.Exclusive))
-            {
-                EmitMemoryCall(Context, nameof(AMemory.SetExclusive), Op.Rn);
-            }
-
-            if (AccType.HasFlag(AccessType.Ordered))
-            {
-                EmitBarrier(Context);
             }
         }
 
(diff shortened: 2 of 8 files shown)


greggameplayer:
@gdkchan for me the deadlock on Super Mario Odyssey is fix too.

greggameplayer:
@gdkchan but it's add a deadlock on Splatoon 2 which now block on this : 
image
with zero fps.

greggameplayer:
@gdkchan but it add also a deadlock on Super Mario Odyssey after hitting the play button

gdkchan:
I'm aware of this, that's why I added "Unfortunately games like Splatoon 2 still deadlocks with this". The only reason splatoon worked before is because SetThreadCoreMask was ignored forcing most (all?) threads to run on the same core, so in this kind of problem is way more unlikely to occur as you don't have two or more threads running at the same time. With SetCoreMask enabled everything was way worse than it is now, this PR is an improvement not a regression.

gdkchan:
I found another possible cause of race conditions... When the mutex address is written by a normal store operation (in this particular case, by HLE code using the WriteInt32 method on the CPU), it's possible that the address was already acquired by a load-acquire + store-release sequence. So the following is possible:

Guest code acquires and reads mutex address
HLE code writes new mutex value
Guest code writes modified value and releases mutex address

The value written by HLE code is trashed.
The exact problem I found during my tests through, was:

1 - Guest code acquires and reads mutex address for CriticalSection.Enter.
2 - Guest code notices that another thread owns the mutex and sets has listeners flag on the mutex value.
3 - Meanwhile another thread calls some svc that causes MutexUnlock to be called.
4 - MutexUnlock notices that no thread is waiting for the mutex (yet) and writes 0 to the mutex address.
5 - Guest code writes modified mutex value with has listeners flag set to mutex address.
^ At this point the 0 mutex value written by MutexUnlock is trashed.
6 - The thread that called MutexUnlock goes to wait.
7 - The thread that called CriticalSection.Enter calls ArbitrateLock and goes to wait aswell, as the new value it just wrote has the has listeners flag set with the old thread handle...
At this point the application is deadlocked.

The fix is introducing a WriteInt32ToSharedAddr method that ensure the exclusive monitor is cleared for HLE code. So the step 5 would fail as it wouldn't have exclusive access to that address anymore, forcing the code to retry and then exit early as it notices that no thread is holding the mutex anymore, thus avoiding the deadlock.

On a real ARM cpu, any store to an address marked as exclusive on the exclusive monitor would "clear" the monitor, meaning that whichever thread that had exclusive access to it will lose that access. So, this is basically what we do here, except that in our case this only works for HLE code... Guest code doing a normal store to an exclusive address will still have the problem mentioned above. However, I'm yet to see game code that does such a thing, so I think this solution should be good enough for now... Implementing this behaviour for every store is most likely too expensive and will hurt perf badly.

After this fix splatoon still deadlocks at the same place. However I'm tired of working on this and overall this PR is an improvement compared to what we had before, so I will merge it and we can address this regression later.

PR #173, opened 2018-06-18
NvGpuEngine3d: Fix uploading vertex arrays without index buffers.

Armada651:
Since this check is preceded by `IndexSize = 1 << IndexSize;` it can't ever be 0. I think a check for IndexCount was intended here.


gdkchan:
Yes that's correct, the `IndexSize` name is not ideal and can lead to that confusion. Althrough iirc the VertexEndPos register should aways be set so in the end the only difference is that it may send more data than necessary.

Thanks for catching and fixing this!

Armada651:
In my tests this resulted in an exception, because it will indeed try to read the index buffer from an invalid address.

PR #174, opened 2018-06-18
Rework smulh to fix an edge case.

riperiperi:
Does the multiplication by casting the inputs to unsigned, then offsetting the result when operands are negative. 

Tested on 10485760 random combinations of numbers, the unit tests and the case that was broken before (long.MinValue * >=0x80000000).


gdkchan:
ptal @LDj3SNuD

LDj3SNuD:
I re-tested Smulh_64bit for further edge cases; (re-tested also Umulh_64bit).
lgtm

PR #175, opened 2018-06-19
Implement GetAvailableLanguageCodeCount2

greggameplayer:
according to switchbrew GetAvailableLanguageCodeCount and GetAvailableLanguageCodeCount2 are the same.

### DIFF
--- a/Ryujinx.HLE/OsHle/Services/Set/ISettingsServer.cs
+++ b/Ryujinx.HLE/OsHle/Services/Set/ISettingsServer.cs
@@ -13,10 +13,11 @@ namespace Ryujinx.HLE.OsHle.Services.Set
         {
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
-                { 0, GetLanguageCode               },
-                { 1, GetAvailableLanguageCodes     },
-                { 3, GetAvailableLanguageCodeCount },
-                { 5, GetAvailableLanguageCodes2    }
+                { 0, GetLanguageCode                },
+                { 1, GetAvailableLanguageCodes      },
+                { 3, GetAvailableLanguageCodeCount  },
+                { 5, GetAvailableLanguageCodes2     },
+                { 6, GetAvailableLanguageCodeCount2 }
             };
         }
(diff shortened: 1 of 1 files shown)


AcK77:
« GetAvailableLanguageCodes2/GetAvailableLanguageCodeCount2 instead of {original commands}.
In official user-processes in the #Language->#LanguageCode conversion function (MakeLanguageCode):
During one-time init, GetAvailableLanguageCodes is used to initialize the LanguageCodes array cache, with max_entries=0xF (buffer size in u64s). [4.0.0+] GetAvailableLanguageCodes2 is now used with max_entries 0x40.
[4.0.0+] When the input #Language is larger than the cached total_entries from the above command output, or #Language is negative, command MakeLanguageCode is used instead of the array. » I don’t thinks it’s really the same method... Some investigation is required!

greggameplayer:
@AcK77 yes but for now we don't know much more on this function. maybe GetAvailableLanguageCodeCount2 is used for GetAvailableLanguageCodes2 and GetAvailableLanguageCodeCount is used for GetAvailableLanguageCodes. I think these two functions just don't return the same value.

AcK77:
As I said, switchbrew docs explained how it works:

```
[4.0.0+] Official user-processes now use GetAvailableLanguageCodes2/GetAvailableLanguageCodeCount2 instead of {original commands}.

In official user-processes in the #Language->#LanguageCode conversion function (MakeLanguageCode):

During one-time init, GetAvailableLanguageCodes is used to initialize the LanguageCodes array cache, with max_entries=0xF (buffer size in u64s). [4.0.0+] GetAvailableLanguageCodes2 is now used with max_entries 0x40.
[4.0.0+] When the input #Language is larger than the cached total_entries from the above command output, or #Language is negative, command MakeLanguageCode is used instead of the array.
```

So it's not the same methods... All you did here is to copy paste some code and not implement the real method. You are free to update it if you can, or close this PR if you can't. Thanks.

greggameplayer:
I close this PR until we find a proper way to implement it

PR #176, opened 2018-06-20
Port from OpenTK.NETCore to OpenTK.NetStandard

ReinUsesLisp:
Thanks to emmauss for the OpenTK fork.

Currently it requires on nuget.org (just like current OpenTK.NETCore dependency) to get it. OpenTK fails to build on Linux (or at least I couldn't build it), but using the NuGet works.

Some changes on Ryujinx had to be done because an up-to-date OpenTK had some changes in enumerations and input API.

About input, using OpenTK.Input.Keyboard.GetState() seems to return an empty state (no keys being pressed). Mouse clicks work, but its relative positions return weird values (moving the window and checking for mouse positions returns negative coordinates). That's why commit gets the states from their events since this seem to give the right values.

Touchscreen input was tested printing the coordinates on a console, not on a real game.

Some visible changes are that mouse no longer jumps when you lose focus on OpenTK's window.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLBlend.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLBlend.cs
@@ -23,8 +23,8 @@ namespace Ryujinx.Graphics.Gal.OpenGL
                 OGLEnumConverter.GetBlendEquation(Equation));
 
             GL.BlendFunc(
-                OGLEnumConverter.GetBlendFactorSrc(FuncSrc),
-                OGLEnumConverter.GetBlendFactorDst(FuncDst));
+                OGLEnumConverter.GetBlendFactor(FuncSrc),
+                OGLEnumConverter.GetBlendFactor(FuncDst));
         }
 
         public void SetSeparate(
@@ -40,10 +40,10 @@ namespace Ryujinx.Graphics.Gal.OpenGL
                 OGLEnumConverter.GetBlendEquation(EquationAlpha));
 
             GL.BlendFuncSeparate(
-                OGLEnumConverter.GetBlendFactorSrc(FuncSrcRgb),
-                OGLEnumConverter.GetBlendFactorDst(FuncDstRgb),
-                OGLEnumConverter.GetBlendFactorSrc(FuncSrcAlpha),
-                OGLEnumConverter.GetBlendFactorDst(FuncDstAlpha));
+                (BlendingFactorSrc)OGLEnumConverter.GetBlendFactor(FuncSrcRgb),
+                (BlendingFactorDest)OGLEnumConverter.GetBlendFactor(FuncDstRgb),
+                (BlendingFactorSrc)OGLEnumConverter.GetBlendFactor(FuncSrcAlpha),
+                (BlendingFactorDest)OGLEnumConverter.GetBlendFactor(FuncDstAlpha));
         }
     }
 }
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
@@ -73,16 +73,16 @@ namespace Ryujinx.Graphics.Gal.OpenGL
             throw new NotImplementedException(Format.ToString());
         }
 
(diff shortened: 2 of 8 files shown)


emmauss:
what about `Keyboard[Key key]` for get input

ReinUsesLisp:
They dropped it

emmauss:
It is the same

ReinUsesLisp:
They dropped it

Ryujinx.Graphics/Gal/OpenGL/OGLBlend.cs
```diff
@@ -21,10 +21,10 @@ public void Disable()
         {
             GL.BlendEquation(
                 OGLEnumConverter.GetBlendEquation(Equation));
-
+
```
gdkchan: Tab on empty line is not necessary.

Ryujinx.Graphics/Gal/OpenGL/OGLBlend.cs
```diff
@@ -21,10 +21,10 @@ public void Disable()
         {
             GL.BlendEquation(
                 OGLEnumConverter.GetBlendEquation(Equation));
-
+    
             GL.BlendFunc(
-                OGLEnumConverter.GetBlendFactorSrc(FuncSrc),
-                OGLEnumConverter.GetBlendFactorDst(FuncDst));
+                (BlendingFactor)OGLEnumConverter.GetBlendFactorSrc(FuncSrc),
```
gdkchan: IMO it would be better to just return `BlendingFactor` directly, so you can remove the `GetBlendFactorDst` method, change `GetBlendFactorSrc` and then change the return type from `BlendingFactorSrc` to `BlendingFactor`. You'll not only remove the need for this cast this way, but also remove one useless method.

Ryujinx.Graphics/Gal/OpenGL/OGLTexture.cs
```diff
@@ -49,7 +49,7 @@ public void Create(long Tag, byte[] Data, GalTexture Texture)
                 GL.CompressedTexImage2D(
                     TextureTarget.Texture2D,
                     Level,
-                    InternalFmt,
+                    (InternalFormat)InternalFmt,
```
gdkchan: IMO its better to change the type at `OGLEnumConverter.GetCompressedTextureFormat` to avoid the need for this cast.

Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
```diff
@@ -162,57 +162,29 @@ public static BlendEquationMode GetBlendEquation(GalBlendEquation BlendEquation)
...
+                case GalBlendFactor.OneMinusDstAlpha:      return BlendingFactor.OneMinusDstAlpha;
+                case GalBlendFactor.ConstantColor:         return BlendingFactor.ConstantColor;
+                case GalBlendFactor.OneMinusConstantColor: return BlendingFactor.OneMinusConstantColor;
+                case GalBlendFactor.ConstantAlpha:         return BlendingFactor.ConstantAlpha;
+                case GalBlendFactor.OneMinusConstantAlpha: return BlendingFactor.OneMinusConstantAlpha;
+                case GalBlendFactor.SrcAlphaSaturate:      return BlendingFactor.SrcAlphaSaturate;
+                case GalBlendFactor.Src1Color:             return BlendingFactor.Src1Color;
+                case GalBlendFactor.OneMinusSrc1Color:     return (BlendingFactor)BlendingFactorSrc.OneMinusSrc1Color;
+                case GalBlendFactor.Src1Alpha:             return BlendingFactor.Src1Alpha;
+                case GalBlendFactor.OneMinusSrc1Alpha:     return (BlendingFactor)BlendingFactorSrc.OneMinusSrc1Alpha;
```
ReinUsesLisp: It's not included for BlendingFactor, luckily it's the same enum value for src and dest

gdkchan: I guess they forgot to add those 2 values on the enum...

ReinUsesLisp:
Yes.

PR #177, opened 2018-06-21
Small OpenGL Renderer refactoring

gdkchan:
Instead of pushing all OpenGL related method calls to a queue inside OpenGL renderer, call them directly. For this we need to ensure that the calls are made from the thread that created the ogl context (in our case, the main thread). So the PFIFO processing thread was removed, and instead a method is exposed that allows the main thread to process the PFIFO command list data. We can remove the PFIFO thread and also a bunch of code at OpenGLRenderer only used to enqueue an method that would be later called through the main thread.

This also seems to fix an issue where some games would be locked to 120 fps instead of 60.

I also added some more checks to avoid unbounded growth of queues (like the fifo queue) and endless issuing of draw commands (that can happen, for example, if macro intepreter enter on a long loop calling VertexEndGl). This happens with a bunch of games and causes the system to freeze. I made it throw an exception which is better than freezing the system, but we should still find the root cause of the issue. Games like Splatoon tries to issue ~96k calls to VertexEndGl and I assume that this is not correct.

FYI @ReinUsesLisp since you're working on graphics stuff aswell.
I'm also interested in suggestion to improve the current approach, since we still need the queue to enqueue actions from the VI service (like when a frame is queue to NVFlinger for rendering) and this can happen from any thread.

While I'm at it I also plan to do more cleanp so this is WIP for now.


ReinUsesLisp:
Now that you are on it, I'd suggest renaming Uniform2F to SetFlip and Uniform1 to EnsureTextureBinding (or something like that) since that's what they are being used for. This allows calls IGalRenderer to be generic instead of OpenGL-specific. (It also helps to implement UBOs.)

About this method, I think it's better for OpenGL. IMO there's a limit in how deep we can emulate Maxwell behaviour from OpenGL's abstraction layers (unless [driver-specific extensions](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_command_list.txt) are used). I know there are logistical limitations, but from Vulkan we could query command buffers from multiple threads.

gdkchan:
Should be ready to merge now.

gdkchan:
Note: The issue with some games running too fast is not fixed. It was before because the FIFO was being processed at a different rate and this caused some games to run at the correct speed (but other slowed down a lot).

PR #179, opened 2018-06-23
Better audren implementation

Cyuubi:
This PR implements better audren functionality, etc.


gdkchan:
The buffer type is still wrong.

PR #181, opened 2018-06-23
Basic Immediate mode GUI 

emmauss:
Another take on a GUI, this time with IMGUI. I have tried to implement most of the basic functions that one would use in a user interface, with the following issues;

-  ~Safely shutting down to UI is not implemented .~
-  Game list do not support xci and nca, only nro(need xci support in emulator first @AcK77 ).
-  There is some delay when starting the emulator and when loading a rom, the screen is not updated during this delay.

Currently, the gui should be okay for loading a rom, configuring controls and pause/resume emulation. I am leaving this here so it can be tested, and those familiar with imgui and gui in general can comment on this.


ChocolArm64/ATranslator.cs
```diff
@@ -21,6 +22,12 @@ public class ATranslator
 
         public bool EnableCpuTrace { get; set; }
 
+        public static long CyclesExecuted { get; set; }
```
gdkchan: This variable seems to be incremented but it is never really used for anything.

ChocolArm64/ATranslator.cs
```diff
@@ -60,6 +67,11 @@ private void ExecuteSubroutineA32(AThreadState State, AMemory Memory)
                 AOpCode OpCode = ADecoder.DecodeOpCode(State, Memory, State.R15);
 
                 OpCode.Interpreter(State, Memory, OpCode);
+
+                CyclesExecuted++;
+
+                while (Pause)
+                    PauseResetEvent.WaitOne(1000);
```
gdkchan: The event is not being used correctly here. It is basically being used as sleep. In this case you could just use `Thread.Sleep` and have the same effect. However I do believe that using the event is a better solution, You can use a ManualResetEvent defaulting to `true` (initialized on the signaled state). Then you can do an wait without timeout here (without the while loop) like: `PauseResetEvent.WaitOne();`. Then, to pause execution you can call `PauseResetEvent.Reset();` and to unpause `PauseResetEvent.Set();`. Due to the fact that the cpu tries to translate entire subroutines at once, theres no guarantees that it will pause immediately.

Ryujinx/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "Ryujinx": {
+      "commandName": "Project",
+      "commandLineArgs": "\"C:\\Users\\Emmanuel Hansen\\AppData\\Roaming\\RyuFs\\sdmc\\switch\\fruity\\fruity.nro\""
```
gdkchan: Hmm I guess that you added this for easier testing? I think this file shouldn't be included on the repository.

emmauss: forgot to delete that

Ryujinx.ImGui/Config.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            Log.SetEnable(LogLevel.Debug,   Convert.ToBoolean(Parser.GetValue("Logging_Enable_Debug")));
+            Log.SetEnable(LogLevel.Stub,    Convert.ToBoolean(Parser.GetValue("Logging_Enable_Stub")));
+            Log.SetEnable(LogLevel.Info,    Convert.ToBoolean(Parser.GetValue("Logging_Enable_Info")));
+            Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.GetValue("Logging_Enable_Warn")));
+            Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.GetValue("Logging_Enable_Error")));
+
+            string[] FilteredLogClasses = Parser.GetValue("Logging_Filtered_Classes").Split(',', StringSplitOptions.RemoveEmptyEntries);
+
+            //When the classes are specified on the list, we only
```
DrHacknik: Mind using `/*` for opening comments and closing them with `*/`? 

It just would look cleaner with that, and when aligned properly. But that's just me.

gdkchan: The entire code base uses the `//` style comments, I don't see any reason to change that now.

DrHacknik: I was unaware that was the scheme used. Sorry~ But that was just my perspective on how I handle Comments. __Seemed more logical in my opinion.__

emmauss:
rebased

emmauss:
shutdown is now fixed

emmauss:
squashed

emmauss:
I guess this is enough to be reviewed and merged. Any devs around?

emmauss:
The file picker now supports drives. I have also renamed the UI folder.

ChocolArm64/ATranslator.cs
```diff
@@ -21,6 +22,8 @@ public class ATranslator
 
         public bool EnableCpuTrace { get; set; }
 
+        public static ManualResetEvent PauseResetEvent = new ManualResetEvent(true);
```
gdkchan: Imo this shouldn't be static, and should be private. It should be possible to pause/unpause individual instances of the translator. Also, Ideally it would expose Pause/Resume methods that would call the Set/Reset methods as needed.

Ryujinx.Graphics/Gal/IGalRenderer.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Ryujinx.Graphics.Gal
 {
-    public interface IGalRenderer
+    public unsafe interface IGalRenderer : IDisposable
```
gdkchan: Why is this unsafe now?

Ryujinx.Graphics/Gal/OpenGL/OGLFrameBuffer.cs
```diff
@@ -23,7 +23,7 @@ public Rect(int X, int Y, int Width, int Height)
             }
         }
 
-        private class FrameBuffer
+        public class FrameBuffer
```
gdkchan: Why does this need to be public?

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -1,21 +1,25 @@
...
-        public byte[] Text { get; private set; }
-        public byte[] RO   { get; private set; }
-        public byte[] Data { get; private set; }
-
-        public int Mod0Offset { get; private set; }
-        public int TextOffset { get; private set; }
-        public int ROOffset   { get; private set; }
-        public int DataOffset { get; private set; }
-        public int BssSize    { get; private set; }
+        public byte[] Text          { get; private set; }
```
gdkchan: Unnecessary spacing.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -1,21 +1,25 @@
...
+        public byte[] RO            { get; private set; }
+        public byte[] Data          { get; private set; }       
 
+        public int Mod0Offset  { get; private set; }
```
gdkchan: Unnecessary spacing.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -39,11 +43,12 @@ public Nro(Stream Input, string Name)
             int DataSize   = Reader.ReadInt32();
             int BssSize    = Reader.ReadInt32();
 
-            this.Mod0Offset = Mod0Offset;
-            this.TextOffset = TextOffset;
-            this.ROOffset   = ROOffset;
-            this.DataOffset = DataOffset;
-            this.BssSize    = BssSize;
+            this.Mod0Offset  = Mod0Offset;
```
gdkchan: Unnecessary spacing. (1 space)

Ryujinx.HLE/OsHle/Horizon.cs
```diff
@@ -161,8 +179,15 @@ internal void ExitProcess(int ProcessId)
...
             if (Processes.TryRemove(ProcessId, out Process))
             {
-                Process.StopAllThreadsAsync();
-                Process.Dispose();
+                try
+                {
+                    Process.StopAllThreadsAsync();
+                    Process.Dispose();
+                }
+                catch (ObjectDisposedException Ex)
```
gdkchan: Ideally this shouldn't be necessary. What is the situation that it tries to dispose a process that is already disposed? I think that the current way to handle it is preventing such cases from happening.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -1,21 +1,25 @@
+using System;
+using System.Drawing;
```
gdkchan: Why is this using necessary? It doesn't seems that you use anything from Drawing here.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -1,21 +1,25 @@
...
 using System.IO;
+using System.Text;
```
gdkchan: Same as above.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -1,21 +1,25 @@
+using System;
```
gdkchan: Probably also not necessary.

Ryujinx.UI/Extensions/Nro.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+    class Nro : HLE.Loaders.Executables.Nro
+    {
+        public  byte[] AssetRomfData { get; set; }
+        public  byte[] IconData      { get; set; }
+        private byte[] NACPData      { get; set; }
+        public  int    AssetOffset   { get; set; }
+
+        public ControlArchive ControlArchive { get; set; }
+
```
gdkchan: IMO the extra new line is not necessary.

Ryujinx.UI/GUI/Widgets/GameList.cs
```diff
@@ -0,0 +1,194 @@
...
+                        }
+
+                        ImGui.EndChildFrame();
+                    }
+                }
+
+                ImGui.EndChildFrame();
+            }
+
+            return new Tuple<bool, string>(false,string.Empty);
```
gdkchan: nit: `(false,string.Empty)` -> `(false, string.Empty)`. Could also use the new named tuples.

Ryujinx.UI/GUI/Widgets/GameList.cs
```diff
@@ -0,0 +1,194 @@
...
+                AppletType = AppletType.Cartridge;
+        }        
+
+        public byte[] GetIconData()
+        {
+            if (IsNro)
+            {
+                return Nro.IconData;
+            }
+            else return null;
```
gdkchan: imo this looks inconsistent. You used curly braces for the if but not for the else.

gdkchan: (There are also other cases)

Ryujinx.UI/GUI/Widgets/HomeUI.cs
```diff
@@ -0,0 +1,105 @@
+﻿using ImGuiNET;
+using System;
+
+namespace Ryujinx.UI
+{
+    partial class EmulationWindow
+    {
+        void RenderMainUI()
+        {
+            ImGui.SetNextWindowPos(System.Numerics.Vector2.Zero, Condition.Always,
```
gdkchan: imo Add a `using System.Numerics` to avoid needing to write it all the time, if possible.

Ryujinx.UI/GUI/WindowHelper.cs
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        private unsafe void RenderImDrawData(DrawData* DrawData)
+        {
+            Vector4 ClearColor = new Vector4(114f / 255f, 144f / 255f, 154f / 255f, 1.0f);
+            GL.Viewport(0, 0, Width, Height);
+            GL.ClearColor(ClearColor.X, ClearColor.Y, ClearColor.Z, ClearColor.W);
+            GL.Clear(ClearBufferMask.ColorBufferBit);
+
+            // We are using the OpenGL fixed pipeline to make the example code simpler to read!
```
gdkchan: Maybe this comment doesn't belong here as this is not example code I believe? Also would be nice to not use legacy features. But this can be done later.

Ryujinx.UI/GUI/Widgets/FilePicker.cs
```diff
@@ -0,0 +1,214 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Numerics;
+
+namespace ImGuiNET
+{
+    /// <summary>
```
gdkchan: I think this should be a normal comment. The point of using this is showing a description of what the class does on the IDE, but this is not the case here.

Ryujinx.UI/EmulationController.cs
```diff
@@ -0,0 +1,74 @@
...
+        public event EventHandler IsShutDown;
+
+        public EmulationController(Switch Ns)
+        {
+            this.Ns = Ns;
+        }
+
+        public void Resume()
+        {
+            lock (Ns)
```
gdkchan: Why is the lock necessary? It doesn't seems that those methods can be accessed by more than one thread.

emmauss:
addressed comments

Ryujinx.UI/Extensions/ControlArchive.cs
```diff
@@ -0,0 +1,54 @@
...
+                    LanguageEntries[index] = new LanguageEntry()
+                    {
+                        AplicationName = Encoding.ASCII.GetString(LanguageReader.ReadBytes(0x200)).Trim('\0'),
+                        DeveloperName  = Encoding.ASCII.GetString(LanguageReader.ReadBytes(0x100)).Trim('\0')
+                    };
+                }
+            }
+        }
+    }
+
```
gdkchan: nit: New line not necessary.

Ryujinx.HLE/Loaders/Executables/Nro.cs
```diff
@@ -2,20 +2,21 @@
...
 {
-    class Nro : IExecutable
+    public class Nro : IExecutable
     {
         public string Name { get; private set; }
 
         public byte[] Text { get; private set; }
         public byte[] RO   { get; private set; }
-        public byte[] Data { get; private set; }
+        public byte[] Data { get; private set; }
```
gdkchan: Spacing after end of line not necessary.

Ryujinx.HLE/OsHle/Horizon.cs
```diff
@@ -161,8 +179,8 @@ internal void ExitProcess(int ProcessId)
 
             if (Processes.TryRemove(ProcessId, out Process))
             {
+
```
gdkchan: nit: New line not necessary.

emmauss:
addressed comments

emmauss:
should be complete now. It should work with the current input changes, but I do not have a controller to test it.

emmauss:
anyone want to test this. appveyor should be reconfigured to build this in addition to the main one.

ReaperOfSouls1909:
appvey doesnt build it

emmauss:
Updated and cleaned up.

PR #183, opened 2018-06-24
Add Sse Opt. for Cmeq_V_2D, Cmgt_V_2D (Reg). Add Sse Opt. for Crc32cb, Crc32ch, Crc32cw, Crc32cx. Add 10 simple tests for Fcmgt, Fcmge, Fcmeq, Fcmle, Fcmlt (S, V) (Reg, Zero). Add 2 Cnt_V tests.

LDj3SNuD:
Update Tester as well.
Nits.


greggameplayer:
Permit Rayman Legends Definitive Edition to go further

ChocolArm64/AOptimizations.cs
```diff
@@ -6,7 +6,15 @@ public static class AOptimizations
 
     public static bool GenerateCallStack = true;
 
-    public static bool UseSse2IfAvailable = true;
+    private static bool UseAllSseIfAvailable = true;
```
gdkchan: IMO it would be better to keep those public to allow SSE optimizations to be enabled or disabled at runtime by users of the CPU.

LDj3SNuD: It would have been useful (also for the tests, to repeat them automatically with and without the optimizations enabled, therefore the "All" flag), but they were not modifiable in real time (verified through debugging), probably because they are static and therefore defined at a certain point during the initialization phase.

PR #186, opened 2018-06-24
Implementation of UBOs instead of uniform constant arrays

ReinUsesLisp:
Like title says, it uses UBOs instead of constant uniforms. It's implemented for Maxwell's constant buffers, not for other uniforms (flip, samplers).

It uses 16 KiB for every used constant buffers (that's an OpenGL standard limit), that's 4096 floats per buffer, hopefully games use less than that. It also uses vec4s instead of floats.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
@@ -145,7 +145,9 @@ namespace Ryujinx.Graphics.Gal.Shader
 
             foreach (ShaderDeclInfo DeclInfo in Decl.Uniforms.Values.OrderBy(DeclKeySelector))
             {
-                SB.AppendLine($"uniform {GetDecl(DeclInfo)}[{DeclInfo.Index + 1}];");
+                SB.AppendLine($"layout (std140) uniform {DeclInfo.Name} {{");
+                SB.AppendLine($"{IdentationStr}vec4 {DeclInfo.Name}_data[{DeclInfo.Index / 4 + 1}];");
+                SB.AppendLine($"}};");
             }
 
             if (Decl.Uniforms.Count > 0)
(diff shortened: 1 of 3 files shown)


Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -145,7 +145,9 @@ private void PrintDeclUniforms()
 
             foreach (ShaderDeclInfo DeclInfo in Decl.Uniforms.Values.OrderBy(DeclKeySelector))
             {
-                SB.AppendLine($"uniform {GetDecl(DeclInfo)}[{DeclInfo.Index + 1}];");
+                SB.AppendLine($"layout (std140) uniform {DeclInfo.Name} {{");
+                SB.AppendLine($"    vec4 {DeclInfo.Name}_data[{DeclInfo.Index / 4 + 1}];");
```
gdkchan: Please use `$"{IdentationStr}` instead of `$"    ` to allow the identation size to be easily adjusted.

Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -534,11 +536,13 @@ private string GetName(ShaderIrOperCbuf Cbuf)
                 //This may not be aways the case.
                 string Offset = "(floatBitsToInt(" + GetSrcExpr(Cbuf.Offs) + ") >> 2)";
 
-                return DeclInfo.Name + "[" + Cbuf.Pos + " + " + Offset + "]";
+                string Index = "(" + Cbuf.Pos + " + " + Offset + ")";
+
+                return $"{DeclInfo.Name}_data[{Index} / 4][{Index} % 4]";
```
gdkchan: since you need to do the division/mod here anyway, you can remove the `>> 2` from offset and use `/ 16` and `% 16`, the generated code would be imo, a bist easier to understand this way.

gdkchan: About the above comment, it's  not as simple  as the buffer position (Cbuf.Pos) is already divided by 4 so doing this change directly won't work.  I think we should decode the byte offset instead of the word offset, then the proposed change would work. But I think this can be done later as it would complicate your PR.

gdkchan:
lgtm, ty!

PR #191, opened 2018-06-26
Implement nvhost-nvdec, nvhost-nvjpg & nvhost-vic

greggameplayer:


greggameplayer:
@gdkchan so you suggest me to do others files to do the same things ?

gdkchan:
You'd need to know how those engines works to implement it. AFAIK is not publicly documented so it requires reverse engineering.

NvHostChannel needs to be changed to support multiple engines instead of just GPU (which is the only one currently supported by the emulator).

greggameplayer:
So i close it until we find a proper way to implement it

PR #192, opened 2018-06-26
Add support for Vertex Program A and other small shader improvements

gdkchan:
This change:

- Adds partial support for vertex program A shaders.
- Adds the FADD_I32 shader instruction.
- Fixes the FFMA_I encoding (it had a bit that was fixed to 0).

Vertex program A seems to be an extra shader stage on maxwell that is not used often. It seems that the gpu actually supports 2 vertex shaders, games like Farming Simulator Switch Edition needs it.

I added support on the GlslDecompiler (as adding support for it on the decoder would be way more complicated I believe).  It basically has an extra overload now that allows passing 2 shader addresses, in this case it translates the Vertex Program A as "main" and then calls Vertex Program B on the exit points. Its still necessary to handle cases like a value being passed on an attribute in VPA and then being read in VPB, then being passed again in VPB and similar cases. The easiest way I could think to do that is creating temps to store the output/input attributes shared between VPA/VPB and then replacing all uses on the IR with those temps, which is still a bit complicated to do.

I'm not very happy with this code so suggestions are aways welcome.

FYI @ReinUsesLisp as I think this may impact your work on SPIR-V aswell.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
@@ -29,8 +29,6 @@ namespace Ryujinx.Graphics.Gal.Shader
 
         private StringBuilder SB;
 
-        private bool NeedsProgramBCall;
-
         public GlslDecompiler()
         {
             InstsExpr = new Dictionary<ShaderIrInst, GetInstExpr>()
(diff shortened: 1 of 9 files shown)


gdkchan:
Should be good to merge now, the mentioned bits that were missing are implemented now.

Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -334,6 +334,14 @@ private void PrintMain()
                 SB.AppendLine(IdentationStr + DeclInfo.Name + " = " + Attr.Name + Swizzle + ";");
             }
 
+            if (Decl.ShaderType == GalShaderType.Vertex)
+            {
+                SB.AppendLine(IdentationStr + "gl_Position.xy *= flip;");
```
ReinUsesLisp: I think "flip" could be FlipUniformName instead

PR #194, opened 2018-06-27
Use structures for GLSL input and output attributes

ReinUsesLisp:
It generates valid but dumb code for non-vertex stages. It has commentaries for what should be changed for Stage B vertex shaders.

Please test it if it's going to be merged, I don't trust GLSL compilers.


ReinUsesLisp:
Closing as it was implemented without structs.

PR #196, opened 2018-06-28
Add Controller Support

Dudejoe870:
This is just Basic Controller support, should work with PS4, PS3, Xbox One, and Xbox 360 Controllers, though only Xbox 360, PS3, and PS4 Controllers have been tested.  Controllers that need to be tested are: Switch Pro Controller, PS2, and any other obscure and weird controller you happen to have.

### DIFF
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -1,5 +1,5 @@
 ﻿using OpenTK.Input;
-using Ryujinx.HLE.Input;
+using Ryujinx.UI.Input;
 using Ryujinx.HLE.Logging;
 using System;
 using System.Collections.Generic;
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -35,6 +35,9 @@ namespace Ryujinx
         public static string Controls_Left_FakeJoycon_GamePadStick_Button;
         public static string Controls_Left_FakeJoycon_GamePadTrigger_ZL;
 
+        public static string Controls_Right_FakeJoycon_GamePadJoystick_R;
+        public static string Controls_Left_FakeJoycon_GamePadJoystick_L;
+
         public static void Read(Logger Log)
         {
             string IniFolder = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
@@ -72,6 +75,9 @@ namespace Ryujinx
             Controls_Left_FakeJoycon_GamePadButton_L      = Parser.Value("Controls_Left_FakeJoycon_GamePadButton_L");
             Controls_Left_FakeJoycon_GamePadStick_Button  = Parser.Value("Controls_Left_FakeJoycon_GamePadStick_Button");
             Controls_Left_FakeJoycon_GamePadTrigger_ZL    = Parser.Value("Controls_Left_FakeJoycon_GamePadTrigger_ZL");
+
+            Controls_Right_FakeJoycon_GamePadJoystick_R   = Parser.Value("Controls_Right_FakeJoycon_GamePadJoystick_R");
+            Controls_Left_FakeJoycon_GamePadJoystick_L    = Parser.Value("Controls_Left_FakeJoycon_GamePadJoystick_L");
 
             string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes").Split(',', StringSplitOptions.RemoveEmptyEntries);
 
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -1,4 +1,5 @@
-﻿using Ryujinx.HLE.Input;
+﻿using OpenTK.Input;
+using Ryujinx.HLE.Input;
 using Ryujinx.HLE.Logging;
 using System;
 using System.Collections.Generic;
@@ -13,6 +14,8 @@ namespace Ryujinx
         public static JoyCon FakeJoyCon { get; private set; }
 
         public static float GamePad_Deadzone;
+        public static bool  GamePad_Enable;
+        public static int   GamePad_Index;
 
         public static void Read(Logger Log)
         {
@@ -30,7 +33,9 @@ namespace Ryujinx
             Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.Value("Logging_Enable_Warn")));
             Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.Value("Logging_Enable_Error")));
 
-            GamePad_Deadzone = (float)Convert.ToDouble(Parser.Value("GamePad_Deadzone"));
+            GamePad_Enable   =        Convert.ToBoolean(Parser.Value("GamePad_Enable"));
+            GamePad_Index    =        Convert.ToInt32  (Parser.Value("GamePad_Index"));
+            GamePad_Deadzone = (float)Convert.ToDouble (Parser.Value("GamePad_Deadzone"));
 
             string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes").Split(',', StringSplitOptions.RemoveEmptyEntries);
 
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -1,5 +1,5 @@
 ﻿using OpenTK.Input;
-using Ryujinx.Input;
+using Ryujinx.HLE.Input;
 using Ryujinx.HLE.Logging;
 using System;
 using System.Collections.Generic;
@@ -14,10 +14,10 @@ namespace Ryujinx
         public static JoyConKeyboard   JoyConKeyboard   { get; private set; }
         public static JoyConController JoyConController { get; private set; }
 
-        public static float GamePad_Deadzone;
-        public static bool  GamePad_Enable;
-        public static int   GamePad_Index;
-        public static float GamePad_Trigger_Threshold;
+        public static float GamePadDeadzone;
+        public static bool  GamePadEnable;
+        public static int   GamePadIndex;
+        public static float GamePadTriggerThreshold;
 
         public static void Read(Logger Log)
         {
@@ -35,10 +35,10 @@ namespace Ryujinx
             Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.Value("Logging_Enable_Warn")));
             Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.Value("Logging_Enable_Error")));
 
-            GamePad_Enable            =        Convert.ToBoolean(Parser.Value("GamePad_Enable"));
-            GamePad_Index             =        Convert.ToInt32  (Parser.Value("GamePad_Index"));
-            GamePad_Deadzone          = (float)Convert.ToDouble (Parser.Value("GamePad_Deadzone"));
-            GamePad_Trigger_Threshold = (float)Convert.ToDouble (Parser.Value("GamePad_Trigger_Threshold"));
+            GamePadEnable            =        Convert.ToBoolean(Parser.Value("GamePad_Enable"));
+            GamePadIndex             =        Convert.ToInt32  (Parser.Value("GamePad_Index"));
+            GamePadDeadzone          = (float)Convert.ToDouble (Parser.Value("GamePad_Deadzone"));
+            GamePadTriggerThreshold  = (float)Convert.ToDouble (Parser.Value("GamePad_Trigger_Threshold"));
 
             string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes").Split(',', StringSplitOptions.RemoveEmptyEntries);
 
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -12,6 +12,8 @@ namespace Ryujinx
     {
         public static JoyCon FakeJoyCon { get; private set; }
 
+        public static float GamePad_Deadzone;
+
         public static void Read(Logger Log)
         {
             string IniFolder = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
@@ -27,6 +29,8 @@ namespace Ryujinx
             Log.SetEnable(LogLevel.Info,    Convert.ToBoolean(Parser.Value("Logging_Enable_Info")));
             Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.Value("Logging_Enable_Warn")));
             Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.Value("Logging_Enable_Error")));
+
+            GamePad_Deadzone = (float)Convert.ToDouble(Parser.Value("GamePad_Deadzone"));
 
             string[] FilteredLogClasses = Parser.Value("Logging_Filtered_Classes").Split(',', StringSplitOptions.RemoveEmptyEntries);
 
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -1,7 +1,7 @@
-﻿using OpenTK.Input;
-using Ryujinx.UI.Input;
+﻿using Ryujinx.UI.Input;
 using Ryujinx.HLE.Logging;
 using System;
+using System.Globalization;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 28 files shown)


Luffykun007:
X, Y, A, B buttons appear to be inverted Y is X and B is A with controller of X-box 360!

Dudejoe870:
In the Configuration file or in-game?  Because the default mapping is A = B, B = A, Y = X, X = Y, (The Original switch button on the left, the mapped button on the right) since the layout of the buttons is different on a Switch Controller then a traditional controller. @Luffykun007

Luffykun007:
@Dudejoe870 in-game X should be Y and A should be B only tested with x-box 360 controller.

Dudejoe870:
@Luffykun007 I can't understand your comment without knowing which is the button being registered and which is the button you're pressing.

Slick9000:
@Luffykun007 This uses Nintendo button layout by default, where the buttons are reversed. So A = B, and X = Y. There _is_ a config file where you could change the layout to suit yourself, though.

Luffykun007:
Dudejoe870 in the game Tiny barbarian when I go in controller settings there show the Y attack buttons and B jump but in the game X Attack and A jump whit this PR.

capturar1

Dudejoe870:
@Luffykun007 That's normal because the button mapping is relative to the positions of the buttons on the switch controller A on the Switch doesn't equal A on the controller, it's B because of the button layout: https://imgur.com/a/JltbcOb see how the buttons are different?  On an Xbox controller A would be B and B would be A and Y would be X and X would be Y.

Ryujinx/Config.cs
```diff
@@ -10,7 +11,13 @@ namespace Ryujinx
 {
     public static class Config
     {
-        public static JoyCon FakeJoyCon { get; private set; }
+        public static JoyConKeyboard   JoyConKeyboard   { get; private set; }
+        public static JoyConController JoyConController { get; private set; }
+
+        public static float GamePad_Deadzone;
```
AcK77: Could you rename `GamePad_Deadzone` > `GamePadDeadzone` and same for other vars names.

Luffykun007:
@ Dudejoe870 Okay then but I'm a little confused yet, it's strange in the game settings show one thing and in the game be another.

Ryujinx/JoyConController.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
AcK77: For now I think we probably need to keep files in `Ryujinx.HLE/Hid/` instead of `Ryujinx` folder. (It's available for other PR files too)

Ryujinx/JoyConReal.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Ryujinx.Input
+{
+    //TODO: Add real Bluetooth Joy Con Support in the far future.
```
AcK77: If real JoyCon are not supported, I guess create an empty file for that is not needed. We just need to create it when we will support real JoyCon.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
 
             Renderer.FrameBuffer.SetWindowSize(Width, Height);
         }
+        
+        private bool IsGamePadButtonPressedFromString(GamePadState gamePad, string str)
```
AcK77: Please, use `PascalCase` instead of `camelCase` here. And `str` can get a more explicit name too.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+            {
+                return GetGamePadTriggerFromString(gamePad, str) >= Config.GamePad_Trigger_Threshold;
+            }
+            else
+            {
+                return (GetGamePadButtonFromString(gamePad, str) == ButtonState.Pressed);
+            }
+        }
+
+        private ButtonState GetGamePadButtonFromString(GamePadState gamePad, string str) //Please make this prettier if you can.
```
AcK77: Please, use `PascalCase` instead of `camelCase` here. And `str` can get a more explicit name too.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+        {
+            ButtonState result = gamePad.Buttons.A;
```
AcK77: Please, use `PascalCase` instead of `camelCase` here.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+                    result = gamePad.DPad.Right;
+                    break;
+                case "Start":
+                    result = gamePad.Buttons.Start;
+                    break;
+                case "Back":
+                    result = gamePad.Buttons.Back;
+                    break;
+                default:
+                    Console.Error.WriteLine("Invalid Button Mapping \"" + str + "\"!  Defaulting to Button A.");
```
AcK77: Use our Logging class here instead of a direct Console write.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+            switch (str)
+            {
+                case "LTrigger":
+                    result = gamePad.Triggers.Left;
+                    break;
+                case "RTrigger":
+                    result = gamePad.Triggers.Right;
+                    break;
+                default:
+                    Console.Error.WriteLine("Invalid Trigger Mapping \"" + str + "\"!  Defaulting to 0.");
```
AcK77: Use our Logging class here instead of a direct Console write.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+            switch (str)
+            {
+                case "LJoystick":
+                    result = gamePad.ThumbSticks.Left;
+                    break;
+                case "RJoystick":
+                    result = new Vector2(-gamePad.ThumbSticks.Right.Y, -gamePad.ThumbSticks.Right.X);
+                    break;
+                default:
+                    Console.Error.WriteLine("Invalid Joystick Axis \"" + str + "\"!  Defaulting the Vector2 to 0, 0.");
```
AcK77: Use our Logging class here instead of a direct Console write.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+                    break;
+                default:
+                    Console.Error.WriteLine("Invalid Button Mapping \"" + str + "\"!  Defaulting to Button A.");
+                    break;
+            }
+
+            return result;
+        }
+
+        private float GetGamePadTriggerFromString(GamePadState gamePad, string str)
```
AcK77: Please, use `PascalCase` instead of `camelCase` here. And `str` can get a more explicit name too.

Dudejoe870:
@Luffykun007 In most situations this mapping is better because muscle memory of the game on the actual system will take over, but all you need to know is that it is completely normal and isn't an issue, if you want to change it go to the Ryujinx.conf and change it yourself.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+        {
+            float result = 0;
```
AcK77: Please, use `PascalCase` instead of `camelCase` here.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+                    Console.Error.WriteLine("Invalid Trigger Mapping \"" + str + "\"!  Defaulting to 0.");
+                    break;
+            }
+
+            return result;
+        }
+
+        private Vector2 GetJoystickAxisFromString(GamePadState gamePad, string str)
+        {
+            Vector2 result = new Vector2(0, 0);
```
AcK77: Please, use `PascalCase` instead of `camelCase` here.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+                    break;
+                default:
+                    Console.Error.WriteLine("Invalid Trigger Mapping \"" + str + "\"!  Defaulting to 0.");
+                    break;
+            }
+
+            return result;
+        }
+
+        private Vector2 GetJoystickAxisFromString(GamePadState gamePad, string str)
```
AcK77: Please, use `PascalCase` instead of `camelCase` here. And `str` can get a more explicit name too.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -55,44 +163,90 @@ protected override void OnUpdateFrame(FrameEventArgs e)
             int LeftJoystickDY = 0;
             int RightJoystickDX = 0;
             int RightJoystickDY = 0;
+            float deadzone = Config.GamePad_Deadzone;
```
AcK77: Please, use `PascalCase` instead of `camelCase` here. And `deadzone` can get a more explicit name too.

Ryujinx/Ui/GLScreen.cs
```diff
@@ -55,44 +163,90 @@ protected override void OnUpdateFrame(FrameEventArgs e)
...
+                if (Keyboard[(Key)Config.JoyConKeyboard.Right.ButtonY])     CurrentButton |= HidControllerButtons.KEY_Y;
+                if (Keyboard[(Key)Config.JoyConKeyboard.Right.ButtonPlus])  CurrentButton |= HidControllerButtons.KEY_PLUS;
+                if (Keyboard[(Key)Config.JoyConKeyboard.Right.ButtonR])     CurrentButton |= HidControllerButtons.KEY_R;
+                if (Keyboard[(Key)Config.JoyConKeyboard.Right.ButtonZR])    CurrentButton |= HidControllerButtons.KEY_ZR;
+            }
+
+            //Controller Input
+            if (Config.GamePad_Enable)
+            {
+                GamePadState gamePad = GamePad.GetState(Config.GamePad_Index);
```
AcK77: Please, use `PascalCase` instead of `camelCase` here.

Ryujinx/Ui/GLScreen.cs
Dudejoe870: I didn't know how to use the logging class, trying to figure it out now.

Ryujinx/JoyConController.cs
gdkchan: If its a config related file then I think it belongs to the UI project. IMO we shouldn't have anything related to user configuration on Ryujinx.HLE, as this is only really used by the interface.

Gamedev1909:
works good with xbox 360 controller wired

Ryujinx/Ui/JoyConKeyboard.cs
```diff
@@ -1,7 +1,6 @@
-//TODO: This is only used by Config, it doesn't belong to Core.
-namespace Ryujinx.HLE.Input
+namespace Ryujinx.UI.Input
 {
-    public struct JoyConLeft
+    public struct JoyConKeyboardLeft
```
gdkchan: Why was this (and the other structs here) renamed to JoyConKeyboard*?

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+            }
+        }
+
+        private ButtonState GetGamePadButtonFromString(GamePadState GamePad, string Button) //Please make this prettier if you can.
+        {
+            ButtonState Result = GamePad.Buttons.A;
+
+            switch (Button)
+            {
+                case "A":
```
gdkchan: imo it would be better if this and the other methods to get the button enum value from a string where not case sensitive,

Ryujinx/Ui/GLScreen.cs
```diff
@@ -44,6 +44,114 @@ protected override void OnLoad(EventArgs e)
...
+            {
+                return GetGamePadTriggerFromString(GamePad, Button) >= Config.GamePadTriggerThreshold;
+            }
+            else
+            {
+                return (GetGamePadButtonFromString(GamePad, Button) == ButtonState.Pressed);
+            }
+        }
+
+        private ButtonState GetGamePadButtonFromString(GamePadState GamePad, string Button) //Please make this prettier if you can.
```
gdkchan: This comment is not necessary...

Ryujinx/Config.cs
```diff
@@ -28,6 +35,13 @@ public static void Read(Logger Log)
             Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.Value("Logging_Enable_Warn")));
             Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.Value("Logging_Enable_Error")));
 
+            GamePadEnable            =        Convert.ToBoolean(Parser.Value("GamePad_Enable"));
+            GamePadIndex             =        Convert.ToInt32  (Parser.Value("GamePad_Index"));
+            GamePadDeadzone          = (float)Convert.ToDouble (Parser.Value("GamePad_Deadzone"), 
+                System.Globalization.CultureInfo.InvariantCulture); // Remember to use Invariant Culture when dealing with parsing doubles.
```
gdkchan: Please add a using, so you can just use `CultureInfo.InvariantCulture` instead of `System.Globalization.CultureInfo.InvariantCulture`.

Ryujinx/Config.cs
```diff
@@ -10,7 +11,13 @@ namespace Ryujinx
 {
     public static class Config
     {
-        public static JoyCon FakeJoyCon { get; private set; }
+        public static JoyConKeyboard   JoyConKeyboard   { get; private set; }
+        public static JoyConController JoyConController { get; private set; }
+
+        public static float GamePadDeadzone;
```
gdkchan: this could be a private set property aswell as you're not setting it from outside of this class.

Ryujinx/Config.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Ryujinx.HLE.Input;
+﻿using OpenTK.Input;
+using Ryujinx.UI.Input;
```
gdkchan: Please place this in alphabetical order. U comes after H.

Ryujinx/Ui/JoyConKeyboard.cs
Dudejoe870: Good point, didn't catch that.

Ryujinx/Ui/GLScreen.cs
Dudejoe870: I just though it looked kinda ugly, but okay I can remove it.

Ryujinx/Ui/JoyConKeyboard.cs
Dudejoe870: Oh because it gave me an error when I put it like that in both classes since the namespace already had that defined in the other class.

Ryujinx/Ui/GLScreen.cs
gdkchan: You could just return the enum value in each case directly. And then placing it all on a single line, like this:
```
case "A": return GamePad.Buttons.A;
```
Then you can throw an exception in the case the value is invalid and is not handled by the switch at the end of the method. This eliminates the need to print to the console (and for a default value).

Ryujinx/Ui/JoyConKeyboard.cs
emmauss: And how is it different from the JoyConController class?

Gamedev1909:
All my controllers work the layout is good and even my broken 360 controller that disconnects works as well

Dantttt:
I tried my Xbox360 controller and it works,the input is working,nice job!

gdkchan:
lgtm, thanks!

PR #197, opened 2018-06-28
Add support for vertex base on indexed draws (fixes #185)

### LINKED ISSUE #185: Stardew Valley broken gfx
sadas
@gdkchan if you know any reason why its like this it be killer to fix because fps is 60

gdkchan:
This adds support for passing the base vertex element base when a draw is made with a index buffer. The only game that is know to use that so far is Stardew Valley. This fixes most observed graphical issues on this game but the screen is still upside down.


Gamedev1909:
axiom verge uses this fix also

Gamedev1909:
and it seems steam world dig 2

PR #199, opened 2018-06-29
Explicit GLSL "position" attribute location

ReinUsesLisp:
A buggy OpenGL driver fails to link with implicit host-required "position" attribute location. This PR forces the attribute to be in location 15.

It may be an argument that if a shader uses 16 attributes they would overlap. The problem is that the driver wouldn't find a position for it either.

I found this issue in Axiom Verge on NVidia (propietary).


gdkchan:
lgtm, thanks!

PR #205, opened 2018-06-29
Add linux-x64 to RID property to make tests works on linux

Thog:
Also explicitly set the RID in all the projects


gdkchan:
lgtm, thanks.

PR #207, opened 2018-06-30
Add FaceAttr in GLSL, unmanaged case in EmitTex and ConstantColorG80 blend factor

ReinUsesLisp:


gdkchan:
Thanks! We still need to investigate whenever the GL values applies to other values (or other registers), if thats the case then I think it may be better to implement a direct case to the GL enums in those cases.

PR #209, opened 2018-07-01
Implement some GPU features

ReinUsesLisp:
Register values were extracted from here:
https://github.com/mesa3d/mesa/blob/af3685d14936844f79e6f372b4b258e29375f21b/src/mesa/drivers/dri/nouveau/nv04_state_raster.c#L38


ReinUsesLisp:
Added commits for implementing:
- Stencil testing
- Depth testing
- Face culling
- Front face (CW or CCW)

Face culling and front face are commented out from VertexEndGl because (IMO) they would mean a performance loss on the CPU side.

ReinUsesLisp:
Fixed the issue that made some games with culling enabled black screen. Those games where using flipping which made geometry's front face swap too, leaving the real front face back and vice versa. Flipping front facing when flipping is enabled did the magic.

Should be ready to review now.

ReaperOfSouls1909:
this has an issue causes black screen in stardew valley before its merge the game was fine 
capture

gdkchan:
After some investigation it seems that this is the same problem caused by the image being or not being flipped (as this affects the order of the vertices of the quad being rendered). We should flip the Y axis and the front face for this game aswell.

PR #210, opened 2018-07-01
Audio Renderer improvements

gdkchan:
The goal here is improving the current implementation of the audio renderer since it is really lacking. So far I only started implementing voices, I don't know how much of it I will implement here yet.


Cyuubi:
Lgtm.

ReaperOfSouls1909:
i have found no regressions in any working games no performance drops Great work !

Luffykun007:
I did not see regression in my tests here either.

PR #215, opened 2018-07-02
Implement Shared Fonts

Thog:
This fully implements shared fonts.

This PR is provided without fonts and therefore cannot compile as it is (We need to discuss about integrating six libre fonts that can fit in 17MB of RAM).

This PR also add Size to HSharedMem.Positions to be able to add fonts to shared zones when RequestLoad is called.

Here is the code that I have used to dump the TTF fonts if you want to test with the real system fonts https://github.com/Thog/nx-ttf-dumper


simonmkwii-dev:
This doesn't work, I compiled it with the fonts extracted from my Switch's shared memory, tested Super Mario Odyssey and shared fonts do not get loaded.

AcK77:
How can you know nothing is loaded? Have you did a memdump? Have you try another things than SMO? 

It's not because SMO does not display text that shared fonts does not work. Problem could be in shaders translator...

ReaperOfSouls1909:
if we have the shared_font.bin where do we put it for this to compile?

AcK77:
Nowhere, it's use TTF files. Please, I know you all want to try all those PR, but if a PR it isn't merge yet, there is some reason. Just be patient and wait the final solution.

Ryujinx.HLE/Font/SharedFontManager.cs
```diff
@@ -20,27 +25,40 @@ public class SharedFontManager
...
+                { SharedFontType.NintendoEx,          GetData("FontNintendoExtended")          }
             };
 
             LoadedFonts       = new uint[FontEmbeddedPaths.Count];
         }
 
+        public byte[] GetData(string FontName)
+        {
+            string FontFilePath = Path.Combine(FontsPath, $"{FontName}.ttf");
+            try
```
gdkchan: I think that instead of using a try/catch, its better to just use `File.Exists` to check whenever the file exists or not.

Ryujinx.HLE/Font/SharedFontManager.cs
```diff
@@ -20,27 +25,40 @@ public class SharedFontManager
...
+            {
+                return File.ReadAllBytes(FontFilePath);
+            } catch (FileNotFoundException e)
+            {
+                throw new SystemResourceNotFoundException($"Font \"{FontName}.ttf\" not found. Please provide it in {FontsPath}", e);
```
gdkchan: Please add quotes around the path (FontsPath), and a dot at the end of the phrase.

Thog:
Should be good now

Ryujinx.HLE/Font/SharedFontManager.cs
```diff
@@ -0,0 +1,165 @@
...
+            for (SharedFontType Type = SharedFontType.JapanUsEurope; Type < FontType; Type++)
+            {
+                Pos += GetFontSize(Type);
+                Pos += 0x8;
+            }
+
+            return Pos;
+        }
+
+        public int Count()
```
gdkchan: Why is this a method? Could be a single line property.

Ryujinx.HLE/Font/SharedFontManager.cs
```diff
@@ -0,0 +1,165 @@
...
+                // Some games don't request a load, so we need to load it here.
+                Load(FontType);
+                return 0;
+            }
+            return LoadedFonts[(int)FontType];
+        }
+
+        public uint GetFontSize(SharedFontType FontType)
+        {
+            return Convert.ToUInt32(FontEmbeddedPaths[FontType].Length);
```
gdkchan: Why not just returning the int directly? If you want to return a uint you could just cast to uint, the `Convert.ToUInt32` is not necessary.

Ryujinx.HLE/Font/SharedFontManager.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            HSharedMem SharedMem = (HSharedMem)sender;
+
+            lock (ShMemLock)
+            {
+                ShMemPositions = SharedMem.GetVirtualPositions();
+
+                (AMemory Memory, long Position, long Size) = ShMemPositions[ShMemPositions.Length - 1];
+
+                for (SharedFontType Type = SharedFontType.JapanUsEurope; (int)Type < LoadedFonts.Length; Type++)
```
gdkchan: if it was me I would just use the int type here, can this cast int to `SharedFontType` when needed for the method call.

PR #216, opened 2018-07-03
Implement CSRNG (Cryptographically Secure Random Bytes)

Cyuubi:


gdkchan:
Just a FYI, rng means "random number generator" and not "random bytes".

AcK77:
It's not `cryptographically secure random number generator` ? Who could be get by `RandomNumberGenerator.Create()`

greggameplayer:
@Cyuubi it's a buffer type-6 and not a buffer type 0x21

Cyuubi:
IIRC, the others didn't work for me.

gdkchan:
I suggest renaming `CspRnd` to `Rng` and using the correct buffer type.

gdkchan:
I just checked the docs and `RNGCryptoServiceProvider` needs to be disposed, which doesn't happen on your code.

AcK77:
Thanks!

PR #217, opened 2018-07-03
Remove tracing functionality from the CPU

gdkchan:
This remove the cpu trace option from the CPU (that was mostly useless anyway as the only thing it did was showing addresses and names if possible from subroutines/basic blocks being executed). It also removes call stack building code, instead a clever method is used to build the call stack now by walking the stack frames and reading LR return addresses from it. The call stack is now a bit different but its IMO better than before:

Old:
```
 nn::svc::aarch64::lp64::Break(nn::svc::BreakReason, unsigned long, unsigned long) (sdk:000a7bc4)
 nn::diag::detail::Abort(void) (sdk:0004546c)
 nn::diag::detail::VAbortImpl(char const*, char const*, char const*, int, nn::Result const*, nn::os::UserExceptionInfo const*, char const*, std::__va_list) (sdk:00044250)
 nn::diag::detail::AbortImpl(char const*, char const*, char const*, int, char const*, ...) (sdk:0004436c)
 Sound::Initialize(Sound::SoundType) (main:002bc230)
 System::Initialize(void) (main:002cbe04)
 nnMain (main:002c4c3c)
 Sub0000000008000090 (rtld:00000090)
 Sub0000000008000064 (rtld:00000064)
 Sub0000000008000050 (rtld:00000050)
 Sub0000000008000030 (rtld:00000030)
 Sub000000000800001c (rtld:0000001c)
```

New:
```
 nn::diag::detail::AbortImpl(char const*, char const*, char const*, int, nn::Result const*, nn::os::UserExceptionInfo const*, char const*, ...) (sdk:000442b4)
 nn::diag::detail::AbortImpl(char const*, char const*, char const*, int) (sdk:000443d8)
 SoundSystemNintendoWare::Initialize(void) (main:002bac48)
 System::Initialize(void) (main:002cbed0)
 nnMain (main:002c4c68)
 Sub00000000080000dc (rtld:000000dc)
```

Removing the old code used to build call stacks on the cpu (called on every branch to subroutine and return) also gives a very small perf boost and also reduce memory usage by the cpu a bit.


gdkchan:
Re-added tracing support on the CPU since some homebrew developers uses it for testing their homebrews. Should be more efficient than before (now it uses binary search to find names on the symbol table instead of adding new entries on a dictionary for each chunck of the function which is just dumb).

gdkchan:
#224 will be replacing this.

PR #222, opened 2018-07-06
Query multiple pages at once with GetWriteWatch

gdkchan:
Should be faster than calling it once per page.

### DIFF
--- a/ChocolArm64/Memory/AMemory.cs
+++ b/ChocolArm64/Memory/AMemory.cs
@@ -33,19 +33,25 @@ namespace ChocolArm64.Memory
 
         private byte* RamPtr;
 
+        private int HostPageSize;
+
         public AMemory()
         {
             Manager = new AMemoryMgr();
 
             Monitors = new Dictionary<int, ArmMonitor>();
 
+            IntPtr Size = (IntPtr)AMemoryMgr.RamSize + AMemoryMgr.PageSize;
+
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Ram = AMemoryWin32.Allocate((IntPtr)AMemoryMgr.RamSize + AMemoryMgr.PageSize);
+                Ram = AMemoryWin32.Allocate(Size);
+
+                HostPageSize = AMemoryWin32.GetPageSize(Ram, Size);
             }
             else
             {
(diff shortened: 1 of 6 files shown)


gdkchan:
Went ahead and made a few improvements to the cache aswell since I didn't really want to open a new PR just for that.

- It seems that on some games different types of data can share the same page (for example, part of a texture and vertex data sharing a page). On the old code it would cause that page to be invalidated every time because the buffer type would be constantly changing triggering a re-upload of the data. With the new code, multiple buffer types are allowed to use a same page. However, this still doesn't solve the problem that any write to that page will invalidate it. This means that, for example, if the first page of a texture is sharing a page with vertex data, and the vertex data is constantly updated, the texture will be invalidated aswell. Not much I can do about that.
- Using the virtual address of the GPU as key for the caches is problematic because we query physical addresses, not virtual ones. The end result is that is there are two or more virtual addresses mapped to the same physical address, and the physical address is modified, only the data using the first virtual address will be invalidated, the rest won't. I don't really remember why I used virtual address, so this PR changes that and uses physical addresses for cache keys. This *seems* to have fixed graphical issues I noticed on One Piece.

ChocolArm64/Memory/AMemory.cs
```diff
@@ -149,49 +155,53 @@ public void WriteInt32ToSharedAddr(long Position, int Value)
...
+
+            Position &= ~HostPageMask;
+
+            Size = EndPos - Position;
+
+            IntPtr[] Addresses  = new IntPtr[(Size + HostPageMask) / HostPageSize];
+
+            AMemoryWin32.IsRegionModified(MemAddress, MemSize, Addresses, out int Count);
+
+            bool[] Modified = new bool[Addresses.Length];
```
ReinUsesLisp: Since calls to IsRegionModified are coherent (called always from the same thread, without overlapping), it may be possible to have an statically allocated result variable for Addresses.Length < N where N is an arbitrary number or previous static variable size.

gdkchan:
I tried the approach above and it didn't improve speed on the game I tested.
- When using a pre-allocated array with fixed size for both `Addresses` and `Modified`, the game being tested ran much slower.
- When using a pre-allocated array with fixed size for both `Addresses` only, it seems to run with the same speed as before.

Peharps with games doing more intensive uses of the GPU, sending a lot of vertex and texture data, this could give a small perf boost. But anyway, I will keep it out for now.

PR #223, opened 2018-07-06
Do not set front face if face culling is not enabled

gdkchan:
This fixes a regression in Sonic Mania due to a invalid FrontFace value.
AFAIK, when face culling is not enabled, whatever is set for FrontFace doesn't matter, so the driver doesn't need to set the register in this case.


gdkchan:
Well, this is really old at this point, and not really relevant for now since face culling code is commented out. So I will just close this.

PR #226, opened 2018-07-07
Disable front facing and face culling to avoid regression

ReinUsesLisp:
There is a regression when face culling is enabled. It may be linked to flipping issues.
Since face culling is something that's very unlikely to cause problems when disabled, this is a solution until flipping issues are solved.

Added a commented section in OGLFramebuffer to disable face culling.


ReaperOfSouls1909:
that works thank you

PR #229, opened 2018-07-08
Blit framebuffer without shaders

ReinUsesLisp:
First of all, this PR ignores framebuffer rotation feature because it can't be implemented from glBlitFramebuffer. That aside, it allows to blit window's framebuffer without shaders and without touching current OpenGL's state (aside of framebuffer bindings).


gdkchan:
Now that #275 is in, can you update to take the changes into account? Also would be nice if hardcoded valeus for the screen size (1280x720) weren't used, or at least use some named constants to make understanding the code easier.

AcK77:
Thanks.

PR #230, opened 2018-07-08
Replace OGL buffer streaming from SubData to MapAndOprhan

ReinUsesLisp:
Currently uniform buffer data (which is sent completely every draw call) is sent to OpenGL API through glBufferSubData which works and it's safe but has an implicit synchronization, I'll try to explain it:

```
Data is sent to const buffers
Draw call is dispatched
...
Data is sent to const buffers
Driver halts the thread waiting for previous draw call to finish the buffer usage
Draw call is dispatched
```

To avoid this there are [many solutions](https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming). This PR uses buffer orphaning, what this does is allocating fresh memory (invalidating) for the buffer to use, avoiding the implicit synchronization.

I tried using a single big buffer for all const buffers with manual synchronization, but it proved to be slower.

The real solution is generating individual buffers for all data from Guest requested to be used by the GPU, binding, caching and updating those individually. It works, but has some issues with a game.

So, for now, I think this the best solution to avoid implicit synchronization.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLShader.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLShader.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 
 namespace Ryujinx.Graphics.Gal.OpenGL
 {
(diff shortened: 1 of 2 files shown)


ReaperOfSouls1909:
busted
this PR breaks stardew valley and adds flicker it breaks more you go into the game but it did fix transitions and made them very smooth

Luffykun007:
Is normal for me with GTX 960 driver 398,36.
capturar3

ReinUsesLisp:
That's the best buffer streaming technique I could think of without using explicit synchronization. I don't trust AMD's propietary implementation nor NVidia's, I do trust Mesa. If it fails there, it'd be better to keep using glBufferSubData.
After doing some more research I found that there are just a few (mobile) drivers that halt on glBufferSubData, most drivers would just keep the data and send it to GPU when the buffer is idle.

PR #235, opened 2018-07-08
AInstEmitSimdCvt: Half-precision to single-precision conversion

MerryMage:
Test examples from [Wikipedia](https://en.wikipedia.org/wiki/Half-precision_floating-point_format#Half_precision_examples).

### DIFF
--- a/ChocolArm64/Instruction/ASoftFloat.cs
+++ b/ChocolArm64/Instruction/ASoftFloat.cs
@@ -225,5 +225,41 @@ namespace ChocolArm64.Instruction
 
             return 2.0 + op1 * op2;
         }
+
+        public static float ConvertHalfToSingle(ushort x)
+        {
+            uint x_sign = (uint)(x >> 15) & 0x0001;
+            uint x_exp = (uint)(x >> 10) & 0x001F;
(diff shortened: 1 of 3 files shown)


ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -225,5 +233,41 @@ public static double RecipStep(double op1, double op2)
...
+        public static float ConvertHalfToSingle(ushort x)
+        {
+            uint x_sign = (uint)(x >> 15) & 0x0001;
+            uint x_exp = (uint)(x >> 10) & 0x001F;
+            uint x_mantissa = (uint)x & 0x03FF;
+
+            if (x_exp == 0 && x_mantissa == 0)
+            {
+                // Zero
+                return Int32BitsToFloat(x_sign << 31);
```
gdkchan: Why not using BitConverter.Int32BitsToSingle?

gdkchan:
lgtm, thank you!

ChocolArm64/Instruction/ASoftFloat.cs
gdkchan: It's a new API only available on .NET Core for now (along with MathF for operations with the single precision type).

PR #241, opened 2018-07-10
Allow copy between frame buffers

gdkchan:
This still needs improvements (de-hardcoding frame buffer sizes, and in general the texture copy engine is really lacking, it needs the ability to convert between formats, change resolution, and so on...). However I don't plan to do this work now.


gdkchan:
Closing in favor of #275 .

PR #242, opened 2018-07-10
[NV] Fix NvHostCtrlGetConfig

Cyuubi:


AcK77:
You read a struct with a size of 0x183 bytes, who looks... A little strange for me!
And you don't trim the readed values so the comparaison is probably always `false`.

AcK77:
According to switchbrew your struct seems to be good:
```
NVHOST_IOCTL_CTRL_GET_CONFIG
Gets configured settings. Not available in production mode.

 struct {
   __in char domain_str[0x41];       // "nv"
   __in char param_str[0x41];
   __out char config_str[0x101];
 };
```

But those values are already present in `Ryujinx.HLE/OsHle/Services/Set/NxSettings.cs` and seems to be not used since the content of this file is generated automatically from a switch.

Cyuubi:
GetConfig is mostly used in launch games.

Cyuubi:
Also, I just checked and the if's are working fine so...

AcK77:
As I said on Discord:

The result need to be wrote in `ConfigString` instead of `ParameterString` and you wrote the whole struct in memory who is not what the call need... Plus, you don't use the NxSetting.cs values, how are set to null because it's not used in production mode.

AcK77:
Closed in favor of better implementation in #243

PR #244, opened 2018-07-10
Loop GLScreen with custom method

ReinUsesLisp:
Bypasses OpenTK's "jail".

Using OpenTK's event system we're able to dispatch GPU calls just once per frame, this PR calls it "as much as possible". This is not ideal because CPU cycles are wasted, but it's better than current approach.

It has included an event system to avoid this (locking host FPS to 60 most of the time) but it deadlocks almost every game so it's disabled.

It also forks OpenTK's event handling and GPU rendering in two different threads, this is mainly because OpenTK malfunctions when sleeps are called in its thread.

### DIFF
--- a/Ryujinx/Ui/GLScreen.cs
+++ b/Ryujinx/Ui/GLScreen.cs
@@ -5,6 +5,9 @@ using Ryujinx.Graphics.Gal;
 using Ryujinx.HLE;
 using Ryujinx.HLE.Input;
 using System;
+using System.Threading;
+
+using Stopwatch = System.Diagnostics.Stopwatch;
 
 namespace Ryujinx
 {
@@ -16,6 +19,8 @@ namespace Ryujinx
         private const float TouchScreenRatioX = (float)TouchScreenWidth  / TouchScreenHeight;
         private const float TouchScreenRatioY = (float)TouchScreenHeight / TouchScreenWidth;
 
+        private const int TargetFPS = 60;
+
         private Switch Ns;
 
         private IGalRenderer Renderer;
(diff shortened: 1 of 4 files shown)


ReaperOfSouls1909:
nice work it gave a speed boost in stardewvalley by up to 10 more frames and even in binding/cave story

ReaperOfSouls1909:
SO here is what my tests came up with gpu intensive games like stardew valley and  Disgaea 5 Complete got faster but cpu intensive games like cave story and binding got slower in the long run so this PR is good and causes some minor issues on the cpu side if you want me to try anything else please ask ive tested all working games

ReinUsesLisp:
Did you test it before or after d2e3f2d commit?

ReaperOfSouls1909:
will redo my tests

ReaperOfSouls1909:
re did my testing 2D is fine and gpu intensive titles hold the boost great fix great PR

Ryujinx/Ui/GLScreen.cs
```diff
@@ -36,13 +45,73 @@ public GLScreen(Switch Ns, IGalRenderer Renderer)
...
+            while (!IsExiting)
+            {
+                if (Ns.WaitFifo())
+                {
+                    Ns.ProcessFrame();
+                }
+
+                Renderer.RunActions();
+
+                if (ResizeEvent.WaitOne(0))
```
gdkchan: Imo a event is not necessary here. You could just use a bool and set it to false inside the condition. However as this is multithreaded you'd need to ensure that the test and set operation would be atomic. You can use the Interlocked methods and a int var for that.

ReaperOfSouls1909:
Can confirm from new tests all is well stardew valley and OP unlimited red and D5 is faster with this pr 
it causes no extra issues that i have found

gdkchan:
lgtm, thanks!

PR #245, opened 2018-07-10
Implement Fminp_V & Fmaxp_V

greggameplayer:


gdkchan:
If the only difference is that one call min and the other calls max, it would IMO be better to create a method to handle both (similar to the other EmitVectorPairwise method on AInstEmitSimdHelper, but for floats). It could be used with FADDP too.

AcK77:
Maybe @LDj3SNuD, @MerryMage or @gdkchan can review it please ?

AcK77:
I got a `System.ArgumentNullException : 'Value cannot be null.'` on the `EmitCall(MethodInfo MthdInfo)` method here...

AcK77:
I close this PR in favor of #412

PR #249, opened 2018-07-10
Add return of Texture Size and Bytes Per Pixel of ASTC2D 5x5, 6x6, 8x8, 10x10 and 12x12

greggameplayer:


Ryujinx.HLE/Gpu/Texture/TextureHelper.cs
```diff
@@ -51,10 +51,7 @@ public static int GetTextureSize(GalTexture Texture)
                 case GalTextureFormat.BC1:
                 case GalTextureFormat.BC4:
                 {
-                    int W = (Texture.Width  + 3) / 4;
-                    int H = (Texture.Height + 3) / 4;
-
-                    return W * H * 8;
+                    return CompressedTextureSize(Texture.Width, Texture.Height, 4, 4);
```
gdkchan: This look wrong, BC1/BC4 textures only uses 8 bytes per block (unlike the other formats that uses 16), So I don't think you can just use the same method.

greggameplayer: Ok i will just add an argument

PR #250, opened 2018-07-12
TimeZone implements cmd 0, 1, 2, 3, 4 and 100

Thog:
The implementation of the TimezoneRule isn't matching hardware but
doesn't need to be accurate (games are only passing the value)

### DIFF
--- a/Ryujinx.HLE/OsHle/Services/Time/ITimeZoneService.cs
+++ b/Ryujinx.HLE/OsHle/Services/Time/ITimeZoneService.cs
@@ -2,6 +2,7 @@ using Ryujinx.HLE.Logging;
 using Ryujinx.HLE.OsHle.Ipc;
 using System;
 using System.Collections.Generic;
+using System.Text;
 
 namespace Ryujinx.HLE.OsHle.Services.Time
 {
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/OsHle/Services/Time/ITimeZoneService.cs
```diff
@@ -11,66 +12,177 @@ class ITimeZoneService : IpcService
 
         public override IReadOnlyDictionary<int, ServiceProcessRequest> Commands => m_Commands;
 
+        private TimeZoneInfo TimeZone          = TimeZoneInfo.Local;
```
gdkchan: You don't need to have all assignments with the `=` aligned. For example here it looks a bit weird that Epoch and TimeZone are aligned like this. I usually only do it when it improves readability. When the variables are not really related I usually add a empty line in between. So, feel free to align it or not when you see fit.

Ryujinx.HLE/OsHle/Services/Time/ITimeZoneService.cs
```diff
@@ -11,66 +12,177 @@ class ITimeZoneService : IpcService
...
+                { 4,   LoadTimeZoneRule          },
+                { 100, ToCalendarTime            },
+                { 101, ToCalendarTimeWithMyRule  }
             };
         }
 
         public long GetDeviceLocationName(ServiceCtx Context)
         {
-            Context.Ns.Log.PrintStub(LogClass.ServiceTime, "Stubbed.");
+            char [] TzName = TimeZone.Id.ToCharArray();
```
gdkchan: nit: `char []` -> `char[]`. Some comment I made for TimeZone also applies here.

Ryujinx.HLE/OsHle/Services/Time/ITimeZoneService.cs
```diff
@@ -11,66 +12,177 @@ class ITimeZoneService : IpcService
...
+
+        public long LoadLocationNameList(ServiceCtx Context)
+        {
+            long BufferPosition = Context.Response.SendBuff[0].Position;
+            long BufferSize     = Context.Response.SendBuff[0].Size;
+
+            int i = 0;
+            foreach (TimeZoneInfo info in TimeZoneInfo.GetSystemTimeZones())
+            {
+                byte [] TzData = Encoding.ASCII.GetBytes(info.Id);
```
gdkchan: nit: `byte []` -> `byte[]`. Some comment I made for TimeZone also applies here.

AcK77:
Thanks a lot!

PR #252, opened 2018-07-13
Implement some shader instructions

ReinUsesLisp:


pendingchaos:
add_with_carry() is basically:
```
uint32_t add_with_carry(uint32_t a, uint32_t b) {
    uint64_t result = a + b;
    if (x) result += carry_flag;
    if (cc && result > 0xffffffff) carry_flag = 1;
    return result & 0xffffffff;
}
```

MerryMage:
[Carry flag](https://en.wikipedia.org/wiki/Carry_flag).

@pendingchaos While using a wider width integer to calculate the carry flag makes sense, you don't have to do that:

```
(uint, bool) AddWithCarry(uint LHS, uint RHS, bool CarryIn) {
    uint Result = LHS + RHS + (CarryIn ? 1u : 0u);
    bool CarryOut = Result < LHS; // If the result is less than one of the inputs, we have overflowed
    return (Result, CarryOut);
}
```

pendingchaos:
Yeah, it wasn't meant to be optimal performance-wise. It was just to explain what it does.

Also, there aren't overflow and zero flag registers and such. Just one general purpose (and as far as I know, one bit) carry flag/condition code/flags/status/whatever register.

Ryujinx.Graphics/Gal/Shader/ShaderOpCodeTable.cs
```diff
@@ -64,20 +64,33 @@ static ShaderOpCodeTable()
             Set("0100110011100x", ShaderDecode.I2i_C);
             Set("0011100x11100x", ShaderDecode.I2i_I);
             Set("0101110011100x", ShaderDecode.I2i_R);
+            Set("0001110xxxxxxx", ShaderDecode.Iadd_I32_1c0);
```
gdkchan: It seems that the only difference between this iadd and the other one with the 32-bits immediate is the `po` modifier. IMO it would be better to just call this one `Iadd_I32` and the other one `Iadd_I32_Po`. Also it should go below Iadd_I.

gdkchan: Also I think the encoding is supposed to be `0001110x0xxxxx`.

gdkchan:
lgtm, thanks! The carry case can be handled later.

PR #258, opened 2018-07-13
Some small shader related fixes

gdkchan:
This fixes the following issues:

- There was no method to unbind shaders. So if a game used for example, a geometry shader once and never anymore, it would be still bound. An unbind method was added to allow unbinding the shader when the enable flag is set to 0 on the register.
- Code like `--2` (double negation) is not valid in glsl it seems. GlslDecompiler is now more strict about adding parenthesis, and will add it to all negative constants. So now instead of `--2` it writes `-(-2)`. This also impacts some unary operations that doesn't need it through.
- Reading/writing the built-in gl_Position is valid on geometry shader, so now this is allowed instead of using the `position` varying. It also now writes it aswell. This caused problems before because the fragment shader would try to use that varying without the geometry shader writing it.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
@@ -216,7 +216,7 @@ namespace Ryujinx.Graphics.Gal.Shader
 
         private void PrintDeclOutAttributes()
         {
-            if (Decl.ShaderType < GalShaderType.Fragment)
+            if (Decl.ShaderType != GalShaderType.Fragment)
             {
                 SB.AppendLine("layout (location = " + GlslDecl.PositionOutAttrLocation + ") out vec4 " + GlslDecl.PositionOutAttrName + ";");
             }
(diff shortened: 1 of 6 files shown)


Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -216,7 +216,7 @@ private void PrintDeclInAttributes()
 
         private void PrintDeclOutAttributes()
         {
-            if (Decl.ShaderType == GalShaderType.Vertex)
+            if (Decl.ShaderType < GalShaderType.Fragment)
```
ReinUsesLisp: I think it would be more readable a `!=` instead of `<`

Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -337,7 +337,10 @@ private void PrintMain()
             if (Decl.ShaderType == GalShaderType.Vertex)
             {
                 SB.AppendLine(IdentationStr + "gl_Position.xy *= " + GlslDecl.FlipUniformName + ";");
+            }
 
+            if (Decl.ShaderType < GalShaderType.Fragment)
```
ReinUsesLisp: Same as above

PR #259, opened 2018-07-13
Link ISslContext via function CreateContext of ISslService

greggameplayer:
the link is required by Doom but this game don't use functions of ISslContext

### DIFF
--- a/Ryujinx.HLE/OsHle/Services/Ssl/ISslContext.cs
+++ b/Ryujinx.HLE/OsHle/Services/Ssl/ISslContext.cs
@@ -1,0 +1,21 @@
+using Ryujinx.HLE.Logging;
+using Ryujinx.HLE.OsHle.Ipc;
+using System.Collections.Generic;
+
+namespace Ryujinx.HLE.OsHle.Services.Ssl
+{
+    class ISslContext : IpcService
+    {
+        private Dictionary<int, ServiceProcessRequest> m_Commands;
+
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/OsHle/Services/Ssl/ISslContext.cs
```diff
@@ -0,0 +1,21 @@
+using Ryujinx.HLE.Logging;
```
AcK77: Using is not necessary.

Ryujinx.HLE/OsHle/Services/Ssl/ISslService.cs
```diff
@@ -14,9 +14,19 @@ public ISslService()
...
             m_Commands = new Dictionary<int, ServiceProcessRequest>()
             {
+                { 0, CreateContext       },
                 { 5, SetInterfaceVersion }
             };
         }
+		
+        public long CreateContext(ServiceCtx Context)
+        {
+            long Id = Context.RequestData.ReadInt64();
```
AcK77: `CreateContext(nn::ssl::sf::SslVersion, u64, pid) -> object<nn::ssl::sf::ISslContext>`

Maybe you should read the version and the u64 (as an Unknown here). It's not really useful but it's better to get something more consistent.

ReaperOfSouls1909:
Doom needs this service #354

AcK77:
Since the current PR need a rebase... I have make a new one with my change.
So I close this PR in favor of #440 .
Thanks!

PR #262, opened 2018-07-14
Minecraft Bedrock edition stubs.

Cyuubi:
Minecraft Bedrock requires these functions and currently doesn't render with these (black screen). But it does play audio after like 3-6 minutes of loading it. (Creates 1815 NvMaps)

### DIFF
--- a/Ryujinx.HLE/OsHle/Services/Am/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/OsHle/Services/Am/IApplicationFunctions.cs
@@ -19,6 +19,8 @@ namespace Ryujinx.HLE.OsHle.Services.Am
                 { 21, GetDesiredLanguage          },
                 { 22, SetTerminateResult          },
                 { 23, GetDisplayVersion           },
+                { 25, ExtendSaveData              },
+                { 26, GetSaveDataSize             },
                 { 40, NotifyRunning               },
                 { 50, GetPseudoDeviceId           },
                 { 66, InitializeGamePlayRecording },
@@ -77,6 +79,25 @@ namespace Ryujinx.HLE.OsHle.Services.Am
             //FIXME: Need to check correct version on a switch.
             Context.ResponseData.Write(1L);
             Context.ResponseData.Write(0L);
+
+            return 0;
+        }
+
+        public long ExtendSaveData(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
+
+            Context.ResponseData.Write(0x400L);
+
+            return 0;
+        }
+
+        public long GetSaveDataSize(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
+
+            Context.ResponseData.Write(0L);
+            Context.ResponseData.Write(0x200L);
 
             return 0;
         }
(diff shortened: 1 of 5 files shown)


Cyuubi:
Update: apparently it is sending vertexs

Ryujinx.HLE/OsHle/Services/FspSrv/IFileSystemProxy.cs
```diff
@@ -57,6 +59,31 @@ public long OpenDataStorageByCurrentProcess(ServiceCtx Context)
...
+        public long OpenDataStorageByDataId(ServiceCtx Context)
+        {
+            StorageTypes StorageId = (StorageTypes) Context.RequestData.ReadByte();
+            long         TitleId   = Context.RequestData.ReadInt64();
+
+            if (StorageId == StorageTypes.NandSystem)
+            {
+                Context.Ns.Log.PrintStub(LogClass.ServiceFs, $"[{StorageId}] Stubbed.");
+
+                Stream FakeStream = new MemoryStream();
```
Thog: This isn't how it should be stub. This can cause issues if games are using anything that isn't titleId 0100000000000806.

Ryujinx.HLE/OsHle/Services/Am/IApplicationFunctions.cs
```diff
@@ -81,6 +83,25 @@ public long GetDisplayVersion(ServiceCtx Context)
...
+            Context.ResponseData.Write(0x400L);
+
+            return 0;
+        }
+
+        public long GetSaveDataSize(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
+
+            Context.ResponseData.Write(0L);
```
Thog: This should match the first parameter sent to this command, otherwise some games will not work.

Ryujinx.HLE/OsHle/Services/Pctl/IParentalControlService.cs
```diff
@@ -37,5 +38,12 @@ public long Initialize(ServiceCtx Context)
 
             return 0;
         }
+
+        public long CheckFreeCommunicationPermission(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServicePctl, "Stubbed.");
```
Thog: This isn't a stub, this command returns an error code if permission isn't granted.

ReaperOfSouls1909:
@Thog you are correct but now that we know what needs to be implemented for the game to work maybe someone will take note and do the work maybe even you as you are dam good at doing so

AcK77:
@ReaperOfSouls1909 Can I know your nickname on the Ryujinx Discord?

ReaperOfSouls1909:
Gamedev19o9 {o}{o}#8462 same as its always been

AcK77:
Okay so I guess I can call you McDev since it’s your real nickname and since you have been banned two times from our discord.
Then, please just be careful on how you speak to other devs here, in the comment above you used an ironic way and I don’t really like that. Thanks. Discussion close, back to the review.

ReaperOfSouls1909:
wait so because i made a complement to thog saying he is skilled i get yelled at for doing nothing wrong... riiight seems logical to someone

AcK77:
I said discussion closed. Thanks.

gdkchan:
I agree with Thog, stubs tend to cause harder to find bugs as the games will be working with invalid data which can lead to all kind of bugs or crashes later on. We already have many stubs and I think work should be focused in a better implementation for those, or whem implementing new services, doing a proper implementation. That said, I don't have anything against stubs for simple functions that we know are "safe" to stub (for example, functions that just changes some internal state we don't need to care about).

PR #265, opened 2018-07-14
Add config key to dump shaders in local directory

ReinUsesLisp:
Shader dumps should be compatible with `nvdisasm`.

### DIFF
--- a/Ryujinx.Graphics/Gal/ShaderDumper.cs
+++ b/Ryujinx.Graphics/Gal/ShaderDumper.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using System.IO;
+
+namespace Ryujinx.Graphics.Gal
+{
+    static class ShaderDumper
+    {
+        private static string RuntimeDir = "";
+
+        private static int DumpIndex = 1;
+
+        public static void Dump(IGalMemory Memory, long Position, GalShaderType Type, string ExtSuffix = "")
+        {
+            if (AGraphicsConfig.ShadersDumpPath == "")
+            {
+                return;
+            }
+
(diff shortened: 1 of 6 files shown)


Ryujinx.Graphics/Gal/ShaderDumper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            using (FileStream Output = File.Create(Path))
+            using (BinaryWriter Writer = new BinaryWriter(Output))
+            {
+                long Offset = 0;
+
+                ulong Instruction = 0;
+
+                //Dump until a NOP instruction is found
+                while (Instruction >> 52 != 0x50b)
```
gdkchan: This is missing one bit on the encoding for NOP, as its currently checking the 12 upper bits instead of the 13 upper bits.

Ryujinx.Graphics/Gal/ShaderDumper.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private static string DumpDir()
+        {
+            if (RuntimeDir == "")
+            {
+                int Index = 1;
+
+                do
+                {
+                    RuntimeDir = AGraphicsConfig.ShadersDumpPath + "/Dumps" + Index.ToString("d2");
```
gdkchan: use `Path.Combine` there that is flexible regarding slashes on the path. For example, this will most likely break if the user set the path as for example `C:\ShaderDumps\`, then it will use `C:\ShaderDumps\/Dumps00` which is not a valid path. `Path.Combine` takes care of that.

PR #266, opened 2018-07-14
Implement IProfile Get and GetBase

emmauss:
This implements Get and GetBase of IProfile. If tests are available, I will add loading of profile image. The default username is "Ryujinx".


emmauss:
needs verification with minecraft bedrock

emmauss:
anyone want to review this?

Ryujinx.HLE/OsHle/Services/Acc/IProfile.cs
```diff
@@ -14,21 +17,55 @@ public IProfile()
...
+            }
+
+            Context.ResponseData.Write(ProfileBase.UserId);
+            Context.ResponseData.Write(ProfileBase.Timestamp);
+
+            int    ByteCount = Encoding.UTF8.GetByteCount(ProfileBase.Username);
+            byte[] Username  = new byte[0x20];
+
+            if (ByteCount > 0x20)
+                Encoding.UTF8.GetBytes(ProfileBase.Username, 0, ProfileBase.Username.Length, Username, 0x20 - ByteCount);
```
gdkchan: nit: Use `{` and `}`. Maybe its better to have a common helper method for this sort of thing on Utilities (as it seems to be a pretty common operation to convert a string to a padded byte array). Also, doesn't the string need to be null terminated?

Ryujinx.HLE/OsHle/Services/Acc/ProfileBase.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Ryujinx.HLE.OsHle.Services.Acc
+{
+    struct ProfileBase
+    {
+        public byte[] UserId;
```
gdkchan: Imo, instead of the byte array its better to use a struct for the user id with 2 longs inside, with the low and high part of the 128-bits value. Then, expose constructors that allows parsing it from a string or another that allows passing 2 long values. They should throw on any invalid case (string having non 0-9, A-F, a-f character, or length being != 16). This prevents user of the code from creating user ids with invalid lengths.

Ryujinx/Config.cs
```diff
@@ -19,6 +19,9 @@ public static class Config
         public static int   GamePadIndex                { get; private set; }
         public static float GamePadTriggerThreshold     { get; private set; }
 
+        public static string UserId                     { get; private set; }
```
gdkchan: After a new line, IMO aligning the `{` with the above lines is not necessary.

Ryujinx/Config.cs
```diff
@@ -132,6 +145,16 @@ public static void Read(Logger Log)
                 }
             };
         }
+
+        public static string GenerateUserId()
+        {
+            byte[] UserId = new byte[16];
+
+            Random Random = new Random();
+            Random.NextBytes(UserId);
```
gdkchan: Note: 0 is not a valid user id, and should not be allowed here.

Ryujinx/Config.cs
```diff
@@ -132,6 +145,16 @@ public static void Read(Logger Log)
...
+
+            return BitConverter.ToString(UserId).ToLower().Replace("-", string.Empty);
```
gdkchan: I guess that `ToLower` is not necessary as it doesn't matter if the characters are lower or uppercased as long they are valid hex values.

emmauss: I am using it for readability. having lowercase hex look nicer to me, since they are logically the same.

Ryujinx/Config.cs
emmauss: okay

Ryujinx/Config.cs
gdkchan: I'm fine with keeping it then.

Ryujinx.HLE/OsHle/Services/Acc/ProfileBase.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    struct UserId
+    {
+        readonly ulong LowBytes;
+        readonly ulong HighBytes;
+
+        public UserId(string UserIdHex)
+        {
+            if (string.IsNullOrWhiteSpace(UserIdHex))
```
gdkchan: IMO just check for null, the exception will be misleading if the string contains whitespace (and it would be already catched by the other checks). Also please use `{` and `}`.

Ryujinx.HLE/OsHle/Services/Acc/ProfileBase.cs
```diff
@@ -0,0 +1,47 @@
...
+    struct ProfileBase
+    {
+        public UserId UserId;
+        public long   Timestamp;
+        public string Username;
+    }
+
+    struct UserId
+    {
+        readonly ulong LowBytes;
```
gdkchan: nit: `private readonly ulong` or just `private ulong`.

Ryujinx.HLE/OsHle/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Globalization;
+using System.Text;
+
+namespace Ryujinx.HLE.OsHle.Utilities
+{
+    internal static class StringUtils
```
gdkchan: Why `internal`? This is not going to be accessed by other projects on the same assembly I believe.

Ryujinx.HLE/OsHle/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        public static byte[] GetFixedLengthBytes(string InputString, int Size, Encoding Encoding)
+        {
+            int ByteCount = Encoding.GetByteCount(InputString);
+
+            byte[] Output = new byte[Size];
+
+            if (ByteCount > Size)
+            {
+                Encoding.UTF8.GetBytes(InputString, 0, InputString.Length, Output, Size - ByteCount);
```
gdkchan: It's aways using UTF8 encoding here even through it accepts an `Encoding` argument.

Ryujinx.HLE/OsHle/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+            else
+            {
+                Output = Encoding.UTF8.GetBytes(InputString);
```
gdkchan: It's aways using UTF8 encoding here even through it accepts an `Encoding` argument.

Ryujinx.HLE/OsHle/Utilities/StringUtils.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using System.Globalization;
+using System.Text;
+
+namespace Ryujinx.HLE.OsHle.Utilities
+{
+    internal static class StringUtils
+    {
+        public static byte[] GetFixedLengthBytes(string InputString, int Size, Encoding Encoding)
```
gdkchan: I think its expected that the strings are null terminated. What happens if a string (without the null terminator) has ByteCount == Size, will the game still be able to read it properly without the null terminator?

Ryujinx/Config.cs
```diff
@@ -35,6 +38,16 @@ public static void Read(Logger Log)
             Log.SetEnable(LogLevel.Warning, Convert.ToBoolean(Parser.Value("Logging_Enable_Warn")));
             Log.SetEnable(LogLevel.Error,   Convert.ToBoolean(Parser.Value("Logging_Enable_Error")));
 
+            Username = Parser.Value("Username");
+            UserId   = Parser.Value("User_Id");
+
+            if (string.IsNullOrWhiteSpace(Username))
+                Username = "Ryujinx";
+
+            if (string.IsNullOrWhiteSpace(UserId) || UserId.Length != 32
```
gdkchan: Instead of `string.IsNullOrWhiteSpace` only a check for null is necessary, the other cases are already covered with the other checks. Also imo put `UserId.Length != 32 ` on a new line. Like:
```csharp
if (UserId == null ||
    UserId.Length != 32 ||
    !UserId.All("0123456789abcdefABCDEF".Contains))
```

Ryujinx.HLE/OsHle/Services/Acc/IProfile.cs
```diff
@@ -14,21 +17,40 @@ public IProfile()
...
-            Context.ResponseData.Write(0L);
-            Context.ResponseData.Write(0L);
-            Context.ResponseData.Write(0L);
-            Context.ResponseData.Write(0L);
-            Context.ResponseData.Write(0L);
+            long Position = Context.Request.ReceiveBuff[0].Position;
+
+            Context.Memory.WriteInt32(Position, 0);
+            Context.Memory.WriteInt32(Position + 4, 1);
+            Context.Memory.WriteByte(Position + 8, 1);
```
gdkchan: This still uses `WriteByte` and will still contain garbage values or the other 3 bytes (assuming that it is read as a 32-bits integer on guest code, which I assume it is).

gdkchan: Also would be nice to have an explanation on what those values means (0, 1 and 1).

emmauss: they are values for this http://switchbrew.org/index.php?title=Account_services#UserData . the first field is unknown, so I used 1. icon is not a mii, so 1 for charactor icon( which we we get from a jpg file. the last one is the background color id. I used one for that.the rest are either unknown or 0s. I dont really know what the 1st and 3rd field expect for valid values, so I used those as sample value.

emmauss: the buffer is for the userdata, not the profilebase. the profile base is written as normal data into the responsedata. it is the same as calling GetBase with the same context.

emmauss: I have initialized the buffer with zeros

emmauss:
It now supports multiple users. notes in config

Cyuubi:
Looks good to me, IMO.

emmauss:
I just want to check if it passes the acc service, not if it goes ingame. this is the only game i know that requests user accounts before getting stuck on something else later on.

Ryujinx.HLE/Settings/SystemSettings.cs
```diff
@@ -1,7 +1,14 @@
...
 namespace Ryujinx.HLE.Settings
 {
     public class SystemSettings
     {
-        public ColorSet ThemeColor;
+        public ColorSet      ThemeColor;
+        public List<Profile> UserProfiles;
+        public int           DefaultUserIndex;
+
+        public Profile ActiveUser { get => UserProfiles[DefaultUserIndex]; }
```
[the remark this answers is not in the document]
gdkchan: Also since you're using properties, at least for ActiveUser, Why not just using auto properties for all fields aswell.

emmauss:
addressed comments

emmauss:
@gdkchan lets just leave the index there. It would be helpful when proper save data is implemented with respect to accounts.

emmauss:
added a default user to config, cause we dont want to create a different save folder anytime we start the same game.

emmauss:
rebased. Also hardcoded a default user

Ryujinx.HLE/Switch.cs
```diff
@@ -67,6 +67,12 @@ public Switch(IGalRenderer Renderer, IAalOutput AudioOut)
 
             Os.FontSharedMem.MemoryMapped   += Font.ShMemMap;
             Os.FontSharedMem.MemoryUnmapped += Font.ShMemUnmap;
+
+            Settings.User = new Profile()
+            {
+                Username = "Ryujinx",
+                UserId   = "00000000000000000000000000000000"
```
gdkchan: Like I said before, 0 is not a valid user id.

emmauss: forgot that.

emmauss: fixed

PR #270, opened 2018-07-15
Fix Issue 251 (FrintP and FrintI/X with FPCR set to "P" with -1 < x < -0 input)

MS-DOS1999:


ChocolArm64/Instruction/AVectorHelper.cs
```diff
@@ -178,7 +178,7 @@ public static double Round(double Value, int Fpcr)
             switch ((ARoundMode)((Fpcr >> 22) & 3))
             {
                 case ARoundMode.ToNearest:            return Math.Round   (Value);
-                case ARoundMode.TowardsPlusInfinity:  return Math.Ceiling (Value);
+                case ARoundMode.TowardsPlusInfinity:  {if(Value < 0.0 && Value > -1.0){return -0.0;}return Math.Ceiling (Value);}
```
gdkchan: Please don't cram everything in one line like that, its hard to read.

riperiperi: Maybe:
`return (Value < 0.0 && Value > -1.0) ? -0.0 : Math.Ceiling(Value);`

gdkchan: This code doesn't work for all cases because on x86 `-0.0 == 0.0`.

PR #272, opened 2018-07-15
hbabi: Implement argv (Fix #94)

### LINKED ISSUE #94: Add support for nro files with romfs
Currently Ryujinx cannot access romfs from an nro that has one.
Works on hardware, and attached is my application that does this.

[LovePotion.zip](https://github.com/gdkchan/Ryujinx/files/1933354/LovePotion.zip)

While I can call `romfsInit`, it just doesn't return the right code (which would be 0).
From there I change to the romfs directory to run the game's files.

Not sure if those details are helpful, but if you need them, there they are 😄

Thog:
This PR implements the argv config key in Ryujinx to make it possible
to load libnx's "romfs" files.

This PR also propagates the file path to the executable instances.

### DIFF
--- a/Ryujinx.HLE/Loaders/Executable.cs
+++ b/Ryujinx.HLE/Loaders/Executable.cs
@@ -2,6 +2,7 @@ using ChocolArm64.Memory;
 using Ryujinx.HLE.Loaders.Executables;
 using Ryujinx.HLE.OsHle;
 using System.Collections.Generic;
+using System.IO;
 
 namespace Ryujinx.HLE.Loaders
 {
@@ -15,6 +16,8 @@ namespace Ryujinx.HLE.Loaders
 
         public string Name { get; private set; }
 
+        public string FilePath { get; private set; }
+
         private AMemory Memory;
 
         public long ImageBase { get; private set; }
@@ -26,8 +29,9 @@ namespace Ryujinx.HLE.Loaders
 
             m_SymbolTable = new Dictionary<long, string>();
 
-            Name = Exe.Name;
+            FilePath = Exe.FilePath;
 
+            Name = Path.GetFileNameWithoutExtension(FilePath.Replace(Homebrew.TemporaryNroSuffix, ""));
             this.Memory    = Memory;
             this.ImageBase = ImageBase;
             this.ImageEnd  = ImageBase;
--- a/Ryujinx.HLE/OsHle/Homebrew.cs
+++ b/Ryujinx.HLE/OsHle/Homebrew.cs
@@ -1,11 +1,14 @@
 ﻿using ChocolArm64.Memory;
+using System.Text;
 
 namespace Ryujinx.HLE.OsHle
 {
     static class Homebrew
     {
+        public const string TemporaryNroSuffix = ".ryu_tmp.nro";
+
         //http://switchbrew.org/index.php?title=Homebrew_ABI
-        public static void WriteHbAbiData(AMemory Memory, long Position, int MainThreadHandle)
+        public static void WriteHbAbiData(AMemory Memory, long Position, int MainThreadHandle, string SwitchPath)
         {
             Memory.Manager.Map(Position, AMemoryMgr.PageSize, (int)MemoryType.Normal, AMemoryPerm.RW);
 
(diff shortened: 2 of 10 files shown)


Ryujinx.HLE/Loaders/Executable.cs
```diff
@@ -14,6 +14,7 @@ class Executable
         public IReadOnlyDictionary<long, string> SymbolTable => m_SymbolTable;
 
         public string Name { get; private set; }
+        public string FilePath { get; private set; }
```
gdkchan: IMO if you need to store the name and the file path, just store the file path, and then extract the name from the file path when needed, its not necessary to store both.

Ryujinx.HLE/OsHle/Homebrew.cs
```diff
@@ -1,9 +1,12 @@
 ﻿using ChocolArm64.Memory;
+using System.Text;
 
 namespace Ryujinx.HLE.OsHle
 {
     static class Homebrew
     {
+        public static readonly string FakeFilePath = "/ryu/homebrew.nro";
```
gdkchan: imo adding `Fake` to var names doesn't sound good. And why is this using a hardcoded path? AFAIK argv should contain the real path to the application on the sdcard. You can build the virtual path from FilePath on the executable and use that.

Ryujinx.HLE/OsHle/Services/FspSrv/HomebrewSdFileSystem.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+using static Ryujinx.HLE.OsHle.ErrorCode;
+
+namespace Ryujinx.HLE.OsHle.Services.FspSrv
+{
+    class HomebrewSdFileSystem : IFileSystem
```
gdkchan: Why is this extra class necessary?

Ryujinx.HLE/OsHle/Services/FspSrv/IFileSystem.cs
```diff
@@ -244,6 +244,11 @@ public long GetEntryType(ServiceCtx Context)
             return 0;
         }
 
+        protected virtual string GetFullPath(ServiceCtx Context, string BasePath, string FileName)
```
gdkchan: Why does this need to be `protected virtual`? Could just be `private static`. Also I'd recommend moving it to the end of the file, and it seems to be in a random place with method before and after this method using it.

Ryujinx.HLE/OsHle/Services/FspSrv/IFileSystemProxy.cs
```diff
@@ -31,7 +31,14 @@ public long SetCurrentProcess(ServiceCtx Context)
 
         public long OpenSdCardFileSystem(ServiceCtx Context)
         {
-            MakeObject(Context, new IFileSystem(Context.Ns.VFs.GetSdCardPath()));
+            if (Context.Process.NeedsHbAbi)
```
gdkchan: This special casing here shouldn't be necessary. We want to have as little special treatment for homebrew as possible on the codebase.

Ryujinx.HLE/OsHle/Horizon.cs
```diff
@@ -87,19 +87,40 @@ void LoadNso(string FileName)
...
+                string TempPath = Ns.VFs.SwitchPathToSystemPath(SwitchPath);
+
+                string SwitchDir = Path.GetDirectoryName(TempPath);
+                if (!Directory.Exists(SwitchDir))
+                {
+                    Directory.CreateDirectory(SwitchDir);
+                }
+                if (File.Exists(TempPath))
+                {
+                    File.Delete(TempPath);
```
gdkchan: Instead of deleting the file, you could remove that part and use `File.Copy` with the overwrite arg set to true.

Ryujinx.HLE/OsHle/Homebrew.cs
```diff
@@ -1,11 +1,14 @@
 ﻿using ChocolArm64.Memory;
+using System.Text;
 
 namespace Ryujinx.HLE.OsHle
 {
     static class Homebrew
     {
+        public static readonly string TemporaryNroSuffix = ".ryu_tmp.nro";
```
gdkchan: This can be just `public const string`.

Thog:
Should be good now ^^

AcK77:
Seems good to me ;)

PR #273, opened 2018-07-15
Zero out bits 63:32 of scalar float operations with SSE intrinsics

gdkchan:
On ARM, all scalar operations will set the unused bits to zero. So on a float scalar operation, bits 31:0 are set to the scalar result, and bits 127:32 are set to 0. On x86 however this doesn't happen, instead it uses the value at the first operand. This was not taken into account when using the SSE intrinsics, and only the zero upper method was being called, which means that only bits 127:64 were zeroed, and bits 63:32 could still contain "garbage". This introduces a new method `VectorZero32_128` that is used to zero out all needed bits for float operations, and for double operations `VectorZeroUpper` is used.

It's unlikely that applications were affected by this unless it accesses the upper bits for some reason. This still needs testing (althrough I ran some games with it and didn't see any regressions).

### DIFF
--- a/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
+++ b/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
@@ -234,7 +234,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse && AOptimizations.UseSse2)
             {
-                EmitSseOrSse2CallF(Context, nameof(Sse.AddScalar));
+                EmitScalarSseOrSse2CallF(Context, nameof(Sse.AddScalar));
             }
             else
             {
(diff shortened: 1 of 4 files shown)


gdkchan:
Heres a test that shows the issue:
```csharp
        [TestCase(
            0x0badc0de3f800000UL, 0x0badc0deaabbccffUL,
            0x0badc0de40000000UL, 0x0badc0decafecafeUL,
            0x0000000040400000UL, 0x0000000000000000UL)]
        public void Fadd_S(ulong A, ulong B, ulong C, ulong D, ulong Result0, ulong Result1)
        {
            uint Opcode = 0x1e222820;
            Vector128<float> V0 = MakeVectorE0E1(0xcafecafecafecafe, 0xbadc0de0badc0de0UL);
            Vector128<float> V1 = MakeVectorE0E1(A, B);
            Vector128<float> V2 = MakeVectorE0E1(C, D);
            AThreadState ThreadState = SingleOpcode(Opcode, V0: V0, V1: V1, V2: V2);
            Assert.Multiple(() =>
            {
                Assert.AreEqual(Result0, GetVectorE0(ThreadState.V0));
                Assert.AreEqual(Result1, GetVectorE1(ThreadState.V0));
            });
        }
```
Fails on (current) master, passes with this PR.

PR #275, opened 2018-07-16
Support deswizzle of sparse tiled textures and some frame buffer fixes

gdkchan:
I plan to improve a bit the mess that is the texture copy engine aswell. Note that this "fix" is just a guess so I'm not sure if it works with all games or breaks something, so testing is aways welcome.


ReaperOfSouls1909:
was not sure what to test so i tested all working games i see nothing broken and no new changes so if you have anything i should look for i will be happy to retest

gdkchan:
Should be ready now (a lot of stuff isn't supported on the copy engine still, but it is very complex and will take a lot of time to implement fully).

PR #277, opened 2018-07-17
Added appveyor configuration file

emmauss:
It also add crosscompile for linux and osx. window rid has been changed to win-x64, because we don't only support win10 but any window OS that supports NetCore.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,10 @@
 version: 1.0.{build}
-branches:
-  only:
-  - master
+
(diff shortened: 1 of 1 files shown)


gdkchan:
Is downloading the 2.1.0 RC1 package before build still necessary, or does AppVeyor already have the 2.1.0 installed? This was necessary when it was still in release candidate but now maybe it's not needed anymore.

emmauss:
Merging this now. Lets hope appveyor does not break.

PR #278, opened 2018-07-17
SQADD/UQADD instructions

MS-DOS1999:


gdkchan:
The implementation looks good, but imo it would be better to modify `EmitSaturatingNarrowOp` to also work without narrowing and use it for those instructions, to avoid introducing yet another method on `AInstEmitSimdHelper`.

For example, I recently modified the [EmitVectorShImmOp](https://github.com/Ryujinx/Ryujinx/blob/master/ChocolArm64/Instruction/AInstEmitSimdShift.cs#L309) method to also optionally add a round constant, by adding flags that controls what the method emits through a enum. I think something similar could be done there.

ChocolArm64/AOpCodeTable.cs
```diff
@@ -373,6 +373,8 @@ static AOpCodeTable()
             SetA64("0x001110<<1xxxxx100000xxxxxxxxxx", AInstEmit.Smlal_V,       typeof(AOpCodeSimdReg));
             SetA64("0x001110<<1xxxxx101000xxxxxxxxxx", AInstEmit.Smlsl_V,       typeof(AOpCodeSimdReg));
             SetA64("0x001110<<1xxxxx110000xxxxxxxxxx", AInstEmit.Smull_V,       typeof(AOpCodeSimdReg));
+            SetA64("01011110xx1xxxxx000011xxxxxxxxxx", AInstEmit.Sqadd_S,       typeof(AOpCodeSimdReg));
+            SetA64("0x001110xx1xxxxx000011xxxxxxxxxx", AInstEmit.Sqadd_V,       typeof(AOpCodeSimdReg));
```
LDj3SNuD: `"0>001110<<1xxxxx000011xxxxxxxxxx"`

ChocolArm64/AOpCodeTable.cs
```diff
@@ -421,6 +423,8 @@ static AOpCodeTable()
             SetA64("0x101110<<1xxxxx101011xxxxxxxxxx", AInstEmit.Uminp_V,       typeof(AOpCodeSimdReg));
             SetA64("0x001110000xxxxx001111xxxxxxxxxx", AInstEmit.Umov_S,        typeof(AOpCodeSimdIns));
             SetA64("0x101110<<1xxxxx110000xxxxxxxxxx", AInstEmit.Umull_V,       typeof(AOpCodeSimdReg));
+            SetA64("01111110xx1xxxxx000011xxxxxxxxxx", AInstEmit.Uqadd_S,       typeof(AOpCodeSimdReg));
+            SetA64("0x101110xx1xxxxx000011xxxxxxxxxx", AInstEmit.Uqadd_V,       typeof(AOpCodeSimdReg));
```
LDj3SNuD: `"0>101110<<1xxxxx000011xxxxxxxxxx"`

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -773,6 +773,115 @@ public static void EmitVectorPairwiseOp(AILEmitterCtx Context, Action Emit, bool
...
+        public static void EmitSaturatingOp(
+            AILEmitterCtx Context,
+            Action        Emit,
+            bool          SignedSrc,
+            bool          SignedDst,
+            bool          Scalar)
+        {
+            AOpCodeSimdReg Op = (AOpCodeSimdReg)Context.CurrOp;
+
+            int Elems = !Scalar ? 8 >> Op.Size : 1;
```
LDj3SNuD:
```
int Bytes = Op.GetBitsCount() >> 3;
int Elems = !Scalar ? Bytes >> Op.Size : 1;
```

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -773,6 +773,115 @@ public static void EmitVectorPairwiseOp(AILEmitterCtx Context, Action Emit, bool
...
+
+            int ESize = 8 << Op.Size;
+
+            long TMaxValue = SignedDst ? (1 << (ESize - 1)) - 1 : (1L << ESize) - 1L;
```
LDj3SNuD: `long TMaxValue = SignedDst ? (1 << (ESize - 1)) - 1 : (long)(ulong.MaxValue >> (64 - ESize));`

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -773,6 +773,115 @@ public static void EmitVectorPairwiseOp(AILEmitterCtx Context, Action Emit, bool
...
+
+            Context.EmitLdc_I8(0L);
+            Context.EmitSttmp();
+
+            for (int Index = 0; Index < Elems; Index++)
+            {
+                AILLabel LblLe    = new AILLabel();
+                AILLabel LblGeEnd = new AILLabel();
+
+                EmitVectorExtract(Context, Op.Rn, Index, Op.Size + 1, SignedSrc);
```
LDj3SNuD: `EmitVectorExtract(Context, Op.Rn, Index, Op.Size, SignedSrc);`

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -773,6 +773,115 @@ public static void EmitVectorPairwiseOp(AILEmitterCtx Context, Action Emit, bool
...
+                EmitVectorExtract(Context, Op.Rm, Index, Op.Size + 1, SignedSrc);
```
LDj3SNuD: `EmitVectorExtract(Context, Op.Rm, Index, Op.Size, SignedSrc);`

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -773,6 +773,115 @@ public static void EmitVectorPairwiseOp(AILEmitterCtx Context, Action Emit, bool
...
+                {
+                    EmitVectorZeroLower(Context, Op.Rd);
+                }
+
+                EmitVectorInsertTmp(Context, Index, Op.Size);
+            }
+
+            Context.EmitLdvectmp();
+            Context.EmitStvec(Op.Rd);
+
```
LDj3SNuD:
```

if ((Op.RegisterSize == ARegisterSize.SIMD64) || Scalar)
{
    EmitVectorZeroUpper(Context, Op.Rd);
}

```

MS-DOS1999:
@LDj3SNuD Thank you for the correction, I will change all that!

Just can you explain to me the meaning of "<" ">"
in "0> 001110 << 1xxxxx000011xxxxxxxxxx"
Gdkchan had explained to me a long time ago, but I forgot. Thank you !

mailwl:
From source comment:

Note: < and > are used on special encodings.
The < means that we should never have ALL bits with the '<' set.
So, when the encoding has <<, it means that 00, 01, and 10 are valid,
but not 11. <<< is 000, 001, ..., 110 but NOT 111, and so on...
For >, the invalid value is zero. So, for >> 01, 10 and 11 are valid,
but 00 isn't.

MS-DOS1999:
@gdkchan I need to make an enum too? Or just a boolean for Narrow / notNarrow?

gdkchan:
IMO this method already has too many bools (SignedSrc, SignedDst, Scalar), and plus another one for Narrow, I think its way more readable if you use an enum for all those flags like I did on the inst I linked.

MS-DOS1999:
`int Bytes = Op.GetBitsCount() >> 3;
int Elems = !Scalar ? Bytes >> Op.Size : 1;`

`long TMaxValue = SignedDst ? (1 << (ESize - 1)) - 1 : (long)(~0UL >> (64 - ESize));`

`EmitVectorExtract(Context, Op.Rn, Index, Op.Size, SignedSrc);`

`EmitVectorExtract(Context, Op.Rm, Index, Op.Size, SignedSrc);`

I tested this and several instructions do not work anymore with these changes, so I did not make these changes.

gdkchan:
I think you should use `Op.Size` instead of `Op.Size + 1` when Narrow is false,

gdkchan:
Closing in favor of #314.

PR #279, opened 2018-07-17
Stub AppletOE UnlockExit

firewolfgames:


Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -36,6 +37,11 @@ public long LockExit(ServiceCtx Context)
             return 0;
         }
 
+        public long UnlockExit(ServiceCtx Context)
+        {
+            return 0;
```
AcK77: You can't just stub it without doing nothing here... I don't know what others dev thinks here... But we need to close the applet in the right way to avoid any game errors.

gdkchan: I'm not sure about what those functions does, but going by the name I guess they just prevents/enables exiting from the application maybe. If that's the case then it's fine to "stub" it as we can't do something like that on the emulator.

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -36,6 +37,11 @@ public long LockExit(ServiceCtx Context)
             return 0;
```
AcK77: If we stub the UnlockExit. Don't forget to add a Logging print here too.

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
firewolfgames: So we have 2 options:
1. Add a bool (IsExitLocked), make the functions toggle it and stub `Exit`
2. Or stub Exit, LockExit and UnlockExit all together

Imo I don't think we need it, like gdk said. So if we're never going to implement it there's no point in logging the stub as well.

AcK77: And why don't add the `IsExitLocked` bool and stub them plus add a stubbed logging as some others method? The point here is to know there is stub if some games execution is broken.

firewolfgames:
Updated

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -31,8 +35,33 @@ public ISelfController()
             LaunchableEvent = new KEvent();
         }
 
+        public long Exit(ServiceCtx Context)
+        {
+            if(IsExitLocked == false)
```
AcK77: You can remove all conditions here and just wrote `Context.Ns.Log.PrintStub(LogClass.ServiceAm, $"Stubbed. Applet requested exit (IsExitLocked = {IsExitLocked})");`

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -31,8 +35,33 @@ public ISelfController()
...
+                Context.Ns.Log.PrintInfo("Applet requested exit, but not allowed (IsExitLocked = true)");
+            }
+        
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
+            return 0;
+        }
+
         public long LockExit(ServiceCtx Context)
         {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
```
AcK77:
```
IsExitLocked = true;

Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
```

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -31,8 +35,33 @@ public ISelfController()
...
+            IsExitLocked = true;
+            return 0;
+        }
+
+        public long UnlockExit(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
```
AcK77:
```
IsExitLocked = false;

Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");
```

firewolfgames:
Done

AcK77:
Not really since you forget all the extra lines between `IsExitLocked` assignment, `PrintStub` and `return 0;`

AcK77:
I have added the `const` by mistake when I make it `private`... So the bool is not needed xD... Stub are always annoying!

Ryujinx.HLE/OsHle/Services/Am/ISelfController.cs
```diff
@@ -31,8 +35,28 @@ public ISelfController()
             LaunchableEvent = new KEvent();
         }
 
+        public long Exit(ServiceCtx Context)
+        {
+            Context.Ns.Log.PrintStub(LogClass.ServiceAm, $"Stubbed. Applet requested exit (IsExitLocked = {IsExitLocked})");
```
gdkchan: Change that to just `Context.Ns.Log.PrintStub(LogClass.ServiceAm, "Stubbed.");` after removing IsExitLocked.

AcK77:
I have done it.

PR #280, opened 2018-07-17
Implement Geometry shaders

ReinUsesLisp:
This PR makes various assumptions in how vertex index are handled in geometry shaders by Maxwell. It stubs/half-implements shader instructions to get the expected vertex index, resulting in a weak implementation. I couldn't find a game actively using a geometry shader to see if the result is as expected, but generated code is valid GLSL.

Maxwell shader binaries do not include primitive input type, so these are stubbed to expect triangle input data.

It uses ARB_enhanced_layouts to set attribute locations (an OpenGL 4.4 extension), so geometry shaders are disabled on drivers without that extension. I tried using reference-by-name but drivers seem to have some differences in how they handle them, specially when a vertex shader does not declare that vaying.

It also changes how shaders are dumped, generating two folders: `Code` and `Full`. `Code` is `nvdisasm` target binaries, `Full` is for `Ryujinx.ShaderTools`. The difference is that the later includes 0x50 bytes of header.

@Subv FYI the assumption to get vertex index is the following:
1. S2R sets a Maxwell-specific magic number to a register, this is stubbed to zero.
2. VMAD does some math multiplying S2R's result it with itself and adding the vertex index (e.g. 1). Because the number is zero, it doesn't matter that it gets multiplied.
3. ISBERD translates VMAD's result into an "accessor" for ALD, this is stubbed to copy registers.
4. ALD is used with ISBERD's result.

There is also VSETP, but this one is not currently implemented. It seems to set a predicate that's then used to set (conditionally) zero to the register that has ISBERD's value. I don't know how this might be used.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
@@ -11,6 +11,7 @@ namespace Ryujinx.Graphics.Gal.Shader
         public const int VertexIdAttr    = 0x2fc;
         public const int FaceAttr        = 0x3fc;
         public const int GlPositionWAttr = 0x7c;
+        public const int LayerAttr       = 0x64;
 
         public const int MaxUboSize = 1024;
 
@@ -210,7 +211,8 @@ namespace Ryujinx.Graphics.Gal.Shader
                     //This is a built-in input variable.
                     if (Abuf.Offs == VertexIdAttr ||
                         Abuf.Offs == InstanceIdAttr ||
-                        Abuf.Offs == FaceAttr)
+                        Abuf.Offs == FaceAttr ||
+                        Abuf.Offs == LayerAttr)
                     {
                         break;
                     }
@@ -254,6 +256,8 @@ namespace Ryujinx.Graphics.Gal.Shader
 
                         m_Attributes.Add(Index, DeclInfo);
                     }
+
+                    Traverse(Abuf, Abuf.Vertex);
                     break;
                 }
 
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
@@ -4,14 +4,13 @@ namespace Ryujinx.Graphics.Gal.Shader
 {
     class GlslDecl
     {
+        public const int LayerAttr       = 0x064;
         public const int TessCoordAttrX  = 0x2f0;
         public const int TessCoordAttrY  = 0x2f4;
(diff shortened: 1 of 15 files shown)


Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
```diff
@@ -11,6 +11,7 @@ class GlslDecl
         public const int VertexIdAttr    = 0x2fc;
         public const int FaceAttr        = 0x3fc;
         public const int GlPositionWAttr = 0x7c;
```
gdkchan: I think this is no longer used and can be removed (I know its not part of this PR but I think its worth to remove it since you're touching this file anyway).

Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
```diff
@@ -11,6 +11,7 @@ class GlslDecl
...
+        public const int LayerAttr       = 0x64;
```
gdkchan: IMO its better to put it above `TessCoordAttrX` so that they are sorted by value.

Ryujinx.Graphics/Gal/Shader/ShaderHeader.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public ShaderHeader(IGalMemory Memory, long Position)
+        {
+            uint CommonWord0 = (uint)Memory.ReadInt32(Position + 0);
+            uint CommonWord1 = (uint)Memory.ReadInt32(Position + 4);
+            uint CommonWord2 = (uint)Memory.ReadInt32(Position + 8);
+            uint CommonWord3 = (uint)Memory.ReadInt32(Position + 12);
+            uint CommonWord4 = (uint)Memory.ReadInt32(Position + 16);
+
+            int  SphType         = ReadBits(CommonWord0,  0, 5);
```
gdkchan: Since the entire header is being read anyway, why not exposing all information through properties on this class? (Instead of just reading it to local vars that are not used).

Ryujinx.Graphics/Gal/Shader/ShaderDecodeSpecial.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using static Ryujinx.Graphics.Gal.Shader.ShaderDecodeHelper;
+
+namespace Ryujinx.Graphics.Gal.Shader
+{
+    static partial class ShaderDecode
+    {
+        public static void Out_R(ShaderIrBlock Block, long OpCode)
+        {
+            //TODO: This registers have to be used for something
```
gdkchan: nit: `This` -> `Those`.

Issue #282, opened 2018-07-18
Optimize AOpCodeTable.GetInst


gdkchan:
Currently, it does a linear search and in the worst case, it will need to compare the opcode value with the value on all the (currently) 837 entries on the opcode table, which is done using a linked list. This number will keep increasing as new instructions are added and it is only going to get slower, so ideally we should find a more efficient way to find instructions on the table from the opcode value.

strubelz:
I have an idea about some kind of search tree structure, I'll check it out tommorow. And I think it should be faster to read from from an ArrayList or Array.

strubelz:
Hmm, I can't think of any solution using a hash table. The entire point of a hash is that is changes completly when one bit of the source is changed. For me it seems pretty hard to do some kind of pattern matching with that, but you guys know more than me.

emmauss:
well, i thought that with some kind of sorted list or dictionary, we could do a binary search or a hash lookup(dictionary uses those, i think). though this is feasible if the expected value of an instruction does not change, and set is not called after the table has been populated.

mailwl:
just an idea - in most cases, only 22bits of opcode is mean for differ operation, so, it's possibly use array of 2^22 elements (32mb?) to find needed opcode in one step, something like
`instr[opcode & 0x003FFFFF]`

MerryMage:
I agree with mailwl, though personally I think only 8 bits are necessary (256 entries for ~800 instruction types), and preferably the most-significant bits.

mailwl:
@MerryMage made good solution, i think thre is no need further optimizations (at least, before the actuil time used for fetching instructions in whole jit compiling)

PR #284, opened 2018-07-18
AOpCodeTable: Speed up instruction decoding

MerryMage:
Ref #282.

We do this by generating a lookup table.

`ToFastLookupIndex` masks the bit `00111111110000000011110000000000` and compresses them to an index. This was chosen so that each bucket has a small number of possible instructions to linearly scan.

Arrays were used in the lookup table instead of linked lists because arrays generally have better memory characteristics (linear scanning is more performant).

### DIFF
--- a/ChocolArm64/AOpCodeTable.cs
+++ b/ChocolArm64/AOpCodeTable.cs
@@ -4,6 +4,7 @@ using ChocolArm64.Instruction;
 using ChocolArm64.Instruction32;
 using ChocolArm64.State;
 using System;
+using System.Collections.Generic;
 
 namespace ChocolArm64
 {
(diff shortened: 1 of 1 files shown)


ChocolArm64/AOpCodeTable.cs
```diff
@@ -533,55 +567,48 @@ private static void Set(string Encoding, AInst Inst, AExecutionMode Mode)
...
         }
 
         public static AInst GetInstA64(int OpCode)
         {
-            return GetInst(InstHeadA64, OpCode);
+            return GetInstFromList(InstA64FastLookup[ToFastLookupIndex(OpCode)], OpCode);
         }
 
-        private static AInst GetInst(TreeNode Head, int OpCode)
+        private static AInst GetInstFromList<T>(T InstList, int OpCode) where T : IEnumerable<InstInfo>
```
gdkchan: Any reason for using a generic type here instead of using something like `IEnumerable<InstInfo>` directly?

ChocolArm64/AOpCodeTable.cs
```diff
@@ -436,27 +437,60 @@ static AOpCodeTable()
...
         }
 
-        private static TreeNode InstHeadA32;
-        private static TreeNode InstHeadA64;
+        private static List<InstInfo> AllInstA32 = new List<InstInfo>();
+        private static List<InstInfo> AllInstA64 = new List<InstInfo>();
+
+        private static int FastLookupSize = 0x1000;
+        private static InstInfo[][] InstA64FastLookup;
+        private static int ToFastLookupIndex(int Value)
```
gdkchan: IMO a better location for this method is below GetInstA64 (where it is used).

ChocolArm64/AOpCodeTable.cs
MerryMage: I was under the mistaken impression that it would be more performant.

ChocolArm64/AOpCodeTable.cs
```diff
@@ -436,27 +437,60 @@ static AOpCodeTable()
...
+                for (int i = 0; i < FastLookupSize; i++)
+                {
+                    if ((i & Mask) == Value)
+                    {
+                        Tmp[i].Add(Inst);
+                    }
+                }
+            }
+
+            InstA64FastLookup = new InstInfo[FastLookupSize][];
```
mailwl: maybe better move assigment to initializer? in any way it's static constructor, and memory will be allocated at class loading

MerryMage:
Updated as requested

PR #285, opened 2018-07-18
Send data to OpenGL host without client-side copies

ReinUsesLisp:
Currently there are three copies being made before data is sent to the GPU:
1. Vmm.ReadBytes copies to a temporary byte[]
2. Gal copies from that byte[] into a buffer in OGLStreamBuffer (cbuf-only step)
3. Buffer is then sent to the OpenGL server
4. The OpenGL driver probably copies it to a temporary address and then sends it to the GPU

Instead of that, this PR gets host memory address and sends that to the OpenGL server (skipping 1 and 2).

### DIFF
--- a/ChocolArm64/Memory/AMemory.cs
+++ b/ChocolArm64/Memory/AMemory.cs
@@ -204,6 +204,11 @@ namespace ChocolArm64.Memory
             return Modified;
         }
 
+        public IntPtr GetHostAddress(long Position)
+        {
+            return (IntPtr)(RamPtr + (ulong)Position);
+        }
+
         public sbyte ReadSByte(long Position)
         {
             return (sbyte)ReadByte(Position);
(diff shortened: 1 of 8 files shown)


ChocolArm64/Memory/AMemory.cs
```diff
@@ -204,6 +204,11 @@ public bool[] IsRegionModified(long Position, long Size)
             return Modified;
         }
 
+        public IntPtr GetHostAddress(long Position)
+        {
+            return (IntPtr)(RamPtr + (ulong)Position);
```
gdkchan: I think that before returning this pointer, at least a check to ensure that the memory range is valid should be made. This can be done with `EnsureRangeIsValid` (a size parameter would need to be passed aswell). As it is now, its possible for this method to return invalid addresses (if `Position` > the ram size for example (4gb)), or if `Position + Size` > 4gb. Both would most likely result in an access violation/crash. The `EnsureRangeIsValid` should be pretty cheap so I don't expect it to decrease perf.

PR #289, opened 2018-07-19
Implement SvcWaitForAddress 0x34

AcK77:
Currently needed by Sonic Mania Plus

### DIFF
--- a/Ryujinx.HLE/OsHle/Kernel/AddressArbiter.cs
+++ b/Ryujinx.HLE/OsHle/Kernel/AddressArbiter.cs
@@ -1,0 +1,94 @@
+﻿using ChocolArm64.Memory;
+using ChocolArm64.State;
+using Ryujinx.HLE.OsHle.Handles;
+
+using static Ryujinx.HLE.OsHle.ErrorCode;
+
+namespace Ryujinx.HLE.OsHle.Kernel
+{
+    static class AddressArbiter
+    {
+        static ulong WaitForAddress(Process Process, AThreadState ThreadState, long Address, ulong Timeout)
+        {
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/OsHle/Kernel/AddressArbiter.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public static ulong WaitForAddressIfLessThan(Process      Process, 
+                                                     AThreadState ThreadState, 
+                                                     AMemory      Memory, 
+                                                     long         Address, 
+                                                     int          Value, 
+                                                     ulong        Timeout, 
+                                                     bool         ShouldDecrement)
+        {
+            int CurrentValue = Memory.ReadInt32(Address);
```
gdkchan: The read-decrement-write operation is not thread safe, so it should be under a lock. You can do this:
```
Memory.SetExclusive(ThreadState, Address);

int CurrentValue = Memory.ReadInt32(Address);

while (true)
{
    if (Memory.TestExclusive(ThreadState, Address))
    {
        if (CurrentValue < Value)
        {
            if (ShouldDecrement)
            {
                Memory.WriteInt32(Address, CurrentValue - 1);
            }
            
            Memory.ClearExclusiveForStore(ThreadState);
        }
        else
        {
            Memory.ClearExclusiveForStore(ThreadState);
            
            return MakeError(ErrorModule.Kernel, KernelErr.InvalidState);
        }

        break;
    }

    Memory.SetExclusive(ThreadState, Address);

    CurrentValue = Memory.ReadInt32(Address);
}
```

PR #293, opened 2018-07-19
Fix SystemPathToSwitchPath platform issues and make sure to delete temporary NRO after sessions dispose

Thog:
This should fix issues on Windows of #272


Ryujinx.HLE/OsHle/Process.cs
```diff
@@ -431,6 +426,11 @@ protected virtual void Dispose(bool Disposing)
                     }
                 }
 
+                if (NeedsHbAbi && Executables[0].FilePath.EndsWith(Homebrew.TemporaryNroSuffix))
```
gdkchan: This will fail is Executables.Count == 0.

PR #295, opened 2018-07-19
Avoid calling buffer binding when shader didn't change

ReinUsesLisp:
It should give a small performance boost on some games without breaking Tiny Nightmare.


gdkchan:
Hmm, I don't understand why the glUseProgram can't be inside the if aswell, since its essentially the same program... But afaik it breaks tiny *nightmare*? Anyway, this change lgtm.

PR #296, opened 2018-07-19
More accurate impl of FMINNM/FMAXNM, add vector variants

gdkchan:
This PR:
- Implements FMAXNM/FMINNM better, previously it did the same thing as FMAX/FMIN, now it should return the first non-NaN value if only one of the operands is a NaN, while the other one would return the first NaN value.
- Handle NaN inputs correctly for FMAX/FMIN.
- Fix encodings for vector variants of FMAX/FMIN, they would allow some invalid encodings.
- Adds the vector variants of FMAXNM/FMINNM.

I implemented it on `ASoftFloat` as it seemed to be the most fiting place for this sort of method. PTAL @MerryMage, @LDj3SNuD.


MerryMage:
Softfloat code LGTM

gdkchan:
After some fixes all tests pass, including the NaN tests that I re-enabled. Should be good now.

LDj3SNuD:
[What do you think about this solution?](https://pastebin.com/SeNd97R7)

Avoid those tedious conversions.
Introduces the shareable method ProcessNaNs().
(Passes the current tests.)

gdkchan:
Sorry for the delay. Implemented suggestions made by @LDj3SNuD. I changed `ProcessNaNs` to return the bool and the result on the out argument (as this is the pattern used on the .net framework and more intuitive). I also optimized the methods for the most common cases (values are not NaN, and values are not zero). All tests are still passing.

LDj3SNuD:
lgtm, maybe all those "x" could be "op_bits": https://github.com/Ryujinx/Ryujinx/blob/c9cd32602e7c9182d2baf8b6b4250902825cb32f/ChocolArm64/Instruction/ASoftFloat.cs#L513-L534

post-pr: now that the class will also be used for other float implementations, it would be useful if those tables were built only if and when their instructions require it.

PR #305, opened 2018-07-26
update encoding for branch instruction

arthurchan35:


gdkchan:
lgtm, thanks for fixing those encodings!
Those are probably leftovers from the times where the instruction table didn't have all the encoding bits, and the instruction using more bits were special cased. I probably forgot to add those bits when I expanded the instruction table.

PR #306, opened 2018-07-26
Implement AcquireNpadStyleSetUpdate event

emmauss:
I don't know where this is set, so I implemented it just like the VSync event. It is probably set at the NPadStyleSet methods, but not much is known.

### DIFF
--- a/Ryujinx.HLE/OsHle/Horizon.cs
+++ b/Ryujinx.HLE/OsHle/Horizon.cs
@@ -25,7 +25,8 @@ namespace Ryujinx.HLE.OsHle
         internal HSharedMem HidSharedMem  { get; private set; }
         internal HSharedMem FontSharedMem { get; private set; }
 
-        internal KEvent VsyncEvent { get; private set; }
+        internal KEvent VsyncEvent              { get; private set; }
+        internal KEvent NpadStyleSetUpdateEvent { get; private set; }
 
         public Horizon(Switch Ns)
         {
@@ -42,7 +43,8 @@ namespace Ryujinx.HLE.OsHle
             HidSharedMem  = new HSharedMem();
             FontSharedMem = new HSharedMem();
 
-            VsyncEvent = new KEvent();
+            VsyncEvent              = new KEvent();
+            NpadStyleSetUpdateEvent = new KEvent();
         }
 
         public void LoadCart(string ExeFsDir, string RomFsFile = null)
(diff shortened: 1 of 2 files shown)


gdkchan:
It is set (event fires) when the `NpadStyleSet` changes. The only effect of not signaling it is that the game will think that it is never changed, which should be fine.

emmauss:
typo in name. sorry

Ryujinx.HLE/OsHle/Horizon.cs
```diff
@@ -124,6 +126,8 @@ public void LoadProgram(string FilePath)
 
         public void SignalVsync() => VsyncEvent.WaitEvent.Set();
 
+        public void SignalNpadStyleSetUpdate() => NpadStyleSetUpdateEvent.WaitEvent.Set();
```
gdkchan: imo this is not necessary (we don't want to expose a method to signal that, it should be signaled internally when it makes sense to do so -- when the "NpadStyleSet" value is changed).

gdkchan: the vsync event is a special case because rendering is not Ryujinx.HLE responsability, its the GUI responsability, to the GUI is also responsible for signaling when vsync occurs for the core.

emmauss: okay, i will move the signal to the his service

Ryujinx.HLE/OsHle/Horizon.cs
```diff
@@ -42,7 +42,7 @@ public Horizon(Switch Ns)
             HidSharedMem  = new HSharedMem();
             FontSharedMem = new HSharedMem();
 
-            VsyncEvent = new KEvent();
+            VsyncEvent              = new KEvent();
```
gdkchan: Why the spacing?

Ryujinx.HLE/OsHle/Services/Hid/IHidServer.cs
```diff
@@ -9,6 +10,8 @@ class IHidServer : IpcService
     {
         private Dictionary<int, ServiceProcessRequest> m_Commands;
 
+        private KEvent NpadStyleSetUpdateEvent { get; set; }
```
gdkchan: What is the point of making this a property when it is private? Could just use a field.

PR #307, opened 2018-07-27
More flexible memory manager

gdkchan:
This has the following improvements:
- Real address translation.
-- We only need to write shared memory data to a single location now.
-- Can have the entire address space available like on the real system.
-- Pre-requisite for multiple processes support.
- Better implementation of the memory svcs.
- Added a memory allocator.
-- No need to use hardcoded addresses anymore.
- GetWriteWatch is no longer necessary.
-- Better performance on OSes other than Windows.

And the following regressions:
- Loss of performance as all memory accesses also needs an address translation.

This is still WIP, and a few bits I plan to implement aren't done yet. Those include the Map/UmapPhysicalMemory svcs and address space mapping more like the real OS.

It also needs the shared font support that @Thog started implementing but its still missing the libre fonts. This is necessary because games will otherwise try to access unmapped memory without it, and as I plan to leave those checks on by default, it is necessary to avoid crashes.

I'm mostly interested in hearing about possible perf regressions. They are expected, but hopefully the drop will be acceptable and worth it. If its too much (like less than half of the speed compared to master) then we can reconsider keeping the old approach, but I really don't want that.


SexyBabes:
FREE SEXY CAM GIRLS
**18+ ONLY**
GO TO http://sexygirls.com/

gdkchan:
Should be pretty much done I guess.
The HLE project is messy right now and I plan to do at least part of the cleanup in another PR. This is already changing too much as is.
It would be nice to test it for:
- Regressions (games that worked before but don't with this PR).
- Performance drop -- this is expected but the diff shouldn't too high.
- Performance boost (on linux).

gdkchan:
I don't plan to do any more changes on this other than something I may have forgot, but it should be pretty much done. It would be nice to get another round of testing (looking for the same things I said above).

PR #308, opened 2018-07-28
Window related changes

ReinUsesLisp:
Changes some frontend related stuff:
* Reworks input and disables unmapped controllers (that's unknown DirectInput controllers). Moves input parsing from GLScreen to a class.
* Alt+Enter or F11 maximizes the window. Pressing escape or one of those combinations again unmaximizes it, pressing escape again closes (as always).

And some other window-related stuff:
* Event polling was done without sleeping the thread, wasting CPU cycles in somethings that's not that important. It should give a performance boost for all games.
* Changes how touchscreen mouse input math is done (avoids float-point math when it's not needed).

Input issues with Cave Story+ when having a (DirectInput) controller plugged in were because triggers where signaled down, making a "moonwalk". It would be nice to have someone with an XInput controller (e.g. xbox) test the input on spacenx, specially both analog sticks.

### DIFF
--- a/Ryujinx/Config.cs
+++ b/Ryujinx/Config.cs
@@ -14,11 +14,6 @@ namespace Ryujinx
         public static JoyConKeyboard   JoyConKeyboard   { get; private set; }
         public static JoyConController JoyConController { get; private set; }
 
-        public static float GamePadDeadzone             { get; private set; }
-        public static bool  GamePadEnable               { get; private set; }
-        public static int   GamePadIndex                { get; private set; }
-        public static float GamePadTriggerThreshold     { get; private set; }
-
         public static void Read(Logger Log)
         {
             string IniFolder = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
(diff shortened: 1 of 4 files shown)


Luffykun007:
Everything seems to work fine for me using an X-Box 360 control!

capturar2

ReaperOfSouls1909:
after farther testing this makes my controllers all 4 move to the left in every game even my keyboard

ReinUsesLisp:
What do you mean with `my controllers all 4 move to the left`?

ReaperOfSouls1909:
on there own they move to the left the direction left as in < way ive tested without this pr and dont have the issue

ReinUsesLisp:
Did you calibrate your controller? Being an XInput device, there shouldn't be any difference in input layout.

ReaperOfSouls1909:
https://youtu.be/ONm0vT1-MIQ tried deleting the conf and setting up my controller end result remains the same by the way the controllers are 2 xbox 360 official and a afterglow xbox 360 purple and xbox one controller elite

ReinUsesLisp:
Sorry to bother you, could you replace GLScreen.cs code (located in `Ryujinx/Ryujinx/Ui`) with [this](https://pastebin.com/aeiY5u2i)? It will print button/axis status.

ReaperOfSouls1909:
[Ryujinx.log](https://github.com/Ryujinx/Ryujinx/files/2243772/Ryujinx.log)
no bother at all and here is the log without me touching the controller or keyboard in game

ReinUsesLisp:
Same as before, but in [JoyConController.cs](https://pastebin.com/xBGqYk4f)

ReinUsesLisp:
I had someone with an xbox 360 controller to test it and it works without issues. What I find weird is that it worked before for you and now it doesn't. If you have multiple controllers plugged in I recommend unplugging them all but one, or playing around with GamePad_Index config value.

PR #311, opened 2018-07-29
Allow loops inside basic blocks

gdkchan:
When theres a loop with a target inside the basic block, we don't need to return to the dispatcher, we can just jump to the offset inside the basic block directly.


SexyBabes:
FREE SEXY CAM GIRLS
**18+ ONLY**
GO TO http://sexygirls.com/

SexyBabes:
FREE SEXY CAM GIRLS
**18+ ONLY**
GO TO http://sexygirls.com/

gdkchan:
This PR has been replaced by #515.

PR #312, opened 2018-07-29
Add ConstBuffer cache

ReinUsesLisp:
Now that Vmm.IsRegionModified is faster ConstBuffer cache doesn't generate a regression performance-wise. For me it gives almost the exact performance, but people with integrated GPUs may benefit from it. It'd be nice if someone with that kind of GPU could test master branch and this one and compare them.


SexyBabes:
FREE SEXY CAM GIRLS
**18+ ONLY**
GO TO http://sexygirls.com/

ReinUsesLisp:
Closing in favor of #319

PR #314, opened 2018-07-30
Add SQADD, UQADD, SQSUB, UQSUB, SUQADD, USQADD, SQABS, SQNEG (Scalar, Vector) instructions; add 24 Tests. Most saturation instructions now on ASoftFallback.

LDj3SNuD:
For the 64bit variants, in addition to having to manage the saturations, you must also manage the underflows and overflows, signed and unsigned.

Add EmitScalarSaturatingNarrowOpZxSx(), EmitVectorSaturatingNarrowOpZxSx().
"Remove" HasFlag() from EmitScalarOp() and EmitScalarOpF().
Nits.

Update Tester.

```
Total number of tests: 53749. Passed: 53493. Not Passed: 0. Ignored: 120.
The tests have been successfully performed.
Test run time: 4.7515 Minutes
```

### DIFF
--- a/ChocolArm64/Instruction/AInstEmitSimdHelper.cs
+++ b/ChocolArm64/Instruction/AInstEmitSimdHelper.cs
@@ -1129,24 +1129,13 @@ namespace ChocolArm64.Instruction
                 throw new InvalidOperationException();
             }
 
-            AILLabel LblFalse = new AILLabel();
-
             EmitVectorExtract(Context, Op.Rn, Index, 3, Signed);
-            Context.Emit(OpCodes.Dup);
-
             EmitVectorExtract(Context, Op.Rm, Index, 3, Signed);
-            ASoftFallback.EmitCall(Context, Signed
-                ? nameof(ASoftFallback.BinarySignedSatQAdd_Sat)
-                : nameof(ASoftFallback.BinaryUnsignedSatQAdd_Sat));
+            Context.EmitLdarg(ATranslatedSub.StateArgIdx);
 
-            Context.Emit(OpCodes.Brfalse_S, LblFalse);
-            EmitSetSaturatedFlag(Context);
-            Context.MarkLabel(LblFalse);
-
-            EmitVectorExtract(Context, Op.Rm, Index, 3, Signed);
             ASoftFallback.EmitCall(Context, Signed
-                ? nameof(ASoftFallback.BinarySignedSatQAdd_Res)
-                : nameof(ASoftFallback.BinaryUnsignedSatQAdd_Res));
+                ? nameof(ASoftFallback.BinarySignedSatQAdd)
+                : nameof(ASoftFallback.BinaryUnsignedSatQAdd));
         }
 
         // TSrcs (64bit) == TDst (64bit); signed, unsigned.
@@ -1159,24 +1148,13 @@ namespace ChocolArm64.Instruction
                 throw new InvalidOperationException();
             }
 
-            AILLabel LblFalse = new AILLabel();
-
             EmitVectorExtract(Context, Op.Rn, Index, 3, Signed);
-            Context.Emit(OpCodes.Dup);
-
             EmitVectorExtract(Context, Op.Rm, Index, 3, Signed);
-            ASoftFallback.EmitCall(Context, Signed
-                ? nameof(ASoftFallback.BinarySignedSatQSub_Sat)
-                : nameof(ASoftFallback.BinaryUnsignedSatQSub_Sat));
+            Context.EmitLdarg(ATranslatedSub.StateArgIdx);
 
-            Context.Emit(OpCodes.Brfalse_S, LblFalse);
-            EmitSetSaturatedFlag(Context);
-            Context.MarkLabel(LblFalse);
-
-            EmitVectorExtract(Context, Op.Rm, Index, 3, Signed);
             ASoftFallback.EmitCall(Context, Signed
-                ? nameof(ASoftFallback.BinarySignedSatQSub_Res)
-                : nameof(ASoftFallback.BinaryUnsignedSatQSub_Res));
+                ? nameof(ASoftFallback.BinarySignedSatQSub)
+                : nameof(ASoftFallback.BinaryUnsignedSatQSub));
         }
 
         // TSrcs (64bit) == TDst (64bit); signed, unsigned.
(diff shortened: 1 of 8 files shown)


ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+            Context.EmitSttmp(); // Saturated = 1
+
+            Context.EmitLdc_I8(TMinValue);
+
+            Context.MarkLabel(LblGeEnd);
+        }
+
+        // TSrc (from 8bit to 64bit) == TDst (from 8bit to 64bit), signed.
+        public static void EmitUnarySignedSatQAbsOrNeg(
+            AILEmitterCtx Context,
```
gdkchan: IMO the arguments don't need to be each one on a new line here (as there are only 2 arguments here, they could be all on the same line).

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+            Context.EmitSttmp(); // Saturated = 1
+
+            Context.EmitLdc_I8(TMaxValue);
+
+            Context.MarkLabel(LblFalse);
+        }
+
+        // TSrcs (64bit) == TDst (64bit), signed or unsigned.
+        public static void EmitBinarySatQAdd(
+            AILEmitterCtx Context,
```
gdkchan: IMO the arguments don't need to be each one on a new line here (as there are only 3 arguments here, they could be all on the same line).

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
 
-                Context.Emit(OpCodes.Br_S, LblGeEnd);
+                Context.MarkLabel(LblGeEnd);
+            }
+        }
 
-                Context.MarkLabel(LblLe);
+        // TSrcs (64bit) == TDst (64bit), signed or unsigned.
+        public static void EmitBinarySatQSub(
+            AILEmitterCtx Context,
```
gdkchan: IMO the arguments don't need to be each one on a new line here (as there are only 3 arguments here, they could be all on the same line).

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+            }
+        }
 
-                if (Scalar)
-                {
-                    EmitVectorZeroLowerTmp(Context);
-                }
+        // TSrcs (64bit) == TDst (64bit), signed or unsigned.
+        public static void EmitBinarySatQAccumulate(
+            AILEmitterCtx Context,
```
gdkchan: IMO the arguments don't need to be each one on a new line here (as there are only 3 arguments here, they could be all on the same line).

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+
+                Context.Emit(OpCodes.Br_S, LblEnd);
+
+                Context.MarkLabel(LblLt);
+
+                // op1 from long.MinValue to (long)ulong.MinValue - 1L
+                // op2 from (ulong)long.MaxValue + 1UL to ulong.MaxValue
+
+                EmitVectorExtractZx(Context, Op.Rd, Index, 3);
+                Context.EmitLdc_I8(long.MaxValue); // TODO: Spostare a variabile?
```
gdkchan: Italian comment?

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1030,445 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+
+                Context.EmitLdc_I8(TMinValue);
+
+                Context.MarkLabel(LblEnd);
             }
+        }
+
+        public static void EmitUpdateFpsrQCFlag(AILEmitterCtx Context)
+        {
+            const int QCFlag = 27;
```
gdkchan: IMO `QCFlagBit` is a better name.

MerryMage:
re: unsetting saturated flag

To my knowledge, the saturated flag is only ever set; most saturated instructions do not unset the flag. An application needs to write to FPSR in order to unset it.

LDj3SNuD:
And in fact EmitUpdateFpsrQCFlag() can never unset the Fpsr QC flag, even with the "existence" of EmitResetSaturatedFlag().
https://github.com/LDj3SNuD/Ryujinx/blob/9864560c52a58eb347a0a85e62991b382f573d7c/ChocolArm64/Instruction/AInstEmitSimdHelper.cs#L1456

EmitResetSaturatedFlag() is supposed to be used before a for loop to reset the temporary variable Tmp1 (used as placeholder for the Saturated flag).
https://github.com/LDj3SNuD/Ryujinx/blob/9864560c52a58eb347a0a85e62991b382f573d7c/ChocolArm64/Instruction/AInstEmitSimdHelper.cs#L1015

SexyBabes:
FREE SEXY CAM GIRLS
**18+ ONLY**
GO TO http://sexygirls.com/

MerryMage:
@LDj3SNuD Thanks for the clarification. Apologies for misreading the implmentation.

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1027,434 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+        // TSrc (16bit, 32bit, 64bit) > TDst (8bit, 16bit, 32bit); signed, unsigned.
+        public static void EmitSatQ(
+            AILEmitterCtx Context,
+            int  SizeDst,
+            bool SignedSrc,
+            bool SignedDst)
+        {
+            if (SizeDst > 2)
+            {
+                throw new PlatformNotSupportedException();
```
gdkchan: I think the appropriate exception is `throw new ArgumentOutOfRangeException(nameof(SizeDst);`.

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -840,65 +1027,434 @@ public static void EmitVectorSaturatingNarrowOpZxZx(AILEmitterCtx Context, Actio
...
+        }
+
+        // TSrcs (64bit) == TDst (64bit); signed, unsigned.
+        public static void EmitBinarySatQAdd(AILEmitterCtx Context, int Index, bool Signed)
+        {
+            AOpCodeSimdReg Op = (AOpCodeSimdReg)Context.CurrOp;
+
+            if (Op.Size < 3)
+            {
+                throw new NotSupportedException();
```
gdkchan: We've been using InvalidOperationException for those cases.

LDj3SNuD:
ASoftFallback version. Let me know which one you prefer.
Retested with the same results.

gdkchan:
It's way easier to understand than the IL version for sure, so that's a big plus. The JIT should be smart enough to inline this method so in the end it probably even has the same perf as the IL version (and emiting the code is faster since you only have to emit a call), overall I think its a better solution.

The only thing I would change is that currently it effectively calls the method twice -- once to get the saturation flag, and another to get the saturated result. This is not very efficient as it does the same work twice. Instead I would suggest passing the `AThreadState` to the saturation methods, and then setting the flag inside those methods, and then instead of returning a tuple, you could just return the long saturated result. Passing it is very simple, you just need to use `Context.EmitLdarg(ATranslatedSub.StateArgIdx);`. Then its just a matter to add a `AThreadState State` argument on the method, and you can access the cpu/thread state from it.

LDj3SNuD:
Address PR feedback.
Update Tests.
Retested with the same results.

gdkchan:
maybe its a good idea to implement `EmitSatQ` on ASoftFallback aswell, since its the last somewhat complex saturation emitter that it still using IL.

LDj3SNuD:
Address PR feedback.
Retested with the same results.

PR #319, opened 2018-08-02
Low level graphics API prerequisites

ReinUsesLisp:
This PR does the following:
* Adds const buffer cache
* Abstracts pipeline into a structure to detect changes and cache them (this includes blending, depth/stencil testing, primitive restart, vertex assembly, const buffer bindings and others)
* Stores flip values into a buffer instead of using uniform constants
* It fixes Puyo Puyo Tetris stencil testing but that is an accident

This is to ease implementation for a low-level graphics (and compute) API. It's highly probably more changes to GPU's LLE and Gal will have to be done for that, but these were the most obvious.
I'll happily listen suggestions for other changes.


ReinUsesLisp:
Rebased.

PR #326, opened 2018-08-04
That Disagea 5 "for bad driver" shader fix

Gummygamer:


ReaperOfSouls1909:
Testing now

Gummygamer:
The story is that, even if I skipped the intro, Disagea 5 crashed, so, with the help of the main devs, I found a way to correct the shader to avoid that.

ReaperOfSouls1909:
your fix is good it causes no issues and even fixed the one piece unlimited red driver crash as well great job

Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -297,7 +297,7 @@ private void PrintDeclAttributes(IEnumerable<ShaderDeclInfo> Decls, string InOut
             {
                 if (DeclInfo.Index >= 0)
                 {
-                    SB.AppendLine("layout (location = " + DeclInfo.Index + ") " + InOut + " " + GetDecl(DeclInfo) + ";");
+                    SB.AppendLine("layout (location = " + DeclInfo.Index + ") " + InOut + " " + "vec4 " + DeclInfo.Name + ";");
```
gdkchan: `+ " " + "vec4 " +` could be just `+ " vec4 " +`

Gummygamer: updated it.

Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -418,7 +418,7 @@ private void PrintAttrToOutput(string Identation = IdentationStr)
 
                 ShaderDeclInfo DeclInfo = KV.Value;
 
-                string Swizzle = ".xyzw".Substring(0, DeclInfo.Size + 1);
+                string Swizzle = ".xyzw";
```
ReinUsesLisp: If all attributes are `vec4` swizzle is not needed

Gummygamer: Eliminated those swizzles now.

gdkchan:
I don't think that's what he meant. You don't need `.xyzw` if the type is aways a vec4, you can just assign directly.

Gummygamer:
Erased those useless ".xyzw" too, sorry for the redundant code.

ReinUsesLisp:
lgtm, but I didn't test it

Gummygamer:
It seems to be fine in my test.

ReinUsesLisp:
Just some small changes and it should be ready for me.
Remove `GetDecl` calls and replace them with `"vec4 " + DeclInfo.Name`. Also remove `ElemTypes` (because it would become unused).
I would have commented this in the code but GitHub doesn't let me comment unmodified code.

Gummygamer:
A full elimination of ElemTypes and GetDecl causes an error because the shader still needs float for one case of declaration, I'm trying to put back the necessary evaluation in the best way possible.

PR #336, opened 2018-08-09
Don't crash when fonts are not found

emmauss:


filfat:
I feel like this could cause numerous problems when an application expects the shared font(s) to be > 0 bytes. Which in turn causes exceptions because of that instead of inaccurate emulation (and some of those invalid error will most likely end up in the issue tracker)

Perhaps hide this under a console argument instead?

gdkchan:
On my map branch, it only reads the font when the game requests it already.

emmauss:
ok. then closing this.

PR #346, opened 2018-08-12
Enable use of C# 7.3

Thealexbarney:
Enable use of C# 7.3 and clean up project files a little.


Thog:
As pointed by @jduncanator, we will eventually move up to support C# 7.3 but for now I think we can close this.

PR #350, opened 2018-08-14
[WIP] Improve dma copies on the gpu

gdkchan:
Some games are not writing anything to the *SizeX/Y registers, however the X/YCount registers seems to be always written, so use that instead. Also, optimized copies when (de)swizzling is not necessary by just copying the data directly.

This improves some games but there are some still some issues that needs to be investigated so this is WIP.


gdkchan:
This has been superseded by #418.

PR #352, opened 2018-08-15
Add SHA256H, SHA256H2, SHA256SU0, SHA256SU1 instructions; add 4 Tests (closed box).

LDj3SNuD:
Test (open box): [sha256-armv8(emu)-master.zip](https://github.com/Ryujinx/Ryujinx/files/2288842/sha256-armv8.emu.-master.zip)

Update Tester.


greggameplayer:
@LDj3SNuD or we can test with a game who use these like Doom

gdkchan:
This is not a good way to test correctness because the game is not going to show error messages if the instruction is not working properly. It may still work even if the impl is not fully correct, or it may not work even if it is correct (due to a different issue).

PR #355, opened 2018-08-16
Code style fixes and nits on the HLE project

gdkchan:
As the title implies this is mostly making changes related to code style. I tried to separate each change per commit to make reviews easier.

The main change is that it now finalizes the Switch object and exits properly from the application instead of forcing it. For that I had to add a synchronization method on the CPU called on each basic block, didn't notice any performance difference.


gdkchan:
This is ready for review. There are other changes that I plan to do but I will do that later on another PR.

Ryujinx.HLE/HOS/Process.cs
```diff
@@ -57,9 +56,7 @@ class Process : IDisposable
 
         private ConcurrentDictionary<long, KThread> Threads;
 
-        private KThread MainThread;
-
-        private List<Executable> Executables;
+         private List<Executable> Executables;
```
emmauss: Not aligned well

Ryujinx/Ui/GLScreen.cs
```diff
@@ -35,13 +34,13 @@ public class GLScreen : GameWindow
...
-        public GLScreen(Switch Ns, IGalRenderer Renderer)
+        public GLScreen(Switch Device, IGalRenderer Renderer)
             : base(1280, 720,
             new GraphicsMode(), "Ryujinx", 0,
             DisplayDevice.Default, 3, 3,
             GraphicsContextFlags.ForwardCompatible)
         {
-            this.Ns       = Ns;
+            this.Device       = Device;
             this.Renderer = Renderer;
```
emmauss: not aligned

PR #358, opened 2018-08-16
Rendertarget attachments, texture and image changes

ReinUsesLisp:
It does the following:
* Shaders write to attachments according to their register, e.g.: 0-3 FragColor0, 4-7 FragColor1
* Abstracts textures into an image, that being understood like an array of data interpreted as pixels
* Textures and rendertargets share the same "texture" (now image) cache, both of them handled as images
* Implements texture types (Snorm, Unorm, Sint, Uint)
* Implements rendertarget formats
* Implements zeta targets (depth + stencil)

It has some performance regressions on some scenes, but I'd try to fix that later.

I tried to implement all rendertargets formats I found, but some may be missing (they will signal an exception). If someone finds one of those, please let me know.


Cyuubi:
Might wanna fix the conflicts  :P

Ryujinx.Graphics/Gal/OpenGL/TCE.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        //TODO: Use a variable value here
+        public const int MaxBpp = 16;
+
+        private static int CopyBuffer = 0;
+        private static int CopyBufferSize = 0;
+
+        public GalImage Image { get; private set; }
+
+        public int Width { get => Image.Width; }
```
gdkchan: This can be simplified to just `public int Width => Image.Width;` (same below).

Ryujinx.Graphics/Gal/OpenGL/TCE.cs
```diff
@@ -0,0 +1,123 @@
+﻿using OpenTK.Graphics.OpenGL;
+using System;
+
+namespace Ryujinx.Graphics.Gal.OpenGL
+{
+    class TCE
```
gdkchan: Maybe we should use a more descriptive name.

Ryujinx.Graphics/Gal/OpenGL/TCE.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                this.InternalFormat = InternalFormat;
+                this.PixelFormat = PixelFormat;
+                this.PixelType = PixelType;
+
+                Initialized = true;
+            }
+        }
+
+        public bool HasColor   { get => ImageFormatConverter.HasColor(Format); }
```
gdkchan: Same comment as above (can be simplifies), also usually properties are located before the methods inside the class (on the top).

Ryujinx.Graphics/Ryujinx.Graphics.csproj
```diff
@@ -13,6 +13,10 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Include="Gal\OpenGL\TCE.cs" />
```
gdkchan: Why was this path added to the project?

Ryujinx.HLE/Gpu/Engines/NvGpuEngine3d.cs
```diff
@@ -161,13 +165,22 @@ private void SetFrameBuffer(NvGpuVmm Vmm, int FbIndex)
...
+
             long Key = Vmm.GetPhysicalAddress(VA);
 
             FrameBuffers.Add(Key);
 
             int Width  = ReadRegister(NvGpuEngine3dReg.FrameBufferNWidth  + FbIndex * 0x10);
             int Height = ReadRegister(NvGpuEngine3dReg.FrameBufferNHeight + FbIndex * 0x10);
-
+            
             float TX = ReadRegisterFloat(NvGpuEngine3dReg.ViewportNTranslateX + FbIndex * 4);
```
gdkchan: This should be `* 8` instead of `* 4` (I know its not related to this PR but maybe its worth fixing now that you're workig on framebuffers).

Ryujinx.HLE/Gpu/Engines/NvGpuEngine3d.cs
```diff
@@ -161,13 +165,22 @@ private void SetFrameBuffer(NvGpuVmm Vmm, int FbIndex)
...
+            }
+
             long Key = Vmm.GetPhysicalAddress(VA);
 
             FrameBuffers.Add(Key);
 
             int Width  = ReadRegister(NvGpuEngine3dReg.FrameBufferNWidth  + FbIndex * 0x10);
             int Height = ReadRegister(NvGpuEngine3dReg.FrameBufferNHeight + FbIndex * 0x10);
-
+
```
gdkchan: nit: TAB not necessary.

PR #367, opened 2018-08-19
Lock GbpQueueBuffer till Vsync is signalled

emmauss:
This change locks nvflinger's queuebuffer till a vsync event is signalled, effectively limiting the game speed to 60 fps. I have not noticed any performance hit if speed is less than 100%, whether the toggle is on or off. `Tab` toggles the emulated vsync.

### DIFF
--- a/Ryujinx.HLE/Switch.cs
+++ b/Ryujinx.HLE/Switch.cs
@@ -62,7 +62,7 @@ namespace Ryujinx.HLE
 
             Hid = new Hid(this, System.HidSharedMem.PA);
 
-            VsyncEvent =  new AutoResetEvent(true);
+            VsyncEvent = new AutoResetEvent(true);
         }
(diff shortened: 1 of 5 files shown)


emmauss:
okay

Ryujinx.HLE/Switch.cs
```diff
@@ -54,6 +61,8 @@ public Switch(IGalRenderer Renderer, IAalOutput AudioOut)
             Statistics = new PerformanceStatistics();
 
             Hid = new Hid(this, System.HidSharedMem.PA);
+
+            VsyncEvent =  new AutoResetEvent(true);
```
gdkchan: nit: Spacing after `=`.

emmauss: fixed.

PR #371, opened 2018-08-21
Avoid gpr overwritting on Ld_C instruction

ReinUsesLisp:
It should fix some vertex explosions.
Thanks to bunnei (from yuzu) for pointing out this issue.


gdkchan:
I think that `GetOperCbuf36` is now unused? If so I think it can be removed. Also it would be nice if `MakeTemporary` had a index argument or something (to allow using more than one temp), then it would just do `0x100 + Index`.

PR #373, opened 2018-08-21
Improve LM implementation

Thog:
- Manage end of the log packet correctly.
- Add drop count parsing.
- Use the correct buffer type. (0x21 not 0x9)


Ryujinx.HLE/HOS/Services/Lm/ILogger.cs
```diff
@@ -49,17 +49,24 @@ public long Log(ServiceCtx Context)
                     LmLogField Field = (LmLogField)Type;
 
                     string FieldStr = string.Empty;
-
-                    if (Field == LmLogField.Skip)
+                    if (Field == LmLogField.Start)
```
gdkchan: nit: Imo add a empty new life before the `if`.

PR #375, opened 2018-08-22
Implement multiple rendertarget attachments and depth writting

ReinUsesLisp:
For depth writting I used [Subv](https://github.com/yuzu-emu/yuzu/blob/b0f7713fce968c83ee105458dfe361644c937e21/src/video_core/renderer_opengl/gl_shader_decompiler.cpp#L941)'s implementation as reference. Worked from it for in-shader color attachments.
Color attachments are now written after `program()` in GlslDecompiler (to be able to mask individual components).
I couldn't find any graphical progresion since the only game I have that uses it (SMO) has other issues, at least there are no reggresions.


Ryujinx.HLE/Gpu/Engines/NvGpuEngine3d.cs
```diff
@@ -102,10 +102,14 @@ private void VertexEndGl(NvGpuVmm Vmm, NvGpuPBEntry PBEntry)
             SetAlphaBlending(State);
             SetPrimitiveRestart(State);
 
-            //Enabling multiple framebuffer attachments cause graphics reggresions
-            SetFrameBuffer(Vmm, 0);
+            for (int FbIndex = 0; FbIndex < 8; FbIndex++)
+            {
+                SetFrameBuffer(Vmm, 0);
+            }
             SetZeta(Vmm);
```
gdkchan: nit: Empty new line after `}`.

PR #368, opened 2018-08-19
Use a program counter to control shaders' flow

ReinUsesLisp:
This should more precise and allow ssy/sync instructions to be implemented.
It may cause performance regressions on shader-limited GPUs.

Thanks Subv for the suggestion.

PR #382, opened 2018-08-24
Implement SSY/SYNC shader instructions

ReinUsesLisp:
This has performance issues on low-end GPUs regardless of their vendor since it includes #368.


gdkchan:
It's worth noting that this has a significant perf impact on igpus, for example, puyo dropped from 60~70 fps on the title screen to 14 fps here. However, it's better to have something thats slow for some users but works, than something that is fast but does not work properly. We can revisit that and implement smarted control flow handling (with if/else/loop generation, or maybe even a mix of both methods) later.

PR #386, opened 2018-08-24
Implements proper save path

emmauss:
This implements proper save paths for games. uses the following format for the save paths generated  fo reach game
`{partition}/save/{saveId}/{userId}/{titleId}/`.
this also supports sdcard system saves, though no process uses it currently : http://switchbrew.org/index.php?title=SD_Filesystem#save


Ryujinx.HLE/FileSystem/Save.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Ryujinx.HLE.HOS.SystemState;
+
+namespace Ryujinx.HLE.FileSystem
+{
+    struct Save
```
gdkchan: Imo it's better to use `{ get; private set; }` (making the struct immutable) and using a constructor to initialize the values.

Ryujinx.HLE/FileSystem/SaveHelper.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+using Ryujinx.HLE.HOS;
```
gdkchan: This should be on the top (alphabetically sorted).

Ryujinx.HLE/FileSystem/SaveHelper.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
```
gdkchan: This should be before `System.Text` (alphabetically sorted).

Ryujinx.HLE/FileSystem/SaveHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+using Ryujinx.HLE.HOS;
+
+namespace Ryujinx.HLE.FileSystem
+{
+    static class SaveHelper
+    {
+        public static string GetSavePath(Save SaveMetaData, ServiceCtx Context)
+        {
+            string BaseSavePath   = "nand";
```
gdkchan: Use the `NandPath` constant inside `VirtualFileSystem`.

Ryujinx.HLE/FileSystem/SaveHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+            long   CurrentTitleId = SaveMetaData.TitleId;
+
+            switch (SaveMetaData.SaveSpaceId)
+            {
+                case SaveSpaceId.NandUser:
+                    BaseSavePath =  Path.Combine(BaseSavePath, "user");
```
gdkchan: I recommend creating constants for those names inside `VirtualFileSystem` too (it's better to have them all on a single place to allow changing directory names easily if necessary).

Ryujinx.HLE/FileSystem/SaveHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+                case SaveSpaceId.SdCard:
+                    BaseSavePath = Path.Combine("sdmc", "Nintendo");
+                    break;
+            }
+
+            BaseSavePath = Path.Combine(BaseSavePath, "save");
+
+            if (SaveMetaData.TitleId == 0 && SaveMetaData.SaveDataType == SaveDataType.SaveData)
+            {
+                CurrentTitleId = Context.Process.MetaData.ACI0.TitleId;
```
gdkchan: Just a note: Currently `MetaData` can be null if a NPDM file is not found. We should either allow for that in all places where `MetaData` is used, or initialize it with default values if an `NPDM` file is not found.

Ryujinx.HLE/FileSystem/VirtualFileSystem.cs
```diff
@@ -1,7 +1,8 @@
 using System;
 using System.IO;
+using Ryujinx.HLE.HOS;
```
gdkchan: This should be on the top (alphabetically sorted).

Ryujinx.HLE/HOS/Services/FspSrv/IFileSystemProxy.cs
```diff
@@ -1,6 +1,9 @@
 using Ryujinx.HLE.HOS.Ipc;
 using Ryujinx.HLE.Logging;
 using System.Collections.Generic;
+using Ryujinx.HLE.FileSystem;
```
gdkchan: Sorting.

Ryujinx.HLE/HOS/Services/FspSrv/IFileSystemProxy.cs
```diff
@@ -1,6 +1,9 @@
...
+
+using Ryujinx.HLE.HOS.SystemState;
```
gdkchan: Sorting.

Ryujinx.HLE/HOS/SystemState/UserId.cs
```diff
@@ -43,6 +43,20 @@ void WriteBytes(long Value)
...
+            Bytes = UserIdBytes;
+
+            UserIdHex = string.Empty;
+
+            foreach (byte Byte in Bytes)
+            {
+                UserIdHex += Byte.ToString("X2");
+            }
+
+            this.Bytes = Bytes;
```
gdkchan: This line is not necessary.

Ryujinx.HLE/HOS/SystemState/UserId.cs
```diff
@@ -43,6 +43,20 @@ void WriteBytes(long Value)
             this.Bytes = Bytes;
         }
 
+        public UserId(byte[] UserIdBytes)
```
gdkchan: I think it would be good to throw an exception for the invalid cases here aswell (since iirc this class is public). Invalid cases are UserIdBytes being null, UserIdBytes.Length != 16 or UserIdBytes having all bytes == 0.

Ryujinx.HLE/FileSystem/VirtualFileSystem.cs
```diff
@@ -44,7 +45,10 @@ public string GetFullPath(string BasePath, string FileName)
 
         public string GetSdCardPath() => MakeDirAndGetFullPath(SdCardPath);
 
-        public string GetGameSavesPath() => MakeDirAndGetFullPath(NandPath);
+        public string GetNandPath() => MakeDirAndGetFullPath(NandPath);
+
+        public string GetGameSavePath(Save SaveMetaData, ServiceCtx Context)
```
gdkchan: I think its better to use more descriptive names for the types, and shorter names for arguments/variables. What about `Save SaveMetaData` -> `SaveInfo Save` or something like that?

Ryujinx.HLE/HOS/SystemState/UserId.cs
emmauss: I am going to remove that method, as the one with longs is sufficient

emmauss:
fixed nits

Ryujinx.HLE/FileSystem/Save.cs
emmauss: I don't know, It will lack some flexibility as it will become readonly. the struct isn't just for the save struct, but may even be used for the savedatainfo struct in future, as their contents are very similar, only need a titleid for savedatainfo.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -103,7 +103,7 @@ void LoadNso(string FileName)
                 }
             }
 
-            if (!MainProcess.MetaData.Is64Bits)
+            if ((MainProcess.MetaData?.Is64Bits).HasValue && !MainProcess.MetaData.Is64Bits)
```
gdkchan: could be just `MainProcess.MetaData?.Is64Bits ?? true` (if NPDM is null, assume 64-bits).

emmauss:
changes are done

gdkchan:
Please fix conflicts.

gdkchan:
I don't think that passing `ServiceCtx` to the `SaveInfo` struct and reading data from there is good design. I was suggesting adding one argument on the constructor for each struct field, then reading data on `LoadSaveDataFileSystem` and passing the value read on the constructor arguments directly.

emmauss:
I removed the context param. How does it look now?

Ryujinx.HLE/FileSystem/VirtualFileSystem.cs
```diff
@@ -1,14 +1,18 @@
...
-        private const string BasePath   = "RyuFs";
-        private const string NandPath   = "nand";
-        private const string SdCardPath = "sdmc";
-        private const string SystemPath = "system";
+        public const string BasePath   = "RyuFs";
+        public const string NandPath   = "nand";
+        public const string SdCardPath = "sdmc";
+        public const string SystemPath = "system";
+
+        public static string SystemNandPath = Path.Combine("nand", "system");
```
gdkchan: Please use the NandPath constant here (and below) aswell.

Ryujinx.HLE/FileSystem/SaveInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+    struct SaveInfo
+    {
+        public long   TitleId { get; private set; }
+        public long   SaveId  { get; private set; }
+        public UserId UserId  { get; private set; }
+
+        public SaveDataType SaveDataType { get; private set; }
+        public SaveSpaceId  SaveSpaceId  { get; private set; }
+
+        public SaveInfo(long TitleId,
```
gdkchan: `long TitleId,` on a new line.

Ryujinx.HLE/FileSystem/VirtualFileSystem.cs
```diff
@@ -50,7 +54,10 @@ public string GetFullPath(string BasePath, string FileName)
 
         public string GetSdCardPath() => MakeDirAndGetFullPath(SdCardPath);
 
-        public string GetGameSavesPath() => MakeDirAndGetFullPath(NandPath);
+        public string GetNandPath() => MakeDirAndGetFullPath(NandPath);
+
+        public string GetGameSavePath(SaveInfo Save, ServiceCtx Context)
```
gdkchan: Isn't this line too long? I think it would be better to use `{ }` and the function body inside.

Ryujinx.HLE/HOS/Services/FspSrv/IFileSystemProxy.cs
```diff
@@ -70,5 +71,22 @@ public long GetGlobalAccessLogMode(ServiceCtx Context)
...
             return 0;
         }
+
+        public void LoadSaveDataFileSystem(ServiceCtx Context)
+        {
+            SaveSpaceId SaveSpaceId = (SaveSpaceId)Context.RequestData.ReadInt64();
+
+            long TitleId = Context.RequestData.ReadInt64();
+
+            UserId UserId = new UserId(Context.RequestData.ReadInt64(), Context.RequestData.ReadInt64());
```
gdkchan: `Context.RequestData.ReadInt64()` on new lines.

emmauss:
fixed

emmauss:
fixed, again

Schroedingers-Cat:
> uses the following format for the save paths generated fo reach game
> `{partition}/save/{saveId}/{userId}/{titleId}/`

Ryujinx v1.1.132 seems to be using a rather different pattern for the save paths that does not include the titleId in any way. It seems to use an increasing counter making it difficult to predict under which directory a specific title will be saved. Has this been changed?  

It would be great to order the saves per titleId as this PR did so saves can be used across different emulators by symlinking the save folder.

Schroedingers-Cat:
> @Schroedingers-Cat this has indeed been changed as this PR is almost 4 years old. The increasing counter is how the Nintendo Switch itself does it. You would be better off asking other emulators to implement their file systems accurately instead 👍

Thanks for the insight. Interesting that the Switch stores the savefiles in such an unpredictable manner.  

If other emulators were to adopt the Switches original structure, what top-level directory would need to be symlinked (or the emulator pointed at) so the emulator understands and sees another emulator's save files?  

What folder of the save directory would need to be synced if users want to sync their save files across devices without syncing hardware-dependent files like compiled shaders (relates to https://github.com/Ryujinx/Ryujinx/issues/3191)?

PR #388, opened 2018-08-27
[WIP] Partial non-2D textures support

ReinUsesLisp:
It adds in-shader support and adds texture targets to OpenGL. It does not modify swizzle math so the implementation is partial.


ReinUsesLisp:
I have to re-implement this completely to make it work, and I should implement texture format families one by one since each one has its own complexities.

PR #391, opened 2018-08-28
Implement loading of profile image

emmauss:
This implements IProfile:10 and 11.
I also renamed the acc:u0 class so it is also used for  acc:u1


emmauss:
Okay. renamed icon to profilepicture. leaving it blank would look too bland. since it is an image, lets let it show something.

gdkchan:
The file is still called `Ryujinx_icon`.

emmauss:
renamed.

PR #392, opened 2018-08-29
Handle Ipa PASS argument as needed in Fragment Shaders

ReinUsesLisp:


gdkchan:
Since you're working on this, can you remove the `position` attribute aswell? The main reason it was added was due to those elements (0x70, 0x74) being accessed, I think that for all other cases, we can hardcode a value of 1.

ReinUsesLisp:
Are you sure about that? It's still being accessed when Ipa's argument is not `Pass`

gdkchan:
Fair enough. I think it doesn't make sense for it to have completely different values based on interpolation mode, and on glsl at least, theres no `gl_Position` built-in on fragment shaders, only `gl_FragDepth`. I think it could be safely removed, but I don't remember if all the cases accessing 0x70/0x74 uses PASS.

PR #394, opened 2018-08-30
Shader invalidation and cache

ReinUsesLisp:
Currently we're ignoring shader changes that can end up in undefined behaviour. This adds shaders to the cache invalidation and keeps tracks of them in OGLShader to avoid rebuilding shaders.

I'm not happy with how the code ended up being, I'll happily listen suggestion.
I think it could end up being better if all LLE GPU emulation is moved from Ryujinx.HLE to Ryujinx.Graphics.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLShader.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLShader.cs
@@ -15,7 +15,9 @@ namespace Ryujinx.Graphics.Gal.OpenGL
 
         public OGLShaderProgram Current;
 
-        private ConcurrentDictionary<long, OGLShaderStage> Stages;
+        private Dictionary<long, List<OGLShaderStage>> Stages;
+
+        private Dictionary<long, OGLShaderStage> TopStages;
 
         private Dictionary<OGLShaderProgram, int> Programs;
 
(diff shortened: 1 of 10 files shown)


Ryujinx.Graphics/Gal/Shader/ShaderDecoder.cs
```diff
@@ -202,5 +199,18 @@ private static bool NodeHasNext(ShaderIrNode Node)
             return Op.Inst != ShaderIrInst.Exit &&
                    Op.Inst != ShaderIrInst.Bra;
         }
+
+        private static unsafe long ReadQWord(byte[] Bytes, long Position)
```
gdkchan: Why is it called QWord? word sizes varies between archs (for example, its 32-bits on arm and 16-bits on x86). Plus imo using unsafe code is not a good pratice, unless its on a performance critical path (not hte case here). You can use `BitConverter.ToInt64` to do the same thing.

Ryujinx.Graphics/Gal/Shader/ShaderHeader.cs
```diff
@@ -136,6 +136,14 @@ public int DepthRegister
             }
         }
 
+        private static unsafe uint ReadWord(byte[] Bytes, long Position)
```
gdkchan: Same comment as `ReadQWord`.

Ryujinx.HLE/Gpu/Engines/NvGpuEngine3d.cs
```diff
@@ -68,6 +70,8 @@ void AddMethod(int Meth, int Count, int Stride, NvGpuMethod Method)
             {
                 UploadedKeys[i] = new List<long>();
             }
+
+            ShadersSize = new Dictionary<long, long>();
```
gdkchan: Imo its better to store sizes on OGLShader and retrieve from there (like what is done for the other cached resources).

Ryujinx.Graphics/Gal/OpenGL/OGLShader.cs
```diff
@@ -29,60 +31,71 @@ public OGLShader(OGLConstBuffer Buffer)
...
+            long          Key,
+            byte[]        BinaryA,
+            byte[]        BinaryB,
+            GalShaderType Type)
         {
-            Stages.GetOrAdd(Key, (Stage) => ShaderStageFactory(Memory, Key, 0, false, Type));
-        }
+            if (Stages.TryGetValue(Key, out List<OGLShaderStage> Cache))
+            {
+                OGLShaderStage CachedStage = Cache.Find((OGLShaderStage Stage) => Stage.EqualsBinary(BinaryA, BinaryB));
```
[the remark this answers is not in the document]
ReinUsesLisp: Multiple shaders are stored in the same key, they are compared to check if it was compiled previously.

gdkchan:
`I think it could end up being better if all LLE GPU emulation is moved from Ryujinx.HLE to Ryujinx.Graphics.` Now that it was moved I think you could improve it then. Please fix conflicts.

jduncanator:
Closing as superseded by #533

PR #395, opened 2018-08-30
Asynchronously log messages to the Console window

jduncanator:
Writing to the Console blocks until the write completes. This has the potential to block any code path that logs to the application logger.

By queuing up log messages in an asynchronous queue and returning without blocking, we can speed up code paths that heavily log to the console (for example stubbed services like ServiceHid).

From testing, this results in a roughly 8% time decrease between Ryujinx startup and the splash screen in Super Mario Odyssey on my system (in Debug) - 00:03:19.591 down to
00:03:04.354. Depending on your system, YMMV.

### DIFF
--- a/Ryujinx/Ui/ConsoleLog.cs
+++ b/Ryujinx/Ui/ConsoleLog.cs
@@ -1,5 +1,6 @@
 using Ryujinx.HLE.Logging;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
 
@@ -7,6 +8,8 @@ namespace Ryujinx
 {
     static class ConsoleLog
     {
+        private static Thread MessageThread;
+        private static BlockingCollection<LogEventArgs> MessageQueue;
         private static Dictionary<LogLevel, ConsoleColor> LogColors;
 
         private static object ConsoleLock;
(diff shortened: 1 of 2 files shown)


Ryujinx/Ui/ConsoleLog.cs
```diff
@@ -21,15 +24,37 @@ static ConsoleLog()
...
+            {
+                while (!MessageQueue.IsCompleted)
+                {
+                    try
+                    {
+                        PrintLog(MessageQueue.Take());
+                    }
+                    catch (InvalidOperationException)
+                    {
+                        // IOE means that Take() was called on a completed collection.
```
gdkchan: Why not using `TryTake`? This way you don't need to catch the exception I assume.

jduncanator:
No need to call CompleteAdding. The thread is a background thread which will exit automatically upon the rest of the application exiting. The handling of IsComplete is there to future proof the implementation in the event someone decides there is an appropriate use of completing the collection.

If you think there is an appropriate place to call CompleteAdding now, let me know, but this was designed to be a drop in replacement for the existing logging implementation.

Ryujinx/Ui/ConsoleLog.cs
jduncanator: `TryTake` does not block if the collection is empty, the blocking nature of `Take` is what makes this work. There is an overload of `TryTake` that takes a "wait" time, however `TryTake` can still result in an IOE, so it's better to just use `Take` and handle the exception.

Ryujinx/Ui/ConsoleLog.cs
```diff
@@ -21,15 +24,37 @@ static ConsoleLog()
                 { LogLevel.Error,   ConsoleColor.Red      }
             };
 
+            MessageQueue = new BlockingCollection<LogEventArgs>();
+
             ConsoleLock = new object();
+
+            MessageThread = new Thread(() =>
+            {
+                while (!MessageQueue.IsCompleted)
```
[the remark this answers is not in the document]
jduncanator: The Thread is a background thread and will be aborted when the application exits.

PR #401, opened 2018-09-03
Set the default culture to InvariantCulture

0x0ade:
This mainly affects some C# string conversions (`ToString` and `Parse`), and I've experienced a few weird issues in other C# projects due to region differences. Setting this as early as possible doesn't have any negative impact and only makes everything behave the same in every region.

In parts of Europe, `,` is used as the decimal separator, not `.`

The most noticeable change:

image

becomes

image


gdkchan:
All places using `ToString` and related methods where such differences could cause problems (like glsl shader generation) should be using InvariantCulture already.

0x0ade:
Ah, true - didn't realize that. Sorry.

PR #402, opened 2018-09-03
Move GPU emulation from Ryujinx.HLE to Ryujinx.Graphics and misc changes

ReinUsesLisp:
In general it's a code cleanup of the GPU code.
* Move GPU emulation to Ryujinx.Graphics
* Texture and image format descriptors are unified in a single file
* Abstract shader OpCode decoding

I've made a revision of the last because a single number being wrong could lead to hard to track bugs. I plan on doing another revision later.


ReinUsesLisp:
OpCode changes were remade with regex instead of manually, so they are safe now. Changed yet again the image table and made Read* symbols internal instead of public.

ReinUsesLisp:
Changed yet again the image table.
GalImageFormat is now an intermediate enumeration between GalTextureFormat and GalSurfaceFormat (prior GalFrameBufferFormat) matching envytools names. Then GalImageFormat gets converted to whatever the backend supports (ASTC gets converted to RGBA8, G8R8 to R8G8), this allows backend-specific optimizations (like using native ASTC when available).

Ryujinx.HLE/Switch.cs
```diff
@@ -1,6 +1,6 @@
 using Ryujinx.Audio;
 using Ryujinx.Graphics.Gal;
-using Ryujinx.HLE.Gpu;
+using Ryujinx.Graphics;
```
gdkchan: nit: Should be before `using Ryujinx.Graphics.Gal;`.

Ryujinx.Graphics/Gal/OpenGL/OGLTexture.cs
```diff
@@ -67,7 +70,17 @@ public void Create(long Key, byte[] Data, GalImage Image)
                         Image.Width,
                         Image.Height, 1);
 
-                    Image.Format = GalImageFormat.A8B8G8R8_UNORM_PACK32;
+                    Image.Format = GalImageFormat.A8B8G8R8 | GalImageFormat.Unorm;
+                }
+                else if (TypeLess == GalImageFormat.G8R8)
```
gdkchan: I think its more efficient (and easier) to swap the components on the texture swizzle (GL_TEXTURE_SWIZZLE_RGBA). I think all maxwell formats are "reversed" so we could just use the "normal" formats for texture and always reverse the swizzle, however this doesn't work for frame and depth buffer output formats, hmm.

Ryujinx.Graphics/Texture/ImageTable.cs
```diff
@@ -0,0 +1,347 @@
+﻿using Ryujinx.Graphics.Gal;
+using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.Graphics.Texture
+{
+    static class ImageTable
```
gdkchan: I'm not sure about the name `ImageTable`. it does contain a image format table, but this class is not just about that, it also exposes a bunch of methods to allow converting between formats and getting some texture info. In fact the tables are private and its mostly a implementation detail, maybe just call it `ImageUtils` or something like that?

Ryujinx.Graphics/Texture/ImageTable.cs
```diff
@@ -0,0 +1,347 @@
...
+
+            throw new NotImplementedException(Format.ToString());
+        }
+
+        public static TextureReaderDelegate GetReader(GalImageFormat Format)
+        {
+            return GetImageDescriptor(Format).Reader;
+        }
+
+        public static int GetImageSize(GalImage Image)
```
gdkchan: If bpp and block size was added to the table aswell, this method could be reduced to just a couple of lines.

Ryujinx.Graphics/Texture/ImageTable.cs
```diff
@@ -0,0 +1,347 @@
...
+                { GalTextureFormat.Astc2D6x5,   GalImageFormat.ASTC_6x5   | Unorm         },
+                { GalTextureFormat.Astc2D8x6,   GalImageFormat.ASTC_8x6   | Unorm         },
+                { GalTextureFormat.Astc2D10x8,  GalImageFormat.ASTC_10x8  | Unorm         },
+                { GalTextureFormat.Astc2D12x10, GalImageFormat.ASTC_12x10 | Unorm         },
+                { GalTextureFormat.Astc2D8x5,   GalImageFormat.ASTC_8x5   | Unorm         },
+                { GalTextureFormat.Astc2D10x5,  GalImageFormat.ASTC_10x5  | Unorm         },
+                { GalTextureFormat.Astc2D10x6,  GalImageFormat.ASTC_10x6  | Unorm         }
+            };
+
+        private static readonly Dictionary<GalImageFormat, ImageDescriptor> s_ImageTable =
```
gdkchan: It would be nice to merge this table with the table above, but there are probably other complications in doing that that I'm missing.

Ryujinx.Graphics/Texture/ImageTable.cs
```diff
@@ -0,0 +1,347 @@
...
+            new Dictionary<GalImageFormat, ImageDescriptor>()
+            {
+                { GalImageFormat.R32G32B32A32, new ImageDescriptor(TextureReader.Read16Bpp,                       true, false, false) },
```
gdkchan: I think it would be nice if instead of `TextureReader.Read*` methods, only bpp/block size was stored on the table. Most of those read methods are almost the same, usually only with the pixel/block size changing. If such info was stored on the table, it would be possible to use a single method to read for all sizes. Currently only rgba5551 and rgb565 formats are special cased, the components are reversed on read, but if this was implemented on swizzle as suggested above those special read methods would no longer be necessary.

The downside is tha they would most likely be slower aswell, but its possible to take advantage of the JIT inlining capabilities and create one method for each size, each calling the common method. After inlining, it should produce an optimial version without uneeded checks for each size.

Ryujinx.Graphics/Gal/OpenGL/OGLTexture.cs
ReinUsesLisp: Since there's that swizzle bug, I'd like to do it when swizzling is safe

PR #403, opened 2018-09-03
Do Tier1 translation in background

strubelz:
Should save a bit of startup time. Maybe we can even do more optimizations when doing Tier1 translation, or do it more often, to improve performance.


gdkchan:
Something similar was implemented at: https://github.com/Ryujinx/Ryujinx/pull/515
Closing...

PR #405, opened 2018-09-05
Optimize BIC, BSL, BIT, BIF, XTN, ZIP, DUP (Gp), FMADD (Scalar) and FCVT (Scalar) using SSE intrinsics

gdkchan:
Didn't notice any performance difference on the few games I tested.
Most of those instructions are not tested because currently there are no tests for them on Ryujinx.Tests.

### DIFF
--- a/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
+++ b/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
@@ -410,8 +410,8 @@ namespace ChocolArm64.Instruction
 
                     Type[] Types = new Type[] { typeof(Vector128<float>), typeof(Vector128<float>) };
 
-                    Context.EmitCall(typeof(Sse).GetMethod( nameof(Sse.MultiplyScalar), Types));
-                    Context.EmitCall(typeof(Sse).GetMethod( nameof(Sse.AddScalar),      Types));
+                    Context.EmitCall(typeof(Sse).GetMethod(nameof(Sse.MultiplyScalar), Types));
+                    Context.EmitCall(typeof(Sse).GetMethod(nameof(Sse.AddScalar),      Types));
 
                     Context.EmitStvec(Op.Rd);
 
@@ -425,8 +425,8 @@ namespace ChocolArm64.Instruction
 
                     Type[] Types = new Type[] { typeof(Vector128<double>), typeof(Vector128<double>) };
 
-                    Context.EmitCall(typeof(Sse2).GetMethod( nameof(Sse2.MultiplyScalar), Types));
-                    Context.EmitCall(typeof(Sse2).GetMethod( nameof(Sse2.AddScalar),      Types));
+                    Context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.MultiplyScalar), Types));
+                    Context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.AddScalar),      Types));
 
                     EmitStvecWithCastFromDouble(Context, Op.Rd);
 
--- a/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
+++ b/ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
@@ -4,6 +4,7 @@ using ChocolArm64.Translation;
 using System;
 using System.Reflection;
 using System.Reflection.Emit;
+using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.X86;
 
 using static ChocolArm64.Instruction.AInstEmitSimdHelper;
@@ -31,7 +32,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse2)
             {
-                EmitSse2Call(Context, nameof(Sse2.Add));
+                EmitSse2Op(Context, nameof(Sse2.Add));
             }
             else
             {
@@ -305,7 +306,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse && AOptimizations.UseSse2)
             {
-                EmitScalarSseOrSse2CallF(Context, nameof(Sse.AddScalar));
+                EmitScalarSseOrSse2OpF(Context, nameof(Sse.AddScalar));
             }
             else
             {
@@ -317,7 +318,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse && AOptimizations.UseSse2)
             {
-                EmitVectorSseOrSse2CallF(Context, nameof(Sse.Add));
+                EmitVectorSseOrSse2OpF(Context, nameof(Sse.Add));
             }
             else
             {
@@ -375,7 +376,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse && AOptimizations.UseSse2)
             {
-                EmitScalarSseOrSse2CallF(Context, nameof(Sse.DivideScalar));
+                EmitScalarSseOrSse2OpF(Context, nameof(Sse.DivideScalar));
             }
             else
             {
@@ -387,7 +388,7 @@ namespace ChocolArm64.Instruction
         {
             if (AOptimizations.UseSse && AOptimizations.UseSse2)
             {
-                EmitVectorSseOrSse2CallF(Context, nameof(Sse.Divide));
+                EmitVectorSseOrSse2OpF(Context, nameof(Sse.Divide));
             }
             else
             {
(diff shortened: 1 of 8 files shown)


LDj3SNuD:
> Most of those instructions are not tested because currently there are no tests for them on Ryujinx.Tests.

BIC => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimdReg.cs#L373-L431
BSL => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimdReg.cs#L553-L611
BIT => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimdReg.cs#L493-L551
BIF => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimdReg.cs#L433-L491
XTN => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimd.cs#L2391-#L2445
ZIP => https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Tests/Cpu/CpuTestSimdReg.cs#L4279-#L4405
Dup_Gp => Still untested.


`UseAllSseIfAvailable = true;`
SDK 2.1.401

[TestsLog](https://pastebin.com/CthqsyNC)
The `Zip2_V_8B_4H_2S` Test fails.
With `UseAllSseIfAvailable = false;` the Test does not fail.

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -36,91 +67,119 @@ public static int GetImmShr(AOpCodeSimdShImm Op)
...
+        }
+
+        public static void EmitLdvecWithUnsignedCast(AILEmitterCtx Context, int Reg, int Size)
+        {
+            Context.EmitLdvec(Reg);
+
+            switch (Size)
             {
-                Context.EmitCall(Type.GetMethod(Name, new Type[] { BaseType }));
+                case 0: AVectorHelper.EmitCall(Context, nameof(AVectorHelper.VectorSingleToByte)); break;
```
LDj3SNuD: Align `break;`.

ChocolArm64/Instruction/AInstEmitSimdHelper.cs
```diff
@@ -36,91 +67,119 @@ public static int GetImmShr(AOpCodeSimdShImm Op)
...
+            Context.EmitStvec(Reg);
+        }
 
-            if (Op.RegisterSize == ARegisterSize.SIMD64)
+        public static void EmitStvecWithUnsignedCast(AILEmitterCtx Context, int Reg, int Size)
+        {
+            switch (Size)
             {
-                EmitVectorZeroUpper(Context, Op.Rd);
+                case 0: AVectorHelper.EmitCall(Context, nameof(AVectorHelper.VectorByteToSingle)); break;
```
LDj3SNuD: Nit: align `break;`.

ChocolArm64/Instruction/AInstEmitSimdMove.cs
```diff
@@ -394,28 +487,53 @@ private static void EmitVectorZip(AILEmitterCtx Context, int Part)
...
-                EmitVectorExtractZx(Context, Op.Rm, Base + Index, Op.Size);
+                Context.EmitCall(typeof(Sse2).GetMethod(Name, Types));
 
-                EmitVectorInsertTmp(Context, Idx + 1, Op.Size);
-                EmitVectorInsertTmp(Context, Idx,     Op.Size);
+                EmitStvecWithUnsignedCast(Context, Op.Rd, Op.Size);
+
+                if (Op.RegisterSize == ARegisterSize.SIMD64)
+                {
+                    EmitVectorZeroUpper(Context, Op.Rd);
```
MerryMage: When `Part == 1 && Op.RegisterSize == ARegisterSize.SIMD64`:

You should do Sse.UnpackLow (**not** Sse.UnpackHigh), then take the shift the upper 64 bits of the result to become the lower 64 bits.

gdkchan:
I addressed PR feedback, additionaly also wrote optimized SSE paths for FCVT and FMADD aswell, however FMADD is not accurate, but should give the same results as the current implementation, once FMA intrinsics are available we can write a better impl (both faster and more accurate). ZIP1/2 are now tested, all test passes. The other instructions remains untested.

ChocolArm64/Instruction/AInstEmitSimdArithmetic.cs
```diff
@@ -397,11 +398,49 @@ public static void Fdiv_V(AILEmitterCtx Context)
...
+
+                if (Op.Size == 0)
+                {
+                    Context.EmitLdvec(Op.Ra);
+                    Context.EmitLdvec(Op.Rn);
+                    Context.EmitLdvec(Op.Rm);
+
+                    Type[] Types = new Type[] { typeof(Vector128<float>), typeof(Vector128<float>) };
+
+                    Context.EmitCall(typeof(Sse).GetMethod( nameof(Sse.MultiplyScalar), Types));
```
LDj3SNuD: Space before `nameof()`?
More below.

gdkchan: Fixed.

gdkchan:
I reverted the changes to `EmitVectorZero32_128` since the instruction is a vector instruction and can't be used to zero the 127:32 bits. I will optimize it later with INSERTPS once .net core 2.2 is out, since currently the intrinsic method is accepting a scalar value as input instead of a vector.

fiigii:
This project seems to use intrinsic via reflection? The performance of reflection intrinsic call is bad (especially for intrinsic that needs JIT-time const at the last parameter, e.g., `Sse2. Shuffle`), as each reflection intrinsic will be compiled to a function call rather than a single instruction.

jduncanator:
@fiigii I'm not sure I follow your logic? This entire project uses reflection, as we are emitting custom MSIL. The JIT (most of the time) compiles these down to single instructions. Have you got an example where you've seen this not to be the case?

gdkchan:
Like @jduncanator said, we "translate" ARM instructions to CIL, the jit is capable of optimizing this code and producing reasonable code most of the time. Emiting calls to the intrinsic methods also produces the expected results, much like code written in C# would.

`especially for intrinsic that needs JIT-time const` Yes, we had this problem with the INSERTPS instruction (and some others I think). However I managed to get the JIT to emit a single INSERTPS instruction by using a method with explicit if/else checks:
```cs
public static Vector128<float> VectorInsertSingle(float Value, Vector128<float> Vector, byte Index)
{
	if (Sse41.IsSupported)
	{
		//Note: The if/else if is necessary to enable the JIT to
		//produce a single INSERTPS instruction instead of the
		//jump table fallback.
		if (Index == 0)
		{
			return Sse41.Insert(Vector, Value, 0x00);
		}
		else if (Index == 1)
		{
			return Sse41.Insert(Vector, Value, 0x10);
		}
		else if (Index == 2)
		{
			return Sse41.Insert(Vector, Value, 0x20);
		}
		else if (Index == 3)
		{
			return Sse41.Insert(Vector, Value, 0x30);
		}
		else
		{
			throw new ArgumentOutOfRangeException(nameof(Index));
		}
	}
}
```
Since the method is inlined, the `Index` value is constant and the JIT is able to eliminate the dead code, with only `Sse41.Insert(Vector, Value, idx);` remaining, and it produces a single instruction in the end.

As for the other instruction I checked the code being produced and it produces more or less the expected results -- no function calls. The code is often not optimal (for example, sometimes useless register moves are produced, etc), but I don't think those are caused by reflection.

So far the biggest complication I've seen calling the intrinsics from IL using the .NET reflection API is caused by the fact it uses generics. Emiting calls to generic methods through reflection is complicated, so we opted to add methods to convert between the types with `StaticCast`, with one method for each type.

fiigii:
> This entire project uses reflection, as we are emitting custom MSIL.
> we "translate" ARM instructions to CIL

Ah, I see, I thought that uses reflection invocation, reflection emit should be fine. Thanks for the explanation.

> Yes, we had this problem with the INSERTPS instruction (and some others I think). 

@gdkchan Actually, I have been aware of this kind of issues https://github.com/dotnet/coreclr/issues/17108. Your scenario is a good example to push our work. Could you please report the issue to coreclr repo when you get a chance?

> The code is often not optimal (for example, sometimes useless register moves are produced, etc), but I don't think those are caused by reflection.

I believe most of these redundant register roundtrips have been solved in the current master branch of coreclr (but still having some memory roundtrips).  The fix will be available in the next released version.

>  Emiting calls to generic methods through reflection is complicated, so we opted to add methods to convert between the types with StaticCast, with one method for each type.

We are changing some of these generic APIs for .NET Core 3.0 https://github.com/dotnet/coreclr/issues/20062, for example, `StaticCast` will be renamed to `As` and all the generic intrinsic (not helpers) will be removed and exploded to non-generic instead.

fiigii:
Thank you so much for experimenting hardware intrinsic in this cool project. You are welcome to report any codegen issue or API design suggestion to coreclr (and @ me).

gdkchan:
Thanks! I will open a issue on the coreclr repo for the const imm case, and will try to report other issues aswell.

I also had a problem using `Sse2.SetZeroVector128<float>();`, it seems to crash the JIT. I later learned that I should be using `Sse.SetZeroVector128();`, however I think that the expected behavior is at least throwing an exception for the unsupported type. I will see if it was already reported/fixed on coreclr.

jduncanator:
@gdkchan This passes all the unit tests we have, and from my testing with ~6 games, none of regressed. I think the "TODOs" you've listed in the conversation on this PR are valid (eg. intrinsics that aren't available yet or intrinsics that are bugged in .NET Core 2.1), but I don't think we should hold up merging this in for that. It's probably best if you opened an issue to track those changes, and we can submit a PR later that addresses them, once .NET Core 2.2 is available.

PR #411, opened 2018-09-10
Partially implement LDG

ReinUsesLisp:
This is obviously based on @gdkchan's implementation but without the other stuff that's needed to show fonts.

### DIFF
--- a/Ryujinx.Graphics/Gal/IGalShader.cs
+++ b/Ryujinx.Graphics/Gal/IGalShader.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 namespace Ryujinx.Graphics.Gal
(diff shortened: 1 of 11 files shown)


Ryujinx.Graphics/Gal/IGalShader.cs
```diff
@@ -8,9 +9,14 @@ public interface IGalShader
 
         void Create(IGalMemory Memory, long VpAPos, long Key, GalShaderType Type);
 
+        ShaderDeclInfo GetGlobalMemoryUsage(long Key);
```
gdkchan: Would be nice to use a `IEnumerable<ShaderDeclInfo>` and allow multiple "global memories", that is, ldgs with different base addresses.

Ryujinx.Graphics/Gal/Shader/GlslDecl.cs
```diff
@@ -281,6 +285,52 @@ private void Traverse(ShaderIrNode[] Nodes, ShaderIrNode Parent, ShaderIrNode No
                     break;
                 }
 
+                case ShaderIrOperGmem Gmem:
+                {
+                    if (Gmem.BaseAddress is ShaderIrOperGpr BaseAddress)
```
gdkchan: My plan for this was extracting the core logic to a separate method, that would be used by TEX.B aswell, since they do similar things -- they go through each opcode on the same basic block, and tries to find the first assignment to the register being used, in this case to figure out from where the base address comes from. On the TEX.B case, its used to find from where the textuer handle comes from.

jduncanator:
As discussed, another approach is required to properly implement the LDG shader instruction. Closing this due to age & inactivity.

PR #412, opened 2018-09-11
Add FMAXP and FMINP (Vector) instructions on the CPU

gdkchan:
I moved logic from FADDP vector to a helper method that is now used by FADDP, FMAXP and FMINP, accepting an `Action` used to emit the appropriate IL opcodes/calls.

It can be extended to support scalar instructions aswell but I'm not sure if it's worth it, since code for scalar can be simplified (for example it doesn't need to store the addition result on a temp vector), so I think that the code reuse opportunity there is low, it's probably better to create a separate method for pairwise operations with scalars.

### DIFF
--- a/ChocolArm64/AOpCodeTable.cs
+++ b/ChocolArm64/AOpCodeTable.cs
@@ -278,10 +278,12 @@ namespace ChocolArm64
             SetA64("0>0011100<1xxxxx111101xxxxxxxxxx", AInstEmit.Fmax_V,        typeof(AOpCodeSimdReg));
             SetA64("000111100x1xxxxx011010xxxxxxxxxx", AInstEmit.Fmaxnm_S,      typeof(AOpCodeSimdReg));
             SetA64("0>0011100<1xxxxx110001xxxxxxxxxx", AInstEmit.Fmaxnm_V,      typeof(AOpCodeSimdReg));
+            SetA64("0>1011100<1xxxxx111101xxxxxxxxxx", AInstEmit.Fmaxp_V,       typeof(AOpCodeSimdReg));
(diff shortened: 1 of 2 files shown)


LDj3SNuD:
Instruction tables are okay.
Also move down the private method EmitAbs().
The EmitVectorPairwiseOpF() method should be moved to AInstEmitSimdHelper.cs and made public.
I would recommend to implement the EmitVectorPairwiseOpF() method in this way (tested), the same one already accepted in a previous PR:

```csharp
        public static void EmitVectorPairwiseOpF(AILEmitterCtx Context, Action Emit)
        {
            AOpCodeSimdReg Op = (AOpCodeSimdReg)Context.CurrOp;

            int SizeF = Op.Size & 1;

            int Words = Op.GetBitsCount() >> 4;
            int Pairs = Words >> SizeF + 2;

            for (int Index = 0; Index < Pairs; Index++)
            {
                int Idx = Index << 1;

                EmitVectorExtractF(Context, Op.Rn, Idx,     SizeF);
                EmitVectorExtractF(Context, Op.Rn, Idx + 1, SizeF);

                Emit();

                EmitVectorExtractF(Context, Op.Rm, Idx,     SizeF);
                EmitVectorExtractF(Context, Op.Rm, Idx + 1, SizeF);

                Emit();

                EmitVectorInsertTmpF(Context, Pairs + Index, SizeF);
                EmitVectorInsertTmpF(Context,         Index, SizeF);
            }

            Context.EmitLdvectmp();
            Context.EmitStvec(Op.Rd);

            if (Op.RegisterSize == ARegisterSize.SIMD64)
            {
                EmitVectorZeroUpper(Context, Op.Rd);
            }
        }
```

The Fmaxp_V and Fminp_V Tests, just implemented (PR#409), have been successfully passed, for all input types and values (Fpcr in reset state, Fpsr ignored).

gdkchan:
Done.

PR #413, opened 2018-09-12
 Move logging to Ryujinx.Common and make it a static class 

ReinUsesLisp:
Moves logger class to a new project, `Ryujinx.Common`.
It adds a message for `per component types` and ignores them using red's type (instead of throwing an exception). Unimplemented surface formats print a message and use RGBA8 as a fallback.

@jduncanator I know you were working on this but I wanted it to be in a separate PR.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/SvcHandler.cs
+++ b/Ryujinx.HLE/HOS/Kernel/SvcHandler.cs
@@ -1,8 +1,8 @@
 using ChocolArm64.Events;
 using ChocolArm64.Memory;
 using ChocolArm64.State;
+using Ryujinx.Common.Logging;
 using Ryujinx.HLE.HOS.Ipc;
-using Ryujinx.HLE.Logging;
 using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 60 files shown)


jduncanator:
This looks good to me. I've tested and it seems to be working, however can't attest to the correctness of the RGBA8 fallback (I'll have to trust that this is working).

Just keep in mind that this will cause a significant amount of work to be done on a few of the currently open PRs that touch any of the affected files here (they'll all need to be rebased), so if there are some PRs that are currently open that could get merged before this, I'd recommend trying to get them in and then merging this.

ReinUsesLisp:
It's not correct in any manner, but when you are debugging a game (that takes quite some time to boot) and hit an unimplemented surface format, it's frustrating. You can just write down what format (or formats) that is and implement them later.

jduncanator:
@ReinUsesLisp This lgtm, would be nice to get this in master. Can you resolve the conflicts?

jduncanator:
I'd vote for splitting this PR up too.

ReinUsesLisp:
Removed the graphics changes.

Ryujinx.HLE/HOS/Kernel/SvcHandler.cs
```diff
@@ -2,7 +2,7 @@
 using ChocolArm64.Memory;
 using ChocolArm64.State;
 using Ryujinx.HLE.HOS.Ipc;
-using Ryujinx.HLE.Logging;
+using Ryujinx.Common.Logging;
```
gdkchan: nit: Sorting.

PR #418, opened 2018-09-16
Allow "reinterpretation" of framebuffer/zeta formats

gdkchan:
Know issues:

- Reverse formats that are not supported by OpenGL are not yet handled (~~will be addressed soon~~).
- ~~Some cases are way slower than before.~~

Also currently NvGpuVmm depends on AMemory, and this caused a bunch of probems. In the future I plan to make NvGpuVmm fully independent and access memory directly (instead of doing double translation and using AMemory). This will simplify things a bit.

I also removed the texture Read* methods, replaced by a single `ReadTexture` method, and reduced `GetSize` by adding block size and bpp information to the image table.


Ryujinx.Graphics/NvGpuEngine3d.cs
```diff
@@ -214,11 +214,17 @@ private void SetZeta(NvGpuVmm Vmm)
         {
             long ZA = MakeInt64From2xInt32(NvGpuEngine3dReg.ZetaAddress);
```
ReinUsesLisp: Not related to these changes, but `ZA` -> `VA`, I didn't realize it stood for virtual address at that time.

Ryujinx.Graphics/Gal/IGalRenderTarget.cs
```diff
@@ -40,12 +36,6 @@ public interface IGalRenderTarget
...
             int  DstY1);
 
-        void GetBufferData(long Key, Action<byte[]> Callback);
-
-        void SetBufferData(
-            long             Key,
-            int              Width,
-            int              Height,
-            byte[]           Buffer);
+        byte[] GetData(long Key);
```
gdkchan: I think it would be better to add that on `OGLTexture`, but it needs to use a dummy framebuffer to bind the texture and then use `glReadPixels`, so I think this is the best place for now.

Ryujinx.Graphics/Texture/ImageUtils.cs
```diff
@@ -197,108 +206,113 @@ public static GalImageFormat ConvertZeta(GalZetaFormat Format)
...
             throw new NotImplementedException(Format.ToString());
         }
 
-        public static TextureReaderDelegate GetReader(GalImageFormat Format)
+        public static byte[] ReadTexture(IAMemory Memory, GalImage Image, long Position)
         {
-            return GetImageDescriptor(Format).Reader;
+            AMemory CpuMemory;
+
+            if (Memory is NvGpuVmm Vmm)
```
gdkchan: It's unfortunate that we need that, it's a complication caused by the fact that `NvGpuVmm` depends on `AMemory`. In the future I plan to make it independent (and hopefully drop `ChocolArm64` dep on `Ryujinx.Graphics`) like I said so this code should go away.

gdkchan:
This is ready for review/testing.

Pending work that I don't plan to do on this PR:
- Remove ChocolArm64 dep on Ryujinx.Graphics, cleanup memory related code.
- Fix some texture formats that are not correct. Some also needs reversing (like rgba5551).
- Also fix inconsistencies on texture format names.
- Simplify `Create` methods on `OGLTexture` to share as much code as possible.
- Improve reinterpretation code (still doesn't work properly with all games), @ReinUsesLisp plans to work on that I think.

Ryujinx.Graphics/Gal/GalTextureFormat.cs
```diff
@@ -7,6 +7,7 @@ public enum GalTextureFormat
         R32G32       = 0x4,
         A8B8G8R8     = 0x8,
         A2B10G10R10  = 0x9,
+        R16_G16      = 0xc,
```
ReinUsesLisp: I know image formats are messed up, but R16_G16 is not formatted like the other GalTextureFormats

PR #425, opened 2018-09-19
Implement DepthWriteMask and add R16G16

ReinUsesLisp:
I couldn't use GalPipelineState here because glClear* requires glDepthMask to be true


gdkchan:
can't you read `DepthWriteEnable` from `GalPipelineState` from `OGLRasterizer`?

ReinUsesLisp:
Yes I can do that. To keep code consistent, should I pass IGalRenderer to every OGL*?

ReinUsesLisp:
Addressed.

PR #427, opened 2018-09-20
Fix multiple rendertargets

ReinUsesLisp:
`RT_SEPARATE_FRAG_DATA` doesn't seem to be set to commercial games. Until we found out why (probably its set by the OS?), this considers it always being on. I also simplified how OpenGL framebuffers are being bound, but it ends up with more superfluous OpenGL calls.
This requires [ARB_viewport_array](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_viewport_array.txt) (or OpenGL 4.1) to set individual viewports per attachment.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
@@ -7,33 +7,23 @@ namespace Ryujinx.Graphics.Gal.OpenGL
         private static bool Initialized = false;
 
         private static bool EnhancedLayouts;
-
+        private static bool ViewportArray;
         private static bool TextureMirrorClamp;
 
-        public static bool HasEnhancedLayouts()
+        public static bool HasEnhancedLayouts()    => Query(ref EnhancedLayouts);
+        public static bool HasViewportArray()      => Query(ref ViewportArray);
+        public static bool HasTextureMirrorClamp() => Query(ref TextureMirrorClamp);
+        
+        private static bool Query(ref bool Extension)
         {
-            EnsureInitialized();
-
-            return EnhancedLayouts;
-        }
-
-        public static bool HasTextureMirrorClamp()
-        {
-            EnsureInitialized();
-
-            return TextureMirrorClamp;
-        }
-
-        private static void EnsureInitialized()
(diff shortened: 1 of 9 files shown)


gdkchan:
glViewportArray (and glViewportIndexed) are only available starting at ogl 4.1. It would be nice if we could check extension and fallback to the old way if it is not supported, to keep compatibility for users with older hardware.

ReinUsesLisp:
Addressed

Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
```diff
@@ -6,21 +6,21 @@ static class OGLExtension
...
+        private static bool s_EnhancedLayouts;
+        private static bool s_ViewportArray;
+        private static bool s_TextureMirrorClamp;
 
-        public static bool HasEnhancedLayouts()    => Query(ref EnhancedLayouts);
-        public static bool HasViewportArray()      => Query(ref ViewportArray);
-        public static bool HasTextureMirrorClamp() => Query(ref TextureMirrorClamp);
+        public static bool EnhancedLayouts    => Query(ref s_EnhancedLayouts);
+        public static bool ViewportArray      => Query(ref s_ViewportArray);
+        public static bool TextureMirrorClamp => Query(ref s_TextureMirrorClamp);
```
jduncanator: You don't need to pass these as `ref` (in fact you can drop `ref` everywhere in this class). You don't modify any of the parameters inside `Query`, you modify their backing fields directly.

Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
```diff
@@ -6,21 +6,21 @@ static class OGLExtension
...
         
         private static bool Query(ref bool Extension)
```
jduncanator: Per previous comment, you can drop `ref` here.

Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
jduncanator: Yep, you're correct. My bad.

PR #429, opened 2018-09-21
Implement friend:a IFriendService: 10101

AcK77:
As I said in Ryujinx/Ryujinx-Games-List#42, ARMS need this call. Since it's an online related thing, we can partially stub it for now ( as we haven't got friends :'( ).
I have refactored other calls a little to get more informations about them, and added symbols too.

I have updated Profile Picture.


Ryujinx.HLE/HOS/Services/Friend/IFriendServiceTypes.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Ryujinx.HLE.HOS.Services.Friend
+{
+    public enum PresenceStatusFilter
```
gdkchan: This shouldn't be `public`.

Ryujinx.HLE/HOS/Services/Friend/IFriendServiceTypes.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        None,
+        Online,
+        OnlinePlay,
+        OnlineOrOnlinePlay
+    }
+
+    public struct FriendFilter
```
gdkchan: This shouldn't be `public`.

Ryujinx.HLE/HOS/Services/Friend/IFriendService.cs
```diff
@@ -31,17 +70,28 @@ public long DeclareCloseOnlinePlaySession(ServiceCtx Context)
...
                 Context.RequestData.ReadInt64());
 
-            //TODO.
-            Context.Device.Log.PrintStub(LogClass.ServiceFriend, "Stubbed.");
+            long Unknown0 = Context.RequestData.ReadInt64();
+
+            long Position = Context.Request.PtrBuff[0].Position;
+            long Size     = Context.Request.PtrBuff[0].Size;
+
+            //Todo: Write the nn::friends::detail::UserPresenceImpl inside the buffer.
```
gdkchan: The name `nn::friends::detail::UserPresenceImpl` is most likely wrong, it looks like a function name.

gdkchan:
We usually create a separate file for each struct, instead of adding multiple structs to a single file.

AcK77:
Since there is just one `enum` and one `struct` I guess one file is enought for now... We can change that later if we will use more `struct` as you have done inside `AudioRenderer`!

At least, other things are fixed!

Ryujinx.HLE/HOS/Services/Friend/IFriendServiceTypes.cs
AcK77: If I put it `private` I got `inaccessible due to its protection`

Ryujinx.HLE/HOS/Services/Friend/IFriendServiceTypes.cs
AcK77: If I put it `private` I got `inaccessible due to its protection`

jduncanator:
Why was the profile picture changed?

AcK77:
Because we spoke about that with gdkchan and we are agreed to say the current one don’t really look good... So why not? It’s better than the older and I don’t want to do one PR just for a picture too.

EmulationChannel:
anotacao 70

PR #430, opened 2018-09-22
Improve kernel events implementation

gdkchan:
This PR:

- Improves the KProcessHandleTable implementation, should be faster and closer to the real thing.
- Adds missing event svcs (CreateEvent and SignalEvent), plus fixes/improves the ones that were already implemented.
- Adds missing KWritableEvent and KReadableEvent objects, the former being used by services, and the later being passed to the application.
- Starts refactoring the svc code, instead of using the `MakeError` method, use a simple enum with kernel error codes. I think this is way better. Additionally, split svc methods to facilitate 32-bits support later. To keep the PR small I only modified the new or corrected methods, I plan to do the rest later.

### DIFF
--- a/Ryujinx.HLE/HOS/Process.cs
+++ b/Ryujinx.HLE/HOS/Process.cs
@@ -71,7 +71,22 @@ namespace Ryujinx.HLE.HOS
 
             TlsPages = new List<KTlsPageManager>();
 
-            HandleTable = new KProcessHandleTable(Device.System);
+            int HandleTableSize = 1024;
+
+            if (MetaData != null)
+            {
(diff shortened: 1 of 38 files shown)


Ryujinx.HLE/HOS/Process.cs
```diff
@@ -71,7 +71,7 @@ public Process(Switch Device, int ProcessId, Npdm MetaData)
 
             TlsPages = new List<KTlsPageManager>();
 
-            HandleTable = new KProcessHandleTable();
+            HandleTable = new KProcessHandleTable(Device.System);
```
Thog: Might be nice to use the data from the NPDM for the HandleTable size.

PR #431, opened 2018-09-23
Implement a conservative cache

ReinUsesLisp:
Instead of destroying unused resources, this PR saves them for future recycle when they are compatible. Core functionality (`ResourcePool.cs` and `OGLCachedResource`) should be working but I still have to clean up its users (e.g.: `OGLTexture`), there are lots of repeated code there.
It should give a performance boost in almost every game that uses GPU resources.
Originally it was planned to recycle compatible textures, but it also constant buffers, vertex and index cache (these last two could be unified).

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
@@ -30,19 +30,11 @@ namespace Ryujinx.Graphics.Gal.OpenGL
             ResourcePool<TKey, TValue>.CreateValue CreateValueCallback,
             ResourcePool<TKey, TValue>.DeleteValue DeleteValueCallback)
         {
-            if (CreateValueCallback == null)
-            {
-                throw new ArgumentNullException(nameof(CreateValueCallback));
-            }
-
-            if (DeleteValueCallback == null)
-            {
-                throw new ArgumentNullException(nameof(DeleteValueCallback));
-            }
-
             Cache = new Dictionary<long, CacheBucket>();
 
-            Pool = new ResourcePool<TKey, TValue>(CreateValueCallback, DeleteValueCallback);
+            Pool = new ResourcePool<TKey, TValue>(
+                CreateValueCallback ?? throw new ArgumentNullException(nameof(CreateValueCallback)),
+                DeleteValueCallback ?? throw new ArgumentNullException(nameof(DeleteValueCallback)));
         }
 
         public void Lock()
@@ -82,7 +74,7 @@ namespace Ryujinx.Graphics.Gal.OpenGL
             {
                 Value = Bucket.Value;
 
-                Value.UpdateStamp();
+                Value.UpdateTimestamp();
 
                 return true;
             }
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
@@ -27,8 +27,8 @@ namespace Ryujinx.Graphics.Gal.OpenGL
         private bool Locked;
 
         public OGLCachedResource(
-            ResourcePool<TKey, TValue>.CreateValue CreateValueCallback,
-            ResourcePool<TKey, TValue>.DeleteValue DeleteValueCallback)
+            Func<TKey, TValue> CreateValueCallback,
+            Action<TValue>     DeleteValueCallback)
         {
             Cache = new Dictionary<long, CacheBucket>();
 
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
@@ -3,57 +3,46 @@ using System.Collections.Generic;
 
 namespace Ryujinx.Graphics.Gal.OpenGL
 {
-    class OGLCachedResource<T>
+    class OGLCachedResource<TKey, TValue>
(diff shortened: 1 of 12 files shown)


Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
```diff
@@ -3,57 +3,46 @@
...
+        private ResourcePool<TKey, TValue> Pool;
 
         private bool Locked;
 
-        public OGLCachedResource(DeleteValue DeleteValueCallback)
+        public OGLCachedResource(
+            ResourcePool<TKey, TValue>.CreateValue CreateValueCallback,
+            ResourcePool<TKey, TValue>.DeleteValue DeleteValueCallback)
         {
+            if (CreateValueCallback == null)
```
gdkchan: I think the null checks can be simplified, for example you can move it to the ResourcePool constructor call:
```cs
Pool = new ResourcePool<TKey, TValue>(
    CreateValueCallback ?? throw new ArgumentNullException(nameof(CreateValueCallback)),
    DeleteValueCallback ?? throw new ArgumentNullException(nameof(DeleteValueCallback)));
```

Ryujinx.Graphics/ResourcePool.cs
```diff
@@ -0,0 +1,153 @@
...
+        bool IsCompatible(T Other);
+    }
+
+    public class Resource
+    {
+        public int Timestamp { get; private set; }
+
+        public bool IsUsed { get; private set; }
+
+        public void UpdateStamp()
```
gdkchan: imo `UpdateTimestamp` is a better name.

Ryujinx.Graphics/ResourcePool.cs
```diff
@@ -0,0 +1,153 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.Graphics
+{
+    public interface ICompatible<T>
```
gdkchan: And reason for this and `Resource` being `public`?

Ryujinx.Graphics/ResourcePool.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                List<TValue> Family = Tuple.Family;
+
+                if (!Resource.IsUsed)
+                {
+                    int TimeDelta = RingDelta(Resource.Timestamp, Timestamp);
+
+                    if ((uint)TimeDelta > MaxTimeDelta)
+                    {
+                        if (!Family.Remove(Resource))
```
gdkchan: Keep in mind that removals from a List are slow. If this is not something that happens often, or the lists are usually small, then it shouldn't be a problem though.

Ryujinx.Graphics/ResourcePool.cs
```diff
@@ -0,0 +1,153 @@
...
+            this.DeleteValueCallback = DeleteValueCallback;
+
+            Entries = new List<(TKey, List<TValue>)>();
+
+            SortedCache = new Queue<(TValue, List<TValue>)>();
+        }
+
+        public TValue CreateOrRecycle(TKey Params)
+        {
+            List<TValue> Family = GetOrAddEntry(Params);
```
gdkchan: Hmm, `Family` is kind of a weird name for that imo, but I can't think on a good name right now... maybe siblings?

Ryujinx.Graphics/ResourcePool.cs
ReinUsesLisp: I plan to move them to `Ryujinx.Common` because they are not a graphics-specific class.

Ryujinx.Graphics/ResourcePool.cs
ReinUsesLisp: Should I use a linked list instead? Access is not an issue since it's always O(n), and it decreases removal/inserts to O(1).

Ryujinx.Graphics/ResourcePool.cs
ReinUsesLisp: `Family` as family of resources but `Siblings` is ok, both are better than former `Values`.

Ryujinx.Graphics/ResourcePool.cs
gdkchan: A linked list should be faster, if this method is executed often and/or lists have a considerable amount of items, then its worth to do the optmization.

ReinUsesLisp: Resource removal is always slow, but there's no need to make it slower.

Ryujinx.Graphics/ResourcePool.cs
```diff
@@ -0,0 +1,161 @@
...
+        where TValue : Resource
+    {
+        private const int MaxTimeDelta      = 5 * 60000;
+        private const int MaxRemovalsPerRun = 10;
+
+        public delegate TValue CreateValue(TKey Params);
+
+        public delegate void DeleteValue(TValue Resource);
+
+        private LinkedList<(TKey, LinkedList<TValue>)> Entries;
```
jduncanator: I'd make these tuples a static type. Tuples have their uses, but they're used so extensively in this class that they warrant their own type. Something like the following would do:

```cs
struct ResourcePoolKey<T1, T2>
```

`(TKey, LinkedList<TValue>)` could be replaced with `new ResourcePoolKey<TKey, TValue>`
`(TValue, LinkedList<TValue>)` could be replaced with `new ResourcePoolKey<TValue, TValue>`

jduncanator:
This lgtm 👀 👍

jduncanator:
Closing as superseded by #533

PR #432, opened 2018-09-23
Misc content loading improvements

Thealexbarney:
Open patched games if opening an XCI with a patch
Update to LibHac 0.1.2
Enable RomFS validation
Allow loading an NCA without a RomFS


jduncanator:
This looks fine to me. I'd recommend exposing `EnableFsIntegrityChecks` in the configuration file so users can disable it if needed/wanted.

emmauss:
lgtm

PR #433, opened 2018-09-23
Tweak cpu cache deletion policy values

gdkchan:
Previous values were too strict and resulted in functions being deleted too often and in stuttering on some games, this improves that.

- Minimum time since last use for deletion was increased from 30 seconds to 2 minutes.
- Max cache size (in ARM code size, not IL or x86 code size unfortunately) was increased from 2 MB to 4 MB.
- Minimum call count for updating cache entries timestamp was decreased from 1000 calls to 250 calls.

### DIFF
--- a/ChocolArm64/ATranslatorCache.cs
+++ b/ChocolArm64/ATranslatorCache.cs
@@ -8,9 +8,9 @@ namespace ChocolArm64
 {
     class ATranslatorCache
     {
-        private const int MaxTotalSize          = 2 * 1024 * 256;
-        private const int MaxTimeDelta          = 30000;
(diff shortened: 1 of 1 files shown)


jduncanator:
I've tested this and it lgtm. I couldn't reproduce any of the mentioned stuttering (at least not in the games that I tested), so can't validate that this fixes the problem, but it doesn't introduce any regressions.

My only gripe would be some comments on the seemingly "unit-less" constants at the top of the file. It's not immediately obvious why `4 * 1024 * 256` unless you know that an ARM64 instruction is 4 bytes wide, so maybe some explanation there would be warranted/

PR #434, opened 2018-09-27
Save Common implementation

emmauss:
This creates a `savecommon` directory for temporary savedata storage, used currently by the NSO NES games. I currently do not know the exact path for save common on console, so any comment about that before this is merged would be good. I also added an optional check for 0 userids, as savecommon uses a 0 userid.


Ryujinx.HLE/HOS/SystemState/UserId.cs
```diff
@@ -13,11 +13,6 @@ public struct UserId
 
         public UserId(long Low, long High)
         {
-            if ((Low | High) == 0)
```
gdkchan: I don't think those checks should be removed, 0 user ids are NOT valid. In this function it is just used as a special value, but the user id itself doesn't exist and is not used at all when this value is 0.

Ryujinx.HLE/HOS/SystemState/UserId.cs
```diff
@@ -50,11 +45,6 @@ public UserId(string UserIdHex)
                 throw new ArgumentException("Invalid user id!", nameof(UserIdHex));
             }
 
-            if (UserIdHex == "00000000000000000000000000000000")
```
gdkchan: Same thing here.

gdkchan:
If you're going to remove the checks, maybe its a good idea to also rename the struct to `UInt128` or something like that, since theres nothing special about it anyway and it can be used to store any 128-bits integer.

emmauss:
Renamed the struct.

PR #435, opened 2018-09-27
Include a more user-friendly key dumping guide

firewolf24:


AcK77:
There is too many way to found keys... Replace a way by another way, it's not really what we need here.
I think it's better to close this PR until we don't get a better guide who explain all the way to get keys.
Thanks

PR #441, opened 2018-10-06
[WIP] Qt Quick based GUI

emmauss:
Another attempt at making a native gui for ryujinx. Currently, only loading and shutting down games are implemented. 
The pros of this toolkit over one used in my previous attempts is mainly due to its native look, and the options for styling. Also, a lot less code needs to be used for close to what one would need with gtk and imgui for designs, code is easy to read, and libraries can be deployed with binaries without source.


maikpla:
need a rebase

PR #443, opened 2018-10-07
NVDEC implementation using FFmpeg

gdkchan:
This implements the Nvidia Decoder (nvdec) using the FFmpeg open source library. Initially it seemed to be the best route, however due to the way how the decoder works, I think FFmpeg is not really suited for that.

The decoder receives a decoder context with all information necessary to decode a single frame, it doesn't store any context internally. This means that, in theory at least, its possible to decode multiple videos by just swapping the context / modifying the state appropriately. However, FFmpeg does store the context internally, and we have no control over that.

The application parses the headers and extracts the relevant data. For H264, it's a single header (the Sequence Parameter Set and the Picture Parameter Set), for VP9 however, things are a bit more complicated since each frame contains a header (two actually, called on the spec as Uncompressed and Compressed header, the later using a Range Coder). To pass this information to FFmpeg we need to re-encode the header, which is sub-optimal and error prone. Ideally, we should be able to use the already extracted data directly.

Another problem also related to the VP9 codec is that each frame can reference and re-use macro blocks from 3 previously decoded frames, thus saving some space. This is passed to nvdec by setting the respective addresses for each reference frame on the registers. However, ffmpeg will store it internally, when a flag on the header signals that it should be saved. The problem is that we don't know which frames are going to be used in the future for reference, so we don't know which flags to set.

So, this is a experimental implementation using FFmpeg, however it is sub-optimal and problematic, so I suggest that we should use another approach, here are some options:
- Building our own decoders. Assuming that they are going to be written in C#, it will most likely going to be too slow, and will also require a lot of work to get working.
- Using a modified version of FFmpeg, that allows passing the extracted frame header data directly, bypassing the header parsing step entirely. This is probably the best option.
- Going with a more LLE approach and emulating the Falcon microcode. Probably the slowest of all the solutions, probably has some challenges I don't know about yet, will require more REing.

The VP9 decoder is not fully working yet, although it can decode keyframes successfully. Getting it to work with an unmodified FFmpeg is almost impossible due to the aforementioned issues. The H264 decoder is simpler and seems to be working properly, at least for the basic use where a single video is decoded at once.

So I'm mainly looking for feedback, what we should do from here, etc. It may be worth to get this in in the meantime. At least most of the required RE work should be already done.

List of codecs supported by official games:
- VP8 (not implemented)
- VP9 (partial implementation)
- H264 (working)
- H265 (not implemented)

Note: You'll need ffmpeg to run it. If you're on windows, you can use the zeranoe builds.


Ryujinx.Graphics/VideoDecoding/H264BitStreamWriter.cs
```diff
@@ -0,0 +1,79 @@
...
+            WriteBits(1, Size);
+
+            Value -= (1u << (Size - 1)) - 1;
+
+            WriteBits((int)Value, Size - 1);
+        }
+
+        private static readonly byte[] ClzNibbleTbl = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
+
+        private static int CountLeadingZeros(int Value)
```
jduncanator: Core 2.1 has a `clz` intrinsic. Would be even better if we could use `bsr` instruction for greater compatibility.

gdkchan: Ok, I will add it. However I won't be able to test as my CPU doesn't support it.

gdkchan:
Rebased.

gdkchan:
I discovered that the G8R8 texture format was wrong. For some reason it would invert the channels and then change it to R8G8. However the OpenGL format used (Rg8) is really G8R8, since in this case the first channel (R) is on the least significant byte (see the A8B8G8R8 format for example). R16G16 is also wrong but I plan to fix that on a different PR. I only fixed this one here because it affects videos.

gdkchan:
It seems that Nintendo optimized access to decoded video image on newer games, instead of doing a copy and returning the data to the game (and converting from yuv420p to rgba8 if requested on the cpu), it now uses the output surface address directly as texture, and uses the rgba8 format directly on vic (rather than doing a converison on the cpu). To support those games I added the rgba8 format aswell. There are a lot of other formats that are not yet supported, but afaik games only uses those two.

gdkchan:
Rebased and removed the WIP status (even though its incomplete, but I don't plan to work for now).

gdkchan:
Rebased for the 6th time.

PR #445, opened 2018-10-08
Implement IRoInterface

Thog:
This is required by Super Mario Party.
This PR also adds MapProcessCodeMemory and UnmapProcessCodeMemory functions in KMemoryManager. Those two calls might not reflect what the SVC of the same names do.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
@@ -178,6 +178,7 @@ namespace Ryujinx.HLE.HOS.Kernel
                     InsertBlock(Src, PagesCount, MemoryState.Heap, MemoryPermission.None);
 
                     CpuMemory.Map(Dst, PA, Size);
+
                     return 0;
                 }
             }
@@ -225,6 +226,7 @@ namespace Ryujinx.HLE.HOS.Kernel
                     InsertBlock(Src, PagesCount, MemoryState.Heap, MemoryPermission.ReadAndWrite);
 
                     CpuMemory.Unmap(Dst, Size);
+
                     return 0;
                 }
             }
@@ -847,6 +849,7 @@ namespace Ryujinx.HLE.HOS.Kernel
             {
                 Result = IsUnmapped(Position, Size);
             }
+
             return Result;
         }
 
--- a/Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
@@ -148,6 +148,90 @@ namespace Ryujinx.HLE.HOS.Kernel
             }
         }
 
+        public long MapProcessCodeMemory(long Dst, long Src, long Size)
+        {
+            lock (Blocks)
+            {
+                long PagesCount = Size / PageSize;
+
+                bool Success = IsUnmapped(Dst, Size);
+
+                Success &= CheckRange(
+                            Src,
+                            Size,
+                            MemoryState.Mask,
+                            MemoryState.Heap,
+                            MemoryPermission.Mask,
+                            MemoryPermission.ReadAndWrite,
+                            MemoryAttribute.Mask,
+                            MemoryAttribute.None,
+                            MemoryAttribute.IpcAndDeviceMapped,
+                            out _,
+                            out _,
+                            out _);
+
+                if (Success)
+                {
+                    long PA = CpuMemory.GetPhysicalAddress(Src);
+
+                    InsertBlock(Dst, PagesCount, MemoryState.CodeStatic, MemoryPermission.ReadAndExecute);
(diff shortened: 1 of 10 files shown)


Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
```diff
@@ -148,6 +148,85 @@ public void HleMapProcessCode(long Position, long Size)
             }
         }
 
+        public long MapProcessCodeMemory(long Dst, long Src, long Size)
+        {
+            long PagesCount = Size / PageSize;
+
+            if (!IsUnmapped(Dst, Size) || !CheckRange(
```
gdkchan: Calls to `CheckRange`, `IsUnmapped` etc should be inside the lock aswell, otherwise it may iterate over the block list while it is being modified. Also the identation for the arguments seems to be off (compared to the other methods).

Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
```diff
@@ -148,6 +148,85 @@ public void HleMapProcessCode(long Position, long Size)
...
+            }
+
+            return 0;
+        }
+
+        public long UnmapProcessCodeMemory(long Dst, long Src, long Size)
+        {
+            long PagesCount = Size / PageSize;
+
+            if (!CheckRange(
```
gdkchan: IMO is looks better to do the two calls separately, and store the results on a bool. See what I did on the other methods with a `bool Success` variable.

Ryujinx.HLE/HOS/Kernel/KMemoryManager.cs
```diff
@@ -755,7 +834,7 @@ private void FreePages(long Position, long PagesCount)
             }
         }
 
-        private bool IsUnmapped(long Position, long Size)
+        public bool IsUnmapped(long Position, long Size)
```
gdkchan: The `CheckRange` call is not under a lock, calling it from outside may lead to race conditions.

Ryujinx.HLE/HOS/Services/Ldr/IRoInterface.cs
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+
+    class NroInfo
+    {
+        public Nro    Executable { get; private set; }
+        public byte[] Hash       { get; private set; }
+        public long   NroAddress { get; private set; }
+        public long   TotalSize  { get; private set; }
+        public long   NroMappedAddress;
```
gdkchan: Why this one is not a property unlike all the others?

Ryujinx.HLE/HOS/Services/Ldr/IRoInterface.cs
```diff
@@ -0,0 +1,447 @@
...
+
+
```
gdkchan: This empty line is not necessary.

Ryujinx.HLE/HOS/Services/Ldr/IRoInterface.cs
```diff
@@ -0,0 +1,447 @@
...
+            NroInfos = new List<NroInfo>(MaxNro);
+        }
+
+        private long ParseNrr(out NrrInfo NrrInfo, ServiceCtx Context, long NrrAddress, long NrrSize)
+        {
+            NrrInfo = null;
+
+            if (NrrSize == 0 || NrrAddress + NrrSize <= NrrAddress || (NrrSize & 0xFFF) != 0)
+            {
+                return 0xA409;
```
gdkchan: It would be nice to at least have some constants (with names) instead of those "magic" values here.

Ryujinx.HLE/HOS/Services/Ldr/IRoInterface.cs
```diff
@@ -0,0 +1,447 @@
...
+            {
+                foreach (byte[] Hash in Info.Hashes)
+                {
+                    if (Hash.SequenceEqual(NroHash))
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
```
gdkchan: Some methods have an empty new line before the return, other doesn't. It would be nice to be consistent.

Ryujinx.HLE/HOS/Services/Ldr/IRoInterface.cs
```diff
@@ -0,0 +1,447 @@
...
+
+                Result = RemoveNrrInfo(NrrHeapAddress);
+            }
+            return Result;
+        }
+
+        // Initialize(u64, pid, KObject)
+        public long Initialize(ServiceCtx Context)
+        {
+            // TODO: we actually ignore the pid and process handle receive, we will need to use them when we will have multi process support.
```
gdkchan: There are a lot of things that are ignored with the current IPC system, if we decide to do a more accurate impl I believe that a lot of boring rewrite will be needed on the HLE services impl.

PR #446, opened 2018-10-10
Add GetAddOnContentListChangedEvent

emmauss:
Also moved aoc to its own namespace. Name says it returns an event, swipc and switchbrew says nothing of the return.


Ryujinx.HLE/HOS/Services/Aoc/IAddOnContentManager.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            AddOnContentListChangedEvent = new KEvent(System);
+        }
+
+        // CountAddOnContent(u64, pid) -> i32
+        public static long CountAddOnContent(ServiceCtx Context)
+        {
+            Context.ResponseData.Write(0);
+
+            Context.Device.Log.PrintStub(LogClass.ServiceNs, "Stubbed.");
```
gdkchan: Shouldn't this use ServiceAoc to match the namespace?

emmauss: okay

Ryujinx.HLE/HOS/Services/Aoc/IAddOnContentManager.cs
```diff
@@ -0,0 +1,64 @@
...
+                { 8, GetAddOnContentListChangedEvent }
+            };
+
+            AddOnContentListChangedEvent = new KEvent(System);
+        }
+
+        // CountAddOnContent(u64, pid) -> i32
+        public static long CountAddOnContent(ServiceCtx Context)
+        {
+            Context.ResponseData.Write(0);
```
AcK77: Maybe you should explicit the `long` here and then modify the `PrintStub` with the value (even if it's 0, it could be a good thing to know that directly in the log).

emmauss:
addressed comments and rebased

Ryujinx.HLE/HOS/Services/Aoc/IAddOnContentManager.cs
```diff
@@ -0,0 +1,64 @@
...
+                { 8, GetAddOnContentListChangedEvent }
+            };
+
+            AddOnContentListChangedEvent = new KEvent(System);
+        }
+
+        // CountAddOnContent(u64, pid) -> i32
+        public static long CountAddOnContent(ServiceCtx Context)
+        {
+            Context.ResponseData.Write((long)0);
```
jduncanator: The signature in the comment above says this should be writing a int (32-bit), yet you're writing a long (64-bit)? Whichever one is incorrect should be updated.

PR #447, opened 2018-10-11
Quads, QuadStrip, const attributes and half-float attributes support

gdkchan:
This PR:
- Fixes Quads and QuadStrip primitive modes on DrawArrays. It was using the respective OpenGL primitive modes, however they were deprecated on OpenGL 3.x, so it doesn't work anymore. I replaced it with multiple calls to DrawArrays with an offset, and a count of 4. It's not optimal, however the only other option I can think of is creating another vertex buffer, or using a index buffer. Draw Elements now doesn't support Quads or QuadStrip, it should throw an exception if those two are used.
- Allow constant attributes to work properly. When the stride is 0, it should read the attribute from the same location every time, effectively being a constant value, however on OpenGL the value 0 being used as stride is special, it assumes that the data is tightly packed.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
@@ -13,7 +13,7 @@ namespace Ryujinx.Graphics.Gal.OpenGL
                 case GalFrontFace.CCW: return FrontFaceDirection.Ccw;
             }
 
-            throw new ArgumentException(nameof(FrontFace));
+            throw new ArgumentException(nameof(FrontFace) + " \"" + FrontFace + "\" is not valid!");
         }
 
         public static CullFaceMode GetCullFace(GalCullFace CullFace)
@@ -25,7 +25,7 @@ namespace Ryujinx.Graphics.Gal.OpenGL
                 case GalCullFace.FrontAndBack: return CullFaceMode.FrontAndBack;
             }
 
-            throw new ArgumentException(nameof(CullFace));
+            throw new ArgumentException(nameof(CullFace) + " \"" + CullFace + "\" is not valid!");
         }
 
         public static StencilOp GetStencilOp(GalStencilOp Op)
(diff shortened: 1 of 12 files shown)


gdkchan:
Added Quads/QuadStrip primitive support on indexed draws aswell, it creates a new index buffer with the same format, but with the 4 points quad indices or 4 + 2 + 2 + 2... (QuadStrip) points quad indices into 3 + 3 points for 2 triangles indices.

It's mostly unstested.

gdkchan:
Added two more fixes (not related to the previous work):
- Support half float types. Previously it would just use a 32-bits float which is not correct. Also throw exceptions for types that are not supported instead of using the wrong type in some cases.
- Properly align the pitch value as required by the Switch GPU, also use the Pitch value from the TIC when possible.

Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
```diff
@@ -13,7 +13,7 @@ public static FrontFaceDirection GetFrontFace(GalFrontFace FrontFace)
                 case GalFrontFace.CCW: return FrontFaceDirection.Ccw;
             }
 
-            throw new ArgumentException(nameof(FrontFace));
+            throw new ArgumentException(nameof(FrontFace) + " \"" + FrontFace + "\" is not valid!");
```
jduncanator: Using string interpolation for exception messages would make these easier to read.

```
throw new ArgumentException($"{nameof(FrontFace)} \"{FrontFace}\" is not valid!");
```

I also tend to shy away from using exclamation marks on the end of exception messages in place of periods, however that's a personal preference.

jduncanator:
This looks good. I've tested the branch, and on a relatively old AMD driver version, these all seem to work correctly. Apart from the nit regarding the exception messages, this looks fine to merge!

jduncanator:
Ignore my comments, I had a tab open from prior to this getting merged, and did not notice it was had already been merged.

PR #449, opened 2018-10-12
Add Fmls_Se, Fmulx_Se/Ve, Smov_S Inst.; Opt. Clz/Clz_V, Cnt_V, Shl_V, S/Ushr_V, S/Usra_V Inst.; Add 11 Tests. Some fixes.

LDj3SNuD:
**Instructions added:**
  Fmls_Se
  Fmulx_Se
  Fmulx_Ve
  Smov_S

**Decoder fixed:**
  AOpCodeSimdRegElemF class

**Tables fixed:**
  Fmla_Se
  Fmla_Ve
  Fmls_Ve
  Fmul_Se
  Fmul_Ve

**Optimizations (Intrinsics):**
  ASoftFallback class:
    CountLeadingZeros() -> Lzcnt.LeadingZeroCount()
    CountSetBits8() -> Popcnt.PopCount()
  AInstEmitSimdShift:
    Shl_V() -> Sse2.ShiftLeftLogical
    Sshr_V() -> Sse2.ShiftRightArithmetic
    Ushr_V() -> Sse2.ShiftRightLogical

**Tests added:**
  SimdIns Category:
    Smov_S_W/X
    Umov_S_W/X
  SimdRegElem Category:
    Mul_Ve
  _New SimdRegElemF Category added:_
    Fmla_Se
    Fmla_Ve
    Fmls_Se
    Fmls_Ve
    Fmul_Se
    Fmul_Ve
    Fmulx_Se
    Fmulx_Ve

**About FP Tests:**
  Now Explicit

Nits.


ChocolArm64/Decoder/AOpCodeSimdRegElemF.cs
```diff
@@ -8,15 +8,26 @@ class AOpCodeSimdRegElemF : AOpCodeSimdReg
 
         public AOpCodeSimdRegElemF(AInst Inst, long Position, int OpCode) : base(Inst, Position, OpCode)
         {
-            if ((Size & 1) != 0)
+            switch ((OpCode >> 21) & 3) // sz:L
```
gdkchan: I think they should both produce the same values for index, what's the advantage of this method over the old one?

ChocolArm64/Instruction/ASoftFallback.cs
```diff
@@ -405,33 +405,54 @@ public static ulong CountLeadingSigns(ulong Value, int Size)
 
         private static readonly byte[] ClzNibbleTbl = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
 
-        public static ulong CountLeadingZeros(ulong Value, int Size)
+        public static ulong CountLeadingZeros(ulong Value, int Size) // Size is 8, 16, 32 or 64 (SIMD&FP or Base Inst.).
```
gdkchan: I wonder if the JIT is smart enough to inline that when `Lzcnt` is available.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -365,8 +365,8 @@ public static float FPMaxNum(float Value1, float Value2, AThreadState State)
         {
             Debug.WriteIf(State.Fpcr != 0, "ASoftFloat_32.FPMaxNum: ");
 
-            Value1.FPUnpack(out FPType Type1, out bool Sign1, out uint Op1);
-            Value2.FPUnpack(out FPType Type2, out bool Sign2, out uint Op2);
+            Value1.FPUnpack(out FPType Type1, out bool _, out uint _);
```
gdkchan: I think when you use `_` specifying the variable type is not necessary, so you can replace `out bool _` with just `out _` etc.

ChocolArm64/Instruction/AInstEmitSimdShift.cs
```diff
@@ -31,12 +32,32 @@ public static void Shl_V(AILEmitterCtx Context)
...
+            if (AOptimizations.UseSse2 && Op.Size > 0)
             {
-                Context.EmitLdc_I4(GetImmShl(Op));
+                Type[] Types = new Type[] { VectorUIntTypesPerSizeLog2[Op.Size], typeof(byte) };
 
-                Context.Emit(OpCodes.Shl);
-            });
+                EmitLdvecWithUnsignedCast(Context, Op.Rn, Op.Size);
+
+	            Context.EmitLdc_I4(GetImmShl(Op));
```
gdkchan: nit: The identation here (and on the line below too) is off.

ChocolArm64/Instruction/AInstEmitSimdShift.cs
```diff
@@ -167,7 +188,30 @@ public static void Sshr_S(AILEmitterCtx Context)
...
+            AOpCodeSimdShImm Op = (AOpCodeSimdShImm)Context.CurrOp;
+
+            if (AOptimizations.UseSse2 && Op.Size > 0
+                                       && Op.Size < 3)
+            {
+                Type[] Types = new Type[] { VectorIntTypesPerSizeLog2[Op.Size], typeof(byte) };
+
+                EmitLdvecWithSignedCast(Context, Op.Rn, Op.Size);
+
+	            Context.EmitLdc_I4(GetImmShr(Op));
```
gdkchan: Ditto.

ChocolArm64/Instruction/AInstEmitSimdShift.cs
```diff
@@ -239,7 +283,29 @@ public static void Ushr_S(AILEmitterCtx Context)
...
-            EmitShrImmOp(Context, ShrImmFlags.VectorZx);
+            AOpCodeSimdShImm Op = (AOpCodeSimdShImm)Context.CurrOp;
+
+            if (AOptimizations.UseSse2 && Op.Size > 0)
+            {
+                Type[] Types = new Type[] { VectorUIntTypesPerSizeLog2[Op.Size], typeof(byte) };
+
+                EmitLdvecWithUnsignedCast(Context, Op.Rn, Op.Size);
+
+	            Context.EmitLdc_I4(GetImmShr(Op));
```
gdkchan: Ditto.

ChocolArm64/Decoder/AOpCodeSimdRegElemF.cs
LDj3SNuD: For make it similar to AOpCodeSimdRegElem, for "UnallocatedEncoding();" and for "if sz:Q == '10' then ReservedValue();" (see fmls_advsimd_elt.html, "Vector, single-precision and double-precision").

ChocolArm64/Instruction/ASoftFallback.cs
LDj3SNuD: What are you suggesting to do?

gdkchan: It might be better to call the intrinsic method on `Clz` directly, instead of calling it on the `CountLeadingZeros` helper method (assuming that the JIT is not capable of inlining it on its own). Another option is forcing inlining with `AggressiveInlining`, but I think in the case the intrinsic is not available and the fallback is taken, its not worth to inline and it will add quite a bunch of code on the caller.

LDj3SNuD: `CountLeadingZeros()` is called by `Clz` (Base Inst.) and by `Clz_V` (SIMD&FP Inst.).

LDj3SNuD: Address PR feedback.

PR #457, opened 2018-10-15
Improve texture tables

gdkchan:
This improves the texture tables, fixes a bunch of errors and inconsistencies.


gdkchan:
Summary of the changes:
- Renamed most formats. Naming was inconsistent, GalTextureFormat used names like A8B8G8R8 while GalSurfaceFormat used RGBA8. Component order was also inconsistent in some cases (like on the aforementioned example, ABGR vs RGBA, even though its the same format). This caused confusion and lead some people to implement the same format twice, and other errors.
- A bunch of formats were wrong on OGLEnumConverter (for example, 16-bits formats using UnsignedByte).
- In order to simplify the renaming process, I removed all unused formats from GalImageFormat. We can add them back as we need them.
- Represent sRGB as an extra flag rather than being part of the format itself, also add more sRGB formats.

Some changes are not directly related to the texture changes:
- Print shader number when a shader dump path is set, this helps debugging with external tools.
- Fix bogus branches when an SSY and BRA was mixed on the same basic block.
- Support multiple attachments on glColorMask.

Of course, I don't have tests for all those texture formats, so it may have introduced regressions, it would be nice if someone could test it with a wide range of games using multiple texture formats.

Ryujinx.Graphics/Gal/GalImageFormat.cs
```diff
@@ -5,89 +5,61 @@ namespace Ryujinx.Graphics.Gal
     [Flags]
     public enum GalImageFormat
     {
-        Snorm  = 1 << 27,
-        Unorm  = 1 << 28,
-        Sint   = 1 << 29,
-        Uint   = 1 << 30,
-        Sfloat = 1 << 31,
+        Astc2DStart,
```
jduncanator: `Astc2DStart` and `Astc2DEnd` are most likely left over idioms from C source code, used for iterating over enum values. This is a pattern that's rarely used in C#, so unless we have a specific use case in the code base, these can probably be removed.

ReinUsesLisp: I added those to easily find if an enum is ASTC or not. All values defined in that enum are arbitrary.

jduncanator: That's fair. If there is an actual use for them, then they should be left there!

gdkchan: Other than doing that, I see 3 others options:
- Doing the same check with Astc2D4x4 / Astc2D12x12. However if someone change the other of those items on the enum, then the check will stop working properly.
- Adding a huge `if` or `switch` checking for all the ASTC formats, this is slower than the current approach, and code is also larger.
- Adding another flag on the enum to endicate ASTC formats, and then number all ASTC formats like this:
```
Astc2D4x4 = Astc | 0,
Astc2D5x4 = Astc | 1,
Astc2D5x5 = Astc | 2,
```
Then we could just test that bit to check whenever the format is ASTC or not. This is slightly faster than the current approach, but the GalImageFormat enum will get larger and peharps that will hurt readability a bit.

gdkchan:
Another thing that may be worth changing here is the name of the other compressed formats. For example BC1, BC2 and BC3 are more commonly called DXT1, DXT3 and DXT5 respectively (and this is also how its called on OpenGL), while BC4 and BC5 are also know as RGTC1 and RGTC2. It may be worth to rename those since I already renamed BC6 and BC7.

PR #459, opened 2018-10-16
Remove IFileSystem::OpenDirectory extraneous check

Thog:
A directory can be open more than one time. This fix issues with
homebrews opening the same directory multiple time. (like ftpd)


gdkchan:
This is probably a copy/paste type of error, the code was most likely copied from OpenFile.

PR #461, opened 2018-10-16
Print stack trace on invalid memory accesses

gdkchan:


gdkchan:
Updated. I had to rebase because, even though it was showing as mergeable, the code was not compiling due to name changes.

Additionaly, I made the following modifications:
- Image base uses code region start again, however start address for 39 bits address spaces was changed from 0 to 0x8000000, this makes catching null dereference and the like possible.
- Print stack trace when a BRK instruction is hit aswell (although I have never seen it being hit).
- Print stack trace when undefined instructions are executed. May help debugging games that executes the "permanently undefined" instruction due to some error.

PR #467, opened 2018-10-20
Implement basic psm service

AcK77:
- Add `IPsmServer`
- Stub `GetBatteryChargePercentage` & `GetChargerType`


Ryujinx.HLE/HOS/Services/ServiceFactory.cs
```diff
@@ -146,6 +147,9 @@ public static IpcService MakeService(Horizon System, string Name)
                 case "pl:u":
                     return new ISharedFontManager();
 
+                case "psm":
```
gdkchan: Sorting here is wrong. Its supposed to be after prepo.

PR #468, opened 2018-10-21
Fix Fcvtl_V and Fcvtn_V; fix half to float conv. and add float to half conv. (full FP emu.). Add 4 FP Tests.

LDj3SNuD:


ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
-                // Denormal
-                x_mantissa <<= 1;
-                while ((x_mantissa & 0x0400) == 0)
+                State.Fpsr |= 1 << UFCBit;
+
+                return FPZero(Sign);
+            }
+
+            uint BiasedExp = (uint)Max(Exponent - MinimumExp + 1, 0);
+            if (BiasedExp == 0u) Mantissa /= Math.Pow(2d, MinimumExp - Exponent);
```
gdkchan: I think it's better for readability and consistency to place the if body on a new line, like:
```cs
if (BiasedExp == 0u)
{
    Mantissa /= Math.Pow(2d, MinimumExp - Exponent);
}
```

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
+            }
+
+            if (Error != 0d)
+            {
+                FPProcessException(FPExc.Inexact, State);
+            }
+
+            return Result;
+
+            int Max(int Left, int Right) => Left >= Right ? Left : Right;
```
gdkchan: Is this faster than using `Math.Max`? Also I think it's better as a separate method, not as a local method.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
+                throw new NotImplementedException("floating-point trap handling");
+            }
+            else
+            {
+                State.Fpsr |= 1 << (int)Exc;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static FPRounding FPRoundingMode(AThreadState State)
```
gdkchan: imo it would be better to add this directly to `AThreadState`.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
+        {
+            return (FPRounding)((State.Fpcr >> (int)FPCR.RMode) & 3);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool GetFpcrFlag(this AThreadState State, FPCR Fpcr)
```
gdkchan: imo it would be better to add this directly to `AThreadState`.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
+            }
+            else
+            {
+                ResultBits = FPRoundCV(Real, State);
+            }
+
+            return ResultBits;
+        }
+
+        private enum FPType
```
gdkchan: We shouldn't need to repeat those on every class. My suggestion is moving those enums to ChocolArm64.State, each one on a separate file.

ChocolArm64/Instruction/ASoftFloat.cs
```diff
@@ -195,41 +195,639 @@ public static double InvSqrtEstimate(double x)
...
+            else
+            {
+                Type = FPType.Nonzero; // Normal.
+                Real = Math.Pow(2d, (int)Exp32 - 127) * (1d + (double)Frac32 * Math.Pow(2d, -23));
+            }
+
+            return Sign ? -Real : Real;
+        }
+
+        private static ushort FPRoundCV(double Real, AThreadState State)
```
gdkchan: The comments on the other FPRoundCv method also applies here.

LDj3SNuD:
Address PR feedback.

PR #472, opened 2018-10-22
Add depth range support on the GPU

gdkchan:
There are multiple (16) depth range values, it assumes that only the first one is used for now.


Ryujinx.Graphics/Gal/OpenGL/OGLPipeline.cs
```diff
@@ -100,6 +100,8 @@ public OGLPipeline(OGLConstBuffer Buffer, OGLRasterizer Rasterizer, OGLShader Sh
                 DepthTestEnabled = false,
                 DepthWriteEnabled = true,
                 DepthFunc = GalComparisonOp.Less,
+                DepthRangeNear = 0,
+                DepthRangeFar = 1,
```
emmauss: align these

gdkchan: Again I was just following what was already there. But I can align everything.

PR #473, opened 2018-10-22
Fix Color Mask values

gdkchan:
This fixes the color mask common flag (it's not another depth mask value, its just used to switch between using just the first value or the entire array, at least according to nouveau). Additionally it initializes the first color mask to 1, 1, 1, 1 by default since some games never writes to this register.

### DIFF
--- a/Ryujinx.Graphics/Gal/GalPipelineState.cs
+++ b/Ryujinx.Graphics/Gal/GalPipelineState.cs
@@ -37,42 +37,42 @@
 
         public GalFrontFace FrontFace;
 
-        public bool CullFaceEnabled;
+        public bool        CullFaceEnabled;
         public GalCullFace CullFace;
 
-        public bool DepthTestEnabled;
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics/Gal/GalPipelineState.cs
```diff
@@ -72,7 +72,7 @@ public class GalPipelineState
         public GalBlendFactor BlendFuncSrcAlpha;
         public GalBlendFactor BlendFuncDstAlpha;
 
-        public ColorMaskRgba ColorMask;
+        public bool ColorMaskCommon;
         public ColorMaskRgba[] ColorMasks;
```
emmauss: I think this should be aligned or moved

gdkchan: Nothing inside this class is aligned, so I was just following that. But I can align all the fields too, if thats going to improve readability.

PR #479, opened 2018-10-24
Timing: Optimize Timestamp Aquisition

jduncanator:
Currently, we make use of `Environment.TickCount` in a number of places. This has some downsides, mainly being that the `TickCount` is a signed 32-bit integer, and has an effective limit of ~25 days before overflowing and wrapping around. Due to the signed-ness of the value, this also caused issues with negative numbers. This resolves these issues by using a 64-bit tick count obtained from Performance Counters (via the `Stopwatch` class). This has the favorable side effect of being significantly more accurate than the `TickCount`.

The biggest benefit of this change is that it ensures the monotonicity of the value returned (unless you leave your computer on for over 90,000 years).

I've left the member name as `ElapsedTicks`, because we originally used `TickCount`, however a more accurate representation would probably be `ElapsedMilliseconds` given we are returning a value in milliseconds and a "tick" from the underlying performance counter is actually a much smaller and more precise value; this could get confusing. If we renamed it, it would also allow us to expose the precise, accurate underlying tick count via `ElapsedTicks` instead of milliseconds. I'm happy to make these changes if you think this is a worthwhile.

### DIFF
--- a/ChocolArm64/ATranslatorCache.cs
+++ b/ChocolArm64/ATranslatorCache.cs
@@ -138,7 +138,7 @@ namespace ChocolArm64
 
                     CacheBucket Bucket = Cache[Node.Value];
 
-                    long TimeDelta = RingDelta(Bucket.Timestamp, Timestamp);
+                    long TimeDelta = Bucket.Timestamp - Timestamp;
 
                     if (TimeDelta <= MinTimeDelta)
                     {
@@ -159,12 +159,7 @@ namespace ChocolArm64
         {
             long timestamp = Stopwatch.GetTimestamp();
 
-            return (long)(timestamp * (1000.0f / Stopwatch.Frequency));
-        }
-
-        private static long RingDelta(long Old, long New)
-        {
-            return New - Old;
+            return timestamp / (Stopwatch.Frequency / 1000);
         }
     }
 }
--- a/ChocolArm64/ATranslatorCache.cs
+++ b/ChocolArm64/ATranslatorCache.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 
@@ -27,7 +28,7 @@ namespace ChocolArm64
 
             public int Size { get; private set; }
 
-            public int Timestamp { get; private set; }
+            public long Timestamp { get; private set; }
 
             public CacheBucket(ATranslatedSub Subroutine, LinkedListNode<long> Node, int Size)
             {
@@ -41,7 +42,7 @@ namespace ChocolArm64
             {
                 this.Node = Node;
 
-                Timestamp = Environment.TickCount;
+                Timestamp = GetTimestamp();
             }
         }
 
@@ -122,7 +123,7 @@ namespace ChocolArm64
 
         private void ClearCacheIfNeeded()
         {
-            int Timestamp = Environment.TickCount;
+            long Timestamp = GetTimestamp();
 
             while (TotalSize > MaxTotalSize)
             {
@@ -137,9 +138,9 @@ namespace ChocolArm64
 
                     CacheBucket Bucket = Cache[Node.Value];
 
-                    int TimeDelta = RingDelta(Bucket.Timestamp, Timestamp);
+                    long TimeDelta = RingDelta(Bucket.Timestamp, Timestamp);
 
-                    if ((uint)TimeDelta <= (uint)MinTimeDelta)
+                    if (TimeDelta <= MinTimeDelta)
                     {
                         break;
                     }
@@ -154,16 +155,16 @@ namespace ChocolArm64
             }
         }
 
-        private static int RingDelta(int Old, int New)
+        private static long GetTimestamp()
         {
-            if ((uint)New < (uint)Old)
-            {
-                return New + (~Old + 1);
-            }
-            else
-            {
-                return New - Old;
-            }
+            long timestamp = Stopwatch.GetTimestamp();
+
+            return (long)(timestamp * (1000.0f / Stopwatch.Frequency));
+        }
+
+        private static long RingDelta(long Old, long New)
+        {
+            return New - Old;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


ChocolArm64/ATranslatorCache.cs
```diff
@@ -154,16 +155,16 @@ private void ClearCacheIfNeeded()
...
-            else
-            {
-                return New - Old;
-            }
+            long timestamp = Stopwatch.GetTimestamp();
+
+            return (long)(timestamp * (1000.0f / Stopwatch.Frequency));
+        }
+
+        private static long RingDelta(long Old, long New)
```
gdkchan: I think this method is not necessary anymore, you can just subtract the timestamps directly on the caller since handling overflows is not necessary anymore with 64-bits timestamps.

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Diagnostics;
+
+namespace Ryujinx.Common
+{
+    public static class PerformanceCounter
+    {
+        /// <summary>
+        /// Gets the number of milliseconds elapsed since the system started.
+        /// </summary>
+        public static long ElapsedTicks
```
gdkchan: IMO ElapsedMilliseconds is a better name, since it is supposed to return the time in ms.

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            get
+            {
+                long timestamp = Stopwatch.GetTimestamp();
+
+                return (long)(timestamp * (1000.0f / Stopwatch.Frequency));
```
gdkchan: you can place `1000.0f / Stopwatch.Frequency` on a private static field at initialization to avoid re-calculating that on every call and doing a expensive division, since the Frequency is not going to change.

Ryujinx.Graphics/Gal/OpenGL/OGLCachedResource.cs
```diff
@@ -171,16 +172,9 @@ private void ClearCacheIfNeeded()
             }
         }
 
-        private int RingDelta(int Old, int New)
+        private long RingDelta(long Old, long New)
```
gdkchan: Same comment as the other one.

Ryujinx.Common/PerformanceCounter.cs
jduncanator: Agreed.

jduncanator:
`GetTimestamp` is guaranteed to use a high precision timing source on Windows XP and up, so we shouldn't ever see a fallback to `DateTime.Ticks` on Windows. Linux uses `clock_gettime` with `CLOCK_MONOTONIC` which means it's not affected by clock changes, and is guaranteed to only count up. I'm not sure on the resolution of the Linux implementation, but I believe it to be on the scale of nano- to micro-seconds.

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Diagnostics;
+
+namespace Ryujinx.Common
+{
+    public static class PerformanceCounter
+    {
+        static PerformanceCounter()
```
gdkchan: I usually place constructor after properties (at least, that's what is being done on the rest of the code), so I think it would be better to move this one.

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        /// Gets the number of milliseconds elapsed since the system started.
+        /// </summary>
+        public static long ElapsedTicks
+        {
+            get
+            {
+                long timestamp = Stopwatch.GetTimestamp();
+
+                return timestamp;
```
gdkchan: Why not just `return Stopwatch.GetTimestamp();`? This property could be just a single line.

Ryujinx.Common/PerformanceCounter.cs
jduncanator: I usually do too, but this entire file is properties and I thought the initialize would get lost at the bottom. I am happy to move it though.

Ryujinx.Common/PerformanceCounter.cs
jduncanator: Yep, I'll change that.

jduncanator:
`GetTimestamp` means we don't have to have a "running" Stopwatch instance anywhere. We don't need any of the functionality of the Stopwatch class so it's easier to just use the `GetTimestamp` static method than keep a Stopwatch around all the time.

`ElapsedTicks` _is_ more precise (has greater resolution) than `ElapsedMilliseconds`, however there are a number of places that currently expect the ticks in _ms_ (eg. the scheduler) so I just did a blanket replacement for that. If we can tweak other places where the timestamp is just some arbitrary number, then we can change those to use `ElapsedTicks`.

gdkchan:
As long `GetTimestamp ` is as precise as `ElapsedTicks`, I'm fine with it.

jduncanator:
Stopwatch uses `GetTimestamp` internally to back `ElapsedTicks` and all of it's other functionality, so it's as precise as it gets!

jduncanator:
Just as a bonus, MacOS is backed by `mach_absolute_time`, which is a monotonic value of time since last reboot.

ChocolArm64/ATranslatorCache.cs
```diff
@@ -154,16 +155,11 @@ private void ClearCacheIfNeeded()
...
         {
-            if ((uint)New < (uint)Old)
-            {
-                return New + (~Old + 1);
-            }
-            else
-            {
-                return New - Old;
-            }
+            long timestamp = Stopwatch.GetTimestamp();
```
AcK77: `timestamp` > `Timestamp`

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        /// <summary>
+        /// Gets the number of milliseconds elapsed since the system started.
+        /// </summary>
+        public static long ElapsedMilliseconds
+        {
+            get
+            {
+                long timestamp = Stopwatch.GetTimestamp();
```
AcK77: `timestamp` > `Timestamp`

Ryujinx.Common/PerformanceCounter.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                return timestamp / TicksPerMillisecond;
+            }
+        }
+
+        static PerformanceCounter()
+        {
+            TicksPerMillisecond = Stopwatch.Frequency / 1000;
```
AcK77: Could you align assign of those vars ?

PR #491, opened 2018-11-07
Better process implementation

gdkchan:
The goal here is implementing processes similar to how the real OS does. The old `Process` implementation was mostly guess work and ended being pretty different, plus it got a bit messy since it was handling NSO/NRO loading, HBABI and debug stuff, which imo doesn't belong there.

I added a new static class (called ProgramLoader) that takes care of loading NSOs now, it was based on Atmosphere ldr implementation. I removed everything related to HBABI initialization in the process, I don't know if its worth maintaining it but if someone feels we should keep it, let me know and I will try to make it work with the new process impl.

The allocator changed too, and we don't have guarantees that the memory region is contiguous with the new one (well, theres no such guarantees with the old one either, but it at least was most of the time). For this reason it was necessary to remove things like `GetHostAddress` that works on the assumption that the memory region being used is contiguous, but it may not be. So now it uses a `TryGetHostAddress` method, and only uses the pointer if the region is contiguous, otherwise it just reads a byte array and uses that. The `ReadBytes` method is capable of reading from non-contiguous regions.


gdkchan:
Rebased and updated main post. I'm yet to see an error caused by the gpu trying to access an resource on non-contiguous memory (GetHostAddress), which is unexpected, peharps we can remove this one later. Theres also the possibility of the cpu accessing an unaligned address crossing a page boundary, and the physical address for those two pages may not be contiguous either, however handling that properly will have a perf hit, and afaik unaligned accesses are pretty rare...

gdkchan:
`GetHostAddress` was replaced with `TryGetHostAddress`, everything planned for this PR should be done now.

Ryujinx.Common/BitUtils.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public static int CountLeadingZeros64(long Value)
+        {
+            return (int)CountLeadingZeros((ulong)Value, 64);
+        }
+
+        private static readonly byte[] ClzNibbleTbl = { 4, 3, 2, 2, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0 };
+
+        private static ulong CountLeadingZeros(ulong Value, int Size) // Size is 8, 16, 32 or 64 (SIMD&FP or Base Inst.).
```
jduncanator: _Another_ `clz` implementation 🙄😛

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -7,36 +7,64 @@
...
 
-        internal KRecursiveLock CriticalSectionLock { get; private set; }
+        internal bool KernelInitialized { get; private set; }
+
+        internal KResourceLimit ResourceLimit { get; private set; }
+
+        internal KMemoryRegionManager[] MemoryRegions { get; private set; }
+
+        internal KMemoryBlockAllocator MemoryBlockAllocatorSys { get; private set; }
+        internal KMemoryBlockAllocator MemoryBlockAllocator2   { get; private set; }
```
jduncanator: Is there a better name for this than `MemoryBlockAllocator2`? (something more descriptive?)

Ryujinx.Common/BitUtils.cs
gdkchan: This one is supposed to be shared with the other projects though (and this eliminate all those CLZ duplications, except for the cpu project that is supposed to be independent).

jduncanator:
This looks good to me, I don't have the deepest understanding of how the Kernel/Process subsystems work on the Switch, but from an overview, this looks all good.

I've tested a few games, and they all seem to boot and go in-game, so the loader changes are at least working, and the rest don't seem to have an affect on the games I've tested.

PR #499, opened 2018-11-15
Improved GPU command lists decoding

gdkchan:
The biggest problem of the old code is that it didn't work when the commands were split across multiple buffers, so I rewrote it based on nouveau documentation and it should match more closely what the real hardware does. Some submission modes are still not implemented (they weren't on the old code either) since it seems that nothing uses them. I also implemented 2 new texture formats (RGBX8 and D24), fixed some small bugs here and there.

Should fix the following issues (needs retesting):
https://github.com/Ryujinx/Ryujinx-Games-List/issues/108
https://github.com/Ryujinx/Ryujinx-Games-List/issues/107
https://github.com/Ryujinx/Ryujinx-Games-List/issues/106
https://github.com/Ryujinx/Ryujinx-Games-List/issues/105
https://github.com/Ryujinx/Ryujinx-Games-List/issues/104
https://github.com/Ryujinx/Ryujinx-Games-List/issues/103
https://github.com/Ryujinx/Ryujinx-Games-List/issues/102
https://github.com/Ryujinx/Ryujinx-Games-List/issues/101
https://github.com/Ryujinx/Ryujinx-Games-List/issues/74
https://github.com/Ryujinx/Ryujinx-Games-List/issues/52
https://github.com/Ryujinx/Ryujinx-Games-List/issues/47


ChocolArm64/Memory/MemoryManager.cs
```diff
@@ -409,9 +409,31 @@ public void WriteVector128(long position, Vector128<float> value)
 
         public void WriteBytes(long position, byte[] data)
```
gdkchan: Had to add this to ensure that all pages being written will be markes as modified on the gpu cache.

Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
```diff
@@ -268,7 +270,7 @@ public static TextureWrapMode GetTextureWrapMode(GalTextureWrap Wrap)
                 case GalTextureFilter.Linear:  return TextureMinFilter.Linear;
             }
 
-            throw new ArgumentException(nameof(MinFilter) + " \"" + MinFilter + "\" is not valid!");
+            return TextureMinFilter.Linear;
```
gdkchan: Not sure why we need this, but all those OpenGL games I tested seems to use 0 as texture min/mag filter.

Ryujinx.Graphics/NvGpuFifo.cs
```diff
@@ -56,148 +68,109 @@ public NvGpuFifo(NvGpu Gpu)
...
                     {
-                        int Position = PBEntry.Arguments[0];
+                        int Position = MethCall.Argument;
 
                         Macros[CurrMacroBindIndex] = new CachedMacro(this, Gpu.Engine3d, Position);
 
                         break;
                     }
+
+                    default: CallP2mfMethod(Vmm, MethCall); break;
```
gdkchan: Not sure why we need that either, but games will send the p2mf commands to other subchannels, other than the ones where the M2MF class id was bound (in particular, it sends it to subchannels 0 and 2, where 2 is M2MF and 0 is 3D). Going by the fact that method offsets don't overlap between the "engines", I believe they are really just a single engine (what nouveau calls PGRAPH), so maybe that doesn't even matter. I can unify all the current "engines" if desired, and that should fix the problem aswell, but I'm not sure if that's the right thing to do.

Ryujinx.Graphics/NvGpuEngine2d.cs
```diff
@@ -86,23 +93,42 @@ private void TextureCopy(NvGpuVmm Vmm)
...
+                DstBlitX,
+                DstBlitY,
+                DstBlitX + DstBlitW,
+                DstBlitY + DstBlitH);
+
+            //Do a guest side copy aswell. This is necessary when
+            //the texture is modified by the guest, however it doesn't
+            //work when resources that the gpu can write to are copied,
+            //like framebuffers.
+            ImageUtils.CopyTexture(
```
gdkchan: It's unfortunate that we need this, but some games will copy the texture using the 2d engine and then use P2MF to send (or modify?) some lines of the texture. Without that all data copied using would be trashed since it would detect the texture as modified on memory. With this copy we at least ensure that we have a valid copy of the texture on guest memory aswell, so if that happens it will still upload the whole texture instead of just a few lines.

Ryujinx.Graphics/Gal/OpenGL/OGLEnumConverter.cs
gdkchan: I believe that I fixed it properly now, added back the exceptions.

Ryujinx.Graphics/NvGpuFifo.cs
jduncanator: If there is no performance penalty to having all the "engines" separate, then I don't see why we wouldn't keep them separate. Helps keep things tidy.

PR #508, opened 2018-11-19
Audio: Track and Call ReleaseCallbacks in the Dummy Audio Output

jduncanator:
We need to signal the guest process when buffers are released to avoid a softlock.


jduncanator:
@gdkchan There is one remaining "hack" in here that makes this simpler to implement. When the game asks for released buffers using `GetReleasedBuffers`, we return _all_ released buffers, not just the ones that the specified track ID has used in the past.

This doesn't seem to affect games, however I'm not sure if we should be handling this correctly in the dummy audio output, or if the current implementation is "good enough".

PR #514, opened 2018-11-22
Update the "no rom selected" message in Program.cs

zephyren25:


gdkchan:
While you're at it, would be nice to also show a message if the user passes an unrecognized file (currently it just open the OpenGL window without displaying anything on the console).

maikpla:
It should be
` Console.WriteLine("Please specify the folder with the IStorage/NCA/XCI/NSP or an NRO/NSO file.");`
Since the NCA,NRO,XCI,NSP is a group and NRO/NSO are homebrews

emmauss:
Closing this as the author has either deleted his fork or set it to private.

PR #515, opened 2018-11-23
[WIP] Implement speculative translation on the CPU

gdkchan:
Should decrease startup time a bit on some games. Basically it translates the code on a separate thread, it uses a queue and adds all branch targets of the function being translated to that queue. They are then translated on another thread while the function is running. If the function happens to jump to any of those branch targets, it will be already translated and ready to run. Of course, since it has no way to know where it will or will not jump, it also does some useless work (translating functions that are never called etc).

### DIFF
--- a/ChocolArm64/Instructions/InstEmitFlow.cs
+++ b/ChocolArm64/Instructions/InstEmitFlow.cs
@@ -17,6 +17,8 @@ namespace ChocolArm64.Instructions
             }
             else
             {
+                context.TranslateAhead(op.Imm);
+
                 context.EmitStoreState();
                 context.EmitLdc_I8(op.Imm);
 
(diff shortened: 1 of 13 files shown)


gdkchan:
Made some changes on the way how it works. I think it's way simpler now. It still translates code on a separate thread, but now it only produces tier 1 (a.k.a. high quality) code there, which takes a bit longer, but should run faster. It only attempts doing that for functions (it tries to translate the entire function), previously it used a count to determine if the code was "hot" and worth optimizing. This had the disvantage that it would waste time optimizing function chuncks, rather than the whole thing.

Also, I changed the way how branches to functions (both the ones with a constant, immediate address and virtual/indirect calls with address from a register) works. Rather than returning all the way to the dispatcher, now it just calls a method that returns the target function from the cache, or translates it if not found, and returns it, then it calls the returned function. This will allow us to do some optimizations in the future, and may bring some modest improvements.

That said, it would be nice to test it with a broader number of games. From my tests, it's more or less the same as master, with some startup and steady state improvement on a few games

gdkchan:
This should be done and ready for review.

PR #519, opened 2018-12-01
Misc. CPU improvements

gdkchan:
- Fixed two race condition issues on TranslatorCache. Initially I rewrote it to use only a `ConcurrentDictionary`, but removals were too slow, so I decided to use the old method again, and fixed the issues.
- Restructured some stuff. The decoder only returns a single basic block (the entry one), all the other ones can be accessed through Next/Branch, so we don't need to return an array with all of them. `ILEmitterCtx` now traverses the graph as it should.
- Renamed some instruction (LdrLit and *_Fix suffixed ones).
- Fixed some styling issues introduced with the naming fix PR.


ChocolArm64/Instructions/InstEmitFlow.cs
```diff
@@ -73,13 +73,10 @@ public static void Blr(ILEmitterCtx context)
             OpCodeBReg64 op = (OpCodeBReg64)context.CurrOp;
 
             context.EmitLdintzr(op.Rn);
-            context.EmitSttmp();
```
gdkchan: This is not needed, because the value of `Rn` will be on the stack when the `Ret` instruction is executed.

jduncanator: Are you sure `EmitStoreState` won't trample the stack when it's storing state?

gdkchan: It shouldn't, unless there is a bug somewhere. When the `Ret` instruction is executed, it should have only a single value on the stack (since the method returns a `long` value which is the current PC). If there was more than that, then the JIT shouldn't even compile the method as it is not valid (however I don't know if it does this check today).

PR #530, opened 2018-12-07
Move MaxUboSize definition

Thog:
This fix a crash on Ryujinx.ShaderTools caused by the absence of an
OpenGL context.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
+++ b/Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
@@ -31,7 +31,7 @@ namespace Ryujinx.Graphics.Gal.Shader
 
         private StringBuilder SB;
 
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics/Gal/Shader/GlslDecompiler.cs
```diff
@@ -106,6 +108,8 @@ public GlslDecompiler()
                 { ShaderIrInst.Utof,   GetUtofExpr   },
                 { ShaderIrInst.Xor,    GetXorExpr    }
             };
+
+            GlslDecompiler.MaxUboSize = MaxUboSize / 16;
```
gdkchan: IMO, its weird that a static member is being set on the constructor. I think its better to either make it a instance member, or allow it to be set externally aswell.

PR #533, opened 2018-12-09
[WIP] Small optimizations on GPU emulation

gdkchan:
This is mostly some small optimizations I was working on, I can only test on my Intel integrated GPU, so I don't know how much it affects others, if at all.
List of attempted optimizations:
- Change resource cache to allow resources to be re-used, by adding a resource pool.
- Allow framebuffers, zeta buffers, const buffers, vertex buffers and index buffers to be re-used.
- VAO (Vertex Array Object) caching - No noticeable improvement, should save a few OpenGL calls though.
- Added a PBO (Pixel Buffer Object) cache.
- Use dirty flags to update the pipeline state, triggered by writes to the respective registers. The old solution was comparing the entire state and only calling GL functions when it changed, which obviously doesn't scale very well (performance can only get worse as more state is implemented due to all those comparisons).
- Use an accurate vertex count on most cases, to reduce the amount of vertex data sent to the GPU.
- Optimized memory modified check method (on some cases just checking was more expensive than sending the data, now its not anymore).
- Remove only the respective key from the fast cache when a const buffer is modified with `CbData`.
- (TODO) 2d software copy is currently commented out, because it was slowing down some games. I still need to think in a better way to do that.


legend800:
You just made Ultra Street Fighter 2 playable!! 200% speed gain! 20 > 60 fps.

Mainline:
2018-12-08 22_01_29-window

PR:
2018-12-08 22_05_59-window

RyadaProductions:
I am currently working on splitting this PR up in multiple smaller PRs. 
Crossreferencing the PRs I create in this reaction for clarity:
#538

jduncanator:
Closing as owner has no intent on rebasing and finalizing.

PR #534, opened 2018-12-09
Misc. Driver/OS Improvement

WilliamWsyHK:
Add a little bit driver method, thus games like Octopath Traveler can be opened.
image

Also, base on #532, I added some improvement so that using TraditionalChinese and SimplifiedChinese won't cause exceptions due to parsing [SystemLanguage](https://switchbrew.org/wiki/Settings_services#LanguageCode) enum as [TitleLanguage](https://switchbrew.org/wiki/Control.nacp#Title_Entry).


WilliamWsyHK:
close this PR as PR dependency exists. Will create new PR with different branch for language config.

PR #538, opened 2018-12-11
Optimized memory modified check

RyadaProductions:
This was initially in some cases more expensive than plainly sending the data. Now it should have way better performance.

First step in splitting up the massive PR #533  from gdkchan
Refactors in the big PR will be added as a seperate PR after all functional changes have been implemented.


RyadaProductions:
I have added these refactors in a seperate commit instead of rebasing them together to keep the history as clean as possible.

jduncanator:
I assume it's building (I didn't check), but so long as the changes are the same as in the other PR, this looks good to me,

gdkchan:
I believe that with those changes `ValueRange` and `ValueRangeSet` can be removed aswell (at `Ryujinx.Graphics`).

RyadaProductions:
ValueRange and ValueRangeSet are currently still in use because in this PR the Cache changes have not been added yet. Those will be in a seperate PR.

gdkchan:
Are you sure? Because they are only used on `NvGpuVmmCache`.

RyadaProductions:
Yes. As far as I gathered the changes in `NvGpuVmmCache` need the new Cache part to properly be able to merge. As it seems like a direct result of the change in Cache implementation.
With this PR I tried to only commit the logic that was related to the change in the memory modified check.

gdkchan:
The changes on NvGpuVmmCache are completely independent from the other changes, and should work fine on they own.

RyadaProductions:
Removed `ValueRange` and `ValueRangeSet` in this commit as they were not used anymore.

PR #539, opened 2018-12-11
Cache changes

RyadaProductions:
Sadly these commits could not be split up, as they rely quite heavily on each other.

* moved constants to a class for constant resource limits
* Added VAO & PBO caching
* Change resource cache to allow resources to be re-used, by adding a resource pool.
* Use dirty flags to update the pipeline state, triggered by writes to the respective registers. The old solution was comparing the entire state and only calling GL functions when it changed, which obviously doesn't scale very well (performance can only get worse as more state is implemented due to all those comparisons).

Another PR based of gdkchan earlier commit:  #533


gdkchan:
About `Commented  out 2d software copy. This was creating slowdowns in specific titles,  and gdkchan is wanting to think about a different implementation.` maybe we should keep it uncommented, because it causes graphical regressions on some games (mainly OpenGL games, I think), however having it causes a big perf drop on others.

gdkchan:
The changes on `NvGpuVmmCache` are completely independent from the other changes, and should work fine on they own.

RyadaProductions:
Not sure how those got in there but those should not be there now.

RyadaProductions:
Updated https://github.com/Ryujinx/Ryujinx/pull/539#issue-237535097

RyadaProductions:
Separated the pipeline state changes and the cache changes into a seperate commit. I didn't create a new PR for it, as it would have to be merged in order of `pipeline changes` -> `cache changes`. So just to be sure I had them share the same PR

RyadaProductions:
I will change the history of this PR once #544 is approved and merged.

jduncanator:
These changes need additional work on them, as well as a rebase. Closing in preference of the original PR.

PR #541, opened 2018-12-13
Change effect states in the Audio Renderer

Dudejoe870:
This is probably part of the issue why SMOs sub-areas don't work, this needs to be tested.  New Game in SMO still doesn't work, could be different with sub-areas though.

Has been tested: doesn't seem like sub-areas work, however, this could still be part of the problem.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Aud/AudioRenderer/IAudioRenderer.cs
+++ b/Ryujinx.HLE/HOS/Services/Aud/AudioRenderer/IAudioRenderer.cs
@@ -38,6 +38,8 @@ namespace Ryujinx.HLE.HOS.Services.Aud.AudioRenderer
 
         private VoiceContext[] _voices;
 
+        private EffectContext[] _effects;
+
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/HOS/Services/Aud/AudioRenderer/IAudioRenderer.cs
```diff
@@ -75,6 +77,8 @@ class IAudioRenderer : IpcService, IDisposable
 
             _voices = CreateArray<VoiceContext>(Params.VoiceCount);
```
emmauss: you could remove the new line, and align this two

Dudejoe870:
Addressed Feedback

BaronKiko:
Can this be merged with a rebase or should it be closed?

Dudejoe870:
It should probably be closed, even though it has no merge conflicts, I'm not really that happy with the code and it is a bit of a hack to be honest.

PR #544, opened 2018-12-14
Pipeline state changes

RyadaProductions:
"Use dirty flags to update the pipeline state, triggered by writes to the respective registers. The old solution was comparing the entire state and only calling GL functions when it changed, which obviously doesn't scale very well (performance can only get worse as more state is implemented due to all those comparisons)." - gdkchan - #533


jduncanator:
These changes need additional work on them, as well as a rebase. Closing in preference of the original PR.

PR #545, opened 2018-12-14
Partially implement LDG

Dudejoe870:
A rebase of #411.


Thog:
As this PR (and the original) doesn't implement LDG completely, I don't think it's a good idea to have it merged as it may cause more issues than solving things.

gdkchan:
Is it worse than not having LDG at all? Btw, I think that implementing LDG "completely", without software rendering is impossible.

AcK77:
Since Shader translation has changed, we can’t merge it anymore... It’s better to close it!

PR #549, opened 2018-12-21
Update to LibHac 0.2.0

Thealexbarney:
Copy/pasted applicable changes from changelog

- Use a new `IStorage` interface throughout the library instead of `Stream`
    - This interface is more similar to the `IStorage` interface used by Horizon.
    - Instead of having a Stream that keeps track of the current position, IStorage accepts an offset, buffer, and length when reading or writing.
    - Unlike a Stream, an IStorage instance can be shared between multiple consumers without issue
    - `Stream.AsStorage()` and `IStorage.AsStream()` methods are provided for switching between `Stream` and `IStorage`
- Improve AES-XTS performance by ~16x
- Improve AES-CTR performance by 2x
- Allow more disposal methods to cascade down to the base storage

### DIFF
--- a/Ryujinx.HLE/FileSystem/Content/ContentManager.cs
+++ b/Ryujinx.HLE/FileSystem/Content/ContentManager.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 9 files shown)


emmauss:
lgtm

PR #550, opened 2018-12-26
[WIP] Improve kernel IPC implementation

gdkchan:
This attempts to implement it the right way. A bunch of things are still missing, and it also needs to be tested properly. Ideally, we should update the current HLE service implementation to work with the new IPC system, but that's rather complicated and I prefer to save it for later. For now I introduced a constant that can be toggled to use the old or new IPC system (by default it uses the old one).


gdkchan:
This should be ready for review.
One thing that is worth mentioning is that I had to add some basic ref counting, because even though .net has a garbage collector and we don't need to worry about object destruction, it's still necessary to release kernel resources (like decrementing counters for limitable resources etc). It's kind of a pain so I tried to add it only where it is *really* required.

PR #551, opened 2018-12-29
Frame times in title bar

BaronKiko:


emmauss:
though this seems like a good idea, you have not given any description of what this is and how it would improve ryujinx.

BaronKiko:
Sorry, it makes it easier to see fluctuations in performance. I implemented it to test how performance is effected for converting quads to triangles. Pr for that coming in a min.

emmauss:
sounds great. could you join the discord server. Makes it easier to discuss this stuff

Ryujinx/Ui/GLScreen.cs
```diff
@@ -246,14 +246,23 @@ private new void RenderFrame()
...
             _device.Statistics.RecordSystemFrameTime();
 
+            // System
             double hostFps = _device.Statistics.GetSystemFrameRate();
+            double hostFrameTime = _device.Statistics.GetSystemFrameTime();
+            double hostAvgFrameTime = _device.Statistics.GetSystemAverageFrameTime();
+
+            // Game
             double gameFps = _device.Statistics.GetGameFrameRate();
+            double gameFrameTime =_device.Statistics.GetGameFrameTime();
```
emmauss: these ones too

Ryujinx/Ui/GLScreen.cs
```diff
@@ -246,14 +246,23 @@ private new void RenderFrame()
 
             _device.Statistics.RecordSystemFrameTime();
 
+            // System
             double hostFps = _device.Statistics.GetSystemFrameRate();
+            double hostFrameTime = _device.Statistics.GetSystemFrameTime();
+            double hostAvgFrameTime = _device.Statistics.GetSystemAverageFrameTime();
```
emmauss: align these

BaronKiko:
Done

Ryujinx/Ui/GLScreen.cs
```diff
@@ -246,14 +246,23 @@ private new void RenderFrame()
...
-            double hostFps = _device.Statistics.GetSystemFrameRate();
-            double gameFps = _device.Statistics.GetGameFrameRate();
+            // System
+            double hostFps           = _device.Statistics.GetSystemFrameRate();
+            double hostFrameTime     = _device.Statistics.GetSystemFrameTime();
+            double hostAvgFrameTime  = _device.Statistics.GetSystemAverageFrameTime();
+
+            // Game
+            double gameFps           = _device.Statistics.GetGameFrameRate();
+            double gameFrameTime     =_device.Statistics.GetGameFrameTime();
```
AcK77: You forgot a space after the =

BaronKiko: Once again done

BaronKiko:
The averages is actually pretty useless because it's just 1000/fps. It's there to help you understand deviation in the immediate frame time (because doing 1000/fps in your head isn't exactly always obvious). The reason immediate frame time is so important is because it helps you measure frame consistency which is arguably more important to overall playablility than frame rate is.
For example if you have a game that runs at 60fps (16ms) you should never see the 100's digit become a 1 because it means some frames are taking over 5 times as long. Likewise if it's all over the place on a static menu then something is probably not right.
Ideally is should go in a bottom bar or something else you can disable but at there isn't any UI outside of the title bar it went there to avoid implementing lots of UI.
Tl;dr: I think it's all useful and needed but it should be moved off the title bar when some UI is in place to house it.

BaronKiko:
Thinking about it, you could use a config option to hide it. You would probably want to disable by default for users but not when doing a regular build in visual studio.
Regardless I think that's all a bit out of scope for this pr so it should probably be done separately.

BaronKiko:
Rip first commit.
I am moving this info to my new profiler instead: https://github.com/Ryujinx/Ryujinx/pull/567

PR #554, opened 2019-01-04
[WIP] Implement hardware devices support on the kernel

gdkchan:
The goal here is implementing the SVCs that are related to communication with hardware devices (*DeviceAddressSpace, QueryIoMapping, IRQ related stuff). This is just the initial work.

It may have some performance regressions on the CPU memory manager since we now need to call methods to handle access to the devices, and this is done by using a null pointer. The page table now stores the guest physical address aswell, to allow communicating with said devices. A IBus interface was added, each device is supposed to implement it, and the CPU calls the methods to read/write registers etc.

The above is by no means the final solution. I plan to investigate ways to improve the memory manager performance, so it will probably be partially rewritten at some point.

TODO:
- Implement map device address space.
- Finish basic interrupt events implementation.
- Finish device base class implementation, implement a dummy device using it initially.
- Investigate ways to do address translation with support for external method calling efficiently on the memory manager.
- Probably more that I forgot right now...

This depends on #550, I had to merge it in so that's why the PR is a bit lengthy.


gdkchan:
Closing since this is really old and I don't plan to finish anymore.

PR #558, opened 2019-01-12
Add file logging and handle unhandled exceptions

emmauss:
This adds logging for unhandled exceptions, and also file logs. You can enable (which is on by default) file logs with the `Enable_File_Log` key in config.


Ryujinx/Program.cs
```diff
@@ -24,6 +24,12 @@ static void Main(string[] args)
 
             Logger.Updated += ConsoleLog.Log;
 
+            if (Logger.EnableFileLog)
```
gdkchan: I guess we should decide whenever or not to allow omiting the `{` and `}`, for example @BaronKiko wanted to do that and the recommendation was not doing it.

Ryujinx/Ui/FileLog.cs
```diff
@@ -0,0 +1,116 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.IO;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+
+namespace Ryujinx
+{
+    static class FileLog
```
gdkchan: Can't we have a single class to handle both console and file log output?

PR #560, opened 2019-01-21
[BLOCKED] .NET Core 3.0

BaronKiko:
[BLOCKED] Decided to wait till .NET 3.0 is fully released before merging. Blocked until then.

Attempts to update to experimental .Net Core 3.0,
This was initially an attempt to make use of the more optimized jit available in 3.0 however the speed difference was worse if anything.

Also as fiigii said:

> using Create to load data from array to vectors is very inefficient. Please use Sse2.LoadVector128 instead.

So every create call needs to be converted where possible. Jd said he would look at it while he applies his changes. Will reference his pr here once it's up.

Replacements made:

- `Sse2.SetVector128` and `Sse2.SetAllVector128` are now `Vector128.Create`
- `Sse.StaticCast<T1, T2>(vector)` is now `vector.As<T2>`
- `Sse.SetScalarVector128` is now `Vector128.CreateScalarUnsafe` (unsure on the unsafe part but changing it doesn't make a difference
- `Sse2.SetZeroVector128<T>()` is now `Vector128<T>.Zero`
- `Sse41.Insert(Vector128<float>, float, byte)` was replaced with `Sse41.Insert(Vector128<float>, Vector128<float>, byte)`, so converted calls to `Sse41.Insert(value, Vector128.CreateScalarUnsafe(data), index)`


ChocolArm64/Instructions/CryptoHelper.cs
```diff
@@ -318,11 +318,11 @@ private static void FromByteArrayToVector(byte[] state, ref Vector128<float> op)
                 throw new PlatformNotSupportedException();
             }
 
-            op = Sse.StaticCast<byte, float>(Sse2.SetVector128(
+            op = Vector128.Create(
                 state[15], state[14], state[13], state[12],
                 state[11], state[10], state[9],  state[8],
                 state[7],  state[6],  state[5],  state[4],
-                state[3],  state[2],  state[1],  state[0]));
+                state[3],  state[2],  state[1],  state[0]).As<float>();
```
fiigii: `Sse2.SetVector128(e15, e14, ..., e0)` should be replaced to `Vector128.Create(e0, e1, ..., e15)`.

fiigii: Yes, the parameter order is different.

fiigii: Btw, using Create to load data from array to vectors is very inefficient. Please use Sse2.LoadVector128 instead.

ChocolArm64/Instructions/VectorHelper.cs
```diff
@@ -497,15 +497,15 @@ public static Vector128<float> VectorInsertSingle(float value, Vector128<float>
...
         {
             //Note: 0b1110 is the mask to zero the upper bits.
-            return Sse41.Insert(vector, value, 0b1110);
+            return Sse41.Insert(vector, Vector128.CreateScalarUnsafe(value), 0b1110);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static Vector128<sbyte> VectorSByteZero()
         {
             if (Sse2.IsSupported)
```
gdkchan: This check if no longer needed as the method is not under the SS2 ISA anymore, so there's a software fallback (or another implementation using arch specific intrinsics) if the CPU doesn't support SSE2.

Could be just:
```cs
public static Vector128<sbyte> VectorSByteZero() => Vector128<sbyte>.Zero;
```

Or maybe we can just remove those methods and just get the property directly.

ChocolArm64/Instructions/VectorHelper.cs
```diff
@@ -571,7 +571,7 @@ public static Vector128<sbyte> VectorSingleToSByte(Vector128<float> vector)
         {
             if (Sse.IsSupported)
```
gdkchan: Same comment as above. The check is no longer needed for those "reinterpret cast" methods.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -489,7 +488,7 @@ protected static Vector128<float> MakeVectorE0(ulong e0)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<ulong, float>(Sse2.SetVector128(0, e0));
```
gdkchan: Order of elements is inverted.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -499,7 +498,7 @@ protected static Vector128<float> MakeVectorE0E1(ulong e0, ulong e1)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<ulong, float>(Sse2.SetVector128(e1, e0));
```
gdkchan: Order of elements is inverted.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -509,7 +508,7 @@ protected static Vector128<float> MakeVectorE1(ulong e1)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<ulong, float>(Sse2.SetVector128(e1, 0));
```
gdkchan: Order of elements is inverted.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -434,7 +434,7 @@ protected static Vector128<float> MakeVectorE0(double e0)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<long, float>(Sse2.SetVector128(0, BitConverter.DoubleToInt64Bits(e0)));
```
gdkchan: Order of elements is inverted.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -444,8 +444,7 @@ protected static Vector128<float> MakeVectorE0E1(double e0, double e1)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<long, float>(
```
gdkchan: Order of elements is inverted.

Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -455,7 +454,7 @@ protected static Vector128<float> MakeVectorE1(double e1)
                 throw new PlatformNotSupportedException();
             }
 
-            return Sse.StaticCast<long, float>(Sse2.SetVector128(BitConverter.DoubleToInt64Bits(e1), 0));
```
gdkchan: Order of elements is inverted.

ChocolArm64/Memory/MemoryManager.cs
```diff
@@ -243,7 +243,7 @@ public Vector128<float> ReadVector8(long position)
         {
             if (Sse2.IsSupported)
             {
-                return Sse.StaticCast<byte, float>(Sse2.SetVector128(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ReadByte(position)));
+                return Vector128.Create(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ReadByte(position)).As<float>();
```
gdkchan: Order of elements is inverted (ReadByte is supposed to be the first one).

ChocolArm64/Instructions/InstEmitSimdLogical.cs
```diff
@@ -310,7 +310,7 @@ public static void Rev16_V(ILEmitterCtx context)
                 context.EmitLdc_I8(14L << 56 | 15L << 48 | 12L << 40 | 13L << 32 | 10L << 24 | 11L << 16 | 08L << 8 | 09L << 0); // maskE1
                 context.EmitLdc_I8(06L << 56 | 07L << 48 | 04L << 40 | 05L << 32 | 02L << 24 | 03L << 16 | 00L << 8 | 01L << 0); // maskE0
 
-                context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.SetVector128), typesSve));
+                context.EmitCall(typeof(Vector128).GetMethod(nameof(Vector128.Create), typesSve));
```
gdkchan: I believe the order is inverted here aswell (you need to swap the two `EmitLdc_I8` calls to swap maskE0/maskE1).

ChocolArm64/Instructions/InstEmitSimdLogical.cs
```diff
@@ -349,7 +349,7 @@ public static void Rev32_V(ILEmitterCtx context)
                     context.EmitLdc_I8(05L << 56 | 04L << 48 | 07L << 40 | 06L << 32 | 01L << 24 | 00L << 16 | 03L << 8 | 02L << 0); // maskE0
                 }
 
-                context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.SetVector128), typesSve));
+                context.EmitCall(typeof(Vector128).GetMethod(nameof(Vector128.Create), typesSve));
```
gdkchan: I believe the order is inverted here aswell (you need to swap the two `EmitLdc_I8` calls to swap maskE0/maskE1).

ChocolArm64/Instructions/InstEmitSimdLogical.cs
```diff
@@ -393,7 +393,7 @@ public static void Rev64_V(ILEmitterCtx context)
                     context.EmitLdc_I8(03L << 56 | 02L << 48 | 01L << 40 | 00L << 32 | 07L << 24 | 06L << 16 | 05L << 8 | 04L << 0); // maskE0
                 }
 
-                context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.SetVector128), typesSve));
+                context.EmitCall(typeof(Vector128).GetMethod(nameof(Vector128.Create), typesSve));
```
gdkchan: I believe the order is inverted here aswell (you need to swap the two `EmitLdc_I8` calls to swap maskE0/maskE1).

jduncanator:
@BaronKiko The intrinsics API directly maps to underlying processor instructions, so it's not as simple as "creating a overload that takes an array". `LoadVector128` can load the vector values directly in from a memory address in a single instruction. Depending on optimizations available in RyuJIT (I'm not sure), `Vector128.Create` will need to issue individual load instructions for each parameter. It's much more preferred to use `LoadVector128` wherever possible.

gdkchan:
We can remove the ISA checks for all the methods that were moved to `Vector128` since they have software fallback and should work everywhere.

BaronKiko:
Ok lots of reordering to do, will take me quite a bit however I was just about to take a break for dinner so will look a little later.
@gdkchan I used unsafe because it seemed less likely to cause an issue without me having to think about every replacement. We can work out what needs to be unsafe later but jd seems to have more grand plans over in dev chat.
@jduncanator I suggested the create thing because it seems like they are trying to get most things into vector128 and it could just grab the address from an array. The main thing to note is getting it into Vector128 rather than Sse because I would have likely used it if I had known about it. Anyway fiigii does .net work so I'm sure he will know what to do.

gdkchan:
We should check the codegen of `Sse41.Insert(value, Vector128.CreateScalarUnsafe(data), index)` since its now taking a vector rather than a float directly, needing the additional `CreateScalarUnsafe` call. I'm affraid it is producing additional instructions now.

jduncanator:
@BaronKiko The reason it's in SSE is because the load from memory instruction (specifically the element "width" of the instruction) is SSE specific, each version of SSE exposes different memory load instructions, so just putting the `SetVector128(short[])` overload on `Vector128` would mean that it would have to be available all the time, which isn't the case (especially with larger element sizes).

BaronKiko:
@gdkchan That was the suggested replacement by the .net devs so I just went with it. If we can start with a vector then all the better. I was just trying to get it working.
@jduncanator Ok, was just a thought. Not really for us to worry about anyway :)

ChocolArm64/Instructions/CryptoHelper.cs
BaronKiko: JD will look at Sse2.LoadVector128 changes with his pr. All the orders have been fixed.

ChocolArm64/Memory/MemoryManager.cs
```diff
@@ -243,7 +243,7 @@ public Vector128<float> ReadVector8(long position)
         {
             if (Sse2.IsSupported)
             {
-                return Sse.StaticCast<byte, float>(Sse2.SetVector128(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ReadByte(position)));
+                return Vector128.Create(ReadByte(position), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0).As<float>();
```
fiigii: Why not `Sse41.Insert(Vector128<byte>.Zero, ReadByte(position), 0).As<float>()`? That is much faster.

fiigii:
> We should check the codegen of Sse41.Insert(value, Vector128.CreateScalarUnsafe(data), index) since its now taking a vector rather than a float directly, needing the additional CreateScalarUnsafe call. I'm afraid it is producing additional instructions now.

`CreateScalarUnsafe` is an intrinsic that directly uses a `Vector128<float>` as `float` since `Vector128<float>`and `float` is both holded in `XMM` registers on the assembly level. So, this intrinsic should not generate additional instructions.

jduncanator:
I think for now, we should leave this PR as a strict "port" across to the new Intrinsics API (to remove as much code change surface as possible from this PR). We can make suggested optimizations by fiigii and others in a separate PR based on this one, which should allow us to more easily diagnose potential issues and determine which change introduced them.

BaronKiko:
Removed more sse checks as per gdk's suggestion.
@gdkchan Regarding CI builds I think we should try modify the build script if it's not going to be too hard.

BaronKiko:
Fixed Appveyor build script

jduncanator:
> Removed more sse checks as per gdk's suggestion.
> @gdkchan Regarding CI builds I think we should try modify the build script if it's not going to be too hard.

You can also remove any check like the following:

```cs
if(Sse.IsAvailable && Sse2.IsAvailable) { }
```

AFAIK, .NET Core / RyuJIT does not support any CPU configurations where SSE would be available but SSE2 isn't. This can be simplified to:


```cs
if(Sse2.IsAvailable) { }
```

jduncanator:
This needs a rebase.

BaronKiko:
Rebased. 
Unfortunatly I couldn't get it to work with the latest preview. To build you need to use v3.0.0-preview-27122-01 available from https://dotnet.microsoft.com/download/dotnet-core/3.0
The reason it doesn't work with preview 2 is because SSE2.Insert an SSE2.Extract no longer take longs and ulongs. Apparently they are being moved to Vector128 however in preview 2, and all the dailies that are currently available, these methods don't exist.

samuelGrahame:
Just to comment why it is slower in .net core:

"Tiered compilation is a feature that enables the runtime to more adaptively use the Just-In-Time (JIT) compiler to get better performance, both at startup and to maximize throughput. It was added as an opt-in feature in .NET Core 2.1 and then was enabled by default in .NET Core 2.2 Preview 2"

https://github.com/dotnet/coreclr/pull/21292

would be linked to above.

samuelGrahame:
what is the speed difference with COMPlus_TieredCompilation=0 apparently there is a bug/issue with performance in .net core 2.2 >= I am interested in the performance difference between .net core 3 and .net core 2 with this option off.

BaronKiko:
No idea how I would set that but your are welcome to try.
I didn't really notice and speed difference without that flag

samuelGrahame:
> Rebased.
> Unfortunately I couldn't get it to work with the latest preview (2). To build you need to use v3.0.0-preview-27122-01 available from https://dotnet.microsoft.com/download/dotnet-core/3.0
> The reason it doesn't work with preview 2 is because SSE2.Insert an SSE2.Extract no longer take longs and ulongs. Apparently they are being moved to Vector128 however in preview 2, and all the dailies that are currently available, these methods don't exist.

ulong/long functions are under "Sse41.X64" - per github: 
https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Runtime/Intrinsics/X86/Sse41.cs

samuelGrahame:
> Rebased.
> Unfortunately I couldn't get it to work with the latest preview (2). To build you need to use v3.0.0-preview-27122-01 available from https://dotnet.microsoft.com/download/dotnet-core/3.0
> The reason it doesn't work with preview 2 is because SSE2.Insert an SSE2.Extract no longer take longs and ulongs. Apparently they are being moved to Vector128 however in preview 2, and all the dailies that are currently available, these methods don't exist.

I think they are under: Sse41.X64

gdkchan:
Yes, they are under `Sse41.X64`. And long/ulong overloads of `PopCount` are also under `Popcnt.X64.PopCount` now, so those calls should be updated aswell (on `Cnt_V`).

gdkchan:
I beleive this is not relevant anymore as we are migrating to a new jit.

PR #498, opened 2018-11-12
Add blx_reg bx bxj

arthurchan35:
From: https://github.com/Ryujinx/Ryujinx/pull/303
resolve confilicts after rebasing to latest code

PR #561, opened 2019-01-22
Add ARM32 support on the translator

gdkchan:
This removes the ARM32 interpreter (which only had 3 instructions implemented), for reasons discussed at #498, and does some basic work to support it on the translator. A bunch of things are still necessary, including:

- Support for different modes (User, System, IRQ, FIQ etc), this is only required for LLE as HLE will only execute on the User mode.
- Support for NEON vector registers and instructions.
- SPSR registers.

Of course, almost all instructions are still not implemented. The goal here is just have the basics to allow others to start implementing more instructions.

Currently all the ARM64 OpCode classes at Decoder are suffixed with `64`, while for ARM32 they are suffixed with either `32` for arm, or `T16` for thumb with 16-bits opcode. However for the instruction emitters, theres no suffix for ARM64, and ARM32 is suffixed with `32`. I think its better to do the same with `OpCode` and only suffix ARM32 ones. Either way, I think the inconsistency should be fixed, I'm open to suggestions.

I changed the opcode table to use the "fast lookup" with arm32 aswell. However it doesn't make much sense with thumb since it is 16-bits (or 32-bits), so I think it would only benefit the 32-bits case, and even then I'm not sure since thumb-2 opcode encodings are awkward. Peharps we should use different values from the opcode for the "fast lookup" when the mode is thumb.


gdkchan:
Regarding the `OpCode` 64 suffix changes, I think it's better to do that later on a separate PR, in order to avoid conflicts with other CPU work.
This should be ready.

EmulationChannel:
Game 32bit Test New Super Mario Bros DELUXE

gdkchan:
As mentioned on the PR description:
```
Of course, almost all instructions are still not implemented.
```
There's still a long way before those 32-bits games start booting on the emulator. A lot of CPU instructions still needs to be implemented. This just adds the foundation to enable us to start implementing those instructions.

emmauss:
@EmulationChannel it does not yet support 32bit games. this is just the base, and no instruction is implemented

PR #565, opened 2019-01-26
Implement some ARM32 memory instructions and CMP

gdkchan:
Implements some memory instructions. Only immediate and register + immediate variants for now (still missing register + shifted register variants). Exclusive access instructions are not implemented. I also had to implement CMP since I needed that to run my tests.


gdkchan:
Renamed the name of ARM32 opcode classes to OpCode32* rather than the old OpCode*32. Main reason is that when sorted alphabetically (which is what the IDEs will usually do), it will show all ARM32 classes close to each other, and also it allows to suffix stuff with numbers at the end (for example, OpCode32Imm8 vs OpCodeImm832, the former is more readable). This created some inconsistency with OpCode*64. Like I said on the previous commit, the plan is removing the `64` suffix there. However I will do that later to avoid conflicts with some pending CPU related PRs.

gdkchan:
It's worth noting that a lot of combinations on memory instructions are listed as unpredictable on the manual, or has behavior that varies between CPUs. A few examples are:
- Load/Store multiple with the register mask = 0.
- Load/Store multiple with Rn on the register mask, where Rn is not the first bit set on the mask.
- Cases where Rn == Rt or Rn == Rt2 (for doubleword loads/stores).
- Cases where Rt == 15 for doubleword loads/stores. Rt2 = Rt + 1, but there's no register 16, so this case is special.  (Rt & 1) != 0 may also not be valid depending on the cpu.
- There are some cases where writing to PC is not "unpredictable" aswell.

I implement them as I seem fit. Compilers shouldn't produce any of those encodings as the program wouldn't work properly on all CPUs, except peharps when a specific CPU is being targeted, and the behavior is well defined, or the software has parts programmed in assembly.

ChocolArm64/OpCodeTable.cs
```diff
@@ -37,18 +37,32 @@ static OpCodeTable()
...
+            SetT32("010001110xxxx000",                 InstEmit32.Bx,     typeof(OpCodeT16BReg));
+            SetA32("<<<<00110101xxxx0000xxxxxxxxxxxx", InstEmit32.Cmp,    typeof(OpCode32AluImm));
+            SetA32("<<<<00010101xxxx0000xxxxxxx0xxxx", InstEmit32.Cmp,    typeof(OpCode32AluRsImm));
+            SetA32("<<<<100xx0x1xxxxxxxxxxxxxxxxxxxx", InstEmit32.Ldm,    typeof(OpCode32MemMult));
+            SetA32("<<<<010xx0x1xxxxxxxxxxxxxxxxxxxx", InstEmit32.Ldr,    typeof(OpCode32MemImm));
+            SetA32("<<<<010xx1x1xxxxxxxxxxxxxxxxxxxx", InstEmit32.Ldrb,   typeof(OpCode32MemImm));
+            SetA32("<<<<000xx1x0xxxxxxxxxxxx1101xxxx", InstEmit32.Ldrd,   typeof(OpCode32MemImm8));
+            SetA32("<<<<000xx1x1xxxxxxxxxxxx1011xxxx", InstEmit32.Ldrh,   typeof(OpCode32MemImm8));
+            SetA32("<<<<000xx1x1xxxxxxxxxxxx1101xxxx", InstEmit32.Ldrsb,  typeof(OpCode32MemImm8));
+            SetA32("<<<<000xx1x1xxxxxxxxxxxx1111xxxx", InstEmit32.Ldrsh,  typeof(OpCode32MemImm8));
```
LDj3SNuD: Maybe remove a space or align with the other `typeof()` of the region below.

gdkchan: I added the extra space to avoid needing re-aligning the table when instructions with longer names are added. But I can re-align it to remove the extra spaces, if desired.

PR #566, opened 2019-01-26
Add Smlal_Ve, Smlsl_Ve, Smull_Ve, Umlal_Ve, Umlsl_Ve, Umull_Ve Inst.; add Tests. Add Sse Opt. for Trn1/2_V and Uzp1/2_V Inst. Nits.

LDj3SNuD:
ChocolArm64\Instructions\CryptoHelper.cs:
`FromVectorToByteArray()` -> Used `Sse2.Store()`
`FromByteArrayToVector()` -> Used `Sse2.LoadVector128()`

ChocolArm64\Instructions\InstEmitSimdArithmetic.cs,
ChocolArm64\Instructions\InstEmitSimdCmp.cs:
Removed some unnecessary `Optimizations.UseSse`

### DIFF
--- a/ChocolArm64/Instructions/InstEmitSimdHelper.cs
+++ b/ChocolArm64/Instructions/InstEmitSimdHelper.cs
@@ -642,21 +642,21 @@ namespace ChocolArm64.Instructions
         {
             OpCodeSimdRegElem64 op = (OpCodeSimdRegElem64)context.CurrOp;
 
-            EmitVectorOpByElem(context, emit, op.Index, false, true);
+            EmitVectorOpByElem(context, emit, op.Index, ternary: false, signed: true);
         }
 
         public static void EmitVectorBinaryOpByElemZx(ILEmitterCtx context, Action emit)
         {
             OpCodeSimdRegElem64 op = (OpCodeSimdRegElem64)context.CurrOp;
 
-            EmitVectorOpByElem(context, emit, op.Index, false, false);
+            EmitVectorOpByElem(context, emit, op.Index, ternary: false, signed: false);
         }
 
         public static void EmitVectorTernaryOpByElemZx(ILEmitterCtx context, Action emit)
         {
             OpCodeSimdRegElem64 op = (OpCodeSimdRegElem64)context.CurrOp;
 
-            EmitVectorOpByElem(context, emit, op.Index, true, false);
+            EmitVectorOpByElem(context, emit, op.Index, ternary: true, signed: false);
         }
 
         public static void EmitVectorOpByElem(ILEmitterCtx context, Action emit, int elem, bool ternary, bool signed)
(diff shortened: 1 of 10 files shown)


ChocolArm64/Instructions/InstEmitSimdHelper.cs
```diff
@@ -809,6 +809,64 @@ public static void EmitVectorWidenRnRmOp(ILEmitterCtx context, Action emit, bool
             context.EmitStvec(op.Rd);
         }
 
+        public static void EmitVectorWidenBinaryOpByElemSx(ILEmitterCtx context, Action emit)
+        {
+            OpCodeSimdRegElem64 op = (OpCodeSimdRegElem64)context.CurrOp;
+
+            EmitVectorWidenOpByElem(context, emit, op.Index, false, true);
```
gdkchan: It would be nice to give explicit name for the boolean arguments, for example:
```cs
EmitVectorWidenOpByElem(context, emit, op.Index, ternary: false, signed: true);
```
Imo it's easier to see what they do this way.

ChocolArm64/Instructions/InstEmitSimdMove.cs
```diff
@@ -465,22 +464,77 @@ private static void EmitVectorTranspose(ILEmitterCtx context, int part)
...
-            int words = op.GetBitsCount() >> 4;
-            int pairs = words >> op.Size;
-
-            for (int index = 0; index < pairs; index++)
+            if (Optimizations.UseSsse3)
             {
-                int idx = index << 1;
+                Type[] GetTypesUpkSfl(int size) => new Type[] { VectorIntTypesPerSizeLog2[size], VectorIntTypesPerSizeLog2[size] };
+
+                long[] masksE1 = new long[]
```
gdkchan: Looks like those arrays are being repeated a bunch of times, it would be nice to remove those duplicates and keep them on a single place (with a more descriptive name).

ChocolArm64/Instructions/InstEmitSimdMove.cs
```diff
@@ -492,26 +546,119 @@ private static void EmitVectorUnzip(ILEmitterCtx context, int part)
...
-            }
+                string nameUpk = part == 0
+                    ? nameof(Sse2.UnpackLow)
+                    : nameof(Sse2.UnpackHigh);
 
-            context.EmitLdvectmp();
-            context.EmitStvec(op.Rd);
+                if (op.RegisterSize == RegisterSize.Simd128)
+                {
+                    long[] masksE1 = new long[]
```
gdkchan: First repetition.

ChocolArm64/Instructions/SoftFallback.cs
```diff
@@ -664,7 +664,7 @@ public static ulong BinaryUnsignedSatQAcc(long op1, ulong op2, CpuThreadState st
 
             for (int bit = highBit; bit >= 0; bit--)
             {
-                if (((value >> bit) & 0b1) != 0)
+                if (((int)(value >> bit) & 0b1) != 0)
```
gdkchan: Out of curiosity, the cast to int brings any speed increase here, compared to just doing the bitwise operation with a 64-bits value?

ChocolArm64/Instructions/SoftFallback.cs
```diff
@@ -698,11 +698,6 @@ public static ulong BinaryUnsignedSatQAcc(long op1, ulong op2, CpuThreadState st
 
         public static ulong CountSetBits8(ulong value) // "size" is 8 (SIMD&FP Inst.).
         {
-            if (value == 0xfful)
```
gdkchan: I guess this is not a common case and the "optimization" is actually harmful?

ChocolArm64/Instructions/InstEmitSimdMove.cs
gdkchan: I think `InstEmitSimdHelper` is a good place to add those arrays shared by multiple instructions impl.

ChocolArm64/Instructions/SoftFallback.cs
```diff
@@ -688,7 +688,7 @@ public static ulong BinaryUnsignedSatQAcc(long op1, ulong op2, CpuThreadState st
             do
             {
                 nibbleIdx -= 4;
-                preCount = ClzNibbleTbl[(value >> nibbleIdx) & 0b1111];
+                preCount = ClzNibbleTbl[(int)(value >> nibbleIdx) & 0b1111];
```
LDj3SNuD: Avoid `conv.ovf.i.un`.

ChocolArm64/Instructions/SoftFallback.cs
LDj3SNuD: Thanks for the review; not really, it's a `conv.i4` instead of a `conv.i8`, adopted more than anything to avoid having to write `if (((value >> bit) & 0b1ul) != 0ul)`.

PR #569, opened 2019-01-31
[WIP] Support loading DLC data from XCIs

gdkchan:
The plan was implementing it properly using the current content manager, but it turns out that it is a bit limited and can only handle NCAs installed to the "nand" (inside ryufs nand path), and can't load AOC (Add-on content) data embedded on the cart itself. So, rather than rewriting some parts of it to handle both cases properly (which would require some mounting support), I just added a quick check that gets the cart data instead depending on the storage id.

I don't plan to work in improving the content manager, but I'm open to suggestions to improve this implementation.

Should help with the following games (needs retesting):

https://github.com/Ryujinx/Ryujinx-Games-List/issues/147

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Ns/IAddOnContentManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Ns/IAddOnContentManager.cs
@@ -14,27 +14,46 @@ namespace Ryujinx.HLE.HOS.Services.Ns
         {
             _commands = new Dictionary<int, ServiceProcessRequest>
             {
-                { 2, CountAddOnContent },
-                { 3, ListAddOnContent  }
+                { 2, CountAddOnContent     },
+                { 3, ListAddOnContent      },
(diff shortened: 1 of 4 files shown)


Ryujinx.HLE/HOS/Services/Ns/IAddOnContentManager.cs
```diff
@@ -14,27 +14,46 @@ public IAddOnContentManager()
...
         {
-            context.ResponseData.Write(0);
-
-            Logger.PrintStub(LogClass.ServiceNs);
+            context.ResponseData.Write(context.Device.System.ContentManager.GetCurrentApplicationAocDataCount());
 
             return 0;
         }
 
         public static long ListAddOnContent(ServiceCtx context)
```
gdkchan: I'm not sure that this implementation is correct.

gdkchan:
Closing since this is really old and I don't plan to finish anymore.

PR #571, opened 2019-02-04
Optimize address translation and write tracking on the MMU

gdkchan:
This is a slightly opimitzed implementation of the address translation.

The previous method uses a 2-level page table (to save memory), all reads/writes are done by calling methods inside the `MemoryManager`. With this change, the page table is now one dimensional, retrieving the base address from it thus requires less instructions.

Previously, it also had to check if the address is aligned, and if it isn't, it does instead 2 accesses of with a smaller size. This was done to handle the cases where the access may cross a page boundary, and the 2 pages may be mapped at non-contiguous physical addresses. It also had to validate if the address is valid, before attempting to access the page table. This is done by checking if the bits above the address space limit are 0. Those two steps were combined into a single check, by combining the unused address space bits and the alignment mask into a single mask. It then ANDs this mask with the address, and check if the result is 0, if so, it takes a fast path, were it just loads the "physical address" from the page table, and then adds the page offset. For the case were the address is invalid or unaligned, it then calls the read/write methods on the `MemoryManager`.

Some problems, and how they were solved:
- Write tracking: We need to be able to detect when the game writes to regions that are used by the gpu. When the game writes there, we need to invalidate the gpu resources, sending the new data to the gpu when needed. Previously this was handled by setting the pointer on the page table of the regions that are mapped on the gpu mem to null. The invalid pointer handler was able to detect that, fire an event to invalidate the gpu resource, then restore the pointer and do the memory access normally.
With this method, this is not necessary anymore. Instead, it uses the Windows API `GetWriteWatch` function, which was designed exactly for this purpose.
- Access to unmapped regions: Unmapped regions of memory will have their pointers set to 0 (null) on the page table, since `VirtualAlloc` ensures that memory is initialized to zero on first access. However this is obviously not a valid address, but the .NET runtime catches the exception raised by an access to the zero page, so in this case it throws a `NullReferenceException`, with a stack trace. This may or may not be good enough...
- Higher memory usage: Physical memory is only really going to be allocated on the pages that are *really* in use, so this method doesn't use more physical memory than the previous one. However it does need to commit memory for the entire page table (around 1GB in the worst case), so the system needs to at least have enough space available on the ram/swap, and it will reserve that. It may be a problem for users with low ram. The biggest problem is the multiple process scenario, in this case, it would commit 1GB *per process*. So, it just uses the more efficient 1GB page table for the game, and any other process should use a multi-level page table, with lower commit size (should be only a few MBs).
- Slow boot times: This address translation method produces more IL per memory access. That means more work for the jit, with in turn means longer compilation times. This is addressed by applying the optimization only for "tier 1" code being translated on the background.


gdkchan:
It's not as good as what I wanted to implement, but at least it's an improvement... Thanks.

jduncanator:
Needs a rebase.

gdkchan:
Ready for review.

samuelGrahame:
just ran sonic mania - in game was 90-100 fps average and now is 115-130 fps on my pc nice work :)

PR #575, opened 2019-02-07
Misc. CPU optimizations

gdkchan:
This implements two optimizations:

- Do not save caller saved registers on the thread state, when the function is "complete".  Caller saved register, as the name implies, should be saved by the caller, after calling any function, and the callee is free to modify those registers. If the function needs to preserve the values on those registers after a call, it is supposed to push it to the stack, and pop after the function call. So, we don't need to save those registers on the thread state, since they should only be used locally (within a function).

Of couse, this optimization is only valid if the code is following the arm64 abi (which all the applications should be following), but anyway, I consider this a rather "dangerous" optimization, so I added a new config entry to disable it, called `enable_aggressive_cpu_opts`.

- Do not save registers that are not read by the callee function before calls, when the callee is know, and we are sure that those registers are going to be written by the callee (or a function called by the callee). This also avoids some unnecessary register writes to the thread state, for the cases where a direct call is made.

- Only "synchronize" the thread when needed. We need to emit some code to ensure that the thread will stop running, so other thread can run on the core when a new thread is selected by the scheduler. This is done by checking some fields on the `ThreadState`, but the previous code was quite inefficient as it was doing those checks at the start of *every* basic block. Now, we just check it at the start of a function, and before a backwards jump (to prevent a long running or infinite loop from keeping the thread going forever).

### DIFF
--- a/Ryujinx/_schema.json
+++ b/Ryujinx/_schema.json
@@ -399,6 +399,17 @@
         false
       ]
     },
+    "enable_aggressive_cpu_opts": {
+      "$id": "#/properties/enable_aggressive_cpu_opts",
+      "type": "boolean",
+      "title": "Enable Aggressive CPU Optimizations",
+      "description": "Enable or disable aggressive CPU optimizations",
+      "default": true,
+      "examples": [
+        true,
+        false
+      ]
+    },
     "controller_type": {
       "$id": "#/properties/controller_type",
       "type": "string",
--- a/Ryujinx/_schema.json
+++ b/Ryujinx/_schema.json
@@ -17,6 +17,7 @@
     "enable_vsync",
     "enable_multicore_scheduling",
     "enable_fs_integrity_checks",
+    "enable_aggressive_cpu_opts",
     "controller_type",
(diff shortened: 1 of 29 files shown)


gdkchan:
Ready for review.

Please note that it still contains `Additionally, as an experiment I disabled the "tier 0", low quality version of the code.`. So I'd be interested to hear if it causes any significant regression on boot times. If not, then I plan to just remove all the code related to "tier 0" and simplify the code. The benefit of doing that is increased performance, and simpler code. The downside is potentially slower boot times.

gdkchan:
Enabled tier 0 again, due to some regressions on boot times that I found while testing it. Also should have fixed the perf regressions caused by the tier 0 "low quality" code. Should be good now, testing is always nice.

Ryujinx/_schema.json
```diff
@@ -399,6 +399,17 @@
         false
       ]
     },
+    "enable_aggressive_cpu_opts": {
```
jduncanator: Need to add this to the `required` property at the top of the schema

PR #577, opened 2019-02-12
Support HomeBrew Loader

Thog:
This PR makes HBL working inside Ryujinx.

It also add back the RomFS support for NROs and fix other errors related to homebrew loading.

### DIFF
--- a/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
+++ b/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
@@ -1,5 +1,6 @@
 ﻿using Ryujinx.HLE.HOS;
 using Ryujinx.HLE.HOS.Services.FspSrv;
+using Ryujinx.HLE.Utilities;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -278,6 +279,31 @@ namespace Ryujinx.HLE.FileSystem
             }
 
             throw new InvalidOperationException($"Path {path} is not a child directory of {_rootPath}");
+        }
+
+
+        public (DateTime, DateTime, DateTime) GetFileTimeStampRaw(string name)
+        {
+            CheckIfDescendentOfRootPath(name);
+
+            DateTime creationDate   = DateTime.UnixEpoch;
+            DateTime modifiedDate   = DateTime.UnixEpoch;
+            DateTime lastAccessDate = DateTime.UnixEpoch;
+
+            if (File.Exists(name))
+            {
+                creationDate = File.GetCreationTime(name);
+                modifiedDate = File.GetLastWriteTime(name);
+                lastAccessDate = File.GetLastAccessTime(name);
+            }
+            else if (Directory.Exists(name))
+            {
+                creationDate = Directory.GetCreationTime(name);
+                modifiedDate = Directory.GetLastWriteTime(name);
+                lastAccessDate = Directory.GetLastAccessTime(name);
+            }
+
+            return (creationDate, modifiedDate, lastAccessDate);
         }
     }
 }
--- a/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
+++ b/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
@@ -281,7 +281,6 @@ namespace Ryujinx.HLE.FileSystem
             throw new InvalidOperationException($"Path {path} is not a child directory of {_rootPath}");
         }
 
-
         public FileTimestamp GetFileTimeStampRaw(string name)
         {
             CheckIfDescendentOfRootPath(name);
--- a/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
+++ b/Ryujinx.HLE/FileSystem/FileSystemProvider.cs
@@ -282,28 +282,33 @@ namespace Ryujinx.HLE.FileSystem
         }
 
 
-        public (DateTime, DateTime, DateTime) GetFileTimeStampRaw(string name)
+        public FileTimestamp GetFileTimeStampRaw(string name)
         {
             CheckIfDescendentOfRootPath(name);
 
-            DateTime creationDate   = DateTime.UnixEpoch;
-            DateTime modifiedDate   = DateTime.UnixEpoch;
-            DateTime lastAccessDate = DateTime.UnixEpoch;
+            DateTime creationDateTime   = DateTime.UnixEpoch;
+            DateTime modifiedDateTime   = DateTime.UnixEpoch;
+            DateTime lastAccessDateTime = DateTime.UnixEpoch;
 
             if (File.Exists(name))
             {
-                creationDate = File.GetCreationTime(name);
-                modifiedDate = File.GetLastWriteTime(name);
-                lastAccessDate = File.GetLastAccessTime(name);
+                creationDateTime   = File.GetCreationTime(name);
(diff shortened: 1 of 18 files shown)


Ryujinx.HLE/FileSystem/FileSystemProvider.cs
```diff
@@ -279,5 +280,30 @@ public void CheckIfDescendentOfRootPath(string path)
 
             throw new InvalidOperationException($"Path {path} is not a child directory of {_rootPath}");
         }
+
+
+        public (DateTime, DateTime, DateTime) GetFileTimeStampRaw(string name)
```
jduncanator: This is probably better suited to having a struct for the return type, rather than 3 date times in a tuple. It can be confusing from the callee which return value is which (given they're all the same type).

Ryujinx.HLE/HOS/HomebrewRomFs.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using System.IO;
+
+namespace Ryujinx.HLE.HOS
+{
+    class HomebrewRomFs : Stream
```
jduncanator: Would be better to call this `HomebrewRomFsStream` to keep in line with naming standards, and to indicate its a stream.

Ryujinx.HLE/HOS/HomebrewRomFs.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        private Stream _baseStream;
+        private long   _basePosition;
```
jduncanator: `_basePosition` is a bit misleading, it's more of an offset. Given the stream it wraps is called `_baseStream` probably best to call this `_positionOffset` or something, otherwise it infers that the position is somehow related to the base stream.

Ryujinx.HLE/HOS/HomebrewRomFs.cs
```diff
@@ -0,0 +1,69 @@
...
+            _basePosition = basePosition;
+
+            _baseStream.Position = _basePosition;
+        }
+
+        public override bool CanRead => _baseStream.CanRead;
+
+        public override bool CanSeek => _baseStream.CanSeek;
+
+        public override bool CanWrite => _baseStream.CanWrite;
```
jduncanator: Set this to false, given `Write` is not implemented.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -549,14 +551,58 @@ public void LoadProgram(string filePath)
 
             bool isNro = Path.GetExtension(filePath).ToLower() == ".nro";
 
-            using (FileStream input = new FileStream(filePath, FileMode.Open))
+            FileStream input = new FileStream(filePath, FileMode.Open);
```
jduncanator: What disposes of `input`, now that it's not wrapped in a using directive? Is it disposed by the device filesystem at applications exit?

Ryujinx.HLE/HOS/Services/Ns/IApplicationManagerInterface.cs
```diff
@@ -9,14 +13,221 @@ class IApplicationManagerInterface : IpcService
...
+                Buffer.BlockCopy(isbnData, 0, isbn, 0, isbnData.Length);
+            }
+
+            context.Memory.WriteBytes(position, isbn);
+            position += isbn.Length;
+
+            context.Memory.WriteByte(position, nacp.StartupUserAccount);
+            context.Memory.WriteByte(position + 1, nacp.TouchScreenUsageMode);
+            context.Memory.WriteByte(position + 2, nacp.AocRegistrationType);
+            position += 3;
```
jduncanator: These position increments can be simplified if you use `position++` in the `Write` calls. eg:

```cs
context.Memory.WriteByte(position++, nacp.AocRegistrationType);
```

Thog:
Addressed comments.

Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcMemory.cs
```diff
@@ -386,6 +386,93 @@ private KernelResult UnmapPhysicalMemory(ulong address, ulong size)
             return _process.MemoryManager.UnmapPhysicalMemory(address, size);
         }
 
+        public KernelResult MapProcessCodeMemory64(int handle, ulong dst, ulong src, ulong size)
+        {
+            return MapProcessCodeMemory(handle, dst, src, size);
+        }
+
+        public KernelResult MapProcessCodeMemory(int handle, ulong dst, ulong src, ulong size)
```
gdkchan: Missing some checks that the kernel does (Dst is inside address space, is inside valid range, is outside alias region, and is outside heap region, and also that size is valid, that is, address + size does not overflow).

Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcMemory.cs
```diff
@@ -386,6 +386,93 @@ private KernelResult UnmapPhysicalMemory(ulong address, ulong size)
...
+
+            return targetProcess.MemoryManager.MapProcessCodeMemory(dst, src, size);
+        }
+
+        public KernelResult UnmapProcessCodeMemory64(int handle, ulong dst, ulong src, ulong size)
+        {
+            return UnmapProcessCodeMemory(handle, dst, src, size);
+        }
+
+        public KernelResult UnmapProcessCodeMemory(int handle, ulong dst, ulong src, ulong size)
```
gdkchan: Ditto.

Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcMemory.cs
```diff
@@ -386,6 +386,93 @@ private KernelResult UnmapPhysicalMemory(ulong address, ulong size)
...
+
+            return targetProcess.MemoryManager.UnmapProcessCodeMemory(dst, src, size);
+        }
+
+        public KernelResult SetProcessMemoryPermission64(int handle, ulong src, ulong size, MemoryPermission permission)
+        {
+            return SetProcessMemoryPermission(handle, src, size, permission);
+        }
+
+        public KernelResult SetProcessMemoryPermission(int handle, ulong src, ulong size, MemoryPermission permission)
```
gdkchan: This one is issing the address space, overflow and permission checks. Valid permissions are values <= 5, that is:
```
0 ---
1 --R
2 -W-
3 -WR
4 X--
5 X-R
```
WX or RWX is not allowed. Additionally, for this svc in particular, The X bit being set is also not allowed. So, only none, read, write or read/write is allowed.

Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcMemory.cs
```diff
@@ -386,6 +386,124 @@ private KernelResult UnmapPhysicalMemory(ulong address, ulong size)
...
+            {
+                return KernelResult.InvalidAddress;
+            }
+
+            if (!PageAligned(size) || size == 0)
+            {
+                return KernelResult.InvalidSize;
+            }
+
+            if (permission > MemoryPermission.ReadAndExecute || ((1 << (int)permission) & 0x2B) == 0)
```
gdkchan: Imo, doing multiple checks if better than `((1 << (int)permission) & 0x2B) == 0`, and this is quite hard to understand.

PR #583, opened 2019-02-13
Fix XMAD shader instruction, gl_FrontFacing and enable face culling

gdkchan:
Some games needs face culling to render correctly. I know it caused issues with some games (not rendering) on the past, but ignoring it also causes issues, so... I think it's better to just enable it, and try fixing any possible issues arising from it later.

### DIFF
--- a/Ryujinx.Graphics/Gal/Shader/ShaderDecodeAlu.cs
+++ b/Ryujinx.Graphics/Gal/Shader/ShaderDecodeAlu.cs
@@ -1268,13 +1268,13 @@ namespace Ryujinx.Graphics.Gal.Shader
                 {
                     ShaderIrOperImm Imm31 = new ShaderIrOperImm(31);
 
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics/Gal/Shader/ShaderDecodeAlu.cs
```diff
@@ -1196,40 +1242,53 @@ private static void EmitXmad(ShaderIrBlock Block, long OpCode, ShaderOper Oper)
...
+                    OperC = new ShaderIrOp(ShaderIrInst.Add, OperC, OperBLsh16);
+
+                    break;
+                }
+
+                case XmadMode.Csfu:
+                {
+                    ShaderIrOperImm Imm31 = new ShaderIrOperImm(31);
+
+                    ShaderIrOp SignAjustA = new ShaderIrOp(ShaderIrInst.Lsr, OperA,  Imm31);
```
jduncanator: Typo here, assume this is supposed to be `SignAdjustA`. Same goes for the other `Ajust` in this block.

PR #588, opened 2019-02-16
Config: Correct MulticoreScheduling configuration casing

jduncanator:
A typo in the MulticoreScheduling configuration property prevented it from being deserialized correctly.

I'll PR some further changes later to better catch these hard to spot issues.


BaronKiko:
Lgtm

PR #592, opened 2019-02-17
Added "ROM not found." message if specified file is invalid.

### LINKED ISSUE #590: Specifying a game path that doesn't exist prints no error message
Launching Ryujinx with a path to a game directory or image that doesn't exist prints no log messages in the console, and the window remains open as if it's loading. This can be confusing, and should print a "ROM not found" message.

AlexLavoie42:
Fixes #590

### DIFF
--- a/Ryujinx/Program.cs
+++ b/Ryujinx/Program.cs
@@ -77,12 +77,12 @@ namespace Ryujinx
                 }
                 else 
                 {
-                    Logger.PrintInfo(LogClass.Application, "ROM not found.");
+                    Logger.PrintWarning(LogClass.Application, "Please specify a valid XCI/NCA/NSP/PFS0/NRO file");
                 }
             }
             else
             {
-                Logger.PrintInfo(LogClass.Application, "Please specify the folder with the NSOs/IStorage or a NSO/NRO.");
+                Logger.PrintWarning(LogClass.Application, "Please specify the folder with the NSOs/IStorage or a NSO/NRO.");
             }
 
             using (GlScreen screen = new GlScreen(device, renderer))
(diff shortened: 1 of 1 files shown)


Ryujinx/Program.cs
```diff
@@ -75,6 +75,9 @@ static void Main(string[] args)
                             break;
                     }
                 }
+                else {
```
jduncanator: New line after else and before the opening brace.

Ryujinx/Program.cs
```diff
@@ -75,6 +75,9 @@ static void Main(string[] args)
...
+                    Logger.PrintInfo(LogClass.Application, "ROM not found.");
```
BaronKiko: This message could do with more detail.
Like the other message below. Something like ""Please specify the game XCI/NSP". Make sure to include all supported formats.
Also this should probably be a warning, not info as you can't do much without a game.

AlexLavoie42: Ya good point, will do.

jduncanator: The other message about no arg passed is an Info as well, if we change one we should change both.

AlexLavoie42: > The other message about no arg passed is an Info as well, if we change one we should change both.

Ya that's why I made it just Info. Should I change them both to warnings or errors?

jduncanator: Make them both warnings. Error is too in your face, imo

AlexLavoie42: Ok true.

AlexLavoie42:
Not sure if I got all the supported extensions

BaronKiko:
lgtm :)

Issue #586, opened 2019-02-16
Scissor Test breaks games that use multiple viewports


jduncanator:
The scissor testing introduced in 0cd5ba03fe4cccd232cdb6aa7e8648b4a08a0ab1 causes some games to display blank, or partially blank. It's believed this is due to issues with multiple viewports.

Sparkle 2 is known to be affected on Intel iGPU’s:

Without scissor test:
![Without Scissor Test](https://user-images.githubusercontent.com/1518948/52894552-6efce600-31ff-11e9-8ee7-53e1b25e74fe.png)

With scissor test:
![With Scissor Test](https://user-images.githubusercontent.com/1518948/52894551-6c01f580-31ff-11e9-87ca-701fbcc98caf.png)

Fixed by #595 

cc: @BaronKiko @gdkchan

BaronKiko:
This can be closed

PR #595, opened 2019-02-18
Fixed Scissor Test on Intel based GPUs

BaronKiko:
I have reworked how it works out which scissor tests are active and added limits in to avoid the issues created on intel gpus listed here: https://github.com/Ryujinx/Ryujinx/issues/586
Hopefully that's the end of it -_-

Needs some handling of errors with unsupported features.


BaronKiko:
Added handling of viewport index not being supported. Falls back to applying first scissor test to all viewports. Gives warning or error in new strict open gl mode.
Please review

BaronKiko:
Updated schema, that should be the last of it

Ryujinx/_schema.json
```diff
@@ -399,6 +399,17 @@
         false
       ]
     },
+    "enable_strict_opengl": {
+      "$id": "#/properties/enable_strict_opengl",
+      "type": "boolean",
+      "title": "Enable Strict OpenGL",
+      "description": "Required OpenGL extensions/features will throw exceptions rather than potentially problamatic fallbacks",
```
gdkchan: Typo: problamatic.

Ryujinx.Graphics/Graphics3d/NvGpuEngine3d.cs
```diff
@@ -415,38 +418,118 @@ private void SetStencil(GalPipelineState State)
...
 
-                    State.ScissorTestY[Index]      = (int)((ScissorVertical & 0xFFFF));                                         // Y, lower 16 bits
-                    State.ScissorTestHeight[Index] = (int)((ScissorVertical >> 16)) - State.ScissorTestY[Index];                // Height, top side is upper 16 bits
+                    int width  = Math.Abs(right - left);
+                    int height = Math.Abs(top   - bottom);
 
-                    // Y coordinates may have to be flipped
-                    if ((int)State.FlipY == -1)
+                    // If the scissor test covers the whole possible viewport, i.e. uninititalized, disable scissor test
+                    if ((width > 16383 && height > 16383) || width <= 0 || height <= 0)
```
gdkchan: I think it would be better to use hex here (0x3FFF rather than 16383). And also using a constant with a descriptive name would be better than adding those "magic numbers" here directly.

Ryujinx.Graphics/Gal/OpenGL/OGLExtension.cs
```diff
@@ -25,7 +28,43 @@ private static bool HasExtension(string Name)
                 }
             }
 
+            Logger.PrintWarning(LogClass.Gpu, $"OpenGL extension {Name} unavailable. You may experience some rendering issues or performance degredation");
+
             return false;
         }
+
+        public static class Strict
```
gdkchan: Is throwing really necessary? I think the warning is enough.

BaronKiko: I think it would be nice to optionally stop using fall backs that may cause rendering, or other, issues.
By default it does only give the warning. This setting can be applied in other areas in future.

Ryujinx/_schema.json
jduncanator: Problematic **

BaronKiko:
Fixed 2 of gdk's issues. Only the outstanding stict stuff. Refer to comment

BaronKiko:
Yet more issues found by @HorrorTroll 
Undertale's rendering was totally broken. I have partially fixed it in latest commit however there is still an issue with borders only showing in the top left. [Example](https://cdn.discordapp.com/attachments/480497101676216340/547301808872095757/unknown.png)
Currently investigating but for now back to WIP ;(

BaronKiko:
Sorted out undertale bug.
Ready for review once more

jduncanator:
Merging this. We can sort out the “strict” quirks if we run into issues with it.

BaronKiko:
Changed the way required extensions are handled. It no longer throws an exception and the Strict OpenGL config option has been removed. Hopefully you are happier with this implementation @gdkchan 
Moved to Math.Clamp over if statements. It's a lot cleaner but technically slower as it always writes the value. The difference is probably negligible.
Also merged with master and fixed conflicts

Ready for another review

PR #596, opened 2019-02-18
Implement ConvertScalingMode properly

gdkchan:
This improves the `ConvertScalingMode` implementation, by correcting some issues, and implementing a missing error check:
- The Src/DstScalingMode enums were swapped.
- One error condition check was missing.

Would be nice to re-test with games that depends on it.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Vi/IApplicationDisplayService.cs
+++ b/Ryujinx.HLE/HOS/Services/Vi/IApplicationDisplayService.cs
@@ -182,7 +182,7 @@ namespace Ryujinx.HLE.HOS.Services.Vi
         {
             SrcScalingMode scalingMode = (SrcScalingMode)context.RequestData.ReadInt32();
 
-            DstScalingMode? convertedScalingMode = ConvetScalingMode(scalingMode);
+            DstScalingMode? convertedScalingMode = ConvertScalingMode(scalingMode);
 
             if (!convertedScalingMode.HasValue)
             {
@@ -197,12 +197,12 @@ namespace Ryujinx.HLE.HOS.Services.Vi
                 return MakeError(ErrorModule.Vi, 6);
             }
 
-            context.ResponseData.Write((ulong)ConvetScalingMode(scalingMode));
+            context.ResponseData.Write((ulong)ConvertScalingMode(scalingMode));
 
             return 0;
         }
 
-        private DstScalingMode? ConvetScalingMode(SrcScalingMode source)
+        private DstScalingMode? ConvertScalingMode(SrcScalingMode source)
         {
             switch (source)
             {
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/HOS/Services/Vi/IApplicationDisplayService.cs
```diff
@@ -204,7 +211,7 @@ public long ConvertScalingMode(ServiceCtx context)
                 case SrcScalingMode.PreserveAspectRatio: return DstScalingMode.PreserveAspectRatio;
             }
 
-            return null;
+            throw new ArgumentException($"Invalid scaling mode \"{source}\" specified.");
```
jduncanator: Should we throw here, or is there some sort of "Generic Error" code we can return from ConvertScalingMode?

gdkchan: Invalid enum values (> 4) returns error code 0x272 on Vi. Maybe we could return the same error code here? It's also possible to implement it as a `Try*` method, return a bool and a out parameter with the converted scaling mode.

jduncanator: I avoided that (out param) in my initial implementation because it made the switch case messy to implement (can't cleanly just return from each case anymore). My initial impl got around this by using a nullable return type, and catching a null in the caller and returning an error message. Perhaps we could do this?

I guess the question is, do we prefer to assert this in the host, or in the guest? I'd imagine returning 0x272 in the guest will cause it to assert?

gdkchan: Imo returning the error code is the right thing to do. This exception will never be hit, because it checks whenever or not the enum value is valid beforehand.

jduncanator: Then I think we should return the error code over throwing an exception. I prefer my initial implementation (returning a nullable) over using an out parameter though, but up to you.

Ryujinx.HLE/HOS/Services/Vi/IApplicationDisplayService.cs
```diff
@@ -180,22 +180,31 @@ public long SetLayerScalingMode(ServiceCtx context)
 
         public long ConvertScalingMode(ServiceCtx context)
         {
-            SrcScalingMode  scalingMode     = (SrcScalingMode)context.RequestData.ReadInt32();
-            DstScalingMode? destScalingMode = ConvetScalingModeImpl(scalingMode);
+            SrcScalingMode scalingMode = (SrcScalingMode)context.RequestData.ReadInt32();
 
-            if (!destScalingMode.HasValue)
+            DstScalingMode? convertedScalingMode = ConvetScalingMode(scalingMode);
```
jduncanator: Looks like there is a typo here, should be `ConvertScalingMode`. Looks like this came across from my code 😰

PR #525, opened 2018-12-04
 Initial non 2D textures support

Thog:
The long-awaited non 2d texture PR is here o/

As stated in the title, this is an initial support, that means that some parts of the codebase haven't been changed to support non 2D textures. (like the copy operation)
Some places in the GLSL decompiler also lack some suffix support.

PR #610, opened 2019-03-03
Compressed cubemap support

BaronKiko:
Implements support for compressed cubemaps.

M2MF doesn't currently support copying compressed textures correctly, however this also affects 2D textures and not just cubemaps so I feel this can be resolved in another PR.

This should allow games that previously booted, but now throw a texture type not implemented exception since the merge of #525, to boot again, excluding games that required Texture Buffer support.

This is basically an implementation of @Thog's code.


jduncanator:
Please rebase on top of master.

PR #650, opened 2019-03-14
Print Guest Stack Trace in ServiceNotImplemented Exception

jduncanator:
When a ServiceNotImplemented exception is thrown, it's useful to obtain a guest stack trace as this usually includes critically important information in understanding what exactly was called and how. The stack trace also includes the SDK function and call stack that was used to call the IPC function, which is useful for deciphering unknown or undocumented service functions.

This PR prints a stack trace in the IPC info dump that is logged when a service function is not implemented. It also adds a few extra functions, and refactors some of the existing code to better accommodate more use-cases (eg. no printing the stack trace directly to the log, rather returning it).

I've also obsoleted the `PrintGuestStackTrace` in `HleProcessDebugger` function, as in nearly every situation, getting the stack trace and then printing it locally allows better control over formatting and logging vs having a function that prints the stack trace directly to the log immediately when called. This function could eventually be removed (the `PrintGuestStackTrace` in `KThread` is now re-implemented using `GetGuestStackTrace`).

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
@@ -67,18 +67,18 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
 
                     string imageName = GetGuessedNsoNameFromIndex(imageIndex);
 
-                    string imageNameAndOffset = $"{imageName}:0x{offset:x8}";
-
-                    trace.AppendLine($" {imageNameAndOffset} {subName}");
+                    trace.AppendLine($"   {imageName}:0x{offset:x8} {subName}");
                 }
                 else
                 {
-                    trace.AppendLine($" ??? {subName}");
+                    trace.AppendLine($"   ??? {subName}");
                 }
             }
 
             //TODO: ARM32.
             long framePointer = (long)threadState.X29;
+
+            trace.AppendLine($"Process: {_owner.Name}, PID: {_owner.Pid}");
 
             while (framePointer != 0)
             {
--- a/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
@@ -4,6 +4,7 @@ using Ryujinx.Common.Logging;
 using Ryujinx.HLE.HOS.Diagnostics.Demangler;
 using Ryujinx.HLE.HOS.Kernel.Memory;
 using Ryujinx.HLE.Loaders.Elf;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -97,6 +96,20 @@ void AppendTrace(long address)
                 framePointer = _owner.CpuMemory.ReadInt64(framePointer);
             }
 
+            return trace.ToString();
+        }
+
+        [Obsolete("Use GetGuestStackTrace")]
```
gdkchan: Atm I don't see a lot of value in obsoleting this method and using the new one, as all the callers are doing the same thing to print the stack trace. But anyway, I think it's better to just remove it if it was already been replaced. What else is using this method right now?

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -97,6 +96,20 @@ void AppendTrace(long address)
...
+        public void PrintGuestStackTrace(CpuThreadState threadState)
+        {
+            EnsureLoaded();
+
+            StringBuilder trace = new StringBuilder();
+
+            trace.AppendLine("Guest stack trace:");
+
```
gdkchan: nit: The new line is not present on the other places where the same code is also used. Imo it should be removed.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -68,13 +67,13 @@ void AppendTrace(long address)
 
                     string imageName = GetGuessedNsoNameFromIndex(imageIndex);
 
-                    string imageNameAndOffset = $"[{_owner.Name}] {imageName}:0x{offset:x8}";
```
gdkchan: Since `_owner.Name` was removed, I think that at the very least is should be present somewhere on the stack trace, otherwise in the case where multiple processes are running, we won't know to which process the stack trace belongs.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
jduncanator: Turns out nothing is using this, I'll remove it.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
jduncanator: I'll print it at the top of the stack trace.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
jduncanator: I've just removed the entire function.

PR #652, opened 2019-03-20
Complete ColorFormat enums

Thog:
After some reversing and hardware testing, I found the previous unknown
color spaces. With this, the ColorFormat definition can be considered
done.


jduncanator:
Thanks!

PR #654, opened 2019-03-28
New shader translator implementation

gdkchan:
This is a WIP implementation of a new shader translator. Some of the goals are:

- Attempt to reconstruct structured control flow on the program. On the old translator, branches are "emulated" with a huge switch, with a case for each possible branch target. This is bad for performance on some GPUs. Here we attempt to reconstruct if/else statements, loops etc.
- Allow some optimizations that are required for the shaders to even compile with some drivers.  For example, AMD requires textureOffset to have constant values on the textureOffset call (also required by the spec), but on Maxwell they are packed on registers rather than being an immediate value (the value on those reigsters is constant however, from what was observed so far).
- There are also opportunity to support things that are not supported by the GPU. For example, half float support requires OpenGL 4.2. However, with some more advanced analysis we can turn most (or all depending on the shader) those FP16 operations into FP32 operations.
- Simpler shaders also means faster compilation.
- Improve readability of the decompiled code. This is important if one wants to understand what the shader is actually trying to do.

Futhermore, this attempts to implement functionality that was missing on the old translator, like for example, condition code flags used on the various integer instructions.

How it works for now:

- The decoder builds a CFG with each block containing a list of OpCodes.
- The translator uses those opcodes to emit code for each respective instruction, producing "Operations", each Operation being composed of a Instruction, a destination Operand, and any number of source Operands.
- There are a bunch Operand types, they are: Constant, LocalVariable, Register, and various "Memory" or "Buffer" operands. "LocalVariables" are in SSA form, they can only be assigned once. However, the translator only uses them for temporary/intermmediate values, and uses "Register" for any value being read/written to a register. "Registers" are changed to "LocalVariable" on the SSA renaming step.
- The instructions includes Branch and MarkLabel instructions. This information is used to build a new CFG.
- From the CFG, the dominators and dominance frontiers are computed.
- After that, SSA renaming is performed, removing all "Register" Operands, and replacing them with "LocalVariable" Operands.
- After that, an optimization step is performed over the code IR.
- The next step is transforming the IR into another IR, I called it "StructuredIr", it's basically an AST.
- The AST transformation step includes structuring the program. All the branch instructions should be eliminated.
- The final step is writing the GLSL code from the AST.

Reasons for some of those choices:

Two IRs:
It's easier to optimize the simple IR, while the tree structure is good for building large expressions. On the first IR, each node can be either an Operation or a Phi function. There's no nesting, and this is simple to handle on the optimizer. On the second IR, you can nest operations to build large expressions creating a tree structure. This can be used to enable expression propagation.

SSA:
It's way easier to perform several optimizations on SSA form. Finding all defs/uses and them making sure you are not breaking anything by moving things around or changing some code is otherwise a nightmare.

In general, I beleive a good amount of work is already done, but there are also a bunch of important things missing that I plan to implement before considering this "ready".

TODO:
- [x] Implement half float instructions (at least the ones that are implemented on the old translator).
- [x] Implement remaining texture instructions that are implemented on the old translator.
- [x] Add support for dual vertex shaders.
- [x] Add support for geometry shaders.
- [x] Properly support shaders with "unstructured" control flow (basically, those that would normally need the use of GOTOs).
- [x] Perform some tests on the implementation.


gdkchan:
This should be ready for review. Other than the testing, I did everything I planned for this PR. Obviously there's still a lot of things that could be improved, but I guess that this PR is already quite large and took too long as it is, so I'd prefer to keep those improvements for the future.

All the instructions that matter for now from the old translator are implemented, so it shouldn't have any regression compared to the old translator.

jduncanator:
Testing
-------

This PR brings with it a significant improvement across the board in AMD GPU performance, with the vast majority of "slow" games I tested going from <10fps up to 50-60fps. Some games still suffer from performance degradation in certain areas or screens, but these are far and few between.

### Celeste

Went from 8fps on the menu up to 55-60fps.

Celeste 2D Menu
Celeste 3D Menu

In-game sits around 60fps with only a few drops below throughout gameplay.

Celeste In-game

### Dragon Quest Heroes I + II Demo

This used to have significant rendering artifacts on the main menu screen, as well as missing particle effects (due to the lack of geometry shader support). We now see a properly rendered menu screen, and particle effects. Performance was fairly decent here on AMD under master, so no noticeable change.

Dragon Quest Heroes I + II Demo master
> master

Dragon Quest Heroes I + II Demo shnew
> shnew

### Axiom Verge

Axiom Verge saw a massive increase from 8fps on master, to nearly 60fps on shnew branch.

Axiom Verge

### Nights of Azure 2: Bridge of the New Moon

This saw a major increase from 10fps to 60fps under this PR. The biggest change is a fix for the incorrect rendering of half float textures, which resolves the blue tint over the entire game.

Nights of Azure 2 master
> master

Nights of Azure 2 shnew
Nights of Azure 2 shnew
> shnew

### Puyo Puyo Tetris!

Puyo used to be pretty hit and miss on AMD, with less than 10fps and a stuttery intro and in-game experience. With this PR, we now get a nice smooth 60fps on the intro and menu, and 30-45fps in-game.

Puyo Puyo Tetris! master
>master

Puyo Puyo Tetris! shnew
Puyo Puyo Tetris! shnew In-game
>shnew


All up, I have to say a massive thanks to @gdkchan for slugging this one out. This is probably the single biggest improvement in performance we've seen in the emulator since it started, and the quality of the generated shader code is miles ahead of anything the old recompiler generated. Great work mate!

PR #655, opened 2019-03-31
Sse optimized all the fp to integer conversion instructions (signed) with Tests (signed & unsigned).

LDj3SNuD:
Fixed a bug about all the fp to integer conversion instructions (signed & unsigned); where all those fp values close to T.MaxValue were mistakenly converted to ~T.MaxValue.

Fixed the instruction table (Fcvtzs_V_Fixed & Fcvtzu_V_Fixed).

Uniformed the use of the zero vector to the float type only.

Nits.

### DIFF
--- a/ChocolArm64/Instructions/InstEmitSimdCvt.cs
+++ b/ChocolArm64/Instructions/InstEmitSimdCvt.cs
@@ -35,7 +35,7 @@ namespace ChocolArm64.Instructions
                     //Single -> Double.
                     Type[] typesCvt = new Type[] { typeof(Vector128<double>), typeof(Vector128<float>) };
 
-                    VectorHelper.EmitCall(context, nameof(VectorHelper.VectorDoubleZero));
+                    VectorHelper.EmitCall(context, nameof(VectorHelper.VectorSingleZero));
                     context.EmitLdvec(op.Rn);
 
                     context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.ConvertScalarToVector128Double), typesCvt));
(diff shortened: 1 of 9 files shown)


ChocolArm64/Instructions/InstEmitSimdCvt.cs
```diff
@@ -730,5 +803,314 @@ private static void EmitI2fFBitsMul(ILEmitterCtx context, int size, int fBits)
...
         }
+
+        private static void EmitSse41Fcvt_Signed_Gp(ILEmitterCtx context, RoundMode roundMode, bool isFixed)
+        {
+            OpCodeSimdCvt64 op = (OpCodeSimdCvt64)context.CurrOp;
+
+            if (op.Size == 0)
+            {
+                Type[] typesCmpMul = new Type[] { typeof(Vector128<float>), typeof(Vector128<float>) };
+                Type[] typesAnd    = new Type[] { typeof(Vector128<long> ), typeof(Vector128<long> ) };
```
gdkchan: nit: Why the spacing before the `)`? IIRC we were aligning it after the comma.

ChocolArm64/Instructions/InstEmitSimdCvt.cs
```diff
@@ -730,5 +803,314 @@ private static void EmitI2fFBitsMul(ILEmitterCtx context, int size, int fBits)
...
+                    context.EmitCall(typeof(Sse).GetMethod(nameof(Sse.Multiply), typesCmpMul));
+                }
+
+                context.EmitCall(typeof(Sse41).GetMethod(GetSse41NameRnd(roundMode), typesRndCvt));
+
+                context.EmitStvectmp();
+                context.EmitLdvectmp();
+
+                // TODO: Use Sse.ConvertToInt64 once it is fixed,
+                // remove the following if/else and comment out the code.
```
gdkchan: `comment out` -> `uncomment`?

jduncanator: Might also be worth adding the "ETA" for the function to be fixed, in this instance, .NET Core 3.0, so that we can quickly find things we can update when .NET Core 3.0 lands.

ChocolArm64/Instructions/InstEmitSimdHelper.cs
```diff
@@ -1367,8 +1367,8 @@ public static void EmitVectorZeroUpper(ILEmitterCtx context, int reg)
         {
             if (Optimizations.UseSse)
             {
-                //TODO: Use Sse2.MoveScalar once it is fixed,
-                //as of the time of writing it just crashes the JIT (SDK 2.1.504).
```
gdkchan: This is another intrinsic that will only be fixed on 3.0. I don't think they backport intrinsic related fixes as they are not officially supported pre-3.0, they were in a "experimental" state.

LDj3SNuD:
Address PR feedback.

PR #656, opened 2019-04-07
Sse optimized the Vector & Scalar fp-to-integer conversion instructions (unsigned); improved the related Tests.

LDj3SNuD:
This PR is intended as a complement to #655.

The Sse optimizations related to the General Purpose fp-to-integer conversion instructions have been removed due to the poor performance improvement in relation to the size and complexity of the signed and unsigned (above all) implementations.

### DIFF
--- a/ChocolArm64/Instructions/InstEmitSimdCvt.cs
+++ b/ChocolArm64/Instructions/InstEmitSimdCvt.cs
@@ -209,7 +209,7 @@ namespace ChocolArm64.Instructions
         {
             if (Optimizations.UseSse41)
             {
-                EmitSse41Fcvt_Signed(context, RoundMode.ToNearest, isFixed: false, scalar: true);
+                EmitSse41Fcvt_Signed(context, RoundMode.ToNearest, scalar: true);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


ChocolArm64/Instructions/InstEmitSimdCvt.cs
```diff
@@ -804,168 +811,140 @@ private static void EmitI2fFBitsMul(ILEmitterCtx context, int size, int fBits)
...
                 context.EmitLdvec(op.Rn);
 
-                context.EmitCall(typeof(Sse2).GetMethod(nameof(Sse2.And), typesAnd));
+                context.EmitCall(typeof(Sse).GetMethod(nameof(Sse.And), types));
 
                 if (isFixed)
                 {
-                    // BitConverter.Int32BitsToSingle(fpScaled) == MathF.Pow(2f, op.FBits)
-                    int fpScaled = 0x40000000 + (op.FBits - 1) * 0x800000;
+                    int fBits = GetImmShr((OpCodeSimdShImm64)op);
```
gdkchan: Rather than having the `isFixed` argument and casting the op to `OpCodeSimdShImm64` to get the fixed point scale value, you could do something like:
```cs
if (op is OpCodeSimdShImm64 fixedOp)
{
    int fBits = GetImmShr(fixedOp);
    //...
}
```
And then the `ifFixed` argument woudn't be needed anymore, as all the `*Fixed` variants will also use this `OpCode*` type.

Peharps the current way is easier to understand though. What do you think?

LDj3SNuD:
Address PR feedback; re-tested.

PR #657, opened 2019-04-11
Add missing TextureCubeMapArray texture type entry

jduncanator:
Add missing `TextureCubeMapArray` texture type alias to other texture creation path.

### DIFF
--- a/Ryujinx.Graphics/Gal/OpenGL/OglTexture.cs
+++ b/Ryujinx.Graphics/Gal/OpenGL/OglTexture.cs
@@ -95,6 +95,7 @@ namespace Ryujinx.Graphics.Gal.OpenGL
                         type,
                         IntPtr.Zero);
                     break;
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics/Gal/OpenGL/OglTexture.cs
```diff
@@ -95,6 +95,7 @@ public void Create(long key, int size, GalImage image)
                         type,
                         IntPtr.Zero);
                     break;
+                case TextureTarget.TextureCubeMapArray:
```
Thog: You should add this comment like in the other definitions.
```suggestion
                // Cube map arrays are just 2D texture arrays with 6 entries
                // per cube map so we can handle them in the same way
                case TextureTarget.TextureCubeMapArray:
```

BaronKiko:
Other than the comment as thog suggested it look good to me.

PR #658, opened 2019-04-14
Config option to ignore missing services

BaronKiko:
Adds a config option to ignore missing services. This allows some games to boot with warnings for missing services.
This also provides a quick way to see how many, and what, services are needed to boot a specific game without this option.

Test case, bloons td 5 doesn't boot because of network service requirements however with this it manages to get in game. While slow and containing some graphical errors in the main menu it seems playable.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/IpcService.cs
+++ b/Ryujinx.HLE/HOS/Services/IpcService.cs
@@ -6,7 +6,6 @@ using Ryujinx.HLE.HOS.Kernel.Ipc;
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Net.Security;
 
 namespace Ryujinx.HLE.HOS.Services
 {
--- a/Ryujinx.HLE/HOS/Services/IpcService.cs
+++ b/Ryujinx.HLE/HOS/Services/IpcService.cs
@@ -6,6 +6,7 @@ using Ryujinx.HLE.HOS.Kernel.Ipc;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Net.Security;
 
 namespace Ryujinx.HLE.HOS.Services
 {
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/HOS/Services/IpcService.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Net.Security;
```
gdkchan: huh

BaronKiko: Not sure how that got added, will remove in 1 sec

Ryujinx.HLE/HOS/Services/IpcService.cs
```diff
@@ -89,13 +89,29 @@ public void CallMethod(ServiceCtx context)
...
 
-                long result = processRequest(context);
+                    result = processRequest(context);
+                }
+                else
+                {
+                    string serviceName;
+                    DummyService dummyService = service as DummyService;
+
+                    serviceName = (dummyService == null) ? serviceName = service.GetType().FullName : dummyService.ServiceName;
```
jduncanator: You are assigning to serviceName twice on this line, you can remove the second assignment, as it does nothing.

```suggestion
                    serviceName = (dummyService == null) ? service.GetType().FullName : dummyService.ServiceName;
```

Ryujinx/Config.jsonc
```diff
@@ -44,6 +44,9 @@
     // Enable or disable aggressive CPU optimizations
     "enable_aggressive_cpu_opts": true,
 
+    // Enable or disable ignoring missing services, this may cause instability however lets some games progress further
```
jduncanator: Can we change this text a bit? I'd prefer it to just say: 
```
Enable or disable ignoring missing services, this may cause instability
```

Ryujinx/_schema.json
```diff
@@ -411,6 +411,17 @@
         false
       ]
     },
+    "ignore_missing_services": {
+      "$id": "#/properties/ignore_missing_services",
+      "type": "boolean",
+      "title": "Ignore Missing Services",
+      "description": "Enable or disable ignoring missing services, this may cause instability however lets some games progress further",
```
jduncanator: Ditto:
```
Enable or disable ignoring missing services, this may cause instability
```

Ryujinx/Configuration.cs
```diff
@@ -11,6 +11,7 @@
 using System.Threading.Tasks;
 using Utf8Json;
 using Utf8Json.Resolvers;
+using Ryujinx.HLE.HOS.Services;
```
gdkchan: nit: Sorting, `using Ryujinx.HLE.HOS.Services;` should be above `using System.Threading.Tasks;`.

BaronKiko: Um, ok. Sure I can do that

BaronKiko:
Implemented comments, ready for another review

PR #659, opened 2019-04-16
Toggle VSync Hotkey

BaronKiko:
Allows you to enable and disable vsync at runtime with the a specified hotkey, tab by default.

### DIFF
--- a/Ryujinx/Ui/GLScreen.cs
+++ b/Ryujinx/Ui/GLScreen.cs
@@ -22,6 +22,8 @@ namespace Ryujinx
 
         private IGalRenderer _renderer;
 
+        private HidHotkeyButtons _prevHotkeyButtons = 0;
+
         private KeyboardState? _keyboard = null;
 
         private MouseState? _mouse = null;
@@ -128,6 +130,7 @@ namespace Ryujinx
 
         private new void UpdateFrame()
         {
+            HidHotkeyButtons     currentHotkeyButtons = 0;
             HidControllerButtons currentButton = 0;
             HidJoystickPosition  leftJoystick;
             HidJoystickPosition  rightJoystick;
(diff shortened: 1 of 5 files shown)


Ryujinx/Ui/GLScreen.cs
```diff
@@ -238,6 +241,15 @@ private new void UpdateFrame()
             HidControllerBase controller = _device.Hid.PrimaryController;
 
             controller.SendInput(currentButton, leftJoystick, rightJoystick);
+
+            // Toggle vsync
+            if ((currentHotkeyButtons & HidHotkeyButtons.ToggleVSync) != 0 &&
```
emmauss: woudn't this toggle vsync, even if the key was pressed with other keys?

jduncanator: Depends, out of the box it’s TAB, which I doubt will be used with another key. But given it can be changed, it might be better to check if the only key pressed is the hotkey?

Also, I prefer Enum.HasFlag over testing the bit directly with an AND but I’m impartial, so doesn’t really matter.

BaronKiko: I didn't know of enum.hasflag. I don't use flags that often. I can use that.

Emmauss has a point, I wonder what would happen if you say, ATL tabbed. I'm not sure only allowing it when pressed on its own is a good idea because you may want to toggle it when walking or something. I could maybe avoid modifier keys pressed in combination? Also I can check if the button press gets through when the window isn't focused

BaronKiko:
Changed to hasflag.
I checked and windows handles alt+tab. Opentk handles out of focus key presses.
I decided it's best to not disable while other keys are pressed. Hopefully you all agree and this is good for merge.

Ryujinx/_schema.json
```diff
@@ -636,9 +636,24 @@
...
+        },
+        "hotkeys": {
+          "$id": "#/properties/keyboard_controls/properties/hotkeys",
+          "type": "object",
+          "title": "Hotkey Controls",
+          "required": [
+            "toggle_vsync"
+          ],
+          "properties": {
+            "button_zr": {
```
jduncanator: This is incorrect, should be `toggle_vsync`
```suggestion
            "toggle_vsync": {
```

PR #664, opened 2019-04-20
Implement IIrSensorServer GetIrsensorSharedMemoryHandle

### LINKED ISSUE #620: irs IIrSensorServer: 304 (GetIrsensorSharedMemoryHandle) not implemented
IIrSensorServer GetIrsensorSharedMemoryHandle is not implemented.

## Service description:
[Switchbrew](https://switchbrew.org/wiki/HID_services#GetIrsensorSharedMemoryHandle) | [Reswitched](https://reswitched.github.io/SwIPC/ifaces.html#nn::irsensor::IIrSensorServer(304))

> Takes a PID-descriptor and an AppletResourceUserId. Returns a SharedMemory handle.
>
> The SharedMemory is mapped with permissions=read-only and size=0x8000.

## Required by:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/114
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/154
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/89

jduncanator:
Resolves #620

### DIFF
--- a/Ryujinx.HLE/HOS/Horizon.cs
+++ b/Ryujinx.HLE/HOS/Horizon.cs
@@ -30,6 +30,7 @@ namespace Ryujinx.HLE.HOS
 
         internal const int HidSize  = 0x40000;
         internal const int FontSize = 0x1100000;
+        internal const int IirsSize = 0x8000;
 
         private const int MemoryBlockAllocatorSize = 0x2710;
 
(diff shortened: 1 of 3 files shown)


jduncanator:
> lgtm (although I don't like the implementation, but that's how the other services using shared mem are implemented so should be ok for now).

I agree, I've never worked on this code before, but it looks like we need a proper way to allocate the shared memory regions, rather than just adding together the offsets of all the previous ones and dumping it all in the constructor of the Horizon class.

I'm up for someone rewriting this, but for now, this is how the rest are done so I think I'll leave it.

PR #677, opened 2019-04-24
Fix GetAudioRendererWorkBufferSize for REV5

### LINKED ISSUE #669: GetAudioRendererWorkBufferSize: Library Revision 0x35564552 is not supported!
## Description:
GetAudioRendererWorkBufferSize: Library Revision 0x35564552 is not supported!

## Attention:
This is unknown function!

## Required by:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/155
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/157
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/161
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/162
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/164

AcK77:
This should be close #669. 
Based on my own RE.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
@@ -100,7 +100,7 @@ namespace Ryujinx.HLE.HOS.Services.Aud
                                (Params.PerformanceManagerCount + 1) + 0x13F) & ~0x3FL;
                 }
 
-                if (IsVariadicCommandBufferSizeSupported)
+                if (isVariadicCommandBufferSizeSupported)
                 {
                     size += Params.EffectCount * 0x840 + 
                             Params.MixCount * 0x5A38 +
--- a/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
@@ -59,7 +59,7 @@ namespace Ryujinx.HLE.HOS.Services.Aud
             if (revision <= Rev)
             {
                 bool isSplitterSupported                  = revision >= 3;
-                bool IsVariadicCommandBufferSizeSupported = revision >= 5;
+                bool isVariadicCommandBufferSizeSupported = revision >= 5;
                 
                 long size;
 
--- a/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
@@ -16,7 +16,7 @@ namespace Ryujinx.HLE.HOS.Services.Aud
                                       ('V' << 16) |
                                       ('0' << 24);
 
-        private const int Rev = 4;
+        private const int Rev = 5;
 
         public const int RevMagic = Rev0Magic + (Rev << 24);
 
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
```diff
@@ -58,8 +58,9 @@ public long GetAudioRendererWorkBufferSize(ServiceCtx context)
 
             if (revision <= Rev)
             {
-                bool isSplitterSupported = revision >= 3;
-
+                bool isSplitterSupported                  = revision >= 3;
+                bool IsVariadicCommandBufferSizeSupported = revision >= 5;
```
jduncanator: Lower case `Is` to match the `isSplitterSupported` above it.``
```suggestion
                bool isVariadicCommandBufferSizeSupported = revision >= 5;
```

Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
```diff
@@ -99,7 +100,21 @@ public long GetAudioRendererWorkBufferSize(ServiceCtx context)
                                (Params.PerformanceManagerCount + 1) + 0x13F) & ~0x3FL;
                 }
 
-                size = (size + 0x1907D) & ~0xFFFL;
+                if (IsVariadicCommandBufferSizeSupported)
+                {
+                    size += Params.EffectCount * 0x840 + 
+                            Params.MixCount * 0x5A38 +
```
jduncanator: Would be nice to align these multiplications. I guess we can't really determine what these constants are either?

Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
```diff
@@ -99,7 +100,21 @@ public long GetAudioRendererWorkBufferSize(ServiceCtx context)
...
+                            Params.SinkCount * 0x148 +
+                            Params.SplitterDestinationDataCount * 0x540 +
+                            Params.VoiceCount * (Params.SplitterCount * 0x68 + 0x2E0) +
+                            ((Params.VoiceCount + Params.MixCount + Params.EffectCount + Params.SinkCount + 0x65) << 6) + 0x3F8;
+                }
+                else
+                {
+                    size += 0x1807E;
```
jduncanator: Is there a reason why this is now `+ 0x1807E` and not `0x1907D` like before?

Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
AcK77: For now we can't determine what there constants are yes... 
But since the RE just look like the pseudocode, it's better for now to leave it as-is because if we need to update it later we will recognize the code easily!

Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
AcK77: image
The reason ^

Ryujinx.HLE/HOS/Services/Aud/IAudioRendererManager.cs
jduncanator: Perhaps its better to break this logic out into another function, like the original code does. `CalculateCommandBufferSize(Params)`?

AcK77: Since the logic is not used at another place (at least in the Emulator), we can keep it here for now and move it later if it's necessary.

HorrorTroll:
Working perfectly with many new games using REV5!

Some old games using old REV doesn't broken, such as:
- Super Mario Odyssey
- Puyo Puyo Tetris S
- ...

PR #680, opened 2019-04-24
Implement gl_ClipDistance on the shader generator

gdkchan:
This implements gl_ClipDistance.
Additionally, for unimplemented attributes, do not produce an invalid assignment for the constant "undef" that is used for unintialized variables. Instead, it now produces a commented out line with the attribute offset, like this:
```glsl
// bad_attr0x2C0 = temp_82;
```

### DIFF
--- a/Ryujinx.Graphics/Shader/CodeGen/Glsl/OperandManager.cs
+++ b/Ryujinx.Graphics/Shader/CodeGen/Glsl/OperandManager.cs
@@ -182,7 +182,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
 
             //TODO: Warn about unknown built-in attribute.
 
-            return "// bad_attr0x" + value.ToString("X");
+            return isOutAttr ? "// bad_attr0x" + value.ToString("X") : "0.0";
         }
 
         public static string GetUbName(GalShaderType shaderType, int slot)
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics/Shader/CodeGen/Glsl/OperandManager.cs
```diff
@@ -172,7 +180,9 @@ private static string GetAttributeName(AstOperand attr, GalShaderType shaderType
                 }
             }
 
-            return DefaultNames.UndefinedName;
+            //TODO: Warn about unknown built-in attribute.
```
Thog: Might be good to put the warning?

gdkchan: there are other places that should warn about unimplemented or unsupported stuff that is also lacking an actual warning. I was planning to implement all of them in one go at some point, producing some sort of report with all warning or errors, that the decompiler would return inside the `ShaderProgram` struct. But I can add a warning here using `Logger.PrintWarning` if desired.

Thog: Ah well it might be better to keep it this way then if you still plan on doing it at some point :+1:

jduncanator: I like the idea of logging warnings and unsupported against the ShaderProgram!

PR #688, opened 2019-05-13
Added missing ++ in BindMacro GPU method

BaronKiko:
As was shown by fincs there should be a ++ in the bind macro method.
There is still another issue but it's 1 bug fixed


BaronKiko:
Can this be merged?

PR #693, opened 2019-05-28
Add a new JIT compiler for CPU code

gdkchan:
### Rationale:

On the old CPU emulator (ChocolArm64), the ARM code was directly translated to CIL, which is in turn compiled by RyuJIT into platform specific code. This has some advantages, and also some disvantages. The main disadvantage is the lack of control. Here is a brief list of issues with the approach used on the old CPU emulator:
- RyuJIT is a bit slow compiling the code. This is undesirable as it may lead to long boot times and lags on the game.
- Sometimes we are not able to emit the optimal code sequence for a given ARM instruction, as we don't control what the JIT is going to produce.
- We can't do AOT builds as System.Reflection.Emit is not supported there.
- We are limited to the platforms that RyuJIT supports.
With our own JIT, we don't have any of those limitations, but the cost is increased complexity as we generate machine code directly. That said, I believe this is the way forward and we can improve it over time to get better performance and throughput.

### Design:

**Translation and compilation:**

ARMeilleure is a ARM CPU emulator. `Meilleure` means better, as it is supposed to be a improvement compared to the old CPU. Currently it can be thought as 2 pieces, the translator and the compiler. My current plan is moving the compiler part into a separate project, to allow it to be reused for other things later. It is already fairly independent, so the work required to move those bits into a new project is minimal.

Step 1 - Translation

Everything starts on the Translator. It works like the old CPU, the decoder decodes the opcodes, builds a CFG with the decoded opcodes, and returns the CFG. It's possible to translate either a single basic block, or an entire function (or approximation of that).

The CFG with the opcodes is then processed by the ARM -> IR translator. It produces a sequence of IR instruction for each ARM instruction. The amount of IR instructions is relatively small, this allows new targets to be easily added. Theres also "intrinsics", they map to a specific instruction on the target arch. For example, on x86, the various SSE SIMD and scalar operations are exposed through those intrinsics. However, to facilitate portability, all the ARM instructions should also have a slow path that doesn't require any platform specific instructions.

Step 2 - Register usage

This step does data flow analysis on the IR CFG and finds the registers that are used on a given function, and also the registers that are written to, and adds the necessary operations to load and store the values from those registers on the execution context.

Step 3 - SSA construction

The optimizer expects the IR to be in SSA form. However, it's also possible to disable the SSA construction step. In this case, no optimization is performed. Instead of doing the SSA construction with the input registers, it just creates one "local variable" for each register. Disabling SSA construction speeds the compilation up a bit.

_The next passes are target specific. Other targets may not need them, or it may be completely different._

Step 4 - Optimization

This does basic optimizations on the code. Some of the optimizations includes dead code elimination, copy propagation, constant folding, and a set of optimizations that are just called "simplification", it removes things like addition, subtraction or shifts by 0, multiplication by 1 etc.

Step 5 - Pre-allocation

Some architectures (like x86) has several restrictions on the registers that can be used in some instruction. For example, on non-immediate shifts, the shift value is implied to be on the CL register. To ensure that the value is on the correct register, we add a copy from the local variable to the correct register, and replace the use with that register.

It is also used to handle things like constant that can't be encoded as immediate on the instruction, either because it is too long, or because it's just not supported on a given instruction.

Step 6 - Register allocation

This allocates one CPU register for each variable. There are currently 2 different register allocator. The first one is Linear scan allocator, and the other one is the Hybrid allocator.

Linear scan:

The linear scan allocator works with lifetime intervals of the variables. When the variable is first assigned, the lifetime of that variable starts. We can say that, at this point, the variable is born. The lifetime of the variable ends on the last use on the code.  In the case of loops, the lifetime of the variable needs to be further extended to the loop end, because it may still be used at some point inside the loop, so it is technically still live.

In order to allocate registers for each variable, we first populate a queue of all variables that needs to be processed, sorted by the lifetime starting point on the program. Given a set F of registers available for allocation, we pick the first register in F and assign it to the current variable being processed, and then add the interval to the set of active intervals. When F is empty, we select the most suitable interval on the set of active intervals, and spill the variable. This will free the register, and the register will then be available again for allocation. We assign that register to the current variable.

This is a over simplified description of the linear scan algorithm. The variant implemented here also includes second-change binpacking and lifetime holes. This means that a variable may be split, and only spilled to memory on a part of it's lifetime, while still getting a chance to live on a register at a later point, while the lifetime holes means that it keeps track of point in the middle of the interval where the variable is not used (which means that the register is free on those regions).

Hybrid:

This is a new register allocation algorithm I came up with. It is meant to be fast (faster than linear scan) at the cost of code quality. The hybrid name is due to the fact that it can be considered both a local and global allocator. A local allocator works only on a single basic block, while loading and spilling variables at the start and end of the basic block. A global allocator works on the context of the whole function, thus producing better results. This is a mix of both.

Given a set of available CPU registers F, we split it in two sets, F1 and F2. The registers on F1 are assigned only for variables that are declared and used only inside a single basic block. The registers on F2 are used for any variable, even those that spans multiple blocks. The F set is not divided in equal parts, F2 should get a much higher number of registers than F1, because variables that are live only inside a single block are usually short-lived, so the register becomes free again very quickly, and we only need a small amount of registers.

Keeping track of variables that are live across multiple blocks is complicated, because in the case of loops, they may actually live even after their last use. We need to do data flow analysis to build accurate lifetime intervals for those variable, which is a costly operation, and this allocator is supposed to be fast. So it instead takes a simpler approach, where lifetime intervals are not even necessary. When a register is assigned to that variable, it is never used again. This means that the register will never become free. This eliminates the need for lifetimes intervals as the register will always hold the correct value, as it is never reused.

Deciding which variable should be assigned a register is complicated for the case described above, as the register is never reused. The approach used is collecting all the variables on a previous pass, and then sorting them by use count in descending order. The variables with the highest use counts are assigned a register, the remaining variables are spilled for their entire lifetime.

This approach ensures that long-lived variables with few uses are spilled, which is exactly what we want. A variable with few, far apart uses wasting a register is the exact kind of thing we don't want to happen. Since no data flow or keep track of active intervals is needed, this approach is much faster than linear scan, while still producing reasonable results. It also works better without SSA transformation, because on the original ARM code, registers are usually reused as much as possible.

Step 7 - Code generation

This emits the final machine code from the IR. The register allocator replaces all "local variable" operands with "register" operands with the final CPU registers. The code generator will emit one or more target instruction for a given IR instruction. It also handles intrinsics and other possible target specific operation. The actual binary code is written by the assembler to the stream.

On x86, the jumps are relative, and the immediate value has multiple sizes, including signed 8-bits and 32-bits. In order to produce the optimal encoding, we first write all the instructions to the stream ignoring all the relative jumps. Then, we rewrite it taking the jumps into account. This is necessary because we need to know the number of bytes the code will take to calculate the jump offset. Additionally, other jumps will also affect that size, so we keep processing them on a loop until a fixpoint is reached.

**Execution:**

Guest function execution:

After the code is compiled, we must allow it to be executed. To do that, we first need to allocate a memory region where the code will reside. Then, we need to reprotect that region to allow it to be executed (the "execute" permission bit must be set). The current allocator is very simple. Freeing memory is not yet supported, so it only needs to keep incrementing the current offset.

The memory pages that are already full (that is, we already wrote code on the entire page) are mapped as RX (Read and execute), as we won't need to write anything there anymore. The current page that is not yet full is mapped as RWX (Read, write and execute). This allow the code that is there to be executed, while we still write new code on the end of the page. The remaining region that is not yet used is mapped as RW (Read and write).

In the future, we should also support invalidating code regions, to support the dynamic code generation scenario. This will require a more robust allocator.

After the code is mapped and reprotected, with `Marshal.GetDelegateForFunctionPointer`, passing a pointer to the location where the code was mapped, we get a delegate that we can use to call the function that was just compiled.

Native interface:

The compiled code also needs to call C# code. There are two types of C# functions we call. One of them is helper methods were more complex ARM instructions are implemented. For example, the AES and SHA instructions are very complex, so emiting IR directly for them is not a good idea, as it would be very large, slow to compile and very complicated to implement, as writing IR emission code is way more complicated than implementing a given algorithm on a high level language directly. The other type of function is the ones necessary to communicate with the "external world", that is, the rest of the emulator. Like, for example, supervisor calls (SVC), firing events for the CPU exceptions (undefined instruction, software breakpoints, and others).

Many of those calls would require calling instance methods of a given object. We can't pass the .NET objects to native code (or at least, I believe that we can't, I never really looked but it sounds like a bad idea), so we use a static class called `NativeInterface` that is exactly as the name suggests, an interface that native code uses to some of those managed methods. Since all the methods there are static, it must keep a per-thread context with the instance of each object it will need to call to, like for example, the `ExecutionContext` and `MemoryManager`.


ARMeilleure/Diagnostics/IRDumper.cs
```diff
@@ -0,0 +1,168 @@
...
+using ARMeilleure.Translation;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ARMeilleure.Diagnostics
+{
+    static class IRDumper
+    {
+        private const string Identation = " ";
```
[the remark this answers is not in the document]
gdkchan: I prefer to keep it, to allow the indentation to be easily changed.

ARMeilleure/Instructions/CryptoHelper.cs
```diff
@@ -0,0 +1,279 @@
...
+            for (int columns = 0; columns <= 3; columns++)
+            {
+                int idx = columns << 2;
+
+                byte row0 = inState[idx + 0]; // A, E, I, M: [row0, col0-col3]
+                byte row1 = inState[idx + 1]; // B, F, J, N: [row1, col0-col3]
+                byte row2 = inState[idx + 2]; // C, G, K, O: [row2, col0-col3]
+                byte row3 = inState[idx + 3]; // D, H, L, P: [row3, col0-col3]
+
+                outState[idx + 0] = (byte)((uint)_gfMul02[row0] ^ _gfMul03[row1] ^ row2 ^ row3);
```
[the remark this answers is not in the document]
gdkchan: I think it's fine as it is, maybe @LDj3SNuD has a opinion here since he wrote this code.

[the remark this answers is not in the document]
LDj3SNuD: `I think it's fine as it is` I think so, too.

ARMeilleure/Optimizations.cs
```diff
@@ -0,0 +1,33 @@
+using ARMeilleure.CodeGen.X86;
+
+namespace ARMeilleure
+{
+    public static class Optimizations
+    {
+        public static bool AssumeStrictAbiCompliance { get; set; }
```
LDj3SNuD: Now, we should assign it a value.

ChocolArm64/Optimizations.cs
```diff
@@ -1,24 +1,27 @@
 using System.Runtime.Intrinsics.X86;
 
-public static class Optimizations
+namespace ChocolArm64
 {
-    public static bool AssumeStrictAbiCompliance { get; set; }
+    public static class Optimizations
+    {
+        public static bool AssumeStrictAbiCompliance { get; set; }
```
LDj3SNuD: Now, we should assign it a value.

ARMeilleure/Optimizations.cs
gdkchan: So it should be set to true by default?

LDj3SNuD: Before it was true by default via `Ryujinx\Config.jsonc` (but this was not always the case).

ChocolArm64/Optimizations.cs
LDj3SNuD: Before it was true by default via `Ryujinx\Config.jsonc`.

PR #697, opened 2019-06-05
Clean Up Build

BaronKiko:
This pr intends to clean up the build directory from AppVeyor builds.
A finished build now looks like this: 
explorer_2019-06-05_05-29-15
The log will appear there after the first run along with the profiler config on a profiled build.
Currently this only works on windows builds because it depends on this wrapper I wrote: https://github.com/BaronKiko/RyujinxWrapper
It needs to be ported to linux, I am hoping somebody else can help me with that as I don't fancy setting up a linux install.
That wrapper also needs an icon set just for prettiness and to be moved to a more sensible place. A fork under the ryujinx account is probably wise.

That is all.

Summarized todo:
- [ ] Linux version of wrapper
- [ ] Wrapper icon
- [ ] Fork wrapper under Ryujinx account
- [ ] Review and cleanup for merge
- [ ] Celebrate

### DIFF
--- a/Ryujinx.Profiler/Profile.cs
+++ b/Ryujinx.Profiler/Profile.cs
@@ -17,9 +17,9 @@ namespace Ryujinx.Profiler
         private static ProfilerSettings _settings;
 
         [Conditional("USE_PROFILING")]
-        public static void Initalize()
+        public static void Initalize(bool wrappedBuild)
         {
-            var config = ProfilerConfiguration.Load(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "ProfilerConfig.jsonc"));
+            var config = ProfilerConfiguration.Load(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, (wrappedBuild) ? "..\\ProfilerConfig.jsonc" : "ProfilerConfig.jsonc"));
 
             _settings = new ProfilerSettings()
             {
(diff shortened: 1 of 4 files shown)


Xpl0itR:
I think you should also have the `Ryujinx.log` file in the root folder instead of the bin folder as that is also often used by users and its annoying having to sort through dlls to find it.

BaronKiko:
It does do that. In this pic there was just no log file yet as I hadn't run the build

Xpl0itR:
perfect 👍

ReaperOfSouls1909:
[ryujinx_FJW_icon.zip](https://github.com/Ryujinx/Ryujinx/files/3263654/ryujinx_FJW_icon.zip)
icon ready and it works for all icon sizes

BaronKiko:
Ah cool, ty
We still aren't sure if this pr will get merged but I can certainly use it :)

ReaperOfSouls1909:
glad to help

jduncanator:
Why do you need a CPP project for the wrapper? You can create processes natively in C# too, makes much more sense to keep everything in a single language.

BaronKiko:
The reason it's in cpp is because I wanted it to run as a single file without any extra dll's or other files in the folder and without requiring the install of any extra libs like .net which I don't think you can really do with C#. Please correct me if I'm wrong, C# is desirable as it would mean different platforms could share the same code.

PythonSmith:
We can make a cmd, powershell and bash script to run them, i think thats the best way to do it 👍

zephyren25:
> We can make a cmd, powershell and bash script to run them, i think thats the best way to do it +1

3 scripts seems a bit overkill (especially powershell), might as well use python

BaronKiko:
Closing because .net 3 single exe mode alleviates this issue

PR #711, opened 2019-06-27
PrntStub: Add a way to print arrays

Thog:
This PR adds support for printing arrays in stubs (useful for IPC InBuffer/InPointer).
This also add an util to parse an array of structure from a BinaryReader.

Thanks to @jduncanator for the DefaultLogFormatter.cs changes.

### DIFF
--- a/Ryujinx.Common/Logging/Formatters/DefaultLogFormatter.cs
+++ b/Ryujinx.Common/Logging/Formatters/DefaultLogFormatter.cs
@@ -33,7 +33,7 @@ namespace Ryujinx.Common.Logging
                         sb.Append(prop.Name);
                         sb.Append(": ");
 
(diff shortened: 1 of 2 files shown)


Ryujinx.Common/Logging/Formatters/DefaultLogFormatter.cs
```diff
@@ -31,7 +32,23 @@ public string Format(LogEventArgs args)
                     {
                         sb.Append(prop.Name);
                         sb.Append(": ");
-                        sb.Append(prop.GetValue(args.Data));
+
+                        if(typeof(Array).IsAssignableFrom(prop.PropertyType))
```
AcK77:
```suggestion
                        if (typeof(Array).IsAssignableFrom(prop.PropertyType))
```

PR #663, opened 2019-04-20
Implement IIrSensorServer GetNpadIrCameraHandle

jduncanator:
Based on RE here: https://gist.github.com/AcK77/20220f58c445bdfb3f97c9fd438de0ea

Resolves #618

PR #712, opened 2019-06-27
irs: Little service cleanup

AcK77:
Based on:
- #663 and #664 as references.
- Some RE from here: https://gist.github.com/AcK77/20220f58c445bdfb3f97c9fd438de0ea

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
@@ -81,7 +81,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid.Irs
         {
             NpadIdType npadIdType = (NpadIdType)context.RequestData.ReadUInt32();
 
-            if (npadIdType >= NpadIdType.Player8 && 
+            if (npadIdType >  NpadIdType.Player8 && 
                 npadIdType != NpadIdType.Unknown && 
                 npadIdType != NpadIdType.Handheld)
             {
--- a/Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
@@ -81,7 +81,9 @@ namespace Ryujinx.HLE.HOS.Services.Hid.Irs
         {
             NpadIdType npadIdType = (NpadIdType)context.RequestData.ReadUInt32();
 
-            if (npadIdType >= NpadIdType.Player8 && npadIdType != NpadIdType.Unknown && npadIdType != NpadIdType.Handheld)
+            if (npadIdType >= NpadIdType.Player8 && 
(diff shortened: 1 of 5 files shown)


Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
```diff
@@ -53,58 +63,39 @@ public long DeactivateIrsensor(ServiceCtx context)
...
         }
 
         // GetNpadIrCameraHandle(u32) -> nn::irsensor::IrCameraHandle
         public long GetNpadIrCameraHandle(ServiceCtx context)
         {
-            uint npadId = context.RequestData.ReadUInt32();
+            NpadIdType npadIdType = (NpadIdType)context.RequestData.ReadUInt32();
 
-            if (npadId >= 8 && npadId != 16 && npadId != 32)
+            if (npadIdType >= NpadIdType.Player8 && npadIdType != NpadIdType.Unknown && npadIdType != NpadIdType.Handheld)
```
gdkchan: imo this should be broken into multiple lines:
```cs
if (npadIdType >= NpadIdType.Player8 &&
    npadIdType != NpadIdType.Unknown &&
    npadIdType != NpadIdType.Handheld)
```

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
```diff
@@ -53,58 +63,41 @@ public long DeactivateIrsensor(ServiceCtx context)
...
         }
 
         // GetNpadIrCameraHandle(u32) -> nn::irsensor::IrCameraHandle
         public long GetNpadIrCameraHandle(ServiceCtx context)
         {
-            uint npadId = context.RequestData.ReadUInt32();
+            NpadIdType npadIdType = (NpadIdType)context.RequestData.ReadUInt32();
 
-            if (npadId >= 8 && npadId != 16 && npadId != 32)
+            if (npadIdType >= NpadIdType.Player8 &&
```
gdkchan: This check is different from the original (and I believe it is incorrent). Before it was checking if the `npadId` is >= 8. However, `NpadIdType.Player8` value is 7, so now its checking whenever it is >= 7.

gdkchan:
At some point, it would be nice to fix the inconsistent naming. For example, I see the `HidControllerId` being used in some places, and something called `NpadId` in other places. Are they the same thing? We should use the official names were possible.

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
AcK77: Nice catch, it was incorrect in the original code, it should check if it > 7!

gdkchan: Checking if the value is > 7 is the same as checking if its >= 8, so I believe it was correct.

PR #724, opened 2019-07-09
Reducing log sizes

BaronKiko:
I have separated logs into a single file per run. It should keep up to 3 runs before deleting them.
I have also disable stubs by default as they are the majority of the logs and they aren't very helpful.
I chose not to create a setting for max number for simplicity. I also chose not to limit the size of individual files as it would require checks during runtime and rewriting the whole file when it overflows. This shouldn't really be a concern with the other changes.

### DIFF
--- a/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
+++ b/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace Ryujinx.Common.Logging
@@ -20,6 +21,20 @@ namespace Ryujinx.Common.Logging
 
         public FileLogTarget(string path, string name, FileShare fileShare, FileMode fileMode)
         {
+            // Ensure directory is present
+            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(path, "Logs"));
+            logDir.Create();
+
+            // Clean up old logs, should only keep 3
+            FileInfo[] files = logDir.GetFiles("*.log").OrderBy((info => info.CreationTime)).ToArray();
+            for (int i = 0; i < files.Length - 2; i++)
+            {
+                files[i].Delete();
+            }
+
+            // Get path for the current time
+            path = Path.Combine(logDir.FullName, $"Ryujinx_{DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss")}.log");
+
             _name      = name;
             _logWriter = new StreamWriter(File.Open(path, fileMode, FileAccess.Write, fileShare));
             _formatter = new DefaultLogFormatter();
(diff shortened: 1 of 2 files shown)


Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -10,12 +12,26 @@ public class FileLogTarget : ILogTarget
...
-        public FileLogTarget(string path, FileShare fileShare, FileMode fileMode)
+        public FileLogTarget(string appDirectory, FileShare fileShare, FileMode fileMode)
         {
+            // Ensure directory is present
+            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(appDirectory, "Logs"));
+            logDir.Create();
+
+            // Clean up old logs, should only keep 3
+            FileInfo[] files = logDir.GetFiles("*.log").OrderBy((info => info.CreationTime)).ToArray();
+            for (int i = 0; i < files.Count() - 2; i++)
```
gdkchan: `files.Length`.

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -10,12 +12,26 @@ public class FileLogTarget : ILogTarget
...
+        public FileLogTarget(string appDirectory)
+            : this(appDirectory, FileShare.Read, FileMode.CreateNew)
         { }
 
-        public FileLogTarget(string path, FileShare fileShare, FileMode fileMode)
+        public FileLogTarget(string appDirectory, FileShare fileShare, FileMode fileMode)
         {
+            // Ensure directory is present
+            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(appDirectory, "Logs"));
+            logDir.Create();
```
gdkchan: Can't you just use `Directory.CreateDirectory`?

BaronKiko: I have to create the directory info class anyway so it's quicker to use it's create function.
Also it's safer as you only have to change 1 place if you change the path.

BaronKiko:
> Imo it doesn't solve the main problem, no file size limit.

The only way to solve that is to keep checking the size and cutting the log. It's possible but it would lead to a lot of runtime overhead and I think the real issue is we log too much.
My thinking is go with this so we solve the main issue of the ever expanding file and if logs remain to be too big, which I'm pretty sure they won't, add a size limit in.

What are people's thoughts on disabling stub logs by default? It seems to be the most controversial change

gdkchan:
This doesn't solve the ever expanding issue unless we only log truly exceptional cases. I can for example write a homebrew that prints a debug message in a log, and the log file will keep expanding forever as that currently has the warning log level. Also disabling stub logging is questionable as they may be part of the problem.
```
The only way to solve that is to keep checking the size and cutting the log. It's possible but it would lead to a lot of runtime overhead and I think the real issue is we log too much.
```
I don't think so. You can just keep a circular buffer with the log entries in memory. The buffer would have a fixed size, and in the case of a crash or emulator exit, you can just write the buffer contents to the log file. You could even log additional information like title id, memory map, threads running and so on.

PR #726, opened 2019-07-10
IPC services refactoring

AcK77:
Reflection concept taken from [SupercellNx](https://github.com/daeken/SupercellNx) with consent of @daeken

- Use custom Attributes to handle services.
- ServiceFactory is not needed anymore.
- Add a way to set the permissions and fix the bsd/friend services to use it.
- Little cleanup.
- C#7.1 is required.


Ryujinx.HLE/HOS/Services/Sm/IUserInterface.cs
```diff
@@ -30,6 +35,12 @@ public IUserInterface()
             };
 
             _registeredServices = new ConcurrentDictionary<string, KPort>();
+
+            _services = AppDomain.CurrentDomain.GetAssemblies()
```
gdkchan: Maybe this can be improved a bit... Maybe it's possible to get types only for the Ryujinx.HLE assembly. And maybe only for the `Ryujinx.HLE.HOS.Services` namespace.

Ryujinx.HLE/HOS/Services/Sm/IUserInterface.cs
```diff
@@ -75,7 +86,26 @@ public long GetService(ServiceCtx context)
             }
             else
             {
-                session.ClientSession.Service = ServiceFactory.MakeService(context.Device.System, name);
+                if (_services.TryGetValue(name, out Type type))
+                {
+                    ServiceAttribute serviceAttribute = (ServiceAttribute)type.GetCustomAttributes().First(service => ((ServiceAttribute)service).Name == name);
```
gdkchan: This code will break if any attribute other than the `ServiceAttribute` is used on those classes, due to the explicit cast to `ServiceAttribute`. I think you should use the overload of `GetCustomAttributes` that takes the attribute type like you did above.

Ryujinx.HLE/Ryujinx.HLE.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <RuntimeIdentifiers>win10-x64;osx-x64;linux-x64</RuntimeIdentifiers>
     <Configurations>Debug;Release;Profile Debug;Profile Release</Configurations>
+    <LangVersion>7.1</LangVersion>
```
gdkchan: What requires C# 7.1 on this code?

AcK77: This line here -> [IUserInterface.cs@L43](https://github.com/Ryujinx/Ryujinx/pull/726/files#diff-5031bc908f17d402a6b32d8f50a0f972R43) as explained here -> [https://stackoverflow.com/a/39386379](https://stackoverflow.com/a/39386379)

PR #728, opened 2019-07-10
Refactoring commands handling

AcK77:
- Use Attribute and Reflection to handle commands ID.
- Add all symbols (from SwIPC so not all time accurate).
- Re-sort some services commands methods.
- Some cleanup.
- Keep some empty constructor for consistency.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/IpcService.cs
+++ b/Ryujinx.HLE/HOS/Services/IpcService.cs
@@ -7,12 +7,14 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using Ryujinx.Profiler;
+using System.Reflection;
+using System.Linq;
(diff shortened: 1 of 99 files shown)


Ryujinx.HLE/HOS/Services/IpcService.cs
```diff
@@ -22,6 +24,13 @@ abstract class IpcService : IIpcService
 
         public IpcService()
         {
+            Commands = Assembly.GetExecutingAssembly().GetTypes()
+                .Where(type => type == GetType())
```
gdkchan: Why is this line needed?

gdkchan: Nevermind, got it now.

PR #731, opened 2019-07-14
Refactoring result codes

AcK77:
- Add a main enum who can handle some orphalin result codes and the default `ResultCode.Success` one.
- Add sub-enum by services when it's needed.
- Remove some empty lines.
- Recast all service calls to ResultCode.
- Remove some unneeded static declaration.
- Delete unused `NvHelper` class.
- Remove `MakeError` function and compute the error code in the enum itself.
- Remove `ErrorModule` enum because it's not needed anymore.


Ryujinx.HLE/HOS/Services/IpcService.cs
```diff
@@ -103,7 +103,7 @@ public void CallMethod(ServiceCtx context)
 
             if (ServiceConfiguration.IgnoreMissingServices || serviceExists)
             {
-                long result = 0;
+                ResultCode result = 0;
```
gdkchan: `= ResultCode.Success;`

PR #733, opened 2019-07-14
Fix private on some call

AcK77:
Since the reflection code didn't take care about `private`, this cause regression, so I have added the flag just in case and fix calls who are declared with `private` to `public`.


gdkchan:
I don't think the `NonPublic` flag should be added. We should just fix the existing methods. Otherwise someone may add a private method later without even noticing that it is incorrect.

PR #740, opened 2019-08-13
Remove old functions from jit cache

emmauss:
This allows removing old functions from the jit cache, and reusing their memory fro new functions.
The allocator has been reworked to use a linked list to keep track of current allocations.

### DIFF
--- a/ARMeilleure/Translation/JitCache.cs
+++ b/ARMeilleure/Translation/JitCache.cs
@@ -19,7 +19,7 @@ namespace ARMeilleure.Translation
 
         private static JitCacheMemoryAllocator _allocator;
 
-        private static List<JitCacheEntry> _cacheEntries;
+        private static Dictionary<int, JitCacheEntry> _cacheEntries;
 
         private static object _lock;
 
@@ -41,7 +41,7 @@ namespace ARMeilleure.Translation
 
             _allocator = new JitCacheMemoryAllocator(CacheSize, startOffset);
 
-            _cacheEntries = new List<JitCacheEntry>();
+            _cacheEntries = new Dictionary<int, JitCacheEntry>();
 
             _lock = new object();
         }
@@ -101,7 +101,7 @@ namespace ARMeilleure.Translation
             {
                 if (TryFind((int)offset, out JitCacheEntry entry))
                 {
-                    _cacheEntries.Remove(entry);
+                    _cacheEntries.Remove((int)offset, out entry);
 
                     int size = checked(entry.Size + (CodeAlignment - 1)) & ~(CodeAlignment - 1);
 
@@ -112,23 +112,16 @@ namespace ARMeilleure.Translation
 
         private static void Add(JitCacheEntry entry)
         {
-            _cacheEntries.Add(entry);
+            _cacheEntries.Add(entry.Offset, entry);
         }
 
         public static bool TryFind(int offset, out JitCacheEntry entry)
         {
             lock (_lock)
             {
-                foreach (JitCacheEntry cacheEntry in _cacheEntries)
+                if(_cacheEntries.TryGetValue(offset, out entry))
                 {
-                    int endOffset = cacheEntry.Offset + cacheEntry.Size;
-
-                    if (offset >= cacheEntry.Offset && offset < endOffset)
-                    {
-                        entry = cacheEntry;
-
-                        return true;
-                    }
+                    return true;
                 }
             }
 
--- a/ARMeilleure/Translation/JitCache.cs
+++ b/ARMeilleure/Translation/JitCache.cs
@@ -17,7 +17,7 @@ namespace ARMeilleure.Translation
 
         private static IntPtr _basePointer;
 
-        private static int _offset;
+        private static JitCacheMemoryAllocator _allocator;
 
         private static List<JitCacheEntry> _cacheEntries;
 
@@ -27,13 +27,19 @@ namespace ARMeilleure.Translation
         {
             _basePointer = MemoryManagement.Allocate(CacheSize);
 
+            int startOffset = 0;
+
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 JitUnwindWindows.InstallFunctionTableHandler(_basePointer, CacheSize);
 
                 // The first page is used for the table based SEH structs.
-                _offset = PageSize;
+                startOffset = PageSize;
             }
+
+            ReprotectRange(startOffset, CacheSize - startOffset);
+
+            _allocator = new JitCacheMemoryAllocator(CacheSize, startOffset);
 
             _cacheEntries = new List<JitCacheEntry>();
 
@@ -51,8 +57,6 @@ namespace ARMeilleure.Translation
                 IntPtr funcPtr = _basePointer + funcOffset;
 
                 Marshal.Copy(code, 0, funcPtr, code.Length);
-
-                ReprotectRange(funcOffset, code.Length);
 
                 Add(new JitCacheEntry(funcOffset, code.Length, func.UnwindInfo));
 
@@ -76,16 +80,7 @@ namespace ARMeilleure.Translation
             {
                 IntPtr funcPtr = _basePointer + pageStart;
 
-                MemoryManagement.Reprotect(funcPtr, (ulong)fullPagesSize, MemoryProtection.ReadAndExecute);
-            }
-
-            int remaining = endOffs - pageEnd;
-
-            if (remaining != 0)
-            {
-                IntPtr funcPtr = _basePointer + pageEnd;
-
-                MemoryManagement.Reprotect(funcPtr, (ulong)remaining, MemoryProtection.ReadWriteExecute);
+                MemoryManagement.Reprotect(funcPtr, (ulong)fullPagesSize, MemoryProtection.ReadWriteExecute);
             }
         }
 
@@ -93,16 +88,26 @@ namespace ARMeilleure.Translation
         {
             codeSize = checked(codeSize + (CodeAlignment - 1)) & ~(CodeAlignment - 1);
 
-            int allocOffset = _offset;
-
-            _offset += codeSize;
-
-            if ((ulong)(uint)_offset > CacheSize)
-            {
-                throw new OutOfMemoryException();
-            }
+            int allocOffset = _allocator.Allocate(codeSize);
 
             return allocOffset;
+        }
+
+        public static void Free(ulong address)
+        {
+            ulong offset = address - (ulong)_basePointer;
+
+            lock (_lock)
+            {
+                if (TryFind((int)offset, out JitCacheEntry entry))
+                {
+                    _cacheEntries.Remove(entry);
+
+                    int size = checked(entry.Size + (CodeAlignment - 1)) & ~(CodeAlignment - 1);
+
+                    _allocator.Free((int)entry.Offset, size);
+                }
+            }
         }
 
         private static void Add(JitCacheEntry entry)
--- a/ARMeilleure/Translation/JitCache.cs
+++ b/ARMeilleure/Translation/JitCache.cs
@@ -93,9 +93,9 @@ namespace ARMeilleure.Translation
             return allocOffset;
         }
 
-        public static void Free(ulong address)
+        public static void Free(IntPtr address)
         {
-            ulong offset = address - (ulong)_basePointer;
+            ulong offset = (ulong)address - (ulong)_basePointer;
 
             lock (_lock)
             {
@@ -119,7 +119,7 @@ namespace ARMeilleure.Translation
         {
             lock (_lock)
             {
-                if(_cacheEntries.TryGetValue(offset, out entry))
+                if (_cacheEntries.TryGetValue(offset, out entry))
                 {
                     return true;
                 }
--- a/ARMeilleure/Translation/JitCacheMemoryAllocator.cs
+++ b/ARMeilleure/Translation/JitCacheMemoryAllocator.cs
@@ -77,7 +77,7 @@ namespace ARMeilleure.Translation
         {
             if ((uint)offset >= (ulong)_size)
             {
-                throw new ArgumentOutOfRangeException();
+                throw new ArgumentOutOfRangeException("offset");
             }
 
             var node = _memoryRanges.First;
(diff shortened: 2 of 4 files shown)


ARMeilleure/Translation/Translator.cs
```diff
@@ -45,15 +54,43 @@ private void TranslateQueuedSubs()
...
             }
         }
 
+        private void CleanUpJit()
+        {
+            while(_threadCount !=0 )
+            {
+                if (_oldFunctions.TryDequeue(out TranslatedFunction function))
+                {
+                    if (function.EntryCount > 0)
```
gdkchan: Without a lock, this is not thread safe. The EntryCount may be 0 when you do this check, but by the time you call `_oldFunctions.Enqueue(function);` it may be already > 0.

ARMeilleure/Translation/JitCache.cs
```diff
@@ -17,7 +17,9 @@ static class JitCache
 
         private static IntPtr _basePointer;
 
-        private static int _offset;
+        //private static int _offset;
```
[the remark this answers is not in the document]
emmauss: completely forgot about this

emmauss:
addressed comments

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -7,7 +7,7 @@ class TranslatedFunction
     {
         public ulong Address => (ulong)Marshal.GetFunctionPointerForDelegate(_func);
 
-        public int EntryCount { get; set; }
+        public int entryCount;
```
gdkchan: Not following naming conventions. Imo:
```cs
private int _entryCount;

public int EntryCount => _entryCount;
```
(use _entryCount internally).

ARMeilleure/Translation/Translator.cs
```diff
@@ -54,39 +50,30 @@ private void TranslateQueuedSubs()
...
-            {
-                if (_oldFunctions.TryDequeue(out TranslatedFunction function))
+                else if (_oldFunctions.TryDequeue(out TranslatedFunction function))
                 {
-                    if (function.EntryCount > 0)
-                    {
-                        _oldFunctions.Enqueue(function);
-                    }
-                    else
+                    lock (function)
```
gdkchan: This lock alone is not going to solve the problem. One thread can still change the value after the check and before the `Free` call. However, there's a lock-less way to solve that problem. You could do `Interlocked.CompareExchange(ref function.EntryCount, -1, 0) != 0`. Then, at the `Execute` method, you could change `Interlocked.Increment(ref entryCount)` to:
```cs
if (Interlocked.Increment(ref EntryCount) == 0)
{
    return address;
}
```
Where address is the address of the guest function.

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -18,17 +18,20 @@ class TranslatedFunction
...
+            _func = func;
             _rejit = rejit;
         }
 
         public ulong Execute(State.ExecutionContext context)
         {
-            Interlocked.Increment(ref entryCount);
+            if (Interlocked.Increment(ref EntryCount) == 0)
+            {
+                return Address;
```
gdkchan: This is wrong, address if supposed to be the guest function address. Currently it is returning the address of the jited function in memory.

emmauss:
Currently, there is a chance of the translator hanging up after some time, esp. during gameplay. There are no craches or errors, just a freeze.. I'm yet to find out what's causing it.

emmauss:
rebased

ARMeilleure/Translation/JitCache.cs
```diff
@@ -76,54 +80,48 @@ private static void ReprotectRange(int offset, int size)
...
-            _cacheEntries.Add(entry);
+            _cacheEntries.Add(entry.Offset, entry);
         }
 
         public static bool TryFind(int offset, out JitCacheEntry entry)
         {
             lock (_lock)
             {
-                foreach (JitCacheEntry cacheEntry in _cacheEntries)
+                if(_cacheEntries.TryGetValue(offset, out entry))
```
gdkchan: `if (`.

ARMeilleure/Translation/JitCacheMemoryAllocator.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            return offset;
+        }
+
+        public void Free(int offset, int size)
+        {
+            if ((uint)offset >= (ulong)_size)
+            {
+                throw new ArgumentOutOfRangeException();
```
gdkchan: Pass the name of the argument that is out of range.

ARMeilleure/Translation/JitCacheMemoryAllocator.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            var node = _memoryRanges.First;
+
+            while (true)
+            {
+                if (node == null)
+                {
+                    throw new ArgumentOutOfRangeException();
```
gdkchan: Pass the name of the argument that is out of range.

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -1,25 +1,41 @@
...
+
+        public int EntryCount;
+
         private const int MinCallsForRejit = 100;
 
         private GuestFunction _func;
 
-        private bool _rejit;
-        private int  _callCount;
+        private ulong _address;
```
gdkchan: Unused?

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -1,25 +1,41 @@
+using System;
 using System.Threading;
+using System.Runtime.InteropServices;
 
 namespace ARMeilleure.Translation
 {
     class TranslatedFunction
     {
+        public ulong Pointer => (ulong)Marshal.GetFunctionPointerForDelegate(_func);
```
gdkchan: If possible it's better to use `IntPtr` for host pointers.

emmauss:
addressed comments

ARMeilleure/Translation/JitCacheMemoryAllocator.cs
```diff
@@ -77,7 +77,7 @@ public void Free(int offset, int size)
         {
             if ((uint)offset >= (ulong)_size)
             {
-                throw new ArgumentOutOfRangeException();
+                throw new ArgumentOutOfRangeException("offset");
```
gdkchan: `nameof(offset)`

ARMeilleure/Translation/JitCacheMemoryAllocator.cs
```diff
@@ -86,7 +86,7 @@ public void Free(int offset, int size)
             {
                 if (node == null)
                 {
-                    throw new ArgumentOutOfRangeException();
+                    throw new ArgumentOutOfRangeException("offset");
```
gdkchan: `nameof(offset)`

ARMeilleure/Translation/JitCacheMemoryAllocator.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                    break;
+                }
+
+                node = node.Next;
+            }
+
+            node.Value = (node.Value.Start, offset - 1);
+        }
+
```
gdkchan: New line not necessary.

emmauss:
addressed comments

emmauss:
Rebased and tested.

riperiperi:
image

Pretty much required for A Hat In Time and Breath of the Wild (both crash shortly after getting ingame on master without increasing jit cache size, though the former has its own problems). I'll review this later today.

ameisen:
When I merged this locally, I ended up with crashes due to addresses being allocated that were invalid. I had to remove the call to free memory in the Allocator; I suspect the Allocator may be buggy but don't have time to investigate. I'd prefer seeing a proper memory allocator there, anyways.

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -1,25 +1,42 @@
...
-            _func  = func;
+            _func = func;
             _rejit = rejit;
+            _address = address;
         }
 
         public ulong Execute(State.ExecutionContext context)
         {
-            return _func(context.NativeContextPtr);
+            if (Interlocked.Increment(ref EntryCount) == 0)
```
ameisen, line 30: This causes a race condition if there is more than one thread pending while it is is being exchanged elsewhere. It can end up being reset back to a value greater than 0.

Change the `== 0` to `>= 0` should resolve it in conjunction with the other change.

ARMeilleure/Translation/Translator.cs
```diff
@@ -45,10 +50,31 @@ private void TranslateQueuedSubs()
...
+                        return func;
+                    });
                 }
                 else
                 {
+                    Queue<TranslatedFunction> skippedFunctions = new Queue<TranslatedFunction>();
+
+                    while (_oldFunctions.TryDequeue(out TranslatedFunction function))
+                    {
+                        if (Interlocked.CompareExchange(ref function.EntryCount, -1, 0) != 0)
```
ameisen, line 66: Change this to replace the value with `(int.MinValue + 1)` or so instead of `-1`. This lets the increments from multiple threads that are pending *never* potentially bring the value above 0 again, preventing a race condition which can cause the function to execute while it's potentially being released.

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -1,25 +1,42 @@
...
+            {
+                return _address;
+            }
+
+            var nextAddress = _func(context.NativeContextPtr);
```
ameisen, line 35: Might make sense to wrap this in a `try-finally`, just to make sure it decrements if there's an exception? I suspect an exception would be fatal, but never hurts to be safe rather than sorry.

ARMeilleure/Translation/TranslatedFunction.cs
gdkchan: I guess you meant `<= 0`?

ARMeilleure/Translation/JitCache.cs
```diff
@@ -93,16 +99,23 @@ private static int Allocate(int codeSize)
...
-            _offset += codeSize;
+            return allocOffset;
+        }
 
-            if ((ulong)(uint)_offset > CacheSize)
+        public static void Free(ulong address)
+        {
+            ulong offset = address - (ulong)_basePointer;
+
+            if (TryFind((int)offset, out JitCacheEntry entry))
```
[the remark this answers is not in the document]
ameisen: This PR turns it into an amortized constant-time operation. You can actually remove the memory-management parts of this PR and it will still do it. If you create an issue, I can make a PR of my local build that does exactly that.

ARMeilleure/Translation/TranslatedFunction.cs
ameisen: Yes.

ARMeilleure/Translation/JitCache.cs
```diff
@@ -76,54 +80,48 @@ private static void ReprotectRange(int offset, int size)
...
+        public static void Free(IntPtr address)
+        {
+            ulong offset = (ulong)address - (ulong)_basePointer;
 
-            if ((ulong)(uint)_offset > CacheSize)
+            lock (_lock)
             {
-                throw new OutOfMemoryException();
-            }
+                if (TryFind((int)offset, out JitCacheEntry entry))
```
ameisen, line 102: `.Remove` already returns both the entry (`out entry`) and if it was present.

This can be simplified to `if (_cacheEntries.Remove((int)offset, out var entry)) ...`, which eliminates a needless query and a use of `TryFind`.

ARMeilleure/Translation/JitCache.cs
```diff
@@ -17,25 +17,31 @@ static class JitCache
...
             {
                 JitUnwindWindows.InstallFunctionTableHandler(_basePointer, CacheSize);
 
                 // The first page is used for the table based SEH structs.
-                _offset = PageSize;
+                startOffset = PageSize;
             }
 
-            _cacheEntries = new List<JitCacheEntry>();
+            ReprotectRange(startOffset, CacheSize - startOffset);
```
ameisen, line 40: I'm just marking this as this use of `ReprotectRange` is one of the reasons it cannot be merged presently. My local branch resolves this.

ARMeilleure/Translation/Translator.cs
```diff
@@ -20,6 +21,8 @@ public class Translator
 
         private ConcurrentDictionary<ulong, TranslatedFunction> _funcs;
 
+        private Queue<TranslatedFunction> _oldFunctions;
```
ameisen, line 24: There can be multiple translation threads at a time, and the atomicity of the factory callback in `AddOrUpdate` is not guaranteed (as per MSDN). This should be a `ConcurrentQueue`.

ARMeilleure/Translation/Translator.cs
ameisen: Adding to review.

ARMeilleure/Translation/Translator.cs
```diff
@@ -45,10 +50,31 @@ private void TranslateQueuedSubs()
...
+                        {
+                            skippedFunctions.Enqueue(function);
+                        }
+                        else
+                        {
+                            JitCache.Free(function.Pointer);
+                        }
+                    }
+
+                    _oldFunctions = skippedFunctions;
```
ameisen, line 76: Not thread-safe.

My local build just has `skippedFunctions` as a `List<TranslatedFunction>`, and then just iterates over it at the end to re-enqueue everything.

ARMeilleure/Translation/JitCache.cs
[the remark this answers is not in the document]
gdkchan: https://github.com/Ryujinx/Ryujinx/issues/1129

ARMeilleure/Translation/Translator.cs
ameisen: Also, as a code-style thing, preferentially change this to `== 0`, and reverse the order of Free/Enqueue, so the positive branch is first.

ARMeilleure/Translation/TranslatedFunction.cs
ameisen: Also, due to a significant number of changes in `master`, this is no longer sufficient to determine if a function is amidst execution, since the address could have been taken by another function and called indirectly, or via tail calls, or so forth. I have not found a reliable way to resolve this.

emmauss:
closing as it needs a major rework.

PR #741, opened 2019-08-18
Discord Presence: Fix a crash when no valid program is loaded

Thog:


Ryujinx/Program.cs
```diff
@@ -108,15 +108,29 @@ static void Main(string[] args)
...
+            if (device.System.State.DiscordIntegrationEnabled)
             {
                 if (File.ReadAllLines(Path.Combine(ApplicationDirectory, "RPsupported.dat")).Contains(device.System.TitleID))
                 {
                     DiscordPresence.Assets.LargeImageKey = device.System.TitleID;
                 }
 
-                DiscordPresence.Details               = $"Playing {device.System.TitleName}";
-                DiscordPresence.State                 = device.System.TitleID.ToUpper();
+                string state = device.System.TitleID;
```
Xpl0itR: I think you should make the Title ID uppercase for aesthetic reasons

Thog: Oh yeah totally missed this thanks!

Ryujinx/Program.cs
```diff
@@ -108,15 +108,29 @@ static void Main(string[] args)
...
+            if (device.System.State.DiscordIntegrationEnabled)
             {
                 if (File.ReadAllLines(Path.Combine(ApplicationDirectory, "RPsupported.dat")).Contains(device.System.TitleID))
                 {
                     DiscordPresence.Assets.LargeImageKey = device.System.TitleID;
                 }
 
-                DiscordPresence.Details               = $"Playing {device.System.TitleName}";
-                DiscordPresence.State                 = device.System.TitleID.ToUpper();
+                string state = device.System.TitleID.ToUpper();
```
Xpl0itR: wouldn't that crash if title id is null since u cant make a null uppercase? change it to `string state = string.IsNullOrWhiteSpace(device.System.TitleID) ? "Ryujinx" : device.System.TitleID.ToUpper();`

PR #742, opened 2019-08-18
Fix a memory corruption in SoundIO wrapper

Thog:
This fix audio slowdown on Unix based platforms where soundio will try
to switch to mono because of the invalid data written.


jduncanator:
This file is auto-generated and will be overwritten when the wrapper is updated. It's best to fix this in the auto-generation code, rather than monkey patching like this. I will investigate on my end and get back to you.

jduncanator:
:+1: Thanks!

PR #768, opened 2019-09-08
bcat:u: Implement EnumerateDeliveryCacheDirectory

### LINKED ISSUE #622: bcat:u IDeliveryCacheStorageService: 10 (EnumerateDeliveryCacheDirectory) not implemented
IDeliveryCacheStorageService EnumerateDeliveryCacheDirectory is not implemented.

## Service description:
[Switchbrew](https://switchbrew.org/wiki/BCAT_services#IDeliveryCacheStorageService) | [Reswitched](https://reswitched.github.io/SwIPC/ifaces.html#nn::bcat::detail::ipc::IDeliveryCacheStorageService(10))

```EnumerateDeliveryCacheDirectory() -> (i32, array<nn::bcat::DirectoryName, 6>)```

Unknown.

## Required by:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/136

### PR DESCRIPTION
**Pull request by AcK77**
Basic implementation `EnumerateDeliveryCacheDirectory` call to `IDeliveryCacheStorageService` according to RE.  (close #622)
I've added some comments in the whole service for when we'll implement a real bcat implementation.
For now, all games who use it isn't playable because of GPU.

### CODE REVIEW

File: Ryujinx.HLE/HOS/Services/Bcat/IDeliveryCacheStorageService.cs [hunk 1]
```diff
@@ -1,7 +1,48 @@
...
+
 namespace Ryujinx.HLE.HOS.Services.Bcat
 {
     class IDeliveryCacheStorageService : IpcService
     {
-        public IDeliveryCacheStorageService() { }
+        private const int DeliveryCacheDirectoriesLimit    = 100;
+        private const int DeliveryCacheDirectoryNameLength = 32;
+
+        private List<string> _deliveryCacheDirectories = new List<string>();
```
jduncanator: Any specific reason this is a List instead of just being an array?

AcK77: Just because I prefere use `_deliveryCacheDirectories.Add("something")` instead of allocate the array and then use the index.

jduncanator: Sure

AcK77:
Yeah there is no goal for now to implement bcat, since every game who use this functionnality don't go ingame because of the GPU. So for now, it's better to leave some `TODO` based on RE and implement the call too... Even if it return no data.

Ryujinx.HLE/HOS/Services/Bcat/IDeliveryCacheStorageService.cs
gdkchan: Assuming that the number of elements is not going to change after you load it, I think its better to use an array since it is more efficient.

AcK77: I've changed it by an array :)

gdkchan:
Hmm, with the list being always empty the code doesn't seems to actually do anything other than writing 0 on the response. I guess it would be the case to print a stub log, unless its fine to return 0 entries.

AcK77:
In the case of your switch don't contain bcat content, the save data don't exist and then it's return an empty array. But I can add a stub log if you want.

gdkchan:
Ok, then it should be fine.

Ryujinx.HLE/HOS/Services/Bcat/IServiceCreator.cs
```diff
@@ -12,9 +17,24 @@ class IServiceCreator : IpcService
...
         public ResultCode CreateBcatService(ServiceCtx context)
         {
-            long id = context.RequestData.ReadInt64();
+            // TODO: Call arp:r GetApplicationLaunchProperty with the pid to get the TitleId.
+            //       Add an instance of nn::bcat::detail::service::core::PassphraseManager.
+            //       Add an instance of nn::bcat::detail::service::ServiceMemoryManager.
+            //       Add an instance of nn::bcat::detail::service::core::TaskManager who load "bcat-sys:/" system save data and open "dc/task.bin". 
+            //       If the file don't exist, create a new one (size of 0x800) and write 2 empty struct with a size of 0x400.
+
+            ApplicationLaunchProperty applicationLaunchProperty = new ApplicationLaunchProperty
```
Thog: As we have multiple instance of this in the codebase, it would be nice to do an helper for arp:r GetApplicationLaunchProperty

Ryujinx.HLE/HOS/Services/Bcat/IServiceCreator.cs
```diff
@@ -23,9 +43,24 @@ public ResultCode CreateBcatService(ServiceCtx context)
...
         public ResultCode CreateDeliveryCacheStorageService(ServiceCtx context)
         {
-            long id = context.RequestData.ReadInt64();
+            // TODO: Call arp:r GetApplicationLaunchProperty with the pid to get the TitleId.
+            //       Add an instance of nn::bcat::detail::service::core::ApplicationStorageManager who load "bcat-dc-X:/" system save data,
+            //       return ResultCode.NullSaveData if failed.
+            //       Where X depend of the ApplicationLaunchProperty stored in an array (range 0-3).
+            //       Add an instance of nn::bcat::detail::service::ServiceMemoryManager.
+
+            ApplicationLaunchProperty applicationLaunchProperty = new ApplicationLaunchProperty
```
Thog: same comment

Ryujinx.HLE/HOS/Services/Bcat/IServiceCreator.cs
AcK77: Done.

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchPropertyHelper.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Ryujinx.HLE.FileSystem;
+using Ryujinx.HLE.Utilities;
+using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.HLE.HOS.Services.Arp
+{
+    static class ApplicationLaunchPropertyHelper
+    {
+        public static ApplicationLaunchProperty GetDefault()
```
jduncanator: I'd prefer these functions to be on the `ApplicationLaunchProperty` type itself, if you're trying to get the default or one by PID, that's the first place I'd look for these functions. It also prevents having heaps of Helper classes floating around.

AcK77: done.

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchPropertyHelper.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Ryujinx.HLE.FileSystem;
+using Ryujinx.HLE.Utilities;
+using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.HLE.HOS.Services.Arp
+{
+    static class ApplicationLaunchPropertyHelper
+    {
+        public static ApplicationLaunchProperty GetByPid(ServiceCtx context)
```
jduncanator: This should also be moved into `ApplicationLaunchProperty`, if we need to move it around later, we can when that PR comes in. No point leaving this one function out on it's own when we don't need it to be.

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchProperty.cs
```diff
@@ -7,5 +9,16 @@ class ApplicationLaunchProperty
         public byte  BaseGameStorageId;
         public byte  UpdateGameStorageId;
         public short Padding;
+
+        public static ApplicationLaunchProperty GetDefault()
```
jduncanator: This can just be a static property:

```
public static ApplicationLaunchProperty Default
{
    get
    {
        return new ApplicationLaunchProperty {
            TitleId             = 0x00,
            Version             = 0x00,
            BaseGameStorageId   = (byte)StorageId.NandSystem,
            UpdateGameStorageId = (byte)StorageId.None
        };
    }
}
```

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchPropertyHelper.cs
AcK77: `ApplicationLaunchProperty` are stored in an array with the pid and the `control.nacp` file (https://switchbrew.org/wiki/Glue_services#arp:r) each time a process is loaded. Since we don't really support multiprocess for now, we have to hardcode the values. But when it will be done, the helper is already here and we'll just have to handle everything inside. 

Put the method who should return one `ApplicationLaunchProperty` from an array of a `ApplicationLaunchProperty`, in the `ApplicationLaunchProperty` itself, make no sense for me.

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchProperty.cs
AcK77: done

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchPropertyHelper.cs
gdkchan: I agree that we don't need a helper for now.

gdkchan:
Would be nice to match the names of the real services on the namespaces (since ARP is part of the Glue service, i'd prefer to see this on `Ryujinx.HLE.HOS.Services.Glue`). I know it's not a new issue and it affects other services too, just noticed it on those ARP changes and I think we should fix that at some point.

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchPropertyHelper.cs
AcK77: done

PR #772, opened 2019-09-19
Misc fixes

LDj3SNuD:
**Fixes:**
- A null reference exception on exit if Discord integration is not enabled
- Not having a logger reference time set to 0 if the execution times of the launcher and the title do not match
- The incorrect build up of debug strings for some services, like `GetCurrentProcessorNumber6`

Nits.


LDj3SNuD:
Ready.

PR #781, opened 2019-09-21
audout:u: Implement SetAudioOutVolume and GetAudioOutVolume

### LINKED ISSUE #777: audout:u IAudioOut: 12 (SetAudioOutVolume) not implemented
audout:u IAudioOut: 12 (SetAudioOutVolume) is not implemented.

## Service description:
[Switchbrew](https://switchbrew.org/wiki/Audio_services#IAudioOut)
```
[6.0.0+] SetAudioOutVolume
```

## Required by:
* https://github.com/Ryujinx/Ryujinx-Games-List/issues/200

AcK77:
- Implementation of `audout:u` SetAudioOutVolume and GetAudioOutVolume (checked with RE). 
- Add Get and Set for Volume into audio backends.
- Cleanup of all audio backends to follow the `IAalOutput` structure and .NET standard.
- Split OpenAL backend into 2 files for consistency.

close #777.

### DIFF
--- a/Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
+++ b/Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
@@ -2,7 +2,6 @@ using OpenTK.Audio;
 using OpenTK.Audio.OpenAL;
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Threading;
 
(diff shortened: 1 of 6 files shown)


Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
```diff
@@ -13,174 +12,38 @@ namespace Ryujinx.Audio
...
-        public OpenALAudioOut()
-        {
-            Context = new AudioContext();
+        /// <summary>
+        /// The <see cref="OpenTK.Audio"/> audio context
+        /// </summary>
+        private AudioContext _context;
 
-            Tracks = new ConcurrentDictionary<int, Track>();
+        /// <summary>
```
gdkchan: We don't need to use the xml comments on private fields, but you could use a regular comment here.

Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
```diff
@@ -197,157 +60,223 @@ public static bool IsSupported
...
+        /// </summary>
+        public float GetVolume() => _volume;
+
+        /// <summary>
+        /// Set playback volume
+        /// </summary>
+        /// <param name="volume">The volume of the playback</param>
+        public void SetVolume(float volume)
+        {
+            _volume = Math.Clamp(_volume - (_volume * volume), 0.0f, 1.0f);
```
gdkchan: Not sure why you subtract the volume here. A brief explanation would be nice.

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
```diff
@@ -135,61 +141,75 @@ public void Stop(int trackId)
...
         /// <param name="bufferTag">The internal tag of the buffer</param>
         /// <param name="buffer">The buffer to append to the track</param>
-        public void AppendBuffer<T>(int trackId, long bufferTag, T[] buffer)
-            where T : struct
+        public void AppendBuffer<T>(int trackId, long bufferTag, T[] buffer) where T : struct
         {
-            if (m_TrackPool.TryGet(trackId, out SoundIoAudioTrack track))
+            if (_trackPool.TryGet(trackId, out SoundIoAudioTrack track))
             {
+                track.AudioStream.SetVolume(_volume);
```
jduncanator: Why do you do this every time a buffer is appended?

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
```diff
@@ -77,12 +69,12 @@ public PlaybackState GetState(int trackId)
...
         public int OpenTrack(int sampleRate, int channels, ReleaseCallback callback)
         {
-            if (!m_TrackPool.TryGet(out SoundIoAudioTrack track))
+            if (!_trackPool.TryGet(out SoundIoAudioTrack track))
             {
                 return -1;
             }
 
-            // Open the output. We currently only support 16-bit signed LE
+            // NOTE: Open the output. We currently only support 16-bit signed LE
```
jduncanator: Why have you prepended `NOTE` before the start of each comment? Seems redundant to me...

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
```diff
@@ -233,10 +253,12 @@ private static bool IsSupportedInternal()
 
             try
             {
-                context = new SoundIO();
-
-                context.OnBackendDisconnect = (i) => {
-                    backendDisconnected = true;
+                context = new SoundIO
```
jduncanator: Why was this reformatted? Now seems out of style with the rest of the project?

Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
jduncanator: Why not? It is displayed in intellisense in any IDE that parses XML documentation. Beats having to go to the code (which may be compiled in another binary) and reading the comment.

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
Thog: Might be a better idea to do something like that in my pov
```suggestion
            // Open the output.
            // FIXME: We currently only support 16-bit signed LE.
```

Ryujinx.Audio/Renderers/OpenAL/OpenALAudioOut.cs
gdkchan: Well, its a private field, so at most it would be necessary to scroll up on the file to read the comment, since its not going to be accessed externally. I have never seem them being used for private fields, so I assumed its not common pratice. Personally I just add them on public members.

gdkchan: On a second thought, I guess theres no reason to *not* have those as xml doc.

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
jduncanator: It's not really a fixme, the entire audio rendering pipeline (service impl and all) only supports 16-bit signed LE at the moment, but I'm not fazed. Just seems pointless to me to add "NOTE" to the start of every comment in the entire codebase.

Thog: Then we should open an issue add a comment on the class documentation

jduncanator: I'm happy with that! :+1:

Ryujinx.Audio/Renderers/SoundIo/SoundIoAudioOut.cs
```diff
@@ -135,61 +146,89 @@ public void Stop(int trackId)
...
+        public void SetVolume(float gain)
         {
-            if (m_TrackPool.TryGet(trackId, out SoundIoAudioTrack track))
+            if (!_volumeChanged)
             {
-                List<long> bufferTags = new List<long>();
+                // Games send a gain value here, so we need to apply it on the current volume value.
+                // In that way we have to multiply the gain by the volume to get the real gain value.
+                // And then we substract the real gain value to our current volume value.
+                _volume = Math.Clamp(_volume - (_volume * gain), 0.0f, 1.0f);
```
jduncanator: This logic seems inverted to me. At the moment, if you pass in a "gain" of `+0.1`, the output will be turned down, and if you pass in a "gain" of `-0.1` the output will be turned up. Seems backwards to me? Where did this come from, the sysmod?

PR #794, opened 2019-10-16
Move InvalidSystemResourceException with the other Exceptions

AcK77:
`InvalidSystemResourceException.cs` was in `Ryujinx.HLE/Resource` who didn't make sense since it's the only file in the folder, and other exceptions class are in `Ryujinx.HLE/Exceptions`.


jduncanator:
Just an FYI, technically that exception isn't properly implemented. It needs the `InvalidSystemResourceException(SerializationInfo info, StreamingContext context) ` constructor overload.

PR #799, opened 2019-10-27
Fix ghost users bug.

LDj3SNuD:
The output buffer for sequential writing of the 8 potential `User Id` (for a total of 128 bytes), for the commands `ListAllUsers` and `ListOpenUsers`, must always be cleaned for a smaller number of `User Id` (generally 1); in order to avoid that the potential random values contained in it may be misinterpreted as true `User Id` by the `GetUserExistence` commands and avoid the Sdk failure of EnsureSaveData.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Account/Acc/IAccountServiceForApplication.cs
+++ b/Ryujinx.HLE/HOS/Services/Account/Acc/IAccountServiceForApplication.cs
@@ -63,19 +63,25 @@ namespace Ryujinx.HLE.HOS.Services.Account.Acc
             long outputPosition = context.Request.RecvListBuff[0].Position;
             long outputSize     = context.Request.RecvListBuff[0].Size;
 
-            ulong offset = 0;
+            ulong offset = 0UL;
 
             foreach (UserProfile userProfile in profiles)
             {
-                if (offset + 0x10 > (ulong)outputSize)
+                if (offset + 0x10UL > (ulong)outputSize)
                 {
                     break;
                 }
 
-                context.Memory.WriteInt64(outputPosition + (long)offset,     userProfile.UserId.Low);
-                context.Memory.WriteInt64(outputPosition + (long)offset + 8, userProfile.UserId.High);
+                context.Memory.WriteInt64(outputPosition + (long)offset,      userProfile.UserId.Low);
+                context.Memory.WriteInt64(outputPosition + (long)offset + 8L, userProfile.UserId.High);
 
                 offset += 0x10;
+            }
+
+            for ( ; offset + 0x10UL <= (ulong)outputSize; offset += 0x10UL)
+            {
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/Account/Acc/IAccountServiceForApplication.cs
```diff
@@ -63,19 +64,21 @@ private ResultCode WriteUserList(ServiceCtx context, IEnumerable<UserProfile> pr
...
 
-            ulong offset = 0;
+            MemoryHelper.FillWithZeros(context.Memory, outputPosition, (int)outputSize);
+
+            ulong offset = 0UL;
 
             foreach (UserProfile userProfile in profiles)
             {
-                if (offset + 0x10 > (ulong)outputSize)
+                if (offset + 0x10UL > (ulong)outputSize)
```
jduncanator: You don’t need the `UL` type suffix here, as `offset` is already an unsigned long.

Ryujinx.HLE/HOS/Services/Account/Acc/IAccountServiceForApplication.cs
```diff
@@ -63,19 +64,21 @@ private ResultCode WriteUserList(ServiceCtx context, IEnumerable<UserProfile> pr
...
                     break;
                 }
 
-                context.Memory.WriteInt64(outputPosition + (long)offset,     userProfile.UserId.Low);
-                context.Memory.WriteInt64(outputPosition + (long)offset + 8, userProfile.UserId.High);
+                context.Memory.WriteInt64(outputPosition + (long)offset,      userProfile.UserId.Low);
+                context.Memory.WriteInt64(outputPosition + (long)offset + 8L, userProfile.UserId.High);
 
-                offset += 0x10;
+                offset += 0x10UL;
```
jduncanator: Same as above.

Ryujinx.HLE/HOS/Services/Account/Acc/IAccountServiceForApplication.cs
```diff
@@ -72,8 +75,8 @@ private ResultCode WriteUserList(ServiceCtx context, IEnumerable<UserProfile> pr
                     break;
                 }
 
-                context.Memory.WriteInt64(outputPosition + (long)offset,     userProfile.UserId.Low);
-                context.Memory.WriteInt64(outputPosition + (long)offset + 8, userProfile.UserId.High);
+                context.Memory.WriteInt64(outputPosition + (long)offset,      userProfile.UserId.Low);
+                context.Memory.WriteInt64(outputPosition + (long)offset + 8L, userProfile.UserId.High);
```
jduncanator: Same here, don't need the `L` type suffix because `offset` is already cast to `long`.

PR #757, opened 2019-09-03
Add features to GUI

Xpl0itR:
### TODO list for this PR
- [x] Change controller image depending on type of controller
- [x] Game list should scan subdirs for games
- [x] Keys file exists checker
- [x] Only shows Program NCAs in Application list
- [x] Error popup for certain errors
- [x] Change shown GUI columns without restarting
- [x] Sort by column if you click on the column header
  - [x] Fix sort on `Time Played`, `Last Played` and `File Size` columns
- [x] Add ability to designate favourite games
- [x] Load metadata from JSONs
- [x] Implement async loading of the game list
- [x] Progress bar and information about what is being loaded

PR #801, opened 2019-10-30
Rename "RyuFs" directory to "Ryujinx" and use the same savedata system the Switch uses

Thealexbarney:
Stores savedata the same way Horizon does. Uses LibHac for savedata FS commands.

Renames the Ryujinx AppData directory from `RyuFs` to `Ryujinx`

Renames the `nand` and `sdmc` directories in the Ryujinx AppData directory to `bis` and `sdcard` respectively.

A one-time migration can be performed to migrate an existing `RyuFs` directory to the new `Ryujinx` directory. The `nand` and `sdmc` directories will be renamed, and existing saves will be imported into the new system.

The old `RyuFs` directory will be left untouched after migration.

Uses LibHac for FS commands:

- EnsureSaveData
- OpenBisFileSystem
- OpenSdCardFileSystem
- DeleteSaveDataFileSystem
- CreateSaveDataFileSystem
- CreateSaveDataFileSystemBySystemSaveDataId
- DeleteSaveDataFileSystemBySaveDataSpaceId
- DeleteSaveDataFileSystemBySaveDataAttribute
- CreateSaveDataFileSystemWithHashSalt
- OpenSaveDataFileSystem
- OpenSaveDataFileSystemBySystemSaveDataId
- OpenReadOnlySaveDataFileSystem
- OpenSaveDataInfoReader
- OpenSaveDataInfoReaderBySaveDataSpaceId
- FindSaveDataWithFilter
- OpenSaveDataInfoReaderWithFilter
- ReadSaveDataInfo


Thealexbarney:
Rebased on latest master

Thealexbarney:
Blocked on #757 because this PR relies on UI files being moved out of the nand directory

Ryujinx/Ui/MainWindow.cs
```diff
@@ -71,16 +70,32 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
             DeleteEvent += Window_Close;
 
             ApplicationLibrary.ApplicationAdded += Application_Added;
-
+
```
AcK77: nit: Why?

Ryujinx/Ui/Migration.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            Device = device;
+        }
+
+        public static bool PromptIfMigrationNeededForStartup(Window parentWindow, out bool isMigrationNeeded)
+        {
+            if (!IsMigrationNeeded())
+            {
+                isMigrationNeeded = false;
+                return true;
```
AcK77: add a new empty line before

Ryujinx/Ui/Migration.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            string oldBasePath = Path.Combine(appDataPath, "RyuFs");
+            string newBasePath = Path.Combine(appDataPath, "Ryujinx");
+
+            string oldSaveDir = Path.Combine(oldBasePath, "nand/user/save");
+
+            CopyRyuFs(oldBasePath, newBasePath);
+
+            var importer = new SaveImporter(oldSaveDir, Device.System.FsClient);
+            return importer.Import();
```
AcK77: add a new empty line before

Ryujinx/Ui/MainWindow.cs
Thealexbarney: It wasn't caught because auto-formatting wasn't run on the code

Ryujinx/Ui/AboutWindow.cs
```diff
@@ -44,7 +44,7 @@ private AboutWindow(Builder builder) : base(builder.GetObject("_aboutWin").Handl
             {
                 IJsonFormatterResolver resolver = CompositeResolver.Create(new[] { StandardResolver.AllowPrivateSnakeCase });
 
-                using (Stream stream = File.OpenRead(System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "RyuFS", "Installer", "Config", "Config.json")))
+                using (Stream stream = File.OpenRead(System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Ryujinx", "Installer", "Config", "Config.json")))
```
Xpl0itR: I don't think changing this one is a good idea because this is the path to a JSON created by the ryujinx installer which is currently deprecated and as such most likely won't be updated to use the new path

emmauss: if the installer is deprecated, then it's not supported, and should not be taken into consideration in future updates, including this one. if we still want to keep recommending that, then it should be updated together with any changes in the main app which concerns it.

PR #803, opened 2019-10-31
Increase IPC buffer size

jduncanator:
This is a hack, but it works for now. We should really determine a way to automatically calculate the required buffer size to avoid situations where specific IPC calls "overflow" this arbitrary size, or determine by means of reverse engineering if there really is a fixed "maximum size" that we could implement here.

### DIFF
--- a/Ryujinx.HLE/HOS/Ipc/IpcHandler.cs
+++ b/Ryujinx.HLE/HOS/Ipc/IpcHandler.cs
@@ -67,7 +67,7 @@ namespace Ryujinx.HLE.HOS.Ipc
 
                         case 3:
                         {
-                            request = FillResponse(response, 0, 0x500);
+                            request = FillResponse(response, 0, 0x1000);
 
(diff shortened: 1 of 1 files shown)


jduncanator:
Upon further investigation, it seems these are hardcoded in most public implementations of the IPC. For example, see these references in Stratosphere: https://github.com/Atmosphere-NX/Atmosphere/search?utf8=%E2%9C%93&q=PointerBufferSize

This may require some reverse engineering to determine if there is a calculable size, or if each service literally has a hardcoded value in the official release.

AcK77:
Each system module got the PointerBufferSize hardcoded (and the size don't depend by services) according to the RE.
There are some value I've found by RE (who match Atmosphère values):
```
Loader   -> 0x400
Spl      -> 0x800
Hid      -> 0x200
Settings -> 0x100
Ldn      -> 0x1000
NvDrv    -> 0x8000
```

PR #804, opened 2019-10-31
Implement a rudimentary applets system

jduncanator:
This adds basic support for applets. I imagine this system will be expanded to support more applets as time goes, and it's shape and form will change. For now, it's enough to support the player select applet, which has been implemented with this PR.

We now throw a `NotImplementedException` when creating an applet instance that we don't implement, rather than blindly stubbing it. This will probably negatively impact the boot status of some games, however over time we should be able to implement the vast majority of the applets.


Ryujinx.HLE/HOS/Applets/AppletManager.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public static IApplet Create(AppletId applet, Horizon system)
+        {
+            if(_appletMapping.TryGetValue(applet, out Type appletClass))
+            {
+                return (IApplet)Activator.CreateInstance(appletClass, system);
+            }
+
+            return new StubApplet();
```
gdkchan: I guess a exception for unimplemented applets would be better.

jduncanator: This will cause a large number of games that currently boot "fine" with a stub, to crash. (We currently have it stubbed this way). I'm fine either way.

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
...
+            _system      = system;
+            _inputStack  = new Queue<IStorage>();
+            _outputStack = new Queue<IStorage>();
+        }
+
+        public ResultCode Start()
+        {
+            _outputStack.Enqueue(new IStorage(BuildResponse()));
+
+            AppletStateChanged?.Invoke(this, null);
```
jduncanator: Is it valid to signal the event before `Start()` has returned? We may need to defer this until after `Start()` returns.

Ryujinx.HLE/HOS/Applets/AppletId.cs
```diff
@@ -0,0 +1,27 @@
...
+        SoftwareKeyboard = 0x11,
+        MiiEdit = 0x12,
+        LibAppletWeb = 0x13,
+        LibAppletShop = 0x14,
+        PhotoViewer = 0x15,
+        Settings = 0x16,
+        LibAppletOff = 0x17,
+        LibAppletWhitelisted = 0x18,
+        LibAppletAuth = 0x19,
+        MyPage = 0x1A
```
Thog: Align everything here

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+namespace Ryujinx.HLE.HOS.Applets
+{
+    internal class PlayerSelect : IApplet
+    {
+        private Horizon _system;
+        private Queue<IStorage> _inputStack;
+        private Queue<IStorage> _outputStack;
```
Thog: Align everything here

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/LibraryAppletCreator/ILibraryAppletAccessor.cs
```diff
@@ -65,9 +69,14 @@ public ResultCode PushInData(ServiceCtx context)
         // PopOutData() -> object<nn::am::service::IStorage>
         public ResultCode PopOutData(ServiceCtx context)
         {
-            MakeObject(context, new IStorage(StorageHelper.MakeLaunchParams()));
+            var result = _applet.PopOutData(out IStorage storage);
```
Thog:
```suggestion
            ResultCode result = _applet.PopOutData(out IStorage storage);
```

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/LibraryAppletCreator/ILibraryAppletAccessor.cs
```diff
@@ -65,9 +69,14 @@ public ResultCode PushInData(ServiceCtx context)
...
 
-            return ResultCode.Success;
+            if(storage != null)
```
Thog: Also check error code here

LDj3SNuD:
```suggestion
            if (storage != null)
```

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
...
+            return ResultCode.Success;
+        }
+
+        private byte[] BuildResponse()
+        {
+            var currentUser = _system.State.Account.LastOpenedUser;
+
+            using (var ms = new MemoryStream())
+            {
+                var writer = new BinaryWriter(ms);
```
LDj3SNuD: Too many `var`.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/LibraryAppletCreator/ILibraryAppletAccessor.cs
gdkchan: `if(` -> `if (`.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/LibraryAppletCreator/ILibraryAppletAccessor.cs
jduncanator: `var` is used here because the ResultCode returned by the IApplet is from a different namespace and it's a complete mess to put the full namespace in before the type here.

Thog: Then directly cast the result

Ryujinx.HLE/HOS/Applets/AppletId.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Ryujinx.HLE.HOS.Applets
+{
+    enum AppletId
```
AcK77: You should add the corresponding TitleId for reference (https://switchbrew.org/wiki/Applet_Manager_services#AppletId) as comment.

```
        OverlayDisplay       = 0x02, // 0x010000000000100C
```

and this file should be in `Am.AppletAE.Types` to avoid duplicate code later.

Ryujinx.HLE/HOS/Applets/AppletManager.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.HLE.HOS.Applets
+{
+    static class AppletManager
+    {
+        private static Dictionary<AppletId, Type> _appletMapping;
```
AcK77:
```suggestion
        private static Dictionary<AppletId, Type> _appletList;
```

Maybe `List` is more revelant than `Mapping`.

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            return ResultCode.Success;
+        }
+
+        private byte[] BuildResponse()
+        {
+            UserProfile currentUser = _system.State.Account.LastOpenedUser;
+
+            using (MemoryStream ms = new MemoryStream())
```
AcK77:
```suggestion
            using (MemoryStream stream = new MemoryStream())
```
as we name all `Stream` like that, it's better to keep this name for consistency.

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                BinaryWriter writer = new BinaryWriter(ms);
+
+                writer.Write((ulong)PlayerSelectResult.Success);
```
AcK77: I think the result of `Player Selector Applet` should be handle in a struct or a class.

As example:
```
struct PlayerSelectorResult 
{
   public ulong   Result;
   public UInt128 UserId;
}
```
Then you can write the struct using helper and the code who write the result is more readable.
Moreover if we will use the same implementation (as you use an interface) for other applets, we can keep the same files structure for consistency.

Ryujinx.HLE/HOS/Applets/StubApplet.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Ryujinx.Common.Logging;
+using Ryujinx.HLE.HOS.Services.Am.AppletAE;
+using System;
+
+namespace Ryujinx.HLE.HOS.Applets
+{
+    internal class StubApplet : IApplet
```
AcK77: It's not needed since the `StubApplet` class isn't use anywhere.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
```diff
@@ -10,7 +11,10 @@ class ILibraryAppletCreator : IpcService
         // CreateLibraryApplet(u32, u32) -> object<nn::am::service::ILibraryAppletAccessor>
         public ResultCode CreateLibraryApplet(ServiceCtx context)
         {
-            MakeObject(context, new ILibraryAppletAccessor(context.Device.System));
+            int appletId          = context.RequestData.ReadInt32();
```
AcK77: You should cast `appletId` as `AppletId` enum.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
```diff
@@ -10,7 +11,10 @@ class ILibraryAppletCreator : IpcService
...
+            int libraryAppletMode = context.RequestData.ReadInt32();
```
AcK77: You should case `libraryAppletMode` with an enum (https://switchbrew.org/wiki/Applet_Manager_services#LibraryAppletMode) who is located in the same `Types` folder near `AppletId` file (as I said above).

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
```diff
@@ -10,7 +11,10 @@ class ILibraryAppletCreator : IpcService
...
+
+            MakeObject(context, new ILibraryAppletAccessor((AppletId)appletId, context.Device.System));
```
AcK77: Why you don't pass both vars?

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Ryujinx.HLE.HOS.Services.Account.Acc;
+using Ryujinx.HLE.HOS.Services.Am.AppletAE;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Ryujinx.HLE.HOS.Applets
+{
+    internal class PlayerSelect : IApplet
```
AcK77: There is a bad design choice here, this is basically how current applets works:
- Call AppletAE.
- Call IAllSystemAppletProxiesService: OpenLibraryAppletProxy (Cmd 200/201)
- Call ILibraryAppletProxy: OpenLibraryAppletSelfAccessor (Cmd 20)
- Call ILibraryAppletSelfAccessor: PopInData (Cmd 0) to get the IStorages sent by the game (You use `PushInData` to enqueue them).
- Call ILibraryAppletSelfAccessor: PushOutData (Cmd 1) to sent the result to the game (You use `PopOutData` to dequeue them).

The best way to do that is to make a basic implementation of those calls, then use static functions for PopInData and PushOutData and call them from your Applet implementation. In that way you manage IStorages directly in the system module as HOS do, and the `Applets` folder just contain things related to applets and nothing more.

More informations (from a quick RE):

PopInData return 

```
0   -> Success
2   -> Not Available Storage 
500 -> Invalid Storage Size
```

PushOutData return 

```
0   -> Success 
500 -> Invalid Storage Size
```

Instead of this issue, everything looks good.

Ryujinx.HLE/HOS/Applets/AppletManager.cs
jduncanator: Well, it maps AppletId to Applet implementation?

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
jduncanator: 👍

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
jduncanator: I'm likely to change all this when swkbd comes along soon, so I'd prefer not to have to swap it all out twice. I'm happy to do it if you feel strongly enough about it, but if not, I'll make these changes once I refactor it all for swkbd.

Ryujinx.HLE/HOS/Applets/StubApplet.cs
jduncanator: 👍

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
jduncanator: We don't do this anywhere else in the code (do we?)

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
jduncanator: This is all going to change with swkbd.

Ryujinx.HLE/HOS/Applets/AppletManager.cs
gdkchan: I think mapping makes more sense, since that is not exactly a list.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
jduncanator: The Player Select applet does not make use of libraryAppletMode. When I add swkbd, I'll likely need to pass it then.

Ryujinx.HLE/HOS/Applets/AppletManager.cs
AcK77: Then it's fine.

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
AcK77: It's better to start with something good instead of refactoring it "later", since we don't know when "later" is ...

Ryujinx.HLE/HOS/Applets/PlayerSelect/PlayerSelect.cs
AcK77: As I said... We don't know when `swkbd` will be implemented.

Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ILibraryAppletCreator.cs
AcK77: We do this at many places in HOS -> [Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs#L54](
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs#L545)

PR #809, opened 2019-11-03
Improve IRoInterface logic

Thog:
This commit contains a little rewrite of IRoInterface to fix some issues
that we were facing on some recent games (AC3 Remastered & Final Fantasy
VIII Remastered)

Also move IRoInterface to the RO module.

Related issues:

- https://github.com/Ryujinx/Ryujinx-Games-List/issues/196

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
@@ -203,7 +203,7 @@ namespace Ryujinx.HLE.HOS.Services.Ro
 
             while (retryCount++ < MaxMapRetries)
             {
-                ResultCode result = MapCodeMemoryInProcess(process, out nroMappedAddress, info.NroAddress, info.NroSize);
+                ResultCode result = MapCodeMemoryInProcess(process, info.NroAddress, info.NroSize, out nroMappedAddress);
 
                 if (result != ResultCode.Success)
                 {
@@ -257,7 +257,7 @@ namespace Ryujinx.HLE.HOS.Services.Ro
             return false;
         }
 
-        private ResultCode MapCodeMemoryInProcess(KProcess process, out ulong targetAddress, ulong baseAddress, ulong size)
+        private ResultCode MapCodeMemoryInProcess(KProcess process, ulong baseAddress, ulong size, out ulong targetAddress)
         {
             KMemoryManager memMgr = process.MemoryManager;
 
@@ -265,11 +265,13 @@ namespace Ryujinx.HLE.HOS.Services.Ro
 
             int retryCount;
 
+            int addressSpacePageLimit = (int)((memMgr.GetAddrSpaceSize() - size) >> 12);
+
             for (retryCount = 0; retryCount < MaxMapRetries; retryCount++)
             {
                 while (true)
                 {
-                    targetAddress = memMgr.GetAddrSpaceBaseAddr() + (ulong)(_random.Next((int)((memMgr.GetAddrSpaceSize() - size) >> 12)) << 12);
+                    targetAddress = memMgr.GetAddrSpaceBaseAddr() + (ulong)(_random.Next(addressSpacePageLimit) << 12);
 
                     if (memMgr.InsideAddrSpace(targetAddress, size) && !memMgr.InsideHeapRegion(targetAddress, size) && !memMgr.InsideAliasRegion(targetAddress, size))
                     {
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
```diff
@@ -187,77 +193,118 @@ public ResultCode ParseNro(out NroInfo res, ServiceCtx context, ulong nroAddress
...
                 {
-                    if (!memMgr.InsideHeapRegion (targetAddress, info.TotalSize) &&
-                        !memMgr.InsideAliasRegion(targetAddress, info.TotalSize))
+                    KernelResult bssMappingResult = memMgr.MapProcessCodeMemory(nroMappedAddress + info.NroSize, info.BssAddress, info.BssSize);
+
+                    if (bssMappingResult == KernelResult.InvalidMemState)
                     {
-                        break;
+                        memMgr.UnmapProcessCodeMemory(nroMappedAddress + info.NroSize, info.BssAddress, info.BssSize);
+                        memMgr.UnmapProcessCodeMemory(nroMappedAddress, info.NroAddress, info.NroSize);
```
AcK77:
```suggestion
                        memMgr.UnmapProcessCodeMemory(nroMappedAddress + info.NroSize, info.BssAddress, info.BssSize);
                        memMgr.UnmapProcessCodeMemory(nroMappedAddress,                info.NroAddress, info.NroSize);
```
Maybe you could align args here to be more readable

Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
```diff
@@ -187,77 +193,118 @@ public ResultCode ParseNro(out NroInfo res, ServiceCtx context, ulong nroAddress
...
+
+                        continue;
+                    }
+                    else if (bssMappingResult != KernelResult.Success)
+                    {
+                        memMgr.UnmapProcessCodeMemory(nroMappedAddress + info.NroSize, info.BssAddress, info.BssSize);
+                        memMgr.UnmapProcessCodeMemory(nroMappedAddress, info.NroAddress, info.NroSize);
```
AcK77: Same as above

Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
jduncanator: Imo, this makes it less readable.

AcK77: As I said, it's just a suggestion...

Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
```diff
@@ -187,77 +193,118 @@ public ResultCode ParseNro(out NroInfo res, ServiceCtx context, ulong nroAddress
...
+            targetAddress = 0;
+
+            int retryCount;
+
+            for (retryCount = 0; retryCount < MaxMapRetries; retryCount++)
             {
-                memMgr.UnmapProcessCodeMemory(targetAddress, info.NroAddress, info.NroSize);
+                while (true)
+                {
+                    targetAddress = memMgr.GetAddrSpaceBaseAddr() + (ulong)(_random.Next((int)((memMgr.GetAddrSpaceSize() - size) >> 12)) << 12);
```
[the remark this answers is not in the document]
gdkchan: I think thats what the real service does.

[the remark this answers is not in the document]
jduncanator: I get that, but is ASLR something we really want in an emulator? Especially given the application of assisting developers to debug their homebrew? At least an option to disable it would be nice.

gdkchan: Could you split that up? I personally think its quite confusing with it all in one line.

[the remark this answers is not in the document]
Thog: Personally I have two reason of using ASLR compared to before.

The first one is that it's more accurate to have ASLR here because ro doesn't have any way to disable ASLR for NRO on the system.
The other reason is that it's also way faster to do that instead of doing a dumb incrementation.

If we need to debug NRO code at runtime we should probably improve our diagnostics capabilities (By improving naming of NRO in stacktraces, adding some way to track them by build id, and by printing the whole process address space state when crashing)

Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
Thog: yeah I will not apply this one

PR #831, opened 2019-11-22
Rewrite the configuration system

### LINKED ISSUE #812: Redesign the configuration system
The current configuration system is pretty limiting (no possibility of reloading of any subsystems, no file watch, ...) and we aren't versioning the configuration.

As such, we need a new reliable configuration system that fit our needs.

Thog:
The configuration system was quite fragile and too dependent on everything, this fix #812 .

The changes:

- The file configuration is now entirely independent from the internal configuration state.
- The file configuration is versioned (current version is 1).
- Every configuration elements are now reactive properties that the emulator can register on to handle initialization and configuration changes.
- The configuration system is now in Ryujinx.Common to be accessible on every projects.
- Discord integration is now independent from the UI and can be reloaded.
- The primary controller is now configurable at runtime (NOTE: the UI currently doesn't have any options to configure real controller).
- The logger is entirely reloadable.
- You can now hotplug your controller when the emulator is running.
- The logger now takes name for every LogTarget to make them removable at runtime.
- The logger now always add the default "console" target to avoid loosing early init logs.
- The configuration system now generates a default file configuration if it's missing or too new.
- General system stability improvements to enhance the user's experience


Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,444 @@
...
+                },
+                Hotkeys     = new KeyboardHotkeys
+                {
+                    ToggleVsync = Key.Tab
+                }
+            };
+
+            Hid.JoystickControls.Value = new NpadController
+            {
+                Enabled           = true,
```
gdkchan: nit: Remove extra space.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,444 @@
...
+                    Stick       = ControllerInputId.Axis0,
+                    StickButton = ControllerInputId.Button8,
+                    DPadUp      = ControllerInputId.Hat0Up,
+                    DPadDown    = ControllerInputId.Hat0Down,
+                    DPadLeft    = ControllerInputId.Hat0Left,
+                    DPadRight   = ControllerInputId.Hat0Right,
+                    ButtonMinus = ControllerInputId.Button6,
+                    ButtonL     = ControllerInputId.Button4,
+                    ButtonZl    = ControllerInputId.Axis2,
+
```
gdkchan: nit: Remove empty line.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,444 @@
...
+                    ButtonZr    = ControllerInputId.Axis5,
+                }
+            };
+        }
+
+        public void Load(ConfigurationFileFormat configurationFileFormat)
+        {
+            if (configurationFileFormat.Version != 1 && configurationFileFormat.Version != 0)
+            {
+                Common.Logging.Logger.PrintWarning(LogClass.Application, $"Unsupported configuration version {configurationFileFormat.Version}, loading default!");
```
gdkchan: No need for the exclamation mark at the end of the message, i'd use a dot.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,444 @@
...
+            Hid.EnableKeyboard.Value               = configurationFileFormat.EnableKeyboard;
+            Hid.KeyboardControls.Value             = configurationFileFormat.KeyboardControls;
+            Hid.JoystickControls.Value             = configurationFileFormat.JoystickControls;
+        }
+
+        public static void Initialize()
+        {
+            if (Instance != null)
+            {
+                throw new InvalidOperationException("Configuration is already initialized!");
```
gdkchan: Same as above.

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -37,20 +37,54 @@ static Logger()
             m_LogTargets = new List<ILogTarget>();
 
             m_Time = Stopwatch.StartNew();
+
+            // logger should log to console by default
+            AddTarget(new AsyncLogTargetWrapper(
+                new ConsoleLogTarget("console"),
+                1000,
+                AsyncLogTargetOverflowAction.Block
```
gdkchan: ` AsyncLogTargetOverflowAction.Block));`

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -32,5 +34,10 @@ public void Dispose()
             _logWriter.Flush();
             _logWriter.Dispose();
         }
+
+        public string GetName()
```
gdkchan: Maybe it's better to just make `_name` public? (And change the name to Name). Could also use a property instead of the readonly field.

Ryujinx.Common/Logging/Targets/ILogTarget.cs
```diff
@@ -5,5 +5,7 @@ namespace Ryujinx.Common.Logging
     public interface ILogTarget : IDisposable
     {
         void Log(object sender, LogEventArgs args);
+
+        string GetName();
```
gdkchan: See above. If you use a property, that would change to `string Name { get; }`.

Ryujinx.Common/Logging/Targets/JsonLogTarget.cs
```diff
@@ -7,10 +7,12 @@ public class JsonLogTarget : ILogTarget
     {
         private Stream _stream;
         private bool   _leaveOpen;
+        private string _name;
```
gdkchan: Same comment as above.

Ryujinx.Common/Ryujinx.Common.csproj
```diff
@@ -27,6 +27,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="JsonPrettyPrinter" Version="1.0.1.1" />
```
gdkchan: Would be nice to drop both dependencies and just use the new .NET Json serializer.

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (titleName != null)
+            {
+                details = $"Playing {titleName}";
+            }
+
+            DiscordPresence.Details               = details;
+            DiscordPresence.State                 = state;
+            DiscordPresence.Assets.LargeImageText = titleName;
+            DiscordPresence.Assets.SmallImageKey  = "ryujinx";
+            DiscordPresence.Assets.SmallImageText = "Ryujinx is an emulator for the Nintendo Switch";
```
gdkchan: I think the correct would be "Ryujinx is a Nintendo Switch emulator".

Ryujinx/Program.cs
```diff
@@ -16,8 +17,30 @@ static void Main(string[] args)
...
+            // Initialize the configuration
+            ConfigurationState.Initialize();
+
+            // Initialize the logger system
+            LoggerModule.Initialize();
+
+            // Initialize Discord integration
+            DiscordIntegrationModule.Initialize();
+
+            string configurationPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json");
```
gdkchan: At some point, I believe all the paths should be handled by VFs (or whatever it will turn into in the future). That will avoid logic to assemble the final paths being scattered across the different projects.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -536,65 +487,70 @@ private void About_Pressed(object o, EventArgs args)
...
-            SwitchSettings.SwitchConfig.GuiColumns[8] = _pathToggle.Active;
+            ConfigurationState.Instance.Ui.GuiColumns.Value[8] = _pathToggle.Active;
+
+            SaveConfig();
+        }
 
-            Configuration.SaveConfig(SwitchSettings.SwitchConfig, System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
+        public static void SaveConfig()
+        {
+            ConfigurationState.Instance.ToFileFormat().SaveConfig(System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
```
gdkchan: Just what I was saying above, again that same code is repeated to get the final config path. Not being able to use `using System.IO;` is also not very nice.

Ryujinx/Ui/SwitchSettings.cs
```diff
@@ -120,59 +116,59 @@ private SwitchSettings(Builder builder, HLE.Switch device) : base(builder.GetObj
...
-            _x1.Label            = SwitchConfig.KeyboardControls.RightJoycon.ButtonX.ToString();
-            _y1.Label            = SwitchConfig.KeyboardControls.RightJoycon.ButtonY.ToString();
-            _plus1.Label         = SwitchConfig.KeyboardControls.RightJoycon.ButtonPlus.ToString();
-            _r1.Label            = SwitchConfig.KeyboardControls.RightJoycon.ButtonR.ToString();
-            _zR1.Label           = SwitchConfig.KeyboardControls.RightJoycon.ButtonZr.ToString();
-
-            _custThemePath.Buffer.Text           = SwitchConfig.CustomThemePath;
-            _graphicsShadersDumpPath.Buffer.Text = SwitchConfig.GraphicsShadersDumpPath;
-            _fsLogSpinAdjustment.Value           = SwitchConfig.FsGlobalAccessLogMode;
+            if (ConfigurationState.Instance.Logger.EnableFileLog)             { _fileLogToggle.Click();        }
```
gdkchan: Would be nice to reformat all those ifs like so:
```cs
if (condition)
{
    // Code...
}
```

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -0,0 +1,90 @@
...
+                State      = "Idling",
+                Timestamps = new Timestamps(DateTime.UtcNow)
+            };
+
+            ConfigurationState.Instance.EnableDiscordIntegration.Event += Update;
+        }
+
+        private static void Update(object sender, ReactiveEventArgs<bool> e)
+        {
+            if (e.OldValue != e.Value)
```
jduncanator: Why do you do this check? The Update handler will only be invoked if the previous value is different from the new value.

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public class ReactiveEventArgs<T>
+    {
+        public T OldValue { get; private set; }
+        public T Value    { get; private set; }
+
+        public ReactiveEventArgs(T oldValue, T value)
+        {
+            OldValue = oldValue;
+            Value    = value;
```
jduncanator: If we're calling the previous value `OldValue`, I think `Value` should be called `NewValue`.

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            ConfigurationState.Instance.Logger.EnableDebug.Event       += ReloadEnableDebug;
+            ConfigurationState.Instance.Logger.EnableStub.Event        += ReloadEnableStub;
+            ConfigurationState.Instance.Logger.EnableInfo.Event        += ReloadEnableInfo;
+            ConfigurationState.Instance.Logger.EnableWarn.Event        += ReloadEnableWarning;
+            ConfigurationState.Instance.Logger.EnableError.Event       += ReloadEnableError;
+            ConfigurationState.Instance.Logger.EnableGuest.Event       += ReloadEnableGuest;
+            ConfigurationState.Instance.Logger.EnableFsAccessLog.Event += ReloadEnableFsAccessLog;
+            ConfigurationState.Instance.Logger.FilteredClasses.Event   += ReloadFilteredClasses;
+            ConfigurationState.Instance.Logger.EnableFileLog.Event     += ReloadFileLogger;
```
jduncanator: I'd implement a single handler for all of these. There is very little overhead in changing the enabled/disable state of a logger, so updating the state for all of them at once, even if only one changes, can make all this neater.

Ryujinx.Common/Ryujinx.Common.csproj
jduncanator: I think this should be done in a separate PR.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -71,7 +65,9 @@ public class MainWindow : Window
         [GUI] CheckMenuItem _fileExtToggle;
         [GUI] CheckMenuItem _fileSizeToggle;
         [GUI] CheckMenuItem _pathToggle;
+        [GUI] Box           _box;
```
Xpl0itR: Remove

Ryujinx/Ui/MainWindow.cs
```diff
@@ -71,7 +65,9 @@ public class MainWindow : Window
...
         [GUI] TreeView      _gameTable;
+        [GUI] GLArea        _glScreen;
```
Xpl0itR: Remove

Ryujinx/Ui/MainWindow.cs
```diff
@@ -91,60 +87,39 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
...
-            if (SwitchSettings.SwitchConfig.GuiColumns.IconColumn)       { _iconToggle.Active       = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.AppColumn)        { _appToggle.Active        = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.DevColumn)        { _developerToggle.Active  = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.VersionColumn)    { _versionToggle.Active    = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.TimePlayedColumn) { _timePlayedToggle.Active = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.LastPlayedColumn) { _lastPlayedToggle.Active = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileExtColumn)    { _fileExtToggle.Active    = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileSizeColumn)   { _fileSizeToggle.Active   = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.PathColumn)       { _pathToggle.Active       = true; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.FavColumn)        { _favToggle.Active       = true; }
```
Xpl0itR: align =

Ryujinx/Ui/MainWindow.cs
```diff
@@ -211,15 +186,24 @@ private void UpdateColumns()
...
-            if (SwitchSettings.SwitchConfig.GuiColumns.DevColumn)        { _versionColumn.SortColumnId    = 4; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.TimePlayedColumn) { _timePlayedColumn.SortColumnId = 5; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.LastPlayedColumn) { _lastPlayedColumn.SortColumnId = 6; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileExtColumn)    { _fileExtColumn.SortColumnId    = 7; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileSizeColumn)   { _fileSizeColumn.SortColumnId   = 8; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.PathColumn)       { _pathColumn.SortColumnId       = 9; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.FavColumn)        { _favColumn.SortColumnId        = 0; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.IconColumn)       { _appColumn.SortColumnId        = 2; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.AppColumn)        { _devColumn.SortColumnId        = 3; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.DevColumn)        { _versionColumn.SortColumnId    = 4; }
```
Xpl0itR: I used the wrong conditions in my PR on these lines, i'd appreciate if you fix this in your PR

Ryujinx/Ui/MainWindow.cs
```diff
@@ -527,7 +481,7 @@ private void FavToggle_Toggled(object sender, ToggledArgs args)
         private void Row_Activated(object sender, RowActivatedArgs args)
         {
             _tableStore.GetIter(out TreeIter treeIter, new TreePath(args.Path.ToString()));
-            string path = (string)_tableStore.GetValue(treeIter, 9);
+            string path = (string)_tableStore.GetValue(treeIter, 8);
```
Xpl0itR: `8` needs to be `9` or you are telling ryujinx to load an app from the file size rather than the path

Ryujinx/Ui/MainWindow.cs
```diff
@@ -607,13 +561,13 @@ private void FullScreen_Toggled(object sender, EventArgs args)
...
-            SwitchSettings settingsWin = new SwitchSettings(_device);
+            SwitchSettings settingsWin = new SwitchSettings();
             settingsWin.Show();
         }
 
-        private void Update_Pressed(object sender, EventArgs args)
+        private void Update_Pressed(object o, EventArgs args)
         {
-            string ryuUpdater = System.IO.Path.Combine(new VirtualFileSystem().GetBasePath(), "RyuUpdater.exe");
+            string ryuUpdater = System.IO.Path.Combine(new VirtualFileSystem().GetBasePath(), "RyuFS", "RyuUpdater.exe");
```
Xpl0itR: this is setting the path to: %appdata%/RyuFs/RyuFs/RyuUpdater.exe

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        private static DiscordRpcClient DiscordClient;
+
+        public static RichPresence DiscordPresence { get; private set; }
+
+        public static void Initialize()
+        {
+            DiscordPresence = new RichPresence
+            {
+                Assets     = new Assets
```
Xpl0itR: does this need the random spaces?

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    DiscordClient?.Dispose();
+
+                    DiscordClient = null;
+                }
+
+                // If we need to activate it and the client isn't active, initialize it
+                if (e.Value && DiscordClient == null)
+                {
+                    DiscordClient   = new DiscordRpcClient("568815339807309834");
```
Xpl0itR: remove random spaces

Ryujinx/Configuration/DiscordIntegrationModule.cs
Thog: It's aligned to the next fields

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                T oldValue = _value;
+                
+                _value = value;
+
+                _readerWriterLock.ReleaseWriterLock();
+
+                if (oldValue == null || !oldValue.Equals(_value))
+                {
+                    Event?.Invoke(this, new ReactiveEventArgs<T>(oldValue, value));
```
jduncanator: For clarity and avoidance of issues in the future, you should probably pass `_value` into the `ReactiveEventArgs` constructor, instead of `value`.

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                _readerWriterLock.AcquireReaderLock(int.MaxValue);
+                T value = _value;
+                _readerWriterLock.ReleaseReaderLock();
+
+                return value;
+            }
+            set
+            {
+                _readerWriterLock.AcquireWriterLock(int.MaxValue);
```
jduncanator: This entire `set` block should probably be wrapped in a Reader lock, and you should `UpgradeToWriterLock` when updating the value, and then `DowngradeFromWriterLock` back again once you've set the value.

Thog:
Ready for review again

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,503 @@
...
+                    FileExtColumn    = new ReactiveObject<bool>();
+                    FileSizeColumn   = new ReactiveObject<bool>();
+                    PathColumn       = new ReactiveObject<bool>();
+                }
+            }
+
+            /// <summary>
+            /// Used to toggle columns in the GUI
+            /// </summary>
+            // TODO: UPDATE THIS TO MASTER
```
AcK77: Leftover? Or it needs to be updated once it’s merged in master?

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,503 @@
...
+                    ButtonZr    = ControllerInputId.Axis5,
+                }
+            };
+        }
+
+        public void Load(ConfigurationFileFormat configurationFileFormat)
+        {
+            if (configurationFileFormat.Version != 1 && configurationFileFormat.Version != 0)
+            {
+                Common.Logging.Logger.PrintWarning(LogClass.Application, $"Unsupported configuration version {configurationFileFormat.Version}, loading default.");
```
AcK77:
```suggestion
                Logger.PrintWarning(LogClass.Application, $"Unsupported configuration version {configurationFileFormat.Version}, loading default.");
```
Should be fine too?

Ryujinx.Common/Configuration/Hid/Key.cs
```diff
@@ -0,0 +1,153 @@
+﻿namespace Ryujinx.Configuration.Hid
+{
+    public enum Key
```
AcK77: I know it will boring, but alignment here could be fine

Ryujinx.HLE/Input/Hid.cs
```diff
@@ -47,18 +49,37 @@ public Hid(Switch device, long hidPosition)
             _keyboardOffset     = HidPosition + HidKeyboardOffset;
         }
 
-        public void InitializePrimaryController(ControllerStatus controllerType)
+        private static ControllerStatus ConvertControllerTypeToState(ControllerType controllerType)
         {
-            ControllerId controllerId = controllerType == ControllerStatus.Handheld ?
+            switch (controllerType)
+            {
+                case ControllerType.Handheld:
```
AcK77: What about the case and the return on the same line, aligned with others?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -91,60 +85,39 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
...
-            if (SwitchSettings.SwitchConfig.GuiColumns.IconColumn)       { _iconToggle.Active       = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.AppColumn)        { _appToggle.Active        = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.DevColumn)        { _developerToggle.Active  = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.VersionColumn)    { _versionToggle.Active    = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.TimePlayedColumn) { _timePlayedToggle.Active = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.LastPlayedColumn) { _lastPlayedToggle.Active = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileExtColumn)    { _fileExtToggle.Active    = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileSizeColumn)   { _fileSizeToggle.Active   = true; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.PathColumn)       { _pathToggle.Active       = true; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.FavColumn)        { _favToggle.Active        = true; }
```
AcK77: Maybe brackets should be removed here?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -211,15 +184,24 @@ private void UpdateColumns()
...
-            if (SwitchSettings.SwitchConfig.GuiColumns.FavColumn)        { _favColumn.SortColumnId        = 0; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.IconColumn)       { _appColumn.SortColumnId        = 2; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.AppColumn)        { _devColumn.SortColumnId        = 3; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.DevColumn)        { _versionColumn.SortColumnId    = 4; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.TimePlayedColumn) { _timePlayedColumn.SortColumnId = 5; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.LastPlayedColumn) { _lastPlayedColumn.SortColumnId = 6; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileExtColumn)    { _fileExtColumn.SortColumnId    = 7; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileSizeColumn)   { _fileSizeColumn.SortColumnId   = 8; }
-            if (SwitchSettings.SwitchConfig.GuiColumns.PathColumn)       { _pathColumn.SortColumnId       = 9; }
+            if (ConfigurationState.Instance.Ui.GuiColumns.FavColumn)        { _favColumn.SortColumnId        = 0; }
```
AcK77: Maybe brackets should be removed here?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -607,11 +559,11 @@ private void FullScreen_Toggled(object sender, EventArgs args)
 
         private void Settings_Pressed(object sender, EventArgs args)
         {
-            SwitchSettings settingsWin = new SwitchSettings(_device);
+            SwitchSettings settingsWin = new SwitchSettings();
             settingsWin.Show();
         }
 
-        private void Update_Pressed(object sender, EventArgs args)
+        private void Update_Pressed(object o, EventArgs args)
```
AcK77: Keep sender isn’t better for consistency?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -633,121 +585,81 @@ private void About_Pressed(object sender, EventArgs args)
...
-
-            Configuration.SaveConfig(SwitchSettings.SwitchConfig, System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
+            ConfigurationState.Instance.Ui.GuiColumns.FileExtColumn.Value = _fileExtToggle.Active;
 
+            SaveConfig();
             UpdateColumns();
         }
 
-        private void FileSize_Toggled(object sender, EventArgs args)
+        private void FileSize_Toggled(object o, EventArgs args)
```
AcK77: Keep sender isn’t better for consistency?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -633,121 +585,81 @@ private void About_Pressed(object sender, EventArgs args)
...
-
-            Configuration.SaveConfig(SwitchSettings.SwitchConfig, System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
+            ConfigurationState.Instance.Ui.GuiColumns.FileSizeColumn.Value = _fileSizeToggle.Active;
 
+            SaveConfig();
             UpdateColumns();
         }
 
-        private void Path_Toggled(object sender, EventArgs args)
+        private void Path_Toggled(object o, EventArgs args)
```
AcK77: Keep sender isn’t better for consistency?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -865,5 +777,10 @@ private static int FileSizeSort(ITreeModel model, TreeIter a, TreeIter b)
                 return 0;
             }
         }
+
+        public static void SaveConfig()
+        {
+            ConfigurationState.Instance.ToFileFormat().SaveConfig(System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
```
AcK77:
```suggestion
            ConfigurationState.Instance.ToFileFormat().SaveConfig(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Config.json"));
```
Since there already is a using

Ryujinx/Ui/NpadController.cs
```diff
@@ -1,160 +1,62 @@
...
-            Index            = index;
-            Deadzone         = deadzone;
-            TriggerThreshold = triggerThreshold;
-            LeftJoycon       = leftJoycon;
-            RightJoycon      = rightJoycon;
+            return _inner.Enabled && Joystick.GetState(_inner.Index).IsConnected;
         }
 
-        public void SetEnabled(bool enabled)
+        public void Initialize()
```
AcK77:
```suggestion
        public void Initialize() { }
```
But maybe it’s fine to remove it, since it doesn’t do anything!

Ryujinx.Common/Configuration/ConfigurationState.cs
Thog: This can't be done because this conflict with the Logger field.

Ryujinx/Ui/MainWindow.cs
Thog: Conflicting with GTK.Widget.IO

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,502 @@
...
+                ControllerType            = Hid.ControllerType,
+                GuiColumns                = new GuiColumns()
+                {
+                    FavColumn        =  Ui.GuiColumns.FavColumn,
+                    IconColumn       =  Ui.GuiColumns.IconColumn,
+                    AppColumn        =  Ui.GuiColumns.AppColumn,
+                    DevColumn        =  Ui.GuiColumns.DevColumn,
+                    VersionColumn    =  Ui.GuiColumns.VersionColumn,
+                    TimePlayedColumn =  Ui.GuiColumns.TimePlayedColumn,
+                    LastPlayedColumn =  Ui.GuiColumns.LastPlayedColumn,
```
Xpl0itR: nit: remove extra space after equals sign

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -37,20 +37,53 @@ static Logger()
             m_LogTargets = new List<ILogTarget>();
 
             m_Time = Stopwatch.StartNew();
+
+            // logger should log to console by default
+            AddTarget(new AsyncLogTargetWrapper(
+                new ConsoleLogTarget("console"),
+                1000,
```
gdkchan: Same comment as before. Maybe we should use that as the default value, since the same value seems to be used everywhere?

Ryujinx.Common/Logging/Targets/AsyncLogTargetWrapper.cs
```diff
@@ -27,6 +27,8 @@ public class AsyncLogTargetWrapper : ILogTarget
 
         private readonly int _overflowTimeout;
 
+        string ILogTarget.Name { get => _target.Name; set => throw new NotImplementedException(); }
```
gdkchan: Maybe using a getter-only property would be better here.

Ryujinx.Common/Logging/Targets/ILogTarget.cs
```diff
@@ -5,5 +5,7 @@ namespace Ryujinx.Common.Logging
     public interface ILogTarget : IDisposable
     {
         void Log(object sender, LogEventArgs args);
+
+        string Name { get; protected set; }
```
gdkchan: `string Name { get; }` maybe? Since the interface is for public access.

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -0,0 +1,57 @@
...
+        private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
+        private T                _value;
+
+        public event EventHandler<ReactiveEventArgs<T>> Event;
+
+        public T Value
+        {
+            get
+            {
+                _readerWriterLock.AcquireReaderLock(int.MaxValue);
```
gdkchan: If you want a infinite timeout you can just pass `Timeout.Infinite` (which is just -1).

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public static implicit operator T(ReactiveObject<T> obj)
+        {
+            return obj.Value;
+        }
+    }
+
+    public class ReactiveEventArgs<T>
+    {
+        public T OldValue { get; private set; }
```
gdkchan: `private set` is not necessary since you initialize it from the constructor.

Ryujinx.HLE/Switch.cs
```diff
@@ -60,6 +64,29 @@ public Switch(IGalRenderer renderer, IAalOutput audioOut)
             VsyncEvent = new AutoResetEvent(true);
         }
 
+        public void Initialize()
+        {
+            System.State.SetLanguage((SystemLanguage)ConfigurationState.Instance.System.Language.Value);
+
+            EnableDeviceVsync = ConfigurationState.Instance.Graphics.EnableVsync;
+
+            // TODO: Make this relodable and implement Docking/UnDocking logic.
```
gdkchan: Typo: relodable. Also UnDocking should be Undocking I believe.

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static RichPresence DiscordPresence { get; private set; }
+
+        public static void Initialize()
+        {
+            DiscordPresence = new RichPresence
+            {
+                Assets     = new Assets
+                {
+                    LargeImageKey  = "ryujinx",
+                    LargeImageText = "Ryujinx is an emulator for the Nintendo Switch"
```
gdkchan: Ryujinx is a Nintendo Switch emulator.

gdkchan: Better to put a string const somewhere, to avoid repetition and the texts being different at some point (as they already are).

Ryujinx/Program.cs
```diff
@@ -16,9 +17,32 @@ static void Main(string[] args)
...
+
+            // Now load the configuration as the other subsystem are now registered
```
gdkchan: subsystem**s**

Ryujinx/Ui/MainWindow.cs
```diff
@@ -187,16 +160,16 @@ private void UpdateColumns()
...
-            if (SwitchSettings.SwitchConfig.GuiColumns.VersionColumn)    { _gameTable.AppendColumn("Version",     new CellRendererText(),   "text",   4); }
-            if (SwitchSettings.SwitchConfig.GuiColumns.TimePlayedColumn) { _gameTable.AppendColumn("Time Played", new CellRendererText(),   "text",   5); }
-            if (SwitchSettings.SwitchConfig.GuiColumns.LastPlayedColumn) { _gameTable.AppendColumn("Last Played", new CellRendererText(),   "text",   6); }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileExtColumn)    { _gameTable.AppendColumn("File Ext",    new CellRendererText(),   "text",   7); }
-            if (SwitchSettings.SwitchConfig.GuiColumns.FileSizeColumn)   { _gameTable.AppendColumn("File Size",   new CellRendererText(),   "text",   8); }
-            if (SwitchSettings.SwitchConfig.GuiColumns.PathColumn)       { _gameTable.AppendColumn("Path",        new CellRendererText(),   "text",   9); }
+            if (ConfigurationState.Instance.Ui.GuiColumns.FavColumn)        { _gameTable.AppendColumn("Fav",         favToggle,                "active", 0); }
```
gdkchan: Could remove the { } here aswell since you changed the other one.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -248,6 +230,9 @@ internal void LoadApplication(string path)
             {
                 Logger.RestartTime();
 
+                // TODO: move this somewhere else + reloadable?
```
gdkchan: Nit: `move` -> `Move`.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -394,6 +346,7 @@ private static void CreateGameWindow()
             }
         }
 
+
```
gdkchan: Double empty line.

Ryujinx/Ui/NpadController.cs
```diff
@@ -1,160 +1,57 @@
 ﻿using OpenTK;
 using OpenTK.Input;
+using Ryujinx.Common.Configuration.Hid;
 using Ryujinx.HLE.Input;
 using System;
 
+using InnerNpadController = Ryujinx.Common.Configuration.Hid.NpadController;
```
gdkchan: Not a huge fan of that, would be nice if we could just use `NpadController` directly, but I guess there are some name clashes, would be nice to rename instead to avoid the problem.

Ryujinx/Ui/SwitchSettings.cs
```diff
@@ -7,17 +7,17 @@
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using Ryujinx.Configuration;
```
gdkchan: Sort usings.

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+
+        private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
+        {
+            if (e.NewValue)
+            {
+                Logger.AddTarget(new AsyncLogTargetWrapper(
+                    new FileLogTarget(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx.log"), "file"),
+                    1000,
```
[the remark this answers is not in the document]
jduncanator: It’s the amount of log messages to buffer asynchronously before dropping log messages if it can’t output them fast enough. The parameter is named properly so I see no reason to move this out to a constant, otherwise we might as well move every integer used in the code out to constants.

Also, this should probably be a change for another PR if you’re really set on changing it.

Ryujinx.Common/Logging/Logger.cs
jduncanator: This code is part of the generic logging code and the amount of messages would be different for whatever code used it. Setting default values is a recommendation to the consumer of the API, and we shouldn’t be making recommendations on this use. Again, the code is properly named and documented, so anyone looking at this code can read the documentation on the function about what that parameter does, no need to make it explicit.

Ryujinx.Common/Logging/Targets/AsyncLogTargetWrapper.cs
jduncanator: It’s implementing an interface.

Ryujinx.Common/Configuration/LoggerModule.cs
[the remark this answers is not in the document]
gdkchan: I usually do that to make clear to someone reading the code what the parameter, without needing to look at the function signature or the implementation.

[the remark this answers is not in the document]
jduncanator: You don’t have to, it’s XML documented so you just need to hover the function in the code and it will tell you what you need to know.

There is no point documenting the code on both ends.

Ryujinx.Common/Logging/Logger.cs
gdkchan: I guess the reasonable default behavior in this case would be never dropping any log message, which isn't what we do actually, so I guess either way we'd need to pass the value to the constructor.

jduncanator: The default behavior is to never drop or block at all, it will continue buffering messages forever. This constructor overload is only used when you want to change the behavior or adjust the default queue limit of infinity.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -0,0 +1,502 @@
...
+            Ui.GuiColumns.PathColumn.Value         = configurationFileFormat.GuiColumns.PathColumn;
+            Ui.GameDirs.Value                      = configurationFileFormat.GameDirs;
+            Ui.EnableCustomTheme.Value             = configurationFileFormat.EnableCustomTheme;
+            Ui.CustomThemePath.Value               = configurationFileFormat.CustomThemePath;
+            Hid.EnableKeyboard.Value               = configurationFileFormat.EnableKeyboard;
+            Hid.KeyboardControls.Value             = configurationFileFormat.KeyboardControls;
+            Hid.JoystickControls.Value             = configurationFileFormat.JoystickControls;
+        }
+
+        public static void Initialize()
```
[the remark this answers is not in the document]
Thog: This is initialized the same way as LoggerModule and I prefer it to be that way (it's more obvious when it's initialized ect)

Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -10,6 +10,8 @@ static class DiscordIntegrationModule
     {
         private static DiscordRpcClient DiscordClient;
 
+        private static string LARGE_DESCRIPTION = "Ryujinx is a Nintendo Switch emulator.";
```
gdkchan: `LargeDescription`.

Ryujinx.Common/Configuration/LoggerModule.cs
jduncanator: This was never addressed. There is a lot of code bloat here that could affect future maintainability. I’d like to see these all use a single handler that simply updates all the Logger states by reading the current logger state, rather than having a specific handler for each event.

Thog: This was never addressed as the logs level **should** be handled separately in my point of view. Handling it in a single handler is going to messier and even more error prone.

PR #833, opened 2019-11-24
Fix timezone month offset by one

Thog:
"Merry Christmas.. or maybe not? ❄️" edition

This PR fix an extraneous incrementation that shouldn't even have existed in the first place.


Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZone.cs
```diff
@@ -1707,7 +1707,7 @@ internal static ResultCode ToCalendarTime(TimeZoneRule rules, long time, out Cal
                 Time = new CalendarTime()
                 {
                     Year   = (short)calendarTime.Year,
-                    Month  = (sbyte)(calendarTime.Month + 1),
```
jduncanator: lmao

PR #842, opened 2019-12-08
Fix ILogger type and size decoding

Thog:
The type and size are custom encoded integer, not byte.

This fix crashes on games that send messages longer than 127 characters.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Lm/LogService/ILogger.cs
+++ b/Ryujinx.HLE/HOS/Services/Lm/LogService/ILogger.cs
@@ -8,7 +8,7 @@ namespace Ryujinx.HLE.HOS.Services.Lm.LogService
     {
         public ILogger() { }
 
-        private int ReadEncodedInt(BinaryReader reader)
+        private static int ReadEncodedInt(BinaryReader reader)
         {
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/Lm/LogService/ILogger.cs
```diff
@@ -8,6 +8,26 @@ class ILogger : IpcService
     {
         public ILogger() { }
 
+        private int ReadEncodedInt(BinaryReader reader)
```
gdkchan: This should be a static method.

Ryujinx.HLE/HOS/Services/Lm/LogService/ILogger.cs
```diff
@@ -8,6 +8,26 @@ class ILogger : IpcService
...
+
+            do
+            {
+                encoded = reader.ReadByte();
+
+                result += (encoded & 0x7F) << (7 * position);
+
+                position++;
+
+            } while ((encoded & -128) != 0);
```
gdkchan: Imo use 0x80 instead of -128, it's easier to understand what it is doing that way.

PR #852, opened 2019-12-25
Fix remap ioctl map offset

gdkchan:
The struct on switchbrew is wrong, the field at 0x8 is not padding. I confirmed looking at nvservices that it is indeed used as a offset, multiplied by some page size.


Yvar192:
i approve.

PR #856, opened 2020-01-02
Implement a new physical memory manager and replace DeviceMemory

gdkchan:
Changes:
- Create new abstraction to access large blocks of memory, replaces DeviceMemory on HLE project.
- The new memory block class is span based, has overloads with generic types instead of one method for each integer type, plus one for structs etc. Less method and more efficient.
- All accesses are properly checked to ensure they are within bounds. Old DeviceMemory had no checks at all.
- CPU memory manager has been fully rewritten. Mostly the same benefits as memory blocks. Generic methods instead of one for each integer type, cleaner and more compact code. Proper bounds checking.
- CPU now makes use of interfaces for the memory allocator and memory block. This allowed OS memory allocation code to be moved to separate project.
- New interface for CPU memory manager. Now the user can implement their own memory managers if they wish to do so.
- Hardware capabilities no longer needs to generate a method to get CPUID, it now uses .NET intrinsic IsSupported properties. (I did this change mainly because I planned to remove the Compile method that calls JitCache, and make JitCache a instance of Translator, but they changed my mind half way since Compile is also used in other places now).

I replaced a lot of usages in the HLE project to use the new memory manager methods. I also had to change a few structures to be blittable (from what I recall, mostly those on the time service). I recommend double checking those just to be sure they are still correct.

PR #857, opened 2020-01-03
HID shared memory improvements

gdkchan:
This uses a struct for the entire shared memory, rather than reading/writing each value or struct separately. Requires #856.

The main issue is not being able to use fixed size arrays with non-primitive types, so I had to add a few structs to act like an array. We need a better solution for that.


gdkchan:
Superseded by #1003.

PR #859, opened 2020-01-04
Add 32 bits support to HleProcessDebugger

Thog:
This adds stacktrace and symbols parsing for 32 bits binaries.

Stacktrace code based on @riperiperi's work.

===
General system stability improvements to enhance the user's experience.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
@@ -311,7 +311,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
 
             while ((ulong)symTblAddr < (ulong)strTblAddr)
             {
-                ElfSymbol sym = GetSymbol(memory, symTblAddr, strTblAddr);
+                ElfSymbol sym = isAArch32 ? GetSymbol32(memory, symTblAddr, strTblAddr) : GetSymbol64(memory, symTblAddr, strTblAddr);
 
                 symbols.Add(sym);
 
@@ -324,7 +324,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
             }
         }
 
-        private ElfSymbol GetSymbol(MemoryManager memory, long address, long strTblAddr)
+        private ElfSymbol GetSymbol64(MemoryManager memory, long address, long strTblAddr)
         {
             int  nameIndex = memory.ReadInt32(address + 0);
             int  info      = memory.ReadByte (address + 4);
@@ -332,6 +332,25 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
             int  shIdx     = memory.ReadInt16(address + 6);
             long value     = memory.ReadInt64(address + 8);
             long size      = memory.ReadInt64(address + 16);
+
+            string name = string.Empty;
+
+            for (int chr; (chr = memory.ReadByte(strTblAddr + nameIndex++)) != 0;)
+            {
+                name += (char)chr;
+            }
+
+            return new ElfSymbol(name, info, other, shIdx, value, size);
+        }
+
+        private ElfSymbol GetSymbol32(MemoryManager memory, long address, long strTblAddr)
+        {
+            int  nameIndex = memory.ReadInt32(address + 0);
+            int  info      = memory.ReadByte (address + 4);
+            int  other     = memory.ReadByte (address + 5);
+            int  shIdx     = memory.ReadByte (address + 12);
+            long value     = memory.ReadByte (address + 13);
+            long size      = memory.ReadInt16(address + 14);
 
             string name = string.Empty;
 
--- a/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
@@ -80,7 +80,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
 
                 while (framePointer != 0)
                 {
-                    if ((framePointer & 7) != 0 ||
+                    if ((framePointer & 3) != 0 ||
                         !_owner.CpuMemory.IsMapped(framePointer) ||
                         !_owner.CpuMemory.IsMapped(framePointer + 4))
                     {
@@ -100,7 +100,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
 
                 while (framePointer != 0)
                 {
-                    if ((framePointer & 15) != 0 ||
+                    if ((framePointer & 7) != 0 ||
                         !_owner.CpuMemory.IsMapped(framePointer) ||
                         !_owner.CpuMemory.IsMapped(framePointer + 8))
                     {
--- a/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
@@ -72,25 +72,47 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
                 }
             }
 
-            // TODO: ARM32.
-            long framePointer = (long)context.GetX(29);
-
             trace.AppendLine($"Process: {_owner.Name}, PID: {_owner.Pid}");
 
-            while (framePointer != 0)
+            if (context.IsAarch32)
             {
-                if ((framePointer & 7) != 0 ||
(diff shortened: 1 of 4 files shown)


Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -72,25 +72,47 @@ void AppendTrace(long address)
...
+                    if ((framePointer & 7) != 0 ||
+                        !_owner.CpuMemory.IsMapped(framePointer) ||
+                        !_owner.CpuMemory.IsMapped(framePointer + 4))
+                    {
+                        break;
+                    }
+
+                    // Note: This is the return address, we need to subtract one instruction
+                    // worth of bytes to get the branch instruction address.
+                    AppendTrace(_owner.CpuMemory.ReadInt32(framePointer + 4) - 4);
```
gdkchan: Do we need to adjust that for thumb? (2 bytes vs 4 bytes).

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -72,25 +72,47 @@ void AppendTrace(long address)
...
+                    framePointer = _owner.CpuMemory.ReadInt32(framePointer);
                 }
+            }
+            else
+            {
+                long framePointer = (long)context.GetX(29);
+
+                while (framePointer != 0)
+                {
+                    if ((framePointer & 15) != 0 ||
```
gdkchan: I noticed the mask was changed from 7 to 15. Could you explain why? Is the 16 bytes alignment for the frame pointer mandated by the abi?

Thog: So after some checks, the stack pointer is enforced to a 16 bytes alignment in 64 bits and 8 bytes in 32 bits. However, no rules are defined for the frame pointer.
As such, we need to use natural alignment here and on the 32 bits path.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
Thog: it is tricky.. Thumb frame pointer isn't in the same register as ARM state, and can be found in different registers..

riperiperi:
I think you need to add different parsing for `ElfSymbol` too. MK8 trace with the current `ElfSymbol`:

```
Process: Application, PID: 81
   sdk:0x00336c88 Sub0000000001fd2c88
   sdk:0x00336b9c Sub0000000001fd2b9c
   sdk:0x0004a308 Sub0000000001ce6308
   sdk:0x00081d3c Sub0000000001d1dd3c
   sdk:0x00089578 Sub0000000001d25578
   sdk:0x000875a8 Sub0000000001d235a8
   sdk:0x000899b0 Sub0000000001d259b0
   sdk:0x0008a904 Sub0000000001d26904
   main:0x004b5774 Sub00000000006bb774
   sdk:0x0006773c Sub0000000001d0373c
   sdk:0x0004fc64 Sub0000000001cebc64
   sdk:0x0004f710 Sub0000000001ceb710
   sdk:0x00051064 Sub0000000001ced064
   sdk:0x00068408 Sub0000000001d04408
   sdk:0x000680dc Sub0000000001d040dc
   main:0x00539700 Sub000000000073f700
```
With a GetSymbol32 function that follows elf32_sym on https://github.com/torvalds/linux/blob/master/include/uapi/linux/elf.h :

```
Process: Application, PID: 81
   sdk:0x00336c88 Sub0000000001fd2c88
   sdk:0x00336b9c Sub0000000001fd2b9c
   sdk:0x0004a308 nn::diag::detail::GetBacktraceImpl(unsigned int*, unsigned int)
   sdk:0x00081d3c nnmem_nlib_debug_backtrace
   sdk:0x00089578 nn::nlibsdk::heap::DataCorruptionError(void const*)
   sdk:0x000875a8 nn::nlibsdk::heap::TlsHeapCentral::CacheSmallMemory_(unsigned char*, unsigned int, bool)
   sdk:0x000899b0 nn::nlibsdk::heap::CentralHeap::Alloc(unsigned int, unsigned int)
   sdk:0x0008a904 nn::mem::StandardAllocator::Allocate(unsigned int, unsigned int)
   main:0x004b5774 malloc
   sdk:0x0006773c nn::fs::detail::Allocate(unsigned int)
   sdk:0x0004fc64 nn::fs::RomFsFileSystem::Initialize(nn::fs::IStorage*, void*, unsigned int, bool)
   sdk:0x0004f710 nn::fs::RomFsFileSystem::Initialize(std::__1::unique_ptr<nn::fs::IStorage, std::__1::default_delete<nn::fs::IStorage> >&&, void*, unsigned int, bool)
   sdk:0x00051064 nn::fs::RomFsFileSystemDual::Initialize(std::__1::unique_ptr<nn::fs::IStorage, std::__1::default_delete<nn::fs::IStorage> >&&, std::__1::unique_ptr<nn::fs::IStorage, std::__1::default_delete<nn::fs::IStorage> >&&, void*, unsigned int, bool)
   sdk:0x00068408 Sub0000000001d04408
   sdk:0x000680dc nn::fs::MountRom(char const*, void*, unsigned int)
   main:0x00539700 Sub000000000073f700
```

Suggested implementation:
```
private ElfSymbol GetSymbol32(MemoryManager memory, long address, long strTblAddr)
        {
            int nameIndex = memory.ReadInt32(address + 0);
            long value = memory.ReadInt32(address + 4);
            long size = memory.ReadInt32(address + 8);
            int info = memory.ReadByte(address + 12);
            int other = memory.ReadByte(address + 13);
            int shIdx = memory.ReadInt16(address + 14);

            string name = string.Empty;

            for (int chr; (chr = memory.ReadByte(strTblAddr + nameIndex++)) != 0;)
            {
                name += (char)chr;
            }

            return new ElfSymbol(name, info, other, shIdx, value, size);
        }
```

It's worth noting that `symEntSize` is 0x10 for the mk8 symbols, so this layout does seem to be used.

riperiperi:
Oh yeah, it's read on HleProcessDebugger:315, I just added a switch if `isAArch32`.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -72,25 +72,47 @@ void AppendTrace(long address)
...
+                    if ((framePointer & 3) != 0 ||
+                        !_owner.CpuMemory.IsMapped(framePointer) ||
+                        !_owner.CpuMemory.IsMapped(framePointer + 4))
+                    {
+                        break;
+                    }
+
+                    // Note: This is the return address, we need to subtract one instruction
+                    // worth of bytes to get the branch instruction address.
+                    AppendTrace(_owner.CpuMemory.ReadInt32(framePointer + 4) - 4);
```
gdkchan: Would be nice to fix the `- 4` subtraction. I think we should just remove it and the comment, as theres no way to know if a given function on the call stack is thumb or arm.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -72,25 +72,47 @@ void AppendTrace(long address)
...
+                        !_owner.CpuMemory.IsMapped(framePointer) ||
+                        !_owner.CpuMemory.IsMapped(framePointer + 8))
+                    {
+                        break;
+                    }
 
-                framePointer = _owner.CpuMemory.ReadInt64(framePointer);
+                    // Note: This is the return address, we need to subtract one instruction
+                    // worth of bytes to get the branch instruction address.
+                    AppendTrace(_owner.CpuMemory.ReadInt64(framePointer + 8) - 4);
```
gdkchan: Same change here.

Ryujinx.HLE/HOS/Kernel/Process/HleProcessDebugger.cs
```diff
@@ -305,5 +342,24 @@ private ElfSymbol GetSymbol(MemoryManager memory, long address, long strTblAddr)
 
             return new ElfSymbol(name, info, other, shIdx, value, size);
         }
+
+        private ElfSymbol GetSymbol32(MemoryManager memory, long address, long strTblAddr)
```
gdkchan: I think we have a helper method to read structs from memory, or am I mistaken? If that's the case maybe we should use it instead, rather than reading field by field. Otherwise, just disregard that comment.

PR #860, opened 2020-01-05
Fix ReactiveObject initial event not being propagated with boolean types

Thog:
This fix the logger configuration initial state being ignored.

===
General system stability improvements to enhance the user's experience.

### DIFF
--- a/Ryujinx.Common/ReactiveObject.cs
+++ b/Ryujinx.Common/ReactiveObject.cs
@@ -6,8 +6,8 @@ namespace Ryujinx.Common
     public class ReactiveObject<T>
     {
         private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
-        private bool             _isInitialized      = false;
-        private T                _value;
+        private bool _isInitialized = false;
+        private T _value;
 
         public event EventHandler<ReactiveEventArgs<T>> Event;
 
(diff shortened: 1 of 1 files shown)


Ryujinx.Common/ReactiveObject.cs
```diff
@@ -6,6 +6,7 @@ namespace Ryujinx.Common
     public class ReactiveObject<T>
     {
         private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
+        private bool             _isInitialized      = false;
         private T                _value;
```
LDj3SNuD: I think in this case the alignments do not improve the readability of the code, so I would propose:
```
        private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
        private bool _isInitialized = false;
        private T _value;
```

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -6,6 +6,7 @@ namespace Ryujinx.Common
     public class ReactiveObject<T>
     {
         private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
+        private bool             _isInitialized      = false;
```
gdkchan: Alignment of the `=` is off too.

PR #868, opened 2020-01-10
Add GPU Information to Logging

AcK77:


Ryujinx.Graphics.OpenGL/Renderer.cs
```diff
@@ -75,9 +76,20 @@ public ulong GetCounter(CounterType type)
 
         public void InitializeCounters()
         {
+            GetGpuInformation();
+
             _counters.Initialize();
         }
 
+        private void GetGpuInformation()
```
gdkchan: Calling it get sounds like it is supposed to return the information, but this is not the case. I suggest calling it `PrintGpuInformation`.

Ryujinx.Graphics.OpenGL/Renderer.cs
```diff
@@ -75,9 +76,20 @@ public ulong GetCounter(CounterType type)
 
         public void InitializeCounters()
```
gdkchan: I suggest renaming to just Initialize, since it's not just initializing the counters anymore.

Ryujinx.Graphics.OpenGL/Renderer.cs
AcK77: Done!

Ryujinx.Graphics.OpenGL/Renderer.cs
AcK77: Done!

PR #871, opened 2020-01-11
Call EnsureApplicationSaveData when launching a game

Thealexbarney:
qlaunch calls EnsureApplicationSaveData when launching an application, so we need to do the same.

Also workaround the lack of a program registry by manually setting the program ID if needed when calling some savedata functions.


gdkchan:
Needs a rebase.

PR #883, opened 2020-01-12
Rework SVC handling

Thog:
This PR rework the svc handling system to prepare for 32 bits support.

Changelog:

- Add RAttribute to handle arguments positioning manually. (needed for 32 bits support)
- Enforce usage of RAttribute.
- Add 32 bits codepath for SVC handling.
- Pre-generate all SVC handler in the static constructor.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcTable.cs
+++ b/Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcTable.cs
@@ -108,7 +108,6 @@ namespace Ryujinx.HLE.HOS.Kernel.SupervisorCall
             MethodInfo methodInfo = typeof(SvcHandler).GetMethod(svcName);
 
             ParameterInfo[] methodArgs = methodInfo.GetParameters();
-            int numArgs = methodArgs.Count(x => !x.IsOut);
 
             ILGenerator generator = method.GetILGenerator();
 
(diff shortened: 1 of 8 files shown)


Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcTable.cs
```diff
@@ -296,22 +318,41 @@ void ConvertToFieldType(Type sourceType)
...
 
-                ConvertToFieldType(locals[index].LocalType);
+                ConvertToFieldType(local.LocalType);
 
                 MethodInfo info = typeof(ExecutionContext).GetMethod(nameof(ExecutionContext.SetX));
 
                 generator.Emit(OpCodes.Call, info);
             }
 
+            bool IsRegisterInUse(int registerIndex)
```
gdkchan: I think this code could be improved. For example, you can set bits on a bitmask on indicating, with a bit set, which registers are in use. Then cleaning the unused regs would just need to check which bits are not set and clear the corresponding register. This eliminates the need for this method as you can set it on the previous loop.

Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcTable.cs
```diff
@@ -105,11 +108,7 @@ static SvcTable()
             MethodInfo methodInfo = typeof(SvcHandler).GetMethod(svcName);
 
             ParameterInfo[] methodArgs = methodInfo.GetParameters();
-
-            if (methodArgs.Length > SvcFuncMaxArguments)
-            {
-                throw new InvalidOperationException($"Method \"{svcName}\" has too many arguments, max is 8.");
-            }
+            int numArgs = methodArgs.Count(x => !x.IsOut);
```
jduncanator: This seems unused.

PR #791, opened 2019-10-14
System firmware installer

emmauss:
This adds support for installing firmware backups and updates. Supported formats include a non-trimmed game card dump(XCI), a firmware update package in nsp format, or a zip archive of the system content registered directory from a switch. The installer can be access through the "Tools" menu. Do note, you would still need the necessary keys for the firmware you are installing.

This PR is ready for testing and reviews.

PR #884, opened 2020-01-12
Fix race condition in ContentManager

Thog:
This fix a race condition happening since #791 when trying to load a
game via command line.


Ryujinx.HLE/FileSystem/Content/ContentManager.cs
```diff
@@ -827,32 +853,36 @@ public SystemVersion GetCurrentFirmwareVersion()
...
+                                }
                             }
-                        }
 
+                        }
                     }
                 }
             }
 
+
```
gdkchan: Why this line was added?

PR #886, opened 2020-01-12
Name all threads

### LINKED ISSUE #874: Debugging improvement: Name all the threads
We should name all the threads that are created on the emulator to improve the debugging experience.
As it is now, it's hard to know what each one is supposed to do from a first glance as they don't have names.

AcK77:
Close #874


jduncanator:
I think it would be a good opportunity to add the thread name to the logger output, perhaps in lieu of the thread ID?

AcK77:
Done!

PR #892, opened 2020-01-17
Update 32 bits syscalls to match 64 bits implementation

Thog:
🚗 
Implement all 32 bits counterparts of the 64 bits syscalls we currently have.


riperiperi:
LGTM - have also verified that it works on my A32 branch with games.

PR #904, opened 2020-01-24
Fix a crash when closing the main UI

Thog:
Also make sure to dispose the OpenAL context to not leak memory when
unloading the emulation context.


ARMeilleure/Instructions/NativeInterface.cs
```diff
@@ -24,6 +24,11 @@ public ThreadContext(ExecutionContext context, MemoryManager memory)
 
                 ExclusiveAddress = ulong.MaxValue;
             }
+
+            public void Dispose()
```
gdkchan: What is going to call this? Looks unused.

PR #905, opened 2020-01-25
prepo IPrepoService accurate parsing for report

### LINKED ISSUE #838: prepo IPrepoService needs an accurate parsing for report
```prepo IPrepoService SaveReportOld/SaveReportWithUserOld/SaveReport/SaveReportWithUser```

https://github.com/Ryujinx/Ryujinx/blob/bb7600e215456fd271ba1ec2f7b4fa5088ec2802/Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs#L82

All those calls have a play report sent by the game, currently the code who parse this report is totally guessed and dont works in all cases.

Informations in the play report aren't really useful for us, that's why it's fine for now.
But for accuracy, it could be nice to get the real code from RE.

### PR DESCRIPTION
**Pull request by AcK77**
I've found they use msgpack for the report, so I've added a nuget package and deserialize the report in the right way.

Close #838

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
+++ b/Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
@@ -1,7 +1,8 @@
-using MessagePack;
+using MsgPack.Serialization;
 using Ryujinx.Common.Logging;
 using Ryujinx.HLE.Utilities;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
 
 namespace Ryujinx.HLE.HOS.Services.Prepo
@@ -78,7 +79,7 @@ namespace Ryujinx.HLE.HOS.Services.Prepo
             return ResultCode.Success;
         }
 
-        public string ReadReportBuffer(byte[] buffer, string room, UInt128 userId)
+        private string ReadReportBuffer(byte[] buffer, string room, UInt128 userId)
         {
             StringBuilder sb = new StringBuilder();
 
@@ -92,7 +93,7 @@ namespace Ryujinx.HLE.HOS.Services.Prepo
 
             sb.AppendLine($" Room: {room}");
 
-            IDictionary<string, object> payload = MessagePackSerializer.Deserialize<IDictionary<string, object>>(buffer);
+            IDictionary<string, object> payload = Deserialize<IDictionary<string, object>>(buffer);
 
             foreach (KeyValuePair<string, object> field in payload)
             {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs [hunk 1]
```diff
@@ -93,103 +92,12 @@ public string ReadReportBuffer(byte[] buffer, string room, UInt128 userId)
 
             sb.AppendLine($" Room: {room}");
 
-            try
+            foreach (KeyValuePair<object, object> field in MessagePackSerializer.Deserialize<dynamic>(buffer))
```
jduncanator:
```suggestion
            var payload = MessagePackSerializer.Deserialize<IDictionary<string, string>>(buffer);
            
            foreach (var field in payload)
```

It's better to statically type the deserialization type when using MessagePack. In this case we wish to indicate that we expect string data in both the key and value. This will help catch issues earlier when the guest sends through data that isn't string key/value pairs.

AcK77:
I'm currently got parsing issues with some games (`MessagePackSerializationException : Invalid primitive bytes.`), I've to find how to fix that. So I put this PR in `blocked` state for now.

AcK77:
Fixed by using another package for MessagePack because the previous one can't deserialize some types of object. After some tests, everything works fine now!

Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
```diff
@@ -92,7 +92,9 @@ public string ReadReportBuffer(byte[] buffer, string room, UInt128 userId)
 
             sb.AppendLine($" Room: {room}");
 
-            foreach (KeyValuePair<object, object> field in MessagePackSerializer.Deserialize<dynamic>(buffer))
+            IDictionary<string, object> payload = MessagePackSerializer.Deserialize<IDictionary<string, object>>(buffer);
```
jduncanator: Why do you need the type here, when the type is specified in the generic return type? Using `var` here would make this much more readable.

Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
```diff
@@ -92,7 +92,9 @@ public string ReadReportBuffer(byte[] buffer, string room, UInt128 userId)
...
+
+            foreach (KeyValuePair<string, object> field in payload)
```
jduncanator: Ditto, the type of the dictionary is defined in the line before, we don't need to specify the KeyValuePair type here, just use `var`.

PR #917, opened 2020-02-02
Compare shader code using a span instead of individual reads.

riperiperi:
This was causing a bottleneck in a few games. The cost of `UpdateShaderState` is now more in line with the other state changes (though it still wins out). (6% -> 0.7% total cpu in same situation)

`GetSubSize` has been updated with a `maxSize` parameter to prevent it from scanning the whole memory space when the caller expects a small size.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/MemoryAccessor.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/MemoryAccessor.cs
@@ -41,7 +41,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
         {
             ulong processVa = _context.MemoryManager.Translate(gpuVa);
 
-            ulong size = Math.Min(_context.MemoryManager.GetSubSize(gpuVa), maxSize);
+            ulong size = Math.Min(_context.MemoryManager.GetSubSize(gpuVa, maxSize), maxSize);
 
             return _context.PhysicalMemory.GetSpan(processVa, size);
         }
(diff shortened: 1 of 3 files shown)


Ryujinx.Graphics.Gpu/Memory/MemoryAccessor.cs
```diff
@@ -41,7 +41,7 @@ public ReadOnlySpan<byte> GetSpan(ulong gpuVa, ulong maxSize)
         {
             ulong processVa = _context.MemoryManager.Translate(gpuVa);
 
-            ulong size = Math.Min(_context.MemoryManager.GetSubSize(gpuVa), maxSize);
+            ulong size = Math.Min(_context.MemoryManager.GetSubSize(gpuVa, maxSize), maxSize);
```
gdkchan: I believe `Math.Min` is not necessary anymore since now it can't return more than `maxSize`.

gdkchan:
Good catch.

About the `GetSubSize` change, at some point we might want to change the way how shader code is read from the shader translator. I was thinking in maybe using a read callback method that would be used to read code from GPU mapped memory. This would eliminate the need for the `GetSubSize` method completely, and defining a max size for the shader. Although I guess reading would be a bit slower with this approach. Just throwing the idea out there if you want to try it at some point.

Ryujinx.Graphics.Gpu/Memory/MemoryAccessor.cs
riperiperi: Removed that, didn't even notice.

PR #918, opened 2020-02-02
Only enumerate cached textures that are modified when flushing.

riperiperi:
Before this change, all items on the `_cache` would be enumerated when flushing. I was getting an absurdly high CPU usage rivalling that of MacroInterpreter in situations with lots of draw calls.

Question: is the locking necessary? I wasn't able to verify if the GPU textures are ever modified from another thread, so I added it in for safety.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -54,9 +54,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         public LinkedListNode<Texture> CacheNode { get; set; }
 
         /// <summary>
-        /// Texture data modified by the GPU.
+        /// Event to fire when texture data modified by the GPU.
         /// </summary>
-        public bool Modified { get; set; }
+        public event Action<Texture> Modified;
 
         /// <summary>
         /// Start address of the texture in guest memory.
@@ -931,6 +931,14 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             _depth  = info.GetDepth();
             _layers = info.GetLayers();
+        }
+
+        /// <summary>
+        /// Signals that the texture has been modified.
+        /// </summary>
+        public void SignalModified()
+        {
+            Modified?.Invoke(this);
         }
 
         /// <summary>
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -54,9 +54,14 @@ namespace Ryujinx.Graphics.Gpu.Image
         public LinkedListNode<Texture> CacheNode { get; set; }
 
         /// <summary>
-        /// Event to fire when texture data modified by the GPU.
+        /// Event to fire when texture data is modified by the GPU.
         /// </summary>
         public event Action<Texture> Modified;
+
+        /// <summary>
+        /// Event to fire when texture data is disposed.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


gdkchan:
Locking is not necessary since the texture cache is only accessed from the render thread, thats why none of the methods are thread safe.

riperiperi:
Roger that, removed.

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -54,9 +54,9 @@ class Texture : IRange, IDisposable
         public LinkedListNode<Texture> CacheNode { get; set; }
 
         /// <summary>
-        /// Texture data modified by the GPU.
+        /// Event to fire when texture data modified by the GPU.
         /// </summary>
-        public bool Modified { get; set; }
```
jduncanator: Just be aware, that in the future, changes like these are breaking, and won't be possible once we release the GPU project as a NuGet package. Not an issue at the moment, but just something to keep in mind.

gdkchan: I don't see how it is breaking since Texture is not public.

jduncanator: I mean this type of change in general (changin the type of a public property/field).

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -35,6 +36,8 @@ class TextureManager : IDisposable
 
         private readonly AutoDeleteCache _cache;
 
+        private readonly HashSet<Texture> _modified;
```
jduncanator: Don't use a HashSet<T>, use a List<T>. The only thing you're doing is iterating over a list at the end of the day, so there is no need to pay the additional cost of hashing and bucket allocation that comes with a HashSet<T>.

Also, Texture does not have a `GetHashCode` override defined, so adding them to a HashSet will be relatively slow.

gdkchan: Using a list will add the same texture a lot of times, and potentially waste memory and would be slower to iterate (not to mention it would flush the same texture again and again).

Also since Texture is a reference type I think `GetHashCode` should be pretty fast? It would be slow if `Texture` was a struct without a custom `GetHashCode` but thats not the case.

riperiperi: Yep, I used a HashSet here since hashing for reference types is really fast, and to avoid the case of a texture being added multiple times.

jduncanator: Valid point on using a HashSet, however you should still have a `GetHashCode` on `Texture` to avoid boxing and unboxing in the HashSet's equality comparer.

riperiperi: I'm not sure I understand, would overriding `GetHashCode` on Texture to be `RuntimeHelpers.GetHashCode(this)` or `base.GetHashCode()` be faster?

gdkchan:
> Only a single nit. I'm a little wary making ad-hoc inconsistent changes to the code like this, where the approach may be better suited to the entire GPU code base, rather than just spot fixing issues like these.

Which other place do you think would benefit for such a change?

---

One thing I forgot, you need to remove the texture from the list of modified textures on calls to `DisposeTextures`, otherwise textures may be added there and never removed (on the cases where flush is never called).

PR #919, opened 2020-02-02
Suppress warnings from fields never used or never assigned (CS0169 and CS0649)

### LINKED ISSUE #879: Suppress warnings about fields never used
Currently we have a considerable warnings during build for fields that are never used.
What happens is that those fields are used, but the compiler has no way to know that (either because they are accessed through reflection or in some other way).
We should suppress those warning messages to have a clean build output.

Cristallix:
Fix #879.

From what I'm seeing it's only useful for HLE and Graphics.Gpu. IMHO it's better to limit the use of those rules on the minimum of projects inside the solution.

### DIFF
--- a/Ryujinx.Debugger/UI/ProfilerWidget.cs
+++ b/Ryujinx.Debugger/UI/ProfilerWidget.cs
@@ -289,7 +289,7 @@ namespace Ryujinx.Debugger.UI
                                 _sortedProfileData = _sortedProfileData.Where((pair => filterRegex.IsMatch(pair.Key.Search))).ToList();
                             }
                         }
-                        catch (ArgumentException argException)
+                        catch (ArgumentException)
                         {
                             // Skip filtering for invalid regex
                         }
--- a/Ryujinx.Debugger/UI/ProfilerWidget.cs
+++ b/Ryujinx.Debugger/UI/ProfilerWidget.cs
@@ -81,6 +81,7 @@ namespace Ryujinx.Debugger.UI
 
         private SkRenderer _renderer;
 
+#pragma warning disable CS0649
         [GUI] ScrolledWindow _scrollview;
         [GUI] CheckButton    _enableCheckbutton;
         [GUI] Scrollbar      _outputScrollbar;
@@ -90,6 +91,7 @@ namespace Ryujinx.Debugger.UI
         [GUI] CheckButton    _showInactive;
         [GUI] Button         _stepButton;
         [GUI] CheckButton    _pauseCheckbutton;
+#pragma warning restore CS0649
 
         public ProfilerWidget() : this(new Builder("Ryujinx.Debugger.UI.ProfilerWidget.glade")) { }
 
--- a/Ryujinx.Graphics.Gpu/DmaPusher.cs
+++ b/Ryujinx.Graphics.Gpu/DmaPusher.cs
@@ -214,6 +214,8 @@ namespace Ryujinx.Graphics.Gpu
                 }
                 else if (_state.MethodCount != 0)
                 {
+                    CallMethod(word);
+
                     if (!_state.NonIncrementing)
                     {
                         _state.Method++;
--- a/Ryujinx.Graphics.Gpu/DmaPusher.cs
+++ b/Ryujinx.Graphics.Gpu/DmaPusher.cs
@@ -81,9 +81,6 @@ namespace Ryujinx.Graphics.Gpu
 
         private DmaState _state;
 
-        private bool _sliEnable;
-        private bool _sliActive;
-
         private bool _ibEnable;
 
         private GpuContext _context;
@@ -217,11 +214,6 @@ namespace Ryujinx.Graphics.Gpu
                 }
                 else if (_state.MethodCount != 0)
                 {
-                    if (!_sliEnable || _sliActive)
-                    {
-                        CallMethod(word);
-                    }
-
                     if (!_state.NonIncrementing)
                     {
                         _state.Method++;
--- a/Ryujinx.Graphics.Gpu/State/CopyBufferSwizzle.cs
+++ b/Ryujinx.Graphics.Gpu/State/CopyBufferSwizzle.cs
@@ -5,7 +5,6 @@ namespace Ryujinx.Graphics.Gpu.State
     /// </summary>
     struct CopyBufferSwizzle
     {
-
 #pragma warning disable CS0649
         public uint Swizzle;
 #pragma warning restore CS0649
--- a/Ryujinx.Graphics.Gpu/State/CopyBufferSwizzle.cs
+++ b/Ryujinx.Graphics.Gpu/State/CopyBufferSwizzle.cs
@@ -5,7 +5,10 @@ namespace Ryujinx.Graphics.Gpu.State
     /// </summary>
     struct CopyBufferSwizzle
     {
+
+#pragma warning disable CS0649
         public uint Swizzle;
+#pragma warning restore CS0649
 
         /// <summary>
         /// Unpacks the size of each vector component of the copy.
--- a/Ryujinx.Graphics.Gpu/State/CopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/State/CopyTexture.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Graphics.Gpu.State
         public int          Stride;
         public int          Width;
         public int          Height;
-#pragma warning restore CS0649
         public GpuVa        Address;
+#pragma warning restore CS0649
     }
 }
--- a/Ryujinx.Graphics.Gpu/State/CopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/State/CopyTexture.cs
@@ -6,13 +6,15 @@ namespace Ryujinx.Graphics.Gpu.State
     struct CopyTexture
     {
         public RtFormat     Format;
-        public Boolean32    LinearLayout;
         public MemoryLayout MemoryLayout;
+#pragma warning disable CS0649
+        public Boolean32    LinearLayout;
         public int          Depth;
         public int          Layer;
         public int          Stride;
         public int          Width;
         public int          Height;
+#pragma warning restore CS0649
         public GpuVa        Address;
     }
 }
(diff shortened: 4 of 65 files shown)


gdkchan:
Might be better to disable it in a case by case basis, instead of doing it for the entire project.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-pragma-warning

Cristallix:
Indeed it's a better idea !

Through all those files I have some notes about them because I can't be sure if it's really useful or if it will be used later.

- IndexBufferState.EndAddress is not used
- ShaderState.Unknown0x8, MaxRegisters, Type, Unknown0x14, Unknown0x18, Unknown0x1c, Unknown0x20, Unknown0x24, Unknown0x28, Unknown0x2c, Unknown0x30, Unknown0x34, Unknown0x38 and Unknown0x3c are not used
- ViewportExtents.X, Width, Y and Height are not used
- Not sure about modifications in KMemoryManager
- NvChannel is not used, should it be removed ?
- NvHostSyncpt._eventMask I don't know how this value is set
- CharacteristicsHeader.BufferAddress is not used but not sure if I should remove it
- NvMapHandle.Handle and Id are not used
- ProController._wired seems to not be useful
- Profile._profileInstance I don't know how this value is set

AcK77:
I haven't check other fields but :

> ProController._wired seems to not be useful

This could be removed because it's never used (a leftover).

> Profile._profileInstance I don't know how this value is set

It sets here in debug build only: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Debugger/Profiler/Profile.cs#L41

Cristallix:
Thanks for those information @AcK77 ! I've updated the pr accordingly, rebased and removed new warnings.

Ryujinx.Graphics.Gpu/DmaPusher.cs
```diff
@@ -107,11 +104,6 @@ private bool Step()
                 }
                 else if (_state.MethodCount != 0)
                 {
-                    if (!_sliEnable || _sliActive)
-                    {
-                        CallMethod(word);
```
gdkchan: Why this call was removed?

gdkchan:
Just to make it more clear. Those structs on the GPU project are cast from a int array here:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/State/GpuState.cs#L294

The actual structs matches internal registers of the switch GPU. There's no public documentation, the best we have is nouveau documentation (nouveau is a open source linux driver for nvidia GPUs):
https://github.com/envytools/envytools/blob/master/rnndb/graph/gf100_3d.xml#L287
Compare that with this struct for example, to see what Im talking about:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/State/RtColorState.cs#L6
So, even if the fields are unused, we can't remove or reorder them, as the structure must match the GPU registers structure.

Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
```diff
@@ -63,7 +63,6 @@ class KMemoryManager
 
         public int AddrSpaceWidth { get; private set; }
 
-        private bool _isKernel;
```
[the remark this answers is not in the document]
Cristallix: Done

Ryujinx.Graphics.Gpu/DmaPusher.cs
Cristallix: I didn't pay attention to that exclamation mark, it's restored now.

Cristallix:
Thanks for your patience and your explanations @gdkchan. In my daily work I don't use struct and I'm not a low level developer so I don't really know how it's working under the hood. It was an interesting things to learn :)

Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
```diff
@@ -62,8 +62,9 @@ class KMemoryManager
         private bool _aslrDisabled;
 
         public int AddrSpaceWidth { get; private set; }
-
+
```
gdkchan: Nit: Remove space.

gdkchan:
This needs a rebase.

Ryujinx.Graphics.Gpu/State/CopyTexture.cs
```diff
@@ -6,13 +6,17 @@ namespace Ryujinx.Graphics.Gpu.State
     struct CopyTexture
     {
         public RtFormat     Format;
+#pragma warning disable CS0649
```
gdkchan, line 9: I think in cases like this, its better to supress the warning for the whole struct, instead of disabling and restoring field by field.

Cristallix: Indeed it's cleaner. I've updated every struct I've modified and wrapped all fields.

Ryujinx.Debugger/UI/ProfilerWidget.cs
```diff
@@ -90,6 +91,7 @@ public class ProfilerWidget : Box
         [GUI] CheckButton    _showInactive;
         [GUI] Button         _stepButton;
         [GUI] CheckButton    _pauseCheckbutton;
+#pragma warning restore CS0649
 
         public ProfilerWidget() : this(new Builder("Ryujinx.Debugger.UI.ProfilerWidget.glade")) { }
```
AcK77, line 96: Maybe we could do something with 
```
        catch (ArgumentException argException)
```
at L292 to remove the warning?

Ryujinx.Graphics.Gpu/State/CopyBufferSwizzle.cs
```diff
@@ -5,7 +5,10 @@ namespace Ryujinx.Graphics.Gpu.State
     /// </summary>
     struct CopyBufferSwizzle
     {
+
```
AcK77, line 8: Remove the extra empty line

Ryujinx.Graphics.Gpu/State/CopyTexture.cs
```diff
@@ -13,6 +14,7 @@ struct CopyTexture
         public int          Stride;
         public int          Width;
         public int          Height;
+#pragma warning restore CS0649
         public GpuVa        Address;
```
AcK77, line 18: You forget one field here

Ryujinx.HLE/HOS/Services/Arp/ApplicationLaunchProperty.cs
```diff
@@ -8,7 +8,6 @@ class ApplicationLaunchProperty
         public int   Version;
         public byte  BaseGameStorageId;
         public byte  UpdateGameStorageId;
-        public short Padding;
```
AcK77, line 11: This field came from RE, we should keep it for reference

Ryujinx.HLE/HOS/Services/Bluetooth/IBluetoothDriver.cs
```diff
@@ -9,8 +9,6 @@ namespace Ryujinx.HLE.HOS.Services.Bluetooth
     [Service("btdrv")]
     class IBluetoothDriver : IpcService
     {
-        private string _unknownLowEnergy;
```
AcK77, line 12: This came from RE, we should keep it for reference!

Ryujinx.Debugger/UI/ProfilerWidget.cs
Cristallix: I can just remove `argException` or we can add log to the console. You have to tell me what you prefer.

AcK77: I think remove `argException` is fine.

Ryujinx.HLE/HOS/Services/Bluetooth/IBluetoothDriver.cs
```diff
@@ -53,8 +55,6 @@ public ResultCode InitializeBluetoothLe(ServiceCtx context)
             }
             else
             {
-                _unknownLowEnergy = "low_energy";
```
AcK77, line 56: You should keep it too, this value came from RE

Ryujinx.HLE/HOS/Kernel/Ipc/KLightSession.cs
```diff
@@ -7,14 +7,10 @@ class KLightSession : KAutoObject
         public KLightServerSession ServerSession { get; }
         public KLightClientSession ClientSession { get; }
 
-        private bool _hasBeenInitialized;
```
Cristallix, line 10: @AcK77 Should I keep those lines too ?

Ryujinx.HLE/Input/Controller/ProController.cs
```diff
@@ -2,17 +2,13 @@
 {
     public class ProController : BaseController
     {
-        private bool _wired = false;
```
Cristallix, line 5: Here too ?

Ryujinx.HLE/HOS/Kernel/Ipc/KLightSession.cs
AcK77: Remove this one is fine ;)

Ryujinx.HLE/Input/Controller/ProController.cs
AcK77: Idk why there is this unused var so remove it is fine too ;)

Cristallix:
Rebased. I guess it should be OK.

Ryujinx.HLE/HOS/Services/Bluetooth/IBluetoothDriver.cs
```diff
@@ -54,7 +56,7 @@ public ResultCode InitializeBluetoothLe(ServiceCtx context)
             else
             {
                 _unknownLowEnergy = "low_energy";
-
+
```
AcK77, line 59: Remove extra space here!

Cristallix: Thanks Visual Studio T_T

PR #920, opened 2020-02-03
[WIP] Single File Distribution for Ryujinx

MelonSpeedruns:
This PR makes Ryujinx compile into one giant .exe with all the DLL and dependancies inside it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,11 +14,11 @@ build_script:
 - ps: >-
     dotnet --version
 
-    dotnet publish -c $env:config -r win-x64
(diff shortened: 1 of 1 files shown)


jduncanator:
I don't think single file builds are the way to go. IIRC, single file publishes are just a compressed archive of all the dependencies, and at _each run_ they all have to be extracted before the executable can boot. Given how long it currently takes to extract the resources from the distribution ZIP, I can't imagine this is a fast process.

If it has no overhead to launching Ryujinx, then I don't see any issues, otherwise I think the solution is to move all the dependencies into a subfolder, and have a single Ryujinx.exe in the main folder.

MelonSpeedruns:
Going to use a better system in another PR.

PR #923, opened 2020-02-06
Extend range of constant buffers used to search for storage buffer descriptors on compute shaders

gdkchan:
This improves the codegen and also fixes some shaders where it uses previously ignored constant buffer ranges to store address and size information of the storage buffers.


gdkchan:
This was superseded by #924.

PR #926, opened 2020-02-09
Implement a shader cache

gdkchan:
This implements a persistent shader cache on disk. Right now it only uses what is know as "binary program" which is GPU specific shader code returned by the driver. Need to check driver support and if a different method will be necessary.

Still need to compare performance and probably optimize it, decide where it should be saved and where paths should be handled etc.


BSoDGamingYT:
Some things that would need adding alongside this would be a right-click GUI option on a game, to open shader cache location. While finding the cache is easy for power users, it should be easily reachable for the average user too.

Tellilum:
Shader heavy games like Xenoblade Chronicles might benefit from an Async shader cache.

Ghostotakoi:
When merged this PR will close https://github.com/Ryujinx/Ryujinx/issues/1542

magiccarpet13:
look i realize this is probably a stupid question to ask on github..but as many times as i have run into this problem i HAVE to ask. How do i add this to my ryujinx? and for future reference, PLEASE someone point me in the direction of some info i can read up on how to add code like this in other departments? i am not new to modding, but i CAN NOT figure out how to use creations like this from github when there is not a download button on the github page! :( i realize it is a shameful lack of knowledge, but PLEASE understand that i HAVE TRIED and simply dont even know the words to google to figure this out! is this something that HAS already been created and that i CAN somehow download and add to my ryujinx? or is this something that is not possible to add yet?

Thog:
Closing as stale.

PR #927, opened 2020-02-09
Improve build identification

### LINKED ISSUE #887: Improve build identification
We need to add the commit hash and build number in our CI builds to make it easier to track issues.

Thog:
Add versioning on CI build.

Fix #887

image

image


Ryujinx/Ryujinx.csproj
```diff
@@ -6,6 +6,7 @@
     <OutputType>Exe</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Configurations>Debug;Release;Profile Debug;Profile Release</Configurations>
+    <Version>1.0.0-dirty</Version>
```
Xpl0itR: Is this the default version information when not being built by appveyor? If so I think it should say something like `Custom-Build`

PR #929, opened 2020-02-11
Support for frame defined swap intervals.

emmauss:
Add support for swap interval defined by the buffer sent to the surface flinger. This should solve issues with 30fps games like FE3H running at 60fps and thus, slowing down internally.


jduncanator:
I don't think this is the correct behaviour, but neither was the behaviour we had before.

In my opinion, the entire vsync code should be rewritten rather than "patched" like this, but so long as this has no regressions (in my testing, some changes I made to the vsync code made games that couldn't hit vsync run significantly slower, by up to 10fps), then I don't see any reason to _not_ merge this.

emmauss:
rebased

jduncanator:
This needs a lot of testing before merging, to ensure we don't regress game performance or break games that depend on vsync events for timing.

Thog:
This is superseded by #981

PR #930, opened 2020-02-12
Some fixes for the new GLRenderer

Thog:
Changelog:
- Fix transparency of the window on some games on Windows.
- Fix escape key not being able to exit emulation.
- Fix inverted logic in fullscreen event handling.
- Fix a race condition when stoping emulation causing a hang.
- Fix a memory leak of the OpenGL context when stoping emulation (saving ~200MB of RAM when stoping emulation).
- Simplify and document behaviours when exiting the emulator while the
emulation is running.

### DIFF
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -63,6 +63,7 @@ namespace Ryujinx.Ui
 
             this.Initialized += GLRenderer_Initialized;
             this.Destroyed += GLRenderer_Destroyed;
+            this.ShuttingDown += GLRenderer_ShuttingDown;
 
             Initialize();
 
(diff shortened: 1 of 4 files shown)


Ryujinx/Ui/GLRenderer.cs
```diff
@@ -282,6 +283,11 @@ public void Render()
 
                     if (_ticks >= _ticksPerFrame)
                     {
+                        // Ensure that the content is Opaque.
+                        GL.ColorMask(false, false, false, true);
+                        GL.ClearColor(0.0f, 0.0f, 0.0f, 1.0f);
+                        GL.ColorMask(true, true, true, true);
```
gdkchan: This will affect the GPU state and probably break rendering in a few cases, since you enable all the color channels, when some of them could be previously disabled before.

jduncanator: Also, you shouldn't need ColorMask at all in this instance. `GL.ClearColor` doesn't actually clear anything, it just sets the color that is used when `GL.Clear(COLOR_BUFFER)` is called.

ColorMask would only affect the actual `GL.Clear`, which doesn't happen at all between you changing the ColorMask here, so not sure if this code does anything apart from changing the ClearColor?

Thog: As this code wasn't working, I reworked it in Window::Present and made sure to restore the previous state of ColorMask.

PR #931, opened 2020-02-13
Replace LinkedList by IntrusiveList to avoid allocations on JIT

gdkchan:


jduncanator:
What happened to implement IEnumerable<T> on the IntrusiveList? I thought that was happening before getting merged?

AcK77:
@gdkchan told me this PR is good for merging. Maybe he plans to do that in another one ?

gdkchan:
You should leave a comment on the PR if it needs to be implemented. In any case, if it was implemented, it would remain unused right now, as the point of those changes is reducing allocations and using foreach does the opposite.

PR #933, opened 2020-02-13
Remove ScopedGlContext

Thog:
As MakeCurrent can be slow to execute, this PR changes GLRenderer to not
have to reexecute MakeCurrent when processing frames.


Ryujinx/Ui/GLRenderer.cs
```diff
@@ -234,14 +234,7 @@ public void Exit()
...
 
             IsStopped = true;
-            IsActive = false;
-
-            using (ScopedGlContext scopedGLContext = new ScopedGlContext(WindowInfo, GraphicsContext))
-            {
-                _device.DisposeGpu();
-            }
-
-            WaitEvent.Set();
```
Thog: This was removed as WaitEvent is only used to notify that the GL context exclusivity have been dropped when starting the widget.

PR #948, opened 2020-02-18
Move status information from the title bar to the new status bar

Xpl0itR:
## Move status information from the title bar to the new status bar
Untitled

_Ac_K Edit:
This is what's the informations currently looks like_
unknown
unknown (1)

### DIFF
--- a/Ryujinx/Ui/MainWindow.glade
+++ b/Ryujinx/Ui/MainWindow.glade
@@ -399,10 +399,10 @@
             </child>
             <child>
               <object class="GtkBox" id="_footerBox">
-                <property name="visible">True</property>
-                <property name="can_focus">False</property>
                 <property name="width_request">1280</property>
                 <property name="height_request">19</property>
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
                 <child>
                   <object class="GtkBox" id="_listStatusBox">
                     <property name="visible">True</property>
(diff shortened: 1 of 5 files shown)


Ryujinx/Ui/MainWindow.glade
```diff
@@ -363,6 +363,8 @@
             <property name="orientation">vertical</property>
             <child>
               <object class="GtkBox" id="_viewBox">
+                <property name="width_request">1280</property>
```
Thog: Those two line will break #944, this need to be remove.

Xpl0itR: Fixed. No idea why that was added, i didn't touch any width/height values

jduncanator:
Use proper separators instead of the pipe character (`|`) in the status bar between each of the items (Vsync, Host FPS and Guest FPS), much like the system version number in the bottom right has to it's left.

Also, it would be worth while adding a host frame time counter to the status bar. Frame time is linear measurement (vs FPS which is exponential), so it's a more useful metric when optimising performance.

Xpl0itR:
>Also, it would be worth while adding a host frame time counter to the status bar. Frame time is a linear measurement (vs FPS which is exponential), so it's a more useful metric when optimising performance.

`PerformanceStatistics` Doesn't have a function that returns the average frame time

jduncanator:
> > Also, it would be worth while adding a host frame time counter to the status bar. Frame time is a linear measurement (vs FPS which is exponential), so it's a more useful metric when optimising performance.
> 
> `PerformanceStatistics` Doesn't have a function that returns the average frame time

It should be fairly easy to add.

BaronKiko:
Iirc the profiler shows frame times so you can use that as a dev. It also shows frame pacing as vertical lines on the graph. So it should be obvious if some profiled event is causing a lag spike
I still think it's a good idea to add it to the main GUI. I did it once but it never got merged because the title at the time was too cluttered and the profiler shows it. With the extra space you can simply use the same metrics the profiler uses (it has an average and an instantaneous value)

Xpl0itR:
Added frame time counter to status bar

Ryujinx/Ui/StatusUpdatedEventArgs.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Ryujinx.Ui
+{
+    public class StatusUpdatedEventArgs : EventArgs
```
gdkchan: Would be better to remove the setters and instead add the constructor, and use the constructor passing all the 3 values.

jduncanator:
The frame time in the status bar updates _way_ too fast to be useful, it needs some sort of averaging (much like the other FPS based values have) applied.

Ryujinx/Ui/StatusUpdatedEventArgs.cs
Xpl0itR: Done

jduncanator:
From some brief testing, it looks like the frame times are limited to a minimum of the FPS cap? The frame time should be a measurement of time actually spent doing the work. At the moment, when the Host FPS is capped at 60fps, the frame time never drops below 16ms, even though it may only be spending, say, 10ms.

gdkchan:
The render thread is blocked to lock the frame rate to 60 fps, so it will never go below that, unless we remove the code that is currently blocking it. In any case I would say this should be done separately, not on this PR.

jduncanator:
> The render thread is blocked to lock the frame rate to 60 fps, so it will never go below that, unless we remove the code that is currently blocking it. In any case I would say this should be done separately, not on this PR.

I would assume that the frame time counter would stop timing after the GPU work has been completed, but before the GPU thread has been blocked, no? If that's the case, the Host frame time shouldn't include any of the time spent whilst the thread is blocked waiting for the next vsync.

gdkchan:
From what I recall no, it doesn;t stop timing.

Xpl0itR:
Due to the frame time counter having an issue and the fact that its addition was not even relevant to this PR in the first place, it has been removed

PR #946, opened 2020-02-17
Use GitHub Releases for official builds 

Thog:
This get ride of AppVeyor for official builds once and for all.

As people have been using AppVeyor for custom build and as number might clash with official releases, the minor version of the project was incremented.

AppVeyor will still be used for PRs.

Also to avoid more confusion between AppVeyor builds and official one, AppVeyor builds are now postfixed with "-prXXX".

PR #903, opened 2020-01-23
[WIP] Implement Application Update System

DrHacknik:
This implements a basic Appveyor (and GitHub releases later on) parser, and update system into Ryujinx. 

Pull request, [811](https://github.com/Ryujinx/Ryujinx/pull/811) needs to be merged (preferably) before this is. As the update system will override all files within the Ryujinx directory. 

**What is working and implemented so far**

- Parse appveyor's API 
- Download the appropriate package for the user's platform 
- Extracting the package to a temporary directory, copying and overwriting the old files, and cleaning the temporary files up. 

**What needs to be refined/implemented**

- [ ]  Parse GitHub Releases (CI) API 

- [x] Clean the codebase a bit, especially in the parsing `UpdateParser.cs` 

- [x] Refine the way exceptions are handled for the updater

- [x] Allow Ryujinx to recover from fatal exceptions and restore previous install files.

- [x] Refine the GtkDialog(s) and how they appear/dispose

- [x] Make the updater 'self-aware' and/or use a better method for executing updates

The updater still needs to be tested thoroughly before merging. Testing will not work as expected for now, with the current Ryujinx builds from appveyor. As they don't have the update management for command line arguments :p

PR #950, opened 2020-02-19
Ryujinx Updater

MelonSpeedruns:
Based on #903
Needs #946 to work


MelonSpeedruns:
Going to have to be rebased when Github releases are a thing.

PR #952, opened 2020-02-26
Implement Fast Paths for most A32 SIMD instructions

riperiperi:
This PR adds x86 intrinsic paths to most of the A32 SIMD instructions. This should bring performance more in line with A64, and the slow/non SSE paths should not be affected.

Since this adds fast paths for most instructions I'll go over the omissions instead. 

Omissions:
- Memory instructions.
- VABS+VNEG integer, VDUP, integer multiply (and fused variants) *
- Vector by scalar for integer *
- Narrow ops (there are only like 2 right now)
- VRECPE.f64, VRSQRTE.f64 (integer is not implemented at all, 64 has no x86 intrinsic)
- VCVT f32<->f64. *
- VCVT f32/64 to integer when using fpscr for rounding mode.
- VCVT integer to f32/64. *
- Anything with round mode "away from zero"
- Move immediate. *
- Shift instructions.
- Integer compare instructions.

Everything else should be covered.

Starred ommissions are ones that don't really have any complicated reason to avoid implementing them. Due to the extract/insert experience you have to go through for using scalars in A32, implementing the scalar forms probably won't be much faster.

Two new intrinsics have been added to the list to allow easier movement of scalar values, since A32 allows you to access them from indexed locations within a V128 and x86 does not:

- insertps
- movss

No tests have been added in this version, though all existing ones (that don't require fastfp = false) should pass. Some tests were added to a32-wip alongside the development of these intrinsics to ensure they were correct, which are currently in master.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
@@ -182,33 +182,79 @@ namespace ARMeilleure.Instructions
             }
         }
 
+        private static (long, long) MaskHelperByteSequence(int start, int length, int startByte)
+        {
+            int end = start + length;
+            int b = startByte;
+            long result = 0;
+            long result2 = 0;
+            for (int i=0; i<8; i++)
+            {
+                result |= (long)((i >= end || i < start) ? 0x80 : b++) << (i * 8);
+            }
+            for (int i = 8; i < 16; i++)
+            {
+                result2 |= (long)((i >= end || i < start) ? 0x80 : b++) << ((i-8) * 8);
+            }
+            return (result2, result);
+        }
+
         public static void Vext(ArmEmitterContext context)
         {
             OpCode32SimdExt op = (OpCode32SimdExt)context.CurrOp;
-
             int elems = op.GetBytesCount();
             int byteOff = op.Immediate;
 
-            Operand res = GetVecA32(op.Qd);
-
-            for (int index = 0; index < elems; index++)
+            if (Optimizations.UseSsse3)
             {
-                Operand extract;
-
-                if (byteOff >= elems)
+                EmitVectorBinaryOpSimd32(context, (n, m) =>
                 {
-                    extract = EmitVectorExtractZx32(context, op.Qm, op.Im + (byteOff - elems), op.Size);
-                }
-                else
-                {
-                    extract = EmitVectorExtractZx32(context, op.Qn, op.In + byteOff, op.Size);
-                }
-                byteOff++;
+                    //writing low to high of d: start <imm> into n, overlap into m
+                    //so rotate n down by <imm>, m up by (elems)-imm
+                    //then OR
 
-                res = EmitVectorInsert(context, res, extract, op.Id + index, op.Size);
+                    (long nMaskHigh, long nMaskLow) = MaskHelperByteSequence(0, elems - byteOff, byteOff);
+                    (long mMaskHigh, long mMaskLow) = MaskHelperByteSequence(elems - byteOff, byteOff, 0);
+                    Operand nMask, mMask;
+                    if (op.Q)
+                    {
+                        nMask = X86GetElements(context, nMaskHigh, nMaskLow);
+                        mMask = X86GetElements(context, mMaskHigh, mMaskLow);
+                    }
+                    else
+                    {
+                        nMask = X86GetAllElements(context, nMaskLow);
+                        mMask = X86GetAllElements(context, mMaskLow);
+                    }
+                    Operand nPart = context.AddIntrinsic(Intrinsic.X86Pshufb, n, nMask);
+                    Operand mPart = context.AddIntrinsic(Intrinsic.X86Pshufb, m, mMask);
+
+                    return context.AddIntrinsic(Intrinsic.X86Por, nPart, mPart);
+                });
+            } 
+            else
+            {
+                Operand res = GetVecA32(op.Qd);
+
+                for (int index = 0; index < elems; index++)
+                {
+                    Operand extract;
+
+                    if (byteOff >= elems)
+                    {
+                        extract = EmitVectorExtractZx32(context, op.Qm, op.Im + (byteOff - elems), op.Size);
+                    }
+                    else
+                    {
+                        extract = EmitVectorExtractZx32(context, op.Qn, op.In + byteOff, op.Size);
+                    }
+                    byteOff++;
+
+                    res = EmitVectorInsert(context, res, extract, op.Id + index, op.Size);
+                }
+
+                context.Copy(GetVecA32(op.Qd), res);
             }
-
-            context.Copy(GetVecA32(op.Qd), res);
         }
 
         public static void Vmov_S(ArmEmitterContext context)
--- a/ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
@@ -41,7 +41,11 @@ namespace ARMeilleure.Instructions
 
         public static void Vadd_S(ArmEmitterContext context)
         {
-            if (Optimizations.FastFP)
+            if (Optimizations.FastFP && Optimizations.UseSse2)
+            {
+                EmitScalarBinaryOpF32(context, Intrinsic.X86Addss, Intrinsic.X86Addsd);
+            }
+            else if (Optimizations.FastFP)
             {
                 EmitScalarBinaryOpF32(context, (op1, op2) => context.Add(op1, op2));
             }
@@ -53,7 +57,11 @@ namespace ARMeilleure.Instructions
 
         public static void Vadd_V(ArmEmitterContext context)
         {
-            if (Optimizations.FastFP)
+            if (Optimizations.FastFP && Optimizations.UseSse2)
+            {
+                EmitVectorBinaryOpF32(context, Intrinsic.X86Addps, Intrinsic.X86Addpd);
+            }
+            else if (Optimizations.FastFP)
             {
                 EmitVectorBinaryOpF32(context, (op1, op2) => context.Add(op1, op2));
             } 
@@ -157,7 +165,14 @@ namespace ARMeilleure.Instructions
 
         public static void Vmov_S(ArmEmitterContext context)
         {
-            EmitScalarUnaryOpF32(context, (op1) => op1);
+            if (Optimizations.FastFP && Optimizations.UseSse2)
+            {
+                EmitScalarUnaryOpF32(context, 0, 0);
+            }
+            else
+            {
+                EmitScalarUnaryOpF32(context, (op1) => op1);
+            }
         }
 
         public static void Vmovn(ArmEmitterContext context)
@@ -167,6 +182,7 @@ namespace ARMeilleure.Instructions
 
         public static void Vneg_S(ArmEmitterContext context)
         {
+            //TODO: intrinsic that XORs the sign bit
             EmitScalarUnaryOpF32(context, (op1) => context.Negate(op1));
         }
 
@@ -225,7 +241,11 @@ namespace ARMeilleure.Instructions
 
         public static void Vdiv_S(ArmEmitterContext context)
         {
-            if (Optimizations.FastFP)
+            if (Optimizations.FastFP && Optimizations.UseSse2)
+            {
+                EmitScalarBinaryOpF32(context, Intrinsic.X86Divss, Intrinsic.X86Divsd);
+            }
+            else if (Optimizations.FastFP)
             {
                 EmitScalarBinaryOpF32(context, (op1, op2) => context.Divide(op1, op2));
             }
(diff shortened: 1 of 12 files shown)


ARMeilleure/Instructions/InstEmitSimdHelper32.cs
```diff
@@ -473,6 +473,446 @@ public static void EmitVectorUnaryNarrowOp32(ArmEmitterContext context, Func1I e
...
+                return context.AddIntrinsic(Intrinsic.X86Movhlps, input, input); // High to low.
+            }
+        }
+
+        public static Operand EmitDoubleWordInsert(ArmEmitterContext context, Operand target, Operand value, int targetV)
+        {
+            Debug.Assert(target.Type == OperandType.V128 && value.Type == OperandType.V128);
+
+            int targetSide = targetV & 1;
+            int shuffleMask = 2 | 0;
```
gdkchan: Not sure why the `| 0`, also other methods (well, the one below) seems to directly use a constant value of 2.

ARMeilleure/Instructions/InstEmitSimdHelper32.cs
```diff
@@ -473,6 +473,446 @@ public static void EmitVectorUnaryNarrowOp32(ArmEmitterContext context, Func1I e
...
+
+            if (!op.Q) // Register swap: move relevant doubleword to destination side.
+            {
+                n = EmitSwapDoubleWordToSide(context, n, op.Vn, op.Vd);
+                m = EmitSwapDoubleWordToSide(context, m, op.Vm, op.Vd);
+            }
+
+            Operand res = vectorFunc(d, n, m);
+
+            if (!op.Q) //register insert
```
gdkchan: `// Register insert.`

ARMeilleure/Instructions/InstEmitSimdMove32.cs
```diff
@@ -128,209 +144,411 @@ public static void Vtbl(ArmEmitterContext context)
...
-                        // Result contains the current state of the vector.
-                        lookupResult = context.VectorExtract(OperandType.I64, res, ix);
-                    } 
-                    else
-                    {
-                        lookupResult = EmitVectorExtract32(context, qx, ix, 3, false); // I64
-                    }
-                    
-                    lookupResult = context.ShiftRightUI(lookupResult, subVecIndexShift); // Get the relevant byte from this vector.
+                    (int vn, int en) = GetQuadwordAndSubindex(op.Vn + i, op.RegisterSize);
```
gdkchan: You can just do `tableTuples[i] = GetQuadwordAndSubindex(op.Vn + i, op.RegisterSize);` directly.

ARMeilleure/Instructions/InstEmitSimdMove32.cs
```diff
@@ -128,209 +144,411 @@ public static void Vtbl(ArmEmitterContext context)
...
 
-            bool overlap = op.Qm == op.Qd;
+                            mask = X86GetScalar(context, maskE0);
+
+                            mask = EmitVectorInsert(context, mask, Const(maskE1), 1, 3);
+                        }
+
+                        if (op.Size < 3)
+                        {
+                            d = context.AddIntrinsic(Intrinsic.X86Pshufb, d, mask);
```
gdkchan: I think its better to just add that on the if above to avoid a useless comparison.

ARMeilleure/Instructions/InstEmitSimdMove32.cs
```diff
@@ -128,209 +144,411 @@ public static void Vtbl(ArmEmitterContext context)
...
+                Operand resM = GetVecA32(op.Qm);
+
+                for (int index = 0; index < elems; index++)
                 {
-                    resM = resD;
-                }
+                    Operand dIns, mIns;
+                    if (index >= pairs)
+                    {
+                        int pind = index - pairs;
```
gdkchan: Maybe `pairIndex`?

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
```diff
@@ -15,7 +15,28 @@ static partial class InstEmit32
...
+            OpCode32SimdS op = (OpCode32SimdS)context.CurrOp;
+
+            if (Optimizations.FastFP && Optimizations.UseSse2)
+            {
+                EmitScalarUnaryOpSimd32(context, (m) =>
+                {
+                    if ((op.Size & 1) == 0)
+                    {
+                        Operand mask = X86GetScalar(context, -0f);
+                        return context.AddIntrinsic(Intrinsic.X86Andnps, mask, m);
```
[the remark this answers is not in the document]
gdkchan: This pattern is used on a bunch of places actually, might be better to extract it to a separate function and move it to one of those helper static classes.

ARMeilleure/Instructions/InstEmitSimdCmp32.cs
```diff
@@ -110,7 +160,14 @@ public static void Vclt_Z(ArmEmitterContext context)
 
             if (op.F)
             {
-                EmitCmpOpF32(context, SoftFloat32.FPCompareLTFpscr, SoftFloat64.FPCompareLTFpscr, true);
+                if (Optimizations.FastFP && Optimizations.UseSse2)
+                {
+                    EmitSse2CmpOpF32(context, CmpCondition.LessThanOrEqual, true);
```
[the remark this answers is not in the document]
riperiperi: Good catch, good thing we have reviews 😌

ARMeilleure/Instructions/InstEmitSimdHelper32.cs
```diff
@@ -473,6 +473,446 @@ public static void EmitVectorUnaryNarrowOp32(ArmEmitterContext context, Func1I e
             context.Copy(GetVecA32(op.Qd), res);
         }
 
+        // Intrinsic Helpers
+
+        public static Operand EmitSwapDoubleWordToSide(ArmEmitterContext context, Operand input, int originalV, int targetV)
```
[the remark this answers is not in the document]
riperiperi: It was originally going to be a swap, but the optimisation I wanted to do by keeping the old value wasn't really possible. I'll change the name to match.

ARMeilleure/Instructions/InstEmitSimdHelper32.cs
riperiperi: I wanted to be explicit about the indices that went in each doubleword, but yeah it just looks silly.

ARMeilleure/Instructions/InstEmitSimdMove32.cs
riperiperi: I think there was originally something in the middle, just forgot to combine them when I removed it.

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
```diff
@@ -477,58 +806,115 @@ public static void Vmls_1(ArmEmitterContext context)
...
+                                    return context.AddIntrinsic(Intrinsic.X86Pshufb, op1, mask);
+                            }
+                            break;
+                        case 1:
+                            // Rev16
+                            mask = X86GetElements(context, 0x0e0f_0c0d_0a0b_0809L, 0x_0607_0405_0203_0001L);
+                            return context.AddIntrinsic(Intrinsic.X86Pshufb, op1, mask);
+                    }
+
+                    throw new InvalidOperationException("Unknown VREV Opcode + Size combo.");
```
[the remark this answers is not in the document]
riperiperi: This is unreachable, the decoder gets size from range -1 to 2 (just due to how the value works, it's 2 - size), and sets undefined when == -1.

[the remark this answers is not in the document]
gdkchan: Then maybe add a comment `// Should be unreachable.` and change `Unknown` to `Invalid`.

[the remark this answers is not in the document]
riperiperi: Oops, I was looking at VDUP which has a similar throw. I need to change the opcode for this one to set undefined as defined in the manual:
`if UInt(op)+UInt(size) >= 3 then UNDEFINED;`
I will also change the comment on this.

ARMeilleure/Instructions/InstEmitSimdCmp32.cs
```diff
@@ -224,8 +281,74 @@ private static void EmitVcmpOrVcmpe(ArmEmitterContext context, bool signalNaNs)
...
+                    Operand cf = context.AddIntrinsicInt(Intrinsic.X86Comissge, n, m);
+                    Operand zf = context.AddIntrinsicInt(Intrinsic.X86Comisseq, n, m);
+                    Operand nf = context.AddIntrinsicInt(Intrinsic.X86Comisslt, n, m);
+
+                    EmitSetFPSCRFlags(context, context.BitwiseOr(
+                        context.ShiftLeft(cf, Const(1)),
+                        context.BitwiseOr(
+                            context.ShiftLeft(zf, Const(2)),
+                            context.ShiftLeft(nf, Const(3))
+                            )
```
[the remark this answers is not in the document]
riperiperi: I'm not sure why I'm doing it like this anyways, since the flags are split again afterwards. It's likely a leftover from when the FPSCR flags were set using a managed call. I'll add a path for setting with four boolean operands.

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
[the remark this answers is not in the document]
riperiperi: I'll move it to `InstEmitSimdHelper.EmitFloatAbs`. There are two related uses in A64, and that class is static imported in A32 as well. I'll hold back on modifying A64 beyond the existing uses of this pattern since it's outside the scope of this PR.

riperiperi:
Addressed comments, and fixed a bug with VCVT_R where op.Opc2 could be >3 as we were using all 3 bits instead of the lower two. (special thanks to the new exception)

ARMeilleure/Instructions/InstEmitSimdCvt32.cs
```diff
@@ -250,6 +250,8 @@ private static FPRoundingMode Opc2ToRoundMode(int opc2)
                 case 0b11:
                     roundMode = FPRoundingMode.TowardsMinusInfinity;
                     break;
+                default:
+                    throw new ArgumentOutOfRangeException($"Round mode {opc2} out of supported range.");
```
gdkchan: `ArgumentOutOfRangeException` constructor takes a argument name, not a exception message.

riperiperi: I just realized too, this function should probably just do the (opc2 & 3) rather than the callers since it does advertise that it takes Opc2 raw.

riperiperi: Actually, & 3 is useful for the slow paths. I'll rename the function and argument.

riperiperi:
A little more remodelling done.

ARMeilleure/Instructions/InstEmitSimdCvt32.cs
```diff
@@ -251,7 +251,7 @@ private static FPRoundingMode Opc2ToRoundMode(int opc2)
                     roundMode = FPRoundingMode.TowardsMinusInfinity;
                     break;
                 default:
-                    throw new ArgumentOutOfRangeException($"Round mode {opc2} out of supported range.");
+                    throw new ArgumentOutOfRangeException($"rm");
```
gdkchan: `nameof(rm)`

gdkchan: On a side note Im not a huge fan of the name "rm" since on the manual its usually used as the name of the second operand register.

riperiperi: On the manual it is called `RM` in caps, distinct from `Rm` in case only. The function is named approprately, but the naming convention on variables obviously doesn't allow that.

ARMeilleure/Instructions/InstEmitSimdHelper.cs
```diff
@@ -1467,6 +1467,14 @@ public static Operand EmitBinarySatQAccumulate(ArmEmitterContext context, Operan
             return context.Call(dlg, op1, op2);
         }
 
+        public static Operand EmitFloatAbs(ArmEmitterContext context, Operand value, bool single, bool vector)
+        {
+            Operand mask = vector ? X86GetAllElements(context, -0f) : X86GetScalar(context, -0f);
```
gdkchan: Just a question, is the mask correct for double values? Because it only pass float here regardless of the value of the single argument.

riperiperi: It's not, I just got excited about not having to write a double ternary for the intrinsic selection since we don't have a scalar andn, and forgot about this one.

ARMeilleure/Instructions/InstEmitSimdCvt32.cs
jduncanator: You can pass a message as the second parameter of `ArgumentOutOfRangeException` that has a more detailed message and include `RM` (in caps) in there for distinction.

PR #953, opened 2020-02-27
Implement SMULWB, SMULWT, SMLAWB, SMLAWT (AArch32)

gdkchan:
- Implements 2 new instructions: SMULWx and SMLAWx.
- Adds tests for the 2 new instructions plus SMLAxy and SMULxy.
- Set Q flag on saturation for SMLAxy that was missing.
- Change the name of some instructions, like SMLAB to SMLA__ (since the SMLAB instruction doesn't exist). Re-align opcode table to accomodate the new names.
- Sort methods inside `InstEmitMul32.cs`.


riperiperi:
Also good with the latest changes.

PR #954, opened 2020-02-28
Add SSAT, SSAT16, USAT and USAT16 ARM32 instructions

gdkchan:
Depends on #953 since those instructions may also set the Q flag.


riperiperi:
LGTM, confirmed working in MK8 version 1.7.1 as well.

PR #960, opened 2020-03-01
Implement VMOVL and VORR.I32 AArch32 SIMD instructions

gdkchan:
Implements those 2 instructions needed by Rune Factory 4 Special. Also:
- Remove `fpBaseSize` argument from `GetSimdImmediateAndSize` since it was unused.
- Make `OpCode32SimdImm` inherit `OpCode32SimdBase` To allow the use of the handy Qx and Ix properties.
- Fix wrong `InstName` rename introduced on #953 (ARM32 had a instruction called SMULH which is the same name of a ARM64 instruction, since SMULH doesn't exist on ARM32 I renamed it, but ARM64 was using the same enum entry. I fixed it by adding a new enum entry for ARM64).
- Minor adjustments to improve code readability.


ARMeilleure/Instructions/InstEmitSimdLogical32.cs
```diff
@@ -64,6 +67,42 @@ public static void Vorr_I(ArmEmitterContext context)
             }
         }
 
+        public static void Vorr_II(ArmEmitterContext context)
+        {
+            OpCode32SimdImm op = (OpCode32SimdImm)context.CurrOp;
+
+            long immediate = op.Immediate;
+
+            // Replicate fields to fill the 64-bits, if size is < 64-bits.
```
riperiperi: Might be a good idea to improve `EmitVectorImmUnaryOp32` to do this as well.

riperiperi: In fact it might be a good idea to make that work on `(imm, Vd)` lambdas, so that you can implement this instruction and perhaps similar ones (are there any?) without duplicating code.

gdkchan: The `AdvSIMDExpandImm` function on the manual always produces a 64-bits value, I don't know why the one on the emulator is different. I will check the other instructions that uses this opcode type and see if they depend on the current behavior to work.

gdkchan: I took a look and there is one case that needs the non-replicated immediate, which is VMOV to single register (the value is zero extended instead). So I guess the best thing here would be adding a method to replicate the immediate on the `IOpCode32SimdImm` interface. Then we can use it for vector move. For single reg move we can keep using the non-replicated immediate property. LMK how I should proceed here.

riperiperi: This function was copied from the A64 immediate decoder, which I meant to make reference the helper. It should generate the same immediates as A32. The one difference is that the floating point immediate sizes are `3` rather than `0` and `1`, which is why it had the weird parameter that didn't really do anything. The A64 implementation uses the size, so I'm assuming it doesn't do replication as that would be wasted work.

I'm good with having some helper to replicate. Replication to I64 happens in a few other places so a quick helper function would be nice to clean up all those occurrences. It's also worth noting that a replicated immediate would still work with the scalar ops since it would truncate on insert, it'd just obviously be wasted work.

My comment was just about changing the mentioned function to do work in 64-bit elements rather than `Size` elements just to make things faster.

PR #966, opened 2020-03-04
Add Layered FS support example

Thealexbarney:
This is a basic example of how one might add Layered FS support.

Layered FS mods go in `Ryujinx/lfsContents/<title_id>`


BlueShine7:
Doesn't see to be working on current Ryujinx's versions but looked promising. Will this Pull Request be updated? It would be really awesome to let the emulator use mods. Thank you!

PR #967, opened 2020-03-04
Fix branch with CC and predicate, and a case of SSY/PBK propagation

gdkchan:
This fixes two things:

First, it had a bug for branches using both a predicate and CC condition. When the predicate was inverted, it was inverting the whole expression, not just the predicate. I'm not aware of any specific issue caused by this bug, I just found it while I was looking at the code and decided to fix.

The other one is more complicated. When the shader has BRX instruction (which is basically a indirect branch with address from a register), we currently try to guess the possible targets. It assumes that the BRX is used for switch jump table optimization, and in this case the possible targets can be guessed by finding the common block that all the target branches to (the switch break). This doesn't work very well when PBK/SSY -> BRK/SYNC sequences are used, because the target address is unknown in this case. For this case it could include wrong targets, and this causes the second pass (after decoding/CFG build) that propagates PBK/SSY addresses to their respective BRK/SYNC instructions to not reach some places that are supposed to be reachable. This in turn causes wrong branches to be inserted on the code. It caused a regression on Crash Team Racing, a infinite loop on the shader that probably happened after BRX support was implemented.

To fix the above issue I changed it to delay as much as possible adding the BRX targets on the list of visited block, to allow them to be visited through the SYNC instruction inside the BRX target block, rather than be considered as BRX target up front and then never be visited again.

### DIFF
--- a/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
@@ -390,7 +390,14 @@ namespace Ryujinx.Graphics.Shader.Decoders
 
             void Push(PathBlockState pbs)
             {
-                if (pbs.Block == null || visited.Add(pbs.Block))
+                // When block is null, this means we are pushing a restore operation.
+                // Restore operations are used to undo the work done inside a block
+                // when we return from it, for example it pops addresses pushed by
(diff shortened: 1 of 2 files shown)


ahmedtalaat327:
Is sound problem solved?!

PR #970, opened 2020-03-06
Add dynamic colors when game row is focused

AcK77:
- Little cleanup of `MainWindows.cs`
- Remove unused Updater items
- Add ColorThief nuget to compute colors easily
- Change the game focused row colors based on the game cover dynamically

okmLoeVLpB

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1,5 +1,5 @@
+using ColorThiefDotNet;
 using Gtk;
-using JsonPrettyPrinterPlus;
 using Ryujinx.Audio;
 using Ryujinx.Common.Logging;
 using Ryujinx.Configuration;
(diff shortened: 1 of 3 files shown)


Ryujinx/Ui/MainWindow.cs
```diff
@@ -627,6 +626,51 @@ private void Row_Clicked(object sender, ButtonReleaseEventArgs args)
...
+                            if (!ctBackground.IsDark)
+                            {
+                                ctText = new QuantizedColor(black, 0);
+                            }
+                        }
+                    }
+                }
+
+                CssProvider provider = new CssProvider();
+                provider.LoadFromData("treeview:hover { background-color: " + ctBackground.Color.ToHexString() + "; color: " + ctText.Color.ToHexString() + ";}");
```
Xpl0itR: nit: personally I would use $ string interpolation here

AcK77: Done :)

AcK77:
This change will be added as a setting and the cleanup will be made in a further PR so I close it for now.

PR #975, opened 2020-03-07
Use a Jump Table for direct and indirect calls/jumps, removing transitions to managed

riperiperi:
**_While this PR is not WIP, I recommend testing on as many games as possible before merging._**

This PR changes execution flow in the ARM translator to attempt to find and directly call compiled functions rather than fall back to ExecuteSingle. Thanks to this, rather than returning to a separate function starting from the return point of a call, execution seamlessly returns to the middle of the caller function without any overhead or additional translation. Additionally, calls, indirect calls and indirect jumps now cache the host function address to avoid calling back to managed to get it.

### Extra Features
Direct calls come with a few drawbacks so I've included a few extra features to mitigate them:

- Background translation can now run on up to 4 threads, based on your logical processor count. HighCq code now takes longer to compile, so to mitigate that on stronger processors we compile with more threads.
- The JitCache now dynamically resizes as more code is added to it, to a limit about 4x as large. This is because lowCq code is taking up more space, though this shouldn't be a concern for too much longer (cache eviction and gdk's x86 opts should help).

### Benefits
In CPU limited games, you should see around a **1.25x to 1.75x** increase in performance on average. Some games are more affected than others (specifically ones limited by guest CPU performance). Super Mario Odyssey (with query hack to remove GPU limit), Toad Tracker, Mario Kart 8 see the upper end of this boost, whereas games such as Pokemon Sw/Sh, Zelda Botw and Splatoon 2 will see a more modest boost in select areas. All games should run at least a little bit faster.

In general, CPU emulation speed is now nearly entirely down to code quality, as the main bottleneck around execution itself has been removed.

Specifically, benefits of the approach of using jump tables:
- Solves the problem of returning to managed to find the next code block, which is faster regardless of approach.
- Entirely implemented in IR - so should be portable if other backends are implemented.
- The table is easy to serialize or deserialize for use in future execution. Loading the table is as easy as storing new function host addresses for the tabled function guest addresses. It can even be relocatable if you pass a base address with the NativeContext. You may need to save and register the same dependant functions to better track usage and fill in functions that you don't have later on in execution.

### Drawbacks
~~**Important note: There is an issue where the emulator will crash when attempting to close the game. I don't know enough about the Horizion threading model to know what's going on, but it seems like a synchronisation issue.**~~ Temporary solution implemented, see comments.

Games will likely boot a wee bit slower and take a while to speed up after getting ingame until optimizations to the JIT compiler are made. (separate PRs) This is because the JIT compiler has more work to do both for initial compilation and high quality background compilation. Before, the translator would stop translating a function once it reached a call instruction, as the return would not be able to rejoin to the existing function anyways. Now, it continues past it (after a sanity check), as we know that the function we called will return to the next instruction. Compile time scales much worse than linearly for an increase in function size, though, so despite an increase in execution performance it will take longer to compile these functions.

As a mitigation, the existing limit of 5000 instructions per function call remains for highCq code. Testing showed that compiling lowCq code as functions rather than blocks showed a marginal increase in boot time and pre-rejit performance with an instruction limit of around 500. This could cause retranslation of blocks in certain functions, though I have not seen any serious case of this happening yet.

**One drawback of doing this at all is that ARM function profiling does not work anymore.** Execution does not return to ExecuteSingle unless an exceptional return happens. This means it is also harder to track when functions are used, which will complicate things for #740. Signalling for this will need to be done in IR, which could impact performance, but we only need to do it for lowCq.

Specifically, drawbacks of the approach of using jump tables:
- Locality of reference: Each jump or call that is made has to pull data off the jump table at a fixed address. This could be fine if your CPU has enough cache to hold the code, jump table, page table and whatever the game is doing in there all at once, but it'll definitely strain things if you have less cache or the game is doing lots of random jumps/accesses. (still need to somehow gauge how much of an issue this is)
- More instructions: Loading from memory and calling is a tad slower from just going for a branch directly. Note that this definitely isn't a problem right now, and won't be for a while or only in very specific cases.

### Jump Tables
Before this PR, transitions to other blocks of code were done via a call to a managed function which did a ConcurrentDictionary lookup. This ended up making up a lot of a program's execution time, so we needed a solution that would work entirely without transitioning to managed in the best case, and that would be as close to the speed of a direct jump/call as it possibly could be. 

The solution was to allocate and manage two jump tables - contiguous memory regions filled with (guestAddress, hostAddress) tuples for identifying the translated function we jump to. Each jump/call made in a HighCq function reserves a space on a table, which is loaded and jumped to as directly as possible (see below). Initially, each entry is populated with the address of a stub - which does the slow function lookup as before - for when the target function is not available in highCq. When HighCq versions of functions are compiled, they are filled into the relevant entries (using a fast "dependants" list built during entry reservation) over the stub so that jumps/calls can be made without touching managed at all.

Direct calls and jumps are simple enough - the compiled function completely ignores the `guestAddress` section of the table and jumps directly to whatever `hostAddress` is at the time (it is modified by the translator when it compiles the function). Indirect calls and jumps (to addresses that are not constant) follow a similar, but different pattern. They use a separate table called the "Dynamic" table, which stores a _list_ of entries for each jump/call. The difference is that these entries _do not start with a guest address set_, as we do not know what address the code will end up calling. When the guest does a call, it attempts to _find or reserve_ a spot on the table for the guest address it is calling, using a compare and swap instruction. If the entry is taken by another guest address, it continues... otherwise, it calls the function in the hostAddress part of the entry directly. If the function found is HighCq, it is placed in the table over the stub so that it can be found much quicker in future.

If an indirect call/jump address cannot be found on its table, then it begrudgingly falls back to looking up the address using the managed dictionary. Right now, the table size is set to 1 to reduce code size and compile time, as this does not have much of an impact on most games. Ideally we should be able to hook LowCq functions to track what addresses they call and guess a better table size (and initial population?) for each function when we get to compiling them in HighCq.

For both direct and indirect calls, the target guest address is passed through on the NativeContext so that the dictionary lookup can be performed if we end up calling the stub.

The jump table can only reference highCq functions, as it is expected that they will never change so that we can put the address in the table when it is available and never have to worry about updating it. Also, only highCq functions can reserve space on the jump table at all, to prevent wasting space and time on code that will eventually be succeeded by a higher quality compilation.

The dynamic table has support for caching multiple function calls from one instruction, though table size has currently been limited to 1. Higher values currently don't see much increase in performance, and the volume of IR emitted greatly slows down compliation time.

Note that the result of all this HighCq/LowCq separation is that all HighCq code _only_ references other HighCq code, making it safe to remove LowCq functions when we don't need them. (separate PR)

### Tail Call rather than Return (lowCq or functions with gaps)

This changes how flow works drastically when using the ARM translator. Rather than relying on returning an address to continue from, the native code itself calls the translator to look up a function address, then Tail Calls it without leaving native.

Only things detected as a function return will use the context.Return(address) function. 
In A64, these are:
- BR LR
- RET
In A32, these are:
- BX LR
- LDM {...PC}
- LDx PC, [SP], #0x4
- Any ALU operation that writes to PC, and uses LR as a register input.

Note that we greatly prefers false positives to false negatives - that is we'd rather return on something that may not really be a return than Tail Call the return. The former can cause the guest stack to be thrown out and slow things down, the latter can (eventually) stack overflow and crash the application if no valid return happens later on. See below for more information.

### The curious case of programs that abuse LR

Not all programs return nicely, specifically A64 RTLD has one particularly nasty instance where it branch links to an address, which then modifies LR and returns to a *different end point*.

This is where the return values come in. Around every native function call, there is a guard making sure that the function exited at the right point. If it didn't, we'll bubble that address down the stack until either:
 - We find the stack frame it returned to. The case in RTLD jumps two stack frames down, so it is easily salvagable.
 - We end up returning to GetOrTranslate.

This can cause a huge performance penalty, as exceptional returns like this have to lowCq from the return point. Specifically - before Tail Call was used for joining lowCq blocks, games would typically tank performance while a new game function was being used, but had not been admitted to highCq yet. In the games I have, these exceptional returns happen very rarely, but they are still a concern and appear to be handled correctly with this approach.

### Interop with other PRs
- Should work immediately with X86 optimizations.
- Reasonably separate from Pooling optimizations (to come in a future PR, which will improve multithreaded translation)
- AOT support is more difficult, but thanks to avoiding ever patching functions or baking in function addresses, it should be easy to store the address tables and repopulate them with - the relocated host functions.
- Support for removing unused functions from the cache requires some nuance for tracking uses. It's also important to note that only lowCq functions should be removed, and relocating - functions in memory may be difficult.

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/TailCallRemover.cs
+++ b/ARMeilleure/CodeGen/Optimizations/TailCallRemover.cs
@@ -1,0 +1,75 @@
+﻿using ARMeilleure.Decoders;
+using System;
+using System.Collections.Generic;
+
+namespace ARMeilleure.CodeGen.Optimizations
+{
+    static class TailCallRemover
+    {
+        public static void RunPass(ulong entryAddress, List<Block> blocks)
+        {
+            // Detect tail calls:
+            // - Assume this function spans the space covered by contiguous code blocks surrounding the entry address.
+            // - Unconditional jump to an area outside this contiguous region will be treated as a tail call.
+            // - Include a small allowance for jumps outside the contiguous range.
+
+            if (!Decoder.BinarySearch(blocks, entryAddress, out int entryBlockId))
+            {
+                throw new InvalidOperationException("Function entry point is not contained in a block.");
+            }
+
+            ulong allowance = 4;
+            Block entryBlock = blocks[entryBlockId];
+            int startBlockIndex = entryBlockId;
+            Block startBlock = entryBlock;
+            int endBlockIndex = entryBlockId;
+            Block endBlock = entryBlock;
+
+            for (int i = entryBlockId + 1; i < blocks.Count; i++) // Search forwards.
+            {
+                Block block = blocks[i];
+                if (endBlock.EndAddress < block.Address - allowance)
+                {
+                    break; // End of contiguous function.
+                }
+
+                endBlock = block;
+                endBlockIndex = i;
+            }
+
+            for (int i = entryBlockId - 1; i >= 0; i--) // Search backwards.
+            {
+                Block block = blocks[i];
+                if (startBlock.Address > block.EndAddress + allowance)
+                {
+                    break; // End of contiguous function.
+                }
+
+                startBlock = block;
+                startBlockIndex = i;
+            }
+
+            if (startBlockIndex == 0 && endBlockIndex == blocks.Count - 1)
+            {
+                return; // Nothing to do here.
+            }
+
+            // Replace all branches to blocks outside the range with null, and force a tail call.
+
+            for (int i = startBlockIndex; i <= endBlockIndex; i++)
+            {
+                Block block = blocks[i];
+                if (block.Branch != null && (block.Branch.Address > endBlock.EndAddress || block.Branch.EndAddress < startBlock.Address))
+                {
+                    block.Branch = null;
+                    block.TailCall = true;
+                }
+            }
+
+            // Finally, delete all blocks outside the contiguous range.
+
+            blocks.RemoveRange(endBlockIndex + 1, (blocks.Count - endBlockIndex) - 1);
+            blocks.RemoveRange(0, startBlockIndex);
+        }
+    }
+}
--- a/ARMeilleure/Instructions/InstEmitFlowHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitFlowHelper.cs
@@ -152,7 +152,7 @@ namespace ARMeilleure.Instructions
             if (isJump)
             {
                 context.Tailcall(funcAddr, context.LoadArgument(OperandType.I64, 0));
-            } 
+            }
             else
             {
                 returnAddress = context.Call(funcAddr, OperandType.I64, context.LoadArgument(OperandType.I64, 0));
@@ -192,21 +192,19 @@ namespace ARMeilleure.Instructions
             // correct, if it isn't we keep returning until we reach the dispatcher.
             Operand nextAddr = Const(GetNextOpAddress(context.CurrOp));
 
-            if (context.CurrBlock.Next != null)
+            // Try to continue within this block.
+            // If the return address isn't to our next instruction, we need to return to the JIT can figure out what to do.
+            Operand lblContinue = Label();
+
+            context.BranchIfTrue(lblContinue, context.ICompareEqual(context.BitwiseAnd(returnAddress, Const(~1L)), nextAddr));
+
+            context.Return(returnAddress);
+
+            context.MarkLabel(lblContinue);
+
+            if (context.CurrBlock.Next == null)
             {
-                // Try to continue within this block.
-                // If the return address isn't to our next instruction, we need to return to the JIT can figure out what to do.
-                Operand lblContinue = Label();
-
-                context.BranchIfTrue(lblContinue, context.ICompareEqual(context.BitwiseAnd(returnAddress, Const(~1L)), nextAddr));
-
-                context.Return(returnAddress);
-
-                context.MarkLabel(lblContinue);
-            }
-            else
-            {
-                // No code following this instruction, ask the translator with return address and jump to it.
+                // No code following this instruction, try and find the next block and jump to it.
                 EmitTailContinue(context, nextAddr);
             }
         }
--- a/ARMeilleure/Instructions/InstEmitFlowHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitFlowHelper.cs
@@ -142,7 +142,7 @@ namespace ARMeilleure.Instructions
 
         public static void EmitCall(ArmEmitterContext context, ulong immediate)
         {
-            EmitJumpTableCall(context, Const(immediate));
+            EmitJumpTableBranch(context, Const(immediate));
         }
 
         private static void EmitNativeCall(ArmEmitterContext context, Operand funcAddr, bool isJump = false)
@@ -158,16 +158,7 @@ namespace ARMeilleure.Instructions
                 returnAddress = context.Call(funcAddr, OperandType.I64, context.LoadArgument(OperandType.I64, 0));
                 context.LoadFromContext();
 
-                // InstEmitFlowHelper.EmitContinueOrReturnCheck(context, returnAddress);
-
-                // If the return address isn't to our next instruction, we need to return to the JIT can figure out what to do.
-                Operand continueLabel = Label();
-                Operand next = Const(GetNextOpAddress(context.CurrOp));
-                context.BranchIfTrue(continueLabel, context.ICompareEqual(context.BitwiseAnd(returnAddress, Const(~1L)), next));
-
-                context.Return(returnAddress);
-
-                context.MarkLabel(continueLabel);
+                EmitContinueOrReturnCheck(context, returnAddress);
             }
         }
 
@@ -189,31 +180,34 @@ namespace ARMeilleure.Instructions
             }
             else
             {
-                EmitJumpTableCall(context, target, isJump);
+                EmitJumpTableBranch(context, target, isJump);
             }
         }
 
-        public static void EmitContinueOrReturnCheck(ArmEmitterContext context, Operand retVal)
+        public static void EmitContinueOrReturnCheck(ArmEmitterContext context, Operand returnAddress)
         {
             // Note: The return value of the called method will be placed
             // at the Stack, the return value is always a Int64 with the
             // return address of the function. We check if the address is
             // correct, if it isn't we keep returning until we reach the dispatcher.
-            ulong nextAddr = GetNextOpAddress(context.CurrOp);
+            Operand nextAddr = Const(GetNextOpAddress(context.CurrOp));
 
             if (context.CurrBlock.Next != null)
             {
+                // Try to continue within this block.
+                // If the return address isn't to our next instruction, we need to return to the JIT can figure out what to do.
                 Operand lblContinue = Label();
 
-                context.BranchIfTrue(lblContinue, context.ICompareEqual(retVal, Const(nextAddr)));
+                context.BranchIfTrue(lblContinue, context.ICompareEqual(context.BitwiseAnd(returnAddress, Const(~1L)), nextAddr));
 
-                context.Return(Const(nextAddr));
+                context.Return(returnAddress);
 
                 context.MarkLabel(lblContinue);
             }
             else
             {
-                context.Return(Const(nextAddr));
+                // No code following this instruction, ask the translator with return address and jump to it.
+                EmitTailContinue(context, nextAddr);
             }
         }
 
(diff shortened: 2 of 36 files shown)


riperiperi:
I may have completely broken tests, so give me one momento while I look into that.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -142,51 +143,210 @@ Operand Inverse(Operand val)
...
+            else
+            {
+                EmitJumpTableBranch(context, target, isJump);
+            }
         }
 
-        private static void EmitContinueOrReturnCheck(ArmEmitterContext context, Operand retVal)
+        public static void EmitContinueOrReturnCheck(ArmEmitterContext context, Operand returnAddress)
         {
             // Note: The return value of the called method will be placed
```
gdkchan: Would be nice to update this comment, it still mentions things like "placed on the Stack" from the days it was still emiting IL and so had to put values on the IL stack.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -142,51 +143,210 @@ Operand Inverse(Operand val)
...
+        {
+            Operand nativeContextPtr = context.LoadArgument(OperandType.I64, 0);
+            context.Store(context.Add(nativeContextPtr, Const(NativeContext.GetCallAddressOffset())), guestAddress);
+
+            EmitNativeCall(context, nativeContextPtr, funcAddr, isJump);
+        }
+
+        private static void EmitBranchFallback(ArmEmitterContext context, Operand address, bool isJump)
+        {
+            address = context.BitwiseOr(address, Const(address.Type, 1)); // Set call flag.
```
gdkchan: Would be better to use the CallFlag constant here.

ARMeilleure/Translation/EmitterContext.cs
```diff
@@ -143,9 +143,29 @@ public Operand Call(Operand address, OperandType returnType, params Operand[] ca
...
+            Array.Copy(callArgs, 0, args, 1, callArgs.Length);
+
+            Add(Instruction.Tailcall, null, args);
+
+            _needsNewBlock = true;
+        }
+
+        public Operand CompareAndSwap(Operand address, Operand expected, Operand desired)
+        {
+            if (desired.Type == OperandType.V128)
```
gdkchan: I think it would be better to use `Instruction.CompareAndSwap` for all sizes (including 128), the backend can decide what to write based on the operand type.

ARMeilleure/Translation/JumpTable.cs
```diff
@@ -0,0 +1,148 @@
...
+        private ConcurrentDictionary<ulong, LinkedList<int>> _dependants; // TODO: Attach to TranslatedFunction or a wrapper class.
+
+        private ReservedRegion _jumpRegion;
+        private ReservedRegion _dynamicRegion;
+        public IntPtr JumpPointer => _jumpRegion.Pointer;
+        public IntPtr DynamicPointer => _dynamicRegion.Pointer;
+
+        public JumpTable()
+        {
+            _jumpRegion = new ReservedRegion(JumpTableByteSize, 65536);
```
gdkchan: Would be better to use a named constant instead of `65536`, something like `private const int MinTableSizeInBytes = 65536;`.

ARMeilleure/Translation/Translator.cs
```diff
@@ -124,12 +141,14 @@ private TranslatedFunction GetOrTranslate(ulong address, ExecutionMode mode)
...
         private TranslatedFunction Translate(ulong address, ExecutionMode mode, bool highCq)
         {
-            ArmEmitterContext context = new ArmEmitterContext(_memory, Aarch32Mode.User);
+            ArmEmitterContext context = new ArmEmitterContext(_memory, _jumpTable, (long)address, highCq, Aarch32Mode.User);
 
             Logger.StartPass(PassName.Decoding);
 
-            Block[] blocks = highCq
-                ? Decoder.DecodeFunction  (_memory, address, mode)
+            bool alwaysFunctions = true;
```
gdkchan: I guess this was added to allow easily toggling between new and old behavior. Might be worth adding a comment. For this kind of thing I usually add a const at the top of the file to allow it to be easily found.

ARMeilleure/Translation/DirectCallStubs.cs
```diff
@@ -0,0 +1,131 @@
...
+            Operand functionAddr = context.Call(new _U64_U64_U64(NativeInterface.GetIndirectFunctionAddress), address, entryAddress);
+
+            // Call and save the function.
+            EmitCall(context, functionAddr, tailCall);
+
+            ControlFlowGraph cfg = context.GetControlFlowGraph();
+
+            OperandType[] argTypes = new OperandType[]
+            {
+                    OperandType.I64
```
gdkchan: Adjust indentation.

riperiperi:
Added a temporary solution to the game exiting issue mentioned in the PR. One option was to add another variable to the native context to track if the thread is running, but I don't want to jump the gun in case we do scheduling timing differently now that instruction length is even less predictable.

This will make sure threads exit correctly, most of the time. There is a potential race where setting the sync counter to 0 does nothing (counter stays at what it was before, thread could take too long to exit), but we need to find a better way to do this anyways. Synchronization frequency has been tightened as we never enter blockwise segments of code. Essentially this means, check every x functions or loop iterations, before lowcq blocks existed and were worth just as much. Ideally it should be done in a better way, since functions can be anywhere from 1 to 5000 instructions. (maybe based on RDTSC host timer, or an interrupt flag from a scheduler thread)

Will now address feedback.

ARMeilleure/Translation/JumpTable.cs
riperiperi: I'm going to make ReservedRegion take an optional parameter here, as the granularity isn't really important. Note that the minimum table size is actually 0, since it only commits memory when it makes the first `ExpandIfNeeded` call.

riperiperi: The default will use a named constant in ReservedRegion. The reason I chose 65536 is windows, so I'll explain that in a neat comment.

ARMeilleure/CodeGen/Optimizations/TailCallRemover.cs
```diff
@@ -0,0 +1,75 @@
+﻿using ARMeilleure.Decoders;
+using System;
+using System.Collections.Generic;
+
+namespace ARMeilleure.CodeGen.Optimizations
```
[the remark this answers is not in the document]
riperiperi: Moved to `ARMeilleure.Decoders.Optimizations`

ARMeilleure/Instructions/InstEmitFlowHelper.cs
riperiperi: Done, also re-privated this function as it's only used within this helper now.

ARMeilleure/Translation/Translator.cs
riperiperi: Added as a const with a comment.

```csharp
private const bool AlwaysTranslateFunctions = true; // If false, only translates a single block for lowCq.
```

ARMeilleure/Translation/EmitterContext.cs
riperiperi: I kept them separate due to different requirements on the PreAllocator for the input and output registers. Would it be alright to swap in there based on operand type?

gdkchan: Yea it should be fine, you can change the number of operands on a operation too.

riperiperi:
Addressed all comments - CompareAndSwap variants are now under the same IR instruction.

ARMeilleure/Translation/DirectCallStubs.cs
```diff
@@ -0,0 +1,131 @@
...
+                argTypes,
+                OperandType.I64,
+                CompilerOptions.HighCq);
+        }
+
+        /// <summary>
+        /// Generates a stub that is used to find function addresses and add them to an indirect table. 
+        /// Used for indirect calls entries (already claimed) when their jump table does not have the host address yet.
+        /// Takes a NativeContext like a translated guest function, and extracts the target indirect table entry from the NativeContext.
+        /// If the function we find is highCq, the entry in the table us updated to point to that function rather than this stub.
```
gdkchan: Typo: `us updated`.

ARMeilleure/Translation/JumpTable.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        public void RegisterFunction(ulong address, TranslatedFunction func) {
+            address &= ~3UL;
+            _targets.AddOrUpdate(address, func, (key, oldFunc) => func);
+            long funcPtr = func.GetPointer().ToInt64();
+
+            // Update all jump table entries that target this address.
+            LinkedList<int> myDependants;
+            if (_dependants.TryGetValue(address, out myDependants)) {
```
gdkchan: `{` on new line.

ARMeilleure/Translation/JumpTable.cs
```diff
@@ -0,0 +1,148 @@
...
+                }
+            }
+        }
+
+        public int ReserveDynamicEntry(bool isJump)
+        {
+            int entry = Interlocked.Increment(ref _dynTableEnd);
+            if (entry >= DynamicTableSize)
+            {
+                throw new OutOfMemoryException("JIT Dynamic Jump Table Exhausted.");
```
gdkchan: nit: `exhausted` in lower case.

ARMeilleure/Translation/JumpTable.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            return entry;
+        }
+
+        public int ReserveTableEntry(long ownerAddress, long address, bool isJump)
+        {
+            int entry = Interlocked.Increment(ref _tableEnd);
+            if (entry >= JumpTableSize)
+            {
+                throw new OutOfMemoryException("JIT Direct Jump Table Exhausted.");
```
gdkchan: nit: `exhausted` in lower case.

ARMeilleure/State/ExecutionContext.cs
```diff
@@ -126,6 +126,12 @@ internal void OnUndefined(ulong address, int opCode)
             Undefined?.Invoke(this, new InstUndefinedEventArgs(address, opCode));
         }
 
+        public void StopRunning()
```
gdkchan: Now that we have this method, maybe the `Running` property should be internal or private.

ARMeilleure/Translation/JumpTable.cs
```diff
@@ -0,0 +1,148 @@
...
+        // These start out as all 0. When an indirect call is made, it tries to find the guest address on the table.
+
+        // If we get to an empty address, the guestAddress is set to the call that we want.
+
+        // If we get to a guestAddress that matches our own (or we just claimed it), the hostAddress is read.
+        // If it is non-zero, we immediately branch or call the host function.
+        // If it is 0, NativeInterface is called to find the rejited address of the call.
+        // If none is found, the hostAddress entry stays at 0. Otherwise, the new address is placed in the entry.
+
+        // If the table size is exhausted and we didn't find our desired address, we fall back to doing requesting
```
gdkchan: nit: `doing requesting` -> `requesting` maybe?

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -142,51 +143,209 @@ Operand Inverse(Operand val)
...
 
-            if (context.CurrBlock.Next != null)
-            {
-                Operand lblContinue = Label();
+            // Try to continue within this block.
+            // If the return address isn't to our next instruction, we need to return to the JIT can figure out what to do.
+            Operand lblContinue = Label();
 
-                context.BranchIfTrue(lblContinue, context.ICompareEqual(retVal, Const(nextAddr)));
+            context.BranchIfTrue(lblContinue, context.ICompareEqual(context.BitwiseAnd(returnAddress, Const(~1L)), nextAddr));
```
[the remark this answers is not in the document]
riperiperi: Changing to `~CallFlag`

riperiperi:
Addressed Feedback.

PR #976, opened 2020-03-08
Fix GTK window crash by using 24 bit surface on unix, 32 bit on windows.

riperiperi:
Just a quick fix to get linux working again. This shouldn't reintroduce transparency issues, as I don't think the 24 bit surface had any. (we made it 32 bit to combat issues on windows)

### DIFF
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -51,7 +51,7 @@ namespace Ryujinx.Ui
         private Input.NpadController _primaryController;
 
         public GLRenderer(Switch device)
-            : base (new GraphicsMode(new ColorFormat()),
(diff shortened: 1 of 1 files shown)


Ryujinx/Ui/GLRenderer.cs
```diff
@@ -51,7 +51,7 @@ public class GLRenderer : GLWidget
         private Input.NpadController _primaryController;
 
         public GLRenderer(Switch device)
-            : base (new GraphicsMode(new ColorFormat()),
+            : base (new GraphicsMode(Environment.OSVersion.Platform == PlatformID.Unix ? new ColorFormat(24) : new ColorFormat()),
```
jduncanator: I'd prefer this to be broken out into a static `GetGraphicsMode` function that returns the correct graphics mode based on OS, rather than trying to squish it all into one line.

PR #978, opened 2020-03-09
CodeGen Optimisations (LSRA and Translator)

riperiperi:
The focus of these optimisations is mainly to reduce the number of allocations done when generating code. These allocations were greatly affecting the performance of other threads in the program, meaning that lowCq code would run much slower while highCq was being compiled, and multiple highCq threads would completely starve each other no matter how many cores you have.

With this change and direct-call, booting games up has more comparable performance to master.

- Use pools for Operands, Operations and Nodes. The allocations for all of these small objects happened pretty much constantly and likely caused a lot of heap contention.
  - This uses a `ThreadStaticPool`, a new class which pools objects in a big list. These objects are never deleted, as it is assumed that future workloads will use a similar number of objects as previous ones.
  - There is a pool of `ThreadStaticPool`s that each thread takes from before it starts doing work. This is to avoid creating individual pools for like, 30 guest threads. When activated, the selected pool is given to the translator thread as a thread static variable (accessed through a static property) for easy access. highCq has a different pool "group" than lowCq, as it will have different characteristics.
- BitMap now processes 64-bit chunks, and is its own enumerator (creating them was creating a lot of garbage)
- A new `SortedIntegerList` is used in `LiveInterval` instead of `SortedSet`, as it makes it easier to iterate and remove ranges. (eg. the Split function) Worst case performance for sorting is slightly worse, but everything else makes up for it. _(not sure whether or not to move negating complexity to the SortedIntegerList (making it something else entirely) or keeping it in the consumers)_
- Max background translation threads has been increased to 4, as it can handle it.

## Downsides
The GC doesn't seem to like what I'm doing with the pools of small objects and may occasionally freeze the program after large bursts of translation (for a fraction of a second). This is exaggerated the more threads there are. It would be ideal if I could just tell the GC to ignore the pool items or move them to the final generation, but that wouldn't be very C#. 😩

### DIFF
--- a/ARMeilleure/Common/BitUtils.cs
+++ b/ARMeilleure/Common/BitUtils.cs
@@ -1,3 +1,5 @@
+using System.Runtime.CompilerServices;
+
 namespace ARMeilleure.Common
 {
     static class BitUtils
@@ -6,17 +8,29 @@ namespace ARMeilleure.Common
 
         private static readonly int[] DeBrujinLbsLut;
 
+        private const long DeBrujinSequence64 = 0x37e84a99dae458f;
+
+        private static readonly int[] DeBrujinLbsLut64;
+
         private static readonly sbyte[] HbsNibbleLut;
 
         static BitUtils()
         {
             DeBrujinLbsLut = new int[32];
+            DeBrujinLbsLut64 = new int[64];
 
             for (int index = 0; index < DeBrujinLbsLut.Length; index++)
             {
                 uint lutIndex = (uint)(DeBrujinSequence * (1 << index)) >> 27;
 
                 DeBrujinLbsLut[lutIndex] = index;
+            }
+
+            for (int index = 0; index < DeBrujinLbsLut64.Length; index++)
+            {
+                ulong lutIndex = (ulong)(DeBrujinSequence64 * (1L << index)) >> 58;
+
+                DeBrujinLbsLut64[lutIndex] = index;
             }
 
             HbsNibbleLut = new sbyte[] { -1, 0, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3 };
@@ -64,6 +78,7 @@ namespace ARMeilleure.Common
             return HbsNibbleLut[value];
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int LowestBitSet(int value)
         {
             if (value == 0)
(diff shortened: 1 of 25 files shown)


ARMeilleure/IntermediateRepresentation/Operand.cs
```diff
@@ -26,42 +26,64 @@ public Operand(OperandKind kind, OperandType type = OperandType.None) : this()
...
+        {
+            With(OperandKind.Constant, OperandType.I32);
             Value = (uint)value;
+            return this;
         }
 
-        public Operand(uint value) : this(OperandKind.Constant, OperandType.I32)
+        public Operand With(uint value)
         {
+            With(OperandKind.Constant, OperandType.I32);
```
gdkchan: I think you could cut down one line on each one of those methods by doing `return With(OperandKind.Constant, OperandType.I32);` etc.

ARMeilleure/ThreadStaticPool.cs
```diff
@@ -0,0 +1,104 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace ARMeilleure
+{
+    internal class ThreadStaticPool<T> where T : class, new()
```
gdkchan: I think `ARMeilleure.Common` is a better place for that.

ARMeilleure/ThreadStaticPool.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        private static readonly int PoolSizeIncrement = 200;
```
gdkchan: Why not `private const`?

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -11,39 +12,78 @@ public Operand Destination
...
 
-        public int DestinationsCount => _destinations.Length;
-        public int SourcesCount      => _sources.Length;
+        public int DestinationsCount => _destinations.Count;
+        public int SourcesCount      => _sources.Count;
 
-        public Node(Operand destination, int sourcesCount)
+        private void Resize(List<Operand> list, int size)
+        {
+            while (list.Count > size)
```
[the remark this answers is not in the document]
riperiperi: Just forgot to change it.

ARMeilleure/IntermediateRepresentation/Operand.cs
riperiperi: Value is set to 0 in the shared `With()` method. I'll change it to take an optional value parameter.

ARMeilleure/Common/BitUtils.cs
```diff
@@ -76,6 +91,19 @@ public static int LowestBitSet(int value)
             return DeBrujinLbsLut[(uint)(DeBrujinSequence * lsb) >> 27];
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int LowestBitSet(long value)
```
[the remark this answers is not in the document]
riperiperi: Will replace all instances, and make HighestBitSet use `31 - LeadingZeroCount`.

riperiperi:
Feedback addressed. I think I prefer inserting negative values into a simple sorted list type rather than creating some complex abstraction to insert and iterate them backwards, so I'm fine with it the way it is.

PR #982, opened 2020-03-11
Implement AESMC, AESIMC, AESE, AESD and VEOR AArch32 instructions

riperiperi:
Adds instructions required for games to perform cryptography, and tests for them. Currently, unicorn doesn't seem to be able to do the crypto instructions in A32, so we're just comparing against expected values for now.

Tested in Mario Kart 8 Deluxe, which uses this for multiplayer comms.

Also adds a test for VORR (non-immediate), since it was free.

### DIFF
--- a/Ryujinx.Tests/Cpu/CpuTest32.cs
+++ b/Ryujinx.Tests/Cpu/CpuTest32.cs
@@ -164,11 +164,11 @@ namespace Ryujinx.Tests.Cpu
             }
         }
 
-        protected void ExecuteOpcodes()
+        protected void ExecuteOpcodes(bool runUnicorn = true)
         {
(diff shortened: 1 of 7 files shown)


Ryujinx.Tests/Cpu/CpuTest32.cs
```diff
@@ -164,11 +164,11 @@ protected void Opcode(uint opcode)
             }
         }
 
-        protected void ExecuteOpcodes()
+        protected void ExecuteOpcodes(bool runUnicorn = true)
```
jduncanator: I think this would be better done as an attribute on the specific test case, something like `[UnicornOptions(Enabled = false)]`.

It would allow us to optionally have additional Unicorn settings in the future as well.

PR #983, opened 2020-03-12
Update window title to include title architecture

Cyuubi:
Also clarifies the GUI window a bit more, just like the Console window.
image

Additionally, it removes the 32-bit application warning because the titlebar shows what architecture the application is now.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Renderer.cs
+++ b/Ryujinx.Graphics.OpenGL/Renderer.cs
@@ -87,6 +87,8 @@ namespace Ryujinx.Graphics.OpenGL
             GpuVendor   = GL.GetString(StringName.Vendor);
             GpuRenderer = GL.GetString(StringName.Renderer);
             GpuVersion  = GL.GetString(StringName.Version);
+
+            Logger.PrintInfo(LogClass.Gpu, $"{GpuVendor} {GpuRenderer} ({GpuVersion})");
         }
 
         public void ResetCounter(CounterType type)
--- a/Ryujinx.Graphics.OpenGL/Renderer.cs
+++ b/Ryujinx.Graphics.OpenGL/Renderer.cs
@@ -19,6 +19,12 @@ namespace Ryujinx.Graphics.OpenGL
 
         internal TextureCopy TextureCopy { get; }
 
+        public string GpuVendor { get; private set; }
+
+        public string GpuRenderer { get; private set; }
+
+        public string GpuVersion { get; private set; }
+
         public Renderer()
         {
             _pipeline = new Pipeline();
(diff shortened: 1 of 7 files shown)


jduncanator:
I don't mind the changes to the game title, title ID and architecture, but I prefer how the Console and GUI titles were before (eg. Ryujinx 1.0.0-dirty and Ryujinx Console).

Cyuubi:
I'll see what others think, dunno right now.

Xpl0itR:
The title architecture in the title bar is a nice addition imo. That being said, I think the other changes to the title bar look worse and should be reverted

riperiperi:
Agree with Xpl0itR

Cyuubi:
I made a poll in the Discord server. (https://discordapp.com/channels/410208534861447168/410432399500115968/687454374360121346)

gdkchan:
My opinion:
- Having the information on the log is still useful, in case we receive a log (due to a user hiting unimpl. instruction for example) without a screenshot, we need to know if the game is 32-bits or 64-bits to properly decode the instruction.
- I think it would be better to put it after the game name like this for example: `Ryujinx | New Super Mario Bros U Deluxe (32-bit)`
- For the other changes, I prefer the way it is right now (on master).

gdkchan:
Guess it's also worth noting that puting the information on the title could also lead to confusion with people thinking they are running a 32-bits build of the emulator.

Ryujinx.Graphics.OpenGL/Renderer.cs
```diff
@@ -71,18 +77,18 @@ public ulong GetCounter(CounterType type)
...
         public void Initialize()
         {
-            PrintGpuInformation();
+            GetGpuInformation();
 
             _counters.Initialize();
         }
 
-        private void PrintGpuInformation()
+        private void GetGpuInformation()
```
AcK77: All those change can be removed because of #984 and you don't use it anymore btw.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -117,6 +117,8 @@ public class Horizon : IDisposable
         public ulong  TitleId { get; private set; }
         public string TitleIdText => TitleId.ToString("x16");
 
+        public string TitleArchitecture { get; private set; }
```
AcK77: This could be a bool and then you can use this bool to write the right text.
Named `TitleIs64Bit`

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -178,13 +178,13 @@ public void Start()
             {
                 parent.Present();
 
-                string titleNameSection = string.IsNullOrWhiteSpace(_device.System.TitleName) ? string.Empty
-                    : " | " + _device.System.TitleName;
+                string titleNameSection = string.IsNullOrWhiteSpace(_device.System.TitleName) ? " - Application"
```
AcK77: I don't think `" - Application"` is necessary

Xpl0itR:
Superseded by a commit in #1023

PR #984, opened 2020-03-12
Add GPU name in status bar

HorrorTroll:
Here some proof:
E6F52DA7-D46E-4906-B6D9-30A4BAEB4416

Still needed more testing on another Nvidia, AMD or Intel GPU anyway!

### DIFF
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -328,7 +328,8 @@ namespace Ryujinx.Ui
                     StatusUpdatedEvent?.Invoke(this, new StatusUpdatedEventArgs(
                         _device.EnableDeviceVsync, 
                         $"Host: {_device.Statistics.GetSystemFrameRate():00.00} FPS", 
-                        $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS"));
+                        $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS",
+                        $"GPU: {_renderer.GpuVendor.Trim().Split(' ')[0]} {_renderer.GpuRenderer.Trim().Split('/')[0]}"));
 
                     _device.System.SignalVsync();
 
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -329,7 +329,7 @@ namespace Ryujinx.Ui
                         _device.EnableDeviceVsync, 
                         $"Host: {_device.Statistics.GetSystemFrameRate():00.00} FPS", 
                         $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS",
-                        $"GPU: {_renderer.GpuVendor.Trim().Split(' ')[0]} {_renderer.GpuRenderer.Trim().Split('/')[0]}"));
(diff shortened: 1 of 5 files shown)


9338088:
real nice love the idea =)

jduncanator:
I feel like this one should be right aligned, next to the System Version.

AcK77:
I think the GPU informations should be parsed just to write the vendor name (NVidia / AMD / Intel / Other), we don’t need the whole thing in the status bar.

HorrorTroll:
Like this?
0AE0F399-B16F-40BA-882F-681F55414217

AcK77:
Yeah I think it’s better than the long GPU text.

gdkchan:
I think if we are going to do that its better to show the GPU model etc, not just the vendor.

HorrorTroll:
i think i prefer with only Vendor, since we still print the GPU on Log already.
DA243FD7-5B71-4743-9A76-8F545C9E5A89

HorrorTroll:
image
image

Changed to what Ac_K suggest, and yeah it look better. Since long GPU name still print on log

GamerzHell9137:
Why not add an option to have it full and short name of the GPU?
For troubleshooting issues when users come to the discord and ask for support, if it was just shown in the status bar it would take less to help them with their issue (If the GPU is a GT series we could outright say its too slow for Ryujinx). Most of the time people don't know what their GPU is so going trough hoops to find out what they are using is gonna be time consuming. I know its pretty optional but it would still be beneficial, i don't see the issue of having the full name on default either, its not eye punching at all, to me at least.

HorrorTroll:
But u still have a long GPU name in Log anyway? And we alway have to send log for support in discord. Not just image about why Ryujinx not working

komawoyo:
Will we be seeing a PR for CPU in status bar too?

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -328,7 +328,8 @@ public void Render()
                     StatusUpdatedEvent?.Invoke(this, new StatusUpdatedEventArgs(
                         _device.EnableDeviceVsync, 
                         $"Host: {_device.Statistics.GetSystemFrameRate():00.00} FPS", 
-                        $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS"));
+                        $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS",
+                        $"GPU: {_renderer.GpuVendor.Trim().Split(' ')[0]}"));
```
AcK77: Instead of trim and split it, you should use `string.Contains()` (in lowercase) to check if the GPU string contains `nvidia`, `amd`, `intel` and then write `NVIDIA`, `AMD`, `Intel` or `Unknown` depend to the result.

HorrorTroll:
Updated to Ac_K suggest. Need review again!

Ryujinx/Ui/GLRenderer.cs
HorrorTroll: Fixed.

HorrorTroll:
i think it good to go now

LePwnerer:
Not good to go yet, check for amd or ati since for example the AMD Radeon RX 5700XT has "ati technologies inc." as vendor

PR #986, opened 2020-03-12
Add Legacy Attributes gl_Color and gl_TexCoord

riperiperi:
Used by Xash3D. (also needs an implementation of Ushl_S to get into the campaign). The game's skinned models don't work yet, but that's likely an issue with it doing really quick back to back immediate draws of generated data (that are quickly replaced with index/vertex data for the next draw).

image

Draft PR because I don't know what adverse affects this might have on some games/gpus. There doesn't seem to be an issue on what I have tested. A better solution might be to make use of these attributes trigger a flag in the context which changes the shader between `core` and `compatibility` dynamically, but I wasn't aware of any appropriate location to set such a flag, and the context is inaccessible when getting a variable name.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -15,7 +15,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
 
         public static void Declare(CodeGenContext context, StructuredProgramInfo info)
         {
-            context.AppendLine("#version 420 core");
+            context.AppendLine("#version 420 compatibility");
             context.AppendLine("#extension GL_ARB_gpu_shader_int64 : enable");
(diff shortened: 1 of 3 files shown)


jduncanator:
What's the status on this one? Is it likely to be something we'll want to merge, or are we better closing this?

riperiperi:
> 
> 
> What's the status on this one? Is it likely to be something we'll want to merge, or are we better closing this?

With resolution scale, I introduced a flag system for features that shaders use. That can be used to optionally emit the compatibility shader header rather than the core one.

riperiperi:
Closing til later. It doesn't look like retail games are using this, at least.

gdkchan:
Someone said it fixes the black screen on Dragon Quest 3.

riperiperi:
Oh.

vonchenplus:
Yes, it fixes the black screen on Dragon Quest 3.

PR #988, opened 2020-03-13
Set the minimum width of the window to 854px

dnmodder:
This allows the window size to fit perfectly on monitors that are less than 1280px wide, like 1024x768 in my case.

**Before:**
Screenshot_20200313_114755

**After:**
Screenshot_20200313_115004


Xpl0itR:
Superseded by #1046

PR #995, opened 2020-03-18
Modify TranslatedFunction.GetPointer () to optimize performance

Chenj168:
Add local var to reduce calling Marshal.GetFunctionPointerForDelegate, Can slightly improve performance.

### DIFF
--- a/ARMeilleure/Translation/TranslatedFunction.cs
+++ b/ARMeilleure/Translation/TranslatedFunction.cs
@@ -9,6 +9,7 @@ namespace ARMeilleure.Translation
         private const int MinCallsForRejit = 100;
 
         private GuestFunction _func;
+        private IntPtr _funcPtr;
 
         private bool _rejit;
         private int  _callCount;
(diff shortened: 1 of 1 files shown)


riperiperi:
LGTM, might be worth measuring any change

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -33,7 +34,9 @@ public bool ShouldRejit()
 
         public IntPtr GetPointer()
         {
-            return Marshal.GetFunctionPointerForDelegate(_func);
+            if (_funcPtr == IntPtr.Zero)
```
jduncanator: To follow the style of the remainder of the code, this if clause should be enclosed.

```cs
if (_funcPtr == IntPtr.Zero) {
    _funcPtr = Marshal.GetFunctionPointerForDelegate(_func);
}
```

gdkchan: With `{` on a separate line.

ARMeilleure/Translation/TranslatedFunction.cs
```diff
@@ -33,7 +34,9 @@ public bool ShouldRejit()
...
+                _funcPtr = Marshal.GetFunctionPointerForDelegate(_func);
+            return _funcPtr;
```
AcK77: Since we are on it, don’t forget to add a new line before a return!

ARMeilleure/Translation/TranslatedFunction.cs
jduncanator: Updated! I've been doing too much JavaScript 🤦‍♂

PR #1000, opened 2020-03-19
Support partial invalidation on texture access

gdkchan:
This changes the way how texture invalidation works a bit to allow partial invalidation of textures. When the texture is modified by the GPU, and modification of the data by guest from the CPU side is detected, instead of re-loading the whole texture and potentially losing modifications made by the host GPU, instead it just copies the specific pages that were detected as modified. This prevents destruction of GPU written data, even if it's not a perfect solution (since it only copies whole pages at once).

In terms of performance I don't expect any noticeable regressions, it will be slower to re-upload textures modified by the host GPU, but re-loading textures from the guest memory is already a somewhat slow operation, so I guess for the part where it stutters due to texture load it could get worse.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -321,20 +321,25 @@ namespace Ryujinx.Graphics.Gpu.Image
             {
                 Span<byte> gpuData = GetTextureDataFromGpu();
 
+                ulong endAddress = Address + Size;
+
                 for (int i = 0; i < modifiedRanges.Length; i++)
                 {
                     (ulong modifiedAddress, ulong modifiedSize) = modifiedRanges[i];
 
+                    ulong endModifiedAddress = modifiedAddress + modifiedSize;
+
                     if (modifiedAddress < Address)
                     {
-                        modifiedSize -= (Address - modifiedAddress);
                         modifiedAddress = Address;
                     }
 
-                    if (modifiedSize > Size)
+                    if (endModifiedAddress > endAddress)
                     {
-                        modifiedSize = Size;
+                        endModifiedAddress = endAddress;
                     }
+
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -554,6 +561,13 @@ public Texture FindOrCreateTexture(TextureInfo info, TextureSearchFlags flags =
 
                         overlap.HostTexture.CopyTo(newView, 0, 0);
 
+                        // Inheritc modification from overlapping texture, do that before replacing
```
riperiperi: `Inherit`?

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -302,15 +302,63 @@ public void SynchronizeMemory()
...
+            // Note that if ASTC is not supported by the GPU we can't read it back since
+            // it will use a different format. Since applications shouldn't be writing
+            // ASTC textures from the GPU anyway, ignoring it should be safe.
+            if (_context.Methods.TextureManager.IsTextureModified(this) && !Info.FormatInfo.Format.IsAstc())
+            {
+                Span<byte> gpuData = GetTextureDataFromGpu();
+
+                for (int i = 0; i < modifiedRanges.Length; i++)
+                {
+                    (ulong address, ulong size) = modifiedRanges[i];
```
jduncanator: I'd suggest renaming these variables something else, the following code is very hard to follow given they're both called "address" and "size".

gdkchan: How it should be called then? `modifiedAddress` and `modifiedSize`?

jduncanator: Yea that'd be fine, just something to easily tell the difference between the two.

PR #996, opened 2020-03-19
Use OpCodeTable to get MakeOp to improve performance

Chenj168:
In original code, to create an opcode instance,
use the OpCodeTable to search type of opcode, after that, 
use ConcurrentDictionary<Type, MakeOp> to search MakeOp of type for create instance,
this code flow can be optimized.

Before:
OpCodeTable.GetInstA64(opcode) 
  -> get type 
     -> _opActivators.GetOrAdd(type) 
        -> get MakeOp
           -> create instance

After:
OpCodeTable.GetInstA64(opcode) 
  -> get MakeOp
     -> create instance

ConcurrentDictionary is no longer needed.

PR #1001, opened 2020-03-20
Move the OpActivator to OpCodeTable class for improve performance

Chenj168:
This modification similar to [PR996](https://github.com/Ryujinx/Ryujinx/pull/996) and seems to work well
Thanks gdkchan for the proposal.


Ryujinx.Graphics.Shader/Decoders/Decoder.cs
```diff
@@ -256,7 +245,12 @@ private static bool BinarySearch(List<Block> blocks, ulong address, out int inde
                     continue;
                 }
 
-                OpCode op = MakeOpCode(opCodeType, emitter, opAddress, opCode);
+                if(opActivator == null)
```
riperiperi: Space between `if` and `(`

PR #1002, opened 2020-03-21
Add warning with an option to disable for settings that impact performance

Xpl0itR:
Untitled


jduncanator:
I think the text for both should be along the lines of:

"You have debug logging enabled, which is designed to be used by developers only. For optimal performance, it's recommended to disable debug logging. Would you like to disable debug logging now?"

PR #1008, opened 2020-03-22
Load default config when an invalid config is found

Xpl0itR:
I've also changed the code that write the config to disk to use the `FileOptions.WriteThrough` flag to write to disk immediately to decrease the chances of corruption

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
@@ -227,9 +227,20 @@ namespace Ryujinx.Configuration
         /// Loads a configuration file from disk
         /// </summary>
         /// <param name="path">The path to the JSON configuration file</param>
-        public static ConfigurationFileFormat Load(string path)
+        public static bool TryLoad(string path, out ConfigurationFileFormat configurationFileFormat)
         {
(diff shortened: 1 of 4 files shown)


Xpl0itR:
> I also think it needs a very obvious log message indicating that the default config was used and not the users config, as I imagine people will pull their hair out if the emulator just blindly uses a different config file than the one they're expecting, and can't figure out why.

Added a log message
> I think in the case that the config file doesn't exist, or it can't be loaded, the default one should only be used temporarily, and should not overwrite the users existing config file.

I've changed my code so that after loading the default config, it is not saved on disk straight away. However it is still possible for users to trigger the config to be saved on disk by modifying and saving it in the GUI.

Xpl0itR:
Rebased and partially rewritten

Thog:
This is still needed today, could be nice to rebase @Xpl0itR

Xpl0itR:
Rebased

riperiperi (approved):
LGTM.

Thog:
Need a final rebase but lgtm

Thog (approved):

Xpl0itR:
@Thog rebased again

AcK77 (approved):

PR #1012, opened 2020-03-23
Initial conditional rendering support

gdkchan:
This adds some support for conditional rendering. Conditonal rendering can be used to disable clears and draws based on some condition. When it is used, most of the time it is used for occlusion queries.

Currently it is reading the counter values from memory. For that to work, report calls needs to always write the value in memory. This is very inefficient as each time report is called, it is necessary to wait until the previous commands have completed in order to read the counter value from the GPU, and then wait it in memory.

A possible solution for the above problem is host conditional rendering (that is, using OpenGL conditional rendering functions directly, and passing the query). I added a "counter cache" that should help making this possible, but there's still some work required to implement that. And this does not solve the pre-existing issue -- counter values read from the CPU side. It still needs the "always write" for that to work. So until this issue is solved, I don't think it's worth to try implementing host conditional rendering (as it would have no performance benefit if the counter is still written to memory all the time).

I recommend testing on a number of games to ensure it does not regress anything.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodConditionalRendering.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodConditionalRendering.cs
@@ -1,0 +1,29 @@
+﻿using Ryujinx.Common.Logging;
+using Ryujinx.Graphics.Gpu.State;
+
+namespace Ryujinx.Graphics.Gpu.Engine
+{
+    partial class Methods
+    {
+        private bool GetRenderEnable(GpuState state)
+        {
+            ConditionState condState = state.Get<ConditionState>(MethodOffset.ConditionState);
+
+            switch (condState.Condition)
+            {
+                case Condition.Always:
+                    return true;
+                case Condition.Never:
+                    return false;
+                case Condition.ResultNonZero:
+                case Condition.Equal:
+                case Condition.NotEqual:
+                    return false; // TODO (should we use the host API?)
+            }
+
+            Logger.PrintWarning(LogClass.Gpu, $"Invalid conditional render condition \"{condState.Condition}\".");
+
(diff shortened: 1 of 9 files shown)


jduncanator:
Look's like MK8D uses some of the additional conditions, might be a useful test for the validity and performance when/if you get around to implementing those.

image

ReaperOfSouls1909:
Needs a rebase to merge with master to test the AMD fix

FernandoS27:
Well to be honest this isn't much of an implementation. Here is how you should do it:

get Condition address and read & flush an array of 6 unsigned ints 32 bits, we will call it cond_queries

for ResultNonZero, you check:
`cond_queries[0] != 0 && cond_queries[1] != 0`

for Equal:
`cond_queries[0] == cond_queries[4] && cond_queries[1] == cond_queries[5]`

for NotEqual:
`cond_queries[0] != cond_queries[4] || cond_queries[1] != cond_queries[5]`

Ryujinx.Graphics.Gpu/Engine/MethodConditionalRendering.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            if (!FindAndFlush(gpuVa) && !FindAndFlush(gpuVa + 16))
+            {
+                return false;
+            }
+
+            ulong x = _context.MemoryAccessor.ReadUInt64(gpuVa);
+            ulong y = _context.MemoryAccessor.ReadUInt64(gpuVa + 16);
+
+            return cond == Condition.Equal ? x == y : x != y;
```
riperiperi, line 67: A bit weird passing the full condition enum here when the implication is that it can only be Equal or NonEqual.

PR #1015, opened 2020-03-23
Index constant buffer vec4s using ternary expressions.

riperiperi:
Should help text rendering in most games on AMD gpus. Need testers.

Draft until we decide the best course of action. (detect amd and switch on/off, or use the slow method at all times / another)

Thanks to Rodrigo for suggesting that this might be the issue.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
@@ -114,6 +114,24 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
             return ubName + "." + GetSwizzleMask(offset & 3);
         }
 
+        private static string GetVec4Indexed(string vectorName, string indexExpr)
+        {
+            bool canIndexVec4 = false;
+            if (canIndexVec4)
+            {
+                return $"{vectorName}[{indexExpr}]";
+            } 
+            else
+            {
+                string result = $"{vectorName}.x";
+                for (int i = 1; i < 4; i++)
+                {
+                    result = $"(({indexExpr}) == {i}) ? ({vectorName}.{GetSwizzleMask(i)}) : ({result})";
+                }
+                return $"({result})";
+            }
+        }
+
         public static string GetConstantBufferName(IAstNode slot, string offsetExpr, ShaderStage stage)
         {
             // Non-constant slots are not supported.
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
```diff
@@ -114,6 +114,24 @@ public static string GetConstantBufferName(int slot, int offset, ShaderStage sta
             return ubName + "." + GetSwizzleMask(offset & 3);
         }
 
+        private static string GetVec4Indexed(string vectorName, string indexExpr)
+        {
+            bool canIndexVec4 = false;
+            if (canIndexVec4)
```
jduncanator: If I was pedantic, I'd say there should be a new line between the declaration and the statement...

jduncanator: From extensive testing, it doesn't seem like this negatively affects performance on NVIDIA GPUs. This can probably be on by default

riperiperi:
Now on by default, removed `[]` indexed path in favor of always using the more compatible one.

Cristallix:
A bug report has been made to AMD to fix this issue ?

Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
gdkchan: I would still prefer to have the indexed path, since the code is easier to read and understand, compared to a bunch of ternary expressions.

riperiperi: So do we want it to still exist, just disabled for now?

gdkchan: I think it could be still used for Intel and NVIDIA since it works fine there.

gdkchan: Let's just keep it as-is for now, since there's currently no easy way to check vendor from the shader translator.

riperiperi: The nvidia branch was just removed - I will revert that commit.

gdkchan: It's fine, just keep it like that. If we decide to add it back later, we can re-add the code.

riperiperi: oops 😂

riperiperi: Don't worry I double reverted it.

PR #1029, opened 2020-03-24
Discord Rich Presence update

Xpl0itR:
- Fix bug when ryujinx is run with an invalid parameter
- Add a few more titles
- Move the list from dat file into the code

We've used 60/150 slots in the discord app, so we have plenty of space to add more.


Ryujinx/Configuration/DiscordIntegrationModule.cs
```diff
@@ -99,12 +98,75 @@ public static void SwitchToMainMenu()
...
+            _discordClient?.SetPresence(DiscordPresence);
         }
 
         public static void Exit()
         {
-            DiscordClient?.Dispose();
+            _discordClient?.Dispose();
         }
+
+        private static readonly string[] SupportedTitles =
```
AcK77: Could be nice to add the title name as a comment near the title id. If we want to change some of them it will be easier to find one.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -359,7 +359,8 @@ internal void LoadApplication(string path)
                 else
                 {
                     Logger.PrintWarning(LogClass.Application, "Please specify a valid XCI/NCA/NSP/PFS0/NRO file.");
-                    End(device);
+                    device.Dispose();
+                    return;
```
AcK77: add a new empty line before

PR #1030, opened 2020-03-24
Fix gpu vendor name parsing

AcK77:
This fix an issue who write the wrong GPU vendor in status bar because of the "ati" in "Corporation".


gdkchan:
To be honest I think this method to identify the GPU is pretty bad. Why don't you just print the full vendor name that OpenGL returns?

AcK77:
I've change it to print the full vendor name ;)

PR #1034, opened 2020-03-25
prepo: Add a MessagePack object formatter

jduncanator:
This adds a user readable formatter for MessagePack objects. It's not the prettiest thing itself, but it works.

This will need significant testing on games that make heavy use of prepo, although I suspect the worst case scenario to be objects that just have their types printed in the output (eg. `System.Byte[]`).

### DIFF
--- a/Ryujinx.Common/Utilities/MessagePackObjectFormatter.cs
+++ b/Ryujinx.Common/Utilities/MessagePackObjectFormatter.cs
@@ -1,0 +1,302 @@
+﻿using MsgPack;
+using System;
+using System.Text;
+
+namespace Ryujinx.Common.Utilities
+{
+    public static class MessagePackObjectFormatter
+    {
+        public static string ToString(this MessagePackObject obj, bool pretty)
+        {
+            if (pretty)
+            {
+                return Format(obj);
+            }
+            else
+            {
+                return obj.ToString();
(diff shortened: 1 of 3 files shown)


Ryujinx.Common/Utilities/MessagePackObjectFormatter.cs
```diff
@@ -0,0 +1,302 @@
...
+                builder.Append(", ");
+            }
+
+            // Remove trailing comma
+            builder.Remove(builder.Length - 2, 2);
+
+            builder.Append(" ]");
+        }
+
+        private static char ToHexChar(int b)
```
gdkchan: You could just use `b.ToString("X1")` instead of this method, but I guess this might be more efficient, so Im fine with either one.

Ryujinx.Common/Utilities/MessagePackObjectFormatter.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                return unchecked((char)('0' + b));
+            }
+            else
+            {
+                return unchecked((char)('A' + (b - 10)));
+            }
+        }
+
+        internal class IndentedStringBuilder
```
gdkchan: `internal` not required.

PR #1041, opened 2020-03-25
Implement NOP and stub DEPBAR shader instructions

### LINKED ISSUE #1038: Implement DEPBAR and NOP shader instructions
Those instruction should be just a no-op, we don't need to translate them to anything on glsl. However they should be still recognized to avoid printing the "Invalid instruction" warnings on the log.

EliseZeroTwo:
Should fix #1038

### DIFF
--- a/Ryujinx.Graphics.Shader/Decoders/OpCodeTable.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/OpCodeTable.cs
@@ -162,7 +162,7 @@ namespace Ryujinx.Graphics.Shader.Decoders
             Set("000000010000xx", InstEmit.Mov,     typeof(OpCodeAluImm32));
             Set("0101110010011x", InstEmit.Mov,     typeof(OpCodeAluReg));
             Set("0101000010000x", InstEmit.Mufu,    typeof(OpCodeFArith));
-            Set("1010000101100x", InstEmit.Nop,     typeof(OpCode));
+            Set("0101000010110x", InstEmit.Nop,     typeof(OpCode));
             Set("1111101111100x", InstEmit.Out,     typeof(OpCode));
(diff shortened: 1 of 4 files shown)


gdkchan:
I don't think a new OpCode* type is necessary for NOP since it does nothing (you can just use the OpCode type there). Same for DEPBAR, although this one does have some parameters that we could decode, but since we aren't going to do anything with them, I'm fine with just doing the same as NOP.

Ryujinx.Graphics.Shader/Decoders/OpCodeTable.cs
```diff
@@ -161,6 +162,7 @@ static OpCodeTable()
             Set("000000010000xx", InstEmit.Mov,     typeof(OpCodeAluImm32));
             Set("0101110010011x", InstEmit.Mov,     typeof(OpCodeAluReg));
             Set("0101000010000x", InstEmit.Mufu,    typeof(OpCodeFArith));
+            Set("1010000101100x", InstEmit.Nop,     typeof(OpCode));
```
gdkchan: The encoding for this instruction seems incorrect, I think it should start with `0101` instead of `1010`.

EliseZeroTwo: image
I think it is correct, here is what I used

gdkchan: I think your output is probably not displaying the leftmost 0 bits, thats why the line for nop is shorter than the others.

EliseZeroTwo: Oh god, I'm sorry

EliseZeroTwo:
@gdkchan fixed

PR #1043, opened 2020-03-25
Make max anisotropy configurable

Xpl0itR:
This PR exposes a config option to make the max anisotropy configurable. The value can be 0-16 or -1 to let the game decide the level of anisotropy. The value is set to -1 by default.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Sampler.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Sampler.cs
@@ -1,4 +1,3 @@
-using OpenTK.Graphics.OpenGL;
 using Ryujinx.Graphics.GAL;
 using System;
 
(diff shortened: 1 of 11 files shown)


Ryujinx.Graphics.Gpu/Image/Sampler.cs
```diff
@@ -40,7 +41,11 @@ public Sampler(GpuContext context, SamplerDescriptor descriptor)
             float maxLod     = descriptor.UnpackMaxLod();
             float mipLodBias = descriptor.UnpackMipLodBias();
 
-            float maxAnisotropy = descriptor.UnpackMaxAnisotropy();
+            float maxRequestedAnisotropy = GraphicsConfig.MaxAnisotropy >= 0 && GraphicsConfig.MaxAnisotropy <= 16 ? GraphicsConfig.MaxAnisotropy : descriptor.UnpackMaxAnisotropy();
+            float maxSupportedAnisotropy = GL.GetFloat((GetPName)All.MaxTextureMaxAnisotropy);
```
gdkchan: This should not call OpenGL directly, this is supposed to be handled on the OpenGL backend (Ryujinx.Graphics.OpenGL), the maximum anisotropy can be returned on the `GpuCapabilities` struct from the `GetCapabilities` method on `Renderer`.

Xpl0itR: 👍

Xpl0itR:
Since you all wanted a GUI option I've added one in this PR (and will look forward to solving the merge conflicts in 849 once again :P)

riperiperi:
Choices aren't limited in GUI, but since the effect is it just picking the next most compatible anisotropy level then I think it's fine.

When we have more options like this we should have an enhancements section or tab, with a disclaimer that they might negatively affect some games.

PR #1046, opened 2020-03-26
Set default width and height based on monitor resolution

Xpl0itR:


jduncanator:
I'm not sure how I feel about this one, it gets a little frustrating when you're repeatedly iterating and developing on something, especially on larger 27"+ monitors where I'd never have the emulator full screened anyway. What was the thinking behind this?

Xpl0itR:
On sub-720p monitors, the GUI is too large to fit on the screen. This PR checks if the usual size would not fit on the monitor and resize if needed. This PR would not affect your larger 27"+ monitors.

jduncanator:
Ahh, I see, it only resizes for monitors less than 1280x760.

PR #1003, opened 2020-03-22
HID SharedMem Rework

mageven:
- Inspired by gdkchan's HID struct PR (#857)
- Implements hook into StyleSetUpdatedEvent
- Add basic controller applet
- Updates sequence number correctly
- Add support for mouse device (untested)
- Add support for multiple controllers (partially tested, as long as game doesn't invoke controller applet)
- Adds auto redirecting primary input to a supported controller if only unsupported controllers are configured

Code seperated to commits for easier review. Please also check for common gotchas as I haven't professionally coded in C#.

Potentially Closes #673 (Please confirm)

PR #1048, opened 2020-03-26
Add the stub applet instead of not implemented exception.

Chenj168:
This is a similar implementation of yuzu, 
the Diablo III Eternal Collection can be played.
hc

### DIFF
--- a/Ryujinx.HLE/HOS/Applets/AppletManager.cs
+++ b/Ryujinx.HLE/HOS/Applets/AppletManager.cs
@@ -24,7 +24,7 @@ namespace Ryujinx.HLE.HOS.Applets
                 return (IApplet)Activator.CreateInstance(appletClass, system);
             }
 
-            throw new NotImplementedException($"{applet} applet is not implemented.");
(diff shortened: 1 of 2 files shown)


B3astTech:
Iirc it was already implemented by someone within a project. It has not been pr'd yet or anything. But a stub of it isnt the best way to go

gdkchan:
Which applet Diablo III requires?

ReaperOfSouls1909:
> Which applet Diablo III requires?

controller applet

gdkchan:
Controller applet is already implemented on #1003, so the only value added here is for other applets (not as frequently used).

jduncanator:
I think as a project overall, we've made a concerted effort going forward to merging as minimal amount of stubs as possible, and avoiding large "catch-all" stubs at all costs (which this sort of is). As @gdkchan mentioned, the last commonly used applet is the controller applet which already has a PR open for it, so I'll close this one.

Thanks for the interest in the project, hope to see more PRs in the future :)

PR #1050, opened 2020-03-26
Workaround for AMD and Intel view format bug

gdkchan:
Intel has a bug where trying to render to a texture view with a format different from it's parent will still use the parent format.

As workaround, this creates a intermmediate texture, copies -> intermmediate, renders and copies intermmediate -> parent.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Debugger.cs
+++ b/Ryujinx.Graphics.OpenGL/Debugger.cs
@@ -34,7 +34,7 @@ namespace Ryujinx.Graphics.OpenGL
             switch (type)
             {
                 case DebugType.DebugTypeError:
-                    Logger.PrintDebug(LogClass.Gpu, fullMessage);
+                    Logger.PrintError(LogClass.Gpu, fullMessage);
                     break;
                 case DebugType.DebugTypePerformance:
                     Logger.PrintWarning(LogClass.Gpu, fullMessage);
--- a/Ryujinx.Graphics.OpenGL/Framebuffer.cs
+++ b/Ryujinx.Graphics.OpenGL/Framebuffer.cs
@@ -10,9 +10,13 @@ namespace Ryujinx.Graphics.OpenGL
 
         private FramebufferAttachment _lastDsAttachment;
 
+        private readonly TextureView[] _colors;
+
         public Framebuffer()
         {
             Handle = GL.GenFramebuffer();
+
+            _colors = new TextureView[8];
         }
 
         public void Bind()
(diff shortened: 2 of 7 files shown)


LePwnerer:
RenderDoc Captures:
image
image

Shaders:
[Shader 2974.txt](https://github.com/Ryujinx/Ryujinx/files/4396587/Shader.2974.txt)
[Shader 2975.txt](https://github.com/Ryujinx/Ryujinx/files/4396588/Shader.2975.txt)

LePwnerer:
Ok nvm it's the same fix for amd.

just check for intel or amd. for linux might aswell check for mesa
```
private static GpuVendor GetGpuVendor() {
            string vendor = GL.GetString( StringName.Vendor ).ToLower();
            if ( vendor == "nvidia corporation" ) {
                return GpuVendor.Nvidia;
            } else if ( vendor == "intel" ) {
                return GpuVendor.Intel;
            } else if ( vendor == "ati technologies inc." || vendor == "advanced micro devices, inc." || vendor == "mesa project" ) {
                return GpuVendor.AMD;
            } else {
                return GpuVendor.Unknown;
            }
        }
```

jduncanator:
I don't think the workaround is required for the AMD Linux drivers, afaik, they don't suffer from the same bug.

LePwnerer:
@jduncanator https://media.discordapp.net/attachments/691556414463803413/693319402212294676/unknown.png

LePwnerer:
https://github.com/LePwnerer/Ryujinx/commit/5cd87cd76d7a7980faaedb41799a9be59eaf439e forked Ryujinx and commited the fix to the current branch

jduncanator:
@LePwnerer In that screenshot, is that the Mesa driver, or the proprietary driver? From the vendor ID, it looks like the proprietary driver. I believe the Mesa driver does not have this bug.

gdkchan:
Used the same workaround for AMD aswell, thanks for testing.

sigexp:
@gdkchan , Hi. Mesa dev here. Can you tell me what were the formats of texture view and its parent, that lead to misrendering?

I'm looking at the renderdoc capture of apitrace, provided in the [mesa issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/2708). I can clearly see the defects, but I've not been able to find the glTextureView call with format that is different from the format of its parent.

Looking briefly at the implementation of glTextureView in Mesa, it seems that it does everything according to the spec. At least, in regard to dealing with formats of texture view. Meaning that if the formats of texture view and the original texture are different, they must be compatible according to the Table 8.22 from section 8.18 in the OpenGL 4.6 spec:

image

It would be really great to know which combination leads to misrendering.

gdkchan:
This is a different issue, I don't think that mesa has this issue. The problem on mesa seems to be that it reports the ASTC compressed format as supported (KHR_texture_compression_astc_ldr extension), this causes the emulator to send compressed astc textures directly, instead of using the internal decompressor and sending rgba8 textures. I was told that mesa has a ASTC decompressor, but it seems to be failing somehow decompressing those textures, and sometimes it returns a GL_OUT_OF_MEMORY error.

Answering your specific question (even if I don't think it applies to mesa), the parent format was rg16f, and the view format was r11g10b10f.

gdkchan:
Tested on Intel UHD Graphics 630 on windows (which has hardware support for the ASTC compressed texture format, and also exposes KHR_texture_compression_astc_ldr on windows), and the icons are not corrupted:
image
So the ASTC texture data that we send is valid.

jduncanator:
Hey @sigexp Sorry that issue is such a mess, it appears some users were overly eager and opened the issue on our behalf before we had a chance to open more detailed ones ourselves.

I've added all the information I currently have to the Mesa issue linked above.

Yoshiepic:
Can someone give a slightly more simplified answer on how to fix this issue? I cannot follow what is being said in this thread in the slightest.

PR #1051, opened 2020-03-27
Fix MME shadow RAM implementation

gdkchan:
This in an improvement of #987, the shadow ram control affects all writes, not only those coming from MME.
In addition to the linked PR, this fixes more issues with some games, like for example stencil mask being destroyed by a stencil clear.
This fixes for examlpe, stencil logo effect on Kill la Kill.

Before:
image
After:
image

Thanks again to @fincs for clarifying that.

### DIFF
--- a/Ryujinx.Graphics.Gpu/State/GpuState.cs
+++ b/Ryujinx.Graphics.Gpu/State/GpuState.cs
@@ -12,7 +12,8 @@ namespace Ryujinx.Graphics.Gpu.State
 
         public delegate void MethodCallback(GpuState state, int argument);
 
-        private int[] _backingMemory;
+        private readonly int[] _memory;
+        private readonly int[] _shadow;
 
         /// <summary>
         /// GPU register information.
@@ -29,14 +30,15 @@ namespace Ryujinx.Graphics.Gpu.State
             public bool Modified;
         }
 
-        private Register[] _registers;
+        private readonly Register[] _registers;
 
         /// <summary>
         /// Creates a new instance of the GPU state.
         /// </summary>
         public GpuState()
         {
-            _backingMemory = new int[RegistersCount];
+            _memory = new int[RegistersCount];
+            _shadow = new int[RegistersCount];
 
             _registers = new Register[RegistersCount];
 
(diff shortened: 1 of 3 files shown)


Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -100,7 +117,7 @@ public int Read(int offset)
         /// <param name="value">Value to be written</param>
         public void Write(int offset, int value)
         {
-            _backingMemory[offset] = value;
+            _memory[offset] = value;
```
jduncanator: This should be tracked in the `_shadow` memory when `ShadowRamControl.Track` or `ShadowRamControl.TrackWithFilter` is set.

Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -109,31 +126,31 @@ public void Write(int offset, int value)
         /// <param name="offset">The offset to be written</param>
         public void SetUniformBufferOffset(int offset)
         {
-            _backingMemory[(int)MethodOffset.UniformBufferState + 3] = offset;
+            _memory[(int)MethodOffset.UniformBufferState + 3] = offset;
```
jduncanator: Same as above, this should be tracked in the `_shadow` memory when `ShadowRamControl.Track` or `ShadowRamControl.TrackWithFilter` is set.

Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -293,7 +310,7 @@ public bool QueryModified(MethodOffset m1, MethodOffset m2, MethodOffset m3, Met
         /// <returns>The data at the specified location</returns>
         public T Get<T>(MethodOffset offset) where T : struct
         {
-            return MemoryMarshal.Cast<int, T>(_backingMemory.AsSpan().Slice((int)offset))[0];
+            return MemoryMarshal.Cast<int, T>(_memory.AsSpan().Slice((int)offset))[0];
```
jduncanator: I assume reads from registers always read from main memory, and not shadow memory? Does this apply to MME reads also, or can MME read from shadow memory directly?

gdkchan: I guess so, if read was done from shadow ram, it would defeat the purpose of having the shadow ram.

jduncanator: My thought was more along the line of, what happens when shadow control is set to Replay, and you read from a register? Should it copy the shadow value to main memory, and then return the updated value? Should it return the original value directly from main memory?

PR #1059, opened 2020-03-27
Added GUI option to open shader folder & Fixed typo

MelonSpeedruns:


jduncanator:
What was your thinking behind adding this feature in? What are users/yourself needing to do with the shader cache that requires this option?

I think this is something we actively want to avoid offering. The shader cache is designed to be something that is completely managed behind the scenes without any user interaction. If there is a time the user has to interact with the shader cache to fix something, manually delete it, or otherwise modify it in any way, then there is a bug with the shader cache that should be fixed.

I want to make it as hard as humanly possible for people to share shader caches in the long run, not only because it's (theoretically) illegal, but also because it has unforeseen consequences when a shader cache is loaded from a different GPU series or vendor, than the user has themsevles.

emmauss:
Also, we currently do not have persistent shader cache, so even this pr is a bit premature

gdkchan:
The proposed change is on the shader cache branch.

komawoyo:
I believe the author may be referencing the cache system built by Cemu which has a 'transferable' cache and 'precompiled' cache system which allows for cache sharing. Only the non-sharable cache is found in the precompiled cache

[https://cemu.info/changelog/cemu_1_7_0.txt](https://cemu.info/changelog/cemu_1_7_0.txt)

https://www.reddit.com/r/CEMUcaches/comments/6223uy/faq/

jduncanator:
Sounds like CEMU's shader cache system stores the original game shaders on disk, and recompiles them at boot. This sounds like it's even more illegal than sharing recompiled shaders 🙄

komawoyo:
I believe yuzu took a similar approach. Who knows what's the best way.

MelonSpeedruns:
Closed it because it would allow sending shader cache easier to users, and therefore encourage piracy.

PR #1062, opened 2020-03-28
Fix metadata.json JsonParsingException crash

Xpl0itR:
Catch JsonParsingException and load defualt metadata in the event that the json gets nulled

### DIFF
--- a/Ryujinx/Ui/ApplicationLibrary.cs
+++ b/Ryujinx/Ui/ApplicationLibrary.cs
@@ -531,7 +531,21 @@ namespace Ryujinx.Ui
 
             using (Stream stream = File.OpenRead(metadataFile))
             {
-                appMetadata = JsonSerializer.Deserialize<ApplicationMetadata>(stream, resolver);
+                try
+                {
+                    appMetadata = JsonSerializer.Deserialize<ApplicationMetadata>(stream, resolver);
+                }
+                catch (JsonParsingException)
+                {
+                    Logger.PrintWarning(LogClass.Application, $"Failed to parse metadata json for {titleId}. Loading defaults.");
+                    
+                    appMetadata = new ApplicationMetadata
+                    {
+                        Favorite   = false,
+                        TimePlayed = 0,
+                        LastPlayed = "Never"
+                    };
+                }
             }
 
             if (modifyFunction != null)
--- a/Ryujinx/Ui/ApplicationLibrary.cs
+++ b/Ryujinx/Ui/ApplicationLibrary.cs
@@ -1,4 +1,3 @@
-using JsonPrettyPrinterPlus;
 using LibHac;
 using LibHac.Common;
 using LibHac.Fs;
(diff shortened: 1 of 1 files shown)


Ryujinx/Ui/ApplicationLibrary.cs
```diff
@@ -532,20 +532,40 @@ internal static ApplicationMetadata LoadAndSaveMetaData(string titleId, Action<A
                 };
 
                 byte[] data = JsonSerializer.Serialize(appMetadata, resolver);
-                File.WriteAllText(metadataFile, Encoding.UTF8.GetString(data, 0, data.Length).PrettyPrintJson());
+                using (TextWriter textWriter = new StreamWriter(File.Create(metadataFile, 65536, FileOptions.WriteThrough)))
```
jduncanator: In my opinion, the metadata.json for games doesn't need to be pretty printed, it's just metadata and has no human readable benefit. This could be greatly simplified to something like:

```cs
using (var stream = File.Create(metadataFile))
using (var writer = new StreamWriter(stream))
{
    JsonSerializer.Serialize(writer, appMetadata, resolver);
}
```

The using block will ensure the StreamWriter is flushed to disk when the scope is ended. FileOptions.WriteThrough only affects file system block cache, and won't affect the StreamWriter's buffering behaviour, so `FileOptions.WriteThrough` is not needed, nor is it really the correct way to approach this. This also saves serializing the entire JSON object into memory, rather writes it directly to disk as it's being serialized.

Ryujinx/Ui/ApplicationLibrary.cs
```diff
@@ -532,20 +532,40 @@ internal static ApplicationMetadata LoadAndSaveMetaData(string titleId, Action<A
...
+                }
             }
 
             if (modifyFunction != null)
             {
                 modifyFunction(appMetadata);
 
-                byte[] saveData = JsonSerializer.Serialize(appMetadata, resolver);
-                File.WriteAllText(metadataFile, Encoding.UTF8.GetString(saveData, 0, saveData.Length).PrettyPrintJson());
+                byte[] data = JsonSerializer.Serialize(appMetadata, resolver);
```
jduncanator: The same comment from above is also applicable here.

Xpl0itR:
Since i was unable to reproduce this issue in the 1st place, I've reverted the change to writing the json as i have no way to test if it fixes the issue or not. The try-catch is sufficient enough to stop the emulator from crashing.

Xpl0itR:
I've managed to reproduce the error by putting the code in an infinite loop and then cutting the power to my PC 😬. The cause and symptoms are identical to this stackoverflow post: https://stackoverflow.com/questions/54078564/file-containing-null-values-instead-of-json-object-after-computer-restart/
As such, I've used `FileOptions.WriteThrough` to instruct the OS to write directly to disk without buffering, this should decrease the chance of the file getting nulled

PR #1066, opened 2020-03-28
Support constant attributes (with a value of zero)

gdkchan:
When a attribute is not used, NVN sets bit 6 on the attribute, which indicates that it uses a constant value. The emulator was ignoring that bit and assuming all attributes would always point to data inside the vertex buffer. This is incorrect and some games relies on the constant behavior -- basically they use on the shader attributes that are "disabled".

This fixes bug on some tiles on Animal Crossing.
Before:
image
After:
image

Recommend testing in a few games to be sure that nothing regressed.

### DIFF
--- a/Ryujinx.Graphics.Gpu/State/VertexAttribState.cs
+++ b/Ryujinx.Graphics.Gpu/State/VertexAttribState.cs
@@ -17,6 +17,15 @@ namespace Ryujinx.Graphics.Gpu.State
         }
 
         /// <summary>
+        /// Unpacks the attribute constant flag.
+        /// </summary>
+        /// <returns>True if the attribute is constant, false otherwise</returns>
+        public bool UnpackIsConstant()
+        {
+            return (Attribute & 0x40) != 0;
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Gpu/State/VertexAttribState.cs
```diff
@@ -16,6 +16,15 @@ public int UnpackBufferIndex()
             return (int)(Attribute & 0x1f);
         }
 
+        /// <summary>
+        /// Unpacks the attribute constant flag.
+        /// </summary>
+        /// <returns>True if the attribute is constant, false otherwise</returns>
+        public bool UnpackIsConstant()
```
jduncanator: Out of curiosity, why would these not be implemented as simple getters? Something like:

```cs
public bool IsConstant => (Attribute & 0x40) != 0;
public int Offset => (int)((Attribute >> 7) & 0x3fff);
// etc...
```

gdkchan: There's no special reason. Not a huge fan of mixing (public) fields and properties though.

PR #1068, opened 2020-03-29
Implement FCMP shader instruction

kimcoreovo:


kimcoreovo:
sry, pressed a wrong button

PR #1080, opened 2020-03-31
Fix tooltip on the anisotropic filtering setting

Xpl0itR:


Ryujinx/Ui/SwitchSettings.glade
```diff
@@ -1472,7 +1472,7 @@
                                       <object class="GtkLabel">
                                         <property name="visible">True</property>
                                         <property name="can_focus">False</property>
-                                        <property name="tooltip_text" translatable="yes">Graphics Shaders Dump Path</property>
+                                        <property name="tooltip_text" translatable="yes">Set the level of Anisotropic Filtering. Or set to Auto to let the game decide</property>
```
gdkchan: I don't think the `Set` is necessary. Could be just `Level of Anisotropic Filtering (set to Auto to use the value requested by the game)`.

Xpl0itR: changed

PR #1082, opened 2020-04-01
Better IPA shader instruction implementation

gdkchan:
Using IPA instructions information to determine the interpolation that should be done on a varying is not reliable, because the compiler may decide to use, for example, IPA.PASS (with subsequent multiply) to do perspective correct interpolation, rather than IPA.

The fix here is using interpolation information from the shader header instead. It seems to affect most of the interpolation process, except for the final multiplication by `1 / gl_FragCoord.w`. So we can just multiply the interpolated value by `gl_FragCoord.w` to undo this last step, and then, for IPA (default mode), we multiply the result by the register (which should contain `1 / gl_FragCoord.w`).

This fixes a issue on ACNH where some objects were not being draw at specific angles.
Before:
image
After:
image
Thanks to VanillaLatte for providing the save.

Also fixes Touhou Kobuto V textures:
Before:
image
After:
image

Depends on #1070.


GamerzHell9137:
Someone posted on the discord that FE is fixed by it 
image
image

AcK77:
C477F845-21F2-4A41-9C0C-BB0DAED71839
Users reports it’s fix footprints in ACNH

ReaperOfSouls1909:
https://cdn.discordapp.com/attachments/480497101676216340/694895314481709076/unknown.png fixes splashing waves https://media.discordapp.net/attachments/480497101676216340/694894797286015016/unknown.png fixes smoke 
https://cdn.discordapp.com/attachments/480497101676216340/694893566442340472/unknown.png fixes waterfall 
https://cdn.discordapp.com/attachments/480497101676216340/694892113015603240/unknown.png fixes airport lighting

Mattscreative:
Screenshot from 2020-04-01 16-01-05
image at night in the airport looks fixed also but house lighting in the day is busted still

GamerzHell9137:
@Mattscreative what do you mean fixed? this is how it looks on real hardware
image

riperiperi:
image
- Lasers in Bowser's Castle.
- Inner part of drift turbo particles (closest to the wheel's point of contact)
- Smoke particles when drifting, start boost, spot turn.
- Some course specific effects.

PR #1093, opened 2020-04-06
Simple GPU fixes

### LINKED ISSUE #875: Missing 3D register RASTERIZE_ENABLE implementation
Register [RASTERIZE_ENABLE](https://github.com/envytools/envytools/blob/d799c46c16e6a04c0395c04018b4c326971d0294/rnndb/graph/gf100_3d.xml#L186) is not implemented. This one maps properly to OpenGL's `GL_RASTERIZER_DISCARD` and has to be enabled accordingly for drawing, it also affects clearing but it isn't tested if this is also true on hardware. If I remember correctly, commercial games assume this register is enabled by default.

It is visible in Xenoblade Chronicles 2 were there is a point in the middle of the screen. This is because the game uses a vertex shader invocation with rasterization disabled as if it were a compute shader.
imagen
imagen
Screenshot taken by Discord user JoyConBoyForever_Octolimar

mageven:
### 1. Implement RasterizeEnable
Implements RasterizeEnable register using `GL_RASTERIZER_DISCARD`. Fixes the micro black hole that appears in XC2 menu and opening sequence.

Before (the small dot near the top of the image)-
xc2-before
After-
xc2-after

RasterizeEnable is set to 1 on init as described in nouveau/mesa/deko3d.
`GL_RASTERIZER_DISCARD` causes `glClear`s to be ignored so it is disabled before host clears ~but not at other places. I think this is the right behavior. [Wiki Source](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glEnable.xhtml)~

EDIT: Apparently it affects blits on some vendors (NVIDIA). Thanks to @gdkchan for promptly testing.

### 2. Match viewport count to hardware
Viewport counts were set to 8 previously. Switch hardware supports max of 16.

### 3. Simplify ScissorTest tracking around Blits
Blittting only uses Viewport 0 according to [wiki](https://www.khronos.org/opengl/wiki/Framebuffer#Blitting). So, simplified the state tracking.

Closes #875 

Thanks to @ReinUsesLisp for the detailed description in the issue and @gdkchan for clarifications.

Please test this PR thoroughly as it is my first GPU PR.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -216,10 +216,14 @@ namespace Ryujinx.Graphics.Gpu.Engine
             CommitBindings();
         }
 
+        /// <summary>
+        /// Updates Rasterizer primitive discard state based on guest gpu state.
+        /// </summary>
+        /// <param name="state">Current GPU state</param>
         private void UpdateRasterizerState(GpuState state)
         {
-            int enable = state.Get<int>(MethodOffset.RasterizeEnable);
-            _context.Renderer.Pipeline.SetRasterizerDiscard(enable == 0);
+            Boolean32 enable = state.Get<Boolean32>(MethodOffset.RasterizeEnable);
+            _context.Renderer.Pipeline.SetRasterizerDiscard(!enable);
         }
 
         /// <summary>
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -108,6 +108,11 @@ namespace Ryujinx.Graphics.Gpu.Engine
                 UpdateShaderState(state);
             }
 
+            if (state.QueryModified(MethodOffset.RasterizeEnable))
+            {
+                UpdateRasterizerState(state);
+            }
+
             if (state.QueryModified(MethodOffset.RtColorState,
                                     MethodOffset.RtDepthStencilState,
                                     MethodOffset.RtControl,
@@ -209,6 +214,12 @@ namespace Ryujinx.Graphics.Gpu.Engine
             }
 
             CommitBindings();
+        }
+
+        private void UpdateRasterizerState(GpuState state)
+        {
+            int enable = state.Get<int>(MethodOffset.RasterizeEnable);
+            _context.Renderer.Pipeline.SetRasterizerDiscard(enable == 0);
         }
 
         /// <summary>
--- a/Ryujinx.Graphics.Gpu/State/GpuStateTable.cs
+++ b/Ryujinx.Graphics.Gpu/State/GpuStateTable.cs
@@ -53,11 +53,11 @@ namespace Ryujinx.Graphics.Gpu.State
         public static TableItem[] Table = new TableItem[]
         {
             new TableItem(MethodOffset.RtColorState,           typeof(RtColorState),          8),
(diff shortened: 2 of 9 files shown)


Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -211,6 +216,12 @@ private void UpdateState(GpuState state)
             CommitBindings();
         }
 
+        private void UpdateRasterizerState(GpuState state)
+        {
+            int enable = state.Get<int>(MethodOffset.RasterizeEnable);
```
gdkchan, line 221: you could use the 'Boolean32' type here (assuming it can only be 0 and 1). Then instead of 'enable == 0', it could be '!enable".

mageven: I wasn't sure if it's non-zero or 1 so, I made it an int

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -980,6 +985,14 @@ public void RestoreScissor0Enable()
             }
         }
 
+        public void RestoreRasterizerDiscard()
+        {
+            if(_rasterizerDiscard)
```
gdkchan, line 990: nit: `if(` -> `if (`.

Ryujinx.Graphics.Gpu/State/GpuStateTable.cs
```diff
@@ -53,11 +53,11 @@ public TableItem(MethodOffset offset, Type type, int count)
...
-            new TableItem(MethodOffset.ViewportExtents,        typeof(ViewportExtents),       8),
+            new TableItem(MethodOffset.ViewportTransform,      typeof(ViewportTransform),     Constants.TotalViewports),
+            new TableItem(MethodOffset.ViewportExtents,        typeof(ViewportExtents),       Constants.TotalViewports),
             new TableItem(MethodOffset.VertexBufferDrawState,  typeof(VertexBufferDrawState), 1),
             new TableItem(MethodOffset.DepthBiasState,         typeof(DepthBiasState),        1),
-            new TableItem(MethodOffset.ScissorState,           typeof(ScissorState),          8),
+            new TableItem(MethodOffset.ScissorState,           typeof(ScissorState),          Constants.TotalViewports),
             new TableItem(MethodOffset.StencilBackMasks,       typeof(StencilBackMasks),      1),
             new TableItem(MethodOffset.RtDepthStencilState,    typeof(RtDepthStencilState),   1),
             new TableItem(MethodOffset.VertexAttribState,      typeof(VertexAttribState),     16),
```
Cristallix, line 63: I know it's not the purpose of the pr, but all TableItem should take values from constants no ?

mageven: Yes, it is good to have them through constants (like RenderTargets and VertexBuffers). It's a bit larger effort than adding them only here. Some loops have the same hardcoded values. I'm still learning about the different associations, so I didn't add them in this PR. Maybe if I touch them in the future, I'll add them.

jduncanator:
This looks fine to me. Just for future, try to keep PRs pure, in that they only make a single set of chanegs (or changes closely related to that single set of changes). It allows use to bisect git much more easily, and resolve future problems down to the offending PR quicker :)

(For example, Match viewport count to hardware change should really be in another PR).

LGTM though, thanks :+1:

PR #1094, opened 2020-04-07
Optimize %x ^ %x = 0

FICTURE7:
Simplify `BitwiseExclusiveOr %x %x` to 0 which incurs a modest diff.

IR:
```patch
-  i64 %12 = BitwiseExclusiveOr i64 %1, i64 %1
-  i64 %13 = BitwiseAnd i64 %12, i64 %1
-  i32 %14 = CompareLess i64 %13, i64 0x0
-  i32 %15 = CompareNotEqual i64 %1, i64 0x0
-  BranchIfTrue i32 %15
+  i32 %12 = CompareLess i64 0x0, i64 0x0
+  i32 %13 = CompareNotEqual i64 %1, i64 0x0
+  BranchIfTrue i32 %13
```

CodeGen:
```patch
 cmp    rbp,0x0
 setae  dl
 movzx  edx,dl
-mov    rcx,rbp
-xor    rcx,rbp
-and    rcx,rbp
+xor    ecx,ecx
 cmp    rcx,0x0
 setl   r8b
 movzx  r8d,r8b
```

In large translations, it can save a few local variables which affects register usage and the amount of spills.

`CompareLess` and friends are not constant folded currently, so the ` cmp    rcx,0x0` remains.

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/Simplification.cs
+++ b/ARMeilleure/CodeGen/Optimizations/Simplification.cs
@@ -24,7 +24,7 @@ namespace ARMeilleure.CodeGen.Optimizations
                     break;
 
                 case Instruction.BitwiseExclusiveOr:
-                    TryEliminateBitwiseExlusiveOr(operation);
+                    TryEliminateBitwiseExclusiveOr(operation);
                     break;
 
                 case Instruction.ConditionalSelect:
@@ -92,7 +92,7 @@ namespace ARMeilleure.CodeGen.Optimizations
             }
         }
 
-        private static void TryEliminateBitwiseExlusiveOr(Operation operation)
+        private static void TryEliminateBitwiseExclusiveOr(Operation operation)
         {
             // Try to recognize and optimize those 2 patterns (in order):
             // x ^ y == 0x00000000 when x == y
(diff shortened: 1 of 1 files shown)


ARMeilleure/CodeGen/Optimizations/Simplification.cs
```diff
@@ -89,6 +92,24 @@ private static void TryEliminateBitwiseOr(Operation operation)
             }
         }
 
+        private static void TryEliminateBitwiseExlusiveOr(Operation operation)
+        {
+            // Try to recognize and optimize those 2 patterns (in order):
+            // x ^ y == 0x00000000  when x == y
```
gdkchan, line 98: nit: Turn double space before `when` into single space.

ARMeilleure/CodeGen/Optimizations/Simplification.cs
```diff
@@ -89,6 +92,24 @@ private static void TryEliminateBitwiseOr(Operation operation)
...
+            // 0x00000000 ^ y == y, x ^ 0x00000000 == x
+            Operand x = operation.GetSource(0);
+            Operand y = operation.GetSource(1);
+
+            if (x == y && x.Type.IsInteger())
+            {
+                operation.TurnIntoCopy(Const(0));
```
gdkchan, line 105: The constant should preserve the type of the operation. So I believe it should be `Const(0, x.Type)`.

gdkchan: I just took a look and found other instances of the same problem there aswell, if you want to fix in the future.

PR #1095, opened 2020-04-07
Implement game card mounting and loading

Xpl0itR:
This PR allows you to mount/dismount an xci file into/from the virtual game card. You can also load an application from the virtual game card, `File > Load Application from Game Card`

However, there is a bug somewhere, possibly in my implementation or elsewhere, where game card dumper homebrew fail to dump the cart.

image


AcK77:
I don’t think it should be implemented as-is. GC Dumper Homebrew use FPS-dev command to mount the game card and read it through other commands... There is no benefit to mount the game card if everything who use it can’t read it.
For now I’ll change this PR to blocked because of that!
Nice try btw!

PR #1097, opened 2020-04-08
Improve V128

FICTURE7:
This new implementation takes advantage of `_e0` and `_e1` being contiguous in memory and turns the accesses into essentially array indexing, through the use of `Unsafe`.

The RyuJIT can now inline most of the methods, unlike before where it would complain.

Inline decisions of RyuJIT before:
```
[0 IL=0000 TR=000000 06000001] [FAILED: ldsfld of value class] V128:get_Zero():V128
[0 IL=0002 TR=000135 06000014] [FAILED: too many basic blocks] V128:GetUInt64(int):long:this
[0 IL=0002 TR=000157 06000013] [FAILED: has switch] V128:GetUInt32(int):int:this
[0 IL=0193 TR=000107 0600000B] [FAILED: too many il bytes] V128:Insert(int,int):this
[0 IL=0209 TR=000117 0600000C] [FAILED: too many basic blocks] V128:Insert(int,long):this
```
Since `GetUInt64` and `GetUInt32` are the leaf methods and they themselves don't inline, `GetFloat` and friends don't inline deeply (or completely).

Inline decisions of RyuJIT after:
```
[1 IL=0000 TR=000000 06000001] [below ALWAYS_INLINE size] V128:get_Zero():V128
[12 IL=0002 TR=000251 06000014] [profitable inline] V128:GetUInt64(int):long:this
[20 IL=0002 TR=000326 06000011] [profitable inline] V128:GetUInt32(int):int:this
[70 IL=0193 TR=000107 06000013] [profitable inline] V128:Insert(int,int):this
[76 IL=0209 TR=000117 06000014] [profitable inline] V128:Insert(int,long):this
```
This should affect other methods working with `V128`. For example:
https://github.com/Ryujinx/Ryujinx/blob/b8e3909d800ff5947683bb169d8efda2ef63d697/ARMeilleure/Memory/MemoryManager.cs#L664-L668
The calls to `GetUInt64` are not currently getting inlined, with this PR it does.

When I have time, I'll try to grab some benchmark numbers.

Also, let me know if the code here is not inline with the coding style of the project. :>

### DIFF
--- a/ARMeilleure/State/V128.cs
+++ b/ARMeilleure/State/V128.cs
@@ -1,20 +1,27 @@
 using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 namespace ARMeilleure.State
 {
+     [StructLayout(LayoutKind.Explicit, Size = 16)]
     public struct V128 : IEquatable<V128>
     {
+        [FieldOffset(0)]
         private ulong _e0;
+        [FieldOffset(8)]
         private ulong _e1;
 
-        private static V128 _zero = new V128(0, 0);
-
-        public static V128 Zero => _zero;
-
-        public V128(float value) : this(value, 0, 0, 0) { }
+        public static V128 Zero => new V128(0, 0);
 
         public V128(double value) : this(value, 0) { }
+        public V128(double e0, double e1)
+        {
+            _e0 = (ulong)BitConverter.DoubleToInt64Bits(e0);
+            _e1 = (ulong)BitConverter.DoubleToInt64Bits(e1);
+        }
 
+        public V128(float value) : this(value, 0, 0, 0) { }
         public V128(float e0, float e1, float e2, float e3)
         {
             _e0  = (ulong)(uint)BitConverter.SingleToInt32Bits(e0) << 0;
@@ -23,20 +30,14 @@ namespace ARMeilleure.State
             _e1 |= (ulong)(uint)BitConverter.SingleToInt32Bits(e3) << 32;
         }
 
-        public V128(double e0, double e1)
+        public V128(long e0, long e1) : this((ulong)e0, (ulong)e1) { }
+        public V128(ulong e0, ulong e1)
         {
-            _e0 = (ulong)BitConverter.DoubleToInt64Bits(e0);
-            _e1 = (ulong)BitConverter.DoubleToInt64Bits(e1);
+            _e0 = e0;
+            _e1 = e1;
         }
 
-        public V128(int e0, int e1, int e2, int e3)
-        {
-            _e0  = (ulong)(uint)e0 << 0;
-            _e0 |= (ulong)(uint)e1 << 32;
-            _e1  = (ulong)(uint)e2 << 0;
-            _e1 |= (ulong)(uint)e3 << 32;
-        }
-
+        public V128(int e0, int e1, int e2, int e3) : this((uint)e0, (uint)e1, (uint)e2, (uint)e3) { }
         public V128(uint e0, uint e1, uint e2, uint e3)
         {
             _e0  = (ulong)e0 << 0;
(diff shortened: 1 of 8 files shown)


FICTURE7:
For comparison, here is the change in codegen for `GetUInt32`

Before:
```asm
G_M31163_IG01:
       56                   push     rsi
       4883EC20             sub      rsp, 32
						;; bbWeight=1    PerfScore 1.25
G_M31163_IG02:
       83FA03               cmp      edx, 3
       7744                 ja       SHORT G_M31163_IG11
       8BC2                 mov      eax, edx
       488D1575000000       lea      rdx, [reloc @RWD00]
       8B1482               mov      edx, dword ptr [rdx+4*rax]
       4C8D05E8FFFFFF       lea      r8, G_M31163_IG02
       4903D0               add      rdx, r8
       FFE2                 jmp      rdx
						;; bbWeight=1    PerfScore 7.25
G_M31163_IG03:
       8B01                 mov      eax, dword ptr [rcx]
						;; bbWeight=0.50 PerfScore 1.00
G_M31163_IG04:
       4883C420             add      rsp, 32
       5E                   pop      rsi
       C3                   ret      
						;; bbWeight=0.50 PerfScore 0.88
G_M31163_IG05:
       488B01               mov      rax, qword ptr [rcx]
       48C1E820             shr      rax, 32
						;; bbWeight=0.50 PerfScore 2.00
G_M31163_IG06:
       4883C420             add      rsp, 32
       5E                   pop      rsi
       C3                   ret      
						;; bbWeight=0.50 PerfScore 0.88
G_M31163_IG07:
       8B4108               mov      eax, dword ptr [rcx+8]
						;; bbWeight=0.50 PerfScore 1.00
G_M31163_IG08:
       4883C420             add      rsp, 32
       5E                   pop      rsi
       C3                   ret      
						;; bbWeight=0.50 PerfScore 0.88
G_M31163_IG09:
       488B4108             mov      rax, qword ptr [rcx+8]
       48C1E820             shr      rax, 32
						;; bbWeight=0.50 PerfScore 2.00
G_M31163_IG10:
       4883C420             add      rsp, 32
       5E                   pop      rsi
       C3                   ret      
						;; bbWeight=0.50 PerfScore 0.88
G_M31163_IG11:
       48B9105C9B38FC7F0000 mov      rcx, 0x7FFC389B5C10
       E833B8B95F           call     CORINFO_HELP_NEWSFAST
       488BF0               mov      rsi, rax
       B901000000           mov      ecx, 1
       48BA30D79938FC7F0000 mov      rdx, 0x7FFC3899D730
       E8BC91CB5F           call     CORINFO_HELP_STRCNS
       488BD0               mov      rdx, rax
       488BCE               mov      rcx, rsi
       E8D958FFFF           call     ArgumentOutOfRangeException:.ctor(String):this
       488BCE               mov      rcx, rsi
       E8E949B55F           call     CORINFO_HELP_THROW
       CC                   int3     
						;; bbWeight=0    PerfScore 0.00
RWD00  dd	0000001Dh ; case G_M31163_IG03
       dd	00000025h ; case G_M31163_IG05
       dd	00000032h ; case G_M31163_IG07
       dd	0000003Bh ; case G_M31163_IG09


; Total bytes of code 136, prolog size 5, PerfScore 31.60, (MethodHash=bdca8644) for method V128:GetUInt32(int):int:this
; ============================================================
```

After:
```asm
G_M31166_IG01:
       4883EC28             sub      rsp, 40
						;; bbWeight=1    PerfScore 0.25
G_M31166_IG02:
       83FA03               cmp      edx, 3
       770B                 ja       SHORT G_M31166_IG05
						;; bbWeight=1    PerfScore 1.25
G_M31166_IG03:
       4863C2               movsxd   rax, edx
       8B0481               mov      eax, dword ptr [rcx+4*rax]
						;; bbWeight=1    PerfScore 2.25
G_M31166_IG04:
       4883C428             add      rsp, 40
       C3                   ret      
						;; bbWeight=1    PerfScore 1.25
G_M31166_IG05:
       E8178EFFFF           call     V128:ThrowIndexOutOfRange()
       CC                   int3     
						;; bbWeight=0    PerfScore 0.00

; Total bytes of code 26, prolog size 4, PerfScore 7.60, (MethodHash=bdca8644) for method V128:GetUInt32(int):int:this
; ============================================================
```

FICTURE7:
For some methods in `V128`:

Before
|    Method |     Mean |     Error |    StdDev |
|---------- |---------:|----------:|----------:|
|      Zero | 1.599 ns | 0.0012 ns | 0.0011 ns |
| GetUInt64 | 3.594 ns | 0.0138 ns | 0.0115 ns |
| GetUInt32 | 3.572 ns | 0.0046 ns | 0.0043 ns |

After
|    Method |      Mean |     Error |    StdDev |
|---------- |----------:|----------:|----------:|
|      Zero | 0.0000 ns | 0.0000 ns | 0.0000 ns |
| GetUInt64 | 0.3313 ns | 0.0015 ns | 0.0012 ns |
| GetUInt32 | 0.5286 ns | 0.0013 ns | 0.0012 ns |

~8.8X speed up.

For the software fallback of SHA256:

Before
|     Method |     Mean |   Error |  StdDev |
|----------- |---------:|--------:|--------:|
| HashChoose | 177.1 ns | 0.13 ns | 0.12 ns |
|  HashLower | 397.3 ns | 0.41 ns | 0.39 ns |

After
|     Method |     Mean |    Error |   StdDev |
|----------- |---------:|---------:|---------:|
| HashChoose | 49.00 ns | 0.038 ns | 0.032 ns |
|  HashLower | 93.54 ns | 0.041 ns | 0.039 ns |

~4X speed up.

For `MemoryManager.WriteVector128`:

Before
|         Method |      Mean |    Error |   StdDev |
|--------------- |----------:|---------:|---------:|
|   AlignedWrite |  35.27 ns | 0.129 ns | 0.120 ns |
| UnalignedWrite | 237.07 ns | 0.220 ns | 0.195 ns |

After
|         Method |      Mean |    Error |   StdDev |
|--------------- |----------:|---------:|---------:|
|   AlignedWrite |  32.12 ns | 0.673 ns | 0.720 ns |
| UnalignedWrite | 229.90 ns | 2.379 ns | 2.226 ns |

I guess the cost of the other calls dwarfs the now inlined `GetUInt64`. This benchmark will be moot anyways when #856 gets merged, but this may give an insight on the speed up the new memory manager might produce especially for unaligned writes.

This PR might also speed up atomic loads on `V128` (`ReadVector128Exclusive`), but I did not measure these.

<details>
<summary>Benchmark Details</summary>

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.720 (1909/November2018Update/19H2)
Intel Pentium CPU G2020 2.90GHz, 1 CPU, 2 logical and 2 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT


```

```csharp

public class V128Benchmark
{
    public int Index { get; set; } = 0;

    [Benchmark]
    public V128 Zero()
    {
        return V128.Zero;
    }

    // Used GetUInt64() & GetUInt32() when measuring master.

    [Benchmark]
    public ulong GetUInt64()
    {
        return default(V128).Get<ulong>(Index);
    }

    [Benchmark]
    public uint GetUInt32()
    {
        return default(V128).Get<uint>(Index);
    }
}

public class WriteV128Benchmark
{
    private const int RamSize = 4 * 1024;

    private IntPtr _ram;
    private MemoryManager _memory;

    public int AlignedPosition { get; set; } = 0x0;
    public int UnalignedPosition { get; set; } = 0x7;

    [GlobalSetup]
    public void Setup()
    {
        _ram = MemoryManagement.Allocate(RamSize);
        _memory = new MemoryManager(_ram);
        _memory.Map(0x0000, 0x0000, RamSize);
    }

    [GlobalCleanup]
    public void Cleanup()
    {
        _memory.Dispose();
    }

    [Benchmark]
    public void AlignedWrite()
    {
        _memory.WriteVector128(AlignedPosition, default);
    }

    [Benchmark]
    public void UnalignedWrite()
    {
        _memory.WriteVector128(UnalignedPosition, default);
    }
}

public class Sha256Benchmark
{
    [Benchmark]
    public V128 HashChoose()
    {
        return SoftFallback.HashChoose(default, 0, default);
    }

    [Benchmark]
    public V128 HashLower()
    {
        return SoftFallback.HashLower(default, default, default);
    }
}

```
</details>

FICTURE7:
If this gets approved I'll try to add documentation, since it seems to be desired judging from the recent PRs.

ARMeilleure/State/V128.cs
```diff
@@ -186,29 +101,20 @@ public override int GetHashCode()
...
-            return other._e0 == _e0 && other._e1 == _e1;
-        }
+        public override int    GetHashCode() => HashCode.Combine(_e0, _e1);
+        public override string ToString()    => $"0x{_e1:X16}{_e0:X16}";
 
-        public override string ToString()
-        {
-            return $"0x{_e1:X16}{_e0:X16}";
-        }
+        private static void ThrowIndexOutOfRange() => throw new ArgumentOutOfRangeException("index");
```
gdkchan, line 118: I would prefer to pass the parameter name as argument to this method, just to be able to use `nameof(index)`.

gdkchan:
> 
> 
> This largely looks good to me, nice improvements! Some things to address are code style, which in some places isn't in line with the remainder of the project.
> 
> I also think the `Insert` method should keep it's name, rather than `Set`, as it better aligns with the naming convention typically used for SIMD registers (where one "inserts" a value into the SIMD register).

I think if is is kept as `Insert` then `Get` should be renamed to `Extract` for consistency. For me either one is fine, I think I have some preference for Get/Set since they are shorter.

FICTURE7:
> I also think the Insert method should keep it's name, rather than Set, as it better aligns with the naming convention typically used for SIMD registers (where one "inserts" a value into the SIMD register).

> I think if it is kept as Insert then Get should be renamed to Extract for consistency.

Sounds good to me.

ameisen:
I have some other improvements for V128, I'll wait for this to merge so I can include it in my PR.

PR #849, opened 2019-12-22
Improvements to input and input configuration in the GUI.

Xpl0itR:
## Improvements to input and input configuration in the GUI.
- Implement support for multiple controllers in addition to the keyboard
- Support multiple keyboards in addition to multiple controllers
- Controller settings window supports configuring both keyboard controlled virtual controllers and controller controlled virtual controllers
- Implement controller profiles

PR #1115, opened 2020-04-14
Update Config.json

Joshi234:
Changed deadzone becuase many people are experiencing drift because its to low

### DIFF
--- a/Ryujinx/Config.json
+++ b/Ryujinx/Config.json
@@ -74,7 +74,7 @@
     "joystick_controls": {
         "enabled": true,
         "index": 0,
-        "deadzone": 0.05,
+        "deadzone": 0.15,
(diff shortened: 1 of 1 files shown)


B3astTech:
Out of many others that don't is also to conclude. If someone has a issue with drift they can change this them self's. Because every controller is diffrent and will need to fine tune it to there own controller

Joshi234:
Ok I think it would be usefule if there is a setting in ryujnix for changing deadzone

B3astTech:
There is already a pr doing so. But it's not ready so it will come avenchally

AcK77:
This issue is already fixed in #849. You will be able to set the deadzone in the GUI.
I close this PR because of that. Thanks for your interest ;)!

PR #1118, opened 2020-04-15
Implement user-defined clipping (gl_ClipDistance) on GL state pipeline

mageven:
Complements PR #680 

> Note that, despite redeclaring the output, you still have to glEnable the specific clip distances you intend to use. If you enable a clip distance, but do not write to it from the last vertex processing stage, you get undefined behavior. 

[Source](https://www.khronos.org/opengl/wiki/Vertex_Post-Processing#User-defined_clipping)

Games I tested set clip mask to `0xff`. I just hope that all games redeclare `gl_ClipDistance` properly.

Tested on NVIDIA only.

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -67,6 +67,8 @@ namespace Ryujinx.Graphics.GAL
 
         void SetUniformBuffer(int index, ShaderStage stage, BufferRange buffer);
 
+        void SetUserClipDistance(int index, bool enableClip);
+
         void SetVertexAttribs(VertexAttribDescriptor[] vertexAttribs);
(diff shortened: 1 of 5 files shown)


gdkchan:
Nice work! However, for that to work, isn't gl_ClipDistance[] support on the shader required aswell?

ReinUsesLisp:
This might cause issues on Intel's Windows driver. When a clip distance is enabled but it's not written by a shader it may crop parts of the screen.  It's probably worth testing it on Intel before merging it.

To workaround this (in case that it happens on Ryujinx) you can disable those indices that are not written in shaders (unless something uses physical attributes).

jduncanator:
Before this gets merged, it really needs confirmation that these changes do not negatively impact Intel based GPUs

PR #1119, opened 2020-04-15
Implement Constant Color blends

mageven:
Also initializes blend states to GL default values.

Haven't seen any game use constant color blend modes, but the initialization may help fix any issues.

Added a local state to constant color because some games spam the same value.

Tested on NVIDIA.

**EDIT:** Adds Set methods to GpuState too

### DIFF
--- a/Ryujinx.Graphics.GAL/BlendDescriptor.cs
+++ b/Ryujinx.Graphics.GAL/BlendDescriptor.cs
@@ -4,7 +4,7 @@ namespace Ryujinx.Graphics.GAL
     {
         public bool Enable { get; }
 
-        public ColorF      ConstantBlend  { get; }
+        public ColorF      BlendConstant  { get; }
         public BlendOp     ColorOp        { get; }
         public BlendFactor ColorSrcFactor { get; }
         public BlendFactor ColorDstFactor { get; }
@@ -14,7 +14,7 @@ namespace Ryujinx.Graphics.GAL
 
         public BlendDescriptor(
             bool        enable,
-            ColorF      constantBlend,
+            ColorF      blendConstant,
             BlendOp     colorOp,
             BlendFactor colorSrcFactor,
             BlendFactor colorDstFactor,
@@ -23,7 +23,7 @@ namespace Ryujinx.Graphics.GAL
             BlendFactor alphaDstFactor)
         {
             Enable         = enable;
-            ConstantBlend  = constantBlend;
+            BlendConstant  = blendConstant;
             ColorOp        = colorOp;
             ColorSrcFactor = colorSrcFactor;
             ColorDstFactor = colorDstFactor;
--- a/Ryujinx.Graphics.GAL/BlendDescriptor.cs
+++ b/Ryujinx.Graphics.GAL/BlendDescriptor.cs
@@ -4,6 +4,7 @@ namespace Ryujinx.Graphics.GAL
     {
         public bool Enable { get; }
 
+        public ColorF      ConstantBlend  { get; }
         public BlendOp     ColorOp        { get; }
         public BlendFactor ColorSrcFactor { get; }
         public BlendFactor ColorDstFactor { get; }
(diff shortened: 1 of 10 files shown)


Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -138,6 +138,16 @@ private void InitializeDefaultState()
             {
                 _backingMemory[(int)MethodOffset.RtColorMask + index] = 0x1111;
             }
+
+            // Default blend states
+            var defaultBlendStateCommon = MemoryMarshal.Cast<BlendStateCommon, int>(MemoryMarshal.CreateReadOnlySpan(ref BlendStateCommon.Default, 1));
+            defaultBlendStateCommon.CopyTo(_backingMemory.AsSpan().Slice((int)MethodOffset.BlendStateCommon, defaultBlendStateCommon.Length));
```
mageven, line 144: Is there a better way to do this?

Ryujinx.Graphics.GAL/ColorF.cs
```diff
@@ -14,5 +14,11 @@ public ColorF(float red, float green, float blue, float alpha)
             Blue  = blue;
             Alpha = alpha;
         }
+
+        public static bool operator ==(ColorF l, ColorF r) => l.Red == r.Red &&
```
mageven, line 18: Didn't use epsilon comparisons here. I hope that's ok

Ryujinx.Graphics.Gpu/State/GpuState.cs
gdkchan: I personally would prefer if a `Set` generic method was added, and then it could just use that to write the BlendState struct to the respective offsets.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -35,6 +35,8 @@ class Pipeline : IPipeline, IDisposable
 
         private bool _scissor0Enable = false;
 
+        ColorF _constantBlend = new ColorF(0,0,0,0);
```
gdkchan, line 38: nit: Space after comma.

Ryujinx.Graphics.GAL/ColorF.cs
gdkchan, line 18: If you are overriding those operators, I think you should override `Equals` and `GetHashCode`. Equals would do the same thing as the `==` operator, and `==` could just call `Equals`, while `!=` would do something like `!l.Equals(r)`. `GetHashCode` can just use `HashCode.Combine(Red, Green, Blue, Alpha)`.

Ryujinx.Graphics.GAL/BlendDescriptor.cs
```diff
@@ -4,6 +4,7 @@ public struct BlendDescriptor
     {
         public bool Enable { get; }
 
+        public ColorF      ConstantBlend  { get; }
```
gdkchan, line 7: Imo `BlendColor` would be a better name. Or maybe `BlendConstant`.

Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -269,6 +277,31 @@ public bool QueryModified(MethodOffset m1, MethodOffset m2, MethodOffset m3, Met
             return modified;
         }
 
+        public bool QueryModified(
```
riperiperi, line 280: Add a summary for this, all the others have one. It's a little redundant, but it will appear when doing autocomplete or hovering function uses.

mageven: Even double-checking wasn't enough. I still missed it.

I wish a params approach was as fast as a hardcoded one. Or at least if you could use other types in generics like in C++.

mageven:
Wish there was a cleaner way to suppress those warnings :\

PR #1131, opened 2020-04-17
Adding .editorconfig so code style can be consistent and localized

ameisen:
Presently, developers have to change their code style for Visual Studio globally for the project, or add a default `.editorconfig` and have it set to be ignored by git, if they have custom C# code styles.

By adding the default `.editorconfig` to the solution, Visual Studio will automatically use the included code style settings for the solution.

It also appears that Visual Studio corrected a few errors in the solution file.


ameisen:
Just noting that this is the default `.editorconfig` as provided from [the Documentation Site formerly known as MSDN](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2019#example-editorconfig-file).

PR #1134, opened 2020-04-18
Indicate more clearly the location of the Ryujinx folder on Windows.

Tom60chat:
Because it is not easy to guess where is the Ryujinx folder.

### DIFF
--- a/KEYS.md
+++ b/KEYS.md
@@ -21,6 +21,8 @@ To dump your `prod.keys` and `title.keys` please follow these following steps.
 11. Copy these files and paste them in `Ryujinx/system`.
 And you're done!
 
+For Windows `Ryujinx/system` is located in `%appdata%/Ryujinx/system`.
+
 ## Title keys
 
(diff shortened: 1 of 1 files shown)


AcK77:
The Ryujinx folder is already explained in the Readme.md and you can open the folder through the GUI so I don’t think it’s necessary to add this again in the Keys.md file.

Tom60chat:
Yep, I did not read the Readme file :/
I did not find the button in the GUI.

ameisen:
Just a thought, not sure where to put the thought exactly, but Ryujinx should have two directories for its state - one for `Roaming` and the other for `Local`. Once we have things like file caches (which I have locally) you don't want those in `Roaming` as they won't necessarily be compatible across systems.

AcK77:
@Tom60chat you can open the Ryujinx folder in the GUI through the menu File > Open Ryujinx Folder.

@ameisen don’t hesitate to open an issue instead of speaking of that here, because things will be forgotten if not. Thanks!

Because of my previous message, I close this PR!

PR #1135, opened 2020-04-18
Improve IRDumper

FICTURE7:
Improves IRDumper to now dump:

MemoryOperand
```wasm
i64 r0 = Load i64 [i64 r1 + i64 r0*8]
```

IntrinsicOperation
```wasm
v128 v1 = Extended.X86Shufps v128 v1, v128 v0, i32 0xF0
```

Jump Table & Dynamic Table
```wasm
i64 %299 = Load i64 0x1DFFBB80018 ;; JMP_TABLE_1+8
...
i64 %289 = Copy i64 0x1DFFCB80010 ;; DYN_TABLE_1
i64 %295 = Copy i64 0x1DFFCB80018 ;; DYN_TABLE_1+8
```

Function names
```wasm
i64 %794 = Copy i64 0x7FFA30A038C4 ;; GetFunctionAddress
```

Multiple destinations
```wasm
i32 r0, i32 r3, i32 r1, i32 r2 = CpuId i32 r0
```

### DIFF
--- a/ARMeilleure/Diagnostics/IRDumper.cs
+++ b/ARMeilleure/Diagnostics/IRDumper.cs
@@ -61,13 +61,15 @@ namespace ARMeilleure.Diagnostics
 
                     string instName = string.Empty;
 
+                    var symbols = new List<(ulong Value, string Name)>();
+
                     if (node is PhiNode phi)
                     {
                         for (int index = 0; index < sources.Length; index++)
                         {
                             string phiBlockName = GetBlockName(phi.GetBlock(index));
 
-                            string operName = GetOperandName(phi.GetSource(index), localNames);
+                            string operName = GetOperandName(phi.GetSource(index), localNames, symbols);
 
                             sources[index] = $"({phiBlockName}: {operName})";
                         }
@@ -78,7 +80,7 @@ namespace ARMeilleure.Diagnostics
                     {
                         for (int index = 0; index < sources.Length; index++)
                         {
-                            sources[index] = GetOperandName(operation.GetSource(index), localNames);
+                            sources[index] = GetOperandName(operation.GetSource(index), localNames, symbols);
                         }
 
                         instName = operation.Instruction.ToString();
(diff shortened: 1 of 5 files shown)


ARMeilleure/Diagnostics/Symbols.cs
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+namespace ARMeilleure.Diagnostics
+{
+    static class Symbols
+    {
+#if M_DEBUG
+
+        private static readonly ConcurrentDictionary<ulong, string> _symbols;
+        private static readonly List<(ulong Start, ulong End, ulong ElementSize, string Name)> _rangedSymbols;
```
gdkchan, line 13: Imo this should use a struct instead of the large tuple.

ARMeilleure/Diagnostics/Symbols.cs
```diff
@@ -0,0 +1,62 @@
+﻿#if M_DEBUG
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+#endif
+
+namespace ARMeilleure.Diagnostics
+{
+    static class Symbols
+    {
+#if M_DEBUG
```
gdkchan, line 10: I don't think puting everything under `M_DEBUG` is necessary. You could just add `[Conditional("M_DEBUG")]` to the `Add` method since it is the one that have most impact. All other functions accessing this are already meant to be used for debugging anyway.

ARMeilleure/Diagnostics/IRDumper.cs
```diff
@@ -141,7 +148,40 @@ private static string GetOperandName(Operand operand, Dictionary<Operand, string
             }
             else if (operand.Kind == OperandKind.Constant)
             {
-                name = "0x" + operand.Value.ToString("X");
+                name = Symbols.Get(operand.Value);
```
gdkchan, line 151: This can have some false positives, and in this case we wouldn't be able to see the actual value. So it might be better to show the hex value first, and possible symbol inside `(` `)`.

gdkchan: Could also put it as comment, for example:
```
i64 r2 = Copy i64 0xABCD0123 ; GetFunctionAddress
```

FICTURE7:
If you wish, I can do a more extensive refactoring of IRDumper.

gdkchan:
> 
> 
> If you wish, I can do a more extensive refactoring of IRDumper, to leverage StringBuilder and such.

Sure, why not.

FICTURE7:
> I think making the IR dumper a struct is a bit weird

Indeed, making it a struct just feels off. I initially had an inner `DumperContext` struct, but I thought it was unnecessary since might as well just make `IRDumper` the struct.

If you guys want to, I can switch back to the `DumperContext` thing and turn `IRDumper` into a static class again.

Or simply make `IRDumper` a class. :>

gdkchan:
Imo just make it a class.

PR #1147, opened 2020-04-22
Add support for dynamic docking/undocking

Thog:
As SurfaceFlinger is now working more accurately, we can now support
dynamic configuration of docking mode :)

### DIFF
--- a/Ryujinx.HLE/HOS/Horizon.cs
+++ b/Ryujinx.HLE/HOS/Horizon.cs
@@ -43,6 +43,7 @@ using NsoExecutable      = Ryujinx.HLE.Loaders.Executables.NsoExecutable;
 using static LibHac.Fs.ApplicationSaveDataManagement;
 using Ryujinx.HLE.HOS.Services.Nv;
 using Ryujinx.Common;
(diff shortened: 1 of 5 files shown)


jduncanator:
As a further extension to @riperiperi's comment, some games with endlessly display the Controller Applet if the user has an invalid controller specified in the settings, and they put the emulator in docked mode. With out current applet implementation (which is more of a stub), this will end up with the game executing the applet _extremely_ quickly in a loop, potentially freezing the entire emulator.

I'm not sure how to best address that, but it's something to be aware of.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -42,6 +42,8 @@
 
 using static LibHac.Fs.ApplicationSaveDataManagement;
 using Ryujinx.HLE.HOS.Services.Nv;
+using Ryujinx.Common;
```
gdkchan, line 45: Sort usings.

Thog:
@riperiperi & @jduncanator Will create an issue about the UI side so we can address this later

PR #1151, opened 2020-04-23
Use the official JSON parser

Thog:
This remove Utf8son and JsonPrettyPrinter dependencies.

NOTE: the standard JSON parser doesn't support configurable
indentation, as a result, all the pretty printed JSON are indented with 2
spaces.


Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -519,49 +518,47 @@ public void LoadNca(Nca mainNca, Nca patchNca, Nca controlNca)
...
+            if (File.Exists(titleUpdateMetadataPath))
             {
-                using (Stream stream = File.OpenRead(Path.Combine(Device.FileSystem.GetBasePath(), "games", mainNca.Header.TitleId.ToString("x16"), "updates.json")))
+                string updatePath = JsonHelper.DeserializeFromFile<TitleUpdateMetadata>(titleUpdateMetadataPath).Selected;
+
+                if (File.Exists(updatePath))
                 {
-                    IJsonFormatterResolver resolver = CompositeResolver.Create(StandardResolver.AllowPrivateSnakeCase);
-                    string updatePath = JsonSerializer.Deserialize<TitleUpdateMetadata>(stream, resolver).Selected;
+                    FileStream file = new FileStream(updatePath, FileMode.Open, FileAccess.Read);
```
gdkchan, line 529: Looks like file is not disposed, is that intentional?

Thog: this was already like that before so I'm not sure

emmauss: you should probably dispose it.

Xpl0itR: Don't dispose. That's the update file, if you dispose it then it can't be read from.

Thog:
Rebased, ready to merge.

PR #1155, opened 2020-04-24
Fix kernel memory allocator block coalescing

gdkchan:
This should fix the block coalescing implementation. The previous implementation was incomplete, missing the `count >= 64`. I probably thought it was some sort of compiler optimization that could be omitted at the time. I rechecked the kernel implementation and implemented the missing parts.

Also thanks to SciresM for mesosphere, an open source implementation of Horizon kernel that helped me double check that the memory allocator implementation was correct.

This should fix `OutOfMemory` error being returned by some games that does allocation frequently using (Un)MapPhysicalMemory syscalls.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryRegionBlock.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryRegionBlock.cs
@@ -12,30 +12,77 @@ namespace Ryujinx.HLE.HOS.Kernel.Memory
         public int   Order;
         public int   NextOrder;
 
-        public bool TryCoalesce(int index, int size)
+        public bool TryCoalesce(int index, int count)
         {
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Kernel/Memory/KMemoryRegionBlock.cs
```diff
@@ -12,30 +12,77 @@ class KMemoryRegionBlock
...
+                while (remaining != 0);
+
+                remaining = count;
+                tempIdx = index;
+
+                do
+                {
+                    Masks[MaxLevel - 1][tempIdx++] = 0;
+
+                    for (int level = MaxLevel - 2; level >= 0; level--, index /= 64)
```
gdkchan, line 44: This should be extracted to a separate function to improve readability and reduce code duplication, there are a bunch of other occurrences on the other file aswell, however I think this can be done later.

PR #1121, opened 2020-04-16
Update .NET Core to 3.1, and update NuGet Packages

ameisen:
This updates the .NET Core framework used from 3.0 to 3.1, and also updates all NuGet packages to their latest non-preview versions.

.NET Core 3.0 support ended March 3, 2020.

PR #1164, opened 2020-04-25
Revert change to package include condition

Xpl0itR:
#1121 made a change to the package include condition, this PR reverts that change which fixes the issue where visual studio builds were missing dependencies.


gdkchan:
Can you explain why this change fixes it? Because to me both should be equivalent.

Xpl0itR:
Because pressing F5 in visual studio does not specify a target platform

PR #1167, opened 2020-04-26
Implement Counter Queue and Partial Host Conditional Rendering

riperiperi:
image

Emulating GPU counters is a bit of a challenge, given all that host APIs give us is a way to get deltas in the counter value over a given range. There's no way to get the value in the middle, so it multiple queries need to be started, stopped and reported to update the emulated counter value. Right now this is done entirely synchronously - we wait on the query value being reported and update the counter value instantly. Unfortunately, this stalls the entire render thread.

Rather than calculating and flushing counters instantly, this PR creates report "events" in a queue, which are processed and flushed to memory by a background thread. Additionally, these events can directly feed the associated queries to host conditional rendering to avoid stalling for the query result on the CPU, under select circumstances.

Greatly improves performance in Super Mario Odyssey and Splatoon 2. It's of note that the performance was in Ryujinx all along, it just wasn't very good at showing it. 😌 

This will affect all games that use conditional rendering in some capacity, so I'd recommend testing a lot of games.

## Approach

Each counter type has its own CounterQueue, which represents a number of events which will influence the value of the gpu counter. Each counter queue stores its immediate counter value privately, but they are modified and reset by the events. Each has a "current" event being built, which represents an opengl query that has been started, but not ended or reported.

Each `ReportCounters` method call requests an event on one of these queues. When requested, the "current" event is ended, and opengl is told to asynchronously write the result to a buffer when it is ready. This event is returned, and indicates the _point_ at which `ReportCounters` was called, letting us use the intermediate result at that point by either flushing it or using it in a host query.

The `ResetCounters` method calls simply tell the "current" event on the relevant queue to restart its query (end then begin), and set a flag to indicate that the counter should be reset to 0 when the event is consumed.

The query for each event reports its result to a Buffer, which is directly mapped into host memory (glBufferStorage + glMapBufferRange) with a Persistent flag. We initially set this to -1 to indicate that it's not ready yet, and the GPU updates it with the final value when it's ready.

A background consumer thread goes through the events on the queue, signalling their results when they are ready. This is done by "consuming" events, which adds their query result onto the counter (optionally resetting to 0 before doing so), signals the `OnResult` event with the counter value at this point, and returns the query object back to a pool to be used by future events.

The Gpu's OnResult event handler flushes the new counter value and timestamp to memory, as long as the region has not been unmapped. For counter type Zero, this whole event malarky is skipped and it just reports the timestamp immediately.

### Host Conditional Rendering
Since we have conditional rendering now, my life was made a little harder. Fortunately, we already had a query cache, which presumably exists precicely for this reason. Events are added to the cache with their gpuVa as a key, and are available when we search for counters to do conditional rendering with. Before, we would expect the counter values to be in memory, as they were processed synchronously in ReportCounters. Now, we either have to flush them to cpu before doing the comparison (slow!) or try to get OpenGL to use the query object directly.

OpenGL can only natively do conditional rendering with a very specific type of query - `SamplesPassed != 0`, or `SamplesPassed == 0` with the Inverted modes. Additionally, since OpenGL queries over a range rather than a running counter, we need the event to fully represent the counter's value without having to add the results of others - it _must_ have the reset flag set. Unsurprisingly, lots of games hit this case, which we can speed up tremendously by just using the query object rather than flushing to CPU.

It may be possible to massage conditional rendering into using more complicated comparisons using a draw call with the counter values as input. I'll leave that as an exercise for the reader - for the cases this approach covers, it is faster anyways.

### Things that didn't work and why
This isn't the cleanest solution that exists, but it is by far the fastest. In order of appearance, all of the issues I encountered:
- Query Objects are not properly sharable, so you can't wait on them on a background thread with a shared context.
- Sync objects are a very clean way to signal the consumer thread when the query is available. However, when I tried to use these, the cost of `glFenceSync` essentially made it only slightly better than not trying at all. `glDeleteSync` also burned a lot of CPU time, depending on where it was placed.
- glBufferSubData and friends had various issues - it seemed to generally miss results, fail to set the uninizialized value at the right time and was generally slow. Also, if we were to busy loop for flushing results ASAP, that would be busy looping on a gl call, which I think would cause problems.
- Busy looping is an option if we need more tight timing, but I disabled it to save some cpu cycles. I have disabled it for now since the flagship ruin-your-life game Splatoon 2 works just fine with a 1ms back-off. It is possible to force wait on flushes from another thread if you really need a result, so it may be possible to add that in future. Note that the thread would not busy loop while it has nothing to wait on.

## Additional Changes
- The VaUnmapped event is now signalled before the actual unmapping occurs. This is a little more useful for finding resources that "forgot" their gpuVa.

## Special Mention
Thanks goes to @ReinUsesLisp for suggesting processing queries in another thread, and pointing out that while Query Objects aren't sharable, Buffers you can write their results into are.


EmulationChannel:
SharedScreenshot
@riperiperi ❤❤❤❤❤❤❤❤❤❤❤

Ryujinx.Graphics.OpenGL/Queries/BufferedQuery.cs
```diff
@@ -0,0 +1,96 @@
...
+            GL.BindBuffer(BufferTarget.QueryBuffer, _buffer);
+
+            Marshal.WriteInt64(_bufferMap, -1L);
+            GL.GetQueryObject(Query, GetQueryObjectParam.QueryResult, (long*)0);
+        }
+
+        public bool TryGetResult(out long result)
+        {
+            result = Marshal.ReadInt64(_bufferMap);
+            if (result == DefaultValue)
```
gdkchan, line 59: Why not just `return result != DefaultValue;`?

Ryujinx.Graphics.OpenGL/Queries/BufferedQuery.cs
```diff
@@ -0,0 +1,96 @@
...
+                data = Marshal.ReadInt64(_bufferMap);
+                if (data == DefaultValue)
+                {
+                    Thread.Sleep(1);
+                }
+            }
+
+            if (iterations >= MaxQueryRetries)
+            {
+                Common.Logging.Logger.PrintError(Common.Logging.LogClass.Gpu, $"Error: Query result timed out! Took more than {MaxQueryRetries}ms.");
```
gdkchan, line 82: Are you sure that the amount of tries will be really equivalent to the time it spent in ms? I get that it sleeps for 1 ms, but I think the time would be a bit off. I'd just call it "tries" instead of "ms".

Ryujinx.Graphics.OpenGL/Queries/CounterQueue.cs
```diff
@@ -0,0 +1,203 @@
...
+            lock (_lock)
+            {
+                if (_queryPool.Count > 0)
+                {
+                    BufferedQuery result = _queryPool.Dequeue();
+                    return result;
+                }
+                else
+                {
+                    Common.Logging.Logger.PrintWarning(Common.Logging.LogClass.Gpu, $"Increasing query pool size: {++totalQuery}");
```
gdkchan, line 85: I would prefer adding `using Ryujinx.Common.Logging;`.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1016,6 +1017,52 @@ public void RestoreRasterizerDiscard()
             }
         }
 
+        public bool TryHostConditionalRendering(object value, object compare, bool isEqual)
```
gdkchan, line 1020: I'm not a huge fan of allowing any object to be passed here, but not sure if there's a better alternative.

riperiperi: Yeah, my dream parameter type was (ICounterEvent | ulong), and then checking, but this isn't typescript. The alternative is overrides for (ICounterEvent, ulong) and (ICounterEvent, ICounterEvent), but the caller has to put them in the right order.

I guess since the only situation where the order is unknown is when the condition isn't comparing with zero, so moving the ordering to the caller might be a cleaner way of doing it.

Ryujinx.Graphics.OpenGL/Queries/BufferedQuery.cs
riperiperi: This originally did something else. I'll clean it up.

Ryujinx.Graphics.OpenGL/Queries/BufferedQuery.cs
riperiperi: That is true, will change that.

Ryujinx.Graphics.OpenGL/Queries/CounterQueue.cs
riperiperi: This isn't actually meant to be here, it was for debugging. 😞

Ryujinx.Graphics.OpenGL/Queries/BufferedQuery.cs
riperiperi: I'll add `using Ryujinx.Common.Logging;` for this one, too.

Ryujinx.Graphics.OpenGL/Pipeline.cs
riperiperi: I've changed this to use 2 overloads, let me know what you think.

gdkchan: Looks good now.

riperiperi:
I added a new method to poll the counters manually (just wakes the consumer threads), which is called when a syncpoint is incremented. Appears to fix some regressions where certain queries worked intermittently in BOTW. (object in water check, foot planting inverse kinematics)

This isn't a perfect solution by a long shot, but it's good enough for now. With read actions in future, we can poll the completed queries when the cpu attempts to read the value, rather than using this heuristic.

(also re-ordered the counter functions in IRenderer)

jduncanator:
I've marked as do not merge, because I think this needs a bit more testing. I will do some over the next couple of days, but would be nice to get a list of games you've already tested for regressions.

ReaperOfSouls1909:
doing major testing now on all my titles

jduncanator:
Matt did a bunch of testing (thanks Matt!), and I'm satisfied that this doesn't introduce any obvious regressions!

PR #1197, opened 2020-05-02
nvservice: add a lock around NvHostEvent and remove release fence on SFv2

Thog:
Classic mistake.. this adds a lock inside NvHostEvent and on the events array as those could be manipulated asynchronously.

This should fix some weird race condition happening (example: event id 0 sometime being invalid between 7PM and 8PM on AC:NH when starting, ect)


Thog:
Users reported that this fix the regression caused by https://github.com/Ryujinx/Ryujinx/commit/81cba3c3df7a1534c6690f69de025beeb6ebd1b9 on AC:NH.

This seems to also have a good performance impact on games.

PR #1198, opened 2020-05-02
Fix depth stencil formats copy by matching equivalent color formats

gdkchan:
The 2D engine does not support depth-stencil formats, so equivalent color formats are used for the copy. This modifies the compatibility check to match equivalent color formats when searching for copy textures.

Should fix broken fog on Pokémon Sword/Shield.

I recommend testing on a few games to ensure no regressions. It might be also worth revisiting this in the future, I don't like this approach to be honest.


riperiperi:
image
Someone on discord said that this happens on Link's Awakening on this PR.

gdkchan:
Rebased.
#1365 seems to have fixed the regression on ZLA, so this is ready for review again.

Doctorwho1909:
am going to test now

Doctorwho1909:
image

Doctorwho1909:
image
Working needed tp purge jit cache

PR #1203, opened 2020-05-03
Refactor shader GPU state and memory access

gdkchan:
This adds a interface that is used to access GPU state and read memory from the shader translator. Eliminates the need to get a huge span of the mapped memory and pass that to the translator. Should fix exceptions throw in some cases when translating shaders, and part of the range was not mapped.

Note that this may now throw when reading a unmapped region on a broken shader. Before it would print a warning and give up.


Thog:
Profiler builds seems broken

gdkchan:
The problem was not profiled builds, the NVDEC project was not compiling (which I didn't notice since it doesn't build when Ryujinx is build anymore). Anyway it's fixed now (also fixed a bug on it).

Ryujinx.Graphics.Gpu/Shader/GpuAccessor.cs
```diff
@@ -0,0 +1,264 @@
...
+            _state = state;
+            _stageIndex = stageIndex;
+        }
+
+        /// <summary>
+        /// Creates a new instance of the GPU state accessor for compute shader translation.
+        /// </summary>
+        /// <param name="context">GPU context</param>
+        /// <param name="state">Current GPU state</param>
+        /// <param name="localSizeX">Local group size X of the computer shader</param>
```
riperiperi, line 42: Computer shader?

PR #1205, opened 2020-05-03
Add Docked/Handheld text on status bar

HorrorTroll:
Since we have dynamic Docked/Handheld already, so i try to add Docked/Handheld text on status bar. 

The preview is on this gif (it was an old gif, but the function is the same)
ezgif com-optimize

### DIFF
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -326,10 +326,7 @@ namespace Ryujinx.Ui
                     _device.ProcessFrame();
                 }
 
-                string dockedMode = "Unknown";
-
-                if (ConfigurationState.Instance.System.EnableDockedMode == true)  dockedMode = "Docked";
-                if (ConfigurationState.Instance.System.EnableDockedMode == false) dockedMode = "Handheld";
+                string dockedMode = ConfigurationState.Instance.System.EnableDockedMode ? "Docked" : "Handheld";
 
                 if (_ticks >= _ticksPerFrame)
                 {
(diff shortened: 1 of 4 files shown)


Ryujinx/Ui/GLRenderer.cs
```diff
@@ -326,6 +326,11 @@ public void Render()
                     _device.ProcessFrame();
                 }
 
+                string dockedMode = "Unknown";
+
+                if (ConfigurationState.Instance.System.EnableDockedMode == true)  dockedMode = "Docked";
+                if (ConfigurationState.Instance.System.EnableDockedMode == false) dockedMode = "Handheld";
```
AcK77, line 332: This should be simplified by a ternary condition:

```csharp
string dockedMode = ConfigurationState.Instance.System.EnableDockedMode ? "Docked" : "Handled";
```

jduncanator:
I think this should go next to Vsync, which is our other "dynamically toggle-able setting".

HorrorTroll:
Already simplified & move next to V-Sync like suggest above
image

Ryujinx/Ui/StatusUpdatedEventArgs.cs
```diff
@@ -5,16 +5,19 @@ namespace Ryujinx.Ui
...
 
-        public StatusUpdatedEventArgs(bool vSyncEnabled, string hostStatus, string gameStatus, string gpuName)
+        public StatusUpdatedEventArgs(bool vSyncEnabled, string dockedMode, string hostStatus, string gameStatus, string gpuName)
         {
             VSyncEnabled = vSyncEnabled;
+            DockedMode   = dockedMode;
             HostStatus   = hostStatus;
             GameStatus   = gameStatus;
             GpuName      = gpuName;
+
```
Thog, line 20: nit useless space here

HorrorTroll: Done 👍

PR #1211, opened 2020-05-04
Improve ShaderDumper

plutooo:
This refactors + changes ShaderDumper. It had to move to Graphics.Shader, for visibility reasons.

It now dumps the following formats:
* Raw shader binary
* IntermediateRepresentation (after CFG, after SSA, after Optimizer, after Lowerer)
* StructuredIR
* GLSL


jduncanator:
Personally I'd prefer to see these changes added to ShaderTools, rather than to core Ryujinx.

A big use case for the shader dumper is offline iteration on the shader decompiler. With the proposed changes, one would have to boot the game, and trigger the use of the shader they cared about each time to see the IR or GLSL dump. It's a much more common use case to dump the shader in binary format that you're interested in once, and then run it iteratively through Ryujinx.ShaderTools to add or develop features in the shader decompiler.

Perhaps the APIs aren't there yet for the shader decompiler to allow that in ShaderTools, but I think that would be a better approach over all.

plutooo:
> Personally I'd prefer to see these changes added to ShaderTools, rather than to core Ryujinx.
> 
> A big use case for the shader dumper is offline iteration on the shader decompiler. With the proposed changes, one would have to boot the game, and trigger the use of the shader they cared about each time to see the IR or GLSL dump. It's a much more common use case to dump the shader in binary format that you're interested in once, and then run it iteratively through Ryujinx.ShaderTools to add or develop features in the shader decompiler.
> 
> Perhaps the APIs aren't there yet for the shader decompiler to allow that in ShaderTools, but I think that would be a better approach over all.

Either way, I think we need to introduce these changes. Without these changes, there would be no way for ShaderTools to hook Translator to get the intermediate IR stages.

Cleaned up the code some.

Ryujinx.Graphics.Shader/Translation/Translator.cs
```diff
@@ -5,6 +5,8 @@
 using Ryujinx.Graphics.Shader.Translation.Optimizations;
 using System;
 using System.Collections.Generic;
+using Ryujinx.Common;
```
gdkchan, line 8: nit: Sort usings.

Thog:
Closed for inactivity, will add one somedays

PR #1236, opened 2020-05-11
Refactor out Application details from Horizon

mageven:
All application details (and its loading) have been refactored out into `ApplicationLoader`. 
Since the emulator (and the Switch?) only maintains one `Application` at a time, the loader itself stores app info.
Hopefully, this also brings the project a step closer to implementing it via services.

I'd like an opinion on adding `ApplicationLoader` to `Switch` and not `Horizon`. Although the kernel maintains everything, I felt it's not part of `Horizon` class as such.


Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+    public class ApplicationLoader
+    {
+        private readonly Switch _device;
+        private readonly ContentManager _contentManager;
+        private readonly VirtualFileSystem _fileSystem;
+
+        public IntegrityCheckLevel FsIntegrityCheckLevel => _device.System.FsIntegrityCheckLevel;
+
+        public ulong TitleId {get; private set; }
```
gdkchan, line 35: nit: `{get;` -> `{ get;`.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+        public void LoadXci(string xciFile)
+        {
+            FileStream file = new FileStream(xciFile, FileMode.Open, FileAccess.Read);
+
+            Xci xci = new Xci(_fileSystem.KeySet, file.AsStorage());
+
+            (Nca mainNca, Nca patchNca, Nca controlNca) = GetXciGameData(xci);
+
+            if (mainNca == null)
```
gdkchan, line 79: Maybe we should handle errors consistently here. For example, `GetXciGameData` can throw `InvalidDataException`. Maybe it's better to catch it here and also print the error on the log? In that case it would be probably useful to have a more descriptive error message.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+            Result result = controlFs.OpenFile(out IFile controlFile, "/control.nacp".ToU8Span(), OpenMode.Read);
+
+            if (result.IsSuccess())
+            {
+                result = controlFile.Read(out long bytesRead, 0, ControlData.ByteSpan, ReadOption.None);
+
+                if (result.IsSuccess() && bytesRead == ControlData.ByteSpan.Length)
+                {
+                    TitleName = ControlData.Value
+                        .Titles[(int) _device.System.State.DesiredTitleLanguage].Name.ToString();
```
gdkchan, line 175: nit: Space after `(int)` is not needed.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+            LoadNca(nca, null, null);
+        }
+
+        public void LoadNsp(string nspFile)
+        {
+            FileStream file = new FileStream(nspFile, FileMode.Open, FileAccess.Read);
+
+            PartitionFileSystem nsp = new PartitionFileSystem(file.AsStorage());
+
+            foreach (DirectoryEntryEx ticketEntry in nsp.EnumerateEntries("/", "*.tik"))
```
gdkchan, line 207: The code here is very similar to the one in `GetXciGameData`. It would be nice if we could change `GetXciGameData` a bit to allow this method to also use it, and avoid some code repetition.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+                Logger.PrintWarning(LogClass.Loader, "NPDM file not found, using default values!");
+
+                metaData = GetDefaultNpdm();
+            }
+            else
+            {
+                metaData = new Npdm(npdmFile.AsStream());
+            }
+
+            List<IExecutable> staticObjects = new List<IExecutable>();
```
gdkchan, line 395: I would rename the list to `nsos` or something. `staticObjects` is not really the correct name. (I know it's a existing issue, just pointing it out).

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+            ProgramLoader.LoadNsos(_device.System.KernelContext, metaData, staticObjects.ToArray());
+        }
+
+        public void LoadProgram(string filePath)
+        {
+            Npdm metaData = GetDefaultNpdm();
+
+            bool isNro = Path.GetExtension(filePath).ToLower() == ".nro";
+
```
gdkchan, line 434: Double empty line.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+            IExecutable staticObject;
```
gdkchan, line 436: Same comment about `staticObject`, and this is a NRO.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+                staticObject = new NsoExecutable(new LocalStorage(filePath, FileAccess.Read));
+            }
+
+            _contentManager.LoadEntries(_device);
+
+            TitleName    = metaData.TitleName;
+            TitleId      = metaData.Aci0.TitleId;
+            TitleIs64Bit = metaData.Is64Bit;
+
+            ProgramLoader.LoadNsos(_device.System.KernelContext, metaData, new IExecutable[] { staticObject });
```
gdkchan, line 523: It's possible to avoid explicitly creating an array here by changing the last param of the `LoadNsos` method to `params` (that will allow it to accept a variable number of NSOs without explicitly creating and passing an array).

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+        private Result EnsureSaveData(TitleId titleId)
+        {
+            Logger.PrintInfo(LogClass.Application, "Ensuring required savedata exists.");
+
+            Uid user = _device.System.State.Account.LastOpenedUser.UserId.ToLibHacUid();
+
+            ref ApplicationControlProperty control = ref ControlData.Value;
+
+            if (LibHac.Util.IsEmpty(ControlData.ByteSpan))
```
gdkchan, line 544: Maybe this can be just `Util.IsEmpty` since theres a LibHac using?

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,577 @@
...
+
+                Logger.PrintWarning(LogClass.Application,
+                    "No control file was found for this game. Using a dummy one instead. This may cause inaccuracies in some games.");
+            }
+
+            FileSystemClient fs = _fileSystem.FsClient;
+
+            Result rc = fs.EnsureApplicationCacheStorage(out _, titleId, ref control);
+
+            if (rc.IsFailure())
```
gdkchan, line 562: Maybe this should return early in case of failure? Otherwise this method can return success even if one of the operations failed.

Ryujinx.HLE/Switch.cs
```diff
@@ -27,6 +27,8 @@ public class Switch : IDisposable
 
         public Horizon System { get; private set; }
 
+        public ApplicationLoader Application {get; private set;}
```
gdkchan, line 30: `{get;` -> `{ get;`. Also we don't need the `private set` when it is only set on the constructor, although that's a existing issue.

Ryujinx.HLE/HOS/ApplicationLoader.cs
mageven: Added a common function for both XCI and NSP

Ryujinx.HLE/HOS/ApplicationLoader.cs
mageven: Added a common function for both XCI and NSP

mageven:
The task of refactoring `ContentManager` and `ApplicationLoader` seems to be never-ending as there's tons of duplicate code.
In the interest of finishing my original PRs (and bothering for re-reviews), I'll stop here. 
Probably better to work on it after they are merged for a proper design anyway.

Also, doesn't this PR need more reviewers requested?

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,537 @@
...
+                ReadControlData(controlNca);
+            }
+            else
+            {
+                ControlData.ByteSpan.Clear();
+            }
+
+            if (TitleId != 0)
+            {
+                EnsureSaveData(new TitleId(TitleId));
```
Thealexbarney, line 303: The result of EnsureSaveData isn't checked

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -0,0 +1,537 @@
...
+                _fileSystem.LoadRomFs(romFsFile);
+            }
+
+            LocalFileSystem codeFs = new LocalFileSystem(exeFsDir);
+
+            LoadExeFs(codeFs, out _);
+
+            if (TitleId != 0)
+            {
+                EnsureSaveData(new TitleId(TitleId));
```
Thealexbarney, line 68: The result of EnsureSaveData isn't checked

Ryujinx.HLE/HOS/ApplicationLoader.cs
mageven: `Result` failures are already logged here:
https://github.com/Ryujinx/Ryujinx/pull/1236/files#diff-09dfcda602c4bda66d1e322dbaa88160R522
https://github.com/Ryujinx/Ryujinx/pull/1236/files#diff-09dfcda602c4bda66d1e322dbaa88160R531

It didn't make sense to stop the application load so I left the old code as is.

PR #1238, opened 2020-05-11
Unwinding Follow-up. Fix a bug in JitUnwindWindows where ...

LDj3SNuD:
... in case of "Vector" unwind codes the remaining unwind codes could be corrupted.
Nits.

### DIFF
--- a/ARMeilleure/Translation/JitUnwindWindows.cs
+++ b/ARMeilleure/Translation/JitUnwindWindows.cs
@@ -1,12 +1,15 @@
-using ARMeilleure.IntermediateRepresentation;
+// https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/build/exception-handling-x64.md
+
+using ARMeilleure.CodeGen.Unwinding;
 using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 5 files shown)


ARMeilleure/Translation/JitUnwindWindows.cs
```diff
@@ -93,59 +96,80 @@ public static void InstallFunctionTableHandler(IntPtr codeCachePointer, uint cod
...
-
-                    _unwindInfo->UnwindCodes[codeIndex++] = uwop;
-                    _unwindInfo->UnwindCodes[codeIndex++] = (ushort)spOffset;
-
-                    spOffset += 16;
+                    case UnwindPseudoOp.SaveXmm128:
+                    {
+                        int stackOffset = entry.StackOffsetOrAllocSize;
+
+                        Debug.Assert(stackOffset % 16 == 0);
```
gdkchan, line 116: I think nothing ensures that `stackOffset` is aligned to 16 bytes right now, for XMM registers that were saved on prolog.

gdkchan: Nevermind, it ensures that SP will be aligned to 16 bytes for the reserved stack region, so it should be fine.

PR #1246, opened 2020-05-14
Do not tail continue when next block is available

FICTURE7:
This is a follow up to #1235.

Before when `Branch` was null but `Next` was not, both path would become tailcall continues, with this patch it does not.

This reduces the amount of translations when falling through is common on conditional branches. Reduced 6546 translations to 5516 translations  which is -15.7%. In terms of dumps, reduces size of textual assembly dump from 265mb (the 6546 translations) to 261mb (the 5516 translations) which is -1.5%. This reduces the throughput and CQ gain from #1235.

This feels like a spooky change, and I may be missing some implementation details here and doing big doodoo.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitFlow.cs
+++ b/ARMeilleure/Instructions/InstEmitFlow.cs
@@ -105,11 +105,11 @@ namespace ARMeilleure.Instructions
             {
                 Operand lblFallthrough = Label();
 
-                // The condition is inverted such that the fallthrough block emitted below the taken block. Sometimes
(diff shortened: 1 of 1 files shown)


FICTURE7:
Looking back at this patch, I think this could be better. I think we can get some gain by using `context.GetLabel(address)` instead of using a "raw/unbound" label. But I'll have to check if this is actually the case. Moving this to draft.

btw, should these tail continues be marked with `allowRejit=true`?

FICTURE7:
Superseded by #1298 (hopefully).

PR #1223, opened 2020-05-06
Remember GUI column sort and separate hotkey settings from emulated input settings

Xpl0itR:
Fix #1214

PR #1247, opened 2020-05-14
Add hotkey to toggle Docked/Handheld

HorrorTroll:
I don't know which key is best to set right now. So i'm set it at F1 key

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationState.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationState.cs
@@ -549,7 +549,7 @@ namespace Ryujinx.Configuration
                         },
                         Hotkeys        = new KeyboardHotkeys
                         {
-                            ToggleVsync = Key.Tab
+                            ToggleVsync  = Key.Tab
                         }
                     }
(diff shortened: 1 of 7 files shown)


Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -548,7 +549,7 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
                         },
                         Hotkeys        = new KeyboardHotkeys
                         {
-                            ToggleVsync = Key.Tab
+                            ToggleVsync  = Key.Tab
```
riperiperi, line 552: What is this aligning with?

riperiperi: Also, this specific instance doesn't set a docked key.

HorrorTroll:
I don't know what to do for migrating, so to let hotkey work right now is open controller config and saved.

HorrorTroll:
Closed until #1223 done

PR #1256, opened 2020-05-17
Add Localization Support

emmauss:
Adds support for localizations.

All ui text can be translated, both static and dynamic. text are extracted using gnu's xgettext and saved in Ryujinx/Ui/locale/messages.pot.

Translations are in the po format(original) and mo format(compiled), and are stored in the following location; `Ryujinx/Ui/locale/<language>/LC_MESSAGES/ryujinx.po`, where `<language>` is the language code, eg "en", "ja", "en-US". The language codes canbe in either the 2 letter format("en") or the region format ("en-US").

Current completed Translations
1. French

This pr is ready for review.


emmauss:
Added French translation by @MelonSpeedruns . PR open for review

jduncanator:
We should be using the standard .NET way of doing translations, rather than using some other translation or localisation system. This allows us to port it between any other GUI system when we inevitably move away from GTK.

emmauss:
Rebased, updated config and added Spanish translation

knuxbbs:
>  This allows us to port it between any other GUI system when we inevitably move away from GTK.

What the problem with GTK? There is some plan to use another GUI system?

Edit a .resx file outside of Windows is still a fairly rudimentary task.

Thog:
As jD said, we should use something more standard in .NET for this.

Closing for inactivity.

PR #1270, opened 2020-05-22
Implement TMML and TMML.B

Thog:
This implement TMML and TMML.B shader instructions

### DIFF
--- a/Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
+++ b/Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
@@ -718,7 +718,7 @@ namespace Ryujinx.Graphics.Shader.Instructions
 
         public static void TmmlB(EmitterContext context)
         {
-            EmitTextureMipMapLevel(context, false);
+            EmitTextureMipMapLevel(context, true);
         }
 
         public static void Tmml(EmitterContext context)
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
```diff
@@ -716,6 +716,130 @@ Operand GetDest()
             }
         }
 
+        public static void TmmlB(EmitterContext context)
+        {
+            EmitTextureMipMapLevel(context, false);
```
gdkchan, line 721:
```suggestion
            EmitTextureMipMapLevel(context, true);
```

Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
```diff
@@ -716,6 +716,130 @@ Operand GetDest()
...
+
+            int handle = !isBindless ? op.Immediate : 0;
+
+            for (int compMask = op.ComponentMask, compIndex = 0; compMask != 0; compMask >>= 1, compIndex++)
+            {
+                if ((compMask & 1) != 0)
+                {
+                    Operand dest = GetDest();
+
+                    // Components z and w aren't standard, we return 0 in this case and add a comment
```
gdkchan, line 812:
```suggestion
                    // Components z and w aren't standard, we return 0 in this case and add a comment.
```

Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
Thog: OOPS

PR #1272, opened 2020-05-23
Memory Read/Write Tracking using Region Handles

riperiperi:
image

**This PR will be filled out with more information as the issues are resolved. It will be unset as WIP when they are too.**

Replaces the existing `QueryMemory` and page dirty flag system to one which tracks allows tracking of arbitrary ranges, separately, with read and write actions that can be performed immediately before a memory access. Each resource can obtain a tracking "handle" that provides access to memory read/write information.

Specifically, a handle currently has two features - a dirty flag and a read action. The dirty flag is set whenever a write is signalled in the region, and the read action is called (and consumed) when either a read or write occurs. This allows us to flush resources which aren't available on the CPU, such as modfied textures, immediately before they are needed.

The tracking is implemented via a memory protection system that's quite simple - when a virtual region contains one or more handles that are listening for a change, then it protects against writes. When this protection is triggered, then the memory tracking is called, the relevant region is found and all child handles are signalled that a write occurred. A similar thing is done for reads, where if a read action is present, the region is pretected from reads. Writes will also trigger read actions, as it is likely that the data present in the location is only being partially overwritten, and we want to flush to correctly represent both the new and old data. (eg. a texture with a pixel set by cpu every frame)

Benefits/Changes in games:
---
- Cost for checking if a texture has changed has reduced GREATLY - now just a single dirty flag check.
  - Performance almost doubles in BOTW in certain areas (16 -> 30 staying still), increases marginally in SMO. (30 -> 37 initial view of metro)
  - Depends on game, and if it was limited by this specific thing. Of course every game will be different.
- Games which read GPU modified textures on the CPU will flush them to memory immediately before the read.
  - Fixes replay thumbnails in MK8D. 
  - Fixes lighting flickering to fully dark when transitioning level section or pausing in SMO.
  - Textures are only flushed if they have to be. The flush happens entirely before any read or write completes, so readbacks and partial modification work perfectly as long as the GPU is in sync.
  - Cost for flushing textures to GPU moves to the CPU as a result. I expect generally better performance as I don't think it will stall the GPU thread when reading, like it does now for partial invalidation and manual flushing.
- Cost for reading buffers has decreased slightly.
- Cost for modifying tracked memory has increased. You'll notice this most on things that do so inefficiently, eg. the broken nvdec videos. Don't worry though, for I have a solution for that part too.

Benefits in future:
---
The main reason for doing this is the future potential of dropping software address translation on certain platforms. This is possible on Unix right now, but is a huge mess on Windows due to some questionable decisions so might be a while before we do anything for that.

This is the reason the tracking has been moved to a software memory protection system, so that it can seamlessly be switched to using OS memory protection over a virtual region of memory. I have already verified that this works on Linux, though I've decided that an integration should come after this PR to keep things simpler. (and actually get it done)

It's also generally nicer to work with than QueryMemory. I imagine it could be used for write tracking on shader code (so we can stop checking equality all the time) and invalidating translated functions in self modifying code.

Issues:
---
- ~Mismatching texture sizes + compression types can cause an out of bounds write when flushing a texture. Will crash on splatoon 2 and loads after cloud subworlds in SMO.~
  - ~The solution to this might be to just flag these textures as ineligible for flushing, for now. The texture cache will likely see a rework soon so I don't see any point band-aiding around the problem as part of this PR.~
- ~Linux does not like creating or using a shared context.~
- ~There is a worrying difference in this PR vs master where one course in MK8D (Mario Circuit) has a small region of its course cast literal noise as a shadow. I'm not sure what change could cause this but I'd like to look into it more.~
- Tests. May wish to do from Ryujinx.Cpu context (tracking writes to a virtual region) and from a Ryujinx.Memory context (individual tests into how each component functions)


riperiperi:
The Linux issue has been resolved by creating a hidden OpenTK.GameWindow for the background context, rather than reusing the existing window. This required an update to the OpenTK fork to fix the X11 backend - the package has been updated in Ryujinx.csproj. The other issues are still WIP.

riperiperi:
So, a few things have changed since the original implementation:

- Integration testing for memory tracking is now present. It tests various properties of tracking handles, and the multi tracking handles. There are comments in each test detailing all the expectations, and useful variable names for anything used in an assertion.
- The texture/buffer management PR has generally improved stability and reduced most of the unnecessary background flushing that was happening with the original version of this PR in some games (Animal Crossing, Mario Kart).
- Resolution scaling is now a thing. I've added a "_hostFlushTexture" to scaled textures, which is created the first time `ExternalFlush` is called for a texture. This is a 1x representation of the texture that we copy to for performing a flush, since we can't exactly force the original to 1x from another thread. Right now this does not result in blacklisting, but I might be able to make it defer a blacklist to happen in future if it becomes an issue.
- Memory tracking is disabled on 3D textures constructed from view slices. This is due to an issue that would cause them to be blanked out, due to invalidation of the data contained within. On master, this invalidation is consumed by the slices when re-rendered, then the 3d texture has the old data anyways. On this PR, modified flags do not get consumed for all overlapping textures, just the one requested. Disabling tracking on the 3D texture retains the behaviour of master, until we come up with a better solution for emulated texture views.

The following additional features have been implemented:
- MethodCopyBuffer's `copy2D` mode now attempts to find a target texture to copy to directly, if the full texture is being copied. This skips triggering read/write tracking for unrelated textures (generally large render targets) which was greatly slowing down NVDEC videos in Mario Kart, SMO, Zelda etc. It is also somewhat faster than master, as it can skip a format conversion by setting the host texture data directly, rather than converting to block layout and back. Generally seems to affect H264 videos in nintendo games.

Certain things in this PR aren't used at the moment:
- Tracking read/write from physical address is possible (and in the tests), but not currently used. 
- There are some mechanisms for triggering read/write tracking using host memory protection, though they are not currently used (this one is likely to be used in a direct followup for linux only). 
- The top bit of the pte (negation bit) is used to indicate that at least one software protection bit is active, but nothing consumes this right now. It could be used for making a quicker decision about page protection flags on read/write, though would need some care to avoid triggering slow paths on reads on write protected memory.

riperiperi:
`MethodCopyBuffer` now triggers read tracking, as it is generally used for format conversion when writing a texture back to the CPU. This fixes texture flushes in at least the following games:

- Link's Awakening (save file thumbnail)
- Zelda BOTW (save file thumbnail)
- Animal Crossing (passport photo)

image
image

Also, when creating a larger buffer out of overlapping ones, the memory is synchronized first to avoid overwriting copies from the overlapping buffers.

EmulationFanatic:
The latest commit fixes https://github.com/Ryujinx/Ryujinx/issues/1600

EmulationFanatic:
The latest commit also fixes https://github.com/Ryujinx/Ryujinx/issues/1297

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -699,6 +697,43 @@ public void Flush(bool tracked = true)
...
+        /// This is used when flushing from memory access handlers.
+        /// </summary>
+        public void ExternalFlush(ulong address, ulong size)
+        {
+            if (!IsModified || _memoryTracking == null)
+            {
+                return;
+            }
+
+            // NOTE: Remove before PR merge (useful for testing, but can spam a lot)
```
riperiperi, line 713: This must be removed after sufficient testing has been performed, before merging.

EmulationFanatic:
This fixes https://github.com/Ryujinx/Ryujinx/issues/1491 now.

Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
```diff
@@ -8,7 +8,9 @@ class TextureCopy : IDisposable
     {
         private readonly Renderer _renderer;
 
+        [ThreadStatic]
```
gdkchan, line 11: Might be better to just have another instance of `TextureCopy` for the other thread than doing this. What's the advantage in sharing the object?

Ryujinx.Memory/Range/INonOverlappingRange.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Ryujinx.Memory.Range
+{
+    interface INonOverlappingRange : IRange
+    {
+        /// <summary>
+        /// Split this region into two, around the specified address. 
+        /// This region is updated end at the split address, and a new region is created to represent past that point.
```
gdkchan, line 7: `This region is updated end at the split address` this sound weird imo.

Ryujinx.Memory/Tracking/AbstractRegion.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Ryujinx.Memory.Range;
+
+namespace Ryujinx.Memory.Tracking
+{
+    abstract class AbstractRegion : INonOverlappingRange
```
gdkchan, line 5: Missing XML docs here and on some methods, constructor, properties etc.

Ryujinx.Memory/Tracking/IMultiRegionHandle.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <param name="modifiedAction">Action to perform for modified regions</param>
+        public void QueryModified(Action<ulong, ulong> modifiedAction);
+
+
+        /// <summary>
+        /// Check if part of the region has been modified within a given range, and perform an action for each.
+        /// The range is aligned to the level of granularity of the contained handles.
+        /// Contiguous modified regions are combined.
+        /// </summary>
+        /// <param name="modifiedAction">Action to perform for modified regions</param>
```
gdkchan, line 25: Missing `address` and `size` descriptions.

Ryujinx.Memory/Tracking/MemoryTracking.cs
```diff
@@ -0,0 +1,322 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Get a list of virtual regions that a handle covers.
+        /// </summary>
+        /// <param name="va">Starting virtual memory address of the handle</param>
+        /// <param name="size">Size of the handle's memory region</param>
+        /// <returns></returns>
```
gdkchan, line 111: Missing.

Ryujinx.Memory/Tracking/MemoryTracking.cs
```diff
@@ -0,0 +1,322 @@
...
+        {
+            List<VirtualRegion> result = new List<VirtualRegion>();
+            _virtualRegions.GetOrAddRegions(result, va, size, (va, size) => new VirtualRegion(this, va, size));
+
+            return result;
+        }
+
+        /// <summary>
+        /// Get a list of physical regions that a virtual region covers.
+        /// Note that this becomes outdated if the virtual or physical regions are unmapped or remapped..
```
gdkchan, line 122:
```suggestion
        /// Note that this becomes outdated if the virtual or physical regions are unmapped or remapped.
```

Ryujinx.Memory/Tracking/MemoryTracking.cs
```diff
@@ -0,0 +1,322 @@
...
+
+                return handle;
+            }
+        }
+
+        /// <summary>
+        /// Signal that a physical memory event happened at the given location.
+        /// </summary>
+        /// <param name="address">Physical address accessed</param>
+        /// <param name="size">Size of the region affected in bytes</param>
```
gdkchan, line 213: This parameter does not exist.

Ryujinx.Memory/Tracking/MemoryTracking.cs
```diff
@@ -0,0 +1,322 @@
...
+            // Signal up the chain to relevant handles.
+
+            lock (TrackingLock)
+            {
+                var results = _virtualResults;
+                int count = _virtualRegions.FindOverlapsNonOverlapping(address, size, ref results);
+
+                if (count == 0)
+                {
+                    _memoryManager.Reprotect(address & ~(ulong)(_pageSize - 1), 4096, MemoryPermission.ReadAndWrite);
```
gdkchan, line 272: Maybe use a constant instead of the `4096` magic number here.

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -71,14 +79,18 @@ public MemoryManager(MemoryBlock backingMemory, ulong addressSpaceSize, InvalidA
...
+            while (remainingSize != 0)
             {
                 _pageTable.Write((va / PageSize) * PteSize, PaToPte(pa));
 
                 va += PageSize;
                 pa += PageSize;
-                size -= PageSize;
+                remainingSize -= PageSize;
             }
+            Tracking.Map(oVa, oPa, size);
```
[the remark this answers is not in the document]
riperiperi: Tracking.Map will use the updated page table - it calls `GetPhysicalRegions` on the memory manager for any virtual regions, assuming that the new mapping is in effect. Same for Unmap.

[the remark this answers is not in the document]
gdkchan: Ok.

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -516,13 +506,24 @@ private ulong GetPhysicalAddressInternal(ulong va)
...
-        /// Marks a region of memory as modified by the CPU.
+        /// Reprotect a region of virtual memory. Sets software protection bits.
         /// </summary>
-        /// <param name="va">Virtual address of the region</param>
-        /// <param name="size">Size of the region</param>
-        public void MarkRegionAsModified(ulong va, ulong size)
+        /// <param name="va">Virtual address base</param>
+        /// <param name="size">Size of the region to protect</param>
+        /// <param name="protection">Memory protection to set</param>
+        public void Reprotect(ulong va, ulong size, MemoryPermission protection)
```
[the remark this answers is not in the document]
riperiperi: This would actually do a host reprotect when doing direct memory mapping, so I'll keep Reprotect in the name: "TrackingReprotect"

Ryujinx.Graphics.Gpu/Engine/MethodCopyBuffer.cs
```diff
@@ -7,6 +8,37 @@ namespace Ryujinx.Graphics.Gpu.Engine
...
+        private const int StrideAlignment = 32;
+        private const int GobAlignment = 64;
+
+        /// <summary>
+        /// Determine if a buffer-to-texture region covers the entirety of a texture.
+        /// </summary>
+        /// <param name="cbp">Copy command parameters</param>
+        /// <param name="tex">Texture to compare</param>
+        /// <param name="linear">True if the texture is linear, false if block linear</param>
+        /// <param name="bpp">Texture bits per pixel</param>
```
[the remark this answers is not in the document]
riperiperi: Autopiloted on this description.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -901,42 +904,65 @@ public Texture FindOrCreateTexture(TextureInfo info, TextureSearchFlags flags =
...
+                }
+                else
+                {
+                    // Bpp may be a mismatch between the target texture and the param. 
+                    // Due to the way linear strided and block layouts work, widths can be multiplied by Bpp for comparison.
+                    // Note: tex.Width is the aligned texture size. Prefer param.XCount, as the destination should be a texture with that exact size.
+
+                    bool sizeMatch = param.XCount * bpp == texture.Info.Width * format.BytesPerPixel && tex.Height == texture.Info.Height;
+                    bool formatMatch = !texture.Info.IsLinear &&
+                                        texture.Info.GobBlocksInTileX == tex.MemoryLayout.UnpackGobBlocksInX() &&
```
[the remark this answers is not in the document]
riperiperi: Okay, was confused why it wasn't used anywhere else. I'll remove it.

Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -48,9 +53,16 @@ public Buffer(GpuContext context, ulong address, ulong size)
 
             Handle = context.Renderer.CreateBuffer((int)size);
 
-            _modifiedRanges = new (ulong, ulong)[size / PhysicalMemory.PageSize];
+            _useGranular = size > GranularBufferThreshold;
 
-            _sequenceNumbers = new int[size / MemoryManager.PageSize];
+            if (_useGranular)
+            {
+                _memoryTrackingGranular = context.PhysicalMemory.BeginSmartGranularTracking(address, size, 4096);
```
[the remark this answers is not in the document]
riperiperi: Nope, the threshold is just there to switch to the single handle tracking to save effort. If we are using the granular tracking, there's not much reason to use a granularity above the page size, from what I can imagine anyways.

Maybe this should just be a sensible default parameter, rather than specifically requested.

Ryujinx.Graphics.OpenGL/BackgroundContextWorker.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            _invokePool.Release(actionComplete);
+        }
+
+        public void Dispose()
+        {
+            _running = false;
+            _signal.Set();
+
+            _disposed.WaitOne();
```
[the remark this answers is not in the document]
riperiperi: No idea why I didn't do that in the first place, I was even planning to 🤔

Ryujinx.Memory/Range/INonOverlappingRange.cs
riperiperi: Missing a "to". `This region is updated to end at the split address`

Ryujinx.Memory/Range/RangeList.cs
```diff
@@ -0,0 +1,342 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Ryujinx.Memory.Range
+{
+    /// <summary>
+    /// Sorted list of ranges that supports binary search.
+    /// </summary>
+    /// <typeparam name="T">Type of the range.</typeparam>
+    class RangeList<T> : IEnumerable<T> where T : IRange
```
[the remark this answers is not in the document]
riperiperi: It would need to be located in Ryujinx.Common, but I am for the change.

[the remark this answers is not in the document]
gdkchan: I think `Ryujinx.Gpu` already has a dependency on `Ryujinx.Memory`.

Ryujinx.Memory/Tracking/MemoryTracking.cs
riperiperi: Will use _pageSize, as that's what this value represents here.

Ryujinx.Memory/Range/RangeList.cs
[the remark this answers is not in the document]
riperiperi: Ah, it's there but not top level. Okay.

Ryujinx.Memory/Range/NonOverlappingRangeList.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// Splits a region around a target point and updates the region list. 
+        /// The original region's size is modified, but its address stays the same.
+        /// A new region starting from the split address is added to the region list and returned.
+        /// </summary>
+        /// <param name="region">The region to split</param>
+        /// <param name="splitAddress">The address to split with</param>
+        /// <returns>The new region (high part)</returns>
+        private T Split(T region, ulong splitAddress)
+        {
+            Remove(region); // TODO: Is this necessary?
```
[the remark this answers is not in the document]
riperiperi: Yes. Will remove comment.

riperiperi:
The PTC version has been changed to 1273 as part of the rebase, as changing the name on a function on the native interface invalidated all existing caches. #1273 isn't a CPU PR so there should be no conflict.

Ryujinx.Graphics.OpenGL/BackgroundContextWorker.cs
```diff
@@ -0,0 +1,99 @@
...
+                100, 100, GraphicsMode.Default,
+                "Background Window", OpenTK.GameWindowFlags.FixedWindow, OpenTK.DisplayDevice.Default,
+                3, 3, GraphicsContextFlags.ForwardCompatible, baseContext, false);
+
+            _window.Visible = false;
+            _context = (GraphicsContext)_window.Context;
+
+            _running = true;
+
+            _signal = new AutoResetEvent(false);
```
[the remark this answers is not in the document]
riperiperi: It is now disposed. (the _disposed handle has been removed in favor of thread.join)

Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
riperiperi: Now uses a separate instance of TextureCopy. This was actually broken, as these handles were changed from static some time ago, and ThreadStatic does nothing on non-static variables.

Ryujinx.Memory.Tests/MockVirtualMemoryManager.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Text;
+
+namespace Ryujinx.Memory.Tests
+{
+    class MockVirtualMemoryManager : IVirtualMemoryManager
+    {
+        public bool NoMappings;
+
+        public MockVirtualMemoryManager(ulong size, int pageSize)
+        {
```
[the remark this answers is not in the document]
riperiperi: Removed in this constructor.

riperiperi:
Feedback addressed.

It's worth noting that there is a new performance regression in Pokemon Sword & Shield, as the game uses the Copy Buffer method to extract a few single pixels from a 1024x1024 depth texture. I'm not sure what the game uses this for, but it _does_ use it in guest code, so the only way to solve the problem is to try optimize the flush by only fetching the single pixel it wants. This probably won't be part of this PR.

Ryujinx.Memory/Tracking/IMultiRegionHandle.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+
+namespace Ryujinx.Memory.Tracking
+{
+    public interface IMultiRegionHandle : IDisposable
```
gdkchan, line 5: Interface has redundant public access modifier.

Ryujinx.Memory/Tracking/IRegionHandle.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Ryujinx.Memory.Tracking
+{
+    public interface IRegionHandle : IDisposable
```
gdkchan, line 5: Interface has redundant public access modifier.

Ryujinx.Memory/Tracking/RegionHandle.cs
```diff
@@ -0,0 +1,135 @@
+﻿using Ryujinx.Memory.Range;
+using System;
```
gdkchan, line 2: This `using` can be removed.

ARMeilleure/Translation/Delegates.cs
```diff
@@ -117,7 +117,7 @@ static Delegates()
             SetDelegateInfo(typeof(NativeInterface).GetMethod(nameof(NativeInterface.GetTpidr32))); // A32 only.
             SetDelegateInfo(typeof(NativeInterface).GetMethod(nameof(NativeInterface.GetTpidrEl0)));
             SetDelegateInfo(typeof(NativeInterface).GetMethod(nameof(NativeInterface.GetTpidrEl032))); // A32 only.
-            SetDelegateInfo(typeof(NativeInterface).GetMethod(nameof(NativeInterface.MarkRegionAsModified)));
+            SetDelegateInfo(typeof(NativeInterface).GetMethod(nameof(NativeInterface.SignalMemoryTracking)));
```
LDj3SNuD, line 120: Although not strictly necessary, it is better to place in alphabetical order, being a SortedList, it speeds things up during static initialization.

riperiperi: Ah, I just renamed this totally blindly. Will fix that.

riperiperi:
Oh. That public access modifier wasn't redundant.

gdkchan (approved):
lgtm, great work!

Ryujinx.Memory/Range/NonOverlappingRangeList.cs
```diff
@@ -0,0 +1,108 @@
...
+                    }
+
+                    if (region.EndAddress > address + size)
+                    {
+                        // Split the region around our end address and take the low half.
+
+                        Split(region, address + size);
+                    }
+
+                    list.Add(region);
```
FICTURE7, line 76: I believe this can ignore the lower region from `list` if `region` is larger than `[address, address + size)` in both directions. i.e, both a lower trim and an upper trim happens in the same iteration.

FICTURE7: Nevermind

Ryujinx/Ryujinx.csproj
```diff
@@ -77,7 +77,7 @@
     <PackageReference Include="GtkSharp" Version="3.22.25.56" />
     <PackageReference Include="GtkSharp.Dependencies" Version="1.1.0" Condition="'$(RuntimeIdentifier)' != 'linux-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'" />
     <PackageReference Include="Ryujinx.Graphics.Nvdec.Dependencies" Version="4.3.0" Condition="'$(RuntimeIdentifier)' != 'linux-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'" />
-    <PackageReference Include="OpenTK.NetStandard" Version="1.0.5.12" />
+    <PackageReference Include="OpenTK.NetStandard" Version="1.0.5.22" />
```
LDj3SNuD, line 80: `Ryujinx.Audio\Ryujinx.Audio.csproj` and `Ryujinx.Graphics.OpenGL\Ryujinx.Graphics.OpenGL.csproj` still use the package version `1.0.5.12`, should they be updated too?

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -516,13 +506,24 @@ private ulong GetPhysicalAddressInternal(ulong va)
...
-        /// <param name="va">Virtual address of the region</param>
-        /// <param name="size">Size of the region</param>
-        public void MarkRegionAsModified(ulong va, ulong size)
+        /// <param name="va">Virtual address base</param>
+        /// <param name="size">Size of the region to protect</param>
+        /// <param name="protection">Memory protection to set</param>
+        public void TrackingReprotect(ulong va, ulong size, MemoryPermission protection)
         {
+            // Protection is inverted on software pages, since the default value is 0.
+            protection = (~protection) & MemoryPermission.ReadAndWrite;
```
LDj3SNuD, line 517: Are round brackets around `~protection` necessary?

Ryujinx.Memory/Tracking/MemoryTracking.cs
```diff
@@ -0,0 +1,321 @@
...
+
+            _virtualRegions = new NonOverlappingRangeList<VirtualRegion>();
+            _physicalRegions = new NonOverlappingRangeList<PhysicalRegion>();
+        }
+
+        private (ulong address, ulong size) PageAlign(ulong address, ulong size)
+        {
+            ulong pageMask = (ulong)_pageSize - 1;
+            ulong rA = address & (~pageMask);
+            ulong rS = ((address + size + pageMask) & (~pageMask)) - rA;
```
LDj3SNuD, line 53: Are round brackets around `~pageMask` necessary?

Ryujinx.Cpu/Tracking/CpuRegionHandle.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Ryujinx.Memory.Tracking;
+using System;
```
LDj3SNuD, line 2: It seems unused.

Ryujinx.Memory.Tests/MockVirtualMemoryManager.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Ryujinx.Memory.Tracking;
+using System;
+using System.Collections.Generic;
+using System.Text;
```
LDj3SNuD, line 4: They seem unused.

Ryujinx.Memory/Tracking/RegionSignal.cs
```diff
@@ -0,0 +1 @@
+﻿public delegate void RegionSignal(ulong address, ulong size);
```
LDj3SNuD, line 1: No namespace?

LDj3SNuD (approved):
Just a few nits or so.

I don't know if this is due to this PR, but when you load a title from the list (double click), the window now seems to lose focus.

lgtm, ty

Ryujinx.Cpu/MemoryManager.cs
riperiperi: Visual studio isn't highlighting them as unnecessary. I think the bitwise operation ordering is enough of a mess to warrant them under any circumstance.

Ryujinx/Ryujinx.csproj
riperiperi: Forgot to consolidate across packages, since I just used the changes in this project. Will do that now.

Ryujinx.Memory/Tracking/MemoryTracking.cs
riperiperi: I'll remove these ones, since `value1 & ~value2` only has one meaning regardless of order

Thog (approved):
lgtm

PR #1275, opened 2020-05-25
Fix GetDesiredLanguage

mageven:
Fixes wrong language selection when application doesn't support set system language.

XC2
xc2-lang-fix

Splatoon 2
splatoon2-lang-fix

Before the fix, the above screens would show Japanese.

Please close this PR if it doesn't meet your standards. Just wrote a quick fix as it was annoying me. I've no interest in expanding this for now.

~~I give up. Can't even name the operation right~~ (Reworded 2nd commit to give proper name)

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -102,9 +102,9 @@ namespace Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.Applicati
             if (((1 << (int)context.Device.System.State.DesiredTitleLanguage) & supportedLanguages) == 0)
             {
                 // Find first supported language from TitleLanguage. TODO: In the future, a GUI could enable user-specified search priority
-                int setIndex = 1;
+                int setIndex = 0;
 
-                while ((supportedLanguages & 1) != 0) // This call isn't perf critical, so no need for bit hacks
+                while ((supportedLanguages & 1) == 0) // This call isn't perf critical, so no need for bit hacks
                 {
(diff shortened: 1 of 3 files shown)


Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
```diff
@@ -79,7 +80,42 @@ public ResultCode EnsureSaveData(ServiceCtx context)
...
+                return ResultCode.Success;
+            }
+
+            // If desired language is not supported by application
+            if (((1 << (int)context.Device.System.State.DesiredTitleLanguage) & supportedLanguages) == 0)
+            {
+                // Find first supported language from TitleLanguage. TODO: In the future, a GUI could enable user-specified search priority
+                int setIndex = 0;
+
+                while ((supportedLanguages & 1) == 0) // This call isn't perf critical, so no need for bit hacks
```
gdkchan, line 107: I agree with the comment, but this could be a single call to `BitOperations.TrailingZeroCount` to get the first least significant bit set efficiently.

gdkchan: Also that would be a infinite loop if `supportedLanguages` was 0. I think this case should be properly handled aswell.

mageven: Ah. Didn't know about `BitOperations.TrailingZeroCount`. Infinite loop is already handled in the previous `if` which checks `supportedLanguages == 0` right?

gdkchan: Yea, sorry, I missed it.

PR #912, opened 2020-01-30
[WIP] Implement triangle_rast_flip bit in YControl

riperiperi:
Flips triangle front faces when the triangle_rast_flip flag is set on YControl. Combined with #910, fixes the graphics in Xenoblade Chronicles 2 (mostly). 

image

This is WIP because I don't have enough games to make sure this doesn't cause regressions - the trigger for doing this is more complicated and we could use more examples. Testing is appreciated.

**Known improvements**
- Xenoblade Chronicles 2: Fixes 3D element cull direction. _YControl 0x10, Viewport Swizzle not used, negative viewport height._
- Arc of Alchemist: Fixes 3D element cull direction.
- Disgaea 4 Complete+: Fixes 3D element cull direction. (some were correct already, this flips the ones that weren't)

**Known unchanged**
- Super Mario Odyssey
- Splatoon 2
- Mario Kart 8 Deluxe
- Captain Toad Treasure Tracker

**Known issues**
- Cave Story +: Only background colour drawn. _YControl 0x10, Viewport Swizzle not used, negative viewport height._
- Pokemon Sword/Shield: All 3D elements missing. (likely culling applied during post processing)
- Troll and I: Culling is incorrectly flipped.
- Pokemon Let's Go! Eevee: Culling is incorrectly flipped.
- Mario Tennis Aces: Culling is incorrectly flipped.

PR #1277, opened 2020-05-25
Fix wrong face culling once and for all

gdkchan:
NVN exposes a function that allows the application to set the origin of the window. It can be set to upper left (following Direct3D conventions), or lower left (following OpenGL). This should be directly equivalent to glClipControl on OpenGL.

The way how this is actually implemented on the GPU, however, is not directly equivalent. The GPU has a bit that changes something on the rasterization stage to take the different origin into account (nouveau calls this bit TriangleRastFlip). It does not flip the viewport, however. Flipping the viewport is done by using a negative value for the viewport scale.

So when the UpperLeft mode is used, NVN does 2 things:
- Negates the viewport Y scale value. This effectively flips the viewport, so all the geometry is flipped on the Y direction. This also affects face culling as the orientation of the triangles will change if you flip them.
- It sets the TriangleRastFlip bit to 0 (to indicate the UpperLeft origin is being used,  LowerLeft is 1). This negates the effects that the above viewport flipping would have on the face culling, as it changes something during rasterization.

So in most cases, directly setting origin to UpperLeft or LowerLeft using glClipControl should be safe and equivalent to NVN behavior, as long the above conditions holds (that is, either TriangleRastFlip is 0 and the viewport scale Y is negative, or TriangleRastFlip is 1 and viewport scale Y is positive). However, some games cheats and this may not be the case.

One example of such a game is Xenoblade. It manages to bypass the step where NVN sets the TriangleRastFlip bit to 0 by only setting the origin to upper left *after* queue initialization. This effectively causes NVN to still use negative viewport scale Y values, while not setting triangle rast flip to 0, because during queue initialization, origin was still lower left.

**Approach used to fix the issue**

We can still use ARB_clip_control to emulate TriangleRastFlip behavior, however the viewport flipping might be undesirable. We can negate that part by flipping the viewport again. To my knowledge, per-viewport flipping is possible, but only using NVIDIA [NV_viewport_swizzle](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_viewport_swizzle.txt) extension, only available on Maxwell and newer. For GPUs without the support, we can still do the flipping on the vertex shader, but it affects all the viewports.

The fix here uses NV_viewport_swizzle (or shader emulation) to flip the viewport as needed. glClipControl is still used, but origin is now set to LowerLeft or UpperLeft based on the value of TriangleRastFlip bit. The various combinations of bits and values that may flip it in the Y direction (negative Y scale, Y negate etc) are propagated to the viewport swizzle Y component.

Note: The shader implementation currently has another issue, on top of not supporting per-viewport flipping. Since we don't recompile shaders when the state changes, the shader may be assuming a stale/incorrect viewport swizzle state. However this is a existing issue.

--------

I recommend testing this on a few games, including the ones that had culling issues, and games that used to work fine, to ensure it does not regress anything. If you find issues, please also note the GPU it was tested on (basically, we need to know if its using NV_viewport_swizzle or the shader implementation).

Should fix culling issues on Xenoblade 2, and maybe other games.

This supersedes #912

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -422,7 +422,13 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
             _context.Renderer.Pipeline.SetDepthMode(depthMode);
 
-            bool flipY = (state.Get<YControl>(MethodOffset.YControl) & YControl.NegateY) != 0;
+            int yControl = state.Get<int>(MethodOffset.YControl);
+
+            bool   flipY  = (yControl & 0b00001) != 0;
+            Origin origin = (yControl & 0b10000) != 0 ? Origin.LowerLeft : Origin.UpperLeft;
+
+            _context.Renderer.Pipeline.SetOrigin(origin);
+
             float yFlip = flipY ? -1 : 1;
 
             Viewport[] viewports = new Viewport[Constants.TotalViewports];
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -422,16 +422,23 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
             _context.Renderer.Pipeline.SetDepthMode(depthMode);
 
-            int yControl = state.Get<int>(MethodOffset.YControl);
+            YControl yControl = state.Get<YControl>(MethodOffset.YControl);
 
-            bool   flipY  = (yControl & 0b00001) != 0;
-            Origin origin = (yControl & 0b10000) != 0 ? Origin.LowerLeft : Origin.UpperLeft;
-
+            bool   flipY  = yControl.HasFlag(YControl.NegateY);
+            Origin origin = yControl.HasFlag(YControl.TriangleRastFlip) ? Origin.LowerLeft : Origin.UpperLeft;
+            
             _context.Renderer.Pipeline.SetOrigin(origin);
 
-            float yFlip = flipY ? -1 : 1;
+            // The triangle rast flip flag only affects rasterization, the viewport is not flipped.
+            // Setting the origin mode to upper left on the host, however, not onlyy affects rasterization,
+            // but also flips the viewport.
+            // We negate the effects of flipping the viewport by flipping it again using the viewport swizzle.
+            if (origin == Origin.UpperLeft)
+            {
+                flipY = !flipY;
+            }
 
-            Viewport[] viewports = new Viewport[Constants.TotalViewports];
+            Span<Viewport> viewports = stackalloc Viewport[Constants.TotalViewports];
 
             for (int index = 0; index < Constants.TotalViewports; index++)
             {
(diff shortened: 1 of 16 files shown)


Ryujinx.Graphics.Gpu/State/GpuState.cs
```diff
@@ -146,6 +146,10 @@ private void InitializeDefaultState(int[] memory)
             {
                 memory[(int)MethodOffset.ViewportExtents + index * 4 + 2] = 0;
                 memory[(int)MethodOffset.ViewportExtents + index * 4 + 3] = 0x3F800000;
+                memory[(int)MethodOffset.ViewportExtents + index * 4 + 3] = 0x3F800000;
```
riperiperi, line 149: This is identical to the above line. (typo?)

Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
```diff
@@ -61,13 +59,40 @@ public Operand GetLabel(ulong address)
 
         public void PrepareForReturn()
         {
-            if (_config.Stage == ShaderStage.Fragment)
+            if (Config.Stage == ShaderStage.Vertex && (Config.Flags & TranslationFlags.VertexA) == 0)
+            {
+                // Here we attempt to implement viewport swizzle on the vertex shader.
+                // Perform permutation and negation of the output gl_Position components.
+                // Note that per-viewport swizzling can't be supported using thhis approach.
```
riperiperi, line 66: `this`

GoffyDude:
Tested this PR via Appveyor build 1.0.4624 on Disgaea 4 Complete Plus vs 1.0.4640 master.
Screenshots Below:

#### Master :
2020-05-27 08_36_45-Ryujinx 1 0 4640 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)

#### PR 1277 build :

2020-05-27 08_39_28-Ryujinx 1 0 4624 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)
__________________________________________________________________________________________________________
#### Master : 
2020-05-27 08_36_53-Ryujinx 1 0 4640 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)

#### PR 1277 Build : 
2020-05-27 08_39_54-Ryujinx 1 0 4624 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)
__________________________________________________________________________________________________________
#### Master : 
2020-05-27 08_36_59-Ryujinx 1 0 4640 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)

#### PR 1277 Build : 
2020-05-27 08_41_09-Ryujinx 1 0 4624 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)
__________________________________________________________________________________________________________
#### Master : 
2020-05-27 08_37_12-Ryujinx 1 0 4640 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)

#### PR 1277 Build : 
2020-05-27 08_40_36-Ryujinx 1 0 4624 - Disgaea 4 Complete+ v1 0 0 (0100A9800E9B4000) (64-bit)

GoffyDude:
Tested again with the latest rebased version (1.0.4641) and everything is still looking good!

Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -422,10 +422,23 @@ private void UpdateViewportTransform(GpuState state)
...
-            float yFlip = flipY ? -1 : 1;
+            YControl yControl = state.Get<YControl>(MethodOffset.YControl);
+
+            bool   flipY  = yControl.HasFlag(YControl.NegateY);
+            Origin origin = yControl.HasFlag(YControl.TriangleRastFlip) ? Origin.LowerLeft : Origin.UpperLeft;
+            
+            _context.Renderer.Pipeline.SetOrigin(origin);
+
+            // The triangle rast flip flag only affects rasterization, the viewport is not flipped.
+            // Setting the origin mode to upper left on the host, however, not onlyy affects rasterization,
```
Cristallix, line 433: `onlyy` => typo

Ryujinx.Graphics.OpenGL/EnumConversion.cs
```diff
@@ -416,5 +416,32 @@ public static TextureTarget Convert(this Target target)
...
+                    return NvViewportSwizzle.ViewportSwizzleNegativeYNv;
+                case ViewportSwizzle.NegativeZ:
+                    return NvViewportSwizzle.ViewportSwizzleNegativeZNv;
+                case ViewportSwizzle.NegativeW:
+                    return NvViewportSwizzle.ViewportSwizzleNegativeWNv;
+            }
+
+            Logger.PrintDebug(LogClass.Gpu, $"Invalid {nameof(ViewportSwizzle)} enum value: {swizzle}.");
+
+            return NvViewportSwizzle.ViewportSwizzlePositiveXNv;
```
Cristallix, line 444: Why not using the default of the switch ?

gdkchan: It prints some information on the log before returning. What would be the advantage of using the default case here?

Cristallix:
```suggestion
                default:
                    Logger.PrintDebug(LogClass.Gpu, $"Invalid {nameof(ViewportSwizzle)} enum value: {swizzle}.");
                    return NvViewportSwizzle.ViewportSwizzlePositiveXNv;
```
You can do your log and then `return ` in the `default`. To me it's more consistent.

Ryujinx.Graphics.Gpu/Engine/Methods.cs
gdkchan: I will fix the typo when I touch this file again, since the PR was already merged.

PR #1283, opened 2020-05-28
Use C#7.3's ReadOnlySpan<byte> optimization for static data

turbedi:
https://vcsjones.dev/2019/02/01/csharp-readonly-span-bytes-static/


jduncanator:
These changes have different meaning:

A `readonly` field cannot have it's reference or value changed. A `ReadOnlySpan` cannot have the value it points to changed. In this instance, the fields are now no longer `readonly`, which means the reference they point to could be changed (eg. `_sbox = new byte[0];`). Removing the `readonly` also prevents the JIT from being able to do a certain class of optimization, where it knows the value will never change.

I think it's best to leave this how it was, usually for code like this, it's been profiled and investigated, and the best approach has been chosen for a reason (eg. better codegen, better perf etc).

jduncanator:
Ignore my previous comment. I didn't notice it was now a property with only a getter, rather than a field.

The change looks good, but I'd still like to see verification that the improvement is positive in the Ryujinx code base.

jduncanator:
Closing this, if the OP wants to update with some demonstration of improvements, more than happy to look at merging! Thanks!

PR #1287, opened 2020-05-29
Implementing Vmad

Shrek5InTheatres2019:
Okay so i kinda deleted my old branch on accident so i rewrote and made it look nicer, it still has the same functionality though and works all the same in my testing

### DIFF
--- a/Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
+++ b/Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
@@ -12,7 +12,40 @@ namespace Ryujinx.Graphics.Shader.Instructions
         public static void Vmad(EmitterContext context)
         {
             // TODO: Implement properly.
-            context.Copy(GetDest(context), GetSrcC(context));
+            OpCodeVideo op = (OpCodeVideo)context.CurrOp;
+
+            Operand result;
+            Operand A = GetSrcA(context);
+            Operand B = GetSrcB(context);
+            Operand C = GetSrcC(context);
+
+            bool signedFinal = isSigned(op.RaType) || isSigned(op.RbType);
+
+            int shr = op.RawOpCode.Extract(51, 2);
+
+            result = context.IMultiply(A, B);
+            result = context.IAdd(result, C);
+
+            if (shr == 1)
+            {
+                result = signedFinal
+                    ? context.ShiftRightS32(result, Const(7))
+                    : context.ShiftRightU32(result, Const(7));
+            }
+            else if (shr == 2)
+            {
+                result = signedFinal
+                    ? context.ShiftRightS32(result, Const(15))
+                    : context.ShiftRightU32(result, Const(15));
+            }
+
+
+            context.Copy(GetDest(context), result);
+        }
+
+        public static bool isSigned(VideoType v)
+        {
+            return v == VideoType.S16 || v == VideoType.S32 || v == VideoType.S8;
         }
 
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
```diff
@@ -12,7 +12,40 @@ static partial class InstEmit
...
+            Operand result;
+            Operand A = GetSrcA(context);
+            Operand B = GetSrcB(context);
+            Operand C = GetSrcC(context);
+
+            bool signedFinal = isSigned(op.RaType) || isSigned(op.RbType);
+
+            int shr = op.RawOpCode.Extract(51, 2);
+
+            result = context.IMultiply(A, B);
```
jduncanator, line 26: Looking at the available documentation on NVIDIA's site, it looks like this multiply has a 128-bit product, I'm not fluent enough with the shader IR format, but I'm not sure if this behaviour is replicated here?

Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
```diff
@@ -12,7 +12,40 @@ static partial class InstEmit
...
+            }
+            else if (shr == 2)
+            {
+                result = signedFinal
+                    ? context.ShiftRightS32(result, Const(15))
+                    : context.ShiftRightU32(result, Const(15));
+            }
+
+
+            context.Copy(GetDest(context), result);
```
jduncanator, line 43: The opcode encodes an optional "saturate" mode, that saturates the final 128-bit result to a 32-bit result (eg. if the 128-bit value is larger than `0xFFFFFFFF`, then the value is set to `0xFFFFFFFF`. Encodings that make use of this feature may result in incorrect results if the saturation is not applied.

Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
```diff
@@ -12,7 +12,40 @@ static partial class InstEmit
...
+            Operand A = GetSrcA(context);
+            Operand B = GetSrcB(context);
+            Operand C = GetSrcC(context);
+
+            bool signedFinal = isSigned(op.RaType) || isSigned(op.RbType);
+
+            int shr = op.RawOpCode.Extract(51, 2);
+
+            result = context.IMultiply(A, B);
+            result = context.IAdd(result, C);
```
jduncanator, line 27: According to NVIDIA documentation, there is an optional "plus-one" encoding available, that adds a constant of `1` to the result, eg. `(a * b) + c + 1`.

Ryujinx.Graphics.Shader/Instructions/InstEmitVideo.cs
```diff
@@ -12,7 +12,40 @@ static partial class InstEmit
         public static void Vmad(EmitterContext context)
         {
             // TODO: Implement properly.
-            context.Copy(GetDest(context), GetSrcC(context));
+            OpCodeVideo op = (OpCodeVideo)context.CurrOp;
+
+            Operand result;
+            Operand A = GetSrcA(context);
+            Operand B = GetSrcB(context);
+            Operand C = GetSrcC(context);
```
jduncanator, line 20: According to NVIDIA documentation, there is an additional encoding available that negates either `(a * b)` or `c` (but never both).

Thog:
Closing for inactivity.

PR #1292, opened 2020-06-01
Improved Logger

mageven:
The goal for this rewrite was zero-cost disabled logs without:
- losing out on code readability 
- large changes to the entire project.

The new code allows that with minimal changes (2 extra chars per call). Existing PRs can be modified with a script.
Since the condition is checked before `Print` call, no parameters are evaluated. So, no str-format/heap-allocs/func-calls if that log-level is disabled. And there's no race hazards.

```
Logger.PrintError() => Logger.Error?.Print()
...
Logger.PrintStub() => Logger.Stub?.PrintStub() // the only one that's different due to overloads
```

Other approaches tried:
1. `FormattableString` instead of `string` (only reduces format overhead; doesn't play well with function overloads)
2. `Logger.IsXXEnabled` bool guards before each `Logger.PrintXX` (too verbose)
3. `PrintXX` nullable delegates (uglier `XX.?Invoke()` syntax every log call)
4. Turning `PrintDebug` into `Debug.Print` equivalents (unchangeable at runtime, doesn't help other levels)
5. `ILevel` interface with dummy `Print` on disable (virt call still exists, all params are still evaluated)

Unfortunately, as most interpolated messages lie outside the print methods, it's not feasible to calculate the perf impact. So, I have no benchmarks. <br/>
Nominal game testing _might_ show improvements and reduce stutters (with most log levels turned off). <br/>
In the worst case, at least the GC will be happier.

IMO, an extra Log Level might be better to add in the future (say `Diagnostic`?)

I can provide a script for automating changes to PRs if this approach is accepted.

Also, adds a tiny log-level marker to each line to filter them easily in log files (which don't have colors). Sample-

```
00:00:00.010 |N| Application PrintSystemInfo: Ryujinx Version: 1.0.0-dirty
00:00:01.192 |N| Application PrintSystemInfo: Operating System: ...
00:00:01.192 |N| Application PrintSystemInfo: CPU: ...
00:00:01.192 |N| Application PrintSystemInfo: Total RAM: ...
00:00:01.193 |N| Application PrintSystemInfo: Logs Enabled: Info, Warning, Error
00:00:00.405 |N| Application LoadApplication: Using Firmware Version: ...
...
00:00:00.421 |I| Loader LoadExeFs: Loading rtld...
00:00:00.423 |I| Loader LoadExeFs: Loading main...
...
00:00:02.593 |G| ServiceLm Log: Guest Log:
  Log level: Info
    Time: 1s
    Line: 29
    Filename: '...'
    Function: 'nninitStartup'
    Module: ''
    Thread: 'MainThread'
    ProgramName: 'PRJ_036'
    Message: '---- nninitStartup Begin ----'
...
00:00:05.076 |S| ServiceNv SetUserData: Stubbed. 
00:00:05.081 |S| ServiceNv SetTimeslice: Stubbed. 
00:00:05.084 |I| ServiceNv Create: Created map 4 with size 0x00020000!
00:00:05.113 |I| ServiceNv Create: Created map 5 with size 0x0004a000!
...
00:00:19.622 |E| GUI.RenderLoop Gpu GetInfo: Invalid texture format 0x8003F3F3 (sRGB: False).
00:00:19.632 |E| GUI.RenderLoop Gpu GetInfo: Invalid texture format 0x8003F3F3 (sRGB: False).
00:00:19.969 |W| GUI.RenderLoop Gpu Log: Shader translator: Invalid instruction at 0x000030 (0x50F41BA06FF70107).
00:00:19.982 |W| GUI.RenderLoop Gpu Log: Shader translator: Invalid instruction at 0x000028 (0x50F41BA06FF70007).
...
```

### DIFF
--- a/Ryujinx.Common/Logging/Logger.cs
+++ b/Ryujinx.Common/Logging/Logger.cs
@@ -56,7 +56,6 @@ namespace Ryujinx.Common.Logging
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public void PrintStub(LogClass logClass, string message = "", [CallerMemberName] string caller = "")
             {
-                // if (_level != LogLevel.Stub) throw new InvalidOperationException("PrintStub must be called Stub log level");
                 if (m_EnabledClasses[(int)logClass])
                 {
                     Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, "Stubbed. " + message)));
@@ -66,17 +65,15 @@ namespace Ryujinx.Common.Logging
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public void PrintStub(LogClass logClass, object data, [CallerMemberName] string caller = "")
             {
-                // if (_level != LogLevel.Stub) throw new InvalidOperationException("PrintStub must be called Stub log level");
                 if (m_EnabledClasses[(int)logClass])
                 {
-                    Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, "Stubbed. "), data));
+                    Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, "Stubbed."), data));
                 }
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public void PrintStub(LogClass logClass, string message, object data, [CallerMemberName] string caller = "")
             {
-                // if (_level != LogLevel.Stub) throw new InvalidOperationException("PrintStub must be called Stub log level");
                 if (m_EnabledClasses[(int)logClass])
                 {
                     Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, "Stubbed. " + message), data));
--- a/Ryujinx.Common/Logging/Logger.cs
+++ b/Ryujinx.Common/Logging/Logger.cs
@@ -18,11 +18,11 @@ namespace Ryujinx.Common.Logging
 
         public struct Log
         {
-            private readonly LogLevel _level;
+            internal readonly LogLevel Level;
 
             internal Log(LogLevel level)
             {
-                _level = level;
+                Level = level;
             }
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -30,7 +30,7 @@ namespace Ryujinx.Common.Logging
             {
                 if (m_EnabledClasses[(int)logClass])
                 {
-                    Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, "", message)));
+                    Updated?.Invoke(null, new LogEventArgs(Level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, "", message)));
                 }
             }
 
@@ -39,7 +39,7 @@ namespace Ryujinx.Common.Logging
             {
                 if (m_EnabledClasses[(int)logClass])
                 {
-                    Updated?.Invoke(null, new LogEventArgs(_level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, message)));
+                    Updated?.Invoke(null, new LogEventArgs(Level, m_Time.Elapsed, Thread.CurrentThread.Name, FormatMessage(logClass, caller, message)));
                 }
             }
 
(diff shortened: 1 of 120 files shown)


jduncanator:
I'm not sure how this is any "less cost" than just doing the check in the PrintWarning function? The compiler still emits a null check that runs every time the log function is encountered, this just hoists it from the PrintXXX function, to the call site. Not sure if I'm missing something crucial here?

mageven:
The less cost comes from nullifying all the str-format/heap-allocs/func-calls that are used for the log calls.

jduncanator:
My big concern with going down this path is that it's extremely easy to write code that looks like this:

```cs
Logger.Error.Print()
```

And it may work perfectly fine on your machine with error logs enabled, but as soon as you disable error logs, you'll get a NRE and crash. There doesn't really seem to be an easy way to prevent this easy to make mistake at compile time.

mageven:
Hence what I wrote in the description:
> There doesn't seem to be a way to enforce a null check on every use. This could potentially be remedied with a simple grep for test builds (for PRs and such).

mageven:
Rebased with a better implementation.

@jduncanator I was just dumb. I didn't need nullable ref types. I get compile-time checks with nullable value types (`struct Log`) and everything works fine now. This should address your primary concerns while maintaining my goals.

9db5b12 is generated with a script.

Also, includes a commit to reduce SendVibrationXX spam (moved to Debug).

@gdkchan I changed `PrintGuestStackTrace` output to guest logs. I hope that's alright.

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -8,27 +8,94 @@ namespace Ryujinx.Common.Logging
...
+        private static readonly Stopwatch m_Time;
 
-        private static readonly bool[] m_EnabledLevels;
         private static readonly bool[] m_EnabledClasses;
 
         private static readonly List<ILogTarget> m_LogTargets;
 
         public static event EventHandler<LogEventArgs> Updated;
 
+
```
gdkchan, line 19: nit: Double empty line not necessary.

mageven:
Rebased. Added a new Log Level "Notice", which is always-enabled. It is used to print system info, enabled log levels, etc... (regardless of current log settings)

Also, UnhandledException and ProcessExit event hooks have been refactored to common functions and to make sure Logger is shutdown as late as possible.

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -88,88 +157,47 @@ public static void Shutdown()
...
-            if (m_EnabledLevels[(int)logLevel] && m_EnabledClasses[(int)logClass])
+            switch (logLevel)
             {
-                Updated?.Invoke(null, new LogEventArgs(logLevel, m_Time.Elapsed, Thread.CurrentThread.Name, message, data));
+                case LogLevel.Debug    : Debug     = enabled ? new Log(LogLevel.Debug)    : new Log?(); break;
+                case LogLevel.Info     : Info      = enabled ? new Log(LogLevel.Info)     : new Log?(); break;
+                case LogLevel.Warning  : Warning   = enabled ? new Log(LogLevel.Warning)  : new Log?(); break;
+                case LogLevel.Error    : Error     = enabled ? new Log(LogLevel.Error)    : new Log?(); break;
+                case LogLevel.Guest    : Guest     = enabled ? new Log(LogLevel.Guest)    : new Log?(); break;
+                case LogLevel.AccessLog: AccessLog = enabled ? new Log(LogLevel.AccessLog): new Log?(); break;
```
AcK77, line 192: Nit
```suggestion
                case LogLevel.AccessLog : AccessLog = enabled ? new Log(LogLevel.AccessLog) : new Log?(); break;
```
(Don’t forget to align other lines)

Ryujinx/Ui/MainWindow.cs
```diff
@@ -661,10 +661,6 @@ private void End(HLE.Switch device)
 
             Profile.FinishProfiling();
             DiscordIntegrationModule.Exit();
-            Logger.Shutdown();
-
-            Ptc.Dispose();
-            PtcProfiler.Dispose();
```
LDj3SNuD, line 667: I would prefer to add the following two lines of code again:
```csharp
            Ptc.Dispose();
            PtcProfiler.Dispose();
```
just before `Application.Quit();`, as these disposition methods are important for the PPTC system, having had experience of untriggered `AppDomain.ProcessExit` event.
The `Logger.Shutdown();` method could be re-added as well.

Note: the PR approval is only about PPTC changes, for the rest I tested the PR locally and found no issues.

mageven: > having had experience of untriggered AppDomain.ProcessExit event.

Didn't know this could happen. Do you recall the circumstances?

In any case, re-added the `Dispose` calls but reordered them so that `Logger.Shutdown()` is last.

LDj3SNuD: Please, as I said it is important that those 2 PPTC disposition methods are placed just before `Application.Quit();` and `Logger.Shutdown();`, thank you.

mageven: Sorry about that. Changed it to your suggestion.

PR #1293, opened 2020-06-03
Adds a Hack: Force Ryujinx to use the internal ASTC Decoder

PaintFoxV:
Gives the User the option to disable HW ASTC Texture Decoding via the GUI.


Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
```diff
@@ -127,6 +127,11 @@ public class ConfigurationFileFormat
         /// </summary>
         public bool IgnoreMissingServices { get; set; }
 
+        /// <summary>
+        /// Enable or disable ignoring missing services
```
riperiperi, line 131: This comment is wrong.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -187,6 +187,11 @@ public class SystemSection
             /// </summary>
             public ReactiveObject<bool> IgnoreMissingServices { get; private set; }
 
+            /// <summary>
+            /// Enable or disable ignoring missing services
```
riperiperi, line 191: This comment is also wrong.

Ryujinx/Ui/SettingsWindow.cs
```diff
@@ -37,6 +37,7 @@ public class SettingsWindow : Window
         [GUI] CheckButton  _multiSchedToggle;
         [GUI] CheckButton  _fsicToggle;
         [GUI] CheckButton  _ignoreToggle;
+        [GUI] CheckButton _astcforcedecodeToggle;
```
riperiperi, line 40: Align with other variables. 

Should probably be capitalized: `_astcForceDecodeToggle`

Ryujinx.Graphics.GAL/Capabilities.cs
```diff
@@ -20,7 +20,14 @@ public struct Capabilities
...
             int   storageBufferOffsetAlignment)
         {
-            SupportsAstcCompression          = supportsAstcCompression;
+            if(Ryujinx.Configuration.ConfigurationState.Instance.System.ForceSWASTC)
+            {
+                SupportsAstcCompression = false;
+            }
+            else
+            {
+                SupportsAstcCompression = supportsAstcCompression;
```
riperiperi, line 29: I really don't like the way the configuration is being accessed here. This if statement could also be removed as follows:

`SupportsAstcCompression = ConfigurationState.Instance.System.ForceSWASTC | supportsAstcCompression;`

jduncanator:
I'd prefer to handle this automatically. @ReinUsesLisp has suggested in the past that there are certain flags one can check to "determine" if the driver/GPU has hardware or software based ASTC support. The best way to handle this would be to use that information to choose driver supported ASTC decoding when it's hardware supported, and Ryujinx's internal decoder when it's software supported.

PaintFoxV:
Since Handling this automatically would be better and its now a Mess im closing this PR

PR #1294, opened 2020-06-04
Faster crc32 implementation

MerryMage:
Implement crc32 in terms of pclmulqdq.

Logic hopefully should be obvious.

---

Edit: It's essentially just a Barrett reduction.

image

The above image is from the Intel whitepaper. My implementation differs from Intel's because I optimized it a bit more with a few tricks.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitHash.cs
+++ b/ARMeilleure/Instructions/InstEmitHash.cs
@@ -4,6 +4,8 @@ using ARMeilleure.Translation;
 using System;
 
 using static ARMeilleure.Instructions.InstEmitHelper;
+using static ARMeilleure.Instructions.InstEmitSimdHelper;
+using static ARMeilleure.IntermediateRepresentation.OperandHelper;
 
 namespace ARMeilleure.Instructions
 {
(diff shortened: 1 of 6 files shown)


ARMeilleure/Instructions/InstEmitHash.cs
```diff
@@ -4,51 +4,162 @@
...
+                crc = context.AddIntrinsic(Intrinsic.X86Pslldq, crc, Const((64 - bitsize) / 8));
+                tmp = context.AddIntrinsic(Intrinsic.X86Pxor, tmp, crc);
+            }
+
+            SetIntOrZR(context, op.Rd, context.VectorExtract(OperandType.I32, tmp, 2));
+        }
+
+        private static void EmitCrc32Optimized64(ArmEmitterContext context, bool castagnoli)
+        {
+            OpCodeAluBinary op = (OpCodeAluBinary)context.CurrOp;
```
LDj3SNuD, line 140: To better identify those constants (and for this method avoid repeating them), I would suggest:
```csharp
            long properName1 = castagnoli ? 0x0DEA713F1 : 0x1F7011641; // u'
            long properName2 = castagnoli ? 0x105EC76F0 : 0x1DB710641; // P'(x) << 1
```
Same for the other method.

ARMeilleure/Instructions/InstEmitHash.cs
```diff
@@ -4,51 +4,162 @@
 using System;
 
 using static ARMeilleure.Instructions.InstEmitHelper;
+using static ARMeilleure.Instructions.InstEmitSimdHelper;
```
LDj3SNuD, line 7: For future reference, I would recommend adding:
`// https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/fast-crc-computation-generic-polynomials-pclmulqdq-paper.pdf`
To be added at the beginning of the file (line #1 followed by new line).

jduncanator:
Do the SSE4.2 CRC instructions not cover any of the ARM variants?

LDj3SNuD:
Yes, only the Castagnoli variants. At the time of the epic ChocolArm64 I implemented this intrinsic optimization https://github.com/Ryujinx/ChocolArm64/blob/master/Instructions/InstEmitHash.cs#L80 ; however, I don't know why it wasn't used in ARMeilleure as well.

PR #1298, opened 2020-06-11
Generalize tail continues

FICTURE7:
Introduce "Exit" blocks (essentially pseudo blocks) which represents a continuation.

The `Branch` and `Next` which would be null (before this PR) are now created as "Exit" blocks and are inserted into the ordered decoded block list. The `Translator` then emits tail continues for these "Exit" blocks and branch instructions can be emitted without the null checks as the "Exit" blocks are address labeled.

Reduces redundant tail continues. For example:
```asm
0x8000ac0:      C8 02 40 F9            ldr	x8, [x22]
0x8000ac4:      1F 1D 00 F1            cmp	x8, #7
0x8000ac8:      E8 17 9F 1A            cset	w8, eq
0x8000acc:      68 C2 00 39            strb	w8, [x19, #0x30]
0x8000ad0:      C8 02 40 F9            ldr	x8, [x22]
0x8000ad4:      1F 1D 00 F1            cmp	x8, #7
0x8000ad8:      20 FB FF 54            b.eq	loc_8000a3c ; <- Exit Block #0
0x8000adc:      1F 45 00 F1            cmp	x8, #0x11
0x8000ae0:      E0 FA FF 54            b.eq	loc_8000a3c ; <- Exit Block #0
0x8000ae4:      49 00 00 14            b	loc_8000c08 ; <- Exit Block #1
```
Before, anything after the first `b.eq	loc_8000a3c` would not be generated (becomes tail continues), and if that was fixed the two `b.eq	loc_8000a3c` would become 2 different tail continues even though they branch to the same address. This PR fixes both of these.

This works by inserting an exit block with address `8000a3c` in the ordered list of decoded blocks and so it is just another block in the block stream. The `Translator` then emits a tail continue for `8000a3c` (as it is an exit block). The two `b.eq	loc_8000a3c` branch instructions can then branch to same IR block.

Should reduce number of compilations by number of conditional branch fall through reached.

Supersedes #1246.

### DIFF
--- a/ARMeilleure/Decoders/Decoder.cs
+++ b/ARMeilleure/Decoders/Decoder.cs
@@ -19,14 +19,15 @@ namespace ARMeilleure.Decoders
 
         public static Block[] DecodeBasicBlock(IMemoryManager memory, ulong address, ExecutionMode mode)
         {
-            Block block = new Block(address);
-
-            FillBlock(memory, mode, block, ulong.MaxValue);
-
-            return new Block[] { block };
+            return Decode(memory, address, mode, highCq: false, singleBlock: true);
         }
 
         public static Block[] DecodeFunction(IMemoryManager memory, ulong address, ExecutionMode mode, bool highCq)
+        {
+            return Decode(memory, address, mode, highCq, singleBlock: false);
+        }
+
+        private static Block[] Decode(IMemoryManager memory, ulong address, ExecutionMode mode, bool highCq, bool singleBlock)
         {
             List<Block> blocks = new List<Block>();
 
@@ -44,7 +45,7 @@ namespace ARMeilleure.Decoders
                 {
                     block = new Block(blkAddress);
 
-                    if (opsCount > instructionLimit || !memory.IsMapped(blkAddress))
+                    if ((singleBlock && visited.Count >= 1) || opsCount > instructionLimit || !memory.IsMapped(blkAddress))
                     {
                         block.Exit = true;
                         block.EndAddress = blkAddress;
@@ -141,7 +142,10 @@ namespace ARMeilleure.Decoders
                 }
             }
 
-            TailCallRemover.RunPass(address, blocks);
+            if (!singleBlock)
+            {
+                TailCallRemover.RunPass(address, blocks);
+            }
 
             return blocks.ToArray();
         }
--- a/ARMeilleure/Decoders/Decoder.cs
+++ b/ARMeilleure/Decoders/Decoder.cs
@@ -42,12 +42,13 @@ namespace ARMeilleure.Decoders
             {
                 if (!visited.TryGetValue(blkAddress, out Block block))
                 {
+                    block = new Block(blkAddress);
+
                     if (opsCount > instructionLimit || !memory.IsMapped(blkAddress))
                     {
-                        return null;
+                        block.Exit = true;
+                        block.EndAddress = blkAddress;
                     }
-
-                    block = new Block(blkAddress);
 
                     workQueue.Enqueue(block);
 
@@ -71,6 +72,8 @@ namespace ARMeilleure.Decoders
                         throw new InvalidOperationException("Found duplicate block address on the list.");
                     }
 
+                    currBlock.Exit = false;
+
                     nBlock.Split(currBlock);
 
                     blocks.Insert(nBlkIndex + 1, currBlock);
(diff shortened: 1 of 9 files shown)


FICTURE7:
Sample [diff](https://www.diffchecker.com/BrIbpJXm) for "Do not synchronize when branching into exit block" bfd6312.

FICTURE7:
481247f fixes a regression introduced in this PR. 

Before tail continues would be emitted immediately so they did not emit a synchronization. However this PR treats tail continues as part of the decoded block stream and so backward branches to these tail continues would still emit the synchronization.

FICTURE7:
https://github.com/Ryujinx/Ryujinx/blob/d99b010c70c7f672b538951c62ea0c780faf7be3/ARMeilleure/Translation/Translator.cs#L237-L242

BTW is this invert transformation correct? Because if we invert the condition we also change which flags we compare against. Perhaps it might safer to change the `BranchIfTrue` emitted by `EmitCondBranch` to a `BranchIfFalse`?

riperiperi:
> 
> 
> https://github.com/Ryujinx/Ryujinx/blob/d99b010c70c7f672b538951c62ea0c780faf7be3/ARMeilleure/Translation/Translator.cs#L237-L242
> 
> BTW is this invert transformation correct? Because if we invert the condition we also change which flags we compare against. Perhaps it might safer to change the `BranchIfTrue` emitted by `EmitCondBranch` to a `BranchIfFalse`?

Invert just flips bit 0. In `InstEmitFlowHelper.GetCondTrue`, you can see that conditions with bit 0 set are identical to those without but with the final result being inverted, so it is identical to switching to a BranchIfFalse. 

Example:
```csharp
                case Condition.Eq: //0
                    value = GetFlag(PState.ZFlag);
                    break;

                case Condition.Ne: //1
                    value = Inverse(GetFlag(PState.ZFlag));
                    break;
```

FICTURE7:
Thanks, makes sense! 👍

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -271,39 +260,48 @@ public static void EmitDynamicTableCall(ArmEmitterContext context, Operand table
             Operand endLabel = Label();
             Operand fallbackLabel = Label();
 
-            Action<Operand> emitTableEntry = (Operand entrySkipLabel) =>
+            void EmitTableEntry(Operand entrySkipLabel)
```
riperiperi, line 263: I always forget you can do this.

ARMeilleure/Decoders/Optimizations/TailCallRemover.cs
```diff
@@ -54,22 +58,32 @@ public static void RunPass(ulong entryAddress, List<Block> blocks)
...
+                    block.Branch.Exit     = true;
+                    block.Branch.TailCall = true;
                 }
             }
 
             // Finally, delete all blocks outside the contiguous range.
-
-            blocks.RemoveRange(endBlockIndex + 1, (blocks.Count - endBlockIndex) - 1);
-            blocks.RemoveRange(0, startBlockIndex);
+            RemoveDeadBlocks(blocks, 0, startBlockIndex - 1);
```
gdkchan, line 85: What's the reasoning behind this change (changing blocks to null instead of completely removing them from the list)?

FICTURE7: Yes, this was intended to be an optimization, since we do this on a 1 by 1 basis, it can get relatively expensive with the redundant copies.

If the possible null entries are not desirable, we can build a new array/list in the tailcall remover itself. Should avoid the redundant copies issue and also improve the case where the tailcall remover nukes a good chunk of the decoded blocks and the Translator ends up iterating over a lot of null blocks.

ARMeilleure/Decoders/Decoder.cs
```diff
@@ -135,7 +132,10 @@ Block GetBlock(ulong blkAddress)
                 }
             }
 
-            TailCallRemover.RunPass(address, blocks);
+            if (!singleBlock)
+            {
+                TailCallRemover.RunPass(address, blocks);
```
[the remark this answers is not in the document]
FICTURE7: Sounds good to me. 👍

FICTURE7:
> but it might be good to get some more testing in anyways

Most certainly. :>

> PTC is finally getting merged soon

🎉

riperiperi:
Needs a rebase due to PTC.

ARMeilleure/Decoders/Decoder.cs
```diff
@@ -17,16 +17,7 @@ static class Decoder
...
-        {
-            Block block = new Block(address);
-
-            FillBlock(memory, mode, block, ulong.MaxValue);
-
-            return new Block[] { block };
-        }
-
-        public static Block[] DecodeFunction(IMemoryManager memory, ulong address, ExecutionMode mode, bool highCq)
+        public static Block[] Decode(IMemoryManager memory, ulong address, ExecutionMode mode, bool highCq, bool singleBlock)
```
FICTURE7, line 20: Should I remove the ability to do single block decodes? Since we do not use it anymore.

riperiperi: No, it may be useful in the future. One potential use is aggressive self modifying code, which homebrew/emulators could do.

PR #1300, opened 2020-06-13
VABS takes one input register, not two.

riperiperi:
This fixes an issue where part of the VABS instruction would be parsed as an input register, triggering an "undefined" instruction specifically when using it with quadword vectors. It registered the phantom input register as invalid, as it was not divisible by two.


riperiperi:
I'll make a follow up PR adding VMVN (register) as well as fixing this, with tests.

LDj3SNuD:
All right, thanks.

gdkchan:
> 
> 
> I'll make a follow up PR adding VMVN (register) as well as fixing this, with tests.

Since you mentioned VMVN, would be nice to also take a look at this (it looks wrong): https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Instructions/InstEmitSimdMove32.cs#L37

jduncanator:
Do we not have tests validating this behaviour?

riperiperi:
> 
> 
> Do we not have tests validating this behaviour?

If a test existed for VABS, it would have caught that the quadword variant doesn't work. However there are a lot of holes in the testing for A32 SIMD instructions right now, specifically floating point ones.

PR #1302, opened 2020-06-15
Update README.md

EmulationFanatic:
Updates out of date documentation in the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Every file related to Ryujinx is stored in the `Ryujinx` folder. Located in `C:\
 
 ## Latest build
 
-These builds are compiled automatically for each commit on the master branch. While we strive to ensure optimal stability and performance prior to pushing an update, even though they are official these builds **may be unstable or completely broken.**
+These builds are compiled automatically for each commit on the master branch. While we strive to ensure optimal stability and performance prior to pushing an update, our automated builds **may be unstable or completely broken.**
 
 The latest automatic build for Windows, macOS, and Linux can be found on the [Official Website](https://ryujinx.org/download).
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ If you use a pre-built version, you can use the graphical interface to run your 
 If you build it yourself you will need to:
 **Step one:** Install the [.NET Core 3.1 (or higher) SDK](https://dotnet.microsoft.com/download/dotnet-core).
 
-**Step two (choose one):**
+**Step two (choose one):**  
 **(Variant one)**
 
 After the installation of the Net Core SDK is done; go ahead and copy the Clone link from GitHub from here (via Clone or Download --> Copy HTTPS Link. You can Git Clone the repo by using Git Bash or Git CMD.
--- a/README.md
+++ b/README.md
@@ -10,32 +10,42 @@
 </h1>
 
 <p align="center">
-    <i>An Experimental Switch emulator written in C#</i><br />
+    <i>An experimental Switch emulator written in C#</i><br />
     <br />
     <img src="https://i.imgur.com/JDLmXJ6.png">
 </p>
 
 <h5 align="center">
-    A lot of games boot, but only some are playable. See the compatiblity list <a href="https://github.com/Ryujinx/Ryujinx-Games-List/issues" target="_blank">here</a>.
+    As of June 2020, Ryujinx goes past menus and in-game on over 1,000 commercial titles. Of those, roughly half are considered playable. See the compatiblity list <a href="https://github.com/Ryujinx/Ryujinx-Games-List/issues" target="_blank">here</a>.
 </h5>
 
 ## Usage
 
-To run this emulator, you need the [.NET Core 3.1 (or higher) SDK](https://dotnet.microsoft.com/download/dotnet-core).
-
-If you use a pre-built version, you can use the graphical interface to run your games and homebrew.
+To run this emulator, we recommend that your PC have at least 8GB of RAM; less than this amount can result in unpredictable behavior and may cause crashes or unacceptable performance.
+If you use a pre-built version, you can use the graphical interface to run your games and homebrew: simply add the directory containing your homebrew or games in the Options > Settings > General tab > Game Directories menu item.
 
 If you build it yourself you will need to:
-Run `dotnet run -c Release -- path\to\homebrew.nro` inside the Ryujinx project folder to run homebrew apps.
-Run `dotnet run -c Release -- path\to\game.nsp/xci` to run official games.
+Step One: Install the [.NET Core 3.1 (or higher) SDK](https://dotnet.microsoft.com/download/dotnet-core).
+Step two (Variant one):
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,42 +10,53 @@
...
+If you use a pre-built version, you can use the graphical interface to run your games and homebrew: simply add the directory containing your homebrew or games in the Options > Settings > General tab > Game Directories menu item.
 
 If you build it yourself you will need to:
-Run `dotnet run -c Release -- path\to\homebrew.nro` inside the Ryujinx project folder to run homebrew apps.
-Run `dotnet run -c Release -- path\to\game.nsp/xci` to run official games.
+**Step one:** Install the [.NET Core 3.1 (or higher) SDK](https://dotnet.microsoft.com/download/dotnet-core).
+
+**Step two (Variant one):**
+
+After the installation of the Net Core SDK is done; go ahead and copy the Clone link from GitHub from here (via Clone or Download --> Copy HTTPS Link. Or you can download the ZIP tarball.) You can Git Clone the repo by using the GitBash, or you may use the second variant.
```
gdkchan, line 32: > Or you can download the ZIP tarball.

This is already said on "variant two". I think it's not necessary to repeat this information here.
> or you may use the second variant.

Not sure if this part is necessary either. We could just state, before step two, that the user should choose one of the variants, instead of doing both.

README.md
```diff
@@ -10,42 +10,53 @@
...
+
+**Step two (Variant two):**
+
+Download the ZIP Tarball. Then extract it to a directory of your choice.
+
+**Step three:**
+
+Build the App using a Command prompt in the ROOT directory. You can quickly access it by Holding shift in explorer (in the Ryujinx directory) then right clicking, and typing the following command
```
gdkchan, line 40: I think it's missing `:` at the end?

gdkchan: It's also unclear what's the "ROOT" directory here. On the line below it says it should be built from the Ryujinx project directory. I would expect the root folder to be the one where the solution is located.

README.md
```diff
@@ -10,42 +10,53 @@
...
  - **Switch Keys**
 
    Everything on the Switch is encrypted, so if you want to run anything other than homebrew, you have to dump encryption keys from your console. To get more information please take a look at our [Keys Documentation](KEYS.md).
 
- - **System Titles**
-
-   Some of our System Module implementations, like `time`, require [System Data Archives](https://switchbrew.org/wiki/Title_list#System_Data_Archives). You can install them by mounting your nand partition using [HacDiskMount](https://switchtools.sshnuke.net/) and copying the content to `Ryujinx/bis/system`.
+ - **Firmware**
+    
+    You need an official Switch firmware by either dumping directly from your Switch or dumping your cartridge game to an XCI file. Install the firmware, after you've installed your keys, from the Tools > Install Firmware menu item.
```
gdkchan, line 59: I think the correct here is "game cartridge" rather than "cartridge game". The instructions could be also a bit more clear on the fact that you can directly install the firmware from a XCI file (as long it was not "trimmed").

README.md
```diff
@@ -59,23 +70,27 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
 
  - **Audio**
 
-   Everything for audio is partially supported. We currently use a C# wrapper for [libsoundio](http://libsound.io/), and we support [OpenAL](https://openal.org/downloads/OpenAL11CoreSDK.zip) (installation needed) too as a fallback. Our current Opus implementation is pretty incomplete.
+   Everything for audio is partially supported. We currently use a C# wrapper for [libsoundio](http://libsound.io/), and we support [OpenAL](https://openal.org/downloads/OpenAL11CoreSDK.zip) (installation needed) too as a fallback. Our current Opus implementation is incomplete.
```
gdkchan, line 73: I think that could just be "Audio is partially supported". "Everything for" is just making the sentence more confusing. Also, currently OpenAL is the main backend, and SoundIO is the fallback (as it doesnt support 5.1 audio output). This could easily change in the not-so-distant future though,

README.md
```diff
@@ -59,23 +70,27 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
...
 
 - **CPU**
 
-  The CPU emulator, ARMeilleure, emulates an ARMv8 CPU, and currently only has support for the new 64-bit ARMv8 instructions (with a few instructions still missing). It translates the ARM code to a custom IR, performs a few optimizations, and turns that into x86 code. To handle that, we use our own JIT called ARMeilleure, which uses the custom IR and compiles the code to x86.
+  The CPU emulator, ARMeilleure, emulates an ARMv8 CPU, and currently only has support for the new 64-bit ARMv8 instructions (with a few instructions still missing). It translates the ARM code to a custom IR, performs a few optimizations, and turns that into x86 code. To handle that, we use our own JIT called ARMeilleure, which uses the custom IR and compiles the code to x86.
```
gdkchan, line 77: > and currently only has support for the new 64-bit ARMv8 instructions (with a few instructions still missing).

This is a bit outdated, we have 32-bit support now, so we support some of the ARMv7 (and older) instructions aswell. The support is way more incomplete than ARMv8-only 64-bit instructions, but some 32-bit games are already playable.

gdkchan: Also I think "the custom IR" should be "a custom IR".

README.md
```diff
@@ -59,23 +70,27 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
...
+   We currently have support for keyboard, mouse, touch input, JoyCon input support emulated through the keyboard, and most controllers. Controller support varies by operating system, as outlined below.  
+   Windows: Xinput-compatible controllers are supported natively; other controllers can be supported with the help of Xinput wrappers such as x360ce.  
+   Linux: most modern controllers are supported.  
+   In either case, you can set up everything inside the input configuration menu.
 
 - **Configuration**
 
-   The emulator has settings for dumping shaders, enabling or disabling some logging, remapping controllers, and more. You can configure all of them through the graphical interface or manually through the config file, `Config.json`.
+   The emulator has settings for enabling or disabling some logging, remapping controllers, and more. You can configure all of them through the graphical interface or manually through the config file, `Config.json`, found in the user's appdata\roaming\ryujinx folder.
```
gdkchan, line 93: That path is only valid for windows, maybe we should be more "OS agnostic" on the readme. It could just say that it can be found on the user folder, and mention that the GUI has a option to open such folder.

EmulationFanatic:
Updated per gdkchan's recommendations.

README.md
```diff
@@ -10,42 +10,54 @@
...
 
  - **Switch Keys**
 
    Everything on the Switch is encrypted, so if you want to run anything other than homebrew, you have to dump encryption keys from your console. To get more information please take a look at our [Keys Documentation](KEYS.md).
 
- - **System Titles**
-
-   Some of our System Module implementations, like `time`, require [System Data Archives](https://switchbrew.org/wiki/Title_list#System_Data_Archives). You can install them by mounting your nand partition using [HacDiskMount](https://switchtools.sshnuke.net/) and copying the content to `Ryujinx/bis/system`.
+ - **Firmware**
+
```
Thog, line 59: extraneous spaces.
```suggestion

```

PR #1303, opened 2020-06-15
Fix VMVN (immediate), Add VPMIN, VPMAX, VMVN (register)

riperiperi:
Just a few more instructions I've seen games using. Should compliment valx's PR pretty well.

VMVN (immediate) has been renamed from `VMVN_I` to `VMVN_II`, to make way for the integer variant, and it has also been fixed to use BitwiseNot rather than xor the input with itself for whatever reason.

I've added new tests for all the instructions listed, and added a better test for VPADD (and its integer variant). The floating point pairwise instruction tests will only pass if FastFP is false, but I have verified that the intrinsic versions do work with regular values that aren't NaNs, Infinities and whatnot.

The non floating point tests also pass with intrinsic paths disabled. (VMVN, pairwise integer)

### DIFF
--- a/Ryujinx.Tests/Cpu/CpuTestSimdMov32.cs
+++ b/Ryujinx.Tests/Cpu/CpuTestSimdMov32.cs
@@ -39,8 +39,8 @@ namespace Ryujinx.Tests.Cpu
                 0b1110_1
             };
 
-
             uint opcode = 0xf2800010u; // VMOV.I32 D0, #0
+
             uint cmodeOp = variants[variant];
 
             if (q)
@@ -49,11 +49,11 @@ namespace Ryujinx.Tests.Cpu
             }
 
             opcode |= ((cmodeOp & 1) << 5) | ((cmodeOp & 0x1e) << 7);
-            opcode |= ((q ? 1u : 0u) << 6);
+            opcode |= (q ? 1u : 0u) << 6;
             opcode |= (imm & 0xf) | ((imm & 0x70) << 12) | ((imm & 0x80) << 16);
 
-            opcode |= ((vd & 0x10) << 18);
-            opcode |= ((vd & 0xf) << 12);
+            opcode |= (vd & 0x10) << 18;
+            opcode |= (vd & 0xf) << 12;
 
             SingleOpcode(opcode);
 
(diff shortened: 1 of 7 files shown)


jduncanator:
Where does the `II` mnemonic come from? Is it used in the ARMv8 documentation?

riperiperi:
> 
> 
> Where does the `II` mnemonic come from? Is it used in the ARMv8 documentation?

In the documentation, variants aren't part of the instruction name. The differentiation is for our implementation only. In documentation, it is called `VMVN (immediate)`, versus `VMVN (register)`. In other instructions, `_I` is used to denote an instruction that operates on i32 - gdk introduced `_II` in a previous PR to denote integer _and_ immediate. The naming is kind of silly but I'd prefer changing it all at once if we come up with something better.

riperiperi:
Rebased on top of PTC. Version gets set to 1.

riperiperi:
I snuck in the opcode for `Nop`, since deadly premonition needs it.

gdkchan:
Needs another version increment now.

riperiperi:
Rebased, PTC version set to 2.

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
```diff
@@ -12,6 +12,31 @@ public sealed class CpuTestSimdReg32 : CpuTest32
...
 #if SimdReg32
 
+#region "ValueSource (Opcodes)"
+        private static uint[] _Vp_Add_Max_Min_F_()
+        {
+            return new uint[]
+            {
+                0xf3000d00u, // VPADD.F32 D0, D0, D0
+                0xf3000f00u, // VPMAX.F32 D0, D0, D0
+                0xf3200f00u, // VPMIN.F32 D0, D0, D0
```
LDj3SNuD, line 22: Nit: remove comma.

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
```diff
@@ -12,6 +12,31 @@ public sealed class CpuTestSimdReg32 : CpuTest32
...
+        // VPADD does not have an unsigned flag, so we check the opcode before setting it.
+        private static uint VpaddI8 = 0xf2000b10u; // VPADD.I8 D0, D0, D0
+
+        private static uint[] _Vp_Add_Max_Min_I_()
+        {
+            return new uint[]
+            {
+                VpaddI8,
+                0xf2000a00u, // VPMAX.S8 D0, D0, D0
+                0xf2000a10u, // VPMIN.S8 D0, D0, D0
```
LDj3SNuD, line 35: Nit: remove comma.

Ryujinx.Tests/Cpu/CpuTestSimdMov32.cs
```diff
@@ -258,6 +258,82 @@ public sealed class CpuTestSimdMov32 : CpuTest32
             CompareAgainstUnicorn();
         }
 
+        [Test, Pairwise, Description("VMVN.<size> <Vt>, <Vm>")]
+        public void Vmvn([Range(0u, 1u, 2u)] uint size,
+                         [Values(0u, 1u, 2u, 3u)] uint vd,
+                         [Values(0u, 2u, 4u, 8u)] uint vm,
+                         [Values()] bool q)
```
LDj3SNuD, line 265: Nit: remove round brackets.

riperiperi:
That is a good idea - it would be nice to have these tests (A32/A64) working in some capacity for the fast paths.

I've fixed the nits you brought up.

PR #1334, opened 2020-06-30
Changes to game list reloading

### LINKED ISSUE #1331: GUI Change - Only Reload Game List if Game Directory is Changed
Currently, you may change and save any number of menu options *while the list is loading* and the list will continue loading anyway. You can even change the game directory but Ryujinx will continue loading the old one until it's finished. The only way to stop this is by closing Ryjujinx and re-opening it.

However, once the list has finished loading, changing and saving any menu option whatsoever will result in a full reload of the list. 
The list also reloads every time after you use Esc to exit emulation.

Please consider changing the behavior to the following:
A list reload is triggered only by changing the game directory. This reload should occur whether the list is currently loading or not.

Xpl0itR:
Only reload list when game dirs are changed

closes #1331

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -321,7 +321,7 @@ namespace Ryujinx.Ui
 
         internal static void UpdateGameTable()
         {
-            if (_updatingGameTable)
+            if (_updatingGameTable || _gameLoaded)
             {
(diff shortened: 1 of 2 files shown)


EmulationFanatic:
Will this also cancel an ongoing list load process if the game dir is changed while it is being performed?

PR #1351, opened 2020-07-02
Support inline index buffer data

gdkchan:
This is used by some OpenGL games, presumably when `glDrawElements` is called on the guest side without a index buffer bound (in this case the data is passed directly on this call). Internally the driver should produce a sequence of `VbElement*` calls on the command buffer to pass the data to the GPU.

There's still a bunch of unknows that would be nice to test on hardware (like, what happens if you call it outside of GlBegin/GlEnd, what's the maximum size you can push through, what happens if you set the other index count before draw, etc). However this should work for all normal usages.

Fixes black screen on some OpenGL, games, including:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/450


riperiperi:
Fixes models in Xash3D when used with https://github.com/Ryujinx/Ryujinx/pull/986 .

image

Ryujinx.Graphics.Gpu/Engine/MethodDraw.cs
```diff
@@ -1,5 +1,9 @@
 using Ryujinx.Graphics.Gpu.State;
 using Ryujinx.Graphics.Gpu.Image;
+using Ryujinx.Graphics.GAL;
+using System;
+using System.Runtime.InteropServices;
+using Ryujinx.Common;
```
riperiperi, line 6: Order usings.

PR #1358, opened 2020-07-04
Fold ZeroExtend8/16/32 imm32/64

FICTURE7:
Can improve codegen of code dealing with a lot of 32 bit integers. For example:

```patch
 block61 (next block62) (branch block63):
- i64 %124 = ZeroExtend32 i32 0x1
- i64 %125 = ZeroExtend32 i32 0x5
- i64 %126 = ZeroExtend32 i32 0x14
- i64 %127 = ZeroExtend32 i32 0x3
  i64 %128 = Add i64 %18, i64 0xE0
  i64 %129 = BitwiseAnd i64 %128, i64 0xFFFFFF8000000000
  BranchIfFalse i64 %129
```

These gets constant folded and copy propagated into their use.

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
+++ b/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
@@ -206,6 +206,39 @@ namespace ARMeilleure.CodeGen.Optimizations
                     }
                     break;
 
+                case Instruction.ZeroExtend16:
+                    if (type == OperandType.I32)
+                    {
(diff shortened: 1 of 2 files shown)


riperiperi:
Oh yeah, just mirroring this here since I forgot to mention:

> This will change compiled code, so PTC version should probably be raised. (the benefits will not apply to existing caches)

jduncanator:
@FICTURE7 I've rebased and bumped the PPTC version in your branch, hope you don't mind 😅

FICTURE7:
@jduncanator Not at all, thank you for doing so!

PR #1359, opened 2020-07-04
Fold ConvertI64ToI32 imm64

FICTURE7:
Creates more opportunities for copy propagation. For example:

```patch
 block82 (branch block85):
- i32 %179 = ConvertI64ToI32 i64 0x0
- Call i64 0x7FF85FA30444, i64 %177, i32 %179 ;; WriteByte
+ Call i64 0x7FF888910444, i64 %171, i32 0x0 ;; WriteByte
  Branch
```

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
+++ b/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
@@ -78,6 +78,13 @@ namespace ARMeilleure.CodeGen.Optimizations
                     }
                     break;
 
+                case Instruction.ConvertI64ToI32:
+                    if (type == OperandType.I32)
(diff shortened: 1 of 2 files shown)


FICTURE7:
> I'm also sure there are some cases where I64-I32 conversion can be dropped when using W registers, though that's outside the scope of constant folding.

Right, I made some experiment with propagating `ZeroExtend32` into their `ConvertI64ToI32` use a while back, and it was a quite successful experiment. Most of the remaining redundant zero extend and convert i64 to i32 pairs lived across phi nodes. But I believe a more appropriate solution would probably be value numbering, which can also eliminate the common case of duplicated `LoadArgument` operations as well.

PR #1360, opened 2020-07-04
A couple fixes for the audio backend GUI option

Xpl0itR:
Detect available audio backends in the gui on a separate thread and hardcode dummy backend in the dropdown and select it by default.


jduncanator:
I suspect gtk does not like updates from non-GUI threads. Using a Task or a separate Thread breaks that. You might need to use Invoke to run the update code on the main thread.

Xpl0itR:
👍

PR #1361, opened 2020-07-04
prepo: Update implementation for 10.x changes

Thog:
On 10.x, Nintendo changed yet another time the prepo SaveReport &
SaveReportWithUser command ids.

This PR add support for command 10104 & 10105 and update naming of the
old variants to match switchbrew.

This fix any game using 10.x based sdknso and telemetry. (like AC:NH 1.3.0)


jduncanator:
If we don't care about the differences, I'd prefer using a single function with two attributes (for both commands), rather than cluttering the code with duplicate methods that are identical.

PR #1362, opened 2020-07-05
Improve kernel WaitSynchronization syscall implementation

gdkchan:
This updates the `WaitSynchronization` syscall implementation to match what the kernel does.
This mostly just add more checks/missing errors that the kernel returns. I don't expect it to improve anything in relation to official games (and hopefully also not regress anything). I recommend testing with a few games if possible, just to be sure that it didn't break anything (I did so myself).


jduncanator:
Kakista reports regressions in MK8D Chinese Version with this PR.

PR #1364, opened 2020-07-05
Fix Fullscreen toggling

emmauss:
Fixes the issue where the fullscreen toggle does not update it's label properly. Also fixes the issue where the menu and footer do not hide when you are starting a game while already in fullscreen.


EmulationFanatic:
As mentioned on Discord, I cannot reproduce the incorrect labeling issue (using the game list menu to enter fullscreen, and the label not updating properly to be "Exit fullscreen") on master.
However, I was able to confirm that using this PR does indeed make the menu & footer disappear properly once a game has been launched.

PR #1365, opened 2020-07-05
Implement Zero-Configuration Resolution Scaling

riperiperi:
sand
Click for a surprise.

# A note to the users

If you're using a **1080p screen**, I recommend **2x scaling** (don't go above, you will lose super sampling). For **4K**, I recommend **2x or greater** - though you should only go higher when your GPU can definitely handle it. Remember, 4x (8K) is _16x the switch's pixel count!_

All of this is very much a work in progress and will improve over time. In cases where the scaling fails, your game will likely just run at 1x. If your game has some weird issue when scaled up ~(like LM3's black screen)~ _(fixed!)_, then try booting with native res and switching when ingame - it may work. Make sure to point out edge cases so that they can be fixed in future!

If your target game has a form of anti aliasing, you probably want to make a mod to disable it so you can get the raw upscaled resolution.

# Zero-Configuration Resolution Scaling

This PR introduces a feature and configuration option that lets you scale compatible render targets in a game using a single multiplier value, which you can set via a combo list from 1x to 4x... or if you're feeling brave, any floating point number.

The goal is to scale shadows and screen-bound render targets using the scaling factor.

image
4a3cac36fadba7d4943ec79d1e9bb55d
https://gyazo.com/4a3cac36fadba7d4943ec79d1e9bb55d

You can change this value while emulation is running to get a feel for how it affects game graphics and performance. The changes are applied instantly, so have fun with it.

## Method
Block chain AI machine learning on the cloud. Just kidding, it's really simple.

Any Texture2D or Texture2DArray format texture is immediately _eligible_ for scaling. When the texture is taken as a render target and it is still eligible (or its first request is as a render target), its host texture is copied to a new one with the configured scale.

Textures have a variety of ways they can become _blacklisted_ from scaling. Any texture that is not of the two formats specified above is immediately blacklisted, and any texture that is a view of a blacklisted texture will also become blacklisted itself. 

Since textures can become views of each other _after_ they are created, this means that a texture could be scaled before it becomes a view. Textures that are modified on the CPU memory after their creation or flushed back to it also become blacklisted before promising the change. When a texture is blacklisted _after_ it is scaled, it is copied back to a host texture with 1x scale. If this happens to a view, its parent will be scaled to 1x and _all views_ will be recreated and blacklisted.

When rendering to a texture, we need to make sure that the scales of all bound targets are equal. To facilitate this, all eligible targets are scaled, we ensure that all _scaled targets_ match the scale in the configuration (that can be changed at runtime, mind), and if one target is blacklisted, then all of them are. This scale is used for a few things:

- Scaling all Viewports
- Scaling all Scissor Rectangles
- Sent to the Fragment Shader (fp_renderScale[0])

When using a scaled texture in a shader, thankfully there are only a few cases in which the scale of the texture can negatively affect anything:

- `gl_FragCoord.xy`: This is in screen space. As in, it is measured in pixels. If we render to a larger target, the values returned by this will be multiplied by the scaling factor, whereas the shaders that the game uses will be expecting the original, unmodified scale. To solve this problem, simply divide by the scaling factor `(fp_renderScale[0])`. Problem solved.
- `texelFetch`: This method of sampling a texture samples using an integer vector. The unit is pixels. Repeat what I said above, but this time we need scaling factors for _each texture passed to the shader_. The texture descriptor index is used to index the renderScale array, which contains scales for bound textures starting at index 1.
  - There are cases where we don't currently allow scaling for shaders that use this feature. See below for details.

## Related Gpu Changes
A few changes were put into place to facilitate this change, and a few others that I'm planning or could be useful in the future. There were also a few texture related bugs that I ran into when ensuring runtime switching worked as intended.

- Fixed a bug where scaled copy's source/destination region would be affected by targets previously bound to the copy framebuffers (eg. an old depth stencil was active during a future copy of a colour texture of different size). Now, they unset their texture after the copy.
- Fixed a bug where transforming a texture to a view after its creation would not set its `_firstLayer`, `_firstLevel` information.
- Added `FeatureFlags` to the IR stage of translation, for consumption in the CodeGenerator. These will typically be used to change the declarations used in the shader on the backend. Here, we use it to add the renderScale uniform and the TexelFetchScale helper when needed.
- Added `TextureUsageFlags` to texture descriptors, set based on how they are used in the texture, and intended for use when the texture is bound. This PR sets and uses a flag to blacklist textures that use texelFetch in unsupported situations.
- Width and Height for the host texture now reflects its scaled width and height.

## Compatibility notes
- Games that use spatially aware effects aggressively (such as Xenoblade's AO) will likely only work correctly with power of two scales (1, 2, 4) - with others they will run into unusual issues. Games that use texelFetch will run into issues in general with non integral scales. With the games I have played, these issues are few and far between.
- Games that use a form of Anti Aliasing will still _think_ they are running at 1x, and will likely blur the image. For these games (such as xenoblade... again) you should use a LayeredFS mod to remove the AA. For completeness, I have attached a mod for XC:DE to this PR.
- Scales of less than 1 can reduce texture quality on rendered textures as well. You have been warned!

## Drawbacks
While every game is technically supported, there are cases where the game will fall back to 1x on targets you might not want to, and the game might have some graphical issues in the worst case. Here's some specific information:

- Only non-indexed textures on the fragment+compute shaders are fully supported for texelFetch scaling. If a texture outside of this category uses texelFetch, the texture will be blacklisted for scaling, as described above.
- Bindless textures are not fully supported, and do not have a blacklist fallback. If they are used with texelFetch, the scale on the texture _will_ be incorrect.
- If a game uses a render target to store data (such as the output of a fragment shader DXT compression algorithm), scaling will not work correctly on it. You can only hope that it will blacklist and render the target again at a later point. This is a theory as I haven't tested any games that do this.

## Screenshots
Here are some pictures at 4K. A lot of these are taken at 4x (8K) downsampled to 4K, which results in a supersampling effect on NVIDIA. Note that you should not run at more than double your output resolution if you want a supersampling effect, so use 2x for 1080p.

[Astral Chain Comparison](https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=8b116232-bfcd-11ea-bf88-a15b6c7adf9a)
astraljd
lm3
**(LM3 _must_ start in 1x scaling mode, then you can change the scale from settings ingame. This is being investigated :) )**
ac4x
bigjpg
bokeh
castle2
mk8

Finally, here's the "remove aa" mod I made for Xenoblade, so you can replicate these results (need to merge `mageven`'s PR)
[XenobladeDE remove-aa.zip](https://github.com/Ryujinx/Ryujinx/files/4875408/XenobladeDE.remove-aa.zip)

### DIFF
--- a/Ryujinx.Graphics.GAL/Format.cs
+++ b/Ryujinx.Graphics.GAL/Format.cs
@@ -214,5 +214,49 @@ namespace Ryujinx.Graphics.GAL
 
             return false;
         }
+
+        public static bool IsDepthOrStencil(this Format format)
+        {
+            switch (format)
+            {
+                case Format.D16Unorm:
+                case Format.D24UnormS8Uint:
+                case Format.D24X8Unorm:
+                case Format.D32Float:
+                case Format.D32FloatS8Uint:
+                case Format.S8Uint:
+                    return true;
+            }
+
+            return false;
+        }
+
+        public static bool HasOneComponent(this Format format)
+        {
+            switch (format)
+            {
+                case Format.R8Unorm:
+                case Format.R8Snorm:
+                case Format.R8Uint:
+                case Format.R8Sint:
+                case Format.R16Float:
+                case Format.R16Unorm:
+                case Format.R16Snorm:
+                case Format.R16Uint:
+                case Format.R16Sint:
+                case Format.R32Float:
+                case Format.R32Uint:
+                case Format.R32Sint:
+                case Format.R8Uscaled:
+                case Format.R8Sscaled:
+                case Format.R16Uscaled:
+                case Format.R16Sscaled:
+                case Format.R32Uscaled:
+                case Format.R32Sscaled:
+                    return true;
+            }
+
+            return false;
+        }
     }
 }
--- a/Ryujinx.Graphics.GAL/Format.cs
+++ b/Ryujinx.Graphics.GAL/Format.cs
@@ -252,6 +252,78 @@ namespace Ryujinx.Graphics.GAL
         }
 
         /// <summary>
+        /// Checks if the texture format is an unsigned integer color format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is an unsigned integer color format, false otherwise</returns>
+        public static bool IsUint(this Format format)
+        {
+            switch (format)
+            {
+                case Format.R8Uint:
+                case Format.R16Uint:
+                case Format.R32Uint:
+                case Format.R8G8Uint:
+                case Format.R16G16Uint:
+                case Format.R32G32Uint:
+                case Format.R8G8B8Uint:
+                case Format.R16G16B16Uint:
+                case Format.R32G32B32Uint:
+                case Format.R8G8B8A8Uint:
+                case Format.R16G16B16A16Uint:
+                case Format.R32G32B32A32Uint:
+                case Format.R10G10B10A2Uint:
+                case Format.R8G8B8X8Uint:
+                case Format.R16G16B16X16Uint:
+                case Format.R32G32B32X32Uint:
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks if the texture format is a signed integer color format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is a signed integer color format, false otherwise</returns>
+        public static bool IsSint(this Format format)
+        {
+            switch (format)
+            {
+                case Format.R8Sint:
+                case Format.R16Sint:
+                case Format.R32Sint:
+                case Format.R8G8Sint:
+                case Format.R16G16Sint:
+                case Format.R32G32Sint:
+                case Format.R8G8B8Sint:
+                case Format.R16G16B16Sint:
+                case Format.R32G32B32Sint:
+                case Format.R8G8B8A8Sint:
+                case Format.R16G16B16A16Sint:
+                case Format.R32G32B32A32Sint:
+                case Format.R10G10B10A2Sint:
+                case Format.R8G8B8X8Sint:
+                case Format.R16G16B16X16Sint:
+                case Format.R32G32B32X32Sint:
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks if the texture format is an integer color format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is an integer color format, false otherwise</returns>
+        public static bool IsInteger(this Format format)
+        {
+            return format.IsUint() || format.IsSint();
+        }
+
+        /// <summary>
         /// Checks if the texture format only has one component.
         /// </summary>
         /// <param name="format">Texture format</param>
--- a/Ryujinx.Graphics.GAL/Format.cs
+++ b/Ryujinx.Graphics.GAL/Format.cs
@@ -162,11 +162,21 @@ namespace Ryujinx.Graphics.GAL
 
     public static class FormatExtensions
     {
+        /// <summary>
+        /// Checks if the texture format is an ASTC format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is an ASTC format, false otherwise</returns>
         public static bool IsAstc(this Format format)
         {
             return format.IsAstcUnorm() || format.IsAstcSrgb();
         }
 
+        /// <summary>
+        /// Checks if the texture format is an ASTC Unorm format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is an ASTC Unorm format, false otherwise</returns>
         public static bool IsAstcUnorm(this Format format)
         {
             switch (format)
@@ -191,6 +201,11 @@ namespace Ryujinx.Graphics.GAL
             return false;
         }
 
+        /// <summary>
+        /// Checks if the texture format is an ASTC SRGB format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is an ASTC SRGB format, false otherwise</returns>
         public static bool IsAstcSrgb(this Format format)
         {
             switch (format)
@@ -215,6 +230,11 @@ namespace Ryujinx.Graphics.GAL
             return false;
         }
 
+        /// <summary>
+        /// Checks if the texture format is a depth, stencil or depth-stencil format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the format is a depth, stencil or depth-stencil format, false otherwise</returns>
         public static bool IsDepthOrStencil(this Format format)
         {
             switch (format)
@@ -231,6 +251,11 @@ namespace Ryujinx.Graphics.GAL
             return false;
         }
 
+        /// <summary>
+        /// Checks if the texture format only has one component.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format only has one component, false otherwise</returns>
         public static bool HasOneComponent(this Format format)
         {
             switch (format)
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -364,7 +364,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
             changedScale |= TextureManager.SetRenderTargetDepthStencil(depthStencil);
 
-            if (changedScale || changedScale)
+            if (changedScale)
             {
                 TextureManager.UpdateRenderTargetScale(singleUse);
                 _context.Renderer.Pipeline.SetRenderTargetScale(TextureManager.RenderTargetScale);
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -324,7 +324,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
             int samplesInX = msaaMode.SamplesInX();
             int samplesInY = msaaMode.SamplesInY();
 
-            bool changedColorScale = false;
+            bool changedScale = false;
 
             for (int index = 0; index < Constants.TotalRenderTargets; index++)
             {
@@ -334,14 +334,14 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
                 if (index >= count || !IsRtEnabled(colorState))
                 {
-                    changedColorScale |= TextureManager.SetRenderTargetColor(index, null);
+                    changedScale |= TextureManager.SetRenderTargetColor(index, null);
 
                     continue;
                 }
 
                 Texture color = TextureManager.FindOrCreateTexture(colorState, samplesInX, samplesInY);
 
-                changedColorScale |= TextureManager.SetRenderTargetColor(index, color);
+                changedScale |= TextureManager.SetRenderTargetColor(index, color);
 
                 if (color != null)
                 {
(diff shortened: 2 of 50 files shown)


Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -216,13 +218,20 @@ private void CommitTextureBindings(TexturePool pool, ShaderStage stage, int stag
 
                 Texture texture = pool.Get(textureId);
 
+                if ((binding.Flags & TextureUsageFlags.ResScaleUnsupported) > 0)
```
Thog, line 221: should be != 0 or use HasFlag

Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -269,6 +283,11 @@ private void CommitImageBindings(TexturePool pool, ShaderStage stage, int stageI
...
```
Thog, line 286: should be != 0 or use HasFlag

Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
```diff
@@ -33,6 +33,11 @@ public TextureCopy(Renderer renderer)
 
             ClearBufferMask mask = GetMask(src.Format);
 
+            if ((mask & (ClearBufferMask.DepthBufferBit | ClearBufferMask.StencilBufferBit)) > 0)
```
Thog, line 36: should be != 0

riperiperi:
> 
> 
> lgtm apart from what I listed. Also could you edit the Config.json for debug builds to reflect the change of version number?

Ah, I forgot there was also a version number in there.

riperiperi:
Addressed. I also included a link to a comparison tool for one of the screenshots.

Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -359,7 +362,16 @@ private void UpdateRenderTargetState(GpuState state, bool useControl)
                 depthStencil = TextureManager.FindOrCreateTexture(dsState, dsSize, samplesInX, samplesInY);
             }
 
-            TextureManager.SetRenderTargetDepthStencil(depthStencil);
+            changedScale |= TextureManager.SetRenderTargetDepthStencil(depthStencil);
+
+            if (changedScale || changedScale)
```
gdkchan, line 367:
```suggestion
            if (changedScale)
```

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -282,14 +311,157 @@ private void RecreateStorageOrView(int width, int height, int depthOrLayers)
...
+
+                Logger.PrintDebug(LogClass.Gpu, $"  Copy performed: {HostTexture.Width}x{HostTexture.Height} to {newStorage.Width}x{newStorage.Height}");
+
+                ReplaceStorage(newStorage);
+
+                // All views must be recreated against the new storage.
+
+                foreach (var view in _views)
+                {
+                    Logger.PrintDebug(LogClass.Gpu, $"  Recreating view ${Info.Width}x${Info.Height} ${Info.FormatInfo.Format.ToString()}.");
```
gdkchan, line 441: Why it has the extra `$`s inside the string?

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -169,18 +174,112 @@ public void SetGraphicsTexturePool(ulong gpuVa, int maximumId)
...
+        /// <param name="singleUse">If this is not -1, it indicates that only the given indexed target will be used.</param> 
+        public void UpdateRenderTargetScale(int singleUse)
+        {
+            // Make sure all scales for render targets are at the highest they should be. Blacklisted targets should propagate their scale to the other targets.
+            bool mismatch = false;
+            bool blacklisted = false;
+            bool hasUpscaled = false;
+            float targetScale = GraphicsConfig.ResScale;
+
+            void considerTarget(Texture target)
```
gdkchan, line 201: Do we use `camelCase` for local functions?

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -262,12 +361,50 @@ private void UpdateRenderTargets()
...
+        /// Note that this is different from being compatible - this elilinates targets that would have detrimental effects when scaled.
+        /// </summary>
+        /// <param name="info">The texture info to check</param>
+        /// <returns>True if safe</returns>
+        public bool UpscaleSafeMode(TextureInfo info)
+        {
+            // While upscaling works for all targets defined by IsUpscaleCompatible, we additionally blacklist targets here that
+            // may have undesirable results (upscaling blur textures) or simply waste GPU resources (upscaling texture atlas).
+
+            if (info.Width / 8 == info.Height / 8 && !(info.FormatInfo.Format.IsDepthOrStencil() || info.FormatInfo.Format.HasOneComponent()))
```
gdkchan, line 385: Out of curiosity, why the division by 8 here?

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -572,6 +724,17 @@ public Texture FindOrCreateTexture(TextureInfo info, TextureSearchFlags flags =
 
                         TextureCreateInfo createInfo = GetCreateInfo(overlapInfo, _context.Capabilities);
 
+                        if (texture.ScaleFactor != overlap.ScaleFactor)
+                        {
+                            // A bit tricky, our new texture may need to contain an existing texture that is upscaled, but isn't itself. 
+                            // In that case, we prefer the higher scale only if our format is render-target-like, otherwise we scale the view down before copy.
+                            texture.PropagateScale(overlap);
+
+                            //float preferredScale = IsUpscaleCompatible(info) ? Math.Max(texture.ScaleFactor, overlap.ScaleFactor) : 1f;
```
gdkchan, line 733: Whats up with this commented out code here?

Ryujinx.Graphics.Gpu/Image/TextureSearchFlags.cs
```diff
@@ -11,6 +11,7 @@ enum TextureSearchFlags
         None     = 0,
         IgnoreMs = 1 << 0,
         Strict   = 1 << 1 | Sampler,
-        Sampler  = 1 << 2
+        Sampler  = 1 << 2,
+        WithUpscale = 1 << 4
```
gdkchan, line 15: nit: Alignment, also could be `1 << 3`, the shift makes it a power of 2.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -54,6 +59,11 @@ internal Pipeline()
             {
                 _componentMasks[index] = 0xf;
             }
+
+            for (int index = 0; index < _fpRenderScale.Length; index++)
+            {
+                _fpRenderScale[index] = 1f;
```
gdkchan, line 65: Shouldn't it set a default for `_cpRenderScale` aswell?

Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
```diff
@@ -219,6 +227,33 @@ private static void DeclareUniforms(CodeGenContext context, StructuredProgramInf
...
+        {
+            if ((context.Config.UsedFeatures & (FeatureFlags.FragCoordXY | FeatureFlags.IntegerSampling)) != 0)
+            {
+                string stage = OperandManager.GetShaderStagePrefix(context.Config.Stage);
+
+                int scaleElements = context.TextureDescriptors.Count;
+                
+                if (context.Config.Stage == ShaderStage.Fragment)
+                {
+                    scaleElements += 1; // Also includes render target scale, for gl_FragCoord.
```
gdkchan, line 240:
```suggestion
                    scaleElements++; // Also includes render target scale, for gl_FragCoord.
```

Ryujinx.Graphics.Shader/Ryujinx.Graphics.Shader.csproj
```diff
@@ -1,5 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <ItemGroup>
```
gdkchan, line 3: What's the extra ItemGroup used for? It's not there for the other embedded glsl files.

Ryujinx.Graphics.Shader/Translation/FeatureFlags.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Ryujinx.Graphics.Shader.Translation
+{
+    /// <summary>
+    /// Features used by the shader that are important for the code generator to know in advance.
+    /// These typically change the declarations in the shader header.
+    /// </summary>
+    [Flags]
+    public enum FeatureFlags
+    {
+        None            = 0,
```
gdkchan, line 12: nit: I think it's not necessary to align this to the other entries.

Ryujinx.Graphics.GAL/Format.cs
```diff
@@ -214,5 +214,49 @@ public static bool IsAstcSrgb(this Format format)
 
             return false;
         }
+
+        public static bool IsDepthOrStencil(this Format format)
```
gdkchan, line 218: This one can replace the `IsDepthStencil` function inside `TexturePool.cs`.

Ryujinx.Graphics.Gpu/Image/Texture.cs
riperiperi: I program in typescript too much. I never saw this message much so didn't even notice they were there.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
riperiperi: This also catches targets that are square-_like_, and less than 8x8 pixels. The cubemap faces in MK8 hit this case at the smaller levels - they become like 1x4 or something despite the fact that they should actually be square. This is because they need to be aligned, but this is just a quick way of detecting it before a sampler changes their size.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
riperiperi: I thought it might be useful at some point, then it wasn't.

Ryujinx.Graphics.OpenGL/Pipeline.cs
riperiperi: Yep... 😩

Ryujinx.Graphics.Shader/Ryujinx.Graphics.Shader.csproj
riperiperi: It kind of did this by itself, I'm not sure. I will remove it.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
gdkchan: Ok, I geti t. NVN only aligns the width though, and the alignment depends on the format. It's 64 / BytesPerPixel for block linear textures, and 32 / BytesPerPixel for pitch linear textures.

Ryujinx.Graphics.Gpu/Image/TextureSearchFlags.cs
riperiperi: Wasn't really thinking with this one.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
riperiperi: You're right, this should be `ConsiderTarget`.

Ryujinx.Graphics.GAL/Format.cs
riperiperi: This function also includes S8Uint, which that one doesn't. Should the case in TexturePool also check for that?

gdkchan: Yes, that was an oversight.

jduncanator:
See our blog post here for more information and some mouth watering screenshots :fire:

[ryujinx-blog-link](https://blog.ryujinx.org/introducing-resolution-scaling/)

TibiaZ:
This is so good! Great job man! 👏🏻👏🏻

PR #1367, opened 2020-07-07
Move the "Audio" setting(s) to a new "Audio" tab

VolcaEM:
While we're at it, also move the "Audio" setting(s) to the "AudioSection" class to also reflect this change in the code too

I'm not used with working with GTK at all so, please, if you find any problems, let me know!

audio_settings


Xpl0itR:
I think it is unnecessary to make a tab for a single option, otherwise I would have done so

Thog:
Closed for inactivity (will be superseded some day)

PR #1369, opened 2020-07-08
Fix PPTC on Windows 7.

LDj3SNuD:
In certain circumstances, the page table base pointer was able to undergo the memory operand containment optimisation, which coalesced the page table pointer into the memory instructions immediate operand. This prevented PPTC from relocating the page table pointer when the cache was restored in a subsequent run. This blacklists the page table pointer from undergoing certain optimisations that could prevent the page table pointer from being relocated by PPTC.

### DIFF
--- a/ARMeilleure/IntermediateRepresentation/Operand.cs
+++ b/ARMeilleure/IntermediateRepresentation/Operand.cs
@@ -10,8 +10,10 @@ namespace ARMeilleure.IntermediateRepresentation
 
         public ulong Value { get; private set; }
 
-        public bool DisableCF { get; private set; }
-        public int? PtcIndex  { get; private set; }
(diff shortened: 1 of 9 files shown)


ARMeilleure/IntermediateRepresentation/Operand.cs
```diff
@@ -10,8 +10,10 @@ class Operand
 
         public ulong Value { get; private set; }
 
-        public bool DisableCF { get; private set; }
-        public int? PtcIndex  { get; private set; }
+        public bool DisableCF         { get; private set; }
```
gdkchan, line 13: I'd prefer to only have a single property (like DisableCF) that is used everywhere, instead of 3. The name would have to be changed, maybe something like "Relocatable".

PR #1370, opened 2020-07-08
Initial transform feedback support

gdkchan:
This is just partial support, and still lacks the following:
- DrawTransformFeedback support.
- Pause/Resume support (basically, not ignoring TfBufferState offset, and implementing the query properly).
Games using either of those will most likely not work properly.

Also did some refactoring:
- Changed the write method on MemoryAccessor to take a generic parameter instead of just `int`.
- Fixed constraints on `GpuState`, it now uses `unmanaged` constraint instead of `struct`.

One major change was changing how varyings are declared on shaders. It now uses floats instead of vec4. This was done to allow specifying a single component of the varying to be written on the transform feedback buffer. It does not allow a single or a set of components to be selected from a vec4, so breaking it down into floats was necessary. Futhermore, the GLSL version was increased to 4.4 since it now makes use of ARB_enhanced_layout feature, to allow manually specifying the varying components.


Doctorwho1909:
https://media.discordapp.net/attachments/480497101676216340/730598582565994547/unknown.png almost works in XC2
https://media.discordapp.net/attachments/480497101676216340/730597642756685844/unknown.png metro 2033/last light almost fully works still missing characters

hellblade is now black when before it rendered but ue4 games work again

PR #1379, opened 2020-07-09
Improve kernel IPC related syscalls

gdkchan:
This adds missing logic to decrement the sessions count, when the session handle is closed on the client side. Also change increment implementation to be lock-less. Thanks to SciresM and Thog for pointing out the issue.

Also implement the following syscalls:
- SendSyncRequestWithUserBuffer
- SendAsyncRequestWithUserBuffer
- ReplyAndReceiveWithUserBuffer

This should not affect official games as they don't use those syscalls.

Special thanks to SciresM for the IPC tests aswell, that helped uncovering a few issues.


Ryujinx.Memory/MemoryBlock.cs
```diff
@@ -127,21 +127,21 @@ public void Write(ulong offset, ReadOnlySpan<byte> data)
...
         /// Copies data from one memory location to another.
         /// </summary>
-        /// <param name="srcOffset">Source offset to read the data from</param>
         /// <param name="dstOffset">Destination offset to write the data into</param>
+        /// <param name="srcOffset">Source offset to read the data from</param>
         /// <param name="size">Size of the copy in bytes</param>
         /// <exception cref="ObjectDisposedException">Throw when the memory block has already been disposed</exception>
         /// <exception cref="ArgumentOutOfRangeException">Throw when <paramref name="srcOffset"/>, <paramref name="dstOffset"/> or <paramref name="size"/> is out of range</exception>
-        public void Copy(ulong srcOffset, ulong dstOffset, ulong size)
+        public void Copy(ulong dstOffset, ulong srcOffset, ulong size)
```
gdkchan, line 135: Fixes a regression, kernel code expects destination to be the first argument, not second. Since right now only kernel calls this, the change is safe (and makes more sense as copy functions, like memcpy for example, has dst as first argument).

PR #1380, opened 2020-07-10
Implement Logical Operation registers and functionality

riperiperi:
Used by XC:DE and XC2 - fixes their interior lighting volumes. Probably fixes a few other things.

Before:
https://gyazo.com/6a5e393ceadb3e9d0b7ec484dbdcd2e4

After:
https://gyazo.com/81a81843cd549e38534ee13a4b219635

(funky paths unrelated)

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -252,6 +252,11 @@ namespace Ryujinx.Graphics.Gpu.Engine
                 UpdateBlendState(state);
             }
 
+            if (state.QueryModified(MethodOffset.LogicOpState))
+            {
+                UpdateLogicOpState(state);
+            }
+
             CommitBindings();
         }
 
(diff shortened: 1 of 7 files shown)


Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -875,6 +880,17 @@ private void UpdateBlendState(GpuState state)
             }
         }
 
+        /// <summary>
+        /// Updates host logical op state, based on guest state.
+        /// </summary>
+        /// <param name="state">Current GPU state</param>
+        public void UpdateLogicOpState(GpuState state)
+        {
+            LogicalOpState blendConstant = state.Get<LogicalOpState>(MethodOffset.LogicOpState);
```
gdkchan, line 889: Why is that called `blendConstant`?

Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -875,6 +880,17 @@ private void UpdateBlendState(GpuState state)
             }
         }
 
+        /// <summary>
+        /// Updates host logical op state, based on guest state.
```
gdkchan, line 884: I think it's better to fully spell "logical operation" on this summary since it is supposed to be more detailed.

Ryujinx.Graphics.Gpu/Engine/Methods.cs
riperiperi: Copied the state.Get<> from above then completely skipped reading the variable name at all.

PR #1383, opened 2020-07-11
Fix folding of ConvertI64ToI32 imm64

FICTURE7:
Fixes an issue with #1359 where by the operand types can get mismatched. For example:

This
```wasm
i32 %52 = ConvertI64ToI32 i64 0x0
```
would become this
```wasm
i32 %52 = Copy i64 0x0
```
Then on the second pass of the optimizer on the block, `ConstantFolding` transforms it to
```wasm
i32 %52 = Copy i32 0x0
```

This can cause asserts in the assembler to be fired (because of src/dst type mismatch) when the optimizer does not do another pass after the folding of `ConvertI64ToI32`. This can happen when a ConvertI64ToI32 got constant folded but no copy propagation or removal of unused nodes has occurred. So `modified` remains false and the pass to retype the copies never happens.

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
+++ b/ARMeilleure/CodeGen/Optimizations/ConstantFolding.cs
@@ -81,9 +81,7 @@ namespace ARMeilleure.CodeGen.Optimizations
                 case Instruction.ConvertI64ToI32:
                     if (type == OperandType.I32)
                     {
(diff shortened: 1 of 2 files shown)


FICTURE7:
Mehh, looking back at this, I think this patch could be cleaner. :< I'll clean it up tomorrow I guess.

riperiperi:
Nice one. I should test in debug mode more often so I can actually see these asserts. 🤪

PR #1400, opened 2020-07-19
New GPFifo and fast guest constant buffer updates

gdkchan:
- Implement GPFifo using structs/enums auto-generated from official NVIDIA definitions.
- Replace MethodInfo with GPFifoClass, also with struct also auto-generated. This is how I would like to change the other engines aswell, in the future.
- Improve GPFifo semaphore implementation, supporting all reduction operations. Acquire is still not supported.
- Add a fast path for guest constant buffer updates. Writes all data at once, instead of one word at a time (more information below).

The fast path for constant buffer updates brings a significant performance boost to games that makes heavy use of this functionality. Obviously, games that do not make use of it are not going to be affected.

One of the affected games is Yume Nikki Dream Diary.
Master:
https://streamable.com/0i8dp7
PR:
https://streamable.com/u93b8l

As this makes changes to GPFifo, it has the potential to break games, so it would be nice to test a few of them.


jduncanator:
Would be nice to get a header at the top of each file that is auto-generated, warning people against manually modifying it.

gdkchan:
Ready for review.

PR #1407, opened 2020-07-20
 Print guest stack trace on invalid memory access 

gdkchan:
- Prints the guest stack trace when a invalid memory access occurs on the guest, instead of just throwing "AccessViolationException".
- Prints the guest stack trace on attempts to execute a undefined instruction (useful for invalid instructions, not so much for unimplemented instructions).
- Change `ArgumentOutOfRangeException` to custom `InvalidMemoryRegionException` on `MemoryBlock`. The former is too generic and would be problematic when using try/catch to handle invalid access.

Should have no performance impact, on the codegen, one jump will change (from less to less than or equal), this forces the slow path to be taken, and the costly validity checks are done there.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
@@ -326,7 +326,7 @@ namespace ARMeilleure.Instructions
             }
             while (bit < context.Memory.AddressSpaceBits);
 
-            context.BranchIfTrue(lblSlowPath, context.ICompareLess(pte, Const(0L)));
+            context.BranchIfTrue(lblSlowPath, context.ICompareLessOrEqual(pte, Const(0L)));
 
             Operand pageOffset = context.BitwiseAnd(address, Const(address.Type, PageMask));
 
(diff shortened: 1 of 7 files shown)


ARMeilleure/Instructions/InstEmitMemoryHelper.cs
```diff
@@ -326,7 +326,7 @@ private static Operand EmitPtPointerLoad(ArmEmitterContext context, Operand addr
             }
             while (bit < context.Memory.AddressSpaceBits);
 
-            context.BranchIfTrue(lblSlowPath, context.ICompareLess(pte, Const(0L)));
+            context.BranchIfTrue(lblSlowPath, context.ICompareLessOrEqual(pte, Const(0L)));
```
jduncanator, line 329: Does this need a PPTC version bump?

gdkchan: It is required to print the stack trace on invalid accesses. But not bumping the version should do no harm, other than keeping the old behavior where it only shows AccessViolationException.

PR #1408, opened 2020-07-20
Texture/Buffer Memory Management Improvements

riperiperi:
image

This PR aims to make texture and buffer memory management better in a few areas, so that we can do things like flush textures without worrying about data corruption, and thus fix a few issues with our AutoDeleteCache having a limited size. The course pictured above generally has the ground incredibly dark, as the course shadow map and lighting cube for the floor are both dropped out of the AutoDeleteCache during loading. No such problem in the screenshot. 😌 

All of these changes are related, which is why they're all part of one PR. There is one change that needs to be here to make this more stable, but more on that later.

1. Overlapping textures that are not  are now removed from the AutoDeleteCache. This gives us a guarantee that no unrelated overlapping textures can flush out of order, when we get around to flushing them, and it also cleans out the texture cache so that it's not a huge mess of tangled overlaps. It can also remove some dead textures, which can help with memory usage.
2. When textures fall out of the AutoDeleteCache, they are now flushed rather than lost. The alternative would be an infinite size texture cache, but then we would need you to have a 32GB page file or something. So reworking things to better facilitate flushing wins.
3. Modification flag is unset when a texture's memory is unmapped. This will stop textures from flushing when their memory is definitely used for something else.
4. Buffer and Texture flushing now writes to memory _without_ triggering write tracking. This avoids textures from setting each other as dirty when flushing to memory - which allows us to flush textures in the auto delete cache in order of use without any repercussions. It has been extended to buffers as it seems to improve performance and remove vertex explosions.
5. An OpenGL resource cache now exists so that textures that are deleted and immediately reused can just use the old texture resource. **This mostly exists as a mitigation to some games aggressively reusing memory used for screen render targets.**
6. With the new overlap rules, we can detect whether or not a texture needs its initial data set or not, based on the presence of existing textures and if their memory has been modified. This can improve performance for res scaling, or games that continually create render targets and set the data of other, unrelated ones. (mostly improves res scaling) **It is needed as part of (5)'s performance mitigation.**
7. View compatibility has been refactored a bit so that it returns an enum that indicates if it's only copy compatible. This avoids calling it multiple times.

## Benefits
image

Here are the benefits, with helpful numbers pointing to which change was responsible:

- All random chance course issues (broken shadows, black actor lighting, black screen) in MK8 are fixed. (1, 2)
- In Super Mario Odyssey, you **no longer need to pause the game or fast travel to fix broken graphics in kingdoms.** (1, 2)
- Generally, games with textures are black or garbage after loading will be improved, but that of course does not include all situations (UE4 and catherine don't count) (1, 2)
- **Vertex explosions have been reduced in Unity games**, and their performance has slightly improved. I am not sure if this has affected any other games. (4)
- Write tracking performance will improve with range-tracking, due to less dead overlapping textures within the same spot. In MK8, thousands of 32x32 textures are rendered to in a sequence to generate static lighting for actors. These take up the same space in memory as the _nvdec video_, causing a major performance hit when we return to the menu. With this change, these overlaps are removed before we start tracking the nvdec video, so it can continue being a nice single memory region rather than thousands of tiny ones. (1)

## Downsides
Some existing issues may be eccentuated:

- Luigi's mansion's dynamic resolution will glow with garbage colours more often, due to a ChangeSize bug. (5)
- Some corruption may happen when flushing a texture, though it is infrequent. (see below)

## Overlap + Flushing Rules

(writeup in progress)

## Random corruptions due to flushing?

Yes, I'm PRing something that has a chance to do this. Play 32 courses in Mario Kart and you will notice something that's off in at least one course, or it may even crash. Here's the reason why that is:

The rules for flushing indicate that we should only flush a texture if its memory has not been modified since the last time it was used (and if the memory was unmapped, which is working fine). What happens if two textures share the same page?

Under the _current method_ of tracking texture modification, two or more textures that share a page can face a dilemma. If one texture on the page consumes the dirty flag, it is consumed for _all other textures on the page_. How do we solve this one?

That's right - it's range tracking time again. Literally cannot get away from this. Essentially, it will work better when the memory write tracking is able to signal to each affected texture individually - so that if one's dirty bit is consumed, the other's is still present. The range tracking PR will be rebased on top of this one.

## Call for Testing
Even with the above issue, it would really benefit us to test this PR to check if there have been any regressions. Since every game uses textures, the surface area for potential regressions is huge. If the game is generally improved, good! If textures are corrupt, missing or the game crashes very quickly, then there's probably an issue.

This may change performance on some games, for better or worse. It will be good to know cases where it kills it.

This will be a part of the range tracking PR in future, which may be a better test target when it's up and running due to flushing being more predicatble. I will comment when that is ready.

### DIFF
--- a/Ryujinx.Graphics.GAL/ITexture.cs
+++ b/Ryujinx.Graphics.GAL/ITexture.cs
@@ -17,5 +17,6 @@ namespace Ryujinx.Graphics.GAL
 
         void SetData(ReadOnlySpan<byte> data);
         void SetStorage(BufferRange buffer);
+        void Release();
     }
 }
--- a/Ryujinx.Graphics.GAL/ITexture.cs
+++ b/Ryujinx.Graphics.GAL/ITexture.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace Ryujinx.Graphics.GAL
 {
-    public interface ITexture : IDisposable
+    public interface ITexture
     {
         int Width { get; }
         int Height { get; }
--- a/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
+++ b/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
@@ -119,7 +119,7 @@ namespace Ryujinx.Graphics.GAL
 
         public override int GetHashCode()
         {
-            return Width | (Height << 16);
+            return HashCode.Combine(Width, Height);
         }
 
         bool IEquatable<TextureCreateInfo>.Equals(TextureCreateInfo other)
--- a/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
+++ b/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
@@ -3,7 +3,7 @@ using System;
 
 namespace Ryujinx.Graphics.GAL
 {
-    public struct TextureCreateInfo
+    public struct TextureCreateInfo : IEquatable<TextureCreateInfo>
     {
         public int Width         { get; }
         public int Height        { get; }
@@ -115,6 +115,30 @@ namespace Ryujinx.Graphics.GAL
         private static int GetLevelSize(int size, int level)
         {
             return Math.Max(1, size >> level);
+        }
+
+        public override int GetHashCode()
+        {
+            return Width | (Height << 16);
+        }
+
+        bool IEquatable<TextureCreateInfo>.Equals(TextureCreateInfo other)
+        {
+            return Width == other.Width &&
+                   Height == other.Height &&
+                   Depth == other.Depth &&
+                   Levels == other.Levels &&
+                   Samples == other.Samples &&
+                   BlockWidth == other.BlockWidth &&
+                   BlockHeight == other.BlockHeight &&
+                   BytesPerPixel == other.BytesPerPixel &&
+                   Format == other.Format &&
+                   DepthStencilMode == other.DepthStencilMode &&
+                   Target == other.Target &&
+                   SwizzleR == other.SwizzleR &&
+                   SwizzleG == other.SwizzleG &&
+                   SwizzleB == other.SwizzleB &&
+                   SwizzleA == other.SwizzleA;
         }
     }
 }
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -355,7 +355,6 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         public void BlacklistScale()
         {
-            if (Info.Width == 1920) { }
             ScaleMode = TextureScaleMode.Blacklisted;
             SetScale(1f);
         }
@@ -542,7 +541,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             // Note that if ASTC is not supported by the GPU we can't read it back since
             // it will use a different format. Since applications shouldn't be writing
             // ASTC textures from the GPU anyway, ignoring it should be safe.
-            if (_context.Methods.TextureManager.IsTextureModified(this) && !Info.FormatInfo.Format.IsAstc())
+            if (IsModified && !Info.FormatInfo.Format.IsAstc())
             {
                 Span<byte> gpuData = GetTextureDataFromGpu(true);
 
@@ -655,6 +654,16 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             IsModified = false;
 
+            if (Info.FormatInfo.Format.IsAstc())
+            {
+                return; // Flushing this format is not supported, as it may have been converted to another host format.
+            }
+
+            if (Info.FormatInfo.IsCompressed)
+            {
+                Logger.PrintError(LogClass.Gpu, "Flushing compressed texture!");
+            }
+
             _context.PhysicalMemory.WriteUntracked(Address, GetTextureDataFromGpu(blacklist));
         }
 
@@ -675,12 +684,16 @@ namespace Ryujinx.Graphics.Gpu.Image
                 BlacklistScale();
                 data = HostTexture.GetData();
             } 
-            else
+            else if (ScaleFactor != 1f)
             {
                 float scale = ScaleFactor;
                 SetScale(1f);
                 data = HostTexture.GetData();
                 SetScale(scale);
+            }
+            else
+            {
+                data = HostTexture.GetData();
             }
 
             if (Info.IsLinear)
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -172,7 +172,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// Initializes the data for a texture. Can optionally initialize the texture with or without data.
         /// If the texture is a view, it will initialize memory tracking to be non-dirty.
         /// </summary>
-        /// <param name="isView">True if the texture is a view or not</param>
+        /// <param name="isView">True if the texture is a view, false otherwise</param>
         /// <param name="withData">True if the texture is to be initialized with data</param>
         public void InitializeData(bool isView, bool withData = false)
         {
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -278,6 +278,8 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             width  <<= _firstLevel;
             height <<= _firstLevel;
+            int blockWidth = Info.FormatInfo.BlockWidth;
+            int blockHeight = Info.FormatInfo.BlockHeight;
 
             if (Info.Target == Target.Texture3D)
             {
@@ -288,7 +290,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 depthOrLayers = _viewStorage.Info.DepthOrLayers;
             }
 
-            _viewStorage.RecreateStorageOrView(width, height, depthOrLayers);
+            _viewStorage.RecreateStorageOrView(width, height, blockWidth, blockHeight, depthOrLayers);
 
             foreach (Texture view in _viewStorage._views)
             {
@@ -306,8 +308,26 @@ namespace Ryujinx.Graphics.Gpu.Image
                     viewDepthOrLayers = view.Info.DepthOrLayers;
                 }
 
-                view.RecreateStorageOrView(viewWidth, viewHeight, viewDepthOrLayers);
+                view.RecreateStorageOrView(viewWidth, viewHeight, blockWidth, blockHeight, viewDepthOrLayers);
             }
+        }
+
+        /// <summary>
+        /// Recreates the texture storage (or view, in the case of child textures) of this texture.
+        /// This allows recreating the texture with a new size.
+        /// A copy is automatically performed from the old to the new texture.
+        /// </summary>
+        /// <param name="width">The new texture width</param>
+        /// <param name="height">The new texture height</param>
+        /// <param name="width">The block width related to the given width</param>
+        /// <param name="height">The block height related to the given height</param>
+        /// <param name="depthOrLayers">The new texture depth (for 3D textures) or layers (for layered textures)</param>
+        private void RecreateStorageOrView(int width, int height, int blockWidth, int blockHeight, int depthOrLayers)
+        {
+            RecreateStorageOrView(
+                BitUtils.DivRoundUp(width * Info.FormatInfo.BlockWidth, blockWidth),
+                BitUtils.DivRoundUp(height * Info.FormatInfo.BlockHeight, blockHeight), 
+                depthOrLayers);
         }
 
         /// <summary>
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -44,6 +44,8 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         public bool IsModified { get; internal set; }
 
+        private bool _everModified;
+
         private int _depth;
         private int _layers;
         private int _firstLayer;
@@ -535,13 +537,13 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             ReadOnlySpan<byte> data = _context.PhysicalMemory.GetSpan(Address, (int)Size);
 
-            // If the texture was modified by the host GPU, we do partial invalidation
+            // If the texture was ever modified by the host GPU, we do partial invalidation
             // of the texture by getting GPU data and merging in the pages of memory
             // that were modified.
             // Note that if ASTC is not supported by the GPU we can't read it back since
             // it will use a different format. Since applications shouldn't be writing
             // ASTC textures from the GPU anyway, ignoring it should be safe.
-            if (IsModified && !Info.FormatInfo.Format.IsAstc())
+            if (_everModified && !Info.FormatInfo.Format.IsAstc())
             {
                 Span<byte> gpuData = GetTextureDataFromGpu(true);
 
@@ -661,7 +663,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (Info.FormatInfo.IsCompressed)
             {
-                Logger.PrintError(LogClass.Gpu, "Flushing compressed texture!");
+                Logger.Error?.PrintMsg(LogClass.Gpu, "Flushing compressed texture!");
             }
 
             _context.PhysicalMemory.WriteUntracked(Address, GetTextureDataFromGpu(blacklist));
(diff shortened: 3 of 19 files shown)


Ryujinx.Graphics.OpenGL/ResourcePool.cs
```diff
@@ -0,0 +1,113 @@
...
+        public TextureCreateInfo Info;
+        public TextureView View;
+        public float ScaleFactor;
+        public int RemainingFrames;
+    }
+
+    /// <summary>
+    /// A structure for pooling resources that can be reused without recreation, such as textures.
+    /// </summary>
+    class ResourcePool
```
gdkchan, line 19: I think this should be disposable, to allow all textures in to pool to be deleted.

Ryujinx.Graphics.OpenGL/ResourcePool.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Attempt to obtain a texture from the resource cache with the desired parameters.
+        /// </summary>
+        /// <param name="info">The creation info for the desired texture</param>
+        /// <param name="scaleFactor">The scale factor for the desired texture</param>
+        /// <returns>A TextureView with the description specified, or null if one was not found.</returns>
+        public TextureView TryGetTexture(TextureCreateInfo info, float scaleFactor)
```
gdkchan, line 65: `Try*` methods usually returns a bool indicating if the operation was successful, and the actual value as a out parameter. I'd just call this `GetTexture`. Or maybe `GetTextureOrNull` if you really want to be explicit about the fact that it can return null.

Ryujinx.Graphics.OpenGL/ResourcePool.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// <summary>
+    /// A structure for pooling resources that can be reused without recreation, such as textures.
+    /// </summary>
+    class ResourcePool
+    {
+        private const int DisposedLiveFrames = 2;
+
+        private object _lock = new object();
```
gdkchan, line 23: nit: `readonly` (on both fields).

Ryujinx.Graphics.OpenGL/ResourcePool.cs
```diff
@@ -0,0 +1,113 @@
...
+        private Dictionary<uint, List<DisposedTexture>> _textures = new Dictionary<uint, List<DisposedTexture>>();
+
+        private uint GetTextureKey(TextureCreateInfo info)
```
gdkchan, line 26: Why not just override `GetHashCode` and `Equals` on `TextureCreateInfo`, and use it as key.

Ryujinx.Graphics.OpenGL/Renderer.cs
```diff
@@ -32,6 +34,7 @@ public Renderer()
             _counters = new Counters();
             _window = new Window(this);
             TextureCopy = new TextureCopy(this);
+            ResourcePool = new ResourcePool();
```
gdkchan, line 37: I think this should be disposed when the renderer is disposed.

Ryujinx.Graphics.GAL/ITexture.cs
```diff
@@ -17,5 +17,6 @@ public interface ITexture : IDisposable
 
         void SetData(ReadOnlySpan<byte> data);
         void SetStorage(BufferRange buffer);
+        void Release();
```
gdkchan, line 20: Is this supposed to be the new way to dispose of textures? If so, I think this interface should stop implementing IDisposable.

Ryujinx.Graphics.OpenGL/Renderer.cs
```diff
@@ -91,6 +101,11 @@ public void UpdateCounters()
             _counters.Update();
         }
 
+        public void PreFrame()
```
gdkchan, line 104: Would be nice to not need this function, maybe it's possible to "tick" the pool on window present? I still need to take a look at the code to figure out what this is used for.

Doctorwho1909:
image
turns out my broken fog was a Driver issue updating fixed the problem I have found no regressions or new issues with this PR great work as always it even caused a big speed boost in some titles

Ryujinx.Graphics.OpenGL/Renderer.cs
riperiperi: I deliberately added this PreFrame function to avoid blocking on anything that would have to wait on GPU commands in the future. It's future sight more than anything (for instance, glSync objects from the previous frame could be disposed here, if we were to generate them throughout the previous frame)

riperiperi:
Will need some investigation into performance implications for the flushes. While it's definitely more accurate to not trash these textures, and the point of overlap is definitely the best way to dispose them, doing so mid-frame could cause a stutter. Some games also aggressively reuse a memory region for a target many times in a frame, such as animal crossing.

Ryujinx.Graphics.OpenGL/Renderer.cs
gdkchan: Wouldn't calling that from the window Present function have the same effect, without needing to introduce a new function that needs to be explicitly called before each frame?

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -1236,6 +1327,29 @@ public bool OverlapsWith(ulong address, ulong size)
...
+        public bool HasViewCompatibleChild(Texture texture)
+        {
+            if (_viewStorage != this || _views.Count == 0)
+            {
+                return false;
+            }
+
+            foreach (Texture view in _views)
+            {
+                if (texture.IsViewCompatible(view.Info, view.Size, out int _, out int _) != TextureViewCompatibility.Incompatible)
```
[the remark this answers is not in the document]
riperiperi: There will be no CopyOnly views in `_views`, but if there were then I'd probably want to count them. (for the case this function is currently used for, at least)

Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -151,7 +151,7 @@ public void Flush(ulong address, ulong size)
 
             byte[] data = _context.Renderer.GetBufferData(Handle, offset, (int)size);
 
-            _context.PhysicalMemory.Write(address, data);
+            _context.PhysicalMemory.WriteUntracked(address, data);
```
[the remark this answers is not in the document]
riperiperi: Noted. I'll add a note in a comment.

Ryujinx.Graphics.GAL/ITexture.cs
riperiperi: Yeah, I guess dispose should only be used internally now. Release indicates that the backing resource isn't being finalized yet.

EmulationFanatic:
This noticeably raised the average FPS of and improved (but did not eliminate) vertex explosions in the game "Good Job!"

ReaperOfSouls1909:
Is this still wip ? Or is more testing needed ?

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -855,35 +927,38 @@ private bool TargetAndSamplesCompatible(TextureInfo info)
...
-        /// <returns>True if a view with the given parameters can be created from this texture, false otherwise</returns>
-        public bool IsViewCompatible(
-            TextureInfo info,
-            ulong       size,
-            out int     firstLayer,
-            out int     firstLevel)
+        /// <param name="first">The first compatibility level</param>
+        /// <param name="second">The second compatibility level</param>
+        /// <returns>The minimum compatibility level of two provided view compatibility results</returns>
+        private TextureViewCompatibility PropagateViewCompatability(TextureViewCompatibility first, TextureViewCompatibility second)
```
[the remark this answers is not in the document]
riperiperi: This function has been moved to that class with the rebase.

Ryujinx.Graphics.OpenGL/ResourcePool.cs
riperiperi: Do I need to override equals here? I think the struct equality would do the job - though it would change from the current behaviour of grouping lists by width and height, to creating dictionary buckets but grouping lists by exact TextureCreateInfo.

riperiperi: Though it would mean I could remove the .Equals check in `GetTextureOrNull`.

gdkchan: AFAIK the default struct equality is not very fast, that's why overriding it is recommended. If the current approach has some advantage, then it's fine for me, but I think doing what I suggested would simplify the code without much of a performance penalty.

jduncanator: For better performance, you need to implement IEquatable<T> (and so override Equals), otherwise the object will be boxed/unboxed to do an object comparison. It's not strictly needed for using it as a key, but it will be better overall.

riperiperi: OK, I did that. The equals override is a bit long but I guess there's no way around that 😅

Ryujinx.Graphics.OpenGL/Renderer.cs
riperiperi: (opengl) Present is techincally always during a frame rather than before. Buffers are only swapped _after_ it completes the present, and the present happens after all fifo calls are processed. If we're doing things here that need to wait for gpu sync, then it will delay swapbuffers and the frame time could be slightly higher. If we do it at the start of the next frame, there's less chance sync is required.

Haven't tested any of this, just a theory. I'm not sure if swap buffers will end up forcing a sync anyways.

Ryujinx.Graphics.GAL/TextureCreateInfo.cs
```diff
@@ -116,5 +116,29 @@ private static int GetLevelSize(int size, int level)
         {
             return Math.Max(1, size >> level);
         }
+
+        public override int GetHashCode()
+        {
+            return Width | (Height << 16);
```
jduncanator, line 122: This can use `HashCode.Combine`

riperiperi:
I've committed an additional fix when using SizeChange on compressed+non-compressed view combinations.

Before, this could set either the compressed or non compressed texture with a size with the wrong size, depending on which texture had its size changed. This caused exceptions when flushing the texture. 

Now it correctly takes the block size into account, assuming that these textures are only related because a pixel in the non-compressed texture represents a block in the compressed one.

riperiperi:
Disregard previous comment about performance issues, there doesn't seem to be much performance impact despite the tendancy for flushing textures midgame. (tested smo, mario kart, acnh). In AC:NH specifically, I was hovering around 25-26 fps in a rather busy area, and with or without this PR it was really margin of error as it fluctuated within that 1 frame difference. VRAM usage may be slightly higher but nothing appears to be leaking.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -147,28 +147,51 @@ public static bool LayoutMatches(TextureInfo lhs, TextureInfo rhs)
         }
 
         /// <summary>
-        /// Checks if the view sizes of a two given texture informations match.
+        /// Obtain the minimum compatibility level of two provided view compatibility results.
+        /// </summary>
+        /// <param name="first">The first compatibility level</param>
+        /// <param name="second">The second compatibility level</param>
+        /// <returns>The minimum compatibility level of two provided view compatibility results</returns>
+        public static TextureViewCompatibility PropagateViewCompatability(TextureViewCompatibility first, TextureViewCompatibility second)
```
gdkchan, line 155: typo: `Compatability`.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -14,6 +14,20 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class TextureManager : IDisposable
     {
+        private struct OverlapInfo
+        {
+            public TextureViewCompatibility Compatibility;
```
gdkchan, line 19: I would make those a get-only auto-property if theres no need to modify those fields.

gdkchan (approved):
lgtm, nice to see the weird texture bugs and vertex explosions finally fixed (to some extent).

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -240,6 +278,8 @@ public void ChangeSize(int width, int height, int depthOrLayers)
         {
             width  <<= _firstLevel;
             height <<= _firstLevel;
+            int blockWidth = Info.FormatInfo.BlockWidth;
+            int blockHeight = Info.FormatInfo.BlockHeight;
```
jduncanator, line 282: Nit: These two local declarations could probably be hoisted to the top of the function.

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -956,6 +1077,8 @@ public void DecrementReferenceCount()
             Debug.Assert(newRefCount >= 0);
 
             DeleteIfNotUsed();
+
+            return newRefCount == 0;
```
jduncanator, line 1081: You may want to consider `newRefCount <= 1`. I know technically this is an illegal state, however it would help prevent bugs in functions that depend on the return value of this function being true when all the ref's have been released. The debug assert will make the issue obvious to developers.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -147,28 +147,51 @@ public static bool LayoutMatches(TextureInfo lhs, TextureInfo rhs)
...
+        /// <param name="first">The first compatibility level</param>
+        /// <param name="second">The second compatibility level</param>
+        /// <returns>The minimum compatibility level of two provided view compatibility results</returns>
+        public static TextureViewCompatibility PropagateViewCompatibility(TextureViewCompatibility first, TextureViewCompatibility second)
+        {
+            if (first == TextureViewCompatibility.Incompatible || second == TextureViewCompatibility.Incompatible)
+            {
+                return TextureViewCompatibility.Incompatible;
+            }
+            else if (first == TextureViewCompatibility.CopyOnly || second == TextureViewCompatibility.CopyOnly)
```
jduncanator, line 161: Assuming this function should "Obtain the minimum compatibility level", should this not check that both textures are CopyOnly compatible, rather than either of them?

```suggestion
            else if (first == TextureViewCompatibility.CopyOnly && second == TextureViewCompatibility.CopyOnly)
```

Ryujinx.Graphics.OpenGL/ResourcePool.cs
```diff
@@ -0,0 +1,122 @@
...
+                if (!_textures.TryGetValue(info, out list))
+                {
+                    return null;
+                }
+
+                foreach (DisposedTexture texture in list)
+                {
+                    if (scaleFactor == texture.ScaleFactor)
+                    {
+                        list.Remove(texture);
```
jduncanator, line 72: Are you sure this is valid? Removing values from a collection whilst they're being iterated over via an IEnumerable is usually illegal.

jduncanator (changes requested):
Nicely done! Appreciate the well documented nature of the code, and everything looks really tidy :) Only a few nits and questions from me here, excuse my ignorance if some of them seem silly.

Ryujinx.Graphics.OpenGL/ResourcePool.cs
riperiperi: We return immediately afterwards. If the loop continued, it would throw.

jduncanator: Ah yep, missed that. :+1:

Ryujinx.Graphics.Gpu/Image/Texture.cs
gdkchan: That would be `<= 0` not `<= 1`.

jduncanator: Yes, that's what I meant, my bad. (I was thinking of -1 in my head)

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
riperiperi: If one texture is CopyOnly and the other is Full, then when considering them both the minimum level of compatibility is copy. It's the same logic as the incompatible check above - it chooses the most restrictive compatibility of the two.

jduncanator: :+1:

Ryujinx.Graphics.Gpu/Image/Texture.cs
riperiperi: It hasn't been triggered yet, but I can do this for safety.

gdkchan: I don't see how `<= 0` is better to be honest. If it's using that for disposal, then it would result in double disposal if there's a bug allowing it to be decremented below 0.

jduncanator: I'm happy with it either way, just a suggestion.

riperiperi: Actually, this returned result isn't used anywhere right now. It is returned from AutoDeleteCache.Remove, but that return value is not used by anything right now (it was at one point)

riperiperi: It was essentially used to check if an AutoDeleteCache remove resulted in the texture being deleted. I guess in that case, <= 0 would make more sense.

jduncanator (approved):

Thog (approved):

PR #1410, opened 2020-07-21
Refactor NativeContext

gdkchan:
Change NativeContext to use a struct cast internally, instead of manually calculating sizes/offset and using Marshal Read and Write methods. This makes extending it on the future easier, and bugs less likely to happen. It also fixes 2 bugs on the previous implementation:
- Calculated NativeContext size was 4 bytes lower than it should, since it was counting `CallAddress` as 4 bytes long, not 8. Probably wasn't an issue since allocations are page aligned (?).
- The offset calculates for `FpFlags` was wrong. It was storing both `Flags` and `FpFlags` at the same offset, which means that their flags could overwrite each other (I think that only affects the Qc flag).

This was part of some older changes that I had here that became stale, figured it was worth to PR this separately.

### DIFF
--- a/ARMeilleure/State/NativeContext.cs
+++ b/ARMeilleure/State/NativeContext.cs
@@ -10,7 +10,7 @@ namespace ARMeilleure.State
         private unsafe struct NativeCtxStorage
         {
             public fixed ulong X[RegisterConsts.IntRegsCount];
-            public fixed ulong V[RegisterConsts.VecRegsCount];
+            public fixed ulong V[RegisterConsts.VecRegsCount * 2];
             public fixed uint Flags[RegisterConsts.FlagsCount];
             public fixed uint FpFlags[RegisterConsts.FpFlagsCount];
             public int Counter;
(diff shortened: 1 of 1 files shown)


ARMeilleure/State/NativeContext.cs
```diff
@@ -1,201 +1,159 @@
...
-            Marshal.WriteInt32(BasePtr, GetCounterOffset(), value);
-        }
+        public int GetCounter() => GetStorage().Counter;
+        public void SetCounter(int value) => GetStorage().Counter = value;
 
-        public static int GetRegisterOffset(Register reg)
+        public unsafe static int GetRegisterOffset(Register reg)
         {
-            int offset, size;
+            if ((uint)reg.Index >= 32)
```
LDj3SNuD, line 117: Maybe that wants to be a constant somewhere.

gdkchan: I added a check using the constant register count for each register type. I think using a single constant of 32 wouldn't make sense when it already has separate counts per register type.

PR #1413, opened 2020-07-21
 Implement inline memory load/store exclusive and ordered

gdkchan:
This improves the speed or exclusive or ordered memory accessed, by not calling a C# helper function. Instead, it now uses a inline implementation.

According to the Arm manual, both exclusive and ordered accesses must be size aligned, otherwise a fault is generated. Those changes take advantage of that, since the address is supposed to be aligned, we don't need to worry about accesses crossing a page boundary, so we can assume the accessed region will be always continugous, and no slow path is required for this case.

For the other 2 cases that requires a slow path (address being write tracked or invalid address), it now does the required checks inline, and if needed calls `MarkRegionAsModified` (for write tracking) and `ThrowInvalidMemoryAccess` for invalid accesses. The latter just throws a exception.

It's worth noting that unaligned ordered accesses will now throw a exception. This is OK according to the manual, but is different from the previous implementation, that did allow unaligned ordered accesses (and in some cases even unaligned exclusive accesses).

Depends on #1410 as this needs to add more fields to the NativeContext (to store exclusive address and value for CAS).


ARMeilleure/Instructions/InstEmitMemoryExHelper.cs
```diff
@@ -1,87 +1,181 @@
...
-        {
-            MethodInfo info = null;
+        private const int ErgSizeLog2 = 4;
 
+        public static Operand EmitLoadExclusive(ArmEmitterContext context, Operand address, bool exclusive, int size)
+        {
             if (exclusive)
             {
-                switch (size)
+                Operand isUnalignedAddr = InstEmitMemoryHelper.EmitAddressCheck(context, address, size);
```
FICTURE7, line 19: I believe this alignment check can be moved into the `if (size == 4)` block, because `EmitReadIntAligned` already emits a check itself.

gdkchan: Done.

ARMeilleure/Instructions/InstEmitMemoryExHelper.cs
```diff
@@ -1,87 +1,181 @@
 ﻿using ARMeilleure.IntermediateRepresentation;
+using ARMeilleure.State;
 using ARMeilleure.Translation;
 using System.Reflection;
```
LDj3SNuD, line 4: Nit: `using System.Reflection;` seems unused.

gdkchan: Removed.

PR #1418, opened 2020-07-23
Implement BGRA texture support

gdkchan:
This adds proper support for BGRA8 textures. OpenGL does not support this format, so before data was just being written as RGBA. This actually emulates BGRA with the following method:
- When the BGRA texture is used as render target, the R and B components are swapped on the pixel shader. New uniforms were added to indicate if the given output N should have their components swapped. Note that this will consume one extra constant buffer, that can't be used for game constant buffers anymore (however this was already the case with resolution scaling enabled).
- When a BGRA texture is bound, the R and B components are swapped on the swizzle parameters.
- When a BGRA texture is copied to another non-BGRA texture (and vice-versa), it first swaps R and B doing a PBO copy to a temporary texture, and then performs the copy as normal, using the temporary texture as source.
- When a BGRA texture is presented to the screen, it does the same as above, but doing a blit to the backbuffer instead of another destination texture.

This fixes all isues I'm aware of games with inverted R and B components. Examples:
Before:
image
After:
image
Before:
image
After:
image
Before:
image
After:
image
Hatsune Miku Project DIVA MEGA39's also had a similar issue on displays, which is now fixed:
image

NOTE: Some of the before screenshots were taken from the gamedb, not by me, so credits goes for those that took said screenshots.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -31,7 +31,7 @@ namespace Ryujinx.Graphics.OpenGL
         private int _boundDrawFramebuffer;
         private int _boundReadFramebuffer;
 
-        private bool[] _fpIsBgra = new bool[8];
+        private int[] _fpIsBgra = new int[8];
         private float[] _fpRenderScale = new float[33];
         private float[] _cpRenderScale = new float[32];
 
@@ -817,7 +817,7 @@ namespace Ryujinx.Graphics.OpenGL
 
                 _framebuffer.AttachColor(index, color);
 
-                _fpIsBgra[index] = color != null ? color.Format.IsBgra8() : false;
+                _fpIsBgra[index] = color != null && color.Format.IsBgra8() ? 1 : 0;
             }
 
             UpdateFpIsBgra();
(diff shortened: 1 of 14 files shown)


Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1112,6 +1119,17 @@ private void EnsureFramebuffer()
             return (_boundDrawFramebuffer, _boundReadFramebuffer);
         }
 
+        private void UpdateFpIsBgra()
+        {
+            if (_program != null)
+            {
+                for (int index = 0; index < _fpIsBgra.Length; index++)
+                {
+                    GL.Uniform1(_program.FragmentIsBgraUniform[index], _fpIsBgra[index] ? 1 : 0);
```
riperiperi, line 1128: Is there a reason these can't be uploaded as a boolean array rather than indexed? I feel like 8 Uniform1 calls per render target change might be overkill.

Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
```diff
@@ -115,22 +115,46 @@ public void PrepareForReturn()
...
                         {
-                            Operand dest = Attribute(AttributeConsts.FragmentOutputColorBase + attachment * 16 + component * 4);
+                            continue;
+                        }
+
+                        int fragmentOutputColorAttr = AttributeConsts.FragmentOutputColorBase + rtIndex * 16;
+
+                        Operand src = Register(regIndex, RegisterType.Gpr);
+
+                        // Perform B <-> R swap if needed, for BGRA formats (not support on OpenGL).
```
riperiperi, line 133: `not supported`

Ryujinx.Graphics.OpenGL/Pipeline.cs
gdkchan: How does one send a boolean uniform array?

PR #1422, opened 2020-07-24
common: Fix WMI exception

### LINKED ISSUE #1353: Emulator fails to start with services related to system info are disabled
This behavior did not occur on an earlier version. The exact error is here
`Unhandled exception. System.TypeInitializationException: The type initializer for 'Ryujinx.Common.SystemInfo.SystemInfo' threw an exception.
 ---> System.Runtime.InteropServices.COMException (0x80070422): The service cannot be started, either because it is disabled or because it has no enabled devices associated with it.`

Issue is caused by #1102

AcK77:
We currently don't check if WMI service is available when we get the CPU name and the RAM size.
This fix the issue by catching all exceptions and set default values instead.

Close #1353

### DIFF
--- a/Ryujinx.Common/SystemInfo/WindowsSystemInfo.cs
+++ b/Ryujinx.Common/SystemInfo/WindowsSystemInfo.cs
@@ -1,3 +1,5 @@
+using Ryujinx.Common.Logging;
+using System;
 using System.Management;
 
 namespace Ryujinx.Common.SystemInfo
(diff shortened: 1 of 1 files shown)


Ryujinx.Common/SystemInfo/WindowsSystemInfo.cs
```diff
@@ -9,14 +11,24 @@ internal class WindowsSysteminfo : SystemInfo
...
+                    RamSize = ulong.Parse(mObject["TotalVisibleMemorySize"].ToString()) * 1024;
+                }
+            }
+            catch (Exception)
             {
-                RamSize = ulong.Parse(mObject["TotalVisibleMemorySize"].ToString()) * 1024;
+                Logger.PrintError(LogClass.Application, "WMI isn't available, system informations will use default values.");
+
+                CpuName = "Unknown";
+                RamSize = 0;
```
Xpl0itR, line 31: Default values are already set here: https://github.com/Ryujinx/Ryujinx/blob/dad59c5ca05beef5daae55a2d98018a3ec37788b/Ryujinx.Common/SystemInfo/SystemInfo.cs#L8-L9

AcK77: But since they are override at the beginning of the class values are reset to `""` and `0`.
I've removed `RamSize` assign since it's pointless.

Ryujinx.Common/SystemInfo/WindowsSystemInfo.cs
```diff
@@ -9,14 +11,23 @@ internal class WindowsSysteminfo : SystemInfo
...
+                    CpuName = mObject["Name"].ToString();
+                }
 
-            foreach (ManagementBaseObject mObject in new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_OperatingSystem").Get())
+                foreach (ManagementBaseObject mObject in new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_OperatingSystem").Get())
+                {
+                    RamSize = ulong.Parse(mObject["TotalVisibleMemorySize"].ToString()) * 1024;
+                }
+            }
+            catch (Exception)
```
gdkchan, line 26: It's usually a better pratice to catch the specific exception you're looking for, rather than catching any exception type.

Ryujinx.Common/SystemInfo/WindowsSystemInfo.cs
```diff
@@ -9,14 +11,23 @@ internal class WindowsSysteminfo : SystemInfo
...
+                    CpuName = mObject["Name"].ToString();
+                }
 
-            foreach (ManagementBaseObject mObject in new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_OperatingSystem").Get())
+                foreach (ManagementBaseObject mObject in new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_OperatingSystem").Get())
+                {
+                    RamSize = ulong.Parse(mObject["TotalVisibleMemorySize"].ToString()) * 1024;
+                }
+            }
+            catch (COMException)
```
Thog, line 26: As we talked in private about it, this should also catch PlatformNotSuportedException. (some Windows 10 variants doesn't support WMI from some report we got)

PR #1428, opened 2020-07-25
Silence several build warnings

gdkchan:
Silences several warnings + minor code cleanup.

### DIFF
--- a/Ryujinx.Common/Memory/StructArrayHelpers.cs
+++ b/Ryujinx.Common/Memory/StructArrayHelpers.cs
@@ -12,504 +12,630 @@ namespace Ryujinx.Common.Memory
     }
     public struct Array2<T> : IArray<T> where T : unmanaged
     {
+#pragma warning disable CS0169
         T _e0;
         Array1<T> _other;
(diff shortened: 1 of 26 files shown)


jduncanator:
Personally, I'd prefer silencing CS0649 globally, rather than littering the code with pragma statements. We can just silence it in Release builds, so that it does not clutter build logs, whilst keeping the warnings in debug builds (in case they are useful to developers).

gdkchan:
I think the warning is useful in some cases (for example, I just found that SubsamplingX/Y is never initialized thanks to them), so I think its valuable to keep them in some cases. I guess keeping them in debug only might be a good compromise.

jduncanator:
So guessing we're not addressing my feedback then @AcK77? Thought the decision was to silence these globally in release mode?

gdkchan:
Imo it was better to merge that as-is. I was just following what was already being done in other places (like the GPU project, introduced on #919). If we are going to just disable them globally, I think that would be better done as a separately anyway, since that also requires removing all pragmas that were already on the code.

PR #1432, opened 2020-07-26
Add the ability to set Player name in settings.

### LINKED ISSUE #1304: Add feature - change username (nickname) with GUI
**Add new feature to change username (nickname) with GUI:**

- Add a place to change the **default username** in the **setting menu** before starting any game, like the below picture

image

- Add ability to change the **username** when the **game asked** using the keyboard with **popup windows** (small windows to enter new nickname), see below picture

image

- Also, add a virtual keyboard to change the nickname when game asking can be controlled through a controller (wireless Xbox), also **keyboard can be used**, see below examples

image

image
.

jgoyvaerts:
Added a player name input in settings screen and wired it so it's used as the default text input.

Fixes #1304 

image

image

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
@@ -13,7 +13,7 @@ namespace Ryujinx.Configuration
         /// <summary>
         /// The current version of the file format
         /// </summary>
-        public const int CurrentVersion = 11;
+        public const int CurrentVersion = 12;
 
         public int Version { get; set; }
 
@@ -81,6 +81,11 @@ namespace Ryujinx.Configuration
         /// Enables or disables logging to a file on disk
         /// </summary>
         public bool EnableFileLog { get; set; }
+
+        /// <summary>
+        /// Change Default Player Name
+        /// </summary>
+        public string PlayerName { get; set; }
 
         /// <summary>
         /// Change System Language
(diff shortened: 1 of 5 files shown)


gdkchan:
Nice work, I think the name could be improved though. Right now it is called "Player name", but the swkbd input text could be used for other things aswell (even if it's the player name most of the time). Maybe it's better to call it "Software keyboard text" or something like that.

jduncanator:
Adding to @gdkchan's comment, I don't think this should be the first element on the System tab. Should probably go under System Time if anywhere. Also, don't forget to update the JSON schema for the configuration file, as it may be used in the future to validate configuration files, and is used to provide IntelliSense and auto-complete for users editing the configuration file directly.

jgoyvaerts:
Superseded by #1434

Kryokina:
Does this autofill name feature not work anymore?  I cannot find the option for Player Name in the System Settings.

MutantAura:
Can we also not comment on years old pull requests in multiple places @Kryokina. Your issue is a support request and does not belong on old prs. Open an issue if this is a core emulator issue.

Kryokina:
@GamerzHell9137 I saw that, but it does not seem to do the autofill feature in-game, so I was unsure if it was supposed to be the same thing.

@MutantAura Apologies, I will open issues where I mention these relevant pull requests from now on instead of commenting on the pull requests themselves.

PR #1433, opened 2020-07-27
CPU: This PR fixes Fpscr, among other things.

LDj3SNuD:
**Ryujinx.Tests.Unicorn\libs\windows\unicorn.dll:**
Rebuilt library, to allow reading and writing of the FPSCR register by the Tester.
```
Windows SDK Version: 10.0.17763.0
Platform Toolset: Visual Studio 2019 (v142)
Release | x64
0 Warnings
```
qemu\target-arm\unicorn_arm.c:
```diff
@@ -90,6 +90,9 @@ int arm_reg_read(struct uc_struct *uc, unsigned int *regs, void **vals, int coun
                 case UC_ARM_REG_FPEXC:
                     *(int32_t *)value = ARM_CPU(uc, mycpu)->env.vfp.xregs[ARM_VFP_FPEXC];
                     break;
+                case UC_ARM_REG_FPSCR:
+                    *(int32_t *)value = vfp_get_fpscr(&ARM_CPU(uc, mycpu)->env);
+                    break;
             }
         }
     }
@@ -146,6 +149,9 @@ int arm_reg_write(struct uc_struct *uc, unsigned int *regs, void* const* vals, i
                 case UC_ARM_REG_FPEXC:
                     ARM_CPU(uc, mycpu)->env.vfp.xregs[ARM_VFP_FPEXC] = *(int32_t *)value;
                     break;
+                case UC_ARM_REG_FPSCR:
+                    vfp_set_fpscr(&ARM_CPU(uc, mycpu)->env, *(uint32_t *)value);
+                    break;
             }
         }
     }
```

**ARMeilleure\Instructions\InstEmitSystem32.cs:**
Fixed `Vmrs` & `Vmsr`, with reference to `Fpscr`, with the introduction of the `EmitGetFpscr()` and `EmitSetFpscr()` methods, allowing the new `CpuTestMisc32.Vmsr_Vcmp_Vmrs()` Test to pass.

**ARMeilleure\Instructions\InstEmitSimdShift32.cs:**
Updated `EmitSatQ()` method to support `Fpscr.Qc` set if saturation occurs;
Updated the related Tests about `Vqrshrn` & `Vqrshrun`.

**ARMeilleure\Instructions\InstEmitSimdCmp.cs:**
Optimized `EmitFcmpOrFcmpe()`.
~~_Details here:_ [FCMP S1, S2](https://pastebin.com/Dx7Pjg7H) | [FCMP S1, #0.0](https://pastebin.com/dcqqkm9Y)~~

**ARMeilleure\Instructions\InstEmitSimdCmp32.cs:**
Optimized NaN path avoiding emitting unnecessary instructions.

Nits.


ARMeilleure/Instructions/InstEmitSimdCmp.cs
```diff
@@ -538,9 +538,12 @@ private static void EmitFcmpOrFcmpe(ArmEmitterContext context, bool signalNaNs)
...
 
-                    Operand cf = context.AddIntrinsicInt(Intrinsic.X86Comissge, n, m);
-                    Operand zf = context.AddIntrinsicInt(Intrinsic.X86Comisseq, n, m);
-                    Operand nf = context.AddIntrinsicInt(Intrinsic.X86Comisslt, n, m);
+                    Operand nCopy = context.Copy(n);
+                    Operand mCopy = cmpWithZero ? null : context.Copy(m);
+
+                    Operand cf = context.AddIntrinsicInt(Intrinsic.X86Comissge, nCopy, cmpWithZero ? context.VectorZero() : mCopy);
+                    Operand zf = context.AddIntrinsicInt(Intrinsic.X86Comisseq, nCopy, cmpWithZero ? context.VectorZero() : mCopy);
+                    Operand nf = context.AddIntrinsicInt(Intrinsic.X86Comisslt, nCopy, cmpWithZero ? context.VectorZero() : mCopy);
```
FICTURE7, line 546:
```suggestion
                    Operand nCopy = context.Copy(n);
                    Operand mCopy = cmpWithZero ? m : context.Copy(m);

                    Operand cf = context.AddIntrinsicInt(Intrinsic.X86Comissge, nCopy, mCopy);
                    Operand zf = context.AddIntrinsicInt(Intrinsic.X86Comisseq, nCopy, mCopy);
                    Operand nf = context.AddIntrinsicInt(Intrinsic.X86Comisslt, nCopy, mCopy);
```

I believe doing this instead might improve the codegen further for the zero case, because it removes some of the redundant `VectorZero` operations. Likewise for the `op.Size == 1` path.

LDj3SNuD: Those repeated `VectorZero` are wanted, otherwise `VectorZero` is simply spilled; the goal here is to avoid or reduce interactions with the stack.

FICTURE7: Thanks, makes sense. 👍

FICTURE7: FWIW I investigated this further, and some of the `Fill`/`VectorZero` in LCQ can be avoided if
```patch
- Operand mCopy = cmpWithZero ? m : context.Copy(m);
+ Operand mCopy = cmpWithZero ? context.VectorZero() : context.Copy(m);
```
is used instead.

```patch
 .L5:
   vmovdqu xmm0,oword [rsp+0x5c]
   vmovdqu xmm4,xmm0
   vxorps xmm5,xmm5,xmm5
   vcomiss xmm4,xmm5
   setnc bpl
   movzx ebp,bpl
-  vxorps xmm5,xmm5,xmm5
   vcomiss xmm4,xmm5
   setz sil
   movzx esi,sil
-  vxorps xmm5,xmm5,xmm5
   vcomiss xmm4,xmm5
   setc dil
   movzx edi,dil
   xor eax,eax
```
And turns out the LSRA can handle the original suggestion without any spills, since the interval of `m` spans across the 2 blocks without any holes. But I suppose it is a fair tradeoff between the CQ modes.

gdkchan: I think keeping the zero value on a register is better, otherwise it will do the zero-ing pattern (with `XORPS`) every time `VectorZero` is used. The only cases where it's not better, is when there's high register pressure that could cause the zero value to be spilled and reloaded, but I think on HighCq, the register allocator should be smart enough to not spill/reload the zero value everytime for such a short interval.

ARMeilleure/Instructions/InstEmitSystem32.cs
```diff
@@ -186,14 +186,14 @@ public static void Vmrs(ArmEmitterContext context)
                 return;
             }
 
-            MethodInfo info;
+            //MethodInfo info;
```
gdkchan, line 189: I noticed this and other commented lines of code. What's the point of keeping it?

ARMeilleure/State/FPSR.cs
```diff
@@ -5,9 +5,11 @@ namespace ARMeilleure.State
     [Flags]
     public enum FPSR : uint
```
gdkchan, line 6: I would be in favor of merging this with `FPState`. The individual N/Z/C/V entries can be added here, and `Nzcv` can be just `NFlag | ZFlag | CFlag | VFlag`.

LDj3SNuD: The issue here is that I would have to merge a [flags] enumeration with a non-[flags] enumeration, this would imply having to re-shift the values being used as indexes in Register() and GetStorage().FpFlags[] for example; I don't know if I've got the idea.

gdkchan: Yea, right. I overlooked the fact that the values on the other enum are log2(N) and not N.

gdkchan:
Would be nice to also get updated binaries for mac and linux, otherwise I'd imagine the new tests would fail on those platforms.

PR #1439, opened 2020-07-29
Use movd,movq for i32/64 VectorExtract %x, 0x0

FICTURE7:
Prefer `movd`/`movq` over `pextrd`/`pextrq` when extracting element 0 of a vector for i32/i64.

```patch
@@ -84,7 +84,7 @@
   jz .L7
 .L5:
   movdqu xmm0,[rsp+0xa0]
-  pextrq rbx,xmm0,0x0
+  movq rbx,xmm0
   mov rbp,WriteUInt64
   mov rax,[rsp+0xbc]
   mov rcx,rax
@@ -1012,7 +1012,7 @@
   movdqu xmm0,xmm4
   movdqu [rsp+0x124],xmm0
   movdqu xmm0,[rsp+0x124]
-  pextrd ebp,xmm0,0x0
+  movd ebp,xmm0
   mov ebp,ebp
   mov rax,rbp
   mov [rsp+0x8c],rax
```

This makes `Instrinsic.X86Cvtsi2si %x` unnecessary since these operations can now be replaced by a `i32/i64 VectorExtract %x, 0x0` instead.

### DIFF
--- a/ARMeilleure/CodeGen/X86/CodeGenerator.cs
+++ b/ARMeilleure/CodeGen/X86/CodeGenerator.cs
@@ -1146,10 +1146,9 @@ namespace ARMeilleure.CodeGen.X86
             Debug.Assert(src1.Type == OperandType.V128);
             Debug.Assert(src2.Kind == OperandKind.Constant);
 
-            int  count = OperandType.V128.GetSizeInBytes() / dest.Type.GetSizeInBytes();
             byte index = src2.AsByte();
 
-            Debug.Assert(index < count);
+            Debug.Assert(index < OperandType.V128.GetSizeInBytes() / dest.Type.GetSizeInBytes());
 
             if (dest.Type == OperandType.I32)
             {
--- a/ARMeilleure/CodeGen/X86/CodeGenerator.cs
+++ b/ARMeilleure/CodeGen/X86/CodeGenerator.cs
@@ -1152,27 +1152,27 @@ namespace ARMeilleure.CodeGen.X86
             {
                 Debug.Assert(index < 4);
 
-                if (HardwareCapabilities.SupportsSse41)
+                if (index == 0)
                 {
-                    context.Assembler.Pextrd(dest, src1, index);
+                    context.Assembler.Movd(dest, src1);
                 }
                 else
                 {
-                    if (index != 0)
+                    if (HardwareCapabilities.SupportsSse41)
+                    {
+                        context.Assembler.Pextrd(dest, src1, index);
+                    }
+                    else
                     {
                         int mask0 = 0b11_10_01_00;
                         int mask1 = 0b11_10_01_00;
 
-                        mask0 = BitUtils.RotateRight(mask0,     index * 2, 8);
+                        mask0 = BitUtils.RotateRight(mask0, index * 2, 8);
                         mask1 = BitUtils.RotateRight(mask1, 8 - index * 2, 8);
 
                         context.Assembler.Pshufd(src1, src1, (byte)mask0);
                         context.Assembler.Movd  (dest, src1);
                         context.Assembler.Pshufd(src1, src1, (byte)mask1);
-                    }
-                    else
-                    {
-                        context.Assembler.Movd(dest, src1);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


FICTURE7:
While looking into this, I noticed that vector stores uses f32/64 as destination type when extracting the element from a vector.
https://github.com/Ryujinx/Ryujinx/blob/991784868f278b62f7e847321f0cfd7309fe2f79/ARMeilleure/Instructions/InstEmitMemoryHelper.cs#L266-L281
However the slow path uses i32/64.
https://github.com/Ryujinx/Ryujinx/blob/991784868f278b62f7e847321f0cfd7309fe2f79/ARMeilleure/Instructions/InstEmitMemoryHelper.cs#L428-L443

This causes extraneous code to emitted in the fast path since f32/64 values live in an XMM register.
```nasm
.L33:
  mov ecx,ebp
  and ecx,0xfff
  mov ecx,ecx
  movq xmm0,xmm6          ; f32 %y = VectorExtract v128 %x, 0x1
  pshufd xmm0,xmm0,0xfc   ;
  movd [rax+rcx],xmm0
.L34:
  mov eax,ebp
  add eax,byte +0x4
  mov ecx,eax
  and ecx,byte +0x3
  test ecx,ecx
  jz .L37
.L35:
  pextrd edx,xmm6,0x1     ; i32 %y = VectorExtract v128 %x, 0x1
  mov rdi,WriteUInt32
  mov ecx,eax
  call rdi
  jmp short .L39
```

ARMeilleure/CodeGen/X86/CodeGenerator.cs
```diff
@@ -1152,52 +1152,52 @@ private static void GenerateVectorExtract(CodeGenContext context, Operation oper
...
-                if (HardwareCapabilities.SupportsSse41)
+                if (index == 0)
                 {
-                    context.Assembler.Pextrd(dest, src1, index);
+                    context.Assembler.Movd(dest, src1);
                 }
                 else
                 {
-                    if (index != 0)
+                    if (HardwareCapabilities.SupportsSse41)
```
gdkchan, line 1161: Might be better to use a `else if` here, just to avoid the nesting.

ARMeilleure/CodeGen/X86/CodeGenerator.cs
```diff
@@ -1152,52 +1152,52 @@ private static void GenerateVectorExtract(CodeGenContext context, Operation oper
...
-                if (HardwareCapabilities.SupportsSse41)
+                if (index == 0)
                 {
-                    context.Assembler.Pextrq(dest, src1, index);
+                    context.Assembler.Movq(dest, src1);
                 }
                 else
                 {
-                    if (index != 0)
+                    if (HardwareCapabilities.SupportsSse41)
```
gdkchan, line 1189: Same comment as above.

gdkchan:
> This causes extraneous code to be emitted sometimes in the fast path since f32/64 values live in an XMM register and may need shuffles.

I guess it depends on how expensive those shuffles are. Using integer there should reduce it to 2 instructions rather than 3, but involves extracting the value into a GPR, not sure if that would be faster. If it's indeed faster or the same as the current approach, I think would be nice to change it, even if just for consistency.

Other CQ issues I found on the code snippet:
 ```nasm
  and eax,0xfff
  mov eax,eax
```
could be just:
 ```nasm
  and eax,0xfff
```
This one should be easy to optimize, we can just turn `ZeroExtend` when source type is I32 and the destination register is the same as the source register a no-op.
```nasm
  cmp rcx,byte +0x0
  setl dl
  movzx edx,dl
  test edx,edx
  jnz .L35
```
that could be just:
```nasm
  cmp rcx,byte +0x0
  jl .L35
```
or (since the constant is zero):
```nasm
  test rcx,rcx
  jl .L35
```
This one comes from the fact that theres no "Branch with comparison" IR instruction. Just branch if zero or branch if non-zero, which is great to reduce number of IR instructions, but requires extra optimization to turn a sequence of ICompare** and BranchIf** into something good. Might be possible to extend `X86Optimizer` to replace a sequence of ICompare (with a single use) and Branch into a special `X86Jcc` node  or something, with the condition from the comparison (inverted if the branch is `BranchIfFalse`). That would generate a single cmp + jcc sequence.

Anyway, this is out of the scope of this PR, just wanted to leave those observations here in case you are interested in other somewhat simple optimization possibilities.

PR #1442, opened 2020-07-30
Improve branch operations

FICTURE7:
~~_This PR is mostly intended for discussion._~~

~~This adds a new instruction called `BranchIf` in order to handle more "sophisticated" branches as discussed [here](https://github.com/Ryujinx/Ryujinx/pull/1439#issuecomment-666065612). I tried avoiding adding a new type in the IR and used a Constant Operand kind to represent the jump condition kind.~~

~~However I am not sure I like exposing/using `X86Condition` so high up in the IR, I think its probably better to reuse the `CompareEqual`, `CompareGreater` and friends set of instructions to represent the condition kind. It may also be worth considering removing the `Compare` prefix from them (i.e, just `Equal`, `Greater` etc), I think it fits the their boolean behavior better.~~

~~If we want to expose exotic/platform specific condition kind, we can use some kind of encoding scheme on the constant representing the condition kind.~~

* Add a new enum called `Comparison` to represent various comparisons.
* Add a new IR instruction called `BranchIf` to represent different branch conditions. (uses `Comparison`)
* Add a new IR instruction called `Compare` which replaces the various `Compare*` variants. (uses `Comparison`)
* Use `test` when comparing against 0 for `BranchIf` and `Compare`.
* Extend the optimizer to propagate `Compare` operations into their `BranchIfTrue/False` use and turn these uses into `BranchIf` operations.


FICTURE7:
Worth noting, this also makes `BranchIfTrue/False` unnecessary, since they can be replaced by a `BranchIf %x, 0, X86Condition.Equal/NotEqual`. Although it will probably reduce throughput slightly, since `BranchIfTrue/False` uses a single operand, but there are a few places where we could compensate.

gdkchan:
If I were to add something like this, I would introduce a new `Condition` enum, with all the conditions, and then encode it with the instruction itself (by reserving a bit field on the integer value, like 3 bits, to insert it). Then replace all ICompare** instruction with a single ICompare, and all BranchIf** with a single BranchIf. Then, for both ICompare and BranchIf, condition would be encoded in said special bit field.

To be effective, it would also require the replacement of all uses of ICompare + BranchIfTrue or BranchIfFalse with BranchIf, at least when the comparison value is not used anywhere else.

FICTURE7:
> I would introduce a new `Condition` enum, with all the conditions, and then encode it with the instruction itself (by reserving a bit field on the integer value, like 3 bits, to insert it).

Right, I initially had the same idea, but this comes at a slight tradeoff in the CodeGenerator because it has to register multiple `BranchIf` variants to the single `GenerateBranchIf`. There also already exists an enum in `ARMeilleure.Decoders`, which may slightly be awkward.

gdkchan:
I have looked at other IRs, and found the following:
**.NET:**
Has both comparison (like the ones we have now prefixed with ICompare), and conditional branch (like BranchIf proposed here). It also has Brtrue and Brfalse that are equivalent to the current BranchIfTrue and BranchIfFalse.
**Java:**
Only has conditional branches (called Ifcc, where cc is the condition). Also has ifne equivalent to the current BranchIfTrue and ifeq equivalent to BranchIfFalse. Looks like it does not have anything like ICompare that we have now.
**LLVM:**
Seems to be the closest to what we have currently. Has a br instruction that takes a "cond" binary value, 1 if true and 0 if false. The other 2 operands are labels for true and false conditions. This is like our current BranchIfTrue and BranchIfFalse combined into a single instruction. It also has a Icmp instruction that produces said binary value used as "cond" on the Br instruction. On the Icmp instruction, the condition is the first operand, so there's only a single "Icmp" instruction, rather than one for each condition (unlike .NET and Java above, and what we currently do).

Well, I guess that can give some idea of how to design it.
I guess one possible way to do it is passing condition as first operand, then we can have a single ICompare and a single BranchIf. This solves this issue:
> Right, I initially had the same idea, but this comes at a slight tradeoff in the CodeGenerator because it has to register multiple BranchIf variants to the single GenerateBranchIf.

However it does not solve the second:
> There also already exists a Condition enum in ARMeilleure.Decoders, which may slightly be awkward.

I can't think in a way to solve both of them. I guess the solution is just using a different name for the new enum. Maybe `Comparison` or `ConditionCode`.

There's also the option of doing it the LLVM way and just optimize the ICompare and BranchIfTrue/False sequence on the backend, as I suggested on the other thread. Might not be ideal since it would be another optimization pass and take some time, and this is a JIT so, changing the IR sounds like a better compromise.

FICTURE7:
7f3b347bfd2748387e85865eddfd45160134fad1
Adds a new `Compare` instruction and `Comparison` enum to the IR, and removes the various `Compare*` variants. They can be represented using these 2 new types. `Comparison` can be converted to `X86Condition`.

1fab76290a4944ccd193251414a1b3af7edf1861
Adds a new `BranchIf` instruction which reuses the `Comparison` enum to represent the jump condition.

They both expect the `Comparison` source operand to be in the 3rd position. This is because the PreAllocator would require extra, imo unwarranted, modifications in order to preallocate them neatly.

FICTURE7:
f8d3951 
Causes this

gdkchan:
Nice work. What do you think about removing the `BranchIfFalse`/BranchIfTrue IR instructions and replace it with BranchIf Equal 0 and BranchIf NotEqual 0?

FICTURE7:
> What do you think about removing the BranchIfFalse/BranchIfTrue IR instructions and replace it with BranchIf Equal 0 and BranchIf NotEqual 0?

Sounds good to me. 👌

FICTURE7:
For ebbbb689
```patch
@@ -175,7 +174,6 @@
   mov r8,rax
   xor r8,byte +0x13
   and rdi,r8
-  test rdi,rdi
   setl dil
   movzx edi,dil
   mov eax,edi
@@ -200,7 +198,6 @@
   mov rax,[rsp+0xd8]
   mov rsi,rax
   and rsi,rbp
-  test rsi,rsi
   jz .L14
 .L13:
   mov rbx,ReadUInt32
```

Since it is somewhat related to this set of changes, I figure I should include it. Something similar can be done for a few more instructions (e.g `sub`) but these cases are quite rare.

PR #1443, opened 2020-07-30
fs: stub ReadSaveDataFileSystemExtraDataWithMaskBySaveDataAttribute

Thog:
This is required by AC:NH for the new online save features of 1.4.0

NOTE: I haven't bothered to add the IPC comment as most of the FS interface is missing that right now. (will surely do something about that in a next PR)


EmulationFanatic:
Tested working.

Issue #913, opened 2020-01-31
Decouple kernel from the HLE project


gdkchan:
We should move the kernel to a separate project. This will make testing easier and also creates a more clear separation of concerns.

There are several steps required to do that:
- [ ] Move kernel state out of the Horizon class. Things like the scheduler, the allocators etc should not be part of the Horizon class, instead it should be part of a separate one for the kernel state.
- [ ] Following that change, we should stop using the Horizon type on the kernel code entirely.
- [ ] We should use the new IPC system that was implemented on the kernel.
Currently we use the old "SendSyncRequest" syscall implementation: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcIpc.cs#L109
We should use the new one instead: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Kernel/SupervisorCall/SvcIpc.cs#L171
This is caused by a misunderstanding of where the IPC handling was made when this code was written. It was assumed that the kernel handles some things that are actually handled by the service. 
- [ ] After those changes, we should also stop using kernel objects on services. Instead we can use the proper syscalls. We'll need a dummy process and thread per service for that.

After all that is done, we can finally move the kernel to it's own project.

PR #1447, opened 2020-08-03
IPC refactor part 1: Use explicit separate threads to process requests

### LINKED ISSUE #1234: Surface Flinger v2 stops emulator to close properly
#981 SurfaceFlinger V2 is making games crash when closing the emulator when clicking X on the main window. Seems like the time when it crashes is when there's some kind of loading when you start the emulation, but it doesn't need to be, it can crash later on too but most frequently its when the emulation starts and around logo/loading screens.
4284, the build before SurfaceFlinger V2 has no issues when closing the emulator.

EDIT: Its confirmed that its happening in other games like Animal Crossing so its probably happening in all games at random times.

Rune Factory 4 Special
image
image

Good Job
image

BOXYBOY+BOXYGIRL
image

gdkchan:
**Main goal:**

Currently, HLE IPC uses the ThreadPool to handle processing of IPC messages. That means it will get a free thread from the pool, and use it for processing. This does not match what the real OS does, and for this reason, an accurate IPC implementation becomes difficult with this method. This change introduces the `ServerBase` class that internally, creates a thread that will handle the IPC processing. By default, all services uses a "CommonServer", and services can optionally create their own "servers" by passing it on the `IpcService` constructor.

This will eventually allow the `ReplyAndReceive` syscall to be used for iPC processing on HLE services aswell, and removing all the special handling that we have to make HLE work right now. This also helps with #913 since we no longer manage IPC processing threads on the kernel.

**Some fixes:**

This also attempts to fix crash/hangs on exit that have been happening for a while. While it was not my intention to look at that as part of those changes, they allowed the removal of the "ThreadCounter" event, which was used on disposal to wait until all threads exited, to dispose of the allocated memory. Removing it uncovered more problems, so I decided to just fix them aswell.

Below you can see some of the issues I fixed:
- `UnpauseAndTerminateAllThreadsExcept` was incorrect, it would only wait the termination of one thread, instead of all threads alive. Fixed using mesosphere as reference.
- `PrepareForTermination` was not calling `RequestInterrupt`, required to make the thread stop and call the ISR (or at least the closest thing we have right now).
- The ISR mentioned above was not calling `HandlePostSyscall`, required to make the thread actually exit.

Aditionally, the CPU emulator was modified to check for exit without executing any more guest code after a syscall or any CPU exception. This ensures that the thread will exit immediate when `StopRunning` is called from kernel. This is required to make the assumption that it's safe to dispose the memory after calling `StopRunning` for all threads that where running kernel code (either due to a syscall or interrupt request) hold. The PTC version was incremented due to this small change.

It's hard to tell if this is going to fix all exit issues since this has so many moving parts, but as far I can tell it's fixed.

Fixes #1234.

**Possible outcomes:**

This has the potential of introducing issues, since now less threads are used for IPC processing (for example, a HLE function blocking indefinitely could hang all IPC processing), it could also cause performance issues. On the other hand, it could also fix issues caused by race conditions on HLE code, since we do the processing in just one or a few threads. For this reason, it would be nice to test it with a few games.


EmulationFanatic:
Tested this on a handful of games and none of them hang on exit anymore.

Thog:
Rebase required

Thog:
Need rebase.

jduncanator (approved):
Looks good. Nice first step to the ultimate goal!

PR #1451, opened 2020-08-04
Replace the Audio Backend ComboBoxText to a ComboBox with a custom model

Xpl0itR:
This PR fixes the issue of the dummy backend being selected by default when OpenAL is unavailable.
All the backends are in the list but OpenAL and SoundIO is unselectable by the user until their IsSupported function return true. If an unavailable backend is selected by the config, it will still be selected in the UI meaning we wont have that same crash from the 1st version.

### DIFF
--- a/Ryujinx/Ui/SettingsWindow.cs
+++ b/Ryujinx/Ui/SettingsWindow.cs
@@ -553,7 +553,9 @@ namespace Ryujinx.Ui
             ConfigurationState.Instance.Graphics.ResScaleCustom.Value          = resScaleCustom;
 
             if (_audioBackendSelect.GetActiveIter(out TreeIter activeIter))
+            {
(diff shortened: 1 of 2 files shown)


Ryujinx/Ui/SettingsWindow.cs
```diff
@@ -527,6 +552,9 @@ private void SaveToggle_Activated(object sender, EventArgs args)
             ConfigurationState.Instance.Graphics.ResScale.Value                = int.Parse(_resScaleCombo.ActiveId);
             ConfigurationState.Instance.Graphics.ResScaleCustom.Value          = resScaleCustom;
 
+            if (_audioBackendSelect.GetActiveIter(out TreeIter activeIter))
```
AcK77, line 555: Could be nice to add some brackets

PR #1458, opened 2020-08-07
IPC refactor part 2: Use ReplyAndReceive on HLE services and remove special handling from kernel

gdkchan:
This continues the work started on #1447.

The main change here is that HLE services now use the `ReplyAndReceive` syscall to receive IPC requests, rather than having special handling on kernel that would call the service function directly after a call to `SendSyncRequest` by the game. This also includes changes to allow services to use syscalls directly.

While working on this, I hit a number of issues, limitations or bugs that I had to either fix or work around:
- PID and process handles was being ignored on a number of services. That was fine before since they had access to the game process directly, however  this is no longer the case. So I made a change on the places I found it was necessary to use the process handle or PID, for identification purposes (as is the case of surface flinger/nvservices) or to access the game process address space (as is the case for audio).
- Transfer memory handles were also ignored. They are still unused, however it now at least closes the handle, as the simple act of the game passing the handle to the service will cause it to be added to the service handle table, which increments the ref count, and the memory can't be unmapped until the service closes the handle. Not closing caused issues on some games as it tried to re-use the transfer memory region, but it was still mapped/locked.
- IPC message serialization had a bug where the written raw data size was not word aligned. This was fixed.
- IPC message deserialization had a bug where the raw data size was too high, because it was also counting the padding size. This was fixed.
- IPC message handling had poor support for pointer/receive list buffers. It was extended just enough to allow proper IPC to work, it's still incomplete but "good enough", at least for now.
- IPC pointer buffer size was too low. This caused failures when trying to send buffers to nvservices, and mapping send and receive buffers on the same region that spans multiple pages is not allowed by the kernel. The higher pointer buffer size makes it use pointer and receive list buffers instead, where it is supported.
- Nvservices was ignoring the path buffer size when reading the path string. This could cause it to read garbage if the string was not properly null-terminated in memory. This was fixed.
- A bug in kernel prevented memory used for temporary server buffers used on misaligned pages from being freed properly. This was fixed.
- A bug in kernel caused the wrong memory region to be freed when a misaligned IPC buffer was used. This was fixed.
- A bug in kernel could cause failures when trying to restore client buffer region permission after a request in some specific cases. This was fixed.

**Changes in preparation for kernel decoupling:**
- Repurposed `IVirtualMemoryManager` and introduced the `IProcessContextFactory`/`IProcessContext` interfaces. Kernel no longer needs to care about how the memory manager is implemented, and soon will no longer need to know about the CPU project. This changes allows the implementation of a separate memory manager used by HLE services, which uses *less memory* due to the fact that it does not use a flat page table, at the cost of performance. This is fine as we are not going to do lot's of sequential read/writes from HLE services, so we don't really need all the performance there. And given the large number of services, the memory usage is more important.

**Future work:**
The next step will be decoupling the kernel project.  This will require changes to make services stop using kernel objects directly. This PR already does some progress on that front, but there's still work to be done (this will be done on part 3). After that I can think about designing a new API that will be used for auto-(de)serialization of IPC messages (this will be done on part 4, or maybe part 3, not sure yet).

**Testing:**
This has more potential to break things than part 1, so I recommend even more throrough testing on this one.

**Acknowledgements & Closing words:**
Some improvements to the kernel used mesosphere as reference. The implementations were it was used as reference was `CreateProcess` and `StartProcess` method body and the `ProcessCreationFlags` enum (it has a different name there).

Thanks to Thog for help troubleshooting some issues.


Thog:
Need rebase and as a note will cause different behaviours in multiple games as the resource applet id isn't 0 anymore.

Ryujinx.HLE/Exceptions/ServiceNotImplementedException.cs
```diff
@@ -57,7 +57,7 @@ private string BuildMessage()
             // Print the IPC command details (service name, command ID, and handler)
             (Type callingType, MethodBase callingMethod) = WalkStackTrace(new StackTrace(this));
 
-            if (callingType != null && callingMethod != null)
+            /*if (callingType != null && callingMethod != null)
```
Thog, line 60: uncomment this?

Ryujinx.HLE/HOS/Kernel/Common/OnScopeExit.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Ryujinx.HLE.HOS.Kernel.Common
```
Thog, line 3: Maybe move this to Ryujinx.Common?

Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
```diff
@@ -2016,6 +2009,14 @@ public KernelResult UnmapNoAttributeIfStateEquals(ulong address, ulong size, Mem
                         FreeSinglePage(_memRegion, ConvertVaToPa(endAddrTruncated));
                     }
 
+                    KernelResult result = DoMmuOperation(
```
Thog, line 2012: Any reasons of moving this? was this a diff compared to re?

Ryujinx.HLE/HOS/Services/Nv/INvDrvServices.cs
```diff
@@ -229,8 +227,9 @@ public ResultCode Open(ServiceCtx context)
             if (errorCode == NvResult.Success)
             {
                 long pathPtr = context.Request.SendBuff[0].Position;
+                long pathSize = context.Request.SendBuff[0].Size;
 
-                string path = MemoryHelper.ReadAsciiString(context.Memory, pathPtr);
+                string path = MemoryHelper.ReadAsciiString(context.Memory, pathPtr, pathSize);
```
Thog, line 232: nice catch 👍

Thog (approved):
Some small changes, otherwise lgtm!

Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
gdkchan: The memory needs to be freed before the unmap operation, because it does address translation of the virtual address to physical to get the address to be freed, and doing that is not possible after unmap. This does not match the actual kernel because it does ref counting of those pages instead (and we don't right now).

Ryujinx.HLE/Exceptions/ServiceNotImplementedException.cs
gdkchan: `Context.Session.Service` is gone now, so I will need to find a new way to make it work.

gdkchan:
Had to change some things to make the emulator actually exit cleanly:
- Make surface flinger stop waiting once dispose is called. Otherwise the thread migh get stuck waiting for a frame to be returned, and this may never happen if the GPU was already disposed.
- Horizon calls dispose on `AudioRendererManager`. This dispose methods eventually disposes a event that might be still used by the command processor, so that might result in a `ObjectDisposedException`. So I changed it to stop the audio renderer and wait the exit of the processing thread, before disposing the event. I'm not sure if this is the correct fix, it seems that `AudioRenderSystem` has the logic to properly dispose it calling stop beforehand, but that is not accessible from the Horizon class.
- `GlRenderer` has a exit method that is supposed to make the update loop exit, however it never waits until the thread actually exits, so its possible that other things are finalized and objects set to null while the thread is still running, and the thread expects the objects to be usable still. This could lead to a `NullReferenceException` in some cases (if the thread does not exit in time). I fixed it by adding a event and waiting for the thread exit. I would prefer to call `Thread.Join`, but it looks like that the thread is internal to the `GLWidget` and can't be easily accessed. If there is a better method please let me know.

@Thog I recommend taking a look at those changes as you wrote some of the code being modified here.

Ryujinx.HLE/HOS/Kernel/Common/OnScopeExit.cs
gdkchan: Is it really necessary? I would expect that to be used only in Horizon code just due to their extensive use of RAII.

AcK77 (approved):
LGTM! Pretty nice work!

Thealexbarney (approved):
Looks fine. I didn't notice any issues

Thog:
Need a rebase and some fresh testing.

PR #1474, opened 2020-08-13
Add missing depth-color conversions in CopyTexture

mageven:
Texture copies between depth and color formats aren't allowed in OGL.

Bgra Support (https://github.com/Ryujinx/Ryujinx/pull/1418) exposed an existing bug in texture copies with a new OGL error-
```
GL_INVALID_OPERATION error generated. The <type> and <format> pair is not suppported
```

The root cause for the error was https://github.com/Ryujinx/Ryujinx/pull/1198, which relaxed format matching rules.
But this wasn't reflected in `Methods.CopyTexture()`. This PR updates that check.

Resolves the above OGL error SPAM, removes blur flicker and fixes performance regression after Bgra support in ZLA.

**Note:** Does not fix all graphical glitches in ZLA.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
@@ -28,9 +28,14 @@ namespace Ryujinx.Graphics.Gpu.Engine
             // When the source texture that was found has a depth format,
             // we must enforce the target texture also has a depth format,
             // as copies between depth and color formats are not allowed.
-            if (srcTexture.Format == Format.D32Float)
+            switch (srcTexture.Format)
             {
-                dstCopyTexture.Format = RtFormat.D32Float;
+                case Format.S8Uint         : dstCopyTexture.Format = RtFormat.S8Uint; break;
+                case Format.D16Unorm       : dstCopyTexture.Format = RtFormat.D16Unorm; break;
+                case Format.D24X8Unorm     : dstCopyTexture.Format = RtFormat.D24Unorm; break;
+                case Format.D32Float       : dstCopyTexture.Format = RtFormat.D32Float; break;
+                case Format.D24UnormS8Uint : dstCopyTexture.Format = RtFormat.D24UnormS8Uint; break;
+                case Format.D32FloatS8Uint : dstCopyTexture.Format = RtFormat.D32FloatS8Uint; break;
             }
 
             Texture dstTexture = TextureManager.FindOrCreateTexture(dstCopyTexture, srcTexture.ScaleMode == Image.TextureScaleMode.Scaled);
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
@@ -30,12 +30,12 @@ namespace Ryujinx.Graphics.Gpu.Engine
             // as copies between depth and color formats are not allowed.
             switch (srcTexture.Format)
             {
-                case Format.S8Uint         : dstCopyTexture.Format = RtFormat.S8Uint; break;
-                case Format.D16Unorm       : dstCopyTexture.Format = RtFormat.D16Unorm; break;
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
```diff
@@ -28,9 +28,14 @@ private void CopyTexture(GpuState state, int argument)
             // When the source texture that was found has a depth format,
             // we must enforce the target texture also has a depth format,
             // as copies between depth and color formats are not allowed.
-            if (srcTexture.Format == Format.D32Float)
+            switch (srcTexture.Format)
             {
-                dstCopyTexture.Format = RtFormat.D32Float;
+                case Format.S8Uint         : dstCopyTexture.Format = RtFormat.S8Uint; break;
```
gdkchan, line 33: nit: We don't usually align the `:`.

Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
```diff
@@ -28,9 +28,14 @@ private void CopyTexture(GpuState state, int argument)
...
-            if (srcTexture.Format == Format.D32Float)
+            switch (srcTexture.Format)
             {
-                dstCopyTexture.Format = RtFormat.D32Float;
+                case Format.S8Uint: dstCopyTexture.Format = RtFormat.S8Uint; break;
+                case Format.D16Unorm: dstCopyTexture.Format = RtFormat.D16Unorm; break;
+                case Format.D24X8Unorm: dstCopyTexture.Format = RtFormat.D24Unorm; break;
+                case Format.D32Float: dstCopyTexture.Format = RtFormat.D32Float; break;
+                case Format.D24UnormS8Uint: dstCopyTexture.Format = RtFormat.D24UnormS8Uint; break;
+                case Format.D32FloatS8Uint: dstCopyTexture.Format = RtFormat.D32FloatS8Uint; break;
```
jduncanator, line 38: You can use a `switch` expression here, rather than a `switch` statement setting the same variable each time:

```cs
dstCopyTexture.Format = srcTexture.Format switch {
    Format.S8Uint => RtFormat.S8Uint
    Format.D16Unorm => RtFormat.D16Unorm
    Format.D24X8Unorm => RtFormat.D24Unorm
    Format.D32Float => RtFormat.D32Float
    Format.D24UnormS8Uint => RtFormat.D24UnormS8Uint
    Format.D32FloatS8Uint => RtFormat.D32FloatS8Uint
};
```

gdkchan: Doesn't switch expressions always requires a default case? (Also, missing comma).

jduncanator: Sorry, was just a "pseudo-code" example that I typed in the reply. They don't require a default case, but you will get a warning if the switch is non-exhaustive. Feel free to add a default case if you want.

Thog: I personally find a switch expression way more readable here tbh.

jduncanator: Given we use switch expressions in heaps of other places, we need to remain consistent. Either decide to use them, or not, otherwise it's difficult to maintain and recommend "best practices". I only recommended it here because GDK makes extensive use of them in other places in the GPU project. I agree the switch expression has readability issues, but it has those issues in all instances, not just in this instance (tbh, there are far more "unreadable" switch expressions elsewhere in the project).

I'm fine for not using it here, but we then need to make an active decision not to use them elsewhere, and phase out their existing use.

Thealexbarney: I feel like this is what switch expressions are made for. Mapping one set of values to another.

mageven: Followed maintainers' wishes

Issue #1472, opened 2020-08-12
nfp:user IUser: 3 (StartDetection) is not implemented


albfan:
## Bug Report

```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> Ryujinx.HLE.Exceptions.ServiceNotImplementedException: The service call is not implemented.

Guest Stack Trace:
Process: Application, PID: 81
   sdk:0x000c43b0 nn::sf::cmif::client::CmifDomainProxyClientCoreProcessor<nn::sf::hipc::client::Hipc2ClientSessionManagedProxyKindBase<nn::sf::hipc::client::Hipc2ProxyKindBase<(nn::sf::hipc::detail::MessageType)4> >, nn::sf::cmif::detail::CoreMethodInfo<std::__1::tuple<nn::sf::cmif::detail::InRawArgumentInfo<8ul, 4ul, 0l> >, 8ul, 0ul, false>, std::__1::tuple<nn::sf::cmif::detail::InRawArgumentInfo<8ul, 4ul, 0l> > >::Process(nn::sf::cmif::client::CmifDomainClientMessage<nn::sf::hipc::client::Hipc2ClientSessionManagedProxyKindBase<nn::sf::hipc::client::Hipc2ProxyKindBase<(nn::sf::hipc::detail::MessageType)4> > >::ProxyBaseObject*, nn::sf::cmif::client::InRawClientArgumentType<8ul, 4ul, void>, unsigned int)
   sdk:0x00133728 Sub000000000ad9a728
   sdk:0x0012f534 nn::nfp::StartDetection(nn::nfp::DeviceHandle const&)
   main:0x0095f6c0 al::NfpControllerThread::executeCommandStartDetection()
   main:0x0095f120 al::NfpControllerThread::executeCommand(sead::Thread*, long)
   main:0x0075b040 sead::Thread::run_()
   main:0x0075a908 sead::Thread::ninThreadFunc_(void*)
   sdk:0x0008a8a8 Sub000000000acf18a8
   sdk:0x0008e958 Sub000000000acf5958
   ??? Sub0000000000000000

Raw Request Data:
00000000   01 00 18 00 02 00 00 00  00 00 00 00 00 00 00 00   ················
00000010   53 46 43 49 00 00 00 00  03 00 00 00 00 00 00 00   SFCI············
00000020   00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00   ················
[...]

PR #1476, opened 2020-08-13
WIP: Avoid errors due to non implemented services on execution

albfan:
Initial solution to avoid services marked as unimplemented make a game to crash

relates with #1472

### DIFF
--- a/Ryujinx.HLE/HOS/Services/IpcService.cs
+++ b/Ryujinx.HLE/HOS/Services/IpcService.cs
@@ -118,7 +118,19 @@ namespace Ryujinx.HLE.HOS.Services
 
                     Profile.Begin(profile);
 
-                    result = (ResultCode)processRequest.Invoke(service, new object[] { context });
+                    try
+                    {
+                        result = (ResultCode)processRequest.Invoke(service, new object[] { context });
+                    }
+                    catch (ServiceNotImplementedException ex)
+                    {
+                        // Skip exceptions on service not implemented on execution
+                        bool debug = false;
+                        if (debug)
+                        {
+                            throw ex;
+                        }
+                    }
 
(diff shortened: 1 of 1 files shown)


albfan:
@AcK77 I suppose a log message in execution to trace unimplemented services would be enough. Maybe if debug mode will be a thing, throw an exception on that case.

AcK77:
Throw a `ServiceNotImplementedException` is an expected behavior, since we don’t want to stub service calls, at least not without checking by RE what the call do! Ignore missing service setting is just an hack who shouldn’t be used. And `try/catch` the exception to ignore it is the same. It’s  not the way we want to go.

Thog:
I agree with @AcK77, the ignore missing service setting is only here to be able to "collect" what call need to be implemented.
Crashing is an expected behaviour and we don't want to rely on any hacks for releases
However, in the future we should try to make the emulator stop emulation and generate an error report to the user to improve the user experience.

albfan:
As feedback, no crashes with this PR. Not sure what those commands represents and what I'm missing, but looks like developers first, ending in a real bad user experience (unneeded crash).

Agreed that keep the codebase maintanable is a goal. Count on me to implement these services. Will continue providing feedback on related issue.

Thog:
@albfan Problem is that if you ignore those services call instead of crashing, the game will receive unexpected data and that would totally cause undefined behaviours as a result. (you could end up corrupting a save in a non recoverable way for example)

That's why, it's safer to crash here.

albfan:
@Thog I see. Will keep in mind. By now is a big improvement for my use case. But not sure how to collect info to understand what the game is trying to do, and what it needs to implement.

Then this report is enough or you need something else? (I can add more traces and collect more info if you find it useful

PR #1480, opened 2020-08-14
Fix asserts on VP9 decoder in debug mode and build warnings

gdkchan:
This fixes some build warnings and debug asserts caused by uninitialized fields on the VP9 decoder.


Ryujinx.Graphics.Nvdec.Vp9/Types/Vp9Common.cs
```diff
@@ -83,8 +83,6 @@ internal struct Vp9Common
         public Ptr<Vp9EntropyProbs> Fc;
         public Ptr<Vp9BackwardUpdates> Counts;
 
-        public bool FrameParallelDecodingMode;
```
gdkchan, line 86: Removed this as it was always false, the only ting it does is disabling frame counts calculation for backward updates on the parser, To my knowledge NVDEC always calculates them, so it behaves as if "frame parallel decoding mode" was always false.

PR #1484, opened 2020-08-18
Improve static branch prediction along fast path for memory accesses

FICTURE7:
According to the Intel Optimization Reference Manual, forward conditional branches are predicted to be not taken and backward conditional branches are predicated to be taken, when they are not in the branch target buffer (See section 3.4.1.2 Static Branch Prediction). This PR changes the order of the blocks to take advantage of this.

Before
```asm
.L4:
  add rbx,byte -0x30
  mov rax,0xffffff8000000007
  mov rcx,rbx
  and rcx,rax
  jz .L7
.L5:
  mov rax,WriteUInt64
  mov rcx,rbx
  mov rdx,r13
  call rax
  jmp short .L9
.L7:
  mov rax,rbx
  shr rax,byte 0xc
  mov rcx,PAGE_TABLE_0
  mov rax,[rcx+rax*8]
  test rax,rax
  jng .L5
.L8:
  mov rcx,rbx
  and rcx,0xfff
  mov [rax+rcx],r13
.L9:
```

After
```asm
.L4:
  add rbx,byte -0x30
  mov rax,0xffffff8000000007
  mov rcx,rbx
  and rcx,rax
  jnz .L8
.L5:
  mov rax,rbx
  shr rax,byte 0xc
  mov rcx,PAGE_TABLE_0
  mov rax,[rcx+rax*8]
  test rax,rax
  jng .L8
.L6:
  mov rcx,rbx
  and rcx,0xfff
  mov [rax+rcx],r13
  jmp short .L9
.L8:
  mov rax,WriteUInt64
  mov rcx,rbx
  mov rdx,r13
  call rax
.L9:
```
Ideally this would be improved by some kind of scheduling, but it is good do this early anyways for throughput reasons. Also, scheduling could order the slow path at the end of the list of blocks so that we can omit the jump to `.L9` in `.L6`.

Took the opportunity to change vector stores to extract as I32/64 instead of F32/64 for vector stores as well.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
@@ -124,21 +124,12 @@ namespace ARMeilleure.Instructions
 
         private static void EmitReadInt(ArmEmitterContext context, Operand address, int rt, int size)
         {
-            Operand isUnalignedAddr = EmitAddressCheck(context, address, size);
-
-            Operand lblFastPath = Label();
             Operand lblSlowPath = Label();
             Operand lblEnd      = Label();
 
-            context.BranchIfFalse(lblFastPath, isUnalignedAddr);
+            Operand isUnalignedAddr = EmitAddressCheck(context, address, size);
 
-            context.MarkLabel(lblSlowPath);
-
-            EmitReadIntFallback(context, address, rt, size);
-
-            context.Branch(lblEnd);
-
-            context.MarkLabel(lblFastPath);
+            context.BranchIfTrue(lblSlowPath, isUnalignedAddr);
 
             Operand physAddr = EmitPtPointerLoad(context, address, lblSlowPath, write: false);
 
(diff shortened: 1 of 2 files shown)


riperiperi (approved):
Straightforward and works in my testing. I'm not aware of any areas that are entirely memory bound, but I wsan't able to notice a difference in sand kingdom initial view, though I am on Zen 2 rather than intel.

FICTURE7:
FWIW, I implement a simple block placement pass which re-orders cold blocks at the end of the list of blocks in a CFG FICTURE7@13914923fa5ff738f38b21e7fc3b63e42eff8eb3. It is fairly fast since it uses nodes (blocks) instead of edges to figure out the placement of blocks. It is also not very clear how more valuable would using edges be in our case. This could probably be done in the `EmitterContext` itself as well to avoid a new pass.

Here is a sample [diff](https://www.diffchecker.com/525h5Nq1). This seems quite promising, because it moves the `EmitSynchronization` stuff and epilogue from call guards away from steamy code in loops.

gdkchan:
The branch elimination pass would be useful for (guest) load/stores with a constant address. In those cases, the branch condition becomes a constant and some paths of that jumps into the fallback code can be removed.

PR #1489, opened 2020-08-21
Implement vfnms

### LINKED ISSUE #1477: Unimplemented VFNMS.F64 instruction

Requires the vfnms.f64 instruction.</br>

```
The instruction at 0x00000000002e05e0 (opcode 0xee908ba1) is undefined!
```

### Instruction name
```
EE 90 8B A1    vfnms.f64 d8, d16, d17
```

### Required by: 

https://github.com/Ryujinx/Ryujinx-Games-List/issues/1295

saldabain:
Closes #1477 

Implements slow path of vfnms (FMA instructions not implemented).


ARMeilleure/Decoders/OpCodeTable.cs
```diff
@@ -836,6 +836,7 @@ static OpCodeTable()
             SetA32("111100101x11xxxxxxxxxxxxxxx0xxxx", InstName.Vext,     InstEmit32.Vext,     typeof(OpCode32SimdExt));
             SetA32("<<<<11101x10xxxxxxxx101xx0x0xxxx", InstName.Vfma,     InstEmit32.Vfma_S,   typeof(OpCode32SimdRegS));
             SetA32("<<<<11101x10xxxxxxxx101xx1x0xxxx", InstName.Vfms,     InstEmit32.Vfms_S,   typeof(OpCode32SimdRegS));
+            SetA32("<<<<11101x01xxxxxxxx10>>x0x0xxxx", InstName.Vfnms,    InstEmit32.Vfnms,    typeof(OpCode32SimdReg));
```
riperiperi, line 839: Vfnms is a scalar instruction, so should use the `OpCode32SimdRegS` decoder.

https://static.docs.arm.com/ddi0487/ea/DDI0487E_a_armv8_arm.pdf#G41.5054688

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
```diff
@@ -263,6 +263,15 @@ public static void Vext(ArmEmitterContext context)
             }
         }
 
+        public static void Vfnms(ArmEmitterContext context)
+        {
+            //TODO: Use FMA Instruction set.
+            EmitVectorTernaryOpF32(context, (op1, op2, op3) =>
```
riperiperi, line 269: This is a scalar instruction. You need to use `EmitScalarTernaryOpF32`. Would also be good to have a FastFP/SSE2 path for this.

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
```diff
@@ -365,6 +365,32 @@ private static IEnumerable<ulong> _1D_F_()
             CompareAgainstUnicorn();
         }
 
+        [Test, Pairwise, Description("VNMFS.<size> <Vd>, <Vm>, <Vn>")]
+        public void Vnmfs([Values(0u)] uint rd,
```
riperiperi, line 369: `Vfnms`, same above. The input is also S rather than V.

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
```diff
@@ -365,6 +365,32 @@ private static IEnumerable<ulong> _1D_F_()
...
+                         [Values(1u, 0u)] uint rn,
+                         [Values(2u, 0u)] uint rm,
+                         [Values(2u, 3u)] uint size,
+                         [Random(RndCnt)] ulong z,
+                         [Random(RndCnt)] ulong a,
+                         [Random(RndCnt)] ulong b)
+        {
+            uint opcode = 0xe900A00; // VNMFS.F32 D0, D0, D0
```
riperiperi, line 377: This is not a valid opcode. The comment is not right either, there is no vector version of this opcode.

The real base instruction would be `ee900a00`, which is `vfnms.f32 s0, s0, s0`. As described above, the fp16 variant is not supported and should be filtered by the decoder.

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
```diff
@@ -365,6 +365,32 @@ private static IEnumerable<ulong> _1D_F_()
...
+
+            opcode |= ((rm & 0xf) << 0) | ((rm & 0x10) << 1);
```
riperiperi, line 379: Please keep in mind that the encoding for registers flips when the size is f64 (Vm:M for fp32, M:Vm for fp64). You can see an example of how this is handled in Movi_S.

riperiperi: (same for d and n)

ARMeilleure/Decoders/OpCodeTable.cs
riperiperi: Additionally, the size should be `1x` rather than `>>`, as we do not allow fp16. (`>>` would allow `01`)

`"<<<<11101x01xxxxxxxx101xx0x0xxxx"`

riperiperi (changes requested):

Ryujinx.Tests/Cpu/CpuTestSimdReg32.cs
saldabain: Thank you for the awesome and clear comments -- I'll get on fixing these right away :)

Thog:
Need rebase.

PR #1504, opened 2020-08-26
IPC refactor part 3: Kernel decoupling

### LINKED ISSUE #913: Decouple kernel from the HLE project

gdkchan:
Depends on #1458.

The main change here is the kernel decoupling (closes #913), it adds 3 new projects:
- Ryujinx.Horizon.Kernel: The kernel, which now only depends on `Ryujinx.Memory`, `Ryujinx.Common` and `Ryujinx.Horizon.Common` (see below).
- Ryujinx.Horizon: Contains reimplementation of SDK library code in C# (to be used by services implementation). Should also have the services implementation in the future.
- `Ryujinx.Horizon.Common`: Common code used by the above 2 projects. It was added to avoid a cyclic dependency between the above 2 projects if they had to reference each other. Contains the `Result` struct and a few other utilities.

Services also now register themselves on SM, instead of SM scanning for services. This allows for a cleaner implementation of SM. It is also possible to run initialization code on the service thread. This is used for creating things like shared memory and events that needs to be accessed by other services or other components of the emulator. The hacky allocation of shared memory was also removed, and the proper syscall is used instead. It also does proper mapping of transfer memory now, instead of directly reading it from the game process memory.

Most of the SDK OS APIs implementation, and some of the new kernel syscalls, used Atmosphere as reference. For now, only what was strictly required was implemented, it will be extended in the future.

KIP loading was disabled for now, as it is supposed to be entirely internal to the kernel, and I still need to come up with a good way to load KIPs without exposing the kernel internal implementation to other projects. It was not really available from the GUI anyway, it was only usable with manual code modification.

WIP until the other parts are reviewed/merged.


Thog:
As Atmosphère was used as a reference we will need to add a reference to it in the THIRDPARTY.md file (https://github.com/Atmosphere-NX/Atmosphere/blob/master/docs/licensing_exemptions/MIT_LICENSE)

gdkchan:
> 
> 
> As Atmosphère was used as a reference we will need to add a reference to it in the THIRDPARTY.md file (https://github.com/Atmosphere-NX/Atmosphere/blob/master/docs/licensing_exemptions/MIT_LICENSE)

Done, not sure if I did it right though...

AcK77:
Needs a rebase since #1458 is merged now!

Benedani:
So, does this do anything with #1769 merged 7 months ago?

gdkchan:
Yes, it has a lot more changes, this was just one of them.
At this point though, I just plan to pull the changes and open individual PRs with them, rebasing this would be a nightmare as the code changed a lot over the past months.

PR #1518, opened 2020-08-31
Clear JIT cache on exit

gdkchan:
This does a number of changes to the CPU JIT:
- Remove the call flag that was used to indicate functions that are call targets among other, used as rejit hint. Since the `GetFunctionAddress` helper functions were introduced, we can just pass a rejit hint flag direct to the translation function. The removal of this flag makes the code a bit simpler, and avoids bugs related to ARM32 thumb mode, that also makes use of the least significant address bit.
- Use proper allocators for the JIT cache and also for the jump table. In the future, this will enable removal of functions from the JIT cache, and it is already used to some extent on this PR, see below.
- Clear the JIT cache and jump tables on exit: When all threads exits, all functions are removed from the cache. This fixes a issue where the JIT would leak memory when stopping emulation, leading to a "out of memory" error on subsequent runs. Note: When stopping emulation, it still seems to leak some memory, but I don't think the CPU is at fault for that.
- Partially remove function from JIT cache when code is unmapped: This fixes a issue that could occur if an NRO (for example) is unmapped while the background translator is still accessing the meemory region. I only observed this issue when loading homebrew with `hbloader`, but it might also affects games that makes use of NROs.


riperiperi (commented):
Looks cool, just two comments. Will need to test this one with and without existing PPTC info to make sure it doesn't have any regressions.

Since you mentioned it could affect NRO unloading, I'd probably recommend people to test smash.

ARMeilleure/Translation/Translator.cs
```diff
@@ -319,5 +360,62 @@ private static void EmitSynchronization(EmitterContext context)
...
+            if (toDelete.Length != 0)
+            {
+                // If rejit is running, stop it as it may be trying to rejit the functions we are
+                // supposed to remove.
+                ClearRejitQueue();
+            }
+
+            foreach (var kv in toDelete)
+            {
+                if (_funcs.TryRemove(kv.Key, out TranslatedFunction func))
```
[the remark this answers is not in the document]
gdkchan: Added a comment.

LDj3SNuD (approved):

riperiperi (approved):
After a few tests, I can confirm everything appears to be working fine. 👍 Still a bit hard to actually close some games, but that's not for this PR to fix.

ARMeilleure/Translation/Translator.cs
```diff
@@ -233,22 +236,47 @@ internal static TranslatedFunction Translate(IMemoryManager memory, JumpTable ju
...
+        private static ControlFlowGraph EmitAndGetCFG(ArmEmitterContext context, Block[] blocks, out Range range)
         {
+            ulong rangeStart = ulong.MaxValue;
+            ulong rangeEnd = 0;
+
             for (int blkIndex = 0; blkIndex < blocks.Length; blkIndex++)
             {
                 Block block = blocks[blkIndex];
 
+                if (rangeStart > block.Address)
```
FICTURE7, line 270: I believe this might need a check to exclude exit blocks (i.e `block.Exit == true`), otherwise the range could be an over estimate.

LDj3SNuD:
Now `while (context.Running && (address & ~1UL) != 0);` @ `Translator.Execute()` @ `ARMeilleure\Translation\Translator.cs` should be updated.

ARMeilleure/Translation/Translator.cs
gdkchan: Not sure if I get it, why is the range over estimated? Is it due to possible "gaps" on the function?

FICTURE7: Right. Exit blocks have addresses which are outside the contiguous guest code region and they are translated to tail continues and not IR corresponding to decoded opcodes. So they do not truly form part of the translated guest code range.

FICTURE7: Take this for example (assume `L0` is the entry point):
```asm
L0:             ; [0x000000, 0x000004)
  ; <>
  b L1
  
  .skip 4096, 0
  
L1:             ; [0x001004, 0x001008)
  ; <>
  ret
```
Block `L0` branches to block `L1`. However since `L0` and `L1` are separated by a gap, `L1` will become an exit block (`TailcallRemover` kicks in). Both `L0` and `L1` will still be in the decoded list of blocks. If we do not exclude `L1`, we will include the gap in the range although we will only generate a tail continue for it.

gdkchan: Thanks for the explanation, should be fixed now.

Thog:
Need rebase. Also any reasons for this to not being merged yet? @riperiperi & @LDj3SNuD

LDj3SNuD:
> Any reasons for this to not being merged yet?

Now that IPC1 has been merged, I see no other particular reason.

However, I have to propose a change to JitTable to avoid reserving new regions for unmenaged tables at every title launch; entries committed from previous titles will be reused, saving memory. (Tested, with/out PPTC.)
[JumpTable.cs.txt](https://github.com/Ryujinx/Ryujinx/files/5611626/JumpTable.cs.txt)

AcK77:
Needs a rebase

gdkchan:
Rebased and added code to dispose the jump table (that was not being done before).

PR #1482, opened 2020-08-18
Fix: Issue #1475 Texture Compatibility Check methods need to be centralized

saldabain:
Closes #1475 

This is a first-time pull request to get acclimated with the project.

Downloaded a code-aligner to undo the spacing changes done by visual studio.

PR #1521, opened 2020-09-01
Fix regression on texture compatibility match checks

gdkchan:
Fixes a regression introduced on #1482 that caused height maps to break in SMO after a few steps (and probably affected other games too), the wrong `SizeMatches` overload was being used, compared to the previous code.


jduncanator (approved):
:+1: Looks like this also fixes a regression in AC:NH.

PR #1522, opened 2020-09-01
SIMD&FP load/store with scale > 4 should be undefined

gdkchan:
Per the manual:
```
if scale > 4 then UNDEFINED; 
```
This was not being enforced, which lead to a exception later on, during emit stage since the size could be invalid. Now it should produce a "undefined" instruction instead for those invalid cases.

~~PPTC increment is not required, since when the invalid encoding is encountered, it throws a exception later on and no code is generated (before this change).~~ Nevermind that, incremented it since I found another case of undefined encoding that was also not being matched:
```
if option<1> == '0' then UNDEFINED;    // sub-word index
```
This one is also fixed now.

### DIFF
--- a/ARMeilleure/Decoders/OpCodeSimdMemImm.cs
+++ b/ARMeilleure/Decoders/OpCodeSimdMemImm.cs
@@ -6,7 +6,16 @@ namespace ARMeilleure.Decoders
         {
             Size |= (opCode >> 21) & 4;
 
-            if (!WBack && !Unscaled && Size >= 4)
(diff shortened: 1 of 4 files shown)


riperiperi (approved):
Pretty straightforward. I'm guessing this might be hit if a function translates out of bounds?

gdkchan:
Yes, it might happen if it attempts to translate "garbage".

LDj3SNuD (approved):

PR #1528, opened 2020-09-03
Fix host stack overflow caused by some recursive guest methods.

LDj3SNuD:
Affected titles (now ingame):
https://github.com/Ryujinx/Ryujinx-Games-List/issues/185
https://github.com/Ryujinx/Ryujinx-Games-List/issues/162 (outdated)
and maybe others


gdkchan (approved):
lgtm, very simple fix. Also nice catch.
It is worth noting that this does not fix all cases of recursion. Maybe it's worth adding a comment on the code explaining the purpose of the check.
Might be better to also do the check on `EmitCall` instead, to allow it to also work for Arm32.

riperiperi:
Returning here will likely clear the stack, and cause all functions that were on it to partially re-translate from the point where they return to. Does doing a tail call also fix the problem? If so, that would be preferred.

gdkchan:
> Anyways, why does the game do this? If it were really doing recursive calls here, wouldn't it also stack overflow on real hardware?
That depends. The stack size is likely different on real hardware. Plus the stack usage by the function is probably going to be much lower compared to a emulator. The emulator has worse code quality, and higher stack usage as consequence (due to spills and so on).

FICTURE7:
We could also turn it into a branch operation in HCQ since we do not need to count for rejit anymore.

riperiperi (approved):
Tested a few games and things are looking good to me.

PR #1530, opened 2020-09-03
Move Logs into Config directory

qurious-pixel:
Allow logs to be written in config folder, leaving Publish folder unchanged.

### DIFF
--- a/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
+++ b/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Common.Logging
         public FileLogTarget(string path, string name, FileShare fileShare, FileMode fileMode)
         {
             // Ensure directory is present
-            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(path, "Logs"));
+            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Ryujinx/Logs"));
             logDir.Create();
 
             // Clean up old logs, should only keep 3
--- a/Ryujinx/Ui/SettingsWindow.cs
+++ b/Ryujinx/Ui/SettingsWindow.cs
@@ -480,7 +480,7 @@ namespace Ryujinx.Ui
 
         private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
-            string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
+            string logPath = System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Logs");
             
             DirectoryInfo directory = new DirectoryInfo(logPath);
             directory.Create();
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -62,6 +62,7 @@ namespace Ryujinx.Common.Configuration
             Directory.CreateDirectory(GamesDirPath = Path.Combine(BaseDirPath, GamesDir));
             Directory.CreateDirectory(ProfilesDirPath = Path.Combine(BaseDirPath, ProfilesDir));
             Directory.CreateDirectory(KeysDirPath = Path.Combine(BaseDirPath, KeysDir));
+            Directory.CreateDirectory(logDirPath = Path.Combine(BaseDirPath, logDir));            
         }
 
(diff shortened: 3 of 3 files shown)


Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -22,7 +22,7 @@ public FileLogTarget(string path, string name)
         public FileLogTarget(string path, string name, FileShare fileShare, FileMode fileMode)
         {
             // Ensure directory is present
-            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(path, "Logs"));
+            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Ryujinx/Logs"));
```
Thog, line 25: Starting with b9398f1f3a13612a0bc3e64205cdb7eca3f48d4d, Ryujinx support custom data directory.
As such you need to modify [AppDataManager](https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Common/Configuration/AppDataManager.cs) to add an util here that use the BasePath.

Ryujinx/Ui/SettingsWindow.cs
```diff
@@ -480,7 +480,7 @@ private void BrowseThemeDir_Pressed(object sender, EventArgs args)
 
         private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
-            string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
+            string logPath = System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "Logs");
```
Thog, line 483: Starting with b9398f1f3a13612a0bc3e64205cdb7eca3f48d4d, Ryujinx support custom data directory.
As such you need to modify [AppDataManager](https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Common/Configuration/AppDataManager.cs) to add an util here that use the BasePath.

Thog (changes requested):
Added some comments that need to be addressed.

I'm also against moving logs to this kind of location due to the idea of it not being tied to the build you are using.

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
qurious-pixel: I did try `./Ryujinx -r <my-data-dir>` but it fails if the directory does not exist and falls back to the original base dir. This would crash the emulator if the fallback tries again to write to write-protected area. 
maybe a non-destructive way to create the directory if doesn't exist.

Also, the first line of the logs indicate the version. 
Maybe you are saying if the user is running multiple versions of Ryujinx?

jduncanator:
I’m fairly against this change. The config file is usually located in AppData, and we already have issues getting users to locate the logs when they’re in the same folder as Ryujinx itself, let alone some hidden folder that’s difficult to find on Windows. The config folder is also shared across multiple different Ryujinx versions.

My preference here would be to leave the default location, but to add a config option that allows you to change the log folder location. If you want to move logs out of the publish folder, you can change your config and put them somewhere else.

qurious-pixel:
I understand the hesitancy to change the location to AppData.
I like how Ryujinx has the open Logs Folder button.
I would like to have a way to set the change location BEFORE launching the App.

If changing the DataDir folder, should the logsDir also change, or should the user also select this?

gdkchan:
Sounds like just adding another command line parameter to allow setting a custom log output folder would solve the issue.

jduncanator:
> I would like to have a way to set the change location BEFORE launching the App.

You can always tweak the Config.json to point to the correct location, before launching Ryujinx.

qurious-pixel:
I can see log on/off switches, but i wasn't able to find where to set the path for logs. I had turned off the logs, originally, but decided it was more important to keep them on, in a writeable directory.

jduncanator:
> I can see log on/off switches, but i wasn't able to find where to set the path for logs. I had turned off the logs, originally, but decided it was more important to keep them on, in a writeable directory.

The feature doesn't exist yet. I meant, if we were to add the option to configure the logs directory via settings, you could always tweak the directory in the config file before launching Ryujinx for the first time.

Shahil-Ayato:
@qurious-pixel are you still interested in this Pull Request?

qurious-pixel:
> @qurious-pixel are you still interested in this Pull Request?

It is necessary for any read-only system like AppImage or Flatpak, but as of now, I am making the change during build. This PR can be dropped.

AcK77:
Superseeded by #2027

PR #1531, opened 2020-09-03
Add support for loading ZRA compressed NCAs

Xpl0itR:
This PR adds support for loading NCA files that have been compressed using the [ZStandard Random Access (ZRA)](https://github.com/zraorg/ZRA) compression format. The [ZRA.NET](https://github.com/zraorg/ZRA.NET) library is used for decompression.

| Original | Compressed |
|-|-|
| NCA | ZCA |
| NSP | ZSP |
| XCI | ZCI |

- ZCA is a plaintext NCA compressed using ZRA. It also contains a ZCA Header which is located in the ZRA Metadata Section. Parsing this header is not *required* for decompression/re-encryption and can be ignored in most cases, however compressors should still generate this header.
- ZSP/ZCI files are functionally identical to NSP/XCI files, but have been renamed to signify they contain ZCA files.

#### ZCA Header
| Offset | Size | Description |
|-|-|-|
| 0x0 | 0x4 | "ZCA0" Magic (ASCII) |
| 0x4 | 0x1 | Section Count |
| 0x5 | 0x19 * Section Count | [Section Info](#section-info) Array |

##### Section Info
| Offset | Size | Description |
|-|-|-|
| 0x0 | 0x8 | Offset (Int64) |
| 0x8 | 0x8 | Size (Int64) |
| 0x10 | 0x1 | Encryption Type |
| 0x11 | 0x8 | AES Counter (UInt64) |

### Notes
- The ZRA compression format isn't final, it should also be reviewed
- Before merging leave a comment so we can increment version numbers in both ZRA and ZRA.NET
- In-emulator compression will be added soon in a later PR
- Tool for compressing/decompression can be found [here](https://github.com/Xpl0itR/ZcaTool) (bare in mind that this tool was hastily written)

***
#### How is this done?
ZSTD has the concept of a [Frame](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#frames) which can be decompressed independently from the rest of the file. A ZSTD archive is made of multiple concatenated frames which are decompressed one after another.  

We exploit that fact to break the file into uniformly sized frames (`Frame Size`) and creating a seek-table which contains the offset of each frame within in the file which can be indexed by simply dividing the offset by the frame size.  

#### Header
We store data that's required for decompression or other functionality inside an archive header, that contains the following:  
* **ZSTD Skippable Frame** - The entire header is inside a ZSTD Skippable Frame so that ZRA is fully compatible with any regular ZSTD decompressor
* **CRC-32 Hash** - A CRC-32 hash of the entire header to ensure integrity of the file is always preserved 
* **Metadata Section** - A section where data which might be used by a ZRA decompressor on the other side but not a part of the archive's contents itself
* **Seek-Table** - A table with 40-bit entries containing the offset of individual frames


Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -238,11 +243,15 @@ private void LoadNca(Nca mainNca, Nca patchNca, Nca controlNca)
 
                     _fileSystem.ImportTickets(nsp);
 
-                    foreach (DirectoryEntryEx fileEntry in nsp.EnumerateEntries("/", "*.nca"))
+                    foreach (DirectoryEntryEx fileEntry in nsp.EnumerateEntries("/", "*.*ca"))
```
bylaws, line 246: Is using a wildcard here the best idea? A '.rca' would still get treated as a normal nca.

Ryujinx/Ui/ApplicationLibrary.cs
```diff
@@ -630,13 +667,17 @@ private static bool IsUpdateApplied(string titleId, out string version)
 
                     _virtualFileSystem.ImportTickets(nsp);
 
-                    foreach (DirectoryEntryEx fileEntry in nsp.EnumerateEntries("/", "*.nca"))
+                    foreach (DirectoryEntryEx fileEntry in nsp.EnumerateEntries("/", "*.*ca"))
```
bylaws, line 670: Same as above

nicoboss:
ZRA sounds very interesting. What are the advantages over the block compressed NSZ/XCZ files that also offer random read access and even uses the same zStandard compression algorithm? Block compressed NSZ/XCZ even allows the user to trade of between compression ratio and access time by specifying the block size and stores uncompressible blocks in plain text. I don't see why we should introduce another file format if there is already a well-established by multiple tools and homebrew supported file format for this exact purpose. If you decide to use block compressed NSZ instead take a look at https://github.com/nicoboss/nsz/blob/master/nsz/BlockDecompressorReader.py and https://github.com/nicoboss/nsz/blob/master/nsz/IndependentNczDecompressor.py - it’s really easy to implement. It also has the advantage that you would only have to make a decompressor as my NSZ tool can already compress them and even offers great user experience as it contains booth a CLI and GUI. It’s 392 commits and a year of development ahead of every new file format. If there are any issues with the block compressed NSZ file format just let me know as I’m willing to modify it in order to fit your needs. Also feel free to contact me if you have any questions how it works and how to implement it. If you really want to stick with your ZRA file format I will replace my block compressed NSZ format with it but it's really annoying that we now switched from NSPZ to NSZ to ZRA all offering the same exact thing with the same compression algorithm.

nicoboss:
How does your file format decrypt the NCA files before compression and is a conversion from ZRA back to the original NCA possible?

NSZP decrypted everything including the NCA header without storing any clues how to re-encrypt it as everything needed can be derived from the Ticket and NCA header. This turned out to be too complicated for most developers. NSZ was made with simplicity in mind and is basically a heavily simplified version of NSPZ. In NSZ the first 0x4000 bytes of every NCA are copied over (independently where the first sector starts). There a header is inserted containing information how to re-encrypt the following data. Afterwards there is a seek-table required for mapping read access from the original offsets to the proper block allowing random read access on compressed data. The remaining of the file are just zStandard compressed blocks of decrypted data mixed with some plain text blocks of uncompressible data on where compression would increase the file size. By default, zStandard level 18 and a block size of 1 MB is used however the user can easily change this. Multithreaded compression and decompression are supported as all blocks are independent. Compared to solid compression only 2 – 6% of compression ratio is lost.

emmauss:
ZRA is a compression format. there is no tampering with the compressed data. Decompression will produce data that's bit-identical to the original

Xpl0itR:
> How does your file format decrypt the NCA files before compression and is a conversion from ZRA back to the original NCA possible?

LibHac is used to convert an NCA to plaintext before compression and used to convert the plaintext back into a bit-identical encrypted NCA after full decompression.

bylaws:
Even without libhac decrypting an NCA is super simple with a tiny bit of parsing taking into account the header and section offsets, encryption is just the same

nicoboss:
That's really nice. ZRA didn't exist by the time NSZP or NSZ got developed and so was never considered. Awesome we now have --plaintext and --ciphertext in LibHac. Just tested them out and they indeed decrypt/encrypt the whole NCA file bit-identical. I might add support for this file format to my NSZ tool if this makes it into Ryujinx/Yuzu and gains some popularity however I will also keep block compressed NSZ/XCZ as it’s already widely used and I can just make them mountable as NSP so they are compatible with any application only supporting NSP. Unlike for ZStandard there is no Python binding for ZRA so I would have to write my own first.
It would be nice if you could add offset, size, cryptoType, cryptoKey and cryptoCounter for every section into the ZRA "Metadata Section" so that information doesn't need to be phrased from the NCA Header and Ticket so it's easier for developers to switch from NSZ to ZRA. But I agree that even without that information it wouldn't be hard to obtain them otherwise.

blawar:
> Hi,
> Was discussing with @pixelylon (the ZRA author) and he wondered whether there was reason for NSZ using the ZSTD block level API over the frame level API? Frames represent an atomic unit of decompressable data and their block contents is automatically chosen to best fit the data being compressed by ZSTD. With the block level API multiple blocks may have to be read to decompress just a single block based on the compressed data, this uncertainty is not exactly great for emulators. Since each frame is individual the frame size set in ZRA ends up being the exact amount of data that is read to access any part of a file.
> 
> To help with emulation and large block sizes we plan on implementing a frame cache some time, normally a large block size would result in the performance of sequential access degrading from decompressing the same frame multiple times but a frame cache would mitigate that by caching the last frames used in ram for quick access.

block level is simpler, easier to implement, and allows the format to support mixed compression algo's (it currently supports zstd and plaintext).  Frame level could have been added to this by adding more complexity to the format, however there is no benefit (as outlined below) and it hurts simplicity.

frame level likely has some advantages were small chunk sizes and resource constrained devices like embedded microcontrollers, however the target platform of NSZ is PC's and a nintendo switch which would show much better performance with large chunk sizes, therefore any inherent performance differences should be negligible.

wrt caching, its basically the same whether you cache a block or a frame, though I would discourage both.   I would recommend you cache the deflated data as the target platforms are not memory bound and you would be sacrificing CPU.

bylaws:
> > Hi,
> > Was discussing with @pixelylon (the ZRA author) and he wondered whether there was reason for NSZ using the ZSTD block level API over the frame level API? Frames represent an atomic unit of decompressable data and their block contents is automatically chosen to best fit the data being compressed by ZSTD. With the block level API multiple blocks may have to be read to decompress just a single block based on the compressed data, this uncertainty is not exactly great for emulators. Since each frame is individual the frame size set in ZRA ends up being the exact amount of data that is read to access any part of a file.
> > 
> > To help with emulation and large block sizes we plan on implementing a frame cache some time, normally a large block size would result in the performance of sequential access degrading from decompressing the same frame multiple times but a frame cache would mitigate that by caching the last frames used in ram for quick access.
> 
> block level is simpler, easier to implement, and allows the format to support mixed compression algo's (it currently supports zstd and plaintext).  Frame level could have been added to this by adding more complexity to the format, however there is no benefit (as outlined below) and it hurts simplicity.
> 
> frame level likely has some advantages were small chunk sizes and resource constrained devices like embedded microcontrollers, however the target platform of NSZ is PC's and a nintendo switch which would show much better performance with large chunk sizes, therefore any inherent performance differences should be negligible.
> 
> wrt caching, its basically the same whether you cache a block or a frame, though I would discourage both.   I would recommend you cache the deflated data as the target platforms are not memory bound and you would be sacrificing CPU.
> 
> edit: And to go back to the simplicity thing:  the best format in the world is useless if it is too complex for others to implement.  ZRA is too complex to be widely used outside of this project, because it leverages libhac to parse and do all of the heavy lifting.  libhac is c#, and no other such library exists for any other language.  NSZ does not have this issue and can be easily used across languages and applications.
> 
> That helper table in the NSZ header (that I have noticed most people do not understand) is indeed redundant, all of its info can be  derived from the NCA itself.  It exists to spare developers who want to use the format from having to re-implement half of FS to restore the original NCA.  All of the info needed to restore it is located in that redundant header.  This header is of no use for emulators and can be safely ignored when used in emulator applications, however NSZ was designed to be easily used in as many applications as possible.  It was not intended to be an adhoc format for a specific application.

I think you misunderstand ZRA and ZCA, the core library implemented in C++ and isn't tied to libhac at all, bindings can easily be written for any language. Having a unified core library also simplifies updates to the format greatly and reduces the chance of bugs as well as making implementation much easier (check out the C/C++ API). Decrypting the NCA and packing it into a ZCA is simple for the target audience of emulators, the VFS frameworks in Yuzu, Ryu, Skyline etc all can easily support it. ZRA itself is also designed to be multi purpose, it makes use of a metadata section to support specialisation and the core library has nothing switch specific. 

Regarding compression algorithms, ZSTD automatically decides the types of blocks that should make up a  frame based on the compressed data, check the ZSTD manual for more info on that.

Also yes, caching a frame of decompressed data was what I was referring to, anything else would be silly.

blawar:
> I think you misunderstand ZRA and ZCA, the core library implemented in C++ and isn't tied to libhac at all, bindings can easily be written for any language. Having a unified core library also simplifies updates to the format greatly and reduces the chance of bugs as well as making implementation much easier (check out the C/C++ API). Decrypting the NCA and packing it into a ZCA is simple for the target audience of emulators, the VFS frameworks in Yuzu, Ryu, Skyline etc all can easily support it. ZRA itself is also designed to be multi purpose, it makes use of a metadata section to support specialisation and the core library has nothing switch specific.
> 
> Regarding compression algorithms, ZSTD automatically decides the types of blocks that should make up a frame based on the compressed data, check the ZSTD manual for more info on that.
> 
> Also yes, caching a frame of decompressed data was what I was referring to, anything else would be silly.

Greetings @bylaws ,

I do not believe there is a misunderstanding, allow me to elaborate.  I understand that ZCA simply decompresses to a plain-text NCA that ryujinx then loads as is--that's it for loading.  However, this fails to take into account non-ryujinx applications wanting to use the format who do need to convert it to the original NCA format.

ZRA appears to be very adhoc for ryujinx's needs, whereas NSZ was designed to be more general purpose.  Overall, it is better for the scene to have a single unified format imo.  NSZ already has an established set of tools as well.

bylaws:
> 
> Greetings @bylaws ,
> 
> I do not believe there is a misunderstanding, allow me to elaborate.  I understand that ZCA simply decompresses to a plain-text NCA that ryujinx then loads as is--that's it for loading.  However, this fails to take into account non-ryujinx applications wanting to use the format who do need to convert it to the original NCA format.
> 
> ZRA appears to be very adhoc for ryujinx's needs, whereas NSZ was designed to be more general purpose.  Overall, it is better for the scene to have a single unified format imo.  NSZ already has an established set of tools as well.

Why exactly is this ad-hoc for Ryujinx? Surely it makes sense for a compression format to spit out exactly what was fed into it. Unless you want forced re-encryption or something which would be kinda silly and would just waste pointless CPU time when everything that deals with encrypted ncas can be accept decrypted ncas with like 2 simple code changes. What applications do you speak of where this would be difficult? I am fairly certain this won't be an issue at all for any emulators.

blawar:
> Why exactly is this ad-hoc for Ryujinx? Surely it makes sense for a compression format to spit out exactly what was fed into it. Unless you want forced re-encryption or something which would be kinda silly and would just waste pointless CPU time when everything that deals with encrypted ncas can be accept decrypted ncas with like 2 simple code changes. What applications do you speak of where this would be difficult? I am fairly certain this won't be an issue at all for any emulators.

Because ZCA only supports easy decompression to a dirty format (plaintext) whose only use case (that I am aware of) is emulators.  Anything else (which is every other application I can think of) requires more, that cannot be done easily with ZCA.  

I am not advocating for forced re-encryption, NSZ works just fine working with the decompressed plaintext.  I am just saying there needs to be an easy way to re-encrypt for other applications that require such functionality.

The Ryujinx fork I made with NSZ support was accomplished by creating a custom stream class for NCZ, spoofing the plaintext byte in the NCA header to be set, then returning plaintext in the read through the implemented Read() function.  It completely ignores the encryption header in the NSZ header because it is not needed for emulators use case, but other applications  *do* use it.

bylaws:
> 
> Because ZCA only supports easy decompression to a dirty format (plaintext) whose only use case (that I am aware of) is emulators.  Anything else (which is every other application I can think of) requires more, that cannot be done easily with ZCA.  
What applications are these? All I can think of is hactool, nstool etc and they can all easily change a few lines to just not expect an encrypted NCA. I'm fairly sure hactool already has support even.
And re-encryption isn't difficult, it's just a case of reading in the header, caching it, then encrypting the header/content with the values for the cached header,if we take into account Nico's suggestion of putting re-encryption info in the metadata section it becomes even easier.
> 
> I am not advocating for forced re-encryption, NSZ works just fine working with the decompressed plaintext.  I am just saying there needs to be an easy way to re-encrypt for other applications that require such functionality.
See above, re-encryption isn't hard, and what apps need the encrypted NCA anyhow?
> 
> The Ryujinx fork I made with NSZ support was accomplished by creating a custom stream class for NCZ, spoofing the plaintext byte in the NCA header to be set, then returning plaintext in the read through the implemented Read() function.  It completely ignores the encryption header in the NSZ header because it is not needed for emulators use case, but other applications  *do* use it.
Such as?
> 
>

blawar:
> What applications are these? All I can think of is hactool, nstool etc and they can all easily change a few lines to just not expect an encrypted NCA. I'm fairly sure hactool already has support even.
And re-encryption isn't difficult, it's just a case of reading in the header, caching it, then encrypting the header/content with the values for the cached header,if we take into account Nico's suggestion of putting re-encryption info in the metadata section it becomes even easier.

It is no where near as simple to re-encrypt.  And it is *definitely* not a matter of caching some headers--you have to fully parse the NCA which is non-trivial.  I have personally wrote libraries to do this stuff twice in two different languages.  The only reason ZCA appears to be simple is because it leverages libhac which does all of the heavy lifting (it is a great library).  Unfortunately no one I can think of uses C# except ryujinx and switch backup manager.  Anyone else trying to implement ZSP outside of C# and emulators is going to have a bad time.

> what apps need the encrypted NCA anyhow?

Applications that manage and manipulate NSP like files.  Applications that convert between the formats, verify files, or extract metadata from the files, or want to install unmodified files.

Some examples include:

Nut, SAK, XCI Explorer, NSCB, Nicoboss's NSZ tool, quite a few private tools which cannot be mentioned here, Switch Backup Manager (possibly, not sure what all it supports), and every homebrew app on the switch that interacts with NSP files.

bylaws:
> And re-encryption isn't difficult, it's just a case of reading in the header, caching it, then encrypting the header/content with the values for the cached header,if we take into account Nico's suggestion of putting re-encryption info in the metadata section it becomes even easier.
> 
> It is no where near as simple to re-encrypt.  And it is *definitely* not a matter of caching some headers--you have to fully parse the NCA which is non-trivial.  I have personally wrote libraries to do this stuff twice in two different languages.  The only reason ZCA appears to be simple is because it leverages libhac which does all of the heavy lifting (it is a great library).  Unfortunately no one I can think of uses C# except ryujinx and switch backup manager.  Anyone else trying to implement ZSP outside of C# and emulators is going to have a bad time.

Having written the VFS for Skyline (c++) I fail to see how parsing the NCA header is really that hard, there isn't too much that needs to be done. And yeah, I fail to see how this will be an issue at all if we include extra info on the metadata section as suggested by Nico which would make it just as simple as NSZ or whatever.

> 
> > what apps need the encrypted NCA anyhow?
> 
> Applications that manage and manipulate NSP like files.  Applications that convert between the formats, verify files, or extract metadata from the files, or want to install unmodified files.


Extracting metadata and converting between formats can both be done simply without re-encryption, for verification you are correct though - this shouldn't be an issue as re-encryption isn't that difficult.

Title installation is entirely out of the scope of the format, we cater to emulators, game RE tools, etc: not glorified piracy tools.

> 
> Some examples include:
> 
> Nut, SAK, XCI Explorer, NSCB, Nicoboss's NSZ tool, quite a few private tools which cannot be mentioned here, Switch Backup Manager (possibly, not sure what all it supports), and every homebrew app on the switch that interacts with NSP files

SAK and XCI Explorer could both easily be modified to support decrypted files from a brief look at them.  The others I suppose would all need re-encryption, but as previously stated that isn't too hard, especially with Nico's suggestions.

blawar:
> Having written the VFS for Skyline (c++) I fail to see how parsing the NCA header is really that hard

I never said it was hard, just non-trivial.  It is too much to ask third party developers to parse those nintendo formats when simply adding the info to the meta would suffice.

No one creates plaintext NSP's, therefore converting between the formats requires re-encryption.  No one uses plain text except ZCA.

I understand that title installer's are outside of the scope of this project, and that was my point about this format being an adhoc format disregarding the use of other applications.

You are right, with NICO's suggestion it is easy.  But at this point, you basically have a blocked NSZ format, and the formats were fragmented the scene for no valid reason other than to be different.

bylaws:
> > Having written the VFS for Skyline (c++) I fail to see how parsing the NCA header is really that hard
> 
> I never said it was hard, just non-trivial.  It is too much to ask third party developers to parse those nintendo formats when simply adding the info to the meta would suffice.

I've mentioned that we'll probably be doing this.

> 
> No one creates plaintext NSP's, therefore converting between the formats requires re-encryption.  No one uses plain text except ZCA.
> 
> I understand that title installer's are outside of the scope of this project, and that was my point about this format being an adhoc format disregarding the use of other applications.
> 
> You are right, with NICO's suggestion it is easy.  But at this point, you basically have a blocked NSZ format, and the formats were fragmented the scene for no valid reason other than to be different.

Not really, see mine and xploiter's earlier posts about the benefit this format provides over nsz.

Feel free to ask mark for any details about the exact design of the format too.

blawar:
> Not really, see mine and xploiter's earlier posts about the benefit this format provides over nsz.

I already negated those benefits, to my earlier reply.  I know you think frame level is advantageous over block level, however it is not.  I've personally done a lot of testing and benchmarking on this.  There is no difference when the chunk size is large enough, and block is simpler to implement.

emmauss:
Please, move any discussion to the discord server. This pr is for reviewing the related feature for ryujinx.

bylaws:
I personally still think it's worth doing format review here, it helps to index all the different comments on possible changes to the format etc

jduncanator:
Why was all the discussion surrounding the merits of each format deleted? To me that seems incredibly on-topic to this discussion, as the merits of each format should be discussed before committing to a specific format.

Ryujinx.HLE/HOS/ApplicationLoader.cs
jduncanator: You can use `nsp.EnumerateEntries("/", "*.nca").Concat(nsp.EnumerateEntries("/", "*.zca"))` instead, preferably broken out onto another line.

user18081972:
> Why was all the discussion surrounding the merits of each format deleted? To me that seems incredibly on-topic to this discussion, as the merits of each format should be discussed before committing to a specific format.

@jduncanator The truth is as follow:
- marysaka is transgender. 
- blawar started the NSZ compression format development. 
- blawar has outed criticism on the LGBT "community".
- marysaka now refuses to support a compression format that is widely adopted and would make Ryujinx a better application. 

They claim the reason for not supporting NSZ is as follows:
image

Which is obviously not true if you take all of the above into account.

GamerzHell9137:
> > Why was all the discussion surrounding the merits of each format deleted? To me that seems incredibly on-topic to this discussion, as the merits of each format should be discussed before committing to a specific format.
> 
> @jduncanator The truth is as follow:
> 
>     * marysaka is transgender.
> 
>     * blawar started the NSZ compression format development.
> 
>     * blawar has outed criticism on the LGBT "community".
> 
>     * marysaka now refuses to support a compression format that is widely adopted and would make Ryujinx a better application.
> 
> 
> They claim the reason for not supporting NSZ is as follows: image
> 
> Which is obviously just an excuse if you take all of the above into account. just replace "pirates" with "transphobes" and the above image shows what marysaka really meant.

Maam, this is a wendy's.

marysaka:
@user18081972

1. no emulator currently support that format.
2. no one in the hacking scene support it apart from pirates.

A compressed format could be doable for archiving purpose, but we aren't going to support a format that is so far not the norm among our peers.

gdkchan:
NCAs already have compression support, so it seems better to just use that to compress the data on games that are not natively compressed rather than developing a new format for compression.

PR #1535, opened 2020-09-05
Relax block ordering constraints

FICTURE7:
Before `block.Next` had to follow `block.ListNext`, now it does not. `block.Next` and `block.Branch` could also be a bit ambiguous when there was a single successor, now they are not.

Instead `CodeGenerator` will now emit the jump instructions when necessary to ensure control flow. This makes control flow and block order modifications easier (see `ControlFlowGraph.SplitEdge(pred, succ)` & `LinearScanAllocator.FirstSuccessor(bb)` for examples). It also eliminates some simple case of redundant unconditional branches.

```patch
@@ -30,8 +30,6 @@
   sub ecx,byte +0x1
   mov [rax+0x400],ecx
 .L4:
-  jmp short .L5
-.L5:
   test rbp,rbp
   setl cl
   movzx ecx,cl
@@ -46,8 +44,8 @@
```

This PR removes the `Branch` instruction. An alternative approach would be to canonicalize flow by using the `Branch` instruction. but I figure its one less operation to traverse, so I removed it. Let me know if you think the alternative solution is better or if there are other better alternatives.

This also improves the liveness analysis done in the LSRA a little bit, when visiting the successors, it was short circuiting early so the result set took a couple of more iterations than usual to stabilize.

### DIFF
--- a/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
@@ -427,18 +427,5 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
         {
             return local.Assignments.Count + local.Uses.Count;
         }
-
-        private static IEnumerable<BasicBlock> Successors(BasicBlock block)
-        {
-            if (block.Next != null)
-            {
-                yield return block.Next;
-            }
-
-            if (block.Branch != null)
-            {
-                yield return block.Branch;
-            }
-        }
     }
 }
--- a/ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
@@ -626,16 +626,11 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
                     continue;
                 }
 
-                bool hasSingleOrNoSuccessor = block.Next == null || block.Branch == null;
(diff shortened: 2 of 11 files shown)


riperiperi:
Nice, code looks good and it's appreciably cleaner.

Just to check JIT performance, I reran a PTC compile for MK8 1.7 with 41920 functions, looks like it took slightly longer to complete, but within margin of error.

This pr:
00:01:49.558 |I| HLE.HostThread.0 Gpu : Completed PTC in 105905ms
Master:
00:01:45.165 |I| HLE.HostThread.0 Gpu : Completed PTC in 102107ms

This cache is a bit of an outlier though, and my machine's thread count also makes it more susceptible to heap contention or whatever is slowing the codegen/lsra down.

riperiperi:
Oh yeah, if anyone wants to test themselves for this or future prs, you can easily just change PTC version then add a stopwatch + log around `Ptc.MakeAndSaveTranslations`. Might even be useful enough to PR in future.

riperiperi (approved):

FICTURE7:
Interesting. I suspect it is because accessing the successors is slightly more expensive, we pay for a couple of bound checks and an extra indirection. Before these were inlined in the structure of `BasicBlock` and the getters of `Next`/`Branch` were probably inlined at the caller site as well.

gdkchan (approved):
lgtm, thanks!
The `Next` and `Branch` blocks are indeed a nuisance (and can be confusing with `Next` and `ListNext`), so glad to see this changed.
FWIW the same improvement can also be applied to the shader translator.

PR #1537, opened 2020-09-05
Do not emit StoreToContext before Return

FICTURE7:
`RegisterUsage` pass considers `Return` as a `StoreToContext`, so we do not need to emit one before a `Return` instruction. Sample [diff](https://www.diffchecker.com/XlZey6rj).


gdkchan (approved):
lgtm, thanks.

jduncanator (approved):
Love simple improvements like this. Thanks for your work! ☺️

riperiperi (approved):
Nice one, lgtm.

PR #1538, opened 2020-09-06
Allow swizzles to match with "undefined" components

riperiperi:
image

Improves rules which try to match formats as identical (not view), such as D32 float -> R32 float. Fixes the Depth of Field and Motion Blur in Burnout Paradise, as a view cannot be created in this case (before depth was just being sampled as a solid colour).

The FormatInfo struct has been updated with a "Components" field, which indicates the maximum number of components that can be "defined" by that format, in RGBA order (2 components defines RG, 3 defines RGB, etc). I'd recommend taking a close examination of this table to see if I got anything wrong. Right now, all depth formats that aren't D32 float are set to 4 just to be safe, which I'm not sure is totally accurate.

The rules mostly focus around comparison of SwizzleR, SwizzleG, SwizzleB, SwizzleA values. If a value is above the number of components (eg. `SwizzleComponent.Green` on a 1 component texture) that value is considered "undefined". 
- When two values are undefined, we consider them a match regardless of value. 
- When two values are defined, they are a match if exactly equal.
- When one is defined and the other not, they are a match if equal, if the defined value is 0 or 1, or if the defined value equals the default for that particular component (eg. SwizzleComponent.Green is the "default" for the 2nd component).

Should ideally test in a few games to see if there are any regressions, but it looks fine here.

Also removes Format.HasOneComponent, since this information is now available via the FormatInfo struct.

(for an example of the motion blur before, see https://youtu.be/PC_dvogy-S4?t=162 )


gdkchan:
I'm not sure about the diverging amount of components for depth and stencil formats. If you want to know how it is setup by NVN, and what it does on the GPU, see this comment: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Image/TexturePool.cs#L224

Logically speaking, depth-only and stencil-only formats would have 1 component, while depth-stencil formats would have 2. For the purposes of the comparison, it should take all 4 into account for any depth-stencil formats, as it might change the depth stencil mode used. Btw, swizzle is currently ignored for depth-stencil formats, it is only used to guess the depth-stencil mode (see the code that I linked).

riperiperi:
> For the purposes of the comparison, it should take all 4 into account for any depth-stencil formats, as it might change the depth stencil mode used.

Would you prefer that the Components for these formats is 2 in the table, but there is a special case extending DS formats to 4 just for the comparison? That's kind of why I used 4 as the default for the other depth related formats.

I can definitely change depth only and stencil only to be 1 component.

ReinUsesLisp:
OpenGL supports swizzles on depth/stencil textures, maybe it can be tested with a nouveau homebrew.

gdkchan:
> Would you prefer that the Components for these formats is 2 in the table, but there is a special case extending DS formats to 4 just for the comparison? That's kind of why I used 4 as the default for the other depth related formats.

I would probably either use 4 for all depth/stencil formats, or use 2 for depth-stencil, and 1 for depth only and stencil only. Right now it's not consistent as depth32f has 4 but thats not the case for the others.

gdkchan:
> OpenGL supports swizzles on depth/stencil textures, maybe it can be tested with a nouveau homebrew.

Is that documented anywhere? Last time I checked, the results were not consistent depeding on the driver or debug being enabled or not, which makes me think it is undefined. Also, NVN does not support it so I don't expect games to normally do something like that.

ReinUsesLisp:
From [ARB_texture_swizzle](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_swizzle.txt):
>    4) How does this interact with depth component textures?
>
>    RESOLVED: The swizzle is applied after the DEPTH_TEXTURE_MODE. This 
>    naturally falls out of specifying the swizzle in terms of Table 3.20.

The inconsistency is in `DEPTH_TEXTURE_MODE`'s default value. Some OpenGL profiles default to `RED`, others to `LUMINANCE`. If `DEPTH_STENCIL_TEXTURE_MODE` is specified (not to be confused with `DEPTH_TEXTURE_MODE`), and the swizzle is managed accordingly, it should be have as expected.

riperiperi:
> I would probably either use 4 for all depth/stencil formats, or use 2 for depth-stencil, and 1 for depth only and stencil only. Right now it's not consistent as depth32f has 4 but thats not the case for the others.

Depth32f has 1 component, that's one of the basis of this fix. The swizzle for the depth texture is RGBA, and the swizzle for the R32f texture is R001. If all components of the d32f texture were "defined", then it would just do exact matching.

I'll just use 2 for depth+stencil. I guess B and A shouldn't be used for these textures anyways.

gdkchan (approved):
lgtm, thanks.

jduncanator (approved):

MelonSpeedruns:
This PR broke Breath of the Wild, causing it to display a white image at all times.
image

PR #1549, opened 2020-09-15
Implement block placement

FICTURE7:
Implement a simple pass which re-orders cold blocks at the end of the list of blocks in the CFG. Sample [diff](https://www.diffchecker.com/tPgkKPTd).

This removes some branching along the fast path for memory accesses and moves `CheckSynchronization` away from loops. It is not clear if we should attempt to re-order the decoded stuff, since presumably the guest compiler should already have placed it in an "optimal" order.

### DIFF
--- a/ARMeilleure/Translation/ControlFlowGraph.cs
+++ b/ARMeilleure/Translation/ControlFlowGraph.cs
@@ -9,24 +9,43 @@ namespace ARMeilleure.Translation
     {
         public BasicBlock Entry { get; }
         public IntrusiveList<BasicBlock> Blocks { get; }
-        public BasicBlock[] PostOrderBlocks { get; }
-        public int[] PostOrderMap { get; }
+        public BasicBlock[] PostOrderBlocks { get; private set; }
+        public int[] PostOrderMap { get; private set; }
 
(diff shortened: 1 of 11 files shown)


ARMeilleure/Translation/ControlFlowGraph.cs
```diff
@@ -9,24 +9,43 @@ class ControlFlowGraph
...
+                    array = new T[length];
+                }
+                
+                // No need to clear the array because all its elements will be written to anyways.
+                return array;
+            }
 
-            visited.Add(entry);
-            blockStack.Push(entry);
+            PostOrderBlocks = Initialize(PostOrderBlocks, Blocks.Count);
```
gdkchan, line 44: Maybe that could use `Array.Resize` instead (would require changing it to a field and adding properties to return the value).

FICTURE7: Oof, completely forgot about `Array.Resize`.

gdkchan (approved):
lgtm, great work.

riperiperi (approved):
Nice feature, looks good from here.

PR #1556, opened 2020-09-18
Better viewport flipping and depth mode detection method

gdkchan:
This does two change, the first one changes the method used to flip viewport, this should fix a bug that causes the screen to be flipped on some games on GPUs without viewport swizzle support (old NVIDIA GPUs, AMD and Intel). It would flip on the shader when the extension was not supported, but due to the lack of proper shader specialization, this didn't work very well.

The new approach uses `glClipControl` again to flip the viewport, this time also flipping front face to negate the effects of the viewport flipping.

---

The second one changes the method used to detect the depth mode that the game is using. Just checking the depth mode register is not very reliable as a issue similar to the one that could happen with the origin could happen here aswell. Basically, the game changing the depth mode after queue initialization would leave this register untouched, but it would still affect the viewport transform. The register only seems to control clamping, what actually defines the range is the viewport transform, so it uses that now to detect the correct depth mode.

This fixes some lighting issues on Zelda Link's Awakening, and maybe other games:
image

Special thanks to Rodrigo for hinting that ZLA could have depth range related issues, and riperiperi for finding the issue and providing a possible fix.

---

Bonus: Swap Near and Far parameters when ScaleZ is negative. The high level APIs sort them such as Near is always less than Far, so this undoes this step. This allows the cases where Near > Far to work properly, although I don't know any game that uses that to test.

I recommend testing this with a few games to be sure there are no regressions related to lighting or viewport flip and culling.

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -50,8 +50,6 @@ namespace Ryujinx.Graphics.GAL
 
         void SetLogicOpState(bool enable, LogicalOp op);
 
-        void SetOrigin(Origin origin);
-
         void SetPointParameters(float size, bool isProgramPointSize, bool enablePointSprite, Origin origin);
 
         void SetPrimitiveRestart(bool enable, int index);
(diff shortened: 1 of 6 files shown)


riperiperi:
This also fixes shadows on Burnout Paradise Remastered.

Before:
image

After:
image

I'll give the code + flipping stuff a more thorough look tomorrow.

Thog (approved):

riperiperi (approved):
Tried a few of the affected games (depth and flipping) and some that weren't, and everything looks fine. Code looks good to me as well, nice work.

PR #1558, opened 2020-09-19
Implement small indexed draws and other fixes

gdkchan:
- Implements a few registers used for indexed draws on the NVIDIA Vulkan driver (DrawIndexedSmall/IncInstance/2), names are my own, I don't know the official ones. I also don't know what the difference between the 2 variants are, they are used  the same way by the driver. The "IncInstance" variants increments the instance value before draw.
- Implement primitive topology override. This is also used by the Vulkan driver instead of setting it on GlBegin, probably due to the way how Vulkan works where the primitive topology is part of the pipeline state, instead of being a draw parameter. This is also possibly used by Direct3D. Current implementation is a guess, I'm not sure if it always overrides the other one if the value is non-zero.
- Refactor `MethodDraw` a bit to allow the new methods to be implemented without making a mess there.
- Fixed a bug where clears would re-bind sampled textures and update render targets. Clears only needs to update the render targets, as they don't sample textures.

Vulkan driver usage of the new registers:
If `firstIndex` fits (that is, the value is < 0x10000), and `indexCount` fits (value < 0x1000), it packs both in a single word and performs the "small indexed draw", otherwise it draws as usual with GlBegin/GlEnd, Note that the value passed to GlBegin is always 0, for this reason it uses the primitive type override. It will use either `DrawIndexedSmall` and `DrawIndexedSmallIncInstance`, or `DrawIndexedSmall2` and `DrawIndexedSmallIncInstance2`. There is a value on the driver internal command buffer struct that selects which, but I don't know what this value means yet.

In general those changes should help games using the Vulkan API on the guest. For example, Turok 2 now renders, and Doom 64 probably does too.


Thog (approved):
lgtm

jduncanator (approved):
This looks good to me. Thanks for the changes.

riperiperi (approved):
Looks good.

PR #1559, opened 2020-09-19
Align register index between output targets on pixel shaders

gdkchan:
This fixes a issue I observed on Turok 2 where the color components were written to the wrong render target components. Aligning the register index to 4 (the number of components/RGBA) fixes the issue.

Tested with #1558:
Before:
image
After:
image
image


Thog (approved):

riperiperi (approved):
Nice catch.

jduncanator (approved):

PR #1561, opened 2020-09-19
Update to OpenTK 4

emmauss:
This updates all projects to OpenTK 4. This brings a lot of changes to how we handle some things Opentk was handling for us in Opentk 3. 
The biggest change is the removal of all platform specific implementations for context, windowing and input. As such, GLWidget had to be updated with it's own context creation and management.
For input, keyboard input is handled by GTK, with joystick/gamepad input being handled by GLFW. Since GTK only supports one virtual keyboard device, which grabs input from all keyboards connected, we lose the ability to configure separate keyboards for seperate players. Joystick input on the other hand has been enhanced, and should support both XInput, DInput and raw joystick devices.
The current Opentk release as of the making of this PR is 4.0.0-pre9.6.


Ryujinx/Ui/GLRenderer.cs
```diff
@@ -421,10 +474,14 @@ private bool UpdateFrame()
 
                 if (inputConfig is KeyboardConfig keyboardConfig)
                 {
+                    if (keyboard.IsAnyKeyDown)
```
gdkchan, line 477: Why is that empty?

Ryujinx/Ui/GdkKeyMapping.cs
```diff
@@ -0,0 +1,213 @@
...
+{
+    public static class GdkKeyMapping
+    {
+        public static Dictionary<Gdk.Key, Key> GdkKeyMappingDictionary { get; set; }
+
+        static GdkKeyMapping()
+        {
+            GdkKeyMappingDictionary = new Dictionary<Gdk.Key, Key>()
+            {
+                {Gdk.Key.VoidSymbol, Key.Unknown },
```
gdkchan, line 17: nit: Space after `{` on all entries.

Ryujinx/Ui/GdkKeyMapping.cs
```diff
@@ -0,0 +1,213 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Ryujinx.Ui
+{
+    public static class GdkKeyMapping
+    {
+        public static Dictionary<Gdk.Key, Key> GdkKeyMappingDictionary { get; set; }
```
gdkchan, line 11: no need for the `set;`

Ryujinx/Ui/Joystick.cs
```diff
@@ -0,0 +1,171 @@
...
+        private static Dictionary<int, IntPtr> _joystickHandles;
+
+        static Joystick()
+        {
+            _joystickHandles = new Dictionary<int, IntPtr>();
+        }
+
+        public static void CloseJoysticks()
+        {
+            lock (_joystickHandles)
```
gdkchan, line 20: Useless function?

Ryujinx/Ui/Joystick.cs
```diff
@@ -0,0 +1,171 @@
...
+using SDL2;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Ryujinx.Ui
+{
+    public static class Joystick
+    {
+        private static Dictionary<int, IntPtr> _joystickHandles;
```
gdkchan, line 11: I think private static member should start in upper case aswell. `JoystickHandles`.

Ryujinx/Ui/JoystickCapabilities.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public JoystickCapabilities(int axisCount, int buttonCount, int hatCount, string name) : this()
+        {
+            AxisCount = axisCount;
+            ButtonCount = buttonCount;
+            HatCount = hatCount;
+            Name = name;
+        }
+
+        public int ButtonCount { get; set; }
```
gdkchan, line 19: Why are those at the end? Also I suspect you don't need the setters. Please remove them if the values are not modified.

Ryujinx/Ui/KeyboardController.cs
```diff
@@ -37,42 +35,57 @@ public ControllerKeys GetButtons()
...
         }
 
+        public bool CheckBounded(Configuration.Hid.Key key, KeyboardState keyboard)
+        {
+            if (key == Configuration.Hid.Key.Unbound || key == Configuration.Hid.Key.Unknown || key == Configuration.Hid.Key.LastKey)
+            {
+                return false;
+            }
+
+           var set = keyboard[(Key)key];
```
gdkchan, line 70: nit: Alignment.

Ryujinx/Ui/KeyboardController.cs
```diff
@@ -37,42 +35,57 @@ public ControllerKeys GetButtons()
...
+            if (set)
```
gdkchan, line 71: Why this is empty?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -181,6 +181,37 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
             _statusBar.Hide();
 
             _uiHandler = new GtkHostUiHandler(this);
+
+            this.KeyPressEvent += MainWindow_KeyPressEvent;
+            this.KeyReleaseEvent += MainWindow_KeyReleaseEvent;
+        }
+
```
gdkchan, line 188: Double empty line.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -181,6 +181,37 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
...
+
+        // Send key events down to the renderer in case main window has grab focus
+        [GLib.ConnectBefore]
+        private void MainWindow_KeyReleaseEvent(object o, KeyReleaseEventArgs args)
+        {
+            if(_glWidget != null)
```
gdkchan, line 194:
```suggestion
            if (_glWidget != null)
```

SeraUX:
Tested on latest ArchLinux

Running a game through the GUI crashes every time with these specific errors found
```
[xcb] Extra reply data still left in queue
[xcb] This is most likely caused by a broken X extension library
[xcb] Aborting, sorry about that.
Ryujinx: xcb_io.c:580: _XReply: Assertion `!xcb_xlib_extra_reply_data_left' failed.
Aborted (core dumped)
```
```
[xcb] Unknown sequence number while processing queue
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
Gdk-Message: 00:42:50.325: Ryujinx: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.
```
Running a game through the terminal with the cmd argument, always just gives a `Segmentation fault (core dumped)` error

emmauss:
Addressed nits. I'll look into the linux bug later

Ryujinx/Ui/GdkKeyMapping.cs
gdkchan: This one was not properly addresses, the suggestion was adding a space after the `{` instead of removing the space before `}`.

GamerzHell9137:
PS4 controller in Dinput mode is not working at all.
And when trying to set buttons it doesn't let you do it and they get forever highlighted.
image

emmauss:
> 
> 
> PS4 controller in Dinput mode is not working at all.
> And when trying to set buttons it doesn't let you do it and they get forever highlighted.
> image

make sure you connect the controller before starting ryujinx. it won't detect anything controller you plug after starting ryujinx

GamerzHell9137:
Couple of users with DS4 can confirm that DS4 controllers don't work at all.
Another thing to note is that there's no default settings for keyboard and everything is unbinded.

qurious-pixel:
DS4 is working, but setting the keybindings binds to onmousedown. Requires manually setting config, or using default config.

OpenTK4 complains of `_renderer.Initialize(_glLogLevel);` for some games, and not others. X11GLContext.

nitroos-ai:
This works great in Ubuntu 20.04.

Mou-Ikkai:
Well, we have 4.5.0 now for OpenTK.

Thog:
We are working on a new way of handling transition from OpenTK 3, closing.

PR #1564, opened 2020-09-21
Texture: Make viewStorage still valid after view removal.

riperiperi:
There's a problem with some games, including `Hatsune Miku: Project DIVA Mega Mix`, where they use a texture after it has been removed as a view. In this state, `viewStorage` should be the texture itself, but it was being set as null. A method for propagating the modified flag used in `SignalModified` requires that this is not null, so it only became a problem after it was added with the texture management PR.

Obviously I don't have this game, so I'll be waiting for testing from someone else to ensure it's fixed. Game issue: https://github.com/Ryujinx/Ryujinx-Games-List/issues/237

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -259,7 +259,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             _views.Remove(texture);
 
-            texture._viewStorage = null;
+            texture._viewStorage = texture;
 
             DeleteIfNotUsed();
         }


GamerzHell9137:
Can confirm, its fixed now.
image
Ohh and the called out PR is wrong.
Its https://github.com/Ryujinx/Ryujinx-Games-List/issues/237

Cronos-0:
Confirm again Previously, when choosing a song or dressing, the game will crash immediately,fixed now 

bandicam 2020-09-22 02-29-39-745

gdkchan (approved):

Thog (approved):

AcK77 (approved):

PR #1560, opened 2020-09-19
hle/ui: Basic multi programs support

Thog:
This PR implement basic multi programs support and related IPC commands.

This make it possible to launch Super Mario 3D All-Stars and allows to switch between included games.
This PR also clean up and unify all NCA and updates logic to improve maintainability.

image

NOTE: This PR supersede #1557

PR #1569, opened 2020-09-22
Return "NotAvailable" when no UserChannel data is present.

riperiperi:
Now when no data is available, it returns NotAvailable like the other `Pop` functions. Fixes crashing regressions caused by https://github.com/Ryujinx/Ryujinx/pull/1560 .

Examples of games that call this when no data is present: Splatoon, Super Kirby Clash

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -57,6 +57,11 @@ namespace Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.Applicati
                     throw new NotImplementedException($"Unknown LaunchParameterKind {kind}");
             }
 
+            if (storageData == null)
+            {
+                return ResultCode.NotAvailable;
+            }
+
             MakeObject(context, new AppletAE.IStorage(storageData));
 
             return ResultCode.Success;
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -53,8 +53,11 @@ namespace Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.Applicati
                     // Only the first 0x18 bytes of the Data seems to be actually used.
                     storageData = StorageHelper.MakeLaunchParams(context.Device.System.State.Account.LastOpenedUser);
                     break;
-                default:
+                case LaunchParameterKind.Unknown:
                     throw new NotImplementedException($"Unknown LaunchParameterKind {kind}");
+
+                default:
+                    return ResultCode.ObjectInvalid;
             }
 
             if (storageData == null)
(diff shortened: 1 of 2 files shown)


jduncanator (approved):

Cronos-0:
Confirm UserChannel empty should be resolved.

bandicam 2020-09-23 10-41-47-546

Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
```diff
@@ -57,6 +57,11 @@ public ResultCode PopLaunchParameter(ServiceCtx context)
                     throw new NotImplementedException($"Unknown LaunchParameterKind {kind}");
```
AcK77, line 57: After a quick RE, we should throw this exception.
This check is made at the beginning of the function instead and we should add it.
```csharp
if ((kind - 1) > 2 )
{
    return ResultCode.ObjectInvalid;
}
```

AcK77 (changes requested):
The returned ResultCode is right according to RE, but there is an extra check as I commented.

Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
gdkchan:
```cs
if (kind != LaunchParameterKind.UserChannel && kind != LaunchParameterKind.PreselectedUser)
{
    return ResultCode.ObjectInvalid;
}
```
is much more readable.

AcK77: Yeah "we shouldn't" sorry for the mistake! I've just given the RE pseudocode :D

gdkchan: Are you sure its doing a `>` comparison? If that's the case then that means there's one kind missing from the enum (with value 3).

AcK77: The value 3 is something but for now it seems reserved/unused.

https://github.com/Ryujinx/Ryujinx/blob/6c9565693fd87ae1af81ed63b5fbdde2a5dbecb8/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/Types/LaunchParameterKind.cs#L3-L8

riperiperi: Can't I just return `ResultCode.ObjectInvalid` from `default:` in the switch statement? ...and add a case for unknown that throws.

AcK77: Yeah it's fine too ;)

riperiperi: Added handling for this.

AcK77 (approved):

gdkchan (approved):
lgtm, thanks.

PR #1572, opened 2020-09-23
Run audren services in a separate thread

mageven:
```
Amadeus brings his own strings
```

Run `audren` services in a separate thread to remove audio stutter when unrelated services calls are spammed.
This stutter is due to reuse of `CommonServer` for `audren` after IPC refactor part 1 #1447 

I take no credit for this fix.

PR #1573, opened 2020-09-24
Isolate more services to separate threads

mageven:
This PR isolates the more active services to their own threads.
The ones added here are Surfaceflinger family, HID and Time Services.

I wouldn't be surprised if these services are on separate threads in the native sysmodules as well.

Expands on #1572

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
@@ -35,7 +35,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid
         private HidAccelerometerParameters _accelerometerParams;
         private HidVibrationValue          _vibrationValue;
 
-        public IHidServer(ServiceCtx context)
+        public IHidServer(ServiceCtx context) : base(new ServerBase("HidServer"))
         {
             _xpadIdEvent                 = new KEvent(context.Device.System.KernelContext);
             _palmaOperationCompleteEvent = new KEvent(context.Device.System.KernelContext);
--- a/Ryujinx.HLE/HOS/Services/Time/IStaticServiceForGlue.cs
+++ b/Ryujinx.HLE/HOS/Services/Time/IStaticServiceForGlue.cs
@@ -15,7 +15,7 @@ namespace Ryujinx.HLE.HOS.Services.Time
         private IStaticServiceForPsc _inner;
         private TimePermissions      _permissions;
 
-        public IStaticServiceForGlue(ServiceCtx context, TimePermissions permissions)
+        public IStaticServiceForGlue(ServiceCtx context, TimePermissions permissions) : base(new ServerBase("TimeServer"))
         {
             _permissions = permissions;
             _inner       = new IStaticServiceForPsc(context, permissions);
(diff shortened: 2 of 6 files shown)


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

jduncanator:
Why were these three services chosen in particular? Without investigation on the console, determining which ones actually run their own IPC servers, I feel like we're clutching at straws.

If we're going to be moving services out, we should be making informed decisions on which ones to move, rather than guessing at which ones seem like they'd have the most impact. Things like the time server here are questionable to me, as they're very infrequently called.

I don't mind making these changes, we just need evidence to back up making them, otherwise we're quickly going to end up back where we were before we merged the IPC changes in the first place.

mageven:
They are not guesses. I will explain when I get some free time.

Fixed `DisplayServer` as it already has root `ServerBase`s in "vi" services

mageven:
First of all, the `ServerBase` PRs are temporary because IPC Pt1 was merged early. 

The current `ServerBase` is not in its final form. Pt1 `ServerBase` is written to model 1 process and 1 thread only. Actual sysmods aren't restricted that way. They can use syscalls to create any number of threads for any number of ipc servers. The set of `ServerBase`s added in this PR is quite conservative. I only separated services in different sysmodules into different threads to avoid any races. Official services actually have even more threads (in a single process) for the ipc servers than what we map here. For eg., HWOpus has its own ipc server thread but I didn't add it since it's still under the same process.

> Why were these three services chosen in particular? 

It's a combination of intuition and testing. I started with the question- "What services would I use separate threads for if I designed HOS?".
For me those were at least- HID, Display services (nvnflinger & nvservice), Audio (audout and audren) and Time.
I added the ones missing in the above list and checked the logs to confirm that `CommonServer` wasn't being overwhelmed.

> Things like the time server here are questionable to me as they're very infrequently called.

TimeServer is used quite a lot. Both `rtc` and `time:_` have their own ipc servers in official firmware from a quick check.
Also, `TimeServer` was already added by @gdkchan in IPC Pt2, so it's not just me.

Thought this reasoning was obvious to maintainers so I skipped it.
Also, AFAIK the number of dedicated threads will only increase as more refactoring is done, not decrease. It's just going to be in proper groups instead of a common large thread pool before the refactor.

jduncanator:
I'm aware of how it all works, and understand that this isn't the "final form". My point question was surrounded around how you came to the decision to break those three specific services out into their own threads, and what drove that decision.

At the end of the day, I'm happy to concede on these ones, but I'd much rather decisions be made based on actual evidence from the kernel and operating system (eg. from RE), especially when the intent of the original IPC clean-up series was to ensure that our IPC system as closely matched the original console as possible. Randomly stabbing in the dark at services that should be split out into individual servers is contradictory to that goal.

I'm happy to mark this as approved, but I would rather just get this completed properly, with the other two PR's in the series reviewed, than hack together more patches in the meantime.

jduncanator (approved):

mageven:
> My point/question was surrounded around how you came to the decision to break those three specific services out into their own threads, and what drove that decision.

Yeah, I answered this too. The other part is there for providing the whole context to non-maintainers.

> Randomly stabbing in the dark at services that should be split out into individual servers is contradictory to that goal.

It's not stabbing in the dark. It was a combination of RE and informed decisions (just as you desired). I just didn't add the usual flavor of docs/explanation this time as it's only temporary.

>  the other two PR's in the series reviewed, than hack together more patches in the meantime.

The other 2 PRs won't be enough since they don't contain all the changes to each IPC service for completing the "explicit"ness.

PR #1578, opened 2020-09-25
Correct opcode for FCVTNS (Scalar)

### LINKED ISSUE #1570: Unimplemented FCVTNS instruction

Requires the *FCVTNS* instruction.</br>

```
The instruction at 0x000000000aa6d684 (opcode 0x1e200148) is undefined!
```

### Instruction name
```
fcvtns w8, s10
```

### Required by: 

https://github.com/Ryujinx/Ryujinx-Games-List/issues/2549

saldabain:
closes #1570

It looks like the opcode for FCVTNS was incorrect. I've corrected it according to the documentation here:

https://developer.arm.com/docs/ddi0596/h/simd-and-floating-point-instructions-alphabetic-order/fcvtns-scalar-floating-point-convert-to-signed-integer-rounding-to-nearest-with-ties-to-even-scalar

---

Please leave comments, as I think there is room for further improvement (such as changing the type from OpCodeSimd to OpCodeSimdCvt)


gdkchan:
The encoding is correct. It is for the SIMD reg scalar variants (under `FCVTNS (vector)` page on the manual). The one on the linked issue is from SIMD reg scalar to general purpose reg (under `FCVTNS (scalar)` page on the manual).

If you want to implement said instruction properly, feel free to open a new PR, but since this is not the correct fix, I will close this one.

PR #1584, opened 2020-09-29
Convert 1D texture targets to 2D

gdkchan:
Some games render to 1D textures. For the GPU, there's no difference between a 1D texture and a 2D texture with a height of 1, when it is being rendered. For this reason we can't determine if the texture is 1D or 2D when it is used as render target or copy texture. That means that if a game renders to a 2D texture, and later tries to sample it as a 1D texture, it will read garbage as 2D and 1D targets are not considered compatible (the host API does not allow that).

To fix this issue I just converted all 1D textures to 2D textures with a height of 1. The generated shader code was also adjusted to expect 2D textures instead of 1D ones. I added a constant that controls whenever it does that for the shader or not, later the constant can be replaced by a flag somewhere to make the behavior easily configurable.

This fixes the fog (?) on Monster Hunter Generations Ultimate.
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
@@ -144,6 +144,22 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             Target target = descriptor.UnpackTextureTarget().Convert((samplesInX | samplesInY) != 1);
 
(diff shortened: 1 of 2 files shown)


riperiperi (approved):
That's a cool trick. Will try a few games to verify, code LGTM.

Thog (approved):

PR #1585, opened 2020-09-29
Replace Host FPS with GPU command queue load ("Fifo %")

riperiperi:
image
image

Host FPS is rather confusing for users as they can easily confuse it with Game FPS, thinking that it represents how many frames the game is putting out. However, it is a useful metric for determining if it's our GPU emulation causing the framerate bottleneck, as the host has time to generate more duplicate frames while waiting for more commands. It is also useful to see in screenshots, as it can indicate games with GPU related framerate issues.

This PR introduces a replacement "Fifo %" which indicates the percentage of time spent executing processing the FIFO queue over a period of time. This is essentially how _saturated_ the graphics processing thread is. A new mechanism for measuring percentage time has been added to the PerformanceStatistics class.

It's calculated by getting the percentage of the time since the last measurement between the start and end marker. This is then averaged similar to FPS, though there's no interpolation from previous values.

The name "Fifo" is very deliberately chosen:

- Anything with GPU in the name would lead people to believe the metric is related to their host GPU, and that they could reduce their load by getting a 3090 or something.
- The real name should be "Graphics Command Queue Load" or something, but that would be too long to fit in the bottom bar.

### Notes:
- **With a low Game FPS:**
  - A high Fifo % (> 75%) indicates that the game is being held back by the graphics processing thread.
  - A low Fifo % indicates that the game is spending a lot of time waiting for commands to be sent to the FIFO, possibly due to jitted code or HLE slowdown.
- **When at full speed:**
  - A low Fifo % indicates a lot of headroom on the graphics processing thread.
  - A high Fifo % indicates that the game is close to slowing down due to the graphics processing thread.
  - Host FPS could not give feedback like this!
- This measurement will still work if we use a separate context for the emulated GPU, unlike Host FPS which would likely just report the output refresh rate.
- Different games have different points where Fifo % indicates a "GPU bottleneck", as commands are not delivered to the GPU instantly.
- Fifo % can also be high if the Host GPU is slow, or the driver is excessively blocking on OpenGL calls. This is important to remember.

### DIFF
--- a/Ryujinx.HLE/PerformanceStatistics.cs
+++ b/Ryujinx.HLE/PerformanceStatistics.cs
@@ -8,7 +8,6 @@ namespace Ryujinx.HLE
         private const double FrameRateWeight = 0.5;
 
         private const int FrameTypeGame   = 0;
-
         private const int PercentTypeFifo = 0;
 
         private double[] _averageFrameRate;
--- a/Ryujinx.HLE/PerformanceStatistics.cs
+++ b/Ryujinx.HLE/PerformanceStatistics.cs
@@ -1,4 +1,5 @@
-﻿using System.Diagnostics;
+﻿using Ryujinx.Common;
+using System.Diagnostics;
 using System.Timers;
 
 namespace Ryujinx.HLE
@@ -27,8 +28,6 @@ namespace Ryujinx.HLE
 
         private double _ticksToSeconds;
 
-        private Stopwatch _executionTime;
-
         private Timer _resetTimer;
 
         public PerformanceStatistics()
@@ -48,10 +47,6 @@ namespace Ryujinx.HLE
             _frameLock   = new object[] { new object() };
             _percentLock = new object[] { new object() };
 
-            _executionTime = new Stopwatch();
-
-            _executionTime.Start();
-
             _resetTimer = new Timer(1000);
 
             _resetTimer.Elapsed += ResetTimerElapsed;
@@ -60,7 +55,7 @@ namespace Ryujinx.HLE
 
             _resetTimer.Start();
 
-            _ticksToSeconds = 1.0 / Stopwatch.Frequency;
+            _ticksToSeconds = 1.0 / PerformanceCounter.TicksPerSecond;
         }
 
         private void ResetTimerElapsed(object sender, ElapsedEventArgs e)
@@ -131,14 +126,14 @@ namespace Ryujinx.HLE
 
         private void StartPercentTime(int percentType)
         {
-            double currentTime = _executionTime.ElapsedTicks * _ticksToSeconds;
+            double currentTime = PerformanceCounter.ElapsedTicks * _ticksToSeconds;
 
             _percentStartTime[percentType] = currentTime;
         }
 
         private void EndPercentTime(int percentType)
         {
-            double currentTime = _executionTime.ElapsedTicks * _ticksToSeconds;
+            double currentTime = PerformanceCounter.ElapsedTicks * _ticksToSeconds;
 
             double elapsedTime = currentTime - _percentLastEndTime[percentType];
             double elapsedActiveTime = currentTime - _percentStartTime[percentType];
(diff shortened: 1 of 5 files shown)


gdkchan (commented):
Nice, I would probably also remove the host FPS calculation logic, unless it is expected to be used elsewhere.

Ryujinx.HLE/PerformanceStatistics.cs
```diff
@@ -74,6 +91,27 @@ private void CalculateAverageFrameRate(int frameType)
...
+            {
+                _averagePercent[percentType] = double.IsNaN(percent) ? 0 : percent;
+
+                _percentCount[percentType] = 0;
+
+                _accumulatedPercent[percentType] = 0;
+            }
+        }
+
         private double LinearInterpolate(double old, double New)
```
gdkchan, line 115: Since you made some changes on the file anyway, could be nice to rename those function args so that `New` follow our naming conventions. The issue is that new is a reserved keywoard, but maybe they could just be called `lhs` and `rhs` or something.

Ryujinx.HLE/PerformanceStatistics.cs
```diff
@@ -74,6 +91,27 @@ private void CalculateAverageFrameRate(int frameType)
...
+            double percent = (_percentStartTime[percentType] == 0) ? 0 : 100;
+
+            if (_percentCount[percentType] > 0)
+            {
+                percent = _accumulatedPercent[percentType] / _percentCount[percentType];
+            }
+
+            lock (_percentLock[percentType])
+            {
+                _averagePercent[percentType] = double.IsNaN(percent) ? 0 : percent;
```
gdkchan, line 107: In which case can it be NaN?

riperiperi: It can only happen with the calculation outside the lock, actually. I can move that into the lock to avoid the issue.

riperiperi:
Host FPS has been removed, but the variables have been left as arrays in case we add more FPS or Percent type readings to the statistics class in future.

gdkchan (approved):
lgtm, thanks.

Thog (approved):
lgtm (rip Host FPS ;-;)

jduncanator:
Is it possible to make it “FIFO %”? Technically it should all be capitals as it’s an abbreviation. 😬

riperiperi:
Wish granted 🌠

jduncanator (approved):
This looks good to me. My only thought, whilst you're in this file, was to remove the standalone Stopwatch, and use the global PerformanceCounter class available in Ryujinx.Common. This way, all of that "performance related timer" code is centralized. You can just use `PerformanceCounter.ElapsedTicks` in place of `_executionTime` and `PerformanceCounter.TicksPerSecond` to calculate how many ticks are in a second.

Ryujinx.HLE/PerformanceStatistics.cs
```diff
@@ -7,16 +7,24 @@ public class PerformanceStatistics
     {
         private const double FrameRateWeight = 0.5;
 
-        private const int FrameTypeSystem = 0;
-        private const int FrameTypeGame   = 1;
+        private const int FrameTypeGame   = 0;
+
```
AcK77, line 11: Remove extra empty line

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -345,13 +347,11 @@ public void Render()
...
                     _device.PresentFrame(SwapBuffers);
 
-                    _device.Statistics.RecordSystemFrameTime();
-
                     StatusUpdatedEvent?.Invoke(this, new StatusUpdatedEventArgs(
                         _device.EnableDeviceVsync,
                         dockedMode,
-                        $"Host: {_device.Statistics.GetSystemFrameRate():00.00} FPS",
                         $"Game: {_device.Statistics.GetGameFrameRate():00.00} FPS",
+                        $"FIFO: {_device.Statistics.GetFifoPercent():0.00} %",
```
AcK77, line 354: Since you change the order of `StatusUpdatedEventArgs` values, you should probably rename the fields names too in Ryujinx/Ui/StatusUpdatedEventArgs.cs

riperiperi:
Feedback addressed.

jduncanator:
Needs a rebase.

riperiperi:
Rebased.

PR #1588, opened 2020-09-30
Fix GetStream implementation

Thog:
Fix wrong argument usage on GetStream.

Doesn't change anything as it's not used currently but a nice catch from jD ^^.


gdkchan (approved):

jduncanator (approved):
🙌🏼

AcK77 (approved):

PR #1591, opened 2020-10-01
Improve BRX target detection heuristics

gdkchan:
Currently we use a heuristic to find indirect branch targets on shader code. This heuristic was failing to find targets if they were located after a "EXIT" instruction. That was happening because "EXIT" was one of the stop conditions that would make it stop looking. This changes the stop condition to instead stop if the memory is unmapped or if the opcode value is 0 (this disassembles as NOP, but I assume that it's a invalid instruction. It's also what seems to be in the memory after the end of the shader, but I'm not sure if that's 100% reliable).

Maybe theres a better way to solve the problem, I'm not a big fan of this solution to be honest...

This fixes a issue with:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/2538
And the game now renders:
image
Might affect other games aswell, of course.

### DIFF
--- a/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
@@ -128,7 +128,7 @@ namespace Ryujinx.Graphics.Shader.Decoders
                 }
 
                 // Do we have a block after the current one?
-                if (!IsExit(currBlock.GetLastOp()) && currBlock.BrIndir != null)
+                if (currBlock.BrIndir != null && HasBlockAfter(gpuAccessor, currBlock, startAddress))
                 {
                     bool targetVisited = visited.ContainsKey(currBlock.EndAddress);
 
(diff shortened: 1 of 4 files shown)


jduncanator (approved):

Ryujinx.Graphics.Shader/Decoders/Decoder.cs
```diff
@@ -154,6 +154,19 @@ Block GetBlock(ulong blkAddress)
             return blocks.ToArray();
         }
 
+        private static bool HasBlockAfter(IGpuAccessor gpuAccessor, Block currBlock, ulong startAdddress)
+        {
+            if (!gpuAccessor.MemoryMapped(startAdddress + currBlock.EndAddress) ||
+                !gpuAccessor.MemoryMapped(startAdddress + currBlock.EndAddress + 7))
```
jduncanator, line 160: This magical `+ 7` could probably have an explanation in a comment, or be extracted to a constant somewhere.

gdkchan: It's just checking if the entire range is mapped, as it reads 8 bytes, it checks if the address of the first and 8th byte are mapped.

Thog (approved):

PR #1593, opened 2020-10-02
Support 2D array ASTC compressed texture formats decoding

gdkchan:
Decoding 2D array ASTC texture was not supported, which means that layers other than the base layer would be completely blank. This adds support for decompression for 2D array textures aswell.

Fixes missing textures in Donkey Kong Country Tropical Freeze.
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -657,6 +657,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     Info.Height,
                     _depth,
                     Info.Levels,
+                    _layers,
                     out Span<byte> decoded))
                 {
                     string texInfo = $"{Info.Target} {Info.FormatInfo.Format} {Info.Width}x{Info.Height}x{Info.DepthOrLayers} levels {Info.Levels}";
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -657,6 +657,7 @@ private ReadOnlySpan<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data
                     Info.Height,
                     _depth,
                     Info.Levels,
+                    _layers,
                     out Span<byte> decoded))
                 {
                     string texInfo = $"{Info.Target} {Info.FormatInfo.Format} {Info.Width}x{Info.Height}x{Info.DepthOrLayers} levels {Info.Levels}";
```
jduncanator, line 663: Might want to add `_layers` to the debug string here. Would also be useful to inline this on the actual Logger call, so that it can be "optimized away" when the Debug log is not enabled.

jduncanator (approved):
:+1:

Ryujinx.Graphics.Gpu/Image/Texture.cs
gdkchan: `Info.DepthOrLayers` already contains the amount of layers for array textures, or the depth for 3D textures (and there's no "3D array" texture).

riperiperi (approved):
Very nice one!

Ryujinx.Graphics.Gpu/Image/Texture.cs
jduncanator: Sounds good to me.

PR #1607, opened 2020-10-11
Avoid buffer reallocations by increasing their sizes

gdkchan:
Some games might use sliding buffers. For example, it may first use a buffer at [0x1000, 0x4000), then another one at [0x2000, 0x5000), then another one at [0x3000, 0x6000) etc. Since they all overlap each other, each new buffer will cause a new one to be created, then the old data from the old buffer to be copied to the new one, and then the old one is destroyed. This is terrible for performance, and Super Mario 3D All-Stars does that several times per frame, so it had a big impact on performance, until it would eventually stabilize (since it does use all the ranges at some point and stars repeating, and at that point the emulator will be using one large buffer, and no re-allocations are necessary).

This change solves this issue by increasing the buffer size (based on the current size) to avoid the need for future re-allocations.

Potential downsides:
- Higher VRAM usage since the buffers might be larger than necessary. Might not make a difference.
- Removing buffers from the list after they are unmapped becomes harder, as it may have some unaccounted ranges. Can be solved by limiting the buffer ranges to mapped ranges. 

Note: This does not increase the peak performance on games, this only improves the massive slowdown on the Super Mario Galaxy and Sunshine title screen, for example.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
@@ -17,6 +17,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
 
         private const ulong BufferAlignmentSize = 0x1000;
         private const ulong BufferAlignmentMask = BufferAlignmentSize - 1;
+        private const ulong BufferMaxSizeGrowth = 0x200000;
 
         private GpuContext _context;
 
@@ -370,7 +371,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
                 alignedEndAddress += BufferAlignmentSize;
             }
 
-            CreateBufferAligned(alignedAddress, alignedEndAddress - alignedAddress);
+            CreateBufferAligned(alignedAddress, alignedEndAddress - alignedAddress, true);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


gdkchan:
Closing as this approach does not work with transform feedback buffers, as the spec states:
```
Buffers should not be simultaneously used for both transform feedback and other purposes in the GL. Specifically, if a buffer object is bound to any of the binding points in an active and not paused transform feedback object, only writes to that binding point are supported, and any other writes to or reads from the buffer generate undefined values. This includes writes to another  transform  feedback binding point when the same buffer is bound to more than one binding point in the same transform feedback object. Other examples of reads or writes that could generate undefined values include ReadPixels to a pixel buffer object binding point and client access to a buffer mapped with MapBuffer. Commands that attempt to read or write to an active and unpaused transform feedback buffer will have undefined results.  Generating an INVALID_OPERATION error is recommended in this case.
```
It might be possible to make it work with transform feedback by spliting the buffers again, or maybe taking a completely different approach and using sparse buffers, but both have their own issues and it's not something I'm willing to explore as part of those changes.

gdkchan:
Reopening since #1634 will fix the transform feedback issue and should make this viable again.

gdkchan:
Rebased and confirmed that the transform feedback related regression was fixed.

jduncanator (approved):

gdkchan:
Closing as theres not much advantage in doing that anymore, as the only game it benefits (afaik) had a performance regression, and the solution was not very good to begin with.

PR #1610, opened 2020-10-12
Fix error when dual source blend is used

gdkchan:
Dual source blend is a feature that allows two color outputs from the pixel shader to be combined using the GPU blending functions. When one of the dual source blending sources is used, it is only valid to have a single draw buffer set, as on this mode you can only output to a single color buffer. So, this change forces a single draw buffer to be set if dual source blending is in use.

This adds a bunch of checks to the set blend mode function which could potentially slow it down, so if someone has a better suggestion on how to do it without all the checks on a potentially how function, let me know.

This fixes black floor among other things on Super Mario Galaxy observatory.
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -557,6 +557,30 @@ namespace Ryujinx.Graphics.OpenGL
                 (BlendingFactorSrc)blend.AlphaSrcFactor.Convert(),
                 (BlendingFactorDest)blend.AlphaDstFactor.Convert());
 
+            static bool IsDualSource(BlendFactor factor)
+            {
+                switch (factor)
+                {
+                    case BlendFactor.Src1Color:
+                    case BlendFactor.Src1ColorGl:
+                    case BlendFactor.Src1Alpha:
+                    case BlendFactor.Src1AlphaGl:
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -557,6 +557,30 @@ public void SetBlendState(int index, BlendDescriptor blend)
...
+                    case BlendFactor.OneMinusSrc1ColorGl:
+                    case BlendFactor.OneMinusSrc1Alpha:
+                    case BlendFactor.OneMinusSrc1AlphaGl:
+                        return true;
+                }
+
+                return false;
+            }
+
+            _framebuffer.SetDualSourceBlend(
```
riperiperi, line 578: `_framebuffer` can be null here, if EnsureFramebuffer was not called elsewhere. This will crash on Mario Kart 8.

riperiperi (changes requested):

Ryujinx.Graphics.OpenGL/Pipeline.cs
gdkchan: Fixed.

Thog (approved):
lgtm, nice catch!

riperiperi (approved):

PR #1613, opened 2020-10-13
Fix output component register on pixel shaders

gdkchan:
This fixes an error I noticed on one of Super Mario Galaxy shaders. When the shader does not output to all the components of a color buffer, the register number still corresponds to the component number. So if, for example, only A (alpha) is written to, the output will be still on R3 rather than R0.

I'm not sure about what exactly this fixes on this game, but mario shadow seems better with this fix. I also tested on other games that had similar issues on the past, and they did not regress.


riperiperi (approved):
Nice catch!

AcK77 (approved):

jduncanator (approved):

PR #1609, opened 2020-10-12
Implement LEA.HI shader instruction

gdkchan:
Implements the LEA.HI shader instruction.
This fixes Rosalina rendering on Super Mario Galaxy.
Before:
image
After:
image

PR #1618, opened 2020-10-15
Implement CAL and RET shader instructions

gdkchan:
#1609 caused a "regression" on FE3H where the game would crash during battle due to what looks like a infinite loop on a compute shader. The `LEA.HI` implementation seems correct, and the shaders on this game does use this instruction. So what seems to be happening here is that implementing this instruction caused a change in the behavior of the shader that caused new paths to be taken on the code. There were however other instructions that were not implemented (namely `CAL` and `RET`), which means that the shader was still incorrect and whenever it would be good enough to at least finish is just a gamble without the instructions properly implemented.

Implementing the instructions fixed the user reported issue, thus apparently at least fixing the regression.

A few notes:
- Using the `EXIT` instruction from a function that is not the main one is considered invalid, I don't know how this behaves on the switch.
- Using the `RET` instruction from the main function is also considered invalid. Same reason as above.
- Using SSY/PBK on the caller, and SYNC/BRK on the callee (expecting to find the caller values on stack) will not work (I assume that this is not valid because the return address will be on the stack?).

Implementation:
- Uses GLSL functions, and call them, as one would expect. Most of the work was adding function support on the IR and backend.
- Register consumed by the function are used as input arguments on the function. They are identified on the `RegisterUsage` pass.
- Registers modified by the function are returned  using several `out` arguments.

Sample of the code generated:
```glsl
void fun5(int a0, out int a1, out int a2, out int a3, out int a4)
{
    uint temp_65;
    int temp_66;
    int temp_67;
    int temp_68;
    precise float temp_69;
    precise float temp_70;
    precise float temp_71;
    precise float temp_72;
    // 0x02DDB8: 0x3828000000270303 Shr
    temp_65 = uint(a0) >> 2;
    // 0x02DDC8: 0x1C00000000370306 Iadd
    temp_66 = int(temp_65) + 3;
    // 0x02DDD0: 0x1C00000000270305 Iadd
    temp_67 = int(temp_65) + 2;
    // 0x02DDD8: 0x1C00000000170304 Iadd
    temp_68 = int(temp_65) + 1;
    // 0x02DDE8: 0xDA00008FFFF70303 Texs
    temp_69 = texelFetch(cp_tex8, Helper_TexelFetchScale(ivec2(int(temp_65), 0), 0), 0).r;
    // 0x02DDF0: 0xDA00008FFFF70606 Texs
    temp_70 = texelFetch(cp_tex8, Helper_TexelFetchScale(ivec2(temp_66, 0), 0), 0).r;
    // 0x02DDF8: 0xDA00008FFFF70505 Texs
    temp_71 = texelFetch(cp_tex8, Helper_TexelFetchScale(ivec2(temp_67, 0), 0), 0).r;
    // 0x02DE08: 0xDA00008FFFF70404 Texs
    temp_72 = texelFetch(cp_tex8, Helper_TexelFetchScale(ivec2(temp_68, 0), 0), 0).r;
    // 0x02DE10: 0xE32000000007000F Ret
    a1 = floatBitsToInt(temp_69);
    a2 = floatBitsToInt(temp_72);
    a3 = floatBitsToInt(temp_71);
    a4 = floatBitsToInt(temp_70);
    return;
}
```
```glsl
// 0x0046F8: 0xE26000296B800040 Cal
fun5(temp_2776, temp_2777, temp_2778, temp_2779, temp_2780);
// 0x004708: 0x1C00000006070702 Iadd
temp_2781 = temp_2507 + 96;
// 0x004710: 0xE26000294B800040 Cal
fun1(temp_2781, temp_2782);
// 0x004718: 0x5C9807800007002B Mov
// 0x004728: 0x1C00000001070702 Iadd
temp_2783 = temp_2507 + 16;
// 0x004730: 0xE260002949800040 Cal
fun1(temp_2783, temp_2784);
```

### DIFF
--- a/Ryujinx.Graphics.Shader/Translation/Translator.cs
+++ b/Ryujinx.Graphics.Shader/Translation/Translator.cs
@@ -31,8 +31,8 @@ namespace Ryujinx.Graphics.Shader.Translation
 
         public static ShaderProgram Translate(ulong addressA, ulong addressB, IGpuAccessor gpuAccessor, TranslationFlags flags)
         {
-            var funcA = DecodeShader(addressA, gpuAccessor, flags | TranslationFlags.VertexA, out ShaderConfig configA);
-            var funcB = DecodeShader(addressB, gpuAccessor, flags, out ShaderConfig config);
+            FunctionCode[] funcA = DecodeShader(addressA, gpuAccessor, flags | TranslationFlags.VertexA, out ShaderConfig configA);
+            FunctionCode[] funcB = DecodeShader(addressB, gpuAccessor, flags, out ShaderConfig config);
 
             config.SetUsedFeature(configA.UsedFeatures);
 
--- a/Ryujinx.Graphics.Shader/Translation/Translator.cs
+++ b/Ryujinx.Graphics.Shader/Translation/Translator.cs
@@ -105,40 +105,6 @@ namespace Ryujinx.Graphics.Shader.Translation
             return new ShaderProgram(spInfo, config.Stage, glslCode, config.Size, sizeA);
         }
 
-        private static void RemoveUnreachable(ref BasicBlock[] blocks)
-        {
-            bool modified;
-
-            do
-            {
-                modified = false;
-
-                for (int i = 0; i < blocks.Length; i++)
-                {
-                    if (blocks[i] != null && !blocks[i].Reachable)
-                    {
-                        blocks[i].Next = null;
-                        blocks[i].Branch = null;
-                        blocks[i] = null;
-                        modified = true;
-                    }
-                }
-            }
-            while (modified);
-
-            int j = 0;
-
-            for (int i = 0; i < blocks.Length; i++)
-            {
-                if (blocks[i] != null)
-                {
-                    blocks[j++] = blocks[i];
-                }
-            }
-
-            Array.Resize(ref blocks, j);
-        }
-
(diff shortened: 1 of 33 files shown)


Thog (approved):

riperiperi (approved):
Lots going on here, but it's looking great. Nice work.

Ryujinx.Graphics.Shader/Translation/Translator.cs
```diff
@@ -14,39 +14,82 @@ public static class Translator
...
         public static ShaderProgram Translate(ulong address, IGpuAccessor gpuAccessor, TranslationFlags flags)
         {
             return Translate(DecodeShader(address, gpuAccessor, flags, out ShaderConfig config), config);
         }
 
         public static ShaderProgram Translate(ulong addressA, ulong addressB, IGpuAccessor gpuAccessor, TranslationFlags flags)
         {
-            Operation[] opsA = DecodeShader(addressA, gpuAccessor, flags | TranslationFlags.VertexA, out ShaderConfig configA);
-            Operation[] opsB = DecodeShader(addressB, gpuAccessor, flags, out ShaderConfig config);
+            var funcA = DecodeShader(addressA, gpuAccessor, flags | TranslationFlags.VertexA, out ShaderConfig configA);
```
riperiperi, line 34: Should `var` be used here? It's not quite as clear as the use below for `cfgs` where the type is nearby and easy to infer.

Issue #1620, opened 2020-10-15
Xenoblade Chronicles Definitive Edition: Textures bug 


eruxxn36:
### What's the issue you encountered?

xenoblade chronicles in Ryujinx (mostly ok)

but sometime has wrong 
 
main problem is in the Texture (like picture)

![001](https://user-images.githubusercontent.com/72922836/96121641-ee860f80-0f22-11eb-90b5-f96ca8d36724.jpg)

![002](https://user-images.githubusercontent.com/72922836/96121660-f6de4a80-0f22-11eb-8352-579f685e5fc4.jpg)

![003](https://user-images.githubusercontent.com/72922836/96121681-fc3b9500-0f22-11eb-8700-965534660362.jpg)

![004](https://user-images.githubusercontent.com/72922836/96121705-02317600-0f23-11eb-9b16-5eeb969d0e7b.jpg)

![005](https://user-images.githubusercontent.com/72922836/96121722-08275700-0f23-11eb-88b5-7387fee3a963.jpg)

![006](https://user-images.githubusercontent.com/72922836/96137147-e207b400-0f2e-11eb-9890-ad7924f79da5.jpg)

### How can the issue be reproduced?

in game

the first village and big map 

### Log file

[Ryujinx_xenoblade.log](https://github.com/Ryujinx/Ryujinx/files/5384660/Ryujinx_xenoblade.log)

### Environment?

- Ryujinx version: ryujinx-1.0.5525-win_x64</br>

- Game version: 1.0</br>

- System Specs:
  - OS: *Windows 7 SP 1 (6.1)*
  - CPU: *Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz*
  - GPU: *NVIDIA GTX 1066*
  - RAM: *16GB*
- Applied Mods : [  No ]

### Additional context?

hope the bug can fix

Thanks the devs for making the Ryujinx

Doctorwho1909:
Also add an image

eruxxn36:
Thank you very much for gdkchan and others response

but I try ryujinx-1.0.5545-win_x64

There are still some problems

shadow

![shadow001](https://user-images.githubusercontent.com/72922836/96422460-9e60c300-122a-11eb-89ee-683342ef01b6.jpg)
![shadow002](https://user-images.githubusercontent.com/72922836/96422478-a3be0d80-122a-11eb-9049-c4ec832436c5.jpg)

log
[...]

PR #1625, opened 2020-10-16
Fix image binding format

### LINKED ISSUE #1539: Kirby Star Allies Fire & Ice Texture Bug

Kirby Star Allies has a gpu bug wherein using the fire sword or ice skating results in black (or sometimes green) blocky artifacts showing on the ground and on characters.

Use any Ryujinx build after 1.0.4639 (the issue begins in 1.0.4640 - https://github.com/Ryujinx/Ryujinx/pull/1216) and use a fire sword effect (directional pad + B for sword dash, or hold B for spinning attack), seen in the first video below.
Or get ice effects at the end of one of the first areas and run back and forth outside the exit (seen in the 2nd video).

### Log file

[KSA.zip](https://github.com/Ryujinx/Ryujinx/files/5179753/KSA.zip)

### Environment?

- Ryujinx version: 1.0.4640 or newer</br>
- Game version: Any</br>

- System Specs:
  - OS: Windows 10 2004
  - CPU: Intel i5-6600K
  - GPU: NVIDIA GTX 1080
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Screenshots/videos:
2020-09-06 06_21_32-Ryujinx 1 0 5271 - Kirby Star Allies v4 0 0 (01007E3006DDA000) (64-bit)
2020-09-06 06_52_27-Ryujinx 1 0 5034 - Kirby Star Allies v4 0 0 (01007E3006DDA000) (64-bit)

https://streamable.com/etjupi
https://streamable.com/l0yzpo

https://github.com/Ryujinx/Ryujinx-Games-List/issues/45

gdkchan:
This fixes a number of related things. First, the `EXT_shader_image_load_store` extension states:
```
If the format layout qualifier is present, 
then the format must match the format of the associated image unit, as 
enumerated in table 8.33, otherwise the access is considered to involve a 
format mismatch
```
In fact the format that was specified on the shader was being ignored, and it was using the format of the image binding instead. So I changed it to use the same format declared on the shader for the image binding aswell, this fixes the mismatch and makes the load/store behave as expected.

The second issue is that there are basically two modes on the SULD/SUST (image load/store) shader instructions. One of them is sized, so it just read/writes a 8/16/32/64/128-bits value in memory directly. For this mode it uses a equivalent integer format with the same number of bytes per pixel as the access. And there is another mode that is supposed to use the binding format. Some shaders might load/store the same image using those two different modes. This requires different image formats, so to make it work I duplicated the image, so that it binds the image twice with the 2 formats that are required. **The downside** with this approach is that it consumes shader image declarations, a very limited resource (you can only use up to 8 per shader), so this change makes hiting this limit easier.

The last issue, that I found on Kirby Star Allies, is that it tries to do  image load/store on a sRGB texture. This is not supported, so I fixed it by using a RGBA format instead. I'm not sure if the switch performs sRGB conversion on image load/store, but I guess it doesn't.

This fixes #1539:
image
This also fixes the glitch on the ground in Xenoblade Chronicles Definitive Edition (one of the issues listed on #1620):
image

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -46,7 +46,7 @@ namespace Ryujinx.Graphics.GAL
 
         void SetIndexBuffer(BufferRange buffer, IndexType type);
 
-        void SetImage(int index, ShaderStage stage, ITexture texture);
+        void SetImage(int index, ShaderStage stage, ITexture texture, Format imageFormat);
 
         void SetLogicOpState(bool enable, LogicalOp op);
 
--- a/Ryujinx.Graphics.Gpu/Engine/Compute.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute.cs
@@ -128,7 +128,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
             {
                 var descriptor = info.Textures[index];
 
-                Target target = GetTarget(descriptor.Type);
+                Target target = ShaderTexture.GetTarget(descriptor.Type);
 
                 if (descriptor.IsBindless)
                 {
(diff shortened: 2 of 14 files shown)


Thog (approved):
lgtm

riperiperi (approved):
LGTM

EmulationFanatic:
Confirmed that it fixes https://github.com/Ryujinx/Ryujinx/issues/1539

PR #1626, opened 2020-10-16
Get rid of Reflection.Emit dependency on CPU and Shader projects

gdkchan:
Makes the projects more AOT friendly, that will be useful if we decide to use AOT in the future.
I also removed several property `private set` that were not required as the values were set on the constructor, should reduce the binary size a bit.
All the bulk changes were done using just regex search & replace.


Thog (approved):
lgtm

LDj3SNuD (approved):

PR #1627, opened 2020-10-16
Ensure storage is set for Buffer Textures when binding an Image.

riperiperi:
This was causing OpenGL errors in UE4 games, and a driver crash on NVIDIA Pascal GPUs (GeForce 1070 etc). With this fixed, you can now get ingame on UE4 games with a Pascal GPU. (does not appear to fix any other issues)

image


Thog (approved):
lgtm

AcK77 (approved):

gdkchan (approved):
lgtm, nice catch. It would be nice to find a better way to handle this in the future (that does not require constant re-binding).

Issue #1611, opened 2020-10-12
Luigi's Mansion 3: Cannot continue after B2 floor with FastFP enabled


albfan:
### What's the issue you encountered?

Older versions of Ryujinx should allow to continue

https://ci.appveyor.com/project/gdkchan/ryujinx/builds/34061417/artifacts

On recent versions the only way to make it work is to disable FastFP

https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Optimizations.cs#L7

```diff
diff --git a/ARMeilleure/Optimizations.cs b/ARMeilleure/Optimizations.cs
index fbbbfdb2..93c5adbb 100644
--- a/ARMeilleure/Optimizations.cs
+++ b/ARMeilleure/Optimizations.cs
@@ -4,7 +4,7 @@ namespace ARMeilleure
 {
     public static class Optimizations
     {
-        public static bool FastFP { get; set; } = true;
+        public static bool FastFP { get; set; } = false;
 
         public static bool UseSseIfAvailable       { get; set; } = true;
         public static bool UseSse2IfAvailable      { get; set; } = true;
```

@jduncanator seems related with Ryujinx/Ryujinx#1335

### How can the issue be reproduced?

Attached to saved state games for game update 1.4.0 to test:

[lm3-before-go-b2-floor.zip](https://github.com/Ryujinx/Ryujinx/files/5367090/lm3-before-go-b2-floor.zip)
[lm3-after-b2-floor.zip](https://github.com/Ryujinx/Ryujinx/files/5367091/lm3-after-b2-floor.zip)

Try to go to b2 floor or load game state after being in b2 floor

### Log file

N/A

### Environment?

- Ryujinx version: 1.0.0-dirty (master)</br>
- Game version: 1.4.0</br>
- System Specs:
  - OS: Linux
  - CPU: i7
  - GPU: Intel
  - RAM: 32GB
- Applied Mods : No

### Additional context?

More people reported already:

https://github.com/Ryujinx/Ryujinx-Games-List/issues/207#issuecomment-703088912

albfan:
A narrow version of the patch fixes it while keeping the game fast

```
diff --git a/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs b/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
index 0d417f70..eb532658 100644
--- a/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
[...]

PR #1630, opened 2020-10-20
Add support for guest Fz (Fpcr) mode through host Ftz and Daz (Mxcsr) modes (fast paths).

LDj3SNuD:
`FADDP V0.2S, V1.2S, V2.2S` (HighCq):
```asm
  vunpcklps       xmm0, xmm6, xmm7
  vxorps          xmm1, xmm1, xmm1
  vmovlhps        xmm6, xmm0, xmm1
  vmovhlps        xmm7, xmm1, xmm0
  mov             ecx, 0x400000
  vmovd           xmm0, ecx
  vshufps         xmm0, xmm0, xmm0, 0
  vcmpunordps     xmm1, xmm6, xmm6
  vpand           xmm2, xmm6, xmm0
  vcmpeqps        xmm0, xmm2, xmm0
  vandps          xmm2, xmm0, xmm1
  vandnps         xmm0, xmm0, xmm1
  mov             ecx, 0x400000
  vmovd           xmm1, ecx
  vshufps         xmm1, xmm1, xmm1, 0
  vcmpunordps     xmm3, xmm7, xmm7
  vpand           xmm4, xmm7, xmm1
  vcmpeqps        xmm1, xmm4, xmm1
  vandnps         xmm1, xmm1, xmm3
  mov             ecx, 0x400000
  vmovd           xmm3, ecx
  vshufps         xmm3, xmm3, xmm3, 0
  vpandn          xmm1, xmm1, xmm2
  vpor            xmm0, xmm1, xmm0
  vblendvps       xmm0, xmm7, xmm6, xmm0
  vpor            xmm8, xmm0, xmm3
  vcmpordps       xmm9, xmm6, xmm7
  movabs          rcx, 0x7ffdddd03884
  call            rcx
  test            eax, eax
  je              LOC_0x2
  vstmxcsr        dword ptr [rsp + 0x40]
  or              dword ptr [rsp + 0x40], 0x8940
  vldmxcsr        dword ptr [rsp + 0x40]
  LOC_0x2:
  vaddps          xmm0, xmm6, xmm7
  test            eax, eax
  je              LOC_0x3
  vstmxcsr        dword ptr [rsp + 0x44]
  and             dword ptr [rsp + 0x44], 0xffff7fbf
  vldmxcsr        dword ptr [rsp + 0x44]
  LOC_0x3:
  vblendvps       xmm0, xmm8, xmm0, xmm9
```

Tests performed successfully in both debug and release mode, with Fpcr.Fz on/off (for all instructions involved).

Closes https://github.com/Ryujinx/Ryujinx/issues/1611.


gdkchan:
Nice one, but I don't see why MXCSR is not set permanently based on the value of the Arm FPCR register (when the register is modified), since it is supposed to affect all instructions (although, I guess it could have odd results for the floating-point operations that requires multiple instructions after translated).

Also, if a new IR instruction needs to be introduced for this, it's better if it is as generic as possible (for example, ideally it should naturally also map to something that can change Arm FPCR register if we decide to support that CPU on the JIT aswell).

It might be also worth setting the rounding mode this way.

LDj3SNuD:
> Nice one, but I don't see why MXCSR is not set permanently based on the value of the Arm FPCR register (when the register is modified), since it is supposed to affect all instructions (although, I guess it could have odd results for the floating-point operations that requires multiple instructions after translated).

Initially I also tried to manage the Fz mode at Msr level, the Tests failed. In fact, this mode should be managed selectively at the level of the single instruction, where the final calculations take place.

> Also, if a new IR instruction needs to be introduced for this, it's better if it is as generic as possible (for example, ideally it should naturally also map to something that can change Arm FPCR register if we decide to support that CPU on the JIT aswell).

While the previous answer should partly answer this question too, the methods added in EmitterContext already allow the complete control of Csr while preserving its pre-existing state; available for a detailed proposal.

> It might be also worth setting the rounding mode this way.

I can investigate the feasibility (also depending on the support of the Tests and the Tester for such a mode), but this will probably see possible application in a subsequent PR.

FICTURE7:
Perhaps to avoid the new IR instructions something like this could be done?

Before PreAllocator
```wasm
i32 %0 = Extended.X86Stmxscr
i32 %1 = Bitwise{Or/And} i32 %0, 0xC0FFEE
Extended.X86Ldmxscr i32 %1
```

After PreAllocator
```wasm
Extended.X86Stmxscr i32 0x00  ;; 0x00 is a stack slot/offset.
i32 %0 = Fill i32 0x00
i32 %1 = Bitwise{Or/And} i32 %0, i32 0xC0FFEE
Spill i32 %0, i32 0x00
Extended.X86Ldmxscr i32 0x00
```
However this will come with some CQ issues, because the value will be loaded in a register. 

Another approach might be to implement a new `OperandKind` for stack slots.
```wasm
Extended.X86Stmxscr i32 $[0x00]  ;; $[<offset>] represents an operand on the stack.
Bitwise{Or/And} i32 $[0x00], i32 0xC0FFEE
Extended.X86Ldmxscr i32 $[0x00]
```
This should remove the `Fill` and `Spill` instructions, because they can be implemented with `Load` and `Store` instructions referring to stack slots. I believe this is also a prerequisite for the register allocators to leverage x86's ability to operate on memory operands.

gdkchan:
> Initially I also tried to manage the Fz mode at Msr level, the Tests failed. In fact, this mode should be managed selectively at the level of the single instruction, where the final calculations take place.

Fair enough, it would be nice to know which instructions are expected to require this, and which criteria is being used to decide that.

> While the previous answer should partly answer this question too, the methods added in EmitterContext already allow the complete control of Csr while preserving its pre-existing state; available for a detailed proposal.

Would be nice to have CPU agnostic instructions. For example, we could have a `FPControlSet` and `FPControlClear` instruction, that contains a enum indicating which flag should be set or cleared, similar to the enum that is used for conditional branch instructions. The CPU backend would then assemble the proper mask for the CPU. I think it would be also reasonable to use the current approach, but move the new instructions to x86 "intrinsics". Right now none of the regular instructions are changing CPU specific state, they are all general purpose operations that can be done on any modern CPU (be it x86, Arm, MIPS, RISC-V or whatever).

> I can investigate the feasibility (also depending on the support of the Tests and the Tester for such a mode), but this will probably see possible application in a subsequent PR.

Ok.

ARMeilleure/IntermediateRepresentation/Operand.cs
```diff
@@ -119,6 +119,18 @@ public double AsDouble()
             return BitConverter.Int64BitsToDouble((long)Value);
         }
 
+        public static Operand operator ~(Operand op)
```
gdkchan, line 122: IMO it would be better to just create a new operand, it's quite weird that only the bitwise not operator is supported, and that it only supports constant ones.

gdkchan (approved):
lgtm, thanks!

LDj3SNuD:
Up.

LDj3SNuD:
Rebased due to .NET 5.0.
1 review by 1 reviewer required.

gdkchan:
Would be nice to get an answer to that:
> Fair enough, it would be nice to know which instructions are expected to require this, and which criteria is being used to decide that.

LDj3SNuD:
Sorry for the delay (lack of time, I was working on the other open PR). If you see the pseudocode of an FP instruction, the control of the subnormals (to flush them, if Fpcr.Fz is on) is put in place both in input (for n, m) and output (for d) and only for the final calculation (for example only for FPAdd, not for all Faddp; excluding also the NaN processing).
For long story, consult: https://software.intel.com/content/www/us/en/develop/articles/x87-and-sse-floating-point-assists-in-ia-32-flush-to-zero-ftz-and-denormals-are-zero-daz.html

Currently the functionality (Fpcr.Fz) has been applied only to the FPMax/FPMin instructions, because from my tests, together with the FP comparison instructions, they are the only ones that require it (for the moment I thought it was sufficient to apply it only to the first ones); Faddp I could have saved it.

However if this solution, for technical reasons unknown to me, is unwelcome; I can simply detect and treat subnormals in a classic way; but this will imply having to emit many more instructions, because this algorithm must be applied 3 times (for n, m and d).

LDj3SNuD:
Reminder: review required.

LDj3SNuD:
Reminder: review required.

riperiperi (approved):
Very nice work! Sorry for the delay, just wanted to set aside the time to test it for myself.

PR #1634, opened 2020-10-24
Fix transform feedback errors caused by host pause/resume and multiple uses

gdkchan:
The NVIDIA driver is returning an error if an attempt is made to resume transform feedback after changing the shader program. I couldn't find anything on the spec that forbids that, it just says:

> An INVALID_OPERATION error is generated by UseProgram if the current transform feedback object is active and not paused;

`glBindBufferRange` with the target set to transform feedback was also returning an error, which also seems to be caused by a transform feedback being currently active. So as a solution for both errors, it now ends and restarts the feedback instead of pausing and resuming it.

The last issue is that having transform feedback buffers used for anything else while the transform feedback is active is not valid and will cause the draw to fail. This is what blocked #1607. This is fixed here by first doing a copy from the TFB to a temporary internal TFB when it is bound, and then doing a copy from the temporary TFB back to the actual TFB after draw.

Special thanks to Rodrigo for the hint on how to do fast buffer re-allocations.

The shader `Program` constructor was also changed to use `glProgramUniform1` rather than `glUniform1`, because the latter requires a program to be currently bound (using `glUseProgram`), which is undesirable (because it changes the current shader program) and fails if the transform feedback is active. The former allows the program handle to be passed so nothing needs to be bound.

Other minor things that were fixed:
- Comparison of `BufferHandle` with null was being done, but `BufferHandle` is a struct. It now compares with `BufferHandle.Null` which is zero. Note that this will change behavior.
- `BufferHandle` has a `Empty` property that was supposed to be static, but wasn't. Fixed that and used it. Does not change behavior.

`SetTransformFeedbackBuffers` was also changed to take multiple buffers instead of just one at a time. This reduces the number of calls, but otherwise should not change behavior.

This fixes T-Posed models on SNK Heroines: Tag Team Frenzy:
Before (image from gamedb, not mine):
image
After:
image
Should also fix the black grass issue on #1620.

### DIFF
--- a/Ryujinx.Graphics.GAL/BufferRange.cs
+++ b/Ryujinx.Graphics.GAL/BufferRange.cs
@@ -4,7 +4,7 @@ namespace Ryujinx.Graphics.GAL
     {
         private static readonly BufferRange _empty = new BufferRange(BufferHandle.Null, 0, 0);
 
-        public BufferRange Empty => _empty;
+        public static BufferRange Empty => _empty;
 
         public BufferHandle Handle { get; }
 
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -75,7 +75,7 @@ namespace Ryujinx.Graphics.GAL
 
         void SetTexture(int index, ShaderStage stage, ITexture texture);
 
-        void SetTransformFeedbackBuffer(int index, BufferRange buffer);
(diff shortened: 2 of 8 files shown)


Thog (approved):
lgtm

riperiperi (approved):
LGTM.

riperiperi (approved):
Nice one, both the Xenoblades also seem good with this change. Would probably need to play for a long time to see if it still breaks after a while, but my 20 mins was fine.

Thog (approved):

PR #1636, opened 2020-10-25
Improve the speed of redundant ASTC texture data updates

gdkchan:
Some games (well, I'm aware of only one) have a large performance hit caused by constant ASTC data updates. Upon investigation, I discovered that on this specific game, the data does not really change, so my guess is that the game is writing the same data over and over in memory on every frame.

In order to avoid paying the cost for ASTC decompression for those redundant updates, I added a check that skips the update if the data did not change. It needs to keep a copy of the data in memory for that. In order to avoid increasing the memory usage in the common case, it only does that if:
- The texture is ASTC compressed.
- The host does not support the format.
- Modification was detected at least a bunch of times before (currently it uses a hardcoded value of 4).

This improves the performance on SNK Heroines (note: tested with #1634):
Before: https://streamable.com/z4z0w0
After: https://streamable.com/ivumcm

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -16,6 +16,11 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class Texture : IRange, IDisposable
     {
+        // How many updates we need before switching to the byte-by-byte comparison
+        // modification check method.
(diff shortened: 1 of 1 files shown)


Thog (approved):

riperiperi (approved):
Pretty straightforward, I think the threshold makes sense. I can see it accidentally getting triggered by something else on the same page, but I don't think it'll be often enough to use that much more memory.

PR #1642, opened 2020-10-26
Avoid sampler conflicts on bindless samplers with the same name

gdkchan:
The shader decompiler currently tries to find the source of the handles for bindless texture accesses. If it's a constant buffer, it turns it into a regular texture access, and instructs the GPU emulator to read the handle from the respective constant buffer, and bind it properly. If said analysis fails however, it will create a sampler for a handle at constant buffer 0 at offset 0 (those values comes from the fact that the values are not changed from their defaults).

Anyway, the problem is that when the analysis fails for multiple texture accesses, it's possible that multiple texture sampling operations with different sampler types will share the sampler/name. This is invalid and causes the shader compilation to fail. This adds the sampler type to the name of the sampler itself to disambiguate.

This is intended to be a stop-gap solution until we get proper bindless texture support.
This improves rendering on Super Mario Party.
Before:
image
After:
image
Obviously some bindless accesses are still broken, but with this change the shader at least compiles and the characters render.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -317,7 +317,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
                     continue;
                 }
 
-                string samplerTypeName = GetSamplerTypeName(texOp.Type);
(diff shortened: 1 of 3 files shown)


Thog (approved):

riperiperi (approved):
Nice catch. LGTM!

Issue #1594, opened 2020-10-02
Updating "-r" portable install erases files in subdirectory "ryujinx\user", restart does not repsect "-r" command line argument


CeruleanSky:
## Bug Report
I have a portable install using the "-r" command line option and after the updater completed and restarted ryujinx I lost files in my portable user directory which is a subdirectory to ryujinx like Yuzu has.  
All files were erases, but the directories were kept.  
The restarted ryujinx does not launch with "-r"  

### What's the issue you encountered?

I run Ryujinx in a portable install with "-r"   
I have a subdirectory called "user" inside my Ryujinx folder.  
Restarting erased all the files but kept the folder structure in my Ryujinx\user folder  

### How can the issue be reproduced?

1. Unzip ryujinx-1.0.5479-win_x64.zip that requires an update to ryujinx-1.0.5480 d:\ryujinx
2. create a ryujinx\user subdirectory
3. create a shortcut that runs "d:\ryujinx.exe -r d:\ryujinx\user"
4. launch "ryujinx.exe -r d:\ryujinx\user" and say yes to updating
5. Check "d:\ryujinx\user" to find all files such as keys, firmware, config, save files, erased, but the folder structure remains
6. Notice a default config is located in %appdata%\Ryujinx if one was not there previously.

### Log file

[na]

### Environment?

- Ryujinx version: 1.0.5479</br>
- System Specs:  
  - OS: Windows 10  
  - CPU: Intel i7  
  - GPU: NVIDIA GTX 1070  
  - RAM:16GB  
- Applied Mods : No  

### Additional context?

Additional info about your environment:</br>
[na]

TSRBerry:
Has this been resolved?

PR #1643, opened 2020-10-27
Preserve command line arguments when updater restarts Ryujinx

CeruleanSky:
Command line arguments are not preserved by the updater, and this causes an issue when the updater restarts Ryujinx in portable mode as it will create/use/modify its default directory instead.

This simple pull addresses the first part of #1594 

As far as the second part of the issue, where the updater perhaps overzealously deletes everything but log files and leaving a graveyard of empty folders, which causes problems in portable installs with a Ryujinx\user subfolder or something else the user places inside the Ryujinx folder is unexpectedly deleted, the idea of doing what fuller installers do comes to mind.

Perhaps an UpdateList.txt containing a list of files that the updater deletes is included in the ryujinx.zip/tar.gz/etc and placed in the unpacked directory for the updater to use upon the next upgrade.

The build system can run something like `forfiles /s /m *.txt /c "cmd /c echo @relpath"` or the correct command for the OS and include it in the distribution, or it could be generated by the updater itself after it unpacks the files in the temporary directory.


gdkchan (approved):
lgtm, thanks.

Thog (approved):

AcK77 (approved):

PR #1646, opened 2020-10-28
Scale texture resolution before sending to backend

gdkchan:
This scales the texture size before sending it to the OpenGL backend. The advantage of doing that is that it's not necessary to scale it on the several places the sizes are used for something on the backend. Right now there was a bug where it would use the unscaled size to calculate the size of the temporary buffer used for BGRA swap PBO copies, which means that resolution scaling wouldn't work on games presenting BGRA textures, like Dragon Marked for Death for example. This PR fixes this issue.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -181,7 +181,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             {
                 Debug.Assert(!isView);
 
-                TextureCreateInfo createInfo = TextureManager.GetCreateInfo(Info, _context.Capabilities);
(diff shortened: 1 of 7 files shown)


riperiperi:
I think the ScaleFactor parameter and associated logic needs to be removed from `TextureCopyUnscaled`, as before it scaled the width/height found in the Info. Now, I think it will scale it twice.

gdkchan:
> 
> 
> I think the `scaleFactor` parameter and associated logic needs to be removed from `TextureCopyUnscaled`, as before it scaled the width/height found in the Info. Now, I think it will scale it twice.

Fixed, I also changed the ceil operation to use `MathF` rather than `Math` as that should be faster, and it doesn't seems necessary to do the operation with double precision.

riperiperi (approved):
Looks good, should make things easier for unscaled PBO copy too.

Thog (approved):

PR #1647, opened 2020-10-30
Separate zeta from color formats

gdkchan:
Those zeta (depth-stencil) formats are only supposed to be valid for depth-stencil render targets. However they all shared the same enum, so it was considered valid for copy textures and even color render targets too, which does not match the GPU behavior at all. This changes move the depth-stencil formats to a separate enum, used by depth-stencil render targets only. While I don't expect this to actually fix any graphical issue, it is technically more accurate.


riperiperi (approved):

Thog (approved):

Thog:
Need a rebase

PR #1648, opened 2020-10-30
Remove unused texture and sampler pool invalidation code

gdkchan:
This is now unused (I think it was unused for a long time actually, I remember introducing them at one point, but in the end I did not use it, and after range tracking, I don't think it will be necessary), so we can remove it. Or, we can keep it too if anyone thinks it could be useful in the future...


Thog (approved):
Seems fine to remove to me 👍

riperiperi (approved):

PR #1650, opened 2020-10-30
CPU (A64): Add FP16/FP32 fast paths (F16C Intrinsics) for Fcvt_S, Fcvtl_V & Fcvtn_V Instructions. Now HardwareCapabilities uses CpuId.

LDj3SNuD:
Nits.

Tests performed successfully in both debug and release mode (for all instructions involved).


ARMeilleure/Instructions/InstEmitSimdCvt.cs
```diff
@@ -60,21 +60,48 @@ public static void Fcvt_S(ArmEmitterContext context)
             }
             else if (op.Size == 0 && op.Opc == 3) // Single -> Half.
             {
-                Operand ne = context.VectorExtract(OperandType.FP32, GetVec(op.Rn), 0);
+                if (Optimizations.UseF16c)
+                {
+                    Debug.Assert(!Optimizations.ForceLegacySse);
```
gdkchan, line 65: Why the assert?

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -4,6 +4,13 @@ namespace ARMeilleure.CodeGen.X86
 {
     static class HardwareCapabilities
     {
+        static HardwareCapabilities()
+        {
+            (_, _, int ecx, _) = X86Base.CpuId(0x00000001, 0x00000000);
```
gdkchan, line 9: Now that we can use the instruction easily, we could just have a enum or 2 with all the feature flags, and basically just do `Enum.HasFlag(FeatureFlags.Avx);` etc. It might be better for consistency, and because they might decide to return false on the `System.Runtime.Intrinsics.X86*` properties even if the CPU supports the ISA extensions, the reasons including:
- CPU issues, like abysmal PDEP performance on AMD.
- We might be running on a runtime that does not support those intrinsics, but that should not stop our own JIT from using those instructions.
None of those are a problem today, but it might be at some point in the future, since its a quite small change anyway I think it could be worth it.

gdkchan (commented):
Nice to see fast path for those instructions, the .NET 5 update might be better as a separate PR, so if something regresses due to the update its easier to track what caused it.

ARMeilleure/Instructions/InstEmitSimdCvt.cs
LDj3SNuD: https://www.felixcloutier.com/x86/vcvtph2ps
https://www.felixcloutier.com/x86/vcvtps2ph

Only Vex encoding, If someone plays with `Optimizations.ForceLegacySse`, the CPU doesn't take it well.

gdkchan: I see, puting the assert on the x86 backend could avoid repeating it on a few places (and also would be consistent with what is done to other VEX-only instructions).

LDj3SNuD: It looks like a change that involves unnecessary code redundancy, currently using IntrinsicType.BinaryImm and IntrinsicType.Unary.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
gdkchan: Btw, would be also good to check `X86Base.IsSupported`. If not, then all the `Supports*` would be == false.

Thog:
I'm going to do a PR for the update of .NET for CI but I don't plan to merge it until a first stable release launch in November 10.

LDj3SNuD:
For those who want to try Ryujinx on `.NET 5.0`: https://ci.appveyor.com/project/gdkchan/ryujinx/builds/36063885/artifacts

shukenmg:
@LDj3SNuD What's the main difference after upgrading to Net 5.0? Performance?

LDj3SNuD:
More stability and enhanced user experience.

shukenmg:
@LDj3SNuD Oh nice. Thanks, gonna try it out now.

shukenmg:
I am noticing a small regression in AOC using Net 5.0, the attack effects are not rendered correctly.
This is on version 1.0.5584 
image
But on your build the effects are transparent.
Is this because the build is not built with the latest commits included?

LDj3SNuD:
For those who want to try Ryujinx on `.NET 5.0`: https://ci.appveyor.com/project/gdkchan/ryujinx/builds/36073152/artifacts (rebased)

shukenmg:
@LDj3SNuD Nice, thanks.

LDj3SNuD:
Up.

LDj3SNuD:
Depends on https://github.com/Ryujinx/Ryujinx/pull/1694

LDj3SNuD:
Completely remove CpuId, from the Assembler?

gdkchan:
> 
> 
> Completely remove CpuId, from the Assembler?

I guess that's fine.

LDj3SNuD:
Rebased due to .NET 5.0.
2 review requests from the 2 reviewers.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -1,20 +1,60 @@
...
-        public static bool SupportsPclmulqdq => Pclmulqdq.IsSupported;
-        public static bool SupportsFma => Fma.IsSupported;
-        public static bool SupportsPopcnt => Popcnt.IsSupported;
-        public static bool SupportsAesni => Aes.IsSupported;
-        public static bool SupportsAvx => Avx.IsSupported;
+        static HardwareCapabilities()
+        {
+            if (!X86Base.IsSupported)
+            {
+                throw new Exception();
```
gdkchan, line 12: I think this should just return. Why does it throw a exception?

riperiperi (approved):
Looks good to me!

gdkchan (approved):
lgtm, thanks!

PR #1655, opened 2020-11-01
Support 3D BC4 and BC5 compressed textures

gdkchan:
On OpenGL, those compressed formats can't be used with 3D textures (attempting to create a storage with them returns `GL_INVALID_OPERATION`), so as a solution, this now decompresses them on the CPU and uses a different format.
Downsides:
- Decompression may be potentially slow, depending on the texture size.
- If the texture is written from the GPU it will not work, it can be made to work by forcing flushes, but that is slow.
A potential solution for the above problems is decompressing on the GPU, but this needs other changes first.

Fixes water on Hyrule Warriors: Ages of Calamity Demo.
Before:
image
After:
image


Thog (approved):

riperiperi (approved):
LGTM. Very weird that OpenGL driver doesn't like this.

Just for completeness in case anyone's looking at a similar issue on the future, Mesa drivers seem to also be missing this on all other DXT formats. Only BC4 and BC5 are done here as the others are fine on NV proprietary drivers.

Ryujinx.Graphics.Texture/BCnDecoder.cs
```diff
@@ -0,0 +1,242 @@
...
+            ReadOnlySpan<ulong> data64 = MemoryMarshal.Cast<byte, ulong>(data);
+
+            Span<byte> rPal = stackalloc byte[8];
+            Span<byte> gPal = stackalloc byte[8];
+
+            int baseOOffs = 0;
+
+            for (int l = 0; l < levels; l++)
+            {
+                int w = BitUtils.DivRoundUp(width, 4);
```
riperiperi, line 119: `BlockWidth`?

PR #1657, opened 2020-11-01
Support res scale on images, correctly blacklist for SUST, move logic out of backend.

riperiperi:
Resolution scaling is now possible for bound images using SULD, and using SUST will force a blacklist. Additionally, scale calculation logic for texture/images has been moved out of the backend into the GPU, as it should be backend agnostic.

When res scaled:
- Fixes normal-map decals in Xenoblade Chronicles: DE.
- Should cause Paper Mario and Dead or Alive to force to 1x resolution instead of just breaking.
- May fix some other weird issues.

Before (2x):
image

After (2x):
image

Should have no effect when res scale is disabled. May do less work when res scale is not needed in the shader, as a result of some restructuring.


gdkchan (approved):
lgtm, thanks. Nice to see some scaling logic moved out of the backend.

Thog (approved):
lgtm!

PR #1658, opened 2020-11-01
Add seamless cubemap flag in sampler parameters.

riperiperi:
Adds the seamless cubemap flag for samplers, which allows linear blending over cubemap edges. This removes hard edges from cubemaps, and specifically improves really low res cubemaps (used for glossy surfaces).

As seen on nouveau, and supposedly set on every texture by NVN.
https://github.com/envytools/envytools/blob/master/rnndb/graph/g80_texture.xml#L389

There is also a global setting for this, but since the per-texture one is set by NVN I just used that instead. Other guest apis may expect the global setting, which should be added separately.

Before:
image

After:
image

(this does not fix missing cubemaps in mario kart 8)


gdkchan (approved):
lgtm, pretty straightforward.

Thog (approved):
lgtm

PR #1666, opened 2020-11-05
Use explicit buffer and texture bindings on shaders

gdkchan:
This change uses explicit binding points on the shader code, rather than calling the OpenGL `glGet*Location` functions with the texture/image/buffer name and assigning a binding point on the backend. The main advantage here is making the backend cleaner and more flexible.

Advantages:
- Code is simpler/smaller.
- Binding should be a little bit faster, as it no longer needs to get the binding point from index/stage, the binding point is passed directly. A few other things were optimized aswell (for example, it passes all buffers to be bound at once instead of one at a time).

Other changes:
- Buffers are properly unbound instead of just keeping the last binding.

I recommend testing on a few games just to be sure nothing regressed.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -6,6 +6,7 @@ using Ryujinx.Graphics.Gpu.Shader;
 using Ryujinx.Graphics.Gpu.State;
 using Ryujinx.Graphics.Shader;
 using System;
+using System.Linq;
 using System.Runtime.InteropServices;
 
 namespace Ryujinx.Graphics.Gpu.Engine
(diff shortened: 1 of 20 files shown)


Thog (approved):
lgtm, it surely help a lot to simplify some logics :+1: 
Still need to do more testing on games

EmulationFanatic:
Seems to work fine in my testing

Ryujinx.Graphics.Gpu/Engine/Methods.cs
```diff
@@ -1005,6 +1009,10 @@ private void UpdateShaderState(GpuState state)
 
                 if (info == null)
                 {
+                    TextureManager.SetGraphicsTextures(stage, Array.Empty<TextureBindingInfo>());
+                    TextureManager.SetGraphicsImages(stage, Array.Empty<TextureBindingInfo>());
+                    BufferManager.SetGraphicsStorageBufferBindings(stage, null);
+                    BufferManager.SetGraphicsUniformBufferBindings(stage, null);
                     continue;
```
Leischii, line 1016: New line missing

riperiperi (approved):
Very nice! No objections here.

PR #1667, opened 2020-11-05
Simplify logic for bindless texture handling

gdkchan:
Currently the shader decompiler has a pass to turn bindless texture accesses into regular texture accesses. This is necessary because emulating bindless texture access on the host is complicated, and we can't just access an arbitrary texture from a guest handle easily.

Currently the information about bindless textures is stored on the texture descriptor. It includes 3 fields just for bindless related information: `IsBindless` indicating if the texture is used on a bindless access, `CbufSlot` with the slot of the constant buffer where the handle is located, and `CbufOffset` with the offset on the constant buffer. There is also a `Handle` field that has the offset of the handle on the "texture constant buffer" for regular texture accesses.

This change removes 2 of those fields:
- `IsBindless` was removed and all the entries are handled the same way, regardless of the texture being used for bindless or regular access.
- `CbufOffset` was removed, now `Handle` servers the same purpose.

The main difference is that the handle for regular textures will be always located on the "texture constant buffer", while for bindless, the handle may be located on any constant buffer. For that, the "CbufSlot" field is used. Now it can have a new value of `-1` that indicates that the handle should be loaded from the "texture constant buffer", meanwhile when it is a non-negative number, it indicates the constant buffer slot to use directly.

Other improvements that this allowed:
- Finding bindless handles for images at any constant buffer is now possible (before it was limited to constant buffer 2 which is what NVN uses as "texture constant buffer").
- It now returns 0 on attempts to use bindless texture with a unknown handle source, and discards the value on image stores for those cases. Before it would just sample garbage and write to some random texture. Note that this might change behavior of games that uses bindless textures, for example, Super Mario Party board shadows are now a uniform color instead of random garbage. I believe this behavior is more desirable as it produces predictable results.

Other small changes (might not belong here but I don't think it's worth opening a separate PR just for them):
- Rename pass that rewrites instructions in order to compensate for missing hardware functionality to `Rewriter`, I think that makes more sense.
- Remove stray space on one file.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
@@ -15,7 +15,13 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl.Instructions
 
             bool isBindless = (texOp.Flags & TextureFlags.Bindless) != 0;
 
-            bool isArray   = (texOp.Type & SamplerType.Array) != 0;
+            // TODO: Bindless texture support. For now we just return 0/do nothing.
+            if (isBindless)
+            {
+                return texOp.Inst == Instruction.ImageLoad ? NumberFormatter.FormatFloat(0) : "// imageStore(bindless)";
+            }
+
+            bool isArray   = (texOp.Type & SamplerType.Array)   != 0;
             bool isIndexed = (texOp.Type & SamplerType.Indexed) != 0;
 
             string texCall = texOp.Inst == Instruction.ImageLoad ? "imageLoad" : "imageStore";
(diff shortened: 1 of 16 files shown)


Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
```diff
@@ -79,7 +85,10 @@ string ApplyScaling(string vector)
                     flags |= TextureUsageFlags.ResScaleUnsupported;
                 }
 
-                context.ImageDescriptors[index] = context.ImageDescriptors[index].SetFlag(flags);
+                if (!isBindless)
```
riperiperi, line 88: Will this condition matter if it returns early above?

riperiperi (approved):
LGTM!

Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
gdkchan: It doesn't matter, I just left it there make things easier when bindless support is eventually added.

Thog (approved):
lgtm

PR #1670, opened 2020-11-06
Correct BPP of buffer to texture copies

gdkchan:
The sizes and regions that are used with the GPU copy engine are pre-multiplied by the amount of bytes per pixel (checked on NVN and deko3d). This means that we shouldn't be passing the BPP value as it was already multiplied with the coords and sizes. This change fixes this by always setting the BPP to 1, which should have the same effect as passing the correct BPP and dividing the width/height and the coords by the BPP.

The component size/count fields are now unused. The copy engine has a feature officially called "remap" that can be used to change the component order. I believe this is where the component sizes and counts come into play, as we don't emulate this feature yet, we don't need to use them.

This change fixes memory corruption caused by too much data being copied in some UE4 games. With this change, Bravely Default II DEMO now goes further for example:
image
Unfortunately it still crashes trying to start a new game.
Remothered Broken Porcelain also goes further, but crashes with the same error as Bravely Default:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyBuffer.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyBuffer.cs
@@ -2,7 +2,6 @@ using Ryujinx.Common;
 using Ryujinx.Graphics.Gpu.State;
 using Ryujinx.Graphics.Texture;
 using System;
-using System.Runtime.Intrinsics;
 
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Nice find. 

Would recommend testing:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/1898

Thog (approved):

PR #1671, opened 2020-11-06
Do not match aligned sizes for buffer texture targets.

riperiperi:
This should fix a random crash in Hyrule Warriors, and potentially other games that use buffer textures. This crashed before as we don't support copying buffer textures.

Buffer textures don't need to do any size change tricks (they're 1D), and their data should be inherited simply by using the same buffer storage.


gdkchan (approved):
lgtm, thanks.

Thog (approved):

PR #1672, opened 2020-11-06
Do not report unmapped pages as dirty

gdkchan:
Some games uses ranges that are partially unmapped for buffers (I don't know if this is normal or a bug to be honest). This change makes it not report unmapped regions as dirty, as attempting to read from those regions will throw. This fixes a regression in Higurashi no Naku Koro ni Hou after range tracking.

### DIFF
--- a/Ryujinx.Cpu/MemoryManager.cs
+++ b/Ryujinx.Cpu/MemoryManager.cs
@@ -1,4 +1,4 @@
-using ARMeilleure.Memory;
+﻿using ARMeilleure.Memory;
 using Ryujinx.Cpu.Tracking;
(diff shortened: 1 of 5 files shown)


Thog (approved):

riperiperi (approved):
This is fine, though there may be an existing bug where mapping a non-dirty virtual region will not reprotect the virtual memory, so the flag will never be set in future.

You may be able to do that by adding `region.UpdateProtection();` after `region.RecalculatePhysicalChildren();` in the `Map()` function in `MemoryTracking.cs`.

riperiperi (approved):

PR #1680, opened 2020-11-07
Synchronize Rasterizer State before Clear

riperiperi:
This fixes shadows sometimes not clearing in Xenoblade 2, as they are usually cleared immediately after a series of commands with Rasterizer Discard enabled.

image


Thog (approved):

gdkchan (approved):
lgtm, nice catch.

PR #1686, opened 2020-11-08
Size hints for copy regions and viewport dimensions to avoid data loss

riperiperi:
Data can be lost if the last sampled size for a texture is smaller than the region it wants to render to, due to the texture being changed to the smaller size. This manifests in games with dynamic resolution as garbage at the bottom right edge, cropping, the image shifting every frame etc. It can also cause data to be lost when copying between textures. This PR attempts to solve this issue by passing in usage hints that indicate the minimum size that the requester would _prefer_, and increase the size to meet it when possible.

Size Hints currently come from two places:
- Textures used as color or depth targets use Viewport 0's dimensions.
- Source and destination textures for copies use the copy region.
  - Source is included because generally the texture needs to be synchronized before copy. Synchronizing with the wrong size could lose data.

Both these hints fit the pattern `Position + Size`, which should result in a size that fully covers the desired region.

The rules for using the size hints are as follows:
- If the texture requested is for a sampler, trust and resize to the dimensions in the given texture info, as we did before.
- If this is not a case and a size hint is not provided, do nothing.
- If it is, grow the existing texture's size to contain the size hint, but with dimensions in the given info as an upper bound.
  - If neither dimension is affected, do nothing.
  - Otherwise resize the texture.

Additionally, when a texture is resized at least once, an additional check will be done when retrieving it from a texture pool to check if its size doesn't match the texture descriptor. This fixes an issue where multiple entries on the pool could point to the same texture at different sizes, without changing to the correct size on use.

The most visible improvements will be in Luigi's Mansion 3 and Kingdom Hearts: Melody of Memory. This also helps texture streaming in unreal engine 3/4 games, and cubemap generation in MK8, though the improvements will be hard to spot through everything else being broken. This PR does not speed up uses of dynamic resolution, it just fixes graphics in games that use it.

I'd recommend testing in a few other games to make sure nothing weird happens.

Before:
https://gyazo.com/387bdbefacad6fb56768eddbe85613a3

After:
https://gyazo.com/ee8dff8573d34883d1d126a98497fb90

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
@@ -1,6 +1,7 @@
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Image;
 using Ryujinx.Graphics.Gpu.State;
+using Ryujinx.Graphics.Texture;
 using System;
 
 namespace Ryujinx.Graphics.Gpu.Engine
@@ -19,9 +20,30 @@ namespace Ryujinx.Graphics.Gpu.Engine
             var dstCopyTexture = state.Get<CopyTexture>(MethodOffset.CopyDstTexture);
             var srcCopyTexture = state.Get<CopyTexture>(MethodOffset.CopySrcTexture);
 
+            var region = state.Get<CopyRegion>(MethodOffset.CopyRegion);
+
+            var control = state.Get<CopyTextureControl>(MethodOffset.CopyTextureControl);
+
+            int srcX1 = (int)(region.SrcXF >> 32);
+            int srcY1 = (int)(region.SrcYF >> 32);
+
+            int srcX2 = (int)((region.SrcXF + region.SrcWidthRF * region.DstWidth) >> 32);
+            int srcY2 = (int)((region.SrcYF + region.SrcHeightRF * region.DstHeight) >> 32);
+
+            int dstX1 = region.DstX;
+            int dstY1 = region.DstY;
+
+            int dstX2 = region.DstX + region.DstWidth;
+            int dstY2 = region.DstY + region.DstHeight;
+
+            // The source and destination textures should at least be as big as the region being requested.
+            // The hints will only resize within alignment constraints, so out of bound copies won't resize in most cases.
+            var srcHint = new Size(srcX2, srcY2, 1);
+            var dstHint = new Size(dstX2, dstY2, 1);
+
             var srcCopyTextureFormat = srcCopyTexture.Format.Convert();
 
-            Texture srcTexture = TextureManager.FindOrCreateTexture(srcCopyTexture, srcCopyTextureFormat);
+            Texture srcTexture = TextureManager.FindOrCreateTexture(srcCopyTexture, srcCopyTextureFormat, true, srcHint);
 
             if (srcTexture == null)
             {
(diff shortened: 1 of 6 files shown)


Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -304,7 +304,7 @@ public static Size GetAlignedSize(TextureInfo info, int level = 0)
         {
             int width = Math.Max(1, info.Width >> level);
             int height = Math.Max(1, info.Height >> level);
-
+
```
gdkchan, line 307: nit: Remove spacing on empty line.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -911,6 +905,44 @@ public Texture FindOrCreateTexture(TextureInfo info, TextureSearchFlags flags =
...
+                    texture.ChangeSize(info.Width, info.Height, info.DepthOrLayers);
+                }
+            }
+            else if (sizeHint != null)
+            {
+                // A size hint indicates that data will be used within that range, at least.
+                // If the texture is smaller than the size hint, it must be enlarged to meet it.
+                // The maximum size is provided by the requested info, which generally has an aligned size.
+
+                int width = Math.Min(Math.Max(sizeHint.Value.Width, texture.Info.Width), info.Width);
```
gdkchan, line 936: Could probably use `Math.Clamp` instead here.

gdkchan (commented):
Looks good, it might be worth noting that NVN only ever aligns the width, so not aligning the height could simplify things a bit (and reduce the number of cases that trigger the size changes). Aligning both width and height is technically more correct though.

Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
```diff
@@ -100,7 +106,7 @@ private void CopyTexture(GpuState state, int argument)
             {
                 srcCopyTexture.Height++;
 
-                srcTexture = TextureManager.FindOrCreateTexture(srcCopyTexture, srcCopyTextureFormat, srcTexture.ScaleMode == TextureScaleMode.Scaled);
+                srcTexture = TextureManager.FindOrCreateTexture(srcCopyTexture, srcCopyTextureFormat, srcTexture.ScaleMode == TextureScaleMode.Scaled, srcHint);
                 if (srcTexture.ScaleFactor != dstTexture.ScaleFactor)
```
Leischii, line 110: Missing new line

gdkchan: I don't think we have any rule for where new lines should be inserted right now.

Leischii: Afaik if statements should be seperated by new lines with a variable declaration infront of it. If I'm wrong, just ignore this comment

Thog (approved):
Appart from gdkchan's comments, lgtm.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
riperiperi: Min can sometimes be greater than Max, so clamp can't be used. However, I noticed that the order I do min and max in could lead to some textures shrinking, which it shouldn't do for textures that aren't for sampler. I've changed the order so that it can only increase the existing size.

gdkchan: Ok, at least the suggestion was useful in the end I guess 😄

gdkchan (approved):
lgtm, thanks.

PR #1687, opened 2020-11-08
Implement ATOM shader instruction

gdkchan:
This implements the ATOM shader instruction, and fixes negative offsets on the ATOMS shader instructions (the offset was not being sign-extended).

The ATOM instruction is similar to ATOMS, but operates on global memory. It can be used to do atomic operations on memory, such as GLSL `atomicAdd` and friends.

I know a bunch of games that uses this (Fire Emblem Three Houses is one of them), but I don't know any specific issue that this change fixes.

Note: The compare and swap variant is still not implemented, but it also doesn't seems to be implemented for the other atomic instructions.

### DIFF
--- a/Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
+++ b/Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
@@ -61,6 +61,8 @@ namespace Ryujinx.Graphics.Shader.Instructions
         {
             OpCodeAtom op = (OpCodeAtom)context.CurrOp;
 
+            ReductionType type = (ReductionType)op.RawOpCode.Extract(49, 2);
+
             int sOffset = (op.RawOpCode.Extract(28, 20) << 12) >> 12;
 
             (Operand addrLow, Operand addrHigh) = Get40BitsAddress(context, op.Ra, op.Extended, sOffset);
@@ -71,7 +73,7 @@ namespace Ryujinx.Graphics.Shader.Instructions
                 context,
                 Instruction.MrGlobal,
                 op.AtomicOp,
-                op.Type,
+                type,
                 addrLow,
                 addrHigh,
                 value);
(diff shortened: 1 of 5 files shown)


riperiperi (approved):
👍

Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
```diff
@@ -57,13 +57,37 @@ public static void Ast(EmitterContext context)
...
+
+            (Operand addrLow, Operand addrHigh) = Get40BitsAddress(context, op.Ra, op.Extended, sOffset);
+
+            Operand value = GetSrcB(context);
+
+            Operand res = EmitAtomicOp(
+                context,
+                Instruction.MrGlobal,
+                op.AtomicOp,
+                op.Type,
```
Thog, line 74: Wrong, this is reusing bits 28 to 30 here, https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Shader/Decoders/OpCodeAtom.cs#L27

the type is stored starting at bit 49 with the given possible values for ATOM (opcode: 0xed)

```cs
enum AtomType
{
        U32 = 0,
        S32 = 1,
        U64 = 2,
        F32 = 3,
}

File: Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs [hunk 2]
```diff
@@ -57,13 +57,37 @@ public static void Ast(EmitterContext context)
...
+            int sOffset = (op.RawOpCode.Extract(28, 20) << 12) >> 12;
+
+            (Operand addrLow, Operand addrHigh) = Get40BitsAddress(context, op.Ra, op.Extended, sOffset);
+
+            Operand value = GetSrcB(context);
+
+            Operand res = EmitAtomicOp(
+                context,
+                Instruction.MrGlobal,
+                op.AtomicOp,
```
Thog, line 73: NOTE here:  ATOM (opcode: 0xed) also have a possible "safe add" (0b1010) operation, would be nice to add it to AtomicOp and add a comment about it. (only valid for this opcode)

Thog (changes requested):
Need changes

Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
gdkchan: Nice catch, I will fix it.

Thog (approved):

PR #1695, opened 2020-11-11
Simplify depth test state updates

gdkchan:
The OpenGL spec states:
> If there is no depth buffer, it is as if the depth buffer test always passes.

Currently, there is some code that disables depth test if there's no depth buffer. This is redundant as the host API will already do that. This change will just set depth test directly instead which reduces the number of API calls and simplifies the code a bit.

Should not cause any change on games.


Thog (approved):
lgtm

Thog:

gdkchan:
Rebased.

riperiperi (approved):
LGTM.

PR #1696, opened 2020-11-12
Fix: Update performance mode when docked mode changes.

riperiperi:
Performance mode was stuck on "default", which caused SMO to always run as if it were in handheld mode. This fixes the system init and docked mode switch to also update the performance mode.


AcK77 (approved):
Thanks for catched and fixed that!

gdkchan (approved):

PR #1697, opened 2020-11-12
Standardize to camel-case: option to check for updates on launch

EmulationFanatic:
This PR updates the "Check for updates on launch" to read instead "Check For Updates On Launch" to be consistent with camel-case options elsewhere in the UI.


EmulationFanatic:
Will open a new PR with just the text changes since Glade did something unintended in my changes.

PR #1703, opened 2020-11-12
Use "Screen Scissor" as size hint for render targets

riperiperi:
"Screen scissor" is the minimum size of all render targets, and is set when any render target is bound on NVN or OpenGL. Since it works on all render texture's _real_ sizes, it is therefore more reliable than viewport 0's width, and is actually set before clear.

This fixes a regression with Hyrule Warriors: Age Of Calamity's cubemaps, which did not set viewport dimensions before clear, so the hint was inaccurate. This resulted in attempting to create a cubemap with rectangular sides (16x8), which is logically and physically impossible. (also it just fails)

A better solution for size rules regarding cubemap views will likely be done later.


gdkchan (approved):
Pretty straightforward, lgtm.

riperiperi:
Regression:
image

Fixed:
image

Thog (approved):

jduncanator (approved):

PR #1704, opened 2020-11-12
Use backup when PTC compression is corrupt

riperiperi:
We have some nice checks for PTC to make sure the data is valid before we load and use it, and a backup to try in case it isn't.

Unfortunately, none of that is used right now because most of the time it throws attempting to decompress an incomplete stream.

This PR catches decompression errors and treats it as a loading failure, which lets it fall back to the backup instead of crashing. The change has been made for both the cache and the info.

### DIFF
--- a/ARMeilleure/Translation/PTC/Ptc.cs
+++ b/ARMeilleure/Translation/PTC/Ptc.cs
@@ -196,7 +196,16 @@ namespace ARMeilleure.Translation.PTC
             {
                 int hashSize = md5.HashSize / 8;
 
-                deflateStream.CopyTo(stream);
+                try 
+                {
+                    deflateStream.CopyTo(stream);
+                } 
+                catch
+                {
+                    InvalidateCompressedStream(compressedStream);
+
+                    return false;
+                }
 
                 stream.Seek(0L, SeekOrigin.Begin);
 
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
lgtm, thanks.

Thog:

riperiperi:
Rebased.

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -196,7 +196,16 @@ private static bool Load(string fileName)
             {
                 int hashSize = md5.HashSize / 8;
 
-                deflateStream.CopyTo(stream);
+                try
```
LDj3SNuD, line 199:
```suggestion
                try
```

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -196,7 +196,16 @@ private static bool Load(string fileName)
             {
                 int hashSize = md5.HashSize / 8;
 
-                deflateStream.CopyTo(stream);
+                try 
+                {
+                    deflateStream.CopyTo(stream);
+                }
```
LDj3SNuD, line 202:
```suggestion
                }
```

ARMeilleure/Translation/PTC/PtcProfiler.cs
```diff
@@ -120,7 +120,16 @@ private static bool Load(string fileName)
             {
                 int hashSize = md5.HashSize / 8;
 
-                deflateStream.CopyTo(stream);
+                try
+                {
+                    deflateStream.CopyTo(stream);
+                }
```
LDj3SNuD, line 126:
```suggestion
                }
```

LDj3SNuD (approved):

PR #1707, opened 2020-11-13
Don't warn when debug feature is enabled in debug builds

Xpl0itR:
This stops the GUI from asking you to disable debug features in debug builds


riperiperi (approved):
😌

Thog (approved):

PR #1711, opened 2020-11-14
Allow copy destination to have a different scale from source

riperiperi:
Will result in more scaled copy destinations, but allows scaling in some games that copy textures to the output framebuffer.

Also, textures with more than one level are no longer blacklisted. This has been changed to more than 3 levels. (indicates a game texture)

- Res scaling now works in Link’s Awakening (causes shadow acne tho)
- Res scaling no longer blacklists in some cutscenes in Super Mario Odyssey.

Needs further testing, so Draft for now.

### DIFF
--- a/Ryujinx.Graphics.GAL/Extents2D.cs
+++ b/Ryujinx.Graphics.GAL/Extents2D.cs
@@ -1,3 +1,5 @@
+using Ryujinx.Common;
+
 namespace Ryujinx.Graphics.GAL
 {
     public struct Extents2D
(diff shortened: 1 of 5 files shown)


Thog:

gdkchan (approved):
lgtm, but looks like `GetScaledHostTexture` does not handle mipmaps (and only copies the first level), is that fine?

riperiperi:
> 
> 
> lgtm, but looks like `GetScaledHostTexture` does not handle mipmaps (and only copies the first level), is that fine?

That is true, though I suppose there's not really a function to do scaled copy between levels. I'll look into it.

gdkchan:
I guess the current copy function can be extended to blit multiple levels aswell. It can also be extended to multiple multiple array layers, then you can remove the function to blit multiple arrays on the `Texture` class.

riperiperi:
Update on scaled copies for switching res scale: The backend now uses glFramebufferTextureLayer to copy multiple texture layers, copies levels individually (and scales the regions).

Remove CopyArrayScaled, since the backend copy handles it now.

gdkchan (approved):
Still looking good.

Thog (approved):
lgtm

gdkchan:
Needs rebase.

PR #1714, opened 2020-11-16
Improvements with new .NET 5 functions or bugfixes

gdkchan:
This makes some small improvements to the code made possible by .NET 5, either because it fixed bugs that would otherwise prevent some code from being used, or because it introduced new functions that can replace existing code.


riperiperi (approved):
Cool, didn't know `SkipLocalsInit` was a thing.

Thog (approved):
lgtm, TIL I learned for SkipLocalsInit

PR #1688, opened 2020-11-08
am/lbl/hid/pctl: Enabled VR Rendering

AcK77:
This PR enable VR rendering on games which support it through the Toy-Con VR Goggles.

Please remember Ryujinx currently don't support console SixAxis sensor and for now, in some games, the view can't be moved.
Some games can enable VR only in Handled mode, BotW can enable it in Docked too.

Everything is implemented accordingly to RE:
- am: ICommonStateGetter: SetVrModeEnabled, BeginVrModeEx, EndVrModeEx (Closes #1605).
- lbl: ILblController: SetBrightnessReflectionDelayLevel, GetBrightnessReflectionDelayLevel, SetCurrentAmbientLightSensorMapping, GetCurrentAmbientLightSensorMapping, SetCurrentBrightnessSettingForVrMode, GetCurrentBrightnessSettingForVrMode, EnableVrMode, DisableVrMode, IsVrModeEnabled.
- pctl: IParentalControlService: ConfirmStereoVisionPermission, ConfirmStereoVisionRestrictionConfigurable, GetStereoVisionRestriction, SetStereoVisionRestriction, ResetConfirmedStereoVisionPermission, IsStereoVisionPermitted.
- hid: IHidServer: ResetSevenSixAxisSensorTimestamp is stubbed because we don't support console SixAxisSensor for now.

Maybe we could add a setting later to enable or disable VR. But I think it's fine to keep this always available since you have to enable it in games.

SMO:
image
image
image
Capture

Zelda BotW:
image
image
image
Capture2

PR #1715, opened 2020-11-16
am: Fix VR enabled by default

AcK77:
This PR fix an issue I've made in #1688 which is enabled VR as default.
It could cause rendering issues in games when VR mode isn't used, as users have reported in Smash (where VR rendering don't seems to works fine anyway).

Before
image

After
image


gdkchan:
How would one enable VR after this change?

AcK77:
Games enable VR by calling `SetVrModeEnabled` with `true` as arg.
I've tested SMO and BotW with this change and both working fine.
Smash too but the rendering is a bit glitched.

gdkchan (approved):
lgtm, but makes me wonder why it was enabled by default before.

riperiperi (approved):
I told you dog!

PR #1716, opened 2020-11-16
Propagate zeta format properly

gdkchan:
Fix a regression introduced on #1647 where the wrong format would be propagated to the destination texture. Should fix performance regression on Zelda Links Awakening.


riperiperi (approved):
Nice catch.

Thog (approved):

PR #1717, opened 2020-11-16
shader cache: Fix invalid virtual address clean up

Thog:
This fix an issue causing the virtual address of texture descriptors to
not be cleaned up when caching and instead cleaning texture format and swizzle.

This should fix duplicate high duplication in the cache for certain
games and possible texture corruption issues.

**THIS WILL INVALIDATE ALL SHADER CACHE LEVELS CONSIDERING THE NATURE OF THE ISSUE**

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/ITextureDescriptor.cs
+++ b/Ryujinx.Graphics.Gpu/Image/ITextureDescriptor.cs
@@ -1,0 +1,10 @@
+﻿namespace Ryujinx.Graphics.Gpu.Image
+{
+    interface ITextureDescriptor
+    {
+        public uint UnpackFormat();
+        public TextureTarget UnpackTextureTarget();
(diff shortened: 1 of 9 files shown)


riperiperi (approved):
😔 Will test later. This should greatly reduce shader cache size.

riperiperi:
Can confirm that this resolves issues with shaders caching multiple times.

gdkchan (approved):

gdkchan:
After thinking a bit more about this problem, I think it's better to just store in the cache the fields that are relevant for shader translation right now, to avoid similar issues in the future. If we need to invalidate peoples caches for this change, it's better to do it now rather than cause another invalidation in the future.

riperiperi (approved):
LGTM, new subset of descriptor looks good and works well in the games I've tested.

PR #1718, opened 2020-11-16
shader cache: Fix possible race causing crashes on manifest at startup

Thog:
This fix a misplace function call ending up causing possibly two write
on the cache.info at the same time.


gdkchan (approved):

riperiperi (approved):
Can confirm that this fixes the crash I was having.

riperiperi (approved):

PR #1721, opened 2020-11-17
Fix buffer to texture copy with remap enabled

gdkchan:
Fix bug introduced on #1670 that affects homebrew using nouveau (and maybe OpenGL games) where it wouldn't copy the full texture data. The other games fixed on the aforementioned PR are still working too (tested one of them).


riperiperi (approved):
Looking good here.

Thog (approved):

PR #1726, opened 2020-11-18
gui: Toggle Docked/Handheld mode and VSync by clicking status bar

AcK77:
This PR just add a way to toggle the docked and the handheld mode by clicking the label in the status bar.
Nothing more.

EDIT: Due to some recommandations, I've added the same behavior for the VSync setting, even if we don't recommand to change it when everything works fine.

docked

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -117,9 +117,9 @@ namespace Ryujinx.Ui
                 End(null);
             }
 
-            _virtualFileSystem = VirtualFileSystem.CreateInstance();
+            _virtualFileSystem      = VirtualFileSystem.CreateInstance();
             _userChannelPersistence = new UserChannelPersistence();
-            _contentManager    = new ContentManager(_virtualFileSystem);
+            _contentManager         = new ContentManager(_virtualFileSystem);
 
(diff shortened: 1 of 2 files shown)


Xpl0itR (approved):
lgtm

Ryujinx/Ui/MainWindow.cs
```diff
@@ -848,6 +848,12 @@ private void Row_Activated(object sender, RowActivatedArgs args)
             LoadApplication(path);
         }
 
+        private void DockedMode_Clicked(object sender, ButtonReleaseEventArgs args)
+        {
+            ConfigurationState.Instance.System.EnableDockedMode.Value = !ConfigurationState.Instance.System.EnableDockedMode.Value;
+            ConfigurationState.Instance.ToFileFormat().SaveConfig(Program.ConfigurationPath);
```
Thog, line 854: This should be volatile

```suggestion
```

Thog (changes requested):

Thog (approved):

Ryujinx/Ui/MainWindow.cs
jduncanator: I agree 👍

jduncanator (approved):

PR #1732, opened 2020-11-18
Perform Compressed<->Uncompressed copies using Pixel Buffer Objects

riperiperi:
This fixes issues where data was lost when copying between non-compressed textures, and a compressed texture with a size that was not block aligned. It also retains slicing behaviour when copying between textures with different size.

- Fixes (most) button and large UI text in Splatoon 2. Some text may still be missing or out of date due to the emulatedViewParent stuff only triggering once. (will be resolved in future)
  - _A copy was added when an emulated view is created, which helps a bit._
- Fixes cubemap mip levels below 4x4 not being copied in Mario Kart 8. Does not have any visual impact right now.
- May improve some games with black small mipmap levels, such as UE3 and UE4 games. It does not fix them entirely.

Before:
image

After:
image

Before:
image

After:
image
(i think some textures are being lost to impossible texture resizes. also the texture swaps are something else entirely)

There is also an additional change that fixes aligned size calculation for compressed textures. It was dividing by the block size twice when getting gob block sizes in `GetBlockLinearAlignedSize`, now it only does so once.

The PBO 2D copy also supports copying between two textures of different sizes, where it will pick the minimum size of the two textutres to copy across. This is done in the following way:
- Dump the entire source texture
- Use GL_UNPACK_ROW_LENGTH and GL_UNPACK_IMAGE_HEIGHT to set the dimensions of the source texture (may need to be converted to block or pixel size based on the destination)
  - Use GL_UNPACK_COMPRESSED_BLOCK_WIDTH and friends to the destination's values to allow the unpack parameters to work on compressed textures.
- Read the texture from the buffer with the copy size used in the TexSubImage call. (not destination size)
- Reset unpack parameters afterwards.

None of the unpack parameters are set if the source and destination size match - it's only for sliced copies.

PS: No Shahil/20191, this does not fix Xenoblade issues.


legend800:
Fixes Yoshi's Craft World black objects and the missing map path dashes:

2020-11-18 18_44_26-Ryujinx 1 0 5762 - Yoshi's Crafted World v1 0 1 (01006000040C2000) (64-bit)

2020-11-18 18_45_59-Ryujinx 1 0 5762 - Yoshi's Crafted World v1 0 1 (01006000040C2000) (64-bit)

Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
```diff
@@ -80,6 +81,116 @@ public TextureCopy(Renderer renderer)
...
+
+            for (int level = 0; level < levels; level++)
+            {
+                // Stop copy if we are already out of the levels range.
+                if (level >= srcInfo.Levels || dstLevel + level >= dstInfo.Levels)
+                {
+                    break;
+                }
+
+                if ((width % blockWidth != 0 || height % blockHeight != 0) && src is TextureView && dst is TextureView)
```
gdkchan, line 157: Could do something like `src is TextureView srcView` to avoid the casts below.

Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
```diff
@@ -147,6 +258,88 @@ public TextureView BgraSwap(TextureView from)
             return to;
         }
 
+        public TextureView PboCopy(TextureView from, TextureView to, int srcLayer, int dstLayer, int srcLevel, int dstLevel, int width, int height)
```
gdkchan, line 261: I think this can be `private static`.

riperiperi: Requires the PBO handle from the instance. I can make it static tho.

gdkchan (approved):
lgtm, thanks. Nice to finally see this fixed.

Thog (approved):
lgtm

PR #1734, opened 2020-11-19
olsc: Add service olsc:u and stub some calls

AcK77:
This PR fix an issue of missing service `olsc:u` added in firmware 10.0.
ACNH 1.6.0 update needs it and some other service calls:

- Initialize
- SetSaveDataBackupSettingEnabled

Both are stubbed since we will not support online savedata backup. 
Checked with RE to return the right result.

ACNH can still be playable now:
image


riperiperi (approved):
LGTM.

Thog (approved):

jduncanator (approved):

PR #1736, opened 2020-11-20
Remove some warnings and cleaning code

AcK77:
This PR fix some warnings:
- SystemInfo which are platform specific, since .NET 5 we have to use `SupportedOSPlatform` attribute.
- `stackalloc` used in a loop could cause a potential stack overflow.
- All not used exceptions have their var removed in Motion client.
- Updater async tasks are currently run synchronously, so I've discarded the await warning by an empty assign.
- `TarInputStream` in the Updater is obsolete now and needs the filename Encoding.

Additionnally to that, I've cleaned up a bit Motion Client code.
BinaryFormatter warnings are still here, I'll let LDj3SNuD handle that.

### DIFF
--- a/Ryujinx.Common/SystemInfo/LinuxSystemInfo.cs
+++ b/Ryujinx.Common/SystemInfo/LinuxSystemInfo.cs
@@ -1,8 +1,10 @@
 using System.IO;
 using System.Linq;
+using System.Runtime.Versioning;
 
 namespace Ryujinx.Common.SystemInfo
 {
+    [SupportedOSPlatform("linux")]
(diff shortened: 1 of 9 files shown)


gdkchan (approved):
lgtm, but would be nice to know from where the OS platform names came from (I suppose we can't just put any name there).

AcK77:
@gdkchan https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer

Thog (approved):

riperiperi (approved):

Xpl0itR (approved):

PR #1738, opened 2020-11-20
am: Cleaning and stub TryPopFromFriendInvitationStorageChannel

### LINKED ISSUE #1571: appletOE IApplicationFunctions: 141 (TryPopFromFriendInvitationStorageChannel) is not implemented

*appletOE* *IApplicationFunctions*: *141* (*TryPopFromFriendInvitationStorageChannel*) is not implemented.</br>

[Switchbrew](https://switchbrew.org/wiki/Applet_Manager_services#TryPopFromFriendInvitationStorageChannel)

### Service description
```
No input, returns an output #IStorage.

This uses the same StorageChannel as #PushToFriendInvitationStorageChannel.

This is exposed by sdknso under nn::friends::. When the storage_size is <0x10, this returns false. data_size = storage_size-0x10; Then the first 0x10-bytes from storage are read to stack, which is later copied to the output Uid on success prior to returning. Then the storage is read into the specified output buffer with storage-offset 0x10, where the size is size = data_size > buf_size ? buf_size : data_size. Reading is skipped if size is zero. Afterwards the used size is written to an output param. On success, true is returned.
```

### Required by: 
https://github.com/Ryujinx/Ryujinx-Games-List/issues/2552

AcK77:
This PR cleans `IApplicationFunctions` a bit. More will be done about the `GetDesiredLanguage` call later.
`TryPopFromFriendInvitationStorageChannel` is stubbed, checked with RE to return the right result code which close #1571.

This should help to progress in related games. I've tested some which are playable now:

image
image
image
image

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -5,7 +5,6 @@ using LibHac.Fs;
 using LibHac.Ns;
 using Ryujinx.Common;
 using Ryujinx.Common.Logging;
-using Ryujinx.HLE.Exceptions;
 using Ryujinx.HLE.HOS.Ipc;
 using Ryujinx.HLE.HOS.Kernel.Common;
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
Nice. I appreciate the helpful comments.

Thog (approved):

PR #1740, opened 2020-11-21
salieri: remove a wrong debug assert

Thog:
This assert is wrong and useless now, remove it because it's annoying in
Debug.


AcK77 (approved):
Thanks for that :)

gdkchan (approved):

riperiperi (approved):
😌

PR #1741, opened 2020-11-21
Interrupt GPU command processing when a frame's fence is reached.

riperiperi:
Experimental (needs testing and approach review). This PR removes the frame timer in GLRenderer and replaces it with a "frame ready" signal that's sent whenever a frame's AndroidFence is completed. When this happens, GPU command processing is interrupted and we present a frame immediately.

- "interrupts" the GPFifo processing, allowing it to check for frames to present
- Window keeps track of # frames with their fences reached. Present only occurs when at least one frame is ready.

This is an alternative to processing and presenting on separate threads. In future this should be rethought since we will have multiple SurfaceFlinger layers to consider, though it might still be useful if we decide to single thread GPU processing between these layers for whatever reason (easier debug?)

### Improvements
Generally improves framerate when the emulator below the game's framerate target due to GPU processing. Will not improve bad framerates to full speed, but may move them _towards_ it.

- Xenoblade DE and 2 have much better utilization and framepacing when they go below 30 fps. Duplicate frames at 25 fps look like 10, so the games are a _lot_ smoother even when they aren't at fullspeed.

Before:
image
After (note FIFO% - higher utilization):
image

- Game speed in Link's Awakening much more stable. It is not perfect - may improve with a sharper vsync clock.
- Frame pacing in windowed mode or when using a high refresh rate (144hz) seems somewhat improved.
- Disabling guest vsync now allows the game to reach your monitor refresh rate, as there is no artificial 60fps limit.
image

### Drawbacks
This changes the render loop, so will affect every game differently. I'd recommend testing a few games to make sure they don't start deadlocking.

- I encountered an issue in Link's Awakening where the game was stuck at 15 fps with vsync enabled. I could not replicate it after the fact, but the `_framesAvailable` count may have got out of sync with the actual frame queue.
- **Does not fix Zelda BOTW**

### TODO
I'm looking into some other frame time behaviours. Also, need to figure out what cause the weird bug in LA I described above.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Window.cs
+++ b/Ryujinx.Graphics.Gpu/Window.cs
@@ -175,7 +175,7 @@ namespace Ryujinx.Graphics.Gpu
         /// <returns>True if a frame is available, false otherwise</returns>
         public bool ConsumeFrameAvailable()
         {
-            if (_framesAvailable > 0)
+            if (Interlocked.CompareExchange(ref _framesAvailable, 0, 0) != 0)
             {
                 Interlocked.Decrement(ref _framesAvailable);
 
--- a/Ryujinx.Graphics.Gpu/Window.cs
+++ b/Ryujinx.Graphics.Gpu/Window.cs
@@ -2,6 +2,7 @@ using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Image;
 using System;
 using System.Collections.Concurrent;
+using System.Threading;
 
 namespace Ryujinx.Graphics.Gpu
 {
@@ -68,6 +69,8 @@ namespace Ryujinx.Graphics.Gpu
         }
 
         private readonly ConcurrentQueue<PresentationTexture> _frameQueue;
+
+        private int _framesAvailable;
 
         /// <summary>
         /// Creates a new instance of the GPU presentation window.
@@ -156,6 +159,30 @@ namespace Ryujinx.Graphics.Gpu
 
                 pt.ReleaseCallback(pt.UserObj);
             }
+        }
+
+        /// <summary>
+        /// Indicate that a frame on the queue is ready to be acquired.
+        /// </summary>
+        public void SignalFrameReady()
+        {
+            Interlocked.Increment(ref _framesAvailable);
+        }
+
+        /// <summary>
+        /// Determine if any frames are available, and decrement the available count if there are.
+        /// </summary>
+        /// <returns>True if a frame is available, false otherwise</returns>
+        public bool ConsumeFrameAvailable()
+        {
+            if (_framesAvailable > 0)
+            {
+                Interlocked.Decrement(ref _framesAvailable);
+
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
+++ b/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
@@ -27,6 +27,7 @@ namespace Ryujinx.HLE.HOS.Services.SurfaceFlinger
         private long _ticks;
         private long _ticksPerFrame;
         private long _spinTicks;
+        private long _1msTicks;
 
         private int _swapInterval;
 
@@ -65,6 +66,7 @@ namespace Ryujinx.HLE.HOS.Services.SurfaceFlinger
 
             _ticks = 0;
             _spinTicks = Stopwatch.Frequency / 500;
+            _1msTicks = Stopwatch.Frequency / 1000;
 
             UpdateSwapInterval(1);
 
@@ -224,7 +226,7 @@ namespace Ryujinx.HLE.HOS.Services.SurfaceFlinger
                     }
                     else
                     {
-                        _event.WaitOne(1);
+                        _event.WaitOne((int)(diff / _1msTicks));
                     }
                 }
             }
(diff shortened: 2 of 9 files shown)


riperiperi:
This has been rebased on top of the stutter fix. Also:

- The wait is now done with EventWait to get 1ms precision
- The C# friendly spinwait has more iterations to be a bit nicer to hyperthreading/power usage
- flipped _device.System?.SignalVsync(); and Compose again. Not sure which order they should really happen in.

riperiperi:
This needs better support for swap interval 0, or when we force it. It should only send the vsync signal event after buffers are swapped by the host.

riperiperi:
XC2 Comparison

Before:
image

After:
image

riperiperi:
I'm happy with how this behaves when vsync is enabled... However, when it is disabled the CPU gets absolutely hammered.

https://i.gyazo.com/c7a65ab9391a283b904eb0f25f7c9563.gif

It may be a better idea to bite the bullet and limit to 1000 vsync signals a second (1ms) rather than signalling as many million as we possibly can.

riperiperi:
Now ready for review.

There is a final component to keep "vsync off" mode cheap - It now waits for a buffer to be queued on the surfaceflinger. This allows us to consume surfaceflinger buffers literally as soon as they are available, and signal a vsync event at the same time (indicating that they have arrived right on time). It also allows us to use a very low amount of CPU for the compose loop, even when running at very high framerates.

Vsync events are still signalled at least every 17ms if a buffer was not queued. This is to prevent games that rely more on the vsync signal from locking... but you should probably enable vsync for those ones, as they will misbehave without it. (Link's Awakening)

This will need to be refactored when we're dealing with multiple layers, but so will much else about the presentation loop, I'm sure.

Ryujinx.Graphics.Gpu/Window.cs
```diff
@@ -157,5 +160,29 @@ public void Present(Action swapBuffersCallback)
...
+
+        /// <summary>
+        /// Determine if any frames are available, and decrement the available count if there are.
+        /// </summary>
+        /// <returns>True if a frame is available, false otherwise</returns>
+        public bool ConsumeFrameAvailable()
+        {
+            if (_framesAvailable > 0)
+            {
+                Interlocked.Decrement(ref _framesAvailable);
```
gdkchan, line 180: If this is supposed to be thread safe, then it should use the value returned from the decrement. I think it would require a `CompareExchange` loop to check if the value is > 0 and decrement if true, to prevent it from going below 0.

Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -187,23 +200,55 @@ private void HandleComposition()
...
+                    _ticks += ticks - lastTicks;
+                    lastTicks = ticks;
+
+                    if (_ticks >= _ticksPerFrame)
+                    {
+                        Compose();
+
+                        _device.System?.SignalVsync();
+
+                        _ticks = Math.Min(_ticks - _ticksPerFrame, _ticksPerFrame * 3);
```
gdkchan, line 229: Maybe add a comment about why the `_ticksPerFrame * 3`, where the 3 came from?

Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -187,23 +200,55 @@ private void HandleComposition()
...
-                Thread.Sleep(1);
+                    // Sleep the minimal amount of time to avoid being too expensive.
+                    long diff = _ticksPerFrame - (_ticks + _chrono.ElapsedTicks - ticks);
+                    if (diff > 0)
+                    {
+                        if (diff < _spinTicks)
+                        {
+                            do
+                            {
+                                Thread.SpinWait(5);
```
gdkchan, line 240: Maybe add a comment about this magic 5 value here.

Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -291,6 +336,12 @@ private void PostFrameBuffer(Layer layer, BufferItem item)
                 Item  = item,
             };
 
+            item.Fence.RegisterCallback(_device.Gpu, () => 
+            {
+                _device.Gpu.Window.SignalFrameReady();
+                _device.Gpu.GPFifo.Interrupt();
```
gdkchan, line 342: Why the GPU command processing needs to be interrupted here?

riperiperi: The render thread is completely sequential, and it will continue running commands until it runs out of them. Frames are only presented when command processing exits, which in master only happens when it runs out of commands. This means that certain games (xenoblade) would process a second frame without ever presenting the first, which lead to some pretty awful pacing issues.

This interrupt specifically yields command processing to present a frame.

Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
riperiperi: This is here to better handle situations where the emulator is suspended due to a debugger, sleep state or some timer mishap. If we miss more than 3 vsync signals, it will only provide the maximum of 3. I'll add a comment.

Ryujinx.Graphics.Gpu/Window.cs
riperiperi: It's meant to consume one. I can just compare and exchange for != 0, and decrement once. (there is only one consumer here anyways, so there will never be a race pushing it under 0)

riperiperi:
Comments addressed. Probably.

gdkchan (approved):
Looking good, great work!

gdkchan:
One of the memory tests failed. I restarted the jobs, but it would be nice to get this addressed (separately, of course).
```
A total of 1 test files matched the specified pattern.
  Failed Multithreading [991 ms]

Failed!  - Failed:     1, Passed:    78, Skipped:     0, Total:    79, Duration: 1 s - Ryujinx.Memory.Tests.dll (net5.0)
```

Thog (approved):

PR #1742, opened 2020-11-21
amadeus: Fix reverb 3d mono wrong delay line offset

Thog:
This fix reverb 3d mono rendering that was using -1 instead of 1 for delay line
offset.

I also did some clean up as this value is now used by all channels
variants.

Now Resident Evil 6 goes in-gane.

image
image


AcK77 (approved):

gdkchan (approved):
Nice catch.

riperiperi (approved):
Nice.

PR #1746, opened 2020-11-23
audio: Cleanup Ryujinx.Audio and fix OpenAL issue

AcK77:
This PR cleanup the Ryujinx.Audio code by removing unused code.
I've cleaned the SoundIO backend too even if some parts of it are auto generated.
I've cleaned Audout service a bit too and fixed an issue about the GetVolume function in OpenAL backend.
Fix ContainsAudioOutBuffer which didn't returns the right value.


gdkchan (approved):
Looks good, although there's still some weird spaces on SoundIO code.
Example:
```cs
return (int) (IntPtr) handle;
```
```cs
return Marshal.PtrToStringAnsi (Marshal.ReadIntPtr (handle, app_name_offset));
```

Thog (approved):

Ryujinx.HLE/HOS/Services/Audio/AudioOutManager/IAudioOut.cs
```diff
@@ -85,7 +85,7 @@ public ResultCode ContainsAudioOutBuffer(ServiceCtx context)
         {
             long tag = context.RequestData.ReadInt64();
 
-            context.ResponseData.Write(_audioOut.ContainsBuffer(_track, tag) ? 1 : 0);
+            context.ResponseData.Write(_audioOut.ContainsBuffer(_track, tag) ? true : false);
```
riperiperi, line 88: If you're using a tertiary statement on this, isn't it already true or false?

riperiperi (approved):
I approve mostly. I commented in one place where you seem to be `bool ? true : false` which is a bit unnecessary.

PR #1748, opened 2020-11-24
am: Fix GetSaveDataSize stub

AcK77:
This PR fix the `IApplicationFunctions::GetSaveDataSize` call of `am` service.
Checked by RE, we previously didn't return the right values.

https://github.com/Ryujinx/Ryujinx-Games-List/issues/2815 goes in game now (and probably some other games) :
image


gdkchan (approved):
lgtm, nice catch!

Thog (approved):

PR #1752, opened 2020-11-26
Fix pre-allocator shift instruction copy on a specific case

gdkchan:
For shift instructions, the pre-allocator needs to insert a copy to ensure the shift value is on the CL register (and you can only use CL for the shift on x86), and additionally, it needs to insert copies to ensure that the first source operand and the destination are on the same register. Right now there are 2 separate functions that handles this, but the problem is that each one of those functions returns a new node that is the node of the last operation is inserted on the list, and due to this behavior, the shift instruction would be skipped on the next call.

To sum it up, the following was possible:
- `HandleDestructiveRegCopy` inserts a copy before and a copy after the operation to ensure dest == src1, then it returns the node of the last copy.
- `HandleConstrainedRegCopy` is then not able to insert the copy to the CL register as we are at the node of the copy inserted on the previous step, with leads to a assert later on.

To fix it I simply changed it to not return the last inserted node. This ensures that all functions will be able to visit the current instruction node. I also added a temp variable that holds the next node, this allows easily skipping any newly inserted nodes that do not need to be visited.

In partice nothing hits this right now as far I'm aware, because `useNewLocal ` must be true for it to insert a node *after* the operation on `HandleDestructiveRegCopy`. And for it to be true, the destination operand and at least one of the sources must be equal. This does not happen right now because the optimizations are still not able to eliminate all copies. Since I could not find evidence that this is being hit right now, I also did not increment the PPTC version, but can do so just to be on the safe side if desired.

### DIFF
--- a/ARMeilleure/CodeGen/X86/Assembler.cs
+++ b/ARMeilleure/CodeGen/X86/Assembler.cs
@@ -899,7 +899,7 @@ namespace ARMeilleure.CodeGen.X86
             }
             else if (source.Kind == OperandKind.Constant)
             {
-                source = source.With((uint)source.Value & (dest.Type == OperandType.I32 ? 0x1f : 0x3f));
+                source = source.With((int)source.Value & (dest.Type == OperandType.I32 ? 0x1f : 0x3f));
             }
 
             WriteInstruction(dest, source, type, inst);
--- a/ARMeilleure/CodeGen/X86/PreAllocator.cs
+++ b/ARMeilleure/CodeGen/X86/PreAllocator.cs
@@ -22,16 +22,20 @@ namespace ARMeilleure.CodeGen.X86
 
             for (BasicBlock block = cctx.Cfg.Blocks.First; block != null; block = block.ListNext)
             {
-                for (Node node = block.Operations.First; node != null; node = node.ListNext)
+                Node nextNode;
+
+                for (Node node = block.Operations.First; node != null; node = nextNode)
                 {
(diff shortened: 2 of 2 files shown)


FICTURE7:
> And for it to be true, the destination operand and at least one of the sources must be equal.

Perhaps unrelated but for some commutative operations, we can swap the operands to avoid some copies. Currently this only happens when the first source is a constant. Should cause a minor improvement for LCQ.
```
 i32 %0 = Add i32 %1, i32 %0

=>

 i32 %0 = Add i32 %0, i32 %1
```

riperiperi (approved):
LGTM

LDj3SNuD (approved):
Do you think it might be worth fixing `Assembler.WriteShiftInst()` as well ?
This line `source = source.With((uint)source.Value & (dest.Type == OperandType.I32 ? 0x1f : 0x3f));` should use a `Const()`, also there is an implicit cast to `long` to handle.

gdkchan:
> Perhaps unrelated but for some commutative operations, we can swap the operands to avoid some copies. Currently this only happens when the first source is a constant. Should cause a minor improvement for LCQ. :>

Good idea, but yea, outside the scope of this PR.

> Do you think it might be worth fixing Assembler.WriteShiftInst() as well ?
This line source = source.With((uint)source.Value & (dest.Type == OperandType.I32 ? 0x1f : 0x3f)); should use a Const(), also there is an implicit cast to long to handle.

Done.

PR #1754, opened 2020-11-27
Prefer truly perfect texture matches over format aliased ones

riperiperi:
When attempting to find "perfectly matching" textures, we actually have a few gives that allow matches that are not exactly perfect. However, this means that multiple textures that are "perfect" can exist, and some might be more preferable to others.

This PR adds a preference system to selection of matching textures, where it will pick the "most perfect" match available in the texture cache. If a match is deemed truly perfect, then it is returned immediately as before. However, if some matches are not perfect, it will return the one with the highest "quality" defined in an enum.

Currently, only format aliasing for sampler/copy is considered an imperfect match. This could be expanded in future to choose better texture matches for texture sizes that match by alignment vs perfectly. I didn't include it here as it'll only be clear if it's needed when doing larger scale changes.

### Example

For the specific case in Zelda, imagine there are 2 textures with the same address and size, (1) and (2), that are half the game resolution.

- The game reduces the main depth buffer into (1) as a D32F depth buffer.
- The game draws (1): D32F into (2): R32F using a fragment shader, rescaling the depth values from [0.5,10] to [0.0,1.0]
- The game samples (2): R32F when doing the final draw, using it for the fog
- **Next frame...**
- The game reduces the main depth buffer into (2) as a D32F depth buffer. 
  - If the (1->2) step happened first, this will create a second texture with the same address and size. 
- The game draws (2): D32F into (1): R32F using a fragment shader, rescaling the depth values from [0.5,10] to [0.0,1.0]
  - Binding (1) as R32F will create a new texture for (1) that is R32F.

We now have 2 distinct versions of (1) and (2), a R32F and D32F version. Currently in master, the last sample for R32F gets the D32F texture _from the last frame_, so not only is it in the incorrect interval/format, it's also in the past. Ideally, the sample should choose the R32F version over the D32F one as it is already in the desired format, but if the D32F one exists in the list first, it will be chosen instead.

For this specific case (where the depth and red textures are used for distinct purposes), recognising these are different textures and picking the best match fixes the issue, and does not need any pesky copies. In future, this should probably create a copy dependency (once those exist) between the D32F and R32F textures, so that the data is not lost between uses of the texture. At least in this specific example, the data is overwritten every time, so it is not needed.

Fixed screenshot, immediately after loading in:
image

Would recommend testing games that use the depth format aliasing functionality. It looks fine in Burnout Paradise and Link's Awakening.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -800,29 +800,31 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         /// <param name="info">Texture information to compare against</param>
         /// <param name="flags">Comparison flags</param>
-        /// <returns>True if the textures are strictly equal or similar, false otherwise</returns>
-        public bool IsPerfectMatch(TextureInfo info, TextureSearchFlags flags)
+        /// <returns>A value indicating how well this texture matches the given info</returns>
+        public TextureMatchQuality IsExactMatch(TextureInfo info, TextureSearchFlags flags)
         {
-            if (!TextureCompatibility.FormatMatches(Info, info, (flags & TextureSearchFlags.ForSampler) != 0, (flags & TextureSearchFlags.ForCopy) != 0))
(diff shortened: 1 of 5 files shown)


Doctorwho1909:
Testing done found no regressions on nvidia or amd

gdkchan (approved):
Nice catch, looks good as a temporary solution.

Thog (approved):

PR #1755, opened 2020-11-27
Implement Force Early Z Test Register

riperiperi:
In normal circumstances, the Z test is performed before the fragment test to avoid running it when possible. However, as this is an optimization, cases where it would change the functionality instead perform the _fragment test_ first, as it can do things like set the output depth, or perform incoherent writes.

This register exists to allow shaders to _force_ the Z test to happen early regardless of whether it will cause different behaviour. In the case of Image Load/Store, performing the Z test first can very much be _desired_ behaviour.

Register can be found here:
https://github.com/envytools/envytools/blob/master/rnndb/graph/gf100_3d.xml#L68

This fixes the interior volumes on Xenoblade for good. The issue was actually found by @ReinUsesLisp [over at yuzu,](https://github.com/yuzu-emu/yuzu/pull/5013/) which after reading two sentences I looked up this register as quickly as possible so that Colony 9 wouldn't look like a kaleidoscope.

Essentially a follow-up to https://github.com/Ryujinx/Ryujinx/pull/1380 in terms of Xenoblade. I've been on this mess for a while, and I honestly can't believe that I didn't notice the depth test was essentially meaningless for the image store as it would run the fragment shader first. Khronos even have an explanation with this exact example on their wiki:

> This feature exists to ensure proper behavior when using Image Load Store or other incoherent memory writing. Without turning this on, fragments that fail the depth test would still perform their Image Load/Store operations, since the fragment shader that performed those operations successfully executed.

🥴 

Before:
https://gyazo.com/aae376b1338e835e20c6d64d5b35c0f8

After:
https://gyazo.com/b0c30a18607046588021f21ed778ddfe

image
Also fixes Xenoblade 2, if you're into that kind of thing.

## Shader Cache
This register directly affects shader compilation, and therefore the shader cache. Fortunately, the default state for this behaviour is _off_, as it is a _force_ rather than _enable_ register (auto enable is done some other way), so all shaders that don't use this functionality will not change. One of the reserved fields in the Gpu Accessor has been changed to represent GPU state flags, which can be populated with more flags as we discover them.

**HOWEVER**

There is a bug with the shader cache where the cached GpuAccessor state _is not part of the hash_, which will mean that existing broken shaders for xenoblade will still be "valid" for the case that this fixes. This should be fixed by a future PR to the shader cache to fix this and _rehash_ any existing shaders, which will not invalidate any of your precious guest cache but will correctly make them invalid only if the new EarlyZ flag is set.

If you're testing, I'd therefore recommend **doing so with shader cache off for now.** (or rebuilding for xenoblade, you insane people) Go bug @Thog to fix the issue. :P

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/Cache/CacheHelper.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/Cache/CacheHelper.cs
@@ -333,6 +333,23 @@ namespace Ryujinx.Graphics.Gpu.Shader.Cache
         }
 
         /// <summary>
+        /// Builds gpu state flags using information from the given gpu accessor.
+        /// </summary>
+        /// <param name="gpuAccessor">The gpu accessor</param>
+        /// <returns>The gpu state flags</returns>
+        private static GuestGpuStateFlags GetGpuStateFlags(IGpuAccessor gpuAccessor)
+        {
+            GuestGpuStateFlags flags = 0;
+
+            if (gpuAccessor.QueryEarlyZForce())
+            {
+                flags |= GuestGpuStateFlags.EarlyZForce;
+            }
+
+            return flags;
+        }
+
+        /// <summary>
         /// Create a new instance of <see cref="GuestGpuAccessorHeader"/> from an gpu accessor.
         /// </summary>
         /// <param name="gpuAccessor">The gpu accessor</param>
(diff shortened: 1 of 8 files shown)


gdkchan (approved):
Looks good.

Thog:
Need rebase now that #1759 is merged 👌

riperiperi:
Rebased, and looking good with shader cache changed. (correctly generates new shaders)

Thog (approved):
lgtm, thanks 👍

PR #1756, opened 2020-11-27
ngct: Fix services and stub calls

AcK77:
This fixes the ngct services placeholder which was wrong.
https://github.com/Ryujinx/Ryujinx-Games-List/issues/2887 needs calls of ngct service to goes further.
- IService Match and Filter calls are stubbed. Checked by RE.
- IServiceWithManagementApi Match and Filter calls are stubbed. Checked by RE.

Now Horace is playable :

image
image
image


Thog (approved):

gdkchan (approved):
lgtm, thanks.

PR #1760, opened 2020-11-28
NVIDIA Thread create/delete stutter workarounds

riperiperi:
When NVIDIA Threaded Optimization is on, the driver will snapshot all threads in the system whenever the application creates or exits any threads. I’m guessing it does this so it knows what all your program’s threads are, but the method it uses to do this snapshots and iterates threads in _the entire system_.

https://docs.microsoft.com/en-us/windows/win32/api/tlhelp32/nf-tlhelp32-createtoolhelp32snapshot

The ThreadPool has something called a "GateThread" which terminates itself after some inactivity. However, it immediately starts up again, since the rules regarding when to terminate and when to start differ. This creates a new thread every second or so.

https://github.com/dotnet/runtime/blob/v5.0.0-rtm.20519.4/src/coreclr/src/vm/win32threadpool.cpp#L1424

The main problem with this is that the thread snapshot can take 70ms, is on the OpenGL thread and will delay rendering any graphics. This is a little over budget on a frame time of 16ms, so creates a large stutter.

The solution is to keep the ThreadPool active so that it never has a reason to terminate the GateThread. I’ve also taken the liberty of disabling tiered compilation, as it appeared to be creating temporary threads.

---

image

For the poor souls plagued with this ailment, your FPS will rise from 57-58 to 60 in games where you can reach it. You should still play in fullscreen to really feel the 60fps, tho

I’d like to also set a minimum on the ThreadPool to stop it from downsizing, but the methods for doing that right now are not a _hard_ minimum, it still slowly kills threads while the pool is inactive. If you have any suggestions on how to improve this, please send your fix to my PO box, or just comment it idk.

image

### DIFF
--- a/Ryujinx/Ryujinx.csproj
+++ b/Ryujinx/Ryujinx.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net5.0</TargetFramework>
@@ -6,6 +6,7 @@
     <OutputType>Exe</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Version>1.0.0-dirty</Version>
+    <TieredCompilation>false</TieredCompilation>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Ryujinx/Ryujinx.csproj
+++ b/Ryujinx/Ryujinx.csproj
@@ -7,6 +7,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Version>1.0.0-dirty</Version>
     <TieredCompilation>false</TieredCompilation>
+    <TieredCompilationQuickJit>false</TieredCompilationQuickJit>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


gdkchan:
Would be nice to file a issue on .NET runtime repo about this.

GamerzHell9137:
Its cleaning up the FPS for me in Mario Kart
Base Profile Screenshot 2020 11 29 - 00 30 52 69

Thog (approved):

jduncanator:
Disabling Tiered Compilation and Quick JIT is a pretty big hammer to fix this problem. Have the effects of disabling this been measured to show that it's a negligible difference?

riperiperi:
I did not notice any change playing games, though I don't know if dotnet has some kind of disk cache that might be saving me. I don't think there's much cold ryu code by the time you get into a game. Quick JIT was disabled to make sure we get the same code quality as if tiered compilation were active.

Either way, any lengthened dotnet JIT stutter is a lot better than the one nvidia is unfortunately enforcing on us.

riperiperi:
The gate thread issue seems to be resolved in ipc-part-2, though it's not clear what fixed it. If this persists after future IPC changes, this workaround may not be needed and just disabling tiered compilation may be good enough.

gdkchan (approved):

PR #1769, opened 2020-12-02
Create a single guest process per IPC message processor

gdkchan:
Should fix a regression introduced on #1458 that caused some games like Pokémon to crash due to a OutOfMemory error being returned from SendSyncRequest. It was getting out of memory because multiple guest processes were being creased for the Bsd service, and the memory used by them was never freed. This change creates just a single process for Bsd and fixes the issue. It's worth noting that #1504 does something similar, so it should fix the issue aswell in theory, but I didn't test yet.


AcK77 (approved):

Doctorwho1909:
issue is fixed the PR works

Thog (approved):

PR #1771, opened 2020-12-02
Cache delegate for QueryModified, use regular multi handle.

riperiperi:
This PR has two simple changes:

- Uses of QueryModified now pass a cached `Action<ulong, ulong>` delegate of a named method for the function that handles modified regions.
  - This is because a profile revealed that a lot of time was being spent creating a new delegate for the lambda function on each call.
- Change Buffer tracking to use simple tracking rather than smart, as the case where the smart tracking is faster has not been discovered yet.
  - The same as: https://github.com/gdkchan/Ryujinx/commit/c77cc956866a7452c161b529e5c691e214b5cbc2
  - It may have also been causing additional issues in games such as Catherine, though this may just be an amplification of an existing problem.

Combined, these changes almost half buffer tracking time in SMO Metro Kingdom's initial view (~11.5% to ~6%). (it is still the same FPS for other reasons, though) There is a small improvement with Xenoblade: DE with the interrupt-gpu branch.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Pool.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Pool.cs
@@ -36,6 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         public ulong Size { get; }
 
         private readonly CpuMultiRegionHandle _memoryTracking;
+        private readonly Action<ulong, ulong> _modifiedDelegate;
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
lgtm 👍

Thog (approved):

PR #1773, opened 2020-12-03
Dummy out occlusion queries with 0 draws, remove glFlush call

riperiperi:
A glFlush was placed before query end to mitigate some potential issues when ending a query after 0 draw calls were made. However, glFlush has a substantial cost that can greatly slow down OpenGL if you're not careful.

This PR adds a draw count, which can be snapshotted when a query begins and compared when it ends. If 0 draws have occurred since we began, simply return the result immediately (0) and don't add it to the queue.

Affected games:
- Super Mario Odyssey (primarily metro kingdom, or when res scaling is active)
  - Specifically, the nvoglv64.dll thread's load is now ~65% of the FIFO processing thread's load in the metro kingdom initial view, rather than being neck and neck with it. This means that perf improvements elsewhere are now visible.
- Splatoon 2
- Mario Kart 8 (many racers, extreme circumstances. in vs race translated cpu code is now always the bottleneck)

Best friends with https://github.com/Ryujinx/Ryujinx/pull/1771

This does not affect games that do not use occlusion queries. There don't seem to be any regressions, can still swim thru ink in splat2 etc. May fix some weird driver bugs with drawless queries. (query timed out?)

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -10,6 +10,8 @@ namespace Ryujinx.Graphics.OpenGL
 {
     class Pipeline : IPipeline, IDisposable
     {
+        internal ulong DrawCount { get; private set; }
+
         private Program _program;
 
         private bool _rasterizerDiscard;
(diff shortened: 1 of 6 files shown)


gdkchan (approved):
Looks good.

Thog (approved):

PR #1778, opened 2020-12-03
Applets: Explicitly yield kthread when waiting on GUI dialogs

mageven:
Following IPC Refactor Pt 2, blocking host calls (such as applet GUI dialogs) cause the entire core to lock as the service processes use cooperative scheduling prio. This PR uses a helper to yield the waiting applet thread.

Also applets now have their own Server; otherwise it'll yield the common server process instead.

Fixes the glitchy audio during FE3H swkbd invoke and similar issues with other applets (post IPC Pt 2).

I'm not sure this is the best solution in the long-term. Will leave the merge decision to gdkchan as I don't know the full scope of upcoming changes.


gdkchan:
Looks good but should be addressed more broadly by #1786.

mageven:
Closing in favor of #1786

PR #1785, opened 2020-12-05
Signal memory tracking before/after mapping into another process

riperiperi:
Fixes a regression from #1458 , where data written by a service would not mark regions as dirty, and reads from the service would not trigger flushes from textures.

This fixes vertex explosions after one race in Mario Kart 8 Deluxe, and may fix some other games with graphical errors introduced in #1458 . The game appears to write buffer data directly from a service.

Easy repro:
VS Race: Hyrule Circuit -> Sunshine Airport -> Yoshi Circuit

Broken example:
https://gyazo.com/a6ce0d7219ebe086dbb48c0324c43e16

You don't need a fixed example, you've all seen Mario Kart 8 before.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
@@ -1656,6 +1656,9 @@ namespace Ryujinx.HLE.HOS.Kernel.Memory
                 }
             }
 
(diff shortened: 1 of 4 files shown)


gdkchan (approved):
lgtm, thanks for looking into this. It might be worth to not mark the region as dirty if the buffer can't be written to by the service (send buffers can't be written to for example).

gdkchan:
Tests build are failing as `MockVirtualMemoryManager` does not implement the new interface method.

Thog (approved):

PR #1786, opened 2020-12-06
Rewrite scheduler context switch code

gdkchan:
Fixes some issues with the current context switch code.
- It was possible for a thread to be scheduled on two cores at once.
  - Now it uses a lock to prevent the thread from being scheduled in two cores at once, like the official OS does.
- Fixed thread CurrentCore value being updated ahead of time. Now it is only updated after the thread is switched to the new core.
- Fixed bugs on some yield functions that could cause the scheduler to select the wrong thread.
- Fixed bugs that would cause KThread and KSession reference counts to be decremented below 0. For KThread, it could cause a exception since the thread was removed from the KProcess thread list twice (only observed in single core mode before the changes).
  - Two asserts were added to ensure the reference count is not decremented below 0, or incremented after the object has been "destroyed".
- Fixed rance on unmap client buffer function (IPC related, not scheduler related, but its a small, simple change as was affecting the tests here aswell, so I decided to just fix it).

Other improvements:
- Core idle time is now properly reported to the guest instead of just returning 0 (I am not sure if games can actually read that).
- `GetCurrentThread` and `GetCurrentProcess` functions were simplified, now it uses a thread local variable instead of looping through all core contexts and try finding the current thread.
- Context switch speed was improved. This is in part thanks to the use of `ManualResetEventSlim`, that greatly improves the speed on context switches in quick succession, and in part to the simplification of the function.
- The dummy thread and process were removed. Now `GetCurrentThread` and `GetCurrentProcess` just returns null if called from a non-guest or non-service thread.
  - Prevents bugs caused by external threads messing with the scheduler.
  - A new function was added that allows creating a guest process/thread from a external thread, this is used to start service faux processes.
- Allow service threads to run in parallel again.
  - They can be still be blocked by kernel in some cases, like when waiting for IPC messages for example, however they are no longer scheduled to cores, which means that they are free to run in parallel.
  - The `YieldUntilCompletion` function was removed as it is no longer needed.
  - This matches the behavior before ipc-part2, and fixes issues caused by this specific ipc-part2 change, like the audio crackling on SMO etc.

Other changes:
- Single core mode was removed. Now there is only multi core mode that is on by default.
  - The GUI checkbox was removed.
  - It is worth noting that the single core mode was not working anyway due to the bug mentioned before.

Some of the code used Atmosphere as reference. The respective license was added to `THIRDPARTY.md`.

This changes a core part of the emulator, as such I recommend throrough testing to try finding any possible regressions.


legend800:
I tested Yoshi Crafted World with this PR vs. today's master and the PR build had constant (every 1-2 secs) micro freezes, making the game unplayable. Master was much smoother. This was on a cached level (played 3+ times) so this is not related to that. 

i-i6700k @4.5, nvidia 1080, 16gb ram, Win10

jduncanator:
> I tested Yoshi Crafted World with this PR vs. today's master and the PR build had constant (every 1-2 secs) micro freezes, making the game unplayable. Master was much smoother. This was on a cached level (played 3+ times) so this is not related to that.
> 
> i-i6700k @4.5, nvidia 1080, 16gb ram, Win10

Thanks for the feedback, it's much appreciated!

riperiperi:
There are micro freezes since https://github.com/Ryujinx/Ryujinx/ pull/1760 is required again, and I noticed yesterday some macro freezes in Hyrule Warriors. We identified that it was due to the SpinWait in KScheduler, so hopefully a future change will improve that soon.

jduncanator:
> There are micro freezes since #1760 is required again, and I noticed yesterday some macro freezes in Hyrule Warriors. We identified that the second one was due to the SpinWait in KScheduler, so hopefully a future change will improve that soon.

Doesn’t this assume the user has NVIDA Threaded Optimisation enabled? Or do the stutters affect all NVIDIA drivers regardless of that setting?

riperiperi:
The default setting is "Auto", which generally works the same as having it enabled. To avoid the stutters, someone would have to manually disable it, and get much worse overall performance as a result too.

Th3d0nGsT3R:
Does this still need testing?

gdkchan:
Ready for review...

legend800:
Micro-freezing issue is all fixed. It's just as, if not more, smooth than Master now. Nice work!

Thog (approved):

riperiperi (approved):
Looks very clean, seems to run quite a bit better than even before IPC2. Nice work! 😌

gdkchan:
Found a race condition in KTimeManager that was causing IPC to deliver invalid messages, among other issues. The issue was on this piece of code:
```cs
                        if (timeUp)
                        {
                            lock (_waitingObjects)
                            {
                                timeUp = _waitingObjects.Remove(next);
                            }
                        }

                        if (timeUp)
                        {
                            next.Object.TimeUp();
                        }
```
The following scenario is possible here:
- TimeManager removes a thread (let's call it thread A) from the `_waitingObjects` list.
- Host OS decides to suspend the TimeManager thread (that runs the code above).
- Thread A runs and calls ` UnscheduleFutureInvocation` to remove itself from the list, as it should not be signaled after the timeout anymore.
- Thread A runs as normal, then decides to make a request to a service, calls `SendSyncRequest`.
- Thread A is suspended and it should wait for the response to the request made.
- Now the host OS decides that the TimeManager thread should continue running.
- TimeManager thread continues, calls `TimeUp` which will signal the thread and wake it up.
- Thread A is awaken, even though it just called SendSyncRequest was just called and the service had no time to process the request and send a response yet.
- Thread A notices the invalid response and crashes.

I fixed the issue by waking the thread inside the lock aswell. This way removing the thread from the list and waking it up is now atomic. Since `TimeUp` will try to acquire the kernel scheduler lock, and operations that take that lock will also call the TimeManager, having two locks here would cause a deadlock, so I reused the scheduler lock there aswell.

This is not a new issue, it has been there for more than 2 years I believe, however it seems to be more frequent with those changes for some reason. Anyway, I decided to include the fix here to avoid a rebase later.

Should fix crashes with `NotImplementedException: Response` and `SetTerminateResult` with error 0x0001a80a in the logs. This bug is not exclusive to IPC so it may also fix other random crashes and softlocks.

PR #1788, opened 2020-12-07
am: Implement GetHealthWarningDisappearedSystemEvent

AcK77:
This implement `IApplicationFunctions GetHealthWarningDisappearedSystemEvent` call which is needed by Mario Kart 8 in chinese version. Call was checked by RE.
We still have to determine where signals some AM events.

Thanks to Kakasita on our Discord for reported this and confirm this works fine!


riperiperi (approved):
Looks good.

The fact that this is used makes you wonder - what would happen if this event were signalled while the game was running? Maybe that is worth a shot on mario kart (who knows, maybe the health warning is about the particle effects. :P )

Thog (approved):

PR #1790, opened 2020-12-08
Implement lazy flush-on-read for Buffers (SSBO/Copy)

riperiperi:
This PR makes it possible for SSBO writes to be read from CPU, and ensures they are not accidentally overwritten. The performance impact should be minimal due to the approach, but more extreme cases could have an impact.

The primary goal is fixing restarting particles that appear in many Nintendo games, caused by CPU writes to the same page that an SSBO write happens to (which causes the whole page to be read in, and replace the SSBO written data with its initial value). I was also hoping it might fix some vertex explosions. Flushing data to be read by CPU is hilariously just a second goal, but it works now too.

Simply adding read/write protection to a region of a buffer every time there is a modification would be a recipe for disaster - a tracked write could happen to the same page of the data immediately after it is written, which would require us to wait on a draw that just happened. This could happen multiple times in a row, which could reduce framerate dramatically. Ideally, we only want to flush the data when the guest _expects_ it to be there and is actually reading/overwriting, rather than whenever it is writing to the same page for an unrelated draw.

So, instead I took an approach that takes advantage of guest syncpoints. The assumption is that data written from a shader will _never_ be tampered with by the guest before the GPU tells it that the command has executed, so during this time it should never be read/overwritten. This way, our syncpoints are finally actually waiting on the GPU (hooray), but only for flushing buffers right now. This functionality can be extended to ensure better sync for other things.

Essentially, the order of operations is:
- Shader or Copy sets a region of the buffer as modified.
- The region is added to the modified range list, with the current "SyncNumber" stored with the region.
  - If the region already exists, its SyncNumber will be updated.
  - When a region slices into another, it will split the existing region into multiple parts.
  - A "SyncAction" is registered, which will add the flush action for this buffer when the next syncpoint is reached.
- When a buffer has queried as modified, any existing modified ranges are excluded from the data load.
  - It is not possible to overwrite GPU modified data until the GPU increments a syncpoint.
- When the GPU increments a syncpoint, it is telling the CPU that the data in the buffers is ready to either be read or replaced.
  - A GL sync object is created. This is waited on before flushing data, to ensure that it is up to date before the flush.
  - All modified ranges will add a read action to their pages that attempts to flush the buffer data, and remove the modified range.
- When the CPU reads/writes to a page containing one of the ranges, it will find all ranges that have sync objects (not == current SyncNumber), and wait on them. Then, it will flush all of their data to memory.
  - When data is flushed, the range is removed from the modified range list. This allows a following write from CPU to overwrite the data (and avoids flushing it again)

In the background, we are lazily deleting sync objects as they are reached in `PreFrame`. Deleting sync objects mid draw is costly, especially if they have not been reached yet, so this is crucial for performance.

## Other Changes
- Buffer copies now set buffer regions as modified and flush lazily. Unity games do this a lot, and so does Pokemon.
- Order of operations when expanding buffers has been changed for more robustness.
- Fixed a bug with Memory Tracking where write tracking (dirty flags) could be lost after read tracking (flush) was triggered.
- GPU MemoryManager reads can now trigger memory tracking (shader data).

## Examples

_**Mario Kart 8**_
**(before):**
- Explosion & snow particles: https://gyazo.com/6a02ef6468ff1208a8120e1823a5a498
- Drift particles: https://gyazo.com/69c80073f6949ab6187f05fc631a0129

**(after):**
- Explosion & snow particles: https://gyazo.com/18e77b5be3ce0f22c65f23ab43e96880
- Drift particles: https://gyazo.com/bc454bdeca06939f7f2886c1fe39e258

_**Super Mario Odyssey**_
**(before):**
- Birds: https://gyazo.com/a610cf477f6897c20adff54634e6540b
- Flagpole: https://gyazo.com/e0c845e8373ca7a7d6300201075e72ef

**(after)**
- Birds: https://gyazo.com/c0eb8fa3953bb8147e65e12b90a7471a
- Flagpole: https://gyazo.com/638476d00221e719552fa6fcaf3e1131

**Link's Awakening Flashing (before):** https://gyazo.com/7c10a670ab62ba9eaed137564c3f246c
**(after):** https://gyazo.com/b4113dab7728ce7076f6141bbacf2b44

**Also fixed:**
- Animal Crossing particles no longer reset.
- Any other nintendo game with resetting particles.
- Vertex explosions in any Unity games that had them.
- A bug was fixed where modified textures could flush over buffer or other texture data and cause corruption/crashes (Mario Kart mostly)
- Hyrule Warriors vertex explosions are somewhat reduced (but still occur rarely)
- Any game that expects buffer data to be available to the CPU will now do better at that. I have no idea what games that could mean.

## Future
- Two resources with tracked regions cannot interact. Ideally, adding a read action would immediately set any handles on the same page as dirty, therefore allowing for buffer to texture flushes. This is not yet possible, but it wasn't possible before anyways.
- We can potentially flush textures directly to buffers now, for MethodCopyBuffer, and have the data readable from CPU. I'm specifically thinking of pokemon here. To do this, the above bullet needs to be completed.

### DIFF
--- a/Ryujinx.Graphics.Gpu/GpuContext.cs
+++ b/Ryujinx.Graphics.Gpu/GpuContext.cs
@@ -4,6 +4,7 @@ using Ryujinx.Graphics.Gpu.Engine.GPFifo;
 using Ryujinx.Graphics.Gpu.Memory;
 using Ryujinx.Graphics.Gpu.Synchronization;
 using System;
+using System.Collections.Generic;
 using System.Threading;
 
 namespace Ryujinx.Graphics.Gpu
@@ -59,6 +60,18 @@ namespace Ryujinx.Graphics.Gpu
         /// </summary>
         internal int SequenceNumber { get; private set; }
 
+        /// <summary>
+        /// Internal sync number, used to denote points at which host synchronization can be requested.
+        /// </summary>
+        internal ulong SyncNumber { get; private set; }
+
+        /// <summary>
+        /// Actions to be performed when a CPU waiting sync point is triggered.
+        /// If there are more than 0 items when this happens, a host sync object will be generated for the given SyncNumber,
+        /// and the SyncNumber will be incremented.
+        /// </summary>
+        internal List<Action> SyncActions { get; }
+
         private readonly Lazy<Capabilities> _caps;
 
         /// <summary>
@@ -87,6 +100,8 @@ namespace Ryujinx.Graphics.Gpu
             _caps = new Lazy<Capabilities>(Renderer.GetCapabilities);
 
             HostInitalized = new ManualResetEvent(false);
+
+            SyncActions = new List<Action>();
         }
 
         /// <summary>
@@ -116,6 +131,37 @@ namespace Ryujinx.Graphics.Gpu
         public void SetVmm(Cpu.MemoryManager cpuMemory)
         {
             PhysicalMemory = new PhysicalMemory(cpuMemory);
+        }
+
+        /// <summary>
+        /// Registers an action to be performed the next time a syncpoint is incremented.
+        /// This will also ensure a host sync object is created, and SyncNumber is incremented.
+        /// </summary>
+        /// <param name="action"></param>
+        public void RegisterSyncAction(Action action)
+        {
+            SyncActions.Add(action);
+        }
+
+        /// <summary>
+        /// Creates a host sync object if there are any pending sync actions. The actions will then be called.
+        /// If no actions are present, a host sync object is not created.
+        /// </summary>
+        public void CreateHostSyncIfNeeded()
+        {
+            if (SyncActions.Count > 0)
+            {
+                Renderer.CreateSync(SyncNumber);
+
+                SyncNumber++;
+
+                foreach (Action action in SyncActions)
+                {
+                    action();
+                }
+
+                SyncActions.Clear();
+            }
         }
 
         /// <summary>
--- a/Ryujinx.Graphics.Gpu/GpuContext.cs
+++ b/Ryujinx.Graphics.Gpu/GpuContext.cs
@@ -137,7 +137,7 @@ namespace Ryujinx.Graphics.Gpu
         /// Registers an action to be performed the next time a syncpoint is incremented.
         /// This will also ensure a host sync object is created, and SyncNumber is incremented.
         /// </summary>
-        /// <param name="action"></param>
+        /// <param name="action">The action to be performed on sync object creation</param>
         public void RegisterSyncAction(Action action)
         {
             SyncActions.Add(action);
--- a/Ryujinx.Graphics.Gpu/GpuContext.cs
+++ b/Ryujinx.Graphics.Gpu/GpuContext.cs
@@ -67,7 +67,7 @@ namespace Ryujinx.Graphics.Gpu
 
         /// <summary>
         /// Actions to be performed when a CPU waiting sync point is triggered.
-        /// If there are more than 0 items when this happens, a host sync object will be generated for the given SyncNumber,
+        /// If there are more than 0 items when this happens, a host sync object will be generated for the given <see cref="SyncNumber"/>,
         /// and the SyncNumber will be incremented.
         /// </summary>
         internal List<Action> SyncActions { get; }
@@ -135,7 +135,7 @@ namespace Ryujinx.Graphics.Gpu
 
         /// <summary>
         /// Registers an action to be performed the next time a syncpoint is incremented.
-        /// This will also ensure a host sync object is created, and SyncNumber is incremented.
+        /// This will also ensure a host sync object is created, and <see cref="SyncNumber"/> is incremented.
         /// </summary>
         /// <param name="action">The action to be performed on sync object creation</param>
         public void RegisterSyncAction(Action action)
--- a/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
@@ -289,16 +289,14 @@ namespace Ryujinx.Graphics.Gpu.Memory
         }
 
         /// <summary>
-        /// Called when the memory for this buffer has been unmapped.
+        /// Called when part of the memory for this buffer has been unmapped.
         /// Calls are from non-gpu threads.
         /// </summary>
-        public void Unmapped()
+        /// <param name="address">Start address of the unmapped region</param>
+        /// <param name="size">Size of the unmapped region</param>
+        public void Unmapped(ulong address, ulong size)
         {
-            _memoryTracking?.Reprotect();
-            _memoryTracking?.RegisterAction(null);
-
-            _memoryTrackingGranular?.QueryModified((address, size) => { });
-            _memoryTrackingGranular?.RegisterAction(Address, Size, null);
+            _modifiedRanges?.Clear(address, size);
         }
 
         /// <summary>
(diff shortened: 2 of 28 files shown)


riperiperi:
>>> A Hat in Time crashes due to shader corruption quite a lot. Perhaps texture and buffer flushes are overlapping (see below), I need to look into it.

This is due to a buffer copy that contains shader code, from some other buffer (that was not modified by an SSBO). The problem is that within the GPU, you don't need to wait for a syncpoint to start using data from a previous operation, it's all expected to be in order. So, I think copying the shader data was being missed.

I've added an optimization, where buffer copies will forego lazy flush if the source region has no modified ranges, and instead copy the data on CPU as well as doing the buffer copy. This removes all tracking overhead, so will be faster in cases like Unity games and Pokemon. It actually doesn't cost more in any situation, as lazy flush will have _eventually_ flushed out the data anyways, and non-gpu data copies are pretty fast anyways.

It has the side effect of fixing A Hat in Time, as the shader data is never written from a shader.

This will still be a problem if shaders are written from SSBO (???), but I don't think it is too likely and wouldn't have worked before anyways. It will be solved in future when I think of a good way of making tracked resources fully aware of each other (future PR? depends on if we have regressions)

Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
```diff
@@ -849,9 +898,9 @@ public void ClearBuffer(GpuVa gpuVa, ulong size, uint value)
         /// <param name="address">Start address of the memory range</param>
         /// <param name="size">Size in bytes of the memory range</param>
         /// <returns>The buffer sub-range for the given range</returns>
-        private BufferRange GetBufferRange(ulong address, ulong size)
+        private BufferRange GetBufferRange(ulong address, ulong size, bool write = false)
```
riperiperi, line 901: Missing Doc

Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
```diff
@@ -861,19 +910,30 @@ private BufferRange GetBufferRange(ulong address, ulong size)
         /// <param name="address">Start address of the memory range</param>
         /// <param name="size">Size in bytes of the memory range</param>
         /// <returns>The buffer where the range is fully contained</returns>
-        private Buffer GetBuffer(ulong address, ulong size)
+        private Buffer GetBuffer(ulong address, ulong size, bool write = false)
```
riperiperi, line 913: Missing Doc

Ryujinx.Graphics.Gpu/Memory/BufferModifiedRangeList.cs
```diff
@@ -0,0 +1,365 @@
...
+        /// <param name="address">Start address to query</param>
+        /// <param name="size">Size to query</param>
+        /// <param name="rangeAction">The action to call for each modified range</param>
+        public void WaitForAndGetRanges(ulong address, ulong size, Action<ulong, ulong> rangeAction)
+        {
+            ulong endAddress = address + size;
+            ulong currentSync = _context.SyncNumber;
+
+            int rangeCount = 0;
+            // Range list must be consistent for this operation
```
riperiperi, line 222: Newline above for clarity

Ryujinx.Graphics.Gpu/Memory/MemoryManager.cs
```diff
@@ -61,11 +61,11 @@ public MemoryManager(GpuContext context)
         /// <param name="gpuVa">GPU virtual address where the data is located</param>
         /// <param name="size">Size of the data</param>
         /// <returns>The span of the data at the specified memory location</returns>
-        public ReadOnlySpan<byte> GetSpan(ulong gpuVa, int size)
+        public ReadOnlySpan<byte> GetSpan(ulong gpuVa, int size, bool tracked = false)
```
riperiperi, line 64: Missing Doc

Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -34,7 +34,7 @@ class ShaderCache : IDisposable
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 1901;
+        private const ulong ShaderCodeGenVersion = 1001;
```
riperiperi, line 37: This version is a placeholder.

riperiperi (commented):
Self comments to correct before undrafting.

riperiperi:
Some newer changes: 

- Flushing happens with page granularity now. Most SSBOs have a small working area, so this avoids some expensive operations when a larger than needed area is set.
- A race condition was fixed (LDN2 testing)
- WaitForIdle now adds a Sync object, which allows for GPU resources to be flushed from buffers within the same frame.
  - SetReference may be a more reliable way to get this for indirect draws, but that will be a future PR

Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -172,15 +327,62 @@ public void Flush(ulong address, ulong size)
...
 
+        /// <summary>
+        /// Align a given address and size region to page boundaries.
+        /// </summary>
+        /// <param name="address">The start address of the region</param>
+        /// <param name="size">The size of the region</param>
+        /// <returns>The page aligned address and size</returns>
+        private (ulong address, ulong size) PageAlign(ulong address, ulong size)
+        {
+            ulong pageMask = (ulong)MemoryManager.PageSize - 1;
```
gdkchan, line 338: Why not just `MemoryManager.PageMask`?

Ryujinx.Graphics.Shader/BufferUsageFlags.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// <summary>
+    /// Flags that indicate how a buffer will be used in a shader.
+    /// </summary>
+    [Flags]
+    public enum BufferUsageFlags
+    {
+        None = 0,
+
+        // Buffer is written to.
+        Write = 1 << 0
```
gdkchan, line 14: Why not turn it into a proper XML doc comment?

Ryujinx.Memory/Range/RangeList.cs
```diff
@@ -12,7 +12,7 @@ public class RangeList<T> : IEnumerable<T> where T : IRange
     {
         private const int ArrayGrowthSize = 32;
 
-        private readonly List<T> _items;
+        protected readonly List<T> _items;
```
gdkchan, line 15: I think the name should be changed to `Items` if its changed from protected, according to the adopted naming conventions.

Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -59,6 +60,18 @@ public sealed class GpuContext : IDisposable
...
         internal int SequenceNumber { get; private set; }
 
+        /// <summary>
+        /// Internal sync number, used to denote points at which host synchronization can be requested.
+        /// </summary>
+        internal ulong SyncNumber { get; private set; }
+
+        /// <summary>
+        /// Actions to be performed when a CPU waiting sync point is triggered.
+        /// If there are more than 0 items when this happens, a host sync object will be generated for the given SyncNumber,
```
gdkchan, line 70: Could use `<see cref="SyncNumber"/>` rather than `SyncNumber` to improve intellisense.

gdkchan (commented):
Great work, would be nice to flush transform feedback and image buffers aswell, but I guess theres a reason this is called SSBO flush.

riperiperi:
Won't buffer textures flush through the Texture flushing mechanisms right now? Granted, there is no sync like there is with the buffers...

gdkchan:
> 
> 
> Won't buffer textures flush through the Texture flushing mechanisms right now? Granted, there is no sync like there is with the buffers...

Yea, right, although I think this can have issues if the buffer is also used for something else, as it could be flushed twice.

gdkchan (approved):
lgtm, great work 👍.

Thog (approved):

PR #1798, opened 2020-12-10
Remove interrupt counting and interrupt CPU immediately

gdkchan:
Currently the CPU has a counter that only checks if a interrupt was requested after N blocks were executed. This was done to prevent slow down in some games (like Sonic Mania) caused by context switch overhead. #1786 improved context switch speed, so I think it's worth revisiting it.

With those changes it simply checks if a interrupt was requested, if so it calls the interrupt handler immediately, without any counting. This can improve performance a little bit for the following reasons:
- Less time waiting to switch threads.
- Less code to check if a interrupt was requested.
- No periodic calls to the `CheckInterrupt` function.

However it can also slow down games doing a *lot* of context switches in rapid succession, so it's worth watching out for that.

### DIFF
--- a/ARMeilleure/State/ExecutionContext.cs
+++ b/ARMeilleure/State/ExecutionContext.cs
@@ -6,13 +6,9 @@ namespace ARMeilleure.State
 {
     public class ExecutionContext
     {
-        private const int MinCountForCheck = 4000;
-
         private NativeContext _nativeContext;
 
         internal IntPtr NativeContextPtr => _nativeContext.BasePtr;
-
-        private bool _interrupted;
 
         private static Stopwatch _tickCounter;
 
@@ -87,8 +83,7 @@ namespace ARMeilleure.State
             _nativeContext = new NativeContext(allocator);
 
             Running = true;
-
-            _nativeContext.SetCounter(MinCountForCheck);
+            _nativeContext.SetInterruptState(0);
         }
 
         public ulong GetX(int index)              => _nativeContext.GetX(index);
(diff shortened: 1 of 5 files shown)


gdkchan:
This is causing significant performance regression on Sonic Mania and BOTW, so on hold for now. We need to increase precision of wait timeouts for that to work, otherwise context switches will happen way too often and hurt performance.

gdkchan:
Making this ready for review again. One of the issues that made this have a large performance regression in some games is that we did not support precise waits (with less than 1 ms), and would just continue execution almost immediately in this case. This could lead to the same threads being rescheduled very often. Now we have more precise waits, spin waiting if the wait amount is less than 1 ms, so this should be less of an issue.

The benefits of this change is that the interruptible safe point checks we need to add to JIT generated code is now a bit simpler/shorter, so it could result in a minor performance improvement.

The games to test are the same problematic ones I mentioned on my previous comment, plus anything really.

Thog (approved):
lgtm, need a rebase

gdkchan:
The issue mentioned above still exists, it causes perf loss in some games.

TSRBerry:
Needs another rebase. Did the situation around performance losses change?

gdkchan:
> Needs another rebase. Did the situation around performance losses change?

I don't think so, but I haven't tested it recently.

ayurai:
so the games that need test is BOTW and Sonic Mania?  Can you rebase to easier test?

gdkchan:
I rebased this and same story. There's still a very significant perf loss on the games that were mentioned.
For example, here's the title screen of Sonic Mania on master:
image
And with this change:
image
So on those screenshots there's a loss of 34 FPS there.
I don't know if that can be solved without generally making syscalls/interrupts faster, which is not exactly simple, and some of the costs are out of our control (such as the .NET native <-> managed transition costs).

PR #1801, opened 2020-12-11
Fix register read after write on STREX implementation

gdkchan:
This fixes a regression introduced on #1413 where the STREX instruction implementation could write the wrong value or write to the wrong address if the register used for Rs (result) was the same register used for either the address or the value. Additionally, this also zeros the upper bits of the exclusive value stored on context, just to be on the safe side, in case there is some application out there mixing LDREX/STREX of different sizes.

This fixes a regression on Sonic Forces, where the game would no longer boot.


riperiperi (approved):
Nice catch.

Thog (approved):

PR #1807, opened 2020-12-11
salieri: Support read-only mode if archive is already opened

Thog:
This improves shader cache resilience when people opens another program that touch the cache.zip.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/Cache/CacheCollection.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/Cache/CacheCollection.cs
@@ -116,6 +116,8 @@ namespace Ryujinx.Graphics.Gpu.Shader.Cache
         /// </summary>
         private ZipArchive _cacheArchive;
 
+        public bool IsReadOnly { get; }
+
         /// <summary>
         /// Immutable copy of the hash table.
         /// </summary>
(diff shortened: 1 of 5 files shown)


riperiperi (approved):
Nice one, this will be useful for testing with multiple instances.

Ryujinx.Graphics.Gpu/Shader/Cache/CacheCollection.cs
```diff
@@ -308,6 +318,20 @@ public void EnsureArchiveUpToDate()
...
             string archivePath = GetArchivePath();
 
+            if (IsReadOnly)
+            {
+                Logger.Warning?.Print(LogClass.Gpu, $"Cache collection archive in read-only, archiving task skipped.");
+
+                return;
+            }
+
+            if (CacheHelper.IsArchiveReadOnly(archivePath))
```
gdkchan, line 328: Whats the difference between this check and the one above?

gdkchan (approved):
lgtm, thanks.

Ryujinx.Graphics.Gpu/Shader/Cache/CacheCollection.cs
Thog: This one ensure the cache wasn't locked later on (by another instance running) while exiting the first instance.
I also wanted to have a different warning for this to be more explicit for the user.

AcK77 (approved):

PR #1808, opened 2020-12-12
Rewrite size for fixed size buffers

gdkchan:
#1458 added more accurate IPC handling. After the changes, services no longer have access to the size of fixed size buffers (thats how its done on hardware, they don't need to know the size since the size is fixed). Services were not prepared to handle that, and we were reading them from the client buffer descriptor. So, after the IPC changes the functions using fixed-size output buffers could return data that is too large (that would result in a error being returned by the kernel as copy would fail), or too small (which would cause data to be missing).

This updates the affected function to write the correct sizes on the pointer buffer descriptors.

### DIFF
--- a/Ryujinx.Cpu/MemoryHelper.cs
+++ b/Ryujinx.Cpu/MemoryHelper.cs
@@ -37,7 +37,7 @@ namespace Ryujinx.Cpu
             }
         }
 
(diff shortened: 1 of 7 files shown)


AcK77 (approved):
Nice catch! Lgtm!

riperiperi (approved):

jduncanator (approved):

PR #1814, opened 2020-12-14
Free up memory allocated by Pools during any PPTC translations at boot time.

LDj3SNuD:
`Ptc.MakeAndSaveTranslations()` no longer uses `Parallel.ForEach()` but `Thread`, to avoid having translation threads with too long a lifetime and too much turnover of the same threads at the end of the job.

Nits.

`ARMeilleure.Common.BitMap` pools are not involved.

```
NES Online:
  7904 (00:25) Translations:
    ~1600 MB: Memory to menu without ResetPools()
    ~800 MB: Memory to menu with ResetPools()
```


ARMeilleure/Common/ThreadStaticPool.cs
```diff
@@ -33,9 +35,10 @@ private static Stack<ThreadStaticPool<T>> GetPools(int groupId)
         public static void PreparePool(int groupId)
         {
             // Prepare the pool for this thread, ideally using an existing one from the specified group.
+
             if (_instance == null)
             {
-                Stack<ThreadStaticPool<T>> pools = GetPools(groupId);
+                var pools = GetPools(groupId);
```
riperiperi, line 41: Switch to var for a function with no obvious type? I don't think we normally do this. (though `Stack<ThreadStaticPool<T>>` was rather verbose.

riperiperi (approved):
Looks good, this one was pretty rough on RAM. Just one small question about the use of `var`, otherwise looks good.

ARMeilleure/Common/ThreadStaticPool.cs
LDj3SNuD: It's just a matter of verbosity, I can change if you want.

Thog (approved):
lgtm!

gdkchan (approved):

Issue #1791, opened 2020-12-08
Unimplemented pmull instruction


AcK77:
### What CPU instruction is missing?

Requires the pmull instruction.</br>

```
Ryujinx.HLE.Exceptions.UndefinedInstructionException: The instruction at 0x000000000f8ead60 (opcode 0x0ee1e002) is undefined!
```

### Instruction name
```
pmull v2.1q, v0.1d, v1.1d
```

### Required by: 
- DOOM Eternal

PR #1817, opened 2020-12-15
CPU (A64): Add Pmull_V Inst. with Clmul fast path for the "1/2D -> 1Q" variant & Sse fast path and slow path for both the "8/16B -> 8H" and "1/2D -> 1Q" variants; with Test.

LDj3SNuD:
Fix V128 Shl/Shr when shift = 0.

A32: Add Vmull_I P64 impl. with Test (P8/P64).

---

Closes https://github.com/Ryujinx/Ryujinx/issues/1791.


LDj3SNuD:
Add Clmul fast path for the 128 bits variant.

gdkchan:
Would be nice to share both the slow and fast paths with the Arm32 implementation (see https://github.com/Ryujinx/Ryujinx/blob/3332b29f01993d22195f417d39ae4074d53fb55c/ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs#L1374), btw, the Arm32 implementation currently lacks the 128-bit support.

LDj3SNuD:
Generally fast paths would not appear to be sharable; however I could open a new PR to add such modified fast paths in A32 as well; slow paths on the other hand are and what I could do is add the 8-bit variant by calling that sharable method in the comment. Let me know.

gdkchan:
> I could do is add the 8-bit variant by calling that sharable method in the comment. Let me know.

Sounds fine. Any reason for not adding the slow path of the 128-bit variant?

LDj3SNuD:
Is it worth it? It would take a lot of IR instructions.

gdkchan:
All other instructions have a slow paths (including ones with hundreds of IR instructions such as the LDn/STn), I don't see why we wouldn't do it for this one.

gdkchan:
Calling a C# implementation should be fine too, I don't expect this instruction to run very often anyway.

LDj3SNuD:
I thought about it but it would have to be run twice, for the low and high 64 bit part; currently we cannot return 2 Operands, can we?

gdkchan:
No, but it can return a `V128` with the 128-bit value.

LDj3SNuD:
And can I use C# intrinsics? If yes, I see no problem then.

gdkchan:
It would defeat the purpose of having the slow path if it depends on CPU specific instructions to work.

LDj3SNuD:
I will see what I can do.

LDj3SNuD:
Added slow path, both variants.

gdkchan (approved):
lgtm, thanks, great work. It would be nice to also update `Vmul_I` to support the 128-bit variant.

Issue #1800, opened 2020-12-11
Confirmation dialog after clicking "Check for Updates" needs two clicks to confirm


voidcat0:
### What's the issue you encountered?

The "Yes" button to confirm that Ryujinx should now restart after the update, needs to be clicked twice to work.
The first click seems to be ignored.
### How can the issue be reproduced?

Click the "Check for Updates" menu and let it search for an update. When Ryujinx asks for confirmation to restart, click the button and notice it's not doing anything. Click again to actually restart.  
_To test this I recommend manually downloading an older version._

### Log file

There is no entry in the log file for this.

### Environment?

- Ryujinx version: 1.0.6020</br>
- Game version: N/A
- System Specs:
  - OS: Operating System: Microsoft Windows 10.0.17763 (X64)
  - CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
  - GPU: NVIDIA Corporation GeForce GTX 1070/PCIe/SSE2 (4.6.0 NVIDIA 457.09)
  - RAM: 16GB (2x8GB - 2666Mhz)

### Additional context?

None

Doctorwho1909:
it's not a bug it's a feature

AcK77:
I'm not able to reproduce this issue, when a new update is available, this window appears:

![image](https://user-images.githubusercontent.com/4905390/102315129-e7f03480-3f73-11eb-8de8-475790dd25d2.png)

Then you click `Yes` and it's download the new version.
When the new version is downloaded, you got this:

![image](https://user-images.githubusercontent.com/4905390/102315225-12da8880-3f74-11eb-814c-819505e3e250.png)

And you have to click `Yes` to restart Ryujinx. It's the expected behavior.

Do I have misunderstood something ?

saldabain:
On the second prompt, the first click on the Yes button appears to be a no op, thats what the issue is about. You need to click it twice to actually restart.

AcK77:
I don't have to click twice here, but I'll try with an old version to see if I can get the issue.

voidcat0:
Not sure what to add to this issue other than: No matter which (publicly available) version I try, I get this behaviour. I just updated to 6127 and I still get this behaviour.
Btw, you don't have to use an "old" version, I only added that line to make sure people know that downloading any version below the most recent would make it find a new version.

AcK77:
Fixed in #1836

PR #1836, opened 2020-12-26
Fix button press issue

AigleSpartiate:
A small code change to fix the issue that was forcing the user to press the "Yes" button twice to restart Ryujinx after it had been updated, instead of only once.
Fixes recently opened issue #1800


Thog:
Nice catch 👍

AcK77 (approved):
Lgtm!

Thog (approved):

PR #1838, opened 2020-12-28
audio: Fix wrong buffer size check

AcK77:
This PR fix a wrong buffer size check in `ListAudioDeviceName` and `ListAudioDeviceNameAuto`.
Now the error isn't reached anymore and the device name is written in the buffer.


gdkchan:
Why not check if before the loop? Like `if (deviceNames.Length * AudioDeviceNameSize > size) { // ... }`.

AcK77:
Closed because of inaccuracy.

PR #1839, opened 2020-12-28
surfaceflinger: Fix fences callback registering issue

AcK77:
This PR fixes a regression introduced in #1741. The actual implementation do the assumption of fences always exists and then it register the callback.
Homebrews may not use fences, so the code crashes when it try to register the callback.

image


Thog (approved):

gdkchan (approved):
lgtm, but I recommend checking with @riperiperi since hes the one that made the change iirc.

riperiperi (approved):
That's reasonable, nice catch.

PR #1844, opened 2020-12-29
Enable Profiled Persistent Translation Cache (PPTC) by default

EmulationFanatic:
Enables PPTC by default. Enough time has passed and enough games tested that it should be the preferred setting for 99.9% of games. In case we are worried about PPTC files being put on the user's disk, shader cache already does this by default and generates about the same size files on average, so this should no longer be a concern.

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationState.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationState.cs
@@ -477,7 +477,7 @@ namespace Ryujinx.Configuration
             CheckUpdatesOnStart.Value              = true;
             Graphics.EnableVsync.Value             = true;
             Graphics.EnableShaderCache.Value       = true;
-            System.EnablePtc.Value                 = false;
+            System.EnablePtc.Value                 = true;
             System.EnableFsIntegrityChecks.Value   = true;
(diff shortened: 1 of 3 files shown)


LDj3SNuD (commented):
I think these lines should also be changed to `true`:
https://github.com/Ryujinx/Ryujinx/blob/EmulationFanatic-patch-1/Ryujinx/_schema.json#L1218
https://github.com/Ryujinx/Ryujinx/blob/EmulationFanatic-patch-1/Ryujinx/Config.json#L26

EmulationFanatic:
> I think these lines should also be changed to `true`:
> https://github.com/Ryujinx/Ryujinx/blob/EmulationFanatic-patch-1/Ryujinx/_schema.json#L1218
> https://github.com/Ryujinx/Ryujinx/blob/EmulationFanatic-patch-1/Ryujinx/Config.json#L26

@LDj3SNuD 
Will make these changes shortly.

AcK77 (approved):
Lgtm! Good idea!

LDj3SNuD (approved):
Without wanting to influence the final choice, I approve from a code point of view.

Thog (approved):

PR #1842, opened 2020-12-29
Audio: Include OpenAL with Ryujinx

jduncanator:
This includes a version of OpenAL, from https://openal-soft.org, with Ryujinx out of the box, removing the requirement to manually install OpenAL.

Please note, OpenAL will now _always_ be the default audio backend, over SoundIO.

PR #1847, opened 2020-12-30
Distribute with OpenAL Soft on Windows

Thog:
This remove dependency on OpenAL EA on Windows and allows to operate with both backend without any user interventions.

Based on #1842 and supersede it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Ryujinx system files are stored in the `Ryujinx` folder. This folder is located 
 
  - **Audio**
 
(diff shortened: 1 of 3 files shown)


jduncanator (approved):

gdkchan:
If we are doing that for OpenAL, then I suppose it should be done for SoundIO aswell.

EmulationFanatic:
Working fine here after removing all traces of OpenAL from my system and then installing this appveyor build.

AcK77 (approved):

PR #1859, opened 2021-01-02
gui: Refactoring Part 1

AcK77:
The GUI is currently based on the original implementation, which means a lot of things can now be cleaned.
To avoid a too big PR, I prefer to purpose those changes now, and will probably do more later if we keep GTK a bit longer.

What's done:

- Implement a Windows helper to force NVidia GPU usage if it can.
- Move DiscordIntegration/Motion/Updater in a Modules subfolder, those are cleaned up too.
- Delete Migration code since it's deprecated now (one year after the original migration).
- Move window to the Windows folder.
- Move widget to the Widgets folder.
- Rename "assets" files to something more cleaner and move them in a "Resources" folder.
- Rewriting the About window to avoid using a glade file, it uses a AboutWindow.Designer.cs instead.
- Rename DllImport of WindowsMultimediaTimerResolution to follow C# convention.
- No more static classes/methods.
- Fix title update selected idem when « No Update » is selected.
- Some cleanup here and there...

I suggests some testing to avoid any further issue.


Ryujinx/Ui/Windows/ControllerWindow.cs
```diff
@@ -417,37 +409,37 @@ private InputConfig GetValues()
...
-                Enum.TryParse(_x.Label,            out Key rButtonX);
-                Enum.TryParse(_y.Label,            out Key rButtonY);
-                Enum.TryParse(_plus.Label,         out Key rButtonPlus);
-                Enum.TryParse(_r.Label,            out Key rButtonR);
-                Enum.TryParse(_zR.Label,           out Key rButtonZr);
-                Enum.TryParse(_rSl.Label,          out Key rButtonSl);
-                Enum.TryParse(_rSr.Label,          out Key rButtonSr);
-
-                int.TryParse(_dsuServerPort.Buffer.Text, out int port);
+                _ = Enum.TryParse(_lStickUp.Label,     out Key lStickUp);
```
gdkchan, line 412: Why the `_ =`?

gdkchan (approved):
lgtm except for the `_ =`.

ShahilSharma:
So far I haven’t noticed any issues and I’ve gone through the entire GUI twice, Nice work!

Thog (approved):

PR #1870, opened 2021-01-07
Allow out of bounds storage buffer access by aligning their sizes

### LINKED ISSUE #1689: Yo Kai Watch 4: Rendering is too bright

[ Issue for Yo Kai Watch 4 when go outside from the main character's home the landscape is going to white n only the map seems normal. ]</br>

[ first the character have story inside the home, and the next story is the character need to buy some item in store, when the character going outside the issue happened ]

### Log file
[Ryujinx_2020-11-08_21-01-54.log](https://github.com/Ryujinx/Ryujinx/files/5507763/Ryujinx_2020-11-08_21-01-54.log)
</br>

### Environment?

- Ryujinx version: 1.0.5639

- Game version: 2.2.0</br>

- System Specs:
  - OS: *(e.g. Windows 10)*
  - CPU: *(e.g. i7-9750)*
  - GPU: *(e.g. NVIDIA GTX 1050 Ti)*
  - RAM: *(e.g. 8GB)*
- Applied Mods : [ No ]

### Additional context?

ryujinx exe Screenshot 2020 11 09 - 09 48 02 07

gdkchan:
Some games such as Yo-kai watch performs a out of bounds storage buffer access. It says that the size is 4 bytes, but accesses (reads and writes) 256 bytes on a compute shader. In fact the game allocates enough memory for that (the buffers are 256 bytes apart), its just the size that is wrong on this particular call. The shader does not use this size for anything, so this is not a problem on the Switch, but on OpenGL, looks like theres some bounds checking going on preventing values after the first 4 bytes from being read/written.

This fixes this particular issue by binding the entire buffer rather than just the range that the game says its using. Unfortunately there is no way to guess the real size of the buffer, so this will just effectively page align the size, which works in this case as the real buffer size is way smaller than a page (256 bytes vs 4096 bytes that is the page size).

This solution is not perfect, it won't work for all possible cases where the size is incorrect, but it should work in all cases observed on this particular game. The only way I can think to get it working in all cases is mirroring the guest memory map on the host, which would require a full rewrite of the buffer management code, and I'm not sure if its even feasible.

Closes #1689

I recommend testing on games that makes use of SSBOs. This should not have any performance impact. I only tested Yo-kai watch 4, but the other Yo-kai watch games with a similar issue might be fixed aswell.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
@@ -591,7 +591,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
 
                 if (bounds.Address != 0)
                 {
-                    sRanges[bindingInfo.Binding] = GetBufferRange(bounds.Address, bounds.Size, bounds.Flags.HasFlag(BufferUsageFlags.Write));
+                    sRanges[bindingInfo.Binding] = GetBufferRange(bounds.Address, bounds.Flags.HasFlag(BufferUsageFlags.Write));
                 }
             }
 
@@ -764,7 +764,9 @@ namespace Ryujinx.Graphics.Gpu.Memory
 
                     if (bounds.Address != 0)
                     {
-                        ranges[bindingInfo.Binding] = GetBufferRange(bounds.Address, bounds.Size, bounds.Flags.HasFlag(BufferUsageFlags.Write));
+                        ranges[bindingInfo.Binding] = isStorage
+                            ? GetBufferRange(bounds.Address, bounds.Flags.HasFlag(BufferUsageFlags.Write))
+                            : GetBufferRange(bounds.Address, bounds.Size, bounds.Flags.HasFlag(BufferUsageFlags.Write));
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
```diff
@@ -821,6 +823,16 @@ public void CopyBuffer(GpuVa srcVa, GpuVa dstVa, ulong size)
             dstBuffer.Flush(dstAddress, size);
         }
 
+        /// <summary>
+        /// Gets a buffer sub-range starting at a given memory address.
+        /// </summary>
+        /// <param name="address">Start address of the memory range</param>
+        /// <returns>The buffer sub-range starting at rhe given memory address</returns>
```
jduncanator, line 830: Typo here, should be `the` not `rhe`

jduncanator (approved):
Looks good to me, I love the more "straight forward" fixes for issues like these.

EmulationFanatic:
From Discord user kakasita: "Fixed most of the cases except:"
image

tfqmstrmnd:
> 
> 
> From Discord user kakasita: "Fixed most of the cases except:"
> image

dis game can't saving...right?

Cronos-0:
Fixed too bright lighting for Yo kai Watch 4 and Yo kai Gakuen Y WaiWai Gakuen Seikatsu.

bandicam 2021-01-10 04-32-29-255
bandicam 2021-01-10 04-39-21-307

It looks like Yo kai Watch 1 still has some display issues.

bandicam 2021-01-10 04-22-09-802
bandicam 2021-01-10 04-24-22-112

Thog (approved):

PR #1878, opened 2021-01-08
Support conditional on BRK and SYNC shader instructions

gdkchan:
Like the BRAnch instruction, BRK and SYNC also support a conditonal, in addition to the predicate, which allows compound predication of the jump. The CC register is set by previous instruction based on the result, setting the NZCV flags. Example:
```
@P0 BRK CC.NEU
```
In this case, the branch is taken if P0 is true *and* the Z flag is clear, as the zero flag should be set if the result is zero, and cleared otherwise. This is usually used with ISET/FSET with `RZ.CC` as destination that will set the zero and negative flags based on the result, which can be -1 (if comparison is true) or 0 (if its false).

This improves rendering on Monster Hunter Rise, but there are still other bugs preventing most objects from being draw.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -34,7 +34,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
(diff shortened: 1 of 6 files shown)


jduncanator:
Should this bump some shader cache version to invalidate existing shader caches?

gdkchan:
> 
> 
> Should this bump some shader cache version to invalidate existing shader caches?

Done.

riperiperi (approved):
Nice catch!

jduncanator (approved):

Issue #1880, opened 2021-01-09
An exception of type 'System.ArgumentOutOfRangeException' occurred in Expection when lauching homebrew games: "Specified argument was out of the range of valid values"


macabeus:
## Bug Report

When launching homebrew games, such as [2048](https://github.com/FlagBrew/2048), there is the following exception:
![image](https://user-images.githubusercontent.com/9501115/104082736-4eddd200-5230-11eb-8c72-2a566b2cae93.png)

PR #1881, opened 2021-01-09
android-fence: call callback when fence is invalid

macabeus:
Closes https://github.com/Ryujinx/Ryujinx/issues/1880

It's the fix made by rip in peri peri on Discord ([message link](https://discord.com/channels/410208534861447168/410208534861447170/795112697674465330))


AcK77 (approved):

riperiperi (approved):
Thanks for PR'ing this. This callback should be called if there are no more fences to wait on, so running the callback immedately if it's invalid makes sense here. (this may actually supersede an earlier PR that changed things in SurfaceFlinger)

PR #1885, opened 2021-01-09
Implement friendlier portable mode

### LINKED ISSUE #1879: Enhance portable mode
## Feature Request

This is a (basic and must-have) feature request that I do supposing that **Ryujinx** still not implemented it (otherwise please let me know):

Please consider to implement portable mode, that is the **bis**, **games**, **sdcard** and the other directories and any other files must be read/written from root directory of **Ryujinx.exe**. 

I cannot think of a good, actual and active emulator that is not portable itself or that does not provide the user a functionality to make it portable, like **PCSX2**, **RPCS3**, **Dolphin**, **Cemu**, **Citra**, **Cxbx-R**, **Xemu*** (**Xemu**'s devs will implement it in a next release) and of course **Yuzu**, just to mention some of them.

Please take into account that a vast amount of users will need emulators to be portable, not to save settings in the user roaming folder like **Ryujinx** does :(

Just let the end user the chance to decide the default behavior of the software in this meaning.

One easy way for all users to set portable mode probably would be to let them create a "portable.txt" file in the root folder of **Ryujinx.exe**, and then **Ryujinx** should do the magic to recognize the installation as a portable mode. Or through an option in the emulator UI.

Just consider this feature request. Thanks in advance.

mageven:
Users seem to have trouble using the "custom data directory" feature for portability. This implements an extra (arguably redundant) way to achieve it. Creating a `portable` dir in the exe base directory now activates portable mode.

This code prioritizes Custom Data argument -> Portable dir -> User Profile.

**EDIT:** Removed first run dialog.

**NOTE1:** I was unable to figure out a way to stop GTK from creating a 'recently-used.xbel' file in user profiles used for Recent items in file dialogs. Besides that, the application should be fully portable.

**NOTE2:** It's still a bit jarring to find the portable directory among the plethora of files and folders. This would have to be resolved in the future.

Closes #1879

Tested on Windows. Should work on others but please do test.

~__I highly recommend blocking this from merge until #1594 is resolved.__~
To unblock this PR, using portable mode blocks updates for now.

### DIFF
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -47,19 +47,7 @@ namespace Ryujinx.Common.Configuration
             string userProfilePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
             string portablePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, DefaultPortableDir);
 
-            if (baseDirPath != null && baseDirPath != userProfilePath)
-            {
-                if (!Directory.Exists(baseDirPath))
-                {
-                    Logger.Error?.Print(LogClass.Application, $"Custom Data Directory '{baseDirPath}' does not exist. Using defaults...");
-                }
-                else
-                {
-                    BaseDirPath = baseDirPath;
-                    Mode = LaunchMode.Custom;
-                }
-            }
-            else if (Directory.Exists(portablePath))
+            if (Directory.Exists(portablePath))
             {
                 BaseDirPath = portablePath;
                 Mode = LaunchMode.Portable;
@@ -69,6 +57,21 @@ namespace Ryujinx.Common.Configuration
                 BaseDirPath = userProfilePath;
                 Mode = LaunchMode.UserProfile;
             }
+
+            if (baseDirPath != null && baseDirPath != userProfilePath)
+            {
+                if (!Directory.Exists(baseDirPath))
+                {
+                    Logger.Error?.Print(LogClass.Application, $"Custom Data Directory '{baseDirPath}' does not exist. Falling back to {Mode}...");
+                }
+                else
+                {
+                    BaseDirPath = baseDirPath;
+                    Mode = LaunchMode.Custom;
+                }
+            }
+
+            BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
             SetupBasePaths();
         }
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -42,7 +42,7 @@ namespace Ryujinx.Common.Configuration
             KeysDirPathUser = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".switch");
         }
 
-        public static void Initialize(string baseDirPath, Func<bool> FirstRunPortableChoice)
+        public static void Initialize(string baseDirPath)
         {
             string userProfilePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
             string portablePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, DefaultPortableDir);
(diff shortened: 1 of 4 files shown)


EliEron:
I'm a bit dubious about the first-run dialog. In my experience users tend to just blindly press "Yes" when they are presented with an unprompted dialogue. Assuming it to be the default behavior without bothering to read the full message. So usage of portable mode will likely skyrocket if this dialogue is added.

EmulationFanatic:
> I'm a bit dubious about the first-run dialog. In my experience users tend to just blindly press "Yes" when they are presented with an unprompted dialogue. Assuming it to be the recommended behavior without bothering to read the full message. So usage of portable mode will likely skyrocket if this dialogue is added.

This doesn't inherently seem to be a big problem, though adding something to the title bar or log indicating if Ryujinx is in portable mode could be useful (not sure how hard that is to implement.)

jduncanator:
I’m not a fan of the first run dialog either. We already prompt users for far too many things on first run, and I think the percentage of users who legitimately want to run Ryujinx in portable mode on first run would be far far lower than those who don’t.

I’m happy with the portable folder change, but don’t think we should be adding yet another forced dialog prompt for a feature rarely used.

Doctorwho1909:
add the ability to go portable in the general tab or under tools

mageven:
Agreed with both @EliEron and @jduncanator's statements. I have removed the dialog.

To unblock this PR (and since I noticed the Updater downloads to system temp directories), I have temporarily disabled updates for portable installs. This requirement can be relaxed when #1594 is resolved.

By the way, I think "default portable mode" isn't a bad idea in the future as there's benefit in reusing the built caches among shared owners. Especially after Ryujinx supports guest user profiles.

> adding something to the title bar or log indicating if Ryujinx is in portable mode could be useful (not sure how hard that is to implement.)

There already is a log that shows launch mode (was an extension to the custom path log).

> add the ability to go portable in the general tab or under tools

There are a few considerations here. The biggest hurdle is adding a menu item implies data migration. So all the data has to be moved/copied to the new location and it'd have to be robust (the dir can grow large). If you mean we have to switch to the new location while the program is running, then some systems would require overhauls. Right now everything is set up on boot. If you mean without data transfer, then all it would do is create an empty directory named "portable" which imo the user can do quite easily.

mageven:
Some old FS code doesn't play well with relative paths which is possible via `-r`. I convert them to absolute paths now.
Fixes https://github.com/Ryujinx/Ryujinx/issues/1879#issuecomment-757577331

AcK77 (approved):
Lgtm!

ElektroStudios:
Any E.T.A for this portable mode feature?

riperiperi (approved):
Looks good here.

CeruleanSky:
> **NOTE1:** I was unable to figure out a way to stop GTK from creating a 'recently-used.xbel' file in user profiles used for Recent items in file dialogs. Besides that, the application should be fully portable.

The file is created with the following call in [gtk recent manger](https://github.com/GNOME/gtk/blob/3.24.8/gtk/gtkrecentmanager.c#L579)
```
  return g_build_filename (g_get_user_data_dir (),
                           GTK_RECENTLY_USED_FILE,
                           NULL);
```
[g_get_user_data_dir](https://github.com/GNOME/glib/blob/2.60.3/glib/gutils.c#L1299) is from glib and it respects the environmental variable `XDG_CONFIG_HOME`

mageven:
@CeruleanSky Thanks for the suggestions.

- I'm a bit concerned by overriding user-set env variables. But, I'm fine with adding it with maintainers' permission. It'll only be overridden for portable mode.
- I was aware of `PROP_FILENAME`. It is sadly installed with `G_PARAM_CONSTRUCT_ONLY` so can't be changed after construction, say for `RecentManager.Default`; and I couldn't install a custom `RecentManager` to `FileChooserDialog`.

In any case, this file and solutions won't affect the long-term as apparently there's plans to move away from GTK.

EmulationFanatic:
I know this is already approved. I'm just adding that I tested the detection of the "portable" folder and associated operation successfully.

Doctorwho1909:
Moved ryujinx to fully portable and it works super well no issues found great work

julianxhokaxhiu:
Just wanted to get notified when this PR will get merged. I hope it will be soon. Thanks!

EmulationFanatic:
> Just wanted to get notified when this PR will get merged. I hope it will be soon. Thanks!

FWIW on the right hand side you can subscribe without having to reply in the thread :)
image

julianxhokaxhiu:
> > Just wanted to get notified when this PR will get merged. I hope it will be soon. Thanks!
> 
> FWIW on the right hand side you can subscribe without having to reply in the thread :)
> image

Thanks :) I'm a GH veteran, but sometimes that doesn't really work, that's why I left a message over here. Still, appreciated.

PR #1892, opened 2021-01-10
Fix compute reserved constant buffer updates

gdkchan:
NVN uses `LoadInlineData` to load the compute QMD structure *and* the reserved driver constant buffer data (thats the only thing it ever seems to use `LoadInlineData` for). Due to the sequence number optimization on buffer updates, that prevents potentially costly modification checks, the compute reserved constant buffers were not being updated, since `LoadInlineData` currently does not update the sequence number. This PR fixes that.

This fixes a bug on Monster Hunter Rise demo where the screen would randomly turn black with white elements due to compute reading values from the wrong SSBO.

The only cases affected by this should be compute reading reserved constant buffers, mainly those using LDG/STG where the shader translator can't figure out which SSBO it is accessing. So, to sum it up, it might also fix other games using compute and SSBOs.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Inline2Memory.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Inline2Memory.cs
@@ -126,6 +126,8 @@ namespace Ryujinx.Graphics.Gpu.Engine
             }
 
             _finished = true;
+
+            _context.AdvanceSequence();
         }
     }
 }


riperiperi:
Very nice catch, would have been hard to trace back.

riperiperi (approved):

AcK77 (approved):

PR #1899, opened 2021-01-12
Pacify updater

mageven:
The current updater marks everything except `.log` files for deletion.
With this PR, it now deletes only select directories (Windows only, Linux and MacOS don't have these) and all top-level files in the base directory. So, if there's other directories (like portable or custom), their contents will be spared.

Also used this opportunity to log errors from the updater (to easily identify future issues).

This PR does not claim to solve #1594 100%. Only to unblock #1885. So, I have not added an auto-close. 
I do think this is enough for most people, but feel free to close this if it isn't the way to go.

### DIFF
--- a/Ryujinx/Modules/Updater/Updater.cs
+++ b/Ryujinx/Modules/Updater/Updater.cs
@@ -7,7 +7,9 @@ using Ryujinx.Common.Logging;
 using Ryujinx.Ui;
 using Ryujinx.Ui.Widgets;
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Net.NetworkInformation;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


CeruleanSky:
An alternative solution that might be worth consideration is to do something more traditional like how uninstallers work. Create an UpdateList.txt containing a list of files that the updater should delete that is bundled, kept, and extracted from the ryujinx.zip/tar.gz/etc the updater downloads.  Then the next time the updater is invoked, if it sees that file in the Ryujinx install location, it can process the listing deleting the proper files from the previous install before going onto the next update step.

The build system can run something like `forfiles /s /m *.* /c "cmd /c echo @relpath" `or the correct command for the OS and include the listing in the distribution, or a listing could be generated by the updater itself after it unpacks the files in the temporary directory to use for the next update.

mageven:
Yes I already considered that. Feel free to make a PR with that approach and I'll close this. I didn't want to spend too much energy on this Updater as I'm not a fan of an application renaming its own dlls/exes while running. So, I'm waiting for a simpler standalone one.

Ryujinx/Modules/Updater/Updater.cs
```diff
@@ -321,6 +320,29 @@ public static bool CanUpdate(bool showWarnings)
...
+        // NOTE: This method should always reflect the latest build layout
+        private static IEnumerable<string> EnumerateFilesToDelete()
+        {
+            var files = Directory.EnumerateFiles(HomeDir); // all files directly in base dir
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // On Windows, GtkSharp.Dependencies adds these extra dirs
+                // the other dependencies are just files in the base dir so they got marked above
+                string[] dependencyDirs = new string[] { "bin", "etc", "lib", "share" };
```
riperiperi, line 332: Perhaps a static string array up at the top of the file? Would be easier to find if we need to change it.

Ryujinx/Modules/Updater/Updater.cs
```diff
@@ -321,6 +320,29 @@ public static bool CanUpdate(bool showWarnings)
             return true;
         }
 
+        // NOTE: This method should always reflect the latest build layout
+        private static IEnumerable<string> EnumerateFilesToDelete()
+        {
+            var files = Directory.EnumerateFiles(HomeDir); // all files directly in base dir
```
riperiperi, line 326: Capital letter, full stop at the end.

riperiperi (approved):
Nice. I didn't actually know that the linux builds had no subdirectories, but now I do. 🤯

AcK77 (approved):

PR #1901, opened 2021-01-12
Implement shader CC mode for ISCADD, X mode for ISETP and fix STL/STS/STG with RZ

gdkchan:
This implements the CC mode on the ISCADD shader instruction (which is supposed to set the NZCV flags based on the post-shift addition result, as far I can tell), and the X (extended) mode on the ISET and ISETP instruction (used for 64-bit and longer comparisons).

This also fixes a bug on the STL, STS and STG shader instruction where using RZ as input with a length greater than 32-bit would only write the value once. Now it writes 0 up to the size of the write (1, 2 or 4 words).

This fixes normal vectors on animated characters in Monster Hunter Rise DEMO improving the rendering a little bit, but there are still other things to be fixed (coming soon).

NOTE: This was a shader bug, so it will invalidate the shader cache.

### DIFF
--- a/Ryujinx.Graphics.Shader/Translation/EmitterContextInsts.cs
+++ b/Ryujinx.Graphics.Shader/Translation/EmitterContextInsts.cs
@@ -406,9 +406,39 @@ namespace Ryujinx.Graphics.Shader.Translation
             return context.Add(Instruction.CompareEqual, Local(), a, b);
         }
 
+        public static Operand ICompareGreater(this EmitterContext context, Operand a, Operand b)
+        {
+            return context.Add(Instruction.CompareGreater, Local(), a, b);
+        }
+
+        public static Operand ICompareGreaterOrEqual(this EmitterContext context, Operand a, Operand b)
+        {
+            return context.Add(Instruction.CompareGreaterOrEqual, Local(), a, b);
+        }
+
+        public static Operand ICompareGreaterOrEqualUnsigned(this EmitterContext context, Operand a, Operand b)
+        {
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Shader/Translation/EmitterContextInsts.cs
```diff
@@ -406,11 +406,41 @@ public static Operand ICompareEqual(this EmitterContext context, Operand a, Oper
             return context.Add(Instruction.CompareEqual, Local(), a, b);
         }
 
+        public static Operand ICompareGreater(this EmitterContext context, Operand a, Operand b)
+        {
+            return context.Add(Instruction.CompareGreater, Local(), a, b);
+        }
+
+        public static Operand ICompareGreaterOrEqual(this EmitterContext context, Operand a, Operand b)
```
riperiperi, line 414: These aren't used, I'm guessing they're just here for completeness?

riperiperi (approved):
Very cool! No issues I can find, I do like seeing TODO comments removed.

Ryujinx.Graphics.Shader/Translation/EmitterContextInsts.cs
gdkchan: Yea, I thought I would need them initially, but turns out it was not necessary. I can remove it or just leave it there, either way is fine to me.

Thog: I think it's fine to me to keep

Thog (approved):

jduncanator (approved):

PR #1902, opened 2021-01-12
Implement clear buffer (fast path)

gdkchan:
NVN uses the DMA engine remap functionality to do buffer clears. Basically you can specify a constant as the copy source, and it willl fill the buffer with the constant value. This is similar to OpenGL `glClearBufferSubData` function. This implements a fast path using this function. The full functionality still needs to be implemented (and it can't be done with the host API alone, since its possible to intermix sources from another buffers and constants, swizzle etc). However all functionality used by NVN should be implemented.

This also fixes a glitch on vector normals on Monster Hunter Rise (which could contain garbage values as the buffer was not being cleared before), this improving the rendering, but there are also other issues that needs to be fixed (texture related).

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -9,6 +9,8 @@ namespace Ryujinx.Graphics.GAL
 
         void BeginTransformFeedback(PrimitiveTopology topology);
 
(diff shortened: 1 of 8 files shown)


riperiperi (approved):
A bit weird having to fully qualify ClearBuffer at the end there (due to the name on the interface, I'm fine with that), but otherwise, nice work. 😌

Thog (approved):

jduncanator (approved):

PR #1905, opened 2021-01-13
Support for resources on non-contiguous GPU memory regions

gdkchan:
This aims to support GPU textures mapped at non-contiguous CPU regions, as some games requires it (such as the Monster Hunter Rise Demo and possibly some UE4 games).

**Problem:**

Games can map GPU virtual memory regions to non-contiguous physical regions using the remap ioctl. For example, we may have a GPU region, from 0 to 0x2000, where the range from 0..0x1000 is mapped to physical address 0x8000, and the range 0x1000..0x2000 is mapped to physical address 0x15000 (the values are made up and are supposed to be just an example). Right now the texture data would be read from the physical range 0x8000..0x10000, which is wrong because only half of the data is mapped there. The correct behavior would be reading the first half of the data from 0x8000..0x9000, and the second half from 0x15000..0x16000.

**Solution:**

The solution here was changing the way how texture addresses (or rather, regions) are represented. Instead of having a single address and size, a new structure was added, called `MultiRange`, that can store multiple ranges of memory (composed of a address and size pair) pointing to where the texture data is located. A new method was added to the GPU memory manager called `GetPhysicalRegions` that can get such ranges from a single GPU virtual memory range. New overloads of the read, write and tracking functions were added that takes a `MultiRange` instead of a single address and size pair.

One thing that it does right now is calculating the offset of a texture view from its address, and the parent texture address. On master, this is as simple as doing `viewOffset = viewTextureAddress - parentTextureAddress`. With `MultiRange`, this is not so simple because there are now multiple, possibly disjoint addresses. So now, what it does is trying to find the view range sequence inside the parent texture range sequence.

Another possible solution would be always using the GPU virtual address and read the data from that address, doing address translation as needed and reading the data from the correct region. However this has some issues. First, memory aliasing (basically, different GPU virtual memory region mapped to the same physical one) wouldn't work properly, as it would create separate textures for the different GPU regions. Meanwhile if we use physical address (in our case this is a CPU virtual address right now), the memory aliasing work because after address translation, the address will be the same, and it will pull the same texture from the cache. The second problem is texture flushing. When reading from CPU, it only has a CPU virtual address to be used for flushing, finding the correct texture on the cache would require some sort of inverse address translation. Furthermore, it would also need a way to handle the possibility of multiple textures mapped into the same address.

The GPU memory manager was updated, now all methods there should correctly support reading or writing data at non-contiguous ranges. Before it just assumed this could never happen.

**Future work:**

Right now this is only supported for textures. It should be extended to buffers aswell.

**Testing:**

While testing, I recommend looking for the following:
- Games that worked properly before, but regressed in some way with those changes.
- Decrease in performance (it is not expected to have any noticeable performance hit).

### DIFF
--- a/Ryujinx.Cpu/MemoryManager.cs
+++ b/Ryujinx.Cpu/MemoryManager.cs
@@ -13,7 +13,7 @@ namespace Ryujinx.Cpu
     /// <summary>
     /// Represents a CPU memory manager.
     /// </summary>
-    public sealed class MemoryManager : IMemoryManager, IVirtualMemoryManager, IDisposable
+    public sealed class MemoryManager : IMemoryManager, IVirtualMemoryManager, IWritableBlock, IDisposable
     {
         public const int PageBits = 12;
         public const int PageSize = 1 << PageBits;
@@ -202,12 +202,12 @@ namespace Ryujinx.Cpu
             WriteImpl(va, data);
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         /// <summary>
         /// Writes data to CPU mapped memory.
         /// </summary>
         /// <param name="va">Virtual address to write the data into</param>
         /// <param name="data">Data to be written</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void WriteImpl(ulong va, ReadOnlySpan<byte> data)
         {
             try
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -368,6 +368,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             SetInfo(new TextureInfo(
                 Info.Address,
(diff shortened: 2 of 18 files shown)


Doctorwho1909:
image
it fixed most of the weird texture issues in hellblade

Niwu34:
DaemonxMachina
Big improvement

riperiperi (approved):
I've looked through it all and everything looks sensible. Performance was a bit of a worry, but I checked in a number of games and it hasn't been affected. Flushing textures also seems to be working as expected. I guess the main cost will be paid when setting up these textures.

I also found that this PR fixes empty/invalid shaders in MH Rise.

Either way, looks great! I was wondering how you'd keep the cpu va overlaps stuff working for this, definitely not disappointed.

Doctorwho1909:
image
with the new changes HellBlade has Perfect textures

Doctorwho1909:
image
spyro is perfect !!

Doctorwho1909:
image

Thog (approved):

PR #1906, opened 2021-01-13
Emulate a circular zone for keyboard analog sticks

mageven:
I saw a few user complaints on discord that diagonal inputs are not working with a keyboard controller in certain games. Some games get confused as the current code emulates a square zone whereas they expect a circular one (which is natural). Thought of using a lookup table too but ultimately this seemed cleaner. 

Also previously, the keyboard analog sticks always preferred down-right. So, if say, up & down were pressed together it'd be registered as down. I changed this behavior to be additive so opposite directions cancel out. There's probably other things that could be improved in the area but I'd like to keep this fix simple.

### DIFF
--- a/Ryujinx/Ui/KeyboardController.cs
+++ b/Ryujinx/Ui/KeyboardController.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using OpenTK;
 using OpenTK.Input;
 using Ryujinx.Common.Configuration.Hid;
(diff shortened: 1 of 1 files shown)


jduncanator (approved):
Nice, elegant fix to the problem.

riperiperi (approved):
Nice one! ☺️

AcK77 (approved):

PR #1909, opened 2021-01-13
Implement Sfdnsres.GetAddrInfo

### LINKED ISSUE #637: sfdnsres IResolver: 6 (GetAddrInfo) not implemented
IResolver GetAddrInfo is not implemented.

## Service description:
[Switchbrew](https://switchbrew.org/wiki/Sockets_services#GetAddrInfo) | [Reswitched](https://reswitched.github.io/SwIPC/ifaces.html#nn::socket::resolver::IResolver(6))

```
GetAddrInfo(bool enable_nsd_resolve, u32, u64 pid_placeholder, pid, buffer<i8, 5, 0> host, buffer<i8, 5, 0> service, buffer<packed_addrinfo, 5, 0> hints) -> (i32 ret, u32 bsd_errno, u32 packed_addrinfo_size, buffer<packed_addrinfo, 6, 0> response)
```

> Takes three type 5 buffers (host, port, and hints), and a type 6 buffer (the output addrinfos). Also takes a u8 (padded to 4 bytes so the next raw parameter can align), a u32, and a u64. The u8 is a boolean for whether to enable "nsd resolve" (1) or not (0). Not sure what the u32 is. It seems to either come from a parameter to GetAddrInfo or be zero. The u64 is most likely a placeholder for the server to copy the PID into and should be zero. Both hints and the output buffer contain serialized addrinfo chains. The hints buffer is sized 0x400 bytes long by default, and the output buffer 0x1000 bytes.
>
>#### Addrinfo Serialization Format
>Each struct addrinfo in the linked list is serialized according to this format and then written to the buffer. All numbers are in network byte order.
>
> | Size (bytes)                      | Name                              | Notes                               |
>| --------------------------------- | --------------------------------- | ----------------------------------- |
>| 4                                 | Magic                             | Needs to be 0xBEEFCAFE. Seriously.  |
>| 4                                 | ai_flags                          |                                     |
>| 4                                 | ai_family                         |                                     |
>| 4                                 | ai_socktype                       |                                     |
>| 4                                 | ai_protocol                       |                                     |
>| 4                                 | ai_addrlen                        |                                     |
>| ai_addrlen ? ai_addrlen : 4       | ai_addr                           |                                     |
>| null-terminated string            | ai_canonname                      |                                     |
> 
> If `ai_addrlen` is zero, `ai_addr` will occupy 4 bytes.
>
> If ai_family is recognized as AF_INET6 (28) or AF_INET (2), ai_addr is read as struct sockaddr_in or struct sockaddr_in6. Otherwise, it's just read as u8[ai_addrlen].
>
> The list should be terminated with a sentinel four-byte zero value.

## Required by:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/91

WilliamWsyHK:
This PR is to implement `nn::socket::resolver::IResolver(6)`, which would close #637.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
@@ -1,5 +1,6 @@
 using Ryujinx.Common.Logging;
 using System;
+using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
@@ -139,40 +140,27 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres
             return result;
         }
 
-        private AddrInfo ParseAddrInfoFromBytes(byte[] addrInfoBytes)
+        private AddrInfo ParseAddrInfoFromBytes(ReadOnlySpan<byte> addrInfoBytes)
         {
-            Span<byte> addrInfoBytesSpan = addrInfoBytes.ToArray().AsSpan();
-            Span<byte> magicSpan = addrInfoBytesSpan.Slice(0, 4);
-            Span<byte> flagsSpan = addrInfoBytesSpan.Slice(4, 4);
-            Span<byte> familySpan = addrInfoBytesSpan.Slice(8, 4);
-            Span<byte> socketTypeSpan = addrInfoBytesSpan.Slice(12, 4);
-            Span<byte> protocolSpan = addrInfoBytesSpan.Slice(16, 4);
-            Span<byte> addrLenSpan = addrInfoBytesSpan.Slice(20, 4);
-
-            if (BitConverter.IsLittleEndian)
-            {
-                magicSpan.Reverse();
-                flagsSpan.Reverse();
-                familySpan.Reverse();
-                socketTypeSpan.Reverse();
-                protocolSpan.Reverse();
-                addrLenSpan.Reverse();
-            }
-
-            int addrLen = BitConverter.ToInt32(addrLenSpan);
+            int magic = BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(0, 4));
+            int flags = BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(4, 4));
+            AddressFamily addressFamily = (AddressFamily)BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(8, 4));
+            int socketType = BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(12, 4));
+            int protocol = BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(16, 4));
+            int addrLen = BinaryPrimitives.ReadInt32BigEndian(addrInfoBytes.Slice(20, 4));
             // If ai_family is recognized as AF_INET6 (28) or AF_INET (2),
             // ai_addr is read as struct sockaddr_in or struct sockaddr_in6.
             // Otherwise, it's just read as u8[ai_addrlen].
-            byte[] addrs = addrInfoBytesSpan.Slice(24, (addrLen != 0 ? addrLen : 4)).ToArray();
-            string canonName = Encoding.ASCII.GetString(addrInfoBytesSpan.Slice(24 + (addrLen != 0 ? addrLen : 4))).Replace("\0", "");
+            byte[] addrs = addrInfoBytes.Slice(24, (addrLen != 0 ? addrLen : 4)).ToArray();
+            string canonName = Encoding.ASCII.GetString(addrInfoBytes.Slice(24 + (addrLen != 0 ? addrLen : 4))).Replace("\0", "");
 
             return new AddrInfo
             {
-                Magic = BitConverter.ToInt32(magicSpan),
-                Flags = BitConverter.ToInt32(flagsSpan),
-                Family = (AddressFamily)BitConverter.ToInt32(familySpan),
-                SocketType = BitConverter.ToInt32(socketTypeSpan),
-                Protocol = BitConverter.ToInt32(protocolSpan),
+                Magic = magic,
+                Flags = flags,
+                Family = addressFamily,
+                SocketType = socketType,
+                Protocol = protocol,
                 Addrs = addrs,
                 CanonName = canonName
             };
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
@@ -218,7 +218,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres
             bufferPosition += 4;
 
             // ai_canonname
-            ReadOnlySpan<byte> canonName = Encoding.ASCII.GetBytes("" + '\0');
+            ReadOnlySpan<byte> canonName = Encoding.ASCII.GetBytes("" + "\0");
             context.Memory.Write((ulong)bufferPosition, canonName);
             bufferPosition += canonName.Length;
 
@@ -466,7 +466,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres
             try
             {
                 hintsAddrInfo = ParseAddrInfoFromBytes(hints);
-                addresses = Dns.GetHostAddresses(Encoding.ASCII.GetString(host));
+                addresses = Dns.GetHostAddresses(Encoding.ASCII.GetString(host).TrimEnd('\0'));
                 addresses = addresses.Where(a => a.AddressFamily == hintsAddrInfo.Family).ToArray();
             }
             catch (SocketException exception)
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
@@ -1,4 +1,5 @@
 using Ryujinx.Common.Logging;
+using Ryujinx.Memory;
 using System;
 using System.Buffers.Binary;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -138,6 +139,94 @@ private List<IPAddress> GetIpv4Addresses(IPHostEntry hostEntry)
             return result;
         }
 
+        private AddrInfo ParseAddrInfoFromBytes(byte[] addrInfoBytes)
```
jduncanator, line 142: This can be a `ReadOnlySpan<byte>`, no need to pass in a full `byte[]`.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -138,6 +139,94 @@ private List<IPAddress> GetIpv4Addresses(IPHostEntry hostEntry)
...
+
+            if (BitConverter.IsLittleEndian)
+            {
+                magicSpan.Reverse();
+                flagsSpan.Reverse();
+                familySpan.Reverse();
+                socketTypeSpan.Reverse();
+                protocolSpan.Reverse();
+                addrLenSpan.Reverse();
+            }
```
jduncanator, line 160: Once you make the argument a ROS, you can use the `BinaryPrimitives` class here to simplify all this logic. For example:

```
int flags = BinaryPrimitives.ReadUInt32BigEndian(addrInfo); // Where addrInfo is a ReadOnlySpan passed into the function
```

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+    //     int              ai_flags;
+    //     int              ai_family;
+    //     int              ai_socktype;
+    //     int              ai_protocol;
+    //     socklen_t        ai_addrlen;
+    //     struct sockaddr *ai_addr;
+    //     char            *ai_canonname;
+    //     struct addrinfo *ai_next;
+    // };
+    class AddrInfo
```
jduncanator, line 14: This should be a struct, and `sockaddr` should also be implemented, and then changed to `public SockAddr Addr;`.

With it being a struct, it may also mean it's directly blittable from a Span, rather than having to use the `ParseAddrInfoFromBytes` function.

jduncanator (changes requested):
Left a few comments, didn't do an indepth review, just some code improvements from a brief glance over it :)

Thanks for your first contribution!

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
WilliamWsyHK: Agreed. This would be changed

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
WilliamWsyHK: Oh I didn't notice in C# we already have such function!

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
WilliamWsyHK: I really would like it could be directly blittable, but given that the fact `struct sockaddr *ai_addr` is a pointer to either IPv4 / IPv6, and knowing that NS only has IPv4 currently, should we just pretend it to be always 4 bytes of IPv4, or leave it as dynamic `byte[]` with either IPv4/IPv6?

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -138,6 +140,81 @@ private List<IPAddress> GetIpv4Addresses(IPHostEntry hostEntry)
...
+        private long SerializeAddrInfo(ServiceCtx context, long offset, AddrInfo hints, IPAddress address)
+        {
+            long originalBufferPosition = context.Request.ReceiveBuff[0].Position + offset;
+            long bufferPosition         = originalBufferPosition;
+            long bufferSize             = context.Request.ReceiveBuff[0].Size;
+
+            // Magic
+            unchecked
+            {
+                context.Memory.Write((ulong)bufferPosition, IPAddress.HostToNetworkOrder((int)0xBEEFCAFE));
```
AcK77, line 178: Why don't use `BinaryPrimitives.ReverseEndianness()` instead of `IPAddress.HostToNetworkOrder()` ?
(same for following lines)

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -138,6 +140,81 @@ private List<IPAddress> GetIpv4Addresses(IPHostEntry hostEntry)
...
+                Flags = flags,
+                Family = addressFamily,
+                SocketType = socketType,
+                Protocol = protocol,
+                Addrs = addrs,
+                CanonName = canonName
+            };
+        }
+
+        private long SerializeAddrInfo(ServiceCtx context, long offset, AddrInfo hints, IPAddress address)
```
AcK77, line 169:
```suggestion
        private long SerializeAddrInfo(ServiceCtx context, long offset, AddrInfo addrInfo, IPAddress address)
```

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public int Magic;
+        public int Flags;
+        public System.Net.Sockets.AddressFamily Family;
```
AcK77, line 18: use a `using` for `System.Net.Sockets`

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+        public int SocketType;
+        public int Protocol;
+        // public int AddrLen;
```
AcK77, line 21: Why is it commented ?

AcK77 (commented):
Do this implementation based on reverse engineering?
Thanks for your first contribution.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
jduncanator: You can also just get a `Span` from the `context.Memory` and use `BinaryPrimitives.WriteUInt32BigEndian`.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
WilliamWsyHK: `Addrs` itself already contains enough length while being parsed, and if serializing, the correct length is also written base on `Array.Length`

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
WilliamWsyHK: Hmm must be thinking in C way when writing this line. I will take the `Span` way then.

WilliamWsyHK:
Yes the implementation is base on reverse engineering, and the whole code is working as in the normal C code, except N was doing some wrapping on the BSD socket library.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
WilliamWsyHK: I think I will call it as `hintsAddrInfo` instead, to let the caller knowing the intent that it is acting like the C `getaddrinfo`, which requires `hints` to work.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo.cs
Thog: I don't think Nintendo will bring support for IPv6 later so I think it could be nice to assume ipv4 all the time

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -138,6 +141,90 @@ private List<IPAddress> GetIpv4Addresses(IPHostEntry hostEntry)
...
+
+            // ai_addr
+            // IPAddress.GetAddressBytes is already in network byte order.
+            context.Memory.Write((ulong)bufferPosition, ipBytes);
+            bufferPosition += 4;
+
+            // ai_canonname
+            ReadOnlySpan<byte> canonName = Encoding.ASCII.GetBytes("" + "\0");
+            context.Memory.Write((ulong)bufferPosition, canonName);
+            bufferPosition += canonName.Length;
```
Thog, line 223: This is highly ineffective. I think that you could do a structure for the non variable parts of this and just write the structure (casting the WritableRegion memory span to that structure and just writing it)

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -353,6 +440,82 @@ public ResultCode GetGaiStringError(ServiceCtx context)
...
 
+        [Command(6)]
+        // GetAddrInfo(bool enable_nsd_resolve, u32, u64 pid_placeholder, pid, buffer<i8, 5, 0> host, buffer<i8, 5, 0> service, buffer<packed_addrinfo, 5, 0> hints) -> (i32 ret, u32 bsd_errno, u32 packed_addrinfo_size, buffer<packed_addrinfo, 6, 0> response)
+        public ResultCode GetAddrInfo(ServiceCtx context)
+        {
+            byte[] host = new byte[context.Request.SendBuff[0].Size];
+            byte[] service = new byte[context.Request.SendBuff[1].Size];
+            byte[] hints = new byte[context.Request.SendBuff[2].Size];
+
+            bool enableNsdResolve = context.RequestData.ReadInt32() == 1;
```
Thog, line 451: NSD resolution need to be supported here

Thog (changes requested):
I think some stuffs could be turned into structure (example AddrInfo). Some logics are missing part because of no reverse. Otherwise lgtm

AcK77:
@WilliamWsyHK are you still interested to working on this?

WilliamWsyHK:
Yes, will try to get some code updated these 2 days

EmulationFanatic:
> Yes, will try to get some code updated these 2 days

Hi @WilliamWsyHK we haven't forgotten your PR here :) we hope you haven't forgotten it either!

AcK77:
Superseded by #2130

PR #1910, opened 2021-01-13
Fix shader LOP3 predicate write condition

gdkchan:
The condition was being read from the wrong bits on the LOP3 instruction.
Note: The shader cache version was updated, shaders will be invalidated (again, sorry about that).


riperiperi (approved):
Nice catch for this one.

Thog (approved):
nice catch!

PR #1911, opened 2021-01-13
Fix mipmap base level being ignored for sampled textures and images

gdkchan:
The base level was being ignored, which meant that it would bind level 0 only, ignoring the level that was specified. This change allows the correct mipmap level to be accessed from the shader.

Summary of the changes:
- Use base level and max level values from the texture descriptor on the texture pool. They were previously ignored, which means it would always bind all the levels of the texture regardless of what the game actually wants.
- Use layer size provided by the game for render targets. Before the layer size was calculated from the texture dimensions. However, for textures were not all levels are used, we can't calculate the layer size as we don't know the real amount of levels, and we don't know the current base level. So instead of calculating it, it uses the value provided by the game. Should fix array textures using a sub-range of levels.

It is worth noting that flush can't currently handle flushing multiple layers with a sub-range of levels correctly, because there are gaps between the layers (for the unused levels) that it will blank out, instead of leaving untouched, due to the fact that it allocates a array and writes it back to memory. A solution to this is passing a span to the linear -> block linear conversion method that would just write the new data to guest memory directly, and this is also more efficient than the current method. I plan to do that in a future PR. It is worth noting that this was already wrong before, but in a different way, since it was simply ignoring the base level and max level completely.

I recommend testing this in a number of games to ensure that nothing regressed, in special those know to do texture flushes. 

Together with #1910, this improves shadows on Monster Hunter Rise Demo.
Before:
image
After:
image
image
image
(also includes ssbo-flush and the other texture fix for those tests).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureInfo.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureInfo.cs
@@ -1,4 +1,5 @@
 using Ryujinx.Graphics.GAL;
+using Ryujinx.Graphics.Texture;
 
 namespace Ryujinx.Graphics.Gpu.Image
 {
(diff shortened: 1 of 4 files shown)


riperiperi (approved):
LGTM 👍

Thog (approved):
lgtm 👍

PR #1912, opened 2021-01-14
Fix mipmap base level being ignored for sampled textures and images

MisterPrintf:


MisterPrintf:
Dammit, my bad. Github client is misbehaving.

PR #1927, opened 2021-01-18
Fix Linux Icon

pineappleEA:
The icon on the Linux build has been broken since #1859 since every mention of Icon was removed.

Current master:
image
Generic icon on window and bar

After fix:
image
The icon is back on the window and the bar


Doctorwho1909:
Works Well Causes no issues

Xpl0itR (approved):
lgtm, nice catch. This shows that we need to test on linux more :p

IvarWithoutBones (approved):

jduncanator (approved):

PR #1930, opened 2021-01-19
Enable parallel ASTC decoding by default

gdkchan:
This improves the speed of ASTC decoding greatly by using multiple CPU cores for the task. I believe it was not enabled by default due to the concern that it would take away CPU that would be otherwise used by the game and background JIT. This is not as much of a issue now that we have PPTC (enabled by default) that eliminates the JIT cost after a few runs.


riperiperi (approved):
Most impactful single letter change ever. LGTM.

AcK77 (approved):

jduncanator (approved):

PR #1934, opened 2021-01-19
Match texture if the physical range is the same

gdkchan:
#1905 changed the behavior of exact texture matches a little bit. Instead of comparing the physical address (CPU VA) and assuming it is contiguous, it was changed to compare the GPU VA instead (as physical regions might not be contiguous). This change makes it closer to the previous behavior, it now compares the physical regions if the virtual address is not the same (see the new `CompareRange` method), and allows a match if the virtual addresses are different, but mapped to the same physical region.

This seems to fix (or at least greatly improve) a regression on Rune Factory 4 Special caused by the aforementioned PR.

NOTE: This has the potential to slow things down if the slow path that calls `CompareRange` is hit frequently, which is the reason I was only compared the GPU virtual addresses.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
@@ -689,11 +689,18 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 if (matchQuality != TextureMatchQuality.NoMatch)
                 {
+                    // If the parameters match, we need to make sure the texture is mapped to the same memory regions.
+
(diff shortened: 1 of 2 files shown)


GamerzHell9137:
Kinda 
image

AcK77 (approved):

riperiperi (approved):
LGTM

GamerzHell9137:
There was a regression in NSMBUD too which is being fixed with the pr

Before:
image

After: 
image

Way of triggering: Getting into a stage and choosing another one after it, textures from the older stage are being applied to the new one.

Thog (approved):
lgtm

Issue #1416, opened 2020-07-23
Button & joystick input stops working spontaneously


EmulationFanatic:
There are a couple of games in which all inputs seem to be working fine, and then spontaneously most input ceases to work at all.
Will continue adding to this list as more are found.

Affected Games:

https://github.com/Ryujinx/Ryujinx-Games-List/issues/1212
https://github.com/Ryujinx/Ryujinx-Games-List/issues/999
https://github.com/Ryujinx/Ryujinx-Games-List/issues/1530

EmulationFanatic:
FastFP off seems to resolve the issue for Friday the 13th and Slayaway.

EmulationFanatic:
Closed by https://github.com/Ryujinx/Ryujinx/pull/1943

PR #1943, opened 2021-01-21
Lower precision of estimate instruction results to match Arm behavior

### LINKED ISSUE #1940: Catherine: Full Body has a serious bug
## Bug Report
The graphics problems fix in new version.
but it still has a serious bug，so the game cannot continue.

In stage 1, When you reach the position in the picture and pull out the stone, the character can’t grab the edge of the stone and fall down. on real machine, he can grab the edge between two stones and move to the next place.

emulator

real machine

Just started.

### Log file
[Ryujinx_1.0.6348_2021-01-20_22-28-05.log](https://github.com/Ryujinx/Ryujinx/files/5843601/Ryujinx_1.0.6348_2021-01-20_22-28-05.log)

### Environment?
- Ryujinx version: 1.0.6348
- Game version: 1.0.0

- System Specs:
  - OS: Windows 10
  - CPU: I3-8350k
  - GPU: NVIDIA GTX 1060
  - RAM: 16GB
- Applied Mods : [ No ]

### Additional context?
thank for the hardwork!

gdkchan:
Both estimate instruction (RECPE and RSQRTE) only produces 8 bits of fraction, the remaining 15 bits are filled with zeros. The equivalent x86 instructions also returns an approximation, but they have a bit more precision so the result is different from Arm. This PR round the result to nearest and limits the number of bits of the exponent to 8, to match Arm. The results are still not the same for all inputs, and making them match would require implementing the soft float implementation inline, which would hurt performance considerably, so thats a tradeoff.

This fixes a logic bug on Catherine Full Body where the player was not able to grab the edge of the blocks after pulling a block. Would like to have someone confirm that its indeed fixed.

Closes #1940.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
@@ -1475,9 +1475,11 @@ namespace ARMeilleure.Instructions
 
             int sizeF = op.Size & 1;
 
-            if (Optimizations.FastFP && Optimizations.UseSse && sizeF == 0)
+            if (Optimizations.FastFP && Optimizations.UseSse41 && sizeF == 0)
             {
-                EmitScalarUnaryOpF(context, Intrinsic.X86Rcpss, 0);
+                Operand res = EmitSse41FP32RoundExp8(context, context.AddIntrinsic(Intrinsic.X86Rcpss, GetVec(op.Rn)), scalar: true);
+
+                context.Copy(GetVec(op.Rd), context.VectorZeroUpper96(res));
             }
             else
             {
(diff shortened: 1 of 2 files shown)


EmulationFanatic:
This PR also fixes Slayaway Camp: Butcher's Cut and Friday the 13th Killer Puzzle game logic issues. https://github.com/Ryujinx/Ryujinx/issues/1416

jduncanator (approved):

riperiperi (approved):
Nice solution to the problem. LGTM

MerryMage:
> making them match would require implementing the soft float implementation inline, which would hurt performance considerably, so thats a tradeoff.
If you want complete accuracy, you can do one Newton round before truncation.
I can implement this if you wish.

gdkchan:
Yes, I would appreciate it, thanks. I am not a huge fan of the current approach to be honest, but I wanted something with minimal performance cost. I think its fine to do something a little bit more expensive if we can match Arm for most inputs.

PR #1929, opened 2021-01-18
am: Implement CreateHandleStorage and fixes

AcK77:
This PR implement ILibraryAppletCreator::CreateHandleStorage call (Close #1896), fixes CreateStorage/CreateTransferMemoryStorage results and fixes the am IStorage Write ReadOnly check.

This fix the boot regression of MHGU introduced in #1868.

image
image

PR #1945, opened 2021-01-22
Fix inverted read only flag in transfer memory creation

Caian:
From https://switchbrew.org/wiki/Applet_Manager_services#CreateTransferMemoryStorage :

```
The input bool controls whether writing to the storage is allowed: #Write will throw an error if this flag is not set.
```

So the state of `isReadOnly` must be inverted.


Leischii:
Any info on what impact this change will have?

EmulationFanatic:
This is for the swkbd/inline kb crash in MHGU multiplayer chat @Leischii

Caian:
@Leischii, as @EmulationFanatic said, PR #1929 caused MHGU to crash with a stack trace dump when trying to write to the newly created buffer.

I decided to turn make this into a separate commit because I don't know if it is exclusive to the inline keyboard, but is also necessary to continue the fix of the inline keyboard itself.

AcK77 (approved):
Nice catch!

jduncanator (approved):

Issue #1705, opened 2020-11-13
Display an On-screen compiling shaders hint


Ghostotakoi:
## Feature Request

Display an On-screen compiling shaders hint or indicator, maybe a number of how many shaders are compiling

#### Overview:
Display an On-screen compiling shaders hint or indicator

#### Smaller Details: 
Display an On-screen compiling shaders hint or indicator

#### Nature of Request:  
- Addition

### Why would this feature be useful?
It could be very useful for the end user and very user friendly

GamerzHell9137:
Fixed by #2507

GamerzHell9137:
fix #2507

Issue #1312, opened 2020-06-20
Implement loading screen


Ghostotakoi:
## Feature Request

Implement a loading screen

#### Overview:
Would be nice instead just a black screen sort of a loading screen, or loading text, or bar, or whatever that indicate that is loading. Also now that FINALLY a shader cache has been implemented, could be awesome to also add a loading shaders on boot bar, and how many has been loaded.

#### Smaller Details: 
Could be awesome to also add a loading shaders on boot bar, and how many shaders has been loaded.

#### Nature of Request:  
- Addition

### Why would this feature be useful?
It could be awesome for the experienced users, and also for the regular user to know how long it will take to boot your game due shaders or just loading the game itself.

Thog:
This could be a possibility if we implement the necessary AM calls probably. However, this need more research to be fully understood + that would require multi layer support to be done correctly.

Ghostotakoi:
@Thog I have update the request with the new template and also added a little request ;)

Thog:
This is implemented as part of the Avalonia new UI, see https://github.com/Ryujinx/Ryujinx/pull/2507

IsaacMarovitz:
Avalonia UI has this feature, and is now the default UI.

PR #1946, opened 2021-01-22
Show progress of PPTC and Shader Cache Loading in Status Bar

Joshi234:
This shows a bar which shows the progress of pptc and of the shader cache loading and shows when shaders are compiling. Unfortunatly there are 2 bugs right now. The first one is that the progress bar starts at about 15% and the other one is that it often crashes while updating the text.bI hope you can help me fix it.
image
image


Dime333:
This is a very Nice feature. I hope something similar can be added that shows the progress when starting in full screen.

Joshi234:
LeverBars needed to get changed to ProgressBars because LeverBars are broken, they can't go lower then 10% and fixed the crashing

Ghostotakoi:
I have mentioned something similar here https://github.com/Ryujinx/Ryujinx/issues/1312 and here https://github.com/Ryujinx/Ryujinx/issues/1705

jduncanator:
Just took a brief look over the code. These changes should not live in the `Logger` class.

There is already a global `UiHandler` on `Device` for displaying the software keyboard user interface, so the status update related code should live there. This can then also be re-used by other components. The shader cache itself should also have an event added to it called `Status` that emits the event whenever the rebuild status changes. 

In a central position, where both the `UiHandler` instance and the shader cache are both accessible, you'd then add a handler to the shader cache's status update event, and call the `UiHandler` instance's `UpdateStatus` method, with the new status.

This way, the Shader Cache and the status message are completely decoupled, and other code in the future can listen for shader cache status changes in a similar way.

Joshi234:
> Just took a brief look over the code. These changes should not live in the `Logger` class.
> 
> There is already a global `UiHandler` on the `Switch` instance for displaying the software keyboard user interface, so the status bar related code should live there. This can then also be re-used by other components. The shader cache itself should also have an event added to it called `Status` that emits the event whenever the rebuild status changes.
> 
> In a central position, where both the `UiHandler` instance and the shader cache are both accessible, you'd then add a handler to the shader cache's status update event, and call the `UiHandler` instance's `UpdateStatus` method, with the new status. The shader cache shouldn't have any status bar specific code in it.
> 
> This way, the shader cache and the status bar are completely decoupled, and other code in the future can listen for shader cache status changes in a similar way.

Alright changed it

saldabain:
It only appears to work in increments of 50 for me, is that intended?

Joshi234:
@saldabain it's intended because updating it for every would be to much, but I gonna decrease it to 10

Ryujinx.Common/Status/StatusChangedEventArgs.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Ryujinx.Common.Status
+{
+    public class StatusChangedEventArgs : EventArgs
+    {
+        public readonly int Current;
+        public readonly int Total;
+        public readonly string ClassName;
```
riperiperi, line 13: Rather than having this be a string, consider using an `enum` of different status types.

```csharp
namespace Ryujinx.Common.Status
{
    public enum StatusType {
        Ptc,
        ShaderCache
    }
}
```

Ryujinx.Common/Status/StatusChanged.cs
```diff
@@ -0,0 +1,23 @@
+﻿
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Ryujinx.Common.Status
+{
+    public static class StatusChanged
```
riperiperi, line 10: Weird name for a static class (sounds more like an event field), though I'm not sure what's better. Maybe `StatusHandler`?

Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -311,6 +315,7 @@ internal void Initialize()
                     _cacheManager.Synchronize();
                 }
 
+                Common.Status.StatusChanged.DisableStatus();
```
riperiperi, line 318: Import `Ryujinx.Common.Status` to avoid qualifying this.

Ryujinx.HLE/IHostUiHandler.cs
```diff
@@ -36,5 +36,13 @@ public interface IHostUiHandler
         /// </summary>
         /// <returns>False when OK is pressed, True when another button (Details) is pressed.</returns>
         bool DisplayErrorAppletDialog(string title, string message, string[] buttonsText);
+
+        /// Shows and changes the status
+        /// </summary>
```
riperiperi, line 41: Closing tag without an opening tag. I guess the above example is also broken, but you can add the `<summary>` for that one too.

Ryujinx.HLE/IHostUiHandler.cs
```diff
@@ -36,5 +36,13 @@ public interface IHostUiHandler
...
+        void ShowStatus(object o, Common.Status.StatusChangedEventArgs args);
```
riperiperi, line 42: Import `Ryujinx.Common.Status` to avoid qualifying this.

Ryujinx.HLE/IHostUiHandler.cs
```diff
@@ -36,5 +36,13 @@ public interface IHostUiHandler
...
+
+        /// Disable the Status
```
riperiperi, line 44: Add `<summary>` above this to fix the xml doc.

Ryujinx/Ui/Applet/GtkHostUiHandler.cs
```diff
@@ -186,5 +189,27 @@ public bool DisplayErrorAppletDialog(string title, string message, string[] butt
...
+
+        public void ShowStatus(object o, StatusChangedEventArgs args)
+        {
+            MainWindow window = ((MainWindow)_parent);
+            Application.Invoke(delegate
+            {
+                window._statusProgressBar.Visible = true;
+                window._statusProgressBar.Fraction = ((double)((double)args.Current / (double)args.Total));
+                window._status.Visible = true;
+                window._status.Text =  args.ClassName+": "+ args.Current.ToString()+ " / "+args.Total.ToString();
```
riperiperi, line 201: Whitespace is really inconsistent here. VS will do it for you if you cut and paste the line into the same location.

```csharp
window._status.Text = args.ClassName + ": " + args.Current.ToString() + " / " + args.Total.ToString();
```

Ryujinx/Ui/Applet/GtkHostUiHandler.cs
```diff
@@ -186,5 +189,27 @@ public bool DisplayErrorAppletDialog(string title, string message, string[] butt
...
+            });
+        }
+
+        public void DisableStatus(object o,EventArgs args)
```
riperiperi, line 205: Space after comma.

Ryujinx/Ui/Applet/GtkHostUiHandler.cs
```diff
@@ -186,5 +189,27 @@ public bool DisplayErrorAppletDialog(string title, string message, string[] butt
 
             return showDetails;
         }
+
+        public void ShowStatus(object o, StatusChangedEventArgs args)
+        {
+            MainWindow window = ((MainWindow)_parent);
```
riperiperi, line 195: Outer brackets not needed.

Ryujinx/Ui/Applet/GtkHostUiHandler.cs
```diff
@@ -186,5 +189,27 @@ public bool DisplayErrorAppletDialog(string title, string message, string[] butt
...
+                window._statusProgressBar.Visible = true;
+                window._statusProgressBar.Fraction = ((double)((double)args.Current / (double)args.Total));
+                window._status.Visible = true;
+                window._status.Text =  args.ClassName+": "+ args.Current.ToString()+ " / "+args.Total.ToString();
+            });
+        }
+
+        public void DisableStatus(object o,EventArgs args)
+        {
+            MainWindow window = ((MainWindow)_parent);
```
riperiperi, line 207: Outer brackets not needed.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -137,7 +139,7 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
             _fullScreen.Activated         += FullScreen_Toggled;
 
             GlRenderer.StatusUpdatedEvent += Update_StatusBar;
-
+
```
riperiperi, line 142: Remove these spaces.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -86,9 +86,11 @@ public class MainWindow : Window
         [GUI] TreeSelection   _gameTableSelection;
         [GUI] ScrolledWindow  _gameTableWindow;
         [GUI] Label           _gpuName;
+        [GUI] public readonly Label _status;
         [GUI] Label           _progressLabel;
         [GUI] Label           _firmwareVersionLabel;
-        [GUI] LevelBar        _progressBar;
+        [GUI] ProgressBar      _progressBar;
+        [GUI] public readonly ProgressBar _statusProgressBar;
```
riperiperi, line 93: This breaks the alignment of all of these variables. Since they have a different visibility, maybe you should put them before or after in their own group, with their own alignment.

Ryujinx.Common/Status/StatusChangedEventArgs.cs
riperiperi: A benefit of using enum values like this is that you leave it up to the UI to convert that to readable text or influence the display (maybe treat these status types differently) without doing an arbitrary string comparison. Makes things a lot easier to translate, as it's not printing a string directly.

riperiperi (changes requested):
This is a really useful feature, and I like the look of the new bar, but I have quite a few comments.

Ryujinx.Common/Status/StatusChanged.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public static event EventHandler<StatusChangedEventArgs> StatusChangedEvent;
+        public static event EventHandler StatusDisableEvent;
+        public static void ChangeStatus(int current, int total, string message)
+        {
+            StatusChangedEvent.Invoke(null, new StatusChangedEventArgs(current, total, message));
+        }
+        public static void DisableStatus()
+        {
+            StatusDisableEvent.Invoke(null,EventArgs.Empty);
```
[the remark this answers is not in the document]
jduncanator: This should also be:
```suggestion
            StatusDisableEvent?.Invoke(null,EventArgs.Empty);
```
to avoid invoking an empty (null) event handler.

Ryujinx/Ui/Applet/GtkHostUiHandler.cs
jduncanator:
```suggestion
                window._status.Text =  $"{args.ClassName}: {args.Current}/{args.Total}";
```

Ryujinx/Ui/MainWindow.cs
jduncanator: This shouldn't be a public readonly field anyway. Expose it via a getter-only property on MainWindow.

Niwu34:
Joshi pptc
Joshi Shader cache bar
works awesome

MelonSpeedruns:
Does this mean the CMD console cna be disabled by default? :O

Joshi234:
> Does this mean the CMD console cna be disabled by default? :O

Theoretically yes, since there is no real need for normal users anymore since everything is in the logs

Shahil-Ayato:
Shouldn’t this get closed if it’s superseded?

PR #1948, opened 2021-01-22
Add support for shader atomic min/max (S32)

gdkchan:
This adds support for atomic minimum and maximum with a signed integer type. The problem of supporting the signed operation is that the storage (both buffer and shared memory) is uint, and the overload of the atomic operation used is decided from the type, so we can only use the uint `atomicMin` and `atomicMax` functions. This adds new functions that can do signed atomic min/max using a CAS loop.

Other changes: Changed some of the helper functions to have `{` on a new line for conssitency with the other ones.

I believe theres no need for shader cache invalidation here as the affected shaders would fail to compile before this change.

Fixes missing graphics on Disgaea 6 Demo.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -157,6 +157,16 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
                 }
             }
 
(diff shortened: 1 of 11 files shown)


Thog (approved):

Cronos-0:
Confirm is fixed. OvO

ฺBefore
bandicam 2021-01-22 18-19-20-928
After
bandicam 2021-01-24 00-38-42-702

jduncanator (approved):

PR #1950, opened 2021-01-23
Fix TZName parsing in TZIF footer

### LINKED ISSUE #1949: POSIX timezone string isn't properly parsed

~The current timezone calculation code (TZIF parser and its dependents) do not account for the footer described here-
https://tools.ietf.org/id/draft-murchison-tzdist-tzif-15.html#rfc.section.3.3~ 
TZIF Parser does read the footer but it's not being used when calculating the final timezone. (apologies to @Thog. I failed to read the code)

In newer firmwares, some TZIF data blocks no longer provide transition times beyond a certain timestamp (much older than `now`), in which case, a POSIX-style timezone string in the footer is (supposed to be) used to deduce current timezone info-
>  Local time for timestamps on or after the last transition is specified by the TZ string in the footer if present and nonempty, and is unspecified otherwise.

~Here is GNU's manual about parsing POSIX-style timeszone strings- 
https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html~

~If anyone is interested: glibc, BSD libc and IANA's Olson code have implementations for it. Note however that glibc is GPL and maybe incompatible with Ryujinx's license.~

The latest reproduction was from user Niwu in discord- 
https://discord.com/channels/410208534861447168/410432399500115968/802226039433461791

To summarize, as of today 'Europe/Berlin' is in `CET`. With firmware 11.0.1, the parser deduces `CEST`, which is wrong.
The reason- In firmware 11.0.1 the corresponding TZIF does not provide transition timestamps beyond year 1996 (hence the need to use posix timezone info), whereas in 10.2.0 it does until 2037.

This bug usually manifests as a +/- DST hours for certain timezones at certain times depending on which transition time the TZIF ended with.

### Current Workarounds

Select any other timezone which has the desired UTC offset or use a lower firmware (10.2.0 seems to work).

mageven:
An unfortunate optimization resulted in `ParsePosixTime` returning false early, making the TZIF footer info go unused for all timezones.
Closes #1949
tzfix

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZone.cs
+++ b/Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZone.cs
@@ -183,11 +183,10 @@ namespace Ryujinx.HLE.HOS.Services.Time.TimeZone
         {
             int i = namePosition;
 
-            char c = name[i];
+            char c;
 
-            while (c != '\0' && !char.IsDigit(c) && c != ',' && c != '-' && c != '+')
+            while ((c = name[i]) != '\0' && !char.IsDigit(c) && c != ',' && c != '-' && c != '+')
             {
-                c = name[i];
                 i++;
             }
 


Thog (approved):
Good catch!

PR #1953, opened 2021-01-24
Enable Docked Mode by Default

ShahilSharma:
Enables Docked Mode by default, There is one game that doesn't render but for most people I feel having it enabled by default isn't an issue :p


gdkchan:
Theres at least one game that doesn't render with docked mode according to reports, fwiw.

ShahilSharma:
> Theres at least one game that doesn't render with docked mode according to reports, fwiw.

Updated the comment since I didn't realize a game has issues with it enabled >.<

jduncanator (approved):
I always wondered why this wasn't the default. I feel if a game doesn't work in Docked mode, then that's a bug, and we shouldn't be trying to work around it with a default that makes 99% of games look worse out of the box.

AcK77 (approved):

gdkchan (approved):

PR #1954, opened 2021-01-24
Reduce JIT memory usage and GC pressure

gdkchan:
WIP, I will add a description when its done.


FICTURE7:
Another approach I tried for the storing and keeping track of `Uses` and `Assignments` was to create a new structure, `UseMap`. But I can't find the branch on which I did that 😩. But it was something like this:

```csharp
class UseMap {
  void Build(ControlFlowGraph cfg);

  bool HasSingleAssignment(Operand op);
  Node GetSingleAssignment(Operand op);

  bool HasUse(Operand op);
  Span<Node> GetUses(Operand op);

  void AddUse(Node node, Operand operand);
  void RemoveUse(Node node, Operand operand);

  void AddAssignment(Node node, Operand operand);
  void RemoveAssignment(Node node, Operand operand);
}

class Node {
  void SetSource(int index, Operand operand, UseMap map = null);
  void SetDestination(int index, Operand operand, UseMap map = null);
}
```

The idea is to move all the use and assignment bookkeeping to the `UseMap` class and the information is updated whenever a `UseMap` instance is passed to the `SetSource` and `SetDestination` methods. It kept the changes to the `Optimizer` minimal if I recall correctly.

gdkchan:
When would it start building the user information though? From what I recall, with those changes, it goes through all blocks to build the definition and use information for all the operations. I think the main advantage of this approach is not needing to build it for LCQ (where optimizations are not performed), which can save some time and memory.

FICTURE7:
It would start building just before the optimizations passes.

>  I think the main advantage of this approach is not needing to build it for LCQ (where optimizations are not performed), which can save some time and memory.

Oof I see. I guess the `UseMap` could have a mode for LCQ where it does not maintain everything. Not maintaining these lists after the optimization passes will also help the pre-allocator and register allocators because changing the local operand on a node does not a cause a linear search to happen.

gdkchan:
Superseded by #2515.

PR #1961, opened 2021-01-25
Implement geometry shader passthrough

gdkchan:
This implements the geometry shader passthrough functionality, which allows passing the geometry inputs directly to the output without any modification. This feature is exclusive to NVIDIA, but should be possible to support on other vendors eventually by generating the code to copy all inputs to the output on the geometry shader.

This improves rendering on Marvel Ultimate Alliance 3 a little bit.
Before (screenshot not mine):
image
After:
image


Thog (changes requested):
lgtm but need shader cache version incrementation

Thog (approved):

jduncanator (approved):

PR #1964, opened 2021-01-26
Support multiple destination operands on shader IR and shuffle predicates

gdkchan:
This support multiple destination operands on the IR rather than just one.
This is required for supporting the predicate writes for the shuffle instruction. Without this change it would either need to calculate the value inline, or add a separate function just to return the bool result. With the change, now the same function can return both values.
As a follow-up I plan to change the call instruciton implementation and 64-bit multiply to also use it, which will simplify some code.


Thog (changes requested):
lgtm, will need shader cache version increment

Thog (approved):

jduncanator (approved):

PR #1968, opened 2021-01-27
PPTC & Pool Enhancements.

LDj3SNuD:
- Fixed memory instability / excessive memory spikes in some "heavy" games after loading/saving .cache files _(fixes issue reported by users)_;
- Fixed limitation of loading/saving large .cache files _(fixes issue reported by users)_;
- Reduced memory usage when loading/saving .cache files;
- Ptc.Load & Ptc.Save now use XXHash128, which is 10 times faster than MD5;
- Fixed redundant saving / log spamming of .info files _(fixes issue reported by users)_;
- Added a simple `PtcFormatter` library for deserialization/serialization, which does not require reflection, in use at PtcJumpTable and PtcProfiler; improves maintainability and simplicity / readability of affected code.
* Improved handling of Pools (Slim) for PPTC, allowing to halve the number of pools in use and thus reducing memory usage and slightly increasing the translation speed (you save ~1 second every 1k translations (for a CPU with 4+4 cores); so for 30k translations you save ~30 seconds (for the same CPU));
* `BitMap` pools are now limited during use and disposed after use;
* Pools Limiter is now configurable.


FICTURE7:
I think you can reduce the pressure on the operand pool further by interning register operands. For example the translation pass and register allocation pass will fetch a new operand from the pool every time they want to refer to a register.

AFAIK they could instead just refer to the same register operand instances, and be shared across threads because register operands are not mutated anywhere. Something similar could be done for some popular constants such as `0` for `BranchIf true/false` and for the page table constants.

EmulationFanatic:
This fixes the audio blips in Tokyo Mirage Sessions #FE Encore.

ARMeilleure/CodeGen/X86/CodeGenerator.cs
```diff
@@ -190,6 +190,11 @@ public static CompiledFunction Generate(CompilerContext cctx, PtcInfo ptcInfo =
 
                 byte[] code = context.GetCode();
 
+                if (ptcInfo != null)
+                {
+                    ptcInfo.Code = code;
+                }
```
Thog, line 196:
```suggestion
                }ptcInfo?.Code = code;
```

Thog (approved):
lgtm

ARMeilleure/CodeGen/X86/CodeGenerator.cs
LDj3SNuD: That causes a `CS0131`.

Thog: oh oki

Ryujinx.Memory/MemoryBlock.cs
```diff
@@ -40,8 +40,6 @@ public MemoryBlock(ulong size, MemoryAllocationFlags flags = MemoryAllocationFla
             }
 
             Size = size;
-
-            GC.AddMemoryPressure((long)Size);
```
riperiperi, line 44: Was there some kind of impact to doing this?

riperiperi (approved):
Looks good to me, nice work.

Ryujinx.Memory/MemoryBlock.cs
LDj3SNuD: I added those add/remove memory pressure methods to make the GC aware of allocated unmenaged memory (to improve garbage collection response); except that not all allocated umenaged memory is freed; the documentation advises against this, so I removed them. I will re-evaluate their reintroduction after #2002 is merged.

PR #1977, opened 2021-01-28
Use a descriptor cache for faster pool invalidation.

riperiperi:
Rather than evaluating info for cached textures and checking for a match to the existing pool texture, saves the descriptor used to create that texture and only invalidates textures if it no longer equals the guest memory descriptor. Also does this comparison for samplers, to avoid recreating them (before, all samplers on a page would be invalidated).

This may improve performance in games that have stutters caused by texture pool invalidation. Not sure what games that will be, but at least this is a lot faster now.

This PR also fixes an issue with Mario Kart 8 Deluxe introduced by #1905 where the course's cubemap array and shadow map could be evicted from the pool while playing, which would cause the course to become rainbow soup... and generally really hard to look at:

image

A way to reproduce this issue on master is to play Toad Harbor and then Wild Woods back to back. The screenshot above is from going to Yoshi Circuit afterwards.

The problem happened because the GpuAddress in the new descriptor would be compared with the GpuAddress in the texture. But... when the texture was placed in the pool, it was matched by _cpu region_... so any pool invalidation afterwards would erroneously evict the texture. This fixes the issue by comparing against the gpu address in the old descriptor instead.

Any issues that existed before #1905 probably have not been fixed, unless they were performance related.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Pool.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Pool.cs
@@ -8,14 +8,16 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// <summary>
     /// Represents a pool of GPU resources, such as samplers or textures.
     /// </summary>
-    /// <typeparam name="T">Type of the GPU resource</typeparam>
-    abstract class Pool<T> : IDisposable
+    /// <typeparam name="T1">Type of the GPU resource</typeparam>
+    /// <typeparam name="T2">Type of the descriptor</typeparam>
+    abstract class Pool<T1, T2> : IDisposable where T2 : unmanaged
     {
         protected const int DescriptorSize = 0x20;
 
         protected GpuContext Context;
 
-        protected T[] Items;
+        protected T1[] Items;
+        protected T2[] DescriptorCache;
 
         /// <summary>
         /// The maximum ID value of resources on the pool (inclusive).
@@ -47,7 +49,8 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             ulong size = (ulong)(uint)count * DescriptorSize;
 
-            Items = new T[count];
+            Items = new T1[count];
+            DescriptorCache = new T2[count];
 
             Address = address;
             Size    = size;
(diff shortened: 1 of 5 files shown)


gdkchan (approved):
Looks good, thanks for fixing this regression.
Do you know what is the default `Equals` equality comparer for structs? I'm just worried that it might be slow since those descriptor structs are not overriding `Equals` (which is usually recommended), and they also do not implement `IEquatable`.

riperiperi:
It’s ValueType.Equals(), it seems to override object.Equals. Not sure if there’s any hidden cost to unboxing to ValueType.

jduncanator:
In general, for `struct`’s containing no reference type fields, it will perform a byte-wise comparison of the entire struct, which should be pretty quick. If the `struct` contains any reference type fields, it will fall back to using reflection to do a field by field comparison of the entire struct, which can be really slow.

The general rule of thumb I tend to follow, is if the struct is used in any code that could potentially be performance sensitive, then I override the default Equals implementation, and implement `IEquatable`. For the couple dozen lines of code it takes to implement, it's usually a safe trade-off to make. All it takes is someone unknowingly adding a reference type field, and without even realising it, performance can jump off a cliff.

In this instance however, it looks like neither of the descriptors contain reference type fields, and the generic type definition for the `Pool` only accepts unmanaged types, so any attempt to add a reference type field to either of the descriptors will result in a compilation error. It's probably safe to not override Equals or implement `IEquatable` in this instance.

Writing this has made me think though, what the performance difference is between the default behaviour of a byte-wise comparison of the struct, and an override of Equals doing a field-wise comparison. I'm curious to see if you could recreate the default byte-wise comparison behaviour whilst implementing `IEquatable` by doing some `Span<T>` magic in the Equal's override.

riperiperi:
I did some benchmarking and it turned out that it was way slower for some reason. I opted instead for a cast to Vector256 and it looks to be around 500x faster in a tight loop of 100,000,000 comparisons. Of course, we aren't in a tight loop, but it's nice to know that it doesn't cost much anymore.
```
Default equality 2089ms
Custom equality 49ms
```

jduncanator:
Might be handy to have a unit test, or some other form of validation, to ensure that the Descriptor struct is always exactly 32 bytes in size, otherwise strange bugs might occur if the struct is larger (or smaller), and cast to a Vector256 for comparison.

This is where I wish C# had the concept of a `static_assert` 😬

jduncanator (approved):
I'm happy with this, if you wanted to humour my pedanticism, you could probably do:

```
Debug.Assert(Unsafe.SizeOf<TextureDescriptor>() == Unsafe.SizeOf<Vector256>());
```

But not a massive drama. Nice work :)

PR #1987, opened 2021-02-01
Validate CPU virtual addresses on access

gdkchan:
The two main changes here are:
- When the PTE is loaded from the page table, the address is now properly validated, by checking if the address is 0. When the address is 0, it calls the slow read/write functions. With this change, for invalid CPU memory accesses, it prints the guest stack trace and throws `InvalidMemoryRegionException` rather than just showing a `AccessViolationException` on the console. Note that this was already working before #1272.
- Move address validation (checking if it is inside the address space and aligned) inside the `EmitPtPointerLoad` function, rather than doing the check before calling it. This allows some code to be simplified. Additionally, the check is now branchless, more on that later.

The guest pointer validation is done by masking the host pointer value based on the guest pointer value. If the guest pointer is invalid, it masks the host pointer with a 0 mask (forcing the value to be 0), otherwise the value is unchanged. The host pointer being 0 forces it to call the slow path (see host pointer validation below).

The host pointer validation is done with a single check (<= 0), since the pointer tag is inserted in the most significant bits, and so the address will be always negative if the page is being tracked. For writes, it needs to left shift the address by 1 before doing the check, since the bit used to indicate if the page is write tracked is bit 62. This eliminates the need for masking the address later on, it can instead just right shift the address by 1 since the most significant bit was already discarded, and bit 62 is known to be 0 if we are doing the fast path.

Below you can see the equivalent C code of what is currently done for a 32-bit load for example.
```c
int read(uint64_t address) {
    const uint64_t* pt = (uint64_t*)0xCAFEull; // Pretend this is the base page table pointer.
    const int64_t ASE = 1ull << 39;
    uint64_t addrRotated = (address >> 2) | (address << 62);
    uint64_t addrShifted = addrRotated >> 10;
    uint64_t base = pt[addrShifted & ((ASE >> 12) - 1)];
    base &= (int64_t)(addrShifted - (ASE >> 12)) >> 63;
    if ((int64_t)base <= 0) {
        return readSlow(address);
    }
    return *(int*)(base + (address & 0xfff));
}
```
The guest address is first rotated, moving the least significant bits to the most significant bits, then the remaining page bits are shifted out. Since the least significant bits were moved, the value will now be very high if the address is not aligned to the access size. We can check if the address is outside of the address space or not aligned with a single check, `addrShifted >= (AddressSpaceSize >> 12)`. We know that x - y will always be negative if x is less than y, and positive otherwise (as long both x and y are positive). So we know that `addrShifted - (AddressSpaceSize >> 12)` will be negative if the address is in range and aligned, and positive otherwise. We can take advantage of that fact and build a mask with the result, by doing a arithmetic right shift and filling the value with the sign bit. So we have `mask = (long)(addrShifted - (AddressSpaceSize >> 12)) >> 63`. With a bitwise and operation, we apply the mask to the host pointer, and this way it will be zero if the guest pointer is invalid, and will not change otherwise.


Thog (commented):
lgtm

Thog (approved):

LDj3SNuD (approved):

PR #1988, opened 2021-02-02
Fix inter-process data copy on non-contiguous physical regions

gdkchan:
This fixes a issue that started happening since #1458, where the inter-process IPC copies would assume that the entire virtual region was contiguous, when it might not be. This causes memory corruption if the region is not contiguous, as the data will be copied to the wrong physical region.

Fixes crash on boot on Bravely Default II Demo, Balan Wonderworld Demo, and possibly other games. Note: It will not fix crashes that already existed before IPC part 2.
image

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Memory/KMemoryManager.cs
@@ -1506,15 +1506,33 @@ namespace Ryujinx.HLE.HOS.Kernel.Memory
                 {
                     KProcess currentProcess = KernelStatic.GetCurrentProcess();
 
-                    serverAddress = currentProcess.MemoryManager.GetDramAddressFromVa(serverAddress);
+                    while (size > 0)
+                    {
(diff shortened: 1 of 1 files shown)


Thog (approved):
nice catch 👍

jduncanator (approved):

jduncanator:
It might be nice to expose a "SafeCopy" function on the VMM, that performs a block copy in the best case, and handles non-continuous memory in the worst case scenario. I have a feeling we'll be needing to handle non-continuous memory more and more, and having the same large loop to handle that everywhere might get a bit tedious and hard to maintain.

gdkchan:
This is copying data from one address space to another. Where would such "SafeCopy" function be located?

jduncanator:
> This is copying data from one address space to another. Where would such "SafeCopy" function be located?

I was thinking something along the lines of:
```
MemoryUtilities.SafeCopy(srcMemoryManager, srcVirtualAddress, dstMemoryManager, dstVirtualAddress, size)
```

That said, @gdkchan and I discussed this on Discord. Given that this is likely the only place to copy between address spaces, we'll leave it as is for now.

Issue #1915, opened 2021-01-14
Excessive VRAM usage in Hyrule Warriors Age of Calamity


skull-01:
### What's the issue you encountered?
While playing Age of Calamity, VRAM usage increases as you continue to play. This happens using native resolution and the resolution scaler. All default settings and using shader cache. Doing nothing special; the issue happens to whatever stage you play.

### How can the issue be reproduced?
Can be reproduced by just playing the game. The longer you play, the more VRAM it uses. The emulator will crash once you’ve out of VRAM.
If you use resolution scaler, VRAM will be used faster.

### Log file
I’m attaching two sets of logs.
1.	Native resolution
2.	Using x3 resolution scale
I am also attaching the VRAM log usage for both cases. 
[Test 1 Native - No Crash Shader Cache on.zip](https://github.com/Ryujinx/Ryujinx/files/5816224/Test.1.Native.-.No.Crash.Shader.Cache.on.zip)
[Test 2 res scale x3 crash.zip](https://github.com/Ryujinx/Ryujinx/files/5816229/Test.2.res.scale.x3.crash.zip)

### Environment
Ryujinx version: 1.0.6318
Age of Calamity 1.01
- System Specs:
  - OS: Windows 10 20H2
  - CPU: i7 9700
  - GPU: NVIDIA RTX 3090
  - RAM: 32 GB
- Applied Mods :  No

gdkchan:
FIxed by #2001.

PR #2001, opened 2021-02-05
Texture Cache: "Texture Groups" and "Texture Dependencies"

riperiperi:
## Intro

[Download](https://w.ryujinx.org/dl/2001)

Our texture cache tries to make sense of a low level collection of textures iteratively as draws and copies come in and attempt to use them. When _part_ of an existing texture is used, or an existing texture is used in another format, a _view_ of the existing texture is created, with the existing texture being the "storage". Existing textures can also _become_ views when a new texture creates them, with that new texture becoming storage. This enables a ton of fun things like writing to mip levels, texture arrays and cubemaps, as drawing to these views will draw to the original texture.

The key fact is, all of these textures _share their data_ through the storage texture. When the data for one view is set, all overlapping views will also see that change - they are linked at their core. Before now, views tracked their memory as if they were completely _separate_:
 - Before range-tracking, they would check for modified flags in a page table shared between all textures. This meant that a dirty flag could be cleared by another overlapping texture, and a different storage texture (or even views on the same page) wouldn't see that modified flag.
 - After range tracking, each view had its _own_ tracked handle, which covered its entire memory range. However, this meant that a dirty flag could be consumed on one view, but still present on another. Because synchronizing texture data for one view affects all overlapping ones, this is wrong too. Due to this aggressively setting texture data to its original state, there are a few workarounds to disable memory tracking for views, which cause problems for games that stream in or replace a lot of overlapping textures.
 
## Texture Groups

The real solution to these problems is to acknowledge the link between views of the same storage, and share all memory tracking between them. The structure used to represent a storage texture and its views is called a **Texture Group**, and acknowledges that a storage texture is made up of a number of sub-images, where each view into the storage will cover one or more of them.

So for each sub-image, we want to know if it has been "dirtied" by a guest memory write, and needs to be reloaded. We also want to know whether a sub-image has been modified, and flush out or preserve that data when possible. We don't want to consume dirty flags that we don't use, and we don't want to miss any from partial texture writes or that were not consumed by a previous view, so each view needs to work on a _range_ of these sub image handles, shared between all views of the storage texture.

However, creating tracking handles for every sub-image of every image would be very wasteful. Most textures will never have views of their mips/layers at all - they just contain static texture data that doesn't overlap with anything. So, textures have three separate levels of subdivision: 
 - One where the whole texture is covered by one handle,
 - One where one subdivision of the texture has a view (handle for each layer/face for 2d, and a handle for each mip level for 3d)
 - One where all layers and mip levels are tracked individually - there is a 1:1 mapping between view offsets and handles.
 
Because the layer/face view ordering changes when a texture is 3d, the presence of mip and layer views are tracked separately, and the difference in calculation happens when calculating tracking handles, and converting back to view indexes for the storage.

When additional views are created, the tracking will be subdivided further if needed, and the existing dirty/modified flags will be migrated onto the new handles.

#### Partial Synchronization
Since a view can cover multiple sub-images within the storage, and we have tracking (or potential modification) for each of these sub-images, it's possible that only a selection of them still have dirty flags. In most cases, we can still synchronize the full view, as it is faster to do than individual sub-images if the data on CPU is indeed the data we want. That is - in most cases.

Say that you have a texture array of 6 2d textures, and the first texture in the array is bound as a render target and rendered to. It consumes a dirty flag and synchronizes that first 2d array item from CPU memory - memory that the game cleared for us before use - then renders to it and flags it as GPU modified. So far so good, we have data in the first and no data in the other 5. But what if we suddenly bind the full texture array? When we check the dirty flags for the 6 overlapping faces of the storage, the first is consumed, but the other 5 are not. Synchronizing the full texture would clear out the data we just rendered into the first 2d item. We know that one of the 6 textures has been modified and not flushed out, so we'll need to _partially sync texture data_ from the CPU to honor the 5 other dirty flags.

To do this, we keep track of which of the overlapping handles had a dirty flag. If _at least one, but not all of them_ had a dirty flag, AND at least one of the non-dirty textures is gpu modified, then we need to selectively sync _only_ the textures with the dirty flags to avoid losing the data for the texture we just wrote.

A few changes were made to achieve this. There is an additional method to set a single 2D slice of a texture on the backend, and the `ConvertToHostCompatibleFormat` function has been updated to take a `level` and optionally convert layout for just a single level/layer, rather than all of them. And finally, this is the reason that `Texture.SynchronizeMemory` doesn't do any synchronizing itself, it calls the `TextureGroup` to decide what to do _first_. This will also help in future for other things, and is made just as fast as before by the per texture `_dirty` flag.

### Additional Changes:
A few changes were made so that the assumptions made by the Texture Group were correctly held:
- RegionHandles now let you register an event that is fired when "Dirty" becomes true. This cannot be unregistered, it's assumed that this callback will be relevant until the handle is disposed.
- RegionHandles can now be reprotected as dirty. This is used for inheriting dirty flags from another handle.
- When a texture with views is converted into a view itself, its views will be migrated to the new storage texture first.
  - This may be responsible for improving gpu memory usage in HW:AOC
  - This may fix some other unusual errors, though the process of "stealing" views is still a little flawed without a copy dependency.

### Fixes:
- Texture swaps in Xenoblade Chronicles: Definitive Edition, and Xenoblade 2
  - May fix texture swaps in other games
- Texture memory leak in Hyrule Warriors: Age Of Calamity (possibly additional change 1)
- Small improvements in A Hat in Time and Mario Kart 8, fixed further by what you see further down in the comments.
- Greatly reduced number of texture tracking handles (and frequency of their creation) when views are created. This will spend less time creating the handles, which locks the tracking structure.

## TODO: Future Texture Changes
None of these issues have anything to do with this PR, I just wanted to mention them to manage expectations (these things are not yet fixed), and to plan out a roadmap for future changes of this kind.

- Texture eviction strategy that keeps textures alive (after falling off the pool) in case they were simply moved to another location in the pool, or temporarily removed.
  - Stutters and performance in Hyrule Warriors: Age Of Calamity
- Texture eviction strategy that evicts textures when an overlapping texture is created that is not compatible in any way (its data will be garbage)
  - VRAM usage in Xenoblade Chronicles: Definitive Edition (it's also a bit slower cos of buffer flush, but that's not related at all)
- Keep textures _in the cache_ when a view is still alive.
 - Textures are still kept _alive_ when any view is still alive, but they are not findable via the texture cache. This can lead to situations where we end up recreating a storage texture that already exists, or creating a recursive view train between formats, and doing a bunch of unnecessary work over time, and adding a bunch of confusion when finding overlaps which could impact changes.

#### Texture Dependencies (actually done, see below)
Texture views only perfectly represent the guest state if a texture can only have one storage. In some cases, a texture may end up having separate potential "storages", due to the following:
 - View texture compatible with two or more disjoint storage layouts (mario kart - offset overlapping texture arrays + cubemaps)
 - View cannot be created due to host incompatibility, such as compressed views of uint textures (mario kart, a hat in time, splatoon 2)
 
This change is made a lot more possible by the changes in this PR. Dependencies would be _between groups_, as they are at their core an attempt to keep two overlapping storage textures in sync.

## Testing requests
You can download a build of this PR here: https://w.ryujinx.org/dl/2001

Please test everything. I wish I was kidding. Particular attention to:

- Resolution Scaling. This will definitely affect texture blacklist rules, so make sure all your fave games still work.
- Games which load and stream a lot of textures. (UE4)

[Download](https://w.ryujinx.org/dl/2001)

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -215,13 +215,13 @@ namespace Ryujinx.Graphics.Gpu.Image
             }
             else
             {
-                // Don't update this texture the next time we synchronize.
-                ConsumeModified();
-
                 _hasData = true;
 
                 if (!isView)
                 {
+                    // Don't update this texture the next time we synchronize.
+                    ConsumeModified();
+
                     if (ScaleMode == TextureScaleMode.Scaled)
                     {
                         // Don't need to start at 1x as there is no data to scale, just go straight to the target scale.
@@ -567,13 +567,10 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// Checks if the memory for this texture was modified, and returns true if it was.
         /// The modified flags are consumed as a result.
         /// </summary>
-        /// <remarks>
-        /// If there is no memory tracking for this texture, it will always report as modified.
-        /// </remarks>
         /// <returns>True if the texture was modified, false otherwise.</returns>
         public bool ConsumeModified()
         {
-            return Group?.ConsumeDirty(this) != false;
+            return Group.ConsumeDirty(this);
         }
 
         /// <summary>
@@ -599,10 +596,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (_hasData)
             {
-                if (Group != null)
-                {
-                    Group.SynchronizeMemory(this);
-                }
+                Group.SynchronizeMemory(this);
             }
             else
             {
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -51,6 +51,11 @@ namespace Ryujinx.Graphics.Gpu.Image
         public TextureScaleMode ScaleMode { get; private set; }
 
         /// <summary>
+        /// Group that this texture belongs to. Manages read/write memory tracking.
+        /// </summary>
+        public TextureGroup Group { get; private set; }
+
+        /// <summary>
         /// Set when a texture has been modified by the Host GPU since it was last flushed.
         /// </summary>
         public bool IsModified { get; internal set; }
@@ -63,10 +68,12 @@ namespace Ryujinx.Graphics.Gpu.Image
 
         private int _depth;
         private int _layers;
-        private int _firstLayer;
-        private int _firstLevel;
+        public int FirstLayer { get; private set; }
+        public int FirstLevel { get; private set; }
 
         private bool _hasData;
+        private bool _trackingEnabled = true;
+        private bool _dirty = true;
         private int _updateCount;
         private byte[] _currentData;
 
@@ -104,8 +111,6 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         public ulong Size => (ulong)_sizeInfo.TotalSize;
 
-        private GpuRegionHandle _memoryTracking;
-
         private int _referenceCount;
 
         /// <summary>
@@ -131,8 +136,8 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             InitializeTexture(context, info, sizeInfo, range);
 
-            _firstLayer = firstLayer;
-            _firstLevel = firstLevel;
+            FirstLayer = firstLayer;
+            FirstLevel = firstLevel;
 
             ScaleFactor = scaleFactor;
             ScaleMode = scaleMode;
@@ -186,8 +191,6 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <param name="withData">True if the texture is to be initialized with data</param>
         public void InitializeData(bool isView, bool withData = false)
         {
-            _memoryTracking = _context.PhysicalMemory.BeginTracking(Range);
-
             if (withData)
             {
                 Debug.Assert(!isView);
@@ -205,6 +208,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             {
                 // Don't update this texture the next time we synchronize.
                 ConsumeModified();
+
                 _hasData = true;
 
                 if (!isView)
@@ -219,6 +223,18 @@ namespace Ryujinx.Graphics.Gpu.Image
                     HostTexture = _context.Renderer.CreateTexture(createInfo, ScaleFactor);
                 }
             }
+        }
+
+        /// <summary>
+        /// Initialize a new texture group with this texture as storage.
+        /// </summary>
+        /// <param name="hasLayerViews">True if the texture will have layer views</param>
+        /// <param name="hasMipViews">True if the texture will have mip views</param>
+        public void InitializeGroup(bool hasLayerViews, bool hasMipViews)
+        {
+            Group = new TextureGroup(_context, this);
+
+            Group.Initialize(ref _sizeInfo, hasLayerViews, hasMipViews);
         }
 
         /// <summary>
@@ -240,8 +256,8 @@ namespace Ryujinx.Graphics.Gpu.Image
                 info,
                 sizeInfo,
                 range,
-                _firstLayer + firstLayer,
-                _firstLevel + firstLevel,
+                FirstLayer + firstLayer,
+                FirstLevel + firstLevel,
                 ScaleFactor,
                 ScaleMode);
 
@@ -259,11 +275,24 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <param name="texture">The child texture</param>
         private void AddView(Texture texture)
         {
-            DisableMemoryTracking();
-
             _views.Add(texture);
 
             texture._viewStorage = this;
+
+            Group.UpdateViews(_views);
+
+            if (texture.Group != null && texture.Group != Group)
+            {
+                Group.Inherit(texture.Group);
+
+                if (texture.Group.Storage == texture)
+                {
+                    // This texture's group is no longer used.
+                    texture.Group.Dispose();
+                }
+            }
+
+            texture.Group = Group;
         }
 
         /// <summary>
@@ -294,12 +323,12 @@ namespace Ryujinx.Graphics.Gpu.Image
             int blockWidth = Info.FormatInfo.BlockWidth;
             int blockHeight = Info.FormatInfo.BlockHeight;
 
-            width  <<= _firstLevel;
-            height <<= _firstLevel;
+            width  <<= FirstLevel;
+            height <<= FirstLevel;
 
             if (Target == Target.Texture3D)
             {
-                depthOrLayers <<= _firstLevel;
+                depthOrLayers <<= FirstLevel;
             }
             else
             {
@@ -310,14 +339,14 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             foreach (Texture view in _viewStorage._views)
             {
-                int viewWidth  = Math.Max(1, width  >> view._firstLevel);
-                int viewHeight = Math.Max(1, height >> view._firstLevel);
+                int viewWidth  = Math.Max(1, width  >> view.FirstLevel);
+                int viewHeight = Math.Max(1, height >> view.FirstLevel);
 
                 int viewDepthOrLayers;
 
                 if (view.Info.Target == Target.Texture3D)
                 {
-                    viewDepthOrLayers = Math.Max(1, depthOrLayers >> view._firstLevel);
+                    viewDepthOrLayers = Math.Max(1, depthOrLayers >> view.FirstLevel);
                 }
                 else
                 {
@@ -334,8 +363,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         public void DisableMemoryTracking()
         {
-            _memoryTracking?.Dispose();
-            _memoryTracking = null;
+            _trackingEnabled = false;
         }
 
         /// <summary>
@@ -393,7 +421,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (_viewStorage != this)
             {
-                ReplaceStorage(_viewStorage.HostTexture.CreateView(createInfo, _firstLayer, _firstLevel));
+                ReplaceStorage(_viewStorage.HostTexture.CreateView(createInfo, FirstLayer, FirstLevel));
             }
             else
             {
@@ -495,7 +523,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     view.ScaleFactor = scale;
 
                     TextureCreateInfo viewCreateInfo = TextureManager.GetCreateInfo(view.Info, _context.Capabilities, scale);
-                    ITexture newView = HostTexture.CreateView(viewCreateInfo, view._firstLayer - _firstLayer, view._firstLevel - _firstLevel);
+                    ITexture newView = HostTexture.CreateView(viewCreateInfo, view.FirstLayer - FirstLayer, view.FirstLevel - FirstLevel);
 
                     view.ReplaceStorage(newView);
                     view.ScaleMode = newScaleMode;
@@ -523,11 +551,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <returns>True if the texture was modified, false otherwise.</returns>
         public bool ConsumeModified()
         {
-            bool wasDirty = _memoryTracking?.Dirty ?? true;
-
-            _memoryTracking?.Reprotect();
-
-            return wasDirty;
+            return Group?.ConsumeDirty(this) != false;
         }
 
         /// <summary>
@@ -544,17 +568,45 @@ namespace Ryujinx.Graphics.Gpu.Image
                 return;
             }
 
-            if (_hasData)
+            if (!_dirty)
             {
-                if (_memoryTracking?.Dirty != true)
-                {
-                    return;
-                }
-
-                BlacklistScale();
+                return;
             }
 
-            _memoryTracking?.Reprotect();
+            _dirty = false;
+
+            if (_hasData)
+            {
+                if (_trackingEnabled && Group != null)
+                {
+                    Group.SynchronizeMemory(this);
+                }
+            }
+            else
+            {
+                Group.ConsumeDirty(this);
+                SynchronizeFull();
+            }
+        }
+
+        /// <summary>
+        /// Signal that this texture is dirty, indicating that the texture group must be checked.
+        /// </summary>
+        public void SignalGroupDirty()
+        {
+            _dirty = true;
+        }
+
+        /// <summary>
+        /// Fully synchronizes guest and host memory. 
+        /// This will replace the entire texture with the data present in guest memory.
+        /// </summary>
+        public void SynchronizeFull()
+        {
+            if (_hasData)
+            {
+                BlacklistScale();
+            }
 
             ReadOnlySpan<byte> data = _context.PhysicalMemory.GetSpan(Range);
 
@@ -596,7 +648,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             BlacklistScale();
 
-            _memoryTracking?.Reprotect();
+            Group.ConsumeDirty(this);
 
             IsModified = false;
 
@@ -606,17 +658,45 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
+        /// Uploads new texture data to the host GPU for a specific layer/level.
+        /// </summary>
+        /// <param name="data">New data</param>
+        /// <param name="layer">Target layer</param>
+        /// <param name="level">Target level</param>
+        public void SetData(ReadOnlySpan<byte> data, int layer, int level)
+        {
+            BlacklistScale();
+
+            HostTexture.SetData(data, layer, level);
+
+            _currentData = null;
+
+            _hasData = true;
+        }
+
+        /// <summary>
         /// Converts texture data to a format and layout that is supported by the host GPU.
         /// </summary>
         /// <param name="data">Data to be converted</param>
         /// <returns>Converted data</returns>
-        private ReadOnlySpan<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data)
+        public ReadOnlySpan<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data, int level = 0, bool single = false)
         {
+            int width = Info.Width;
+            int height = Info.Height;
+
+            int depth = single ? 1 : _depth;
+            int layers = single ? 1 : _layers;
+            int levels = single ? 1 : Info.Levels;
+
+            width = Math.Max(width >> level, 1);
+            height = Math.Max(height >> level, 1);
+            depth = Math.Max(depth >> level, 1);
+
             if (Info.IsLinear)
             {
                 data = LayoutConverter.ConvertLinearStridedToLinear(
-                    Info.Width,
-                    Info.Height,
+                    width,
+                    height,
                     Info.FormatInfo.BlockWidth,
                     Info.FormatInfo.BlockHeight,
                     Info.Stride,
@@ -626,11 +706,11 @@ namespace Ryujinx.Graphics.Gpu.Image
             else
             {
                 data = LayoutConverter.ConvertBlockLinearToLinear(
-                    Info.Width,
-                    Info.Height,
-                    _depth,
-                    Info.Levels,
-                    _layers,
+                    width,
+                    height,
+                    depth,
+                    levels,
+                    layers,
                     Info.FormatInfo.BlockWidth,
                     Info.FormatInfo.BlockHeight,
                     Info.FormatInfo.BytesPerPixel,
@@ -650,11 +730,11 @@ namespace Ryujinx.Graphics.Gpu.Image
                     data.ToArray(),
                     Info.FormatInfo.BlockWidth,
                     Info.FormatInfo.BlockHeight,
-                    Info.Width,
-                    Info.Height,
-                    _depth,
-                    Info.Levels,
-                    _layers,
+                    width,
+                    height,
+                    depth,
+                    levels,
+                    layers,
                     out Span<byte> decoded))
                 {
                     string texInfo = $"{Info.Target} {Info.FormatInfo.Format} {Info.Width}x{Info.Height}x{Info.DepthOrLayers} levels {Info.Levels}";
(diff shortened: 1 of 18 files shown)


Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                bool modified = group.Modified;
+
+                bool handleDirty = false;
+
+                foreach (CpuRegionHandle handle in group.Handles)
+                {
+                    if (handle.Dirty)
+                    {
+                        handle.Reprotect();
+                        dirty |= true;
```
gdkchan, line 121: Same comment applies here too.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                    texture.SynchronizeFull();
+                }
+            }
+        }
+
+        /// <summary>
+        /// Synchronize part of the storage texture, represented by a given range of handles.
+        /// Only handles marked by the _loadNeeded array will be synchronized.
+        /// </summary>
+        /// <param name="baseRegion"></param>
```
gdkchan, line 154: Missing description for parameters.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                    handle.RegisterAction((address, size) => FlushAction(group, address, size));
+                }
+            }
+        }
+
+        /// <summary>
+        /// Evaluate the range of tracking handles which a view texture overlaps with.
+        /// </summary>
+        /// <param name="texture">The texture to get handles for</param>
+        /// <returns></returns>
```
gdkchan, line 214: Missing.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+            {
+                if (_loadNeeded[baseRegion + i])
+                {
+                    var info = GetHandleInformation(baseRegion + i);
+                    int offsetIndex = info.index;
+
+                    // Only one of these will be greater than 1, as partial sync is only called when there are sub-image views.
+                    for (int layer = 0; layer < info.layers; layer++)
+                    {
+                        for (int level = 0; level < info.levels; level++)
```
gdkchan, line 170: Looking at this made me realise that theres a bug on the way `allOffsets` length is calculated right now, it doesn't take into account the fact that the layers count halve for 3D textures at each mip level... So this will probably read the wrong offset (zeros) while iterating it (for 3D textures with levels > 1).

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+        }
+
+        /// <summary>
+        /// Get the range of offsets for a given mip level of a 3D texture.
+        /// </summary>
+        /// <param name="level">The level to return</param>
+        /// <returns>Start index and count of offsets for the given level</returns>
+        private (int index, int count) Get3DLevelRange(int level)
+        {
+            int index = 0;
```
gdkchan, line 269: Why not just:
```cs
int index = _layers * level;
int count = Math.Max(_layers >> level, 1);
```

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                if (_hasLayerViews)
+                {
+                    // NOTE: Will also have mip views, or only one level in storage.
+
+                    index = handleIndex;
+                    baseLevel = 0;
+
+                    int layerLevels = _levels;
+
+                    while (handleIndex >= layerLevels)
```
gdkchan, line 305: You can get rid of this loop by taking the difference, and then using multiplication, shift, etc.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+
+            ReplaceHandles(handles);
+        }
+
+        /// <summary>
+        /// A flush has been requested on a tracked region. Find an appropriate view to flush.
+        /// </summary>
+        /// <param name="handle">The handle this flush action is for</param>
+        /// <param name="address">The address of the flushing memory access</param>
+        /// <param name="address">The size of the flushing memory access</param>
```
gdkchan, line 639:
```suggestion
        /// <param name="size">The size of the flushing memory access</param>
```

Ryujinx.Graphics.Gpu/Image/TextureInfo.cs
```diff
@@ -232,6 +232,27 @@ public static int GetLayers(Target target, int depthOrLayers)
             }
         }
 
+        /// <summary>
+        /// Gets the number of 2d slices of the texture.
```
gdkchan, line 236:
```suggestion
        /// Gets the number of 2D slices of the texture.
```

Ryujinx.Memory/Tracking/RegionHandle.cs
```diff
@@ -120,11 +138,22 @@ public bool OverlapsWith(ulong address, ulong size)
             return Address < address + size && address < EndAddress;
         }
 
+        bool disposed;
```
gdkchan, line 141: I think this should be called `_disposed` and moved next to the other fields.

Ryujinx.Memory/Tracking/RegionHandle.cs
```diff
@@ -120,11 +138,22 @@ public bool OverlapsWith(ulong address, ulong size)
...
+
         /// <summary>
         /// Dispose the handle. Within the tracking lock, this removes references from virtual and physical regions.
         /// </summary>
         public void Dispose()
         {
+            if (disposed)
+            {
+                throw new Exception("Double Dispose!");
```
gdkchan, line 150:
```suggestion
                throw new ObjectDisposedException();
```

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+
+            return (index, count);
+        }
+
+        /// <summary>
+        /// Get view information for a single tracking handle.
+        /// </summary>
+        /// <param name="handleIndex">The index of the handle</param>
+        /// <returns>The layers and levels that the handle covers, and its index in the offsets array</returns>
+        private (int baseLayer, int baseLevel, int levels, int layers, int index) GetHandleInformation(int handleIndex)
```
gdkchan, line 286: `(int BaseLayer, int BaseLevel, int Levels, int Layers, int Index)`

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                return (texture.FirstLevel + (texture.FirstLayer) * levelHandles, targetLevelHandles + (targetLayerHandles - 1) * levelHandles);
+            }
+        }
+
+        /// <summary>
+        /// Get the range of offsets for a given mip level of a 3D texture.
+        /// </summary>
+        /// <param name="level">The level to return</param>
+        /// <returns>Start index and count of offsets for the given level</returns>
+        private (int index, int count) Get3DLevelRange(int level)
```
gdkchan, line 267: `(int Index, int Count)`

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,669 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Evaluate the range of tracking handles which a view texture overlaps with.
+        /// </summary>
+        /// <param name="texture">The texture to get handles for</param>
+        /// <returns></returns>
+        private (int baseHandle, int regionCount) EvaluateRelevantHandles(Texture texture)
```
gdkchan, line 215: `(int BaseHandle, int RegionCount)`

gdkchan (commented):
Great work, Xenoblade has never looked so good. I will test some games later.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
riperiperi: Each level should have less layers than the last. Imagine a 3d texture with a depth of 4, and 4 mip levels:

```
[
1, 2, 3, 4, 5, 6, 7, 8,
9, 10, 11, 12,
13, 14,
15
]
```
Calculating the index like that will result in an incorrect offset index for the second mip. The count needs to be reduced each time so it's like (8 + 4 + 2) etc.

Ryujinx.Memory/Tracking/RegionHandle.cs
riperiperi: This is actually a leftover from testing, where a double dispose was causing issues. I didn't plan to make it into an actual change.

riperiperi: Though, throwing an `ObjectDisposedException` would make it obvious if this error were encountered again.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
gdkchan: Ah ok. Its worth noting it isn't actually stored like that on the `AllOffsets` array right now, as I mentioned before.

Illynir:
No problem on FETH, Buried Stars and MHR Demo.

Xenoblade 1 and 2 are almost perfect now.

Leischii:
Already mentioned on Discord but to make sure it's tested once fixed: AC-NH crashes with this after the daily announcement.

scese250:
No regression found for Girls und Panzer DTM DX. 

No changes for DOA Xtreme 3 Scarlet unfortunately.

ReaperOfSouls1909:
No regressions in ue4 titles and nothing improved

CryptForge:
Age of Calamity appears to gain a small increase in performance and better stability with this PR. Especially when resolution is scaled up.

LukeWarnut:
I found a regression in 20XX with this PR.

Before PR:
image
[20XX.log](https://github.com/Ryujinx/Ryujinx/files/6013459/20XX.log)

After PR:
image
[20XX-2001.log](https://github.com/Ryujinx/Ryujinx/files/6013442/20XX-2001.log)

gdkchan:
Can you include steps to reproduce the issue? And also a save if necessary/possible.

LukeWarnut:
> Can you include steps to reproduce the issue? And also a save if necessary/possible.

Those screenshots were taken in the hub world after beating the tutorial. (Full screenshot below)
Here's my save file that has the tutorial completed:
[0.zip](https://github.com/Ryujinx/Ryujinx/files/6013541/0.zip)
image

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
riperiperi: Will be resolved in a future commit.

riperiperi:
Don't review right now, there are more changes to come.

riperiperi:
# Copy Dependencies

Copy Dependencies are an extension of the "CopyOnly" compatibility that was previously used to copy 3D texture slices to the full 3D texture on creation, and the emulated views (copy driven) we used for compressed<->uncompressed views, but in a more general sense.

Firstly, the CopyOnly compatibility has been extended to include view-parent pairs with mismatching compressed/uncompressed formats. It is also set when a texture is to be taken as view of a new texture, but it is already a view of another, as we cannot have a texture with two storage textures.

Before, CopyOnly compatibility would only be honored at the time of texture creation, and _only_ when a storage texture was created, consuming the copyable views. Updates to these "views" would not cause future copies to the parent. Because of this, there were a number of workarounds that attempted to avoid losing this initial data by synchronizing from memory, as future draws to update it (presumably after a cpu memory clear) would be lost.

To solve this problem, we use the flexibility of Texture Group Handles to implement "copy-dependencies" - a sort of contract that indicates that a write to one sub-image of the storage texture _must_ result in a copy to a sub-image of another texture before the next time it is used. In the context of Texture Groups, the dependency is between two TextureGroupHandles, and copies are deferred to dependent textures when SignalModified is called. This sets the `_dirty` flag on overlapping textures, which will indicate that it should check its dependencies to see if 

This fits nicely in with the partial texture updates added by Texture Groups. Deferred copies are only performed if their source handle has not been dirtied after its last use, and will override dirty flags on the side of the copy target (as due to the copy source dirty flags being from the same memory region, the copy will be newer). It's even possible for multiple copies to be performed _alongside_ synchronizing several sub-images of a texture from memory. The BBC will have to do better than strikingly similar nature documentaries narrated by a guy with a nice voice to top how beautiful this all is.

Note that creating a copy dependency forces the texture group to use the most granular handle subdivision, as the copies happen for a single 2D sub-image.

## Additional Fixes

- Tracking can no longer be disabled on a texture.
- View compatibility now also checks:
  - If the levels/layers of the view are in bounds of the parent. If they are not, the texture is incompatible.
  - If the view has more than 1 slice, its LayerSize must match the parent's (otherwise only the base layer is a view).
- As a result of copy dependencies doing the work instead, the emulated view parent has been removed.
- Fixed many issues with 3D texture views.
  - View offsets are now properly calculated for 3D textures with mipmaps.
  - View offsets are now properly transformed into layer/level for 3D textures. (also, the all offsets binary search now works correctly)
  - Fixed some issues with 3D textures and Texture Groups.
- Render targets now notify the texture when it is _bound_, and _unbound_. Copies performed while a texture is bound are repeated every time, rather than being cleared, as the texture is still being written to.
- Other minor rearrangements and fixes.

# Comparisons

Screenshots were taken with a build before some final changes.

## Mario Kart 8 Deluxe
- Fixed majority of missing faces on ingame cubemaps (4800% more cubemap minimum)
- Fixed issue where characters/karts would become partially or fully black after cycling through character select for a while.
- Fixed colour levels being incorrect for most courses. (rendered 3D texture LUT)
(there is still an issue where certain interior lighting can infrequently cause a rainbow effect that is being looked into, and one face of a cubemap's mip layers may be lost past the first race.)

### Before
Character Select:

Electrodrome:


Royal Raceway:


### After
Character Select (source: just trust me bro):

Electrodrome:


Royal Raceway:



## A Hat in Time
- Majority of texture issues fixed. Some texture related crashes fixed.
- Game still devolves into madness after a while of playing... but it's a lot better.

### Before
Title Screen:

Mafia Town:

...just kidding (3 crashes later):


### After
Title Screen:

Mafia Town:


## Super Mario Odyssey
- Fixed colour levels being incorrect for most kingdoms. (rendered 3D texture LUT)
- Fixed various issues caused by a rendered mipmapped 3d perlin noise texture. (that's a lot of words)
  - Fog in all kingdoms is now wispy, rather than filled with harsh lines and unrealistic changes in intensity.
  - Fixed all effects relating to the Seaside Kingdom water. (beach waves, caustics, surface colour)
  - Fixed "dirty" decals that are applied to mario in certain situations. (smoke, fog)

### Before
Fog (metro):

Fog (wooded):
https://gyazo.com/0f4c0c1e8084e6b491e7a05f11c37bef
Seaside Waves:
https://gyazo.com/f9133048bf9173085ea52ce25850faa1

### After
Fog (metro):

Fog (wooded):
https://gyazo.com/0ef88252830c80f8c94d59051d3a8570
Seaside Waves:
https://gyazo.com/d757c516ba23bc3355e9c4edf33d5639
Seaside Caustics:
https://gyazo.com/6b3b9d1b83dbb68150c21c18eb305f4b

## Splatoon 2
- Fixed an issue where the plaza would be really dark on the first load, or randomly on future loads/levels.
- Fixed issues where text on the large splatter buttons would not appear, or would be incorrect. (this also affects large text in multiplayer)
- Fixed colour levels being incorrect occasionally. (rendered 3D texture LUT)

### Before
Plaza (1st load):

Menu buttons:

### After
Plaza (1st load):

Menu buttons:


## Xenoblade Chronicles: Definitive Edition (maybe 2?):
- Fixed colour levels being subtly incorrect. (rendered 3D texture LUT, every frame)

### Before

### After


There might be a better example somewhere else.

## Animal Crossing: New Horizons
- Fixed DOF only using the first blur level, rather than all of them. (a copy each frame is required to do this)
  - Only really noticeable for very far away, or cutscenes that enforce a heavy DOF. Pay attention to the museum at the back.
### Before
Pay attention to the museum.
https://cdn.discordapp.com/attachments/728011297857339422/811957312377585694/unknown.png
### After
Pay attention to the museum.
https://cdn.discordapp.com/attachments/728011297857339422/811957328995024936/unknown.png

## Luigi's Mansion 3:
- May have fixed some lighting or room transition related issues.

## Pikmin 3 Deluxe:
- Fixed intro video showing a black screen.
- May fix subtle issues with the KopPad.

### After (black screen before)


May fix similar issues with other Nintendo first party games. Changes should be tested with games with texture issues.

# Note

This isn't a perfect solution, but it has the potential to be one. I'll be looking into the weird remaining issues, but what I have right now is very reviewable and should work better in most cases.

legend800:
Fixes missing menu text in Fast RMX.

riperiperi:
> 
> 
> > Can you include steps to reproduce the issue? And also a save if necessary/possible.
> 
> Those screenshots were taken in the hub world after beating the tutorial. (Full screenshot below)
> Here's my save file that has the tutorial completed:
> [0.zip](https://github.com/Ryujinx/Ryujinx/files/6013541/0.zip)
> image

Can you retest this with the latest commit? I'm not too sure what would cause it.

Niwu34:
Master:
20XX - Master

Pr 2001 []Ryujinx 1.0.6668:
20XX - Texture

riperiperi:
Latest commit seems to fix remaining issues with Mario Kart 8: "and one face of a cubemap's mip layers may be lost past the first race."

Shahil-Ayato:
Fixes #1620

Shahil-Ayato:
Could this potentially fix #1915?

riperiperi:
> 
> 
> Could this potentially fix #1915?

Yes.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -382,10 +403,22 @@ public static bool ViewLayoutCompatible(TextureInfo lhs, TextureInfo rhs, int le
         /// </summary>
         /// <param name="lhs">Texture information of the texture view</param>
         /// <param name="rhs">Texture information of the texture view</param>
-        /// <returns>True if the formats are compatible, false otherwise</returns>
-        public static bool ViewFormatCompatible(TextureInfo lhs, TextureInfo rhs)
+        /// <returns>How view compatible the formats are</returns>
```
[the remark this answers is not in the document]
riperiperi: Yes, it just reads rather unusually since it starts with "how". The returned value is a measure of the view compatibility of the formats, so the function returns how "view compatible" the formats are. I can think of another way to word it.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -841,24 +943,27 @@ public Texture FindOrCreateTexture(TextureSearchFlags flags, TextureInfo info, i
...
+
                 // We need to synchronize before copying the old view data to the texture,
                 // otherwise the copied data would be overwritten by a future synchronization.
                 texture.InitializeData(false, setData);
 
                 for (int index = 0; index < viewCompatible; index++)
                 {
                     Texture overlap = _textureOverlaps[index];
+
+                    if (overlap.Info.Target == Target.Cubemap && texture.Info.Target == Target.CubemapArray) { }
```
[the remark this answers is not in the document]
riperiperi: It just keeps happening 🥴

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -544,18 +590,46 @@ public void SynchronizeMemory()
...
+            {
+                return;
+            }
+
+            _dirty = false;
+
             if (_hasData)
             {
-                if (_memoryTracking?.Dirty != true)
+                if (Group != null)
```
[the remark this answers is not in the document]
riperiperi: `Group` actually isn't set null anymore. I'll change this and ConsumeModified to reflect that.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -0,0 +1,985 @@
...
+        }
+
+        /// <summary>
+        /// Calculate a single view's data size. This is used to better affirm the bounds of 2D sub-images,
+        /// and is particularly useful for layer strided Texture2DArrays, where a handle's calculated size
+        /// shouldn't cover anything between layers, such as mip levels.
+        /// </summary>
+        /// <param name="level">The level of the view</param>
+        /// <returns>The view's size in bytes</returns>
+        private int CalculateViewDataSize(int level)
```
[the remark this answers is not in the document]
riperiperi: I've addressed this by making the size calculator output an array of `Slice Sizes` for each mip level, which can be indexed by level rather than using this function. This is given to the texture group as part of initializating the storage.

riperiperi:
Should be all comments addressed. There shouldn't be any regressions, but it's worth testing anyways.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -235,6 +235,27 @@ public static TextureViewCompatibility ViewSizeMatches(TextureInfo lhs, TextureI
...
         }
 
+        /// <summary>
+        /// Checks if the potential child texture fits within the level and layer bounds of the parent.
+        /// </summary>
+        /// <param name="parent">Texture information for the parent</param>
+        /// <param name="child">Texture information for the child</param>
+        /// <param name="layer">Base layer of the child texture</param>
+        /// <param name="level">Base level of the child texture</param>
+        /// <returns>Full compatiblity if the child's layer and level count fit within the parent, incompatibile otherwise</returns>
```
gdkchan, line 245: Typo: `incompatibile`.

Ryujinx.Graphics.Gpu/Image/TextureDependency.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// One side of a two-way dependency between one texture view and another.
+    /// Contains a reference to the handle owning the dependency, and the other dependency.
+    /// </summary>
+    class TextureDependency
+    {
+        /// <summary>
+        /// The handle that owns this dependency.
+        /// </summary>
+        public TextureGroupHandle Handle;
+        /// <summary>
```
gdkchan, line 13: Missing empty line separation before this line.

Shahil-Ayato (approved):

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -0,0 +1,327 @@
...
+            SignalModified();
+
+            // Note: Bind count currently resets to 0 on inherit for safety, as the handle <-> view relationship can change.
+            _bindCount = Math.Max(0, _bindCount + (bound ? 1 : -1));
+        }
+
+        /// <summary>
+        /// Signal that a copy dependent texture has been modified, and must have its data copied to this one.
+        /// </summary>
+        /// <param name="copyFrom">The texture handle that must defer a copy to this one.</param>
```
gdkchan, line 146:
```suggestion
        /// <param name="copyFrom">The texture handle that must defer a copy to this one</param>
```

gdkchan (approved):
lgtm other than the few comments.
Awesome work fixing all those bugs 👍

michaelpegaso:
With the latest updates it has stopped having effect, for example in splatton 2 the problem of shadows in the initial square of the game is once again.

Leischii:
> 
> 
> With the latest updates it has stopped having effect, for example in splatton 2 the problem of shadows in the initial square of the game is once again.
> 
> I have tried the "Output Windows x64 (Release)" version and everything seems to work perfect.

What do you mean by "latest updates"? Did you use the built in updater? If that's the case, then the changes made in this Pr have been overwritten with latest master, which doesn't have these changes.

Thog (approved):
lgtm

PR #2005, opened 2021-02-08
Proper bindless texture support

gdkchan:
This is some WIP changes that I started around one year ago, but never finished due to what seems to be a driver bug. It allows for proper bindless textures support by using thee bindless texture extension on the host.

This is required to make some games render properly, example: Super Mario Party (has memory leak right now with those changes), some Otome games (Enchanted in the Moonlight, Diabolik Lovers, and more).

I will add more information if/when the TODOs are completed.

TODO:
- Figure out why it is leaking VRAM memory on NVIDIA.
- Figure out why it is crashing randomly on NVIDIA (might b e connected to the above).
- Only load from the pool textures with the same target as the shader (might reduce VRAM usage in some cases).
- Better texture and sampler validation (reduce false positives, less memory usage and sync overhead, etc).

Bindless 3D textures that are rendered to will not work properly without copy dependency (example: Luigi's Mansion 3).

[Download](https://w.ryujinx.org/dl/2005)


LukeWarnut:
Improves rendering in the Balan Wonderworld Demo.
image

LukeWarnut:
Improves rendering in Tetris 99
image
image

gdkchan:
Fixes missing clouds on https://github.com/Ryujinx/Ryujinx-Games-List/issues/3599
image

gdkchan:
Superseded by https://github.com/Ryujinx/Ryujinx/pull/3001.

Shadowzyzy:
> 

your links arent working, do you know any way to solve this black screen problem on this specific visual novel called diabolik lovers chaos lineage? the dialogue and the opening appears but no image

PR #2009, opened 2021-02-10
Enable multithreaded VP9 decoding

gdkchan:
This adds some optimisations to the VP9 decoder:
- Support multithreaded tile decoding. Improves speed by decoding multiple tiles in parallel, on separate CPU cores.
- Add `SkipLocalsInit` attribute to several inverse transform functions with stackallocs. Gives a minor performance improvement by avoiding zeroing the stack.

Most of the benefit comes from the multithreaded decoding. I had an speedup of about 2.5x decoding 1080p videos here (measured with standalone decoding, it is likely lower on the emulator). I would expect similar improvements on other CPUs with more than 6 cores.

**Some technical notes:**

Due to the way how the work is distributed, uneven thread counts does not yield much improvement. For example, a thread count of 2 and a thread count of 3 will give pretty much the same performance, for a video with 4 tile columns, because one thread will need to decode 2 tiles, and the other will need to decode 1. Considering that they all take the same time to decode, then theres no advantage is having 3 threads, because it will still need to way for the one decoding 2 tiles to finish. The code tries to compensate this by sorting the tiles, so that the thread that needs to process 2 tiles only processes the smallest ones (which are assumed to be the fastest to decode).

This is pretty much a port of libvpx multithreaded tile column decoding, but I made it more .NET-like, by using `Parallel.For` rather than manually creating threads and using condition variablles to wait for work completion, etc.

**What to test:**

Any game using VP9 videos. Here's some examples:
- Super Smash Bros Ultimate
- Pokémon Let's Go Eevee/Pikachu
- Zelda Link's Awakening
- Fire Emblem Three Houses
- Tokyo Mirage Session #FE Encore
- Animal Crossing New Horizons
- Mr. Driller no Drillland

One should look for performance improvements, and also possible regressions. For example, Zelda Link's Awakening intro video is very choppy on master right now, it should be better with this change.

[Download](https://w.ryujinx.org/dl/2009)

### DIFF
--- a/Ryujinx.Graphics.Nvdec.Vp9/DecodeFrame.cs
+++ b/Ryujinx.Graphics.Nvdec.Vp9/DecodeFrame.cs
@@ -1,13 +1,14 @@
 ﻿using Ryujinx.Common.Memory;
+using Ryujinx.Graphics.Nvdec.Vp9.Common;
+using Ryujinx.Graphics.Nvdec.Vp9.Dsp;
+using Ryujinx.Graphics.Nvdec.Vp9.Types;
+using Ryujinx.Graphics.Video;
 using System;
 using System.Buffers.Binary;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using Ryujinx.Graphics.Nvdec.Vp9.Common;
-using Ryujinx.Graphics.Nvdec.Vp9.Dsp;
-using Ryujinx.Graphics.Nvdec.Vp9.Types;
-using Ryujinx.Graphics.Video;
+using System.Threading.Tasks;
 using Mv = Ryujinx.Graphics.Nvdec.Vp9.Types.Mv;
 
 namespace Ryujinx.Graphics.Nvdec.Vp9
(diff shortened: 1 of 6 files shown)


Thog (approved):
lgtm :+1:

Benedani:
This turns the SSBU intro cutscene from a laggy mess to full speed with a few hiccups on my i7-4770s. The experience should be flawless on a more modern CPU. Nice!

EmulationFanatic:
Biggest change for me was with Link's Awakening, whose opening scene was essentially a slideshow.

Without PR:
https://user-images.githubusercontent.com/62343878/107509658-15b2bd80-6b60-11eb-89ff-de64e0254444.mp4

With PR:

https://i.imgur.com/RhvMtuU.mp4

EmulationFanatic:
This also fixes the slowness/choppy audio in the opening scene on Tokyo Mirage Sessions:

Without PR:

https://i.imgur.com/mHpPEj8.mp4

With PR:

https://i.imgur.com/qDiSwL5.mp4

Leischii:
Results in Smash Bros Ultimate (Sorry for the blackscreens at the start of the video, mistake during recording):

Intro video before: https://i.imgur.com/QEiDYes.mp4

Intro video after: https://i.imgur.com/4M78XCK.mp4

How to play video before: https://i.imgur.com/LmzRbE1.mp4

How to play video after: https://i.imgur.com/977gO14.mp4

Love the changes. Most notable change is the how to play video when someone is being kicked of the stage.
GJ gdkchan :)

Leischii:
Animal Crossing New Horizon:
Introduction video (Again, sry for the brief blackscreen in the beginning):

Before: https://i.imgur.com/UhcMohj.mp4

After: https://i.imgur.com/11p49qD.mp4

Runs without any stutters after the first few seconds now :)

AcK77 (approved):

PR #2014, opened 2021-02-11
prepo: Implement GetSystemSessionId and add perms

AcK77:
This PR implement `GetSystemSessionId` of prepo service accordingly to RE.
During my research I've pointed out what the result code `90` is and where it's used in our actual implementation. So I've added it with some missing services too.

This should let "Super Mario 3D World + Bowser's Fury" run without `Ignore Missing Services` enabled. 

Thanks to Ga1act1c from our Discord for let us know the issue.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
+++ b/Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
@@ -10,16 +10,17 @@ using System.Text;
 
 namespace Ryujinx.HLE.HOS.Services.Prepo
 {
-    [Service("prepo:a",  -1)] // 1.0.0-5.1.0
-    [Service("prepo:a2", -1)] // 6.0.0+
-    [Service("prepo:m",   6)]
-    [Service("prepo:u",   1)]
-    [Service("prepo:s",   2)]
+    [Service("prepo:a",  PrepoServicePermissionLevel.Admin)] // 1.0.0-5.1.0
+    [Service("prepo:a2", PrepoServicePermissionLevel.Admin)] // 6.0.0+
+    [Service("prepo:m",  PrepoServicePermissionLevel.Manager)]
+    [Service("prepo:u",  PrepoServicePermissionLevel.User)]
+    [Service("prepo:s",  PrepoServicePermissionLevel.System)]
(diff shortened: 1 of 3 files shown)


Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
```diff
@@ -57,6 +75,26 @@ public ResultCode GetTransmissionStatus(ServiceCtx context)
...
+        public ResultCode GetSystemSessionId(ServiceCtx context)
+        {
+            if ((_permission & 1) == 0)
+            {
+                return ResultCode.PermissionDenied;
+            }
+
+            byte[] randomBuffer = new byte[8];
+
+            new Random().NextBytes(randomBuffer);
```
riperiperi, line 89: We alright with this being randomly generated each time? Surely two requests in a row would obtain the same system session ID? (though i'm not sure at what point a session ID would be generated)

Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
```diff
@@ -5,23 +5,36 @@
...
-    [Service("prepo:u")]
+    [Service("prepo:a",  -1)] // 1.0.0-5.1.0
+    [Service("prepo:a2", -1)] // 6.0.0+
+    [Service("prepo:m",   6)]
+    [Service("prepo:u",   1)]
+    [Service("prepo:s",   2)]
     class IPrepoService : IpcService
     {
-        public IPrepoService(ServiceCtx context) { }
+        private long _permission;
```
gdkchan, line 20: Would be better to create a enum for those permissions.

Ryujinx.HLE/HOS/Services/Prepo/IPrepoService.cs
gdkchan, line 89: Is there any advantage in using random values here? I'm not a huge fan of that since its just another thing that could cause random and hard to reproduce bugs (since each run will have a different value).

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
LGTM.

PR #2017, opened 2021-02-13
[draft] InstEmitSimdArithmetic: One Newton step in Frsqrte_S

merryhime:
Clamping the input then doing a Newton step on the output before rounding improves accuracy.

For a demonstration of various approaches I had previously tested, see: https://gist.github.com/MerryMage/6c0d69bb400397b3d1014f19b797b03b


merryhime:
The speed:accuracy tradeoff of the above is probably not worth it. lmk if you want a fully accurate impl.

PR #2024, opened 2021-02-16
Allow modding AddOnContent RomFS

mageven:
This approach should work fine assuming games behave properly and don't spam `OpenDataStorageByDataId`.
Slow path is only hit when mods are detected.


ai-Robot-2029:
I tested it and the both parts of the mod loaded. 
Also tried to play some other games like SMO, SM3DW&BF, MK8D, Hollow Knight and all worked fine. 
There were no crashes and didn't notice anything different from how the games run on master.

AcK77 (approved):

Thog (approved):

PR #2026, opened 2021-02-16
Fix memory tracking performance regression

gdkchan:
Fix a regression introduced on #1987 due to the read/write flags being inverted.
It is not possible to track memory writes without also tracking reads, but tracking reads without tracking writes is possible. The previous code assumed the opposite, which lead to it taking the slow path for memory access when it shouldn't.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitMemoryHelper.cs
@@ -334,14 +334,14 @@ namespace ARMeilleure.Instructions
             {
                 if (write)
                 {
-                    pte = context.ShiftLeft(pte, Const(1));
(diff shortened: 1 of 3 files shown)


jduncanator (approved):

riperiperi (approved):
That's better. It's worth noting the why this is:

- Write tracking is used to set dirty flags
- Read tracking is used to flush data before access

In all cases, we want to flush the data before access. By flushing on write, we can properly cover situations where data is being partially overwritten - the parts that aren't written will contain the flushed data. 

If write tracking is also signalled on read protection only, then we'll end up hitting the slow path and not doing anything (no flushing read action, no write and reprotect) and it will _not_ unprotect the region of memory that was accessed, which will just cause every memory access in sequence for that page to hit the slow path...

Issue #1526, opened 2020-09-03
[Feature Request] Change File Log location directory


qurious-pixel:
#### Overview:
Move logging out of Publish folder into config folder

#### Smaller Details: 
Placing logging in a known writeable area.

#### Nature of Request:  
- Change
Ryujinx/Ui/SettingsWindow.cs
Ryujinx.Common/Logging/Targets/FileLogTarget.cs

### Why would this feature be useful?
Sandboxed apps such as AppImages and FlatPaks cannot write to the read-only section.
This crashes the emulator.

macabeus:
Related PR: https://github.com/Ryujinx/Ryujinx/pull/1530

qurious-pixel:
closing this since it can be selected `return AppDataManager.BaseDirPath;` by setting ChannelOwner.

      private const string FlatHubChannelOwner = "flathub";

TSRBerry:
Yes, but I still think we shouldn't log to the "installation" directory in any case. So I'd like to keep your feature request around for that if that's fine for you!

PR #2027, opened 2021-02-16
Add custom log dir

Joshi234:
This adds an option to set your own log path.
Closes https://github.com/Ryujinx/Ryujinx/issues/1526


EmulationFanatic:
Requires a restart of the application after setting the Logs path. I suppose this makes sense, but it would be nice if it somehow took effect immediately, or automatically restarted Ryujinx after setting.

Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
```diff
@@ -78,6 +78,11 @@ public class ConfigurationFileFormat
         /// </summary>
         public bool LoggingEnableFsAccessLog { get; set; }
 
+        /// <summary>
+        /// Sets the log dir
+        /// </summary>
+        public string CustomLogDir { get; set; }
```
jduncanator, line 84: Probably best to use the full word "Directory", there isn't any reason to shorten it here.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -138,6 +138,11 @@ public class LoggerSection
             /// </summary>
             public ReactiveObject<bool> EnableFsAccessLog { get; private set; }
 
+            /// <summary>
+            /// Sets a custom directory where logs get saved
+            /// </summary>
+            public ReactiveObject<string> CustomLogDir { get; private set; }
```
jduncanator, line 144: Ditto.

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -74,11 +74,21 @@ private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
         {
             if (e.NewValue)
             {
+                if (ConfigurationState.Instance.Logger.CustomLogDir.Value == "" || ConfigurationState.Instance.Logger.CustomLogDir.Value==null) {
```
jduncanator, line 77: This can just be `String.IsNullOrEmpty(ConfigurationState.Instance.Logger.CustomLogDir.Value)`

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -74,11 +74,21 @@ private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
...
                 Logger.AddTarget(new AsyncLogTargetWrapper(
-                    new FileLogTarget(AppDomain.CurrentDomain.BaseDirectory, "file"),
+                    new FileLogTarget(Path.Combine(AppDomain.CurrentDomain.BaseDirectory+"Logs"), "file"),
                     1000,
                     AsyncLogTargetOverflowAction.Block
                 ));
+                }
+                else
+                {
+                    Logger.AddTarget(new AsyncLogTargetWrapper(
```
jduncanator, line 86: You should check that the directory exists before using it, otherwise the application will just crash. If the folder doesn't exist, it should fallback to using the default logs directory.

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -74,11 +74,21 @@ private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
         {
             if (e.NewValue)
             {
+                if (ConfigurationState.Instance.Logger.CustomLogDir.Value == "" || ConfigurationState.Instance.Logger.CustomLogDir.Value==null) { 
                 Logger.AddTarget(new AsyncLogTargetWrapper(
-                    new FileLogTarget(AppDomain.CurrentDomain.BaseDirectory, "file"),
+                    new FileLogTarget(Path.Combine(AppDomain.CurrentDomain.BaseDirectory+"Logs"), "file"),
```
jduncanator, line 79: Why was "Logs" appended here? This is incorrect anyway, it needs a path separator after the BaseDirectory, it should be `Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs")`

Ryujinx/Program.cs
```diff
@@ -122,6 +122,14 @@ static void Main(string[] args)
...
             }
 
+            if (!(ConfigurationState.Instance.Logger.CustomLogDir.Value == "" | ConfigurationState.Instance.Logger.CustomLogDir.Value == null)) {
+                Logger.AddTarget(new AsyncLogTargetWrapper(
+                new FileLogTarget(ConfigurationState.Instance.Logger.CustomLogDir.Value, "file"),
+                    1000,
+                    AsyncLogTargetOverflowAction.Block
+                ));
+            }
+
```
jduncanator, line 132: Why is the File log target added twice? Once here and once in the LoggerModule?

Ryujinx/Ui/MainWindow.cs
```diff
@@ -915,8 +915,15 @@ private void Open_Ryu_Folder(object sender, EventArgs args)
 
         private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
-            string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
-
+            string logPath;
+            if (!(ConfigurationState.Instance.Logger.CustomLogDir.Value == "" | ConfigurationState.Instance.Logger.CustomLogDir.Value == null))
```
jduncanator, line 919: Ditto.

Ryujinx/Program.cs
```diff
@@ -122,6 +122,14 @@ static void Main(string[] args)
                 ConfigurationState.Instance.Ui.StartFullscreen.Value = true;
             }
 
+            if (!(ConfigurationState.Instance.Logger.CustomLogDir.Value == "" | ConfigurationState.Instance.Logger.CustomLogDir.Value == null)) {
```
jduncanator, line 125: Ditto.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -479,6 +486,7 @@ public void LoadDefault()
             Logger.EnableError.Value               = true;
             Logger.EnableGuest.Value               = true;
             Logger.EnableFsAccessLog.Value         = false;
+            Logger.CustomLogDir.Value              = "";//When set to nothing it saves into the Ryujinx application folder
```
jduncanator, line 489: I don't think we need this comment here. If it's needed, it should be a XML doc on the CustomLogDirectory field itself.

jduncanator (changes requested):

Ryujinx.Common/Configuration/LoggerModule.cs
EmulationFanatic: It created the folder automatically when I tested it.

jduncanator: Best not to assume this behaviour, that seems like a side affect of how the file is opened.

Ryujinx/Program.cs
Joshi234: Because here it gets added when logging to file gets enabled

jduncanator: This was never required before, and the logging worked fine. Why is it required now? I don't see any other changes that would require this to be the case.

Ryujinx/Ui/Windows/SettingsWindow.glade
```diff
@@ -2034,6 +2034,40 @@
...
+                            </object>
+                            <packing>
+                              <property name="expand">False</property>
+                              <property name="fill">True</property>
+                              <property name="padding">13</property>
+                              <property name="position">0</property>
+                            </packing>
+                          </child>
+                          <child>
+                            <object class="GtkEntry" id="_customLogDir">
```
jduncanator, line 2056: This should be a folder/file browser entry, not a free text entry.

Ryujinx.Common/Configuration/LoggerModule.cs
Joshi234: It already does this in FileLogTarget

jduncanator: I'm more thinking along the lines of: The user enters `ImBeingAGiantIdiot` into the custom log directory field. This isn't even a valid path, and passing this down to the underlying log target will most likely result in a crash.

It should validate the path exists, is valid, and if it isn't, should either fallback to the default directory, or create it.

Ryujinx/Program.cs
Joshi234: Yeah I was pretty dumb

mageven:
Tying logging to yet another component that can fail (one that does so frequently too looking at support questions on discord) doesn't sound like a good idea to me.

This makes support harder too as there's not one common path to recommend to users for getting logs. It's already difficult as it is now (even considering "Open Log Folder" which isn't always possible)

Also, I'm trying to understand why a regular user needs this (or why a GUI config I guess). Logs are pretty lightweight so the reason isn't lack of space.

If you want custom log locations, I suggest passing it via command line. It'll work without side-effects at least (like requiring a reboot on location change). But that's kind of redundant, when redirection works just as well.

This also doesn't close #1526 imo. They wanted logs in the data directory (for a different purpose).

jduncanator:
> Tying logging to yet another component that can fail (one that does so frequently too looking at support questions on discord) doesn't sound like a good idea to me.
> 
> This makes support harder too as there's not one common path to recommend to users for getting logs. It's already difficult as it is now (even considering "Open Log Folder" which isn't always possible)
> 
> Also, I'm trying to understand why a regular user needs this (or why a GUI config I guess). Logs are pretty lightweight so the reason isn't lack of space.
> 
> If you want custom log locations, I suggest passing it via command line. It'll work without side-effects at least (like requiring a reboot on location change). But that's kind of redundant, when redirection works just as well.
> 
> This also doesn't close #1526 imo. They wanted logs in the data directory (for a different purpose).

I'm sort-of of the same thinking, although I think we can go a step further than the command line option.

What was suggested in that issue was a configuration option in the config file. I think that's fine, and I think we should implement that, but I too question the requirement for an actual GUI option to configure that. The user in that issue would then be able to set their logging directory in the config file before booting Ryujinx, and have everything they wanted (I think this was for AppImage support, which needs everything in specific locations).

I'm sort of averse to adding too many command line options, unless we want to go all in and have everything configurable by a command line option too.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -915,7 +915,23 @@ private void Open_Ryu_Folder(object sender, EventArgs args)
 
         private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
-            string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
+            string logPath;
```
AcK77, line 918: Add new empty line after

Ryujinx/Ui/MainWindow.cs
```diff
@@ -915,7 +915,23 @@ private void Open_Ryu_Folder(object sender, EventArgs args)
...
+            if (!String.IsNullOrEmpty(ConfigurationState.Instance.Logger.CustomLogDirectory.Value))
+            {
+                if (Directory.Exists(ConfigurationState.Instance.Logger.CustomLogDirectory)) { 
+                    logPath = ConfigurationState.Instance.Logger.CustomLogDirectory;
+                }
+
```
AcK77, line 924: Remove extra empty line

Ryujinx/Ui/MainWindow.cs
```diff
@@ -915,7 +915,23 @@ private void Open_Ryu_Folder(object sender, EventArgs args)
...
+                else
+                {
+                    logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
+                }
+            }
+
```
AcK77, line 930: Remove extra empty line

Ryujinx/Ui/MainWindow.cs
```diff
@@ -915,7 +915,23 @@ private void Open_Ryu_Folder(object sender, EventArgs args)
 
         private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
-            string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
+            string logPath;
+            if (!String.IsNullOrEmpty(ConfigurationState.Instance.Logger.CustomLogDirectory.Value))
```
AcK77, line 919: You should group the both if condition, then if the directory exists and it’s not empty (since here the string can’t be null) it use the custom path, else, it use the default one. Btw the { should be on a new line.

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -36,7 +47,14 @@ public FileLogTarget(string path, string name, FileShare fileShare, FileMode fil
             path = Path.Combine(logDir.FullName, $"Ryujinx_{version}_{DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss")}.log");
 
             _name      = name;
-            _logWriter = new StreamWriter(File.Open(path, fileMode, FileAccess.Write, fileShare));
+            try {
```
AcK77, line 50: The { should be on a new line

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -36,7 +47,14 @@ public FileLogTarget(string path, string name, FileShare fileShare, FileMode fil
             path = Path.Combine(logDir.FullName, $"Ryujinx_{version}_{DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss")}.log");
 
             _name      = name;
-            _logWriter = new StreamWriter(File.Open(path, fileMode, FileAccess.Write, fileShare));
+            try { 
+                _logWriter = new StreamWriter(File.Open(path, fileMode, FileAccess.Write, fileShare));
+            }
+            catch
+            {
+                return;
```
AcK77, line 55: Why do you silent all exception, it’s usually a bad practice, you should fix the issue instead of silent it, if it’s possible!

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -19,9 +19,20 @@ public FileLogTarget(string path, string name)
...
+            if (Directory.Exists(path))
+            {
+                 logDir = new DirectoryInfo(path);
+            }
+            else
+            {
+                logDir = new DirectoryInfo(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs"));
+                logDir.Create();
+            }
+
```
AcK77, line 34: Remove the extra 2 empty lines

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -7,7 +7,7 @@ namespace Ryujinx.Common.Logging
 {
     public class FileLogTarget : ILogTarget
     {
-        private readonly StreamWriter  _logWriter;
```
AcK77, line 10: Why it’s not readonly anymore?

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -7,7 +7,7 @@ namespace Ryujinx.Common.Logging
...
+        private StreamWriter  _logWriter;
```
AcK77, line 10: Alignment is wrong (but you shouldn’t change that line)

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -19,9 +19,20 @@ public FileLogTarget(string path, string name)
 
         public FileLogTarget(string path, string name, FileShare fileShare, FileMode fileMode)
         {
+            DirectoryInfo logDir;
+
             // Ensure directory is present
-            DirectoryInfo logDir = new DirectoryInfo(Path.Combine(path, "Logs"));
-            logDir.Create();
+            if (Directory.Exists(path))
```
AcK77, line 25: Why do you change that if the path is passed as argument? You should change the path where this method is called instead.

CeruleanSky:
>If you want custom log locations, I suggest passing it via command line.

If you use the command line switch for a custom user directory could the log folder just go in there?  It seems that solution works fine for other emulators, and it would probably work for the flatpack/appimage users as well.  Then there would be only two places it would be found, either the default, or an intentionally picked, and thus hopefully unconfusing place along side every other user file for portable/custom/flatpack/etc users.

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -70,20 +71,59 @@ private static void ReloadFilteredClasses(object sender, ReactiveEventArgs<LogCl
...
         }
 
-        private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
+        public static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
         {
             if (e.NewValue)
             {
-                Logger.AddTarget(new AsyncLogTargetWrapper(
-                    new FileLogTarget(AppDomain.CurrentDomain.BaseDirectory, "file"),
+                if (Logger.HasTarget("file"))
```
AcK77, line 78: You should just check the path, if the custom isn’t sets, use the default one, check if the folder exists (as jD suggested) and then start the FileLogger. You duplicate code for nothing here and it could be done much easier.

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -137,6 +137,15 @@ private static ILogTarget GetTarget(string targetName)
             return null;
         }
 
+        public static bool HasTarget(string target)
+        {
+            if (GetTarget(target) != null)
+            {
+                return true;
+            }
+            return false;
```
AcK77, line 146: Add a new empty line before

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -70,20 +71,59 @@ private static void ReloadFilteredClasses(object sender, ReactiveEventArgs<LogCl
             }
         }
 
-        private static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
+        public static void ReloadFileLogger(object sender, ReactiveEventArgs<bool> e)
```
AcK77, line 74: Why it’s public now?

AcK77 (changes requested):
A bunch of things need to be fixed. Maybe you should change the logic here and there to get something which work easier without some not needed code!
Nice try, I hope you will find a nice way to fix everything 😊!

mageven:
> I'm sort of averse to adding too many command line options, unless we want to go all in and have everything configurable by a command line option too.

I'd like it if there's a cli-only Ryujinx in the long term.

> If you use the command line switch for a custom user directory could the log folder just go in there? It seems that solution works fine for other emulators, and it would probably work for the flatpack/appimage users as well. Then there would be only two places it would be found, either the default, or an intentionally picked, and thus hopefully unconfusing place along side every other user file for portable/custom/flatpack/etc users.

@CeruleanSky's solution seems like a good compromise to me. The log files have version info in and on them so that's not an issue.
But I'm fine with a config option as long as it is robust and doesn't miss any logs.

Joshi234:
And what about an advanced tab in the settings for settings that most people won't use but might be useful to some people?

EmulationFanatic:
> And what about an advanced tab in the settings for settings that most people won't use but might be useful to some people? I don't see why adding a the option in the config file but not in the settings when it's already implemented, it just makes it less user friendly

I think the GUI field is the most user friendly option here and should be kept. Having config file options and a command line parameter on top of it would be great too. This way it be doesn't matter how a user enjoys using Ryujinx most; they can have the feature in any context.

Joshi234:
So what should we do now?

EmulationFanatic:
> So what should we do now?

I think adding a command line switch would satisfy any outstanding requirements (and it would overrule any GUI setting).

Ryujinx.Common/Configuration/LoggerModule.cs
```diff
@@ -18,6 +18,7 @@ public static void Initialize()
             ConfigurationState.Instance.Logger.EnableFsAccessLog.Event += ReloadEnableFsAccessLog;
             ConfigurationState.Instance.Logger.FilteredClasses.Event   += ReloadFilteredClasses;
             ConfigurationState.Instance.Logger.EnableFileLog.Event     += ReloadFileLogger;
+            ConfigurationState.Instance.Logger.CustomLogDirectory.Event      += ReloadFileLoggerNewDirectory;
```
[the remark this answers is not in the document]
Joshi234: Is it possible to have an unknown type for `ReactiveEventArgs<bool> e?`

qurious-pixel:
Could it be done via command-line switch like `-r` and `-f`
[Custom Data Dir](https://github.com/Ryujinx/Ryujinx/blob/5ab5b0e709c3667ca85a6727620ff26f074e4ce5/Ryujinx/Program.cs#L38-L57)
<details><summary>Code Block</summary>

            {
                string arg = args[i];

                if (arg == "-r" || arg == "--root-data-dir")
                {
                    if (i + 1 >= args.Length)
                    {
                        Logger.Error?.Print(LogClass.Application, $"Invalid option '{arg}'");

                        continue;
                    }

                    baseDirPathArg = args[++i];
                }
                else if (arg == "-f" || arg == "--fullscreen")
                {
                    startFullscreenArg = true;
                }
                else if (launchPathArg == null)
                {
                    launchPathArg = arg;
                }
            }
</details>

Add something like `-l` and `--user-log-dir`

Joshi234:
Closing this since it's kind a stupid, maybe in the future when it's needed I gonna implement it

PR #2029, opened 2021-02-17
settings: Fix returned buffer size of GetFirmwareVersion

AcK77:
This PR fix an issue introduced on last IPC rewrite PRs where some returned buffer size have to be explicit now.
GetFirmwareVersion/GetFirmwareVersion2 missed this and could crashes some homebrew because libnx have to check the firmware version to allow some call usage (with `hosversionBefore()`).


Thog (approved):
lgtm, it would be nice to upgrade the default version tho

AcK77:
> lgtm, it would be nice to upgrade the default version tho

This will be done in another PR later, since it’s just an edge case where there is no firmware installed

PR #2036, opened 2021-02-19
nifm: Fix returned buffer size of GetClientId

AcK77:
This PR fix an issue introduced on last IPC rewrite PRs where some returned buffer size have to be explicit now.
`GetClientId` without an explicit buffer size returns some garbage data to the guest and then `nifm` service crashes because of a wrong ClientId.

Horizon Chase Turbo and Doom regression are fixed.  (Probably some other games too)
image


gdkchan (approved):
Nice catch 👍

Thog (approved):

PR #2039, opened 2021-02-20
Add MotionButton to simulate shaking of controller

CeruleanSky:
Adds a hotkey(inspired by yuzu feature) that simulates shaking of a motion devices.  It is currently defaults to Grave " ` " .  The ranges are from shaking my phone while running the sixaxis homebrew from https://github.com/switchbrew/switch-examples/tree/master/hid/sixaxis .  A hotkey was chosen because a person might not have a joystick button to assign or may not be using keyboard input as a controller.

It works for my purposes as hitting a key is easier than using my feet to shake my phone, perhaps this will work for others as well.

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationState.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationState.cs
@@ -517,7 +517,8 @@ namespace Ryujinx.Configuration
             Hid.EnableKeyboard.Value               = false;
             Hid.Hotkeys.Value = new KeyboardHotkeys
             {
-                ToggleVsync = Key.Tab
+                ToggleVsync = Key.Tab,
+                MotionButton = Key.Grave
             };
             Hid.InputConfig.Value = new List<InputConfig>
             {
@@ -716,7 +717,7 @@ namespace Ryujinx.Configuration
 
                 configurationFileFormat.Hotkeys = new KeyboardHotkeys
                 {
-                    ToggleVsync = Key.Tab
+                    ToggleVsync = Key.Tab,
                 };
 
                 configurationFileUpdated = true;
@@ -800,6 +801,18 @@ namespace Ryujinx.Configuration
                 Common.Logging.Logger.Warning?.Print(LogClass.Application, $"Outdated configuration version {configurationFileFormat.Version}, migrating to version 22.");
 
                 configurationFileFormat.HideCursorOnIdle = false;
+
+                configurationFileUpdated = true;
+            }
+
+            if (configurationFileFormat.Version < 23)
+            {
+                Common.Logging.Logger.Warning?.Print(LogClass.Application, $"Outdated configuration version {configurationFileFormat.Version}, migrating to version 23.");
+
+                configurationFileFormat.Hotkeys = new KeyboardHotkeys
+                {
+                    MotionButton = Key.Grave,
+                };
 
                 configurationFileUpdated = true;
             }
--- a/Ryujinx.Common/Configuration/ConfigurationState.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationState.cs
@@ -717,7 +717,7 @@ namespace Ryujinx.Configuration
 
                 configurationFileFormat.Hotkeys = new KeyboardHotkeys
                 {
-                    ToggleVsync = Key.Tab,
+                    ToggleVsync = Key.Tab
                 };
 
                 configurationFileUpdated = true;
@@ -811,7 +811,8 @@ namespace Ryujinx.Configuration
 
                 configurationFileFormat.Hotkeys = new KeyboardHotkeys
                 {
-                    MotionButton = Key.Grave,
+                    ToggleVsync = Key.Tab,
+                    MotionButton = Key.Grave
                 };
 
                 configurationFileUpdated = true;
--- a/Ryujinx/Modules/Motion/MotionDevice.cs
+++ b/Ryujinx/Modules/Motion/MotionDevice.cs
@@ -97,7 +97,6 @@ namespace Ryujinx.Modules.Motion
             value.X = float.IsNegative(value.X) ? MathF.Ceiling(value.X * power) / power : MathF.Floor(value.X * power) / power;
             value.Y = float.IsNegative(value.Y) ? MathF.Ceiling(value.Y * power) / power : MathF.Floor(value.Y * power) / power;
             value.Z = float.IsNegative(value.Z) ? MathF.Ceiling(value.Z * power) / power : MathF.Floor(value.Z * power) / power;
-
             return value;
         }
     }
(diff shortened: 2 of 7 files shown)


AcK77:
Is that needed/useful for anything else than testing a sixaxis homebrew ?

CeruleanSky:
Any software that has a shake controller motion gesture and you don't want to setup a motion server or actually have to shake your device physically. Super Mario Odyssey for example has shaking your controller as a modifier for many action.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -716,7 +717,7 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
 
                 configurationFileFormat.Hotkeys = new KeyboardHotkeys
                 {
-                    ToggleVsync = Key.Tab
+                    ToggleVsync = Key.Tab,
```
riperiperi, line 720: I don't think we usually put a comma on the last item of a list.

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -804,6 +805,18 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
                 configurationFileUpdated = true;
             }
 
+            if (configurationFileFormat.Version < 23)
+            {
+                Common.Logging.Logger.Warning?.Print(LogClass.Application, $"Outdated configuration version {configurationFileFormat.Version}, migrating to version 23.");
+
+                configurationFileFormat.Hotkeys = new KeyboardHotkeys
+                {
+                    MotionButton = Key.Grave,
```
riperiperi, line 814: This is a redefinition - it replaces the `KeyboardHotkeys` object above which defines `ToggleVsync`, so after this is done it will be entirely undefined. You should set `ToggleVsync` here too.

Ryujinx/Modules/Motion/MotionDevice.cs
```diff
@@ -67,6 +67,29 @@ public void Poll(InputConfig config, int slot)
...
+
+            if (!config.EnableMotion)
+            {
+                Accelerometer = new Vector3();
+                Gyroscope = new Vector3();
+
+                return;
+            }
+
+            Gyroscope = new Vector3(random.Next(-2000, 2000) * 0.001f);
```
riperiperi, line 83: What format is gyro data in? If it's a normalized vector, this could produce some undesirable results.

Ryujinx/Modules/Motion/MotionDevice.cs
```diff
@@ -67,6 +67,29 @@ public void Poll(InputConfig config, int slot)
...
+            Accelerometer = new Vector3(random.Next(-2000, 2000) * 0.001f);
+            Rotation = Vector3.Zero;
+
+            Orientation[0] = 1f;
+            Orientation[4] = 1f;
+            Orientation[8] = 1f;
+
```
riperiperi, line 90: Unnecessary space.

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -605,7 +605,14 @@ private bool UpdateFrame()
 
                 currentButton |= _device.Hid.UpdateStickButtons(leftJoystick, rightJoystick);
 
-                motionDevice.Poll(inputConfig, inputConfig.Slot);
+                HotkeyButtons currentHotkeyButtons = KeyboardController.GetHotkeyButtons(OpenTK.Input.Keyboard.GetState());
+                if (currentHotkeyButtons.HasFlag(HotkeyButtons.MotionButton))
```
riperiperi, line 609: Nit: add a newline above

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -605,7 +605,14 @@ private bool UpdateFrame()
...
+                {
+                    motionDevice.PollRandomData(inputConfig);
+                }
+                else {
```
riperiperi, line 613: This bracket should be on the next line.

riperiperi (changes requested):
This is particularly useful, nice work. One thing though:

Are we OK with making this a hotkey rather than a controller button? Perhaps each player would want to have their own shake button, and bind it to their controller rather than the keyboard.

I'm also not sure about the aggressive gyro rotation. I'm not sure how many games depend on gyro change to detect motion controls that could be triggered by violent shaking rather than precision, but it will definitely cause the screen to shake horribly in games with gyroscope aiming. Maybe accelerometer only could work?

Ryujinx.Common/Configuration/ConfigurationState.cs
CeruleanSky: Slipped in when experimenting to figure out how config worked. Fixed.

Ryujinx.Common/Configuration/ConfigurationState.cs
CeruleanSky: Done and removed trailing comma as well.

Ryujinx/Modules/Motion/MotionDevice.cs
CeruleanSky: I tested Yuzu, whose feature inspired this, on the sixaxis homebrew, their motion button feature did not produce a normalized vector, but they did clamp the components in the range of -1 to 1.  I chose -2 to 2 from shaking my phone and watching the values produced by the net.sshnuke.dsu.MotionSource Android software in the sixaxis homebrew. 

I'll hold off for now, but if your testing with software and hardware indicates that it should be normalized, that can be done easily.

CeruleanSky:
>Are we OK with making this a hotkey rather than a controller button? Perhaps each player would want to have their own shake button, and bind it to their controller rather than the keyboard.

I generally use [antimicro](https://github.com/AntiMicro/antimicro/) for binding a key to a controller.  Perhaps it would be preferred to override or dual purpose a controller button instead. I picked hotkey so it works for keyboard and controller users and doesn't interfere with their usual controls.  As far as multiple people, I did not really have multiple people with their controllers in mind for this. I'm sure they would enjoy such a thing, but at least from me, it will have to be a future feature, as I'm not really setup to test it. 

>I'm also not sure about the aggressive gyro rotation. I'm not sure how many games depend on gyro change to detect motion controls that could be triggered by violent shaking rather than precision, but it will definitely cause the screen to shake horribly in games with gyroscope aiming. Maybe accelerometer only could work?

I'll be glad to adjust things, I could even offer config presets, accel only, gyro only, harmonic oscillation, etc , but I don't really have much to test with, and was just using Yuzu's values in the sixaxis homebrew as guidance, which does activate both the gyro and the acceleration.   As far as the aggressiveness, while the ranges worked for my purposes, if it is causing issues, perhaps my phone was shaken a bit harder than needed when I was sliding it on the table....

CeruleanSky:
Due to the dropping of OpenTK3 which includes a newer motion system, a new input system, and due to the lack of my ability to test and develop it for multiple controllers, I am closing this.  Maybe the new system will allow an easier to add motion button in case anybody else wants to remake this.

CeruleanSky:
Due to the dropping of OpenTK3 which includes a newer motion system, a new input system, and due to the lack of my ability to test and develop it for multiple controllers, I am closing this. Maybe the new system will allow an easier to add motion button in case anybody else wants to remake this.

PR #2044, opened 2021-02-21
Fix virtual address overflow near ulong limit

Caian:
Some address arithmetic used in `MemoryManager` are prone to overflow when the virtual address resides in the last page possible of a 64bit address. This causes, for instance, `IsRangeMapped` to return true even though the address is not mapped, which can lead to a bad logic and a crash.

This PR creates an overflow-safe way of counting pages for the affected functions and replaces the address-based loop in `IsRangeMapped` with page-based loop.

### DIFF
--- a/Ryujinx.Cpu/MemoryManager.cs
+++ b/Ryujinx.Cpu/MemoryManager.cs
@@ -345,6 +345,23 @@ namespace Ryujinx.Cpu
             return ref _backingMemory.GetRef<T>(GetPhysicalAddressInternal(va));
         }
 
+        /// <summary>
+        /// Cumputes the number of pages in a virtual address range.
+        /// </summary>
+        /// <param name="va">Virtual address of the range</param>
+        /// <param name="size">Size of the range</param>
+        /// <param name="startVa">The virtual address of the beginning of the first page</param>
+        /// <remarks>This function does differentiate between allocated and unallocated pages.</remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int GetPagesCount(ulong va, ulong size, out ulong startVa)
+        {
+            // WARNING: Always check if ulong does not overflow during the operations.
+            startVa = va & ~(ulong)PageMask;
+            ulong vaSpan = (va - startVa + size + PageMask) & ~(ulong)PageMask;
+
+            return (int)(vaSpan / PageSize);
+        }
+
         private void ThrowMemoryNotContiguous() => throw new MemoryNotContiguousException();
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -358,11 +375,7 @@ namespace Ryujinx.Cpu
                 return false;
             }
 
-            ulong endVa = (va + (ulong)size + PageMask) & ~(ulong)PageMask;
-
-            va &= ~(ulong)PageMask;
-
-            int pages = (int)((endVa - va) / PageSize);
+            int pages = GetPagesCount(va, (ulong)size, out va);
 
             for (int page = 0; page < pages - 1; page++)
             {
@@ -396,11 +409,7 @@ namespace Ryujinx.Cpu
                 return null;
             }
 
-            ulong endVa = (va + size + PageMask) & ~(ulong)PageMask;
-
-            va &= ~(ulong)PageMask;
-
-            int pages = (int)((endVa - va) / PageSize);
+            int pages = GetPagesCount(va, size, out va);
 
             List<(ulong, ulong)> regions = new List<(ulong, ulong)>();
 
@@ -485,11 +494,9 @@ namespace Ryujinx.Cpu
                 return true;
             }
 
-            ulong endVa = (va + size + PageMask) & ~(ulong)PageMask;
+            int pages = GetPagesCount(va, size, out va);
 
-            va &= ~(ulong)PageMask;
-
-            while (va < endVa)
+            for (int page = 0; page < pages; page++)
             {
                 if (!IsMapped(va))
                 {
--- a/Ryujinx.Cpu/MemoryManager.cs
+++ b/Ryujinx.Cpu/MemoryManager.cs
@@ -346,12 +346,12 @@ namespace Ryujinx.Cpu
         }
 
         /// <summary>
-        /// Cumputes the number of pages in a virtual address range.
+        /// Computes the number of pages in a virtual address range.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -345,6 +345,23 @@ public WritableRegion GetWritableRegion(ulong va, int size)
             return ref _backingMemory.GetRef<T>(GetPhysicalAddressInternal(va));
         }
 
+        /// <summary>
+        /// Cumputes the number of pages in a virtual address range.
```
Thog, line 349:
```suggestion
        /// Computes the number of pages in a virtual address range.
```

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -345,6 +345,23 @@ public WritableRegion GetWritableRegion(ulong va, int size)
...
+        /// </summary>
+        /// <param name="va">Virtual address of the range</param>
+        /// <param name="size">Size of the range</param>
+        /// <param name="startVa">The virtual address of the beginning of the first page</param>
+        /// <remarks>This function does differentiate between allocated and unallocated pages.</remarks>
```
Thog, line 354:
```suggestion
        /// <remarks>This function does differentiate between allocated and unallocated pages.</remarks>
```

Caian: Sorry @Thog but I can't see the suggested change.

Thog: Seems like my suggestion got lost in the middle 😕

Caian: Oh ok, thanks, fixed :)

Thog (approved):

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -345,6 +345,23 @@ public WritableRegion GetWritableRegion(ulong va, int size)
...
 
+        /// <summary>
+        /// Computes the number of pages in a virtual address range.
+        /// </summary>
+        /// <param name="va">Virtual address of the range</param>
+        /// <param name="size">Size of the range</param>
+        /// <param name="startVa">The virtual address of the beginning of the first page</param>
+        /// <remarks>This function does not differentiate between allocated and unallocated pages.</remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int GetPagesCount(ulong va, ulong size, out ulong startVa)
```
gdkchan, line 356:
```suggestion
        private static int GetPagesCount(ulong va, ulong size, out ulong startVa)
```

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -345,6 +345,23 @@ public WritableRegion GetWritableRegion(ulong va, int size)
...
+        {
+            // WARNING: Always check if ulong does not overflow during the operations.
+            startVa = va & ~(ulong)PageMask;
+            ulong vaSpan = (va - startVa + size + PageMask) & ~(ulong)PageMask;
```
gdkchan, line 360: This can still overflow if for example, `va` is 0 and `size` is 0xFFFFFFFFFFFFFFFF (not that I expect those values to be ever used in pratice, but it would be nice to have the function behave correct for any possible input). In that case the amount of pages also wouldn't fit on a `int` which is a different problem. Might be better to just throw for large range sizes.

Caian: I agree that the function should be robust, but I'm worried about throwing an exception. For instance, I think IsRangeMapped should just return false for any unrealistic va / size combination.

What you think changing `size` from `ulong` to `uint` and adding:

```C#
private bool ValidateSize(ulong size)
{
    return size < _addressSpaceSize;
}
```

Or even

```C#
private bool ValidateAddressAndSize(ulong va, ulong size)
{
    ulong endVa = va + size;
    return endVa >= va && endVa >= size && endVa < _addressSpaceSize;
}
```

gdkchan: Personally I prefer the last one.

Caian: One correction though is that `endVa <= _addressSpaceSize`

Caian:
@gdkchan I noticed that `AddressSpaceManager` has the same address calculations as `MemoryManager`, should those be fixed as well?

gdkchan:
> 
> 
> @gdkchan I noticed that `AddressSpaceManager` has the same address calculations as `MemoryManager`, should those be fixed as well?

Sure.

Caian:
> > @gdkchan I noticed that `AddressSpaceManager` has the same address calculations as `MemoryManager`, should those be fixed as well?
> 
> Sure.

Done

gdkchan (approved):
lgtm, thanks.

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -627,17 +676,20 @@ public CpuSmartMultiRegionHandle BeginSmartGranularTracking(ulong address, ulong
...
+            AssertValidAddressAndSize(va, size);
+
             // We emulate guard pages for software memory access. This makes for an easy transition to
             // tracking using host guard pages in future, but also supporting platforms where this is not possible.
 
             // Write tag includes read protection, since we don't have any read actions that aren't performed before write too.
             long tag = (write ? 3L : 1L) << PointerTagBit;
 
-            ulong endVa = (va + size + PageMask) & ~(ulong)PageMask;
+            int pages = GetPagesCount(va, (uint)size, out va);
```
riperiperi, line 687: `va` is used by `Tracking.VirtualMemoryEvent(va, size, write);`, and must be the original va passed to this function.

Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -649,7 +701,7 @@ public void SignalMemoryTracking(ulong va, ulong size, bool write)
                     break;
                 }
 
-                va += PageSize;
+                pageStart++;
```
riperiperi, line 704: Removing the increment `va` here fixes a bug with memory tracking where the wrong base address was supplied to the tracking structure. Unfortunately, another issue was introduced by overwriting va above.

Ryujinx.Cpu/MemoryManager.cs
riperiperi: You should use another variable for the va returned by this function.

PR #2052, opened 2021-02-23
Relocate IHostUiHandler to Ryujinx.Common and reimplement loading progress bars over it

mageven:
This PR contains 2 changes.

1. The first change moves `IHostUiHandler` to a `Common.HostUiBridge` as it's not just `HLE` that needs to use it.
    
    This also changes `ExecuteProgram` to not use `Switch` class as we could've used `UserChannelPersistence` directly. If ever there's a need to change it, we can always rely on a lambda.

2. The first change allows implementing progress bars cleanly. This is done via a `StartProgressReport` call. It's used for both PTC and Shader cache and follows a polling method with customizable refresh rate (currently 100ms).

The original idea is thanks to @Joshi234's PR (#1946) which while functionally great, needed strange constructs because I didn't make the first change earlier.  This PR supersedes #1946 and so I've added appropriate credits to Joshi234 as part of the commits.

Right now, progress logs are only printed when there's no GUI. I'm not that fond of this solution and am open to feedback.


gdkchan:
I still think this should not be in `Ryujinx.Common`. I think the projects just just either expose a event that the GUI can subscribe to and receive progress updates, or allow passing a interface with a callback that would be called when progress changes.

mageven:
> 
> 
> I still think this should not be in `Ryujinx.Common`. I think the projects just just either expose a event that the GUI can subscribe to and receive progress updates, or allow passing a interface with a callback that would be called when progress changes.

I will look into it. I was too concerned about not having everything linking to UI in one class.

Issue #1676, opened 2020-11-07
Linux - Ryujinx isn't marked as executable after using built in updater


RickAndTired:
### What's the issue you encountered?

On Linux - Ryujinx isn't marked as executable after updating with the built in updater

### How can the issue be reproduced?

- Click Help 
- Check for Updates 
- Update
- Restart Ryujinx (doesn't work)
- The new Ryujinx isn't marked as executable

### Log file

NA

### Environment?

- Ryujinx version: 1.0.5636</br>
- System Specs:
  - OS: *Ubuntu 20.10*
  - CPU: *Ryzen 5 2600*
  - GPU: *RX 570*
  - RAM: *16GB*
- Applied Mods :  No 

### Additional context?

When downloading Ryujinx from the website the Ryujinx file is marked as executable

RickAndTired:
Update: This has been partially fixed.

When restarting through the update - everything works correctly. Ryujinx is marked as executable.

If you say no to restarting when running the update then Ryujinx will not be marked as executable.

PR #2055, opened 2021-02-23
Fix updater breaking user execute permissions on Unix

pineappleEA:
The updater currently sets the binary's (the _"Ryujinx"_ file on Unix) permissions to executable (for the current user) only when the Yes button is clicked on the restart dialog, since the part of the code that performs this task is inside the _YesButton_Clicked_ void in _UpdateDialog.cs_, which means that if you click No, close the updater, kill Ryujinx, etc. you will get a permission denied error the next time you attempt to run it.

This pr moves the part responsible for setting this permission to _Updater.cs_ so it is run after the updater finishes and before the restart dialog.

While when you build Ryujinx the executable has -rwxr-xr-r (any user can execute it), the updater sets the permissions to -rwxr--r-- (only the current user can execute it), this wasn't changed.

_P.S. A white space was also removed._

Fixes https://github.com/Ryujinx/Ryujinx/issues/1676

### DIFF
--- a/Ryujinx/Modules/Updater/Updater.cs
+++ b/Ryujinx/Modules/Updater/Updater.cs
@@ -2,6 +2,7 @@ using Gtk;
 using ICSharpCode.SharpZipLib.GZip;
 using ICSharpCode.SharpZipLib.Tar;
 using ICSharpCode.SharpZipLib.Zip;
+using Mono.Unix;
 using Newtonsoft.Json.Linq;
 using Ryujinx.Common.Logging;
 using Ryujinx.Ui;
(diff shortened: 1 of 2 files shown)


jduncanator (approved):

Ryujinx/Modules/Updater/Updater.cs
```diff
@@ -325,6 +326,18 @@ private static void DoUpdateWithSingleThread(UpdateDialog updateDialog, string d
             }
         }
         
+        private static void SetUnixPermissions()
+        {
+            string ryuBin  = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx");
```
riperiperi, line 331: Double space after `ryuBin`. The place it was copied from had this to align to other declarations, but they don't exist here.

Ryujinx/Modules/Updater/Updater.cs
```diff
@@ -325,6 +326,18 @@ private static void DoUpdateWithSingleThread(UpdateDialog updateDialog, string d
...
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                UnixFileInfo unixFileInfo = new UnixFileInfo(ryuBin);
+                unixFileInfo.FileAccessPermissions |= FileAccessPermissions.UserExecute;
+            }
+
```
riperiperi, line 338: Unnecessary newline

riperiperi (approved):
Sensible fix, nice one.

PR #2056, opened 2021-02-24
Force command flush after creating a syncpoint

riperiperi:
May fix sync object timeouts on AMD.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Sync.cs
+++ b/Ryujinx.Graphics.OpenGL/Sync.cs
@@ -26,6 +26,8 @@ namespace Ryujinx.Graphics.OpenGL
                 Handle = GL.FenceSync(SyncCondition.SyncGpuCommandsComplete, WaitSyncFlags.None)
             };
 
+            GL.ClientWaitSync(handle.Handle, ClientWaitSyncFlags.SyncFlushCommandsBit, 0); // Force commands to flush up to the syncpoint.
+
             lock (Handles)
             {
                 Handles.Add(handle);
--- a/Ryujinx.Graphics.OpenGL/Sync.cs
+++ b/Ryujinx.Graphics.OpenGL/Sync.cs
@@ -26,7 +26,8 @@ namespace Ryujinx.Graphics.OpenGL
                 Handle = GL.FenceSync(SyncCondition.SyncGpuCommandsComplete, WaitSyncFlags.None)
             };
 
-            GL.ClientWaitSync(handle.Handle, ClientWaitSyncFlags.SyncFlushCommandsBit, 0); // Force commands to flush up to the syncpoint.
+            // Force commands to flush up to the syncpoint.
+            GL.ClientWaitSync(handle.Handle, ClientWaitSyncFlags.SyncFlushCommandsBit, 0);
 
             lock (Handles)
             {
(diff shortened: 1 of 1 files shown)


Mou-Ikkai:
I'll try this later this day, seeing as I get about a hundred errors when I play Persona 5 Strikers on my 5700xt

riperiperi:
This does not appear to fix the issue. Unfortunately random stabs in the dark is all I can do at the moment.

Doctorwho1909:
@EmulationFanatic do you have an image of what i have to check is fixed?

EmulationFanatic:
image
This was happening on a 6800XT in xenoblade. Other AMD GPUs seem to exhibit similar but not identical behavior

Doctorwho1909:
> image
> This was happening on a 6800XT in xenoblade. Other AMD GPUs seem to exhibit similar but not identical behavior

ok i will test now

Doctorwho1909:
So played for 15 mins combat and such and no issues here is the log 
[Ryujinx_1.0.0+a3c440e_2021-03-10_21-24-28.log](https://github.com/Ryujinx/Ryujinx/files/6118992/Ryujinx_1.0.0%2Ba3c440e_2021-03-10_21-24-28.log) if you need more games tested let me know

Doctorwho1909:
Final Post it's good to go on linux also tested fedora and arch and ubuntu even clear linux

Mou-Ikkai:
I will soon test Persona 5 Strikers with my 5700XT.

Ext3h:
It should work around the issue, but it doesn't look like a sane solution.

You are now waiting in the same thread which has committed the work for the fence to be signaled. That means, you are effectively waiting for the GPU to signal you "I'm idle again".

You can't do that, or every usage of a fence is going to kill performance as you break any form of batching / pipelining.

What you can do, is issuing a regular flush, *without* waiting for the fence. That is enough to submit the fence (as a scheduling primitive) to the scheduler, so it becomes await-able in the first place. Even that is a a last-effort though, as you should avoid explicit flush as long as you can.

What you actually should be doing, is to double-track the fence as a software synchronization primitive *until the next flush on the originally issuing context*, and only then to signal the (client side) software fence, which may then allow waiting for the OpenGL fence object.

Take special care when the fence is awaited on the same thread, in that case you will need to flush explicitly when starting to wait - that is what the `GL_SYNC_FLUSH_COMMANDS_BIT` flag is actually for! You should not be using the `GL_SYNC_FLUSH_COMMANDS_BIT` when a flush was already performed.

riperiperi:
The timeout given to this command is a 0, so it does not _wait_ for the fence, it just flushes the commands and (needlessly) checks if the the syncpoint is reached. From my testing, glFlush was more expensive than doing this, and I haven't seen any performance impact on the nvidia driver from using the glClientWaitSync trick.

The flush can't come from the background thread that handles flush-on-access, as it uses a different context. It's too late at that point, unless we were to intervene and force a flush on the render thread using some kind of action queue. Flushes from the render thread were fine due to already using this flag when waiting on the fence, this issue was only happening with the background thread used to emulate unified memory with flushes.

There are no guarantees that a flush is needed for any one sync point, as that is defined by what the guest decides to access. We don't actually flush anywhere else.

Leischii:
Tested AC:NH, Smash Ultimate and Odyssey on Nvidia. No regressions found.

gdkchan (approved):
lgtm, this was also a issue on Intel, nice to see if fixed.

Thog (approved):

Kryptuq:
slight performance regression happening here for MH Rise:
unknown
unknownn

^only tested with master build
same results tho with merging PR#2138 & PR#1865 with 60fps mod enabled.

PR #2057, opened 2021-02-25
Add progress reporting to PTC and Shader Cache

mageven:
Both of them expose two events, one for state change and the other for progress which are consumed by `MainWindow`.
Both PTC and Shader Cache report progress events on a timer (currently 100ms).
This is because PTC/ShaderCache has (or eventually will have) parallel processing. So, it's not easy to serialize it per-change.

GUI changes and initial concept courtesy of @Joshi234 (credited changes) from #1946 which this supersedes.

### DIFF
--- a/ARMeilleure/Translation/PTC/Ptc.cs
+++ b/ARMeilleure/Translation/PTC/Ptc.cs
@@ -776,7 +776,7 @@ namespace ARMeilleure.Translation.PTC
 
             ThreadPool.QueueUserWorkItem(TranslationLogger, profiledFuncsToTranslate.Count);
 
-            PtcTranslationStateChanged(true);
+            PtcTranslationStateChanged?.Invoke(true);
 
             void TranslateFuncs()
             {
@@ -826,7 +826,7 @@ namespace ARMeilleure.Translation.PTC
             threads.Clear();
 
             _loggerEvent.Set();
-            PtcTranslationStateChanged(false);
+            PtcTranslationStateChanged?.Invoke(false);
 
             PtcJumpTable.Initialize(jumpTable);
 
@@ -846,7 +846,7 @@ namespace ARMeilleure.Translation.PTC
 
             do
             {
-                PtcTranslationProgressChanged(_translateCount, profiledFuncsToTranslateCount);
+                PtcTranslationProgressChanged?.Invoke(_translateCount, profiledFuncsToTranslateCount);
             }
             while (!_loggerEvent.WaitOne(refreshRate));
 
--- a/ARMeilleure/Translation/PTC/Ptc.cs
+++ b/ARMeilleure/Translation/PTC/Ptc.cs
@@ -70,6 +70,10 @@ namespace ARMeilleure.Translation.PTC
 
         internal static PtcState State { get; private set; }
 
+        // Progress update events
+        public static event Action<bool> PtcTranslationStateChanged;
+        public static event Action<int, int> PtcTranslationProgressChanged;
+
         static Ptc()
         {
             InitializeMemoryStreams();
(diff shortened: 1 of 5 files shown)


ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -833,15 +840,22 @@ void TranslateFuncs()
...
+            const int refreshRate = 100; // ms
 
             int profiledFuncsToTranslateCount = (int)state;
 
-            do
+            try
+            {
+                do
+                {
+                    PtcTranslationProgressChanged(_translateCount, profiledFuncsToTranslateCount);
```
gdkchan, line 851:
```suggestion
                    PtcTranslationProgressChanged?.Invoke(_translateCount, profiledFuncsToTranslateCount);
```
From what I recall, the event will be null if theres no listeners, so it shouldn't attempt to do the call in this case.

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -833,15 +840,22 @@ void TranslateFuncs()
...
+                }
+                while (!_loggerEvent.WaitOne(refreshRate));
+            }
+            catch (Exception e)
```
gdkchan, line 855: Why does this need to catch exception? I think the event handling code should make sure that no exceptions go uncatched.

Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -79,6 +79,24 @@ public sealed class GpuContext : IDisposable
         /// </summary>
         internal Capabilities Capabilities => _caps.Value;
 
+        /// <summary>
+        /// Passthrough for ShaderCache.ShaderCacheStateChanged event
```
gdkchan, line 83: This should have a proper description of the event. `ShaderCache.ShaderCacheStateChanged` is not public, so imo doesn't make sense to mention it in public facing documentation.

Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -79,6 +79,24 @@ public sealed class GpuContext : IDisposable
...
+        /// </summary>
+        public event Action<bool> ShaderCacheStateChanged
+        {
+            add => Methods.ShaderCache.ShaderCacheStateChanged += value;
+            remove => Methods.ShaderCache.ShaderCacheStateChanged -= value;
+        }
+
+        /// <summary>
+        /// Passthrough for ShaderCache.ShaderCacheProgressChanged event
```
gdkchan, line 92: Same comment as above.

Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -314,7 +329,33 @@ internal void Initialize()
...
+        /// <summary>
+        /// Raises ShaderCacheProgressChanged events periodically.
+        /// </summary>
+        private void ProgressLogger(object state)
+        {
+            const int refreshRate = 100; // ms
+
+            int totalCount = (int)state;
+
+            try
```
gdkchan, line 348: The same comments I made for PTC events also applies here.

ARMeilleure/Translation/PTC/Ptc.cs
mageven: Thanks for catching! `myEvent += myHandler` works without needing a null check and this inconsistency bothers me.

ARMeilleure/Translation/PTC/Ptc.cs
mageven: I considered progress reporting to be non-critical. If I don't catch it, it will reach `AppDomain.UnhandledException` and halt the emulator. If you prefer it that way, I can change it.

gdkchan: Yes, that's what I prefer. The event handles shouldn't throw, and if it does, then we should fix it. Btw, I think if the handler throws, then the other listeners won't be called.

mageven: Done. Rebased and removed commit with the change.

gdkchan (approved):
lgtm, thanks.

LDj3SNuD (approved):
lgtm, ty

AcK77 (approved):

PR #2058, opened 2021-02-25
Fix "Hide Cursor on Idle" inefficiency

mageven:
Both `DateTime.Now` and `ReactiveObject.get` (due to RW lock) are pretty heavy calls and are called from a sensitive location `UpdateFrame()` (runs every ~1ms and is responsible for handling input) which is already in a bad enough state.

There's loads of ways to combat that. With this PR, I have simplified the logic and moved the checks to `HandleScreenState()` which runs in the GUI thread.

### DIFF
--- a/Ryujinx/Ui/GLRenderer.cs
+++ b/Ryujinx/Ui/GLRenderer.cs
@@ -5,20 +5,23 @@ using OpenTK;
 using OpenTK.Graphics;
 using OpenTK.Graphics.OpenGL;
 using OpenTK.Input;
+using Ryujinx.Common;
 using Ryujinx.Common.Configuration;
 using Ryujinx.Common.Configuration.Hid;
 using Ryujinx.Configuration;
 using Ryujinx.Graphics.OpenGL;
-using Ryujinx.HLE;
 using Ryujinx.HLE.HOS.Services.Hid;
 using Ryujinx.Modules.Motion;
 using Ryujinx.Ui.Widgets;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading;
 
 namespace Ryujinx.Ui
 {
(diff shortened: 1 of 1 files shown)


Thog (approved):
lgtm

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -65,7 +66,11 @@ static GlRenderer()
 
         private readonly ManualResetEvent _exitEvent;
         
-        private Gdk.Cursor _invisibleCursor = new Gdk.Cursor (Gdk.Display.Default, Gdk.CursorType.BlankCursor);
+        // Hide Cursor
+        const int CursorHideIdleTime = 8; //seconds
```
AcK77, line 70:
```suggestion
        const int CursorHideIdleTime = 8; // seconds
```

Ryujinx/Ui/GLRenderer.cs
```diff
@@ -101,6 +106,27 @@ public GlRenderer(Switch device, GraphicsDebugLevel glLogLevel)
             _glLogLevel = glLogLevel;
 
             _exitEvent = new ManualResetEvent(false);
+
+            _hideCursorOnIdle = ConfigurationState.Instance.HideCursorOnIdle;
+            _lastCursorMoveTime = Stopwatch.GetTimestamp();
+            ConfigurationState.Instance.HideCursorOnIdle.Event += HideCursorStateChanged;
```
AcK77, line 112: Add a new empty line before

AcK77 (changes requested):

riperiperi (approved):
LGTM, with agreement with Ac_K's comments.

I didn't review the original, but I did see this code after it merged and had similar concerns. Good to see it fixed up.

AcK77 (approved):

PR #2070, opened 2021-02-28
Dummy CI test for artifacts

Thog:
Please ignore.


Thog:
Good bot, everything working as expected 👍

PR #2073, opened 2021-03-01
Fix regression in SignalMemoryTracking

Caian:
This PR solves the regression introduced by #2044. As mentioned in the PR, altering the value of `va` breaks the `Tracking.VirtualMemoryEvent` call.

This PR was tested with Monster Hunter Rise DEMO as it is known to manifest the regression.

I also looked for similar regressions in `MemoryTracking` and `AddressSpaceManager`.

### DIFF
--- a/Ryujinx.Cpu/MemoryManager.cs
+++ b/Ryujinx.Cpu/MemoryManager.cs
@@ -684,7 +684,7 @@ namespace Ryujinx.Cpu
             // Write tag includes read protection, since we don't have any read actions that aren't performed before write too.
             long tag = (write ? 3L : 1L) << PointerTagBit;
 
-            int pages = GetPagesCount(va, (uint)size, out va);
+            int pages = GetPagesCount(va, (uint)size, out ulong _);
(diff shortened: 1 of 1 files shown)


Ryujinx.Cpu/MemoryManager.cs
```diff
@@ -684,7 +684,7 @@ public void SignalMemoryTracking(ulong va, ulong size, bool write)
             // Write tag includes read protection, since we don't have any read actions that aren't performed before write too.
             long tag = (write ? 3L : 1L) << PointerTagBit;
 
-            int pages = GetPagesCount(va, (uint)size, out va);
+            int pages = GetPagesCount(va, (uint)size, out ulong _);
```
gdkchan, line 687:
```suggestion
            int pages = GetPagesCount(va, (uint)size, out _);
```
No need to specify the type if the value will be discarded.

riperiperi (approved):

Thog (approved):

PR #1850, opened 2020-12-31
Prevent Windows Display Sleep

EliEron:
This PR prevents Windows from turning off the display while a game is running. 

I also looked into implementing this for Linux but determined it to be too complex, as it requires D-Bus communication, and it would have been hard for me to test since I don't currently have any Linux test environment setup.

The placement and use of the class is based on the WindowsMultimediaTimerResolution class, as that is also a Windows exclusive feature that calls native code. Windows itself keeps tracks of threads that are preventing the display from sleeping and resumes normal behavior when they exit, so it was not necessary to create a method to manually enable the display sleep.

I happened to noticed some minor errors in the GTK Glade file while I was looking at it (One wrong and one missing tooltip) which I decided I might as well fix.

I am pretty new to Git and code contributions in general so I apologize if I have made any mistakes, and I welcome any feedback.

This PR partially fixes issue #1829, I say partially because it only implements it for Windows, and not for Linux And macOS.

PR #2080, opened 2021-03-05
Prevent Display Sleep on Linux

pineappleEA:
#1850 addressed the issue for Windows users, this one aims to do so for Linux ones.

While the Windows code runs only once, in the MainWindow, this isn't possible with X11 (I know, I hate Xorg too). Instead, we have to reset the counter constantly so I do it on the UpdateFrame loop.

It's important to note that IntPtr.Zero makes libX11 look at the DISPLAY environmental variable for the display (if you only have one screen it's :0).

### DIFF
--- a/Ryujinx.Common/System/DisplaySleep.cs
+++ b/Ryujinx.Common/System/DisplaySleep.cs
@@ -26,7 +26,7 @@ namespace Ryujinx.Common.System
         static extern IntPtr XOpenDisplay(IntPtr display);
         
         [DllImport("libX11", EntryPoint = "XCloseDisplay")]
-        static extern IntPtr XCloseDisplay(IntPtr display);
+        static extern int XCloseDisplay(IntPtr display);
 
         static public void Prevent()
         {
@@ -48,7 +48,7 @@ namespace Ryujinx.Common.System
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
 			{
-                XCloseDisplay(IntPtr.Zero);
+                XCloseDisplay(_display);
 			}
         }
 
--- a/Ryujinx.Common/System/DisplaySleep.cs
+++ b/Ryujinx.Common/System/DisplaySleep.cs
@@ -6,9 +6,9 @@ namespace Ryujinx.Common.System
 {
     public class DisplaySleep
     {
-        private static Timer _resetTimer;
+        static private Timer _resetTimer;
 
-        private static IntPtr _display;
+        static private IntPtr _display;
 
         [Flags]
         enum EXECUTION_STATE : uint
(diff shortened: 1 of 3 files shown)


CeruleanSky:
Any reason for not using a timer of a few hundred milliseconds (System.Threading.Timer , System.Timers.Timer, or maybe there is a better one) and not calling this from MainWindow.cs around where the windows one is called or another place?

I saw [log message on a recent pull request](https://github.com/Ryujinx/Ryujinx/pull/2058#issue-579870534) that the update frame code is pretty abused as is, and it might be better to keep more things out of there when it can be done.

pineappleEA:
The PR is open for review again!

Radically changed the way it works, Instead of Opening a Display , sending the Reset signal every frame on UpdateFrame() of GLRenderer, a function that starts the timer on MainWindow starts and stops at the same time with the Windows functions.  The timer Opens a Display, sends the Reset signal and Closes the Display every 25s. Since some laptops may dim the screen after 30s of idle, 25s feels about right.

gdkchan:
Is that really how applications usually disable the screen saver on linux? It looks very hacky with the call every 25 seconds.

jms-c:
What about XScreenSaverSuspend?
That would stop the screensaver too. (however, this solution is also quite popular it seems)

pineappleEA:
I've attempted every possible solution. The best solution is to send a DBus message according to freedesktop, sadly GNOME doesn't respect that and GNOME has the biggest share when it comes to desktop environments. Indeed it is hacky and I despise the solution, Xorg indeed is at a terrible state. Regarding what other software does, MPV sends the reset signal every 10s and bsnes a fake keypress every 20 (which is the same as sending a reset signal I guess).

AcK77 (approved):

Xpl0itR (changes requested):
As previously discussed on discord, using a timer to invoke a dll every 25 seconds isn't a good idea. You should instead use dbus, `org.gnome.SessionManager.Inhibit` on gnome and `org.freedesktop.ScreenSaver.Inhibit` elsewhere.

pineappleEA:
I believe treating each compositor differently is a worse solution. For example, during my test I found out dbus is broken on kwin too, it expects qdbus.

Alkaid-Benetnash:
I currently use `https://github.com/FeralInteractive/gamemode` to deal with screensaver on Gnome 3.38.
According to [their source code](https://github.com/FeralInteractive/gamemode/blob/b11d2912e280acb87d9ad114d6c7cd8846c4ef02/daemon/gamemode-dbus.c#L711), gamemode seems to only use the freedesktop dbus interface.
Not sure why @pineappleEA found gnome not respecting that API.

pineappleEA:
@Alkaid-Benetnash I tested gamemoderun on KDE and indeed prevents sleep. It's nice to hear GNOME works with `org.freedesktop.ScreenSaver.Inhibit` and doesn't require `org.gnome.SessionManager.Inhibit`. I'll take a closer look at what gamemode does and if that's the case someone should tell BSNES and MPV devs too since they do similar calls every some seconds

AcK77:
Closed due to inactivity and a hacky way to handle things. Feel free to fix things and reopen it later! Thanks.

Issue #1996, opened 2021-02-04
Smash Ultimate reproducible crash


ravenCondol:
### What's the issue you encountered?

Had 4 CPU's fighting each other in a Best of 5, random-stage gamerule set (to generate shader cache), but it always crashes when trying to return to main menu after set finish.

Only setting changed was enabling docked mode.

### How can the issue be reproduced?

Not sure if all this is necessary, but all I did was have 4 CPUs duke it out using following ruleset:

![Basic rule-set tab](https://puu.sh/Hdsqq/94d05e83a6.png)
![Advanced rule-set tab](https://puu.sh/HdsqV/f212398d99.png)

Then after one of them wins it crashes.

### Log file

Using LDN2.1 build:
[Ryujinx_1.0.0-ldn2.1_2021-02-03_17-45-21.log](https://github.com/Ryujinx/Ryujinx/files/5922890/Ryujinx_1.0.0-ldn2.1_2021-02-03_17-45-21.log)

Using latest (1.0.6471):
[Ryujinx_1.0.6471_2021-02-03_18-12-27.log](https://github.com/Ryujinx/Ryujinx/files/5922885/Ryujinx_1.0.6471_2021-02-03_18-12-27.log)

### Environment?

- Ryujinx version: Both **1.0.0-ldn2.1** and **1.0.6471**
- Game version: 10.1.0
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 5 **5600x** with base clock frequencies of 3700 Mhz, boost of up to 4700Mhz
  - GPU: NVIDIA RTX **3070** 8GB GDDR6 VRAM
  - RAM: 32GB of CL16 3600MHz G.SKILL **Ripjaws** V Series
- Applied Mods : None

### Additional context?

No, that is all.

EmulationFanatic:
I was able to reproduce this crash using the above settings. 

```
00:15:05.986 |E| GUI.RenderLoop Cpu InvalidAccessHandler: Invalid memory access at virtual address 0x00000000E32E0000.
00:15:05.998 |E| GUI.RenderLoop Application : Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: Attempted to access an invalid memory region.
   at Ryujinx.Memory.MemoryBlock.ThrowInvalidMemoryRegionException() in C:\projects\ryujinx\Ryujinx.Memory\MemoryBlock.cs:line 292
[...]

Issue #1871, opened 2021-01-07
Contra: Rogue Corps regression: crashes before getting in-game now


EmulationFanatic:
### What's the issue you encountered?

[Contra: Rogue Corps](https://github.com/Ryujinx/Ryujinx-Games-List/issues/1729) used to be playable, but now crashes during/after intros before going in-game.

### How can the issue be reproduced?

Use any version of the emulator after version 1.0.5525 and boot the game; the range-tracking update ([PR #1272](https://github.com/Ryujinx/Ryujinx/pull/1272)/1.0.5543) was the cause of this regression. 

### Log file

[CRC.log](https://github.com/Ryujinx/Ryujinx/files/5781415/CRC.log)

### Environment?

- Ryujinx version: 1.0.6235
- Game version: 1.0.0
- System Specs:
  - OS: Windows 10 20H2
  - CPU: Ryzen 7 5800X
  - GPU: NVIDIA GTX 1080
  - RAM: 32GB
- Applied Mods : No

### Additional context?

This issue was largely tracked down by Github user [@Johnwrnd](https://github.com/Johnwrnd)

gdkchan:
Should be fixed by #2083.

PR #2083, opened 2021-03-06
Improve handling for unmapped GPU resources

riperiperi:
- Fixed a memory tracking bug that would set protection on empty PTEs. (causes some buffer tracking crashes)
- When a texture's memory is (partially) unmapped, all pool references are forcibly removed and the texture must be rediscovered to draw with it. This will also force the texture discovery to always compare the texture's range for a match.
- RegionHandles now know if they are unmapped, and automatically unset their dirty flag when unmapped.
- Partial texture sync now loads only the region of texture that has been modified. Unmapped memory tracking handles cause dirty flags for a texture group handle to be ignored.

Buffers can try to sync memory that's unmapped if the size is too large, which it usually is for UE4 games, as it seems to pass in a large uniform buffer size no matter what is actually in the slot. These fixes prevent that from happening.

Textures can try to sync memory that's unmapped if a smaller view is created within a larger texture, but part of that larger texture has been unmapped. This can also happen due to GPU VA for a texture being mapped to something else, but its pool entry remaining unchanged. It's important that we do partial sync only with the data that we know is available to avoid an invalid memory access.

This greatly improves the emulator's stability for newer UE4 games, such as Bravely Default II. Does not fix _all_ crashes, and in BD2 there are still a number of graphical issues such as vertex explosions, missing facial animations and missing shadows.

Fixes texture swaps in Bravely Default II. This is specifically due to the change that removes textures from pools on unmap, and forcing compare range on textures where the gpu va has been remapped.

May improve the situation for other games that crash due to buffer sync. Not animal crossing.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
@@ -13,7 +13,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     class TexturePool : Pool<Texture, TextureDescriptor>
     {
         private int _sequenceNumber;
-        ConcurrentQueue<Texture> _dereferenceQueue = new ConcurrentQueue<Texture>();
+        private readonly ConcurrentQueue<Texture> _dereferenceQueue = new ConcurrentQueue<Texture>();
 
         /// <summary>
         /// Intrusive linked list node used on the texture pool cache.
--- a/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
@@ -2,6 +2,7 @@ using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Texture;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Ryujinx.Graphics.Gpu.Image
@@ -12,6 +13,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     class TexturePool : Pool<Texture, TextureDescriptor>
     {
         private int _sequenceNumber;
+        ConcurrentQueue<Texture> _dereferenceQueue = new ConcurrentQueue<Texture>();
 
         /// <summary>
         /// Intrusive linked list node used on the texture pool cache.
@@ -53,6 +55,8 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 TextureInfo info = GetInfo(descriptor, out int layerSize);
 
+                ProcessDereferenceQueue();
+
                 texture = Context.Methods.TextureManager.FindOrCreateTexture(TextureSearchFlags.ForSampler, info, layerSize);
 
                 // If this happens, then the texture address is invalid, we can't add it to the cache.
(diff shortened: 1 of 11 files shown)


gdkchan:
This should improve stability on Super Smash Bros Ultimate. Anyone testing those changes, please report your findings on that game if you try it (and whenever you had crashes before this change).

Ryujinx.Memory/Range/MultiRange.cs
```diff
@@ -75,6 +76,51 @@ public MultiRange(MemoryRange[] ranges)
             }
         }
 
+        /// <summary>
+        /// Gets a slice of the multi-range.
+        /// </summary>
+        /// <param name="offset">Offset of the slice into the multi-range in bytes</param>
+        /// <param name="size">Size of the slice in bytes</param>
+        /// <returns>A new multirange representing the given slice of this one</returns>
```
gdkchan, line 84:
```suggestion
        /// <returns>A new multi-range representing the given slice of this one</returns>
```
To match the summary.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -12,6 +13,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     class TexturePool : Pool<Texture, TextureDescriptor>
     {
         private int _sequenceNumber;
+        ConcurrentQueue<Texture> _dereferenceQueue = new ConcurrentQueue<Texture>();
```
gdkchan, line 16:
```suggestion
        private readonly ConcurrentQueue<Texture> _dereferenceQueue = new ConcurrentQueue<Texture>();
```
For consistency.

Ryujinx.Memory/Range/MultiRange.cs
```diff
@@ -75,6 +76,51 @@ public MultiRange(MemoryRange[] ranges)
...
+                        ranges.Add(new MemoryRange(range.Address, Math.Min(size, range.Size)));
+                        size -= range.Size;
+                    }
+                    else if (offset < range.Size)
+                    {
+                        ranges.Add(new MemoryRange(range.Address + offset, Math.Min(size, range.Size - offset)));
+                        size -= range.Size;
+                    }
+
+                    if (size <= 0)
```
gdkchan, line 112: Why not just `size == 0`? It can't be less than 0 since the value is unsigned.

gdkchan: I guess this is supposed to have a cast to long since `size -= range.Size;` can make the value "negative".

gdkchan:
Should fix #1871.

Shahil-Ayato:
This should fix #1996

Doctorwho1909:
Smash doesnt crash on classic now

Shahil-Ayato:
After a bunch of testing with smash ultimate, this greatly improves the stability of the game and fixes a lot of crashes.

Ryujinx.Memory/Range/MultiRange.cs
```diff
@@ -75,6 +76,52 @@ public MultiRange(MemoryRange[] ranges)
...
+                {
+                    throw new ArgumentOutOfRangeException(nameof(size));
+                }
+
+                return new MultiRange(_singleRange.Address + offset, size);
+            }
+            else
+            {
+                var ranges = new List<MemoryRange>();
+                foreach (MemoryRange range in _ranges)
```
Thog, line 99: Little nit

```suggestion

                foreach (MemoryRange range in _ranges)
```

Thog (approved):
lgtm

gdkchan (approved):
lgtm, thanks. Nice to see those intermittent buffer crashes fixed.

PR #2085, opened 2021-03-06
Remove unused physical region tracking

riperiperi:
(based on top of https://github.com/Ryujinx/Ryujinx/pull/2083, as some of the same functions are changed)

Memory tracking initially supported tracking physical regions of memory, as a potential goal was to replace the memory tracking part of software address translation, and rely on host memory protection of the switch "physical memory" to signal tracking when needed. This would have simplified memory instructions in the JIT, but it turned out that OS memory protection and signals cost a lot more than just doing the protection in software.

Virtual Regions created within the tracking structure would also maintain a collection of Physical Regions that make it up. These are calculated at creation, and recalculated any time the virtual region is mapped or unmapped. This physical region recalculation is quite expensive, and becomes a large problem when the guest decides to map/unmap cpu va at a very granular level.

UE4 games are a good example at games which constantly map and unmap memory that is used for GPU resources. Unfortunately, it also does this at quite a granular scale (occasionally single 4096 byte page at a time, covering a very large resource), so the physical regions ended up being recalculated very often, which would cause much longer loading times and ingame stutters. 

This PR removes physical region tracking entirely, as it is not used and can cause a lot of unnecessary work to be done. This will speed up memory unmap/remap of virtual addresses that were once used for GPU resources, as well as reduce the cost of creating new tracking handles.

This also changes the order in which unmapping events are called. Tracking regions/handles are now notified about an unmap _before_ they are actually unmapped, which could reduce races where unmapped memory still has a dirty flag.

Note: Physical region tracking could still be useful for multi process shared memory, though if you know what you're doing, you could potentially track the VA mapping in each address space and link them together, similar to what service calls are doing right now. Would probably save a few headaches rather than trying to reintroduce and speed up the physical memory calculation.


gdkchan (approved):
lgtm, thanks!

Thog (approved):
lgtm 👍

PR #2087, opened 2021-03-07
Flip component mask if target is BGRA.

riperiperi:
For OpenGL, we use RGBA textures as BGRA and swap the components on draw, write, copy etc. However, we forgot to swap the component mask, so when a texture meant to draw to only "blue", it ended up drawing to only "red" and vice versa.

On its own, this doesn't do much as far as I know. Combined with better bindless texture handling, it will lead to fixing shadows in BD2's battles.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -862,7 +862,14 @@ namespace Ryujinx.Graphics.OpenGL
 
                 _framebuffer.AttachColor(index, color);
 
-                _fpIsBgra[index] = color != null && color.Format.IsBgra8() ? 1 : 0;
+                int isBgra = color != null && color.Format.IsBgra8() ? 1 : 0;
+
+                if (_fpIsBgra[index] != isBgra)
+                {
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1233,11 +1240,15 @@ private void PostDraw()
 
         private void RestoreComponentMask(int index)
         {
+            // If the bound render target is bgra, swap the red and blue masks.
+            bool redMask = _fpIsBgra[index] == 0 ? (_componentMasks[index] & 1u) != 0 : (_componentMasks[index] & 4u) != 0;
```
riperiperi, line 1244: Maybe just `bool redMask =  (_componentMasks[index] & (_fpIsBgra[index] == 0 ? 1u : 4u)) != 0;`?

gdkchan (approved):
lgtm, nice catch!

riperiperi:
Switched up the code to make it look a little nicer. Should be functionally identical.

When bindless accesses are working:
image

jduncanator (approved):

PR #2088, opened 2021-03-08
Improve Buffer Textures and flush Image Stores 

riperiperi:
Fixes a number of issues with buffer textures:

- Reworked Buffer Textures to create their buffers in the TextureManager, then bind them with the BufferManager later.
  - Fixes an issue where a buffer texture's buffer could be invalidated after it is bound, but before use.
- Fixed width unpacking for large buffer textures. The width is now 32-bit rather than 16.
- Force buffer textures to be rebound whenever any buffer is created, as using the handle id wasn't reliable, and the cost of binding isn't too high.

Fixes vertex explosions and flickering animations in UE4 games. May affect games that use ImageStore, as now those textures are flushed on read/write.

Before:
image
image
image
(maybe one day I'll open the game on master to get a better comparison shot for this, but you get the gist)

After:
image
image

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Image/TextureBuffer.cs
+++ b/Ryujinx.Graphics.OpenGL/Image/TextureBuffer.cs
@@ -6,12 +6,17 @@ namespace Ryujinx.Graphics.OpenGL.Image
 {
     class TextureBuffer : TextureBase, ITexture
     {
+        private Renderer _renderer;
         private int _bufferOffset;
         private int _bufferSize;
+        private int _bufferCount;
 
         private BufferHandle _buffer;
 
-        public TextureBuffer(TextureCreateInfo info) : base(info) {}
+        public TextureBuffer(Renderer renderer, TextureCreateInfo info) : base(info)
+        {
+            _renderer = renderer;
+        }
 
         public void CopyTo(ITexture destination, int firstLayer, int firstLevel)
         {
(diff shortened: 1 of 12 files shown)


Thog (approved):
lgtm

Ryujinx.Graphics.OpenGL/Image/TextureBuffer.cs
```diff
@@ -50,16 +55,16 @@ public void SetData(ReadOnlySpan<byte> data, int layer, int level)
 
         public void SetStorage(BufferRange buffer)
         {
-            if (buffer.Handle == _buffer &&
-                buffer.Offset == _bufferOffset &&
-                buffer.Size == _bufferSize)
+            if (_buffer != BufferHandle.Null && _renderer.BufferCount == _bufferCount)
```
gdkchan, line 58: Why is it not comparing the offsets and size anymore? It should rebind aswell if the buffer is the same, but the range changed.

riperiperi: I can add that, but it's not possible for the size to change, and the buffer offset changing is implied by a buffer combine, which will increment the buffer count.

gdkchan: I was talking from the API perspective, someone calling the function would not expect it to not update the binding if the offset or size changed, which could lead to someone inadvertently inserting bugs on the code later on if they need to change something related to buffer textures.

gdkchan (approved):
lgtm, thanks. Nice catch!

PR #2089, opened 2021-03-08
Traverse PhiNodes for Bindless Elimination

riperiperi:
This allows bindless handles to be found for image/texture instructions with predicates, when the assignment of the texture handle is within the same predicate. This is a little rough, since it's made to target a specific pattern.

If the handle source is a PhiNode rather than an Operation, we try to find which phi source is relevant to the instruction. This source is the latest source block that has a branching condition matching the texture access block.

This seems to cover the remaining bindless handles that compilers seem to be creating as an optimization. Note that this is more likely to make a shader create too many samplers, and fail to link.

The only way to solve the issue with creating too many samplers, is either to play bindless textures straight and support bindless on them, or to emulate more then 32 samplers with host bindless texture access of our own design. This isn't in the scope of this PR.

Will affect newer UE4 games, and games by NdCube (Super Mario Party, Clubhouse Games)

## Before:
image
(note, shadows and eyes. this screenshot is at night, so the overall brightness shouldn't be compared)

## After:
image

### DIFF
--- a/Ryujinx.Graphics.Shader/Translation/Optimizations/BindlessElimination.cs
+++ b/Ryujinx.Graphics.Shader/Translation/Optimizations/BindlessElimination.cs
@@ -5,6 +5,66 @@ namespace Ryujinx.Graphics.Shader.Translation.Optimizations
 {
     class BindlessElimination
     {
+        private static Operation FindBranchSource(BasicBlock block)
+        {
+            foreach (BasicBlock sourceBlock in block.Predecessors)
+            {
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
lgtm, nice way to fix this, I tried to fix it before by merging if blocks, but did not think about just extending the elimination pass to catch more cases for some reason. I'm just a bit worried that this might cause regressions due to the higher amount of samplers that might be produced (and cause the shader to fail to compile if > 32), but that was technically already a issue before.

Thog (approved):

PR #2091, opened 2021-03-09
Fix lineSize for LinearStrided -> Linear conversion

riperiperi:
Fixes a possible crash when width is greater than stride, which can happen due to alignment when copying textures with the copy buffer method.

Fixes `System.ArgumentOutOfRangeException` in `ConvertLinearStridedToLinear` called from `CopyBuffer`. May fix another issue when outStride < stride.


Thog (approved):

gdkchan (approved):
lgtm, thanks.

Issue #2045, opened 2021-02-21
Bindless textures break the shader cache


EmulationFanatic:
### What's the issue you encountered?

Any game with bindless textures will render the shader cache unusable for that game. The only solution is purging the shader cache prior to booting the game each time, or disabling shader cache in the emulator options before launching the game.

### How can the issue be reproduced?

Play a game with bindless textures (Super Mario Party is one example), then exit emulation/close the emulator and launch the game again.

### Log file

[Ryujinx_1.0.6662_2021-02-21_08-46-07.log](https://github.com/Ryujinx/Ryujinx/files/6018003/Ryujinx_1.0.6662_2021-02-21_08-46-07.log)

### Environment?

- Ryujinx version: 1.0.6662
- Game version: N/A
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 7 5800X
  - GPU: GTX 1080
  - RAM: 32GB
- Applied Mods : No

Shahil-Ayato:
Could this possibly be fixed with [#2005](https://github.com/Ryujinx/Ryujinx/pull/2005) ?

EmulationFanatic:
> Could this possibly be fixed with [#2005](https://github.com/Ryujinx/Ryujinx/pull/2005) ?

According to gdk, no. There will still need to be a fix like a blacklist preventing bindless textures from entering the shader cache or some other solution.

PR #2097, opened 2021-03-10
Salieri: Detect and avoid caching shaders using bindless textures

Thog:
Currently the shader cache doesn't have the right format to support
bindless textures correctly and may cache shaders that it cannot rebuild
after host invalidation.

This PR address the issue by blacklisting shaders using bindless
textures.

PS: It is planned to readd support for those shaders in the cache later.


Shahil-Ayato:
Going to assume that this should fix #2045

EmulationFanatic:
> Going to assume that this should fix #2045

It does, but there will need to be a separate issue opened for implementing proper support for bindless in the shader cache.

Doctorwho1909:
image
it works very well

AcK77 (approved):

riperiperi (approved):

PR #2099, opened 2021-03-12
Improve linear texture compatibility rules

riperiperi:
Fixes an issue where small or width-aligned (rather than byte aligned) textures would fail to create a view of existing data. Creates a copy dependency as size change may be risky.

- View layout compatibility is now determined by the stride shifted by the level, rather than a stride caculated from the level's width.
- Linear textures are considered copy compatible when they have matching stride.

Caused missing lens flare in Mario Kart 8 Deluxe and Splatoon 2.

Fixes an issue where Cross Code would display a black screen.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Constants.cs
+++ b/Ryujinx.Graphics.Gpu/Constants.cs
@@ -69,5 +69,15 @@ namespace Ryujinx.Graphics.Gpu
         /// Maximum size of gl_ClipDistance array in shaders.
         /// </summary>
         public const int TotalClipDistances = 8;
+
+        /// <summary>
+        /// Byte alignment for texture stride.
+        /// </summary>
+        public const int StrideAlignment = 32;
+
+        /// <summary>
+        /// Byte alignment for block linear textures
+        /// </summary>
+        public const int GobAlignment = 64;
     }
 }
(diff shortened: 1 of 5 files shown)


Doctorwho1909:
image
that sunset looking very beautiful !

riperiperi:
Need to change some size change rules - Cross Code is now visible but ends up crashing later into the run.

michaelpegaso:
It is a great advance, it looks great with the reflections of the sun, in splatoon 2 you see the reflections of but at the same time you see 2 transparent circles that should not be there. I'm trying to upload a photo for you to see.

[https://ibb.co/jw60qJz](url)

riperiperi:
2021031302335800_s
The transparent circles are meant to be there, it's part of the lens flare. Here's a screen capture from the switch (handheld mode)

michaelpegaso:
> 2021031302335800_s
> Los círculos transparentes están destinados a estar ahí, es parte del destello de la lente. Aquí hay una captura de pantalla del interruptor (modo portátil)

Everything cleared up. Thank you very much for answering and for the hard work of improving the emulator. Greetings.

riperiperi:
Removed the size change when creating a copy dependency, as the copy to the target (potentially different sized) texture can properly deal with cropping by itself. There are also a few other guarantees unrelated to this PR that help this:

- Data is only copied from another texture if the data has not been modified on CPU since the last draw to it. This means that rather than copying from a poorly cropped texture and losing data, we will be reading uncropped data from memory.
- After a size change when getting texture for sampler, the size is changed _before_ synchronizing the data. This means that the new size will sync all the data in the new size, rather than sync cropped and scale up.

gdkchan (approved):
lgtm, thanks! It would be nice to add a constant for the "32" constant alignment value, to make the code more clear.

Thog (approved):
Seems fine to me 👍

PR #2101, opened 2021-03-13
Override openal-soft configuration

Thog:
This enforce speaker mode to avoid weird audio quality reduction with
output considered as headphones by Windows (It applies a HRTF or crossfeed filter supposed to improve audio quality, might be a bug on their end)


AcK77 (approved):

jms-c (approved):
This looks like a simple fix, I approve.
(even if I don't have write-access)

jduncanator (approved):

PR #2105, opened 2021-03-14
Remove ignore service hack

Thog:
This have been around and doesn't make sense much those days.

Additionally, people weren't reporting missing services anymore causing
more trouble for the team.

This PR entirely remove the hack feature.


gdkchan:
Imo it's better to keep the option, at least for now. There's a number of games that needs it enabled to work, especially on the LDN build that is not often updated. This also helps making newer games bootable. If the problem is just the lack of missing reports, it can be solved using other methods (like telemetry). This is a hack, but its a pretty harmless one since it can't possibly possibly make anything run worse than it already does, in the worst case it will replace a early crash with a crash or instability at a later time.

amayra:
this emulator is an experimental Switch emulator so this hack is not good for future development plus people who complain about this RP are the same people who don't report anything and come for a FREE game only

Shahil-Ayato:
> this emulator is an experimental Switch emulator so this hack is not good for future development plus people who complain about this RP are the same people who don't report anything and come for a FREE game only

I think your missing the point here, this is good for the short term for a lot of reasons, one this can get games booting day 1 and help a lot with new games so the end user doesn’t have to wait for the services to be implemented and it can also generate some decent publicity for Ryujinx. Two this can greatly benefit the end user, a good example was when MH rise came out people came to the discord to report crashes due to the missing services and not a lot of people have this hack on as it’s disabled by default, looking at the LDN build especially since it’s not updated regularly it would be awful to remove this hack since game updates need this to boot, in the future when major game releases stop coming out I can see this hack going away but for now I’d prefer this hack to stay as it’s very convenient for the end user.

choppymaster:
Oh, I think this is still good that this option is there. Otherwise, many games will encounter crashes and makes some games unplayable.
Nope, people will, and is reporting missing services.  Some of them are not implemented. Hope for more reports.

Ghostotakoi:
Is OK if you want to remove it, IMO at least should be keeped as a togglable option, you could hide it under developer mode or something if you don't want that people turn it on unintentionally, but I think it should stay for a while.

doctorwhosthat:
I honestly don't know whether people are actually missing the point to this...

Though, what exactly does this achieve in any way and what benefit does it have on the end-user?

Doctorwho1909:
> I honestly don't know whether people are actually missing the point to this...
> 
> Though, just what exactly will this actually achieve and how is it likely to effect the end user?
> 
> If that actually made no sense, then I do apologise.

Some games that are played by alot of users well use ignored services and even ldn needs it still and smash also and missing services are reported even with it so thats not a reason to get rid of it as goffy and sera and many others users report those all the time. Just let it stay hidden not gone

jduncanator:
> Imo it's better to keep the option, at least for now. There's a number of games that needs it enabled to work, especially on the LDN build that is not often updated. This also helps making newer games bootable. If the problem is just the lack of missing reports, it can be solved using other methods (like telemetry). This is a hack, but its a pretty harmless one since it can't possibly possibly make anything run worse than it already does, in the worst case it will replace a early crash with a crash or instability at a later time.

I agree with gdkchan here, I'd prefer leaving this in place, and letting those that do gam testing to pick up missing services. We have a game database that is full of reports of missing services, and ongoing testing is performed by the community, so I don't think it's entirely accurate that "no one reports them anymore".

I agree that in the future, it would be ideal to remove this functionality, however I'm failing to see the benefits of removing this today.

gdkchan:
Btw, one of the reasons this option was added was to easily log all missing service functions, which allows implementing them all at once, rather than crashing, seeing the missing service function, implementing/stubbing, and then relaunching and repeating the process over and over until everything is implemented, which may take a lot longer.

Mou-Ikkai:
Thog: Haha, opinion go brrrrrrrrrr.
This entire github: *Everyone disliked that*

Mou-Ikkai:
But for all seriousness, I get what Thog is going for. It's not used THAT much. But for the people that do use it, it can seriously handicap them in their possibility to playing a game with an unimplemented service. Don't fight over this. It's a 50/50 on this change.

vickyorlo:
If the issue is that users do not report missing services enough, just have an obtrusive pop-up that shows up every time a service is ignored.

Then they will come and report them. A lot.

jms-c:
I agree to hiding the feature (or removing from gui and editing the settings json) but not completely removing it.

MelonSpeedruns:
Same here.

Th3d0nGsT3R:
The idea of removing it from the gui but leaving it implemented within the code for testing reasons seems like the best compromise. The reasons for which have already been highlighted by other people.

PR #2115, opened 2021-03-16
[Suggestion] Add base classes to simplify implementing proper disposable patterns.

hanabi1224:
Disposable pattern is the way MS officially recommends to implement IDisposable interface, and it will give some performance boost in certain cases.

e.g. In current MemoryBlock.cs a finalizer ```~MemoryBlock()``` is implemented, but ```GC.SuppressFinalize(this)``` is not called in ```Dispose()``` which actually leads to perf panelty of GC because GC will put the object in Finalize queue when finalizer is implemented.

> Empty finalizers should not be used. When a class contains a finalizer, an entry is created in the Finalize queue. When the finalizer is called, the garbage collector is invoked to process the queue. An empty finalizer just causes a needless loss of performance.

In this case, resources are manually released by calling ```dispose()```, no need to let GC call finalizer again.

The base classes can be utilized to simpliy the code to implement a proper disposable pattern.

references:
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors
http://joeduffyblog.com/2005/04/08/dg-update-dispose-finalization-and-resource-management/

### DIFF
--- a/Ryujinx.Common/DisposableBase.cs
+++ b/Ryujinx.Common/DisposableBase.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace Ryujinx
+namespace Ryujinx.Common
 {
     /// <summary>
     /// A base class to implement IDisposable pattern for classes with only managed resources.
--- a/Ryujinx.Common/DisposableBase.cs
+++ b/Ryujinx.Common/DisposableBase.cs
@@ -1,0 +1,41 @@
+﻿using System;
+
+namespace Ryujinx
+{
+    /// <summary>
+    /// A base class to implement IDisposable pattern for classes with only managed resources.
+    /// Use DisposableUnmanagedBase when there are unmanage resources to dispose.
+    /// </summary>
+    public abstract class DisposableBase : IDisposable
+    {
+        /// <summary>
+        /// A flag that indicates whether object has been disposed.
+        /// </summary>
+        public bool Disposed { get; private set; }
+
+        protected void Dispose(bool disposing)
+        {
+            if (Disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                DisposeManaged();
+            }
+
+            Disposed = true;
+        }
+
+        /// <summary>
+        /// Dispose managed resources.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(disposing: true);
+        }
+
+        protected abstract void DisposeManaged();
+    }
+}
--- a/Ryujinx.Common/DisposableBase.cs
+++ b/Ryujinx.Common/DisposableBase.cs
@@ -4,8 +4,10 @@ namespace Ryujinx
 {
     /// <summary>
     /// A base class to implement IDisposable pattern for classes with only managed resources.
-    /// Use DisposableUnmanagedBase when there are unmanage resources to dispose.
     /// </summary>
+    /// <remarks>
+    /// Use <see cref="DisposableUnmanagedBase"/> when there are unmanaged resources to dispose.
+    /// </remarks>
     public abstract class DisposableBase : IDisposable
     {
         /// <summary>
--- a/Ryujinx.Common/DisposableUnmanagedBase.cs
+++ b/Ryujinx.Common/DisposableUnmanagedBase.cs
@@ -1,0 +1,50 @@
+﻿using System;
+
+namespace Ryujinx
+{
+    /// <summary>
+    /// A base class to implement IDisposable pattern for classes with unmanaged resources.
+    /// Use DisposableBase when there are only manage resources to dispose.
+    /// </summary>
+    public abstract class DisposableUnmanagedBase : IDisposable
+    {
+        /// <summary>
+        /// A flag that indicates whether object has been disposed.
+        /// </summary>
+        public bool Disposed { get; private set; }
+
+        protected void Dispose(bool disposing)
+        {
+            if (Disposed)
+            {
+                return;
+            }
+
+            if (disposing)
+            {
+                DisposeManaged();
+            }
+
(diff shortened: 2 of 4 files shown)


saldabain (approved):

saldabain:
@gdkchan This looks interesting

Ryujinx.Common/DisposableBase.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+
+namespace Ryujinx
```
gdkchan, line 3: Imo the namespace should be `Ryujinx.Common` (same for `DisposableUnmanagedBase`).

Ryujinx.Common/DisposableBase.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// A base class to implement IDisposable pattern for classes with only managed resources.
+    /// Use DisposableUnmanagedBase when there are unmanage resources to dispose.
```
gdkchan, line 7: Maybe put that under `<remarks></remarks>`, with `<see cref="DisposableUnmanagedBase"/>`?

Ryujinx.Common/DisposableUnmanagedBase.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+
+namespace Ryujinx
+{
+    /// <summary>
+    /// A base class to implement IDisposable pattern for classes with unmanaged resources.
+    /// Use DisposableBase when there are only manage resources to dispose.
```
gdkchan, line 7:
```suggestion
    /// Use DisposableBase when there are only managed resources to dispose.
```
This also could be under "remarks".

Ryujinx.Common/DisposableBase.cs
hanabi1224: I manually changed it to Ryujinx because IDisposable is everywhere and it will make it much easier to switch later, otherwise every file needs ```using Ryujinx.Common```

Ryujinx.Common/DisposableBase.cs
hanabi1224: Sure

Ryujinx.Common/DisposableUnmanagedBase.cs
hanabi1224: Good catch

Ryujinx.Common/DisposableBase.cs
hanabi1224: Fixed

Thealexbarney:
Has there been any testing done on the performance impact of adding a finalizer to every `IDisposable`? And is the base meant to be thread-safe or not?
One downside to this is that the lack of multiple-inheritance in C# means any class using this can't inherit from anything else. Could a source generator maybe be a better option?

hanabi1224:
@Thealexbarney 
> Has there been any testing done on the performance impact of adding a finalizer to every ```IDisposable```?

Only ```DisposableUnmanagedBase``` implements a finalizer, with ```GC.SuppressFinilizer(this)``` in Dispose(). As long as ```Dispose()``` is called which is expected, no finalizer overhead is introduced.  The finalizer is a safeguard to ensure unmanaged resources are disposed even when people forget to call ```Dispose()```, which is not an expected case that finalizer would add performance overhead in. 

>  And is the base meant to be thread-safe or not?

Can you please elaborate on how this is relevant here? 

> One downside to this is that the lack of multiple-inheritance in C#

I looked at the code base, there're no classes with base that implements IDisposable.  Only some Disposable structs may not fit.

> Could a source generator maybe be a better option?

Visual studio 2019 has the option to generate IDisposable patterns already, the base classes are introduced to reduce code duplication. Personally, I don't think it's better.

Thealexbarney:
> Only `DisposableUnmanagedBase` implements a finalizer

Ah, I must have missed that somehow.

> As long as `Dispose()` is called which is expected, no finalizer overhead is introduced. The finalizer is a safeguard to ensure unmanaged resources are disposed even when people forget to call `Dispose()`, which is not an expected case that finalizer would add performance overhead in.

There's always overhead for using an object with a finalizer. For example allocation is usually an order of magnitude or 2 slower.

>> And is the base meant to be thread-safe or not?
>
> Can you please elaborate on how this is relevant here?

In case multiple places call `Dispose` at about the same time. You could argue that you've probably got other issues if that happens, but it's [easy enough to do](https://github.com/Thealexbarney/LibHac/blob/d2c327377d8e9d2d90b922f0c80beb5f582a61f7/src/LibHac/Fs/IStorage.cs#L147).

hanabi1224:
@Thealexbarney There's a similar discussion in this [post](http://joeduffyblog.com/2005/04/08/dg-update-dispose-finalization-and-resource-management/) that whether ```GC.SuppressFinalize(this)```  should be called before ```Dispose(true)``` or after. It suggests ```GC.SuppressFinalize(this)``` come later to make sure ```Dispose``` finishes successfully.

The implementation is IStorage.cs has a problem when ```Dispose``` finished with an error (and a retry might work) , however, the _disposedState is changed beforehand and there will never be a chance to retry (in the finializer)

So is it a valid scenario to call ```Dispose``` concurrently?  I don't think so, there're many .net analyzers that warn on ```undisposed disposables```, ```class that owns disposables should be disposable as well``` cases, and suggests good designs when IDisposable objects are not properly disposed by its owner (the warning rule is kind of similar to rust ownership maybe?) , which any concurrent call to ```Dispose``` would violate and get warned.

Thog:
Closing as stale.

PR #2117, opened 2021-03-17
[Perf] KTimeManager.UnscheduleFutureInvocation

hanabi1224:
Profiling shows that linq in KTimeManager.UnscheduleFutureInvocation is doing a lot of unnecessary allocations, this PR is an attempt to make some improvement.  Ideally, a PrioriyQueue seems to be the best fix but a hashmap would make an easy improvement without introducing 3rd packages or own PQ implementation.

Before:

After:

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -8,28 +8,15 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 {
     class KTimeManager : IDisposable
     {
-        private class WaitingObject
-        {
-            public IKFutureSchedulerObject Object { get; }
-            public long TimePoint { get; set; }
-
-            public WaitingObject(IKFutureSchedulerObject schedulerObj, long timePoint)
-            {
-                Object = schedulerObj;
-                TimePoint = timePoint;
-            }
-        }
-
         private readonly KernelContext _context;
         // TODO: PriorityQueue will have the best performance here.
-        private readonly IDictionary<int, WaitingObject> _waitingObjects;
+        private readonly HashSet<IKFutureSchedulerObject> _waitingObjects = new HashSet<IKFutureSchedulerObject>();
         private AutoResetEvent _waitEvent;
         private bool _keepRunning;
 
         public KTimeManager(KernelContext context)
         {
             _context = context;
-            _waitingObjects = new Dictionary<int, WaitingObject>();
             _keepRunning = true;
 
             Thread work = new Thread(WaitAndCheckScheduledObjects)
@@ -44,10 +31,9 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
         {
             long timePoint = PerformanceCounter.ElapsedMilliseconds + ConvertNanosecondsToMilliseconds(timeout);
 
-            var key = schedulerObj.GetHashCode();
             lock (_context.CriticalSection.Lock)
             {
-                if (_waitingObjects.TryGetValue(key, out var existing))
+                if (_waitingObjects.TryGetValue(schedulerObj, out var existing))
                 {
                     if (timePoint < existing.TimePoint)
                     {
@@ -56,7 +42,8 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
                 }
                 else
                 {
-                    _waitingObjects.Add(key, new WaitingObject(schedulerObj, timePoint));
+                    schedulerObj.TimePoint = timePoint;
+                    _waitingObjects.Add(schedulerObj);
                 }
             }
 
@@ -65,14 +52,12 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 
         public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
         {
-            var key = schedulerObj.GetHashCode();
-            if (_waitingObjects.ContainsKey(key))
+            lock (_context.CriticalSection.Lock)
             {
-                lock (_context.CriticalSection.Lock)
-                {
-                    // Call TimeUp here?
-                    _waitingObjects.Remove(key);
-                }
+                // Not calling TimeUp here, 
+                // TimeUp should only be called when the timeout expires,
+                // if it expires (the scheduler object may be signaled and remove itself from the list before that happens).
+                _waitingObjects.Remove(schedulerObj);
             }
         }
 
@@ -82,14 +67,13 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
             {
                 while (_keepRunning)
                 {
-                    KeyValuePair<int, WaitingObject> nextPair;
+                    IKFutureSchedulerObject next;
                     lock (_context.CriticalSection.Lock)
                     {
                         // TODO: PriorityQueue will has the best performance here.
-                        nextPair = _waitingObjects.OrderBy(x => x.Value.TimePoint).FirstOrDefault();
+                        next = _waitingObjects.OrderBy(x => x.TimePoint).FirstOrDefault();
                     }
 
-                    var next = nextPair.Value;
                     if (next != null)
                     {
                         long timePoint = PerformanceCounter.ElapsedMilliseconds;
@@ -98,11 +82,15 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
                             _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
                         }
 
-                        lock (_context.CriticalSection.Lock)
+                        var timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
+                        if (timeUp)
                         {
-                            if (_waitingObjects.Remove(nextPair.Key))
+                            lock (_context.CriticalSection.Lock)
                             {
-                                next.Object.TimeUp();
+                                if (_waitingObjects.Remove(next))
+                                {
+                                    next.TimeUp();
+                                }
                             }
                         }
                     }
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -11,24 +11,25 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
         private class WaitingObject
         {
             public IKFutureSchedulerObject Object { get; }
-            public long TimePoint { get; }
+            public long TimePoint { get; set; }
 
             public WaitingObject(IKFutureSchedulerObject schedulerObj, long timePoint)
             {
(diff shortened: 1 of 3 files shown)


Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -43,19 +44,35 @@ public void ScheduleFutureInvocation(IKFutureSchedulerObject schedulerObj, long
...
         public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
         {
-            lock (_context.CriticalSection.Lock)
+            var key = schedulerObj.GetHashCode();
+            if (_waitingObjects.ContainsKey(key))
             {
-                _waitingObjects.RemoveAll(x => x.Object == schedulerObj);
+                lock (_context.CriticalSection.Lock)
+                {
+                    // Call TimeUp here?
```
gdkchan, line 73: No, `TimeUp` should only be called when the timeout expires, *if* it expires (the scheduler object may be signaled and remove itself from the list before that happens).

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -43,19 +44,35 @@ public void ScheduleFutureInvocation(IKFutureSchedulerObject schedulerObj, long
...
             }
 
             _waitEvent.Set();
         }
 
         public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
         {
-            lock (_context.CriticalSection.Lock)
+            var key = schedulerObj.GetHashCode();
+            if (_waitingObjects.ContainsKey(key))
```
gdkchan, line 69: Why is this check necessary? `Remove` will just do nothing if the `key` doesn't exist anyway.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -65,32 +82,27 @@ private void WaitAndCheckScheduledObjects()
...
                     if (next != null)
                     {
                         long timePoint = PerformanceCounter.ElapsedMilliseconds;
-
                         if (next.TimePoint > timePoint)
                         {
                             _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
                         }
 
-                        bool timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
```
gdkchan, line 84: Why was this check removed?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -11,24 +11,25 @@ class KTimeManager : IDisposable
...
             {
-                Object    = schedulerObj;
+                Object = schedulerObj;
                 TimePoint = timePoint;
             }
         }
 
         private readonly KernelContext _context;
-        private readonly List<WaitingObject> _waitingObjects;
+        // TODO: PriorityQueue will have the best performance here.
```
gdkchan, line 24: FWIW a `PriorityQueue` was (or will, its still in preview) added on .NET 6. Anyway, I think a `HashSet` might be a better alternative than a `Dictionary`, since it seems that the key is just the object hash code anyway.

riperiperi: `HashSet` will also have proper handling for hash collisions.

riperiperi: Hmm... I guess it's taking the hash of the schedulerObj, rather than the WaitingObject.

hanabi1224: What actually being stored is wrapper of obj+timepoint, a hashset can only be used when wrapper's ```GetHashCode``` is overriden to return ```obj.GetHashCode()```, that way the data structure does not present the logic by itself, it's implies ```wrapper.GetHashCode() == wrapper.obj.GetHashCode() ```

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
hanabi1224: When I attach a debugger, I can see in a good percent of cases, key is not found, that's why I added key check before entering lock which is more expensive

gdkchan: The problem is that `Dictionary` is not thread safe, so I believe that calling `ContainsKey` in one thread and adding elements in another is not going to be thread safe either.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
hanabi1224: ```_waitEvent.WaitOne((int)(next.TimePoint - timePoint));``` implies the condition is satisfied. Even if this sys api is considered untrustworthy, we should change ```if (next.TimePoint > timePoint)``` to ```while``` loop, otherwise the most expensive ```_waitingObjects.OrderBy``` is called again

gdkchan: `_waitEvent` is also signaled on `ScheduleFutureInvocation` (because a object with a lower time point might have been added, so it needs to re-check), that's why the check is necessary.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
hanabi1224: Ummm, I think it's just a classic double-checked locking pattern, can you please elaborate in which case it causes an issue here?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
hanabi1224: Sounds good

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
gdkchan: I'm just saying that calling `ContainsKey` in one thread, and then something like `Add` in another at the sime time might cause problems, simply because `Dictionary` is not thread safe, so it makes no guarantees that something like this will work properly.

hanabi1224: I believe a concurrent read-only operation may result in the wrong result, which is handled by the double lock pattern, note that ```Remove()``` function checks key existence again and return boolean.  As long as it causes no crash / panic / UB, it should be fine right?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
hanabi1224: Ah I c, that makes sense

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
gdkchan: I think you're making an assumption about the implementation here. You assumed that calling `ContainsKey` from one thread while items are added on another is fine. I believe it should be, but if the API is not documented to be safe to use in this way, it may break in future updates for example. Bugs caused by race conditions are hard to reproduce by nature, that's why I prefer to be on the safe side with those things.

hanabi1224: > If you are only reading from a shared collection, then you can use the classes in the System.Collections.Generic namespace.

https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/

gdkchan: This is doing read *and* write to a shared collection though (`ContainsKey` would be read, `Add` and `Remove` would be write).

FICTURE7: I think this can throw an `InvalidOperationException` for concurrent operation, even for `ContainsKey`.

hanabi1224: I agree that concurrent write ```Add``` or ```Remove``` causes issues, but all write ops are inside lock block here.  And I don't think one write + multiple read in the same time causes ```InvalidOperationException``` (I have never met such issue with this assumption tho), I would definitely file a bug if that happens because that design just simply disallow using the double-check locking pattern with a non concurrent collection to reduce locking overhead before concurrent collections are introduced into stdlib :)

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -65,31 +67,29 @@ private void WaitAndCheckScheduledObjects()
...
                         long timePoint = PerformanceCounter.ElapsedMilliseconds;
-
                         if (next.TimePoint > timePoint)
                         {
                             _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
                         }
 
-                        bool timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
-
+                        var timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
```
Thog, line 85: To reduce diff
```suggestion
                        bool timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
```

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -65,31 +67,29 @@ private void WaitAndCheckScheduledObjects()
             {
                 while (_keepRunning)
                 {
-                    WaitingObject next;
-
+                    IKFutureSchedulerObject next;
```
Thog, line 70: line return

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
Thog: I guess this todo can be removed now?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -65,31 +67,29 @@ private void WaitAndCheckScheduledObjects()
...
                     lock (_context.CriticalSection.Lock)
                     {
+                        // TODO: PriorityQueue will has the best performance here.
```
Thog, line 73: I guess this todo can be removed now?

Thog (approved):
lgtm

EmulationFanatic:
@hanabi1224 are you still interested in this PR?

Thog:
Closing as stale, will be investigated again in the future.

PR #2121, opened 2021-03-18
Ipc: Remove IIpcService interface

AcK77:
This PR remove the IIpcService.cs interface usage which isn't needed anymore since the interface is only used by IpcService class. We can use it directly.


Thog (approved):
lgtm

gdkchan (approved):

PR #2138, opened 2021-03-25
Add support for HLE macros and accelerate MultiDrawElementsIndirectCount

gdkchan:
This adds support for HLE macros. This is basically an implementation of the macro code in C#, that can take better advantage of the host API and improve performance of some draw operations, like for example multi-draws. With the old implementation, each multi-draw would run the macro code and performs N separate draws. With an HLE macro, it's now possible to just call the host API multi-draw function directly, which not only reduces the number of API calls, but also enables passing the GPU indirect buffers directly, which eliminates the need for buffer flush and greatly improves the performance of games that does indirect draws.

For now, the only HLE macro function implemented is `MultiDrawElementsIndirectCount`, used by Monster Hunter Rise.

While working on this, I had the following problems, I found solutions for them but they are not exactly ideal:
- Unknown index buffer size. Right now, the index count is used to calculate the size, however, on indirect draws, this size is stored on a GPU buffer. Right now the code just assumes those are written from the CPU, which is the case for Monster Hunter Rise, but it could break for other games.
- OpenGL does not allow the index buffer offset to be set. It can only bind from the start of the buffer. Right now I copy the index buffer sub-range to a temporary buffer and bind that before each draw. This is not very efficient. Vulkan should not have this issue.

TODO:
- Find a way to handle `startOffset != 0` on the indirect draw.
- Flush indirect buffer data? This is required for correctness, but right now it takes all the performance benefit of the change, needs further investigation.

### DIFF
--- a/Ryujinx.Common/Hash128.cs
+++ b/Ryujinx.Common/Hash128.cs
@@ -9,6 +9,12 @@ namespace Ryujinx.Common
         public ulong Low;
         public ulong High;
 
+        public Hash128(ulong low, ulong high)
+        {
+            Low = low;
+            High = high;
+        }
+
         public override string ToString()
         {
             return $"{High:x16}{Low:x16}";
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -33,6 +33,9 @@ namespace Ryujinx.Graphics.GAL
 
         void EndTransformFeedback();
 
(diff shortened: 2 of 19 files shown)


flavien-marcantoni (approved):

MetrosexualGarbodor (approved):

Kevinjeter:
i dont get it, when i download the artifacts do i do anything else?

gdkchan:
Superseded by #2557.

PR #2144, opened 2021-03-26
Use copy dependencies for the Intel/AMD view format workaround

riperiperi:
Currently, on Intel/AMD (windows, not mesa) we must perform a copy whenever rendering to a view with a different format from its storage. This is rather slow because it happens after every draw. This change moves this behaviour out of the backend, so that it can be covered by Copy Dependencies, which are much faster as they only copy on read, rather than after every write.

Upsides:
- Greatly speeds up drawing to views with a different format for AMD/Intel, as the copy is only performed when the texture with the other format is read, rather than after every draw.
- Removes the SignalModified calls and checks from GAL, which may speed up drawing a little on NVIDIA where it did not matter.
  - Removing this makes the code in the gl backend a little nicer too.
- This may avoid other weird issues relating to view format mismatches, for example if there is an issue with image store. 

Downsides:
- Have to check capabilities within texture view compatibility check, which is a tad messy.
- May create copy dependencies when they are not needed, for example between the format matching storage that is rendered to, and a different format view which is only read from.

I'd recommend testing on a few games on AMD/Intel and Windows. Mario Kart 8 definitely has a boost, but other games may get worse (captain toad treasure tracker?).

### DIFF
--- a/Ryujinx.Graphics.GAL/Capabilities.cs
+++ b/Ryujinx.Graphics.GAL/Capabilities.cs
@@ -5,6 +5,7 @@ namespace Ryujinx.Graphics.GAL
         public bool SupportsAstcCompression          { get; }
         public bool SupportsImageLoadFormatted       { get; }
         public bool SupportsNonConstantTextureOffset { get; }
+        public bool SupportsMismatchingViewFormat    { get; }
         public bool SupportsViewportSwizzle          { get; }
 
         public int   MaximumComputeSharedMemorySize { get; }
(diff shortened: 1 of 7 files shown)


EmulationFanatic:
Improved MK8D on a Ryzen 7 5800X/RX 5500XT combo from 16fps average on master to 32fps average with this PR.

infirit:
I'll post more fps numbers when I have time to try other games. One thing I did notice that recent mainline builds improved performance for me on AMD. It would be good if the builds could be added based on a more recent git revision.

vsync off otherwise default options

Rev f9e5ffd (intel-amd-copy)
Luigi's Mansion
Title/Gameload screen 25-30 (fluctuating based on movement)
Lobby wide camera angle 5-7 fps
Lobby close camera angle 8-10 fps

Version 1.0.6823
Luigi's Mansion
Title/Gameload screen 26-30 (fluctuating based on movement)
Lobby wide camera angle 6/8 fps
Lobby close camera angle 10-11 fps

Overall slightly better on 1.0.6823.

riperiperi:
I'm undrafting this, since though it's a bit more general than the workaround it replaces, it's a clear improvement.

Thog (approved):

PR #2145, opened 2021-03-26
Shader Cache: Catch bindless textures early

mageven:
This PR moves the bindless check stopgap from translation to decode phase and disables cache-related processing early.

Fixes MHR v1.1.1 shader cache crash and possibly other bindless texture related shader cache crashes.

Thanks to @gdkchan for suggestions.

Since this affects shader caches, needs some testing to see if there are regressions.


Thog (approved):

riperiperi (approved):
Sensible.

PR #1942, opened 2021-01-20
Add items to standard logs

Joshi234:
This is useful if for troubleshooting issues.

Closes https://github.com/Ryujinx/Ryujinx/issues/1728

PR #2146, opened 2021-03-26
Improves logging to console.

Pixxel123:
This PR aims to expose more user changeable information in the logs, mainly so that the upcoming Discord bot (https://github.com/Ryujinx/ryuko-ng/pull/3) can use this information for analysis.

This also simplifies the hotkey states logging from #1942 (which I have now removed), as it uses the already recorded configuration states and leverages the already existing `ReactiveObject.cs` change state.

Both UI menu item changes and hotkey changes are logged in real time once settings are saved.

New information includes:

- File Logging Enabled
- File Integrity Checks Enabled
- File Global Access Log Mode
- Resolution scale
- Resolution scale custom
- Aspect Ratio
- Anisotrophic Filtering
- Docked/Undocked
- PPTC enabled/disabled
- Shader Cache enabled/disabled
- V-Sync enabled/disabled
- Audio Backend
- Ignore Missing Services enabled/disabled

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationState.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationState.cs
@@ -91,10 +91,10 @@ namespace Ryujinx.Configuration
             {
                 GuiColumns        = new Columns();
                 ColumnSort        = new ColumnSortSettings();
-                GameDirs          = new ReactiveObject<List<string>>();
-                EnableCustomTheme = new ReactiveObject<bool>();
-                CustomThemePath   = new ReactiveObject<string>();
-                StartFullscreen   = new ReactiveObject<bool>();
+                GameDirs          = new ReactiveObject<List<string>>(); //{ Name = "Game dir", Category = "UI" };
+                EnableCustomTheme = new ReactiveObject<bool>(); //{ Name = "Custom Theme", Category = "UI" };
+                CustomThemePath   = new ReactiveObject<string>(); //{ Name = "Custom Theme", Category = "UI" };
+                StartFullscreen   = new ReactiveObject<bool>(); //{ Name = "Fullscreen", Category = "UI" };
             }
         }
 
@@ -163,8 +163,8 @@ namespace Ryujinx.Configuration
                 EnableGuest        = new ReactiveObject<bool>();
                 EnableFsAccessLog  = new ReactiveObject<bool>();
                 FilteredClasses    = new ReactiveObject<LogClass[]>();
-                EnableFileLog      = new ReactiveObject<bool>();
-                GraphicsDebugLevel = new ReactiveObject<GraphicsDebugLevel>();
+                EnableFileLog      = new ReactiveObject<bool>(); //{ Name = "File Log", Category = "Logger" };
+                GraphicsDebugLevel = new ReactiveObject<GraphicsDebugLevel>() { Name = "Graphics level", Category = "Logger" };
             }
         }
 
@@ -268,7 +268,7 @@ namespace Ryujinx.Configuration
 
             public HidSection()
             {
-                EnableKeyboard = new ReactiveObject<bool>();
+                EnableKeyboard = new ReactiveObject<bool>() { Name = "Keyboard", Category = "Input" };
                 Hotkeys        = new ReactiveObject<KeyboardHotkeys>();
                 InputConfig    = new ReactiveObject<List<InputConfig>>();
             }
@@ -316,13 +316,13 @@ namespace Ryujinx.Configuration
 
             public GraphicsSection()
             {
-                ResScale          = new ReactiveObject<int>();
-                ResScaleCustom    = new ReactiveObject<float>();
-                MaxAnisotropy     = new ReactiveObject<float>();
-                AspectRatio       = new ReactiveObject<AspectRatio>();
-                ShadersDumpPath   = new ReactiveObject<string>();
-                EnableVsync       = new ReactiveObject<bool>();
-                EnableShaderCache = new ReactiveObject<bool>();
+                ResScale          = new ReactiveObject<int>() { Name = "Res Scale", Category = "Graphics" };
+                ResScaleCustom    = new ReactiveObject<float>() { Name = "Custom Scale", Category = "Graphics" };
+                MaxAnisotropy     = new ReactiveObject<float>() { Name = "Anisotrophy", Category = "Graphics" };
+                AspectRatio       = new ReactiveObject<AspectRatio>() { Name = "Aspect", Category = "Graphics" };
+                ShadersDumpPath   = new ReactiveObject<string>(); //{ Name = "Dump Path", Category = "Graphics" };
+                EnableVsync       = new ReactiveObject<bool>() { Name = "VSync" , Category = "Graphics"};
+                EnableShaderCache = new ReactiveObject<bool>() { Name = "Shader Cache", Category = "Graphics" };
             }
         }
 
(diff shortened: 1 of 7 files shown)


AcK77:
`Ryujinx.HLE/HOS/Horizon.cs` Needs a rebase

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -30,12 +35,17 @@ public T Value
...
 
                 _isInitialized = true;
-                _value         = value;
+                _value = value;
 
                 _readerWriterLock.ReleaseWriterLock();
 
                 if (!oldIsInitialized || !oldValue.Equals(_value))
                 {
+                    if (!string.IsNullOrEmpty(Name))
```
AcK77, line 44: Add brackets hover the next 3 lines

Ryujinx/Ui/Windows/SettingsWindow.cs
```diff
@@ -432,17 +430,6 @@ private void SaveSettings()
             ConfigurationState.Instance.Graphics.ResScale.Value                = int.Parse(_resScaleCombo.ActiveId);
             ConfigurationState.Instance.Graphics.ResScaleCustom.Value          = resScaleCustom;
 
-            if (_audioBackendSelect.GetActiveIter(out TreeIter activeIter))
```
AcK77, line 435: You should add this next code back, because if you don't, the audio backend isn't configurable anymore using the GUI.
```
            if (_audioBackendSelect.GetActiveIter(out TreeIter activeIter))
            {
                ConfigurationState.Instance.System.AudioBackend.Value = (AudioBackend)_audioBackendStore.GetValue(activeIter, 1);
            }
```

(https://github.com/Ryujinx/Ryujinx/pull/1942/files#diff-52233a1b2858093f071f6439da4b0c888e8fdc4859d60ecdf73b6089c43d739bL434)

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -8,6 +9,10 @@ public class ReactiveObject<T>
         private ReaderWriterLock _readerWriterLock = new ReaderWriterLock();
         private bool _isInitialized = false;
         private T _value;
+        public string _name { get; private set; }
```
AcK77, line 12: Add a new empty line before

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -30,12 +35,17 @@ public T Value
                 bool oldIsInitialized = _isInitialized;
 
                 _isInitialized = true;
-                _value         = value;
```
AcK77, line 33: Could be nice to keep the alignment here

Ryujinx.Common/Configuration/ConfigurationState.cs
```diff
@@ -91,10 +91,10 @@ public UiSection()
             {
                 GuiColumns        = new Columns();
                 ColumnSort        = new ColumnSortSettings();
-                GameDirs          = new ReactiveObject<List<string>>();
-                EnableCustomTheme = new ReactiveObject<bool>();
-                CustomThemePath   = new ReactiveObject<string>();
-                StartFullscreen   = new ReactiveObject<bool>();
+                GameDirs          = new ReactiveObject<List<string>>(); //{ Name = "Game dir", Category = "UI" };
```
AcK77, line 94: I don't think those commented args are really useful. They could be added easily if needed so it's probably better to remove those.
(Same suggestion for all other fields)

AcK77 (changes requested):
Good idea, it's way better than the current way. I've just some changes requested ;)!

Ryujinx.Common/Configuration/ConfigurationState.cs
Pixxel123: Would it be worth having a comment at the start that states something like "Add a name and category to items that should appear in the log"?

AcK77: You can remove the commented lines and keep only one you have done, since they are useful one! The rest of fields doesn’t really needs logging, and if they needs to be logged one day, we could easily understand how do it by looking to other declared one. (Even more when you considering this will be only done by developers)

Ryujinx.Common/Logging/LogClass.cs
```diff
@@ -57,6 +57,7 @@ public enum LogClass
         ServiceTime,
         ServiceVi,
         SurfaceFlinger,
-        Vic
+        Vic,
+        Configuration
```
AcK77, line 61: Sort this alphabetically

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -9,6 +10,17 @@ public class ReactiveObject<T>
...
         private T _value;
 
+        private string _loggedName;
+
+        public ReactiveObject(string loggedName = "")
+        {
+            _loggedName = loggedName;
+        }
+
+        private string _name { get; set; }
```
AcK77, line 20: `_name` and `Name` aren't used and can be removed

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -36,11 +48,20 @@ public T Value
...
+                        //Some reactive object names are left commented out here, since they don't need to be exposed to the end user but the option is there for devs
+                        Logger.Info?.Print(LogClass.Configuration, $"{_loggedName} set to: {_value}");
+                    }
+
                     Event?.Invoke(this, new ReactiveEventArgs<T>(oldValue, value));
                 }
             }
         }
 
+        public string LoggedName { get; }
```
AcK77, line 63: This is not used and can be removed too

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -36,11 +48,20 @@ public T Value
 
                 if (!oldIsInitialized || !oldValue.Equals(_value))
                 {
+                    if (!string.IsNullOrEmpty(_loggedName))
+                    {
+                        //When values are changed, logged to console. If value does not have a Name, no log is printed
```
AcK77, line 53: I don't think those 2 lines of comment are really revelent. They could be removed.

AcK77 (changes requested):

AcK77 (approved):
Lgtm! Thanks!

Thog (approved):

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -9,6 +10,13 @@ public class ReactiveObject<T>
         private bool _isInitialized = false;
         private T _value;
 
+        private string _loggedName;
+
+        public ReactiveObject(string loggedName = "")
```
jduncanator, line 15: I don't really like the idea of this information being stored in the ReactiveObject itself, the ReactiveObject is designed to be reusable as a generic component, so having it intrinsically linked to "logging" in some way breaks the concept of separation of concerns. Worst case, I'd rather this just be called `name`, rather than `loggedName`.

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -36,6 +44,11 @@ public T Value
 
                 if (!oldIsInitialized || !oldValue.Equals(_value))
                 {
+                    if (!string.IsNullOrEmpty(_loggedName))
+                    {
+                        Logger.Info?.Print(LogClass.Configuration, $"{_loggedName} set to: {_value}");
```
jduncanator, line 49: I think this should be handled in the Configuration code. You can add an additional event handler to the `Event` event of each configuration object, and then log the changes there. The way this currently is, any other reuse of ReactiveObject for anything not related to configuration will potentially spam these log messages.

Ryujinx.Common/ReactiveObject.cs
jduncanator: Perhaps you could use a `DisplayNameAttribute` on each configuration field in `ConfigurationState` and use that to determine a "friendly name" for the configuration object?

Ryujinx.Common/ReactiveObject.cs
```diff
@@ -36,7 +36,7 @@ public T Value
 
                 if (!oldIsInitialized || !oldValue.Equals(_value))
                 {
-                    Event?.Invoke(this, new ReactiveEventArgs<T>(oldValue, value));
+                      Event?.Invoke(this, new ReactiveEventArgs<T>(oldValue, value));
```
gdkchan, line 39: Why the extra space here?

gdkchan (commented):
Needs a rebase now that a new setting has been added.

jms-c (approved):
Looks good to me! 
Have a nice day!

riperiperi (approved):
Looks good, will be useful seeing this info on logs.

PR #2147, opened 2021-03-27
Fix ZN flags set for shader instructions using RZ.CC dest

gdkchan:
This fixes a bug where the Z/N flags would be set to a incorrect value if the destination register of some shader instructions are RZ. This was happening because it was comparing the RZ register value, rather than the actual operation result.

While I was at it, I also fixed a bug on pixel shaders with calls where it would set the output pixel values before every return, rather than just on the main function.

I found the first bug while looking at one of the Monster Hunter Rise compute shaders, but I did not notice any improvement after fixing this.


Thog (approved):
nice catch!

riperiperi (approved):

PR #2151, opened 2021-03-29
Optimize PrintRoSectionInfo

mageven:
Recent applications seem to ship large binaries and I noticed `PrintRoSectionInfo` is slowing down loads.

- Remove unnecessary `Span.ToArray()`
- Skip a search if name is already found (previously always searched)
- Use `CultureInvariant` with `IgnoreCase` for safety
- Use `Regex.Match` when only first matches are needed. Using `Count` bypasses laziness.
- Use `RegexOptions.Compiled`. Helped in tests, probably because this operates on large "strings". 

It's a shame that .net Regex is pretty inflexible and works only on strings. A considerable portion of execution time is still due to `ASCII.GetString()`.

My tests indicate this method is 40-60% faster now.

### DIFF
--- a/Ryujinx.HLE/Loaders/Executables/NsoExecutable.cs
+++ b/Ryujinx.HLE/Loaders/Executables/NsoExecutable.cs
@@ -4,7 +4,6 @@ using LibHac.FsSystem;
 using LibHac.Loader;
 using Ryujinx.Common.Logging;
 using System;
-using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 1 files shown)


jduncanator:
If anyone's interested in future work, the Regex is probably simple enough to be reimplemented naively to operate only on Span's.

gdkchan (approved):
lgtm, any load speed improvement is welcome 👍

riperiperi (approved):
Nice one.

AcK77 (approved):

PR #2153, opened 2021-03-29
Patch up DNS Blacklist

mageven:
Very simple change. Hostname resolution IPC calls don't seem to have any filtering on the input name so, this PR makes the DNS blacklist case-insensitive to catch them properly. Also added a few extra perf-oriented options to them- `ExplicitCapture` as we only use `IsMatch`, `Compiled` as applications can spam it.

A quick test using a single regex with alternation, something like-
```csharp
Regex blockedHosts = new Regex(string.Join('|', patternsArray), options);
```
yielded no clear winner with a small sample of tests I tried (complex state machine or more backtracking?)
As that's more error-prone and confusing anyway, decided not to use it. If you still prefer a single regex, let me know.


AcK77 (approved):

AcK77:
Nice catch!

Thog (approved):

PR #2155, opened 2021-03-30
Improve `StoreToContext` emission

FICTURE7:
Hoist `StoreToContext` in dynamic branch fast & slow paths out into their predecessor. Reduces register pressure, code size and compile time a little bit because we're throwing less stuff down the pipeline. Sample [diff](https://www.diffchecker.com/17Or5Uog).

Some thing similar can be done with `LoadFromContext` emitted for dynamic calls, by sinking them into the successor. We need to store the return address of the calls (fast & slow paths) into a single variable and apply the call guard check on it. This is essentially tail merging and could be applied to tail continues as well. However we get limited by the `EmitterContext` and `SsaConstruction` because local variables are assumed to be in SSA early on, so they cannot participate in phi nodes.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitFlowHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitFlowHelper.cs
@@ -147,10 +147,8 @@ namespace ARMeilleure.Instructions
             EmitJumpTableBranch(context, Const(immediate), isRecursive);
         }
 
-        private static void EmitNativeCall(ArmEmitterContext context, Operand nativeContextPtr, Operand funcAddr, bool isJump = false)
+        private static void EmitNativeCall(ArmEmitterContext context, Operand nativeContextPtr, Operand funcAddr, bool isJump)
         {
-            context.StoreToContext();
-
             if (isJump)
             {
                 context.Tailcall(funcAddr, nativeContextPtr);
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
lgtm, great improvement 👍 
While you are at it, maybe you could make `EmitDynamicTableCall` private, it confused me a bit when I was looking through it (and I do find this part of the code a bit confusing in general, I opened https://github.com/Ryujinx/Ryujinx/issues/2158 improvement proposal).
> It might get superseded by #1954 though.

I'm not sure about those changes myself. Would be nice to turn `Operand` into a struct and get rid of the pools, but the approach that I used to build definitions/uses seems more error prone.
In any case, any change that reduces the allocation is welcome, since its sort of a huge problem right now (especially with PPTC rebuild, I constantly get out of memory here with only 8GB of ram...).

FICTURE7:
> While you are at it, maybe you could make EmitDynamicTableCall private

Good call, I missed that.

> and I do find this part of the code a bit confusing in general

heh, same

> Would be nice to turn Operand into a struct and get rid of the pools

One of the thing I wish we did is to turn `Operand` into a struct, allocate an array (perhaps pinned) of `Operand`. Then use an arena style allocator on that array, which fits quite nicely with the allocation patterns of the JIT. We then refer to these `Operand` through a 32-bit index or maybe a new type `OperandRef`. Plus I think we can pack `Operand` into 16 bytes. Caveat is that we have to explicitly "dereference" the index into the array which makes `InstEmitter` code not nice.

As for `Operation`s, most of the time, they are binary or unary operations (I haven't collected any metrics but it is pretty apparent from the IR dumps & from x86' semantics). We could probably get away with classes and create a `BinaryOperation`/`UnaryOperation` subclass. It should not contain the `_sources` and `_destinations` list of `Node` but instead contains the `Operand` fields inlined. Should save us a couple of allocations of `List<T>`. Maybe have a `NaryOperation` which is the current `Operation : Node` implementation for rarer cases.

I tried doing that in the past but ran into certain cases which became more complicated to handle and did not follow through. e.g here:
https://github.com/Ryujinx/Ryujinx/blob/69f8722e795b0d76b84efa4de52e199e9441fca7/ARMeilleure/CodeGen/X86/PreAllocator.cs#L284-L311
The operation was originally binary but turned ternary so the entire operation should probably be replaced or something.

riperiperi (approved):
Nice catch, was definitely a bit messy here before.

PR #2156, opened 2021-03-30
Hold reference for render targets in use

gdkchan:
`SetRenderTargets` might be called on the backend with a texture that was already "released". This happens because the active render targets are stored on the `TextureManager`, but this never increments/decrements the reference count, so those textures might be deleted if they fall out of the auto delete cache. This changes increments and decrements the reference count to prevent that from happening.

Found this while working on Vulkan (using a freed texture would cause a crash), not sure if its causing anything spooky on the OpenGL backend (I guess its less likely to crash there since textures are reused).

Also moved `SetRenderTargetDepthStencil` close to the color one, as the 2 are related and puting them together makes the code easier to read imo.


Thog (approved):
nice catch

riperiperi (approved):

PR #1928, opened 2021-01-18
Add the TamperMachine module for runtime mods and cheats

Caian:
This is an implementation of a small programmable thread that enables runtime memory modifications and cheats. It is named Tamper instead of Cheat because some modifications are not necessarily cheats, they alter the game to provide different experiences, some possible only with emulators, like unlocked FPS mods.

Although standalone programs such as CheatEngine exist, they can be a hassle for some users and for some anti-cheat systems.

This implementation is integrated with the `ModLoader` class and shares the mod folder already available in Ryujinx.

It implements full support Atmosphere style cheats (except game pausing/resume): 
https://github.com/Atmosphere-NX/Atmosphere/blob/master/docs/features/cheats.md

## Installation of Atmosphere Cheats

Two good places to acquire Atmosphere cheats are:

- https://github.com/WerWolv/EdiZon_CheatsConfigsAndScripts/tree/master/Cheats (general switch cheats)
- https://yuzu-emu.org/wiki/switch-mods/ (emulator-specific mods such as high FPS, ultrawide, etc...)

Inside the cheat `zip`, `rar`, or other archive format, there is usually a file structure like this:

```
Cheat_Name\cheats\0123456789ABCDEF.txt
```

Where `0123456789ABCDEF` will vary according to the game and the cheat (this is the build id of the executable the cheat attaches itself to). A cheat can also contain several `txt` files. **Cheats are game-specific and version-specific, they will not work if transferred from one game to another or to an incompatible update of the game.**

To install the cheat: 

- Open Ryujinx, right-click the game you wish to install the cheat and select to open its `Mod directory`;
- A directory like this will open: `Ryujinx\mods\contents\XXXXXXXXXXXXXXXX\`, `XXXXXXXXXXXXXXXX` will be the id of the chosen game;
- Extract the cheat archive to the mod directory, keeping the cheat name: `Ryujinx\mods\contents\XXXXXXXXXXXXXXXX\Cheat_Name\cheats\<all the txt files>`.

## Activating/Deactivating cheats

Currently, all cheats are automatically enabled and cannot be disabled, so mixing, for instance, 60, 90 and 120 fps cheats may cause a conflict. 

One alternative is to create a directory `disabled_cheats` in the mod directory and move all unused cheats there. Another one is to prepend something to the file name, for instance `off_0123456789ABCDEF.txt` as a valid cheat must not have anything but `0123456789ABCDEF.txt` in its name.

[...]

PR #2159, opened 2021-04-01
Fix Tamper Machine crash and timings

Caian:
Discord user mageven reported two issues with the Tamper Machine that are addressed in this PR:

- The first one is a crash when using unpacked games, caused by a missing `return` statement.
- The second one was an inconsistency between the sleep duration in Tamper Machine (1ms) versus Atmosphere (~83.33ms). Time sensitive cheats (if any) may behave differently.

The first issue was reproduced and tested unpacking a game and running it.

For the second issue, I measured the average time between sleeps in the Tamper Machine and it's around 83.7~83.9ms on Windows/Release, even with multiple cheats running, so the timing is now between spec and no significant drift from the `Sleep` function itself was observed.

I re-tested the cheats from my previous PR #1928, those being:

- 60fps mod in MHGU
- 99 lives / 999 seconds / start as penguin cheats in NSMBD
- 10 coins cheat in MK8D

### DIFF
--- a/Ryujinx.HLE/HOS/TamperMachine.cs
+++ b/Ryujinx.HLE/HOS/TamperMachine.cs
@@ -13,6 +13,8 @@ namespace Ryujinx.HLE.HOS
 {
     public class TamperMachine
     {
+        private const int TAMPERMACHINE_SLEEP_MS = 1000 / 12;
+
         private Thread _tamperThread = null;
         private ConcurrentQueue<ITamperProgram> _programs = new ConcurrentQueue<ITamperProgram>();
         private long _pressedKeys = 0;
@@ -76,7 +78,7 @@ namespace Ryujinx.HLE.HOS
                 if (sleepCounter == 0)
                 {
                     sleepCounter = _programs.Count;
-                    Thread.Sleep(1);
+                    Thread.Sleep(TAMPERMACHINE_SLEEP_MS);
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


AcK77 (approved):

Caian:
> thank you
> i found an issues! cheating not work for this format
> [invincible]
> 040B0000 02430204 1E249000

Thanks, this is a known issue to be addressed in another PR.

Myoko:
> > thank you
> > i found an issues! cheating not work for this format
> > [invincible]
> > 040B0000 02430204 1E249000
> 
> Thanks, this is a known issue to be addressed in another PR.

thats good news, thank you

Ryujinx.HLE/HOS/TamperMachine.cs
```diff
@@ -13,6 +13,8 @@ namespace Ryujinx.HLE.HOS
 {
     public class TamperMachine
     {
+        private const int TAMPERMACHINE_SLEEP_MS = 1000 / 12;
```
Thog, line 16:
```suggestion
        // 83ms
        private const int TamperMachineSleepMs = 1000 / 12;
```

Ryujinx.HLE/HOS/TamperMachine.cs
```diff
@@ -76,7 +78,7 @@ private void TamperRunner()
                 if (sleepCounter == 0)
                 {
                     sleepCounter = _programs.Count;
-                    Thread.Sleep(1);
+                    Thread.Sleep(TAMPERMACHINE_SLEEP_MS);
```
Thog, line 81:
```suggestion
                    Thread.Sleep(TamperMachineSleepMs);
```

Thog (approved):
I think the constant need to be in PascalCase but otherwise lgtm :+1:

Caian:
Thanks @Thog, fixed.

PR #2163, opened 2021-04-01
Implement shader HelperThreadNV

gdkchan:
From the [spec](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_shader_thread_group.txt):
```
The variable gl_HelperThreadNV specifies if the current thread is a helper
thread.  In implementations supporting this extension, fragment shader
invocations may be arranged in SIMD thread groups of 2x2 fragments called
"quad".  When a fragment shader instruction is executed on a quad, it's
possible that some fragments within the quad will execute the instruction
even if they are not covered by the primitive.  Those threads are called
helper threads.  Their outputs will be discarded and they will not execute
global store functions, but the intermediate values they compute can still
be used by thread group sharing functions or by fragment derivative
functions like dFdx and dFdy.
```
On Monster Hunter Rise, this seems to be used to ignore such "helper threads" on shuffle operations. Ignoring this register caused some data to not be written into a SSBO, which would later cause a compute shader to calculate the wrong instance count for instanced indirect multi-draws, and cause the "flickering" effect where objects would randomly disappear, because the instance count was incorrect.

This change fixes the flickering on this game, which now appears to be graphically perfect (apart from the seemingly random corrupt cubemap texture?).

~~This extension is only supported by NVIDIA. I'm not sure if it's possible to emulate it on other vendors, I believe it's better to just return a constant value on those (objects will likely flicker but at least the shader will compile, which should keep the current behavior).~~ It uses `gl_HelperInvocation` now, which is supported by all vendors (thanks to @ReinUsesLisp for the suggestion), however this requires GLSL 4.5, **so the version requirement has been increased from 4.4 to 4.5**.


michaelpegaso:
Tested and works perfect. Thank you very much for your work.

Thog (approved):

jduncanator (approved):

PR #2165, opened 2021-04-02
Only flush commands when creating sync on Intel/AMD (windows)

riperiperi:
One particular game seemed to slow down due to flushing commands on sync creation, so this PR disables it on the platforms that don't seem to need it, such as NVIDIA's proprietary drivers and Mesa. My guess is that the NVIDIA driver is always flushing commands anyways, and telling it to do so manually slows it down.


gdkchan (approved):
lgtm, tbh I would prefer to not do this since according to the spec, flushing after sync creation would be the right thing to do, but it doesn't seems to cause issue on NVIDIA to not do so, and it has higher performance, so.

riperiperi:
Yes, I would like if this weren't faster, too.

jduncanator (approved):

PR #1865, opened 2021-01-04
PPTC meets ExeFS Patching.

LDj3SNuD:
Changelog:
- PPTC now supports Mods;
- PPTC already supported Cheats;
In both cases all mods or cheats of any complexity should work, but they should not change the external execution flow (between functions), otherwise functions could be profiled which should not or cannot be translated; I am not aware of such applications in practice.
- Refactoring of PtcProfiler.cs (the internal structure of the .info files now follows that of the .cache files; use of Hash128 instead of MD5; etc.) and Follow-up of Ptc.cs;
- *As a result of the aforementioned refactoring of PtcProfiler.cs, but also as a result of https://github.com/Ryujinx/Ryujinx/pull/2190, users' .info files will be invalidated*.

PR #2168, opened 2021-04-03
PPTC vs. giant ExeFS.

LDj3SNuD:
- PPTC can now handle any size of JitCache;
Allows to avoid memory peaks due to the resizing of the internal MemoryStream array (and its possible waste), its allocation in the LOH region of the heap; and allows to reduce the size of the .cache file if many translation invalidations are involved (LowCq -> HighCq over several runs).
- Added an outer header structure and added the hashes for both this new structure and the existing "inner" header structure;
- Hashes are now handled on a per-section basis and no longer for the entire uncompressed stream.
* Added a new heuristic algorithm for calculating the number of threads for parallel translations that also takes into account the user's free physical memory and not just the number of CPU cores.

For those games with a large ExeFS, this PR fixes:
```
|N| Application : Unhandled exception caught: System.IO.IOException: Stream was too long.
   at System.IO.MemoryStream.Write(ReadOnlySpan`1 buffer)
   at ARMeilleure.Translation.PTC.Ptc.WriteInfoCodeRelocUnwindInfo(UInt64 address, UInt64 guestSize, Boolean highCq, PtcInfo ptcInfo) in ARMeilleure\Translation\PTC\Ptc.cs:line 873
   at ARMeilleure.Translation.Translator.Translate(IMemoryManager memory, JumpTable jumpTable, UInt64 address, ExecutionMode mode, Boolean highCq) in ARMeilleure\Translation\Translator.cs:line 257
   at ARMeilleure.Translation.PTC.Ptc.<>c__DisplayClass78_0.<MakeAndSaveTranslations>g__TranslateFuncs|0() in ARMeilleure\Translation\PTC\Ptc.cs:line 794
```
```
|E| Application : Unhandled exception caught: System.OutOfMemoryException: Insufficient memory to continue the execution of the program.
   at System.Runtime.InteropServices.Marshal.AllocHGlobal(IntPtr cb)
   at ARMeilleure.Translation.PTC.Ptc.Save(String fileName) in ARMeilleure\Translation\PTC\Ptc.cs:line 428
   at ARMeilleure.Translation.PTC.Ptc.PreSave() in ARMeilleure\Translation\PTC\Ptc.cs:line 396
```
However, this PR does not fix `System.OutOfMemoryException: JIT Cache exhausted.` (JitCache related).


HiHiKazumaDesu:
Hello, I tested this PR with my MHR PPTC. which weights about 2 GB, I'm not sure if this is a Rebase of PR #1865 but when I try to open the game PTC isn't being loaded, is this intended?

Myoko:
This pull conflicts with PR #1865

Shahil-Ayato:
> This pull conflicts with PR #1865

No it doesn’t, #1865 depends on this being merged.

Myoko:
> > This pull conflicts with PR #1865
> 
> No it doesn’t, #1865 depends on this being merged.

ok thanks for that tip

ARMeilleure/Translation/Translator.cs
```diff
@@ -103,6 +103,7 @@ public void Execute(State.ExecutionContext context, ulong address)
 
                 if (Ptc.State == PtcState.Enabled)
                 {
+                    Debug.Assert(_funcs.Count == 0);
```
Thog, line 106: Is it intended to stay here?

Thog (approved):

ARMeilleure/Translation/Translator.cs
LDj3SNuD: why not lol

riperiperi (approved):
Looks good to me, glad to see PPTC getting some more love.

One thing that might be a bit annoying - free memory being used to determine the PPTC compile paralellism might be a bit unpredictable. For instance, I'm usually above 70% memory usage, but I have a lot of swap space, and windows memory compression helps a lot. The free memory measurement doesn't really take account of these things. Though to be honest, I should probably upgrade to 32GB...

LDj3SNuD:
I have excluded free virtual memory for performance reasons (at least this applies to my years-old computer). The algorithm can be improved through user feedback. In your case it should not change anything if you have 4 GiB of free physical memory (30% of 16 GiB): you get a weight of 75% to apply to the number of your cores; so still 18 as per master (3/4 of 24).

PR #2169, opened 2021-04-03
Make jit cache non-static

### LINKED ISSUE #1823: Make JIT cache non-static
#### Overview:
On the CPU (ARMeilleure project), the JitCache is responsible for storing the JIT generated code. Currently it is static and has a 2GB region reserved for code of all `Translator` instances. It should be made non-static, and a instance of the cache should be stored per `Translator` instance.

This way, we don't run out of reserved memory if we have multiple `Translator` instances. Additionally, it makes disposal simpler as one can just dispose the entire cache when the translator is disposed, and the memory can be reclaimed by the OS.

I expect this to be fairly easy to do overall, but it requires the removal of the unwinder that is registered on windows to handle the unwinding of JIT generated functions. See: https://github.com/Ryujinx/Ryujinx/blob/5e724cf24e3d696b95be859c055a617e5d37bf80/ARMeilleure/Translation/JitUnwindWindows.cs#L62

I'm not sure if windows allows removing it, so peharps this part of the code should be kept static.

For anyone interested, I have started some work on this, but I am not working on it right now (and don't plan to anytime soon), I can make the code available somewhere.

#### Nature of Request:  
Fix/refactoring.

Avoids out of memory errors with multiple translators, makes disposal simpler.

Twilight-Struggle:
close #1823
~~JitUnwindWindows.cs needs to be modified.~~
~~Not for review.~~
Ready for review.

### DIFF
--- a/ARMeilleure/Translation/Cache/JitUnwindWindows.cs
+++ b/ARMeilleure/Translation/Cache/JitUnwindWindows.cs
@@ -7,7 +7,7 @@ using System.Runtime.InteropServices;
 
 namespace ARMeilleure.Translation.Cache
 {
-    class JitUnwindWindows
+    class JitUnwindWindows : IDisposable
     {
         private const int MaxUnwindCodesArraySize = 32; // Must be an even value.
 
@@ -53,6 +53,10 @@ namespace ARMeilleure.Translation.Cache
             IntPtr context,
             string outOfProcessCallbackDll);
 
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
+        private static unsafe extern bool RtlDeleteFunctionTable(
+            RuntimeFunction* functionTable);
+
         private GetRuntimeFunctionCallback _getRuntimeFunctionCallback;
 
         private int _sizeOfRuntimeFunction;
@@ -60,6 +64,8 @@ namespace ARMeilleure.Translation.Cache
         private unsafe RuntimeFunction* _runtimeFunction;
 
         private unsafe UnwindInfo* _unwindInfo;
+
+        private bool disposedValue = false;
 
         public JitUnwindWindows(IntPtr codeCachePointer, uint codeCacheLength, IntPtr workBufferPtr, JitCache jitCache)
         {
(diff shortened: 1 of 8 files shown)


ARMeilleure/Translation/Cache/JitUnwindWindows.cs
```diff
@@ -184,5 +196,40 @@ private static ushort PackUnwindOp(UnwindOp op, int prologOffset, int opInfo)
...
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                bool result;
+
+                unsafe
+                {
+                    result = RtlDeleteFunctionTable(
```
riperiperi, line 208: Should probably just be on one line.

riperiperi:
Unwinding does appear to be fine. I think my rebase was broken because you merged upstream rather than rebasing, but that's entirely fine.

gdkchan:
I think the `ClearJitCache` function can be remved now, theres no need to unmap each function one by one if the entire cache is freed afterwards anyway.

Twilight-Struggle:
rebased!

gdkchan:
> 
> 
> I think the `ClearJitCache` function can be removed now, theres no need to unmap each function one by one if the entire cache is freed afterwards anyway. But you will need to make sure the rejit queue is empty before disposing the JIT cache.

What about this?

Twilight-Struggle:
> What about this?

Sorry. I haven't checked it yet. But I will try it.

Twilight-Struggle:
> I think the `ClearJitCache` function can be removed now, theres no need to unmap each function one by one if the entire cache is freed afterwards anyway. But you will need to make sure the rejit queue is empty before disposing the JIT cache.

I've understood that within `ClearJitCache`, `_jitCache.Unmap` is not necessary.
And I also think I should run `ClearRejitQueue` to make the rejit queue empty.
`CallCounter?.Dispose` is still a necessary process, so I'm going to remove only `_jitCache.Unmap` from `ClearJitCache`.
However, this idea seems to contradict the "ClearJitCache function can be removed now".
Am I thinking about something wrong?

gdkchan:
`CallCounter?.Dispose` doesn't seems necessary as the `CountTable` is disposed after the `JitCache` is disposed anyway.

Twilight-Struggle:
While I was trying to eliminate ClearJitCache(), I found a bug that "Stop Emulation" did not work.
This also happens in the current commit (not eliminating ClearJitCache()).
This bug occurs in ServerBase.cs:70. "System.NullReferenceException: 'Object reference not set to an instance of an object.'"
In my environment, the bug occurred when I launched SSBU, helloworld, and SSBU in that order.
I don't understand the contents of ServerBase.cs, so I haven't been able to pinpoint the cause, but I figured out it happens after applying the #2241 pull request.

riperiperi:
With POWER, things have changed a little here. You'll note that now the first creation of the Translator instance creates and installs a signal handler, which isn't great for disposing the JitCache.

You'll need to uninstall this signal handler and reinstall it whenever a new JitCache is created for now. There's no method for uninstalling the signal handler right now, but there are platform specific `RestoreExceptionHandler` etc that can be used. Ideally in future these signal handlers should be on a small JitCache instance for shared functions and native handlers, as they will be used between all running processes, but that's a while off.

AcK77:
Needs a rebase

Twilight-Struggle:
Rebased and now signal handler is installed and uninstalled when JitCache is created and disposed.
Bacause of aforementioned issue, I cannot test JitCache dispose completely.
But it occasionally succeeds that run and stop emulation in Windows.
I have not tried it on Linux. I'm going try it with virtualbox, etc.

Benedani:
> I have not tried it on Linux. I'm going try it with virtualbox, etc.

That isn't going to work well due to the outdated OpenGL support.
I tested it, it opens a game and then stops emulation correctly.

Twilight-Struggle:
Thank you for testing on Linux!

How can I get this PR to complete?
I wish I could fix the aformentioned problem, but it would be difficult or very time consuming...

AcK77:
Needs a rebase

marysaka:
Closing as stale.

PR #2172, opened 2021-04-03
Convert struct declarations to read-only in ARMeilleure where trivially appropriate

yunnypuff:
https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code
- Convert `struct` declarations to `readonly` where trivially appropriate
- Convert usages of said structs to use the `in` keyword where appropriate
- The above provides compiler optimizations to use only references instead of copy semantics due to immutability guarantee
- Added dictionary lookup of JitCache by offset/index instead of a BinarySearch each time
- Observed 3-5fps improvement in a separate branch (with HLE macro) on MHR, but otherwise not rigorously profiled

More info on readonly/int:
> Because int is no larger than a reference in most modern machines, there is no benefit to passing a single int as a readonly reference.
> Add the in modifier to pass an argument by reference and declare your design intent to pass arguments by reference to avoid unnecessary copying. You don't intend to modify the object used as that argument.
> This practice often improves performance for readonly value types that are larger than IntPtr.Size. For simple types (sbyte, byte, short, ushort, int, uint, long, ulong, char, float, double, decimal and bool, and enum types), any potential performance gains are minimal. In fact, performance may degrade by using pass-by-reference for types smaller than IntPtr.Size.

Some additional considerations from XNA days:
https://docs.microsoft.com/en-us/archive/blogs/ricom/performance-quiz-11-ten-questions-on-value-based-programming-solution


yunnypuff:
Noticed performance degradation during the initial JitCache populating phase due to the addition of the dictionary mapping and adjustments. Going to test removing it and see what the tradeoff is.

yunnypuff:
Gonna close this PR out as I don't have any conclusive proof that performance has improved in any statistically significant way. Should've taken "profile first, optimize later" to heart.

Did find some interesting tidbits while studying some changes on the side:
* RyuJIT will inline certain small functions that operate on struct fields, but not inline the same functions if they were operating on auto properties (which are readonly)
* On Win10 RyuJIT x64 I would have to get close to a struct ~64bytes in size before seeing a performance gain in passing by reference. At struct sizes ~128 bytes I would see a significant performance gain. But relatively speaking most of the structs in the codebase seem to be < 64bytes.

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8700K CPU 3.70GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.201
  [Host]    : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT
  RyuJitX64 : .NET Core 5.0.4 (CoreCLR 5.0.421.11614, CoreFX 5.0.421.11614), X64 RyuJIT

Job=RyuJitX64  Jit=RyuJit  Platform=X64  

```
|                          Method |     Mean |    Error |   StdDev | Code Size |
|-------------------------------- |---------:|---------:|---------:|----------:|
|            TestValueLargeStruct | 582.2 μs |  2.33 μs |  1.95 μs |     327 B |
|    TestValueLargeReadOnlyStruct | 597.3 μs | 11.83 μs | 11.06 μs |     327 B |
| TestValueLargeReadOnlyStruct128 | 920.8 μs | 18.11 μs | 29.76 μs |     360 B |
|      TestRefLargeReadOnlyStruct | 562.7 μs |  1.88 μs |  1.75 μs |     255 B |
|   TestRefLargeReadOnlyStruct128 | 564.6 μs | 10.89 μs | 10.19 μs |     327 B |

TestValue = pass by value
TestRef = pass by ref (in)
LargeStruct = 64bytes
LargeStruct128 = 128bytes

I've ensured the JITed code emitted a separate function call to the method being tested instead of getting inlined, which was originally causing a wild difference.

PR #2174, opened 2021-04-04
Allow DRAM size to be increased from 4GB to 6GB

gdkchan:
This exposes an option on the UI, that allows increasing the amount of the emulated Switch memory from 4GB to 6GB. Since no retail unit has more than 4GB (afaik), I'm not sure how safe exposing extra memory to the game is (they may assume there's 4 and break if more is available). No problems have been observed so far using it, but I decided to put it under "hacks" regardless.

image

The amount of memory available for the application is increased from 3.2GB to 4.5GB (the remaining memory is reserved for the system, and with HLE, its almost entirely unused).

There is a total of 5 configurations, but from those only 2 are useful for games, the other distributes the memory differently, and gives more memory to applets or the system, which is not very useful for us, so I decided to not expose them on the UI.

This is useful for mods that requires more memory than the Switch has (like Monster Hunter Rise 4K resolution mod, for example).


Thog (approved):

riperiperi (approved):
Cool, this is rather useful. Looks like it's working fine.

PR #2177, opened 2021-04-04
Allow parallel shader compilation when loading a shader cache

riperiperi:
When your shader cache version is out of date, or your host shader cache is no longer valid for your GPU, you're subjected to the slow process of compiling all of the saved guest shaders. This PR allows programs to report their compilation process without blocking, and use the OpenGL's capability for multithreading shader compilation with `ARB_parallel_shader_compile`.

**This PR does not affect runtime shader compilation time.**

- ShaderCache Initialize now dispatches up to 8 concurrent tasks and shader compilations, and periodically checks them for completion to start the next step or add the compiled shader to the cache.
  - 8 seems to be the max level of multithreading that NVIDIA allows. We choose the minimum of 8 and the # of logical cores.
- Programs no longer check their link status on compilation.
- Uniform locations are now initialized on the first bind of a program, rather than on creation.
  - These two changes allow programs to start compilation without blocking.
- There is a new IProgram method for checking link status of a shader, optionally blocking. This is used internally when attempting to use a program.
  - If `ARB_parallel_shader_compile` is not supported, the link check will always block.

### Numbers
When rebuilding from guest with no driver cache, NVIDIA compilation is 4-5x faster on capable hardware.
AMD windows compilation is about 2x faster.
Unknown on Intel.

**Mario Kart 8:**
- No Host Cache, no Driver Cache
  - **Parallel:** 1:43.130
  - **Master:** 7:58.810
- No Host Cache, but Driver Cached
  - **Parallel:** 18.590
  - **Master:** 26.918
- Host Cache
  - **Parallel:** 8.697
  - **Master:** 9.051
  
**Splatoon 2:**
- No Host Cache, no Driver Cache
  - **Parallel:** 5:39.208
  - **Master:** 23:09.293

### Note on ShaderCache.cs
I don't like the current Initialize function, but I don't know how to fix it. I'd split it into individual steps, but there's so much shared state that capturing variables with lambdas seems like it works better than moving all the state variables to the task class, as that would be very arbitrary and weird.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -197,11 +197,6 @@ namespace Ryujinx.Graphics.Gpu.Shader
 
                                         task.OnCompiled(hostProgram, (bool isNewProgramValid, ShaderCompileTask task) =>
                                         {
-                                            if (!isNewProgramValid)
-                                            {
-                                                return true;
-                                            }
-
                                             // As the host program was invalidated, save the new entry in the cache.
                                             hostProgramBinary = HostShaderCacheEntry.Create(hostProgram.GetBinary(), new ShaderCodeHolder[] { shader });
 
@@ -365,11 +360,6 @@ namespace Ryujinx.Graphics.Gpu.Shader
 
                                         task.OnCompiled(hostProgram, (bool isNewProgramValid, ShaderCompileTask task) =>
                                         {
-                                            if (!isNewProgramValid)
-                                            {
-                                                return true;
-                                            }
-
                                             // As the host program was invalidated, save the new entry in the cache.
                                             hostProgramBinary = HostShaderCacheEntry.Create(hostProgram.GetBinary(), shaders);
 
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Ryujinx.Graphics.Gpu.Shader
 {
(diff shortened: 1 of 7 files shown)


Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -102,234 +103,322 @@ internal void Initialize()
...
 
-                        // Reconstruct code holder.
-                        for (int i = 0; i < cachedShaderEntries.Length; i++)
+                                        return false; // Not finished: still need to compile the host program.
+                                    });
+
+                                    return false; // Not finished: translating the program.
+                                }
+                            });
+
```
gdkchan, line 235: nit: Empty line.

gdkchan (commented):
One thing that I noticed (on master) is that trying to close the emulator while theres shader compilation going just makes it hang (on master). Perhaps its something that should be fixed separately, but it would be nice to at least expose some way to cancel the shader compilation tasks.

riperiperi:
> 
> 
> One thing that I noticed (on master) is that trying to close the emulator while theres shader compilation going just makes it hang (on master). Perhaps its something that should be fixed separately, but it would be nice to at least expose some way to cancel the shader compilation tasks.

It might be enough to be able to stop the dispatch loop and let the currently ongoing compilations complete, since we can't exactly stop program linking, and cancelling a task is its own bag of worms.

riperiperi:
I've added some utilities for controlling NVIDIA threaded optimisation at boot, to stop it randomly disabling under the default `Auto` setting. This was happening with increased frequency when doing a lot of parallel shader compilation, but does not disable when the setting is explicitly `On`. These also allow the disabling of threaded optimisation, which will be useful in future.

riperiperi:
Fixed a bug where deleting files from the guest cache would cause compilation to get stuck forever.

Ryujinx.Common/GraphicsDriver/DriverUtilities.cs
```diff
@@ -0,0 +1,22 @@
...
+    public static class DriverUtilities
+    {
+        public static void ToggleOGLThreading(bool enabled)
+        {
+            Environment.SetEnvironmentVariable("mesa_glthread", enabled.ToString());
+            Environment.SetEnvironmentVariable("__GL_THREADED_OPTIMIZATIONS", enabled ? "1" : "0");
+
+            try
+            {
+                NVThreadedOptimization.SetThreadedOptimization(enabled);
```
gdkchan, line 14: Can `ForceDedicatedGpu.Nvidia` be removed now? Since this will force nvapi to be loaded anyway.

Thog (approved):

gdkchan (approved):
lgtm, great improvement 👍

PR #2180, opened 2021-04-06
Inline software keyboard without input pop up dialog

Caian:
This is the draft for the proof of concept that I posted on Discord a few days ago. The idea is to use the software keyboard in inline mode without the pop up dialog, since a way to display the text is already provided by the games.

The initial implementation already supports special characters, moving the text cursor with the arrow keys of the keyboard and a custom indirect layer to show the keyboard message:

specialchars

This PR extends the `IHostUiHandler` interface to allow the creation of a `IDynamicTextInputHandler`, that reads raw data from the keyboard directly and converts it into text. Behind the scenes the key strokes are being forwarded to a custom `Gtk Entry` class.

The keyboard renderer takes advantage of the constant `A8B8G8R8` pixel format of the indirect layer to fill the bytes directly. It currently uses `System.Drawing` as a quick way to render text, but can be improved and refactored to perform more tasks and better integrate with the rendering mechanisms provided by Ryujinx.

The state machine used by the inline keyboard was also simplified, fixing a reported keyboard bug in Monster Hunter XX. This also seems to fix a soft-lock in Monster Hunter Gen U that happens when the player spams the key to show the keyboard.

The following items are still missing from the PR and will be implemented:

- Limit input text size according to Calc message;
- Initial input text according to Calc message;
- Failsafe in case of soft-lock;
- **Bug** Gnosia and Monster Hunter Rise do not update the text while typing (but it is correctly entered);

### DIFF
--- a/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/InlineResponses.cs
+++ b/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/InlineResponses.cs
@@ -38,9 +38,12 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             return (uint)text.Length; // Return the cursor position at the end of the text
         }
 
-        private static void WriteStringWithCursor(string text, BinaryWriter writer, uint maxSize, Encoding encoding)
+        private static void WriteStringWithCursor(string text, uint cursor, BinaryWriter writer, uint maxSize, Encoding encoding)
         {
-            uint cursor = WriteString(text, writer, maxSize, encoding);
+            uint length = WriteString(text, writer, maxSize, encoding);
+
+            if (cursor > length)
+                cursor = length;
 
             writer.Write(cursor); // Cursor position
         }
@@ -72,7 +75,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             }
         }
 
-        public static byte[] ChangedString(string text, InlineKeyboardState state)
+        public static byte[] ChangedString(string text, uint cursor, InlineKeyboardState state)
         {
             uint resSize = 6 * sizeof(uint) + MaxStrLenUTF16;
 
@@ -80,7 +83,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             using (BinaryWriter writer = new BinaryWriter(stream))
             {
                 BeginResponse(state, InlineKeyboardResponse.ChangedString, writer);
-                WriteStringWithCursor(text, writer, MaxStrLenUTF16, Encoding.Unicode);
+                WriteStringWithCursor(text, cursor, writer, MaxStrLenUTF16, Encoding.Unicode);
                 writer.Write((int)0); // ?
                 writer.Write((int)0); // ?
 
@@ -88,7 +91,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             }
         }
 
-        public static byte[] MovedCursor(string text, InlineKeyboardState state)
+        public static byte[] MovedCursor(string text, uint cursor, InlineKeyboardState state)
         {
             uint resSize = 4 * sizeof(uint) + MaxStrLenUTF16;
 
@@ -96,13 +99,13 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             using (BinaryWriter writer = new BinaryWriter(stream))
             {
                 BeginResponse(state, InlineKeyboardResponse.MovedCursor, writer);
-                WriteStringWithCursor(text, writer, MaxStrLenUTF16, Encoding.Unicode);
+                WriteStringWithCursor(text, cursor, writer, MaxStrLenUTF16, Encoding.Unicode);
 
                 return stream.ToArray();
             }
         }
 
-        public static byte[] MovedTab(string text, InlineKeyboardState state)
+        public static byte[] MovedTab(string text, uint cursor, InlineKeyboardState state)
         {
             // Should be the same as MovedCursor.
 
@@ -112,7 +115,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             using (BinaryWriter writer = new BinaryWriter(stream))
             {
                 BeginResponse(state, InlineKeyboardResponse.MovedTab, writer);
-                WriteStringWithCursor(text, writer, MaxStrLenUTF16, Encoding.Unicode);
+                WriteStringWithCursor(text, cursor, writer, MaxStrLenUTF16, Encoding.Unicode);
 
                 return stream.ToArray();
             }
@@ -145,7 +148,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             }
         }
 
-        public static byte[] ChangedStringUtf8(string text, InlineKeyboardState state)
+        public static byte[] ChangedStringUtf8(string text, uint cursor, InlineKeyboardState state)
         {
             uint resSize = 6 * sizeof(uint) + MaxStrLenUTF8;
 
@@ -153,7 +156,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             using (BinaryWriter writer = new BinaryWriter(stream))
             {
                 BeginResponse(state, InlineKeyboardResponse.ChangedStringUtf8, writer);
-                WriteStringWithCursor(text, writer, MaxStrLenUTF8, Encoding.UTF8);
+                WriteStringWithCursor(text, cursor, writer, MaxStrLenUTF8, Encoding.UTF8);
                 writer.Write((int)0); // ?
                 writer.Write((int)0); // ?
 
@@ -161,7 +164,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             }
         }
 
-        public static byte[] MovedCursorUtf8(string text, InlineKeyboardState state)
+        public static byte[] MovedCursorUtf8(string text, uint cursor, InlineKeyboardState state)
         {
             uint resSize = 4 * sizeof(uint) + MaxStrLenUTF8;
 
@@ -169,7 +172,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             using (BinaryWriter writer = new BinaryWriter(stream))
             {
                 BeginResponse(state, InlineKeyboardResponse.MovedCursorUtf8, writer);
-                WriteStringWithCursor(text, writer, MaxStrLenUTF8, Encoding.UTF8);
+                WriteStringWithCursor(text, cursor, writer, MaxStrLenUTF8, Encoding.UTF8);
 
                 return stream.ToArray();
             }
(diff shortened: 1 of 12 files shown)


Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/InlineResponses.cs
```diff
@@ -38,9 +38,12 @@ private static uint WriteString(string text, BinaryWriter writer, uint maxSize,
...
         }
 
-        private static void WriteStringWithCursor(string text, BinaryWriter writer, uint maxSize, Encoding encoding)
+        private static void WriteStringWithCursor(string text, uint cursor, BinaryWriter writer, uint maxSize, Encoding encoding)
         {
-            uint cursor = WriteString(text, writer, maxSize, encoding);
+            uint length = WriteString(text, writer, maxSize, encoding);
+
+            if (cursor > length)
+                cursor = length;
```
AcK77, line 46: Adds brackets

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardApplet.cs
```diff
@@ -375,159 +378,156 @@ private void OnBackgroundInteractiveData(byte[] data)
                             {
                                 _encoding = Encoding.UTF8;
                             }
+                        }
+                        // Make the state transition.
```
AcK77, line 382: Add a new empty line before

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardApplet.cs
```diff
@@ -375,159 +378,156 @@ private void OnBackgroundInteractiveData(byte[] data)
...
+                                Logger.Debug?.Print(LogClass.ServiceAm, $"Calc during state {state} is probably a dummy");
+                            }
+                            else
                             {
-                                shouldShowKeyboard = true;
+                                string newText = _keyboardBackgroundCalc.InputText;
+                                uint cursorPosition = (uint)_keyboardBackgroundCalc.CursorPos;
+                                _dynamicTextInputHandler?.SetText(newText);
+
+                                // TODO(Caian): Update the text in the with the proper response
```
AcK77, line 398: `in the` what ? :)

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardRenderer.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                // Fill the keyboard area with a transparent black color.
+                SolidBrush backgroundBrush = new SolidBrush(Color.FromArgb(120, 0, 0, 0));
+                gfx.FillRectangle(backgroundBrush, keyboardRectangle);
+
+                string line1Text = "Please use the Keyboard to input text...";
+                string line2Text = "Press Return to accept or Page Up to cancel.";
+
+                // Draw the first line larger then the second.
+                System.Drawing.Font line1Font = new System.Drawing.Font("Tahoma", 40);
```
AcK77, line 65: What if the font doesn't exists on the system ?
(I think the text sizes are too huge too)

Ryujinx.HLE/HOS/Services/Vi/RootService/IApplicationDisplayService.cs
```diff
@@ -239,18 +241,45 @@ public ResultCode ConvertScalingMode(ServiceCtx context)
             return null;
         }
 
+        private ulong GetA8B8G8R8LayerSize(int width, int height, out int pitch, out int alignment)
+        {
+            const int   defaultAlignment = 0x1000;
+            const ulong defaultSize = 0x20000;
```
AcK77, line 247:
```suggestion
            const ulong defaultSize      = 0x20000;
```

Ryujinx.HLE/HOS/Services/Vi/RootService/IApplicationDisplayService.cs
```diff
@@ -239,18 +241,45 @@ public ResultCode ConvertScalingMode(ServiceCtx context)
...
+
+            alignment = defaultAlignment;
+            pitch     = BitUtils.AlignUp(BitUtils.DivRoundUp(width * 32, 8), 64);
+
+            int memorySize           = pitch * BitUtils.AlignUp(height, 64);
```
AcK77, line 252:
```suggestion
            int   memorySize         = pitch * BitUtils.AlignUp(height, 64);
```

Ryujinx.HLE/HOS/Services/Vi/RootService/IApplicationDisplayService.cs
```diff
@@ -239,18 +241,45 @@ public ResultCode ConvertScalingMode(ServiceCtx context)
...
 
+            long layerWidth        = context.RequestData.ReadInt64();
+            long layerHeight       = context.RequestData.ReadInt64();
+            long layerHandle       = context.RequestData.ReadInt64();
             long layerBuffPosition = context.Request.ReceiveBuff[0].Position;
             long layerBuffSize     = context.Request.ReceiveBuff[0].Size;
 
-            // Fill the layer with zeros.
-            context.Memory.Fill((ulong)layerBuffPosition, (ulong)layerBuffSize, 0x00);
+            // TODO(Caian): Support other handles other than keyboard.
```
AcK77, line 271: I'm pretty sure the `handle` should be something related to the applet.

Ryujinx.HLE/HOS/Services/Vi/RootService/IApplicationDisplayService.cs
```diff
@@ -239,18 +241,45 @@ public ResultCode ConvertScalingMode(ServiceCtx context)
...
+
+            ulong size = GetA8B8G8R8LayerSize((int)layerWidth, (int)layerHeight, out int pitch, out int alignment);
+
+            Debug.Assert(layerBuffSize == (long)size);
+
+            Span<byte> graphics = SoftwareKeyboardRenderer.GetGraphicsA8B8G8R8((int)layerWidth, (int)layerHeight, pitch, (int)layerBuffSize);
```
AcK77, line 277: Maybe you should check the `handle` to be sure you should returns something related to the SoftwareKeyboard Applet ?

Ryujinx.HLE/HostTextInputChangedHandler.cs
```diff
@@ -0,0 +1,7 @@
+using Ryujinx.HLE.HOS.Applets;
+using Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.ApplicationProxy.Types;
+
+namespace Ryujinx.HLE
```
AcK77, line 4: This should be in the SoftwareKeyboard Applet folder maybe ?

Ryujinx.HLE/IDynamicTextInputHandler.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Ryujinx.HLE
```
AcK77, line 3: This should be in the SoftwareKeyboard Applet folder maybe ?

Ryujinx/Ui/Applet/GtkDynamicTextInputHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [GLib.ConnectBefore()]
+        private void HandleKeyPressEvent(object o, KeyPressEventArgs args)
+        {
+            if (args.Event.Key == Gdk.Key.Return)
+            {
+                InvokeTextChanged(true, false);
+                _inputToTextEntry.Text = "";
+            }
+            else if (args.Event.Key == Gdk.Key.Page_Up) // TODO other key?
```
AcK77, line 29: Needs to be determined.

AcK77 (changes requested):
Awesome works, but there are some edge cases which have to be handled properly :)

Caian:
@AcK77 The `handle` value is the same for all applets because it is hard coded in `ILibraryAppletAccessor::GetIndirectLayerConsumerHandle` and there is no interface for them to exchange this information.

I had some ideas about this:

Global handle identifiers: 

Applets expose a handle value associated with its type, from a global enum, i.e. `AppletKindHandle.Keyboard`, so `Vi` can lookup and dispatch to static renderers. 

This would be the easiest approach as I can create an interface for such applets with a single function `GetIndirectLayerConsumerHandle`, then `ILibraryAppletAccessor` can check with the `as` operation if the IApplet object also implements this other interface.

The drawback is that the rendering is static and this can lead to a bad programming pattern where the applet instances send the commands to the renderer directly, if there are multiple instances of the same kind of applet then it would introduce concurrency issues. I also think that probing objects for its interface shades the relation between classes.

Global handle identifiers++:

I turn the interface `IApplet` into an abstract class `AppletBase` to include a default implementation of the aforementioned function that returns `AppletKindHandle.Invalid` (or throws an exception if it makes sense). The same concurrency problems exist, but the interface is clearer.

Applet IPC:

The handle is unique and point to a specific instance of the applet. This would be far more complicated to do and there is also the question of whether it makes sense to represent an instance of an applet with its indirect layer handle.

Ryujinx.HLE/HostTextInputChangedHandler.cs
Caian: I imagined it as something that any applet could use, like IHostUiHandler.

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardRenderer.cs
Caian: The appearance of the layer is still an open topic

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardRenderer.cs
```diff
@@ -0,0 +1,718 @@
...
+            _padPressedPen       = new Pen(borderColor, _padPressedPenWidth);
+
+            _inputTextFontSize = 20;
+            _padButtonFontSize = 24;
+
+            string font = uiTheme.FontFamily;
+
+            _messageFont     = new System.Drawing.Font(font, 26                , FontStyle.Regular, GraphicsUnit.Pixel);
+            _inputTextFont   = new System.Drawing.Font(font, _inputTextFontSize, FontStyle.Regular, GraphicsUnit.Pixel);
+            _labelsTextFont  = new System.Drawing.Font(font, 24                , FontStyle.Regular, GraphicsUnit.Pixel);
```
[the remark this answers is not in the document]
AcK77: As you want but in your code the extra spaces should be after the comma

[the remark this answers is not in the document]
Caian: Thanks @AcK77 , done.

Caian:
Changed the shortcut that cycles through input modes to make it work with #2428

Caian:
Merged master

Thog (approved):
lgtm

PR #2183, opened 2021-04-07
(CPU) Fix CRC32 instruction when constant values are used as input

### LINKED ISSUE #1457: Intrinsics & ConstantRegCopy
Currently the PreAllocator does not handle constant operands on intrinsics and this can be problematic when optimizations are applied. I ran into this by forcing HCQ in the Translator and running the tests, some of the CRC32 tests fails.

Here is a repro.
```csharp
public static void Test()
{
  static ControlFlowGraph GetCFG()
  {
    var emitter = new EmitterContext();

    var op0 = emitter.Copy(Const(0L)); // These gets copy propagated into their use in HCQ. i.e, res.
    var op1 = emitter.Copy(Const(0L));

    var res = emitter.AddIntrinsicLong(Intrinsic.X86Crc32, op0, op1);

    emitter.Return(res);

    return emitter.GetControlFlowGraph();
  }

  var cfg0 = GetCFG();
  var cfg1 = GetCFG();

  var argTypes = new OperandType[] { OperandType.I64 };
  var retType  = OperandType.I64;

  Compiler.Compile<GuestFunction>(cfg0, argTypes, retType, CompilerOptions.None);   // Ok
  Compiler.Compile<GuestFunction>(cfg1, argTypes, retType, CompilerOptions.HighCq); // Fail
}
```

If we get unlucky, the 1st operand becomes a constant and the instruction is silently incorrectly encoded when the 2nd operand is a register (asserts fires in Debug builds). If the 2nd operand becomes a constant an exception in the Assembler is thrown.

I believe most use of intrinsics in the code do not run into this issue because they work on V128 which cannot be constants. Or work on I32/64/FP32/64 which have been extracted from a V128 and so these also cannot be constants.

gdkchan:
There is a bug with the CRC32 intrinsic were the JIT does not force a copy when a constant value is used as input for this instruction. On debug, this will cause an assert as this specific instruction does not support immediate operands on x86, and in release, it will just generate invalid code. This fixes the issue by forcing the value to be copied to a register, and then replacing the operation input with the register.

Fixes #1457

A user ("yunny") reported the assert firing while rebuilding PPTC cache for Monster Hunter Rise, I have reproduced the issue and verified that the assert no longer fires with the fix, however, I also did not notice any visible improvement on the game.

I also ran all CRC32 tests locally in both LCQ and HCQ modes, and they all pass.

### DIFF
--- a/ARMeilleure/CodeGen/X86/PreAllocator.cs
+++ b/ARMeilleure/CodeGen/X86/PreAllocator.cs
@@ -135,7 +135,7 @@ namespace ARMeilleure.CodeGen.X86
 
         private static void HandleConstantRegCopy(IntrusiveList<Node> nodes, Node node, Operation operation)
         {
-            if (operation.SourcesCount == 0 || IsIntrinsic(operation.Instruction))
+            if (operation.SourcesCount == 0 || IsXmmIntrinsic(operation))
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Nice catch.

LDj3SNuD (approved):

PR #2189, opened 2021-04-09
Added progress reporter for loading existing translations from PTC

yunnypuff:
- Refactored existing progress reporter in PTC
- Added reporting progress when loading existing translations from PTC

### DIFF
--- a/ARMeilleure/Translation/PTC/Ptc.cs
+++ b/ARMeilleure/Translation/PTC/Ptc.cs
@@ -66,9 +66,6 @@ namespace ARMeilleure.Translation.PTC
 
         internal static PtcState State { get; private set; }
 
-        // Progress reporting helpers
-        private static volatile int _translateCount;
-        private static volatile int _translateTotalCount;
         public static event Action<PtcLoadingState, int, int> PtcStateChanged;
 
         static Ptc()
@@ -527,12 +524,17 @@ namespace ARMeilleure.Translation.PTC
             _relocsStream.Seek(0L, SeekOrigin.Begin);
             _unwindInfosStream.Seek(0L, SeekOrigin.Begin);
 
+            // Get total entries to process for progress reporting and for looping.
+            int infoEntriesCount = GetInfosEntriesCount();
+
+            using ProgressReporter progressReporter = ProgressReporter.CreateAndStart(infoEntriesCount, PtcStateChanged);
+
             using (BinaryReader infosReader = new(_infosStream, EncodingCache.UTF8NoBOM, true))
             using (BinaryReader codesReader = new(_codesStream, EncodingCache.UTF8NoBOM, true))
             using (BinaryReader relocsReader = new(_relocsStream, EncodingCache.UTF8NoBOM, true))
             using (BinaryReader unwindInfosReader = new(_unwindInfosStream, EncodingCache.UTF8NoBOM, true))
             {
-                for (int i = 0; i < GetInfosEntriesCount(); i++)
+                for (int i = 0; i < infoEntriesCount; i++)
                 {
                     InfoEntry infoEntry = ReadInfo(infosReader);
 
@@ -570,6 +572,8 @@ namespace ARMeilleure.Translation.PTC
                         StubReloc(infoEntry.RelocEntriesCount);
                         StubUnwindInfo(unwindInfosReader);
                     }
+
+                    progressReporter.Increment();
                 }
             }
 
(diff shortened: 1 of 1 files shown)


ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -986,5 +949,103 @@ public static void Dispose()
                 DisposeMemoryStreams();
             }
         }
+
+        /// <summary>
+        /// Class for reporting PTC translation / loading item progress to the UI
+        /// </summary>
+        private class ProgressReporter : IDisposable
```
Thog, line 956: Would be better to move that to the Ryujinx project and provide an interface so we remove this from ARMeilleure (This will reduce intrusive changes that we made on ARMeilleure for progress bars)

Thog (changes requested):

ARMeilleure/Translation/PTC/Ptc.cs
yunnypuff: Any recommended location for the Interface and the class itself?

Perhaps under Ryujin.Common.Utilities or System?

I think no matter what we do the minimum we need is to have some kind of event handler delegate that Ptc can report progress to. Are the UI updates from the event delegate being marshalled back onto the main UI thread? If so then perhaps there isn't even a need for a background progress reporter thread.

In terms of the interface -- we only want the progress reporter when Ptc is engaged in code sections with reportable progress, really. So at some point Ptc will need to instantiate the ProgressReporter -- which forces a concrete dependency on the type. Or we can make some kind of IProvider interface too.

Start-stop pattern?
```
interface IBackgroundProgressReporter
{
  void Start();
  void Increment(int progress);
  void Stop(); // Or do we prefer the Dispose pattern?
}

void Foo()
{
  var reporter = new ProgressReporter(); // whatever type we choose
  reporter.Start();
  foreach ()
  {
    reporter.Increment();
  }
  reporter.Stop(); // if we forget to call Stop() then our impl might have a background thread running forever.
                           // so maybe reason to put a finalizer in
}
```

Or Disposable pattern?
```
interface IBackgroundProgressReporter : IDisposable
{
  void Start();
  void Increment(int progress);
}

void Foo()
{
  using (var reporter = new ProgressReporter())
  {
    reporter.Start();
    foreach ()
    {
      reporter.Increment();
    }
  } // As long as we properly use "using" we should be ok.
}
```
Lastly, I'm thinking I'll take a look at usages of other progress bars (at the moment the obvious ones are Shader and Ptc) and see if I can provide something more universal. Let me know if you  are aware of other progress-bar like features that could share the same interface or behaviors.

yunnypuff: I did a little more digging and feel like we have a few options. All of them require PTC (or Shader) to know that some system may be interested in its progress, but the exact connection and responsibilities are divided differently. It's important to note that none of the options introduce *tight coupling* between MainWindow and PTC.

Option A
1. Ptc knows that someone might be interested in its progress (via the event delegate), 
2. Ptc knows that internally it has lots of worker threads who can report progress from any number of those threads, so it internally steps in front of potential concurrency issues
3. Ptc knows that whoever is interested in the progress likely isn't interested in every unit progress made, so it reports at some interval it pleases

This is basically what we've got now and what this PR is just a minor refactor of.
I'd argue maybe point 3 maybe outside of the responsibility of Ptc, but in order to take care of point 2, we can take care of point 3 trivially.

Option B
1. Ptc knows that someone might be interested in its progress (via the event delegate), 
2. Ptc knows that internally it has lots of worker threads who can report progress, but it's just gonna call the event delegate every time any event is worth reporting.
3. MainWindow needs to know that the event delegate could be called from any thread, and may need to be concurrency-safe.
4. MainWindow knows it doesn't need to know about every bit of progress, so it can choose to update the UI at the interval it chooses.

I'm not a huge fan of point 3 here, but the other bits sound ok

Option C
1. Ptc knows that someone might be interested in its progress via some interface provider. It will use the provider to instantiate a type and report to it when it expects to do a bunch of work worth reporting.
2. ProgressReporter is an middle-man type that can be made thread-safe and the appropriate delegates are installed by the MainWindow. It'll have to ensure the delegates are called on the main UI thread.
3. ProgressReporter can be customized by the MainWindow to report at intervals MainWindow pleases
4. MainWindow only needs to know about installing the correct ProgressReporterProvider to Ptc.

This method has the potential of producing the most re-usable components, but probably the most work. In the end Ptc remains just as coupled with the fact that someone may be interested in knowing about its progress.

Any preferences here?

Issue #2158, opened 2021-03-31
Fast CPU emulator JIT function lookup


gdkchan:
#### Overview:
The CPU emulator works by recompiling Arm code into x86 code that can run on the target platform. Evidently, the code needs to be able to quickly find the respective x86 function for a given Arm function on the in-memory JIT cache. Currently this uses a *dictionary* on the `Translator` class, to find those functions. This is reasonably fast, but not as fast as it could be. Furthermore, the dictionary being a managed object, and only being accessible through managed functions, means it can't be directly accessed from the native x86 JIT generated code.

Getting the x86 function pointer for a given Arm function thus requires calling `NativeInterface.GetIndirectFunctionAddress` and `NativeInterface.GetFunctionAddressWithHint`. Doing so is actually *very slow*, because it requires native <-> managed transitions (this is a cost paid every time native code is called from managed, and vice-versa). On top of that, we pay the cost of a bunch more calls and dictionary lookup, which is not actually that bad, but can become a issue if this is called often.

In order to minimise the cost of doing this, the translator will attempt to cache those function pointer in a small jump table, a small region of memory reserved for pointers. They work like so:
- For direct calls (when the target address is constant, like when the `BL` Arm instruction is used), it reserves one entry on the table that will hold the pointer, but is initially 0.  The JIT generated code checks if the value there is 0, if so, it calls `GetFunctionAddress`, stores the pointer there, and calls it, otherwise it just calls the loaded pointer directly. This avoids the managed call cost on the 2nd run onwards.
- For indirect calls (when the target address is unknown and comes from a register, like when the `BLR` Arm instruction is used), it also reserves one entry on the dynamic table, which holds an address pair. The first address is the Arm function address, while the second one is the host function address. Since it only reserves one entry, it can only possibly cache a single address, so if multiple functions address are called from the same call site, it will still be doing a lot of managed calls to `GetIndirectFunctionAddress`, which will hurt performance.

Here, I propose a solution that not only greatly simplifies all this logic, but also improves performance by addressing its shortcomings.

#### Smaller Details: 
The proposed change replaces the dictionary that is currently used for function lookup with a multi-level table. Hash tables are O(1) in the best case, and O(n) in the worst case (if the hash function is extremely poor and all elements ends being shoved in the same bucket), meanwhile we have constant O(1) complexity with the multi-level table approach. However, the biggest benefit here is that the multi-level table can be **easily accessed from native, JIT generated code, without requiring any loops**.

[...]

PR #2190, opened 2021-04-09
Add inlined on translation call counting

FICTURE7:
Contributes to #2158.

* Add re-usable structures to do on translation counting ~(might want to make `Counter` a generic though)~.
* Add on translation call counting.
* Change call counting behavior to be associated with nodes (i.e head of translations) instead of edges (i.e out of translation branches). Before if all incoming branches to a translation was not hinted for rejit, it would never tier up even though it could be hot.
  * **Pros**: Improves hot code discovery.
  * **Cons**: Increase in code size. -- Because we may discover more code to rejit, and the decoder may produce a far larger output for HCQ than for LCQ.

PPTC integration is done this way for the sake of simplicity and because it allows `EntryTable` to be dynamic/expandable. ~although it currently is not.~

Call counting is not implemented as an atomic operation currently, so `EnqueueForRejit` needs to ensure the rejit queue contains unique requests so as to not waste time compiling the same stuff. I held back from implementing an IR instruction for this because I figure it would require some discussion to model it in IR, and I am not entirely if its worth. If you think so let me know :>

### DIFF
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -1,0 +1,48 @@
+﻿namespace ARMeilleure.Common
+{
+    /// <summary>
+    /// Represents an 8-bit counter.
+    /// </summary>
+    class Counter
+    {
+        private readonly int _index;
+        private readonly EntryTable<byte> _countTable;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Counter"/> class from the specified
+        /// <see cref="EntryTable{byte}"/> instance and index.
+        /// </summary>
+        /// <param name="countTable"><see cref="EntryTable{byte}"/> instance</param>
+        /// <param name="index">Index in the <see cref="EntryTable{TEntry}"/></param>
+        private Counter(EntryTable<byte> countTable, int index)
+        {
+            _countTable = countTable;
+            _index = index;
+        }
+
+        /// <summary>
+        /// Gets a reference to the value of the counter.
+        /// </summary>
+        public ref byte Value => ref _countTable.GetValue(_index);
+
+        /// <summary>
+        /// Tries to create a <see cref="Counter"/> instance from the specified <see cref="EntryTable{byte}"/> instance.
+        /// </summary>
+        /// <param name="countTable"><see cref="EntryTable{TEntry}"/> from which to create the <see cref="Counter"/></param>
+        /// <param name="counter"><see cref="Counter"/> instance if success; otherwise <see langword="null"/></param>
+        /// <returns><see langword="true"/> if success; otherwise <see langword="false"/></returns>
+        public static bool TryCreate(EntryTable<byte> countTable, out Counter counter)
+        {
+            if (countTable.TryAllocate(out int index))
+            {
+                counter = new Counter(countTable, index);
+
+                return true;
+            }
+
+            counter = null;
+
+            return false;
+        }
+    }
+}
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -40,7 +40,7 @@ namespace ARMeilleure.Common
         /// Gets a reference to the value of the counter.
         /// </summary>
         /// <exception cref="ObjectDisposedException"><see cref="Counter{T}"/> instance was disposed</exception>
-        public ref T Value 
+        public ref T Value
         {
             get
             {
@@ -49,7 +49,7 @@ namespace ARMeilleure.Common
                     throw new ObjectDisposedException(null);
                 }
 
-                return ref _countTable.GetValue(_index); 
+                return ref _countTable.GetValue(_index);
             }
         }
 
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -6,8 +6,9 @@ namespace ARMeilleure.Common
     /// Represents a numeric counter.
     /// </summary>
     /// <typeparam name="T">Type of the counter</typeparam>
-    class Counter<T> where T : unmanaged
+    class Counter<T> : IDisposable where T : unmanaged
     {
+        private bool _disposed;
         private readonly int _index;
         private readonly EntryTable<T> _countTable;
 
@@ -26,7 +27,18 @@ namespace ARMeilleure.Common
         /// <summary>
         /// Gets a reference to the value of the counter.
         /// </summary>
-        public ref T Value => ref _countTable.GetValue(_index);
+        public ref T Value 
+        {
+            get
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(null);
+                }
+
+                return ref _countTable.GetValue(_index); 
+            }
+        }
 
         /// <summary>
         /// Tries to create a <see cref="Counter"/> instance from the specified <see cref="EntryTable{byte}"/> instance.
@@ -63,6 +75,19 @@ namespace ARMeilleure.Common
             counter = null;
 
             return false;
+        }
+
+        /// <summary>
+        /// Releases all resources used by the <see cref="Counter{T}"/> instance.
+        /// </summary>
+        public void Dispose()
+        {
+            if (!_disposed)
+            {
+                _countTable.Free(_index);
+
+                _disposed = true;
+            }
         }
     }
 }
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -3,7 +3,7 @@
 namespace ARMeilleure.Common
 {
     /// <summary>
-    /// Represents a numeric counter.
+    /// Represents a numeric counter which can be used for instrumentation of compiled code.
     /// </summary>
     /// <typeparam name="T">Type of the counter</typeparam>
     class Counter<T> : IDisposable where T : unmanaged
@@ -27,6 +27,7 @@ namespace ARMeilleure.Common
         /// <summary>
         /// Gets a reference to the value of the counter.
         /// </summary>
+        /// <exception cref="ObjectDisposedException"><see cref="Counter{T}"/> instance was disposed</exception>
         public ref T Value 
         {
             get
@@ -41,10 +42,10 @@ namespace ARMeilleure.Common
         }
 
         /// <summary>
-        /// Tries to create a <see cref="Counter"/> instance from the specified <see cref="EntryTable{T}"/> instance.
+        /// Tries to create a <see cref="Counter{T}"/> instance from the specified <see cref="EntryTable{T}"/> instance.
         /// </summary>
-        /// <param name="countTable"><see cref="EntryTable{T}"/> from which to create the <see cref="Counter"/></param>
-        /// <param name="counter"><see cref="Counter"/> instance if success; otherwise <see langword="null"/></param>
+        /// <param name="countTable"><see cref="EntryTable{T}"/> from which to create the <see cref="Counter{T}"/></param>
+        /// <param name="counter"><see cref="Counter{T}"/> instance if success; otherwise <see langword="null"/></param>
         /// <returns><see langword="true"/> if success; otherwise <see langword="false"/></returns>
         /// <exception cref="ArgumentNullException"><paramref name="countTable"/> is <see langword="null"/></exception>
         /// <exception cref="ArgumentException"><typeparamref name="T"/> is unsupported</exception>
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -42,6 +42,39 @@ namespace ARMeilleure.Common
         }
 
         /// <summary>
+        /// Releases all resources used by the <see cref="Counter{T}"/> instance.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Releases all unmanaged and optionally managed resources used by the <see cref="Counter{T}"/> instance.
+        /// </summary>
+        /// <param name="disposing"><see langword="true"/> to dispose managed resources also; otherwise just unmanaged resouces</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!_disposed)
+            {
+                // The index into the EntryTable is essentially an unmanaged resource since we allocate and free the
+                // resource ourselves.
+                _countTable.Free(_index);
+
+                _disposed = true;
+            }
+        }
+
+        /// <summary>
+        /// Frees resources used by <see cref="Counter{T}"/> instance.
+        /// </summary>
+        ~Counter()
+        {
+            Dispose(false);
+        }
+
+        /// <summary>
         /// Tries to create a <see cref="Counter{T}"/> instance from the specified <see cref="EntryTable{T}"/> instance.
         /// </summary>
         /// <param name="countTable"><see cref="EntryTable{T}"/> from which to create the <see cref="Counter{T}"/></param>
@@ -76,19 +109,6 @@ namespace ARMeilleure.Common
             counter = null;
 
             return false;
-        }
-
-        /// <summary>
-        /// Releases all resources used by the <see cref="Counter{T}"/> instance.
-        /// </summary>
-        public void Dispose()
-        {
-            if (!_disposed)
-            {
-                _countTable.Free(_index);
-
-                _disposed = true;
-            }
         }
     }
 }
--- a/ARMeilleure/Common/Counter.cs
+++ b/ARMeilleure/Common/Counter.cs
@@ -1,20 +1,23 @@
-﻿namespace ARMeilleure.Common
+﻿using System;
+
+namespace ARMeilleure.Common
 {
     /// <summary>
-    /// Represents an 8-bit counter.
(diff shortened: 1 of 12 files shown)


ARMeilleure/Common/Counter.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    throw new ObjectDisposedException(null);
+                }
+
+                return ref _countTable.GetValue(_index); 
+            }
+        }
+
+        /// <summary>
+        /// Tries to create a <see cref="Counter"/> instance from the specified <see cref="EntryTable{byte}"/> instance.
```
gdkchan, line 44:
```suggestion
        /// Tries to create a <see cref="Counter"/> instance from the specified <see cref="EntryTable{T}"/> instance.
```

ARMeilleure/Common/Counter.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        /// <param name="countTable"><see cref="EntryTable{TEntry}"/> from which to create the <see cref="Counter"/></param>
```
gdkchan, line 46:
```suggestion
        /// <param name="countTable"><see cref="EntryTable{T}"/> from which to create the <see cref="Counter"/></param>
```

ARMeilleure/Common/EntryTable.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// </summary>
+    /// <typeparam name="TEntry">Type of the entry in the table</typeparam>
+    class EntryTable<TEntry> where TEntry : unmanaged
+    {
+        private int _freeHint;
+        private readonly TEntry[] _table;
+        private readonly BitMap _allocated;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EntryTable{TValue}"/> class with the specified capacity.
```
gdkchan, line 17:
```suggestion
        /// Initializes a new instance of the <see cref="EntryTable{TEntry}"/> class with the specified capacity.
```

ARMeilleure/Common/EntryTable.cs
```diff
@@ -0,0 +1,100 @@
...
+                throw new ArgumentOutOfRangeException(nameof(capacity));
+            }
+
+            _freeHint = 0;
+            _allocated = new BitMap();
+            _table = GC.AllocateArray<TEntry>(capacity, pinned: true);
+        }
+
+        /// <summary>
+        /// Tries to allocate an entry in the <see cref="EntryTable{TValue}"/>. Returns <see langword="true"/> if
```
gdkchan, line 34:
```suggestion
        /// Tries to allocate an entry in the <see cref="EntryTable{TEntry}"/>. Returns <see langword="true"/> if
```

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -701,15 +726,16 @@ private static UnwindInfo ReadUnwindInfo(BinaryReader unwindInfosReader)
             return new UnwindInfo(pushEntries, prologueSize);
         }
 
-        private static TranslatedFunction FastTranslate(ReadOnlySpan<byte> code, ulong guestSize, UnwindInfo unwindInfo, bool highCq)
+        private static TranslatedFunction FastTranslate(ReadOnlySpan<byte> code, Counter<uint> callCounter, ulong guestSize,
```
gdkchan, line 729: We usually put each argument on a separate line when it becomes too long, rather than spliting it like that.

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -753,7 +779,8 @@ private static void StubUnwindInfo(BinaryReader unwindInfosReader)
             }
         }
 
-        internal static void MakeAndSaveTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory, JumpTable jumpTable)
+        internal static void MakeAndSaveTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory,
```
gdkchan, line 782: Same comment as above.

ARMeilleure/Translation/Translator.cs
```diff
@@ -22,36 +22,47 @@ namespace ARMeilleure.Translation
...
 
             _funcs = new ConcurrentDictionary<ulong, TranslatedFunction>();
             _oldFuncs = new ConcurrentQueue<KeyValuePair<ulong, IntPtr>>();
 
+            _backgroundSet = new ConcurrentDictionary<ulong, object>();
             _backgroundStack = new ConcurrentStack<RejitRequest>();
             _backgroundTranslatorEvent = new AutoResetEvent(false);
             _backgroundTranslatorLock = new ReaderWriterLock();
 
+            CountTable = new EntryTable<uint>(capacity: 4 * 1024 * 1024);
```
gdkchan, line 64: I'd prefer if a constant was defined for `4 * 1024 * 1024` on the top of the class.

ARMeilleure/Translation/Translator.cs
```diff
@@ -80,6 +100,28 @@ private void TranslateStackedSubs()
...
+                    {
+                        var now = Stopwatch.GetTimestamp();
+
+                        if (now < nextUpdate)
+                        {
+                            _nextUpdate = nextUpdate;
+                        }
+                        else
+                        {
+                            Ryujinx.Common.Logging.Logger.Info?.Print(
```
gdkchan, line 115: Not sure if logging rejit requests is desirable, especially as "Info". It can certainly be useful, but I don't think its something we'd want to log by default.

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -701,15 +726,16 @@ private static UnwindInfo ReadUnwindInfo(BinaryReader unwindInfosReader)
             return new UnwindInfo(pushEntries, prologueSize);
         }
 
-        private static TranslatedFunction FastTranslate(ReadOnlySpan<byte> code, ulong guestSize, UnwindInfo unwindInfo, bool highCq)
+        private static TranslatedFunction FastTranslate(ReadOnlySpan<byte> code, Counter<uint> callCounter, ulong guestSize,
+            UnwindInfo unwindInfo, bool highCq)
```
FICTURE7, line 730:
```suggestion
        private static TranslatedFunction FastTranslate(
            ReadOnlySpan<byte> code,
            Counter<uint> callCounter,
            ulong guestSize,
            UnwindInfo unwindInfo,
            bool highCq)
```

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -753,7 +779,8 @@ private static void StubUnwindInfo(BinaryReader unwindInfosReader)
             }
         }
 
-        internal static void MakeAndSaveTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory, JumpTable jumpTable)
+        internal static void MakeAndSaveTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory,
+            JumpTable jumpTable, EntryTable<uint> countTable)
```
FICTURE7, line 783:
```suggestion
        internal static void MakeAndSaveTranslations(
            ConcurrentDictionary<ulong, TranslatedFunction> funcs,
            IMemoryManager memory,
            JumpTable jumpTable,
            EntryTable<uint> countTable)
```

ARMeilleure/Translation/PTC/Ptc.cs
```diff
@@ -513,7 +515,8 @@ private static void WriteHeader(Stream stream)
             }
         }
 
-        internal static void LoadTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory, JumpTable jumpTable)
+        internal static void LoadTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory,
+            JumpTable jumpTable, EntryTable<uint> countTable)
```
FICTURE7, line 519:
```suggestion
        internal static void LoadTranslations(ConcurrentDictionary<ulong, TranslatedFunction> funcs, IMemoryManager memory,
            JumpTable jumpTable, EntryTable<uint> countTable)
```
```suggestion
        internal static void LoadTranslations(
            ConcurrentDictionary<ulong, TranslatedFunction> funcs, 
            IMemoryManager memory,
            JumpTable jumpTable,
            EntryTable<uint> countTable)
```

ARMeilleure/Translation/Translator.cs
```diff
@@ -22,36 +22,47 @@ namespace ARMeilleure.Translation
 {
     public class Translator
     {
+        private long _nextUpdate;
```
FICTURE7, line 25:
```suggestion
        private const int CountTableCapacity = 4 * 1024 * 1024;

        private long _nextUpdate;
```

ARMeilleure/Translation/Translator.cs
FICTURE7, line 64:
```suggestion
            CountTable = new EntryTable<uint>(CountTableCapacity);
```

ARMeilleure/Translation/Translator.cs
FICTURE7: I decided to remove it, there are probably better ways to instrument this with event counters.

gdkchan (approved):
lgtm, thanks!

gdkchan:
Tests are failing, apparently the changes are causing a hard crash when they are executed. I recommend trying to run them locally.

FICTURE7:
It was not disposing the `Counter<T>` instances and since they had finalizers, they would survive a gen0 collection and so would the `EntryTable<T>`. This meant that memory would fluctuate a lot more during the tests because `Translator` instances (and thus `EntryTable<T>`) are constantly allocated, and the `EntryTable<T>` instances would live up to at least gen1. I guess the windows CI did not like this fluctuation, since I could not reproduce this once even with server GC toggled on.

ARMeilleure/Translation/Translator.cs
```diff
@@ -368,29 +393,46 @@ private static ControlFlowGraph EmitAndGetCFG(ArmEmitterContext context, Block[]
...
+        internal static void EmitRejitCheck(ArmEmitterContext context, out Counter<uint> counter)
         {
-            long countOffs = NativeContext.GetCounterOffset();
+            const int MinsCallForRejit = 100;
 
-            Operand countAddr = context.Add(context.LoadArgument(OperandType.I64, 0), Const(countOffs));
+            if (!Counter<uint>.TryCreate(context.CountTable, out counter))
+            {
+                return;
+            }
```
LDj3SNuD, line 403: I think we should raise an exception here, in the hypothetical case of not being able to allocate a further entry in the count table, also to avoid not being able to have rejits. In this way we can realise the limit and increase the capacity of the table.

ARMeilleure/Common/Counter.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            _countTable = countTable;
+            _index = index;
+        }
+
+        /// <summary>
+        /// Gets a reference to the value of the counter.
+        /// </summary>
+        /// <exception cref="ObjectDisposedException"><see cref="Counter{T}"/> instance was disposed</exception>
+        public ref T Value
```
LDj3SNuD, line 31: Nit:
```suggestion
        public ref T Value
```

ARMeilleure/Common/Counter.cs
```diff
@@ -0,0 +1,114 @@
...
+        public ref T Value 
+        {
+            get
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(null);
+                }
+
+                return ref _countTable.GetValue(_index);
```
LDj3SNuD, line 40: Nit:
```suggestion
                return ref _countTable.GetValue(_index);
```

ARMeilleure/Common/EntryTable.cs
```diff
@@ -0,0 +1,102 @@
...
+        public EntryTable(int capacity)
+        {
+            if (capacity < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(capacity));
+            }
+
+            _freeHint = 0;
+            _allocated = new BitMap();
+            _table = GC.AllocateArray<TEntry>(capacity, pinned: true);
```
LDj3SNuD, line 30: The approach is certainly interesting (https://tooslowexception.com/pinned-object-heap-in-net-5/) but tests (see below) show a significant reduction in performance linearly dependent on the capacity of the table (currently 16 MiB).
It would be advisable instead to use `_table = Marshal.AllocHGlobal(capacity);` and add a Dispose() method `Marshal.FreeHGlobal(_table);`.
To avoid wasting 16 MiB per run without restarting the emulator, the constructor of this class could be called in the `if (Interlocked.Increment(ref _threadCount) == 1)` block (@ Translator.Execute()) and the Dispose() method in the `if (Interlocked.Decrement(ref _threadCount) == 0)` block.

```
GitHub CI Artifact | GitHub CI Tests Time

#2190:
ubuntu-latest (Debug)    | 12m 42s
ubuntu-latest (Release)  | 6m 30s
macOS-latest (Debug)     | 13m 25s
macOS-latest (Release)   | 9m 57s
windows-latest (Debug)   | 9m 20s
windows-latest (Release) | 13m 6s

#master:
ubuntu-latest (Debug)    | 2m 2s
ubuntu-latest (Release)  | 2m 40s
macOS-latest (Debug)     | 1m 57s
macOS-latest (Release)   | 1m 53s
windows-latest (Debug)   | 2m 37s
windows-latest (Release) | 2m 17s
```
```
CountTableCapacity | Local Tests Time

4 * 1024 * 1024 | ~10:00
...
1000000 | 3:36
500000  | 2:42
250000  | 2:02
125000  | 1:40
62500   | 1:33
31250   | 1:28
```

LDj3SNuD (changes requested):

ARMeilleure/Common/EntryTable.cs
FICTURE7: Sounds good to me.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -218,9 +218,8 @@ public static void EmitTailContinue(ArmEmitterContext context, Operand address,
                 {
                     context.StoreToContext();
 
-                    Operand fallbackAddr = context.Call(typeof(NativeInterface).GetMethod(allowRejit
-                        ? nameof(NativeInterface.GetFunctionAddress)
-                        : nameof(NativeInterface.GetFunctionAddressWithoutRejit)), address);
+                    Operand fallbackAddr = context.Call(
+                        typeof(NativeInterface).GetMethod(nameof(NativeInterface.GetFunctionAddress)), address);
```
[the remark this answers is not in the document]
FICTURE7:
```suggestion
                    Operand fallbackAddr = context.Call(typeof(NativeInterface).GetMethod(nameof(NativeInterface.GetFunctionAddress)), address);
```

ARMeilleure/Common/EntryTable.cs
FICTURE7: FWIW this could also be fixed by implementing a memory pool, but busting out a memory pool to get a completely managed table seemed more like wishful thinking than anything else. Much like getting the tests to re-use the same `Translator` instance.

Kinda bummbed out I was not suspicious of the tests runtime. Thanks a lot for noticing and looking into this. 👍

FICTURE7:
Here's a chart of the working set of bcab67f (red) vs 5f5abaa (blue). Let me know if this is a concern, I'll revert 5f5abaa if so.

bcab67f vs 5f5abaa

LDj3SNuD:
I think the previous solution (on bcab67f) is preferable.
The current expandability of the table is welcome.
I also raise the question of the usefulness of using a template for the EntryTable class; because if we had a table with fixed-size entries (4 or 8 bytes per entry, for example) we could simplify a few things, not only in the class itself but also externally; we could send around `Value` as `IntPtr` type and for example we'd have `Operand address = Const(counter.Value.ToInt64(), true, Ptc.CountTableIndex);` (@ Translator.EmitRejitCheck()) while also avoiding
```csharp
        public static unsafe Operand Const<T>(ref T reference, int? index = null)
        {
            return Operand().With((long)Unsafe.AsPointer(ref reference), index != null, index);
        }
```
(@ OperandHelper.cs)
and in Ptc.PatchCode() we'd have `imm = (ulong)callCounter.Value.ToInt64();`.
This is not a prerequisite for PR approval.

FICTURE7:
The idea behind making `EntryTable` a generic was to make it a re-usable structure. For example it could re-used to replace the jump tables we currently use, and could be useful for various overnight experiments. Like implementing profiling/instrumentation of compiled code or implementing a constant pool.

As for using `IntPtr` as return value, it seems to me quite unideal to pass raw pointers around when we could be passing proper references. Plus it allows us to read and write to these references directly from C# without using things like `Marshal`. E.g here:
https://github.com/Ryujinx/Ryujinx/blob/dc8c781d7a8a6778e493cca798cf2a7d298983de/ARMeilleure/Instructions/NativeInterface.cs#L256-L262

LDj3SNuD (approved):

LDj3SNuD:
lgtm, ty

gdkchan (approved):

PR #2198, opened 2021-04-11
Fix sub-image copies on intel GPUs

gdkchan:
When doing sub-image copies with `glCopyImageSubData` on Intel, the driver seems to ignore the specified `dstLayer` for some reason (and also `dstLevel` I'm guessing), and always copies the data to the base layer. This problem doesn't happens when passing the storage handle rather than the view handle, so to fix this, the change adds a special path for intel, that passes storage handles rather than view handles, and adds the view base layer and level to the source and destination layer and level. In theory, this should work on all vendors without any performance penalty, but I decided to play safe here and only do this for Intel, to avoid the risk of breaking it for other vendors.

Some comparison screenshots:
Before:
image
image
After:
image
image

**Note:** This only affects Intel iGPUs, other vendors are not affected.
**Note 2:** `glGetTexSubImage` also seems to be ignoring the layer/level, which means that PBO copies have the same issues.
It is a simple change that improves things for Intel on OpenGL, even if we have Vulkan coming, so why not, right?

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Image/ITextureInfo.cs
+++ b/Ryujinx.Graphics.OpenGL/Image/ITextureInfo.cs
@@ -5,6 +5,10 @@ namespace Ryujinx.Graphics.OpenGL.Image
     interface ITextureInfo
     {
         int Handle { get; }
+        int StorageHandle { get; }
+        int FirstLayer => 0;
+        int FirstLevel => 0;
+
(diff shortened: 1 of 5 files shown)


riperiperi (approved):
Simple enough, nice catch!

Thog (approved):
tested and lgtm 👍

PR #2200, opened 2021-04-12
Improve shader global memory to storage pass

gdkchan:
The emulator currently has a optimisation pass that tries to pattern match global memory access patterns with a base address from a constant buffer. This has the same problem that bindless elimination had before #2089. Since the pass couldn't see beyond phi nodes. anything with control flow divergence would make it stop looking and give up, which would in turn result in the slow fallback path being used for the global memory access. This change uses the same trick used on bindless elimination to enable the pass to see through phi nodes and catch more cases. This allows the elimination to match, for example, cases where the global memory access and the base address are on different `if` blocks.

**Potential improvements:**
- Reduced code size.
    - May improve compile times as a result.
- Improved code execution speed (probably only matters for integrated/mobile GPUs).
- Reduced the number of bindings used (mitigates out of bindings compilation error on drivers with a low limit, like Intel that only supports 16 max).
- Better buffer management (no need to sync/write track unused buffers).

**Notes:**

I had to change the order of the optimisations a bit, and run the general opts. before the global elimination pass (as doing so increases the cases it can match, otherwise theres too much "noise" to match anything, such as redundant additions and copies), and after the pass, as the global elimination creates a few redundant operations, and running the other opts. again can catch more cases and further reduce code size.

The way how the global eliminations finds the base address was also changed, before it would just look through all operations until it was able to find a constant buffer operand or a phi node was hit, and it did that before any other opt. was done. Now it looks for specific patterns, which is more efficient, but could potentially reduce the number of cases it can match.

I recommend testing this with a few games, since I didn't do much beyond basic testing myself.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -35,7 +35,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2163;
+        private const ulong ShaderCodeGenVersion = 2200;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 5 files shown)


Thog (approved):

EmulationFanatic:
Didn't see any issues in my testing (tested Resident Evil Revelations 2 - a very shader heavy game - for about 30 minutes).

riperiperi (approved):
Nice one. Would be interesting to see a situation where the fallback is required now where it wasn't before, and how many of the fallbacks are removed by traversing phi nodes.

I tried a few games and there don't appear to be any regressions. I also noticed that the shader cache version wasn't changed, so if there are any issues after merging it's worth noting that they won't be visible for people with older shader caches.

PR #2211, opened 2021-04-14
Reduce number of dlls on distribution with single file publish

gdkchan:
This enables the single file publish feature at the project level to embed the managed dependencies on the executable, and remove some unused dependencies/files.

Here is what the publish folder looks like with this change:
image

Needs feedback, currently I still need to decide what to do about the PDBs:
- Embeding them requires adding `<DebugType>embedded</DebugType>` to each project which is not exactly ideal.
- Also, embeding is not working properly, stack traces are not showing the file and line numbers.

Easiest solution is simply including the PDBs on the output if we decide this information is desirable to have on release aswell, but it will be one extra file per project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Test
         run: dotnet test -c "${{ matrix.configuration }}"
       - name: Publish
-        run: dotnet publish -c "${{ matrix.configuration }}" -r "${{ matrix.DOTNET_RUNTIME_IDENTIFIER }}" -o ./publish /p:Version="1.0.0" /p:SourceRevisionId="${{ steps.git_short_hash.outputs.result }}" /p:ExtraDefineConstants=DISABLE_UPDATER Ryujinx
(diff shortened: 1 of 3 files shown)


Ghostotakoi:
👏Finally👏

WilliamWsyHK:
Can the platform-specific file not be output to other OS (e.g. `dylib` for macOS only, `so` for *nix only, instead of being included in Windows build; and `libMonoPosixHelper.dll` is not suppose to be included in Windows also)?

Thog:
> 
> 
> Can the platform-specific file not be output to other OS (e.g. `dylib` for macOS only, `so` for *nix only, instead of being included in Windows build; and `libMonoPosixHelper.dll` is not suppose to be included in Windows also)?

This is caused by us packing libsoundio in the repo. If it was handled correctly as a nuget package or via the correct directory and library map this wouldn't be an issue.

Can you open an issue about that?

Thog (approved):
lgtm

It should also be noted that this doesn't cause any LGPL incompatibilities as single file publish still dynamically link dependencies internally.

Daniel-iel (approved):

MelonSpeedruns:
Bump for this. I'd really like if it was approved.

AcK77 (approved):

GamerzHell9137:
Autoupdater is broken with this and Ryujinx can't be launched anymore.

AcK77:
I can upgrade from 1.0.6928 to 1.0.6929 without any issues here.

GamerzHell9137:
@AcK77 seems like its random. 2nd try it only crashed after renaming old files but could be launched afterwards. Normally you see "Would you like to restart Ryujinx now?"

PR #2213, opened 2021-04-16
Intel: OpenGL: No sub-image copies fix on Linux

A-w-x:
Intel OpenGL is fully working in Linux, lets not break it!

Fixes multiple regressions in Linux

ps: please do not apply Windows fixes on Linux

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
+++ b/Ryujinx.Graphics.OpenGL/Image/TextureCopy.cs
@@ -205,7 +205,7 @@ namespace Ryujinx.Graphics.OpenGL.Image
                     int copyWidth = sizeInBlocks ? BitUtils.DivRoundUp(width, blockWidth) : width;
                     int copyHeight = sizeInBlocks ? BitUtils.DivRoundUp(height, blockHeight) : height;
 
-                    if (HwCapabilities.Vendor == HwCapabilities.GpuVendor.Intel)
+                    if (!OperatingSystem.IsLinux() && HwCapabilities.Vendor == HwCapabilities.GpuVendor.Intel)
                     {
                         GL.CopyImageSubData(
                             src.StorageHandle,


riperiperi:
Does this cause any actual difference? Also, aren't intel GPUs on linux detected as "xorg" and "mesa", thus the vendor will be "unknown"?

A-w-x:
I didn't look at how you are detecting the gpu, but that code runs in linux and it is causing glitches

riperiperi:
Can you elaborate on the glitches, such as the games affected and screenshots of the issues? A copy from/to a storage texture should be identical to a copy from/to a view, so even if the workaround were running, it should have no effect.

A-w-x:
yeah, i don't the device now, but for example, lets use mario kart (that is also shown in the PR)

when you are in the character selection, characters are black
an other example, the semaphore when you start a race shows a few violet triangles when it should be a red circle (not sure how to explain better)

riperiperi:
I think a before and after screenshot is needed, as well as what the gpu info string printed in the log says.

A-w-x:
you mean this?

`Gpu PrintGpuInformation: Intel Mesa Intel(R) Iris(R) Plus Graphics (ICL GT2) (4.6 (Compatibility Profile) Mesa 21.0.2)`


master: (you can also see in race colors are a bit dark)

Screenshot_2021-04-16_18-48-20
Screenshot_2021-04-16_18-49-17


this PR:
Screenshot_2021-04-16_18-46-35

Screenshot_2021-04-16_18-45-46

gdkchan:
Actually fixing the storage copy would be more valuable than simply disabling the fix on linux, since it benefits both linux and windows.

A-w-x:
true, but still the issue the fix is trying to fix is non-existent in the first place on linux, so no need for this anyway
or i am missing something?

windows and linux drivers are not on par, they are two different things atm

Benedani:
Are you sure it will work across all Intel GPUs with adequate GLSL support?

gdkchan:
> 
> 
> true, but still the issue the fix is trying to fix is non-existent in the first place on linux, so no need for this anyway
> or i am missing something?
> 
> windows and linux drivers are not on par, they are two different things atm

Yea, it is not required on linux because that was already working there, I meant that fixing the issue would benefit windows users, and fix the regression on linux. Anyway, I think it would be more valuable to introduce those changes on `HwCapabilities.GetGpuVendor()` instead, that way it would disable *all* Intel workarounds on linux, rather than just this specific one. Maybe its worth renaming the `GpuVendor` enum name from `Intel` to `IntelWindows` if you do this.

PR #2216, opened 2021-04-16
Fix intel view copy workaround.

riperiperi:
The texture target must be taken from the storage rather than the view, when using the storage handle for the copy.

Screenshots with workaround artificially enabled on NVIDIA.

Before:
image

After:
image

Related to https://github.com/Ryujinx/Ryujinx/pull/2213 , but mesa should probably be explicitly detected to disable other workarounds anyways (specifically the view format one).


gdkchan (approved):
lgtm, nice catch.

Thog (approved):

PR #2217, opened 2021-04-16
Only enable clip distance if written to on shader

gdkchan:
This changes clip distances to only be enabled if they are actually written to on the vertex shader, since otherwise the results are undefined according to the spec.
No changes were observed on NVIDIA, but on Intel, this fixes flashing triangles on games that uses custom clip distances (most first party games).

Before:
image
After:
image

Should be tested on all games that were fixed by the original implementation: https://github.com/Ryujinx/Ryujinx/pull/1118
(I was not able to reproduce the bug on Super Mario Odyssey even without clip distances).

I also changed the way how InstanceId use is reported slightly, it now uses `FeatureFlags` to be consistent with how the other built-in uses are reported.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -39,6 +39,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
         private bool _isAnyVbInstanced;
         private bool _vsUsesInstanceId;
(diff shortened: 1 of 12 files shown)


riperiperi (approved):
LGTM, there don't seem to be regressions in the games I tested.

Thog (approved):

Thog:
Need rebase.

PR #2223, opened 2021-04-18
Amadeus: Allow out of bound read on empty delay lines

Thog:
This allows to handle an OOB with delay lines when DelayTimeMax = 0.
On real hardware, it will end up reading garbage at the given user work buffer address.
As we do not use those buffers and allocate them ourself for simplicy, this could possibly cause a crash.

Proposed solution here is to only increase the size of _workBuffer by one like what is done in DelayLineReverb3d already.

This fixes FEZ. (Ryujinx/Ryujinx-Games-List#3526)

image


AcK77 (approved):

riperiperi (approved):
Fair enough.

PR #2225, opened 2021-04-18
JIT Cache Exhausted Temporary Fix.

LDj3SNuD:
Temporary increase of JIT cache size to 3 GiB.
When the extra size is no longer needed, just edit this single line:
`private const uint CacheSize = 3U * 1024 * 1024 * 1024; // Default cache size = 2047 * 1024 * 1024.`
(@ ARMeilleure/Translation/Cache/JitCache.cs#L17).

### DIFF
--- a/ARMeilleure/Translation/Cache/CacheEntry.cs
+++ b/ARMeilleure/Translation/Cache/CacheEntry.cs
@@ -6,12 +6,12 @@ namespace ARMeilleure.Translation.Cache
 {
     struct CacheEntry : IComparable<CacheEntry>
     {
-        public int Offset { get; }
-        public int Size   { get; }
+        public uint Offset { get; }
+        public uint Size   { get; }
 
         public UnwindInfo UnwindInfo { get; }
 
-        public CacheEntry(int offset, int size, UnwindInfo unwindInfo)
+        public CacheEntry(uint offset, uint size, UnwindInfo unwindInfo)
         {
             Offset     = offset;
             Size       = size;
--- a/ARMeilleure/Translation/Cache/CacheMemoryAllocator.cs
+++ b/ARMeilleure/Translation/Cache/CacheMemoryAllocator.cs
@@ -8,10 +8,10 @@ namespace ARMeilleure.Translation.Cache
     {
         private struct MemoryBlock : IComparable<MemoryBlock>
         {
-            public int Offset { get; }
-            public int Size { get; }
+            public uint Offset { get; }
+            public uint Size { get; }
 
-            public MemoryBlock(int offset, int size)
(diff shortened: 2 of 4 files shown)


FIocker:
Dozens of Smash matches at 4k. No crashes. 

Would usually crash after 3-4 matches.

Very nais.

gdkchan (commented):
I think the int -> uint changes are valuable, but the fact that the 2GB limit still remains for the unwind table is unfortunate. Did you try passing a uint there? There's nothing on the documentation that says if the value is supposed to be signed or unsigned. It might be possible to simply call `RtlInstallFunctionTableCallback` twice, offseting the address the 2nd time (and limiting the size to min(2GB, remainingSize).

manfred-exz:
Works well for me. Fix one 100% happening "JIT Cache exhausted" issue in Monster Hunter Rise.

manfred-exz:
I was always getting `System.OutOfMemoryException: JIT Cache exhausted` while compiling `PPTC` in `Monster Hunter Rise 1.1.2`.
This pull request fixed it.

Thog (approved):

FICTURE7:
Not sure where to post this but, we can implement some sort of garbage collector on top of the JIT cache and free memory only when we need to, e.g when the allocator fails to allocate memory. Here is a rough prototype FICTURE7@2e894ab9c727eb7f0d1411bad5c1c253ddc0a855. However the fragmentation in the code cache gets rough very quickly. Perhaps could be worked around with different allocation strategy? 🤔

gdkchan:
> 
> 
> Not sure where to post this but, we can implement some sort of garbage collector on top of the JIT cache and free memory only when we need to, e.g when the allocator fails to allocate memory. Here is a rough prototype [FICTURE7/Ryujinx@2e894ab](https://github.com/FICTURE7/Ryujinx/commit/2e894ab9c727eb7f0d1411bad5c1c253ddc0a855). However the fragmentation in the code cache gets rough very quickly. Perhaps could be worked around with different allocation strategy? 🤔

Currently the only memory that can be deleted is the one used for redundant translations of the same function, which I assume is a rare occurrence and wouldn't be enough to make a difference on games where we get cache exhaustion. I think making it effective would require some way to remove LowCq functions from the cache (either because they have been replaced by HighCq or because they are no longer need, which should be the case for cold boot initialization functions for example).

Daniel-iel (approved):

Benedani:
With #2286, this should no longer be needed.

Daniel-iel:
So let's close the PR

LDj3SNuD:
PR was opened, as a pre-POWER temporary solution, to help some users complete their most demanding games in terms of JitCache; it can now be closed.

PR #2228, opened 2021-04-19
Add multi-level function table

### LINKED ISSUE #2158: Fast CPU emulator JIT function lookup

FICTURE7:
Adds a multi-level function table as described in #2158. ~~The levels are hardcoded into the algorithms, I can make it configurable if this is undesirable.~~

**TODO**
- [x] Add PPTC support.
- [x] Add documentation to `AddressTable<T>`.
- [x] Add fallback for out of range addresses.
- [x] Improve codegen for out of translation branches.

~~Feel free to review/comment even though this is in draft.~~

Closes #2158.


gdkchan:
> The levels are hardcoded into the algorithms, I can make it configurable if this is undesirable.
Yea, I would prefer if it had some constants for that (it doesn't need to be something that is configurable at runtime, just some compile time constants that we can easily change to control for example, the amount of bits for each level etc).

FICTURE7:
PPTC could support the shorted table walk path if we stored the guest address in the `PtcIndex` or something, but the change seemed a little intrusive over the PPTC machinery. If its desired I can implement.

gdkchan:
Btw I think with this change we can enable single block  translation in LowCq (not suggesting doing it as part of this PR, it should probably be a separate change), but it would be nice to know if doing so saves us anything in compilation time and/or code size.

FICTURE7:
Right with this change LCQ should perform much better and have lower latency, plus we can change the current return behavior to be a virtual jump instead:
https://github.com/Ryujinx/Ryujinx/blob/89791ba68dba70999643c5d876e9329b385c6e8a/ARMeilleure/Instructions/InstEmitFlow.cs#L65-L70
This should allow us to remain in unmanaged for far longer and possibly get rid of this loop:
https://github.com/Ryujinx/Ryujinx/blob/89791ba68dba70999643c5d876e9329b385c6e8a/ARMeilleure/Translation/Translator.cs#L158-L162
Care should be taken around calls though because it will probably stack dive. I guess we can do the virtual jump return only if the call depth is 1 or avoid trying to mimic the guest calls stack by always doing jumps.

However I suspect that if we enable single block decode in LCQ, the JIT code will overlap more, at least with the current rejit policy. Take the following flow for example: `block A -> block B -> block C`. If block A becomes hot, block B and C too will probably become hot, and since presumably HCQ will not be single block decode, an HCQ compile at block A will overlap [A, B, C] and a compile B will overlap [B, C]. I guess we have to figure a new rejit policy or do fancy tracing?

gdkchan:
> I guess we have to figure a new rejit policy or do fancy tracing?

I guess in a ideal world, the decoder would be able to identify and decode a function perfectly, and then we'd just need to rejit call target addresses. But there are several complications, not only because the decoder might hit the instruction limit or mess up in some other way and not decode the entire function, but also due to things like tail calls, where the call is just a B or BR instruction jump.

I think it will need some way to identify all the blocks of a function, and then only rejit from the functions entry point. Perhaps after changing LCQ to use single blocks, we can find a way to represent the "connection" of the functions on the cache, and build some sort of control flow graph with them. Block A functoin would be connected to B, which would be connected to C etc. A rejit request of C could just walk the graph backwards all the way up to the entry point (block A function) and request a rejit of that instead. I guess I should think a bit more about that, but it does sound like something that could work.

Another thing that we will need to eventually take into account is the removal of LCQ functions (and also of HCQ at some point, for cache invalidation). this change should hopefully make the task a little bit easier, since one no longer needs to remove entries from the jump tables.

FICTURE7:
Here's another idea which should be closer to what we currently do. If an edge single block becomes hot, do an HCQ decode and record that interval in some data structure. If the single block was already overlapping in 
Although there could be other gotchas to this.

LDj3SNuD:
About PPTC there is a bug in the code in that the `PtcProfiler.UpdateEntry()` method is never called, so rejits (even across multiple runs) are not profiled.
Also, to avoid repeating code, it might be useful to make the `Translator.RegisterFunction()` method static so that it can also be called from `TranslateFuncs()` (@ Ptc.MakeAndSaveTranslations()). It's also understood that the RegisterFunction() method should remove calls to PtcProfiler methods, so that they can be called from the appropriate locations (always post-translate, because the profiled info must be translate-proof).

FICTURE7:
Yep, there are other issues with PPTC the integration atm. Will push the changes later today or tomorrow.

EmulationFanatic:
I can't launch any games with this PR. On first boot they just hang after a moment, and on 2nd boot they crash without any error in the console/log.

FICTURE7:
> I can't launch any games with this PR. On first boot they just hang after a moment, and on 2nd boot they crash without any error in the console/log.

Looks like I broke the table walk during d310ea666d9b443cff322514a4a2695aa3944b0e and hopefully this is the only thing I broke. :eyes: Thanks a lot for testing this. 👍

Also if the extra complexity of d310ea666d9b443cff322514a4a2695aa3944b0e seems unwarranted or concerning, I can add some extra constraints and simplify it to almost what is was.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -208,164 +173,114 @@ public static void EmitTailContinue(ArmEmitterContext context, Operand address)
...
+                        index = context.BitwiseAnd(context.ShiftRightUI(guestAddress, Const(level.Index)), mask);
 
-            EmitBranchFallback(context, address, isJump);
+                        if (i < context.FunctionTable.Levels.Length - 1)
+                        {
+                            page = context.Load(OperandType.I64, context.Add(page, context.ShiftLeft(index, Const(3))));
 
-            context.MarkLabel(endLabel);
-        }
+                            context.BranchIfFalse(lblFallback, page);
```
riperiperi, line 230: It may be possible to remove all branching here, which is useful for the most common path where everything is defined anyways:

- Add "dummy tables", where every value points to a fallback stub function.
  - Empty top level points to a dummy 2nd level, where all entries point to a dummy 3rd level... etc
- A function for finding, executing and recording the function can be used as a default value. The address can be passed on the native context, like it is with jump tables.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -208,164 +173,114 @@ public static void EmitTailContinue(ArmEmitterContext context, Operand address)
...
+            Operand lblFallback = Label();
+            Operand lblEnd = Label();
+
+            // If address is mapped onto the function table, we can do an inlined table walk. Otherwise we fallback
+            // onto the translator.
+            if (context.FunctionTable.IsMapped(guestAddress.Value))
             {
-                if (i == JumpTable.DynamicTableElems - 1)
+                // If the guest address specified is a constant, we can skip the table walk.
+                if (!context.HasPtc && guestAddress.Kind == OperandKind.Constant)
```
riperiperi, line 204: It would be ideal to have this working with PTC, as a relocatable constant. The table search is much slower than the existing behaviour for constant table lookup that jump tables provide, and this would bring it much closer.

riperiperi (commented):
Nice work! I have a few suggestions.

Goffy was testing the branch and noticed that it was a bit slower, which hopefully can be improved with some small changes. It is faster with PTC disabled (after running for a while), so the constant table lookup is definitely valuable.

Goffy also noted that this branch seems to have random 300ms pauses in smash where everything stops - including the music. At a glance, I'm not sure what that is, but it's something to look into.

FICTURE7:
> Goffy was testing the branch and noticed that it was a bit slower, which hopefully can be improved with some small changes.

We might be able to do that recoup the performance by switching guest returns to virtual jumps instead.

> Goffy also noted that this branch seems to have random 300ms pauses in smash where everything stops - including the music. At a glance, I'm not sure what that is, but it's something to look into.

Quite odd, it could be because of its emitting more IR putting pressure on the pools and the GC.

FICTURE7:
I'll break this up and add PPTC support for the shorted table walk later today or tomorrow to this PR, then follow up with other optimization PR(s). Just so this remains manageable and regression are caught early.

EmulationFanatic:
@FICTURE7 Thanks for fixing the game booting issue. I wanted to share with you the results of some performance testing I did of `94a9e17` vs. master, as I was curious about the potential benefits. Note that all shaders were pre-built for this test as I wanted to rule out any associated FPS drops.

**SSBU 10.0.0 torture test (8 players on Fountain of Dreams)**
PPTC On (full purging of PPTC including .info file before building new PPTC with this commit):
image

PPTC Off:
image

**Animal Crossing: New Horizons - 10 minutes idling in-game at the starting point (Leischii's save with a full island)**:
PPTC Off:
image

At this point I stopped testing as it seems there is a common theme and I felt you should know sooner rather than later.

ARMeilleure/Common/AddressTable.cs
```diff
@@ -0,0 +1,256 @@
...
+                Mask |= level.Mask;
+            }
+        }
+
+        /// <summary>
+        /// Determines if the specified <paramref name="address"/> is mapped on to the table.
+        /// </summary>
+        /// <param name="address">Guest address</param>
+        /// <returns><see langword="true"/> if is mapped; <see langword="false"/> otherwise</returns>
+        public bool IsMapped(ulong address)
```
gdkchan, line 129: I would probably call this `IsValid`, since if the address is out of range then it is invalid.

ARMeilleure/Common/AddressTable.cs
```diff
@@ -0,0 +1,256 @@
...
+        public ref TEntry GetValue(ulong address)
+        {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException(null);
+            }
+
+            if (!IsMapped(address))
+            {
+                throw new ArgumentException($"Address 0x{address:X2} is not mapped onto the table.", nameof(address));
```
gdkchan, line 150: Why the `X2` here? From what I recall, this will force the value to have at most 2 trailings zeros (if the value happens to be < 0x10).

ARMeilleure/Common/AddressTable.cs
```diff
@@ -0,0 +1,256 @@
...
+    {
+        /// <summary>
+        /// Default levels used by <see cref="AddressTable{TEntry}"/>.
+        /// </summary>
+        private static readonly Level[] DefaultLevels = new[]
+        {
+            new Level(39,  9),
+            new Level(30,  9),
+            new Level(21,  9),
+            new Level( 2, 19)
```
gdkchan, line 21: This should take into account thumb code, that may be 2-byte aligned. Maybe change the range to `1, 19` if the translator is 32-bit?

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -208,164 +173,109 @@ public static void EmitTailContinue(ArmEmitterContext context, Operand address)
...
 
-                Operand funcAddr = context.Load(OperandType.I64, tableEntryPtr);
+                // Try to continue within this block.
+                // If the return address isn't to our next instruction, we need to return so the JIT can figure out
+                // what to do.
+                Operand lblContinue = context.GetLabel(nextAddr.Value);
 
-                // Call the function directly. If it's not present yet, this will call the direct call stub.
-                EmitNativeCallWithGuestAddress(context, funcAddr, address, isJump);
+                // We need to clear out the call flag for the return address before comparing it.
```
gdkchan, line 275: Which call flag? I already removed it with my last JIT cache changes.

ARMeilleure/Common/AddressTable.cs
FICTURE7: I am not very familiar with the HLE, and unsure of how to pass the hint to the translator its mostly targeting 32-bits. Presumably this https://github.com/Ryujinx/Ryujinx/blob/ec97a8a1faef122c0f13d881ecc7f74cd20d95a0/Ryujinx.HLE/HOS/ArmProcessContextFactory.cs#L7-L13
would have to pass on the information?

gdkchan: Yea, it is availabe on `KProcess.Flags.HasFlag(ProcessCreationFlags.Is64Bit)`. I guess another alternative is getting it from the `ExecutionContext` of the first thread to call the translator execute method (this would also require lazy initialization of the address table). This would also be assuming that all threads using the translator will be either 32-bit or 64-bit, which should be always the case, but from a design perspective, I do think its better to just pass the information on the constructor.

gdkchan:
Been testing some games on this branch, and so far I only noticed a significant improvement with ACA NEO GEO.
master:
image
PR:
image
Around +6 FPS ingame on this title.
No regressions observed so far (in performance or games breaking). Good job.

Leischii:
Regression also seems to be fixed for AC, maybe even slightly better fps now:
https://i.gyazo.com/thumb/1000/2d63752f04f4225b77e806b76eeda54e-jpg.jpg

Leischii:
Regression also seems to be fixed for AC, maybe even slightly better fps now:

PR:
https://i.gyazo.com/thumb/1000/2d63752f04f4225b77e806b76eeda54e-jpg

ARMeilleure/Diagnostics/EventSources/AddressTableEventSource.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Diagnostics.Tracing;
```
LDj3SNuD, line 1: EventCounters are welcome (they are the new Performance Counters in .NET "Core"); using ad hoc EventSources it would also be possible to integrate/replace the current Log system, with an out-of-proc consumer in a separate process, etc.
However I would recommend the following pattern https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters#conditional-counters if possible, as most of our users would not use them in practice.

FICTURE7: This uses a polling event source and the consumer will poll it as needed, so the only thing we'll really be saving is the construction the `PollingCounter` instance when the process is not being monitored. The resulting overhead will still be the same. Unless if you insist. :>

LDj3SNuD: I will not insist :), however
`Conditional counters are counters that are conditionally instantiated, a micro-optimization. The runtime adopts this pattern for scenarios where counters are normally not used, to save a fraction of a millisecond.`
in fact
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs
which also uses PollingCounter

FICTURE7: Right, AFAIK the runtime folks do it this way because they want shave a couple of ms and allocations of overhead from having to initialize these counters when they are not needed. I think we've got some ways to go until we these kinds of strategies pays off. But then again, its quite a small change.

ARMeilleure/Translation/TranslatorStubs.cs
```diff
@@ -0,0 +1,256 @@
...
+        public TranslatorStubs(Translator translator)
+        {
+            if (translator == null)
+            {
+                throw new ArgumentNullException(nameof(translator));
+            }
+
+            _translator = translator ?? throw new ArgumentNullException(nameof(translator));
+            _dispatchStub = new(GenerateDispatchStub, isThreadSafe: true);
+            _dispatchLoop = new(GenerateDispatchLoop, isThreadSafe: true);
```
FICTURE7, line 91:
```suggestion
            _translator = translator ?? throw new ArgumentNullException(nameof(translator));
            _dispatchStub = new(GenerateDispatchStub, isThreadSafe: true);
            _dispatchLoop = new(GenerateDispatchLoop, isThreadSafe: true);
```

gdkchan (commented):
I took a look at the implementation and everything looks reasonable to me, good job.

I checked the amount of memory used by the table using the event counters that you added, the results were a little bit unexpected. It uses a pretty high amount of memory at boot (a few MBs usually), and after that, the amount of memory used doesn't change much. I would expect it to gradually increase as the game is played. It might be worth trying to tweak the level sizes to maybe improve the table distribution and maybe decrease the upfront memory usage. (Tested with PPTC disabled).

**Edit:**
I made a quick test here, and reduced the size of the lowest level, and redistributed the other sizes. The table memory usage dropped from 110MB to around 66MB ingame, so it might be worth tweaking it a bit. However making the value too low would increase the amount of allocations a lot, which is not great either. I guess we need to find a balance.

FICTURE7:
I added an extra level to the table, seems to produce ~3x decrease in memory use in most cases. One quick way to get the memory usage is to get a large PPTC info and doing something like this:
```csharp
foreach (var k in PtcProfiler.ProfiledFuncs) funcTable.GetValue(k.Key);
```
The PPTC cache need not be purged because the table walk is only done in the dispatch routine, which is compiled with every `Translator` instance. So this may allow us to tune the levels automatically by analyzing the PTC info at initialization.

gdkchan:
Are you sure the allocations are not a problem now though? because smaller blocks means more allocations, I'm just a bit worried about the extra allocations negatively affecting the boot times.

FICTURE7:
Hmm, the allocations should be smaller so it should be cheaper to initialize (since we have to fill less memory). But I'll try to get some numbers just for good measure.

To be honest I have been second guessing this change and wondering if we could get away with the original dense jump tables (even considering the possibility of patching the `jmp`s into relative jumps with consideration for unlinking etc) but use some kind of branch predictor for indirect branches. Because AFAIK the complication with cache invalidation was to unlink the dynamic table.

FICTURE7:
This PR against master, seems to consistently regress boot time by ~0.974 times on average. I also checked how boot time would perform with a 64-bit table (since it emits less IR for a direct branch) and with an optimization which folds a load into a `MemoryOperand` on `Call`s and `Tailcall`s, and it seems to consistently perform better than master by ~1.07 times on average.

The maximum I could get the 64-bit table was ~50mb, whilst the 32-bit one ~30mb (the current jump tables reserves 32mb).

> To be honest I have been second guessing.

Maybe its alright after all. 🤔 Either way, if the reduction in boot time is a concern, I could implement some kind of intrinsic for memory accesses, and hopefully this should compensate.

FWIW the reason why the `DispatchLoop` does not improve performance much is because the translation does a pretty good job at staying parallel with the guess call stack, most call guard succeeds (> 90%). One of the cases which causes the call guards to fail is when there is a recursive call.

https://github.com/Ryujinx/Ryujinx/blob/acb1647c86ee8410817ae22b38fcd481a07fe84d/ARMeilleure/Instructions/InstEmitFlowHelper.cs#L143-L148

gdkchan:
> 
> 
> > Why does it need to be a intrinsic though? Is it not possible to just fold the load into the call?
> 
> Oh, I meant the guest memory accesses as in `InstEmitMemoryHelper`, instead of expanding them as BBs and IR operations.

I already thought about this, but couldn't think in a good way to move the complexility down the pipeline by exposing a specialized IR instruction. Do you have anything specific in mind?

FICTURE7:
I am thinking of something like this and similar for stores.

```llvm
i32 %result = TranslatorIntrinsic.Load8 i64 %guestAddress
i32 %result = TranslatorIntrinsic.Load16 i64 %guestAddress
i32 %result = TranslatorIntrinsic.Load i64 %guestAddress
i64 %result = TranslatorIntrinsic.Load i64 %guestAddress
v128 %result = TranslatorIntrinsic.Load i64 %guestAddress

;; variants for aligned/exclusive accesses as well?
```

The register allocators and pre-allocator will have to be modified to treat these as calls since they may call into the `Translator` (multiple times too with different signatures, `SignalMemoryTracking` & `ThrowInvalidMemoryAccess`). They may also have to be modfied to understand "temporaries" which are just short lived locals which are consumed by the `CodeGenerator`. For example something like this:

Before register allocation:
```llvm
i32 %result = TranslatorIntrinsic.Load8 i64 %guestAddress {i32 %t0, i64 %t1, i64 %t2, ...}
```

After register allocation:
```llvm
i32 r4 = TranslatorIntrinsic.Load8 i64 r3 {i32 r0, i64 r1, i64 r2, ...}
```

The `CodeGenerator` can then use these temporaries to generate the machine code.  `HybridAllocator` may not like that cause it has some limitation with the number of operands live in registers, if I understand correctly. But we may be able to avoid having to modify the register allocators by clobbering some `LocalVariable`s? It'll generate more IR which has to be travsered though (should be better than currently still).

```llvm
i32 %t0 = Clobber
i64 %t1 = Clobber
i64 %t2 = Clobber
...
i32 %result = TranslatorIntrinsic.Load8 i64 %guestAddress, i32 %t0, i64 %t1, i64 %t2, ...
```

This will also tighten the coupling of the backend with the translator, as is at least. There is also the thing with hot/cold splitting, don't think it will be a problem to implement in the `CodeGenerator` itself though.

gdkchan:
> The register allocators and pre-allocator will have to be modified to treat these as calls since they may call into the Translator (multiple times too with different signatures, SignalMemoryTracking & ThrowInvalidMemoryAccess).

This is where most of the complication comes from. I guess the IR instruction should take the function pointer it should call on failure, but in the end it would be a very weird instruction, as it does so much as it would need to be so specific for the emulator use case. That's why I'm not a fan of it, I think its better to build something more generic that can be reused for other things if needed.

I think the simplest solution here (and one that does not require any new IR instruction) is just calling the memory read/write functions directly in LCQ. I have already verified that this brings some improvements for boot times, but obviously everything else is a lot slower as a result, as memory accesses becomes super expensive. So I think a possible solution here is just emiting a few small native functions for memory read/write, that does not need to pay the managed transition cost most of the time. I think it would be a good balance, but I haven't tried it, so can't say for sure how well it would work. Ensuring that this is performant would probably require some optimizations on the handling of calls too.

> HybridAllocator may not like that cause it has some limitation with the number of operands live in registers, if I understand correctly.

Yea, it is sort of weird. Originally my idea for this allocator was spliting the list of available registers in half. One half would be used for block local variables (basically short-lived variables that don't span blocks), and the other half would be used for long-lived variables that does span multiple blocks. The long-lived variables would be sorted by use count (the ones with highest number of uses being on the top), so the ones with most uses would get a register, while the ones on the bottom of the list would live on the stack. In theory, that should be very good because we can basically reuse the register allocation from the original program (where the compiler should have attempted to use the lowest possible amount of registers, which in the translator results in a few long-lived variables with a lot of uses), and also take into account the large number of short-lived variables introduced by the translation process. That is also why it is called *hybrid* allocator.

Unfortunately some issue at the time (I don't remember exactly what) prevented me from finishing it. So as it stands right now, it only does local block allocation. You will notice that there are quite a bunch of things there are unused right now. I also wanted to work on some GPU related stuff at the time, after spending ~2 months or so in the JIT, which made just take the easiest path. Anyway, I just figured this tidbit could be useful for understanding it.

> We may perhaps need more granularity with e.g an TranslatorIntrinsic.LoadPage or something if we started bundling memory accesses together like FICTURE7/Ryujinx@7c81f27. Also if there is any interest over that change, I could polish it up and send a PR for it.

I think it would be better to try what I proposed above first, with the native function calls, and then build this optimization on top of it if we decide to go this route. It could have something like a `Read2x32` and `Read2x64` function. Multiple returns values are a bit tricky though.

riperiperi (approved):
Looking brilliant! The slowdowns from before seem to be gone, and performance is generally better than master. The managed run loop is a good idea and probably does save time in edge cases.

I've got a few comments about some future direction, but I'm happy with what's here now.

- I still think that the lookup in the table could be faster by cutting out the null checks for table levels by using dummy levels. By how much, I'm not sure.
- It's definitely faster for compilation to be calling a stub for the function lookup, but I'm interested in what the effect on execution performance is of a jump/call versus inlining the lookup.
  - Removing null check branches could make this have less of an impact on compilation time.
- It would be beneficial if we could have less levels on the table, as it's obviously more to go through for every call. Might be worth tweaking around with to see if we can get a configuration that appropriately covers a game's small code region without using too much additional memory.

ARMeilleure/Translation/Translator.cs
```diff
@@ -406,7 +437,7 @@ internal static void EmitRejitCheck(ArmEmitterContext context, out Counter<uint>
 
             Operand lblEnd = Label();
 
-            Operand address = Const(ref counter.Value, Ptc.CountTableIndex);
+            Operand address = Const(ref counter.Value, Ptc.CountTableSymbol);
```
LDj3SNuD, line 440:
```suggestion
            Operand address = !context.HasPtc ? Const(ref counter.Value) : Const(ref counter.Value, Ptc.CountTableSymbol);
```
Otherwise Operand will be Relocatable and we may lose optimisations.

ARMeilleure/Instructions/InstEmitFlowHelper.cs
```diff
@@ -196,176 +160,65 @@ public static void EmitVirtualJump(ArmEmitterContext context, Operand target, bo
...
-
-                context.Branch(endLabel);
-            }
+            Operand hostAddress;
 
-            // Currently this uses a size of 1, as higher values inflate code size for no real benefit.
-            for (int i = 0; i < JumpTable.DynamicTableElems; i++)
+            // If address is mapped onto the function table, we can skip the table walk. Otherwise we fallback
+            // onto the dispatch stub.
+            if (guestAddress.Kind == OperandKind.Constant && context.FunctionTable.IsValid(guestAddress.Value))
```
LDj3SNuD, line 186:
```csharp
            if (guestAddress.Kind == OperandKind.Constant && context.FunctionTable.IsValid(guestAddress.Value))
            {
                Operand hostAddressAddr = !context.HasPtc
                    ? Const(ref context.FunctionTable.GetValue(guestAddress.Value))
                    : Const(ref context.FunctionTable.GetValue(guestAddress.Value), new Symbol(SymbolType.FunctionTable, guestAddress.Value));

                hostAddress = context.Load(OperandType.I64, hostAddressAddr);
            }
            else
            {
                hostAddress = !context.HasPtc
                    ? Const((long)context.Stubs.DispatchStub)
                    : Const((long)context.Stubs.DispatchStub, Ptc.DispatchStubSymbol);
            }
```

FICTURE7:
I slapped together a prototype to enable single block decodes in LCQ 9a4d9d5481b952032ce965747f3f400362fdd3fd. It works very similarly to how it was before #2190. A flag `InstrumentEdge`, is stored on the `NativeContext` which the rejit check will check to determine if it should count this transition. An edge is counted if it is a
* Direct/Indirect call
* Direct/Indirect jump
* A continuation from an HCQ translation

The last one is because, usually the decoder will produce an incomplete graph of the function (if the decode started at the entry of the function even) due to gaps or indirect branches, so in an attempt to have the whole function in HCQ, the continuations of the translation is also counted. You can think of the continuation as an edge to another part of the function's CFG which isn't available.

Haven't got the chance to thoroughly check if this causes performance regressions, but boot time against this PR as base seem to have improved by 24%-40%

gdkchan:
How easy always using "continuations" for LCQ (including Arm BL/BLR calls and RET returns) would be? I ask because I think theres no benefit in making an actual call and growing the stack if we only have a single block, and theres nothing after the call itself. On top of that, there are some games that currently triggers a stack overflow while executing guest code, presumably due to recursive calls that ends going too deep, so this would be a possible solution for this issue aswell (atleast assuming that the recursive code never leaves LCQ).

FICTURE7:
> How easy always using "continuations" for LCQ (including Arm BL/BLR calls and RET returns) would be?

I cannot think of any gotchas with doing that, so I think relatively easy. However even in LCQ, it does a pretty good job at staying parallel with the guest call stack. Since we do a call only on BL/BLR and return on RET, the continuations are direct branches which the decoder has not explored. As we do tailcalls for continuations, we do not deviate from the guest call stack.

So if also do continuations, we will lose hardware assisted returns and won't emit code pass a BL/BLR instruction in LCQ. However we will be more resilient against large stack usage, so I am not sure.

gdkchan:
> So if also do continuations, we will lose hardware assisted returns and won't emit code pass a BL/BLR instruction in LCQ. However we will be more resilient against large stack usage, so I am not sure.

Was thinking about this, I guess the problem is HCQ code calling LCQ. It using cotinuations for returns there would be bad as the stack would keep growing. It's a shame because it had some potential to bring some improvement to LCQ when using single blocks.

EmulationFanatic:
@FICTURE7 can you rebase?

FICTURE7:
Rebased.

EmulationFanatic:
Since this seems to be the final state of the PR, I figured I'd update my numbers.

**SSBU Stats**
Boot time
Master: 48.09s
PR: 47.1s

Avg FPS during torture test:
Master: 52.3
PR: 54.6

**ACNH Stats**
Boot time
Master: 68.58s
PR: 64.91s

Avg FPS during afternoon on Leischii's island:
Master: 59.3
PR: 59.1

These certainly paint the PR's performance in a much better light than last month when I first had compared. Great work!

riperiperi (approved):
Looking great here, very nice to see! Improvements seem most notable in guest emulators like the NEO GEO ones, presumably due to the small function size, though I can imagine there are a few games that run weirdly slow that could also be due to the current slow indirect lookups.

FICTURE7:
> Improvements seem most notable in guest emulators like the NEO GEO ones

I am guessing it is using a dispatch table for its interpreter which the old approach does not like because it would have multiple targets and so would hit managed from time to time.

I'll add your suggestion of adding dummy tables in another PR, and I think the table walk could be sped up by adding some kind of per `NativeContext` TLB which shorts to the last level or something.

PR #2231, opened 2021-04-20
Miria: Do not try to query keyboard input when controller isn't set

Thog:
This fix a possible crash with raw keyboard input since Miria.
I took the liberty to move the the keyboard update to avoid possible duplicate update if we end up having two keyboards in use.


riperiperi (approved):
LGTM. Still a bit weird that this is tied to a controller, but this works for now.

Thog:
> 
> 
> LGTM. Still a bit weird that this is tied to a controller, but this works for now.

Yeah might be worth

AcK77 (approved):

PR #2233, opened 2021-04-21
mm: Cleaned up the service a bit

AcK77:
This PR just cleaned up the `mm:u` services a bit with some unknown values which is now known. Nothing more.


gdkchan (approved):
lgtm, thanks.

Thog (approved):

PR #2237, opened 2021-04-22
Allow any shader SSBO constant buffer slot and offset

gdkchan:
The Switch GPU supports memory access on shaders from anywhere on the GPU mapped memory. However, games don't just pass raw GPU pointers to the shaders, they use "storage buffers", which are buffers that can be read and written to by shaders. The game binds the storage buffer to binding N, and then the shader can access the data through binding N. The information for each storage buffer (base address and size) is stored on a driver reserved constant buffer (constant buffer #0 for NVN), and the shader reads the base address from there.

The shader translator tries to undo what the compiler does by trying to find from where the base address is read, this way it can find the binding number, and then bind all the buffers upfront as needed. Right now the translator is making assumptions about where this information is stored. It assumes that it is always stored on the constant buffer and range that NVN uses. This doesn't work for other APIs that uses different locations to store this information (such as homebrew APIs).

This change removes the assumption about NVN, and for each global memory access, it will just use whichever constant buffer it can find, and assume that contains the storage buffer information (base address and size). The information is propagated to the GPU emulator to allow the buffer to be bound. The `BufferDescriptor` struct was updated to hold this information.

Benefits:
SSBOs on APIs other than NVN works, as long the shader translator can find from where the base address comes from.

Issues to be discussed:
Being too lenient about where the storage buffer information can possibly be can lead to false positives (for example, it may find a constant buffer and assume it holds the base address, when it doesn't). It might be better to limit the match to only match for constant buffer 0 (which is NVN reserved constant buffer, and I believe other APIs uses this constant buffer for this purpose aswell).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Compute.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute.cs
@@ -89,11 +89,8 @@ namespace Ryujinx.Graphics.Gpu.Engine
             {
                 BufferDescriptor sb = info.SBuffers[index];
 
-                ulong sbDescAddress = BufferManager.GetComputeUniformBufferAddress(0);
-
-                int sbDescOffset = 0x310 + sb.Slot * 0x10;
-
-                sbDescAddress += (ulong)sbDescOffset;
+                ulong sbDescAddress = BufferManager.GetComputeUniformBufferAddress(sb.SbCbSlot);
+                sbDescAddress += (ulong)sb.SbCbOffset * 4;
 
                 SbDescriptor sbDescriptor = _context.PhysicalMemory.Read<SbDescriptor>(sbDescAddress);
 
--- a/Ryujinx.Graphics.Gpu/Engine/Methods.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Methods.cs
@@ -327,11 +327,8 @@ namespace Ryujinx.Graphics.Gpu.Engine
                 {
                     BufferDescriptor sb = info.SBuffers[index];
 
-                    ulong sbDescAddress = BufferManager.GetGraphicsUniformBufferAddress(stage, 0);
-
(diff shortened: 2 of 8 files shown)


TSRBerry:
Needs a rebase

Niwu34:
image
^^
was able to fix the menu not showing up by deleting `vulkan_nvidia.data` and `vulkan_nvidia.toc`  and letting the emu rebuild those files at boot helped!

The game crashes in the loading screen after the menu now tho!
image

gdkchan:
> First time booting BOTW lead to this: image was able to fix the menu not showing up by deleting `vulkan_nvidia.data` and `vulkan_nvidia.toc` and letting the emu rebuild those files at boot
> 
> The game crashes in the loading screen after the menu now tho! image

should be fixed now, can you restest?

Niwu34:
image
compiled this build before the third commit but seems to be fixed now

riperiperi:
image

Seems to still work properly rebased. There are other issues I needed to patch around to get melonDS running again though:

- ConstantBufferUpdater batched buffer size is 512 int32s. This was an oversight in the way the current implementation was programmed, and how most games end up using it.
  - Solution: increase the buffer size (which i have done to get this screenshot), or rethink how this is done. Probably will be the latter since I think it can still be faster.
- WaitForIdle is used as a barrier for indirect compute. I don't think this is the correct barrier, but it should probably work anyways. A change to master makes it defer emitting sync on WaitForIdle until the current block of compute is finished, as it is generally spammed... Which isn't good for normal games, but working around it isn't good for this homebrew.
  - Flush deferred WaitForIdle on command buffer fetch (?) or improve sync in general so this doesn't hurt as bad.
  
These obviously have nothing to do with this PR. NVN games seem to still be working fine, and I'm doing more testing to verify.

riperiperi (approved):
LGTM, seems good in the games I tested. Should make things a lot nicer for homebrew.

marysaka (approved):

PR #2241, opened 2021-04-24
SM instance & TIPC fixes

Thog:
This PR addresses the following issues:
- SM was previously instancied once and reused on all sessions. This
  could cause inconsistency on the service initialization.
- TIPC replies were not matching what is generated on hardware.


AcK77 (approved):

gdkchan (approved):
Not a huge fan of the way some of those things are handled, but it's fine for now.

PR #2246, opened 2021-04-25
Initialize hid inputs on activation to avoid spurious inputs

Caian:
This PR addresses ghost typing and movement issues in Black Legend caused by leaving the hid shared memory completely zeroed.

image

I added code to `HidServer` to fill the entries of the hid devices when their respective `Activate` function is called.


AcK77 (approved):

gdkchan (approved):
lgtm, nice catch.

PR #2250, opened 2021-04-28
nsd: Cleanup the service and implement some calls

### LINKED ISSUE #2247: nsd:u IManager: 100 (GetApplicationServerEnvironmentType) is not implemented
nsd:u IManager: 100 (GetApplicationServerEnvironmentType) is not implemented

[Switchbrew](https://switchbrew.org/wiki/Sockets_services#nsd:u.2C_nsd:a) | 

## Service description:
```

```

## Required by:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/2552

AcK77:
This PR cleanup the `nsd` service a bit and implement / stub some calls (checked with RE):

- Added placeholders for lastest firmware calls,
- `GetSettingName` is removed because of a bad previous implementation (doesn't seems to be used by any games so there is no regression).
- `SetChangeEnvironmentIdentifierDisabled`, `WriteSaveDataToFsForTest`,`DeleteSaveDataOfFsForTest`, and `IsChangeEnvironmentIdentifierDisabled` are stubbed.
- `GetApplicationServerEnvironmentType` is implemented (closes #2247).

Now PGA TOUR 2K21 goes in-game (then deadlock):
Capture

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Nsd/IManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Nsd/IManager.cs
@@ -3,6 +3,7 @@ using Ryujinx.Cpu;
 using Ryujinx.HLE.Exceptions;
 using Ryujinx.HLE.HOS.Services.Settings;
 using Ryujinx.HLE.HOS.Services.Sockets.Nsd.Manager;
+using Ryujinx.HLE.HOS.Services.Sockets.Nsd.Types;
 using System.Runtime.InteropServices;
 using System.Text;
 
@@ -369,11 +370,11 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Nsd
 
             byte environmentType = identifier.Substring(0, 2) switch
             {
-                "lp" => 1,
-                "sd" => 2,
-                "sp" => 3,
-                "dp" => 4,
-                _    => 0
+                "lp" => (byte)ApplicationServerEnvironmentType.Lp,
+                "sd" => (byte)ApplicationServerEnvironmentType.Sd,
+                "sp" => (byte)ApplicationServerEnvironmentType.Sp,
+                "dp" => (byte)ApplicationServerEnvironmentType.Dp,
+                _    => (byte)ApplicationServerEnvironmentType.None
             };
 
             context.ResponseData.Write(environmentType);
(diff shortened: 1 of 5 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Nsd/IManager.cs
```diff
@@ -11,57 +12,64 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Nsd
...
-
-            if (result == ResultCode.Success)
-            {
-                byte[] settingNameBuffer = Encoding.UTF8.GetBytes(settingName + '\0');
-
-                context.Memory.Write(outputPosition, settingNameBuffer);
-            }
-
-            return result;
+            throw new ServiceNotImplementedException(this, context, false);
```
Thog, line 57: Any reasons of removing the implementation here?

Ryujinx.HLE/HOS/Services/Sockets/Nsd/IManager.cs
```diff
@@ -247,37 +284,113 @@ public ResultCode ReadSaveDataFromFsForTest(ServiceCtx context)
...
+            //       Returns ResultCode.NullOutputObject if failed.
+
+            ResultCode result = _fqdnResolver.GetEnvironmentIdentifier(out string identifier);
+
+            if (result != ResultCode.Success)
+            {
+                return result;
+            }
+
+            byte environmentType = identifier.Substring(0, 2) switch
```
Thog, line 370: an enum ``ApplicationServerEnvironmentType`` would be better here.

Thog (changes requested):

Ryujinx.HLE/HOS/Services/Sockets/Nsd/IManager.cs
AcK77: The current implementation is totally wrong and any games seems to currently uses it, so it's better to remove it.

Ryujinx.HLE/HOS/Services/Sockets/Nsd/Manager/FqdnResolver.cs
```diff
@@ -56,7 +31,10 @@ public ResultCode GetEnvironmentIdentifier(ServiceCtx context, out string identi
 
         public ResultCode Resolve(ServiceCtx context, string address, out string resolvedAddress)
         {
-            if (address != "api.sect.srv.nintendo.net" || address != "conntest.nintendowifi.net")
+            if (address != "api.sect.srv.nintendo.net"     ||
```
gdkchan, line 34: This check seems incorrect. `x != y || x != z` will always be true, regardless of the value in "x" because the value in "x" will always be either different from "y" or different from "z", since it can't be equal to both.

Thog (approved):

gdkchan (approved):
lgtm, thanks.

PR #2259, opened 2021-05-01
Update to FFmpeg 4.4.0

Thog:
As the title say


AcK77 (approved):

gdkchan (approved):
lgtm. Also unrelated, but it would be nice to silence ffmpeg stdout output, since it messes with console logs.

Thog:
> 
> 
> lgtm. Also unrelated, but it would be nice to silence ffmpeg stdout output, since it messes with console logs.

Would be nice to make an issue for that yeah or at least catch them and reformat so it could not break the logger coloring and all

PR #2266, opened 2021-05-05
ffmpeg: Redirect log output

### LINKED ISSUE #2263: Redirect or disable FFMPEG log output
#### Overview:
The FFMPEG log output, that reports errors in the decoding process, should be ideally redirected to the emulator and printed using the existing Ryujinx logging functions, that ways it is logged to the file too, and does not mess with the console logging.

#### Smaller Details: 
Currently FFMPEG is writing the log information directly to stdout, which messses the emulator console logging, see the image below for an example.
image
This is happening because there are two threads fighting for the output (FFMPEG's and the emulator log thread).
If not possible to redirect FFMPEG output, I think it should be just disabled.

Apparently this can se used to set a custom log callback:
[https://libav.org/documentation/doxygen/master/group__lavu__log.html](https://libav.org/documentation/doxygen/master/group__lavu__log.html)
In special, the `av_log_set_callback` function.

This makes the log output cleaner.
Having the output on the log file could be also useful to diagnose video decoding glitches.

Note for new contributors: The project that makes use of FFMPEG right now is `Ryujinx.Graphics.Nvdec.H264`.
We use a custom built version of FFMPEG on Windows which can be found at: https://github.com/Ryujinx/Ryujinx.Graphics.Nvdec.Dependencies
On other OSes it will just use whatever the user has installed.

AcK77:
This PR redirect the FFmpeg logging to our own log manager. It fixes #2263 and makes the log output cleaner.

image

### DIFF
--- a/Ryujinx.Common/Logging/LogClass.cs
+++ b/Ryujinx.Common/Logging/LogClass.cs
@@ -9,6 +9,7 @@ namespace Ryujinx.Common.Logging
         Cpu,
         Font,
         Emulation,
+        FFmpeg,
         Gpu,
         Hid,
         Host1x,
--- a/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
+++ b/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
@@ -7,8 +7,6 @@ namespace Ryujinx.Graphics.Nvdec.H264
 {
     unsafe class FFmpegContext : IDisposable
     {
-        private delegate av_log_set_callback_callback_func LoggingDelegate(void* p0, int level, string format, byte* vl);
-
         private readonly av_log_set_callback_callback _logFunc;
         private readonly AVCodec* _codec;
         private AVPacket* _packet;
--- a/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
+++ b/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Graphics.Nvdec.H264
 
         public FFmpegContext()
         {
-            _logFunc = Logging;
+            _logFunc = Log;
 
             // Redirect log output
             ffmpeg.av_log_set_level(ffmpeg.AV_LOG_MAX_OFFSET);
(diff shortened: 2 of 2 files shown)


Ryujinx.Common/Logging/LogClass.cs
```diff
@@ -9,6 +9,7 @@ public enum LogClass
         Cpu,
         Font,
         Emulation,
+        FFmpeg,
```
gdkchan, line 12: I would just use `Nvdec` rather than adding a new entry here.

Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
```diff
@@ -19,6 +28,42 @@ public FFmpegContext()
             _packet = ffmpeg.av_packet_alloc();
         }
 
+        private void Logging(void* p0, int level, string format, byte* vl)
```
gdkchan, line 31: I would just call the function `Log`, since its in line with how the other functions are named.

Ryujinx.Common/Logging/LogClass.cs
AcK77: Since it’s logs from the ffmpeg dependencies I think a new entry make sense, using nvdec isn’t totally true!

gdkchan: Either way is fine for me I guess, but I don't think we'll be using FFmpeg forever, so this entry will probably be removed at some point in the future.

gdkchan (approved):
lgtm, thanks.

Thog (approved):

Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
```diff
@@ -1,16 +1,25 @@
...
         private readonly AVCodec* _codec;
         private AVPacket* _packet;
         private AVCodecContext* _context;
 
         public FFmpegContext()
         {
+            _logFunc = Log;
+
+            // Redirect log output
+            ffmpeg.av_log_set_level(ffmpeg.AV_LOG_MAX_OFFSET);
```
CeruleanSky, line 20: Is formatting the trace log optimized away by csharp if unused or nearly instant by the Log function? If not maybe this can be set to an appropriate level by a switch determined by the Ryujinx.Common.Logging Level.  so  the Log function's `if (level > ffmpeg.av_log_get_level()){ return;}` can do so immediately instead of processing the log just to throw it away afterwards if a person isn't doing trace logging.

AcK77: Our Logger currently don’t use levels, they are more like categories which can be enabled/disabled separatly and not a real level as ffmpeg. That’s why I’ve selected the higher log level for ffmpeg and then parse all messages.

PR #2269, opened 2021-05-06
misc: Move configuration management to the Ryujinx project

Thog:
This decouple the Ryujinx configuration instance from every subprojects of Ryujinx and move it to the Ryujinx project.

This allow more flexibility on the way we create an emulation context and could allow new window backend to be created without relying on the UI only configuration.

As always need heavy testing.

NOTE: Some class weren't moved from the Ryujinx.Common project as those are way more integrated in some subsystem we have (like HLE HID)


Ryujinx.HLE/HLEConfiguration.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// Aspect Ratio applied to the renderer window by the SurfaceFlinger service.
+        /// </summary>
+        public AspectRatio AspectRatio { internal get; set; }
+
+        /// <summary>
+        /// An action called when HLE force a refresh of output after docked mode changed.
+        /// </summary>
+        public Action RefreshInputConfig { internal get; set; }
+
+        public HLEConfiguration(VirtualFileSystem virtualFileSystem,
```
AcK77, line 138: I think this should be named differently, but I have no idea about a potential name ^^'.

AcK77 (approved):
Instead of one remark, it's lgtm!

Ryujinx.HLE/HLEConfiguration.cs
```diff
@@ -0,0 +1,179 @@
...
+{
+    /// <summary>
+    /// HLE configuration.
+    /// </summary>
+    public class HLEConfiguration
+    {
+        /// <summary>
+        /// The virtual file system used by the FS service.
+        /// </summary>
+        /// <remarks>This cannot be changed after <see cref="Switch"/> instantiation.</remarks>
```
gdkchan, line 25: The parts that can't be changed after instantiation should be probably passed on the constructor (and be private), so that it's not even possible to change them afterwards. (Just a note, not saying that it has to be done *now*).

gdkchan: Nevermind, I missed that those are already `internal`. Not sure why they are part of `HLEConfiguration` though.

Thog: It's mostly to have all elements that form the "configuration" of the ``Switch`` instance in one place and to only have one thing to pass to create the emulation context.

PR #2279, opened 2021-05-09
Add `BIC/ORR Vd.T, #imm` fast path

FICTURE7:
Noticed this while skimming through the code.


ARMeilleure/Instructions/InstEmitSimdLogical.cs
```diff
@@ -64,10 +65,35 @@ public static void Bic_V(ArmEmitterContext context)
...
+            {
+                OpCodeSimdImm op = (OpCodeSimdImm)context.CurrOp;
+
+                int eSize = 8 << op.Size;
+
+                Operand d = GetVec(op.Rd);
+                Operand imm = eSize switch {
+                    16 => X86GetAllElements(context, (short)~op.Immediate),
+                    32 => X86GetAllElements(context, (int)~op.Immediate),
+                    _ => throw new Exception()
```
gdkchan, line 78: I think this should throw a more specific exception with a message. From what I recall, other places on the code throw `InvalidOperationException` on cases like this. But then I looked around, and the similar cases I could find (like the CLZ instruction implementation) is just setting the operand to null.

ARMeilleure/Instructions/InstEmitSimdLogical.cs
```diff
@@ -278,7 +304,32 @@ public static void Orr_V(ArmEmitterContext context)
...
+            {
+                OpCodeSimdImm op = (OpCodeSimdImm)context.CurrOp;
+
+                int eSize = 8 << op.Size;
+
+                Operand d = GetVec(op.Rd);
+                Operand imm = eSize switch {
+                    16 => X86GetAllElements(context, (short)op.Immediate),
+                    32 => X86GetAllElements(context, (int)op.Immediate),
+                    _ => throw new Exception()
```
gdkchan, line 317: Same comment.

ARMeilleure/Instructions/InstEmitSimdLogical.cs
FICTURE7: I think I'll change it to `InvalidOperationException` instead because CLZ implementation returns null when the instruction set extension is not supported. Perhaps we should look into decoupling these emitters from `OpCode/context.CurrOp` and friends in the hopes of someday removing/flattening them? 👉👈

FICTURE7: I was thinking of flattening the inheritance tree of `OpCode` and turning it into a struct and thinking of removing the dependency of these functions from `OpCode`. E.g these
https://github.com/Ryujinx/Ryujinx/blob/a8022ca3a1c9c2f312855c7676454507031be644/ARMeilleure/Instructions/InstEmitSimdHelper.cs#L795-L813
But after further thought it seems like more trouble than its worth.

gdkchan (approved):
Would be nice to include a exception message aswell, like `$"Invalid element size {eSize}.", but other than that lgtm.

riperiperi (approved):
Cool.

AcK77:
Needs a rebase!

PR #2282, opened 2021-05-12
Use branch instead of tailcall for recursive calls

FICTURE7:
Use a branch instead of doing a tailcall for recursive calls. This avoids having to store the dispatch address, storing the registers, going through the epilogue and then the reverse (loading the registers, going through the prologue) when the call is considered recursive.

The rejit check is moved down into the entry block so that the rejit behavior remains the same as before.


riperiperi (approved):
Nice one. It's worth nothing that returning from these recursive calls (assuming they do pop return addresses) will probably cause us to fall all the way back to the dispatcher, but it did that before anyways.

gdkchan (approved):
Nice one.

PR #2285, opened 2021-05-13
Fold constant offsets and group constant addresses

gdkchan:
This makes a few minor improvements to the x86 backend, reducing the generated code size a little bit, and potentially improving performance (nothing really noticeable though).

Changes:
- Allow `add rax, 0x123`followed by `add rax, rbx` to be fully contained into the memory operand. Previously it was only able to contain the `add rax, rbx` leaving the constant addition. Now it can be fully contained as `mov rax, [rax+rbx*1+0x123]`.
- On blocks where the same constant value is used more than once, if the constant can't fit on a 32-bit immediate, instead of inserting a `mov` before each use, it now inserts a single move per block, and keeps the constant in a register.
- Improved codegen for pre-indexed pair Arm loads and stores a little bit. Instead of adding the offset to the base register, producing the address, and then adding another offset to that address, it now adds both offsets to the base register directly. This avoids a dependency on the previous operation, and in some cases might also lead to less registers being needed (as the register used for the first address becomes free right after the first access).

### DIFF
--- a/ARMeilleure/CodeGen/X86/X86Optimizer.cs
+++ b/ARMeilleure/CodeGen/X86/X86Optimizer.cs
@@ -1,6 +1,7 @@
 ﻿using ARMeilleure.CodeGen.Optimizations;
 using ARMeilleure.IntermediateRepresentation;
 using ARMeilleure.Translation;
+using System.Collections.Generic;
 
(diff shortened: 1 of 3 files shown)


Thog (approved):

riperiperi (approved):
Nice one. 😌

PR #2288, opened 2021-05-16
Allow `LocalVariable` to be assigned more than once

FICTURE7:
This allows us to write flow controls like loops and if-elses with `LocalVariables` participating in phi nodes.


FICTURE7:
I also added `GetLocalNumber` to operand because I thought it would help make the distinction between when the code is referring a 32-bit constant or a local number.

gdkchan (approved):
lgtm, thanks. This limitation was a problem when implementing a few Arm instructions.

riperiperi (approved):
Nice one, this will prove useful. 😌

PR #2290, opened 2021-05-18
Move shader resource descriptor creation out of the backend

gdkchan:
This moves all the logic used to create buffer and texture descriptors that the GPU emulation uses to know the buffers and textures/images that should be bound, from the GLSL backend to the translator.

The advantage of doing this is that we don't need to duplicate the same logic when adding the new backends (SPIR-V and ARB).

This is just refactoring, shouldn't change behavior, however I do recommend testing some games to be sure that nothing regressed with those changes.


Ryujinx.Graphics.Shader/StructuredIr/StructuredProgram.cs
```diff
@@ -169,23 +139,12 @@ AstTextureOperation GetAstTextureOperation(TextureOperation texOp)
 
                 if (operation is TextureOperation texOp)
                 {
-                    if (texOp.Inst == Instruction.ImageLoad || texOp.Inst == Instruction.ImageStore)
```
gdkchan, line 172: `texOp.Inst == Instruction.ImageStore` was removed because on this path it checks if `dest != null`, and dest is always null for `ImageStore`, so the instruction should ever be `ImageStore` here.

Thog (approved):
lgtm

riperiperi (approved):
Looking great, tested a few games with res scale and buffer flush. All seems good, after the latest commits. Nice to see this moved somewhere more shared.

PR #2292, opened 2021-05-19
ffmpeg: Attempt to fix RootPath on some linux distributions

### LINKED ISSUE #2284: ffmpeg cant find its libraries on PopOS 20.10

even after https://github.com/Ryujinx/Ryujinx/commit/1769510a280ab33cb7b31a1f3e17fdb09f5b87cf, ffmpeg crashes on popos 20.10 not being able to find its libraries in `/lib`
looking more into it, setting the ffmpeg.RootPath to `/usr/lib/x86_64-linux-gnu/` in the `/Ryujinx/Program.cs` file fixes the issue on PopOS
big thanks to @EmulationFanatic for testing on PopOS 20.10

run any ryujinx version after 1.0.6857 on PopOS 20.10

AcK77:
This should fixes #2284 where the lib folder set in RootPath isn't the same between some linux distributions.
Since I don't use linux I can't really test it so feedbacks are really appreciate.

### DIFF
--- a/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
+++ b/Ryujinx.Graphics.Nvdec.H264/FFmpegContext.cs
@@ -16,8 +16,6 @@ namespace Ryujinx.Graphics.Nvdec.H264
 
         public FFmpegContext()
         {
-            SetRootPath();
-
             _logFunc = Log;
 
             // Redirect log output
@@ -32,7 +30,12 @@ namespace Ryujinx.Graphics.Nvdec.H264
             _packet = ffmpeg.av_packet_alloc();
         }
 
-        private void SetRootPath()
+        static FFmpegContext()
+        {
+            SetRootPath();
+        }
+
+        private static void SetRootPath()
         {
             if (OperatingSystem.IsLinux())
             {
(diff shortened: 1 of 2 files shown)


Niwu34:
Bildschirmfoto von 2021-05-19 23-52-11
works :DD

SeraUQ:
confirmed it fixes the ffmpeg issue on NixOS
great job in figuring it out :+1:

edisionnano:
Can confirm it works on glorious Arch
ack

EmulationFanatic:
Works great on Pop OS!

gdkchan (approved):
I'm not a fan of solutions that needs to spawn a process and parse its output as they are fragile and slow by nature, but I also don't know a better way to handle all the possible different paths where the user may have this installed on linux, so approving. I guess another possible solution would be just shiping FFMpeg ourselves for linux aswell.

SeraUQ:
works as expected on NixOS in both cases of ffmpeg being installed in the system and not :ok_hand:

narutorun-19:
It fixed the ffmpeg issue on Ubuntu 20.04. Thanks a lot~

Thog (approved):

PR #2296, opened 2021-05-20
Reduce `Operand[]` & `List<Operand>` allocations

FICTURE7:
This reduces the amount of `Operand[]` and `List<Operand>` allocated in `Node` by the making general case not require a `List<Operand>`, i.e 1 destination operand and 3 source operands are inlined in the `Node` structure itself. In the case where there are more than 1 destination and 3 sources, the extra operands are stored in an array. This array is partitioned such that the start of the array contains sources and then destinations.

Master:
Screenshot 2021-05-20 145743

PR:
Screenshot 2021-05-20 151200

If the extra complexity is concerning, I can write some unit tests. An alternative, perhaps simpler, would be to allocate a large `Operand[]` block and represent the sources and destinations by allocating 2 `ArraySegment` into that block using arena allocator. Then reset the arena after each compilation.

Something like that should probably be done for `BasicBlock._successors` because its allocating a lot.

### DIFF
--- a/ARMeilleure/IntermediateRepresentation/Node.cs
+++ b/ARMeilleure/IntermediateRepresentation/Node.cs
@@ -13,8 +13,10 @@ namespace ARMeilleure.IntermediateRepresentation
         private int _destCount;
         private Operand _dest0;
         private Operand _src0;
+#pragma warning disable CS0414
         private Operand _src1;
         private Operand _src2;
+#pragma warning restore CS0414
         private Operand[] _operands;
 
         public Node ListPrevious { get; set; }
@@ -72,9 +74,6 @@ namespace ARMeilleure.IntermediateRepresentation
             {
                 operands[i] = null;
             }
-
-            _ = _src1;
-            _ = _src2;
         }
 
         public Node With(Operand dest, int srcCount)
@@ -103,7 +102,7 @@ namespace ARMeilleure.IntermediateRepresentation
         {
             if ((uint)index >= _destCount)
             {
-                ThrowIndexOutOfRange();
+                ThrowIndexOutOfRange(nameof(index));
             }
 
             if (index < InlinedDestinationsCount)
@@ -126,7 +125,7 @@ namespace ARMeilleure.IntermediateRepresentation
         {
             if ((uint)index >= _srcCount)
             {
-                ThrowIndexOutOfRange();
+                ThrowIndexOutOfRange(nameof(index));
             }
 
             if (index < InlinedSourcesCount)
@@ -304,7 +303,7 @@ namespace ARMeilleure.IntermediateRepresentation
             int n = _operands != null ? _operands.Length : 0;
 
             // If no extra operands are needed, allow the array to be garbaged collected if its large enough.
-            if (count == 0 && n >= 4)
+            if (count == 0)
             {
                 _operands = null;
             }
@@ -424,6 +423,6 @@ namespace ARMeilleure.IntermediateRepresentation
             }
         }
 
-        private static void ThrowIndexOutOfRange() => throw new ArgumentOutOfRangeException("index");
+        private static void ThrowIndexOutOfRange(string param) => throw new ArgumentOutOfRangeException(param);
     }
 }
--- a/ARMeilleure/IntermediateRepresentation/Node.cs
+++ b/ARMeilleure/IntermediateRepresentation/Node.cs
@@ -1,201 +1,319 @@
 using System;
-using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 namespace ARMeilleure.IntermediateRepresentation
 {
     class Node : IIntrusiveListNode<Node>
     {
+        private const int InlinedDestinationsCount = 1;
+        private const int InlinedSourcesCount = 3;
+
+        private int _srcCount;
+        private int _destCount;
+        private Operand _dest0;
+        private Operand _src0;
+        private Operand _src1;
+        private Operand _src2;
+        private Operand[] _operands;
+
         public Node ListPrevious { get; set; }
         public Node ListNext { get; set; }
 
         public Operand Destination
         {
-            get => _destinations.Count != 0 ? GetDestination(0) : null;
+            get => _destCount != 0 ? GetDestination(0) : null;
             set => SetDestination(value);
         }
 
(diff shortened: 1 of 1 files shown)


FICTURE7:
Here are some metrics I collected in regards to how many `Operand`s do `Node`s take after `Translation` pass and `RegisterAllocation` pass.

For sources:
SourcesCount | Nodes after `Translation` | Nodes after `RegisterAllocation`
-- | -- | --
0 | 751369 | 480558
1 | 8443397 | 22593290
2 | 6180681 | 12213110
3 | 1491437 | 1391472
4 | 0 | 8

For destinations:
DestinationsCount | Nodes after `Translation` | Nodes after `RegisterAllocation`
-- | -- | --
0 | 2513091 | 11079686
1 | 14327743 | 25584642
2 | 0 | 258

This does not include `PhiNode`s because these will almost always allocate the extra array. So the majority of `Operand[]` allocations remaining are in ABI handling and large `PhiNode`s.

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -1,201 +1,319 @@
...
+
+            // Store _operands in a local variable to avoid bound checks below.
+            var operands = _operands;
+
+            for (int i = 0; i < operands.Length; i++)
+            {
+                operands[i] = null;
+            }
+
+            _ = _src1;
```
gdkchan, line 76: Whats the point of this?

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -1,201 +1,319 @@
...
+            }
+
+            if (index < InlinedDestinationsCount)
+            {
+                return ref Unsafe.Add(ref _dest0, index);
+            }
+
+            ref Operand extra = ref MemoryMarshal.GetArrayDataReference(_operands);
+
+            return ref Unsafe.Add(ref extra, Math.Max(0, _srcCount - InlinedSourcesCount) + index - InlinedDestinationsCount);
```
gdkchan, line 116: Why not just returns `ref _operands[...]` directly?

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -1,201 +1,319 @@
...
+            _srcCount = count;
+        }
 
-                AddUse(newOp);
+        private Operand[] EnsureExtra(int count)
+        {
+            int n = _operands != null ? _operands.Length : 0;
+
+            // If no extra operands are needed, allow the array to be garbaged collected if its large enough.
+            if (count == 0 && n >= 4)
```
gdkchan, line 307:
```suggestion
            if (count == 0 && n >= InlinedDestinationsCount + InlinedSourcesCount)
```

gdkchan: Assuming that I understand where the value came from.

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -305,5 +423,7 @@ private void RemoveUse(Operand op)
                 }
             }
         }
+
+        private static void ThrowIndexOutOfRange() => throw new ArgumentOutOfRangeException("index");
```
gdkchan, line 427: Maybe pass the argument name as parameter here? As it could easily change and the exception message could get out of sync with the code.

gdkchan (commented):
Always nice to see JIT memory allocations and usage being reduced.

ARMeilleure/IntermediateRepresentation/Node.cs
FICTURE7: This is the scuffed way of suppressing defined but unused warnings. I'll change it to how these warnings are normally suppressed in the project.

ARMeilleure/IntermediateRepresentation/Node.cs
FICTURE7: Sounds good.

ARMeilleure/IntermediateRepresentation/Node.cs
FICTURE7: Actually this should be changed to always release the array. This path is probably never hit with the current configuration, 1 destination and 3 sources inlined. IIRC the maximum size I've seen the extra array be is 2 for operations.

ARMeilleure/IntermediateRepresentation/Node.cs
FICTURE7: I was trying to avoid the bound checks since we already do it ourselves. I'll change it to `ref _operands[...]` directly if you think doing it this way is too scuffed/fragile.

gdkchan: In general I prefer clarity, but if this is being done for the sake of performance I'm not against it. Those functions are called a lot, so we need all the performance we can get there.

ARMeilleure/IntermediateRepresentation/Node.cs
```diff
@@ -1,201 +1,318 @@
...
+            _srcCount = count;
+        }
 
-                _sources[index] = newOp;
+        private Operand[] EnsureExtra(int count)
+        {
+            int n = _operands != null ? _operands.Length : 0;
 
-                AddUse(newOp);
+            // If no extra operands are needed, allow the array to be garbaged collected if its large enough.
```
riperiperi, line 305: `garbage`

riperiperi, line 305:
```suggestion
            // If no extra operands are needed, allow the array to be garbage collected if its large enough.
```

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
Looking great apart from the typo. 🥴

FICTURE7:
Looking back at this, I think a more generic data structure would still be fine, something like a `struct InlinedList<T>`. Then this could be re-used for `BasicBlock.Predecessors/Successors` and `Operand.Uses/Assignments` as well with minimal changes.

gdkchan:
Do you plan to do this change, or the PR can be considered ready for merge?

FICTURE7:
Oof, I meant to move this draft. I'll do it as apart of this PR itself.

FICTURE7:
Superseded by #2515.

PR #2298, opened 2021-05-20
Fix constant buffer array size when indexing is used and other buffer descriptor and resolution scale regressions

gdkchan:
Fixes a potential regression introduced on #2290 where the constant buffer array size would be incorrect on games that uses non-constant constant buffer slots. Note that the only game that is know to do that is Super Mario 3D All-Stars, and from what I was able to test, it is fine even without that fix, probably because all the constant buffers that it accesses on the shader using the non-constant indexing are also used elsewhere.

While testing, I also found a shader cache related issue. The `QueryConstantBufferUse` value is not saved on the cache, which causes it to use the default one of `0xffff`, which equals to 16 constant buffers used. This is too high, higher than the amount of uniform buffers allowed on the host, so it may cause a compilation error. It seems to be a existing issue that would trigger if the shader cache has to recompile from guest source. To fix it I just changed the default value to `0`, which seems to work on Super Mario Galaxy, but the correct fix here would be storing the value on the cache.

Other regressions fixed:
- Only report storage buffers (and constant buffers when non-constant indexing is used) that are actually used on the shader, instead of all the ones that consumes bindings.
- Add back some resolution scaling checks that were missing.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -36,7 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2290;
+        private const ulong ShaderCodeGenVersion = 2298;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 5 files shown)


Thog (approved):
Nice catch, I suppose we will need to create an issue about the cache missing this.

Tho if we do modify that, we will need to invalidate all shader cache around so might be worth looking into other missing spots when doing so.

riperiperi (approved):
LGTM, looks like the regressions in Xenoblade are sorted.

PR #2301, opened 2021-05-20
Fix dimensions check for scale eligibility

riperiperi:
Fix a regression with res scale where the old logic from the glsl emitter for detecting 2d array textures, but our new dimension count did not include a dimension for array index anyways.

Fixes res scale in the warriors games. 1D or 3D sampled textures should remain blacklisted.


gdkchan (approved):
Thanks for fixing this!

Thog (approved):
👍

PR #2302, opened 2021-05-21
Use a different method for out of bounds blit

gdkchan:
The guest OpenGL driver does out of bounds copies when copying textures for layout conversion. The way how this is supposed to work is that the copy should wrap to the next line of the texture and continue copying from there. Before this was being emulated on the host with two blits, which did work but had the issue that the second copy required the source texture to have one extra line. This extra line could sometimes have pixels fall into unmapped memory regions, which would cause a crash with a invalid memory region exception.

This uses a new approach that only requires a single bit. It simply adds an offset to the texture address, that way the copy is "in bounds" and no wrap to the next line is necessary. It also doesn't need to read any extra pixel.

This fixes the crash due to invalid memory regions being accessed on Rune Factory 5 and Shantae (the GBC emulator). Also tested another 2 OpenGL games, and they are fine with this change.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
@@ -1,3 +1,4 @@
+using Ryujinx.Common;
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Image;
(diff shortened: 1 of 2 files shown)


riperiperi (approved):
Very cool, seems to fix an issue in undertale where part of the sprite sheet for characters would be missing.

AcK77 (approved):

PR #2303, opened 2021-05-21
Fix non-independent blend state not being updated

gdkchan:
The entries for non-independent blend state were missing from the state table, which means it wouldn't detect modification for most of those blend registers, and this not update the blend state. This only affects OpenGL games as NVN and Vulkan always have independent blend enabled (which is working properly).

This fixes broken blend on some OpenGL games, like Code of Princess EX.
Before (from gamedb):
image
With PR:
image


riperiperi (approved):
Nice catch.

GamerzHell9137:
Arcaea fixed 
image

AcK77 (approved):

PR #2304, opened 2021-05-21
Add another Depth32F texture format variant

gdkchan:
This is used by Yokai Watch 1.
Together with #2302 and #2303, it improves rendering on this game.
Before (from gamedb):
image
With PR:
image

I actually implemented it quite some time ago, but the game was still very broken so I didn't submit it for this reason. Now it seems much better with the other fixes I linked.

Theres still a issue similar to what was recently fixed on Super Mario Party where the screen is overlayed by a solid colour.


riperiperi (approved):
Nice!

AcK77 (approved):

AeroRush:
Now it is playable, but it has slow loading times, is there any solution to that?

PR #2305, opened 2021-05-22
Improve accuracy of reciprocal step instructions

gdkchan:
Improves the accuracy of the FRECPS and FRSQRTS Arm instructions, by also handling the cases where one of the inputs is infinity and other is 0, as per the manual.

This fixes a bug on Rune Factory 5 where the characters skirt and other parts of the characters clothing would be lifted or stretched.
Before:
image
After:
image

Other changes:
- Fixed a oversight on the implementation of FRECPE and FRSQRTE rounding function, where it was using the CMPEQW SSE instruction instead of CMPEQD. It should use the dword variant since it is comparing 32-bit values. This didn't actually cause any major problems because it is just used to select between the rounded or non-rounded values, so it just causes the result to have more precision than it should in some rare cases. No changes noticed on the games tested.
- Renamed the rounding function mentioned above to match how the other functions are named.

What to test in Rune Factory 5:
- If all cases of broken clothes animation were fixed.
- If the was any performance loss with this change. It can cause some loss since the implementation now uses more instructions, but only if the instruction is used often enough.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
@@ -1518,9 +1518,12 @@ namespace ARMeilleure.Instructions
 
         public static void Frecps_S(ArmEmitterContext context) // Fused.
         {
-            if (Optimizations.FastFP && Optimizations.UseSse2)
+            if (Optimizations.FastFP && Optimizations.UseSse41)
             {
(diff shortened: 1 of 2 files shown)


riperiperi (approved):
Nice one! There doesn't seem to hage been much perceptible cost for doing this on Animal Crossing, but it's probably worth testing other games too.

jduncanator (approved):

PR #2306, opened 2021-05-23
Compare aligned size for largest mip level when considering sampler resize

riperiperi:
When selecting a texture that's a view for a sampler resize, we should take care that resizing it doesn't change the aligned size of any larger mip levels.

This PR covers two cases:
- When creating a view of the texture, we check that the aligned size of the view shifted up to level 0 still matches the aligned size of the container. If it does not, a copy dependency is created rather than resizing.
- When searching for a texture for sampler, textures that do _not_ match our aligned size when both are shifted up by its base level are not considered an exact match, as resizing the found texture will cause the mip 0 aligned size to change. It will create a copy dependency view instead.

Fixes graphical errors and crashes (on flush) in various Unity games that use render-to-texture. This likely also fixes infrequent graphical errors and crashes in Mario Kart 8 Deluxe.

Please test across various games to ensure there are no regressions.

### Before
image

```
Fatal error. Internal CLR error. (0x80131506)
```

### After:
image

image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
@@ -343,17 +343,16 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
-        /// Gets the aligned sizes of the specified texture information, shifted to the largest mip from a given level.
+        /// Gets the aligned sizes for the given dimensions, using the specified texture information.
         /// The alignment depends on the texture layout and format bytes per pixel.
         /// </summary>
         /// <param name="info">Texture information to calculate the aligned size from</param>
-        /// <param name="level">Mipmap level for texture views. Shifts the aligned size to represent the largest mip level</param>
-        /// <returns>The aligned texture size of the largest mip level</returns>
-        public static Size GetLargestAlignedSize(TextureInfo info, int level)
+        /// <param name="width">The width to be aligned</param>
+        /// <param name="height">The height to be aligned</param>
+        /// <param name="depth">The depth to be aligned</param>
+        /// <returns>The aligned texture size</returns>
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -327,6 +342,44 @@ public static bool SizeMatches(TextureInfo lhs, TextureInfo rhs, bool alignSizes
...
         }
 
+        /// <summary>
+        /// Gets the aligned sizes of the specified texture information, shifted to the largest mip from a given level.
+        /// The alignment depends on the texture layout and format bytes per pixel.
+        /// </summary>
+        /// <param name="info">Texture information to calculate the aligned size from</param>
+        /// <param name="level">Mipmap level for texture views. Shifts the aligned size to represent the largest mip level</param>
+        /// <returns>The aligned texture size of the largest mip level</returns>
+        public static Size GetLargestAlignedSize(TextureInfo info, int level)
```
gdkchan, line 352: Maybe create a method that takes width, height and depth as args, and make `GetLargestAlignedSize` and `GetAlignedSize` call it? To avoid repeating the code.

riperiperi:
Can confirm that this fixes missing reflection cubemap faces in races after the first, as I played... a lot of them. Frequency of objects appearing rainbow is likely significantly reduced.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
riperiperi: Sure.

gdkchan (approved):
lgtm, nice catch.

GamerzHell9137:
CLR Error fixed in Rune Factory 5

jduncanator (approved):

PR #2309, opened 2021-05-24
amadeus: Update to REV9

Thog:
This implements all the changes made with REV9 on 12.0.0.

### DIFF
--- a/Ryujinx.Audio/Renderer/Parameter/Effect/LimiterParameter.cs
+++ b/Ryujinx.Audio/Renderer/Parameter/Effect/LimiterParameter.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Copyright (c) 2019-2021 Ryujinx
 //
 // This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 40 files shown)


Ryujinx.Audio/Renderer/Parameter/Effect/LimiterParameter.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// The attack time.
+        /// <remarks>This is in microseconds.</remarks>
+        /// </summary>
+        public int AttackTimeUs;
+
+        /// <summary>
+        /// The release time.
+        /// <remarks>This is in microseconds.</remarks>
+        /// </summary>
+        public int ReleaseTimeUs;
```
AcK77, line 72: Sometimes you specified "Us" in the name, sometimes you don’t even if the comment say it’s in "micro second", maybe you should do the same thing everything for consistency? Imo removing "Us" is better.

AcK77 (approved):
Just one nit, instead of that it lgtm!

Ryujinx.Audio/Renderer/Parameter/Effect/LimiterParameter.cs
Thog: totally missed to clean that out thanks 👍

Thealexbarney (approved):
LGTM other than the `Memory<T>.Span` thing we talked about

PR #2324, opened 2021-05-30
Use single block translations in LCQ

FICTURE7:
Use single block translations in LCQ. This changes the rejit policy to be very similar to how it was before #2190. A flag `RejitHint`, is stored on the `NativeContext` which the rejit check will check to determine if it should count this transition. A transition is counted if it is a:

* Direct/Indirect call
* Indirect jump
* A continuation from an HCQ translation

Still haven't got the chance to thoroughly check this causes performance regression or collected metrics in regards to guest bytes jitted and stuff like that. Expect PTC profile info to blow up.

### DIFF
--- a/ARMeilleure/Decoders/Block.cs
+++ b/ARMeilleure/Decoders/Block.cs
@@ -5,23 +5,19 @@ namespace ARMeilleure.Decoders
 {
     class Block
     {
-        public ulong Address    { get; set; }
+        public ulong Address { get; set; }
         public ulong EndAddress { get; set; }
 
-        public Block Next   { get; set; }
+        public Block Next { get; set; }
         public Block Branch { get; set; }
 
         public bool Exit { get; set; }
 
         public List<OpCode> OpCodes { get; }
 
-        public Block()
+        public Block(ulong address)
         {
             OpCodes = new List<OpCode>();
-        }
-
-        public Block(ulong address) : this()
-        {
             Address = address;
         }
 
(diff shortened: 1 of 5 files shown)


FICTURE7:
Did some measurements on Dead Cells:

Master:
&nbsp; | **LCQ** | **HCQ** | **LCQ+HCQ**
-- | -- | -- | --
**Translations** | 22043 | 8551 | 30594
**In Bytes** | 10429460 | 4092112 | 14521572
**Out Bytes** | 159602910 | 58042421 | 217645331
**Out/In Ratio** | 15.30308472 | 14.18397664 | 14.98772523

PR:
&nbsp; | **LCQ** | **HCQ** | **LCQ+HCQ**
-- | -- | -- | --
**Translations** | 167041 | 8414 | 175455
**In Bytes** | 3455356 | 3473880 | 6929236
**Out Bytes** | 90803637 | 45810040 | 136613677
**Out/In Ratio** | 26.27909744 | 13.18699552 | 19.71554685

The total number guest bytes consumed and host bytes produced is reduced overall as expected. The number of translations tiering up to HCQ decreases a little bit which is also expected, worth looking into which cases are not getting tiered up anymore. Could also be that execution took different paths between the two runs.

gdkchan:
Nice, it would be good to also have a comparison of the boot times on master VS with this change (I would expect them to decrease with this change).

FICTURE7:
Oof, thanks for testing. I'll move this to draft until I get some spare time to look deeper into this.

FICTURE7:
IIRC POWER reduced both compile time and execution time, to the point where doing single block decodes is not always a win. The HybridAllocator also shines much better because the BBs are longer.

After further thought, I think a better long term solution would be to improve the decoder at detecting function boundaries.

PR #2328, opened 2021-05-31
Update to LibHac 0.13.1

### LINKED ISSUE #2272: fsp-srv IFileSystemProxy: 34 (GetCacheStorageSize) is not implemented
fsp-srv IFileSystemProxy: 34 (GetCacheStorageSize) is not implemented

[Switchbrew](https://switchbrew.org/wiki/Filesystem_services#fsp-srv) | [SwIPC](https://reswitched.github.io/SwIPC/ifaces.html#nn::fssrv::sf::IFileSystemProxy(34))

## Service description:
```
N/A
```

## Required by:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/165 v1.0.26

Thealexbarney:
Update the LibHac dependency to version 0.13.1. This brings a ton of improvements and changes such as:
- Refactor `FsSrv` to match the official refactoring done in FS.
- Change how the `Horizon` and `HorizonClient` classes are handled. Each client created represents a different process with its own process ID and client state.
- Add FS access control to handle permissions for FS service method calls.
- Add FS program registry to keep track of the program ID, location and permissions of each process.
- Add FS program index map info manager to track the program IDs and indexes of multi-application programs.
- Add all FS IPC interfaces.
- Rewrite `Fs.Fsa` code to be more accurate.
- Rewrite a lot of `FsSrv` code to be more accurate.
- Extend directory save data to store `SaveDataExtraData`
- Extend directory save data to lock the save directory to allow only one accessor at a time.
- Improve waiting and retrying when encountering access issues in `LocalFileSystem` and `DirectorySaveDataFileSystem`.
- More `IFileSystemProxy` methods should work now.
- Probably a bunch more stuff.

On the Ryujinx side:
- Forward most `IFileSystemProxy` methods to LibHac.
- Register programs and program index map info when launching an application.
- Remove hacks and workarounds for missing LibHac functionality.
- Recreate missing save data extra data found on emulator startup.
- Create system save data that wasn't indexed correctly on an older LibHac version.

`FsSrv` now enforces access control for each process. When a process tries to open a save data file system, FS reads the save's extra data to determine who the save owner is and if the caller has permission to open the save data. Previously-created save data did not have extra data created when the save was created.
With access control checks in place, this means that processes with no permissions (most games) wouldn't be able to access their own save data. The extra data can be partially created from data in the save data indexer, which should be enough for access control purposes.
This should be tested to make sure the extra data creation works properly with everything.

Needs testing with games that previously had FS-related issues. Many issues might have gone away due to general improvements or added IPC methods.

Closes #2272


Sekkei:
> Extend directory save data to lock the save directory to allow only one accessor at a time.
Could I propose to not do that ? In case of savefile manipulation while running. I belive an alternative secure writing method would be to create a new file and if successful written, it is allowed to delete (to bin?) and rename the file(s)

jduncanator:
> > Extend directory save data to lock the save directory to allow only one accessor at a time.
> 
> Could I propose to not do that ? In case of savefile manipulation while running. I belive an alternative secure writing method would be to create a new file and if successful written, it is allowed to delete (to bin?) and rename the file(s)

I believe @Thealexbarney's goal with libhac is to match 1:1 what the actual console firmware does, compromising this in certain ways to make things easier defeats the purpose of that, and can introduce weird bugs.

Thealexbarney:
> > Extend directory save data to lock the save directory to allow only one accessor at a time.
> 
> Could I propose to not do that ? In case of savefile manipulation while running. I belive an alternative secure writing method would be to create a new file and if successful written, it is allowed to delete (to bin?) and rename the file(s)

Not exactly sure what you mean, but I'm talking about allowing the save data to only be opened once in the context of Horizon.

For save image files opening the file for write automatically means that nobody else can open the file at the same time. This isn't true for directories, so the directory save data will keep a read/write handle to a `.lock` file in the save directory to provide the same behavior. This ends up fixing a few weird bugs in some games, as Nintendo's `DirectorySaveDataFileSystem` design ends up basically failing dramatically if the same save is opened multiple times at once.

You can always go ahead and edit files while the game is running, but that's basically undefined behavior. No guarantees can be made that stuff will continue to work properly.

Sekkei:
> allowing the save data to only be opened once in the context of Horizon.

That was not clear for me from only reading this post, excuse the misunderstanding

Thealexbarney:
Didn't feel like blocking a LibHac release on this any longer, so this PR's now using the release LibHac 0.13.0 build. The rebase seems to have gone fine, so the PR should be mergeable in its current form

Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
```diff
@@ -211,6 +218,31 @@ public ResultCode GetSaveDataSize(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(27)] // 5.0.0+
+        // CreateCacheStorage(u16 index, s64 save_size, s64 journal_size) -> ()
```
AcK77, line 222:
```suggestion
        // CreateCacheStorage(u16 index, s64 save_size, s64 journal_size) -> u64, u64
```

Ryujinx.HLE/HOS/Services/Arp/LibHacIReader.cs
```diff
@@ -47,5 +42,26 @@ public Result GetApplicationControlPropertyWithApplicationId(out ApplicationCont
...
+
+        public LibHacArpServiceObject(LibHacIReader serviceObject)
+        {
+            _serviceObject = serviceObject;
+        }
+
+        public Result GetServiceObject(out object serviceObject)
+        {
+            serviceObject = _serviceObject;
+            return Result.Success;
```
AcK77, line 64: Add a new empty line before

Ryujinx.HLE/HOS/Services/Fs/IFileSystemProxy.cs
```diff
@@ -94,58 +101,69 @@ public ResultCode OpenBisFileSystem(ServiceCtx context)
...
+            Result result = _baseFileSystemProxy.Target.OpenBisFileSystem(out ReferenceCountedDisposable<IFileSystem> fileSystem, in path, bisPartitionId);
+            if (result.IsFailure()) return (ResultCode)result.Value;
 
             MakeObject(context, new FileSystemProxy.IFileSystem(fileSystem));
 
             return ResultCode.Success;
         }
 
+        [CommandHipc(12)]
+        public ResultCode OpenBisStorage(ServiceCtx context)
```
AcK77, line 115: Could be nice to get the symbol for consistency (and for other calls which doesn't have one too)

AcK77 (approved):
Lgtm except some nits :)! Awesome works as usual!

GamerzHell9137:
Is it known which games have FS-related issues?

AcK77:
Reminder: Fix this https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Services/Mii/MiiDatabaseManager.cs#L104 before merging.

Thealexbarney:
I had to add a fix for correcting the save data extra data for miiedit since the commit that allows running it was merged before this one.

Thealexbarney:
Also I'm making a bugfix libhac release, so once that's on NuGet commit 5227752 can be removed

Thog (approved):

PR #2329, opened 2021-05-31
Fix shaders with mixed PBK and SSY addresses on the stack

gdkchan:
PBK and SSY shader instruction pushes an address to a stack, which is poped by BRK and SYNC instructions respectively. When the addresses on stack are mixed, only the address for the matching instruction should be used.
Consider the following example:
- PBK pushes address 0xA00.
- SSY pushes address 0xB00.
- BRK is executed.
The current behaviour is that the shader translator assumes that BRK would jump to 0xB00, which is incorrect as the address was pushed by a SSY instruction, not PBK. This changes makes it discard the 0xB00 address, then use the 0xA00 since the type of this instruction matches.

Fixes pink sky on Dead or Alive Xtreme 3.
Before (from gamedb):
image
PR:
image
Things to look for while testing:
- Any regression caused by this change.
- Shader compilation being slower, since it needs to do a bit more work on the decoder now (in special the function that propagates the PBK and SSY addresses). I wouldn't expect a noticeable impact though.

### DIFF
--- a/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
@@ -278,7 +278,7 @@ namespace Ryujinx.Graphics.Shader.Decoders
                 OpCode op = makeOp(emitter, opAddress, opCode);
 
                 // We check these patterns to figure out the presence of bindless access
-                hasBindless |= (op is OpCodeImage image && image.IsBindless) || 
+                hasBindless |= (op is OpCodeImage image && image.IsBindless) ||
                     (op is OpCodeTxd txd && txd.IsBindless) ||
                     (op is OpCodeTld4B) ||
                     (emitter == InstEmit.TexB) ||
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Shader/Decoders/Decoder.cs
```diff
@@ -332,35 +332,39 @@ private enum RestoreType
...
 
             public PathBlockState(Block block)
             {
-                Block         = block;
-                _restoreType  = RestoreType.None;
-                _restoreValue = 0;
+                Block             = block;
+                _restoreType      = RestoreType.None;
+                _restoreValue     = 0;
+                _restoreMergeType = 0;
```
jduncanator, line 344: Something feels a bit "icky" assigning an integer to an enum type. I'd prefer this explicitly use `MergeType.Brk`, or `default(MergeType)` rather than blindly assigning 0.

riperiperi (approved):
Neat improvement here. 👍

Ryujinx.Graphics.Shader/Decoders/Decoder.cs
gdkchan: I assigned 0 because the value doesn't matter in this case as its not used. Anyway I've changed it to `default` instead.

Thog (approved):

PR #2330, opened 2021-05-31
Inherit buffer tracking handles rather than recreating on resize.

riperiperi:
This greatly speeds up games that constantly resize buffers, and removes stuttering on games that resize large buffers occasionally:

- Large improvement on Super Mario 3D All-Stars. Slowdown now is just due to the creation of the OGL buffer resources. (#1663 needed for best performance)
- Improvement to Hyrule Warriors: AoC, and UE4 games. These games still stutter in general due to texture creation/loading.
- Small improvement to other games, potential 1-frame stutters avoided.

`ForceSynchronizeMemory`, which was added with POWER, is no longer needed. Some tests have been added for the MultiRegionHandle.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
@@ -3,6 +3,8 @@ using Ryujinx.Graphics.GAL;
 using Ryujinx.Memory.Range;
 using Ryujinx.Memory.Tracking;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Ryujinx.Graphics.Gpu.Memory
 {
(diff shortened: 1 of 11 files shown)


Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -78,13 +81,45 @@ public Buffer(GpuContext context, ulong address, ulong size)
...
             }
             else
             {
                 _memoryTracking = context.PhysicalMemory.BeginTracking(address, size);
+
+                if (baseHandles != null)
+                {
+                    _memoryTracking.Reprotect(false);
+
+                    foreach (IRegionHandle handle in baseHandles)
```
gdkchan, line 113: Maybe its possible to simplify this? Like, with `anyDirty |= handle.Dirty;` (initially set to false), then `_memoryTracking.Reprotect(anyDirty);`. But maybe I'm missing something here.

riperiperi: The concern is that a part of the buffer is dirtied while we're iterating over this. Imagine the following situation:

- Handle 1 dirty flag read (false)
- Handle 2 dirty flag read (false)
- Handle 3 dirty flag read (false)
- Handle 2 is dirtied
- Handle 4 dirty flag read (false)

Without the new handle also being active, the new write to handle 2 will be dropped.

ericdallo:
any news on this @riperiperi ?

gdkchan (approved):

Thog (approved):

PR #2331, opened 2021-05-31
Use Quads on OpenGL host when supported.

riperiperi:
Improves OpenGL performance on FAST RMX and Xenoblade DE/2. Will probably only work on NVIDIA GPUs, but the emulated quads path will still be valid for other GPUs. Note: SLOW RMX gets a bit faster in handheld mode.

I'd recommend checking on platforms without supported host quads (AMD? Intel? Mesa?) to make sure a GL error is actually thrown when attempting GL.Begin(PrimitiveType.Quads).


eddylance:
I found some very significative improvements in Xenoblade Chronicles DE. From 35fps to 50 on cities. Testing on a 5800x and a GTX 1070 Ti. Will go on testing.

eddylance:
I got a blown up texture in XCDE after a cutscene, which fixed when I restarted the emulation and didn't appear again. Never got this kind of error on LDN 2.3, but don't know if it's specific to this build, since I couldn't reproduce it.

ReinUsesLisp:
All desktop platforms support quads natively.

gdkchan (approved):
Seems good.

Thog (approved):
lgtm, working fine here

riperiperi:
> All three major desktop platforms support quads natively.
Cool. gdk said a while back that an older intel driver did not support quads, so it's probably better to be safe than sorry with the check.

PR #2333, opened 2021-05-31
Add TouchScreen Manager

emmauss:
This pr moves touchscreen updates to the Input project.


Ryujinx.Input/IMouseDriver.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Drawing;
+using System.Numerics;
+
+namespace Ryujinx.Input
+{
+    public interface IMouseDriver: IGamepadDriver, IDisposable
+    {
+        public bool[] Buttons { get; }
```
Thog, line 9: Remove all public here, those are uneeded in an interface

Thog: I kind would prefer that you do something like what was done in the ``IGamepad`` and ``IKeyboard`` interface that create a snapshot instance instead of having access directly given like that imo.

That would allow to avoid possible weird thing happening if any values change during ``TouchScreenManager`` update.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -240,7 +240,9 @@ private MainWindow(Builder builder) : base(builder.GetObject("_mainWin").Handle)
 
             Task.Run(RefreshFirmwareLabel);
 
-            InputManager = new InputManager(new GTK3KeyboardDriver(this), new SDL2GamepadDriver());
+            var mouseDriver = new Gtk3MouseDriver(this);
```
Thog, line 243:
```suggestion
            Gtk3MouseDriver mouseDriver = new Gtk3MouseDriver(this);
```

Thog: Does this actually work on the render widget?
I remember having to register on it instead of the main one for keyboard...
~~Also did you checked if the touch area is the same size as the widget one in the end? I think it could be possible to have mapping issue on the main window.~~ Just saw that you grab the allocation size after 👍

Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -413,6 +386,7 @@ public void Start()
         public void Exit()
         {
             NpadManager?.Dispose();
+            TouchScreenManager?.Dispose();
```
Thog, line 389: Invert with ``NpadManager`` dispose (as you initalize ``TouchScreenManager`` after ``NpadManager`` in ``Initialize``.

Thog (changes requested):
One less thing to have duplicate between UIs 👍 
Just have some comments but otherwise the code lgtm, will need to test it

Ryujinx/Ui/MainWindow.cs
emmauss: I does work on the render widget, but I'll switch it to the widget for consistency.

Ryujinx.Input/IMouseDriver.cs
emmauss: I've split it into driver and device classes

AcK77 (changes requested):

emmauss:
reviews addressed.

Ryujinx/Input/GTK3/GTK3Mouse.cs
```diff
@@ -0,0 +1,83 @@
...
+        public Size ClientSize => _driver.GetClientSize();
+
+        public Vector2 GetPosition()
+        {
+            return _driver.CurrentPosition;
+        }
+
+        public GamepadStateSnapshot GetMappedStateSnapshot()
+        {
+            throw new System.NotImplementedException();
```
AcK77, line 35: What about `using System;`?

AcK77 (approved):

Ryujinx/Input/GTK3/GTK3Mouse.cs
emmauss: done

Thog (approved):

PR #2335, opened 2021-06-01
Fix texture blit off-by-one errors

gdkchan:
NVN adds an offset to the source X/Y coordinates of the blit, in order to ensure that the blit will be "centered" (sampling from the center of the pixel rather than the corner). This is done by adding `RatioX / 2.0` and `RatioY / 2.0` to the X/Y coordinates, where `RatioX` and `RatioY` are calculated as `SourceRegionX / DestinationRegionX` and `SourceRegionY / DestinationRegionY` respectively. This changes removes said offset by subtracting the offset value from the coordinates.

This fixes some off-by-one errors that caused the destination textures to have a black border. When used to blit mipmap levels, this could propagate and cause the whole texture to be black in the end, as it would usually use the previous level as source.

This only affects copies between regions of different sizes, as when the sizes are equal, the offset is 0.5 (which is truncated to 0).

Fixes screen being too bright on Fast RMX.

Master:
image
PR:
image

Fixes Mii character face on Mario Kart 8 Deluxe.

Master:
image
PR:
image

Note: Issue found by @riperiperi.
I recommend testing on a few games, including those using the OpenGL and Vulkan APIs, as I'm not sure how those APIs handles those blits.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/MethodCopyTexture.cs
@@ -24,11 +24,29 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
             var control = state.Get<CopyTextureControl>(MethodOffset.CopyTextureControl);
 
-            int srcX1 = (int)(region.SrcXF >> 32);
-            int srcY1 = (int)(region.SrcYF >> 32);
+            bool originCorner = control.UnpackOriginCorner();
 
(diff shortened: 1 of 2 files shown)


Thog (approved):

riperiperi (approved):
Looks good here, in the NVN and OGL games I have at least.

VitoItalianGamer:
I think this post was the solution for me but the download link for windows is broken. Can you help me please?

AcK77:
This is already merged in our main build. GitHub is for development so if you’ll need support or anything else please join our discord server. Thanks.

PR #2336, opened 2021-06-02
Workaround for cb indexing error on shader cache rebuild

gdkchan:
This contains a workaround for a bug that can occur if a shader cache rebuild is triggered (due to a cache version change) on a shader that uses constant buffer slot indexing. Currently the shader cache does not store the constant buffer use information, so it doesn't know the array size. This changes the default mask to one with the 13 least significant bits set, since a maximum of 13 uniform buffers can be used on OpenGL (on NVIDIA, other vendors are usually a bit higher).

Additionally, this also has a different change that validates the constant buffer slots, and logs a warning if its invalid (and then replaces the value by 0). This prevents invalid constant buffer slots to be output on the buffer descriptor, which would later cause a `IndexOutOfRangeException` when attempting to bind it. Of course this should never happen, and is a indication that the shader code is corrupted, but at least it prevents the emulator from crashing for those invalid shaders. I'm happy to remove the change if its not seem as appropriate.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -36,7 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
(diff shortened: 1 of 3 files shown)


gdkchan:
The first issue will be fixed *properly* by #3194. I can open a separate PR for the second one if desired.

PR #2339, opened 2021-06-02
Support bindless textures with separate constant buffers for texture and sampler

gdkchan:
Some games stores the handles on different constant buffers, which was not supported.
This extends the shader bindless elimination to support this case, and also extends the GPU emulation to support reading the handles from different buffers.
Fixes black screen on Final Fantasy XII. Other games with black textures might be affected as well.
image
image
I believe updating the shader cache version is not needed, as shaders with bindless access are not cached right now anyway.
Should be tested on other games that uses bindless textures to ensure they didn't regress.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -12,6 +12,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         private const int HandleHigh = 16;
         private const int HandleMask = (1 << HandleHigh) - 1;
 
+        private const int SlotHigh = 16;
+        private const int SlotMask = (1 << SlotHigh) - 1;
+
         private GpuContext _context;
 
         private bool _isCompute;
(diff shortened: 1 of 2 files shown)


RyzenDew:
bioshock 1-2-3 all render now
with slight texture issues

meisiek:
XCOM 2 menu is now longer invisible and you can get a bit further in game.
Civilization VI shows some graphics as well, but its all red colored.

Couldnt find anything broken. Looks good for me.

meisiek:
Crash Bandicoot 4 looks worse after the intro:

grafik

Master looks much better, but the colors are also a bit off there.

EmulationFanatic:
> Crash Bandicoot 4 looks worse after the intro:
> 
> grafik
> 
> Master looks much better, but the colors are also a bit off there.

Crash Bandicoot 4 looks differently almost every time you boot it. Boot it on master a few times and you'll see what I mean.

riperiperi (approved):
Nice catch, looks good to me.

AcK77 (approved):

PR #2358, opened 2021-06-10
Unbind input on middle click

ooa113y:
This PR allows to "unbind" inputs in the `ControllerWindow` by middle-clicking while an input is selected.

This is my first code contribution to Ryujinx, so I apologize if the code is not compliant with some internal standard, or even if it's just outright bad for some reason that I don't understand.

### DIFF
--- a/Ryujinx/Ui/Windows/ControllerWindow.cs
+++ b/Ryujinx/Ui/Windows/ControllerWindow.cs
@@ -105,6 +105,7 @@ namespace Ryujinx.Ui.Windows
         private IGamepadDriver _gtk3KeyboardDriver;
         private IGamepad _selectedGamepad;
         private bool _mousePressed;
+        private bool _middleMousePressed;
 
         public ControllerWindow(MainWindow mainWindow, PlayerIndex controllerId) : this(mainWindow, new Builder("Ryujinx.Ui.Windows.ControllerWindow.glade"), controllerId) { }
 
@@ -112,6 +113,7 @@ namespace Ryujinx.Ui.Windows
         {
             _mainWindow = mainWindow;
             _mousePressed = false;
+            _middleMousePressed = false;
             _selectedGamepad = null;
 
             // NOTE: To get input in this window, we need to bind a custom keyboard driver instead of using the InputManager one as the main window isn't focused...
@@ -856,6 +858,11 @@ namespace Ryujinx.Ui.Windows
                     {
                         button.Label = pressedButton;
                     }
+                    else if (_middleMousePressed)
+                    {
+                        button.Label = "Unbound";
+                        _middleMousePressed = false;
+                    }
 
                     ButtonPressEvent -= MouseClick;
                     keyboard.Dispose();
@@ -883,6 +890,10 @@ namespace Ryujinx.Ui.Windows
         private void MouseClick(object sender, ButtonPressEventArgs args)
         {
             _mousePressed = true;
+            if (args.Event.Button == 2)
+            {
+                _middleMousePressed = true;
+            }
         }
 
         private void SetProfiles()
--- a/Ryujinx/Ui/Windows/ControllerWindow.cs
+++ b/Ryujinx/Ui/Windows/ControllerWindow.cs
@@ -860,6 +860,7 @@ namespace Ryujinx.Ui.Windows
                     {
                         button.Label = pressedButton;
                     }
+
                     _middleMousePressed = false;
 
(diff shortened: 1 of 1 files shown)


Ryujinx/Ui/Windows/ControllerWindow.cs
```diff
@@ -856,6 +858,11 @@ private void HandleButtonPressed(ToggleButton button, bool forStick)
                     {
                         button.Label = pressedButton;
                     }
+                    else if (_middleMousePressed)
```
emmauss, line 861: this can be simplified with a ternary conditional operator.

Ryujinx/Ui/Windows/ControllerWindow.cs
```diff
@@ -105,13 +105,15 @@ public class ControllerWindow : Window
...
         private bool _mousePressed;
+        private bool _middleMousePressed;
 
         public ControllerWindow(MainWindow mainWindow, PlayerIndex controllerId) : this(mainWindow, new Builder("Ryujinx.Ui.Windows.ControllerWindow.glade"), controllerId) { }
 
         private ControllerWindow(MainWindow mainWindow, Builder builder, PlayerIndex controllerId) : base(builder.GetObject("_controllerWin").Handle)
         {
             _mainWindow = mainWindow;
             _mousePressed = false;
+            _middleMousePressed = false;
```
emmauss, line 116: This is not necessary.

emmauss (changes requested):

Ryujinx/Ui/Windows/ControllerWindow.cs
jduncanator: Technically neither is `_mousePressed = false` above it. Either both should be removed, or this change left.

Ryujinx/Ui/Windows/ControllerWindow.cs
ooa113y: Sorry but this confuses me slightly.

There are two conditions here, one that checks for `pressedButton` and one that checks for `_middleMousePressed`, with different results. Should I merge them together into something like `button.Label = buttonPressed != "" ? pressedButton : _middleMousePressed ? "Unbound": button.Label`? To me personally that seems much more confusing to read, but I'll do it of course.

Or do you mean just this specific addition? In that case we end up with `button.Label = _middleMousePressed ? "Unbound" : button.Label` which is still somewhat awkward in my eyes since it repeats `button.Label` twice.

I'm not against changing this, I'm just not entirely sure *what* the change should be here 😅

Ryujinx/Ui/Windows/ControllerWindow.cs
ooa113y: Removed both of these, confirmed the window functions fine without them.

Ryujinx/Ui/Windows/ControllerWindow.cs
```diff
@@ -852,10 +852,15 @@ private void HandleButtonPressed(ToggleButton button, bool forStick)
...
-                    if (pressedButton != "")
+                    if (_middleMousePressed)
+                    {
+                        button.Label = "Unbound";
+                    }
+                    else if (pressedButton != "")
                     {
                         button.Label = pressedButton;
                     }
+                    _middleMousePressed = false;
```
AcK77, line 863: Add a new empty line before

Ryujinx/Ui/Windows/ControllerWindow.cs
```diff
@@ -883,6 +888,10 @@ private void ButtonForStick_Pressed(object sender, EventArgs args)
         private void MouseClick(object sender, ButtonPressEventArgs args)
         {
             _mousePressed = true;
+            if (args.Event.Button == 2)
```
AcK77, line 891: `_middleMousePressed = args.Event.Button == 2`

AcK77 (changes requested):

Ryujinx/Ui/Windows/ControllerWindow.cs
ooa113y: Done.

Ryujinx/Ui/Windows/ControllerWindow.cs
ooa113y: Done.

AcK77 (approved):

riperiperi (approved):
Nice one!

PR #2360, opened 2021-06-11
Fix separate bindless sampler at offset 0

gdkchan:
When the (separate) sampler bindless handle comes from the offset 0 of the constant buffer, it was being ignored which caused the wrong sampler to be used. This correct the bug by adding 1 to the offset on the shader translator, and then subtracting 1 on the GPU emulator to ensure that it is never 0 if a separate bindless sampler is used.

Fix incorrect sampling on Final Fantasy XII.
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -473,7 +473,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     ? bufferManager.GetComputeUniformBufferAddress(samplerBufferIndex)
                     : bufferManager.GetGraphicsUniformBufferAddress(stageIndex, samplerBufferIndex);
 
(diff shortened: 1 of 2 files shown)


RyzenDew:
image
not sure if it's an issue with PR but papers on the ground are black from a far and then good up close

ryulovesjinx:
I do not know if it is a coincidence, but I test most of the PRs to see if it will fix Dragon Ball FighterZ characters selection screen/loading battle. And this one stop the flickering at least. Don't know if it is a coincidence though. The screen is only red now.
(Tested on Ubuntu 20.04)
Screenshot from 2021-06-11 09-54-59

RyzenDew:
image
his hip area is missing also

RyzenDew:
image
fixes random black graphic issues

riona222:
I Test ffvii still the same problem black secan

Thog (approved):
lgtm

riperiperi (approved):

PR #2362, opened 2021-06-12
Pass all inputs when geometry shader passthrough is enabled

gdkchan:
Not passing those inputs was generating a link error on the shaders, as the fragment shader was consuming inputs that were not explicitly passed from the geometry shader. This change makes it pass all possible inputs when the passthrough is enabled.

Fixes some missing graphics on Game Builder Garage.
Before:
image
After:
image
Note: Doesn't fix apple not showing on the first tutorial, this looks more like a logic bug rather than just something not being rendered (but existing).


Leischii:
From user on discord:
Also fixes missing graphics in the labo vr kit software
https://cdn.discordapp.com/attachments/480497101676216340/853267532537659412/unknown.png

Thog (approved):

riperiperi (approved):
Nice catch.

PR #2367, opened 2021-06-14
End shader decoding when reaching a block that starts with an infinite loop (after BRX)

riperiperi:
The NV shader compiler puts these at the end of shaders. This is needed when a shader includes a BRX instruction, as we continue decoding past all exits in order to catch potential branch targets. Before, this would terminate when it encountered a 0, but some shaders did not have a 0 at the end and ended up decoding forever, possibly because the shader data _did_ end at the infinite loop, and the "instruction" afterwards was just leftover data from earlier rather than 0.

Note that a better solution would probably be identifying the BRX targets rather than continuing to decode like this, but this workaround is functional for now.

A few games with complex shaders should be tested, maybe Fire Emblem to make sure it still works. Fixes Dark Devotion.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -36,7 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2329;
+        private const ulong ShaderCodeGenVersion = 2330;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -36,7 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2330;
+        private const ulong ShaderCodeGenVersion = 2367;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
--- a/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
@@ -9,6 +9,8 @@ namespace Ryujinx.Graphics.Shader.Decoders
 {
     static class Decoder
     {
+        public const ulong ShaderEndDelimiter = 0xe2400fffff87000f;
+
         public static Block[][] Decode(IGpuAccessor gpuAccessor, ulong startAddress, out bool hasBindless)
         {
             hasBindless = false;
@@ -190,7 +192,7 @@ namespace Ryujinx.Graphics.Shader.Decoders
 
             ulong inst = gpuAccessor.MemoryRead<ulong>(startAdddress + currBlock.EndAddress);
 
-            return inst != 0UL;
+            return inst != 0UL && inst != ShaderEndDelimiter;
         }
 
         private static bool BinarySearch(List<Block> blocks, ulong address, out int index)


Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -36,7 +36,7 @@ class ShaderCache : IDisposable
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2329;
+        private const ulong ShaderCodeGenVersion = 2330;
```
riperiperi, line 39:
```suggestion
        private const ulong ShaderCodeGenVersion = 2367;
```

Thog (approved):

FernandoS27:
You may want to put this behind an option. A BRX can just jump anywhere even outside a shader though. The truth is that either you HLE figure it out or you just put branches to everywhere in the shader or an exception: the GL driver gives no fucks but with VK, it's all about repeating a more conservative command list.

riperiperi:
That was definitely a lot of words. I'll make a best guess response.

We make assumptions based on the shader compiler that all commercial games are using, that only apply when the BRX instruction is present. This was already done for blocks starting with the "instruction" `0`, which technically would cover shader compilers that don't emit this loop, and has been reliable so far for indicating the end of a shader.

Not sure why you are talking about vulkan, but the games I've tested with _guest_ vulkan, official opengl and nouveau opengl are perfectly fine with this change. If you have one that isn't, feel free to share.

AcK77 (approved):

FernandoS27:
> That was definitely a lot of words. I'll make a best guess response.
> 
> We make assumptions based on the shader compiler that all commercial games are using, that only apply when the BRX instruction is present. This was already done for blocks starting with the "instruction" `0`, which technically would cover shader compilers that don't emit this loop, and has been reliable so far for indicating the end of a shader.
> 
> Not sure why you are talking about vulkan, but the games I've tested with _guest_ vulkan, official opengl and nouveau opengl are perfectly fine with this change. If you have one that isn't, feel free to share.

That's totally fine but as I told gdk on my younger days: you can HLE it. Since NVN compiler puts the BRX table next to it, this is fine to HLE. Nouvaue, doesn't use this doe but CUDA does, it's just that CUDA is mostly disabled for Switch Developers but not for homebrew. If you want to support to support CUDA, you need to figure it can jump anywhere but that's not possible in OHL, In Vulkab, you can `kinda` catch it.

riperiperi:
I'm not sure what the value in this comment is. Doing things the way we do it now works for nearly all of the switch library, and the case(s?) where it does not work can be covered by the alternate approach I indicated that we might try in the PR body. There is no downside to doing this, thus there is no need for a configurable option, and it's definitely not a "mistake". Saying "you could really suffer if you considered that shaders can jump to either side of gpu memory" is interesting in theory, but it does not exist in practice...

This PR is just adding an additional delimiter for our decoder anyways, there is no functional change to the way BRX is handled.

FernandoS27:
> I'm not sure what the value in this comment is. Doing things the way we do it now works for nearly all of the switch library, and the case(s?) where it does not work can be covered by the alternate approach I indicated that we might try in the PR body. There is no downside to doing this, thus there is no need for a configurable option, and it's definitely not a "mistake". Saying "you could really suffer if you considered that shaders can jump to either side of gpu memory" is interesting in theory, but it does not exist in practice...
> 
> This PR is just adding an additional delimiter for our decoder anyways, there is no functional change to the way BRX is handled.

are you emulating the switch or the the switch's games that's the question I left you.

ReinUsesLisp:
Kinda late, some shaders are delimited by `0xE2400FFFFF07000` aside from `0xE2400FFFFF87000F` (mind the 8->0 between F and 7).

riperiperi:
> are you emulating the switch or the the switch's games that's the question I left you.

Do you think this is some kind of gotcha? Both switch emulators are running, filled with assumptions and expectations that neither of them _would run performantly without_. If you're looking to emulate the hardware of the switch rather than the software, I'd suggest looking instead at emulating everything outside of the userland before making complaints that if you fuzzed the Tegra GPU with a shader that goes out of its way to misbehave, you might break graphics rendering that is going to be a mismatch for the real thing anyway unless you do everything in software like you're "supposed to".

> Kinda late, some shaders are delimited by `0xE2400FFFFF07000` aside from `0xE2400FFFFF87000F` (mind the 8->0 between F and 7).

Thanks for the tip.

PR #2371, opened 2021-06-16
hid/irs: Stub IsFirmwareUpdateAvailableForSixAxisSensor and CheckFirmwareVersion

LukeFZ:
This pull request stubs nn::hid::IHidServer::IsFirmwareUpdateAvailableForSixAxisSensor and nn::irsensor::IIrSensorServer::CheckFirmwareVersion, to make games which use these commands run without the "Ignore Missing Services" setting enabled.

This is my first pull request/contribution to Ryujinx, so feedback on what I did right/wrong would be much appreciated.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
@@ -22,6 +22,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid
         private bool _unintendedHomeButtonInputProtectionEnabled;
         private bool _vibrationPermitted;
         private bool _usbFullKeyControllerEnabled;
+        private bool _isFirmwareUpdateAvailableForSixAxisSensor;
 
         private HidNpadJoyAssignmentMode      _npadJoyAssignmentMode;
         private HidNpadHandheldActivationMode _npadHandheldActivationMode;
@@ -46,6 +47,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid
             _npadJoyAssignmentMode      = HidNpadJoyAssignmentMode.Dual;
             _npadHandheldActivationMode = HidNpadHandheldActivationMode.Dual;
             _gyroscopeZeroDriftMode     = HidGyroscopeZeroDriftMode.Standard;
+            _isFirmwareUpdateAvailableForSixAxisSensor = false;
 
             _sensorFusionParams  = new HidSensorFusionParameters();
             _accelerometerParams = new HidAccelerometerParameters();
@@ -575,17 +577,15 @@ namespace Ryujinx.HLE.HOS.Services.Hid
         }
 
         [CommandHipc(83)] // 6.0.0+
-        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId) -> bool UpdateAvailable
+        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId, pid) -> bool UpdateAvailable
         public ResultCode IsFirmwareUpdateAvailableForSixAxisSensor(ServiceCtx context)
         {
             int  sixAxisSensorHandle  = context.RequestData.ReadInt32();
             long appletResourceUserId = context.RequestData.ReadInt64();
 
-            bool updateAvailable = false;
+            context.ResponseData.Write(_isFirmwareUpdateAvailableForSixAxisSensor);
 
-            context.ResponseData.Write(updateAvailable);
-
-            Logger.Stub?.PrintStub(LogClass.ServiceHid, new { appletResourceUserId, sixAxisSensorHandle, updateAvailable });
+            Logger.Stub?.PrintStub(LogClass.ServiceHid, new { appletResourceUserId, sixAxisSensorHandle, _isFirmwareUpdateAvailableForSixAxisSensor });
 
             return ResultCode.Success;
         }
--- a/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
@@ -574,6 +574,22 @@ namespace Ryujinx.HLE.HOS.Services.Hid
             return ResultCode.Success;
         }
 
+        [CommandHipc(83)] // 6.0.0+
+        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId) -> bool UpdateAvailable
+        public ResultCode IsFirmwareUpdateAvailableForSixAxisSensor(ServiceCtx context)
+        {
+            int  sixAxisSensorHandle  = context.RequestData.ReadInt32();
+            long appletResourceUserId = context.RequestData.ReadInt64();
+
+            bool updateAvailable = false;
+
+            context.ResponseData.Write(updateAvailable);
+
+            Logger.Stub?.PrintStub(LogClass.ServiceHid, new { appletResourceUserId, sixAxisSensorHandle, updateAvailable });
+
+            return ResultCode.Success;
+        }
+
         [CommandHipc(91)]
         // ActivateGesture(nn::applet::AppletResourceUserId, int Unknown0)
         public ResultCode ActivateGesture(ServiceCtx context)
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -574,6 +574,22 @@ public ResultCode IsSixAxisSensorAtRest(ServiceCtx context)
...
         }
 
+        [CommandHipc(83)] // 6.0.0+
+        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId) -> bool UpdateAvailable
+        public ResultCode IsFirmwareUpdateAvailableForSixAxisSensor(ServiceCtx context)
+        {
+            int  sixAxisSensorHandle  = context.RequestData.ReadInt32();
+            long appletResourceUserId = context.RequestData.ReadInt64();
+
+            bool updateAvailable = false;
```
AcK77, line 584: Maybe you could use a global private var named `_isFirmwareUpdateAvailableForSixAxisSensor` here

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
```diff
@@ -75,6 +75,20 @@ public ResultCode GetNpadIrCameraHandle(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(314)] // 3.0.0+
+        // CheckFirmwareVersion(nn::irsensor::IrCameraHandle, nn::irsensor::PackedMcuVersion, nn::applet::AppletResourceUserId)
+        public ResultCode CheckFirmwareVersion(ServiceCtx context)
+        {
+            PlayerIndex irCameraHandle        = (PlayerIndex)context.RequestData.ReadInt32();
```
AcK77, line 82: This should be casted as an int

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
```diff
@@ -75,6 +75,20 @@ public ResultCode GetNpadIrCameraHandle(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(314)] // 3.0.0+
+        // CheckFirmwareVersion(nn::irsensor::IrCameraHandle, nn::irsensor::PackedMcuVersion, nn::applet::AppletResourceUserId)
+        public ResultCode CheckFirmwareVersion(ServiceCtx context)
```
AcK77, line 80: https://switchbrew.org/wiki/HID_services#CheckFirmwareVersion

This call take a PID as param, you should add it in the symbol and check if the order your read them is good, because usually there is a ulong placeholder to store it in the IPC call (try to look in other services to see how we did that)

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -574,6 +574,22 @@ public ResultCode IsSixAxisSensorAtRest(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(83)] // 6.0.0+
+        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId) -> bool UpdateAvailable
+        public ResultCode IsFirmwareUpdateAvailableForSixAxisSensor(ServiceCtx context)
```
AcK77, line 579: https://switchbrew.org/wiki/HID_services#IsFirmwareUpdateAvailableForSixAxisSensor

This call take a PID as param, you should add it in the symbol and check if the order your read them is good, because usually there is a ulong placeholder to store it in the IPC call (try to look in other services to see how we did that)

(Some other calls of this service are wrong too)

AcK77 (changes requested):

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
LukeFZ: Not quite sure what you mean here. Should I just remove the cast to PlayerIndex and just use the int itself?

AcK77: Yes, the value cast is an `int` not a `PlayerIndex`

LukeFZ: Okay, thanks!

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -46,6 +47,7 @@ public IHidServer(ServiceCtx context) : base(context.Device.System.HidServer)
             _npadJoyAssignmentMode      = HidNpadJoyAssignmentMode.Dual;
             _npadHandheldActivationMode = HidNpadHandheldActivationMode.Dual;
             _gyroscopeZeroDriftMode     = HidGyroscopeZeroDriftMode.Standard;
+            _isFirmwareUpdateAvailableForSixAxisSensor = false;
```
Thog, line 50: Degroup from previous assignation or align with them

```suggestion

            _isFirmwareUpdateAvailableForSixAxisSensor = false;
```

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
Thog: It is not necessarily present and looking at the layout I didn't found anything for it but yeah always something to be aware of ^^

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -574,6 +576,20 @@ public ResultCode IsSixAxisSensorAtRest(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(83)] // 6.0.0+
+        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::SixAxisSensorHandle, nn::hid::AppletResourceUserId, pid) -> bool UpdateAvailable
```
Thog, line 580: Inverted parameters in definition
```suggestion
        // IsFirmwareUpdateAvailableForSixAxisSensor(nn::hid::AppletResourceUserId, nn::hid::SixAxisSensorHandle, pid) -> bool UpdateAvailable
```

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -574,6 +576,20 @@ public ResultCode IsSixAxisSensorAtRest(ServiceCtx context)
...
+        public ResultCode IsFirmwareUpdateAvailableForSixAxisSensor(ServiceCtx context)
+        {
+            int  sixAxisSensorHandle  = context.RequestData.ReadInt32();
```
Thog, line 583: you need to skip 4 bytes because of padding
```suggestion
            int  sixAxisSensorHandle  = context.RequestData.ReadInt32();
            context.RequestData.BaseStream.Position += 4;
```

Ryujinx.HLE/HOS/Services/Hid/Irs/IIrSensorServer.cs
Thog: It is not necessarily present and looking at the layout I didn't found anything for it but yeah always something to be aware of ^^

Thog (changes requested):
Mostly lgtm, only need some tiny fixes 👍

Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
LukeFZ: Thanks for confirming my suspicion! I noticed in testing that padding was probably necessary to get accurate numbers, but since all other calls dont have this padding I decided against adding it. Should I apply this padding to the other commands which don't currently have it?

Thog: It really depends of the offset at which a field is placed and the next field size.

Basically this follow standard C alignment *most* of the time.

example 1 (padding):
```c
struct A {
    int  field_0x0;
                    // padding of 4 bytes right after field_0x0 to meet the alignment requirements for field_0x8. (sizeof(long) - sizeof(int) = 4)
    long field_0x8;
}
```

example 2 (no padding):
```c
struct A {
    int  field_0x0;
    int  field_0x4;
                    // no padding after field_0x4 because it meet the alignment requirements for field_0x8. (sizeof(long) - (sizeof(int) + sizeof(int)) = 0)
    long field_0x8;
}
```

LukeFZ: Alright, so just add the sizes of all previous fields together and check how many padding bytes to add to get to a multiple of 8 bytes, correct? 

Adding the same padding to IsSixAxisSensorAtRest and ResetSixAxisSensorFusionParameters (among others) fixes their applet ids as well, so I can confirm that this would fix other commands as well. Should I open up another PR for those fixes, or just add them to this one? (Personally I'd open up another PR, to not clutter up this one with unrelevant changes)

Thog: Another PR would be better 👍

LukeFZ: Great! I'll do that then.

AcK77 (approved):
Thanks for your contribution, lgtm!

Thog (approved):

PR #2384, opened 2021-06-21
account: Fix wrong condition in TrySelectUserWithoutInteraction

### LINKED ISSUE #2320: Shantae Half Genie Hero softlocks on title screen if multiple user profiles are present

The game softlocks on the title screen where it says "Press A" if the emulator has multiple user profiles. Having a non-default user profile selected it not required to trigger this, and profile will do, as long theres more than one. Deleting all user profiles other than the default one is enough to make the game work properly.

Make sure there is more than one user profile created. Then start the game and try to press A on the title screen. The game will softlock. Music still plays on the background, FPS is still "active" (which means that the GPU is still receiving commands), but progressing on the game is not possible.

### Environment?

- Ryujinx version: 1.0.6895
- Game version: 1.0
- Applied Mods: No

AcK77:
Since the implementation of User Profiles, we can get more than one profile stored. This PR fixes a wrong condition in `TrySelectUserWithoutInteraction`.

Closes #2320

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Account/Acc/ApplicationServiceServer.cs
+++ b/Ryujinx.HLE/HOS/Services/Account/Acc/ApplicationServiceServer.cs
@@ -62,7 +62,7 @@ namespace Ryujinx.HLE.HOS.Services.Account.Acc
 
             foreach (UserProfile userProfile in profiles)
             {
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
Cool, nice catch.

Thog (approved):

PR #2387, opened 2021-06-22
ipc: Remove size checks for buffer type 0x21/0x22

AcK77:
Since original IPC implementation doesn't check the buffers size, there is no reason to check them so I've removed it. Checking the buffers addresses could prevent to unexpected behaviors.

That fixes a bsd service issue with some homebrews and some games included Knockout City (https://github.com/Ryujinx/Ryujinx-Games-List/issues/3622) which is now bootable (but not playable since online is required):

image


gdkchan:
You can put the `if` on a single line now. Also the second address check is useless since it returns 0 anyway if the address is 0.

gdkchan (approved):

Thog (approved):

PR #2372, opened 2021-06-17
Initial support for GPU channels

gdkchan:
This adds some basic support for GPU channels. Each channel has *separate state*. Due to the fact that everything was using a single channel before, games that made use of multiple Queues (each Queue is a channel) would end with corrupt state. The most common situation that causes this is Queue A setting up its state, then a new queue (Queue B) being created, and then setting up its own state, and corrupting Queues A state in the process.

This fixes the problem by adding a new `GpuChannel` class, and moving almost all of the state that shouldn't be shared there. A few things such as draw state were not moved yet, but those shouldn't cause issues as this state is reset often. GPU address spaces are also still shared, but this is a separate issue, and queues can share an address space (address spaces are created per Device rather than per Queue).

Things that still need to be done:
- ~~Separate host state. OpenGL doesn't allow this, at least not easily. The backend could expose a `CreatePipeline` method, and allow the creation of separate pipelines. Supporting this on OpenGL would require keeping track of each pipeline state and setting it on switch as needed, or using multiple contexts. Vulkan allows the creation of multiple queues, where each queue has separate state, but queues have a limit that varies per vendor.~~ This has been solved by forcing the host state to be updated on every channel switch.
- Separate remaining state on frontend (mostly draw related state).
- Separate GPU address spaces (this is likely the most complicated part). Likely requires the memory tracking to work over physical memory instead, otherwise the the caches are dependant on the guest process address space, which means that those resources can't be shared between processes.

This fixes a number of games that previously had `InvalidMemoryRegionException`s caused by addresses from already closed channels being used, due to all queues sharing the same channel.

Fixes https://github.com/Ryujinx/Ryujinx-Games-List/issues/2901
Now goes ingame.
image
image
image
(Note: Still has an issue when a save is loaded where the scene is mostly black. Needs further investigation.)

Fixes https://github.com/Ryujinx/Ryujinx-Games-List/issues/1965
Now goes ingame.
image
image
image
image

Fixes https://github.com/Ryujinx/Ryujinx-Games-List/issues/3415
Now goes ingame.
image
image
image
image
image
image

Likely also fixes (needs re-testing, I don't have those games):
https://github.com/Ryujinx/Ryujinx-Games-List/issues/82
https://github.com/Ryujinx/Ryujinx-Games-List/issues/745
https://github.com/Ryujinx/Ryujinx-Games-List/issues/924
https://github.com/Ryujinx/Ryujinx-Games-List/issues/3077
Might also fix Nekopara vol. 1/2/3 save issue?

PR #2394, opened 2021-06-23
Initial support for separate GPU address spaces

### LINKED ISSUE #1822: Separate GPU address spaces on Nvservices
#### Overview:
Nvservices supports creating separate GPU address spaces. This is done through `NvHostAsGpuDeviceFile`. One can create a address space, and then use the `BindChannel` ioctl to bind it to a channel. The channel will then use that address space when submiting commands. For non-GPU devices (such as NVDEC and VIC), I believe that it will create an address space as part of the channel creation. Also, unlike the GPU that has a 40-bit address space, those devices only have a 32-bit address space.

Right now there is a single GPU address space that is shared by everything. They should be separated, and one `MemoryManager`/`NvMemoryAllocator` should be created per `NvHostAsGpuDeviceFile` or `NvHostChannelDeviceFile`. This will require some changes to the GPU and NVDEC/VIC projects, like for example, making the `Submit` methods used for command submission also take a `MemoryManager` instance that will be used internally.

For anyone interested, I have started some work on this, but I am not working on it right now (and don't plan to anytime soon), I can make the code available somewhere.

#### Nature of Request:  
Fix/refactoring.

This will match hardware behavior and is more accurate.

gdkchan:
Continuation of #2372 (and depends on it, of course).  This adds support for separate GPU address spaces, so channels may also have a different address space bound, while it was sharing the same address space between all channels and processes before.

This is not exactly how I envisioned it. I would prefer to do all the mapping over physical memory rather than process CPU virtual memory, however this is not really simple due to the way how memory read/write tracking works right now. The current solution keeps the `PhysicalMemory` class that is really wrapping a process virtual memory manager, and for this reason, resource sharing between processes is not possible. New methods have been introduced on the GPU to allow explicit registration and unregistration of processes, and such physical memory wrappers are keyed by PID on the GPU.

Shader cache loading should behave the same, it will load the caches for the processes currently loaded when the function is called. If more processes are registered after that, they won't have their shader caches loaded. I believe it can be easily changed to always load it by moving the call to the `RegisterProcess` function, however I haven't really looked at the shader cache load code, so I'm not sure if it is prepared to handle that. I also believe the UI is not prepared to show load progress of the cache after a game has been launched.

The NVDEC/VIC instances are now per-process rather than shared. This was done because it needs memory tracking to work, which in turn needs the process memory manager instace, so that means it needs to be created per-process. It can be made shared again (imo the correct way to do it) if the memory tracking is modified to operate over physical memory instead. This means that multi-process video decoding should work (however I haven't tested that). Simultaneous H264 video decoding on the same process is still broken.

Summary of the changes on the GPU project:
- Shader/Buffer/Texture caches have been moved to physical memory (as they are cached by CPU VA).
- Counter cache has been moved to the memory manager (as they are cached by GPU VA).
- Physical memory has been moved to the memory manager.
- Memory manager has been moved from the GPU context to the GPU channel.
- A new `BindMemory` method has been added to the GPU channel to allow changing the current memory manager.
- Physical memory is now reference counted as otherwise `UnregisterProcess` would dispose the CPU memory manager too early (while it was still in use by a GPU channel).
- Most other changes are just boring and trivial changes to accomodate for all those members being moved.

Testing is always welcome to ensure that nothing broke on the refactoring.
This is not expected to fix any game, but is a step towards multi-process support.

Closes #1822.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Compute.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute.cs
@@ -16,7 +16,9 @@ namespace Ryujinx.Graphics.Gpu.Engine
         /// <param name="argument">Method call argument</param>
         public void Dispatch(GpuState state, int argument)
         {
-            FlushUboDirty(state.Channel.MemoryManager);
+            var memoryManager = state.Channel.MemoryManager;
+
+            FlushUboDirty(memoryManager);
 
             uint qmdAddress = (uint)state.Get<int>(MethodOffset.DispatchParamsAddress);
 
@@ -43,7 +45,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
                 state.Channel.BufferManager.SetComputeUniformBuffer(index, gpuVa, size);
             }
 
-            ShaderBundle cs = ShaderCache.GetComputeShader(
+            ShaderBundle cs = memoryManager.Physical.ShaderCache.GetComputeShader(
                 state,
                 shaderGpuVa,
                 qmd.CtaThreadDimension0,
@@ -82,7 +84,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
                 cbDescAddress += (ulong)cbDescOffset;
 
-                SbDescriptor cbDescriptor = _context.PhysicalMemory.Read<SbDescriptor>(cbDescAddress);
+                SbDescriptor cbDescriptor = state.Channel.MemoryManager.Physical.Read<SbDescriptor>(cbDescAddress);
 
                 state.Channel.BufferManager.SetComputeUniformBuffer(cb.Slot, cbDescriptor.PackAddress(), (uint)cbDescriptor.Size);
             }
@@ -97,7 +99,7 @@ namespace Ryujinx.Graphics.Gpu.Engine
 
                 sbDescAddress += (ulong)sbDescOffset;
 
-                SbDescriptor sbDescriptor = _context.PhysicalMemory.Read<SbDescriptor>(sbDescAddress);
+                SbDescriptor sbDescriptor = state.Channel.MemoryManager.Physical.Read<SbDescriptor>(sbDescAddress);
 
                 state.Channel.BufferManager.SetComputeStorageBuffer(sb.Slot, sbDescriptor.PackAddress(), (uint)sbDescriptor.Size, sb.Flags);
             }
--- a/Ryujinx.Graphics.Gpu/Engine/Compute.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute.cs
@@ -16,11 +16,11 @@ namespace Ryujinx.Graphics.Gpu.Engine
         /// <param name="argument">Method call argument</param>
         public void Dispatch(GpuState state, int argument)
         {
-            FlushUboDirty();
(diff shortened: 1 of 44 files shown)


ger96rh:
image

I tried the story mode with this pr in SSBU and now it can be played before it gave black screens or graphic errors, great job

GamerzHell9137:
@ger96rh you could always enter adventure mode but you can't get into fights in adventure mode.

ger96rh:
To your surprise I can enter the fights without any problem by trying this pr

GamerzHell9137:
@ger96rh you might have ignore missing services on. Says 

`00:04:26.166 |E| HLE.OsThread.16 Application : Unhandled exception caught: Ryujinx.HLE.Exceptions.ServiceNotImplementedException: Ryujinx.HLE.HOS.Services.Caps.IAlbumApplicationService: 142`
 for me

Thog:
> @ger96rh you might have ignore missing services on. Says
> 
> `00:04:26.166 |E| HLE.OsThread.16 Application : Unhandled exception caught: Ryujinx.HLE.Exceptions.ServiceNotImplementedException: Ryujinx.HLE.HOS.Services.Caps.IAlbumApplicationService: 142`
> for me

It would be nice to get full logs + version so we could implement the missing services call :+1:

EliasM108:
> > @ger96rh you might have ignore missing services on. Says
> > `00:04:26.166 |E| HLE.OsThread.16 Application : Unhandled exception caught: Ryujinx.HLE.Exceptions.ServiceNotImplementedException: Ryujinx.HLE.HOS.Services.Caps.IAlbumApplicationService: 142`
> > for me
> 
> It would be nice to get full logs + version so we could implement the missing services call 👍

You need to have Ignore missing services checked in order to enter any fight in adventure mode, else it crashes the emulator.
Log:
[Ryujinx_1.0.0+048e776_2021-06-24_13-58-24.log](https://github.com/Ryujinx/Ryujinx/files/6708634/Ryujinx_1.0.0%2B048e776_2021-06-24_13-58-24.log)

meisiek:
> To your surprise I can enter the fights without any problem by trying this pr

I can confirm that after some tests. With version 1.0.0, no DLC and "ignore missing services" adventure mode fights are working. Without "ignore missing services" it still crashes when entering a fight.

riperiperi (approved):
Looking good from here, hopefully leading to some great things here.

I've noticed a lot of places where 4 references have to be followed to reach physical memory resources for the channel, sometimes multiple times in the same function. I don't think these get optimized out, so it might be a good idea to pull things out into local variables when possible, or even store/pass the associated resource directly instead of the _state.

tazevam:
It may be related to InvalidMemoryRegionException, here is also feedback.
I tried the latest version, but DQ11S still crashed after setting up the save. It may be due to the different game version. Mine is the Japanese version (1006C300E9F0000).
In addition, the test process of the new version is very smooth, the loading speed is much faster, RYUJINX is better to use than in the past, thank you for your hard work. 
[Ryujinx_1.0.0+048e776_2021-06-25_01-25-57.log](https://github.com/Ryujinx/Ryujinx/files/6713604/Ryujinx_1.0.0%2B048e776_2021-06-25_01-25-57.log)

Thog (approved):

PR #2404, opened 2021-06-24
Fix texture sampling with depth compare and LOD level or bias

gdkchan:
Some overloads of the texture sample functions (in special, those that take a 2D or Cube shadow sampler, with a LOD level or LOD bias parameter) requires an extension, namely the [texture shadow LOD](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_shadow_lod.txt) extension. This PR enables the extension on shaders, fixing a compilation failure that would otherwise occur on those shaders. Additionally, since the extension is not widely supported, this removes the LOD parameters if the GPU doesn't support the extension. Note that it was already removing the parameter in one case (any array shadow sampler that had explicit LOD levels). Now this case is handled properly if the extension is supported, which is more accurate.

Fixes lighting on Mario Golf Super Rush, and possibly other games.
Before:
image
image
After:
image
image

### DIFF
--- a/Ryujinx.Graphics.GAL/Capabilities.cs
+++ b/Ryujinx.Graphics.GAL/Capabilities.cs
@@ -4,8 +4,9 @@ namespace Ryujinx.Graphics.GAL
     {
         public bool SupportsAstcCompression          { get; }
         public bool SupportsImageLoadFormatted       { get; }
-        public bool SupportsNonConstantTextureOffset { get; }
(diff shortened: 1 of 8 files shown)


AcK77 (approved):

Thog (approved):

riperiperi (approved):
LGTM here, nice catch.

PR #2406, opened 2021-06-25
Add support for custom line widths

gdkchan:
Adds support for custom line widths on the GPU.
Fixes incorrect grid line thickness on Mario Golf Super Rush.
Before:
image
image
After:
image
image


Thog (approved):
nice catch!

PR #2286, opened 2021-05-14
POWER - Performance Optimizations With Extensive Ramifications

riperiperi:
This PR adds an additional type of memory manager, the HostMappedMemoryManager, which maps directly into a reserved "address space" range in the host memory, rather than into an emulated physical memory and accessing it with a software page table. Kernel memory management has been refactored to support this.

This PR is a joint effort - gdkchan did all the kernel refactoring, memory instruction changes and new memory manager, and I made changes necessary to make the memory tracking function performantly with host memory protection, and the Windows 10 emulated shared memory using placeholders.

More information on these changes and their implications will be posted in a comment later. Check the commit history for more detailed attribution.

For the record, I didn't come up with the name. 👼 

# Technical overview
## Current approach
One area of Ryujinx CPU emulation that is in dire need of improvements right now is memory access. Currently, guest memory access is implemented using inline software translation. This means that on each memory access, we need to do address validation, address translation and software protection for memory tracking. This approach has some benefits, but it also has some issues.

### Main benefits:

- Enhanced control over the memory mapping and address translation process. Allows things like guest shared memory to be easily implemented on all OSes.
- Memory reprotection is fast. This benefits GPU emulation as we need to track CPU writes to memory regions that the GPU can access, to invalidate data on GPU memory.

### Notable downsides:

- Poor performance, as it needs to produce roughly 20 x86 instructions per single Arm memory access instruction.
- Inflated code size, as all those instructions end up in memory (and also on disk if PPTC is enabled). This contributes to "JIT Cache Exhausted" crashes.
- Slow boot times because we throw more code into the pipeline, the JIT takes more time which means longer loads and more stutters without PPTC (first run).
- Slow PPTC build times (with PPTC), for the same reason mentioned above.

## How can we improve this?
We can take advantage of the host CPU MMU to make the address translation for us, instead of implementing it in software. This can be done by replicating the guest memory mappings on the host. Basically, we reserve a region of memory in the emulator process that can fit the entire guest address space, at a memory location "BASE", and then for each map that the guest tries to do at an address "A", we can just map "BASE + A" on the host. Similarly, on the CPU emulator, for each access to access "A", we just need to access "BASE + A".

The changes accomplish this by introducing a new memory manager called "HostMappedMemoryManager", which functions as mentioned above. The old "MemoryManager" is still there as there are some cases where it can be useful, such as situations that require memory aliasing to work (like services that need to access memory shared with other processes, if we ever decide to run services on the emulator).

[...]

PR #2407, opened 2021-06-25
Flush UBO updates more frequently

gdkchan:
This fixes a regression introduced on #2286. This PR made buffer updates are batched, but sometimes the updates are delayed way too much, and when it gets to actually write the data, the buffer was already unmapped, leading to a NRE as the address is invalid and the buffer no longer exists. This attempts to solve this by making those updates happen more frequently (at the end of every command buffer).

Note: This has the potential to reduce performance, so it has to be tested to make sure that performance was not regressed. If it was, then it will need to be fixed using a different method.

Affected games:
https://github.com/Ryujinx/Ryujinx-Games-List/issues/876 (probably?)
https://github.com/Ryujinx/Ryujinx-Games-List/issues/1096
Maybe more...

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/GPFifo/GPFifoProcessor.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/GPFifo/GPFifoProcessor.cs
@@ -126,6 +126,8 @@ namespace Ryujinx.Graphics.Gpu.Engine.GPFifo
                     }
                 }
             }
+
+            _context.Methods.FlushUboDirty(MemoryManager);
         }
 
         /// <summary>


riperiperi:
Maybe it would work doing this in the GPU semaphore release methods? I imagine that the guest won't unmap the buffer til it's sure the commands are done, and that would be one way to know.

gdkchan:
Switched to a new approach that flushes UBO updates only on unmap, this approach is not going to regress performance. Verified working on Wonderboy. Ready for review.

zombie167:
Now the game can be entered.Thank you very much!

sshot-3
sshot-4

BigheadSMZ:
Confirming that this fixes booting "Wonder Boy: The Dragon's Trap".

Thog (approved):

riperiperi (approved):
I can't think of a better way to do this right now, so LGTM.

PR #2410, opened 2021-06-25
Implement 12.0.0 hwopus functions

bylaws:
This is based off of my RE of 12.0.2 audio services, the newly added parameter can be safely ignored due to ryujinx not using fixed-size I/O buffers. I also included a minor fixup to avoid returning incorrect error codes to callers as HOS always aborts (aside from one special case however nothing actually returns that 😖)

This is completely untested as I don't own any games using the new 12.0.0+ SDK so it should probably be tested on one of them before merging.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
@@ -1,4 +1,6 @@
+using Ryujinx.Common;
 using Ryujinx.HLE.HOS.Services.Audio.HardwareOpusDecoderManager;
+using Ryujinx.HLE.HOS.Services.Audio.Types;
 
 namespace Ryujinx.HLE.HOS.Services.Audio
 {
@@ -26,11 +28,38 @@ namespace Ryujinx.HLE.HOS.Services.Audio
         // GetWorkBufferSize(bytes<8, 4>) -> u32
         public ResultCode GetWorkBufferSize(ServiceCtx context)
         {
-            // Note: The sample rate is ignored because it is fixed to 48KHz.
+            // NOTE: The sample rate is ignored because it is fixed to 48KHz.
             int sampleRate    = context.RequestData.ReadInt32();
             int channelsCount = context.RequestData.ReadInt32();
 
             context.ResponseData.Write(GetOpusDecoderSize(channelsCount));
+
+            return ResultCode.Success;
+        }
+
+        [CommandHipc(4)] // 12.0.0+
+        // InitializeEx(OpusParametersEx, u32, handle<copy>) -> object<nn::codec::detail::IHardwareOpusDecoder>
+        public ResultCode InitializeEx(ServiceCtx context)
+        {
+            OpusParametersEx parameters = context.RequestData.ReadStruct<OpusParametersEx>();
+
+            // UseLargeFrameSize can be ignored due to not relying on fixed size buffers for storing the decoded result.
+            MakeObject(context, new IHardwareOpusDecoder(parameters.SampleRate, parameters.ChannelCount));
+
+            // Close transfer memory immediately as we don't use it.
+            context.Device.System.KernelContext.Syscall.CloseHandle(context.Request.HandleDesc.ToCopy[0]);
+
+            return ResultCode.Success;
+        }
+
+        [CommandHipc(5)] // 12.0.0+
+        // GetWorkBufferSizeEx(OpusParametersEx) -> u32
+        public ResultCode GetWorkBufferSizeEx(ServiceCtx context)
+        {
+            OpusParametersEx parameters = context.RequestData.ReadStruct<OpusParametersEx>();
+
+            // NOTE: The sample rate is ignored because it is fixed to 48KHz.
+            context.ResponseData.Write(GetOpusDecoderSize(parameters.ChannelCount));
 
             return ResultCode.Success;
         }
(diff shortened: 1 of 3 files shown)


Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
```diff
@@ -35,6 +37,33 @@ public ResultCode GetWorkBufferSize(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(4)]
+        // InitializeEx(OpusParametersEx, u32, handle<copy>) -> object<nn::codec::detail::IHardwareOpusDecoder>
+        public ResultCode InitializeEx(ServiceCtx context)
+        {
+            OpusParametersEx parameters = context.RequestData.ReadStruct<OpusParametersEx>();
+
+            // UseLargeFrameSize can be ignored due to not relying on fixed size buffers for storing the decoded result
```
AcK77, line 46:
```suggestion
            // UseLargeFrameSize can be ignored due to not relying on fixed size buffers for storing the decoded result.
```

Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
```diff
@@ -35,6 +37,33 @@ public ResultCode GetWorkBufferSize(ServiceCtx context)
...
+            return ResultCode.Success;
+        }
+
+        [CommandHipc(5)]
+        // GetWorkBufferSizeEx(OpusParametersEx) -> u32
+        public ResultCode GetWorkBufferSizeEx(ServiceCtx context)
+        {
+            OpusParametersEx parameters = context.RequestData.ReadStruct<OpusParametersEx>();
+
+            // Note: The sample rate is ignored because it is fixed to 48KHz.
```
AcK77, line 61:
```suggestion
            // NOTE: The sample rate is ignored because it is fixed to 48KHz.
```

Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
```diff
@@ -35,6 +37,33 @@ public ResultCode GetWorkBufferSize(ServiceCtx context)
             return ResultCode.Success;
         }
 
+        [CommandHipc(4)]
```
AcK77, line 40:
```suggestion
        [CommandHipc(4)] // 12.0.0+
```

Ryujinx.HLE/HOS/Services/Audio/IHardwareOpusDecoderManager.cs
```diff
@@ -35,6 +37,33 @@ public ResultCode GetWorkBufferSize(ServiceCtx context)
...
+            // UseLargeFrameSize can be ignored due to not relying on fixed size buffers for storing the decoded result
+            MakeObject(context, new IHardwareOpusDecoder(parameters.SampleRate, parameters.ChannelCount));
+
+            // Close transfer memory immediately as we don't use it.
+            context.Device.System.KernelContext.Syscall.CloseHandle(context.Request.HandleDesc.ToCopy[0]);
+
+            return ResultCode.Success;
+        }
+
+        [CommandHipc(5)]
```
AcK77, line 55:
```suggestion
        [CommandHipc(5)] // 12.0.0+
```

Ryujinx.HLE/HOS/Services/Audio/Types/OpusParametersEx.cs
```diff
@@ -0,0 +1,17 @@
...
+    struct OpusParametersEx
+    {
+        public int SampleRate;
+        public int ChannelCount;
+
+        [MarshalAs(UnmanagedType.U4)]
+        public bool UseLargeFrameSize;
+
+        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 4)]
+        public char[] Padding1;
```
AcK77, line 15: You can use a `int` here

Ryujinx.HLE/HOS/Services/Audio/HardwareOpusDecoderManager/IHardwareOpusDecoder.cs
```diff
@@ -128,6 +128,10 @@ public ResultCode DecodeInterleavedOriginal(ServiceCtx context)
                     context.ResponseData.Write(outConsumed);
                     context.ResponseData.Write(outSamples);
                 }
+                else
+                {
+                    throw new InvalidOperationException("Opus DecodeInterleavedOriginal failed!");
```
AcK77, line 133: I don't think it's the right thing to do here, since there is a returned `ResultCode` if something goes wrong in `DecodeInterleavedOriginal` and there is a `OpusException` if the decoding fail too.
(Same comment for the three other exceptions you added above)

bylaws: This is what hos does, Decode returns a result code however it's ignored by the service and it will always abort instead in case of error

bylaws: If you want I can drop all things that return the result code and abort directly instead

Thog (approved):
other than @AcK77 comments lgtm.

I don't think the exception would be an issue imo considering games will just abort if an error is returned tbh.

Ryujinx.HLE/HOS/Services/Audio/Types/OpusParametersEx.cs
Thog: or better ``Array4<byte>``

bylaws:
feedback addressed

AcK77 (approved):

PR #2411, opened 2021-06-25
Implement "hide UI" option

ooa113y:
This implements a menu option to hide the emulator UI (status/menu bar) while the game is running. Additionally, this implements focusing the menu bar on Alt keypress, which will also show the UI if it's hidden. As a neat side effect, you can now navigate the menu bar with the keyboard, but the main purpose of this is to allow UI visibility to be controlled independently of fullscreen.


EmulationFanatic:
Doesn't F4 toggle between handheld/docked mode?

ooa113y:
> Doesn't F4 toggle between handheld/docked mode?

That's F9.

Leischii:
Would it be possible to add scnresshots of this change to the post? Since it's an interface change.

ooa113y:
Not sure what to screenshot, it's just gonna look like a window without the widgets... But sure I guess?

image

jduncanator:
This ones a bit of an interesting one. I'm always wary of introducing functionality like this that isn't obvious. For example, imagine if you accidentally hit "F4". Because it hides all the menus, there is no easy way of undoing what you've done unless you happen to know what key you hit, or you happen to know that the "F4" hotkey is what you need to press to bring the menus back.

At least with things like Docked mode or Vsync, even if you don't know what hotkey you hit, you can still set these options back via the menu and/or status bar.

It's probably not a deal breaker, but something to think about.

Shahil-Ayato:
> This ones a bit of an interesting one. I'm always wary of introducing functionality like this because the interaction isn't immediately obvious. For example, imagine if you accidentally hit "F4". Because it hides all the menus, there is no easy way of undoing what you've done unless you happen to know what key you hit, or you happen to know that the "F4" hotkey is what you need to press to bring the menus back.
> 
> At least with things like Docked mode or Vsync, even if you don't know what hotkey you hit, you can still set these options back via the menu and/or status bar.
> 
> It's probably not a deal breaker, but something to think about.

I guess a better idea would be to have this as a setting in the GUI and have it off by default? That way most can avoid mishaps like this

jduncanator:
> > This ones a bit of an interesting one. I'm always wary of introducing functionality like this because the interaction isn't immediately obvious. For example, imagine if you accidentally hit "F4". Because it hides all the menus, there is no easy way of undoing what you've done unless you happen to know what key you hit, or you happen to know that the "F4" hotkey is what you need to press to bring the menus back.
> > At least with things like Docked mode or Vsync, even if you don't know what hotkey you hit, you can still set these options back via the menu and/or status bar.
> > It's probably not a deal breaker, but something to think about.
> 
> I guess a better idea would be to have this as a setting in the GUI and have it off by default? That way most can avoid mishaps like this

Perhaps the hotkey could be unbound by default?

There's also the possibility of adding an additional menu option to the window title bar that would let you enter and exit the mode, although this'll be platform specific, so it'd involve Win32 calls and I'm not sure what the equivalent is on Linux. macOS doesn't have this problem as the menu bar is always visible.

image

ooa113y:
I understand the concerns here but I'm not sure what an adequate solution would be.

I'll point out that this setting at least isn't persistent (wouldn't make sense for it to be as you *need* the menus outside of the actual emulation context), so worst case scenario in the current situation you just close Ryujinx and open it again.

Having the hotkey be unbound would need an implementation of an entire hotkey configuration menu in the GUI because otherwise the feature would be too involved to enable, and I'm not sure I can tackle something that major.

Having this be a menu option in the GUI is kind of contradictory since, well, once you click that option the GUI would disappear and you wouldn't be able to bring it back up, unless I'm misunderstanding the suggestion.

I've tried adding a MessageDialog pop-up to confirm the action when the key is pressed, but that just froze Ryujinx and I'm not overly sure why, so that was probably a bad idea.

alololan:
How about making pressing 'Esc' reset the GUI to default.
That's what normally people would press usually. Not sure if its doable cause I dont know much about programming.

ooa113y:
`Esc` is already bound to be an exit button.

GamerzHell9137:
Why not add an option to enable/disable the status bar? 
image

ooa113y:
Just removing the statusbar feels half-baked to me. You get rid of one thing that takes up space but you still have another one.

jduncanator:
> Just removing the statusbar feels half-baked to me. You get rid of one thing that takes up space but you still have another one.

How about this then. Most applications will bring up the top menu bar if it's hidden when you press the "Alt" key. If you add an entry into the "Options" menu, and bring the menu bar up when you press the "Alt" key, then you can always get out of "zen mode" without knowing the hotkey. One could argue you trade knowing one hotkey for another, however I feel "Alt" is far more common knowledge than a custom hotkey for Ryujinx.

You could extend this further to display the menu bar when pressing "Alt" when in full screen mode, which has been another request I've seen asked about a couple of times.

Two birds, one stone?

ooa113y:
That sounds good to me. I'll see if I can figure this out sometime soon.

riperiperi (approved):
LGTM, cool to see. Not sure about the practicality but other people might be able to weigh on on that.

emmauss (approved):

Thog (approved):

PR #2422, opened 2021-06-29
Separate GPU engines and make state follow official docs (part 1/2)

gdkchan:
This change separate the GPU engines (except 3D, which will be done separately since it is way more complex), and uses structs auto-generated from the official NVIDIA headers. The goal is making the code more organized, and the implementation closer to the real GPU.

Highlight of the changes:
- `GpuState` replaced by `DeviceState<T>`, where `T` is a per-engine state struct (auto-generated from official headers). This was done to all engines except 3D.
- Register write trigger methods are no longer shared between all engines. For example, writing to the register that triggers compute dispatch from the 3D or 2D engine no longer works, or trying to draw/blit from compute. Before that was allowed (which is completely incorrect).
- Engine mapping to sub-channels is now fixed as per the documentation (Compute on sub-channel 1, I2M on 2, 2D on 3, DMA on 4 etc).
- More state is now per-channel rather than shared. In special, the internal state kept by I2M for copies is now per-channel, however I don't expect that to actually make any visible difference, as normally on channel switch, the copy would be already over, and the state would then be reset anyway on the next transfer.
- `DeviceState` now takes a optional log callback on the constructor, which can be used to log all register writes, useful for debugging. Nothing uses this right now, but one can easily pass a callback method and use the functionality for debugging purposes.

I'm not aware of any game fixed by this, but it could potentially fix some issues caused by GPU operations being incorrectly triggered due to trigger methods being shared between sub-channels.

Testing is always welcome to ensure the refactoring didn't cause regressions.


Ryujinx.Graphics.Gpu/Image/TextureCache.cs
```diff
@@ -722,22 +751,32 @@ private void ChangeSizeIfNeeded(TextureInfo info, Texture texture, bool isSample
...
         {
-            ulong address = _context.MemoryManager.Translate(cbp.DstAddress.Pack());
+            ulong address = memoryManager.Translate(gpuVa);
 
             if (address == MemoryManager.PteUnmapped)
             {
                 return null;
             }
 
-            int bpp = swizzle.UnpackDstComponentsCount() * swizzle.UnpackComponentSize();
```
gdkchan, line 739: I believe this was incorrect before, as the caller only calculates the BPP this way when `Remap` is enabled. Fixed by passing the BPP directly instead of calculating it again here.

Thog (approved):

riperiperi (approved):
LGTM, lots here but all looks good.

PR #2428, opened 2021-06-29
Implement a "Pause Emulation" option & hotkey

mpnico:
Add a new menu item to pause/resume emulation :
image

Add also a new hotkey to toggle pause (default is F5, can be changed in Config.json file)

~~Implementation is not perfect as Audio continues to play a bit after pause.~~

Closes Ryujinx#1604

### DIFF
--- a/ARMeilleure/State/ExecutionContext.cs
+++ b/ARMeilleure/State/ExecutionContext.cs
@@ -145,12 +145,12 @@ namespace ARMeilleure.State
             _nativeContext.SetCounter(0);
         }
 
-        public void Suspend()
+        public static void SuspendCounter()
         {
             _tickCounter.Stop();
         }
 
-        public void Resume()
+        public static void ResumeCounter()
         {
             _tickCounter.Start();
         }
--- a/ARMeilleure/State/ExecutionContext.cs
+++ b/ARMeilleure/State/ExecutionContext.cs
@@ -145,6 +145,16 @@ namespace ARMeilleure.State
             _nativeContext.SetCounter(0);
         }
 
+        public void Suspend()
+        {
+            _tickCounter.Stop();
+        }
+
+        public void Resume()
+        {
+            _tickCounter.Start();
+        }
+
         public void Dispose()
         {
             _nativeContext.Dispose();
--- a/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
+++ b/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
@@ -15,6 +15,7 @@ namespace Ryujinx.Audio.Backends.OpenAL
         private readonly ALDevice _device;
         private readonly ALContext _context;
         private readonly ManualResetEvent _updateRequiredEvent;
+        private readonly ManualResetEvent _pauseEvent;
         private readonly ConcurrentDictionary<OpenALHardwareDeviceSession, byte> _sessions;
         private bool _stillRunning;
         private Thread _updaterThread;
@@ -24,6 +25,7 @@ namespace Ryujinx.Audio.Backends.OpenAL
             _device = ALC.OpenDevice("");
             _context = ALC.CreateContext(_device, new ALContextAttributes());
             _updateRequiredEvent = new ManualResetEvent(false);
+            _pauseEvent = new ManualResetEvent(true);
             _sessions = new ConcurrentDictionary<OpenALHardwareDeviceSession, byte>();
 
             _stillRunning = true;
@@ -86,6 +88,11 @@ namespace Ryujinx.Audio.Backends.OpenAL
         public ManualResetEvent GetUpdateRequiredEvent()
         {
             return _updateRequiredEvent;
+        }
+
+        public ManualResetEvent GetPauseEvent()
+        {
+            return _pauseEvent;
         }
 
         private void Update()
--- a/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
+++ b/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
@@ -139,6 +139,8 @@ namespace Ryujinx.Audio.Backends.OpenAL
 
                 ALC.DestroyContext(_context);
                 ALC.CloseDevice(_device);
+
+                _pauseEvent.Dispose();
             }
         }
 
--- a/Ryujinx.Audio/AudioManager.cs
+++ b/Ryujinx.Audio/AudioManager.cs
@@ -15,7 +15,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 //
 
-using Ryujinx.Audio.Integration;
 using System;
 using System.Threading;
 
@@ -45,11 +44,6 @@ namespace Ryujinx.Audio
         /// The worker thread in charge of handling sessions update.
         /// </summary>
         private Thread _workerThread;
-
-        /// <summary>
-        /// The audio device driver.
-        /// </summary>
-        private IHardwareDeviceDriver _deviceDriver;
 
         /// <summary>
         /// Create a new <see cref="AudioManager"/>.
@@ -84,15 +78,14 @@ namespace Ryujinx.Audio
         /// <summary>
         /// Initialize update handlers.
         /// </summary>
-        /// <param name="deviceDriver ">The device driver to get the driver event when an audio buffer finished playing/being captured and the pause event</param>
+        /// <param name="updatedRequiredEvent ">The driver event that will get signaled by the device driver when an audio buffer finished playing/being captured</param>
         /// <param name="outputCallback">The callback to call when an audio buffer finished playing</param>
         /// <param name="inputCallback">The callback to call when an audio buffer was captured</param>
-        public void Initialize(IHardwareDeviceDriver deviceDriver, Action outputCallback, Action inputCallback)
+        public void Initialize(ManualResetEvent updatedRequiredEvent, Action outputCallback, Action inputCallback)
         {
             lock (_lock)
             {
-                _deviceDriver = deviceDriver;
-                _updateRequiredEvents[0] = deviceDriver.GetUpdateRequiredEvent();
+                _updateRequiredEvents[0] = updatedRequiredEvent;
                 _actions[0] = outputCallback;
                 _actions[1] = inputCallback;
             }
--- a/Ryujinx.Audio/AudioManager.cs
+++ b/Ryujinx.Audio/AudioManager.cs
@@ -123,7 +123,7 @@ namespace Ryujinx.Audio
         }
 
         /// <summary>
-        /// Stop updating the <see cref="AudioManager"/> without stopping worker thread.
+        /// Stop updating the <see cref="AudioManager"/> without stopping the worker thread.
         /// </summary>
         public void StopUpdates()
         {
--- a/Ryujinx.Audio/AudioManager.cs
+++ b/Ryujinx.Audio/AudioManager.cs
@@ -84,7 +84,7 @@ namespace Ryujinx.Audio
         /// <summary>
         /// Initialize update handlers.
         /// </summary>
-        /// <param name="updatedRequiredEvent ">The driver event that will get signaled by the device driver when an audio buffer finished playing/being captured</param>
+        /// <param name="deviceDriver ">The device driver to get the driver event when an audio buffer finished playing/being captured and the pause event</param>
         /// <param name="outputCallback">The callback to call when an audio buffer finished playing</param>
         /// <param name="inputCallback">The callback to call when an audio buffer was captured</param>
         public void Initialize(IHardwareDeviceDriver deviceDriver, Action outputCallback, Action inputCallback)
@@ -105,8 +105,6 @@ namespace Ryujinx.Audio
         {
             while (true)
             {
-                //_deviceDriver.GetPauseEvent().WaitOne();
-
                 int index = WaitHandle.WaitAny(_updateRequiredEvents);
 
                 // Last index is here to indicate thread termination.
--- a/Ryujinx.Audio/AudioManager.cs
+++ b/Ryujinx.Audio/AudioManager.cs
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 //
 
+using Ryujinx.Audio.Integration;
 using System;
 using System.Threading;
 
@@ -44,6 +45,11 @@ namespace Ryujinx.Audio
         /// The worker thread in charge of handling sessions update.
         /// </summary>
         private Thread _workerThread;
+
+        /// <summary>
+        /// The audio device driver.
+        /// </summary>
+        private IHardwareDeviceDriver _deviceDriver;
 
         /// <summary>
         /// Create a new <see cref="AudioManager"/>.
@@ -81,11 +87,12 @@ namespace Ryujinx.Audio
         /// <param name="updatedRequiredEvent ">The driver event that will get signaled by the device driver when an audio buffer finished playing/being captured</param>
         /// <param name="outputCallback">The callback to call when an audio buffer finished playing</param>
         /// <param name="inputCallback">The callback to call when an audio buffer was captured</param>
-        public void Initialize(ManualResetEvent updatedRequiredEvent, Action outputCallback, Action inputCallback)
+        public void Initialize(IHardwareDeviceDriver deviceDriver, Action outputCallback, Action inputCallback)
         {
             lock (_lock)
             {
-                _updateRequiredEvents[0] = updatedRequiredEvent;
+                _deviceDriver = deviceDriver;
+                _updateRequiredEvents[0] = deviceDriver.GetUpdateRequiredEvent();
                 _actions[0] = outputCallback;
                 _actions[1] = inputCallback;
             }
@@ -98,6 +105,8 @@ namespace Ryujinx.Audio
         {
             while (true)
             {
+                //_deviceDriver.GetPauseEvent().WaitOne();
+
                 int index = WaitHandle.WaitAny(_updateRequiredEvents);
 
                 // Last index is here to indicate thread termination.
--- a/Ryujinx.Audio/AudioManager.cs
+++ b/Ryujinx.Audio/AudioManager.cs
@@ -45,6 +45,8 @@ namespace Ryujinx.Audio
         /// </summary>
         private Thread _workerThread;
 
+        private bool _isRunning;
+
         /// <summary>
         /// Create a new <see cref="AudioManager"/>.
         /// </summary>
@@ -52,6 +54,7 @@ namespace Ryujinx.Audio
         {
             _updateRequiredEvents = new ManualResetEvent[2];
             _actions = new Action[2];
+            _isRunning = false;
 
             // Termination event.
             _updateRequiredEvents[1] = new ManualResetEvent(false);
@@ -72,6 +75,7 @@ namespace Ryujinx.Audio
                 throw new InvalidOperationException();
             }
 
+            _isRunning = true;
             _workerThread.Start();
         }
 
@@ -96,7 +100,7 @@ namespace Ryujinx.Audio
         /// </summary>
         private void Update()
         {
-            while (true)
+            while (_isRunning)
             {
                 int index = WaitHandle.WaitAny(_updateRequiredEvents);
 
@@ -116,6 +120,14 @@ namespace Ryujinx.Audio
                     _updateRequiredEvents[0].Reset();
                 }
             }
+        }
+
+        /// <summary>
+        /// Stop updating the <see cref="AudioManager"/> without stopping worker thread.
+        /// </summary>
+        public void StopUpdates()
+        {
+            _isRunning = false;
         }
 
         public void Dispose()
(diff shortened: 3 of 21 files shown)


Thog:
This is not enough to pause emulation.
If you want to pause emulation entirely you need to pause all guest processes running.
The way you handle it will likely make some games crashes

mpnico:
Thx for pointing this out, I let it as a draft for now and will investigate more on other threads to pause.

Ryujinx.Audio/Integration/IHardwareDeviceDriver.cs
```diff
@@ -46,5 +47,6 @@ IHardwareDeviceDriver GetRealDeviceDriver()
         {
             return this;
         }
+
```
riperiperi, line 50: Unnecessary space.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -456,5 +456,25 @@ protected virtual void Dispose(bool disposing)
...
         }
+
+        public void TogglePauseEmulation(bool pause)
+        {
+            lock (KernelContext.Processes)
+            {
+                foreach (KProcess process in KernelContext.Processes.Values)
+                {
+                    process.SetThreadActivity(pause);
+                }
```
riperiperi, line 467: Maybe add a new line after this.

riperiperi (commented):
Looks good at a glance, still need to test more games. Nice to see pausing using SetActivity working.

I think that games should be sent an applet message when they resume from pause - to simulate returning from a sleep or the home menu. Maybe do the same signal as "Simulate Wake-up Message"?

Also, audren doesn't seem to be pausing... Try going ingame in Mario Kart 8 and pausing, the looping sound will continue playing for the kart engines etc.

Ryujinx.Audio/AudioManager.cs
```diff
@@ -98,6 +105,8 @@ private void Update()
         {
             while (true)
             {
+                //_deviceDriver.GetPauseEvent().WaitOne();
```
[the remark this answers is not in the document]
mpnico: It was to pause some audio but wasn't necessary during my tests. Removed

mpnico:
Thx for review. During dev of this PR, I was able to reproduce the engine sound loop in MK8D but it should be fixed by the pause event in AudioProcessor loop [L208](https://github.com/Ryujinx/Ryujinx/pull/2428/files#diff-7f3a3e2811dc95012e0ceade98a4844040b163b4bb7b6854f13e73a7d15d3b73R208). I can't reproduce it right now.

Btw i fixed suggested changes and add the suggested wake up event on resume.

LukeWarnut:
I tested the feature with the following games. It had no issues with any of them.

Arcade Archives Donkey Kong (1.0.0)
Balan Wonderworld Demo (1.0.0)
Clubhouse Games: 51 Worldwide Classics (1.1.2)
DOOM 64 (1.0.2)
Mario Kart 8 Deluxe (1.7.2)
Polyroll (1.3)
Shovel Knight: Treasure Trove (3.00)
Slay the Spire (2.0.3)
Tetris 99 (2.1.0)

MutantAura:
Tested in the following games:

**Mario Kart 8 Deluxe (1.7.1)** - No issues
**Pokemon Mystery Dungeon: Rescue Team DX (1.0.2)** - no issues with the feature but this build displays some ghosting around characters not seen in 1.0.6957. Could just be a base issue. (see screenshots below)
**Super Smash Bros. Ultimate (12.0.0)** - No issues
**The Legend of Zelda: Breath of the Wild (1.6.0)** (static 30fps mod) - No issues

Audio was fully synced after multiple pauses and multiple minutes of pausing. Worked fine in all cutscenes and gameplay that I tested across each game.

MutantAura:
To add an issue to my previous comment going from windowed to fullscreen or visa versa is broken. The graphics do not switch and remain in a semi bugged state until emulation is resumed. 
image
and 
image

mpnico:
Not sure if I can do much about this. You can observe the same behavior when starting a game during shader loading. Maybe resume then immediaty pause again when pressing F11 ? I'll give it a try.

emmauss:
To solve that, you would need to keep drawing the last frame until a new one is ready. so the rendering thread can't just be paused

riperiperi (approved):
LGTM. The pausing in MK8 is working now, not sure why it wasn't before...

I tried a few more obscure games like LM3 (crashes when gpu stalls) and HW:AoC (runs game logic asynchronously to rendering) and they both work properly when the game is paused for a while. Something of interest is that HW:AoC's dynamic resolution seems to plummet after pausing for a while - not sure why. Obviously, it quickly raises afterwards.

As for the issue where the screen is not updated when paused, I don't think it's in the scope of this PR to resolve. The same thing happens when an applet stops the game from rendering, or if there is any other reason for the game to stop rendering. I feel like this should be done all-in-one with changes to allow compositing multiple layers and supporting applet capture services.

FICTURE7:
Perhaps `_tickTimer` in `ExecutionContext` of must be paused as well?

https://github.com/Ryujinx/Ryujinx/blob/0d841c8d5104a09d2733c0e78f6d5b7ebc8fee3e/ARMeilleure/State/ExecutionContext.cs#L17-L41

riperiperi:
> Something of interest is that HW:AoC's dynamic resolution seems to plummet after pausing for a while - not sure why. Obviously, it quickly raises afterwards.

I tested this on console, and it does indeed greatly lower the resolution after being in the home menu for a few seconds. This behaviour might be expected.

Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
```diff
@@ -16,6 +16,7 @@ public class OpenALHardwareDeviceDriver : IHardwareDeviceDriver
         private ALDevice _device;
         private ALContext _context;
         private ManualResetEvent _updateRequiredEvent;
+        private ManualResetEvent _pauseEvent;
```
gdkchan, line 19: This must be disposed when `Dispose` is called.

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
```diff
@@ -1082,5 +1082,13 @@ private void UndefinedInstructionHandler(object sender, InstUndefinedEventArgs e
         }
 
         protected override void Destroy() => Context.Dispose();
+
+        public void SetThreadActivity(bool pause)
```
gdkchan, line 1086: Would be nice to implement this like the kernel does. Just calling `SetActivity` directly might resume threads that were previously paused by the game (I have only ever seen Unity games using this, and I'm not sure for what).
You can use Atmosphere implementation as reference:
https://github.com/Atmosphere-NX/Atmosphere/blob/master/libraries/libmesosphere/source/kern_k_process.cpp#L981
The equivalent for their `m_suspend_request_flags` on Ryujinx is `_forcePauseFlags`. The suspend and resume methods are just 3 lines of code, so should be simple to implement, assuming that everything else is already in place (should be, but I don't think process pausing was ever tested before).

mpnico: You mean to store previous `_forcePauseFlags` in each `KThread` in case of pause, to then be able to resume only previously running threads ?

gdkchan: I meant implementing the process `SetActivity` function like the Switch kernel in general. When a process is suspended, the `_forcePauseFlags` value of all the threads of the process is changed to indicate that the thread is suspended, and then it calls `AdjustScheduling` to make the thread actually suspend. The code I linked should be equivalent to how it's done on the Switch, and can be used as reference.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -456,5 +456,27 @@ protected virtual void Dispose(bool disposing)
...
+
+                if (pause)
+                {
+                    Device.AudioDeviceDriver.GetPauseEvent().Reset();
+                }
+                else
+                {
+                    SimulateWakeUpMessage();
```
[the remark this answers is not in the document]
mpnico: This was suggested by @riperiperi a few comments below.

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
mpnico: Seems to work correctly !

Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs
```diff
@@ -471,6 +471,29 @@ public void SetPriority(int priority)
             KernelContext.CriticalSection.Leave();
         }
 
+        public void Suspend()
+        {
+            _forcePauseFlags |= ThreadSchedState.ThreadPauseFlag;
```
gdkchan, line 476: Looks mostly correct, but the function should be taking a `ThreadSchedState` argument. For the process `SetActivity`, it should be passing `ThreadSchedState.ProcessPauseFlag` and for the thread `SetActivity`, it should be passing `ThreadSchedState.ThreadPauseFlag`. This is what prevents for example, process `SetActivity` set to true resuming a thread that was previously paused by thread `SetActivity` set to false. On the code I linked, it passes `SuspendType_Process` from process and `SuspendType_Thread` from `KThread::SetActivity`. Also:
- `KProcess.SetActivity` should call `Context.CriticalSection.Enter` at the start, and `Context.CriticalSection.Leave` before return to ensure that the scheduler lock is acquired, otherwise it might have a race between other thread scheduling operations.
- It should also `lock (_threadingLock)` to avoid races with the `AddThread` method, since the linked list is not thread safe.

mpnico: Ok, I miss that part of the code. It's done and it's working. Thank you for the clear explanations.

gdkchan:
I'm testing it right now, so far the only issues I have found are the following:
- The F5 hotkey doesn't work here. Pressing it does nothing here. Clicking Actions > Pause Emulation and Actions > Resume Emulation works properly. Maybe it's related to my config file? (I did not touch it, btw).
- On Hatsune Miku, pausing and resuming on a song causes the pause menu to open. It's a side effect of calling `SimulateWakeUpMessage` since the same thing happens if I click on "Simulate wake-up message". Dunno if the behaviour is desirable or not.

gdkchan:
Also getting a NRE here when stopping emulation or closing the emulator:
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Ryujinx.Audio.Renderer.Dsp.AudioProcessor.Dispose(Boolean disposing) in C:\Users\gdk\Documents\Ryujinx\Ryujinx.Audio\Renderer\Dsp\AudioProcessor.cs:line 257
   at Ryujinx.Audio.Renderer.Server.AudioRendererManager.Dispose(Boolean disposing) in C:\Users\gdk\Documents\Ryujinx\Ryujinx.Audio\Renderer\Server\AudioRendererManager.cs:line 348
   at Ryujinx.HLE.HOS.Horizon.Dispose(Boolean disposing) in C:\Users\gdk\Documents\Ryujinx\Ryujinx.HLE\HOS\Horizon.cs:line 448
   at Ryujinx.HLE.Switch.Dispose(Boolean disposing) in C:\Users\gdk\Documents\Ryujinx\Ryujinx.HLE\Switch.cs:line 159
   at Ryujinx.Ui.MainWindow.CreateGameWindow() in C:\Users\gdk\Documents\Ryujinx\Ryujinx\Ui\MainWindow.cs:line 861
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
   ```

**Comment by mpnico** [comment]
For the hotkey it's a config problem I should address. For the `SimulateWakeUpMessage` I think we should remove it as the pause from emulator should be _invisible_ to the game.

And I will fix the NRE. Thank you for testing.

**Comment by riperiperi** [comment]
My argument for leaving the wakeup message in is that games will deal better with a large jump in time if they know that it's due to the application being "resumed", as it is when the system resumes from sleep or returns from the home menu.

**Comment by MutantAura** [comment]
I feel like surely if the game is going to respond to an emulator action that should be completely separate (like actually pausing in-game the frame after) this feature loses some of the reason it exists? Pause and resume play should ideally be unknown to the game you're playing. Of course I have no idea of the technicalities like peri mentioned of a game suddenly getting a large time jump but I feel this is a pretty massive trade to eliminate the chance of that being an issue? 

I could leave something paused overnight with an older commit if needed like Sword/Shield maybe and see how it handles long-term time jumps.

**Comment by riperiperi** [comment]
It's not a massive trade at all - it's the game's expected behaviour for being interrupted. Some games even restart their audio renderer when this happens.

There is a lot to consider changing if you want pausing to be truly "invisible" rather than imitating suspending the application as you would on the switch, including all timers, timestamps and the RTC.

**Comment by gdkchan** [comment]
If a large jump in time can cause game issues, then we should just pause the timer aswell IMO. However I believe games should be prepared to handle this regardless of those messages, since it's only going to receive the message *after* all threads have resumed. Games usually either have a dedicated thread for this, that wakes up on the message receive event, which would then need to take action after finding the application was just resumed, or periodic checks on one thread (usually on the main loop).

**Comment by mpnico** [comment]
For now I removed the `SimulateWakeUpMessage`. I also added a pause in the `_tickCounter` in `ExecutionContext` as suggested by @FICTURE7 which fix some animations jump in Monster Hunter Rise after resuming the game.

File: ARMeilleure/State/ExecutionContext.cs [hunk 8]
```diff
@@ -145,6 +145,16 @@ public void StopRunning()
             _nativeContext.SetCounter(0);
         }
 
+        public void Suspend()
```
gdkchan, line 148: Those methods should be static, since the tick counter is static.

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
```diff
@@ -1082,5 +1082,68 @@ private void UndefinedInstructionHandler(object sender, InstUndefinedEventArgs e
         }
 
         protected override void Destroy() => Context.Dispose();
+
+        public KernelResult SetActivity(bool pause)
+        {
+            if (Flags.HasFlag(ProcessCreationFlags.IsApplication))
```
gdkchan, line 1088: This check should be inside `foreach (KProcess process in KernelContext.Processes.Values)` rather than here. We should still support `SetActivity` in any `KProcess` imo, as this is the kernel behaviour.

ARMeilleure/State/ExecutionContext.cs
gdkchan: Also they should be called `SuspendTimer` and `ResumeTimer` in my opinion, to be clear about what is being suspended here. `SuspendCounter` and `ResumeCounter` sounds fine too.

mpnico:
Fixed config migration bug and applied suggestions

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -463,16 +463,22 @@ public void TogglePauseEmulation(bool pause)
             {
                 foreach (KProcess process in KernelContext.Processes.Values)
                 {
-                    process.SetActivity(pause);
+                    if (!process.Flags.HasFlag(ProcessCreationFlags.IsApplication))
```
gdkchan, line 466: This check seems inverted. The game process is the one with `IsApplication` flag set, not the other way around.

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -463,16 +463,22 @@ public void TogglePauseEmulation(bool pause)
...
+                    {
+                        // only game process should be paused
```
gdkchan, line 468:
```suggestion
                        // Only game process should be paused.
```

Ryujinx.HLE/HOS/Horizon.cs
mpnico: That what I thought but the previous problem with Monster Hunter Rise and non schedulable threads is only fixed that way and pause seems to work perfectly like this. Hard to understand for now.

gdkchan: If the KProcess has the IsApplication flag set, then all its KThreads should have IsSchedulable as true. And if it is *not* set, then all its KThreads should have IsSchedulable as false. If this is not the case today, then something else is wrong. Anyway, pausing the service processes would also cause the game to pause since none of its service requirests will be processed, causing all threads trying to make a service request to block, but some of its threads could remain active which is undesirable.

mpnico: I can confirm with MHR, I have a thread with `IsApplication` flag set (flags are 1100111) and owner process has `IsScheculable` to `false`

gdkchan: I took a look at this, the problem is that `MakeUnschedulable` is called before the thread exits, to make sure the thread can't be scheduled again (sinnce it will exit, so it is not supposed to be running again). `MakeUnschedulable` sets `_forcedUnschedulable` to true, and `IsSchedulable` does this check:
```cs
public bool IsSchedulable => _customThreadStart == null && !_forcedUnschedulable;
```
The problem is that `AdjustScheduling` will try to set or reset the event directly if the thread is not "schedulable". This is required for HLE threads, as they are not assigned to a core like the game guest threads, but still need to suspend and resume for some syscalls.

My proposed fix is simply not doing anything if `_forcedUnschedulable` is true. You can just replace the respective code in the `AdjustScheduling` function with the snippet below:
```cs
            if (!IsSchedulable)
            {
                if (!_forcedUnschedulable)
                {
                    // Ensure our thread is running and we have an event.
                    StartHostThread();

                    // If the thread is not schedulable, we want to just run or pause
                    // it directly as we don't care about priority or the core it is
                    // running on in this case.
                    if (SchedFlags == ThreadSchedState.Running)
                    {
                        _schedulerWaitEvent.Set();
                    }
                    else
                    {
                        _schedulerWaitEvent.Reset();
                    }
                }

                return;
            }
```
Then on Horizon you can change the check to `if (process.Flags.HasFlag(ProcessCreationFlags.IsApplication))`. I have verified that MHR pauses and unpauses properly with this change.

gdkchan:
I just noticed that closing the emulator *while* the game is paused doesn't work properly, it just hangs.

Ryujinx.HLE/HOS/Horizon.cs
mpnico: Done, thank you for investigating.

mpnico:
It's fixed, good catch :)

gdkchan:
Getting this exception when I click on "Stop emulation":
```
00:01:30.581 |E| Application : Unhandled exception caught: System.ObjectDisposedException: Safe handle has been closed.
Object name: 'SafeHandle'.
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at System.StubHelpers.StubHelpers.SafeHandleAddRef(SafeHandle pHandle, Boolean& success)
   at Interop.Kernel32.SetEvent(SafeWaitHandle handle)
   at System.Threading.EventWaitHandle.Set()
   at Ryujinx.HLE.HOS.Horizon.TogglePauseEmulation(Boolean pause)
```

mpnico:
It should be better now

gdkchan (approved):
lgtm, thanks.
One thing that I noticed though, is that you can hear the game music continuing right after you click stop emulation with the game paused. Not a big issue really, but I personally prefer that it would be better if the game being unpaused was not noticeable by the user.

Ryujinx.Audio/AudioManager.cs
```diff
@@ -45,6 +46,11 @@ public class AudioManager : IDisposable
         /// </summary>
         private Thread _workerThread;
 
+        /// <summary>
+        /// The audio device driver.
+        /// </summary>
+        private IHardwareDeviceDriver _deviceDriver;
```
Thog, line 52: Added but unused, this should be removed.

Ryujinx.Audio/AudioManager.cs
```diff
@@ -78,14 +84,15 @@ public void Start()
...
+        /// <param name="deviceDriver ">The device driver to get the driver event when an audio buffer finished playing/being captured and the pause event</param>
         /// <param name="outputCallback">The callback to call when an audio buffer finished playing</param>
         /// <param name="inputCallback">The callback to call when an audio buffer was captured</param>
-        public void Initialize(ManualResetEvent updatedRequiredEvent, Action outputCallback, Action inputCallback)
+        public void Initialize(IHardwareDeviceDriver deviceDriver, Action outputCallback, Action inputCallback)
         {
             lock (_lock)
             {
-                _updateRequiredEvents[0] = updatedRequiredEvent;
+                _deviceDriver = deviceDriver;
```
Thog, line 94: Added but unused, this should be removed.

Thog (changes requested):
Some minor clean up to do, otherwise lgtm

mpnico:
`AudioManager` code cleaned, thank you @Thog.

@gdkchan I agree about the stop button. If it is not get merged before next week, I should be able to find a solution for that, as I can't work on it this week.

PR #2433, opened 2021-06-30
Honour copy dependencies when switching render target

riperiperi:
When switching from one render target to another, when both have a copy dependency to each other, a copy can be deferred on the second target when unbinding the first.

Before, this would not be honoured before binding the new texture, so the copy would stay deferred until the render targets change again, at which point it would copy in old data and essentially clear all the draws done during that time.

This change runs synchronize memory to make sure that copies are honoured. This can cause a redundant copy, but this is a rare situation and it's better than it breaking for now.

This should fix miiedit on AMD/Intel GPUs on windows. May fix other games, or perhaps rare copy dependency bugs on NVIDIA too.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
@@ -154,8 +154,12 @@ namespace Ryujinx.Graphics.Gpu.Image
             if (_rtColors[index] != color)
             {
                 _rtColors[index]?.SignalModifying(false);
-                color?.SynchronizeMemory();
-                color?.SignalModifying(true);
+
+                if (color != null)
+                {
+                    color.SynchronizeMemory();
+                    color.SignalModifying(true);
+                }
 
                 _rtColors[index] = color;
             }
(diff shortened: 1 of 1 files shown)


Thog (approved):

Leischii:
Fixes missing menu items in Pokemon Snap with Amd gpu

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -154,6 +154,7 @@ public bool SetRenderTargetColor(int index, Texture color)
             if (_rtColors[index] != color)
             {
                 _rtColors[index]?.SignalModifying(false);
+                color?.SynchronizeMemory();
```
gdkchan, line 157: I think it would be better to do `if (color != null)` here, otherwise it will do 2 null checks (checked via SharpLab disasm).

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -175,6 +176,7 @@ public bool SetRenderTargetDepthStencil(Texture depthStencil)
             if (_rtDepthStencil != depthStencil)
             {
                 _rtDepthStencil?.SignalModifying(false);
+                depthStencil?.SynchronizeMemory();
```
gdkchan, line 179: Same comment.

riperiperi:
Addressed feedback.

gdkchan (approved):
Thanks for the quick fix.

PR #2419, opened 2021-06-27
no name: Mii Editor applet support

AcK77:
# Project no name

Why "no name"? Because it's the Miis default name in the Switch service so yeah, let's introduce a standalone mode for official Mii Editor applet.
Since Ryujinx requires a Switch firmware to be installed, we can run applets directly from the firmware's files. In that way we can run the Mii Editor applet in a standalone mode.
That allows you to create, edit, rename, save, move and delete Miis in the `mii` service system save data, which is required by some games such as Mario Kart 8 Deluxe, Miitopia or more recently Mario Golf (in adventure mode). Now save files aren't needed.

image
image
image
image
image
image
image
image

Launching applets from games is not supported. For this reason, if you play a game that requires having Miis created, you need to first launch the Mii applet (from the File > Launch Mii Applet menu), create a Mii of your liking, and then launch the game. The Mii should be available for selection when you start a new game. Not following those steps will result in a crash when the game requests the Mii.

image
image

### What's implemented
- A bunch of services calls needed to get it working.
- You can run applets directly from the File menu.
  
image

- When you install a new (or old) firmware, all caches related to the applet are purged automatically to avoid any issues.
- Touch is supported aswell, that means you can drag and drop Miis with your mouse to sort them.

image

### Known Issues
- Mii Editor from Firmware < 3.0.1 seems to crash with a GPU issue.
- If you press B on the main screen, it will try to exit the applet and crash because of unimplemented call of am service.
- "Send/Receive" currently don't work due to unimplemented call of nifm service.
- If you try to "Copy Mii from amiibo" in "Create New Mii", you will be able to scan a virtual amiibo by checking "Show All Amiibo", but the returned mii values are wrong.
- That doesn't let you run the Mii Editor applet from a running game.

Tests are welcome, don't hesitate to post any feedbacks or issues that aren't known.

Thanks to @riperiperi , @Thog and @gdkchan for their precious help as usual.

image
image
image
image

PR #2414, opened 2021-06-27
aoc/am: Cleanup aoc service and stub am calls

AcK77:
This PR implement aoc call `GetAddOnContentListChangedEventWithProcessId` (Closes #2408) and `CreateContentsServiceManager`. Additionnally, a big cleanup (checked by RE on latest firmware) is made on the whole service. I've added `CountAddOnContent`, `ListAddOnContent` and `GetAddonContentBaseId` for games which require version `1.0.0-6.2.0` too.

Am service call `ReportUserIsActive` is stubbed (checked by RE, closes #2413).

Since some logic in the service (aoc) which handle the DLCs has been changed, it could be nice to have some testing to be sure there is no regression.

Tony Hawk’s Pro Skater 1+2 is now bootable:
image

PR #2434, opened 2021-07-01
aoc: Fixes some inconsistencies

AcK77:
This PR fixes an wrong returned value (introduced in #2414) which cause some DLC not recognized in some games like Super Robot War T.
In #2414 I've removed the EventHandle check too, because it could cause some issues, but sadly it doesn't do the job so I reverted changes. It should fix Diablo III: Eternal Collection.

Revert a change made to desired title language in Traditional Chinese in #2419.


Thog (approved):

riperiperi (approved):
LGTM.

emmauss (approved):

PR #2438, opened 2021-07-02
changing the session cleaning to fix a crash while running Game Builder Garage

Gummygamer:


Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -269,7 +277,7 @@ private bool Process(int serverSessionHandle, ulong recvListAddr)
                     {
                         disposableObj.Dispose();
                     }
-                    _sessions.Remove(serverSessionHandle);
+                    //_sessions.Remove(serverSessionHandle);
```
emmauss, line 280: removed commented out code

Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -78,7 +78,15 @@ public void AddSessionObj(KServerSession serverSession, IpcService obj)
...
-            _sessions.Add(serverSessionHandle, obj);
+            try
+            {
+                _sessions.Add(serverSessionHandle, obj);
+            }
+            catch
+            {
+                _sessions.Remove(serverSessionHandle);
+                _sessions.Add(serverSessionHandle, obj);
+            }
```
Benedani, line 89: There is something called `_sessions.ContainsKey`. No need for the slower try-catch.

Ryujinx.HLE/HOS/Services/ServerBase.cs
Benedani, line 280: Are you sure this won't cause any regressions?

Benedani (changes requested):

Ryujinx.HLE/HOS/Services/ServerBase.cs
Gummygamer: It would easily cause a crash if it weren't by the approach of removing sessions if they collide in key. As it is I'm more concerned with possible performance issues like in memory usage or speed. I'm really thinking it might be there a better way to manage that memory.

Ryujinx.HLE/HOS/Services/ServerBase.cs
Gummygamer: Changed.

Ryujinx.HLE/HOS/Services/ServerBase.cs
Gummygamer: changed.

Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -269,7 +277,6 @@ private bool Process(int serverSessionHandle, ulong recvListAddr)
                     {
                         disposableObj.Dispose();
                     }
-                    _sessions.Remove(serverSessionHandle);
```
AcK77, line 272: if the `IpcMessageType` is `HipcCloseSession` of `TipcCloseSession` I'm pretty sure the session have to be removed here and not in the `AddSessionObj` method.

Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -78,7 +78,15 @@ public void AddSessionObj(KServerSession serverSession, IpcService obj)
         public void AddSessionObj(int serverSessionHandle, IpcService obj)
         {
             _sessionHandles.Add(serverSessionHandle);
-            _sessions.Add(serverSessionHandle, obj);
+            if(!_sessions.ContainsKey(serverSessionHandle))
```
AcK77, line 81:
```suggestion
            if (!_sessions.ContainsKey(serverSessionHandle))
```

Imo this looks like more as a hack than a real fix :/

AcK77 (changes requested):

Ryujinx.HLE/HOS/Services/ServerBase.cs
Gummygamer: I understand and would love some help in reaching a better solution if possible. This PR avoided crashes and, because I remove sessions when they collide in key, it remains viable to run. That's why it's submitted at all, but I know it deserves polish.

Ryujinx.HLE/HOS/Services/ServerBase.cs
Gummygamer: Interesting distinction in terms of message type. Should I add a branch (if) to detect that?

AcK77: It's just a way to handle both type, there is no distinction in terms of functionality here.

Gummygamer: Now I've checked the context of the code in question and I suppose not removing the sessions in those conditions is core to my solution to the crashes. The crashes happened when a session key was not found in the dictionary. So...maybe Ryujinx was trying to close sessions too soon without my PR? Again, I would love help with a "smoother" solution if the current one is too rough.

AcK77:
Closing as stale and because this isn't a proper fix.

PR #2440, opened 2021-07-03
Separate GPU engines (part 2/2)

gdkchan:
**Overview**

This is a continuation of #2422, this time for the 3D engine. Code has been moved from the catch-all `Methods` class, to the per-engine Threed (3D) class. This time, there's no official headers sadly, so the structures were generated from the old code.

**Highlights**

- 3D engine now also uses `DeviceState`, like the change done to the other engines on #2422.
  - At the time of writing, NVIDIA has not published any official documentation or definitions for this engine, so this is still a structure based on information obtained  from reverse-engineering and guesswork.
  - State is now per-channel rather than shared. With this change, now all the GPU emulation state is per-channel, and nothing is shared.
  - The `Methods` class is now gone, it has been replaced by per-engine classes.
  - New state update method. The previously method required explictly calling `QueryModified` on each state group to check if the state was modified. The new method uses a bitmask, and sets a bit for each modified state group. Before draw, it only needs to check and clear the bitmask bits, until it is 0. If we have N state groups to check, the old method always did N checks, the new method does only one check in the best case (nothing was modified), and N checks in the worst case (everything was modified). Should be more efficient.
  - The force shader update flag was removed, it can now just set the respective bit on the bitmask to force a shader state update.
  - `Ryujinx.Graphics.Gpu.State` namespace is gone, state has either been moved inside the classes they are used, inside the per-engine namespaces, or to the `Ryujinx.Graphics.Gpu.Engine.Types` namespace if it is a type shared by more than one engine.
- Fixed an oversight where the host Y flip was not being updated on writes to the `YNegate` register.
  - Fixes rendering on several upside down OpenGL games (see screenshots below).
  - Also fixes rendering on Game Tengoku CruisinMix Special (which was looking completely corrupt before).
- Add missing `PrimitiveTypeOverrideEnable` register, used to enable the primitive type override used by the Vulkan driver (thanks to ByLaws for testing on hardware).
  - Previously it was simply checking if the `PrimitiveTypeOverride` had a valid value, and using it if so, which is not correct.
  - Likely what fixes Turok 2 menu (see screenshot below), however it was not confirmed.
- Channel state is now properly initialized with the same values that Nvservices uses for the registers, rather than guessed values.
- `DeviceState` register read/write functions were optimized, since those operations are very frequent on the 3D engine.
  - The `RegisterAttribute` and `AccessControl` were removed as nothing was using them. It was intended to control whenever a register can be read and written, useful for write-only registers that always read as 0, or read-only registers that can't be modified. The reason for the removal is that it requires additional checks on the read and write funcitons. 

**Fixes**

Fixes some upside down games.
Before:
image
After:
image

Fixes rendering on Game Tengoku CruisinMix Special.
Before:
image
After:
image

Fixes Turok 2 menu.
Before:
image
After:
image


GamerzHell9137:
PR fixed upside down in Dragon Quest Builders!### 
image

Niwu34:
image
Fixes 20XX being upside down

MarieTheLegend:
PR fixed upside down in Asterix & Obelix XXL 2

Bild_2021-07-03_230916

zombie167:
This PR fixes the problem of the upside down images of the game BLADE ARCUS Rebellion from Shining

t2

gdkchan:
Rebased and updated.

Ryujinx.Graphics.Device/DeviceState.cs
```diff
@@ -102,20 +102,43 @@ public void Write(int offset, int data)
...
-#endif
+                DebugWrite(alignedOffset, data);
 
                 GetRefUnchecked<int>(alignedOffset) = data;
 
                 Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(_writeCallbacks), (IntPtr)index)?.Invoke(data);
             }
         }
 
+        public bool WriteWithRedundancyCheck(int offset, int data)
```
FICTURE7, line 113: If this is changed to something like `Write(int offset, int data, out bool changed)`, the JIT will turn the call to the write callback to a tailcall and will not setup a call frame.

Ryujinx.Graphics.Device/DeviceState.cs
```diff
@@ -102,20 +102,43 @@ public void Write(int offset, int data)
...
+        {
+            bool changed = false;
+            uint index = (uint)offset / RegisterSize;
+
+            if (index < _size)
```
FICTURE7, line 118: `_size` can become a compile time constant by turning it in an expression. Perhaps an inlineable property e.g:

```csharp
private uint Size => (uint)(Unsafe.SizeOf<TState>() + RegisterSize - 1) / RegisterSize;
```

gdkchan:
Thanks for the suggestions. For the record, this is the new codegen (for `DeviceState<GPFifoClassState>.Write`):
```x86asm
00007FF7C85F2660  shr         edx,2  
00007FF7C85F2663  cmp         edx,4Ah  
00007FF7C85F2666  jae         Ryujinx.Graphics.Device.DeviceState`1[[Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoClassState, Ryujinx.Graphics.Gpu]].Write(Int32, Int32)+03Dh (07FF7C85F269Dh)  
00007FF7C85F2668  mov         eax,edx  
00007FF7C85F266A  shl         eax,2  
00007FF7C85F266D  cmp         dword ptr [rcx],ecx  
00007FF7C85F266F  lea         r9,[rcx+28h]  
00007FF7C85F2673  mov         eax,eax  
00007FF7C85F2675  mov         dword ptr [r9+rax],r8d  
00007FF7C85F2679  mov         rcx,qword ptr [rcx+10h]  
00007FF7C85F267D  cmp         dword ptr [rcx],ecx  
00007FF7C85F267F  add         rcx,10h  
00007FF7C85F2683  mov         edx,edx  
00007FF7C85F2685  mov         rax,qword ptr [rcx+rdx*8]  
00007FF7C85F2689  test        rax,rax  
00007FF7C85F268C  jne         Ryujinx.Graphics.Device.DeviceState`1[[Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoClassState, Ryujinx.Graphics.Gpu]].Write(Int32, Int32)+02Fh (07FF7C85F268Fh)  
00007FF7C85F268E  ret  
00007FF7C85F268F  mov         rcx,qword ptr [rax+8]  
00007FF7C85F2693  mov         edx,r8d  
00007FF7C85F2696  mov         rax,qword ptr [rax+18h]  
00007FF7C85F269A  jmp         rax
00007FF7C85F269D  ret  
```

Ryujinx.Graphics.Device/DeviceState.cs
```diff
@@ -79,48 +69,71 @@ public DeviceState(IReadOnlyDictionary<string, RwCallback> callbacks = null, Act
...
-                if (_writeCallbacks.TryGetValue(alignedOffset, out Action<int> write))
-                {
-                    write(data);
-                }
+            if (index < Size)
+            {
+                uint alignedOffset = index * RegisterSize;
+                DebugWrite(alignedOffset, data);
+
+                ref var storage = ref GetRefUnchecked<int>(alignedOffset);
```
FICTURE7, line 118: If you feel like carrying RyuJIT a little more you can do this
```csharp
ref var storage = ref GetReIntAlignedUnchecked(index);

private ref int GetRefIntAlignedUncheck(ulong index)
{
    return ref Unsafe.Add(ref Unsafe.As<TState, int>(ref State), (IntPtr)index);
}
```
It will fold the `alignedOffset` multiplication on the memory operand itself. Changing `index` to `ulong` will avoid an extra zero-extending `mov`.

Also using
```csharp
changed = Interlocked.Exchange(ref storage, data) != data;
```
did not seem profitable here, in case anyone was wondering.

riperiperi (approved):
Nice one! I've tested this on a few games, but I'll give it a test on a couple more to make sure nothing has regressed.

Ryujinx.Graphics.Device/DeviceState.cs
gdkchan: Thanks for the suggestion. Initially I misunderstood it, and thought that just changing the type to ulong would be enough to get it folded, but it wasn't.
Anyway, before this change:
```x86asm
00007FF7C85DD0E0  cmp         dword ptr [rcx],ecx  
00007FF7C85DD0E2  lea         r10,[rcx+28h]  
00007FF7C85DD0E6  mov         eax,eax  
00007FF7C85DD0E8  add         rax,r10  
00007FF7C85DD0EB  cmp         dword ptr [rax],r8d  
00007FF7C85DD0EE  setne       r10b  
00007FF7C85DD0F2  mov         byte ptr [r9],r10b
00007FF7C85DD0F5  mov         dword ptr [rax],r8d  
```
After:
```x86asm
00007FF97B7B447B  cmp         dword ptr [rcx],ecx  
00007FF97B7B447D  lea         rax,[rcx+28h]  
00007FF97B7B4481  mov         r10d,edx  
00007FF97B7B4484  lea         rax,[rax+r10*4]  
00007FF97B7B4488  cmp         dword ptr [rax],r8d  
00007FF97B7B448B  setne       r10b  
00007FF97B7B448F  mov         byte ptr [r9],r10b  
00007FF97B7B4492  mov         dword ptr [rax],r8d 
```
The raw instruction count on that part is the same, but it should allow the removal of the `shl` instruction. I'm not sure why it is not combining:
```x86asm
00007FF97B7B447D  lea         rax,[rcx+28h]  
00007FF97B7B4484  lea         rax,[rax+r10*4]  
```
Into:
```x86asm
00007FF97B7B447D  lea         rax,[rcx+rdx*4+28h]  
```
I guess it's because index is a uint, and it can't figure out that the upper bits are already zero, so it does this zero extension.

FICTURE7: I believe this is because RyuJIT does not do forward substitution. Some of these days, it'll be able to do it 😩.

Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -0,0 +1,1044 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Update()
+        {
+            // The vertex buffer size is calculated using a different
+            // method when doing indexed draws, so we need to make sure
+            // to update the vertex buffers if we are doing a regular
+            // draw after a indexed one and vice-versa.
+            if (_drawState.DrawIndexed != _prevDrawIndexed)
+            {
+                _updateTracker.ForceDirty(VertexBufferStateIndex);
```
gdkchan, line 187: Note that before it updated the vertex buffer state every time the index buffer state was modified. This improves the situation by only doing so on non-indexed -> indexed or indexed -> non-indexed draw transition. This could potentially cause a change in behavior though, just something to keep in mind.

Thog (approved):

Benedani:
Getting a 10% boost on heavy areas in Super Mario Odyssey. Small, but a boost nonetheless.

Issue #2436, opened 2021-07-02
Monster hunter stories 2 demo :  Resolution scaling issue during battle


c3ntrin0:
Game works perfectly with resolution scaling  in village and open world. 
However during battle resolution scaling makes the monster invisible, and need to switch to native res to fix it.

I believe full release will have same issue.
This is a great game to be played with high res, please fix the res scaling issue.
![image](https://user-images.githubusercontent.com/68693384/124274315-b0de1b00-db73-11eb-8573-310263ea27d1.png)
![image](https://user-images.githubusercontent.com/68693384/124274379-c6534500-db73-11eb-99b8-603e1f48377b.png)
![image](https://user-images.githubusercontent.com/68693384/124274438-d8cd7e80-db73-11eb-8151-05ef52a2da8b.png)

PR #2441, opened 2021-07-03
Unscale textureSize when resolution scaling is used

gdkchan:
Monster Hunter Rise has a shader that divides `gl_FragCoord` by screen size, then multiplies it by the texture size and uses the result on `texelFetch`. This doesn't work because while `gl_FragCoord` is unscaled, the `textureSize` isn't. This change also unscales the size which fixes the issue.
Before:
image
After:
image
Might also fix a similar issue on Monster Hunter Stories 2 (#2436). Needs testing.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -37,7 +37,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2441;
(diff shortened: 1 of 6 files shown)


zombie167:
This PR cannot be completed fix the problem that the image of Monster Hunter Stories 2 in the battle under high resolution.It has been improved.Unfortunately, the problem also occurs at the native resolution now.

s1
s2
s3
s4

gdkchan:
This build is outdated, this issue was already fixed on the latest one.

zombie167:
> This build is outdated, this issue was already fixed on the latest one.

Thank you very much! The problem has been fixed
s5
s6

riperiperi (approved):
Looks good here.

Thog (approved):

Linear524:
This commit broke The Touryst graphics:


Before it was fine and rendered image was like this:

PR #2452, opened 2021-07-08
Fix depop mixing

Kelebek1:
This PR does 2 things:

1. Depopping gets stuck due to rounding. `MathF.Round(a, 0);` still rounds the number up when above .5. Specifically at 13, depopping gets stuck forever. `0.962189 * 13 = 12.508457`, which rounds back to 13. Instead, take the floor result, as I think was the intention in RoundDown, and depopping can now clear itself.

2. Take the absolute of the depop value instead. Without the abs, this function always flips a negative depop into a positive, which seems wrong, as well as doing negatives minus negatives and such. I'm not 100% sure on this one, but this seems more right, so please check this part.


Thog:
Inaccurate and doesn't match hw behavior.

PR #2464, opened 2021-07-11
Fix virtual memory allocation being out of range

gdkchan:
This fixes a bug where virtual memory allocation could be out of range, as it was using the memory block address without enforcing it to be inside the requested range.

This fixes memory corruption that leads to crashes in some games that makes assumptions about the virtual memory, without first checking if it is in use. As far I can tell, only games using `MapPhysicalMemory`/`UnmapPhysicalMemory` are affected, but it could affect more too.

Fixes crash on boot on Disaster Report 4, now ingame (however suffers from typical UE4 texture corruption, so not playable).
image
Likely also affects other games, it's worth trying games that crashes very early on boot due to invalid memory region exceptions or aborts, especially if they are UE4 games.


riperiperi (approved):
Nice catch.

Thog (approved):

Benedani:
Fixes Mario Kart 8 Deluxe crashes on every 2nd boot

PR #2465, opened 2021-07-11
shadertools: Prepare for new target Languages and APIs

Thog:
This improves shader tools command line by adding support for target language and api.

### DIFF
--- a/Ryujinx.ShaderTools/Program.cs
+++ b/Ryujinx.ShaderTools/Program.cs
@@ -32,13 +32,13 @@ namespace Ryujinx.ShaderTools
             [Option("target-language", Required = false, Default = TargetLanguage.Glsl, HelpText = "Indicate the target shader language to use.")]
             public TargetLanguage TargetLanguage { get; set; }
 
-            [Option("target-api", Required = false, Default = TargetApi.OpenGL, HelpText = "Indicate the target graphic api to use.")]
+            [Option("target-api", Required = false, Default = TargetApi.OpenGL, HelpText = "Indicate the target graphics api to use.")]
             public TargetApi TargetApi { get; set; }
 
-            [Value(0, MetaName = "input", HelpText = "Input to load.", Required = true)]
+            [Value(0, MetaName = "input", HelpText = "Binary Maxwell shader input path.", Required = true)]
             public string InputPath { get; set; }
 
-            [Value(1, MetaName = "output", HelpText = "Output to store.", Required = false)]
+            [Value(1, MetaName = "output", HelpText = "Decompiled shader output path.", Required = false)]
             public string OutputPath { get; set; }
         }
 
--- a/Ryujinx.ShaderTools/Program.cs
+++ b/Ryujinx.ShaderTools/Program.cs
@@ -1,4 +1,5 @@
-﻿using Ryujinx.Graphics.Shader;
+﻿using CommandLine;
+using Ryujinx.Graphics.Shader;
 using Ryujinx.Graphics.Shader.Translation;
 using System;
 using System.IO;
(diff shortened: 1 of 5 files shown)


Ryujinx.ShaderTools/Program.cs
```diff
@@ -23,28 +24,70 @@ public GpuAccessor(byte[] data)
...
+        private class Options
         {
-            if (args.Length == 1 || args.Length == 2)
+            [Option("compute", Required = false, Default = false, HelpText = "Indicate that the shader is a compute shader.")]
+            public bool Compute { get; set; }
+
+            [Option("target-language", Required = false, Default = TargetLanguage.Glsl, HelpText = "Indicate the target shader language to use.")]
+            public TargetLanguage TargetLanguage { get; set; }
+
+            [Option("target-api", Required = false, Default = TargetApi.OpenGL, HelpText = "Indicate the target graphic api to use.")]
```
gdkchan, line 35:
```suggestion
            [Option("target-api", Required = false, Default = TargetApi.OpenGL, HelpText = "Indicate the target graphics api to use.")]
```

Ryujinx.ShaderTools/Program.cs
```diff
@@ -23,28 +24,70 @@ public GpuAccessor(byte[] data)
...
+            public TargetApi TargetApi { get; set; }
+
+            [Value(0, MetaName = "input", HelpText = "Input to load.", Required = true)]
```
gdkchan, line 38:
```suggestion
            [Value(0, MetaName = "input", HelpText = "Binary Maxwell shader input path.", Required = true)]
```

Ryujinx.ShaderTools/Program.cs
```diff
@@ -23,28 +24,70 @@ public GpuAccessor(byte[] data)
...
+            public string InputPath { get; set; }
+
+            [Value(1, MetaName = "output", HelpText = "Output to store.", Required = false)]
```
gdkchan, line 41:
```suggestion
            [Value(1, MetaName = "output", HelpText = "Decompiled shader output path.", Required = false)]
```

Ryujinx.ShaderTools/Program.cs
```diff
@@ -23,28 +24,70 @@ public GpuAccessor(byte[] data)
...
-                TranslationOptions options = new TranslationOptions(TargetLanguage.Glsl, TargetApi.OpenGL, flags);
-                string code = Translator.CreateContext(0, new GpuAccessor(data), options).Translate(out _).Code;
+            if (options.OutputPath == null)
+            {
+                if (program.BinaryCode != null)
+                {
+                    using Stream outputStream = Console.OpenStandardOutput();
 
-                Console.WriteLine(code);
+                    outputStream.Write(program.BinaryCode);
```
gdkchan, line 66: Isn't this just going to print garbage on the console?

Thog: That will write the binary. It could be nice to pipe operation like with the SPIR-V disassembler without having to rely on a file in the middle.

jduncanator: In the case where you need to write binary data to stdout, it's better to use `Console.OpenStandardOutput()` and write directly to that stream. That ensures the binary is written out untouched, Console.WriteLine has the potential to do ASCII <-> Unicode conversion depending on the current console codepage settings, and would also append a CRLF to the end of the binary.

jduncanator: Nevermind, ignore my last comment. I should have looked at the full diff 😅

gdkchan (approved):
lgtm, thanks.

jduncanator:
For what it's worth, the semi-defacto command line parser for .NET Core console applications is `System.CommandLine`, but for something that's a developer tool, it likely doesn't matter much :shrug:

gdkchan:
This is the same library used on the headless project.

AcK77 (approved):

PR #2467, opened 2021-07-11
Implement a fast path for I2M transfers

gdkchan:
This improves the performance of Inline-to-Memory engine data transfers by pusing all the data at once, instead of word by word. This is similar to the optimization that is done for constant buffer updates.

Mostly expected to affect Switch games using the OpenGL API as those use Inline-to-Memory a lot more. NVN only uses it to write the compute QMD structure afaik (which is small). Might reduce the loading times or stutters on said OpenGL games a little bit. I noticed that they have an issue where they simply pause for a few seconds sometimes though, and it doesn't seems to be a GPU related issue.


Thog (approved):

riperiperi (approved):
Looking good here.

PR #2471, opened 2021-07-12
Fix shader compilation on shaders that uses rectangle textures

### LINKED ISSUE #2469: The Tourist render to alpha bu
## Bug Report

[ If any section does not apply, replace its contents with "N/A". ]</br>
[ Lines between [ ] (square brackets) should be removed before posting. ]

[ Describe the issue in detail and what you were doing beforehand. ]</br>
[ Did you make any changes related to Ryujinx itself? ]</br>
[ If so, make sure to include details relating to what exactly you changed. ]

[ Include a detailed step by step process for recreating your issue. ]

### Log file

### Environment?

- Ryujinx version: 1.0.X</br>
[ Replace X's with the Ryujinx version at time of crash. ]
- Game version: X.X.X</br>
[ Replace X's with the game version at time of crash. ]
- System Specs:
  - OS: *(e.g. Windows 10)*
  - CPU: *(e.g. i7-6700)*
  - GPU: *(e.g. NVIDIA RTX 2070)*
  - RAM: *(e.g. 16GB)*
- Applied Mods : [ Yes (Which ones) / No ]

### Additional context?

[ Any other information relevant to your issue. ]

gdkchan:
Fixes a regression introduced on #2441 where shaders using rectangle textures wouldn't have their `textureSize` used for the coordinates normalization operation flagged and wouldn't be able to get the texture scale. While I was at it, I also fixed a issue that I noticed where the constant buffer slot wouldn't be added to the new texture operations inserted on the rewrite pass, which could cause them to fail with bindless textures.

Fixes regression on The Touryst. One thing to note is that I found high memory usage when testing this title. I'm not sure if it's a leak, but as I have a low amount of RAM, this leads to graphical glitches and crashes. However I believe this is not a new issue, someone else reported high memory usage on this game too a while ago.

Closes #2469

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -38,7 +38,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2439;
+        private const ulong ShaderCodeGenVersion = 2469;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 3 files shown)


Linear524:
Thanks !:) Looks like it works fine again in this build )
About memory leak - yeah ! It is memory leak and I think it somehow connected to dynamic resolution scaling in this title (Touryst).
When running this game on original Switch hardware I noticed very frequent jumps between screen resolution modes and shadow resolution in order to keep speed at constant 60 fps...

riperiperi (approved):

Thog (approved):

Niwu34:
Looking good
unknown

PR #2473, opened 2021-07-13
Support non-contiguous copies on I2M and DMA engines

gdkchan:
Makes more copies respect non-contiguous GPU virtual memory regions, instead of assuming it is contiguous. After this, the main thing missing will be support for non-contiguous buffers, which will be more complicated, and will be done on a separate PR.

This also fixes a bug that I noticed on I2M, where the copy would be incorrect for block linear textures, if destination X was not a multiple of 16. This was fixed by aligning the start of the vector copy, which should start and end of a multiple of 16. I'm not aware of any games that uses it with a X value that is not a multiple of 16 though (it is usually 0), so maybe nothing was affected by this bug.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -152,14 +152,11 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                     dst.MemoryLayout.UnpackGobBlocksInZ(),
                     dstBpp);
 
-                ulong srcBaseAddress = memoryManager.Translate(srcGpuVa);
-                ulong dstBaseAddress = memoryManager.Translate(dstGpuVa);
-
                 (int srcBaseOffset, int srcSize) = srcCalculator.GetRectangleRange(src.RegionX, src.RegionY, xCount, yCount);
                 (int dstBaseOffset, int dstSize) = dstCalculator.GetRectangleRange(dst.RegionX, dst.RegionY, xCount, yCount);
 
-                ReadOnlySpan<byte> srcSpan = memoryManager.Physical.GetSpan(srcBaseAddress + (ulong)srcBaseOffset, srcSize, true);
-                Span<byte> dstSpan = memoryManager.Physical.GetSpan(dstBaseAddress + (ulong)dstBaseOffset, dstSize).ToArray();
+                ReadOnlySpan<byte> srcSpan = memoryManager.GetSpan(srcGpuVa + (ulong)srcBaseOffset, srcSize, true);
+                Span<byte> dstSpan = memoryManager.GetSpan(dstGpuVa + (ulong)dstBaseOffset, dstSize).ToArray();
 
                 bool completeSource = IsTextureCopyComplete(src, srcLinear, srcBpp, srcStride, xCount, yCount);
                 bool completeDest = IsTextureCopyComplete(dst, dstLinear, dstBpp, dstStride, xCount, yCount);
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
```diff
@@ -189,36 +190,43 @@ private void FinishTransfer()
...
 
-                        Span<byte> pixel = data.Slice(srcOffset, 16);
+                    for (; x < x2Trunc; x += 16, srcOffset += 16)
+                    {
+                        int dstOffset = dstCalculator.GetOffset(x, y);
 
-                        _channel.MemoryManager.Physical.Write(dstAddress, pixel);
+                        ulong dstAddress = _dstGpuVa + (uint)dstOffset;
+
+                        memoryManager.Write(dstAddress, MemoryMarshal.Cast<byte, Vector128<byte>>(data.Slice(srcOffset, 16))[0]);
```
Thog, line 220: TIL that is possible with a Vector128

Thog (approved):

AcK77 (approved):

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -152,14 +152,11 @@ private void LaunchDma(int argument)
...
-                ulong srcBaseAddress = memoryManager.Translate(srcGpuVa);
-                ulong dstBaseAddress = memoryManager.Translate(dstGpuVa);
-
                 (int srcBaseOffset, int srcSize) = srcCalculator.GetRectangleRange(src.RegionX, src.RegionY, xCount, yCount);
                 (int dstBaseOffset, int dstSize) = dstCalculator.GetRectangleRange(dst.RegionX, dst.RegionY, xCount, yCount);
 
-                ReadOnlySpan<byte> srcSpan = memoryManager.Physical.GetSpan(srcBaseAddress + (ulong)srcBaseOffset, srcSize, true);
-                Span<byte> dstSpan = memoryManager.Physical.GetSpan(dstBaseAddress + (ulong)dstBaseOffset, dstSize).ToArray();
+                ReadOnlySpan<byte> srcSpan = memoryManager.GetSpan(srcGpuVa + (uint)srcBaseOffset, srcSize, true);
+                Span<byte> dstSpan = memoryManager.GetSpan(dstGpuVa + (uint)dstBaseOffset, dstSize).ToArray();
```
FICTURE7, line 159: Could this `dstSpan` come from a `WritableRegion` instead? This was something I noticed while taking a stab at adding memory pooling FICTURE7@f45899e3dd2dc533aa. These allocations can get large enough to end up in the large object heap, and when there are many of them being allocated the GC will do a gen2 collection which can be expensive.

gdkchan: Yes, but the write tracking becomes a bit weird. Well, right now the `WritableRegion` is not write tracked at all (it does not call `SignalMemoryTracking` when the data is written back or is done with). For this to work it would need to be changed to signal the region for modification on dispose. But there is another problem, if the `dstSpan` is only partially written (should be possible if the dst X/Y are non-zero), then it would need to trigger flush on read aswell (although, it doesn't do that right now, so to simply match the current behavior, just signaling it on write-back/dispose is enough).

FICTURE7: I see, thanks for the explanation. 👍

PR #2477, opened 2021-07-14
Close transfer memory properly on nvservices

gdkchan:
Fixes an oversight where it was getting the transfer memory handle from the wrong location, leading to only the process handle being closed.

Fixes crash with some homebrew that initializes nvservices more than once, I'm not sure if games can do this, and I'm not aware of any game being affected by this change.


Thog (approved):
nice catch

AcK77 (approved):

Issue #2454, opened 2021-07-08
Dragon Ball FighterZ - Red Screen / Flickering


codbugpatrol:
### What's the issue you encountered?
While on character selection screen and on fight loading, the game do not render the background correctly, flicking the correct image or with an static red screen. (**This part is fixed** (PR #2481) : Also during the game play, after a special attack, sometimes the white flickering starts and visual glitches begin.)

![DragonBallFighterZ_RedBackground_Flickering](https://user-images.githubusercontent.com/85196112/124982275-e5872200-e00c-11eb-8579-623c80996417.jpg)
![DragonBallFighterZ_RedBackground_Flickering_2](https://user-images.githubusercontent.com/85196112/124982279-e750e580-e00c-11eb-9570-826a0b672498.jpg)
![DragonBallFighterZ_RedBackground_Flickering_3](https://user-images.githubusercontent.com/85196112/124982286-e8821280-e00c-11eb-8968-2329690470ca.jpg)
![DragonBallFighterZ_RedBackground_Flickering_4](https://user-images.githubusercontent.com/85196112/124982295-e9b33f80-e00c-11eb-8641-1b5576ab36e7.jpg)

**This part is fixed** (PR #2481):
Visual Glitches after a special attack (it seems related to the above issue, as starts with flickering):
TEST WITH VEGITO (Potara Version), because it also, after any special, cause irreparable visual glitches in the round.
![image](https://user-images.githubusercontent.com/85196112/125681980-67f59e90-a172-4606-9ecc-c81fdf3f2367.png)

### How can the issue be reproduced?
Go on character selection screen in any mode. 
Select any characters you like and load the fight.

### Log file

00:01:01.014 |E| GUI.RenderLoop Gpu GLERROR: DebugSeverityHigh: GL_INVALID_VALUE error generated. <offset> is negative or <size> is less than or equal to zero or <offset> is not an integer multiple of TEXTURE_BUFFER_OFFSET_ALIGNMENT.
CallStack=   at System.Environment.get_StackTrace()
   at Ryujinx.Graphics.OpenGL.Debugger.GLDebugHandler(DebugSource source, DebugType type, Int32 id, DebugSeverity severity, Int32 length, IntPtr message, IntPtr userParam)
   at Ryujinx.Graphics.OpenGL.Image.TextureBuffer.SetStorage(BufferRange buffer)
   at Ryujinx.Graphics.Gpu.Memory.BufferManager.CommitBufferTextureBindings()
[...]

PR #2478, opened 2021-07-14
Ensure that DMA copy target textures are kept alive or flushed

gdkchan:
There is a bug where textures that are modified on the fast path of DMA copies might suffer from data loss, if the texture is deleted, because the data is never flushed. The fix here is adding the texture to the "auto delete cache", so that the data is eventually flushed when it falls out of the cache, and the texture is kept alive for longer, so even if it is removed from the texture pool, chances are that it will still exist on the next use (thus not even needing to flush and reload the data).

This fixes black portraits on Legend of Mana, might also fix the other graphical glitch that was reported.
Before:
image
After:
image
Needs more testing to confirm that all instances of the bug where fixed.

It is worth nothing that images might have a similar issue. If a image is modified by a shader, but not accessed by the CPU, it will eventually have its data lost forever if it is removed from the pool. This change does not address this issue.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -816,6 +816,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 if (match)
                 {
+                    _cache.Lift(texture);
                     return texture;
                 }
             }


AcK77 (approved):

Thog (approved):

riperiperi:
Hmm, this could be an issue in general. I didn't realize that only textures that are bound as render targets get added to the `AutoDeleteCache`, this could also result in data lost from image stores and texture copy. It's worth investigating for other games.

riperiperi (approved):

codbugpatrol:
@gdkchan, I normally test every PR to see if the by any change, fixes the https://github.com/Ryujinx/Ryujinx/issues/2454. (I love this game a lot). But this time I've noted something different. This PR may be fixing something in that.
During testing, when I use a special attack, somethings brakes, the screen start to blink in white for a while. After that something brakes.
This PR seems to be fixing, or it is just a coincidence a little bit of the image, not fixing at all, but at least part, that's good!

(This problem is random, sometimes occurs, sometimes not).

In the mainline before Special:
db_ryujinx_before

In the mainline after Special:
db_ryujinx_after

In this PR before Special:
db_ryujinx_before_gdk

In this PR after Special:
db_ryujinx_after_gdk

gdkchan:
> 
> 
> Hmm, this could be an issue in general. I didn't realize that only textures that are bound as render targets get added to the `AutoDeleteCache` - this could also result in data lost from image stores and 2d blit. It's worth investigating for other games.

2D blit is added to the `AutoDeleteCache` right now https://github.com/Ryujinx/Ryujinx/blob/208ba1dde2b9a4d31446ace2bba8f0d641d2e300/Ryujinx.Graphics.Gpu/Image/TextureCache.cs#L444
But you are right about images (as I noted on the PR description). I'm not sure if they should be added to the `AutoDeleteCache` too, I feel that images will need a different solution.

PR #2481, opened 2021-07-16
Flush buffers and texture data through a persistent mapped buffer.

riperiperi:
This change gets data for buffers and textures by first copying it to a mapped persistent buffer, waiting on a fence, and copying the data out of it. This seems to be magnitudes faster than downloading the data directly with the gl methods, so it should improve games and situations where data is being flushed back from the GPU a lot.

There is one buffer used per opengl context for flushing data. For one instance of the Renderer, there is a `_main` and a `_background` buffer, so that the main rendering thread and the background context can both flush at the same time without stepping on each other's toes. Each buffer sizes up to match the biggest thing it has flushed so far, similar to the PBO copy buffer for textures. One benefit of doing it this way is that 4 instances of the same 1080p texture will use the _same_ 8MB buffer to flush out the data, rather than one for each texture or setting it up at the time.

Affected Games:
- Skyward Sword HD is greatly improved.
- Xenoblade 2 (and likely DE, to some extent).
- Pokemon Sword is much closer to its performance in LDN build.
- I am not sure how Monster Hunter Rise is affected, but it will be. It probably will not be as fast as LDN, but those changes can apply on top of this just fine.
- Small improvement to any game that uses buffer flush.

Additional changes:
- Linear to LinearStrided conversion no longer copies the data when it has the same stride.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/PersistentBuffers.cs
+++ b/Ryujinx.Graphics.OpenGL/PersistentBuffers.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using OpenTK.Graphics.OpenGL;
+using Ryujinx.Graphics.GAL;
+using Ryujinx.Graphics.OpenGL.Image;
+
+namespace Ryujinx.Graphics.OpenGL
+{
+    class PersistentBuffers : IDisposable
+    {
+        private PersistentBuffer _main = new PersistentBuffer();
+        private PersistentBuffer _background = new PersistentBuffer();
+
+        public PersistentBuffer Default => BackgroundContextWorker.InBackground ? _background : _main;
+
+        public void Dispose()
+        {
+            _main?.Dispose();
+            _background?.Dispose();
+        }
+    }
+
+    class PersistentBuffer : IDisposable
+    {
+        private IntPtr _bufferMap;
+        private int _copyBufferHandle;
+        private int _copyBufferSize;
+
+        private void EnsureBuffer(int requiredSize)
+        {
+            if (_copyBufferSize < requiredSize && _copyBufferHandle != 0)
+            {
+                GL.DeleteBuffer(_copyBufferHandle);
+
+                _copyBufferHandle = 0;
+            }
+
+            if (_copyBufferHandle == 0)
+            {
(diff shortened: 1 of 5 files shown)


codbugpatrol:
This solves part of the https://github.com/Ryujinx/Ryujinx/issues/2454. Explaning:
Fixes: 
- Visual glitches after special attacks (e.g. Vegito Potara).
Do not fix: 
- Character menu selection, but it make is even more smooth, fast.

riperiperi:
This should probably be tested more carefully on AMD GPUs, as it's possible that something could break there due to this being a newer OGL feature.

Ryujinx.Graphics.OpenGL/PersistentBuffers.cs
```diff
@@ -0,0 +1,113 @@
...
+            byte[] data = new byte[size];
+
+            view.WriteToPbo(0, false);
+
+            GL.BindBuffer(BufferTarget.PixelPackBuffer, 0);
+
+            GL.MemoryBarrier(MemoryBarrierFlags.ClientMappedBufferBarrierBit);
+
+            IntPtr sync = GL.FenceSync(SyncCondition.SyncGpuCommandsComplete, WaitSyncFlags.None);
+            GL.ClientWaitSync(sync, ClientWaitSyncFlags.SyncFlushCommandsBit, 1000000000);
```
gdkchan, line 72: Shouldn't this log an error on timeout? Since the data will probably be incomplete.

riperiperi: Addressed, moved this to a helper method since it's duplicated.

gdkchan (approved):
lgtm, thanks! I just tried SSHD and I'm getting 60 fps here, its the first time I try the game though, so I'm not sure how much I will get on master. Tested Catherine too since it seems to flush buffers somewhat frequently, it doesn't seems any better or worse.

Points of improvement for the future:
- Make the `GetData` methods take a `Span<byte>`, that will allow them to copy to guest memory directly (the caller would need to use `GetWritableRegion` and pass its span).
- Use a similar approach for `glTexSubImage` and friends, this will likely speedup texture uploads.

Thog (approved):

PR #2489, opened 2021-07-18
Only use persistent buffers to flush on NVIDIA and Windows+AMD

riperiperi:
It seems like this method of flushing data is much slower on Mesa drivers, and slightly slower on Intel Windows. Have not tested Intel Mesa, but I'm assuming it is the same as AMD. The game in question is Skyward Sword HD, as it had the greatest effect from the last PR.

This also adds vendor detection for AMD on Unix, which counted as "Unknown" before.


TacoCake:
I've tested this (on linux with AMD cpu and gpu) and my FPS went from 2-6 fps to well over 50+ fps.

TacoCake (approved):

Thog (approved):

gdkchan (approved):
lgtm, thanks.

PR #2493, opened 2021-07-18
Clamp controller sticks to circle, instead of square

emmauss:
Xinput devices provides axis data with square ranges, ie +-32767. Moving the sticks to diagonal ends means that it will have the max range values for both axes. This causes fast motion of player characters in games that do not clamp their input. 

This fixes it by clamping the input to a circle. Fixes Link sprinting without using up stamina when the controller sticks are at diagonals

### DIFF
--- a/Ryujinx.Input/HLE/NpadController.cs
+++ b/Ryujinx.Input/HLE/NpadController.cs
@@ -3,6 +3,7 @@ using Ryujinx.Common.Configuration.Hid;
 using Ryujinx.Common.Configuration.Hid.Controller;
 using Ryujinx.Common.Configuration.Hid.Controller.Motion;
 using Ryujinx.HLE.HOS.Services.Hid;
+using SixLabors.ImageSharp;
 using System;
 using System.Numerics;
 using System.Runtime.CompilerServices;
@@ -331,28 +332,6 @@ namespace Ryujinx.Input.HLE
             }
         }
 
-        private static short ClampAxis(float value)
-        {
-            if (value <= -short.MaxValue)
-            {
-                return -short.MaxValue;
-            }
-            else
-            {
-                return (short)(value * short.MaxValue);
-            }
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
-        {
-            return new JoystickPosition
-            {
-                Dx = ClampAxis(MathF.Abs(x) > deadzone ? x : 0.0f),
-                Dy = ClampAxis(MathF.Abs(y) > deadzone ? y : 0.0f)
-            };
-        }
-
         public GamepadInput GetHLEInputState()
         {
             GamepadInput state = new GamepadInput();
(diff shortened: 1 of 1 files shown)


Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -3,6 +3,7 @@
 using Ryujinx.Common.Configuration.Hid.Controller;
 using Ryujinx.Common.Configuration.Hid.Controller.Motion;
 using Ryujinx.HLE.HOS.Services.Hid;
+using SixLabors.ImageSharp;
```
AcK77, line 6: Is it really needed? If yes, could it not be replaced with something else?

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -385,16 +364,57 @@ public GamepadInput GetHLEInputState()
             }
             else if (_config is StandardControllerInputConfig controllerConfig)
             {
-                (float leftAxisX, float leftAxisY) = State.GetStick(StickInputId.Left);
+                (float leftAxisX,  float leftAxisY)  = State.GetStick(StickInputId.Left);
                 (float rightAxisX, float rightAxisY) = State.GetStick(StickInputId.Right);
 
-                state.LStick = ApplyDeadzone(leftAxisX, leftAxisY, controllerConfig.DeadzoneLeft);
-                state.RStick = ApplyDeadzone(rightAxisX, rightAxisY, controllerConfig.DeadzoneRight);
+                state.LStick = ClampToCircle(ApplyDeadzone(leftAxisX, leftAxisY,  controllerConfig.DeadzoneLeft));
```
AcK77, line 370:
```suggestion
                state.LStick = ClampToCircle(ApplyDeadzone(leftAxisX, leftAxisY, controllerConfig.DeadzoneLeft));
```

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -385,16 +364,57 @@ public GamepadInput GetHLEInputState()
...
+            if (value <= -short.MaxValue)
+            {
+                return -short.MaxValue;
+            }
+            else
+            {
+                return (short)(value * short.MaxValue);
+            }
+        }
+
```
AcK77, line 399: Remove extra empty line

AcK77 (changes requested):

Ryujinx.Input/HLE/NpadController.cs
emmauss: addressed

Ryujinx.Input/HLE/NpadController.cs
emmauss: addressed

Ryujinx.Input/HLE/NpadController.cs
emmauss: I don't know how that got there. removed

AcK77 (approved):
Lgtm! Thanks!

MelonSpeedruns:
LGTM, tested with Skyward Sword HD, and no super-walking when the joystick is diagonal!

Thog (approved):

PR #2494, opened 2021-07-18
Return mapped buffer pointer directly for flush, WriteableRegion for textures

riperiperi:
A few changes here to generally improve performance, even for platforms not using the persistent buffer flush.

- Texture and buffer flush now return a ReadOnlySpan<byte>. It's guaranteed that this span is pinned in memory, but it will be overwritten on the next flush from that thread, so it is expected that the data is used before calling again.
- As a result, persistent mappings no longer copy to a new array - rather the persistent map is returned directly as a Span<>. A similar host array is used for the glGet flushes instead of allocating new arrays each time.
- Texture flushes now do their layout conversion into a WriteableRegion when the texture is not MultiRange, which allows the flush to happen directly into guest memory rather than into a temporary span, then copied over. This avoids another copy when doing layout conversion.
- The WritableRegion can now be created with or without triggering tracking. The `IWritableBlock` interface has a method implementation for an untracked write that will just call the tracked write if no unique implementation is provided.

Overall, this saves 1 data copy for buffer flush, 1 copy for linear textures with matching source/target stride, and 2 copies for block textures or linear textures with mismatching strides.

On platforms that don't flush data with the Persistent Buffers, this will still save on layout conversion, and there will be no allocations for buffer/texture flush. Everyone's a winner.

# Comparisons
I totally forgot to do these the last time, so let's get a running comparison of how some affected games are doing:

### Master 1.0.6969 (unfortunately Not Nice as it was before all the persistent buffer stuff)
image
### Master (current)
image
### This PR
image

### Master 1.0.6969
image
### Master (current)
image
### This PR
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -878,11 +878,10 @@ namespace Ryujinx.Graphics.Gpu.Image
             {
                 MemoryRange subrange = Range.GetSubRange(0);
 
-                WritableRegion region = _physicalMemory.GetWritableRegion(subrange.Address, (int)subrange.Size, tracked);
-
-                GetTextureDataFromGpu(region.Memory.Span, tracked, texture);
-
-                region.Dispose();
+                using (WritableRegion region = _physicalMemory.GetWritableRegion(subrange.Address, (int)subrange.Size, tracked))
+                {
+                    GetTextureDataFromGpu(region.Memory.Span, tracked, texture);
+                }
             }
             else
             {
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -4,6 +4,7 @@ using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Memory;
 using Ryujinx.Graphics.Texture;
 using Ryujinx.Graphics.Texture.Astc;
+using Ryujinx.Memory;
 using Ryujinx.Memory.Range;
 using System;
 using System.Collections.Generic;
@@ -821,14 +822,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 return; // Flushing this format is not supported, as it may have been converted to another host format.
             }
 
-            if (tracked)
-            {
-                _physicalMemory.Write(Range, GetTextureDataFromGpu(tracked));
-            }
-            else
-            {
-                _physicalMemory.WriteUntracked(Range, GetTextureDataFromGpu(tracked));
-            }
+            FlushTextureDataToGuest(tracked);
         }
 
         /// <summary>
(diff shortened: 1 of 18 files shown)


Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -864,19 +858,57 @@ public void ExternalFlush(ulong address, ulong size)
...
+        /// </remarks>
+        /// <param name="tracked">True if writing the texture data is tracked, false otherwise</param>
+        /// <param name="texture">The specific host texture to flush. Defaults to this texture</param>
+        private void FlushTextureDataToGuest(bool tracked, ITexture texture = null)
+        {
+            if (Range.Count == 1)
+            {
+                MemoryRange subrange = Range.GetSubRange(0);
+
+                WritableRegion region = _physicalMemory.GetWritableRegion(subrange.Address, (int)subrange.Size, tracked);
```
gdkchan, line 881: Why not `using WritableRegion region` without the explicit dispose.

Ryujinx.Graphics.Texture/LayoutConverter.cs
```diff
@@ -359,6 +359,7 @@ public static class LayoutConverter
         }
 
         public static ReadOnlySpan<byte> ConvertLinearToBlockLinear(
```
gdkchan, line 361: Can the return be removed on those methods now that its possible to specify the output storage?

gdkchan: I guess the callers would need to be changed to allocate themselves rather than passing an empty span, but imo its better than making the function either use the supplied storage or allocate its own if empty.

gdkchan (commented):
> Texture and buffer flush now return a ReadOnlySpan. It's guaranteed that this span is pinned in memory, but it will be overwritten on the next flush from that thread, so it is expected that the data is used before calling again.

The restriction is a bit unfortunate, it doesn't seems that unlikely that someone would inadvertently try to call the methods for 2 differents textures sequentially. It is not a problem right now though, as it is always called, then written back and done with.

> The WritableRegion can now be created with or without triggering tracking. The IWritableBlock interface has a method implementation for an untracked write that will just call the tracked write if no unique implementation is provided.

Cool, I actually commented briefly about this on https://github.com/Ryujinx/Ryujinx/pull/2473#discussion_r671696879
The copies on DMA could use it, instead of the span that is copied into a new array and then written back. Eventually, I2M should be able to use it too. I believe the `SignalMemoryTracking` call when the `WritableRegion` is created is not necessary if the span is fully written (that is, every single byte is overwritten without caring about the current value), it is only needed if the span is partially written. Whenever or not doing this little optimization depends on the `SignalMemoryTracking` cost though. If it's cheap, I guess its not worth the hassle.

I think VIC copies could use that too, I will give it a try after this is merged.

Ryujinx.Graphics.Texture/LayoutConverter.cs
riperiperi: There is a small problem with that and the linear -> linear strided conversion, where it does not allocate storage but instead uses the input storage when no conversion is needed. The caller would need to check this case.

gdkchan: I think its better to check this case on the caller anyway, since you can avoid doing the call.

Thog (approved):

gdkchan (approved):
lgtm, thanks.

PR #2496, opened 2021-07-20
Replace BGRA and scale uniforms with a uniform block

gdkchan:
Currently we have some "loose" uniforms on shaders that are being used for the texture scales, and the value that indicates if a texture is BGRA or not. They require a `glGetUniformLocation` to get the location the compiler assigned to them, and also requires setting the new uniform data every time the shader change. The motivator for this change though, was the fact those uniform variables are not support on SPIR-V, and I believe its better to handle things in a similar way between the backends. Plus, if we want to produce SPIR-V from GLSL, this change is necessary (otherwise it will fail to compile).

With the new approach, a single buffer that can fit everything (the render scales, IsBGRA, and a new alpha test field) is created. Data is written to the buffer using the current `SetBufferData` methods. The shader now reserves the binding 0 for this new uniform block, which allows the buffer is only bound once (right after it is created) since the binding is the same across all shaders. The `SetUniformBuffers` method previously assumed the binding always started at 0. This is no longer the case due to it reserving binding 0, so the method was changed to take a `first` parameter that tells the first binding number. For consistency, `SetStorageBuffers` was also changed. On the GPU emulator, the buffer manager no longer assumes that the bindings are in sequence, it can now handle "gaps" on the bindings, by making one backend `Set*Buffers` call per sequential range.

On compute shaders, the uniform block layout was made to match the fragment one. So, since it doesn't need the alpha test and IsBGRA fields, it simply inserts an array at the bedding in place of the fields that are not needed. That way, the scales falls at the same offset as the fragment shader block. This makes things simpler as we don't need to know the shader type to set this data.

The alpha test field was added in preparation for a future change were I will add alpha test emulation code at the end of the fragment shaders, which is required on OpenGL if the GPU/driver doesn't support it (it's a deprecated feature). On Vulkan, it is also needed as there's no support for alpha test at all there.

On the backend, some `int` and `float` arrays had to be changed to `Vector4<int>` and `Vector4<float>`, to match the uniform buffer layout. On OpenGL, uniforms have its elements aligned to 16 bytes by default, even array elements are aligned to 16, so this changes makes the C# array match the OpenGL uniform layout.

**What to test:**
In general, I recommend testing a few games to ensure they didn't regress, including making sure that resolution scale still works as before, and also testing games using BGRA textures, and ensure they also work as before.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -80,11 +80,17 @@ namespace Ryujinx.Graphics.OpenGL
             for (int index = 0; index < _fpRenderScale.Length; index++)
             {
                 _fpRenderScale[index].X = 1f;
+                _fpRenderScale[index].Y = 0f;
+                _fpRenderScale[index].Z = 0f;
+                _fpRenderScale[index].W = 0f;
             }
 
             for (int index = 0; index < _cpRenderScale.Length; index++)
             {
                 _cpRenderScale[index].X = 1f;
+                _cpRenderScale[index].Y = 0f;
+                _cpRenderScale[index].Z = 0f;
+                _cpRenderScale[index].W = 0f;
             }
 
             _tfbs = new BufferHandle[Constants.MaxTransformFeedbackBuffers];
@@ -1192,9 +1198,9 @@ namespace Ryujinx.Graphics.OpenGL
         {
             static void Copy(float[] from, int fromIndex, Vector4<float>[] to, int toIndex, int count)
             {
-                for (int i = 0; i < count; i++)
+                for (int index = 0; index < count; index++)
                 {
-                    to[toIndex + i].X = from[fromIndex + i];
+                    to[toIndex + index].X = from[fromIndex + index];
                 }
             }
 
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -42,8 +42,7 @@ namespace Ryujinx.Graphics.OpenGL
         }
 
         private Vector4<int>[] _fpIsBgra = new Vector4<int>[SupportBuffer.FragmentIsBgraCount];
-        private Vector4<float>[] _fpRenderScale = new Vector4<float>[65];
-        private Vector4<float>[] _cpRenderScale = new Vector4<float>[64];
+        private Vector4<float>[] _renderScale = new Vector4<float>[65];
 
         private TextureBase _unit0Texture;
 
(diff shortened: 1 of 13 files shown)


Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -823,9 +834,6 @@ public void SetProgram(IProgram program)
             {
                 _program.Bind();
             }
-
-            UpdateFpIsBgra();
```
gdkchan, line 827: This shouldn't be needed now that all shaders are using the same binding/buffer, but would be nice to get a confirmation.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -31,9 +33,17 @@ class Pipeline : IPipeline, IDisposable
...
+        private struct Vector4<T>
+        {
+            public T X;
+            public T Y;
+            public T Z;
+            public T W;
+        }
+
+        private Vector4<int>[] _fpIsBgra = new Vector4<int>[SupportBuffer.FragmentIsBgraCount];
+        private Vector4<float>[] _fpRenderScale = new Vector4<float>[65];
```
gdkchan, line 45: Maybe those can be combined now? I'm not sure why it was using separate arrays and uniforms for compute and fragment scales in the first place, since the two can never execute together.

Btw, the size is a bit higher than needed, I'm not sure if that is intentional, but the image limit is 8, so the total is 1 + 32 + 8 = 41. However there are games that currently get past the 32 samplers limit, so I guess that "fixing" this would make them throw IOR exception, so I'm going to leave it as-is.

ActualMandM:
Can confirm that Project DIVA MEGA39's BGRA still works fine
2021-07-21_17-36-42_Ryujinx

Thog (approved):

Ryujinx.Graphics.OpenGL/Pipeline.cs
riperiperi: I think these should be combined now. I'd leave the size as is for now, but yeah that was a mistake originally.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1179,37 +1183,50 @@ private void EnsureFramebuffer()
...
+                }
+
+                return changed;
+            }
+
+            switch (stage)
             {
-                GL.Uniform1(_program.FragmentIsBgraUniform, 8, _fpIsBgra);
+                case ShaderStage.Fragment:
+                    if (Copy(scales, 0, _fpRenderScale, 1, textureCount + imageCount))
```
riperiperi, line 1207: This might break when switching between fragment and compute. Combining the arrays will fix it.

riperiperi (approved):
Looking good, nice one! I think fp and cp scale arrays should be combined, otherwise the current change detection might fail to update the scales to the appropriate array.

Ryujinx.Graphics.OpenGL/Pipeline.cs
gdkchan: Done.

riperiperi (approved):
LGTM. I'm a little worried that updating the res scale buffer now will upload 4x the number of bytes, but I don't think it seems to be a problem at the moment.

PR #2501, opened 2021-07-24
Add a Multithreading layer for the GAL, multi-thread shader compilation at runtime

riperiperi:
_Note - Draft since the PR needs testing, and also I need to do another reading pass over it._

This PR adds a custom threading layer between the FIFO processing thread and the backend. This replaces all backend methods with a quick dispatch that sends a command to a queue to be processed in another thread. This already exists in some drivers, but by implementing it ourselves we gain more consistency and some... additional benefits.

The goal is not _necessarily_ better peak performance, but rather **Multithreaded Shader Compilation at runtime**, reduction of some rather nasty stuttering from NVIDIA's own multithreading, and taking control of our own destiny. With some drivers, and future backends, peak performance improvements are achieved too.

This can be enabled or disabled via an option in the settings called **"Backend Threading"**. The default option is "Auto", which always means on right now, but could change for some vendors or backends.

Also, due to related changes, the initial view in **Super Mario Odyssey's Metro Kingdom** has went from 50-51 fps to 57-58 fps for me. With our threading off, it is now 55.

This feature is very complicated and should be tested on most games. By design it should not cause any issues, so if it does please report them so that they can be fixed!

# Multithreaded Shader Compilation
One of the main benefits of doing our own multithreading is being able to continue throwing draws at the backend without waiting for programs to compile, which allows us to begin multiple shader compilations in parallel at runtime without skipping draws.

**_If you've been using our shader cache for a while, you might have forgotten how bad the first run actually is. Please take a look at the videos for a refresher._**

This does not eliminate stuttering, due to GLSL compilation being very slow, but it does reduce it significantly when multiple new shaders appear in the same frame (4-5x in best case scenarios).

Notably, this improvement could extend to SPIR-V (vulkan) and potentially GP5/GLASM (ARB, throw as many abbreviations as you want at it) shaders with little additional effort, reducing their low compilation times even further.

Here are some examples of popular games with an empty shader cache (but with PPTC enabled, to remove its influence on results). I've also timed a few key stutters to show where the improvement lies.

## Splatoon 2
This game essentially stops at the moment the camera pans down to the plaza. VERY long pauses when loading game levels or effects ingame result in disconnections with LDN, which will happen _much less_ as a result of the magnitudes faster shader compilation.

### Master
https://youtu.be/WbWjnl4_ebU

### threaded-gal
https://youtu.be/GROH9XhgX0I

#### Loading Screen Pause until the title screen fades in
- Master: 27.333 - 2:32.533 = 125.2 seconds
- GAL: 24.183 - 58.733 = 34.5 seconds **(3.63x)**

#### L+R until player controllable
- Master: 2:37.033 - 4:53.000 = 135.967 seconds
- GAL: 1:03.566 - 1:34.050 = 30.484 seconds **(4.46x)**

#### First step
- Master: 4:53.000 - 4:53.583 = 0.583 seconds
- GAL: 1:34.050 - 1:34.400 = 0.350 seconds **(1.66x)**

#### Loading single player hub
- Master: 5:15.583 - 6:17.333 = 61.75 seconds
- GAL: 1:52.800 - 2:09.166 = 16.366 seconds **(3.77x)**

#### Shooting ink for the first time
- Master: 6:30.100 begin - 6:30.350 arm raise - 6:30.733 bullet appears - 6:32.366 gun particle = 2.266 seconds
- Master: 6:32.433 splash 1 - 6:33.500 = 1.067 seconds

- GAL: 2:16.733 begin - 2:16.800 arm raise - 2:17.216 bullet appears - 2:17.600 extra - 2:18.083 gun particle = 1.35 seconds **(1.68x)**
- GAL: 2:18.166 splash 1 - 2:18.700 = 0.534 seconds **(2.00x)**

## Monster Hunter Rise Demo
This game has an incredibly long first draw, resulting in the _entire title screen song_ playing before it draws any frames, which could make people think the emulator was entirely broken. A similar pause happens when loading levels for the first time, or attack effects, which tends to cause disconnections on LDN. These should be improved.

### Master
https://youtu.be/I6RBCrfinJY

### threaded-gal
https://youtu.be/B42gZNlToeU

#### Intro pause
- Master: 33.650 - 1:33.266 = 59.616 seconds
- GAL: 28.350 - 41.416 = 13.066 seconds **(4.56x)**

#### Hunter preview
- Master: 2:10.450 - 2:30.466 = 20.016 seconds
- GAL: 54.066 - 1:00.033 = 5.967 seconds **(3.35x)**

#### Quest initial black screen
- Master: 2:44.016 - 3:23.900 = 39.884 seconds
- GAL: 1:13.050 - 1:26.150 = 13.1 seconds **(3.04x)**

#### Bird pauses
These stutters tend to be ASTC texture decoding related, but they are still impacted by GAL threading slightly.
**Master:**
3:24.483 - 3:26.566 = 2.083 seconds
3:26.916 - 3:28.066 = 1.15 seconds
3:28.116 - 3:29.650 = 1.534 seconds

**GAL:**
1:26.766 - 1:27.850 = 1.084 seconds **(1.92x)**
1:28.100 - 1:29.233 = 1.133 seconds **(1.02x)**
1:29.266 - 1:30.566 = 1.3 seconds **(1.18x)**

## Mario Kart 8 Deluxe
This game's shader compile stutters are greatly reduced, in the menu and at the start of the race. The wait at the end of the loading screen is _much_ shorter.

### Master
https://youtu.be/yN6IXA6HFqI

### threaded-gal
https://youtu.be/ZD_wcFhRIUc

#### Logo to Menu
- Master: 5.300 - 8.966 = 3.666 seconds
- GAL: 5.166 - 6.333 = 1.167 seconds **(3.14x)**

#### L+R
- Master: 12.850 - 13.633 = 0.783 seconds
- GAL: 9.566 - 9.800 = 0.234 seconds **(3.35x)**

#### Character Stutters
- Master: 19.716 - 20.816 = 1.1 seconds
- GAL: 15.666 - 16.200 = 0.534 seconds **(2.00x)**

#### Loading Screen End
- Master: 36.300 - 53.800 = 17.5 seconds
- GAL: 30.933 - 36.766 = 5.833 seconds **(3.00x)**

#### GO!
- Master: 1:08.983 - 1:12.083 = 3.1 seconds
- GAL: 50.200 - 51.300 = 1.1 seconds **(2.82x)**

# NV Threaded Stutter Fixed
The NVIDIA threaded optimization on windows is very fast, but it has one fatal flaw that impacts some games more than others. Whenever a thread is created or destroyed, the driver appears to snapshot processes in the system, causing a 50-100ms stutter.

Every game does it at some point - some once a minute, some more frequently. When a game does this, you'll notice. Here are some videos of the worst offenders:

## Pokemon Sword
The game constantly stutters for the entire runtime. The stutter is very noticeable when vsync is off and running at a high framerate, but in normal gameplay tends to result in a framerate of about 29 instead of 30. This stutter is entirely fixed by our custom backend threading.

### Master
https://youtu.be/Y02xV9hgBs0

### threaded-gal
https://youtu.be/vPqO5nrqymk

## Super Mario Odyssey (Kingdom Selection)
When selecting a kingdom, a number of threads are spun up to handle the decoding of the video showing a preview of the kingdom. This causes a very noticeable hitch when scrolling through kingdoms. This stutter is entirely fixed by our custom backend threading.

### Master
https://youtu.be/bQWoFBCJ3t8

### threaded-gal
https://youtu.be/8hC5rL1sPz4

# Performance
Since the goal of this PR is to replace the threaded optimization in the driver, it will inevitably have an effect on performance. This table is not complete, as I do not have access to certain platforms to test them.

An "Auto" setting exists so that we can adjust which platforms our threading is enabled on by default, should any of them be considerably slower. Right now, "Auto" means "On" for all platforms, though we don't know yet how mesa will behave, so that may change. Linux users, please assemble.

**Note that when running our threaded optimization, you MUST disable the threading in your driver. We do this for NVIDIA windows automatically, but for mesa you must set `mesa_glthread` to `false`, on linux nvidia you must set `__GL_THREADED_OPTIMIZATIONS` to 0. "Game Mode" stuff on linux tends to set these to true, so watch out for that.**

| Platform  | Intel | Nvidia | AMD |
| --------- | ----- | ------ | --- |
| Windows | _Very_ Slightly Faster* | Similar** | **Much Faster** |
| Linux       | Unknown* | **Faster*** | Unknown* |

\* Platform has GL threaded optimisation of its own, and comparison is done against cases with that optimization enabled.
\*\* Similar peak performance, but with _much_ less stuttering as seen in above videos. Performance is generally better (+3fps around Mario Odyssey.) though some games are unaffected - don't expect _every_ game to be the same. Your mileage may vary, and you can turn off the feature if it affects anything negatively (but it shouldn't!)

Note that the performance is only going to improve from here. Situations that are less than favorable are likely to improve, and since we are no longer working with a multithreading black box, we have the tools to improve it.

# Use for Vulkan and the future
**Vulkan** does not have any kind of driver threading of its own in any implementation - it is always assumed that the user is building and submitting commands in the most efficient way possible, such as in a standalone dispatcher thread. People are used to the performance they get from a GL multithreaded implementation, where the GL commands essentially cost nothing on our FIFO processing thread, so without any multithreading of our own a worse experience is guaranteed. This PR provides that dispatcher thread without any duplicate code, including the possibility for compiling multiple pipelines at once for faster shader compilation. 

Now that we have our own multithreading implementation, it will be much easier for us to find out which methods OGL costs are actually coming from, rather than stare at a profile where it randomly blocks execution on any number of them. We have a much easier path to tightening up both the commands and the command dispatching to achieve better performance, and potentially even help the driver threaded path along the way. One such change has been included that does this already!

Finally, the threaded GAL layer can be placed in front of any backend with minimal changes, including anything we might come up with in future. This ensures a quicker path to good performance in the future on more exotic backends.

# Call for testing!

Looking for tests in a wide range of games to make sure nothing has broken or is notably slower. If anything is slightly slower, that is fine as this feature will improve as we move into the future, and the user can disable it.

Specifically looking for AMD/Intel Linux users to test this change! You will need to disable the driver's multithreading using `mesa_glthread=false` to get the most out of ours. If performance is not ideal, we can make `Auto` mean off on mesa drivers, but only if you tell us.


JumpmanSr:
We spoke a bit on Discord about this but I'll provide a bit more detail here.

PC Specs:
CPU: 5900x
GPU 3090 FE
RAM: 32GB 3200Mhz

In testing "Hyrule Warriors: Age of Calamity v1.0.0" I found that with the background threading set to on/auto the game had a major improvement compared to mainline or previous versions of Ryujinx (in fact it made the game experience significantly better when compared to mainline). To quantify the difference, I went and stopped emulation. Then I turned background threading to off and restarted Ryujinx. Afterwards, I purged both PPTC and Shader Cache, and after getting past the menu and into the first stage, as soon as the world loads it crashes. I thought that was weird, so I decided to try again once again purging everything. Crashed in the same spot. 

Alright weird, so I turn background threading back on, launch up the game, play a little, it works just fine and works through restarts. So I turn background threading off again, same crash and in the same place. It's persisted through restarting my system as well. 

Attached you will find: Crash logs, and 2 pictures on ingame with BG threading on, the other at the location every crash has been. I hope this can help you resolve this issue because it does bring improvement in many titles. 

Log file:
[Ryujinx_1.0.0+e381dc2_2021-07-24_18-43-03.log](https://github.com/Ryujinx/Ryujinx/files/6873151/Ryujinx_1.0.0%2Be381dc2_2021-07-24_18-43-03.log)

Working: (Small mention, this would have the same results as the artifacts posted compared to the build here, I took these SS's before testing on the artifact and it was the same result) 
bton working

Crash:
btoff crash

If you need anything additional from me feel free to contact me on discord (Jumpman) or anywhere. I will update this with any edits if I find any information relating to this.

tazevam:
I tested Catherine FB, set backend multithreading to ON, cleared the PPTC and shader cache before starting the game, and then the game crashed soon after it started. 

[Ryujinx_1.0.0+e381dc2_2021-07-25_10-37-33.log]

log:
(https://github.com/Ryujinx/Ryujinx/files/6873304/Ryujinx_1.0.0%2Be381dc2_2021-07-25_10-37-33.log)

molitar:
Legend of Mana crash after winning battle.

[Ryujinx_1.0.0+e381dc2_2021-07-25_02-31-32.zip](https://github.com/Ryujinx/Ryujinx/files/6873481/Ryujinx_1.0.0%2Be381dc2_2021-07-25_02-31-32.zip)

riperiperi:
> 
> 
> Legend of Mana crash after winning battle.
> 
> [Ryujinx_1.0.0+e381dc2_2021-07-25_02-31-32.zip](https://github.com/Ryujinx/Ryujinx/files/6873481/Ryujinx_1.0.0%2Be381dc2_2021-07-25_02-31-32.zip)

This crash is related to NvDrvServices rather than this PR. This shouldn't change anything going on there.

riperiperi:
> 
> 
> (Crash when Background Threading is Set to Off in Hyrule Warriors: Age of Calamity)
> 
> We spoke a bit on Discord about this but I'll provide a bit more detail here.
> 
> ...

Should be addressed in the latest commit.

JumpmanSr:
> > (Crash when Background Threading is Set to Off in Hyrule Warriors: Age of Calamity)
> > We spoke a bit on Discord about this but I'll provide a bit more detail here.
> > ...
> 
> Should be addressed in the latest commit.

I can confirm after re-testing that the crash has been fixed and with BG threading off it is in parity with mainline. With BG threading on, its night and day how much faster shaders are compiled for this game.

MutantAura:
Overall improvements seen in games I've tested including SSBU, MK8D and LoZ:BoTW.

BoTW comparison: 
https://user-images.githubusercontent.com/44103205/126918333-9a977c02-262a-4131-bf1a-839badf63394.mp4

Major stutters still obviously present but the overall time they consume is greatly reduced across the board. A fresh 8 player smash game was actually _moving_ fairly quickly compared to master which basically stays stationary for a solid 20s at the beginning of a match.

However as noted over discord some graphical regression seen in Breath of the Wild with water.
Master (correct):  image

PR: image

No other regressions spotted in the games I've tested.

Mou-Ikkai:
Decided to try one of the more laggier/desync songs of Megamix, Ghost Rule. (Shader folder was deleted before each execution)
GAL Multithreaded seems to desync a bit less.

Main: https://www.youtube.com/watch?v=hWbvB3Z-pQk
GAL: https://www.youtube.com/watch?v=7dp1s1H5ZEc

gdkchan:
The NV stutter workaround thread can be removed now, I believe?

riperiperi:
> 
> 
> The NV stutter workaround thread can be removed now, I believe?

Only when our threading is on. When it is off, our NVAPI profile turns on nvidia's threading, which still needs this workaround.

riperiperi:
There has always been the possibility that conditional rendering could use a query object just as it is disposed by the counter queue. I've pushed a change that makes it so that when the host decides to use host conditional rendering, the query object is reserved so that it cannot be deleted. 

Counter events can now optionally start reserved, as the threaded implementation can reserve them before the backend creates them, and there would otherwise be a short amount of time where the counter queue could dispose the event before a call to reserve it could be made. Reservation is denied if the result is already available.

Ryujinx.Graphics.GAL/Multithreading/Commands/SetRenderTargetsCommand.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public void Set(TableRef<ITexture[]> colors, TableRef<ITexture> depthStencil)
+        {
+            _colors = colors;
+            _depthStencil = depthStencil;
+        }
+
+        public static void Run(ref SetRenderTargetsCommand command, ThreadedRenderer threaded, IRenderer renderer)
+        {
+            renderer.Pipeline.SetRenderTargets(command._colors.Get(threaded).Select(color => ((ThreadedTexture)color)?.Base).ToArray(), command._depthStencil.GetAs<ThreadedTexture>(threaded)?.Base);
```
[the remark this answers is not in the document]
riperiperi: I'll leave this as a nice to have for now. One idea is having a render target array attached to the threaded pipeline so one doesn't have to be allocated every time, but that is rather messy, and would assume that the size is never going to change (it doesn't... but there's no guarantee it won't in future)

Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs
```diff
@@ -0,0 +1,227 @@
...
+                TextureBarrierTiledCommand.Run(ref GetCommand<TextureBarrierTiledCommand>(memory), threaded, renderer);
+            Lookup[(int)CommandType.TryHostConditionalRendering] = (Span<byte> memory, ThreadedRenderer threaded, IRenderer renderer) =>
+                TryHostConditionalRenderingCommand.Run(ref GetCommand<TryHostConditionalRenderingCommand>(memory), threaded, renderer);
+            Lookup[(int)CommandType.TryHostConditionalRenderingFlush] = (Span<byte> memory, ThreadedRenderer threaded, IRenderer renderer) =>
+                TryHostConditionalRenderingFlushCommand.Run(ref GetCommand<TryHostConditionalRenderingFlushCommand>(memory), threaded, renderer);
+            Lookup[(int)CommandType.UpdateRenderScale] = (Span<byte> memory, ThreadedRenderer threaded, IRenderer renderer) =>
+                UpdateRenderScaleCommand.Run(ref GetCommand<UpdateRenderScaleCommand>(memory), threaded, renderer);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
```
gdkchan, line 221: Out of curiosity, why those flags are needed here? Is the JIT not inlining this function by default?

Ryujinx.Graphics.GAL/Multithreading/CommandType.cs
```diff
@@ -0,0 +1,94 @@
+﻿namespace Ryujinx.Graphics.GAL.Multithreading
+{
+    enum CommandType : byte
+    {
+        Action = 0x00,
```
gdkchan, line 5: Why are the entries using explicit values here? If we need to add a new entry on the middle here, then we'll need to adjust all the values of subsequent entries which is quite a lot of manual work.

Ryujinx.Graphics.GAL/Multithreading/SyncMap.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.Threading;
+
+namespace Ryujinx.Graphics.GAL.Multithreading
+{
+    class SyncMap
+    {
+        private HashSet<ulong> _inFlight = new HashSet<ulong>();
+        private AutoResetEvent _inFlightChanged = new AutoResetEvent(false);
```
gdkchan, line 9: This is not being disposed.

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -0,0 +1,434 @@
...
+        private const int MaxRefsPerCommand = 2;
+        private const int QueueCount = 10000;
+
+        private int _elementSize;
+        private IRenderer _baseRenderer;
+        private Thread _gpuThread;
+        private bool _disposed;
+        private bool _running;
+
+        private AutoResetEvent _frameComplete = new AutoResetEvent(true);
```
gdkchan, line 36: This is not being disposed.

Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
```diff
@@ -222,7 +222,10 @@ private void DrawBegin(bool incrementInstance, bool resetInstance, PrimitiveTopo
                 _instanceIndex = 0;
             }
 
-            _context.Renderer.Pipeline.SetPrimitiveTopology(topology);
+            if (_drawState.Topology != topology)
+            {
+                _context.Renderer.Pipeline.SetPrimitiveTopology(topology);
+            }
 
             _drawState.Topology = topology;
```
gdkchan, line 230: Might aswell only set this if the value actually changed.

Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
```diff
@@ -174,7 +175,7 @@ void resultHandler(object evt, ulong result)
 
                 if (counter?.Invalid != true)
                 {
-                    _channel.MemoryManager.Write(gpuVa, counterData);
+                    _channel.MemoryManager.WriteUntracked(gpuVa, MemoryMarshal.Cast<CounterData, byte>(MemoryMarshal.CreateSpan(ref counterData, 1)));
```
gdkchan, line 178: Out of curiosity, why is that untracked now?

Ryujinx.Graphics.GAL/IRenderer.cs
```diff
@@ -8,11 +8,13 @@ public interface IRenderer : IDisposable
     {
         event EventHandler<ScreenCaptureImageInfo> ScreenCaptured;
 
+        bool PreferThreading { get; }
```
gdkchan, line 11: Instead of adding this, I'd probably expose a capability like `IsDriverMultithreaded` or something that would be used on the decision of whenever to enable GAL threading or not.

Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs
riperiperi: This function contained a huge switch statement before - now it's notably simpler. I don't know if they do anything as-is.

Ryujinx.Graphics.GAL/Multithreading/CommandType.cs
riperiperi: Between categories the value jumps to a different value, but I guess the commands after the first in each category could have it omitted.

Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
riperiperi: This was causing a deadlock. That was a while ago though, and I forgot the reason.

Ryujinx.Graphics.GAL/Multithreading/CommandType.cs
gdkchan: Any special reason for it needing categories?

riperiperi: No technical reason, it's just neater looking at raw values. The "categories" are each resource type, for example texture or counter event.

Ryujinx.Graphics.GAL/IRenderer.cs
riperiperi: That's not what this boolean means. There's no way of detecting right now if the driver is actively multithreading - this just defines what the result of Auto is for this backend, and potentially based on the vendor string, as I thought there might be certain GPUs or drivers which had a clear disadvantage no matter what. As of right now the value is always true.

gdkchan: Right, I just used it as an example because I only imagine we'd want to turn it off if the driver is multithreading *and* we have no way to turn it off. But I don't think this is the case for any driver today, so I don't know when we'd want to turn it off.

Ryujinx.Graphics.GAL/Multithreading/CommandType.cs
gdkchan: Why not just remove the explicit values then, I honestly don't see any benefit in having them. If one is looking at raw values, its easy enough to just load the enum file and lookup what the value corresponds to. Having all values explicit means that we'd either need to renumber everything if a value is added to the middle/start of the enum, or that we'd always need to add at the end (which means new values are no longer organized in categories). Having a explicit value just for the first entry of each category means that one could easily, inadvertently have overlapping values on the enum and create hard to track bugs.

Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs
```diff
@@ -0,0 +1,227 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Graphics.GAL.Multithreading
+{
+    static class CommandHelper
+    {
+        private delegate void CommandDelegate(Span<byte> memory, ThreadedRenderer threaded, IRenderer renderer);
+
+        private static CommandDelegate[] Lookup = new CommandDelegate[256];
```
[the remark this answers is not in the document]
riperiperi: Maybe `Total` would be a better word? Though that would be out of line with the other occurances where we do something similar to this.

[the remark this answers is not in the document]
gdkchan: It's also possible to use `Enum.GetValues()` to get the maximum value and then add 1. This doesn't require any additional entry on the enum. Maybe something like `(int)Enum.GetValues<CommandType>()[^1] + 1` .

Ryujinx.Graphics.GAL/Multithreading/CommandType.cs
riperiperi: I'll remove the values then. With the values removed, using a count value on `_lookup` also makes more sense, as there are no gaps.

gdkchan (approved):
lgtm, thanks. Nice to finally be able to *take control of our own destinty*.

Thog (approved):
lgtm

PR #2504, opened 2021-07-25
Hide UI rework/arrow key fix

### LINKED ISSUE #2502: Left and right arrow keys brings up File menu after running the game
The arrow keys ← → on the keyboard keep bringing up the Ryujin menu at the top during gameplay, which causes a loss of control in any game. My workaround was to change the directional pads to a different set of keys, but I would rather go back to the arrow keys.

1. Run any game.
2. Press the left and right arrow keys.

### Log file

[Ryujinx_1.0.6976_2021-07-24_14-59-00.log](https://github.com/Ryujinx/Ryujinx/files/6873118/Ryujinx_1.0.6976_2021-07-24_14-59-00.log)
[Ryujinx_1.0.6976_2021-07-24_15-19-44.log](https://github.com/Ryujinx/Ryujinx/files/6873127/Ryujinx_1.0.6976_2021-07-24_15-19-44.log)

### Environment?

- Ryujinx Version: 1.0.6976

- System Specs:
Operating System: Microsoft Windows 10.0.19042 (X64)
CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz ; 8 logical
RAM: Total 16335 MB ; Available 10817 MB
NVIDIA Corporation NVIDIA GeForce GTX 1060 with Max-Q Design
- Applied Mods : [ romfs patch for audio ]

ooa113y:
This fixes #2502, which was a regression caused by #2411 (I'm sorry! 😅)

For one, Alt no longer focuses the menu bar (this is the part of my implementation that caused the issue). That functionality wasn't really relevant to the original intent of #2411, and the main purpose of the PR remains usable. Focus on Alt is something I might fix later, as I can't seem to figure out what broke it in the first place.

Secondly, this reworks the entire flow of how the "Show UI" hotkey works. Note that it is specifically a "Show UI" hotkey, and hiding the UI still has to be done via the Actions menu, as we don't want users to trigger it by accident (there was some relevant discussion in #2411).
1. It is now a legitimate entry in the `KeyboardHotkeys` struct, rather than being hacked on via `MainWindow`. Since we no longer need menubar access for this, this makes more sense.
2. The key is now configurable (via `Config.json`).
3. The key now defaults to F4. I found several annoyances with Alt, as it's part of the language-switching combination and also getting in the way of taking a (regular, non-Ryujinx) screenshot. Something that's not commonly used made more sense. Of course, due to the point 2 above, users are welcome to switch it back to Alt if they are so inclined.
4. The key is read from the config file at startup and inserted into the menu label so that the user is clear on the exact key that needs to be pressed, even if it's changed from the default. The user has to know how to bring the UI back.

Additionally, I ended up fixing a problem with the implementation of the `KeyboardHotkeyState` enum which would cause it to break were we to add any more hotkeys. The default values set here are 1, 2, 3 etc, so adding a third hotkey would make it automatically press the first two, as 1+2 would equal to 3. I switched the values to powers of 2, which is a safe approach for this kind of thing.

### DIFF
--- a/Ryujinx/Ui/RendererWidgetBase.cs
+++ b/Ryujinx/Ui/RendererWidgetBase.cs
@@ -571,6 +571,12 @@ namespace Ryujinx.Ui
                     Renderer.Screenshot();
                 }
 
+                if (currentHotkeyState.HasFlag(KeyboardHotkeyState.ShowUi) &&
+                    !_prevHotkeyState.HasFlag(KeyboardHotkeyState.ShowUi))
+                {
+                    (Toplevel as MainWindow).ToggleExtraWidgets(true);
+                }
+
                 _prevHotkeyState = currentHotkeyState;
             }
 
@@ -596,9 +602,10 @@ namespace Ryujinx.Ui
         [Flags]
         private enum KeyboardHotkeyState
         {
-            None,
-            ToggleVSync,
-            Screenshot
+            None = 0,
+            ToggleVSync = 1,
+            Screenshot = 2,
+            ShowUi = 4
         }
 
         private KeyboardHotkeyState GetHotkeyState()
@@ -613,6 +620,11 @@ namespace Ryujinx.Ui
             if (_keyboardInterface.IsPressed((Key)ConfigurationState.Instance.Hid.Hotkeys.Value.Screenshot))
             {
                 state |= KeyboardHotkeyState.Screenshot;
+            }
+
+            if (_keyboardInterface.IsPressed((Key)ConfigurationState.Instance.Hid.Hotkeys.Value.ShowUi))
+            {
+                state |= KeyboardHotkeyState.ShowUi;
             }
 
             return state;
--- a/Ryujinx/Ui/RendererWidgetBase.cs
+++ b/Ryujinx/Ui/RendererWidgetBase.cs
@@ -603,9 +603,9 @@ namespace Ryujinx.Ui
         private enum KeyboardHotkeyState
         {
             None = 0,
-            ToggleVSync = 1,
-            Screenshot = 2,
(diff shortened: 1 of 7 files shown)


Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -596,9 +602,10 @@ private bool UpdateFrame()
         [Flags]
         private enum KeyboardHotkeyState
         {
-            None,
-            ToggleVSync,
-            Screenshot
+            None = 0,
+            ToggleVSync = 1,
+            Screenshot = 2,
+            ShowUi = 4
```
AcK77, line 608: Why not 3?

AcK77: Nvm it’s at the end of the description, maybe we should use a bit flag instead.

ooa113y: Isn't this what we're effectively doing here? Is there a better approach? Googling "C# bit flags" nets me a ton of results for the `[Flags]` attribute that we're using here and not much else at all.

gdkchan: > 
> 
> Nvm it’s at the end of the description, maybe we should use a bit flag instead.

The values are already power of 2, therefore they can already function as "bit flags". Maybe using `1 << 0`, `1 << 1` and `1 << 2` would be a bit more clear though (and it's what we do for other enums that are used as flags).

ooa113y: Adjusted.

Thog (approved):

AcK77 (approved):

PR #2505, opened 2021-07-26
Use fixed bindings on shaders

gdkchan:
Depends on https://github.com/Ryujinx/Ryujinx/pull/2496

This removes the need for the `TranslationCounts` class on the shader translator, which is used to keep track of each binding used by each stage. Instead, it assigns specific binding ranges for each stage (for example, 0-17 ubo bindings for vertex, 18-35 for fragment, etc).

This has two advantages. First, the most obvious one that the code is smaller/simpler. And second, We reduce the number of GL calls we need to do to bind resources, since the binding numbers are now fixed, the likelyhood that a different shader will use the same textures/images etc at the same binding points is way higher, and in those cases we don't need to do anything if the resource is already bound. This could result in an improvement for uniform and storage buffers aswell, but right now, all of them are always bound on a shader change, and improving this is outside of the scope of this PR, but this change makes it more feasible.

As usual testing with a few games to ensure they did not regress is welcome.


riperiperi:
As-is, this breaks image bindings on NVIDIA.

image

```
Fragment info
-------------
0(22) : error C3012: invalid value '8' for layout qualifier 'binding'
0(23) : error C3012: invalid value '9' for layout qualifier 'binding'
```

gdkchan:
Closing due to the NVIDIA issue with image bindings. I will just keep opengl binding numbering as-is and use the new scheme for Vulkan only.

PR #2506, opened 2021-07-26
Update TamperMachine and disable write-to-code prevention

Caian:
This PR focuses on improving reporting for cheats and updating the cheat instructions to match [Atmosphere's implementation](https://github.com/Atmosphere-NX/Atmosphere/blob/master/docs/features/cheats.md). Those are:

- A warning is now issued when the cheat tries to write to code region instead of requiring log analysis in debug mode;
- The cheat names are now printed in the log for better error detection and support;
- Implemented new addressing modes `Alias` and `Aslr`;
- Implemented `else` instruction for conditionals.

Also, with #1865 merged, it is now safe to enable cheats to write to code regions (with the aforementioned warning). This is a preparation for the upcoming full support, but does not mean that all cheats will work (some of them may start working).

### DIFF
--- a/Ryujinx.HLE/HOS/Tamper/AtmosphereCompiler.cs
+++ b/Ryujinx.HLE/HOS/Tamper/AtmosphereCompiler.cs
@@ -9,14 +9,36 @@ namespace Ryujinx.HLE.HOS.Tamper
 {
     class AtmosphereCompiler
     {
-        public ITamperProgram Compile(IEnumerable<string> rawInstructions, ulong exeAddress, ulong heapAddress, ITamperedProcess process)
+        private ulong    _exeAddress;
+        private ulong    _heapAddress;
+        private ulong    _aliasAddress;
+        private ulong    _aslrAddress;
+        ITamperedProcess _process;
+
+        public AtmosphereCompiler(ulong exeAddress, ulong heapAddress, ulong aliasAddress, ulong aslrAddress, ITamperedProcess process)
         {
-            Logger.Debug?.Print(LogClass.TamperMachine, $"Executable address: {exeAddress:X16}");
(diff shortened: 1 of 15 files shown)


JumpmanSr:
Great work, did a little bit of testing and in MH:Rise and Pokemon Sword/Shield (two of the games I tested) it works without any lasting effects or anything like that. (after a restart and proper disable there are 0 issues with any bleed-over translation or anything)

Infinite wirebug which was broken before:
image

Monster HP displayed as potion count: 
image

Additionally, all encounter codes in Pokemon now work, as well as a plethora of other games in which they use code that modifies the code-region.

Ryujinx.HLE/HOS/Tamper/AtmosphereCompiler.cs
```diff
@@ -9,14 +9,36 @@ namespace Ryujinx.HLE.HOS.Tamper
 {
     class AtmosphereCompiler
     {
-        public ITamperProgram Compile(IEnumerable<string> rawInstructions, ulong exeAddress, ulong heapAddress, ITamperedProcess process)
+        private ulong    _exeAddress;
+        private ulong    _heapAddress;
+        private ulong    _aliasAddress;
+        private ulong    _aslrAddress;
+        ITamperedProcess _process;
```
AcK77, line 16:
```suggestion
        private ITamperedProcess _process;
```
and realign var names

AcK77 (approved):
Just a nit, instead of that it LGTM!

Ryujinx.HLE/HOS/Tamper/AtmosphereCompiler.cs
Caian: Fixed

Thog (approved):

PR #2515, opened 2021-07-29
Reduce JIT GC allocations

FICTURE7:
This PR reduces interactions with the GC and reduces memory usage during compilation of CPU translations. It is very to similar to #1954.

* Turn `Operand` and `Operation` into unmanaged structs. They now wrap a pointer to the actual data. This makes them pointer size and fast to pass around whilst also maintaining the reference semantics when they used to be classes.
* Replace the object pools with an arena allocator. Allocating in the arena is relatively fast. It could be made faster by adding a fast path. The arena will pool a certain number of pages and free them if they're not used for more than 5s.
* Add interning of `Constant` and `Register` operands. These kinds of operand are immutable, so they can be shared, instead of allocating a new one. The intern table has 256 slots and will linear probe up to 8 slots, which is quite cheap to look up.
* `RegisterUsage` pass has been optimized a bit to use SSE when available.
* `Optimizer` pass has been optimized a bit also. By iterating in reverse order, some dependency chains can be killed in a single pass. Also it uses the arena to allocate scratch memory now.
* Use a new heuristic for selecting the number thread when building the PPTC cache.

#### Caveats
The new approach brings a set of caveats as well.

* Having multiple compilation on a single thread can be problematic, because the arena may be reset before the other compilations is completed. This can cause the operands and operation in the other compilation to refer to the void. The old pools could also do that, but they would not point the void though.
* Pattern matching will not do a null so it can lead to `NullReferenceException`.

I've got some ideas on preventing the first issue from happening, but I'll leave that in a follow up PR (assuming its feasible).

#### Effects
Reduces PPTC rebuild duration by 20%-60%. It may also help with boot time and stutters when hitting new code paths.

#### TODO
- [x] Unpack `ArenaList<T>` in `Operation` to avoid holes.
- [x] Unpack `ArenaList<T>` in `Operand` to avoid holes.
- [x] Wrap operations with phi instruction with a `PhiOperation` structure or something.
- [x] Get rid of `ThreadStaticPool` & co.
- [x] Intern popular constants (like 0 and 1) and register `Operand`s.
  - [x] Investigate closed-hashing to reduce waste in the intern table.
- [x] ~~Investigate use of a singly linked list for `Operand.Uses` and `Operand.Assignments`.~~<br>Does not seem to be worth it. `Operand`'s size goes from 56 to 40 bytes, but d-cache efficiency goes down enough to degrade PPTC rebuild durations.
- [x] ~~Investigate moving `Operand.Symbol` out into an auxiliary table.~~<br>Would require a rather extensive change to PPTC to be neatly implemented. Maybe in another PR.
- [x] ~~Investigate 32-bit indices/relative pointers in the arena instead of full 64-bit pointer.~~<br>Does not seem to be worth it. Now that the arena allocates in pages, it is too expensive to dereference them.
- [x] Investigate better defaults for arena size for the different types.
- [x] Investigate an LRU style removal mechanism to the `Arena` allocator.
- [x] Investigate impact of memory usage and GC collections during PPTC rebuild.
- [x] Check this does not produce codegen diffs (apart from cf3d94612308f8d619bbd0ed1e669ccf5540589d).


gdkchan (commented):
It seems reasonable to me, I had a similar idea in the past, but I didn't really want to pass pointers everywhere through the JIT. Wrapping them inside the struct is a nice solution to this problem. There are some simple ways to reduce memory usage a bit I believe, like making `Intrinsic` and `Instruction` `ushort`, ~~and it's probably possible to make `Operand` more compact without much effort too (I think I did that in my PR), maybe that way you can make them fit on the 16MB allocation~~ (nevermind, I just noticed you already did that). In general the direction looks good to me, quite nice surprise to see those object pools finally gone 👍

FICTURE7:
Here is some data I collected regarding the arena sizes, aggregating SMO, MHR, ACNH, MK8 and Dead Cells consisting of ~200k compilations including both LCQ and HCQ. There are 4 arenas:
* Operands - Stores `Operand.Data` structs
* Operations - Stores `Operation.Data` structs
* References - Stores references to operands and operations
* Default - Stores anything else, like bitmaps

64KiB for each arena is enough to compile ~65% of the translations and 512KiB enough for ~98%. So a 64KiB page size and pooling up to 8 pages seems appropriate.

<details>
<summary>Measurements</summary>

Everything here is in KiB.
```
Sample count: 205293
Arena - Default
 Maximum: 27785.34375
 Average: 73.31081903087153
 Deviation: 511.58269551939964

 25% Q: 0.390625
 50% Q: 1.59375
 75% Q: 11.078125
 99% Q: 1594.171874999942

 32KiB:  0.8524213142422613
 64KiB:  0.8990485375670011
 128KiB: 0.933341988612126
 256KiB: 0.9579724391193872
 512KiB: 0.9742196338817071


Arena - Operands
 Maximum: 7904.109375
 Average: 93.43151620037706
 Deviation: 176.33259256753956

 25% Q: 8.484375
 50% Q: 27.46875
 75% Q: 102.75
 99% Q: 559.0737499999996

 32KiB:  0.5369580997013733
 64KiB:  0.6663662537206549
 128KiB: 0.779566723339163
 256KiB: 0.8551587398245264
 512KiB: 0.9876603352608477


Arena - Operations
 Maximum: 4025.9375
 Average: 70.29273516244221
 Deviation: 116.4411050174129

 25% Q: 6.484375
 50% Q: 21.5234375
 75% Q: 79.4140625
 99% Q: 379.0078125

 32KiB:  0.5887720255464673
 64KiB:  0.7126753153118043
 128KiB: 0.8134727898944333
 256KiB: 0.9099248318093466
 512KiB: 0.9946412825846556


Arena - References
 Maximum: 16143.5
 Average: 98.92535984091917
 Deviation: 181.10505836355966

 25% Q: 8.09375
 50% Q: 27.609375
 75% Q: 104.3046875
 99% Q: 582.5690625

 32KiB:  0.5378154944878284
 64KiB:  0.6654601433213331
 128KiB: 0.7780955118305866
 256KiB: 0.8497561783576019
 512KiB: 0.9847568847339884


Arena - Total
 Maximum: 42087.5078125
 Average: 281.9054287232734
 Deviation: 667.2509880725881

 25% Q: 23.2734375
 50% Q: 78.0
 75% Q: 290.453125
 99% Q: 1674.2293749999992

 32KiB:  0.3081603523113123
 64KiB:  0.4650635982325976
 128KiB: 0.6071767840875322
 256KiB: 0.7257652004891048
 512KiB: 0.8187340760840441
```

#### Charts

Outliers were removed in these charts, otherwise a magnifying glass would be needed to inspect them. Each bar represents 4KiB.

Default
Operands
Operations
References
Total

I can upload the data-set here incase someone has got a better idea on how to tune this.
</details>

Niwu34:
Getting a crash when loading my PPTC with your latest commit (I purged it)

```
00:00:02.372 |I| HLE.GuestThread.14 Ptc MakeAndSaveTranslations: 0 of 109287 functions translated | Thread count: 32
00:00:07.131 |E| Application : Unhandled exception caught: System.OutOfMemoryException: Insufficient memory to continue the execution of the program.
   at ARMeilleure.Common.ArenaAllocator.ThrowOutOfMemory() in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\Common\ArenaAllocator.cs:line 118
   at ARMeilleure.Common.ArenaAllocator.Allocate(Int32 size) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\Common\ArenaAllocator.cs:line 33
   at ARMeilleure.IntermediateRepresentation.Operand.Add[T](T& item, T*& data, UInt16& count, UInt16& capacity) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\IntermediateRepresentation\Operand.cs:line 272
   at ARMeilleure.IntermediateRepresentation.Operand.AddUse(Operation operation) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\IntermediateRepresentation\Operand.cs:line 215
   at ARMeilleure.CodeGen.X86.X86Optimizer.RunPass(ControlFlowGraph cfg) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\CodeGen\X86\X86Optimizer.cs:line 14
   at ARMeilleure.CodeGen.X86.CodeGenerator.Generate(CompilerContext cctx, PtcInfo ptcInfo) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\CodeGen\X86\CodeGenerator.cs:line 106
   at ARMeilleure.Translation.Compiler.Compile(ControlFlowGraph cfg, OperandType[] argTypes, OperandType retType, CompilerOptions options, PtcInfo ptcInfo) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\Translation\Compiler.cs:line 58
   at ARMeilleure.Translation.Translator.Translate(UInt64 address, ExecutionMode mode, Boolean highCq) in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\Translation\Translator.cs:line 296
   at ARMeilleure.Translation.PTC.Ptc.<>c__DisplayClass72_0.<MakeAndSaveTranslations>g__TranslateFuncs|0() in C:\Users\Niwu\Desktop\Ryujinxpptc brrr\ARMeilleure\Translation\PTC\Ptc.cs:line 813
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

FICTURE7:
Thanks for testing, this should be fixed in f3f80289fed46f8f3497ce9722ebf7f34997325b, along with some other issues introduced in 66b8810e1f9e02ad815e67b567e4d9a62affb6d9.

Niwu34:
Thanks for the quick response:
The issue was fixed!!!

Here some tests on the latest commit:

SSBU - 110k PPTC
Master: 1:38,63 sec
#2515:  58,42 sec

Splatoon 2 - 37k  PPTC
Master: 18,33 sec
#2515: 9,32 sec

Monster Hunter Stories 2 - 35k PPTC
Master: 16,38 sec
#2515: 8,83 sec

FICTURE7:
I've been chasing a diff and turns out it was a codegen issue. The use of `With` here appears to be inoffensive
https://github.com/Ryujinx/Ryujinx/blob/70f79e689bc947313aab11c41e59928ce43be517/ARMeilleure/CodeGen/X86/Assembler.cs#L912-L927
but can cause issues. Because the `With` method will mutate the instance, and since the instance may be referred to in another operation. That other operation will end up with the mutated `Operand`. Here is an example:

Before optimizer:
```wasm
i32 %1 = ZeroExtend i32 0xA63AAC
i32 %2 = Compare i32 %1, i32 0x0, Equal
i32 %3 = ShiftRightUI i32 %4, i32 %1
BranchIf %2, 0x0, NotEqual
```

After optimizer:
```wasm
;; `i32 %1 = ZeroExtend i32 0xA63AAC` gets folded into a Copy of 0xAC, then the constant operand 0xAC
;; gets propagated into its uses.
;; `i32 %2 = Compare i32 %1, i32 0x0, Equal` gets propagated into the BranchIf.
i32 %3 = ShiftRightUI i32 %4, i32 0xAC
BranchIf 0xAC, 0x0, Equal
;; The constant 0xAC in the ShiftRightUI and BranchIf refers to the same Operand instance. Since
;; WriteShiftInst will mask the shift. The BranchIf is generated as `BranchIf 0xC, 0x0, Equal` instead.
```

___

This change does not produce diffs minus the one above and d801f56d4863301c1bf64569b33a2676e442d6b5. So far, I've only seen the above issue in only one case in ~1.33GB of codegen dumps on Windows. This was tested by dumping the translations sorted by address in a binary file and patching all the relocatables to 0. Both master and this PR produced the same hash on the dump when d801f56d4863301c1bf64569b33a2676e442d6b5 is excluded.

gdkchan:
Does that lead to bugs on master? From what I recall, it does not reuse operands if they have the same constant there.
Btw, which kind of diffs is the `RegisterUsage` pass change expected to produce?

FICTURE7:
Yeah, the bug happens on master only, and the diffs from d801f56 is just a modest one, it emits a single `LoadArgument` instead of two (one for integer registers and one for vector registers).

I've seen the `RegisterPass` emit some quite extreme stuff when there are many returns and many registers to store, tail merging appeared quite attractive. I also think we should move the store of registers closer to their last definition or something which would reduce register pressure a bit at context stores, should be easy in SSA form (I think). Maybe I'll look into another approach to the pass itself.

gdkchan:
> I've seen the `RegisterPass` emit some quite extreme stuff when there are many returns and many registers to store, tail merging appeared quite attractive. I also think we should move the store of registers closer to their last definition or something which would reduce register pressure a bit at context stores, should be easy in SSA form (I think). Maybe I'll look into another approach to the pass itself.

I thought about this, I think the only issue is placing those stores inside of a loop. Probably isn't too hard to check if the definition is inside a loop, this is a similar problem to the decision of where to spill variables during register allocation. Also from what I recall, the code is not fully in SSA form at this point (as it still has the Arm registers rather than the SSA variables).

FICTURE7:
> I thought about this, I think the only issue is placing those stores inside of a loop.

Oof, right.

> this is a similar problem to the decision of where to spill variables during register allocation.

Speaking of register allocators, I'll take a stab at reduce the memory footprint (managed memory at least) of the LSRA next in another PR, since this one is getting a bit large.

___
I investigated the impact this change has on memory usage and GC allocations during rebuilds. Since rebuilds are faster, memory grows faster. However the memory use reached at the end of the rebuild is lower than that of master as expected.

#### Charts
This was collected using `dotnet-counters` at 1 sec interval (it is the lowest `dotnet-counters` allows). The plots starts and ends with PPTC rebuilds.
<details>
<summary>ANCH</summary>

acnh
</details>

<details>
<summary>MHR</summary>

mhr
</details>

<details>
<summary>Dead Cells</summary>

dc
</details>

gdkchan:
> Speaking of register allocators, I'll take a stab at reduce the memory footprint (managed memory at least) of the LSRA next in another PR, since this one is getting a bit large.

Nice. Yea, it is pretty intense from what I recall, although the memory usage is not as bad since the option to map on the host address space directly (removing the need for all the address translation code per guest memory access). From what I recall, most of the memory usage on LSRA is from `LiveInterval`s.

Maybe with those changes, PPTC can use a higher number of threads for compilation, since the current constraint was due to the high memory usage and large amount of "garbage" produced iirc.

LDj3SNuD:
> Maybe with those changes, PPTC can use a higher number of threads for compilation, since the current constraint was due to the high memory usage and large amount of "garbage" produced iirc.

Well maybe https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Translation/PTC/DegreeOfParallelism.cs could be adjusted, although I don't think it can be over-adjusted, perhaps because of locks contentions: maybe https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L25 should be analysed.

But what would make the difference is restoring this constant https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Decoders/Decoder.cs#L16 to its original value (5000 if I'm not mistaken), as I had to reduce it at the launch of PPTC, over a year ago now; however, boot times and memory usage should be re-analysed.

FICTURE7:
> Maybe with those changes, PPTC can use a higher number of threads for compilation, since the current constraint was due to the high memory usage and large amount of "garbage" produced iirc.

Rebuilding PPTC is CPU-bound so the optimal number of threads to use in theory is the number of cores available. If we have too many threads, we waste time doing context switches and if do not have enough, we underutilize the CPU. Although this is not very clear when a GC and simultaneous multi-threading is around (and things like turbo boost, thermals etc). 

> Well maybe https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Translation/PTC/DegreeOfParallelism.cs could be adjusted, although I don't think it can be over-adjusted

Took a closer look at `DegreeOfParallelism` (DOP), my understanding is that `GetWeight` should return a value in `[0,100]` but if `gib` is larger than `(100 - _coefficient) / IncrementByGiB`, it will return a value larger than 100. If a machine got a bit more than 6 GiB of available memory, the DOP will be greater than the number of logical cores the machine has.

It appears to me that machines with a lot of available memory (> 6 GiB) and relatively low core counts will be "penalized" because the DOP will be high. And machines with low memory available (< 6 GiB) and high core counts will be "penalized" because the DOP will be low.

If someone has 12 GiB of available memory and 8 logical cores, the degree of parallelism will be 14. There will be 6 extra threads which am unsure is really speeding things up.

dop

I think something like `dop = Math.Max(1, Environment.ProcessorsCount - 1)` should be enough, a core is left around for things like the GC and gives mobile CPUs some breathing room to not down throttle too much.

I maybe missing something here though.

LDj3SNuD:
The Available Memory is the free physical (not virtual) memory; the heuristic algorithm was thus designed to be centred on the characteristics of the previous one (75% weight to be applied to the number of cores @ 4 GiB) and to allow a linear (configurable) increase in the degree of parallelism to try to satisfy the requests of some users with capable hardware who complained of low CPU utilisation during PPTC translations.
It is true that after POWER and your PR, re-tuning or a new heuristic algorithm may be required; feel free to change that and if we receive any complaints via Discord, from 80% of users who have PPTC enabled, we will let you know here.
(ah, nice graph)

FICTURE7:
Sounds good, if the new heuristic is doodoo I'll revert to the old one.

ARMeilleure/Common/ArenaAllocator.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            int currentTime = Environment.TickCount;
+
+            // Free pooled pages that has not been used in a while. Remove pages at the back first, because we try to
+            // keep the pages at the front alive, since they're more likely to be hot and in the d-cache.
+            for (int i = _pages.Count - 1; i >= 0; i--)
+            {
+                PageInfo info = _pages[i];
+
+                if (currentTime - info.LastUse >= 5000)
```
gdkchan, line 108: I'm not a huge fan of time based "garbage collection" methods myself, as if there is a bug somewhere (related to the freed memory), it becomes very hard to reproduce it on a different machine, or even on a later run. I personally would prefer if it used something like the number of compilations performed, or something like that, instead of time elapsed, if possible. Althought in this case, I guess using the number of translations would delay the memory being freed for longer...

gdkchan (commented):
Imo, the PPTC version should be changed, as this contains a bugfix, and also, if there's some issue caused by the changes, we'd want to catch it sooner rather than later.

ARMeilleure/Common/ArenaAllocator.cs
FICTURE7: > I'm not a huge fan of time based "garbage collection" methods myself, as if there is a bug somewhere (related to the freed memory), it becomes very hard to reproduce it on a different machine, or even on a later run.

I do not think this is going to be too problematic because after a reset, all memory previously allocated should be considered freed anyways. And resets happens after each compilation, so the page is not freed mid compilation. Although there are no mechanism to prevent someone from doing it right now (the object pools could also be reset mid compilation and cause issues), We can use reference counting to prevent the allocators from being reset mid compilation.

>  I personally would prefer if it used something like the number of compilations performed, or something like that, instead of time elapsed, if possible.

This is something I was going to look into when adding a fast path, and looking for cheaper alternatives to `Environment.TickCount`. I was thinking of something like this: each page has a flag `unused` and a counter. After each reset, if the page was `unused` , the counter is incremented. If that counter reaches a certain threshold, the page is freed. This makes codegen for the fast path smaller than currently, see [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYG3XNgBmMJqQYBBWF2znJkiGGwZoDAN40GnhmKi8FTk1wMKA4wDAYABWwAcxgASS5DCDcPLzSdBmAATwwYACpIiF4eGQ1qNPTdYhQCqNjZDGwoDDKKrwzq2piYAFEuABNWtqZdYvCAVS4OXBhB1LaMsYYRblyoIc8AX3mGHZ8/AMjuhKSGAH0xbrKdjpqLOwcAgAol3F4ALxgASjdhnbTeIYGE8Lt0WBEiiUoAwANTAjhjL5vT4MAA850usXB3T6/R+7nKw3Sd1guA4knCAF4MWCIWNSv8/oSiZ5QVi6VDYdSngieEiPjANiy2YJJtNZgxqXQhUTGQsAOwMUnklpyhjbZkVNXERW2eyOXKyewAdyeyO+Qo1aR2AG0ALIwDAACwg/Ti4kkTwdztd7rEkgA8mI+BAuLgWAA5CAJSTFYrRL4AXT2vn8uSYdz1j0NJpePAY5vx2sV/RghmwKstNE2QA).

FICTURE7: I implemented a fast path and a new heuristic which is more aggressive at freeing memory. The new heuristic uses time and number of resets to figure out how aggressive it should be when freeing memory.

If we used a fixed threshold (for number of resets unused) to decide when to free the pages, we will have to make a "larger" trade off on performance between when rebuilding PPTC and when hitting new code paths. If the threshold is low then we free aggressively (causing allocations for next compilation), causing a degradation in rebuild durations. If the threshold is high then we keep memory around for longer.

Let me know if you do not like the use of time here again, I think I'll swap to a fixed high threshold.

gdkchan (approved):
lgtm. thanks. Tested a few games here (both with PPTC on and off) and they are all working properly 👍 Also tested 32-bit titles for good measure. A few users tested it too and the PPTC compilation time halved for some of them. On my tests with PPTC off, games were also booting quickly and seemed more responsive.

riperiperi (approved):
Fantastic work! Some notable improvements here with PTC off, too.

### AOC
Master: https://www.youtube.com/watch?v=e3A6P3BnSTQ
PR: https://www.youtube.com/watch?v=zz-4tzTIy1g
This game's an outlier as its translations are really rough. You spend a long time ingame before it gets through the rejit queue and the game finally speeds up. 

In this example, the game's framerate is much higher while running LowCq translations, and it takes less time to reach a stable 30fps (a few seconds shorter). I think in this case, the multithreaded rejit was impacting the other emulator threads significantly.

The recording for this is a bit scuffed on the PR video- I think there are 2 stutters due to my network drive having issues: one during loading and one ingame.

Thealexbarney:
The JIT has been hitting a lot of `EnsureSameType` asserts since this PR. I think SSBU 1.0.0 hits one a second or 2 after booting

FICTURE7:
Thanks for the heads up, will look into it.

ARMeilleure/IntermediateRepresentation/Operand.cs
```diff
@@ -143,6 +161,158 @@ internal void NumberLocal(int number)
...
+
+                return;
+            }
+
+            // Could not add item in the fast path, fallback onto the slow path.
+            ExpandAdd(item, ref data, ref count, ref capacity);
+
+            static void ExpandAdd(T item, ref T* data, ref ushort count, ref ushort capacity)
+            {
+                ushort newCount = checked((ushort)(count + 1));
```
gdkchan, line 281: This seems to be throwing on this game: https://github.com/Ryujinx/Ryujinx-Games-List/issues/3796
More than 65535 uses?

FICTURE7: Oof, would not be surprised if it was the case. Would be interesting to see what kind of translation it is trying to compile, to know how we end up with more than 65535 uses.

PR #2374, opened 2021-06-19
Add Direct Mouse Support

emmauss:
Add direct mouse support, similar to the direct keyboard support. This enables games and homebrew that support usb mouse to have that option enabled.
Touchscreen is disabled when direct mouse is enabled. Also, the real cursor is hidden while inside the client area, but gtk seems to reset the mouse cursor state often, so you will see the mouse pop in once in a short while.

PR #2527, opened 2021-08-04
Make audio disposal thread safe on all 3 backends

gdkchan:
This fixes a bunch of issues, mostly related to the audio backends.

First, `SDLHardwareDeviceSession` can't be disposed twice. Attempting to do so will cause a crash due to a double free. This was happening most of the time here when I use the SDL2 backend. What was causing the double dispose was the session being disposed from the audio out server thread, and from some other thread disposed the `Switch` instance (which disposes the audio device driver, which in turn disposes all the existing sessions). There is a race when the driver and session dispose methods are called the same time, which in turn can cause the session dispose to be called twice, and then free it twice, causing the crash.

OpenAL sessions would deadlock most of the time when disposed. This caused the emulator to hang while closing when the OpenAL backend was used here.

SoundIO seems to be... fine? I rarely use it, but on the few tests I made the disposal seemed to work fine. Maybe the library already handles double free internally.

Anyway, I fixed all the issues above by changing how the session disposal works. First I got rid of the lock on driver, and replaced the sessions list with a thread safe data structure (`ConcurrentDictionary`). `Unregister` is still thread safe because the `TryRemove` method is guaranteed to be thread safe. `Unregister` now returns a bool indicating if the session was still on the list, if it was not on the list, that means another thread tried to dispose it already -- and won the race, so we don't need to do anything in this case.

This fixes the emulator deadlocking on exit trying to dispose OpenAL (can't deadlock when you don't have locks ;) ) and also fixes the crash on exit when SDL2 is disposed.

Additionally, this fixes a NRE on `Window.Cursor = null`. To be honest, I'm not sure why this code was added, `Window` is null at that point... I added a null check there, but it would be nice to hear from the author of this code, I think the line should be removed.

### DIFF
--- a/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
+++ b/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceDriver.cs
@@ -3,7 +3,7 @@ using Ryujinx.Audio.Common;
 using Ryujinx.Audio.Integration;
 using Ryujinx.Memory;
 using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
 using static Ryujinx.Audio.Integration.IHardwareDeviceDriver;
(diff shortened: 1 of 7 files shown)


CeruleanSky:
For context it seems @emmauss added the windows.cursor = null code when adding cursor hiding for their direct mouse support [in this commit](https://github.com/Ryujinx/Ryujinx/pull/2374/commits/4ec55d82f3d34d47017aea50697b5526984f0419)

Thog (approved):

AcK77 (approved):

gdkchan:
Had a talk with emmauss, and according to him the `Window.Cursor = null;` line is useless (because it already shows the cursor on the MouseLeave event), so I have removed it.

PR #2532, opened 2021-08-06
Use a new approach for shader BRX targets

gdkchan:
**Overview:**

BRX is basically a indirect branch shader instruction. The branch target offset comes from a register. So the problem here is that we have no way to know where the branch may land, as the offset is unknown at compile time. Before, it would make assumptions about the possible target offsets from the program layout. Basically it assumed that everything after the BRX itself was a potential target, until the target branch of the first branch (the "merge" point). This works reasonably well for simple programs, but falls apart for more complex ones.

The new approach makes assumptions about the code the compiler is going to generate for those instructions. It assumes that the BRX is preceded by a LDC instruction (used to load the target offset from the constant buffer), which is preceded by a SHL instruction (used to shift left the index by 2 to get the byte offset), which is then preceded by a IMNMX instruction (used to enforce that the index is in range).

The new approach requires accessing the constant buffer (1) data, which is not present on the current shader cache. In order to keep the new shader cache compatible with older versions of the emulator, I simply appended the new constant buffer data to the end of the code data. Older versions will just read the code with the constant buffer data at the end (and should ignore it since the shader ends before that). New versions are aware of this and should read the constant buffer data from the end of the code. One of the reserved fields was used for the constant buffer size. Compiling an old cache on the new version will fail for shaders with the BRX instruction (trying to read the constant buffer will throw IOR as the size is 0), the cache loader detects that (`task.IsFaulted` is true) and simply deletes the incompatible shader entries from the cache.

Special thanks to @mageven for finding the BRX issue on Cadence of Hyrule.

**Improvements:**

Fixes several graphical glitches on Hatsune Miku Project DIVA MEGA 39's:
Broken visualizer on remote control clip:
Before:
image
After:
image

Missing transparency, lights etc on Piano Forte Scandal clip:
Before:
image
After:
image

Missing effects in Theo clip:
Before:
image
After:
image
(The light trail was always complete before, instead of following Miku's finger).

Other lighting issues (MMORPG Addict Anthem clip):
Before:
image
After:
image

Other lighting issues:
Before:
image
After:
image

Fixes Bowwow being a white glowing ball on Zelda's Link Awakening:
Before:
image
image
After:
image
image

Fixes reflection bug on Cadence of Hyrule:
Before:
image
After:
image

And more...

**What to test**

Any game really. This change the way how the BRX shader instruction is implemented, so any game using this instruction has the potential to get better or worse (hopefully only for the better). A `Failed to find targets for BRX instruction at 0xN.` warning will be printed on the console/log if it fails to find BRX targets, so watch out for that.

It is worth to also test if the shader caches are intact when switching between this and older versions.


codbugpatrol:
@gdkchan, great job, really works for chomp ball in Link's Awakening. Thought that would fix the https://github.com/Ryujinx/Ryujinx/issues/2454 , but unfortunatelly it does not. =(

Mou-Ikkai:
The MMORPG Addict's Anthem - Missing tunnel
The tunnel effect used in this song was missing its walls, this PR also fixed that issue.

Before:
image
After:
image

Thog (approved):

ActualMandM:
HIBANA's lights are fixed with this PR

Before:
2021-08-09_00-59-49_Ryujinx-bak

After:
2021-08-09_00-56-55_Ryujinx

riperiperi (approved):
LGTM, cool approach! Everything I've tested so far works. Invalidating only the BRX shaders also seems to be fine in the games I tested.

PR #2534, opened 2021-08-09
Initial tessellation shader support

### LINKED ISSUE #2498: Missing tesselation shader in LM3
Missing tesselation shader in Luigis mansion 3

This game uses tesselation function and vertex texture fetch (for dynamic deformation)
Currently Ryu shows empty space in sand level instead of correct deform-able sand.

And this is how it works on Switch hardware and Yuzu emu:

So we are asking for tesselation implementation in Ryujinx :)    ...Please :)

gdkchan:
**Overview**

This adds support for tessellation shaders (control and evaluation stages, also known as hull and domain), which is the only shaer type that was not yet support. Most of the work here was just adding declarations that are specific to those stages, and also improving the implementation of a few instruction. `VSETP` is not implemented (but not fully, see `TODO`, but all functionality needed by those shaders should be implemented), ALD/AST now respects the `P` (patch) modified (exclusive to tessellation, accesses per-patch attributes), and the `O` (output) modifier, used to load data from an output attribute (rather than from the input, which is the default behaviour). Some built-ins commonly used on those stages have also been added.

**What it fixes**

The only known visible improvement right now is Luigi's Mansion 3 (which uses tessellation for the sand etc).
image
Closes #2498.

**What to test**

Games that uses tessellation (Luigi's Mansion 3, Mario Golf: Super Rush), games that uses geometry shader might be worth testing aswell, as they use `VSETP` and and shares some stuff done on the tessellation stages. UE4 games usually uses geometry shaders.


Ryujinx.Graphics.Shader/CodeGen/Glsl/CodeGenContext.cs
```diff
@@ -72,11 +73,15 @@ public void LeaveScope(string suffix = "")
 
         private static int FindDescriptorIndex(TextureDescriptor[] array, AstTextureOperation texOp)
         {
+            // Can we trust the type on this texture operation?
+            // If not then just ignore the type.
+            bool accurateType = !texOp.Inst.IsTextureQuery();
```
gdkchan, line 78: A bit unrelated, but I just found this bug while testing. The texture query "TextureOperation" type may not match the sampler type. That would make the search fail for a texture query that needs the scale value, returning -1. Using -1 as index is invalid and causes the shader to fail to compile. In the future, a better way to address this is probably getting the texture type from the texture pool (texture query instructions don't have a type unfortunately).

RinMaru:
ryujinx_capture_2021-08-09_06-22-11
amd rx 480

Linear524:
Yup ! It works now ! :)
Thank you GDKCHAN ! (^_^)
And not only sand levels, there are other levels with empty textures works now too ! For exapmle - museum floor now visible and last bossfight level floor now visible too :)
Don't know how these levels was related to tessellation issue, but now they all works perfect ! :)
Luigi's Mansion 3 now fully playable :)  (except characters reflection issue, but it looks like texture cache issue)

Linear524:
My rig is:
- SYS: Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz | 14 Threads | 31.83 GiB RAM | 
- SYS: Operating system: Windows, Major: 10, Minor: 0, Build: 18362
- GPU: 'NVIDIA GeForce GTX 1080' running on driver 471.41.0.0
- SYS: Firmware version: 9.1.0

Linear524:
Is it possible to move this into master ?

jduncanator:
This is still completely broken on AMD.

mrjo118:
Tested on Nvidia Geforce GT 1030 and get the same result as @Linear524, working great except the shadow map issue. It looks like the shadow map issue mentioned has been fixed in mainline build 1.0.7020, and also the change in 1.0.7022 made a small difference to the rendering (only thing I noticed was when you are inside the elevator and push a button, Luigi's hat flashed black, but with that fix that doesn't happen now, although something about the animation is a little off... possibly just a stutter). I gave it a go cherry picking the commits from those two builds in GIT and putting it on this pull request, and I get perfect results (at least for the average eye).

So once this is fixed for AMD gpu's and merged on the mainline build, this game should be pretty much free of any rendering glitches :)

MarkLanders777:
Can this pr be rebased?

gdkchan:
Rebased.

Ryujinx.Graphics.Shader/CodeGen/Glsl/CodeGenContext.cs
gdkchan: I think this is not needed anymore as the TXQ sampler type should be correct now. I will remove it later.

Linear524:
Wow ! This PR is finally merged to Master ! (^_^)
Thank you guys ! (^_^) Ryujinx is the BEST !!! :)

Linear524:
Unfortunately - trailing shadow bug is still here...
ryujinx_capture_2021-11-08_19-24-26
ryujinx_capture_2021-11-08_19-24-38
ryujinx_capture_2021-11-08_19-27-10

It looks like shadow buffer is not being updated correctly, or not being flushed from previous frame...

jduncanator:
> Unfortunately - trailing shadow bug is still here... ryujinx_capture_2021-11-08_19-24-26 ryujinx_capture_2021-11-08_19-24-38 ryujinx_capture_2021-11-08_19-27-10
> 
> It looks like shadow buffer is not being updated correctly, or not being flushed from previous frame... (P.S. - on my hardware this issue appears after 2 or 3 minutes after game start. But on other hardware this issue occurs a bit faster. So I think it might depend on VRAM size...)

Best to open a new issue for this if you believe it's a regression. It's unlikely to be looked at here.

INFINITY99K:
Still exists on master 1.0.7104

riperiperi:
This PR does not aim to fix issues with shadow maps not clearing.

INFINITY99K:
Thanks for clarification!

PR #2537, opened 2021-08-09
Use "Undesired" scale mode for certain textures rather than blacklisting

riperiperi:
There are a number of textures where it is definitely _possible_ to scale them, but we don't scale them due to the potential for wasted work (depth of field need not be high res) and undesired effects (blur shaders do not behave correctly when scaled, scaling texture atlases results in ugly linear blending blur).

This PR gives these types of texture a new scale mode, rather than blacklisting them entirely. This mode is "Undesired", as we don't _want_ to scale the texture, but if it's bound alongside another texture that is scaled, we'd rather scale them _both_ than blacklist and potentially lose res scale entirely.

### Effects:
- Games that reuse depth stencil between a lot of unrelated draws to textures of different sizes will no longer blacklist from scaling forever. (Bayonetta 2)
- Textures that are detected as being atlas-like or dof-like will not be locked out of scaling forever - if a depth stencil target is bound when drawing, they can gain res scale. This might result in more 3D UI elements seeing resolution scale. (Xenoblade DE menus, though they're still heavily antialiased)
- Some other games that randomly blacklist to 1x, or don't scale at all may now work.

### Other changes:
- The approach for blur/dof texture detection has been updated to take into account width alignment, as it can greatly alter the aspect ratio. This fixes small textures failing to detect, which would have cascaded across mipmap texture views and forced textures to scale.
- Render target scale is now only sent to the backend when its value has changed, rather than each time the scales are evaluated.
- Scissor and Viewport are only recalculated when the value changes too.

image

### Testing
Please make sure that no existing games break as a result! Test your favourites. 😌

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
@@ -141,6 +141,16 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
+        /// Check if a texture's scale must be updated to match the configured res scale.
+        /// </summary>
+        /// <param name="texture">The texture to check</param>
+        /// <returns>True if the scale needs updating, false if the scale is up to date</returns>
+        private bool ScaleNeedsUpdated(Texture texture)
+        {
+            return texture != null && !(texture.ScaleMode == TextureScaleMode.Blacklisted || texture.ScaleMode == TextureScaleMode.Undesired) && texture.ScaleFactor != GraphicsConfig.ResScale;
+        }
+
+        /// <summary>
         /// Sets the render target color buffer.
         /// </summary>
         /// <param name="index">The index of the color buffer to set (up to 8)</param>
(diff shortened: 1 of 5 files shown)


riperiperi:
Note: This currently reveals an issue in Vulkan regarding the scaled copy - just a heads up if you wanted to test that. It'll probably be fixed soon on that PR.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -140,6 +140,16 @@ public void SetGraphicsTexturePool(ulong gpuVa, int maximumId)
             _gpBindingsManager.SetTexturePool(gpuVa, maximumId);
         }
 
+        /// <summary>
+        /// Check if a texture's scale must be updated to match the configured res scale.
```
gdkchan, line 144:
```suggestion
        /// Check if a texture's scale must be updated to match the configured resolution scale.
```

gdkchan (approved):
Looks good, thanks.

Thog (approved):

legend800:
While Bayo 2 does upres now, it seems like there's an issue with menus where it slows down to ~10 fps at 2x. On Master, it's full speed at 2x. While Master might not be doing the same thing here, that's a pretty huge drop on a simple screen. In-game is about 30fps as comparison. 

PR:
bayo 2x

Master:
Bayo 2 720p

As a note, upressing still doesn't affect CTR track elements:
CTR

riperiperi:
Bayonetta seems to be slow due to BGRA on opengl, so it will get (pixel count)x slower with res scale. I think it will be fine on vulkan.

Issue #1109, opened 2020-04-13
Reading uninitialised variables leads to misrendering on Mesa drivers


sigexp:
I've been analysing renderdoc capture of an Animal Crossing apitrace, provided in the [mesa issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/2708). Some textures appear to have severe misrenderings such as random black noise over the texture, that on some GPUs might cover the texture completely. 

<details>
<summary>Misrendering example</summary>

![Ry_misrender](https://user-images.githubusercontent.com/10928134/79140925-726b0d80-7dc1-11ea-8692-3f6fe4746ecb.jpg)

</details>

Looking at the call, where one of the major misrenderings first appeared, I've found that the fragment shader operates on a variable, passed from the vertex shader, where it wasn't properly initialised. Initialising this variable does indeed fix the defects for a particular texture:

<details>
<summary>Partially fixed</summary>

![Ry_partfixed](https://user-images.githubusercontent.com/10928134/79142301-f4f4cc80-7dc3-11ea-9492-21356a28d56c.jpg)

</details>

Applying `glsl_zero_init` workaround in `00-mesa-defaults.conf` also fixes the problem.

I've attached [both vertex and fragment shaders](https://github.com/Ryujinx/Ryujinx/files/4470760/Ryujinx_issue.zip). The variable, that was causing trouble is `out_attr2`: only `x` component was initialised in the vertex shader.

According to the GLSL spec, uninitialised variables have undefined value. From Section 5.8 "Assignments" of the GLSL 4.60.7 spec (page 122 of the PDF):

> Reading a variable before writing (or initializing) it is legal, however the value is undefined.

jduncanator:
Thanks for reporting the issue @sigexp. I'm currently working on (re)implementing our shader cache, and this issue actually leads into some work that has to be done anyway, so I'll tackle this issue there!

ReinUsesLisp:
I think Nvidia GPUs automatically default output attributes as `vec4(0,0,0,1)` (or `vec4(0)`, it might be different than `gl_Position`), similar to what D3D seems to do.

It probably is a matter of initializing output attributes.

I'm working on 3D unit tests, I can report in a few weeks if it's `vec4(0)` or `vec4(vec3(0), 1)` what it defaults to.

PR #2538, opened 2021-08-10
Make sure attributes used on subsequent shader stages are initialized

### LINKED ISSUE #1895: Hatsune Miku Mega39 - AMD GPU Incorrect/Broken Render

Loading in any module will produce eyes that aren't rendered correctly on an AMD GPU.</br>

Incorrect Render 1</br>

Another notably example is one particular module that has white eyes, though it is rendered with the same error shown above</br>
Incorrect Render 2</br>

There is a seperate issue opened currently on yuzu that has a similar error like the one shown that I can link if asked for.

Either enter the customization menu or use the default module and enter a song (In PV mode or rhythm mode)

### Log file

[Ryujinx_1.0.6291_2021-01-11_12-50-56.log](https://github.com/Ryujinx/Ryujinx/files/5797792/Ryujinx_1.0.6291_2021-01-11_12-50-56.log)

### Environment?

- Ryujinx version: 1.0.6291</br>
- Game version: 1.0.0</br>
- System Specs:
  - OS: Windows 7
  - CPU: i3-6100
  - GPU: AMD R7 360 2GB VRAM
  - RAM: 8GB
- Applied Mods : No

### Additional context?

None that I'm aware

gdkchan:
Sometimes, an shader attribute is consumed on a later stage, but the previous stage does not write it. On NVIDIA, all attributes that are not written to are initialized with a default value of (0, 0, 0, 1). This is not the case for other vendors however, and according to the spec, the values on them are "undefined". This causes problems for AMD and Intel as the attributes have different values.

The current approach taken is assuming all the first 16 attributes are used, so initializes the first 16.  That's because on desktop, usually only 16 attributes are supported (at least on OpenGL, Vulkan might allow more?). This causes shaders to fail to compile if the previous stage does not write the attribute at all. For example, one shader in Yo-kai Watch 4 fails to compile due to this, I'm not sure for what the shader is used though, the game seems to render fine regardless.

Additionally, there is another problem where partially written attributes are not initialized. So if the vertex shader writes X, Y and Z, and the  fragment shader reads W, the value will be undefined. This causes eyes to be solid blue on Hatsune Miku, and shadows to break in Zelda Link's Awakening on Intel and AMD.

This change fixes all the issues above by propagating attribute use information backwards (from later stages to early ones, like from fragment to vertex etc). That way, all attributes that are used on the subsequent stages can be properly initialized. Attribute use information is now computed very early (on the decode step) rather than later (on the structured program construction step), this was done to allow the uses to be propagated after decode, but before translation. Attribute initialization is also done earlier now, it is done at the IR level instead of the backend, this reduces code duplication between GLSL and SPIR-V.

So, to sum it up, those issues should be fixed:
- Shadows not working on Zelda Link's awakening (Intel and AMD).
- Hatsune Miku eyes being black (Intel and AMD).
- Shader failing to compile on Yo-kai Watch 4 (all vendors, no visible effect though?).

Probably more...

Would be nice to get a confirmation that those games were fixed on AMD, as I did not test them *on AMD*, and they still don't render properly on Intel OpenGL due to other issues (I will be fixing some of them soon), also helps Vulkan.

Fixes #1895 and #1109.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -38,7 +38,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2494;
+        private const ulong ShaderCodeGenVersion = 2538;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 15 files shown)


SimpleName140:
Tested this PR, and can confirm on my hardware that it does fix it. Thanks for the fix :+1: 
AmdFix1
AmdFix2

Thog (approved):

AcK77 (approved):

RinMaru:
this works on AMD i can confirm. RX 480

PR #2539, opened 2021-08-10
Workaround for cubemap view data upload bug on Intel

gdkchan:
Setting data to faces > 0 on a cubemap view doesn't work on Intel, so as workaround, this sets the data on the storage instead.
Fixes eyes rendering of Hatsune Miku on Intel.
On master it flickers between black and solid blue.
With #2538, it stops flickering and renders black instead:
image
With #2538 *and* this change, it renders correctly:
image
Only Intel on windows is affected, and only on OpenGL.


codbugpatrol:
Great job thou. This PR only works for Intel and AMD right ?  Dreaming about these PRs that fixes flickering on DBZ char menu selection screen and fight loading. I'm testing every PR to see if the dream come true.

gdkchan:
It only affects Intel on Windows. AMD does not have this bug.

Thog (approved):

riperiperi (approved):

PR #2540, opened 2021-08-10
Workaround for Intel FrontFacing built-in variable bug

gdkchan:
This is a workaround for yet another bug on Intel where `gl_FrontFacing` returns incorrect values (it seems to be flipped?), the solution here is casting the bool to float and then bitcasting to integer and checking if its non-zero. The solution was found by trial and error, if someone knows a better way, please let me know.

Fixes rendering of mario mustache on Super Mario Odyssey on Intel (affects both Vulkan and OpenGL):
Before:
image
(Note that the screenshot is a bit old, the shadow glitch on cappy was already fixed).
After:
image

This also fixes the "dead" palette on Zelda Link's Awakening, however the game is not rendering correctly on OpenGL, so the fix is only visible on Vulkan.

Other changes here includes:
- Added 2 new functions to the IGpuAccessor interface, used to check if the host has the vector indexing bug (present on AMD) and the front facing bug (present on Intel).
- LDC now uses vector indexing of the form `data[offset >> 2][offset & 3]` on NVIDIA and Intel, which produces better codegen (and makes shaders easier to read).
- Functions that just returns a host capability on `GpuAcessor` and `CachedGpuAccessor` were moved to a common base class to avoid duplication and bugs. In fact, one of them (the one to check the texture shadow LOD capability) was missing from the `CachedGpuAccessor`, which likely means those shaders were broken on shader cache rebuild on the vendors/drivers that does not support the extension (as the default interface implementation returns true). Now this is fixed.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
@@ -117,8 +117,13 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
             return $"{GetUbName(stage, slot, cbIndexable)}[{offset >> 2}].{GetSwizzleMask(offset & 3)}";
         }
 
-        private static string GetVec4Indexed(string vectorName, string indexExpr)
+        private static string GetVec4Indexed(string vectorName, string indexExpr, bool indexElement)
         {
+            if (indexElement)
+            {
+                return $"{vectorName}[{indexExpr}]";
+            }
+
             string result = $"{vectorName}.x";
             for (int i = 1; i < 4; i++)
             {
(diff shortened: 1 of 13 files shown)


Thog (approved):

Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
```diff
@@ -198,6 +203,15 @@ public static string GetAttributeName(AstOperand attr, ShaderConfig config, bool
                             case AttributeConsts.PositionY: return "(gl_FragCoord.y / fp_renderScale[0])";
                             case AttributeConsts.PositionZ: return "gl_FragCoord.z";
                             case AttributeConsts.PositionW: return "gl_FragCoord.w";
+
+                            case AttributeConsts.FrontFacing:
+                                if (config.GpuAccessor.QueryHostHasFrontFacingBug())
+                                {
+                                    // This is required for Intel on Windows, gl_FrontFacing sometimes returns incorrect
+                                    // (flipped) values. Doing this seems to fix it.
+                                    return "(-floatBitsToInt(float(gl_FrontFacing)) < 0)";
```
riperiperi, line 212: This is very cursed. I imagine some over-eager optimization on intel's part is breaking this.

riperiperi (approved):
Nice catch. Glad to see the vector indexing come back, too.

Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
Benedani: what even

my brain cells

PR #2541, opened 2021-08-11
Improve firmware install error due to outdated keys

ooa113y:
We have a constant of `UserError.FirmwareParsingFailed`, which when used produces the following message:

> Ryujinx was unable to parse the provided firmware. This is usually caused by outdated keys.

This is a better description of the firmware installation problems that a lot of users face than the currently used `UserError.NoKeys`, that simply states:

> Ryujinx was unable to find your 'prod.keys' file

This PR changes the generated error message to the more relevant one.


gdkchan (approved):
Nice. Weird that the error message exists, but was not used anywhere before.

riperiperi (approved):

Issue #2191, opened 2021-04-09
Ignore games when path not found (.symlink with size 0 kb)


Cloneg:
#### Overview:
- Ignore games with path not found ( .symlink with size 0) ( in my case it is because they are in a external hard drive not connected bc I keep some games in my main drive and the other as symlink in the same folder)</br>
![Captura de pantalla (389)](https://user-images.githubusercontent.com/67025623/114221731-48e0b580-9933-11eb-959b-25796129a392.png)

#### Smaller Details: 
- Ignore thoses games with path not found and don't show them in the game list (ryujinx alredy does this) and don't show the Ryujinx error window every time you open Ryujinx, maybe keep the first log in yellow if it's necessary.</br>

![Captura de pantalla (387)](https://user-images.githubusercontent.com/67025623/114221690-3d8d8a00-9933-11eb-9fa5-a4ba5bb60d34.png)

### Why would this feature be useful?
Users won't need to close the Ryujinx error window **every** time they open Ryujinx if they have games in a hard drive not connected but they want to play other games that are available.

![Captura de pantalla (388)](https://user-images.githubusercontent.com/67025623/114221768-56963b00-9933-11eb-8677-1845dd45a830.png)

AcK77:
Fixed in #2547

PR #2547, opened 2021-08-12
Remove file error popup

bobhope9848:
Address #2191 
Added file size check to now only load in files more than zero bytes in length.


gdkchan:
I think we need a general solution for invalid/corrupt files, rather than just ignoring empty files.

bobhope9848:
> I think we need a general solution for invalid/corrupt files, rather than just ignoring empty files.

Alright how about upon loading games into the application for the first time we create a hash of the loaded game files and then report which files are corrupted? Or maybe you're looking for just a generic read the header of the file and see if it contains game data?

gdkchan:
It already handles the corrupt and empty files. As far I can tell, the annoyance the user wants addressed is the dialog box every time the emulator opens and loads the game list. In this case the solution is simply removing the dialog box. My opinion is that it should only be shown only the first time (after you add the games directory and the list loads for the first time). After that, it's not very useful to have the same error message again and again. Anyway, I wouldn't be against just removing it entirely, it is not very useful.

bobhope9848:
> It already handles the corrupt and empty files. As far I can tell, the annoyance the user wants addressed is the dialog box every time the emulator opens and loads the game list. In this case the solution is simply removing the dialog box. My opinion is that it should only be shown only the first time (after you add the games directory and the list loads for the first time). After that, it's not very useful to have the same error message again and again. Anyway, I wouldn't be against just removing it entirely, it is not very useful.

I think I might just go ahead and remove the message entirely. I'll work on both solutions just in case but it sounds like it doesn't really have a purpose anymore.

bobhope9848:
Alright removed the popup from happening and reverted my previous change. Still writes the error to log so I think this is a good solution should the user want to investigate why their files aren't showing up. LMKWYT

AcK77 (approved):

riperiperi (approved):
LGTM

Issue #2508, opened 2021-07-26
Consider using DSA functions on OpenGL


gdkchan:
#### Overview:
OpenGL 4.5 introduced support for DSA functions. Basically, they allow the state of the OpenGL objects to be changed without needing to bind them beforehand. On older versions of OpenGL, the typical way to copy data from one buffer to another for example, is:
```cs
GL.BindBuffer(BufferTarget.CopyReadBuffer, sourceBufferHandle);
GL.BindBuffer(BufferTarget.CopyWriteBuffer, destinationBufferHandle);

GL.CopyBufferSubData(BufferTarget.CopyReadBuffer, BufferTarget.CopyWriteBuffer, srcOffset, dstOffset, size);
```
Using the equivalent [DSA function](https://www.khronos.org/opengl/wiki/Direct_State_Access) here, we can reduce the number of calls required:
```cs
GL.CopyNamedBufferSubData(sourceBufferHandle, destinationBufferHandle, srcOffset, dstOffset, size);
```
Reducing the number of calls can bring a small performance improvements on games that does a large number of binding/buffer copy/texture copy/upload etc operations.

One thing to note too, is that it binds texture at binding point 0 before each draw/dispatch:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.OpenGL/Pipeline.cs#L1229
This is done because binding the texture to modify its state would also replace the texture previously bound at 0. With DSA this is no longer a problem as the texture handle is passed directly, without binding, so this can be removed.

The reason the new functions are not used right now, is because we used to target OpenGL 4.3, and older drivers that does not support DSA. Nowadays though, we require OpenGL 4.5, so it shouldn't be an issue anymore.

#### Nature of Request:  
Code and performance improvement. Passing the handles directly instead of binding them is also less error prone, as its not possible to leave things bound and later have their state unintentionally changed.

### Why would this feature be useful?
Makes the code more clear and faster.

ReinUsesLisp:
A few things to have in mind when transitioning into DSA:
- `CreateFramebuffers` is broken on Nvidia. It will create an [EXT_framebuffer_object](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_framebuffer_object.txt) instead of a core framebuffer. These can't have attachments with mismatching sizes. A workaround is to call `GenFramebuffers` + `BindFramebuffer(READ_FRAMEBUFFER, fbo)`.
[...]

PR #2549, opened 2021-08-13
Implements Direct State Access and Multi Bind

mpnico:
### Implements OpenGL 4.4 Multi Bind and OpenGL 4.5 DSA
The goal of this PR is to reduce GL calls by using OpenGL 4.4 and 4.5 new extensions (`ARB_multi_bind` and `ARB_direct_state_access`), closing #2508.

Multi bind allows to bind multiple buffers in one call whereas DSA allows to change states of objects without binding them (buffers / textures, framebuffers, …).

Unfortunately, I didn't see major performance increase during my tests, do not expect more than a few %.

About VertexArray: the current implementation is not 100% full DSA. The "DSA" version of `glEnableVertexAttribArray` / `glDisableVertexAttribArray` are not used because of a big negative performance impact (around 15% on SMO).

About Framebuffer: as stated by @ReinUsesLisp in #2508, I encountered a bug with `CreateFramebuffers` so I made an exception only for NVidia.

### What to test
Basically every games for potential regressions. My system is NVidia only so AMD/Intel tests are needed. 

During my tests, I also encountered a strange flickering bug in Monster Hunter Rise which disappears when activating then disabling OpenGL log level. I assume it is a driver bug with my system but this game should be tested.


A-w-x:
intel linux

links awakening
[Ryujinx_1.0.0+be5dffc_2021-08-16_06-05-29.log](https://github.com/Ryujinx/Ryujinx/files/6992031/Ryujinx_1.0.0%2Bbe5dffc_2021-08-16_06-05-29.log)
Screenshot_2021-08-16_06-09-52

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1006,6 +994,7 @@ public void SetTransformFeedbackBuffers(ReadOnlySpan<BufferRange> buffers)
             }
 
             int count = Math.Min(buffers.Length, Constants.MaxTransformFeedbackBuffers);
+            int[] bufferHandles = new int[count];
```
riperiperi, line 997: Use `stackalloc` to avoid allocating arrays in the heap each time this is called.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1118,18 +1109,29 @@ private void SetBuffers(int first, ReadOnlySpan<BufferRange> buffers, bool isSto
         {
             BufferRangeTarget target = isStorage ? BufferRangeTarget.ShaderStorageBuffer : BufferRangeTarget.UniformBuffer;
 
+            int[] bufferHandles = new int[buffers.Length];
+            IntPtr[] bufferOffsets = new IntPtr[buffers.Length];
+            IntPtr[] bufferSizes = new IntPtr[buffers.Length];
```
riperiperi, line 1114: `stackalloc` these - SetBuffers is called a lot and allocating arrays each time isn't great.

Ryujinx.Graphics.OpenGL/Framebuffer.cs
```diff
@@ -17,9 +17,28 @@ class Framebuffer : IDisposable
...
+            {
+                // Nvidia bug, cf. https://forums.developer.nvidia.com/t/bug-spec-violation-checknamedframebufferstatus-returns-gl-framebuffer-incomplete-dimensions-ext-under-gl-4-5-core/58647
+                GL.GenFramebuffers(1, &localHandle);
+                GL.BindFramebuffer(FramebufferTarget.Framebuffer, localHandle);
+            }
+            else
+            {
+                GL.CreateFramebuffers(1, &localHandle);
+            }
+
```
riperiperi, line 34: Double newline.

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -276,8 +253,12 @@ private void ReadFrom2D(IntPtr data, int layer, int level, int width, int height
             TextureTarget target = Target.Convert();
 
             int mipSize = Info.GetMipSize2D(level);
+            int maxLevels = 1 + (int)Math.Floor(Math.Log2(width * height));
```
riperiperi, line 256: This calculation doesn't make any sense. Given a 256x256 texture, this will calculate 17 max levels... which is pretty far off.

```
256
128
64
32
16
8
4
2
1
```
There are 9 mip levels in this situation. With your current calculation, the correct result is obtained with 256x1.

Ryujinx.Graphics.OpenGL/VertexArray.cs
```diff
@@ -20,9 +20,18 @@ class VertexArray : IDisposable
         private uint _vertexAttribsInUse;
         private uint _vertexBuffersInUse;
 
+        public unsafe int Create()
+        {
+            int localHnadle;
```
riperiperi, line 25: `localHandle`

riperiperi (commented):
Just a quick initial review. I've yet to do more testing, but I can confirm the Link's Awakening issue also happens on NVIDIA.
image

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
mpnico: Well this is a commit mistake. This was a test to avoid flood of OpenGL errors in logs in SMO which happens on master too. No relation with DSA so I will remove it.

mpnico:
Fixed Link's Awakening and applied code review changes.

mpnico:
I found a workaround for MHR flickering. I also pushed a few more fixes

Ryujinx.Graphics.OpenGL/Buffer.cs
```diff
@@ -8,21 +8,18 @@ static class Buffer
...
+            valueArr[0] = value;
+
+            GL.ClearNamedBufferSubData<uint>(
+                destination.ToInt32(),
+                PixelInternalFormat.Rgba8ui,
+                (IntPtr)offset,
+                size,
+                PixelFormat.RgbaInteger,
+                PixelType.UnsignedByte,
+                valueArr.ToArray());
```
gdkchan, line 22: Calling `ToArray` here negates the benefits of using the stackalloc (as you will allocate a new array on the heap anyway).

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -833,7 +833,7 @@ public void Flush(bool tracked = true)
         /// </summary>
         public void ExternalFlush(ulong address, ulong size)
         {
-            if (!IsModified)
+            if (!IsModified || HostTexture.isDisposed())
```
gdkchan, line 836: There is no need for this `isDisposed` method (btw the correct name would be `IsDisposed` as funciton names always start with upper case). You can just have a bool field and set it to true when `DisposeTextures` is called. Then here it would just check `if (!IsModified || _disposed)`.

gdkchan: Also I'm not sure of the effects of leaving `IsModified` set to true. Maybe it could cause the texture to be flushed all the time. @riperiperi should be able to give a better suggestion here.

Ryujinx.Graphics.OpenGL/Buffer.cs
mpnico: This is the fix for Monster Hunter Rise. Maybe using `MemoryMarshal.GetReference` is better ?

gdkchan: If theres an overload that takes ref, you can just pass `ref valueArr[0]`. Or well, just `ref value` since theres only one value.

mpnico: Method support to pass `ref value` but does not fix the bug that way. `MemoryMarshal.GetReference` does fix it but I don't know it efficiency.

gdkchan: I think passing the array is just fixing the issue by chance. It's possible that the driver is reading past the pointer that we pass. Try changing the stackalloc size from 1 to 4, and passing the pointer directly, see if that fixes it.

Ryujinx.Graphics.Gpu/Image/Texture.cs
mpnico: I did some tests and, at least in MK8D where I found those disposed textures, and no excessive flush happen so I will make the well advised changes and won't touch `IsModified`

Ryujinx.Graphics.OpenGL/Buffer.cs
mpnico: Good idea but unfortunately it doesn't fix the bug. But you gave me the idea to change `PixelInternalFormat` to a 1 component only (`GL_R32UI`) format and it fixes the bug. I don't know if it's very less efficient though.

mpnico: Finally the best fix is using `MemoryMarshal.GetReference` or not using DSA here. I choose the first one.

mpnico:
Rebased and fixed a problem with some UE games. It need some more tests on AMD as I also pushed a potential fix for garbage pixels at edges of the window.

TSRBerry:
Needs a rebase

marysaka:
Closing as stale.

PR #2552, opened 2021-08-14
Enable transform feedback buffer flush

gdkchan:
Transform feedback buffers can be modified by the GPU, so they must have the `write` flag set to true to indicate they might have been modified.

Seems to fix vertex explosion in SNK Heroines Tag Team Frenzy.
Before:
image
After:
image
Worth noting that it seems somewhat random, but I was getting the same corruption with this character selection. There's still some vertex explosions on some special effects, but I suspect this is a different problem since those seems more consistent.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferManager.cs
@@ -503,7 +503,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


riperiperi (approved):

Thog (approved):

zzgundamx:
Tested this with Age of Calamity where there was vertex explosions when piloting Divine Beasts. The explosions are still present - but it is a little bit better as I could see where I was going where before the explosions would just block the screen completely.

PR #2557, opened 2021-08-15
Add support for HLE macros and accelerate MultiDrawElementsIndirectCount #2

mpnico:
This PR is a rebase of previous #2138 with the ability to check for HLE code support of each macro on current renderer before using it.

The other caveats from previous PR are still here.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
@@ -305,6 +305,63 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         }
 
         /// <summary>
+        /// Performs a indirect multi-draw, with parameters from a GPU buffer.
+        /// </summary>
+        /// <param name="engine">3D engine where this method is being called</param>
+        /// <param name="topology">Primitive topology</param>
+        /// <param name="indirectBuffer">GPU buffer with the draw parameters, such as count, first index, etc</param>
+        /// <param name="parameterBuffer">GPU buffer with the draw count</param>
+        /// <param name="maxDrawCount">Maximum number of draws that can be made</param>
+        /// <param name="stride">Distance in bytes between each element on the <paramref name="indirectBuffer"/> array</param>
+        public void MultiDrawIndirectCount(
+            ThreedClass engine,
+            int indexCount,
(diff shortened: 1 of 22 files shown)


Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
```diff
@@ -297,6 +297,11 @@ public void ClearBuffer(MemoryManager memoryManager, ulong gpuVa, ulong size, ui
             buffer.SignalModified(address, size);
         }
 
+        public BufferRange GetGpuBufferRange(MemoryManager memoryManager, ulong gpuVa, ulong size)
```
gdkchan, line 300: Missing XML docs on this function.

gdkchan (approved):
lgtm other than the comment.

jduncanator:
Do the TODO's from the initial PR not still apply? I feel like rebasing this and not including the original TODO's in the description glosses over the fact that the original PR wasn't yet complete.

Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
```diff
@@ -26,6 +26,8 @@ class DrawManager
 
         private int _instanceIndex;
 
+        private const int IndexBufferCountMethodOffset = 0x5f8;
```
jduncanator, line 29: Where does this magic constant come from? Should it live here in the DrawManager?

gdkchan:
> 
> 
> Do the TODO's from the initial PR not still apply? I feel like rebasing this and not including the original TODO's in the description glosses over the fact that the original PR wasn't yet complete.

For `Flush indirect buffer data? This is required for correctness, but right now it takes all the performance benefit of the change, needs further investigation.` the decision was just that we should just assume that the game is not going to modify the index buffer range on the indirect buffer for now, as there is otherwise no other way to handle this (other than mapping the guest memory directly on the host GPU). As for the other issue, I think it was not addressed, good point.

Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
gdkchan: It's the register offset.

riperiperi (approved):
LGTM.

PR #2563, opened 2021-08-17
Optimize LSRA

FICTURE7:
Right now ~50% of time is spent in the LSRA when rebuilding the PPTC cache. This PR attempts to optimize the LSRA.

#### TODO
- [ ] Investigate a linked-list for `Interval._ranges`. Should make splitting faster.
- [ ] Investigate a skip-list for `Interval._ranges` to enable fast look ups. (Apparently `List<T>.BinarySearch` will use `Array.BinarySearch` internally which is not the fastest implementation available. `Span<T>.BinarySearch` uses a different one which is faster).
- [ ] Optimize `BitMap.Enumerator`. A fair bit of time is spent there.

Let me know you got some other ideas on how to speed this up.


Niwu34:
Just a few quick tests on the current state of the PR:

PPTC count: 111.628 (Smash)
Master: 48,02s
Pr2563: 26,06s

PPTC count: 37.430 (Splatoon 2)
Master:  7,13s
Pr2563: 4,32s

PPTC count: 74.829(MHR)
Master: 32,20s
Pr2563: 19,23s

Again great Improvements!

FICTURE7:
I am marking this ready for review, since its been hanging for a while. There are still some areas where the LSRA could be optimized:
1. **Active Set**: On x86 the active set will have at most ~32 bits set. It should probably be swapped to a more dense data structure which can be iterated through faster.
2. **Inactive Set**: Most variables are in SSA form, so there usually is a small number of intervals which are inactive, except in loops. It too, should probably be swapped by a more dense data structure.
3. **Interval List**: The majority of time is now spent inserting the split child into the unhandled interval set. It should probably be swapped for a linked list instead. Or the `LiveIntervalList` could be optimized a bit more and be reused for that. It currently works well for small lists; the unhandled interval set can get quite large.

gdkchan:
> On x86 the active set will have at most ~32 bits set. It should probably be swapped to a more dense data structure which can be iterated through faster.

The problem with that is that some architectures has more registers, like Arm64 for example that has 32 + 32 (SIMD). So having a fixed limit here would make supporting such architectures harder. Realistically I don't think we'll ever need more than 64, so on places where we use a bit mask to track used registers, I thin using an `ulong` is fine. I was actually going to make a similar comment on your Hybrid Allocator PR since you made a change there that makes it limited to 16 registers.

FICTURE7:
I was thinking more in the lines of having a flat hash set using linear probing, should still be expandable and have somewhat constant access time and be fast to iterate through.

> I was actually going to make a similar comment on your Hybrid Allocator PR since you made a change there that makes it limited to 16 registers.

True. Although we can expand that to 32 registers per domain if we swapped the `int`s to `long`s. We can also introduce a `RegisterSet/Map` type if needed to abstract that part out (Wish we got C++ templates, would make writing these low-overhead data-structures less painful 😩).

ARMeilleure/CodeGen/RegisterAllocators/LiveInterval.cs
```diff
@@ -1,341 +1,289 @@
...
-        private List<LiveRange> _ranges;
+            public UseList Uses;
+            public LiveIntervalList Children;
 
-        private SortedList<int, LiveInterval> _childs;
+            public Operand Local;
+            public Register Register;
 
-        public bool IsSplit => _childs.Count != 0;
+            public byte IsFixed;
```
gdkchan, line 29: Any special reason for using a `byte` here rather than `bool`? Afaik both occupies 1 byte of memory.

ARMeilleure/CodeGen/RegisterAllocators/LiveInterval.cs
```diff
@@ -1,341 +1,289 @@
...
+            // Make sure the iterator in the new split is pointing to the start.
+            result.Reset();
 
-            return right;
+            return result;
         }
 
         private void AddSplitChild(LiveInterval child)
         {
             Debug.Assert(!child.IsEmpty, "Trying to insert a empty interval.");
```
gdkchan, line 284:
```suggestion
            Debug.Assert(!child.IsEmpty, "Trying to insert an empty interval.");
```

ARMeilleure/Common/BitMap.cs
```diff
@@ -170,12 +171,12 @@ public Enumerator GetEnumerator()
 
         public struct Enumerator : IEnumerator<int>
         {
-            private int _index;
+            private long _index;
```
gdkchan, line 174: Why was the `_index` type changed to `long`? Was it to avoid sign extension on the array access?

ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
```diff
@@ -227,8 +272,6 @@ private bool TryAllocateRegWithoutSpill(AllocationContext context, LiveInterval
             }
             else if (selectedNextUse < current.GetEnd())
             {
-                Debug.Assert(selectedNextUse > current.GetStart(), "Trying to split interval at the start.");
```
gdkchan, line 230: Why was this assert removed?

FICTURE7: This assert is always true, because it is in the else of the `if (selectedNextUse <= current.GetStart())`. So I decided to remove it.

ARMeilleure/Common/BitMap.cs
FICTURE7: Yes, it was to avoid sign extensions.

ARMeilleure/CodeGen/RegisterAllocators/LiveInterval.cs
FICTURE7: Nope, will change it to `bool`.

PR #2564, opened 2021-08-18
Clean the SD card save directory when opening the emulator

Thealexbarney:
Cleaning the directories ended up being the best solution because none of these three other options when encountering a file with a different key worked with every program's file system handling logic. Fixes Ryujinx/Ryujinx-Games-List#3678

- Deleting the file
- Deleting the file and creating an empty one
- Deleting the file and creating one of the same size

### DIFF
--- a/Ryujinx.HLE/FileSystem/EncryptedFileSystemCreator.cs
+++ b/Ryujinx.HLE/FileSystem/EncryptedFileSystemCreator.cs
@@ -9,8 +9,6 @@ namespace Ryujinx.HLE.FileSystem
 {
     public class EncryptedFileSystemCreator : IEncryptedFileSystemCreator
     {
-        public EncryptedFileSystemCreator() { }
-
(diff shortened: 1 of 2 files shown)


Benedani:
Is this what the Switch does? If it's not, which I'm 99.99999999% sure it doesn't, then this process shouldn't be done in the emulator.

What if someone doesn't want the SD card to be cleaned?

jduncanator:
> Is this what the Switch does? If it's not, which I'm 99.99999999% sure it doesn't, then this process shouldn't be done in the emulator.
> 
> What if someone doesn't want the SD card to be cleaned?

This PR only removes cached saves from the `save` subdirectories on the SD Card, it doesn't wipe the entire SD card.

Thealexbarney:
The only saves on the SD card are cache storage which can be cleared at any time on an actual Switch, either by removing the SD card or removing the `Nintendo` folder on the SD card

riperiperi (approved):
This is fine as a workaround for now, as a few games are not working as is.

Thog (approved):

PR #2566, opened 2021-08-19
Motion button

Nuschelbacke:
would be nice to have this on the master branch, its useful for some games, mario odyssey for example.

### DIFF
--- a/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
+++ b/Ryujinx.Common/Configuration/ConfigurationFileFormat.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Configuration
         /// <summary>
         /// The current version of the file format
         /// </summary>
-        public const int CurrentVersion = 22;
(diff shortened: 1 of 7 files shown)


Benedani:
needs a rebase

CeruleanSky:
I closed this at the time and have no plans to rebase it because to do it justice, it would need multiple controller support, but I lack games to test with.  Though it works(ed) fine with single player games like Mario Odyssey. Additionally with front-end changes like Avalonia upcoming and changes to how input works since then a motion button implementation probably needs to just be redone.

This was made a hotkey as at the time, and probably still now though I haven't checked,  a controller couldn't take input from a mix of keyboard keys and a gamepad, hopefully hybrid keyboard+gamepad support will be added someday as many gamepads have less buttons than a Switch and reassigning a controller button to motion means a loss of function of whatever that button used to do.

riperiperi:
You shouldn't re-open someone else's pull request unless you intend to finish it.

PR #2568, opened 2021-08-19
Remove pool cache entries for incompatible overlapping textures

### LINKED ISSUE #2344: Xenoblade Chronicles 2 memory leak

Once into a big map, while walking around the Ryujinx.exe process will slowly but steadily exhaust vram then main memory, possibly resulting in crash of the emulator process and sometimes the entire desktop environment. After the memory leak the emulator cannot be shut down normally. </br>
No change is made to the emulator or its settings. </br>

The memory leak will not happen until Gormott Province Upper Level, the first big map of the game. Once there, walking around, looking around or entering battle could trigger the leak. Having more member in the party seems to worsen the case, the game never crashes this way with a single character in the party.

### Log file

[Ryujinx_1.0.6906_2021-06-04_21-58-36.log](https://github.com/Ryujinx/Ryujinx/files/6599328/Ryujinx_1.0.6906_2021-06-04_21-58-36.log)
The emulator is forcefully terminated.

### Environment?

- Ryujinx version: 1.0.6906</br>
- Game version: both the NA version (0100e95004038000) 2.1.0 and the Japanese version (0100f3400332c000) 2.0.2.</br>
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 3600
  - GPU: NVIDIA Geforce 1660 super
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Whenever the memory leak happens the amount of memory occupied by Ryujinx.exe reported by task manager and resource monitor diverge. The number reported by task manager usually caps at ~4GB but the number reported by resource monitor will grow unbounded.

riperiperi:
This greatly reduces memory usage in games that aggressively reuse memory without removing dead textures from the pool, such as the Xenoblade games, UE3 games, and to a lesser extent, UE4/unity games.

This change stops memory usage from ballooning in xenoblade and some other games. It will also reduce texture view/dependency complexity in some games - for example in MK8D it will reduce the number of surface copies between lighting cubemaps generated for actors.

There shouldn't be any performance impact from doing this, though the deletion and creation of textures could be improved by improving the OpenGL texture storage cache, which is very simple and limited right now. This will be improved in future.

I've been meaning to do this for a while, it's just recently I got around to doing it, and realized we can use the modified flags for some neat tricks. The change itself is pretty simple - if a texture being overlapped by a new sample has been modified by cpu, then its modified data is likely destined for the new texture being sampled, and the old texture should be deleted as it contains invalid data and likely won't be used for some time.

Another potential error has been fixed with the texture cache, which could prevent data loss when data is interchangably written to textures from both the GPU and CPU. It was possible that the dirty flag for a texture would be consumed without the data being synchronized on next use, due to the old overlap check. This check no longer consumes the dirty flag.

Please test a bunch of games to make sure they still work, and there are no performance regressions. Most games I have tested do not trigger texture deletion during gameplay at all - the exception obviously being games that stream textures, and triggering it doesn't appear to hinder performance or cause issues.

Closes #2344

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -43,7 +43,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 oldestTexture.SynchronizeMemory();
 
-                if (oldestTexture.IsModified && !oldestTexture.ConsumeModified())
+                if (oldestTexture.IsModified && !oldestTexture.CheckModified(true))
                 {
                     // The texture must be flushed if it falls out of the auto delete cache.
                     // Flushes out of the auto delete cache do not trigger write tracking,
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -252,7 +252,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 if (!isView)
                 {
                     // Don't update this texture the next time we synchronize.
-                    ConsumeModified();
+                    CheckModified(true);
 
                     if (ScaleMode == TextureScaleMode.Scaled)
                     {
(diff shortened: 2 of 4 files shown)


gdkchan (approved):
lgtm, no adverse effects observed on the few games that I tested. And as expected, Xenoblade Chronicles  2 is running better here (no longer crashes after a few minutes in Torigoth, and performance seems better to, probably because I have a low amount of memory available (only 8GB of RAM and 6GB of VRAM)).

gdkchan:
Closes #2344, please include that in the description.

Thog (approved):
lgtm

PR #2569, opened 2021-08-19
Fix GetHandleInformation for mipmapped 3d textures

riperiperi:
Got this the wrong way round - was causing games to try synchronize mipmap levels of like 52 on a 3d texture with 6 levels. Also, corrected the variable name in the method that _was_ working.

This was probably causing minor issues in various games. SMO encountered the issue but I'm not sure if it was visible anywhere. Discovered by @mpnico when checking reasons for OpenGL errors.


gdkchan (approved):
Nice catch.

Thog (approved):

PR #2571, opened 2021-08-20
Fix type mismatch in `BitwiseAnd` simplification

FICTURE7:
`TryEliminateBitwiseAnd` would turn the `BitwiseAnd` operation into a copy of the wrong type. E.g:

Before `Simplification`:
```llvm
i64 %0 = BitwiseAnd i64 0x0, i64 %1
```

After `Simplication`:
```llvm
i64 %0 = Copy i32 0x0
```

Since the with the changes in #2515, we iterate in reverse order and `Simplication`, `ConstantFolding` does not indicate if it modified the CFG, the second pass to "retype" the copy into the proper destination type does not happen.

This also blocked copy propagation since its destination type did not match with its source type. But in the cases I've seen, the
`PreAllocator` would insert a copy for the propagated constant, which results in no diffs.

Since the copy remained as is, asserts are fired when generating it.

### DIFF
--- a/ARMeilleure/CodeGen/Optimizations/Simplification.cs
+++ b/ARMeilleure/CodeGen/Optimizations/Simplification.cs
@@ -71,7 +71,7 @@ namespace ARMeilleure.CodeGen.Optimizations
             }
             else if (IsConstEqual(x, 0) || IsConstEqual(y, 0))
             {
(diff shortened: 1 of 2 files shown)


Thog (approved):
nice catch

riperiperi (approved):
Nice catch!

PR #2582, opened 2021-08-22
Added fallbacks for all Audio Backends

Mou-Ikkai:
I added fallbacks for all backends, I discussed with GDKchan and they wanted a priority of:  SDL2 -> OpenAL -> SoundIO -> Dummy.
So I did my best to match that. The configuration default was also set to SDL2 due to all the fallbacks being added, so doing this would help new users, while not crippling the new users with incompatible audio devices.

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -408,10 +408,38 @@ namespace Ryujinx.Ui
                 if (SDL2HardwareDeviceDriver.IsSupported)
                 {
                     deviceDriver = new SDL2HardwareDeviceDriver();
-                }
+                }    
                 else
                 {
-                    Logger.Warning?.Print(LogClass.Audio, "SDL2 audio is not supported, falling back to dummy audio out.");
+                    Logger.Warning?.Print(LogClass.Audio, "SDL2 is not supported, trying to fall back to OpenAL.");
+                
+                    if (OpenALHardwareDeviceDriver.IsSupported)
+                    {
+                        Logger.Warning?.Print(LogClass.Audio, "Found OpenAL, changing configuration.");
+
+                        ConfigurationState.Instance.System.AudioBackend.Value = AudioBackend.OpenAl;
+                        SaveConfig();
+
+                        deviceDriver = new OpenALHardwareDeviceDriver();
+                    }
+                    else
+                    {
+                        Logger.Warning?.Print(LogClass.Audio, "OpenAL is not supported, trying to fall back to SoundIO.");
+                    }
+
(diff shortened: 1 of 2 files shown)


Ryujinx/Ui/MainWindow.cs
```diff
@@ -408,10 +408,38 @@ private void InitializeSwitchInstance()
                 if (SDL2HardwareDeviceDriver.IsSupported)
                 {
                     deviceDriver = new SDL2HardwareDeviceDriver();
-                }
+                }
```
gdkchan, line 411:
```suggestion
                }
```

Ryujinx/Ui/MainWindow.cs
```diff
@@ -408,10 +408,38 @@ private void InitializeSwitchInstance()
...
+                        SaveConfig();
+
+                        deviceDriver = new OpenALHardwareDeviceDriver();
+                    }
+                    else
+                    {
+                        Logger.Warning?.Print(LogClass.Audio, "OpenAL is not supported, trying to fall back to SoundIO.");
+                    }
+
+                    if (SoundIoHardwareDeviceDriver.IsSupported)
```
gdkchan, line 430: This should be inside the else above.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -422,7 +450,35 @@ private void InitializeSwitchInstance()
...
+                        SaveConfig();
+
+                        deviceDriver = new SDL2HardwareDeviceDriver();
+                    }
+                    else
+                    {
+                        Logger.Warning?.Print(LogClass.Audio, "SDL2 is not supported, trying to fall back to OpenAL.");
+                    }
+
+                    if (OpenALHardwareDeviceDriver.IsSupported)
```
gdkchan, line 469: This should be inside the else above.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -433,8 +489,22 @@ private void InitializeSwitchInstance()
...
+                        SaveConfig();
+
+                        deviceDriver = new SDL2HardwareDeviceDriver();
+                    }
+                    else
+                    {
+                        Logger.Warning?.Print(LogClass.Audio, "SDL2 is not supported, trying to fall back to SoundIO.");
+                    }
+                    
                     if (SoundIoHardwareDeviceDriver.IsSupported)
```
gdkchan, line 508: This should be inside the else above.

Mou-Ikkai:
And I broke it, wooo.

Thog (approved):

AcK77 (approved):

PR #2590, opened 2021-08-25
Very basic performance boosts.

Mou-Ikkai:
A few `.ConfigureAwait(false)`s placed into the HLE.
Performance boost? Maybe on really weak CPUs.


AcK77:
Since `.ConfigureAwait(false)` remove the `SynchronizationContext` it could be probably good to keep it in HLE service. Otherwise it shouldn't be an issue for URL call methods.

AcK77:
Until we aren't sure of the performance boost and if uses `ConfigureAwait(false)` do really something in our case, I close the PR. Feel free to comment it with more informations related to the changes. Thanks.

Issue #2600, opened 2021-08-28
No More Heroes 3 intense glare issue


dondonpow:
## Bug Report
https://files.catbox.moe/lfn2fa.PNG
the camera is pointing up. if it was pointing down, the white would fill up the entire screen.

### What's the issue you encountered?

In a lot of scenes, the glare is so intense that it blinds most of the screen.

### How can the issue be reproduced?

Emulate No More Heroes 3 and get to the part on the spaceship.

### Log file

N/A
### Environment?

- Ryujinx version: 1.0.7018</br>
- Game version: 1.0</br>
- System Specs:
  - OS: *Windows 10*
  - CPU: *i5-7400*
  - GPU: *NVIDIA GeForce GTX 1660*
  - RAM: *16GB*
- Applied Mods : none

### Additional context?

N/A, my environment is fairly normal.

gdkchan:
Fixed on #2601.

PR #2601, opened 2021-08-29
Avoid deleting textures when their data does not overlap.

riperiperi:
It's possible that while two textures start and end addresses indicate an overlap, that the actual data contained within them is sparse due to a layer stride. One such possibility is array slices of a cubemap at different mip levels - they overlap on a whole, but the actual texture data fills the gaps between each other's layers rather than actually overlapping.

This fixes issues with UE4 games having incorrect lighting (solid white screen or really dark shadows). There are still remaining issues with games that use the 3D texture prebaked lighting, such as THPS1+2, and games that use SUATOM/SURED like BD2 and Triangle Strategy.

This PR also fixes a bug with TexturePool's resized texture handling where the base level in the descriptor was not considered. This was causing data loss in some UE4 cubemaps.

#### Before

image

#### After

image

#### Before (on a bad day)
image

#### Before (on a good day)
image

#### After
image

### DIFF
--- a/Ryujinx.Graphics.Texture/SizeInfo.cs
+++ b/Ryujinx.Graphics.Texture/SizeInfo.cs
@@ -96,17 +96,18 @@ namespace Ryujinx.Graphics.Texture
 
         public IEnumerable<Region> AllRegions()
         {
-            for (int i = 0; i < AllOffsets.Length; i++)
+            if (_is3D)
             {
-                int offset = AllOffsets[i];
-
-                if (_is3D)
+                for (int i = 0; i < _mipOffsets.Length; i++)
                 {
-                    yield return new Region(offset, SliceSizes[i / _depth]);
+                    yield return new Region(_mipOffsets[i], SliceSizes[i]);
                 }
-                else
+            }
+            else
+            {
+                for (int i = 0; i < AllOffsets.Length; i++)
                 {
-                    yield return new Region(offset, SliceSizes[i % _levels]);
+                    yield return new Region(AllOffsets[i], SliceSizes[i % _levels]);
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


codbugpatrol:
I've rushed here to test this one, I've really thought that this was the one that maybe fixes the https://github.com/Ryujinx/Ryujinx/issues/2454 issue, but it isn't.  =(

codbugpatrol:
Jump Force is now rendering and we can see the players. Lot of visual glitches, but at least is not a pink screen ou black.
Great Job!

image
image
image

gdkchan:
Closes #2600 I guess?

I tested this on the problematic UE4 games I have here (except Yoshi since you already tested that).
Disaster Report 4, before (randomly, old screenshot):
image
Now:
image

No More Heroes 3 also seems fine now:
image
image
Before it would have corruption with either the screen being completely white or glowing objects.

Ryujinx.Graphics.Texture/SizeInfo.cs
```diff
@@ -1,4 +1,6 @@
 using System;
+using System.Collections;
```
gdkchan, line 2: I think this `using` is not necessary?

gdkchan (approved):
Nice catch 👍 
It's very nice to see this issue finally fixed. There's no reason to be afraid of UE4 games now (or maybe not yet, there's still the buffer mess).
LGTM other than the comments.

Torphedo:
Nice! I have a copy of A Hat In Time, so I'll test that out tomorrow. I haven't actually tried it in Ryujinx at all yet, so not sure if it's even problematic rn.

Thog (approved):

Ryujinx.Graphics.Texture/SizeInfo.cs
riperiperi: Yes, it was used temporarily at some point.

PR #2090, opened 2021-03-09
Implement Shader Instructions SUATOM and SURED

riperiperi:
Fixes morph target animation (used for facial animations, eyes) and anything else using atomic image store (probably on buffer textures) in UE4 games. Should affect BD2 and I think Trials of Mana. Yoshi's Crafted World uses these, but I'm not sure what for. May affect other games.

envytools and nvdisasm were used to determine the opcode formats. 

BD2 example:

https://user-images.githubusercontent.com/6294155/110537362-81b22280-811a-11eb-845b-bd67aba491f3.mp4

https://user-images.githubusercontent.com/6294155/110537430-91316b80-811a-11eb-91cf-aa62db561999.mp4

Note: Currently does not support FP or 64-bit types (which may not even exist for textures). Not sure if we should, but that's why the PR is currently draft.

PR #2605, opened 2021-08-30
Fix out-of-bounds shader thread shuffle

gdkchan:
Fixes some issues with thread shuffles, mainly shuffle up that was not taking negative threads IDs into account. If the source thread ID is negative after the subtraction by the index, it is *not* valid (but it was returning valid before because the comparison was unsigned, and 0xFFFFFFFF (-1) for example, is >= 0. I also made it *always* read the value from the source thread, rather than doing it conditionally, as the latter also caused graphical glitches.

Fixes vertex explosion on Marvel Ultimate Alliance 3.
Before (from gamedb):
image
Now:
image

When combined with #2090, also fixes block flickering on shadows. See screenshots below.
master:
image
(Lighting is broken, water is dark etc).
riperiperi's PR without this fix:
image
(Lighting works but flickers, see the lamp pole only being half lit, the water with weird colouring, etc).
riperiperi's PR with this fix:
image
(100% correct as far I can tell).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -40,7 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2546;
+        private const ulong ShaderCodeGenVersion = 2605;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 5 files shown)


Thog (approved):
nice catch

GamerzHell9137:
With the two PR's merged together in NoMoreHeroes 3, the game will render the "blue fog" correctly.
Its absent on Master.

Master:
image
PR's merged:
image

riperiperi (approved):
Very nice catch.

riperiperi:
image
Fixes the skater being too dark in the title screen of THPS 1+2, and some ingame lighting.

codbugpatrol:
Just for testing knowledge: This DOES NOT fix the green light on Jump Force or background character screen selection on Dragon Ball Fighter Z yet. That big vertex explosion on MK 11 is also present.  (I Know, I like fighting games)
image
image

codbugpatrol:
> Just for testing knowledge: This DOES NOT fix the green light on Jump Force or background character screen selection on Dragon Ball Fighter Z yet. That big vertex explosion on MK 11 is also present. (I Know, I like fighting games)

Did not know how to merge PRs, so I did not test this with SUATON / SUARED PR from @riperiperi

PR #2609, opened 2021-08-30
Add Linux Unicorn patch + desc.

mgielda:
As per the Discord conversation, adding some info on compiling Unicorn with the necessary patch on Linux.

### DIFF
--- a/Ryujinx.Tests.Unicorn/libs/README.md
+++ b/Ryujinx.Tests.Unicorn/libs/README.md
@@ -1,3 +1,22 @@
-The pre-compiled dynamic libraries in this directory are licenced under the GPLv2.
+# Unicorn
 
-The source code for windows/unicorn.dll is available at: https://github.com/MerryMage/UnicornDotNet/tree/299451c02d9c810d2feca51f5e9cb6d8b2f38960
+Unicorn is a CPU simulator with bindings in many languages, including
+C#/.NET.
+It is used by the Ryujinx test suite for comparative testing with its built-in
+CPU simulator, Armeilleure.
+
+## Windows
+
+On Windows, Unicorn is shipped as a pre-compiled dynamic library (`.dll`), licenced under the GPLv2.
+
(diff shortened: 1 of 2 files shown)


jduncanator:
Hmmm... This raises some interesting points. I've never noticed that QEMU (and then by extension Unicorn and the .NET wrapper) are all GPLv2 licensed. We will likely have to remove the Unicorn-based tests, as GPL is not compatible with the MIT license (only the other way around).

mgielda:
I am not a lawyer of course, but from my point of view you are using Unicorn only to test against, it is not delivering any of the Ryujinx features as you have a separate simulator. Of course separating the Unicorn tests out and potentially licensing them under GPL might be a good idea just to keep things clean. This stackexchange item:

https://opensource.stackexchange.com/questions/7503/implications-of-using-gpl-licenced-code-only-during-testing

seems to support this view.

So if keeping the Unicorn tests separate is something you are able to do without a bug overhead, it's probably a good idea.

jduncanator:
> I am not a lawyer of course, but from my point of view you are using Unicorn only to test against, it is not delivering any of the Ryujinx features as you have a separate simulator. Of course separating the Unicorn tests out and potentially licensing them under GPL might be a good idea just to keep things clean. This stackexchange item:
> 
> https://opensource.stackexchange.com/questions/7503/implications-of-using-gpl-licenced-code-only-during-testing
> 
> seems to support this view.
> 
> So if keeping the Unicorn tests separate is something you are able to do without a big overhead, it's probably a good idea.

I agree with your point of view as well. The plan is to move the CPU JIT and it's tests out of the main Ryujinx project anyway, so I don't have any issues merging this as is for now, and addressing that in the future.

Thanks for your work!

Thog (approved):

AcK77 (approved):

gdkchan (approved):
lgtm, but we should probably try to upstream those changes.

PR #2610, opened 2021-08-30
Update game metadata when stopping emulation

Nistenf:
Game metadata (specifically, time played staying at 0s was the obvious symptom) was not being updated when clicking Actions->Stop Emulation. This works as expected when exiting the emulator because UpdateGameMetadata is called from MainWindow.End, which is called from both Exit_Pressed  and Window_Close.

This PR updates the game metadata in this case

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1276,6 +1276,10 @@ namespace Ryujinx.Ui
 
         private void StopEmulation_Pressed(object sender, EventArgs args)
         {
+            if (_emulationContext != null)
+            {
+                UpdateGameMetadata(_emulationContext.Application.TitleIdText);
+            }
             RendererWidget?.Exit();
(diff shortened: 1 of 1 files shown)


Ryujinx/Ui/MainWindow.cs
```diff
@@ -1276,6 +1276,10 @@ private void Window_Close(object sender, DeleteEventArgs args)
 
         private void StopEmulation_Pressed(object sender, EventArgs args)
         {
+            if (_emulationContext != null)
+            {
+                UpdateGameMetadata(_emulationContext.Application.TitleIdText);
+            }
```
AcK77, line 1282: Add a new empty line after

Nistenf: Done!

AcK77 (approved):

riperiperi (approved):

PR #2612, opened 2021-08-31
Fix GTK3 mapping for single quote key

### LINKED ISSUE #2514: Quotes are not detecting in Ryujinx
## Bug Report

Ryujinx is not detecting Quotes even though it is defined.

Ryujinx is not detecting quotes(  '  ) when inputted.

You can first encounter this issue in the Input configuration where it won't detect quotes when reassigning keys.

### Log file
No Log

### Environment?

- Ryujinx version: 1.0.06cd3ab</br>
- System Specs:
  - OS: *Void Linux*
- Applied Mods : No

### Additional context?

I tested whether quotes are detecting by adding Console.WriteLine("Ok"); at the beginning of IsPressed() block and except for **quotes**, other cases showed the result **Ok** and When I created a custom profile called Me.json and Added Quote, eventhough it was displayed in Input Session I was not able to use it in games. I attached my finding along with this bug report.
Screenshot_20210729_220550
Screenshot_20210729_221616
Screenshot_20210729_221951
Screenshot_20210729_223642

Nistenf:
The single quote key (') was incorrectly mapped to the gtk key quotedbl. This fixes #2514

### DIFF
--- a/Ryujinx/Input/GTK3/GTK3MappingHelper.cs
+++ b/Ryujinx/Input/GTK3/GTK3MappingHelper.cs
@@ -134,7 +134,7 @@ namespace Ryujinx.Input.GTK3
             GtkKey.bracketleft,
             GtkKey.bracketright,
             GtkKey.semicolon,
-            GtkKey.quotedbl,
+            GtkKey.quoteright,
             GtkKey.comma,
             GtkKey.period,
             GtkKey.slash,


Nistenf:
The ' key can be correctly mapped in the controller configuration, and it works ingame

AcK77 (approved):

Thog (approved):
Nice catch, I may have missed that when I initially worked on it :sweat_smile:

emmauss (approved):

Narugakuruga:
Caps lock is also not properly mapped. Can this be fixed too?

Nistenf:
Is capslock supposed to be bindable? The GTK event passed to `OnKeyPress` has a key value of `VoidSymbol` when pressing capslock. This is on a Windows machine

Sakura0134:
Yep, that was the problem. @AcK77 Can you link #2514 with this pr?

Sakura0134 (approved):

PR #2613, opened 2021-09-01
Fix TXQ for 3D textures.

riperiperi:
Assumes the texture is 3D if the component mask contains Z.

This fixes a bug in UE4 games where parts of the map had garbage pointers to lighting voxels, as the lookup 3D texture was not being initialized. Most notable game is THPS1+2.

May need another PR to keep image store data alive and properly flush it in order using the AutoDeleteCache.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureTarget.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureTarget.cs
@@ -1,4 +1,5 @@
 using Ryujinx.Graphics.GAL;
+using Ryujinx.Graphics.Shader;
 
 namespace Ryujinx.Graphics.Gpu.Image
 {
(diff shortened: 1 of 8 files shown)


riperiperi:
Before:
image

After:
image

RyzenDew:
image
fixes outlast 2

Thog (approved):
nice catch

codbugpatrol:
Wow! Rip did it again!

This finally closes: https://github.com/Ryujinx/Ryujinx/issues/2454
image
image
image

codbugpatrol:
(I've deleted my last comment because, the testing was done in the wrong PR)

codbugpatrol:
This PR creates a vertex explosion on Jump Force now.

image
image

riperiperi:
Run the game again, I think it's unlikely that the vertex explosion is caused by writing to a 3d texture.

gdkchan:
I think a better way to fix this is changing `QueryIsTextureBuffer` to `QueryTextureTarget` and returning the target there, on the GPU acessor. Then TXQ would call it to get the actual texture target. The way it is right now would fail for 2D array textures for example, as it would assume it is 3D. I'm not sure how that would play with the shader cache though, but I think only the shaders that use TXQ would need to be invalidated (which shouldn't be many).

Also, if you do this change, then `accurateType` can be completely removed from `SetUsedTextureOrImage`. The only reason it exists right now is because the TXQ target may be incorrect, so it's there to prefer target from actual texture sampling operations (which does provide a target).

codbugpatrol:
@riperiperi , you were right. No regression here then. Thanks. 
But I did not have any other games in UE4 that might helping in the testing. The ones I've tested with are most fighting games, I didn't notice any changes, which is good too.

image

gdkchan (approved):
lgtm, thanks! Nice catch.

PR #2617, opened 2021-09-03
Log shader compile errors with Warning level

riperiperi:
These are infrequent enough that I think it's worth dumping any errors into the log. They also keep causing graphical glitches, and the only indication that anything went wrong is a debug log that is never enabled.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Program.cs
+++ b/Ryujinx.Graphics.OpenGL/Program.cs
@@ -115,9 +115,8 @@ namespace Ryujinx.Graphics.OpenGL
 
             if (status == 0)
             {
-                // Use GL.GetProgramInfoLog(Handle), it may be too long to print on the log.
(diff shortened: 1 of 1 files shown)


gdkchan:
I think its fine to log it under warning, but might be a good idea to check if the error message is too large, and truncate it in those cases.

Thog:
@riperiperi What should we do about this PR? Should it be closed, or do you want to integrate @gdkchan suggestion and mark it as ready?

riperiperi:
Updated with a character limit. Things are a bit different now with Vulkan - it already logs compile errors when the GLSL compiler is active, but we use SPIR-V anyways. Not sure if there's a meaningful way to log errors from pipeline compiles across vendors.

AcK77 (approved):

TSRBerry (approved):

PR #2623, opened 2021-09-07
Account for negative strides on DMA copy

gdkchan:
Some OpenGL games (such as https://github.com/Ryujinx/Ryujinx-Games-List/issues/3704) are using negative stride values. This would cause the copy to advance backwards, which I imagine might be useful if the copy is also supposed to flip the image vertically. This should actually "just work" with our simple offset calculation, but does not due to the fact that the current code uses spans, and the code to get the span is not aware of the fact the stride (and consequently the size) might be negative.

To account for that, the following changes were made.
- `Math.Abs` is used on the stride to calculate the size, to ensure it is positive.
- If stride is negative, the base offset is adjusted to the real start offset of the copy.

Nothing else needs to be changed, as the base pointers and the offsets returned by the offset calculator already accounts for negative strides (the former will add an offset to make the pointer point to the last line of the image, while the latter already returns negative offsets, that when added to the pointer will fall into the correct location).

While I was at it, I also changed the flush call on `InlineToMemory` to use the GPU memory manager rather than the physical one, to account for non-contiguous memory.

With this change the game I mentioned before (IdolDays) no longer crashes when trying to open the log or load/save the game, although it does have other issues there that don't seem to be caused by this change (for example, texture/buffer flush for the thumbnails seems unreliable).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -76,6 +76,13 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
         {
             if (linear)
             {
+                // If the stride is negative, the texture has to be flipped, so
+                // the fast copy is not trivial, use the slow path.
+                if (stride <= 0)
+                {
+                    return false;
+                }
+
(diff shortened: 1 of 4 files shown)


Thog (approved):
lgtm

riperiperi (approved):
Nice catch.

PR #2624, opened 2021-09-07
Fast path for Inline2Memory buffer write that skips write tracking

riperiperi:
This PR adds a method to PhysicalMemory that attempts to write all cached resources directly, so that memory tracking can be avoided. The goal of this is both to avoid flushing buffer data, and to avoid raising the sequence number when data is written, which causes buffer and texture handles to be re-checked.

It just uses the existing ForceDirty method that is used for updating buffer data with the ConstantBufferUpdater. This avoids going through the tracking structure, reprotection and triggering read actions. The last part is the most important - as we know the exact range being written to, we don't need to flush the entire pages that we're writing to - just punch a hole in the "modified" ranges and write the data directly.

This currently only targets buffers, with a side check on textures that falls back to a tracked write if any exist within the target range. It's not expected to write textures from here - this is just a mechanism to protect us if someone does decide to do that. It's possible to add a fast path for this in future (and for ShaderCache, once that starts using tracking)

The forced read before Inline2Memory begins has been skipped, as the data is fully written when the transfer is completed anyways. This allows us to flush on read in emergency situations, but still write the new data over the flushed data.

Improves performance on Xenoblade 2 and DE, which were flushing buffer data on the GPU thread when trying to write compute data. May improve performance in other games that write SSBOs from compute, and update data in the same/nearby pages often.

Super Smash Bros Ultimate should probably be tested to make sure the vertex explosions haven't returned, as I think that's what this AdvanceSequence was for.

### Before
image

### After
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
@@ -110,9 +110,6 @@ namespace Ryujinx.Graphics.Gpu.Engine.InlineToMemory
 
             ulong dstGpuVa = ((ulong)state.OffsetOutUpperValue << 32) | state.OffsetOut;
 
-            // Trigger read tracking, to flush any managed resources in the destination region.
-            _channel.MemoryManager.GetSpan(dstGpuVa, _size, true);
-
             _dstGpuVa = dstGpuVa;
             _dstX = state.SetDstOriginBytesXV;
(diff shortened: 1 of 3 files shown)


gdkchan (approved):
lgtm, thanks!
I also tested a few games that could be potentially affected negatively by this change (MHR and OpenGL games), and they are still working fine.

riperiperi:
Requires #2641 to avoid a performance regression in SMO.

AcK77 (approved):

Issue #1125, opened 2020-04-16
Separate ARM translator from IR and code generation


gdkchan:
Currently the CPU emulator can be thought in two parts:
- ARM code -> IR translation: That's the code inside the Instructions folder, it is used to translate the ARM code to IR that we can later optimize and turn into x86 code.
- IR optimization and X86 backend: That's what actually generates the machine code.

I think they should be separated into 2 projects. First because having a IR -> x86 compiler might be useful for a lot of different things. We can use it to easily generate code at runtime, implement JIT for other CPUs if we want etc. Second it makes testing easier, currently we have tests for the ARM instructions, but we would also benefit from tests to individual IR instructions.

Thaodan:
Would this prepare for ARM support e.g. running the emulator on a mobile Linux OS?

gdkchan:
No, it would allow the JIT to be used by other parts of the project that also need runtime code generation (such as GPU macro emulation, however this already uses .NET JIT, it could be useful in the future if we decide to use a AOT approach where .NET JIT is not available).

FICTURE7:
I am interested into doing this, unless someone else is already working it. I think `Compiler`, `CompilerOptions`, `CompilerContext` can be moved to `ARMeilleure.CodeGen` and `EmitterContext`, `ControlFlowGraph` can be moved to `ARMeilleure.IntermediateRepresentation`. Some analysis can be moved there as well, like SSA construction/deconstruction perhaps under a new namespace `ARMeilleure.CodeGen.Analysis`. There is also PTC, I guess it should be included with the backend?

AcK77:
@FICTURE7 Could you contact me on discord please 😊?

FICTURE7:
I do not have a discord at the moment, maybe I should make one some of these heh. You can reach me through email (whenever I check them <_< atleast).

gdkchan:
> 
> 
> I am interested into doing this, unless someone else is already working it. I think `Compiler`, `CompilerOptions`, `CompilerContext` can be moved to `ARMeilleure.CodeGen` and `EmitterContext`, `ControlFlowGraph` can be moved to `ARMeilleure.IntermediateRepresentation`. Some analysis can be moved there as well, like SSA construction/deconstruction perhaps under a new namespace `ARMeilleure.CodeGen.Analysis`. There is also PTC, I guess it should be included with the backend?

[...]

PR #2625, opened 2021-09-08
Refactor `PtcInfo`

FICTURE7:
This change reduces the coupling of `PtcInfo` and the backend by moving relocation tracking to the backend. `RelocEntry`s remains as `RelocEntry`s through out the pipeline until it actually needs to be written to the PTC streams. Keeping this representation makes inspecting and manipulating relocations after compilations less painful. This is something I needed to do to patch relocations to 0 to diff dumps.

I think we should get rid of `CodeGenContext` and move jump management to the `Assembler`. Then add an `Assembler.MarkLabel(Operand)`, handle the `OperandKind.Label` case in `Assembler.Jcc/Jmp` and enable branch tightening when PTC is enabled.

Contributes to #1125.


gdkchan:
> I think we should get rid of `CodeGenContext` and move jump management to the `Assembler`. Then add an `Assembler.MarkLabel(Operand)`, handle the `OperandKind.Label` case in `Assembler.Jcc/Jmp` and enable branch tightening when PTC is enabled.

That would be nice, it is a bit unfortunate that we generate slightly worse code with PPTC enabled right now. Speaking about that, I wonder if it's possible to keep constant folding enabled for those relocatable pointers too. It should be possible by using a relative relation, so it would subtract the base pointer before saving, and adding the base pointer upon relocating on load. The problem is that it needs to know the base pointer to do that...

The change itself looks fine to me (I'll look at it more troroughly later today), I was never a fan of the coupling of PPTC with the  JIT, although at least it's fairly minimal right now.

FICTURE7:
I'll do that change in a PR that optimizes the `Assembler`. Apparently we spend as much time writing X86 code as allocating registers in the `HybridAllocator`.

> The problem is that it needs to know the base pointer to do that...

Right, `Symbol`'s size would probably be increased by a couple of bytes, which I think makes moving it out into a separate per compilation table worth it. `Operand` can refer to the symbol using its index in that table.

gdkchan (approved):
lgtm, thanks! 👍

LDj3SNuD (approved):

PR #2627, opened 2021-09-09
Use shader subgroup extensions if shader ballot is not supported

gdkchan:
Uses new subgroup extension (that AFAIK were introduced for SPIR-V) if the old `ARB_shader_ballot` extension is not supported. This is required for Vulkan. Additionally, this benefits Intel on OpenGL as they don't support the `ARB_shader_ballot` extension on the proprietary driver for some reason, but does support the new extensions.

Games that makes use of shuffles and lane ID access and masks should work better on Intel Windows OpenGL now.
For example, Astral Chain works on Intel OpenGL now.
Before:
image
After:
image

Other vendors should not be affected as they support the shader ballot extension (which is still preferred).

### DIFF
--- a/Ryujinx.Graphics.GAL/Capabilities.cs
+++ b/Ryujinx.Graphics.GAL/Capabilities.cs
@@ -9,6 +9,7 @@ namespace Ryujinx.Graphics.GAL
         public bool SupportsImageLoadFormatted { get; }
         public bool SupportsMismatchingViewFormat { get; }
         public bool SupportsNonConstantTextureOffset { get; }
+        public bool SupportsShaderBallot { get; }
         public bool SupportsTextureShadowLod { get; }
         public bool SupportsViewportSwizzle { get; }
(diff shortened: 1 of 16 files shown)


Thog (approved):

riperiperi (approved):
LGTM. With the way the capability is worded, I guess using the ARB methods is still preferred when both are available?

gdkchan:
> 
> 
> LGTM. With the way the capability is worded, I guess using the ARB methods is still preferred when both are available?

Yes, as I mentioned in the description. I'm not aware of any benefit of using `KHR_shader_subgroup_basic`/`ballot` if `ARB_shader_ballot` is available, and it's better to not risk breaking things in this case by changing the extension imo. There is also `KHR_shader_subgroup_shuffle` that can be used for efficient vendor agnostic shuffle operations, but it does not have the `threadIdValid` output parameter like the NVIDIA extension, so anything using it would still need to compute that.

PR #2629, opened 2021-09-09
Allow textures to have their data partially mapped

### LINKED ISSUE #2463: Miitopia Crashing at a consistent point.

Ryujinx is crashing at a consistent spot in Miitopia.</br>
I've left Ryujinx on stock settings aside from controls.</br>

Start Miitopia with a save that has not cleared the underground maze yet. Play normally until you get to a gate in said underground maze and try to proceed. Your party will walk up, you will hear one of them talk but it will freeze before any speech bubble comes up. After that it just closes the entire emulator.

### Log file

[Ryujinx_1.0.6959_2021-07-11_08-56-35.log](https://github.com/Ryujinx/Ryujinx/files/6796953/Ryujinx_1.0.6959_2021-07-11_08-56-35.log)</br>

### Environment?

- Ryujinx version: 1.0.6959 </br>
- Game version: 1.0.3 </br>
- System Specs:
  - OS: Windows 10 Pro
  - CPU: Intel Core I7
  - GPU: GTX 1660 TI
  - RAM: 2933 MHZ (up to 32GB) DDR4
- Applied Mods : [ No ]

### Additional context?

[ N/A ]

### LINKED ISSUE #3077: Star Ocean FDR Crashing

 In almost the beginning of the game, the first area when you get let outside the starting city, it crashes right after successfully winning a battle and the screen is transitioning back to the map. This doesn't happen 100% of the time, though, I'd say around every 1 in 6 battles I experience it.

I only changed the screen to docked mode

1. Enter a battle outside of town 2. Finish the battle and wait for the screen to transition back

### Log file

[Ryujinx_1.1.16_2022-02-03_09-49-49.log](https://github.com/Ryujinx/Ryujinx/files/7995769/Ryujinx_1.1.16_2022-02-03_09-49-49.log)

### Environment?

- Ryujinx version: 1.1.16
- Game version: 1.0.0
- System Specs:
  - OS: Windows 10 Pro 20H2
  - CPU: Ryzen 7 5800x
  - GPU: RTX 3080 FE
  - RAM: 16GB
- Applied Mods : None

gdkchan:
We are currently hiting a case in some games where the texture data is only partially mapped, so for example, some mipmap levels are not mapped, which causes an exception to be throw on the call to `GetPhysicalRegions`, as it expects the full texture region to be mapped.

This change allows the textures to be created even if their data is not fully mapped. For the ranges that are not mapped, the data is read as 0, and writes goes nowhere. Mapped regions are read/written as normal. This should match hardware behaviour, which needs to work this way most likely to support sparse textures, as unmapped regions being accessed shouldn't cause a crash.

A few methods were adjusted to take this into account:
- `GetPhysicalRegions` now returns `ulong.MaxValue` and the size for unmapped ranges, instead of throwing.
- `CompareRange` can now handle `ulong.MaxValue` unmapped ranges, by checking if the GPU VA range is also unmapped.
- All methods on the `PhysicalMemory` that takes a `MultiRange` now ignores `ulong.MaxValue` ranges (except the read method, which now effectively reads them as 0).
- `TextureGroup` also now ignores `ulong.MaxValue` ranges when "generating handles".
- `MemoryRange` overlap check now returns false if any of the range overflows (which is the case when the address is `ulong.MaxValue`, which is the special value used for unmapped ranges).

**Open questions:**
- Should we print a warning for partially mapped textures? In some cases, it might be caused by a bug on the emulator, so perhaps having a warning could be useful.
- Should we handle cases where the *start* address is unmapped? My opinion is yes, but right now the fast path just looks for the base address and tries to find a perfect match, or returns null if unmapped, so changing that could degrade performance.
- With the current approach, it is still possible to write the data from the GPU, and it would read as whatever was written there later. This does not match hardware behaviour. Is it worth addressing this? It would require using sparse textures on the host, and I haven't looked at what vendor/GPU support for this extension looks like.

**What to test:** Games that had the `InvalidMemoryRegionException: The specified GPU virtual memory range [...] is not fully mapped.` crash, and games that torture the texture cache like UE4 games would be good candidates.

Fixes #2463, fixes #3077.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -40,6 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         private readonly PhysicalMemory _physicalMemory;
 
         private readonly MultiRangeList<Texture> _textures;
+        private readonly HashSet<Texture> _partiallyMappedTextures;
 
         private Texture[] _textureOverlaps;
         private OverlapInfo[] _overlapInfo;
@@ -57,6 +58,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             _physicalMemory = physicalMemory;
 
             _textures = new MultiRangeList<Texture>();
+            _partiallyMappedTextures = new HashSet<Texture>();
 
             _textureOverlaps = new Texture[OverlapsBufferInitialCapacity];
             _overlapInfo = new OverlapInfo[OverlapsBufferInitialCapacity];
@@ -74,17 +76,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             Texture[] overlaps = new Texture[10];
             int overlapCount;
 
-            MultiRange unmapped;
-
-            try
-            {
-                unmapped = ((MemoryManager)sender).GetPhysicalRegions(e.Address, e.Size);
-            }
-            catch (InvalidMemoryRegionException)
-            {
-                // This event fires on Map in case any mappings are overwritten. In that case, there may not be an existing mapping.
-                return;
-            }
+            MultiRange unmapped = ((MemoryManager)sender).GetPhysicalRegions(e.Address, e.Size);
 
             lock (_textures)
             {
@@ -94,6 +86,24 @@ namespace Ryujinx.Graphics.Gpu.Image
             for (int i = 0; i < overlapCount; i++)
             {
                 overlaps[i].Unmapped(unmapped);
+            }
+
+            // If any range was previously unmapped, we also need to purge
+            // all partially mapped texture, as they might be fully mapped now.
+            for (int i = 0; i < unmapped.Count; i++)
+            {
+                if (unmapped.GetSubRange(i).Address == MemoryManager.PteUnmapped)
+                {
+                    lock (_partiallyMappedTextures)
+                    {
+                        foreach (var texture in _partiallyMappedTextures)
+                        {
+                            texture.Unmapped(unmapped);
+                        }
+                    }
+
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


johnwhite5116:
Zero no Kiseki does not crash anymore but it's frozen with black screen and won't proceed.
[Zero with PR.log](https://github.com/Ryujinx/Ryujinx/files/7140178/Zero.with.PR.log)
froze
console

gdkchan:
The issue on this game is a bit more involved. It has a very large 2D array texture that takes a long time to upload the data to GPU. It also uses a lot of memory for flush here (and I get OOM). That explains why it is partially mapped. The texture upload speed can be improved (I did so locally for testing). Improving the flushes will be more complicated though, perhaps we should do partial flushes like we do partial sync.

Anyway, with this, the fast data uploads, flushes disabled + #2630 I can get the game to work:
image

bylaws:
I recommend keeping track of if mappings were mapped by nvdrv as sparse in the PTE, then print a warning only  if they weren't.

Basically if AllocSpace is called with flags & 2 map the region as sparse, track this, then in UnmapBuffer and the `!handle` case in Remap map the region back as sparse.

https://github.com/skyline-emu/skyline/blob/nvdrv-gpu-misc/app/src/main/cpp/skyline/services/nvdrv/devices/nvhost/as_gpu.cpp

Ctrl-F `sparse` here

riperiperi:
Just duplicating feedback here in case it is lost somehow:

> maybe the MultiRangeList could just ignore ranges that are invalid, on lookup and insertion
> so the overlaps check only counts mapped ranges

>it looks like some weird behaviour is happening in the IntervalTree, but it might work in our favor:
>
>- 0xFFFFFFFFFFFFFFFF base address entries appear at the rightmost node in the tree. (expected)
>- the "Max" entry in the tree nodes (used for determining how far to search to the left) can only be influenced by the EndAddress. If the size of an invalid range is > 0, then this will always be very low, as it'll overflow. 
>- searching for invalid ranges does not work, as the overlap check fails. the above could be contributing to this
>
>things will probably work as they are now, but I'm still for explicitly ignoring ranges with the invalid base address in MultiRangeList, rather than ignoring them due to the numbers breaking down

gdkchan:
I have added explicit checks to `MultiRangeList` to ignore memory ranges that are unmapped (those with ulong.MaxValue address). I'm not a huge fan of the way how those unmapped ranges are represented right now, but the alternative is adding another field on the struct just for that, and imo that is worse.

Testing to confirm this is still fixing the crash on Miitopia would be appreciated, I can't test this game easily right now.

GamerzHell9137:
Fixes the issue in Miitopia 
image

riperiperi (approved):
Looks good from a code perspective. Probably worth testing with some unreal engine games since they have a fun time with mappings.

Thog (approved):
code lgtm

Issue #2628, opened 2021-09-09
The Legend of Heroes: Zero no Kiseki


johnwhite5116:
### What's the issue you encountered?

The Legend of Heroes: Zero no Kiseki crashes when launching.

### Log file

[Zero.log](https://github.com/Ryujinx/Ryujinx/files/7139113/Zero.log)

### Environment

- Ryujinx version: 1.0.7024
- Game version: 1.0.1
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 9 3900X
  - GPU: Nvidia RTX 3090
  - RAM: 32GB
- Applied Mods : No

### Additional context

Interestingly, The Legend of Heroes: Ao no Kiseki seems bootable and playable, which is released later than Zero no Kiseki and also using same game engine.

![Ao](https://user-images.githubusercontent.com/75278127/132747191-5302b0f6-37cd-45f2-b484-1fb2153a9809.png)

PR #2630, opened 2021-09-10
Attach a single layer for FBs with mismatching targets

gdkchan:
While looking at the issues on #2628, I found a "framebuffer is not complete" type of error, so I tried calling `glCheckFramebufferStatus` and it returned this:
> GL_FRAMEBUFFER_INCOMPLETE_LAYER_TARGETS is returned if any framebuffer attachment is layered, and any populated attachment is not layered, or if all populated color attachments are not from textures of the same target. 

Basically different attachment targets is not supported, you can't have a 2D and 2D array texture attached for example. This change forces it to attach only the base layer if there is a mismatch, which avoids the error.

Additionally I have made some optimizations to the `TextureCreateInfo` struct access, as the codegen for the newly introduced `IsLayered` property was pretty bad due to struct copies.


Thog (approved):
lgtm

riperiperi (approved):
LGTM, nice catch.

gdkchan:
Going to close this as #2646 solves the issue on the only game that is known to be affected by this change in a much better way, and I don't like doing this kind of change without a test case. It is worth noting that layered depth buffers are not supported right now, I think should be looked at separately as it could cause this kind of framebuffer completness issue (as the color target would be an array, but the depth buffer wouldn't).

PR #2632, opened 2021-09-10
Remove motion filter reset

mpnico:
The motion filter is reset on a regular basis in order to avoid drift when the controller is in a rest position. This behavior has a downside (at least) on PS4/PS5 controller in native SDL motion on Splatoon 2, where your aim is reset to the ground after the filter reset. It does work properly with Pro Controller though.

Removing the reset fix Splatoon 2 and doesn't seems to break other controller or other game but it needs to be tested on various configurations.

### DIFF
--- a/Ryujinx.Input/Motion/MotionInput.cs
+++ b/Ryujinx.Input/Motion/MotionInput.cs
@@ -39,8 +39,6 @@ namespace Ryujinx.Input
 
                         Rotation = Vector3.Zero;
 
-                        _filter.Reset();
-
                         _calibrationFrame = 0;
                     }
                 }


Ryujinx.Input/Motion/MotionInput.cs
```diff
@@ -39,8 +39,6 @@ public void Update(Vector3 accel, Vector3 gyro, ulong timestamp, int sensitivity
 
                         Rotation = Vector3.Zero;
 
-                        _filter.Reset();
```
emmauss, line 42: if this is being removed, then the rest of the reset can be removed

Thog:
@mpnico Any plans on working on this again?

mpnico:
Unfortunatly I have no free time since a few weeks but I'd like to fix this before the end of the year.

AcK77:
Closed due to inactivity (after some DM with the author)

PR #2507, opened 2021-07-26
Avalonia UI

emmauss:
Spanking new ui using avalonia. 

For the most part, this is a port of the gtk ui, with some avalonia specific changes here and there. The controller settings window has been merged with the settings window, and now has a new design which some of you may be familiar with. A new game load screen for showing pptc and cache progress has been added. There are other stuff that I may have forgotten, but it should be there somewhere.

Screenshot from 2021-07-26 16-19-05
Screenshot from 2021-07-26 16-19-09
Screenshot from 2021-07-26 16-19-36
Screenshot from 2021-07-26 16-27-54
Screenshot from 2021-07-26 16-28-07
Screenshot from 2021-07-26 16-28-46
Screenshot from 2021-07-26 16-32-00

There are a few issues with avalonia, and some have been reported and others we had to work around with.

- Although Avalonia has an opengl widget, they do not have a native opengl widget, and instead, draws what you have rendered unto a texture and draws that instead in a control during their render pass. Because we need full control over the opengl context, and also because we would need something that supports vulkan, we are using an embedded window to provide the opengl context and vulkan surface. This is what most other ui toolkits provide out of the box, but is not available on Avalonia.
- Due to the above, focus on the window can be finicky at times, due to how different os handles child windows.

This is currently beta and is open for testing and design review.

Currently Reported Issues

- [x] Tab crashes the emulator while game is running
- [x] Hide UI is not implemented
- [x] Text in ui may be vertically misaligned
- [x] System version on taskbar text isn't aligned properly and not hidden while loading cache
- [x] Line under Memory Manager Mode is way too thick/big
- [x] Windows can be resized below the minimum
- [x] Exiting Fullscreen mode is impossible until you click the screen after going into fullscreen mode
- [x] Game control should keep focus
- [x] Changing the aspect ratio from the taskbar is not working (and when changing in settings window)
- [x] Controller once configured can't be used (DS4 controller)
- [x] Setting 2nd player crashes the emulator
- [x] Loading previous input profile crashes the emulator
- [x] Editing input settings in-game will cause the emulator to not register any inputs even after saving until the game has been restarted.
- [x] Seperators in setting window is too large and conspicuous.
- [x] Make sections in settings more noticeable from other sections. 
- [x] remove max resize limit in input config

Closes #1103, closes #1114, closes #1312, closes #1699, closes #2112, closes #2810, closes #2975, closes #3053.

PR #2633, opened 2021-09-11
gui: Replace FileChooserDialog with FileChooserNative

### LINKED ISSUE #2530: The game directory cannot be specified for the drive where the pagefile.sys exists

When I select the drive where the page file exists, I get an error and cannot select the game directory.

err

This error also occurs in the folder selection dialog that appears when you install the firmware.

1. In the settings on the Windows side, set the page file on the drive to be tested.
1. On the Ryujinx side, call the game directory selection dialog from the settings and select the drive specified in 1. in the tree view.
1. An error dialog is displayed and the game directory cannot be selected.

### Environment?

- Ryujinx version: 1.0.6982 (Probably also happens in older versions)
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 5 3600
  - GPU: NVIDIA GeForce GTX 1660 Ti
  - RAM: 16GB
- Applied Mods : No

### LINKED ISSUE #2584: [Feature Request] Change GUI popup when adding/locating files
### What feature are you suggesting?
Change GUI popup when adding/locating files 
#### Overview:
I found the Ryujinx file addition/location GUI popup to be rather tedious to use. Especially when it came to finding folders that were multiple directories deep, or if the file in question is located inside a “hidden folder.” 

I recommending  changing to the windows file addition/location GUI popup. I have attached an image below of what this popup looks like for the program "Snagit Editor" just in case. 
#### Smaller Details: 
N/A
#### Nature of Request:  
- Change
Using the windows GUI, would make finding files a breeze. Having access to the address bar would make copying and pasting directory quick and easy. The ability to go up and down directories from the address bar is also a fast way to navigate folders. Any files in the users “Quick Access” toolbar would also be readily available in the left column. 

Windows GUI

AcK77:
We currently use the FileChooser from GTK, which is a bit mess. Instead of it we could use the native FileChooser from all specifics OS. This is what this PR attempt to fix.

image

It could be nice to get a test under linux since I've only tested it under Windows without any issues.

Fixes #2584 and fixes #2530

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1201,15 +1201,20 @@ namespace Ryujinx.Ui
 
         private void Load_Application_File(object sender, EventArgs args)
         {
-            using (FileChooserDialog fileChooser = new FileChooserDialog("Choose the file to open", this, FileChooserAction.Open, "Cancel", ResponseType.Cancel, "Open", ResponseType.Accept))
+            using (FileChooserNative fileChooser = new FileChooserNative("Choose the file to open", this, FileChooserAction.Open, "Open", "Cancel"))
             {
-                fileChooser.Filter = new FileFilter();
-                fileChooser.Filter.AddPattern("*.nsp");
-                fileChooser.Filter.AddPattern("*.pfs0");
-                fileChooser.Filter.AddPattern("*.xci");
-                fileChooser.Filter.AddPattern("*.nca");
-                fileChooser.Filter.AddPattern("*.nro");
-                fileChooser.Filter.AddPattern("*.nso");
+                FileFilter filter = new FileFilter()
+                {
+                    Name = "Switch Executables"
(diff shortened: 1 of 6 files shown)


edisionnano:
No regressions observed in Linux

MutantAura:
Is this not already fixed by #2507 ?

AcK77:
It is, but since Avalonia is a large change, there is more chances to get this PR merged faster than Avalonia and so issues will be fixed faster too.

Rstar284:
can say that is works just fine on Linux

emmauss (approved):

RyzenDew:
Can we select multiple folders to add ? If so I'm down cause gtk can do that and its useful

AcK77:
> Can we select multiple folders to add ? If so I'm down cause gtk can do that and its useful

Yes

Thog (approved):

PR #2634, opened 2021-09-12
hos: Cleanup the project

AcK77:
Since a lot of changes has been done on the HOS project, there are some leftover here and there, class just used in one service, things at wrong places, and more.
This PR fixes that, additionnally to that, I've realigned some vars because I though it make the code more readable.


Ryujinx.HLE/Exceptions/InvalidStructLayoutException.cs
```diff
@@ -7,9 +7,8 @@ public class InvalidStructLayoutException<T> : Exception
     {
         static readonly Type _structType = typeof(T);
 
-        public InvalidStructLayoutException(string message) : base(message) {}
+        public InvalidStructLayoutException(string message) : base(message) { }
         
-        public InvalidStructLayoutException(int expectedSize) :
```
gdkchan, line 12: I prefer how it was before, otherwise the line becomes too long and you need to scroll horizontally to read it.

Ryujinx.HLE/Exceptions/ServiceNotImplementedException.cs
```diff
@@ -18,38 +18,29 @@ internal class ServiceNotImplementedException : Exception
         public ServiceCtx Context { get; }
         public IpcMessage Request { get; }
 
-        public ServiceNotImplementedException(IpcService service, ServiceCtx context)
```
gdkchan, line 21: I prefer how it was before, otherwise the line becomes too long and you need to scroll horizontally to read it.

gdkchan (commented):
Here's a general comment about the alignment:

While I don't mind it and agree that it makes the code easier to read in certain cases, it is a bit counter-productive as it doesn't play nice with the IDE and formatting tools. For example, Visual Studio will want to de-align them if the file is for some reason auto-formatted. Also if a type is renamed for example, everything needs to be manually re-aligned on all places where it is used. In general I think it's better to not bother as we save time not needing to manually align the code constantly.

AcK77:
> it is a bit counter-productive as it doesn't play nice with the IDE and formatting tools.
Yeah I know, that's why I don't want to bother other contributors in PR but I still prefere to add them manually sometimes when I do some cleanup or refactoring, there is no big deal of doing that ;)

Ryujinx.HLE/Loaders/Mods/IPSwitchPatcher.cs
```diff
@@ -7,11 +7,19 @@ namespace Ryujinx.HLE.Loaders.Mods
...
+
+        private enum Token
+        {
+            Normal,
+            String,
+            EscapeChar,
+            Comment
+        }
+
         readonly StreamReader _reader;
```
Thog, line 20:
```suggestion
        private readonly StreamReader _reader;
```

Thog (approved):

gdkchan (approved):
lgtm, thanks.

PR #2637, opened 2021-09-13
Optimize `HybridAllocator`

FICTURE7:
This brings a set of optimizations to the `HybridAllocator` which increases the throughput of LCQ, and usually shaves a couple of seconds from boot time. I also disabled `BlockPlacement` in LCQ to get some extra throughput since with the host mapped memory manager, there is a lot less cold code to split from hot code.

### DIFF
--- a/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
@@ -3,6 +3,7 @@ using ARMeilleure.Translation;
 using System;
 using System.Diagnostics;
 using System.Numerics;
+using System.Runtime.CompilerServices;
 using static ARMeilleure.IntermediateRepresentation.Operand.Factory;
 using static ARMeilleure.IntermediateRepresentation.Operation.Factory;
 
@@ -10,9 +11,6 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
 {
     class HybridAllocator : IRegisterAllocator
     {
-        private const int RegistersCount = 16;
-        private const int MaxIROperands  = 4;
-
         private struct BlockInfo
         {
             public bool HasCall { get; }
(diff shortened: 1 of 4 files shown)


ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
```diff
@@ -86,6 +86,11 @@ public AllocationResult RunPass(ControlFlowGraph cfg, StackAllocator stackAlloc,
             var localInfo = new LocalInfo[cfg.Blocks.Count * 3];
             int localInfoCount = 0;
 
+            void ThrowNotVisisted()
+            {
+                throw new InvalidOperationException("Local was not visisted yet. Used before defined?");
```
Mou-Ikkai, line 91: Fix typo: "Local was not visisted yet." -> "Local was not visited yet."
Comparison of word: visisted -> visited.

FICTURE7, line 91:
```suggestion
                throw new InvalidOperationException("Local was not visited yet. Used before defined?");
```

ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
```diff
@@ -86,6 +86,11 @@ public AllocationResult RunPass(ControlFlowGraph cfg, StackAllocator stackAlloc,
             var localInfo = new LocalInfo[cfg.Blocks.Count * 3];
             int localInfoCount = 0;
 
+            void ThrowNotVisisted()
```
Mou-Ikkai, line 89: Same here on these two.
ThrowNotVisisted -> ThrowNotVisited
```suggestion
            void ThrowNotVisited()
```

ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
```diff
@@ -105,7 +110,7 @@ void SetVisited(Operand local)
 
                 if (!IsVisited(local))
                 {
-                    throw new InvalidOperationException("Local was not visisted yet. Used before defined?");
+                    ThrowNotVisisted();
```
Mou-Ikkai, line 113: Same here on these two.
ThrowNotVisisted -> ThrowNotVisited
```suggestion
                    ThrowNotVisited();
```

ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
```diff
@@ -86,6 +86,11 @@ public AllocationResult RunPass(ControlFlowGraph cfg, StackAllocator stackAlloc,
             var localInfo = new LocalInfo[cfg.Blocks.Count * 3];
             int localInfoCount = 0;
 
+            void ThrowNotVisited()
+            {
+                throw new InvalidOperationException("Local was not visited yet. Used before defined?");
+            }
+
```
jduncanator, line 93: Given this is only used once, I'd rather just keep the throw inline.

FICTURE7: I was hoping this would reduce the IL size enough to convince the inliner it is a good idea to inline `GetLocalInfo`, but apparently not and neither is `IsVisited` inlined. Quite unfortunate given how trivial they are; a lot of code (relative to the actual function size) is emitted to handle the ABI. If only we could annotate these functions with `AggressiveInlining`😩.

I think the new IR is less inliner friendly, it requires 2 `ldfld` to load an actual field (one for the `Data*` and one for the actual field in `Data*`). The old IR required only a single `ldfld`.

Anyways, splitting the throw into a different function nonetheless still saves a couple of pushes and pops in the prologue and epilogue. I think I'll move these methods out into the scope of the class and mark them as aggressive inlining.

gdkchan: I wonder why this one is not using `Debug.Assert`, since this case is something that should never happen. Could be worth changing it if the method is hot, just to avoid the check.

FICTURE7: Yeah I think using an assert here sounds good, although it was not enough to get inlined IIRC. 😞

jduncanator: I'm happy with the assert too :)

gdkchan (approved):
lgtm, thanks. Great improvements for boot times with PPTC off on some titles!

AcK77 (approved):

PR #2638, opened 2021-09-13
Adjustments to framerate metric and addition of frametime

MutantAura:
The current framerate indictor is a weighted average which uses a decay rate of 0.5 weighting frame to frame. This causes the perceived FPS and Ryujinx's reported FPS to always feel slightly off from one another and as such the bottom value can never really be fully trusted at a glance. Frametime is usually much more valuable in determining how "smoothly" a game is actually running and so the following changes were made.

Changes:

- FPS monitor now shows instantaneous FPS rather than any form of weighted average
- FPS monitor now only shows integer values (much easier to quickly check if full framerate is achieved)
- Frametime has been added as a metric with a 2 decimal place precision factor
- All performance metrics now update every 750ms rather than 1000ms

Update rate was increased (to an arbitrarily smaller value) to allow frametime to be a more accurate indicator of how smooth a game is running and with the adjustment to FPS precision this means that the bottom bar isn't just flashing different numbers constantly.
image

### Comparison of old method vs new using Skyward Sword HD and pausing emulation:

Old:
https://user-images.githubusercontent.com/44103205/133149297-292e60fb-cb2a-47ba-8520-3a2296ee2374.mp4

New: 
https://user-images.githubusercontent.com/44103205/133149432-71fd671e-6de3-4b45-ab8f-c85184665096.mp4

### More dynamic example videos

MHR:
https://user-images.githubusercontent.com/44103205/133150008-c17544bb-ac37-49fc-9a5e-672a53bb0b7b.mp4

HW AoC:
https://user-images.githubusercontent.com/44103205/133150502-305afb93-f838-4817-b1f8-f2e8b3e1cfe8.mp4

Advice and opinions are very much needed :)


Ryujinx.Headless.SDL2/WindowBase.cs
```diff
@@ -198,7 +198,8 @@ public void Render()
                             Device.EnableDeviceVsync,
                             dockedMode,
                             Device.Configuration.AspectRatio.ToText(),
-                            $"Game: {Device.Statistics.GetGameFrameRate():00.00} FPS",
+                            $"Game: {Device.Statistics.GetGameFrameRate():0} FPS",
```
AcK77, line 201: I’d do something like `Game: 00.00 FPS (00.00ms)`

Ryujinx/Ui/MainWindow.cs
```diff
@@ -1105,6 +1106,7 @@ private void Update_StatusBar(object sender, StatusUpdatedEventArgs args)
             Application.Invoke(delegate
             {
                 _gameStatus.Text  = args.GameStatus;
+                _gameTime.Text  = args.GameTime;
```
AcK77, line 1109: Alignment is wrong here

Ryujinx.Headless.SDL2/WindowBase.cs
MutantAura: I half agree on this. Personally I still feel like Game: `0 FPS (00.00ms)` works better for general users. 
Quick to glance at framerate with frametime acting as your precision tool. I'd say most people just want to glance down and see 60FPS or 30FPS rather than 59.89FPS which is effectively 60 as far as the user cares.

That being said it's not a hill I'd die on so I'm willing to hear more people's views on the matter. Updated with the main idea anyway.

Ryujinx/Ui/MainWindow.cs
MutantAura: Fixed by deletion lol.

Ryujinx.Headless.SDL2/WindowBase.cs
gdkchan: Agree but I would add a space before the ms `00.00 FPS (00.00 ms)`.

Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -427,7 +427,7 @@ public void Render()
                             Device.EnableDeviceVsync,
                             dockedMode,
                             ConfigurationState.Instance.Graphics.AspectRatio.Value.ToText(),
-                            $"Game: {Device.Statistics.GetGameFrameRate():00.00} FPS",
+                            $"Game: {Device.Statistics.GetGameFrameRate():0} FPS ({Device.Statistics.GetGameFrameTime():00.00} ms)",
```
AcK77, line 430: I'm still for `00.00 FPS`, as gdkchan agreed too.

Mou-Ikkai: > I'm still for `00.00 FPS`, as gdkchan agreed too.

One up here, I'd like 00.00 as well.

MutantAura:
FPS monitor is back at `00:00`. 
I still have my reservations on this but it seems I'm in a minority. :)

Ryujinx/Ui/RendererWidgetBase.cs
MutantAura: Adjusted.

AcK77 (approved):
Lgtm! Thanks!

gdkchan (approved):
lgtm, thanks.

Ryujinx.Headless.SDL2/WindowBase.cs
gdkchan: There are cases where having the FPS fraction might be desirable, like the rare cases where game runs at less than 1 FPS due to some issue, just rounding it to 0 would appear as if it was not rendering anything. Although since it shows the frame time now, I guess this is not a problem anymore. I think one alternative that could make everyone happy is simply making this configurable on the settings. It would have 5 formats that the user can choose from:
XX.XX FPS (XX.XX ms)
XX FPS (XX.XX ms)
XX FPS (XX ms)
XX.XX FPS
XX FPS

PR #2641, opened 2021-09-14
Implement and use an Interval Tree for the MultiRangeList

### LINKED ISSUE #1925: Range list interval overlaps optimization
#### Overview:
Currently, the range list implementation performs linear search on the `FindOverlaps` method. The purpose of this method is finding all ranges that overlaps with the range that is passed as argument. This can be improved with the use of interval trees, a data structure designed for fast insertion, removal and lookup of intervals.

#### Smaller Details: 
This can be currently used by `RangeList` and `MultiRangeList`, which can be found on the `Ryujinx.Memory` project, under `Range`. The method to be optimized is the `FindOverlaps` method. There is also a `FindOverlapsNonOverlapping`, which can find overlap in O(log2(n)+m) time in the worst case, where n is the total number of intervals in the list, and m the number of overlaps. However this only works if none of the intervals in the list overlap with each other, which is the case for the buffer cache, but not for the texture cache. A fast `FindOverlaps` implementation would make the `FindOverlapsNonOverlapping` method obsolete, and all uses could be replaced by `FindOverlaps`.

#### Nature of Request:  
This is a optimization. It should greatly improve speed of texture creation. It will not improve peak performance, but should reduce stuttering caused by texture creation a little bit.

riperiperi:
This PR implements an augmented interval tree based off of the existing TreeDictionary and uses it for the MultiRangeList. This greatly speeds up texture overlap checks, as they can't use the non-overlapping fast path that buffers and tracking handles can use. Like the tree dictionary, it is based on a red-black tree and is self balancing.

This supports multi-ranges by inserting the resource into the interval tree multiple times; one for each sub-range. When searching for a range with multiple subranges, multiple queries are made and the results are merged.

Fixes a potential performance regression with #2624, as checking if a texture is in a range becomes much cheaper (5% -> 0.02%). This will also speed up texture/view creation, and notifying textures that they have been unmapped, which might reduce stuttering in unreal engine games.

Supercedes #2204.
Closes #1925

### DIFF
--- a/Ryujinx.Common/Collections/IntervalTree.cs
+++ b/Ryujinx.Common/Collections/IntervalTree.cs
@@ -112,8 +112,6 @@ namespace Ryujinx.Common.Collections
             }
 
             Insert(start, end, value);
-
-            //ValidateTree();
         }
 
         /// <summary>
@@ -122,6 +120,7 @@ namespace Ryujinx.Common.Collections
         /// <param name="key">Key of the node to remove</param>
         /// <param name="value">Value to remove</param>
         /// <exception cref="ArgumentNullException"><paramref name="key"/> is null</exception>
+        /// <returns>Number of deleted values</returns>
         public int Remove(K key, V value)
         {
             if (key == null)
@@ -139,7 +138,7 @@ namespace Ryujinx.Common.Collections
         /// <summary>
         /// Adds all the nodes in the dictionary into <paramref name="list"/>.
         /// <br></br>
-        /// The nodes will be added in Sorted by Key Order.
+        /// The nodes will be added and sorted by Key Order.
         /// </summary>
         public List<RangeNode<K, V>> AsList()
         {
@@ -167,7 +166,9 @@ namespace Ryujinx.Common.Collections
 
             return list;
         }
+
         #endregion
+
         #region Private Methods (BST)
 
         /// <summary>
@@ -370,7 +371,7 @@ namespace Ryujinx.Common.Collections
         }
 
         /// <summary>
-        /// Removes instances of <param name="value"> from the dictionary after searching for it with <param name="key">.
+        /// Removes instances of <paramref name="value"> from the dictionary after searching for it with <paramref name="key">.
         /// </summary>
         /// <param name="key">Key to search for</param>
         /// <param name="value">Value to delete</param>
@@ -470,7 +471,7 @@ namespace Ryujinx.Common.Collections
         }
 
         /// <summary>
-        /// Finds the node whose key immediately less than <paramref name="node"/>.Key.
+        /// Finds the node whose key is immediately less than <paramref name="node"/>.
         /// </summary>
         /// <param name="node">Node to find the predecessor of</param>
         /// <returns>Predecessor of <paramref name="node"/></returns>
@@ -685,7 +686,7 @@ namespace Ryujinx.Common.Collections
 
         #region Safety-Methods
 
-        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against nullpointerexceptions.
+        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullPointerExceptions.
 
         /// <summary>
         /// Returns the color of <paramref name="node"/>, or Black if it is null.
--- a/Ryujinx.Common/Collections/IntervalTree.cs
+++ b/Ryujinx.Common/Collections/IntervalTree.cs
@@ -454,13 +454,8 @@ namespace Ryujinx.Common.Collections
         /// </summary>
         /// <param name="node">Root Node</param>
         /// <returns>Node with the maximum key in the tree of <paramref name="node"/></returns>
-        /// <exception cref="ArgumentNullException"><paramref name="node"/> is null</exception>
         private static IntervalTreeNode<K, V> Maximum(IntervalTreeNode<K, V> node)
         {
-            if (node == null)
-            {
-                throw new ArgumentNullException(nameof(node));
-            }
             IntervalTreeNode<K, V> tmp = node;
             while (tmp.Right != null)
             {
--- a/Ryujinx.Common/Collections/IntervalTree.cs
+++ b/Ryujinx.Common/Collections/IntervalTree.cs
@@ -137,32 +137,13 @@ namespace Ryujinx.Common.Collections
 
         /// <summary>
         /// Adds all the nodes in the dictionary into <paramref name="list"/>.
-        /// <br></br>
-        /// The nodes will be added and sorted by Key Order.
         /// </summary>
+        /// <returns>A list of all RangeNodes sorted by Key Order</returns>
         public List<RangeNode<K, V>> AsList()
         {
             List<RangeNode<K, V>> list = new List<RangeNode<K, V>>();
 
-            Queue<IntervalTreeNode<K, V>> nodes = new Queue<IntervalTreeNode<K, V>>();
-
-            if (this._root != null)
-            {
-                nodes.Enqueue(this._root);
-            }
-            while (nodes.Count > 0)
-            {
-                IntervalTreeNode<K, V> node = nodes.Dequeue();
-                list.AddRange(node.Values);
-                if (node.Left != null)
-                {
-                    nodes.Enqueue(node.Left);
-                }
-                if (node.Right != null)
-                {
-                    nodes.Enqueue(node.Right);
-                }
-            }
+            AddToList(_root, list);
 
             return list;
         }
@@ -170,6 +151,25 @@ namespace Ryujinx.Common.Collections
         #endregion
 
         #region Private Methods (BST)
+
+        /// <summary>
+        /// Adds all RangeNodes that are children of or contained within <paramref name="node"/> into <paramref name="list"/>, in Key Order.
+        /// </summary>
+        /// <param name="node">The node to search for RangeNodes within</param>
+        /// <param name="list">The list to add RangeNodes to</param>
+        private void AddToList(IntervalTreeNode<K, V> node, List<RangeNode<K, V>> list)
+        {
+            if (node == null)
+            {
+                return;
+            }
+
+            AddToList(node.Left, list);
+
+            list.AddRange(node.Values);
+
+            AddToList(node.Right, list);
+        }
 
         /// <summary>
         /// Retrieve the node reference whose key is <paramref name="key"/>, or null if no such node exists.
@@ -228,9 +228,9 @@ namespace Ryujinx.Common.Collections
                     // Contains this node. Add overlaps to list.
                     foreach (RangeNode<K,V> overlap in node.Values)
                     {
-                        if (end.CompareTo(overlap.Start) > 0 && start.CompareTo(overlap.End) < 0)
+                        if (start.CompareTo(overlap.End) < 0)
                         {
-                            if (overlaps.Length == overlapCount)
+                            if (overlaps.Length >= overlapCount)
                             {
                                 Array.Resize(ref overlaps, overlapCount + ArrayGrowthSize);
                             }
@@ -686,7 +686,7 @@ namespace Ryujinx.Common.Collections
 
         #region Safety-Methods
 
-        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullPointerExceptions.
+        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullReferenceExceptions.
 
         /// <summary>
         /// Returns the color of <paramref name="node"/>, or Black if it is null.
--- a/Ryujinx.Common/Collections/IntervalTree.cs
+++ b/Ryujinx.Common/Collections/IntervalTree.cs
@@ -1,0 +1,819 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+
+namespace Ryujinx.Common.Collections
+{
+    /// <summary>
+    /// An Augmented Interval Tree based off of the "TreeDictionary"'s Red-Black Tree. Allows fast overlap checking of ranges.
+    /// </summary>
+    /// <typeparam name="K">Key</typeparam>
+    /// <typeparam name="V">Value</typeparam>
+    public class IntervalTree<K, V> where K : IComparable<K>
+    {
+        private const int ArrayGrowthSize = 32;
+
+        private const bool Black = true;
+        private const bool Red = false;
+        private IntervalTreeNode<K, V> _root = null;
+        private int _count = 0;
+
+        public int Count => _count;
+
+        public IntervalTree() { }
+
+        #region Public Methods
+
+        /// <summary>
+        /// Gets the values of the interval whose key is <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">Key of the node value to get</param>
+        /// <param name="overlaps">Overlaps array to place results in</param>
+        /// <returns>Number of values found</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="key"/> is null</exception>
+        public int Get(K key, ref V[] overlaps)
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+
+            IntervalTreeNode<K, V> node = GetNode(key);
+
+            if (node == null)
+            {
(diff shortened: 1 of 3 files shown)


Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+            }
+
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            Insert(start, end, value);
+
+            //ValidateTree();
```
gdkchan, line 116: Leftover?

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+        }
+
+        /// <summary>
+        /// Removes the given <paramref name="value"/> from the tree, searching for it with <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">Key of the node to remove</param>
+        /// <param name="value">Value to remove</param>
+        /// <exception cref="ArgumentNullException"><paramref name="key"/> is null</exception>
```
gdkchan, line 124: Missing `returns` doc (also why is the return value an `int`? Can it remove more than one item?).

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+
+            _count -= removed;
+
+            return removed;
+        }
+
+        /// <summary>
+        /// Adds all the nodes in the dictionary into <paramref name="list"/>.
+        /// <br></br>
+        /// The nodes will be added in Sorted by Key Order.
```
gdkchan, line 142:
```suggestion
        /// The nodes will be added and sorted by Key Order.
```
Maybe?

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+                if (node.Right != null)
+                {
+                    nodes.Enqueue(node.Right);
+                }
+            }
+
+            return list;
+        }
+        #endregion
+        #region Private Methods (BST)
```
gdkchan, line 171: nit: I'd add an empty line before to separate the 2.

riperiperi:
Closes #1925

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+                parent.Right = newNode;
+            }
+
+            PropagateIncrease(newNode);
+            _count++;
+            return newNode;
+        }
+
+        /// <summary>
+        /// Removes instances of <param name="value"> from the dictionary after searching for it with <param name="key">.
```
gdkchan, line 373:
```suggestion
        /// Removes instances of <paramref name="value"> from the dictionary after searching for it with <paramref name="key">.
```

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+            while (tmp.Right != null)
+            {
+                tmp = tmp.Right;
+            }
+
+            return tmp;
+        }
+
+        /// <summary>
+        /// Finds the node whose key immediately less than <paramref name="node"/>.Key.
```
gdkchan, line 473:
```suggestion
        /// Finds the node whose key is immediately less than <paramref name="node"/>.
```

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,819 @@
...
+                node.Parent = left;
+
+                PropagateFull(node);
+            }
+        }
+        #endregion
+
+        #region Safety-Methods
+
+        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against nullpointerexceptions.
```
gdkchan, line 688:
```suggestion
        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullPointerExceptions.
```

gdkchan (commented):
Some comments might also apply to `TreeDictionary`. Would be nice to update it too if so.

Ryujinx.Common/Collections/IntervalTree.cs
riperiperi: It can remove more than one item - it's allowed that multiple values exist on one node, and delete iterates through all of them.

Ryujinx.Common/Collections/IntervalTree.cs
gdkchan: I just noticed this should be called `NullReferenceException` actually. `NullPointerException` is the name of the equivalent Java exception.

Ryujinx.Common/Collections/IntervalTree.cs
jduncanator: Key Order should also be lower case: `sorted by key order.`

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+            if (node.Values.Count > overlaps.Length)
+            {
+                Array.Resize(ref overlaps, node.Values.Count);
+            }
+
+            int overlapsCount = 0;
+            foreach (RangeNode<K, V> value in node.Values)
+            {
+                overlaps[overlapsCount++] = value.Value;
+            }
```
FICTURE7, line 59: Perhaps something like this?

```suggestion
            int overlapsCount = node.Values.Count;
            
            if (overlapsCount > overlaps.Length)
            {
                overlaps = node.Values.ToArray();
            }
            else
            {
                node.Values.CopyTo(overlaps);
            }
```

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+
+            return removed;
+        }
+
+        /// <summary>
+        /// Adds all the nodes in the dictionary into <paramref name="list"/>.
+        /// <br></br>
+        /// The nodes will be added and sorted by Key Order.
+        /// </summary>
+        public List<RangeNode<K, V>> AsList()
```
FICTURE7, line 143: Is this not doing a BFS traversal of the tree, so the nodes wont be ordered by key?

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+
+            bool endsOnRight = end.CompareTo(node.Start) > 0;
+            if (endsOnRight)
+            {
+                if (start.CompareTo(node.End) < 0)
+                {
+                    // Contains this node. Add overlaps to list.
+                    foreach (RangeNode<K,V> overlap in node.Values)
+                    {
+                        if (end.CompareTo(overlap.Start) > 0 && start.CompareTo(overlap.End) < 0)
```
FICTURE7, line 231: Consider removing the `end.CompareTo(overlap.Start)`, because all `RangeNode<,>` in `node.Values` will start at `node.Start`, i.e: `node.Start == overlap.Start` (Assuming I understood properly).

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+                        {
+                            if (overlaps.Length == overlapCount)
```
FICTURE7, line 233: Consider changing this to `>=` instead of `==` because something like `var overlaps = new T[0]; Get(start, end, ref overlaps, 6)` may throw `IndexOutOfRangeException` or ensure there is enough space up to `overlapCount` in `Get`.

Ryujinx.Memory/Range/MultiRangeList.cs
```diff
@@ -97,22 +81,45 @@ public int FindOverlaps(ulong address, ulong size, ref T[] output)
...
+                            output[insertPtr++] = output[newI];
+                        }
                     }
 
-                    output[outputIndex++] = item;
+                    overlapCount = insertPtr;
+                }
+                else
+                {
+                    overlapCount = newOverlaps;
```
FICTURE7, line 118: Is this not going to ignore duplicates for the first range or single range multi-ranges?

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+                node.Parent = left;
+
+                PropagateFull(node);
+            }
+        }
+        #endregion
+
+        #region Safety-Methods
+
+        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullPointerExceptions.
```
FICTURE7, line 689:
```suggestion
        // These methods save memory by allowing us to forego sentinel nil nodes, as well as serve as protection against NullReferenceExceptions.
```

Ryujinx.Common/Collections/IntervalTree.cs
riperiperi: Can't do this - the `.Value` property must be extracted from the range node.

FICTURE7: Oof 😢

Ryujinx.Memory/Range/MultiRangeList.cs
riperiperi: Hmm, I suppose duplicate entries can exist in the tree if a multi-range were inserted. (i was only accounting for duplicate ranges added by multiple sub-range queries) I'm not a fan of iterating everything here but I guess it's unavoidable.

Ryujinx.Common/Collections/IntervalTree.cs
riperiperi: That is true - I was bee-lining the end overlap check as each individual element does have a different end. I'll remove this.

Ryujinx.Common/Collections/IntervalTree.cs
riperiperi: It is unexpected for someone to do that, but you're right.

Ryujinx.Memory/Range/MultiRangeList.cs
FICTURE7: Feel free to discard any of these comments btw, because I am unware of many of the underlying details.

Ryujinx.Common/Collections/IntervalTree.cs
riperiperi: This was copied from the IntervalTree, so I think that might be wrong too. It would be easier to do the DFS, as well..

Ryujinx.Common/Collections/IntervalTree.cs
```diff
@@ -0,0 +1,820 @@
...
+
+        /// <summary>
+        /// Returns the node with the largest key where <paramref name="node"/> is considered the root node.
+        /// </summary>
+        /// <param name="node">Root Node</param>
+        /// <returns>Node with the maximum key in the tree of <paramref name="node"/></returns>
+        /// <exception cref="ArgumentNullException"><paramref name="node"/> is null</exception>
+        private static IntervalTreeNode<K, V> Maximum(IntervalTreeNode<K, V> node)
+        {
+            if (node == null)
```
FICTURE7, line 460: Consider removing the null check here because its a private method and the null check is already done where it is used.

Ryujinx.Memory/Range/MultiRangeList.cs
```diff
@@ -97,22 +81,45 @@ public int FindOverlaps(ulong address, ulong size, ref T[] output)
...
+                    // Filter new overlaps - remove duplicates.
+                    int insertPtr = overlapCount;
+
+                    for (int newI = overlapCount; newI < newOverlaps; newI++)
                     {
-                        Array.Resize(ref output, outputIndex + ArrayGrowthSize);
+                        bool duplicate = false;
+                        for (int oldI = 0; oldI < overlapCount; oldI++)
+                        {
+                            if (output[newI].Equals(output[oldI]))
```
FICTURE7, line 101: Consider changing this to use `EqualityComparer<T>.Default.Equals(,)` because the JIT will try harder to devirtualize it. See [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8lAbhvoYGUALbKAA4AZbMAB0AJQCuAOwwBLfDHbUaAbQBS8jAHEYMmFHlgAFBgCeAmBABmJgIIBKRwF0awCBAA2DALJ0AHgAVAD4TIIZsNAYI4EcGAF4QmgZUyLEAUQBHKWwvXBM4lXUtXX1DYzNLazsnV3dPH19yYLCIqJiGOMTk6jSGbNyvbXMAYQh8AX5DVrEAERgbbCkvDEycvIKOopoOAGYGXAwoKTAMBnsGEAYASUHsDFEvGAD7EIYAbxS04gP5OQYADUVP1fgx/ucAKogtLfVJgjzeAYbfIOBgQDA8QyOOGfXGggDs6MxhjEgMSCSBDAAZNTiVioGJIRSGNDcQBfGjsoA). Will require `T` to implement `IEquatable<T>` to benefit, although this mostly affects when `T` is a struct.

If you have trust issues with the JIT, consider hoisting `output[newI]` out of the loop manually. I don't think its smart enough to eliminate the bound checks for each access.

riperiperi: In our case, T is always a class. Would I need to implement `IEquatable<T>` to get it to use the object reference comparer?

gdkchan (approved):
lgtm, thanks.
Very nice to see this finally using a more adequate data structure instead of checking every single item on the list.
I can confirm that the loading time of the Mario Golf courses is much shorter now.

Ryujinx.Memory/Range/MultiRangeList.cs
jduncanator: No, by default `EqualityComparer<T>.Default` will use `ObjectEqualityComparer<T>` unless T is an enum, Nullable, or otherwise assignable to `IEquatable<T>`. `ObjectEqualityComparer<T>` defers to `T.Equals` which will do a reference comparison by default unless otherwise overridden.

FICTURE7: Yeah what @jduncanator said. Some other scuffed options to improve performance
1. Use `ReferenceEquals(,)` when `T` is a reference type. Will break `Equal` overrides.
2. Wrapping the reference type in a struct. This will prevent the JIT from using shared generics and generate new code for it, whilst having more knowledge about the generic context. See [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8lAbhvqfIDoAlAVwB2GAJb4Y3AMIR8ABxEAbGFADKygG4iwMXO2o0A2gFkYGABYQAJgEk5CgBQnzV27IUB5WaIiDc3AIIA5oGwuLgi6jDWggoignGBAJQAujTAEBAKDEZ0ADwAKgB89nEM+QzYaAyl5cCJDAC8hTQMrRXcAKIAjvzYCrj2dXqGThY2do6mY64eXiI+fkEhOuGR0bHxgkmp1OmZ2eQFxTUVVSd1jc3UbQzdvbEYAJ7SctiwUEfcACIwAGbY/AUGE6PT6A0qDCGNBGUxcE1GcLcnm8vgCwVCqyiMTiCRSaQyWSMpCOJUEZVO1TJtXqTRabV4f2UMEE2juYPsEKh+l2BIYAGk6PZ/BU6FVhdhyDTCtlBdhRRVJXo9lk+eQhSKxQqpQcOfKJYklby+aR1XLNfrLtkTWatYb9nyAMz2ABCkPlruAkstRjVwHlnoN0NoDqYpAYwoA3nTWsQQ3EMAwAGp6G6xykJgCqegAvkGAPR5hgAdSg2FkofDlIqDFwGCg/DACdwEAYAHczMzqgmRLgGPxcDBLFXsAxAszlFoquYYAwAFLWfI0AttxRZYAz37QbRDjAt3CyGBgER9EQALxn04YkEsMDO3BgIAYgggCcCLfMUAg/ECZhrZjeg5LoWY6CBOYB+EBQEMGoM6OtwHAhrW9aNgwrqPtYbIYNgwBKLkzrSlG1xtGm4rDERMYhsqtygv0LoMC+HZQIk0YMIRNypgA7NWDQNPR05QNw2Apm0ubUNmQA).

Yeah, not exactly ideal; feel free to ignore.

Thog (approved):

PR #2642, opened 2021-09-15
Array based RangeList that caches Address/EndAddress

riperiperi:
This PR changes the RangeList to cache the Address and EndAddress within the list itself rather than accessing them from the object's properties. It also changes the RangeList to be backed by an array containing structs with the above information, back to back. This improves memory locality when binary searching through the array list, so I reckon it'll avoid slow cache misses from jumping to a bunch of resources miles apart to check their address and end address. It might cost more to add or remove items to large lists, but our use case is mostly searching.

This array list is used in a few places: Memory tracking, windows emulated view + placeholder tracking, buffer modified list and buffer lookup. It's this last use that is most important - we were losing quite a bit of time looking up buffers by CPU VA when binding buffers (uniforms primarily). Note that these cases are all non-overlapping ranges - the Textures range list is covered by the other PR with the IntervalTree.

In isolation, this was more than 2x faster than the RangeList that checks using the interface. In practice I'm seeing much better results than I expected. The array is used because checking it is slightly faster than using a list, which loses time to struct copies, but I still want that data locality.

A method has been added to the list to update the cached end address, as some users of the RangeList currently modify it dynamically.

Greatly improves performance in Super Mario Odyssey (~1.25x), Xenoblade (should also use my other PRs in combination) and most other GPU limited games. Improvement will generally depend on how many buffers the game binds and how many draws it does. Give your favourite 98% FIFO game a shot.

### Before:
image

### After:
image

## Xenoblade Chronicles: Definitive Edition
I picked a taxing area that didn't benefit much from the I2M change to demonstrate the change in raw GPU throughput. Note that to get the best performance, you must also have the I2M change.
### Before (master):
image

### #2624 (apologies for different fps measurement ui, someone decided to rapidly change it 😡 ):
image

### After + #2624 + #2641:
image

### DIFF
--- a/Ryujinx.Memory/Range/RangeList.cs
+++ b/Ryujinx.Memory/Range/RangeList.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace Ryujinx.Memory.Range
 {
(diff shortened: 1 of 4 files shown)


Ryujinx.Memory/Range/RangeList.cs
```diff
@@ -10,18 +11,40 @@ namespace Ryujinx.Memory.Range
     /// <typeparam name="T">Type of the range.</typeparam>
     public class RangeList<T> : IEnumerable<T> where T : IRange
     {
-        private const int ArrayGrowthSize = 32;
+        private readonly struct RangeItem<T> where T : IRange
```
gdkchan, line 14: Can you rename `T` to something like `TItem` or `TValue` to silence the warning?

Ryujinx.Memory/Range/RangeList.cs
```diff
@@ -10,18 +11,40 @@ namespace Ryujinx.Memory.Range
...
+        {
+            public readonly ulong Address;
+            public readonly ulong EndAddress;
+
+            public readonly T Item;
```
gdkchan, line 19: I'd probably call this one `Value` to avoid the ambiguity of something like `_items[index].Item`.

gdkchan (approved):
lgtm, thanks. The games I tested are working properly 👍 
Very nice improvement, wasn't expecting it either.

Thog (approved):

PR #2645, opened 2021-09-15
Fix problems added by Pause

mpnico:
This PR fixes two problems :
- Hiding Pause/Resume entry in Action menu doesn't work properly after switching fullscreen mode. With this PR, the entries are always shown but contextually disabled
- Since #2428, Ryujinx can't be properly closed because of a remaining audio thread. This should be fixed now.


Thog (approved):

riperiperi (approved):

gdkchan (approved):
Nice, thanks for fixing it quickly. I tested here and its exiting properly again now.

PR #2646, opened 2021-09-16
Only make render target 2D textures layered if needed

### LINKED ISSUE #2628: The Legend of Heroes: Zero no Kiseki

gdkchan:
There are some cases where games have bogus values written as the render target texture depth. This can cause very large 2D array textures to be created, which not only is bad for performance, but can also cause OOM and other issues.

Usually the non-base layers of the texture doesn't are not accessed at all, as it will only render to a single layer. It only matters when the shader writes to `gl_Layer` with a non-zero value. Doing so will modify the target layer. Knowing this, I have modified the code to only ever use 2D array when one of the VTG (vertex, tessellation or geometry) shaders writes to `gl_Layer`.

This fixes #2628 which was caused by an 1080p array texture with 257 layers (which would take several GBs of memory and cause all sorts of issues).

image

586fdd5dbdd9c95bc2a78342d02f7c9f34edb3e3 fixes an unrelated issue that I found while testing this game. When the game uses multiple channels and there is a channel swap, the topology value was not being updated on the host, making it use the wrong primitive type for some draws. A new bool was added to force an update if there is a channel swap.

I'm not aware of any game that actually tries to render to a non-base layer of an array texture, so not sure what should be tested here. Might be worth testing Luigi's Mansion 3.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
@@ -13,6 +13,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         private readonly GpuChannel _channel;
         private readonly DeviceStateWithShadow<ThreedClassState> _state;
         private readonly DrawState _drawState;
+        private bool _topologySet;
 
         private bool _instancedDrawPending;
         private bool _instancedIndexed;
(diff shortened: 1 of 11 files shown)


Linear524:
Luigis Mansion 3 works fine with this. 
But LM3 needs PR #2534 to be merged for 100% correct overall image...

riperiperi (approved):
Nice catch. Will need to try some games.

GamerzHell9137:
This is fixing a regression with RF4S on newlywed mode when multithreaded graphics backend is enabled

Before:
image
After:
image

AcK77 (approved):

PR #2647, opened 2021-09-16
Set texture/image bindings in place rather than allocating and passing an array

riperiperi:
This was allocating multiple arrays per draw or compute invocation. Noticed it on a profile, its cost was significant but not mind blowing. Figured like the other change, removing these allocations might have some small effects on performance outside this method. Good to remove costs like this either way.

I've changed the SetTextures and SetImages method to instead rent the bindings array for modification, now called RentTextureBindings, which is in charge of ensuring the arrays are the correct size, and clearing the state arrays. This is done to set the data directly, rather than allocate or copy into the bindings manager. 

They now use arrays that are pre-allocated with a default size, but can be increased in size to fit shaders that bind way more textures, such as bindless accesses.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -11,6 +11,9 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class TextureBindingsManager : IDisposable
     {
+        private const int InitialTextureStateSize = 32;
+        private const int InitialImageStateSize = 8;
+
         private const int HandleHigh = 16;
         private const int HandleMask = (1 << HandleHigh) - 1;
 
@@ -43,6 +46,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         private readonly TextureStatePerStage[][] _textureState;
         private readonly TextureStatePerStage[][] _imageState;
 
+        private int[] _textureBindingsCount;
+        private int[] _imageBindingsCount;
+
         private int _textureBufferIndex;
 
         private bool _rebind;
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -81,25 +99,57 @@ public TextureBindingsManager(GpuContext context, GpuChannel channel, TexturePoo
...
         /// <summary>
-        /// Binds textures for a given shader stage.
+        /// Rents the texture bindings array for a given stage, so that they can be modified.
         /// </summary>
         /// <param name="stage">Shader stage number, or 0 for compute shaders</param>
-        /// <param name="bindings">Texture bindings</param>
-        public void SetTextures(int stage, TextureBindingInfo[] bindings)
+        /// <param name="count">The number of bindings needed</param>
+        /// <returns>The texture bindings array</returns>
+        public TextureBindingInfo[] RentTextureBindings(int stage, int count)
```
gdkchan, line 107: Is it not possible to deduplicate some of this code by having a common function where you pass array references (`ref T[]`)? I can understand the duplication if that causes worse code to be generated though.

riperiperi: I'm not sure it would be much nicer - 3 array references would be going into the method. Unsure about performance implications. 😕

gdkchan: Ok then.

gdkchan (approved):
Looks good, does it bring any noticeable improvement?

riperiperi:
In SMO I noticed more of a difference in the FIFO% than the framerate - with the other PRs merged, capturing the wire in metro it was 75-76% a lot more often (with the other PRs) where the main hung around 77-81% and spiked. Nothing as dramatic as swapping out the range list. The framerate does seem to fluctuate a lot... might be due to the conditional rendering.

Thog (approved):

PR #2649, opened 2021-09-17
Refactor `CompilerContext`

FICTURE7:
This refactors `CompilerContext` to play a more "important" role and some other changes.
* Renamed `Allocators` to `CompilerAllocators` and is now reference counted.
* `CompilerContext` is now a disposable class. It will manage a reference count to the arenas. It also has a finalizer, so the GC will decrement the count if the context is not disposed explicitly. Meaning memory is not leaked when the arenas are not referenced anymore.
* Renamed `CompilerContext` to `Compiler`.

Some other stuff to note:
* `Compiler` can be used for per compilation symbol table. (Needed to decouple `Delegates` from backend).
* `Compiler` can be used for per compilation logging and metrics.
* Since the lifetime of the IR starts with a `Compiler`, the `Compiler` is created before the `EmitterContext` is created. Alternatively, `EmitterContext` could return a `Compiler`.

Contributes to #1125


ARMeilleure/CodeGen/CompilerContext.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    class CompilerContext : IDisposable
+    {
+        [ThreadStatic]
+        private static CompilerAllocators _allocators;
+
+        /// <summary>
+        /// Gets the <see cref="CompilerAllocators"/> associated the current thread.
+        /// </summary>
+        public static CompilerAllocators Allocators
```
gdkchan, line 17:
```cs
public static CompilerAllocators Allocators => _allocators ??= new CompilerAllocators();
```

FICTURE7: Ohh, neat. Yeah this method is quite hot, although the majority of the cost is accessing the thread static. I think we can get some pretty good gains by avoiding thread statics.

I was planning on making `Operand.Factory` and `Operation.Factory` non static and use a local reference to the allocators, should make accessing them less expensive. The instances would reside in `CompilerContext` then. `EmitterContext` would probably have to be augmented to hand out `Constant` and `Label` as well.

gdkchan: I decided to take a closer look at the code generated for the null coalescing version, and I'm not sure why it is generating more code. At a first glance it should be better, because at the IL level it `Dup`s the value and avoids the second load of the static field when the value is non-null. But for some reason it decides to insert an extra `mov`, which also causes an extra register to be saved/restored. Here is the code I used for the test:

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHuGBtAFQAWsbABMAyhmwYAlmAC6DAA5QZAN2kwm5JAwjAAVjDAYGAfQwxcGANxceK9Zr2Hjpi1Yyk71Hk0bEOi5GJgwAMtgAXgCeAJIAdjIYktJyABQAlBz2ftwyAGYMaR7WDAC8ZQzxAK4ANrVZnL65LSWmlfEwAO7Bbpk+LdwAvjm5xADs5pbWAzwjzdz02rr6IaYRMQlJKbJgAHJ1tSwQ2LVWYDLxAOaZ2Qtjk20MAPzPHd29Jv2j835Lq25wlE4okMLcmoMGAUim1SOUOodGqNIbDylUPgCvhlZi1foMJlNPN4fqN/q5QhsQUkDvVjqdzpcbkj7n9HtMvC83uiepiwdiSdQhkA==
It does:
```x86asm
    L0021: mov rdi, rax
    L0024: lea rcx, [rsi+8]
    L0028: mov rdx, rdi
```
 rather than:
```x86asm
     L001e: lea rcx, [rsi+8]
    L0022: mov rdx, rax
```
Which causes `rdi` to be saved/restored on the prolog/epilog too. Dunno why it does this, I think their register allocator messed up again.

The other diff is going from:
```x86asm
    L0005: mov rsi, rcx
    L0008: cmp qword ptr [rsi+8], 0
```
to:
```x86asm
    L0006: mov rsi, rcx
    L0009: mov rax, [rsi+8]
    L000d: test rax, rax
```
Which is actually expected, since it actually returns the `rax` value directly for the non-null case instead of loading it again (which is good). At least it does remove the second memory load, so maybe it is faster despite that.

I also added a non-static version for comparison. It gets rid of the call on the hot path which is good, so I guess it is worth going that route.

FICTURE7: Thing is the thread static will go through the CLR, where it contains a platform agnostic implementation of thread local storage. This is where the majority of the cost resides, so there is not really much we can do. But yeah if we're optimizing this method, we can manually do function outlining and move the cold code into its own method, see [LazyInitStaticSplit](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYmpBkIYBvGgzMMA2gBUAFrGwATWRmx8wAXQZiovBa5hM5EgMEMAAVjBgGAwA+hiqGBrU5l4+fvEh4ZHRcQmkSSk6gcGhEVEMADLYAF4AngCSXLwYzq7KABQAlMamKea8AGYM7bm40QC84wxcHJKS3SbJfcujE9MwAO6ZZRhdBcsMAL69y8QA7LHxY/tmx0tmRcwlWeVVdY3NrW4AcrOSIthpLgwLwuABzLo9e59c6XBIMAD8CKmXE222yexORyxjyC6NeNQaTRaLjcsjEkmakMWBwYg2GqwYAEIUX9OljabDVjcYRc3kTPqTlOTKRg2NhYJjoXcUhYALIwDA2CAOeriSTtBVKlVqikAeTEfAgXFwLG+EEalKa4M67ix3l8/mK+Oi/I+JLaYBFzXFkoWHPMXKua1RW1KGM6PJlgd04YJ72J1IDZnpI2D+km0zZUNpKVWGfWYZeu0jyex0MKF3zUZxseLlUJ7t+cwBQJB4KTFcDVfTiORhZdUpSd0OQA=).

marysaka:
What should be done about this PR? @gdkchan 

@FICTURE7 any plans to continues it?

marysaka:
Closing as stale.

PR #2653, opened 2021-09-17
Share scales array for graphics and compute

gdkchan:
Fixes an issue where scales might not be properly updated on games that uses compute. The issue happens because the backend is using a single array to store both fragment and compute scales, while the GPU emulation is using 2 (and is working with the assumption that the backend also uses 2). This change simply shares the same array for both compute and graphics, which solves the problem and the scale changing on one will also force the other to be updated later.

This fixes resolution scale issues on NI no Kuni 2:
Before:
image
After:
image

This might potentially also fix the resolution scale regression that was reported on Discord on Mario Rabbids Kingdom Battle, but I did not test it myself.
This is potentially a regression caused by #2595, but I did not confirm.


Thog (approved):

riperiperi (approved):
LGTM, nice catch.

Issue #2660, opened 2021-09-19
`System.FormatException` when stopping emulation and the system date format is too custom


ExE-Boss:
### What's the issue you encountered?

Since version 1.0.7026, it’s no longer possible to stop emulation by using “Actions” → “Stop Emulation” when using the [Holocene calendar](https://en.wikipedia.org/wiki/Holocene_calendar).

### How can the issue be reproduced?

While running any game and the date format string is set to `1yyyy-MM-dd`, press “Actions” → “Stop Emulation”.

```reg
Windows Registry Editor Version 5.00

; Use the Holocene calendar
; https://en.wikipedia.org/wiki/Holocene_calendar
[HKEY_CURRENT_USER\Control Panel\International]
"sLongDate"="dddd, 1yyyy-MM-dd"
"sShortDate"="1yyyy-MM-dd"
"sYearMonth"="1yyyy-MM"
```

### Log file

```log
Marshaling activate signal
00:04:35.022 |E| Application : Unhandled exception caught: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.FormatException: String '12021-09-19 17:30:00' was not recognized as a valid DateTime.
   at System.DateTimeParse.Parse(ReadOnlySpan`1 s, DateTimeFormatInfo dtfi, DateTimeStyles styles)
   at Ryujinx.Ui.MainWindow.<>c.<UpdateGameMetadata>b__93_0(ApplicationMetadata appMetadata)
   at Ryujinx.Ui.App.ApplicationLibrary.LoadAndSaveMetaData(String titleId, Action`1 modifyFunction)
   at Ryujinx.Ui.MainWindow.UpdateGameMetadata(String titleId)
   at Ryujinx.Ui.MainWindow.StopEmulation_Pressed(Object sender, EventArgs args)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at GLib.Signal.ClosureInvokedCB(Object o, ClosureInvokedArgs args)
[...]

PR #2661, opened 2021-09-19
Always use ISO 8601 as the DateTime format string

ExE-Boss:
Fixes <https://github.com/Ryujinx/Ryujinx/issues/2660>

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -857,7 +857,7 @@ namespace Ryujinx.Ui
 
                 _applicationLibrary.LoadAndSaveMetaData(_emulationContext.Application.TitleIdText, appMetadata =>
                 {
-                    appMetadata.LastPlayed = DateTime.UtcNow.ToString();
+                    appMetadata.LastPlayed = DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'Z'");
                 });
             }
         }


Ryujinx/Ui/MainWindow.cs
```diff
@@ -857,7 +857,7 @@ public void LoadApplication(string path, bool startFullscreen = false)
 
                 _applicationLibrary.LoadAndSaveMetaData(_emulationContext.Application.TitleIdText, appMetadata =>
                 {
-                    appMetadata.LastPlayed = DateTime.UtcNow.ToString();
+                    appMetadata.LastPlayed = DateTime.UtcNow.ToString("yyyy'-'MM'-'dd'T'HH':'mm':'ss'Z'");
```
AcK77, line 860: What about that?
```suggestion
                    appMetadata.LastPlayed = DateTime.UtcNow.ToString("o");
```

ExE-Boss: `ToString("o")` includes microseconds.

AcK77: Right! Then it should be `"yyyy-MM-ddTHH:mm:ssZ"` instead of `"yyyy'-'MM'-'dd'T'HH':'mm':'ss'Z'"`

ExE-Boss: @AcK77
The `'` are necessary to escape the `-`, `:`, `T`, and `Z`, it’s what the [`System.Globalization.CultureInfo.InvariantCulture`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture)’s [`DateTimeFormat`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.datetimeformatinfo) is configured as.

AcK77: image
Both produce the same result, but well...

TSRBerry:
Closing this, since #4861 was merged.

PR #2664, opened 2021-09-19
Upload textures data through a persistent mapped buffer

mpnico:
Following #2481, this PR use a persistent mapped buffers to handle textures upload. The buffer is split into 3 parts, textures are uploaded to one part and once this part is full and can't handle the next texture upload, the next part is used and the previous is protected by a fence. Part size is 16Mb and textures larger than that are managed as they were before. The number and the size of those parts could be adjusted if needed after more tests.

This could increase performances for some games that constantly upload textures.

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Image/TextureView.cs
+++ b/Ryujinx.Graphics.OpenGL/Image/TextureView.cs
@@ -249,25 +249,40 @@ namespace Ryujinx.Graphics.OpenGL.Image
 
         public void SetData(ReadOnlySpan<byte> data)
         {
-            unsafe
+            if (HwCapabilities.UsePersistentBufferForFlush && data.Length < PersistentTextureBuffer.TextureMaxSize)
             {
-                fixed (byte* ptr = data)
+                _renderer.PersistentBuffers.Textures.SetTextureData(this, data, data.Length);
+            }
+            else
+            {
+                unsafe
                 {
-                    ReadFrom((IntPtr)ptr, data.Length);
+                    fixed (byte* ptr = data)
+                    {
+                        ReadFrom((IntPtr)ptr, data.Length);
+                    }
                 }
             }
         }
 
         public void SetData(ReadOnlySpan<byte> data, int layer, int level)
         {
-            unsafe
-            {
-                fixed (byte* ptr = data)
-                {
-                    int width = Math.Max(Info.Width >> level, 1);
-                    int height = Math.Max(Info.Height >> level, 1);
+            int width = Math.Max(Info.Width >> level, 1);
+            int height = Math.Max(Info.Height >> level, 1);
+            int mipSize = Info.GetMipSize2D(level);
 
-                    ReadFrom2D((IntPtr)ptr, layer, level, width, height);
+            if (HwCapabilities.UsePersistentBufferForFlush && mipSize < PersistentTextureBuffer.TextureMaxSize)
+            {
+                _renderer.PersistentBuffers.Textures.SetTextureData(this, data, mipSize, layer, level, width, height);
+            }
+            else
+            {
+                unsafe
+                {
+                    fixed (byte* ptr = data)
+                    {
+                        ReadFrom2D((IntPtr)ptr, layer, level, width, height);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


jduncanator:
This should be thoroughly tested across GPU drivers and vendors. iirc, yuzu tried implementing something similar, however the performance across vendors and driver versions was a wash, wish some be way slower.

JumpmanSr:
Specs:
5900X @ 4.5GHz
RTX 3090 FE (471.96)
32 GB DDR4 RAM @ 3200 MHz

Upon doing some initial testing on XBDE and Astral Chain using default values I noticed the following: 
Frame time appeared to be much smoother, but there was some hitching that was hurting the FPS I assume due to textures being loaded in and out of the buffer. 

So I did some experimenting, I know I have a very beefy PC so I set the max texture size for 256mb and the TextureMaxNumber to be 6 (may up this further since int.max is the upper limit in terms of the buffer size). This almost entirely removed and hitching in XBDE while keeping the frametime low, and improved FPS in places that used a lot of textures. In XBDE the wide panoramic camera shots and cutscenes now run at full speed or very close to it (58~60 fps in 4x docked) and the menu where the character models are was now also running at full speed. In Astral Chain the opening cutscene has much less desync (although still off). I'll be looking for other improvements but at least on nvidia hardware this seems to work very well.

I hope if this is implemented, users are allowed to configure the texture size and the max number as tweaking that setting based on individual hardware would most likely have the optimal result vs a one size fit's all solution.

gdkchan:
I noticed some improvements to video playback with this change.

Master:

https://user-images.githubusercontent.com/5624669/134791394-5f0eaf4d-f0a5-4c6a-8372-e9116c927460.mp4

PR:

https://user-images.githubusercontent.com/5624669/134791405-a53dd61a-6d11-41d0-8b7d-95b88133052a.mp4

Helpmysonplease:
> Specs: 5900X @ 4.5GHz RTX 3090 FE (471.96) 32 GB DDR4 RAM @ 3200 MHz
> 
> Upon doing some initial testing on XBDE and Astral Chain using default values I noticed the following: Frame time appeared to be much smoother, but there was some hitching that was hurting the FPS I assume due to textures being loaded in and out of the buffer.
> 
> So I did some experimenting, I know I have a very beefy PC so I set the max texture size for 256mb and the TextureMaxNumber to be 6 (may up this further since int.max is the upper limit in terms of the buffer size). This almost entirely removed and hitching in XBDE while keeping the frametime low, and improved FPS in places that used a lot of textures. In XBDE the wide panoramic camera shots and cutscenes now run at full speed or very close to it (58~60 fps in 4x docked) and the menu where the character models are was now also running at full speed. In Astral Chain the opening cutscene has much less desync (although still off). I'll be looking for other improvements but at least on nvidia hardware this seems to work very well.
> 
> I hope if this is implemented, users are allowed to configure the texture size and the max number as tweaking that setting based on individual hardware would most likely have the optimal result vs a one size fit's all solution.



> Specs: 5900X @ 4.5GHz RTX 3090 FE (471.96) 32 GB DDR4 RAM @ 3200 MHz
> 
> Upon doing some initial testing on XBDE and Astral Chain using default values I noticed the following: Frame time appeared to be much smoother, but there was some hitching that was hurting the FPS I assume due to textures being loaded in and out of the buffer.
> 
> So I did some experimenting, I know I have a very beefy PC so I set the max texture size for 256mb and the TextureMaxNumber to be 6 (may up this further since int.max is the upper limit in terms of the buffer size). This almost entirely removed and hitching in XBDE while keeping the frametime low, and improved FPS in places that used a lot of textures. In XBDE the wide panoramic camera shots and cutscenes now run at full speed or very close to it (58~60 fps in 4x docked) and the menu where the character models are was now also running at full speed. In Astral Chain the opening cutscene has much less desync (although still off). I'll be looking for other improvements but at least on nvidia hardware this seems to work very well.
> 
> I hope if this is implemented, users are allowed to configure the texture size and the max number as tweaking that setting based on individual hardware would most likely have the optimal result vs a one size fit's all solution.

hey, could you explain how can i change these settings?

TSRBerry:
Needs a rebase

marysaka:
Closing as stale.

PR #2666, opened 2021-09-20
Fix PTC count table relocation patching

FICTURE7:
Fix an issue introduced in #2190 where by 2 different count table entry addresses were used for LCQ functions. E.g:

```asm
 .L1:
   mov rbp,COUNT_TABLE_0   ;; This gets an address.
   mov ebp,[rbp]
   lea esi,[rbp+1]
   mov rdi,COUNT_TABLE_1   ;; This gets another address.
   mov [rdi],esi
   cmp ebp,64h
   je near .L34
```

This caused LCQ functions to not tier up when they're loaded from the PTC cache. This does not happen when they're freshly compiled.

This PR fixes the issue by ensuring only a single counter is created per translation when they're loaded from the cache.

### DIFF
--- a/ARMeilleure/Translation/PTC/Ptc.cs
+++ b/ARMeilleure/Translation/PTC/Ptc.cs
@@ -681,7 +681,10 @@ namespace ARMeilleure.Translation.PTC
                 }
                 else if (symbol == CountTableSymbol)
                 {
-                    callCounter = new Counter<uint>(translator.CountTable);
+                    if (callCounter == null)
+                    {
+                        callCounter = new Counter<uint>(translator.CountTable);
+                    }
 
                     unsafe { imm = (IntPtr)Unsafe.AsPointer(ref callCounter.Value); }
                 }


LDj3SNuD (approved):
Good catch!
When I revised #2190 I was convinced that you could only have one `CountTableSymbol` per function, so the now updated code should be fine; but evidently what happens is that we have 2 of those symbols per function, one for the Load instruction and one for the Store instruction; *when really one should be enough*.
image
Anyway, both .info and .cache files are already able to handle lcq/hcq retranslations so the respective flags do not need to be updated.

FICTURE7:
> I thought that you could only have one `CountTableSymbol` per function

Same 😩. Good to note that the new optimization added to `X86Optimizer` does not catch this case because it only hoist constants on `Add` instructions.

gdkchan (approved):
Nice catch.
About that optimization, there are some instructions where the operands are supposed to be a constant, so doing it for all the instructions wouldn't be valid, that's probably why I only enabled it for `Add`. It could also be enabled for other instructions, but I think I was just trying to keep the diff minimal back then.

PR #2671, opened 2021-09-20
NVDEC (H264): Use separate contexts per channel and decode frames in DTS order

### LINKED ISSUE #2077: Decoding issues when two or more videos are playing at once

When two or more videos are being decoded at one time, the videos fail to decode properly. The emulator's console also gets bombarded with h264 decoding errors.

1. Boot up *初音ミク Project DIVA MEGA39's* (or *Hatsune Miku: Project DIVA Mega Mix*)
2. Go to the song list
3. Play *骸骨楽団とリリア* (or *Gaikotsu Gakudan to Riria*)

### Log file

[Ryujinx_1.0.6773_2021-03-04_00-06-19](https://github.com/Ryujinx/Ryujinx/files/6081854/Ryujinx_1.0.6773_2021-03-04_00-06-19.log)

### Environment?

- Ryujinx version: 1.0.6773</br>
- Game version: 1.0.7</br>
- System Specs:
  - OS: *Windows 10*
  - CPU: *AMD Ryzen 9 3950X*
  - GPU: *NVIDIA RTX 2080*
  - RAM: *32GB*
- Applied Mods: Yes
  - *1920x1080*

### Additional context?

2021-03-04_00-09-47_Ryujinx

2021-03-04_00-22-02_Ryujinx

### LINKED ISSUE #2656: No More Heroes 3 Video Cutscene Playback Issue

At a certain point in one of the cutscenes after killing the first boss (Mr. Blackhole), when the game is supposed to play a pre-recorded video instead of rendering in-game, the game appears to have trouble choosing the right video, causing it to glitch out and overlay the videos on each other. This happens again a little later when another pre-recorded video is supposed to play.

Watch the cutscenes after killing Mr. Blackhole.

### Environment?

- Ryujinx version: 1.0.7047
- Game version: 1.0.1
- System Specs:
  - OS: Windows 10 (Fully Updated)
  - CPU: AMD 5900X
  - GPU: NVIDIA RTX 3090
  - RAM: 32GB
- Applied Mods : No

### Additional context?

Video Example:
https://www.youtube.com/watch?v=kIE0QZrZlUk

Here is the log file:
[video issue.log](https://github.com/Ryujinx/Ryujinx/files/7192915/video.issue.log)

gdkchan:
This PR contains 2 changes, explained individually below. The changes are dependant of each other, which is why I decided to do both on a single PR.

**Multiple FFMPEG contexts (one per channel)**

On the initial NVDEC implementation, it was noted that multiple H264 videos playing simultaneously would not work, as all the videos shared the same FFMPEG context. At the time it was not a big deal because there were no games known to do such a thing. Eventually we did find games that needs to decode multiple videos at once (see #2077), and this started becoming an issue.

This change does nothing more than what was originally suggested by myself on the NVDEC PR. Each channel creates a NVDEC *context*, which is identified by an ID. I could simply return an object, but it would require adding more dependencies to the projects (as the type needs to be accessible) which is undesirable, so I just used an integer ID.

Internally, NVDEC gets the currently bound context from the ID that was set. The context object stores the FFMPEG context.

**Frame ordering issues**

When I implemented H264 support on NVDEC, I noticed that the frames were not in the correct order. At the time I tried to fix it, but could not find the root cause of the issue, so to avoid further delaying it I used a hacky solution where it would ignore the VIC input surface address, and instead use the address of the last NVDEC frame decoded.

This approach has issues, the most noticeable one being that it can lead to the presentation of *duplicate frames*, because if there is more than one consecutive VIC copy operation, it will copy the same frame more than once. The end result is that the H264 videos are usually not smooth, and the frame pacing is irregular.

On top of the existing problems, it also has another issue when multiple videos are decoded at once. There is no guarantee that the NVDEC decode and VIC copy for a given channel will happen in order. That is, the following scenario is possible:

```
Channel A NVDEC decodes video A
Channel B VIC copies video B frame
Channel B NVDEC decodes video B
Channel A VIC copies video A frame
```

The current setup would copy the video A frame into video B destination, which is incorrect.

So all those issues made me dig into this problem again. Fortunately, this time the endeavor was a bit more fruitful.

The problem is that FFMPEG will deliver the frame in PTS (Presentation Time Stamp) order, while NVDEC is supposed to output them in DTS (Decoder Time Stamp) order. That is, not all frames on a H264 video are decoded in the same order they are supposed to be displayed on the screen, but FFMPEG always returns them in display order, which does not match the order that NVDEC is supposed to, or that the game expects.

~~The ideal solution would be making FFMPEG output them in DTS order, however I was not able to find a way to do that (but I did not look much, and I'm not really a FFMPEG expert, so its possible that there is a way, if you know how please let me know).~~

~~The current solution is simply delaying writing the frame until it is actually returned by FFMPEG. This is done by using the `coded_picture_number` field on the `AVFrame` which contains the decoder frame index, not in presentation order but in the stream order, which is what we want.~~

The current solution calls `h264_decode_frame` with a packet size of 0 to force the delayed frame to be returned immediately.

**What does this fix?**

- Decoding multiple videos at once no longer causes image corruption.
- As a result of the hack removal, H264 video playback should be smoother now, without duplicate frames (please note that the game may still decide to duplicate frames or not present anything if the decoding is too slow, YMMV).
- Some minor issues like a few games flashing a green frame when the video starts has also been fixed.
- Add missing `field_pic_order_in_frame_present_flag` flag to the stream PPS. Fixes decoding errors on Layton's Mystery Journey, but the video is still not rendered properly due to VIC Issues.
- Add new `Trim` method on NVDEC's `SurfaceCache` to allow cached frames to be freed. It is called every time a context is destroyed. Cache access now uses a lock to make it thread safe, as `Trim` may be called from outside the NVDEC thread.

**What to test?**

Any game using H264, even better if it decodes multiple videos at once. Please make sure there was no frame ordering issues or other image corruption introduced as part of this change.

Thanks to @jduncanator for helping with FFMPEG.

Fixes #2077.
Fixes #2656.

### DIFF
--- a/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
+++ b/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
@@ -2,7 +2,6 @@
 using Ryujinx.Graphics.Device;
 using Ryujinx.Graphics.Gpu.Memory;
 using Ryujinx.Graphics.Nvdec.Image;
-using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
@@ -17,8 +16,6 @@ namespace Ryujinx.Graphics.Nvdec
         private long _currentId;
         private ConcurrentDictionary<long, NvdecDecoderContext> _contexts;
         private NvdecDecoderContext _currentContext;
-
-        public event Action<FrameDecodedEventArgs> FrameDecoded;
 
         public NvdecDevice(MemoryManager gmm)
         {
@@ -66,20 +63,15 @@ namespace Ryujinx.Graphics.Nvdec
             switch (codecId)
             {
                 case CodecId.H264:
-                    H264Decoder.Decode(this, _currentContext, _rm, ref _state.State);
+                    H264Decoder.Decode(_currentContext, _rm, ref _state.State);
                     break;
                 case CodecId.Vp9:
-                    Vp9Decoder.Decode(this, _rm, ref _state.State);
+                    Vp9Decoder.Decode(_rm, ref _state.State);
                     break;
                 default:
                     Logger.Error?.Print(LogClass.Nvdec, $"Unsupported codec \"{codecId}\".");
                     break;
             }
-        }
-
-        internal void OnFrameDecoded(CodecId codecId, uint lumaOffset, uint chromaOffset)
-        {
-            FrameDecoded?.Invoke(new FrameDecodedEventArgs(codecId, lumaOffset, chromaOffset));
         }
     }
 }
--- a/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
+++ b/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
@@ -31,6 +31,7 @@ namespace Ryujinx.Graphics.Nvdec
         {
             long id = Interlocked.Increment(ref _currentId);
             _contexts.TryAdd(id, new NvdecDecoderContext());
+
             return id;
         }
 
--- a/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
+++ b/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
@@ -40,6 +40,8 @@ namespace Ryujinx.Graphics.Nvdec
             {
                 context.Dispose();
             }
+
+            _rm.Cache.Trim();
         }
 
         public void BindContext(long id)
--- a/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
+++ b/Ryujinx.Graphics.Nvdec/NvdecDevice.cs
@@ -3,14 +3,20 @@ using Ryujinx.Graphics.Device;
 using Ryujinx.Graphics.Gpu.Memory;
 using Ryujinx.Graphics.Nvdec.Image;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading;
 
 namespace Ryujinx.Graphics.Nvdec
 {
-    public class NvdecDevice : IDeviceState
+    public class NvdecDevice : IDeviceStateWithContext
     {
         private readonly ResourceManager _rm;
         private readonly DeviceState<NvdecRegisters> _state;
+
+        private long _currentId;
+        private ConcurrentDictionary<long, NvdecDecoderContext> _contexts;
+        private NvdecDecoderContext _currentContext;
 
         public event Action<FrameDecodedEventArgs> FrameDecoded;
 
@@ -21,6 +27,30 @@ namespace Ryujinx.Graphics.Nvdec
             {
                 { nameof(NvdecRegisters.Execute), new RwCallback(Execute, null) }
             });
+            _contexts = new ConcurrentDictionary<long, NvdecDecoderContext>();
+        }
+
+        public long CreateContext()
+        {
+            long id = Interlocked.Increment(ref _currentId);
+            _contexts.TryAdd(id, new NvdecDecoderContext());
+            return id;
+        }
+
+        public void DestroyContext(long id)
+        {
+            if (_contexts.TryRemove(id, out var context))
+            {
+                context.Dispose();
+            }
+        }
+
+        public void BindContext(long id)
+        {
+            if (_contexts.TryGetValue(id, out var context))
+            {
+                _currentContext = context;
+            }
         }
 
         public int Read(int offset) => _state.Read(offset);
@@ -36,7 +66,7 @@ namespace Ryujinx.Graphics.Nvdec
             switch (codecId)
             {
                 case CodecId.H264:
-                    H264Decoder.Decode(this, _rm, ref _state.State);
+                    H264Decoder.Decode(this, _currentContext, _rm, ref _state.State);
                     break;
                 case CodecId.Vp9:
                     Vp9Decoder.Decode(this, _rm, ref _state.State);
--- a/Ryujinx.Graphics.Nvdec/Types/H264/ReferenceFrame.cs
+++ b/Ryujinx.Graphics.Nvdec/Types/H264/ReferenceFrame.cs
@@ -6,7 +6,7 @@ namespace Ryujinx.Graphics.Nvdec.Types.H264
     {
 #pragma warning disable CS0649
         public uint Flags;
-        public Array2<uint> FIeldOrderCnt;
+        public Array2<uint> FieldOrderCnt;
         public uint FrameNum;
 #pragma warning restore CS0649
 
--- a/Ryujinx.Graphics.Nvdec/Types/H264/ReferenceFrame.cs
+++ b/Ryujinx.Graphics.Nvdec/Types/H264/ReferenceFrame.cs
@@ -1,12 +1,15 @@
-﻿namespace Ryujinx.Graphics.Nvdec.Types.H264
+﻿using Ryujinx.Common.Memory;
+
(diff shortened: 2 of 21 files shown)


GamerzHell9137:
Project Diva's song runs properly now
image

jduncanator:
You can probably bypass `av_send_packet` and `av_receive_frame` completely, and just call `AVCodecContext->decode` directly. If you know when you need to flush a delayed frame, you can send an `AVPacket` with `size == 0` (and `data == null`) to force the H264 decoder to return you the delayed frame immediately.

Whilst it's not exactly getting ffmpeg to return you frames in DTS order, rather than PTS order, it should work around the issue with ffmpeg delaying P-frames until the end of the GOP.

Ryujinx.Graphics.Nvdec/Types/H264/ReferenceFrame.cs
```diff
@@ -1,12 +1,15 @@
...
 {
     struct ReferenceFrame
     {
 #pragma warning disable CS0649
-        public uint Unknown0;
-        public uint Unknown4;
-        public uint Unknown8;
-        public uint UnknownC;
+        public uint Flags;
+        public Array2<uint> FIeldOrderCnt;
```
mzji, line 9:
```suggestion
        public Array2<uint> FieldOrderCnt;
```

codbugpatrol:
Unfortunatelly I did not notice any benefitts for Tony Hawk 1+2 video still stutter. And also I've noticed some really fast glitches and hard to get in LM3 and some audio desyncronization.

Tony Hawk 1+2:
image

LM3:
image


 On the other hand, I did saw some improvements on Jump Force video, it is smoothier:

image

codbugpatrol:
Just for information, MK 11 videos are working, but no benefit noticed, have stutters and are not smooth yet.  Same behavior as mainline.  
image

gdkchan:
> 
> 
> You can probably bypass `av_send_packet` and `av_receive_frame` completely, and just call `AVCodecContext->decode` directly. If you know when you need to flush a delayed frame, you can send an `AVPacket` with `size == 0` (and `data == null`) to force the H264 decoder to return you the delayed frame immediately.
> 
> Whilst it's not exactly getting ffmpeg to return you frames in DTS order, rather than PTS order, it should work around the issue with ffmpeg delaying P-frames until the frames PTS is met.
> 
> For a bit more info on why this works. We can see [here](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/h264dec.c#L1005-L1006) that the H264 decoder will branch into a `send_next_delayed_frame` function when passed an `AVPacket` with size zero.
> 
> The problem using `av_send_packet` and sending an `AVPacket` with a size of zero, is that ffmpeg treats this "special packet" as an indicator of an EOF. This causes ffmpeg to flush the decode context and refuse to decode any future frames.
> 
> We can bypass that issue by calling directly into the H264 decoder, as the special case EOF handling for an `AVPacket` with size 0 is done in the `av_send_packet` code path, before passing the packet through to the H264 decoder.
> 
> This is a bit of a hack, as it depends on the internal behaviour of the H264 decoder, ~and it will likely not work for any of the other H264 decoders in ffmpeg (ie. hardware accel decoders)~ _hwaccel in ffmpeg is handled at a lower level than `AVCodecContext->decode`, so this should work for any implementation of h264 decode in ffmpeg_, but should get us where we need to be without modifying ffmpeg.

Thanks, it did work, however I also had to set `has_b_frames` to 0, otherwise `h264_select_output_frame` would try to pick a delayed frame and fail because all of them were already consumed. Setting it to zero worked on all the videos I tested so far.

Ryujinx.Graphics.Nvdec/NvdecDevice.cs
```diff
@@ -21,6 +24,32 @@ public NvdecDevice(MemoryManager gmm)
...
                 { nameof(NvdecRegisters.Execute), new RwCallback(Execute, null) }
             });
+            _contexts = new ConcurrentDictionary<long, NvdecDecoderContext>();
+        }
+
+        public long CreateContext()
+        {
+            long id = Interlocked.Increment(ref _currentId);
+            _contexts.TryAdd(id, new NvdecDecoderContext());
+            return id;
```
AcK77, line 34: Add an empty line before

AcK77 (approved):

Ryujinx.Graphics.Nvdec/NvdecDevice.cs
gdkchan: Why?

AcK77: For consistency?

Thog: I do agree that on the whole codebase we always have an empty line return before a return so would be better keeping i that way.

Thog (approved):
lgtm

riperiperi:
Very nice! Video presentation a lot smoother in SMO, Mario Kart 8, Xenoblade etc. All the games I have with H264 seem to work fine.

Some things have been fixed since the last time I tested:

- Pikmin 3 Demo's first few video frames aren't green, but there is a weird gray box at the top left. This might have been an issue before.
- SMO no longer flashes the wrong frame sometimes when switching kingdoms in kingdom select.
  - It still runs out of hardware syncpoints after switching video a lot of times, but it did that before. Can't really tell what the impact of this is.

riperiperi (approved):

gdkchan:
> 
> 
> Very nice! Video presentation a lot smoother in SMO, Mario Kart 8, Xenoblade etc. All the games I have with H264 seem to work fine.

XCDE uses VP9.

riperiperi:
> 
> 
> > Very nice! Video presentation a lot smoother in SMO, Mario Kart 8, Xenoblade etc. All the games I have with H264 seem to work fine.
> 
> XCDE uses VP9.

Hmm, I'm guessing it was probably the 60fps mod then, I think frame pacing is always a bit weird with 30fps right now.

gdkchan:
> 
> 
> > > Very nice! Video presentation a lot smoother in SMO, Mario Kart 8, Xenoblade etc. All the games I have with H264 seem to work fine.
> > 
> > 
> > XCDE uses VP9.
> 
> Hmm, I'm guessing it was probably the 60fps mod then, I think frame pacing is always a bit weird with 30fps right now.

Depending on what was the last time you tested, it could have been made smoother by the multithreaded VP9 frame decoded (#2009).

codbugpatrol:
Retested MK 11 with all the changes made since my last comment. It got much better, but still sttuter, some errors are thrown in the log.
image

00:04:31.850 |S| HLE.OsThread.15 ServiceMm Get: Stubbed. {id: 3}
00:04:31.850 |S| HLE.OsThread.15 ServiceMm SetAndWait: Stubbed. {id: 3 ; frequenceHz: 429488704 ; timeout: -1}
00:04:31.850 |S| HLE.OsThread.15 ServiceMm Get: Stubbed. {id: 3}
00:04:31.853 |E| Ryujinx.NvdecProcessor FFmpeg Log: [h264 @ 000002712a870cc0] illegal short term buffer state detected
00:04:31.854 |S| HLE.OsThread.15 ServiceMm SetAndWait: Stubbed. {id: 3 ; frequenceHz: 429488704 ; timeout: -1}
00:04:31.854 |S| HLE.OsThread.15 ServiceMm Get: Stubbed. {id: 3}
00:04:31.938 |S| HLE.OsThread.15 ServiceMm SetAndWait: Stubbed. {id: 3 ; frequenceHz: 83588704 ; timeout: -1}

PR #2672, opened 2021-09-20
gpu: Ensure that syncpoint callbacks will be call in the correct order

Thog:
This fix an oversight I made initially that could cause syncpoint waiter callbacks to be called in the wrong order.

Could fix some weird syncpoint related crashes (still need confirmation)

As always needs testing.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Synchronization/Syncpoint.cs
+++ b/Ryujinx.Graphics.Gpu/Synchronization/Syncpoint.cs
@@ -52,7 +52,25 @@ namespace Ryujinx.Graphics.Gpu.Synchronization
                         Callback  = callback
                     };
 
-                    _waiters.Add(waiterInformation);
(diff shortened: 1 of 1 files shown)


gdkchan:
What's wrong with calling them in the "wrong" order? As long it is really experied, I don't see why the order would matter.
Anyway, if you store the syncpoints sorted by threshold, there are other optimizations you can do here:
- `RegisterCallback` can do binary search to find the first syncpoint >= the threshold value, instead of checking all ofthem.
- `UnregisterCallback` can also use binary search to find the element to be removed faster, and use `RemoveAt` to remove by index once found.
- `Increment` can now break as soon as a syncpoint with a threshold > than the current one is found, as they are sorted, if you iterate from the start of the list. Instead of the `RemoveAll` call, you can use `RemoveRange` to remove the range of expired syncpoints.

The above changes would make it much more efficient in the presence of a large amount of syncpoint waiters.

Thog:
Closing as it doesn't matter and I do not plan to work on this anymore.

PR #2680, opened 2021-09-23
Add `Operand.Label` support to `Assembler`

FICTURE7:
This adds label support to `Assembler` and enables branch tightening when compiling with relocatables.

The instruction table was split into another file, let me know in case this is undesirable. `CodeGenContext` is now smaller, and could probably be merged into `CodeGenerator` itself.


gdkchan:
The code looks fine to me, but I'd rename the `Assembler.Table.cs` file to something like `AssemblerTable.cs`, as we don't use this naming scheme anywhere else.

ARMeilleure/CodeGen/X86/CodeGenContext.cs
```diff
@@ -1,118 +1,69 @@
...
+            int calleeSaveRegionSize = BitOperations.PopCount((uint)intMask) * 8 + xmmSaveRegionSize + 8;
 
             int argsCount = maxCallArgs;
 
             if (argsCount < 0)
             {
-                // When the function has no calls, argsCount is -1.
-                // In this case, we don't need to allocate the shadow space.
+                // When the function has no calls, argsCount is -1. In this case, we don't need to allocate the shadow
+                // space.
```
riperiperi, line 51: I feel like manual wrapping is preferrable to this. Breaking into a new line on commas and fullstops looks nicer than orphaning words.

riperiperi (approved):
LGTM, nice to see this changed.

This might significantly conflict with Ttc, so @LDj3SNuD should probably have a look.

ARMeilleure/CodeGen/X86/CodeGenContext.cs
FICTURE7: The style around comments is something I was wondering about. I wrap these comment blocks at column 120; I think its an OK tradeoff between not wasting vertical space and keeping stuff within the viewport without having a vertical scroll. But yeah, if the consensus is breaking at periods or commas, am fine by it.

LDj3SNuD:
> This might significantly conflict with Ttc (#2373), so @LDj3SNuD should probably have a look.

Already needs to be redone. I would like to remove the functionality to reuse a translation, with guest relocations, when a module is reloaded and preserve the functionality to reuse the same slot in JitCache to avoid exhaustion, as reported for Smash in Discord.

gdkchan (approved):
lgtm, thanks 👍

AcK77:
Needs a rebase

PR #2684, opened 2021-09-25
Replace CacheResourceWrite with more general "precise" write

### LINKED ISSUE #2690: Text corrupts in NooDS menu after a few seconds

When launching [NooDS](https://github.com/Hydr8gon/NooDS) (homebrew DS emulator) to the file selection menu, the menu is all fine as expected but after around 5 seconds the text becomes corrupted with these black boxes as shown below.
noodscorruptedtext

Open [NooDS](https://github.com/Hydr8gon/NooDS) and wait ~5 seconds after the menu loads in.

### Log file
[log](https://github.com/Ryujinx/Ryujinx/files/7230882/Ryujinx_1.0.7047_2021-09-26_01-53-40.log)

### Environment?

- Ryujinx version: 1.0.7047</br>
- System Specs:
  - OS: Windows 10 21H1
  - CPU: R7 3700X
  - GPU: GTX 1070 - 471.96
  - RAM: 16 GB

riperiperi:
The goal of CacheResourceWrite was to notify GPU resources when they were modified directly, by looking up the modified address/size in a structure and calling a method on each resource. The downside of this is that each resource cache has to be queried individually, they all have to implement their own way to do this, and it can only signal to resources using the same PhysicalMemory instance.

This PR adds the ability to signal a write as "precise" on the tracking, which signals a special handler (if present) which can be used to avoid unnecessary flush actions, or maybe even more. For buffers, precise writes specifically do not flush, and instead punch a hole in the modified range list to indicate that the data on GPU has been replaced.

The downside is that precise actions must ignore the page protection bits and always signal - as they need to notify the target resource to ignore the sequence number optimization.

I had to reintroduce the sequence number increment after I2M, as removing it was causing issues in rabbids kingdom battle. However - all resources modified by I2M are notified directly to lower their sequence number, so the problem is likely that another unrelated resource is not being properly updated. Thankfully, doing this does not affect performance in the games I tested. I'd like to remove this again once I figure out what the problem actually is.

This should fix regressions from #2624. Test any games that were broken by that. **(RF4, rabbids kingdom battle)**

I've also added a sequence number increment to ThreedClass.IncrementSyncpoint, as it seems to fix buffer corruption in OpenGL homebrew. (this was a regression from removing sequence number increment from constant buffer update - another unrelated resource thing)

Fixes #2690

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/GpuRegionHandle.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/GpuRegionHandle.cs
@@ -4,6 +4,9 @@ using System;
 
 namespace Ryujinx.Graphics.Gpu.Memory
 {
+    /// <summary>
+    /// A tracking handle for a region of GPU VA, represented by one or more tracking handles in CPU VA.
+    /// </summary>
     class GpuRegionHandle : IRegionHandle
     {
         private readonly CpuRegionHandle[] _cpuRegionHandles;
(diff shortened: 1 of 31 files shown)


Ryujinx.Graphics.Gpu/Memory/GpuRegionHandle.cs
```diff
@@ -49,6 +49,14 @@ public void RegisterAction(RegionSignal action)
             }
         }
 
+        public void RegisterPreciseAction(PreciseRegionSignal action)
```
gdkchan, line 52: Missing XML docs.

riperiperi: I notice none of these methods have docs right now 🤔

GamerzHell9137:
RF4 Live2D elements are fixed, 3D yet to be fixed
image

AcK77 (approved):

PR #2687, opened 2021-09-25
Force copy when auto-deleting a texture with dependencies

riperiperi:
When a texture is deleted by falling to the bottom of the AutoDeleteCache, its data is flushed to preserve any GPU writes that occurred. This ensures that the data appears in any textures recreated in the future, but didn't account for a texture that already existed with a copy dependency.

This change forces copy dependencies to complete if a texture falls out from from the AutoDeleteCache. (not removed via overlap, as that would be wasted effort) This is done via a full sync of the storage texture right now, as that's the best way to ensure all the rules are followed.

Fixes broken lighting caused by pausing in SMO's Metro Kingdom. May fix some other infrequent issues. 

I'd recommend testing games that do a lot of render to texture / copies, like 1st party games and UE4 games.

### Before (pause+unpause)
image
(too bright, weird directional lighting on mario)

### After (pause+unpause)
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -41,14 +41,14 @@ namespace Ryujinx.Graphics.Gpu.Image
             {
                 Texture oldestTexture = _textures.First.Value;
 
-                oldestTexture.SynchronizeMemory();
-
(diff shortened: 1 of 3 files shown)


NikkMann:
Looks like this fixed the occasional bugged out lighting on Travis when loading into his hotel in No More Heroes 3. Nice.

AcK77 (approved):

PR #2692, opened 2021-09-26
clkrst: Stub/Implement IClkrstManager and IClkrstSession calls

### LINKED ISSUE #2686: pcv clkrst: 0 (OpenSession) is not implemented

pcv clkrst: 0 (OpenSession) is not implemented.  [8.0.0+]</br>  

[Switchbrew](https://switchbrew.org/wiki/PCV_services#OpenSession)

### Service description
```
Takes an input DeviceCode and an input u32. Returns an #IClkrstSession.
```

### Required by: 
[NooDS](https://github.com/Hydr8gon/NooDS)

AcK77:
This PR stubs and implements some clkrst call because they are used to overclock the Switch hardware and it's pointless in our case as we emulate the system.
Everything was done checked by RE.

Fixes #2686


riperiperi:
image
Seems to fix the related homebrew (along with #2684 which fixes the opengl issue) though it does run quite slow.

riperiperi (approved):
LGTM. Maybe there should be a default `_clockRate` for some device codes? Not sure what a game might do if it read out `0` as the clock rate without setting it.

AcK77:
> LGTM. Maybe there should be a default `_clockRate` for some device codes? Not sure what a game might do if it read out `0` as the clock rate without setting it.

https://github.com/retronx-team/sys-clk/blob/master/sysmodule/src/clock_manager.cpp#L144-L150

They seems to handle this case.

Thog (approved):

PR #2693, opened 2021-09-26
Use normal memory store path for DC ZVA

riperiperi:
Seems like this is used as an optimized way to clear memory in homebrew applications. Unfortunately, calling the software fallback method every 8 bytes was not very optimal.

The existing EmitStore is used by passing in ZR as the register to get a 0 write.


gdkchan (approved):
Nice, I forgot this was a thing.

Thog (approved):

LDj3SNuD (approved):

PR #2695, opened 2021-09-29
Improve `HyrbidAllocator`

FICTURE7:
Currently once a local has been marked for spilling, it is spilled for its entire lifetime, i.e: a spill when it is defined and fill when it is used. This simplifies and speeds up allocation a lot, but at the cost of CQ.

This PR attempts to improve CQ of the allocator by tracking more precisely the lifetime of locals in registers as they are allocated. It removes the notion of spill set, instead every register in the free set is a candidate for spilling and filling. The allocator has an active set like the LSRA, and has 2 primary operations: `DefRegister` and `UseRegister`. The allocation is done as we're walking through the operations of the block in forward order. The free set still have the constraints of the current approach.

### `DefRegsister`
It is called when a local is defined/assigned:
1. If the local already has a register assigned, return that register.
2. If there is a free register (i.e: not in active set), return that register.
   * This could be based on a heuristic, but is not at the moment.
   * The register is moved to the active set.
3. Otherwise select a register from the active set to spill, return that register.
   * This is based on a simple heuristic.

### `UseRegister`
It is called when a local is used:
1. If the local already has a register, return that register.
2. If it does not, allocate one using `DefRegister` , fill the register, return that register.
   * Using `DefRegister` will take care of the case where we need to spill.
3. If the local has reached the last use, we can free the register and remove it from the active set.
   * The register is spilled when the local is not a block local, this is to avoid issues with loops.

After allocating the operations in the block, the active set may not be empty. In this case we spill everything in it. This is essentially a local linear scan second chance register allocator. Since when locals are evicted, they have chance to live in another register later on.

___
In the majority of cases this is an improvement over the current allocator, in term of code size (7415 improved, 1948 regressed).

#### CQ Improvements
Typical improvements looks like this.

<table>
  <tr><th>Master</th><th>PR</th></tr>
  <tr>
  <td>

  ```asm
  .L17:
    mov rax,9150000h
    mov [rsp+60h],rax
    mov rax,[rsp+60h]
    lea rbp,[rax+53Eh]
    mov rax,rbp
    mov [rsp+60h],rax
    mov rax,801D6BCh
    mov [rsp+68h],rax
    mov rbp,[rsp+4Ch]
    mov rax,[rsp+60h]
    mov [rbp],rax
    mov rax,[rsp+68h]
    mov [rbp+0F0h],rax
    mov rbx,[rsp+4Ch]
    mov qword [rbx+408h],8DE5730h
    mov rbp,DISPATCH_TABLE_8DE5730
    mov rbp,[rbp]
    mov rcx,rbx
    call rbp
    mov rdx,rax
    mov [rsp+0A0h],rdx
    mov rax,[rsp+0A0h]
    cmp rax,801D6BCh
    je near .L19
  ```
  </td>
  <td>

  ```asm
  .L17:
    mov rax,9150000h
    lea rcx,[rax+53Eh]
    mov rax,rcx
    mov rcx,801D6BCh
    mov rdx,[rsp+44h]
    mov [rdx],rax
    mov [rdx+0F0h],rcx
    mov [rsp+58h],rax
    mov [rsp+60h],rcx
    mov rbx,[rsp+44h]
    mov qword [rbx+408h],8DE5730h
    mov rbp,DISPATCH_TABLE_8DE5730
    mov rsi,[rbp]
    mov rcx,rbx
    call rsi
    mov rbx,rax
    mov [rsp+98h],rbx
    mov rax,[rsp+98h]
    mov [rsp+98h],rax
    cmp rax,801D6BCh
    je near .L19
  ```
  </td>
  </tr>
  <tr>
  <td>

  ```asm
  .L5:
    mov rbp,[rsp+4Ch]
    mov rsi,PAGE_TABLE
    mov rbp,[rsi+rbp+8]
    mov rax,rbp
    mov [rsp+60h],rax
    mov rax,[rsp+60h]
    mov [rsp+60h],rax
    mov rax,[rsp+60h]
    test rax,rax
    jne near .L7
  ```
  </td>
  <td>

  ```asm
  .L5:
    mov rax,[rsp+54h]
    mov rcx,PAGE_TABLE
    mov rdx,[rcx+rax+8]
    mov rax,rdx
    mov [rsp+60h],rax
    test rax,rax
    jne near .L7
  ```
  </td>
  </tr>
</table>

#### CQ Regressions
I have also seen some regressions, but I did not have the chance to investigate them yet.
* Because of the sub-rule of rule 3 of `UseRegister` we over spill sometimes.

  <table>
    <tr><th>Master</th><th>PR</th></tr>
    <tr>
    <td>
    
    ```asm
    .L2:
      mov ebp,[rsp+84h]
      sub ebp,1
      mov rax,[rsp+7Ch]
      mov [rax+400h],ebp
    ```
    </td>
    <td>
    
    ```asm
    .L2:
      mov eax,[rsp+84h]
      sub eax,1
      mov rcx,[rsp+54h]
      mov [rsp+54h],rcx  ;; This is the last use and a non-block local so we spill.
      mov [rcx+400h],eax
    ```
    </td>
    </tr>
  </table>
  This could be fixed by doing a more precise liveness analysis and checking if the local lives out the block, but it might throw throughput out of the window.
* It uses more registers than the current approach which causes epilogue size to increase.

#### Throughput
I also did not have the chance to investigate throughput or boot time yet. Will probably close this if I can't get it to be close to the current approach, if it happens to be slower.

### DIFF
--- a/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
@@ -41,10 +41,10 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
 
             public bool IsBlockLocal => _first == _last;
 
-            public LocalInfo(OperandType type, int uses, int blkIndex)
+            public LocalInfo(Operand local, int uses, int blkIndex)
             {
                 Uses = uses;
-                Type = type;
+                Type = local.Type;
 
                 UsesAllocated = 0;
                 Sequence = 0;
@@ -72,7 +72,8 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
             }
         }
 
-        private const int MaxIROperands = 4;
+        private const int RegistersCount = 16;
+
         // The "visited" state is stored in the MSB of the local's value.
         private const ulong VisitedMask = 1ul << 63;
 
@@ -173,7 +174,7 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
                                 }
 
                                 SetVisited(dest);
-                                GetLocalInfo(dest) = new LocalInfo(dest.Type, UsesCount(dest), block.Index);
+                                GetLocalInfo(dest) = new LocalInfo(dest, UsesCount(dest), block.Index);
                             }
                         }
                         else if (dest.Kind == OperandKind.Register)
@@ -193,7 +194,8 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
                 _blockInfo[block.Index] = new BlockInfo(hasCall, intFixedRegisters, vecFixedRegisters);
             }
 
-            int sequence = 0;
+            Operand[] intActive = new Operand[RegistersCount];
+            Operand[] vecActive = new Operand[RegistersCount];
 
             for (int index = cfg.PostOrderBlocks.Length - 1; index >= 0; index--)
             {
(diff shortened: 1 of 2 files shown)


ARMeilleure/CodeGen/RegisterAllocators/HybridAllocator.cs
```diff
@@ -296,154 +246,265 @@ Operand AllocateRegister(Operand local)
                         }
                     }
 
+                    // If the operation is folded to a fill, no need ot inspect sources; since sources of fills are
```
Mou-Ikkai, line 249:
```suggestion
                    // If the operation is folded to a fill, no need to inspect sources; since sources of fills are
```

Mou-Ikkai (commented):
Small grammar correction.

gdkchan:
> This could be fixed by doing a more precise liveness analysis and checking if the local lives out the block, but it might throw throughput out of the window.

I think it can simply not spill if the variable is never assigned to inside the block. For that it only needs to keep track of the blocks where it is assigned on. If the variable only has a single assignment, then this could be a single block index on the local info.

Tests are failing on linux and mac, apparently it triggers an IOR exception on the register allocator.

FICTURE7:
> I think it can simply not spill if the variable is never assigned to inside the block. For that it only needs to keep track of the blocks where it is assigned on. If the variable only has a single assignment, then this could be a single block index on the local info.

Yeah I was thinking of this as well, but it will probably catch less cases. Still better than what it is currently and a lot cheaper than the liveness analysis.

> Tests are failing on linux and mac, apparently it triggers an IOR exception on the register allocator.

Yeah it is because on system V, all vectors are caller saved and the RA will not consider caller saved registers. So the free set for vector registers is empty and cannot evict anything. On master it will go through the spill set then filled and spilled, so special handling needed.

gdkchan:
I have been thinking about this, and I think for functions without loops, we can simply extend this approach to work across basic blocks. I'm not sure if it would be worth it, it depends on how many functions have backwards branches in general, but it's something that could help us remove spilling across blocks without much effort.

FWIW, when I was looking at register allocation in 2019, I found a similar register allocation strategy being used on TCC. You can read more about it here if you are interested: https://www.complang.tuwien.ac.at/Diplomarbeiten/falbesoner14.pdf The relevant chapters are 6 and 7. They propose a simple approach to find the live intervals even in the presence of backwards branches.

FICTURE7:
Actually loops and if-else like also CFG requires sub-rule of rule 3 of `UseRegister`. So can only do that if the control flow is a straight line of code. More precisely (I believe), we can only do these kinds of allocation on traces. A trace is a set of basic-block which has a single entry and contains no internal branches, but it can contain external branches. Having no internal branches means we do not need to handle resolution code at edges of the BBs.

Here is a thesis I found with similar goals: https://ssw.jku.at/General/Staff/Eisl/papers/phdthesis.pdf. Relevant chapters is 4, although I have not read it thoroughly myself.

> FWIW, when I was looking at register allocation in 2019, I found a similar register allocation strategy being used on TCC. You can read more about it here if you are interested: https://www.complang.tuwien.ac.at/Diplomarbeiten/falbesoner14.pdf The relevant chapters are 6 and 7. They propose a simple approach to find the live intervals even in the presence of backwards branches.

Took a quick glance at it; a very neat approach. This may come in handy at some point.

> Edit: I just noticed that you already do something like this on your latest commit 😅

Yeah, it works on extended basic blocks which should be able to handle straight translations, although the constraints on having matching fixed register set could be relaxed by spilling fixed registers of the next block which are active.

gdkchan:
Needs a rebase.

PR #2701, opened 2021-10-03
Don't force scaling on 2D copy sources

riperiperi:
Some games (GameMaker Studio) build texture atlases out of sprites during initialization, using the 2D copy method. These copies are done from textures loaded into memory, not rendered, so they are not scaled to begin with.

I had set srcTexture in these copies to force scaling, but really it only needs to scale if the texture already exists and was scaled by rendering or something else. I just set that to false, so it doesn't change if the texture is scaled or not. This will also avoid the destination being scaled if the source wasn't. The copy can handle mismatching scales just fine.

This prevents scaling artifacts in GMS games, and maybe others (not Super Mario Maker 2, that has another issue).

Before (2x):
image

After (2x):
image

This will probably not prevent scale blacklisting (forced 1x) in Deltarune Chapters 1&2.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
@@ -119,7 +119,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Twod
                 srcCopyTexture,
                 offset,
                 srcCopyTextureFormat,
-                true,
+                false,
                 srcHint);
 
             if (srcTexture == null)


codbugpatrol:
Rip, I was trying your PR here and while I was scaling from 2x to 4x Dragon Ball FighterZ, I noticed that the background was almost fixed. Got confused here. It seems that the background problem seems to be related to something that 4x may enhance.

1x or 2x:

https://user-images.githubusercontent.com/85196112/135772082-0c1d2e80-02b7-4ce7-bcb9-02ccec617f6c.mp4


4x :

https://user-images.githubusercontent.com/85196112/135772089-3ad59e14-a787-4a32-a5c4-4faca78de986.mp4

gdkchan (approved):
Nice.

Thog (approved):

PR #2705, opened 2021-10-04
spl: Implement IGeneralInterface GetConfig

AcK77:
This PR implement the GetConfig call of the spl service. This is currently needed for some homebrews which currently needs Ignore Missing Service to boot. Now it's fixed.
Implementation was done using Atmosphère code and REing too.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Spl/IGeneralInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Spl/IGeneralInterface.cs
@@ -1,13 +1,182 @@
-﻿namespace Ryujinx.HLE.HOS.Services.Sm
+﻿using Ryujinx.Common.Logging;
+using Ryujinx.HLE.FileSystem.Content;
+using Ryujinx.HLE.HOS.Kernel.Common;
+using Ryujinx.HLE.HOS.Services.Spl.Types;
+
+namespace Ryujinx.HLE.HOS.Services.Spl
 {
     [Service("spl:")]
     [Service("spl:es")]
     [Service("spl:fs")]
     [Service("spl:manu")]
     [Service("spl:mig")]
-    [Service("spl:ssl")] 
+    [Service("spl:ssl")]
     class IGeneralInterface : IpcService
     {
         public IGeneralInterface(ServiceCtx context) { }
+
+        [CommandHipc(0)]
+        // GetConfig(u32 config_item) -> u64 config_value
+        public ResultCode GetConfig(ServiceCtx context)
+        {
+            ConfigItem configItem = (ConfigItem)context.RequestData.ReadUInt32();
+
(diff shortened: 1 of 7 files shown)


Ryujinx.HLE/HOS/Services/Spl/IGeneralInterface.cs
```diff
@@ -1,13 +1,182 @@
...
+
+            SystemVersion version = context.Device.System.ContentManager.GetCurrentFirmwareVersion();
+
+            switch (configItem)
+            {
+                case ConfigItem.DisableProgramVerification:
+                    configValue = 0;
+                    break;
+                case ConfigItem.DramId:
+                    if (context.Device.Configuration.MemoryConfiguration.ToKernelMemorySize() == MemorySize.MemorySize8GB)
```
gdkchan, line 70: Can you store `context.Device.Configuration.MemoryConfiguration.ToKernelMemorySize()` on a local variable and use that, to avoid calling it twice?

Ryujinx.HLE/HOS/Services/Spl/IGeneralInterface.cs
```diff
@@ -1,13 +1,182 @@
...
+                    return SmcResult.NotImplemented;
+                case ConfigItem.QuestState:
+                    return SmcResult.NotImplemented;
+                case ConfigItem.RegulatorType:
+                    return SmcResult.NotImplemented;
+                case ConfigItem.DeviceUniqueKeyGeneration:
+                    return SmcResult.NotImplemented;
+                case ConfigItem.Package2Hash:
+                    return SmcResult.NotImplemented;
+                case ConfigItem.ExosphereApiVersion:
```
gdkchan, line 119: Why do we need those exosphere config items? IMO we should aim to just support what the official firmware does, without any atmosphere extensions. Is any standard homebrew actualy requiring that to work?

AcK77: Yeah some homebrews apparently use that to know the atmosphère version

gdkchan: Which homebrew requires it? I still think it's not something we need to support as it's not something that exists on the official firmware.

AcK77: https://github.com/ultimate-research/UltimateModManager/blob/08999dbc5b1858207a636909a78f51933ad83cb9/source/utils.h#L208

gdkchan: This code is trying to detect the current CFW in use and even emulators. Why do we need to pretend it is running atmosphere? I don't think we should aim to support this tbh.

gdkchan (approved):
lgtm, thanks.

Thog (approved):

PR #2640, opened 2021-09-14
vi: Unify resolutions values and accurate implementation of them.

AcK77:
To continue what was made in #2618, I've REd `vi` service a bit. Now values and checks related to displays are more accurate.

- `am`  GetDefaultDisplayResolution / GetDefaultDisplayResolutionChangeEvent have more informations on what the service does.
- `vi:u/vi:m/vi:s` GetDisplayService are now accurate.
- `IApplicationDisplay` GetRelayService, GetSystemDisplayService, GetManagerDisplayService, GetIndirectDisplayTransactionService, ListDisplays, OpenDisplay, OpenDefaultDisplay, CloseDisplay, GetDisplayResolution are now properly implemented.
- Some other calls are cleaned or have extra checks accordingly to RE.

Additionnaly, `IFriendService` have some wrong aligned things, and `pm:info` service placeholder was missing.

PR #2708, opened 2021-10-05
Fix DisplayInfo struct

gdkchan:
Fixes a regression caused by #2640. The `DisplayInfo` struct was incorrect, the `Array40` struct was used for the name fixed buffer, that has a size of 0x40 bytes. However, the size on the `Array` struct name is decimal, not hexadecimal, so the correct type here is `Array64`.

Fixes regression that would cause Dragon Ball Xenoverse 2 to no longer boot, as it would pass a invalid size of 0 to surface flinger initialization, which would later cause other failures.

I also removed the `Fill` call as it was useless, the `Write` already writes all the 0x60 bytes. Also fixed the write address, that was not taking the current `i` value into account, although it was not an issue right now as only one entry is ever written on `ListDisplays` currently.


Thog (approved):
lgtm

AcK77 (approved):

PR #2751, opened 2021-10-18
Clean up rejit queue

FICTURE7:
Clean up management of the rejit queue by creating a new class `TranslatorQueue` and separating it from `Translator`.


FICTURE7:
Test failures is related; will try to get a fix tomorrow.

gdkchan:
Since I was requested to review, the reason I did not approve is because my feedback on https://github.com/Ryujinx/Ryujinx/pull/2751#discussion_r734891666 was not addressed.

FICTURE7:
> Since I was requested to review, the reason I did not approve is because my feedback on https://github.com/Ryujinx/Ryujinx/pull/2751#discussion_r734891666 was not addressed.

Yeah, apologies for that. Finally got the chance to address this.

ARMeilleure/Translation/Translator.cs
```diff
@@ -317,6 +272,27 @@ internal TranslatedFunction Translate(ulong address, ExecutionMode mode, bool hi
...
+                    EnqueueForDeletion(key, oldFunc);
+                    return func;
+                });
+
+                if (PtcProfiler.Enabled)
+                {
+                    PtcProfiler.UpdateEntry(request.Address, request.Mode, highCq: true);
+                }
+
+                RegisterFunction(request.Address, func);
```
gdkchan, line 292: This code looks a little bogus, it seems that nothing is stopping a scenario like:
- Thread A translates a guest function and adds it to `Functions`.
- Thread B translates the same guest function and replaces it on `Functions`.
- Thread B enqueues the function produced by thread A for deletion registers the new function on the table.
- Thread A registers the function on the table but it was already enqueued for deletion by B.

Not a new problem of course, and maybe this scenario is not even possible right now, just think it's worth to point that out somewhere.

gdkchan (approved):
lgtm, thanks for addressing this!

ARMeilleure/Translation/Translator.cs
FICTURE7: True it does look suspicious, I guess `Functions` and `FunctionTable` needs to be updated atomically together.

gdkchan: `GetOrTranslate` does something different, it deletes the *new* function instead, rather than the old one, and uses `GetOrAdd` for that. I think this approach does not have issues.

riperiperi: We determine whether or not to rejit a function by checking if its call count is _equal_ to the number required for rejit, which generally means that a guest function is only translated this way once. However, the call counter isn't atomic right now:

```
            Operand curCount = context.Load(OperandType.I32, address);
            Operand count = context.Add(curCount, Const(1));
            context.Store(address, count);
            context.BranchIf(lblEnd, curCount, Const(MinsCallForRejit), Comparison.NotEqual, BasicBlockFrequency.Cold);
```

So it is possible that this could call multiple times. I think due to the nature of this kind of race, it's not possible for it to miss the MinCallsForRejit. I think an issue could happen here on the existing code, but I think you'd need to trigger this weird case, and then you'd need to trigger the case gdk mentioned on top of that, so it'll be super rare.

riperiperi (approved):
LGTM, nice to see this cleaned up.

Thog (approved):

ARMeilleure/Translation/Translator.cs
FICTURE7: Here is a potential fix. Although I believe it ultimately depends on the semantics of `EnqueueForDeletion` and when the deletion happens.

```patch
diff --git a/ARMeilleure/Translation/Translator.cs b/ARMeilleure/Translation/Translator.cs
index ee8e3e8b..0e579faa 100644
--- a/ARMeilleure/Translation/Translator.cs
+++ b/ARMeilleure/Translation/Translator.cs
@@ -14,6 +14,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Threading;
 using static ARMeilleure.IntermediateRepresentation.Operand.Factory;
 
@@ -277,11 +278,21 @@ namespace ARMeilleure.Translation
             while (_threadCount != 0 && Queue.TryDequeue(out RejitRequest request))
             {
                 TranslatedFunction func = Translate(request.Address, request.Mode, highCq: true);
+                TranslatedFunction deleteFunc = null;
 
                 Functions.AddOrUpdate(request.Address, func.GuestSize, func, (key, oldFunc) =>
                 {
-                    EnqueueForDeletion(key, oldFunc);
-                    return func;
+                    if (oldFunc.HighCq)
+                    {
+                        JitCache.Unmap(func.FuncPtr);
+                        func = oldFunc;
+                        return oldFunc;
+                    }
+                    else
+                    {
+                        deleteFunc = oldFunc;
+                        return func;
+                    }
                 });
 
                 if (PtcProfiler.Enabled)
@@ -290,6 +301,11 @@ namespace ARMeilleure.Translation
                 }
 
                 RegisterFunction(request.Address, func);
+
+                if (deleteFunc != null)
+                {
+                    EnqueueForDeletion(request.Address, deleteFunc);
+                }
             }
         }
```

PR #2763, opened 2021-10-20
Add support for render scale to vertex stage.

riperiperi:
Occasionally games read off textureSize on the vertex stage to inform the fragment shader what size a texture is without querying in there. Scales were not present in the vertex shader to correct the sizes, so games were providing the raw upscaled texture size to the fragment shader, which was incorrect.

One downside is that the fragment and vertex support buffer description must be identical, so the full size scales array must be defined when used. I don't think this will have an impact though. Another is that the fragment texture count must be updated when vertex shader textures are used. I'd like to correct this so that the update is folded into the update for the scales.

Also cleans up a bunch of things, like it making no sense to call CommitRenderScale for each stage.

Fixes render scale causing a weird offset bloom in Super Mario Party and Clubhouse Games. Clubhouse Games still has a pixelated look in a number of its games due to something else it does in the shader.

Before:
image

After:
smp3x

### DIFF
--- a/Ryujinx.Graphics.GAL/SupportBufferUpdater.cs
+++ b/Ryujinx.Graphics.GAL/SupportBufferUpdater.cs
@@ -1,0 +1,93 @@
+﻿using Ryujinx.Graphics.Shader;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Graphics.GAL
+{
+    public class SupportBufferUpdater : IDisposable
+    {
+        public SupportBuffer Data;
+        public BufferHandle Handle;
+
+        private IRenderer _renderer;
+        private int _startOffset;
+        private int _endOffset;
+
+        public SupportBufferUpdater(IRenderer renderer)
+        {
+            _renderer = renderer;
+            Handle = renderer.CreateBuffer(SupportBuffer.RequiredSize);
+        }
+
+        private void MarkDirty(int startOffset, int byteSize)
+        {
+            int endOffset = startOffset + byteSize;
+
+            if (_startOffset == -1)
+            {
+                _startOffset = startOffset;
+                _endOffset = endOffset;
+            }
+            else
+            {
+                if (startOffset < _startOffset)
+                {
+                    _startOffset = startOffset;
+                }
+
+                if (endOffset > _endOffset)
+                {
+                    _endOffset = endOffset;
+                }
+            }
+        }
+
+        public void UpdateFragmentRenderScaleCount(int count)
+        {
+            if (Data.FragmentRenderScaleCount.X != count)
+            {
+                Data.FragmentRenderScaleCount.X = count;
+
+                MarkDirty(SupportBuffer.FragmentRenderScaleCountOffset, sizeof(int));
+            }
+        }
+
+        private void UpdateGenericField<T>(int baseOffset, ReadOnlySpan<T> data, Span<T> target, int offset, int count) where T : unmanaged
+        {
+            data.Slice(0, count).CopyTo(target.Slice(offset));
+
+            int elemSize = Unsafe.SizeOf<T>();
+
+            MarkDirty(baseOffset + offset * elemSize, count * elemSize);
+        }
+
+        public void UpdateRenderScale(ReadOnlySpan<Vector4<float>> data, int offset, int count)
+        {
+            UpdateGenericField(SupportBuffer.GraphicsRenderScaleOffset, data, Data.RenderScale.ToSpan(), offset, count);
+        }
+
+        public void UpdateFragmentIsBgra(ReadOnlySpan<Vector4<int>> data, int offset, int count)
+        {
+            UpdateGenericField(SupportBuffer.FragmentIsBgraOffset, data, Data.FragmentIsBgra.ToSpan(), offset, count);
+        }
+
+        public void Commit()
+        {
+            if (_startOffset != -1)
+            {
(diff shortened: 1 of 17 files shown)


Shahil-Ayato:
This fixes the weird line artifacts when you upscale in in HW:AOC but it seems to also cause rainbow textures even on native res, I tested the latest master and the latest PR build
CPU i7 8700
GPU GTX 1070
Driver 510.31

Master 
https://imgur.com/a/dSA1BD9
https://imgur.com/a/y50LnuT

Pr build 
https://imgur.com/a/7SarbS9
https://imgur.com/a/IGLm8h6

Pr build native 
https://imgur.com/a/xdejMRu

riperiperi:
I've changed the support buffer to be represented by a managed struct, and have its management handled by a SupportBufferUpdater in GAL that tracks regions of the support buffer that have been dirtied, and commits them before a draw or compute dispatch.

Now ready for review (and re-test!)

Ryujinx.Graphics.GAL/SupportBufferUpdater.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Ryujinx.Graphics.GAL
+{
+    public class SupportBufferUpdater : IDisposable
+    {
+        public SupportBuffer Data;
+        public BufferHandle Handle;
+
+        private IRenderer _renderer;
+        private int _startOffset;
```
gdkchan, line 14: Shouldn't it be initialized to -1?

Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -232,42 +232,44 @@ private bool UpdateScale(Texture texture, TextureBindingInfo binding, int index,
...
 
-                            if (activeTarget != null && activeTarget.Info.Width / (float)texture.Info.Width == activeTarget.Info.Height / (float)texture.Info.Height)
+                            if (scale != 1)
                             {
-                                // If the texture's size is a multiple of the sampler size, enable interpolation using gl_FragCoord. (helps "invent" new integer values between scaled pixels)
-                                result = -scale;
-                                break;
+                                Texture activeTarget = _channel.TextureManager.GetAnyRenderTarget();
+
+                                if (activeTarget != null && activeTarget.Info.Width / (float)texture.Info.Width == activeTarget.Info.Height / (float)texture.Info.Height)
```
gdkchan, line 251:
```suggestion
                                if (activeTarget != null && (activeTarget.Info.Width / (float)texture.Info.Width) == (activeTarget.Info.Height / (float)texture.Info.Height))
```
Easier to understand imo.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -95,13 +88,13 @@ internal Pipeline()
             _tfbTargets = new BufferRange[Constants.MaxTransformFeedbackBuffers];
         }
 
-        public void Initialize()
+        public void Initialize(Renderer renderer)
         {
-            _supportBuffer = Buffer.Create(SupportBuffer.RequiredSize);
-            GL.BindBufferBase(BufferRangeTarget.UniformBuffer, 0, Unsafe.As<BufferHandle, int>(ref _supportBuffer));
+            _supportBuffer = new GAL.SupportBufferUpdater(renderer);
```
gdkchan, line 93:
```suggestion
            _supportBuffer = new SupportBufferUpdater(renderer);
```

Ryujinx.Graphics.Shader/Ryujinx.Graphics.Shader.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="CodeGen\Glsl\HelperFunctions\TexelFetchScale_vp.glsl" />
```
gdkchan, line 8: Odd this ItemGroup was only added for this single file. I guess it's a VS thing?

Ryujinx.Graphics.Shader/SupportBuffer.cs
```diff
@@ -1,18 +1,49 @@
...
         // One for the render target, 32 for the textures, and 8 for the images.
         public const int RenderScaleMaxCount = 1 + 32 + 8;
 
-        public const int RequiredSize = FragmentRenderScaleOffset + RenderScaleMaxCount * FieldSize;
+        static SupportBuffer()
+        {
+            FieldSize = Unsafe.SizeOf<Vector4<float>>();
+            RequiredSize = Unsafe.SizeOf<SupportBuffer>();
+
+            FragmentAlphaTestOffset = (int)Marshal.OffsetOf<SupportBuffer>(nameof(FragmentAlphaTest));
```
gdkchan, line 35: I would recommend doing something similar to what is done for the `NativeCtxStorage` on the CPU emulator, where it uses `Unsafe.ByteOffset`. I guess it would be a bit more complicated though, as you need actual references rather than just passing the field name. AFAIK `Marshal.OffsetOf` returns an offset for the equivalent marshalled C struct, which might not match the layout of the managed struct, so the offset could be different depending on the types used on the struct.

Ryujinx.Graphics.Shader/Ryujinx.Graphics.Shader.csproj
gdkchan: FWIW I really want to just use const string for those helper functions on a static class. I think it would be easier to manage.

riperiperi:
Feedback addressed. I'm using `Unsafe.ByteOffset` by creating a temporary instance of the struct. It's not too messy.

gdkchan (approved):
lgtm, thanks.

Thog (approved):

PR #2764, opened 2021-10-20
Clean up `JitCache`

FICTURE7:
Rebase of #2169 with a different approach at managing the `NativeSignalHandler` and some minor clean ups.


marysaka:
Will need a revamp and rebase

PR #2770, opened 2021-10-22
sdl2: Enable GameCube controller support

Thog:
This PR set the SDL2 hint to allow GameCube controller to be supported.

Feedback as always needed.


Thog:
Will need a new iteration and I do not plan to work on that any time soon.

PR #2801, opened 2021-11-01
Add support for advanced blend (part 1/2)

### LINKED ISSUE #2803: MP Superstars issues

There's the issue of black water effects trailing the boats in the minigame 'Puddle Paddle'. 
Which can be seen in this image: ryujinx_capture_2021-11-02_10-59-21
I am using Ryujinx LDN2.4 in handheld mode at 2x resolution. However, the issue still occurs with resolution scaling disabled and enabling docked mode.

There is also an issue with the warp and squish effects used in 2 player games when someone wins/loses where it tanks the performance from 60fps to 8fps, however it only occurs during that squish transition which I'm not too concerned about. I do not have an image for this as it's performance related

It seems that just starting the minigame 'Puddle Paddle' should show the issue for the water bug

For the squish transition tanking performance, you can see that best in the minigame 'Spin Doctor' once you win or lose the minigame

### Environment?

- Ryujinx version: 1.0.LDN-2.4
- Game version: 1.1.0</br>
- System Specs:
  - OS: Windows 10
  - CPU: i5-4570
  - GPU: NVIDIA GTX 1050ti
  - RAM: 16GB
- Applied Mods : No

gdkchan:
The Switch GPU supports fully programmable blend using microcode, however no API actually exposes this functionality. Instead, they expose the functionality documented on the advancled blend extension, with several different blending modes in addition to the regular ones. This implements "HLE" support for this functionality by matching microcode and using the respective advanced blend mode on the host. The plan is implementing the "LLE" approach on part 2/2, for devices where the extension is not supported, or when the match fails. It is worth noting that only NVIDIA supports this extension on desktop. Other vendors support a subset of it, that is missing several blend modes and other parameters.

This fixes some transparency issues on Mario Party Superstars:
- Black particles on Bowser explosion minigame (I forgot the name).
- Black water on Puddle Paddle and Pushy Penguins minigames.

Before:
image
image
image
After:
image
image
image

Fixes #2803.


smiglo5:
Awesome PR!
If you somehow manage to fix also missing shaders and Eatsa Pizza issue that I've mentioned in these posts
https://github.com/Ryujinx/Ryujinx-Games-List/issues/3752#issuecomment-955780588

https://github.com/Ryujinx/Ryujinx-Games-List/issues/3752#issuecomment-956130809

I believe all game will be perfectly emulated

Bowser's game is Bowser's Big Blast by the way :P

GamerzHell9137:
> 
> 
> Awesome PR! If you somehow manage to fix also missing shaders and Eatsa Pizza issue that I've mentioned in these posts [Ryujinx/Ryujinx-Games-List#3752 (comment)](https://github.com/Ryujinx/Ryujinx-Games-List/issues/3752#issuecomment-955780588)
> 
> [Ryujinx/Ryujinx-Games-List#3752 (comment)](https://github.com/Ryujinx/Ryujinx-Games-List/issues/3752#issuecomment-956130809)
> 
> I believe whole game will be perfectly emulated
> 
> Bowser's game is Bowser's Big Blast by the way :P

The game has flashings too.

legend800:
Does this fix Super Mario Party black squares too? PR builds is no longer downloadable. 
2022-02-06 11_57_56-Ryujinx 1 1 0+74fb8fc - Super Mario Party v1 0 0 (010036B0034E4000) (64-bit)

gdkchan:
This is ready to review now. I added a hash table based approach to find the blend op from the microcode. So the assembler and the large functions class is no longer needed. I'm not sure if I should leave it in the code or remove it (it could be useful for re-generating the hashes).

Also I'm not sure what to do when the advanced blend extension is not supported for now.

Ryujinx.Graphics.OpenGL/EnumConversion.cs
```diff
@@ -34,6 +34,126 @@ public static TextureWrapMode Convert(this AddressMode mode)
...
+                    return NvBlendEquationAdvanced.HslHueNv;
+                case AdvancedBlendOp.HslSaturation:
+                    return NvBlendEquationAdvanced.HslSaturationNv;
+                case AdvancedBlendOp.HslColor:
+                    return NvBlendEquationAdvanced.HslColorNv;
+                case AdvancedBlendOp.HslLuminosity:
+                    return NvBlendEquationAdvanced.HslLuminosityNv;
+            }
+
+            Logger.Debug?.Print(LogClass.Gpu, $"Invalid {nameof(AdvancedBlendOp)} enum value: {op}.");
```
marysaka, line 135: If that's really not expected, might be worth making this an error or warning instead?

Also it seems that the Vulkan backend use an helper function for that, could be nice to match both it in the future.

Ryujinx.Graphics.OpenGL/EnumConversion.cs
```diff
@@ -34,6 +34,126 @@ public static TextureWrapMode Convert(this AddressMode mode)
...
+            {
+                case AdvancedBlendOverlap.Uncorrelated:
+                    return All.UncorrelatedNv;
+                case AdvancedBlendOverlap.Disjoint:
+                    return All.DisjointNv;
+                case AdvancedBlendOverlap.Conjoint:
+                    return All.ConjointNv;
+            }
+
+            Logger.Debug?.Print(LogClass.Gpu, $"Invalid {nameof(AdvancedBlendOverlap)} enum value: {overlap}.");
```
marysaka, line 152: Same comment as before.

marysaka (approved):
lgtm,

I think the assembler code and  the  the large functions class should stay here.

About if the extension isn't supported, I guess it's fine to keep it as this but it would be nice to have a warning to indicate that the extension is missing. (maybe something more generic that print all extensions not present in the logs?)

Ryujinx.Graphics.OpenGL/EnumConversion.cs
gdkchan: That's what almost all the other conversion methods does on this class, so I just did the same thing here. Some of those values come directly from GPU registers (not the case for advanced blend) so they could spam the log if the game is doing something invalid, that's why they are "Debug" right now.

For advanced blend, those values being invalid should be impossible as it matches the advanced blend operation from the microcode. If the microcode match fails it just does nothing right now (maybe it should warn? or debug log, not sure).

Ryujinx.Graphics.Gpu/Engine/Threed/Blender/UcodeAssembler.cs
```diff
@@ -0,0 +1,306 @@
+
```
riperiperi, line 1: Newline at start of file?

riperiperi (approved):
Looking good. I wonder if AMD's windows driver will follow suit if they do end up implementing fragment shader interlock on linux, then we could do the fallback.

Ryujinx.Graphics.OpenGL/EnumConversion.cs
marysaka: I see.

I think it's fine without the log for microcode too anyway.

PR #2804, opened 2021-11-02
Nickname! - Init Amiibos with Profile's name!

Mou-Ikkai:
# No Name? Nickname!
*Totally not a joke to an earlier implementation*

Instead of the user being a blank "No Name", I decided to attempt to change a Todo to be actually functional now that we have UserProfiles. Now when amiibos request to have their owner's name, instead of "No Name" being used, it uses the name of the current profile!

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
+++ b/Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
@@ -72,8 +72,7 @@ namespace Ryujinx.HLE.HOS.Services.Nfc.Nfp
 
             charInfo.SetFromStoreData(StoreData.BuildDefault(utilityImpl, 0));
 
-            // TODO: Maybe change the "no name" by the player name when user profile will be implemented.
-            // charInfo.Nickname = Nickname.FromString("Nickname");
+            charInfo.Nickname = Nickname.FromString("Nickname");
 
             RegisterInfo registerInfo = new RegisterInfo()
             {
--- a/Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
+++ b/Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
@@ -85,8 +85,6 @@ namespace Ryujinx.HLE.HOS.Services.Nfc.Nfp
                 Reserved2       = new Array58<byte>()
             };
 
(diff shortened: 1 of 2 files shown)


Mou-Ikkai:
Oh, how fun. It compiles on my main PC, but soon breaks on here after my issues with RUNNING it on my PC.

Mou-Ikkai:
- Halted until later -

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
```diff
@@ -72,8 +72,7 @@ public static RegisterInfo GetRegisterInfo(string amiiboId)
 
             charInfo.SetFromStoreData(StoreData.BuildDefault(utilityImpl, 0));
 
-            // TODO: Maybe change the "no name" by the player name when user profile will be implemented.
-            // charInfo.Nickname = Nickname.FromString("Nickname");
+            charInfo.Nickname = Nickname.FromString("Nickname");
```
EliEron, line 75: I'm not familiar with the Nickname class, but wouldn't this just set the nickname to the literal string _Nickname_? 

Unless I'm missing something `Nickname.FromString(nickname);`  is probably the intended code.

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
```diff
@@ -85,7 +84,7 @@ public static RegisterInfo GetRegisterInfo(string amiiboId)
                 Reserved1       = new Array64<byte>(),
                 Reserved2       = new Array58<byte>()
             };
-
+
```
gdkchan, line 87:
```suggestion

```

gdkchan (approved):
Seems reasonable.

riperiperi (approved):
LGTM.

What is the difference between charInfo and registerInfo name? Is it the name given to the amiibo and the name given to the registering user?

Mou-Ikkai:
> What is the difference between charInfo and registerInfo name? Is it the name given to the amiibo and the name given to the registering user?

registerInfo is all of the amiibo's own data. Like how you can see that it's name is always "Ryujinx" since we don't have the Amiibo Editor cabinet yet. While charInfo is the owner's data. (I believe it just means character info, since amiibos also store the Mii's data onto their bin files, which lets you transfer your Mii to another person's switch.)

PR #2807, opened 2021-11-03
Removed not accurate stub log

LuisMayo:
Hi
A really tiny thing I saw while fiddling around on the codebase

As far as I'm aware. The `GetAddrInfoRequestImpl` method is indeed working.

Unless the stub logging is there because of the method ignoring things, which I don't think it's the case.

Sorry if this is not useful and thanks for maintaining this great project.
Thanks.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
@@ -410,8 +410,6 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres
 
             ulong pidPlaceHolder = context.RequestData.ReadUInt64();
 
-            Logger.Stub?.PrintStub(LogClass.ServiceSfdnsres, new { enableNsdResolve, cancelHandle, pidPlaceHolder, host, service });
-
             IPHostEntry hostEntry = null;
 
             NetDbError netDbErrorCode = NetDbError.Success;


jduncanator:
In this case, I think it would be appropriate to remove the stub log, as the functionality does look fairly complete. One thing to keep in mind for the future though, is that a Stub log doesn't necessarily indicate that a function is not implemented at all, but it can also mean that the function has been stubbed with something that isn't accurate to the original Switch OS implementation. In those instances, even if the function looked relatively complete and "working", it may still be prudent to leave the Stub log in there to indicate that function needs revisiting in the future.

LuisMayo:
Ok yeah.

I thought that may the case, that functions implemented with the minimum logic but not being accurate enough may have an "stub" warning just as you said. But I opened the PR anyway so I could be educated.

Merging or not it's at your own discretion now then

Thog:
Not quite sure if we need this stub log (cc @AcK77)

imo we can remove it

Thog (approved):

LuisMayo:
After doing this I was doing some job trying to get an [alternative server for SM35](https://github.com/kinnay/SMB35) working

I have had some issues with the DNS resolver, with the game outputting a 2306-0520 error. Which may mean that this function is indeed not be working properly after all?

Maybe the stub was warranted

AcK77:
> Not quite sure if we need this stub log (cc @AcK77)
> 
> imo we can remove it and make an issue for it

Maybe informations could be useful to be printed in log, if not, I’m fine with remove it ;)

gdkchan:
Superseded by #2943 (also the function implementation was fixed on #2936, but it is still not complete).

PR #2808, opened 2021-11-03
Clamp number of mipmap levels to avoid API errors due to invalid textures

gdkchan:
This clamps the amount of levels to avoid OpenGL errors on textures with an invalid number of mipmap levels. I'm not aware of any commercial game which such error, but there are some mods that do use an invalid mipmap level count for some reason, and that causes the textures to be solid black as the texture initialization will fail. The PR fixes this issue.

While technically the mod is incorrect, I think it's fine to do this on our side as the actual GPU will not attempt to validate the levels count, so the texture would work there.

### DIFF
--- a/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
+++ b/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
@@ -1,5 +1,6 @@
 using Ryujinx.Common;
 using System;
+using System.Numerics;
 
 namespace Ryujinx.Graphics.GAL
 {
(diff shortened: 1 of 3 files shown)


Vaskiemaia:
This PR solves the issue with some Mario Kart 8 Deluxe mods displaying black textures.

Master: https://i.imgur.com/MyFQdE5.png

PR: https://i.imgur.com/49Fil5h.png

As it is i will have to do further tests to see if it solves the issue with all mods, but it definitely solved on all the ones i tested so far!

riperiperi (approved):
Nice catch, would be better if mods didn't do this though. 🥴

Thog (approved):

PR #2817, opened 2021-11-05
Support shader gl_Color, gl_SecondaryColor and gl_TexCoord built-ins

gdkchan:
Implements those built-ins, that was mostly used for fixed-function functionality before fragment shaders were a thing. One could set `gl_Color` and `gl_SecondaryColor` from a vertex shader to set the parameters used for the fixed-function lighting. Nowadays it's pretty much useless since you can just write a fragment shader and do whatever you want there, but some OpenGL games make use of it.

This is similar to #986, but this implementation does not make use of the compatibility profile, instead it passes the values on regular user attributes that are otherwise not used on the next stage. That means it should work on OpenGL without needing a compatibility profile, and it should also work on Vulkan.

This fixes black screen on a few OpenGL games, such as:
rRootate Reloaded:
image
This one only uses `gl_Color` as the game doesn't use textures at all.
Dragon Quest III (and possibly the older ones too?):
image
image
image
This one uses both `gl_Color` and `gl_TexCoord`.


Thog (approved):

riperiperi:
image
This also fixes Xash3D. Looked like it had the FrontColor/BackColor, but it is fine on the latest commit.

riperiperi (approved):
Nice to see this working.

PR #2822, opened 2021-11-07
Fix InvocationInfo on geometry shader and bindless default integer const

### LINKED ISSUE #2821: (Regression) Worm Jazz - 01009CD012CC0000 missing graphics
## Bug Report
the game is missing graphics
the last good version is 1.0.7077
the bad version is when it first started is 1.0.7078
Initial tessellation shader support (#2534)

just load the game

[ Include a detailed step by step process for recreating your issue. ]

### Log file

### Environment?

- Ryujinx version: 1.0.7098

- Game version: 1.4.1
[ Replace X's with the game version at time of crash. ]
- System Specs:
  - OS: *( Windows 10)*
  - CPU: *(e.g. i7-6700)*
  - GPU: *(AMD Raedon RX580)*
  - RAM: *(e.g. 16GB)*
- Applied Mods : [  No ]

### Additional context?

[ Any other information relevant to your issue. ]

gdkchan:
This fixes a regression introduced on #2534 where shaders would be generated trying to access `gl_PatchPrimitivesIn` on geometry shaders, because the invocation info system register is accessed on those shaders, however this built-in can only be accessed on tessellation evaluation shaders according to the spec.

Also contains a fix for another issue I found while testing this, bindless images with unknown handle source uses a 0 constant right now. However, it was using a float constant for images with integer formats, which is incorrect. Change it to use the proper integer type, this fixes some shader compilation errors.

The first issue apparently doesn't affect NVIDIA, as apparently their compiler doesn't complain on OpenGL and it "just works".
On AMD and Intel, this fixes a regression that broke geometry shaders, they caused some UE4 games to only show a black screen there because those games writes a 3D LUT texture from a draw that uses geometry shaders (to specify the slice). This is also most likely what caused the regression on Vulkan.

Fixes #2821.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -40,7 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2768;
+        private const ulong ShaderCodeGenVersion = 2822;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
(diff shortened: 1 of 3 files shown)


benderscruffy:
the unreal engine games that are missing graphics are showing now
ryujinx_capture_2021-11-07_18-31-19
ryujinx_capture_2021-11-07_18-34-38
ryujinx_capture_2021-11-07_18-36-17
ryujinx_capture_2021-11-07_18-39-19
ryujinx_capture_2021-11-07_18-50-22
ryujinx_capture_2021-11-07_18-50-56

Thog (approved):

riperiperi (approved):
LGTM

Linear524:
Trailing shadow bug is still here... Is it possible to fix it 
ryujinx_capture_2021-11-08_19-24-38
?

Issue #2744, opened 2021-10-16
Hyrule Warriors Definitive Edition - In-stage missions do not progress, various softlocks


vickyorlo:
## Bug Report

</br>

### What's the issue you encountered?
Once in a while, seemingly at random, once a mission (in-level task) is completed, the following mission does not trigger, causing a softlock. On occasion the softlock will resolve itself after a couple minutes of nothing happening, but it feels rare. I have however found one softlock that is fully reproducible on the "Powers Collide" story mode mission, where one of the engineers that are supposed to run towards a certain point of the map simply stands there waiting for God knows what, even though there are no enemies on his way.

![softlock](https://user-images.githubusercontent.com/18310449/137594376-4024f8e7-34bb-4d2d-84cf-e739b65aca68.png)

The engineer in question is on the right side of the map, the blue bullseye icon covered by an arrow. He should be with his other two blue bullseye icon friends. Also note to the left the red dot on the map, which is supposed to represent a group of enemies. It is a lie, however, there are no enemies in that whole section of the map anymore.

I have confirmed that this issue does not appear on real hardware, and I have transplanted a suspend save to Yuzu - the issue does not happen there either. I have also cleared my PPTC as people used to report issues with the game related to it, but it did not help.

### How can the issue be reproduced?

1. Download the included save file
2. When in the main menu, pick Resume to load the suspend save
3. The Engineers should appear momentarily. The save puts you in the corridor the problematic engineer goes through, and it needs to be cleared of enemies. 
4. Even if the enemies are cleared, the eastern Engineer will not move past about halfway of the hallway, as on the screenshot.
5. One of the other Engineers might die and fail the mission, but this is unrelated. This can be easily enough prevented by defeating the other enemies on their paths.

### Log file

[Ryujinx_1.0.7074_2021-10-16_17-58-58.log](https://github.com/Ryujinx/Ryujinx/files/7357889/Ryujinx_1.0.7074_2021-10-16_17-58-58.log)

### Environment?

- Ryujinx version: 1.0.7074</br>
- Game version: 1.0.1</br>
- System Specs:
  - OS: Arch Linux
  - CPU: Ryzen 5600x
  - GPU: GTX1080
  - RAM: 32GBB
- Applied Mods : No

### Additional context?

Additional info about your environment:</br>
The save file with a suspend save:
[hw save file.zip](https://github.com/Ryujinx/Ryujinx/files/7357823/hw.save.file.zip)

[...]

PR #2828, opened 2021-11-09
kernel: Fix sleep timing accuracy

Thog:
This PR corrects some mistake while comparing my reversing of kernel 13.x with our own.

``WaitAndCheckScheduledObjects`` timing accuracy was also improved.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -8,6 +8,8 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 {
     class KTimeManager : IDisposable
     {
+        public static readonly long DefaultTimeIncrementNanoSeconds = ConvertGuestTicksToNanoSeconds(2);
+
         private class WaitingObject
         {
             public IKFutureSchedulerObject Object { get; }
@@ -41,7 +43,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 
         public void ScheduleFutureInvocation(IKFutureSchedulerObject schedulerObj, long timeout)
         {
-            long timePoint = PerformanceCounter.ElapsedMilliseconds + ConvertNanosecondsToMilliseconds(timeout);
+            long timePoint = PerformanceCounter.ElapsedTicks + ConvertNanosecondsToHostTicks(timeout);
 
             lock (_context.CriticalSection.Lock)
             {
@@ -74,14 +76,19 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 
                     if (next != null)
                     {
-                        long timePoint = PerformanceCounter.ElapsedMilliseconds;
+                        long timePoint = PerformanceCounter.ElapsedTicks;
 
                         if (next.TimePoint > timePoint)
                         {
-                            _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
+                            int ms = (int)((next.TimePoint - timePoint) / PerformanceCounter.TicksPerMillisecond);
+
+                            if (ms > 0)
+                            {
+                                _waitEvent.WaitOne(ms);
+                            }
                         }
 
-                        bool timeUp = PerformanceCounter.ElapsedMilliseconds >= next.TimePoint;
+                        bool timeUp = PerformanceCounter.ElapsedTicks >= next.TimePoint;
 
                         if (timeUp)
                         {
@@ -117,6 +124,22 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
         public static long ConvertMillisecondsToNanoseconds(long time)
         {
             return time * 1000000;
+        }
+
+        public static long ConvertNanosecondsToHostTicks(long ns)
+        {
+            long nsDiv = ns / 1000000000;
+            long nsMod = ns % 1000000000;
+            long tickDiv = PerformanceCounter.TicksPerSecond / 1000000000;
+            long tickMod = PerformanceCounter.TicksPerSecond % 1000000000;
+
+            long baseTicks = (nsMod * tickMod + PerformanceCounter.TicksPerSecond - 1) / 1000000000;
+            return (nsDiv * tickDiv) * 1000000000 + nsDiv * tickMod + nsMod * tickDiv + baseTicks;
+        }
+
+        public static long ConvertGuestTicksToNanoSeconds(long ticks)
+        {
+            return (long)Math.Ceiling(ticks * (1000000000.0 / 19200000.0));
         }
 
         public static long ConvertHostTicksToTicks(long time)
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -26,6 +26,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
         private readonly List<WaitingObject> _waitingObjects;
         private AutoResetEvent _waitEvent;
         private bool _keepRunning;
+        private long _enforceWakeupFromSpinWait;
 
         public KTimeManager(KernelContext context)
         {
@@ -48,6 +49,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
             lock (_context.CriticalSection.Lock)
             {
                 _waitingObjects.Add(new WaitingObject(schedulerObj, timePoint));
+                Interlocked.Exchange(ref _enforceWakeupFromSpinWait, timeout < 1000000 ? 1 : 0);
             }
 
             _waitEvent.Set();
@@ -63,14 +65,17 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 
         private void WaitAndCheckScheduledObjects()
         {
+            SpinWait spinWait = new SpinWait();
+            WaitingObject next;
+
             using (_waitEvent = new AutoResetEvent(false))
             {
                 while (_keepRunning)
                 {
-                    WaitingObject next;
-
                     lock (_context.CriticalSection.Lock)
                     {
+                        Interlocked.Exchange(ref _enforceWakeupFromSpinWait, 0);
+
                         next = _waitingObjects.OrderBy(x => x.TimePoint).FirstOrDefault();
                     }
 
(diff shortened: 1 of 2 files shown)


gdkchan:
Could potentially unblock #1798.

gdkchan (approved):
lgtm, thanks.

riperiperi:
Greatly improves loading speeds in BOTW 1.0.0. Before they were very long and inconsistent, now they are short (3-5x faster) and consistent. This will likely also help ingame streaming, and I noticed a small performance boost ingame. (36-38 -> 40-42)

I took some rough measurements from booting the game to loading a save. "To Load Game" is the time until the save game starts loading (spamming a), "To 100% progress" is the time for the BOTW logo to fill entirely on the loading screen, and "To Ingame" is the time from that to the game beginning to fade in.

I noticed that with a hack to make the game run at 30 instead of 20, "to ingame" shortens to about 5.5 seconds. 

## PR

to load game: 15.42s
to 100% progress: 21.85s
to ingame: 8.74s

to load game: 14.78s
to 100% progress: 22.38s
to ingame: 8.89s

to load game: 14.55s
to 100% progress: 22.18s
to ingame: 8.89s

## Master

to load game: 15.00s
to 100% progress: 1:01.00s
to ingame: 9.58s

to load game: 15.34s
to 100% progress: 1:36.42s
to ingame: 15.27s

to load game: 14.37s
to 100% progress: 59.75s
to ingame: 7.96s

Mou-Ikkai:
Hatsune Miku: Project DIVA Megamix seems to no longer crash for the first time on boot, forcing you to press stop emulation and rerun! 
Master: 10/10 times it froze on first run, but ran properly after "Stop Emulation" and relaunching the game.
PR: 0/10 times it froze on first run. This has fixed the first-time startup crash!

jojorne:
Tested. Fixed boot on all my games.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -74,14 +76,19 @@ private void WaitAndCheckScheduledObjects()
...
 
                         if (next.TimePoint > timePoint)
                         {
-                            _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
+                            int ms = (int)((next.TimePoint - timePoint) / PerformanceCounter.TicksPerMillisecond);
+
+                            if (ms > 0)
+                            {
+                                _waitEvent.WaitOne(ms);
+                            }
```
riperiperi, line 88: Maybe should SpinWait a little when ms is 0, so that the critical section lock isn't hammered so hard? Not sure how much is responsive enough.

Thog: Not sure...

vickyorlo:
Incidentally, this fixes #2744, and all the other issues in Hyrule Warriors like the timer running slow, various mission progression softlocks/lagginess, and enemy spawning speed. However it feels like the game's performance has dropped - I'm getting framerates below 50 in busy scenes (which, considering it's a musou game, is almost always) when I used to see 60fps with drops to maybe 55 most of the time. It's hard to say, since with this fix the amount of enemies on screen has increased.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -61,27 +65,49 @@ public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
 
         private void WaitAndCheckScheduledObjects()
         {
+            SpinWait spinWait = new SpinWait();
```
riperiperi, line 68: Since this is defined here, it might make sense to reset the spin wait when the spin loop exists, so that it won't yield on its first spin when doing another wait later on.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -8,6 +8,8 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
 {
     class KTimeManager : IDisposable
     {
+        public static readonly long DefaultTimeIncrementNanoSeconds = ConvertGuestTicksToNanoSeconds(2);
```
riperiperi, line 11: `Nanoseconds` rather than `NanoSeconds`?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -41,11 +44,12 @@ public KTimeManager(KernelContext context)
 
         public void ScheduleFutureInvocation(IKFutureSchedulerObject schedulerObj, long timeout)
         {
-            long timePoint = PerformanceCounter.ElapsedMilliseconds + ConvertNanosecondsToMilliseconds(timeout);
+            long timePoint = PerformanceCounter.ElapsedTicks + ConvertNanosecondsToHostTicks(timeout);
 
             lock (_context.CriticalSection.Lock)
             {
                 _waitingObjects.Add(new WaitingObject(schedulerObj, timePoint));
+                Interlocked.Exchange(ref _enforceWakeupFromSpinWait, timeout < 1000000 ? 1 : 0);
```
riperiperi, line 52: If something is scheduled with >=1ms timeout, then this will actually _clear the spin wait wakeup, even if there might be scheduled invocation <1 still present in `_waitingObjects`. You could just use an if statement to enforce this rather than the ternary statement.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -61,27 +65,49 @@ public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
...
 
                     if (next != null)
                     {
-                        long timePoint = PerformanceCounter.ElapsedMilliseconds;
+                        long timePoint = PerformanceCounter.ElapsedTicks;
 
                         if (next.TimePoint > timePoint)
                         {
-                            _waitEvent.WaitOne((int)(next.TimePoint - timePoint));
+                            int ms = (int)((next.TimePoint - timePoint) / PerformanceCounter.TicksPerMillisecond);
```
riperiperi, line 88: I wonder if 1ms waits will get to this point with 0.99ms remaining, round down to 0 and have to do the spin (maybe similar concerns for longer waits, where it could spin the slightly less than 1ms remainder). I guess it's fine for now, just something to think about. (could a small bias to this calculation help?)

FICTURE7: Perhaps keep the unit in ticks and convert to ms only in the `WaitOne` call?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
FICTURE7: Also is it desirable to do that? I think it only makes sense to interrupt the spin wait if the new time point is smaller than the time point on which the spin wait is waiting on, no?

riperiperi: That is true, but that might be hard to atomically check. The cost of this wakeup is just another run of the `WaitAndCheckScheduledObjects` loop, before it goes back into the spin wait til time is up.

d3xMachina:
Pokemon Brillant Diamond does not start with this PR. It stay blocked at a black screen even after waiting 5 minutes.
Also if I try to do "Stop emulation" and close Ryujinx while there is the black screen, it does not close Ryujinx unless I close the command line window.

On the latest Ryujinx version it start after like 1 min and the following messages are displayed before loading : 
> 00:00:21.405 |E| HLE.OsThread.9 Gpu WaitOnSyncpoint: Wait on syncpoint 1 for threshold 2 took more than 1000ms, resuming execution...
> 00:01:01.362 |E| HLE.OsThread.9 Gpu WaitOnSyncpoint: Wait on syncpoint 1 for threshold 6 took more than 1000ms, resuming execution...

Here are my logs (not much to see) : 
[Ryujinx_1.0.0+10c19b1.log](https://github.com/Ryujinx/Ryujinx/files/7572775/Ryujinx_1.0.0%2B10c19b1_2021-11-19_21-14-05.-.Copie.log)

The logs after 5 mins are after clicking stop emulation.

d3xMachina:
I made a fix for the overflow on the timer.
Also I changed the wait so that the last millisecond we wait is done by spinning even when the wait is higher than 1ms for more precision.
No issue so far.

Here is the patch (rename the extension to .patch) :
[timer_patch.txt](https://github.com/Ryujinx/Ryujinx/files/7573759/timer_patch.txt)

Thog:
> I made a fix for the overflow on the timer. Also I changed the wait so that the last millisecond we wait is done by spinning even when the wait is higher than 1ms for more precision. ScheduleFutureInvocation now only interrupt the wait if necessary. No issue so far.
> 
> Here is the patch (rename the extension to .patch) : [timer_patch.txt](https://github.com/Ryujinx/Ryujinx/files/7574436/timer_patch.txt)
> 
> Edit : I added a _waitEvent.Reset() in WaitAndScheduledObject in lock to avoid the case where _waitEvent.Set() is triggered again after _waitEvent.WaitOne() has been interrupted which would cause the next wait to be instantly canceled if higher than 1 ms.
> 
> I didn't completly explain why my patch fix the issue. The bug there was with the code is that _enforceWakeupFromSpinWait was only triggered if another timeout was of less than 1 ms. Instead it should have been canceled when a new scheduled invocation started sooner. With the overflow it would go spinning and only _enforceWakeupFromSpinWait was able to make it go out since the scheduled time was far in the future.

Fixed the issue on https://github.com/Ryujinx/Ryujinx/pull/2828/commits/1f175bf452075b46e6ef7068585afaae7db54228.
I fixed it by removing the unnecessary cast and by capping the upper timeout limit to 0x7FFFFFFF like it used to be. (see ConvertNanosecondsToMilliseconds behaviour)

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -119,6 +151,22 @@ public static long ConvertMillisecondsToNanoseconds(long time)
...
+            long tickDiv = PerformanceCounter.TicksPerSecond / 1000000000;
+            long tickMod = PerformanceCounter.TicksPerSecond % 1000000000;
+
+            long baseTicks = (nsMod * tickMod + PerformanceCounter.TicksPerSecond - 1) / 1000000000;
+            return (nsDiv * tickDiv) * 1000000000 + nsDiv * tickMod + nsMod * tickDiv + baseTicks;
+        }
+
+        public static long ConvertGuestTicksToNanoseconds(long ticks)
+        {
+            return (long)Math.Ceiling(ticks * (1000000000.0 / 19200000.0));
```
lostromb, line 167: Where does this 19200000 come from? Is it based on `Stopwatch.Frequency`?

Thog: This is the value of ``CNTFRQ_EL0`` on the Switch

AcK77 (approved):

PR #2831, opened 2021-11-10
Fix shader integer from/to double conversion

gdkchan:
This fixes the `Invalid reinterpret cast from "F32" to "F64"` exception that would happen on any shader using I2F or F2I instructions to convert between double and integer types. That was happening because those float conversion instructions have a 32-bit float type, while it was trying to store them as a double, which made it try to reinterpret the value. It was fixed by adding new instructions for explicit conversion from and to double precision.


MadDragon1846:
This lets World War Z load into the menus instead of crashing before the title screen. (Unfortunately though the game runs really bad in the menus, then throws up an error when you try to start a match).

Thog (approved):
nice catch

riperiperi (approved):

Issue #2509, opened 2021-07-27
Anistropic filtering is bugged


GamerzHell9137:
### What's the issue you encountered?

Anistropic filtering is not applied when changed mid game session.

### How can the issue be reproduced?
Just try to change it mid emulation.

### Environment?

- Ryujinx version: 1.0.6976

emmauss:
Is this on master or the new ui pr?

GamerzHell9137:
> 
> 
> Is this on master or the new ui pr?

Seems like master too so its not an Avalonia issue atm.

Shahil-Ayato:
Fixed by #2832

PR #2832, opened 2021-11-10
Limit Custom Anisotropic Filtering to mipmapped textures with many levels

riperiperi:
There's a major flaw with the anisotropic filtering setting that causes @GamerzHell9137 to report graphical bugs that otherwise wouldn't be there, because he just won't set it to Auto. This should fix those issues, hopefully.

These bugs are generally because anisotropic filtering is enabled on something that it shouldn't be, such as a post process filter or some data texture. This PR maintains two host samplers when custom AF is enabled, and only uses the forced AF one when the texture is 2d and fully mipmapped (goes down to 1x1). This is because game textures are the ideal target for this filtering, and they are typically fully mipmapped, unlike things like screen render targets which usually have 1 or just a few levels.

This also only enables AF on mipmapped samplers where the filtering is bilinear or trilinear. This should be self explanatory.

**This PR also allows the changing of Anisotropic Filtering at runtime, and you can immediately see the changes.** All samplers are flushed from the cache if the setting changes, causing them to be recreated with the new custom AF value. This brings it in line with our resolution scale. :relieved:

Before (16x Handheld):
image

1x Handheld:
image

16x Handheld:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Sampler.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Sampler.cs
@@ -9,15 +9,17 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class Sampler : IDisposable
     {
+        private const int MinLevelsForAnisotropic = 5;
+
         /// <summary>
         /// Host sampler object.
         /// </summary>
-        private ISampler _hostSampler { get; }
+        private readonly ISampler _hostSampler;
 
         /// <summary>
         /// Host sampler object, with anisotropy forced.
         /// </summary>
-        private ISampler _anisoSampler { get; }
+        private readonly ISampler _anisoSampler;
 
         /// <summary>
         /// Creates a new instance of the cached sampler.
@@ -99,7 +101,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
         /// <summary>
         /// Determine if the given texture can have anisotropic filtering forced.
-        /// Filtered textures that we might want to force anisotropy on should have a full set of mip levels.
+        /// Filtered textures that we might want to force anisotropy on should have a lot of mip levels.
         /// </summary>
         /// <param name="texture">The texture</param>
         /// <returns>True if anisotropic filtering can be forced, false otherwise</returns>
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Gpu/Image/Sampler.cs
```diff
@@ -11,7 +12,12 @@ class Sampler : IDisposable
         /// <summary>
         /// Host sampler object.
         /// </summary>
-        public ISampler HostSampler { get; }
+        private ISampler _hostSampler { get; }
```
gdkchan, line 15: Can you make those readonly fields now that they are private?

Ryujinx.Graphics.Gpu/Image/Sampler.cs
```diff
@@ -61,15 +64,65 @@ public Sampler(GpuContext context, SamplerDescriptor descriptor)
...
+
+        /// <summary>
+        /// Determine if the given texture can have anisotropic filtering forced.
+        /// Filtered textures that we might want to force anisotropy on should have a full set of mip levels.
+        /// </summary>
+        /// <param name="texture">The texture</param>
+        /// <returns>True if anisotropic filtering can be forced, false otherwise</returns>
+        private static bool AllowForceAnisotropy(Texture texture)
+        {
+            if (!(texture.Info.Target == Target.Texture2D || texture.Info.Target == Target.Texture2DArray))
```
gdkchan, line 108: `Texture` has a `Target` property, any reason to not use that?

gdkchan:
Closes https://github.com/Ryujinx/Ryujinx/issues/2509?

GamerzHell9137:
x16 Anistropic Filtering gang member reporting in!

Games that introduced issues are not introducing issues anymore!
Games tested:
The Ninja Saviors: Return of the Warriors
Disgaea 1 Complete
Little Nightmares II

Some comparisons:
ryujinx_capture_2021-11-10_15-47-46
ryujinx_capture_2021-11-10_19-37-02

image

image
image
image

One thing that im not sure about is Dungeon Defenders. The issue with the lighting might be related to something else
**Auto PR:**

ryujinx_capture_2021-11-10_19-51-55
**x16 PR:**

ryujinx_capture_2021-11-10_19-52-13

Other people say that game's like FE Three Houses don't do x16 AF anymore when enabled

Master
image

PR
image

Mario Kart 8 doens't have that issue
Auto PR

ryujinx_capture_2021-11-10_20-03-19
x16 PR:
ryujinx_capture_2021-11-10_20-03-31

gdkchan (approved):
Nice, glad to see this one fixed!

GamerzHell9137:
Retested all of my games. No regression with the newest changes and Fire Emblem got fixed.

Thog (approved):

PR #2843, opened 2021-11-14
Move texture anisotropy check to SetInfo

riperiperi:
Rather than calculating this for every sampler, this PR calculates if a texture can force anisotropy when its info is set, and exposes the value via a public boolean.

This should help texture/sampler heavy games when anisotropic filtering is not Auto, like UE4 ones (or so i hear?). There is another cost where samplers are created twice when anisotropic filtering is enabled, but I'm not sure how relevant this one is.

Could use some performance testing. (SMTV vs master before the AF stuff?)


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

Thog (approved):

PR #2860, opened 2021-11-19
Allow setting texture data from 1x to fix some textures resetting randomly

riperiperi:
Expected targets:

- Deltarune Chapter 1+2
- Crash Team Racing
- Those new pokemon games idk (reset scale after battle/menu?)

This won’t help all games- for example SMTV issues. This might also be an aggressive solution, so it would be good to test if this has negative effects in other games.

I won’t be able to complete this for a bit so testing would help me make a decision when I can.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
@@ -349,6 +349,11 @@ namespace Ryujinx.Graphics.Gpu.Engine.Twod
                 return;
             }
 
+            if (srcTexture.Info.Samples > 1 || dstTexture.Info.Samples > 1)
+            {
+                srcTexture.PropagateScale(dstTexture);
+            }
+
             float scale = srcTexture.ScaleFactor;
             float dstScale = dstTexture.ScaleFactor;
 
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -122,6 +122,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         private Target   _arrayViewTarget;
 
         private ITexture _flushHostTexture;
+        private ITexture _setHostTexture;
 
         private Texture _viewStorage;
 
@@ -554,9 +555,10 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// Copy the host texture to a scaled one. If a texture is not provided, create it with the given scale.
         /// </summary>
         /// <param name="scale">Scale factor</param>
+        /// <param name="copy">True if the data should be copied to the texture, false otherwise</param>
         /// <param name="storage">Texture to use instead of creating one</param>
         /// <returns>A host texture containing a scaled version of this texture</returns>
-        private ITexture GetScaledHostTexture(float scale, ITexture storage = null)
+        private ITexture GetScaledHostTexture(float scale, bool copy, ITexture storage = null)
         {
             if (storage == null)
             {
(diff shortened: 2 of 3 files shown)


saldabain:
I've noticed that the screen is totally black while a pokemon is evolving in bdsp. Not sure if it is specific to this PR, but this is the one I use.

Nothing noteworthy appears in the logs.

Leischii:
I can confirm saldabain's comment. With this Pr, evolution and egg hatching animations are completly black. Not present with master or ldn.

gdkchan:
Does that happen while in native resolution, scaled or both?

saldabain:
> Does that happen while in native resolution, scaled or both?

I've only tried it out in a scaled resolution. (Standard, not custom)

Niwu34:
Can easily be reproduced in the Pokémon Starter selection as well
It will show something if you start the game on Native res,
it will show a black screen if you choose to go into the game with 4x res

riperiperi:
I've pushed a fix for the screen going black. It seems that several scenes use MSAA targets (which were not allowed to scale), and opengl was having trouble blitting from an MSAA target (1x) to a screen bound target (2x). Now, MSAA targets are allowed to scale, and 2d copies force scale to propagate so they both match.

riperiperi:
Note: right now, msaa textures cannot retain their data through a scale change.

GamerzHell9137:
With the new changes RF5 is upscaling properly now!

PR 
image

Master
image

Cloratus:
How I put that fix in my game? it is a offline build?

poudink:
the builds are available in the first reply of this thread (the one from the github-actions bot). you can also build from source yourself.

0rb1t5l:
Sometimes Caligula Effect 2 renders graphics incorrectly like this after a scene change, using 2x scaling and auto anisotropic filtering, never had this happen on master. Using ryujinx-Release-1.0.0+e99bd25-win_x64 on 1060 GPU.
ryujinx_capture_2021-11-28_12-43-36

ActualMandM:
Sonic Colors: Ultimate now displays 3D when upscaling with the PR, even though it still shows it at 1x scale

PR:
2021-12-03_18-30-01_Ryujinx

Master:
2021-12-03_18-29-08_Ryujinx

Niwu34:
Fixes the reset of the res scaler after each battle in My Hero One's Justice 2

0rb1t5l:
This fixes reset scale for SMT V as well.

ActualMandM:
Fixes Cruis'n Blast upscaling

Master:
master

PR:
pr

GamerzHell9137:
Monopoly Madness is properly upscaling too now 
image

rasmysamy:
I have built and tested the latest commit, I can confirm that the evolutions, etc... work well on BDSP for me. Thank you very much for this fix, it's a huge improvement and I hope it can be merged soon!

ActualMandM:
Fixes *The Stanley Parable Ultra Deluxe* upscaling

Master:
Master

PR:
Pull Request

CrusadingNinja:
Download links are expired

riperiperi:
I did find a case that triggers it constantly - The NSO N64 emulator. However... I don't think this should be happening, even if it is taking a slower path by upscaling the manually set data.

image

riperiperi:
I've fixed the bug above and also added some code to automatically blacklist after too many writes. My theory for this happening in Unity games is that it triggers when a render target is prepared for new usage, perhaps by the guest clearing its data. Deltarune might be similar, as the menu select for that game essentially switches between different games. Hopefully the current tuning should be good enough to keep these games scaling.

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #2861, opened 2021-11-19
Dynamically increase buffer size when resizing

gdkchan:
Some games have a buffer usage pattern where it uses small adjacent buffers with increasing addresses. Right now this causes a new buffer to be created each time, merging the existing overlapping buffers. So that is effectively resizing the existing buffer all the time, page by page which is very inefficient.

This change attempts to address that by dynamically increasing the size based on the current buffer size. Currently it uses a growth rate of 1.5x. So if the existing buffer has 64KB and it attempts to extend it by one page to 68KB, then the new size will be 64KB * 1.5 = 96KB. Further attempts to extend it again by 4KB to 72KB, 76KB, 80KB etc, will be effectively a no-op since the 96KB buffer can already fit all the data.

The downsize of this approach is that it can use a bit more memory (which might not be necessary if the extended region is never used).

This reduces the load time (from black screen to Game Freak logo etc) on Pokémon Brilliant Diamond and Shining Pearl by almost half with PPTC enabled (from 1m34s to 52s here). Also greatly improves title screen animation and overall performance on Super Mario Galaxy, which has a similar issue. Other games might also be improved.

This is very similar to #1607, but with some small differences on the implementation.

I'd recommend testing this on a number of games to make sure they did not break.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
@@ -17,6 +17,8 @@ namespace Ryujinx.Graphics.Gpu.Memory
         private const ulong BufferAlignmentSize = 0x1000;
         private const ulong BufferAlignmentMask = BufferAlignmentSize - 1;
 
+        private const ulong MaxDynamicGrowthSize = 0x100000;
+
         private readonly GpuContext _context;
         private readonly PhysicalMemory _physicalMemory;
 
@@ -170,6 +172,19 @@ namespace Ryujinx.Graphics.Gpu.Memory
 
                 if (_bufferOverlaps[0].Address > address || _bufferOverlaps[0].EndAddress < endAddress)
                 {
+                    if (overlapsCount == 1)
+                    {
+                        // Try to grow the buffer by 1.5x of its current size.
+                        // This improves performance in the cases where the buffer is resized often by small amounts.
+                        ulong existingSize = _bufferOverlaps[0].Size;
+                        ulong growthSize = (existingSize + Math.Min(existingSize >> 1, MaxDynamicGrowthSize)) & ~BufferAlignmentMask;
+
+                        size = Math.Max(size, growthSize);
+                        endAddress = address + size;
+
+                        overlapsCount = _buffers.FindOverlapsNonOverlapping(address, size, ref _bufferOverlaps);
+                    }
+
                     for (int index = 0; index < overlapsCount; index++)
                     {
                         Buffer buffer = _bufferOverlaps[index];
@@ -183,7 +198,9 @@ namespace Ryujinx.Graphics.Gpu.Memory
                         }
                     }
 
-                    Buffer newBuffer = new Buffer(_context, _physicalMemory, address, endAddress - address, _bufferOverlaps.Take(overlapsCount));
+                    ulong newSize = endAddress - address;
+
+                    Buffer newBuffer = new Buffer(_context, _physicalMemory, address, newSize, _bufferOverlaps.Take(overlapsCount));
 
                     lock (_buffers)
                     {
@@ -202,7 +219,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
                         buffer.DisposeData();
                     }
 
-                    newBuffer.SynchronizeMemory(address, endAddress - address);
+                    newBuffer.SynchronizeMemory(address, newSize);
 
                     // Existing buffers were modified, we need to rebind everything.
                     NotifyBuffersModified?.Invoke();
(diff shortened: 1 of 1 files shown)


MutantAura:
Reduces BDSP boot times by around 1/3 for my system (28s->8s) but also noticed a regression in Pokemon Sw/Sh.
On entering Hammerlocke the game will now display a similar GPU syncpoint log and now takes much longer to load into the town. Used to be fairly instant and now there is _significant_ hang.

Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
```diff
@@ -170,6 +172,19 @@ private void CreateBufferAligned(ulong address, ulong size)
 
                 if (_bufferOverlaps[0].Address > address || _bufferOverlaps[0].EndAddress < endAddress)
                 {
+                    if (overlapsCount == 1)
```
riperiperi, line 175: Maybe this should also check that the new range starts _after_ the existing overlap. It doesn't make sense to extend forward when the range is behind the existing overlap.

Generally, maybe this could be more specfic about expansion rules to avoid unnecessarily joining buffers in games that behave well, which is most of them. Maybe it would need to increase end address (not start) multiple times to trigger this? How large is the _new_ region in each buffer expansion in the games that do this? (pokemon, mario galaxy)

Did you find out what buffer type is doing this in Pokemon? I remember that in galaxy it was vertex buffers.

gdkchan: > Generally, maybe this could be more specfic about expansion rules to avoid unnecessarily joining buffers in games that behave well, which is most of them. Maybe it would need to increase end address (not start) multiple times to trigger this?

I tried to make it as general as possible to avoid targeting scenarios on this specific game (to increase the likelyhood of it benefiting another title that does something similar, but slightly different. I don't have any specific title in mind, it's just the direction I was going for). I can make it more specific by only triggering for overlaps at the end (afterall, extending the end if the overlap is at the start makes no sense). I could add a counter and only trigger this if the buffer was resized N times too (and maybe additionally check if it was extended by one page or something, if we really want to target this game as much as possible).

In general I don't like this solution as it is not easily predictable, it might just decide to trigger depending on the usage pattern of the game, and affect memory usage and/or performance as a result, which is why I closed it the last time. I would prefer something like sparse buffers where the whole mapped region is reserved upfront and then committed on demand. But that has other issues, like not being supported by all vendors, and I think creating large sparse buffers (like 1GB+) is slow too.

> How large is the new region in each buffer expansion in the games that do this, and could that be used to detect this expansion? (pokemon, mario galaxy)

On Pokémon it's one page (actually less than that, but it gets rounded to one page). On Mario Galaxy I haven't looked.

> Did you find out what buffer type is doing this in Pokemon? I remember that in galaxy it was vertex buffers.

Constant buffers (aka uniform buffers).

Cloratus:
How make the changes?

Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
gdkchan: I made some changes to restrict the cases where the dynamic expansion is triggered a bit more. Now it only expands if the overlap happens at the end of the existing buffer, and only if the size delta between the new and existing buffers is of at most 2 pages (this is required to handle for example, the non page aligned range crossing a page boundary, and then having a size of 2 pages after being page aligned).

I have verified that this still provides the improvement for both Pokemon Brilliant Diamond and Super Mario Galaxy.
Other ways to further restrict it includes:
- Only dynamically resize if the buffer already had at least N resizes (maybe also with the size constraint?).
- There was something else, but I forgot right now... I will edit if/when I remember.

Might be worth checking if it still has the regression on Sword/Shield.
Marked as ready for review as I don't have any other changes planned, unless something comes up on review or the regression on Sword/Shield is still there.

MutantAura:
Update to my previous comment. Sw/Sh load times are back to normal and the BdSp boot times remain significantly reduced.
No other issues in any other games tested.

Thog (approved):

MutantAura:
Unsure if this was always the case or if a change has made it so but the _first_ boot of BDSP on a new build (I'm actually unsure what causes it. It's independent of PPTC anyway) will still be very slow. The next boots will however be perfectly acceptable and show the changes here. 

Boot 1: 

https://user-images.githubusercontent.com/44103205/155180119-303dd63c-0eca-428f-b383-3bbd287d0230.mp4

Boot 2 (and beyond):

https://user-images.githubusercontent.com/44103205/155180173-13257fdc-5114-45bb-ab5a-45d1ef1e9015.mp4

Is this expected behaviour? Seems odd to me that sometimes I'll boot the game and it will take much longer again.

AcK77 (approved):

Issue #1366, opened 2020-07-06
Bugs in DLC Json loading


mageven:
After the PR for storing DLCs to json was merged, here are some of the bugs I've noticed (list isn't exhaustive)-

1. `DlcNcaList` can be `null` for various reasons.
https://github.com/Ryujinx/Ryujinx/blob/43b78ae157eed5c9436dc19a6d498655891202d8/Ryujinx.HLE/HOS/ApplicationLoader.cs#L264
Pretty sure this causes the following exception (screen taken from discord)
![34fccbc9c46cbdd1](https://user-images.githubusercontent.com/62494521/86566729-c1bb7700-bf87-11ea-84a1-76857d42eeed.png)
Also here-
https://github.com/Ryujinx/Ryujinx/blob/43b78ae157eed5c9436dc19a6d498655891202d8/Ryujinx/Ui/DlcWindow.cs#L87

2. `containerPath` can be invalid for various reasons. Probably the same applies to `AddAocData()` above.
https://github.com/Ryujinx/Ryujinx/blob/43b78ae157eed5c9436dc19a6d498655891202d8/Ryujinx.HLE/FileSystem/Content/ContentManager.cs#L243-L259

3. In Manage DLC window, containers do not have the proper initial state on open. They should at least be on/off if all children are on/off respectively.

4. Currently `dlc.json` stores and uses the nca file when loading enabled DLC. Problem with this approach is DLC can be updated which changes the nca filenames. This would cause issues.

perkel666:
I have this bug. I renamed DLC and folder with DLC and now all games with DLC crash on startup throwing problems about renaming stuff using old names.

PR #2867, opened 2021-11-21
Don't blow up everything if a DLC file is moved or renamed.

lostromb:
If a DLC file goes missing, the user can't even open the DLC dialog to fix it, which leads to bad user experience. This PR adds File.Exists() checks to both the DLC dialog window and the content loader to ignore missing files rather than crashing the program.

### DIFF
--- a/Ryujinx.HLE/HOS/ApplicationLoader.cs
+++ b/Ryujinx.HLE/HOS/ApplicationLoader.cs
@@ -335,7 +335,14 @@ namespace Ryujinx.HLE.HOS
                 {
                     foreach (DlcNca dlcNca in dlcContainer.DlcNcaList)
                     {
-                        _device.Configuration.ContentManager.AddAocItem(dlcNca.TitleId, dlcContainer.Path, dlcNca.Path, dlcNca.Enabled);
+                        if (File.Exists(dlcContainer.Path))
+                        {
+                            _device.Configuration.ContentManager.AddAocItem(dlcNca.TitleId, dlcContainer.Path, dlcNca.Path, dlcNca.Enabled);
+                        }
+                        else
+                        {
+                            Logger.Warning?.Print(LogClass.Application, $"Cannot find AddOnContent file {dlcContainer.Path}. It may have been moved or renamed");
+                        }
                     }
                 }
             }
--- a/Ryujinx.HLE/HOS/ApplicationLoader.cs
+++ b/Ryujinx.HLE/HOS/ApplicationLoader.cs
@@ -341,7 +341,7 @@ namespace Ryujinx.HLE.HOS
                         }
                         else
                         {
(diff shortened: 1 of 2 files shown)


gdkchan:
Potentially closes #1366, but someone will need to check if the other issues have already been fixed.

lostromb:
oh wow there have been a lot of bugs filed for this.

I would also like to fix the weirdness of "enabled DLCs are not marked as enabled when you open the dialog" but I don't understand the data flow well enough to implement it (I assume it would require adding new enabled/disabled flag to the json schema). Also this data is stored as `struct` which seems like an odd choice to me

lostromb:
My other concern is that allowing games to boot with missing DLC might somehow mess up the save state on the guest if it's not equipped to handle that scenario. Like for example if you have a Torna The Golden Country save but then you remove that DLC from the content loader, maybe the XC2 main game corrupts or deletes that save file somehow. Which makes me wonder if it would be better to either allow the crash on startup or (somewhat better user experience) refuse to boot until the missing DLC errors are resolved.

gdkchan:
It doesn't delete the save file, the game tells you that you need the DLC to load the save. Other games could behave differently, of course.

lostromb:
Right.... I'm leaning towards just showing a clear error to the user and aborting boot just to be on the absolute safe side... but I don't know how to do that in this environment. So I might alter the runtime check to log a clear error to console and then stop.

gdkchan:
I think it should be handled in the same way as missing update files for consistency. IIRC it just reverts to the base version if an update is missing, without showing errors. Logging an error is fine.

lostromb:
fixed the "enabled" flag as well as I could, which addresses one more feedback on #1366

Caian (approved):
LGTM, congratulations!

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -335,7 +335,14 @@ private void LoadNca(Nca mainNca, Nca patchNca, Nca controlNca)
...
                     foreach (DlcNca dlcNca in dlcContainer.DlcNcaList)
                     {
-                        _device.Configuration.ContentManager.AddAocItem(dlcNca.TitleId, dlcContainer.Path, dlcNca.Path, dlcNca.Enabled);
+                        if (File.Exists(dlcContainer.Path))
+                        {
+                            _device.Configuration.ContentManager.AddAocItem(dlcNca.TitleId, dlcContainer.Path, dlcNca.Path, dlcNca.Enabled);
+                        }
+                        else
+                        {
+                            Logger.Warning?.Print(LogClass.Application, $"Cannot find AddOnContent file {dlcContainer.Path}. It may have been moved or renamed");
```
AcK77, line 344:
```suggestion
                            Logger.Warning?.Print(LogClass.Application, $"Cannot find AddOnContent file {dlcContainer.Path}. It may have been moved or renamed.");
```

AcK77 (approved):
LGTM instead of a little nit

Thog (approved):

Issue #2863, opened 2021-11-20
Texture disposal hangs Ryujinx with stop emulation on certain games after playing for some time (including Xenoblade, SMTV, Pokemon)


d3xMachina:
### What's the issue you encountered?

Clicking on "Stop emulation" makes Ryujinx hang without closing any of its windows.

### How can the issue be reproduced?

It happens sometimes when you do the above when playing a game (I did it with Pokemon Brillant Diamond).

### Stack trace (logs don't show anything)

> Ryujinx.Graphics.GAL.dll!Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.New<Ryujinx.Graphics.GAL.Multithreading.Commands.Texture.TextureReleaseCommand>() Line 162   C#
> Ryujinx.Graphics.GAL.dll!Ryujinx.Graphics.GAL.Multithreading.Resources.ThreadedTexture.Release() Line 112   C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.Image.Texture.DisposeTextures() Line 1486   C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.Image.Texture.Dispose() Line 1516  C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.Image.TextureCache.Dispose() Line 1093   C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.Memory.PhysicalMemory.Destroy() Line 325 C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.GpuContext.RunDeferredActions() Line 246 C#
> Ryujinx.Graphics.Gpu.dll!Ryujinx.Graphics.Gpu.GpuContext.Dispose() Line 270   C#
> Ryujinx.HLE.dll!Ryujinx.HLE.Switch.DisposeGpu() Line 151 C#
> Ryujinx.dll!Ryujinx.Ui.GlRenderer.Dispose(bool disposing) Line 124   C#
> [External Code]   
> Ryujinx.dll!Ryujinx.Ui.MainWindow.SwitchToGameTable() Line 910 C#
> Ryujinx.dll!Ryujinx.Ui.MainWindow.CreateGameWindow.AnonymousMethod__89_0(object <p0>, System.EventArgs <p1>) Line 966   C#
> [External Code]   
[...]

PR #2868, opened 2021-11-21
Use a semaphore signal instead of Thread.Sleep to handle blocking on threaded renderer queue

lostromb:
Rather than doing `Thread.Sleep` to wait on the queue if the threaded renderer gets backed up, use a semaphore to manage the number of available resources, so waiters will be automatically signaled as soon as the queue is available again. This should theoretically reduce microstutter in GPU heavy workloads.

### DIFF
--- a/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
+++ b/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
@@ -34,6 +34,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
         private bool _running;
 
         private AutoResetEvent _frameComplete = new AutoResetEvent(true);
+        private SemaphoreSlim _queueSlotsAvailable = new SemaphoreSlim(QueueCount);
 
         private ManualResetEventSlim _galWorkAvailable;
         private CircularSpanPool _spanPool;
@@ -136,7 +137,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
                     }
 
                     _consumerPtr = (_consumerPtr + 1) % QueueCount;
-
+                    _queueSlotsAvailable.Release();
                     Interlocked.Decrement(ref _commandCount);
                 }
             }
@@ -154,13 +155,9 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
         internal ref T New<T>() where T : struct
         {
-            while (_producerPtr == (_consumerPtr + QueueCount - 1) % QueueCount)
-            {
-                // If incrementing the producer pointer would overflow, we need to wait.
-                // _consumerPtr can only move forward, so there's no race to worry about here.
-
-                Thread.Sleep(1);
-            }
+            // If incrementing the producer pointer would overflow, we need to wait.
+            // This semaphore controls access to the fixed number of queue resources
+            _queueSlotsAvailable.Wait();
 
             int taken = _producerPtr;
             _lastProducedPtr = taken;
@@ -432,6 +429,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
             // Dispose events.
             _frameComplete.Dispose();
+            _queueSlotsAvailable.Dispose();
             _galWorkAvailable.Dispose();
             _invokeRun.Dispose();
 
--- a/Ryujinx.Graphics.Gpu/GpuContext.cs
+++ b/Ryujinx.Graphics.Gpu/GpuContext.cs
@@ -255,7 +255,6 @@ namespace Ryujinx.Graphics.Gpu
         /// </summary>
         public void Dispose()
         {
-            Renderer.Dispose();
             GPFifo.Dispose();
             HostInitalized.Dispose();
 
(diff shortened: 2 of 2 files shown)


lostromb:
I would like to produce some evidence that this actually improves performance measurably, but as this code path only triggers in worst-case scenarios (when queue is absolutely full) it's difficult to quantify. I can say that if I artificially restrict the queue to 500 elements then this change does improve FPS and reduce FIFO load. It doesn't trigger the block as often at the current queue size of 10000 but it does still happen, usually at the end of loading screens or when something new pops in and triggers a bunch of render calls.

GamerzHell9137:
@lostromb by any chance are you in the Ryujinx discord?

lostromb:
> @lostromb by any chance are you in the Ryujinx discord?

Yeah as Durandal

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -136,7 +137,7 @@ public void RenderLoop()
                     }
 
                     _consumerPtr = (_consumerPtr + 1) % QueueCount;
-
+                    _queueSlotsAvailable.Release();
```
lostromb, line 140: check that ordering is correct here on these atomic operations

d3xMachina:
When I do "Stop emulation" with this PR in any game it hangs and I get the following error : 

> |E| Application : Unhandled exception caught: System.ObjectDisposedException: The semaphore has been disposed.
>       at System.Threading.SemaphoreSlim.CheckDispose()
>       at System.Threading.SemaphoreSlim.Wait(Int32 , CancellationToken )
>       at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.New[T]()
>       at Ryujinx.Graphics.GAL.Multithreading.Resources.ThreadedProgram.Dispose()
>       at Ryujinx.Graphics.Gpu.Shader.ShaderBundle.Dispose()
>       at Ryujinx.Graphics.Gpu.Shader.ShaderCache.Dispose()
>       at Ryujinx.Graphics.Gpu.Memory.PhysicalMemory.Destroy()
>       at Ryujinx.Graphics.Gpu.GpuContext.RunDeferredActions()
>       at Ryujinx.Ui.GlRenderer.Dispose(Boolean disposing)
>       at GLib.Object.Dispose()
>       at Ryujinx.Ui.MainWindow.SwitchToGameTable()
>       at Gtk.Application.InvokeCB.Invoke()
>       at GLib.Timeout.TimeoutProxy.Handler()

d3xMachina:
I think the last commit should fix this issue #2863

riperiperi:
The purpose of this sleep is to save cycles if the backend thread is too far behind command submission. There are so many commands in the queue, that we can afford to wait 1ms for them to complete before we start submitting again.

I suppose it is possible that the backend queue somehow empties itself in that 1ms, but I think it's really unlikely.

I have issues with this approach:

- The semaphore release will happen on every completed command. This loop is incredibly hot, so this will have a performance impact even if the queue is not full.
- The GPU thread will be spinning _very_ aggressively when commands are full, as it'll be woken every time _one_ command is completed. Since this is a slim semaphore, it will be spin waiting, so the GPU and backend threads will both be at 100%, instead of potentially 50% and 100%.

There is room for improving this, as it's not impossible that all commands are consumed in 1ms, or so many commands are consumed that the queue empties and the waiting is the other way around, but I think a solution to this should try to avoid signals for each and every command. Queue submission should resume after _hundreds_ of command slots are available, rather than just one, so that the gpu can proceed unimpeded and the thread can truly yield to lower CPU usage.

I think you can trigger the GPU backend being behind on super mario odyssey's metro kingdom in certain locations. It might be worth logging the number of commands before (max count) and after the 1ms wait in master when the wait triggers, to see how many it gets though.

lostromb:
The main reasoning I had for this change is the fact that `Thread.Sleep` doesn't guarantee an exact amount of time waited, and under the default settings in Windows (without some winmm timer fiddling which I see you have already done), if you ask for a 1ms wait you actually get a 16ms wait. Even with the winmm tweak, `Thread.Sleep(1)` still takes 2ms, and I was worried about that delay causing microstutter. But I also do see your point about the semaphore causing too many unnecessary wakeups and leading to extra overhead. I suppose the ideal here would be some kind of semaphore that did wakeups in batches of 100 or something.....pretty sure that doesn't exist.

I was looking at thread sleep and timers in general and started working on this [experimental change](https://github.com/lostromb/Ryujinx/commit/ebf0afda0ecd2ddf6d139a2eca08d9455aa09195).
It would formalize the winmm stuff a little better and provide a centralized place for fulfilling high precision sleeps like here or in the audio renderer. I was originally thinking it would help smooth out platform-specific behavior but I tested recently on Linux and their timings are pretty much exactly what you ask for. But maybe that would be a better implementation of this change.

lostromb:
Will rescope this to just the change in renderer disposal order to try and address #2863, then maybe an overall revisit of `Thread.Sleep` later

PR #2876, opened 2021-11-23
Fix FLO.SH shader instruction with a input of 0

gdkchan:
This instruction is supposed to return -1 when the input is 0, but `FLO.SH` (leading zeros count,) was implemented using `findMSB` and then XORing the output index by 31 to get the inverse. While this gives correct results for results >= 0, it would be incorrect for a result of -1.

The new implementation now uses `findLSB` (trailing zeros count) with a bit-reversed input. If the input is 0, then the bit-reverse is not going to change it, and it will still produce the correct -1 result. Non-zero inputs will be bit-reversed and the trailing zeros count can then give the same result a leading zeros count instruction would.

One thing to note is that `findLSB` is implemented on NVIDIA by using `BFE.BREV x, x, 0x2000` followed by `FLO.SH x, x`, so we end producing a `findLSB(bitfieldReverse(bitfieldReverse(x)))` pattern, it might be worth to optimize out the double bit-reversal at some point.

I'm not sure which games this affects as this bug was found using a fuzzer. Testing is welcome.


riperiperi (approved):
Nice catch.

Thog (approved):

AcK77 (approved):

PR #2885, opened 2021-11-28
Fix SUATOM and other texture shader instructions with RZ dest

gdkchan:
When the register is RZ, it should set the destination operand to null, because setting it to a constant of zero would cause it to generate `0 = ...;` which is obviously not valid GLSL code. I found the issue only on SUATOM, but other texture instructions also had a similar issue, but unlike SUATOM, I don't think it would be ever RZ on  regular shader code, as doing a texture sample but discarding the results makes no sense.

I'm not sure which games this affects as this bug was found using a fuzzer. Testing is welcome.


Thog (approved):

AcK77 (approved):

riperiperi (approved):
Nice.

PR #2886, opened 2021-11-28
Reduce some unnecessary allocations in DMA handler

lostromb:
Reduce cases where unnecessary copies of data are made in DMA handler, and also add a fast path for pure linear transfers where we can bypass the stride offset calculator on each element and thus use a faster copy method. Should reduce load times slightly and give better performance / less GC stutter during FMVs

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -1,11 +1,9 @@
 ﻿using Ryujinx.Common;
-using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.Device;
 using Ryujinx.Graphics.Gpu.Engine.Threed;
 using Ryujinx.Graphics.Gpu.Memory;
 using Ryujinx.Graphics.Texture;
 using System;
-using System.Buffers;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Runtime.Intrinsics;
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -214,8 +214,6 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                 bool completeSource = IsTextureCopyComplete(src, srcLinear, srcBpp, srcStride, xCount, yCount);
                 bool completeDest = IsTextureCopyComplete(dst, dstLinear, dstBpp, dstStride, xCount, yCount);
 
-                Logger.Info?.PrintMsg(LogClass.Gpu, $"DMA Copy srcSize {srcSize} dstSize {dstSize} srcBpp {srcBpp} dstBpp {dstBpp} cS {completeSource} cD {completeDest} sL {srcLinear} dL {dstLinear}");
-
                 if (completeSource && completeDest)
                 {
                     var target = memoryManager.Physical.TextureCache.FindTexture(
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -211,9 +211,10 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                 bool completeSource = IsTextureCopyComplete(src, srcLinear, srcBpp, srcStride, xCount, yCount);
                 bool completeDest = IsTextureCopyComplete(dst, dstLinear, dstBpp, dstStride, xCount, yCount);
 
+                ReadOnlySpan<byte> srcSpan = memoryManager.GetSpan(srcGpuVa + (ulong)srcBaseOffset, srcSize, true);
+
                 if (completeSource && completeDest)
                 {
-                    ReadOnlySpan<byte> srcSpan = memoryManager.GetSpan(srcGpuVa + (ulong)srcBaseOffset, srcSize, true);
                     var target = memoryManager.Physical.TextureCache.FindTexture(
                         memoryManager,
                         dst,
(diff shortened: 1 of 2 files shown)


lostromb:
whoops, this may be a bit premature... the "1bpp fast path" seems to break unreal engine games...

Mou-Ikkai:
> whoops, this may be a bit premature... the "1bpp fast path" seems to break unreal engine games...

No biggie! No need to delete, just report issues, then fix them before saying it's ready by commenting or editing your main message!

lostromb:
Regression should be fixed now. There are many other code paths related to texture copy / DMA transfer that I would like to optimize - the easiest way I would likely go about it would be to make a static `ArrayPool` instance reserved for large texture copies. That should go a long way to reducing the runtime overhead and stutter we see in games now.

For instance, when some FMVs are playing and decode video to texture, our garbage collection performance is absolutely atrocious as we're allocating 80+ MB of transient byte buffers onto the large object heap per second. But that would have to be more involved as it would have to catch each instance of texture data becoming obsolete and return the backing array to the pool.

Thog:
Kind of would prefer both changes to be separated in two PRs @lostromb as people working on the kernel side aren't necessary the same one as the one working on the gpu side 😅

Ryujinx.HLE/HOS/Kernel/Threading/KSynchronization.cs
```diff
@@ -58,48 +59,54 @@ public KernelResult WaitFor(Span<KSynchronizationObject> syncObjs, long timeout,
             }
             else
             {
-                LinkedListNode<KThread>[] syncNodes = new LinkedListNode<KThread>[syncObjs.Length];
```
Thog, line 61: We know that we have 64 elements max (see usage of this API in the codebase)

I wonder if it would be possible to use a ``stackalloc`` here?

lostromb: That was my first thought actually, but I didn't know the limits.

lostromb:
Reverted kernel changes for now as they are lower priority

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -224,43 +225,64 @@ private void LaunchDma(int argument)
...
                         }
                     }
+
                     return true;
                 }
 
+                // OPT: This allocates a (potentially) huge temporary array and then copies an existing
+                // region of memory into it, data that might get overwritten entirely anyways. Ideally this should
+                // all be rewritten to use pooled arrays, but that gets complicated with packed data and strides
+                Span<byte> dstSpan = memoryManager.GetSpan(dstGpuVa + (ulong)dstBaseOffset, dstSize).ToArray();
```
gdkchan, line 284: the `MemoryManager` has `GetWritableRegion`, which will return a `Memory<byte>` of the memory region. If the region is contiguous, there's no copy or array allocations. If the region is not contiguous though, it will allocate a byte array and do a copy of the data there. So using it could solve the issue in *some* cases.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using Ryujinx.Common;
+using Ryujinx.Common.Logging;
```
gdkchan, line 2: Not needed? I don't see any log calls.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -1,8 +1,10 @@
...
 using Ryujinx.Graphics.Device;
 using Ryujinx.Graphics.Gpu.Engine.Threed;
 using Ryujinx.Graphics.Texture;
 using System;
+using System.Buffers;
```
gdkchan, line 7: Also not needed?

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
lostromb: The `ToArray()` is the killer here.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
lostromb: oh yeah, this was for development. can remove

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
jduncanator: I think what @gdkchan is suggesting here (correct me if I'm wrong), is getting a `Memory<byte>` of the destination address, and passing that into `Convert<T>`, allowing it to work in the underlying memory directly. This would remove the need for any copy at all (or at least would abstract the copy away behind `GetWritableRegion` for the worst case.)

gdkchan: Yea, but how rare the non-contiguous buffers are depends on the game. On UE4 games they are pretty common, since this engine likes remapping sub-ranges of memory for some reason. That's why it might not be an ideal solution still, but in that case we could look into adding a pooled array on top of it. Using the `WritableRegion` requires 0 copies when the region is contiguous, so it is even better than a pooled array for that case (which would still require copies), so it should be preferred if possible.

FICTURE7: Perhaps unrelated, but I remember looking at this in the past and I thought of having a `VMemory<T>` data structure or something to avoid these redundant copies.

```csharp
[Flags]
enum SignalType {
  Read = 1,
  Write = 2,
  Precise = 4
}

struct VMemory<T> : IEnumerable<Span<T>> {
  IVirtualMemoryManager MemoryManager { get; }
  bool IsMapped => ...;
  ulong Address { get; }
  int Length { get; }

  // Optional: For when VMemory<T> is small and contiguous, so we avoid the overhead from SpanEnumerator.
  bool IsSingle => ...;
  Span<T> SingleSpan => ...;

  // Optional: If we're caching the result of IsSingle & SingleSpan, then it may happen that they go stale
  // because memory mappings have changed.
  //
  // I am thinking when the VMemory<T> is stored in another object to track their data for long durations.
  void Invalidate();

  void Signal(SignalType type);
  void Reprotect(MemoryPermission permission);

  // Optional: Low priority APIs but nice to have.
  ref T this[int index] { get; }
  VMemory<T> Slice(int start);
  VMemory<T> Slice(int start, int length);
  T[] ToArray();

  SpanEnumerator GetEnumerator() => new SpanEnumerator(this);
  
  // An iterator which iterates over the pages within this VMemory<T>. For regions which are contiguous,
  // return a single Span<T> for them.
  // 
  // Ideally a zero GC alloc iterator. Might need to modify IVirtualMemoryManager a bit for this to work
  // for all VirtualMemoryManager with no allocations.
  struct SpanEnumerator : IEnumerator<Span<T>> {
    // ...
  }
}

static class SpanVMemoryExtensions {
  // Zero alloc copy from a Span<T> to a possibly non-contiguous memory region.
  static void CopyTo(this Span<T> self, VMemory<T> dest) {
    // Optional: Add a fast path for when dest IsSingle.
    foreach (Span<T> page in dest) {
      self.Slice(0, page.Length).CopyTo(dest);
      self = self.Slice(page.Length);
    }
  }
}
```

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
lostromb: removed.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
lostromb: removed

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
lostromb: I mean, that sounds kind of like a duplicate of what MemoryManager already does. I had considered it, but it seems much easier for code that's expecting contiguous Span<byte> to copy to a pooled buffer when needed and then just use that for the algorithm, since memory copies are relatively cheap and using a pooled buffer properly would mean amortized zero allocations.

jduncanator: If you're optimising the code, then you might as well optimise it fully, rather than just taking the path with least effort. The code can operate on the destination memory directly, we should be doing that in every case that it's possible, and handling the less common case (non-contiguous memory) separately if required.

I don't like the idea of the call site handling renting an array, copying the memory, and then freeing the array. It sounds prone to error and memory leaks. If we're going down that path, I'd prefer an API be exposed via the `MemoryManager` itself that controls the lifetime of that rented buffer, and the copying in and out of the memory internally.

lostromb: Okay, see #2893 for the more fully optimized version. I agree it would be nice to rewrite `GetWritableRegion` to return some kind of union of the direct access span (if possible) and a pooled buffer containing a copy of the data if non-contiguous. But I also wanted to kind of keep the scope of this change to be smaller and mainly catch the one bogus case of texture->texture copies always allocating a full copy of the data which never gets used.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -224,43 +223,64 @@ private void LaunchDma(int argument)
...
+                        memoryManager.Write(dstGpuVa + (ulong)dstBaseOffset, srcSpan);
                         return;
                     }
                 }
 
                 unsafe bool Convert<T>(Span<byte> dstSpan, ReadOnlySpan<byte> srcSpan) where T : unmanaged
                 {
-                    fixed (byte* dstPtr = dstSpan, srcPtr = srcSpan)
+                    if (srcLinear && dstLinear &&
+                        srcBpp == dstBpp)
```
riperiperi, line 239: Unusual wrapping here.

riperiperi (approved):
Looks good here. The biggest change to note seems to be that dstSpan was allocated even when the fast paths are used, which is unnecessary and slowing down copies that use the fast path (videos).

gdkchan (approved):
lgtm, thanks! Sorry for the delay.

PR #2893, opened 2021-12-01
Creating a central pool for large buffers and using these pooled buffers for moving texture data around.

lostromb:
This will likely conflict heavily with changes in the Vulkan branch so I am just throwing this out as a draft for now. I know the comments on that branch mentioned reducing copies for texture and maybe vertex data too, and I don't know if there is a solid plan for that, but consider this as one proposed design?

This PR reduces the CLR heap allocation rate in Mario Odyssey by a whopping 40-50%, from ~166MB / sec to ~81MB / sec. This means less GC stutter, better performance on things like FMVs which do frequent large buffer->texture copies, and likely better performance on CPUs where memory is a bottleneck (haven't fully tested perf yet).

### DIFF
--- a/Ryujinx.Common/Pools/BufferPool.cs
+++ b/Ryujinx.Common/Pools/BufferPool.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using System.Buffers;
+
+namespace Ryujinx.Common.Pools
+{
+    /// <summary>
+    /// A centralized, static pool for getting multipurpose buffers of any type.
+    /// "But why does this exist when you have <see cref="ArrayPool{T}"/>?" This actually just wraps around
+    /// ArrayPool. The difference is that this also tracks the requested size of the allocated
+    /// buffer, so you can pass it around and users of it know how much of the data is actually valid,
+    /// compared to ArrayPool which just gives you an arbitrary sized array. Also, returning a strongly
+    /// typed <see cref="PooledBuffer{T}"/> codifies the fact that the buffer is part of a pool, and that code
+    /// can more clearly manage lifetime and disposal automatically.
+    /// </summary>
+    public static class BufferPool<T>
+    {
+        public const int DEFAULT_BUFFER_SIZE = 65536;
+
+        private static readonly PooledBuffer<T> EMPTY_POOLED_BUFFER = new PooledBuffer<T>(Array.Empty<T>(), -1);
+
+        /// <summary>
+        /// Returns a disposable <see cref="PooledBuffer{T}" /> instance containing an array of at least the specified number of elements.
+        /// </summary>
+        /// <param name="minimumRequestedSize">The minimum number of elements that you require</param>
+        /// <returns>A pooled buffer of at least the requested size. Buffers are not cleared between rentals, so it may initially contain garbage.</returns>
+        public static PooledBuffer<T> Rent(int minimumRequestedSize = DEFAULT_BUFFER_SIZE)
+        {
+            if (minimumRequestedSize < 0)
+            {
+                throw new ArgumentOutOfRangeException("Requested buffer size must be a positive number");
+            }
+
+            if (minimumRequestedSize == 0)
+            {
+                return EMPTY_POOLED_BUFFER;
+            }
+
+            PooledBuffer<T> returnVal = new PooledBuffer<T>(ArrayPool<T>.Shared.Rent(minimumRequestedSize), minimumRequestedSize);
+#if DEBUG
+            returnVal.MarkRented();
+#endif
+            return returnVal;
+        }
+    }
+}
--- a/Ryujinx.Common/Pools/BufferPool.cs
+++ b/Ryujinx.Common/Pools/BufferPool.cs
@@ -19,11 +19,14 @@ namespace Ryujinx.Common.Pools
         private static readonly PooledBuffer<T> EMPTY_POOLED_BUFFER = new PooledBuffer<T>(Array.Empty<T>(), -1);
 
         /// <summary>
-        /// Returns a disposable <see cref="PooledBuffer{T}" /> instance containing an array of at least the specified number of elements.
+        /// Returns a disposable <see cref="PooledBuffer{T}" /> instance backed by a contiguous array of at least the requested number of elements.
+        /// The internal array is only exposed as a <see cref="Span{T}"/> to prevent user code from storing references to it outside
+        /// of the potential lifetime of the pooled handle.
         /// </summary>
         /// <param name="minimumRequestedSize">The minimum number of elements that you require</param>
-        /// <returns>A pooled buffer of at least the requested size. Buffers are not cleared between rentals, so it may initially contain garbage.</returns>
-        public static PooledBuffer<T> Rent(int minimumRequestedSize = DEFAULT_BUFFER_SIZE)
+        /// <param name="clearArray">Whether to clear the contents of the buffer before returning. If false (by default), the buffer may initially contain garbage.</param>
+        /// <returns>A pooled buffer of exactly the requested size.</returns>
+        public static PooledBuffer<T> Rent(int minimumRequestedSize = DEFAULT_BUFFER_SIZE, bool clearArray = false)
         {
             if (minimumRequestedSize < 0)
             {
@@ -35,7 +38,13 @@ namespace Ryujinx.Common.Pools
                 return EMPTY_POOLED_BUFFER;
             }
 
-            PooledBuffer<T> returnVal = new PooledBuffer<T>(ArrayPool<T>.Shared.Rent(minimumRequestedSize), minimumRequestedSize);
+            T[] array = ArrayPool<T>.Shared.Rent(minimumRequestedSize);
+            if (clearArray)
+            {
+                Array.Clear(array, 0, minimumRequestedSize);
+            }
+
+            PooledBuffer<T> returnVal = new PooledBuffer<T>(array, minimumRequestedSize);
 #if DEBUG
             returnVal.MarkRented();
 #endif
--- a/Ryujinx.Common/Pools/BufferPool.cs
+++ b/Ryujinx.Common/Pools/BufferPool.cs
@@ -45,7 +45,7 @@ namespace Ryujinx.Common.Pools
             }
 
             PooledBuffer<T> returnVal = new PooledBuffer<T>(array, minimumRequestedSize);
-#if DEBUG
+#if TRACK_BUFFERPOOL_LEAKS
             returnVal.MarkRented();
 #endif
             return returnVal;
--- a/Ryujinx.Common/Pools/BufferPool.cs
+++ b/Ryujinx.Common/Pools/BufferPool.cs
@@ -6,11 +6,12 @@ namespace Ryujinx.Common.Pools
     /// <summary>
     /// A centralized, static pool for getting multipurpose buffers of any type.
     /// "But why does this exist when you have <see cref="ArrayPool{T}"/>?" This actually just wraps around
-    /// ArrayPool. The difference is that this also tracks the requested size of the allocated
-    /// buffer, so you can pass it around and users of it know how much of the data is actually valid,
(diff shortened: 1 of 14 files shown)


Ryujinx.Common/Pools/PooledBuffer.cs
```diff
@@ -0,0 +1,121 @@
...
+            Dispose(true);
+#if TRACK_BUFFERPOOL_LEAKS
+            GC.SuppressFinalize(this);
+#endif
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            // prevent multiple disposal atomically
+            if (Interlocked.CompareExchange(ref _disposed, 1, 0) != 0)
```
gdkchan, line 106: This could be just `Interlocked.Exchange(ref _disposed, 1) != 0` since it is only ever set to 1.

Ryujinx.Common/Pools/BufferPool.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// of the potential lifetime of the pooled handle.
+        /// </summary>
+        /// <param name="minimumRequestedSize">The minimum number of elements that you require</param>
+        /// <param name="clearArray">Whether to clear the contents of the buffer before returning. If false (by default), the buffer may initially contain garbage.</param>
+        /// <returns>A pooled buffer of exactly the requested size.</returns>
+        public static PooledBuffer<T> Rent(int minimumRequestedSize = DEFAULT_BUFFER_SIZE, bool clearArray = false)
+        {
+            if (minimumRequestedSize < 0)
+            {
+                throw new ArgumentOutOfRangeException("Requested buffer size must be a positive number");
```
gdkchan, line 34: The single string overload of `ArgumentOutOfRangeException` expects a parameter name, not a exception message, see: [https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=net-6.0#System_ArgumentOutOfRangeException__ctor_System_String_](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=net-6.0#System_ArgumentOutOfRangeException__ctor_System_String_)

Ryujinx.Common/Pools/BufferPool.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// A centralized, static pool for getting multipurpose buffers of any type.
+    /// "But why does this exist when you have <see cref="ArrayPool{T}"/>?" This actually just wraps around
+    /// ArrayPool. The difference is that the rented array is hidden inside of a <see cref="PooledBuffer{T}"/>
+    /// handle and is only exposed as a <see cref="Span{T}"/>. This makes it harder to leak references to the
+    /// pooled array, which makes it a bit easier to track ownership. The PooledBuffer handle is <see cref="IDisposable"/>
+    /// so code analysis can enforce better practice as well. In addition, it also tracks the initial requested
+    /// size of the allocated buffer, so you can pass it around and users of it know how much of the data
+    /// is actually valid, compared to ArrayPool which just gives you an arbitrary sized array.
+    /// </summary>
+    public static class BufferPool<T>
```
gdkchan, line 16: This doesn't seems to be too different from .NET own `ArrayPool`, see https://docs.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-6.0
Granted I only gave a quick look at this implementation, any reason we can't use the .NET one?

Ryujinx.Graphics.GAL/ITexture.cs
```diff
@@ -16,8 +17,8 @@ public interface ITexture
 
         ReadOnlySpan<byte> GetData();
 
-        void SetData(ReadOnlySpan<byte> data);
-        void SetData(ReadOnlySpan<byte> data, int layer, int level);
+        void SetData(PooledBuffer<byte> data);
```
gdkchan, line 20: Personally I would prefer if all the interfaces on the GAL namespace kept accepting a `ReadOnlySpan<byte>` as it makes the API more flexible, otherwise we *must* use a pooled buffer to pass the data, and passing data sourced from another place becomes difficult.

gdkchan (commented):
FWIW Ryujinx.Graphics.Vic already has a BufferPool, used for a similar purpose, maybe changing it to use the new buffer pool should be considered here too.

gdkchan:
> I know the comments on that branch mentioned reducing copies for texture and maybe vertex data too, and I don't know if there is a solid plan for that, but consider this as one proposed design?

Which comment are you referring to? I don't recall any plan to reduce copies as part of Vulkan change (except for maybe GPU side copies of texture data for example, which has nothing to do with this change).

Ryujinx.Common/Pools/BufferPool.cs
lostromb: oh right

Ryujinx.Common/Pools/BufferPool.cs
lostromb: I thought the comment explained the answer to this pretty clearly. 

1. It's absolutely clear to all users that you get a handle to a pooled array rather than just a loose `byte[]`
2. The buffer's length is well-defined
3. The underlying array is only exposed as a `Span` so it's harder to leak references to the pooled buffer
4. It implements IDisposable so static code analysis can enforce better usage

gdkchan: Yea sorry, now I noticed the comment explains exactly what I asked.

Ryujinx.Graphics.GAL/ITexture.cs
lostromb: Yes, I agree. The main reason I had to change it was to accommodate the `TextureSetDataCommand`, which needs to hold a longer-lived reference to the data. This could be reverted if `TextureSetDataCommand` instead made its own copy of the `ReadOnlySpan` that was passed to it -- which it already does today, but we could make it so the copy goes to a pooled buffer internally to amortize the allocation. Or possibly change this to accept `Memory<byte>` instead, although that makes ownership of a data a little harder to manage.

lostromb:
I guess I was thinking of the "avoids copies" language, but yeah, that's for the backend whereas this change affects graphics common.

jduncanator:
What's wrong with `MemoryPool<T>`?

Ryujinx.Common/Pools/BufferPool.cs
jduncanator: MemoryPool<T> is exactly that, built on ArrayPool<T> internally. It seems like you've just reimplemented MemoryPool<T>.

FICTURE7:
I prototyped using `MemoryPool<T>` in the past, a while back, see https://github.com/FICTURE7/Ryujinx/tree/proto/arena-alloc. IIRC by doing that we essentially trade off between possibly blasting the LOH and Gen2 collections to blasting Gen0s, because we trade the large buffer allocation to a `IMemoryOwner<T>`. 

At the time I was looking to see if we could avoid these Gen0 by swapping the `IMemoryOwner` to a struct or something, i.e: perhaps using a struct as a buffer handle, some sort of new primitive. Which would probably cause of lot of coupling between the GAL and the new memory allocation machinery. But I believe such coupling would not be an issue since these APIs usually just need to get a `ReadOnlySpan<T>` somehow, i.e: they already depend on a concrete type anyway.

There is also quite a bit of rippling happening on the API surface to pass around the `MemoryPool<byte>` which could probably be avoided by centralizing the `MemoryPool<T>` somewhere in the GAL itself. This is also why I wanted to look for a new memory allocation machinery, because we're already coupling the GAL with `MemoryPool<T>` somewhat.

The prototype uses an arena style allocator - probably not necessary.

Anyways the Gen0 collections incurred by the `IMemoryOwner<T>`, and in this case `PooledArray<T>`, are still much better than allocating the array upfront and possibly blasting the LOH and Gen2.

Ryujinx.Common/Pools/BufferPool.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public const int DEFAULT_BUFFER_SIZE = 65536;
+
+        private static readonly PooledBuffer<T> EMPTY_POOLED_BUFFER = new PooledBuffer<T>(Array.Empty<T>(), -1);
```
FICTURE7, line 20: The length needs to be `0` instead of `-1` no? Otherwise `BufferPool<byte>.Rent(0).AsSpan()` will throw an exception.

lostromb: oh yeah I'm dumb. I adapted this code from somewhere else and forgot to check this

Ryujinx.Graphics.GAL/ITexture.cs
lostromb: Reverted this.

Ryujinx.Common/Pools/PooledBuffer.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// Represents a reusable pooled buffer of some type T that was rented from a <see cref="BufferPool{T}"/>.
+    /// This buffer should be disposed when you are done using it, after which it is returned to the pool to be reclaimed.
+    /// </summary>
+    /// <typeparam name="T">The type of data that this buffer holds.</typeparam>
+    public class PooledBuffer<T> : IDisposable
+    {
+        private int _disposed = 0;
+
+        internal PooledBuffer(T[] data, int length)
```
FICTURE7, line 21: Since `Dispose` already depends on `ArrayPool<T>.Shared.Return`, might as well just remove the `T[] data` parameter and call the `ArrayPool<T>.Shared.Rent` in the constructor itself, i.e: make it a `PooledBuffer(int length)`.

lostromb:
> What's wrong with `MemoryPool<T>`? What additional functionality or benefits do we get from reimplementing _another_ pool type in Ryujinx.Common?

I expected to find such a thing in the `Ryujinx.Common.Pools` namespace. Searching for files named MemoryPool.cs....... where is it?

Ryujinx.Common/Pools/PooledBuffer.cs
lostromb: Yeah, fair point. The original code I ported from had the arrays actually be returned to `BufferPool<T>` when they were done, but that's no longer the design.

jduncanator:
> > What's wrong with `MemoryPool<T>`? What additional functionality or benefits do we get from reimplementing _another_ pool type in Ryujinx.Common?
> 
> I expected to find such a thing in the `Ryujinx.Common.Pools` namespace. Searching for files named MemoryPool.cs....... where is it?

It's in System.Buffers, it's provided by the .NET framework.

lostromb:
Well, perf benchmarks didn't show any of the gains I was hoping for. Fart.
image

lostromb:
> What's wrong with `MemoryPool<T>`? What additional functionality or benefits do we get from reimplementing _another_ pool type in Ryujinx.Common?

I didn't know about this, so thanks for bringing it up. However, `MemoryPool` doesn't _quite_ solve the problem entirely, for the same reasons that `ArrayPool` on its own doesn't work. The problem is that when you ask `MemoryPool` for a buffer, it will almost always give you more memory than what you ask for. The way the current code is structured, we're allocating arrays with exact sizes, and then passing them around with the implicit contract that "this array contains the exact data for a single decoded ASTC texture" (or whatever), so in `SetData` we just copy the entire array to the host GPU and that's the texture. But if the array gets replaced with an `IMemoryOwner`, the intended length gets lost and we end up using running into garbage regions, reading more data than intended to various places in DMA, GPU, texture decode, etc. My proposal with `PooledBuffer` here is to just wrap around the runtime-provided pool, while also giving us some benefits like tracking the intended length, and also we could do things like implement a finalizer on the handle to catch cases where the pool is leaking, which should help debugging.

lostromb:
Overall this seems to give about a 1-2% FPS boost, depending on the game. I assume it changes frame pacing as well but I haven't got clear data on that yet.

PR #2895, opened 2021-12-02
nfp: Name support

jozz024:
This pull request adds initial name support for amiibo files.


jduncanator:
This change-set will result in the name of existing Amiibo to be `null` when they're read back in from disk, as the existing JSON files will not contain a `name` property. You'll need to add some migration code to default existing Amiibo (those with `null` names) to `Ryujinx`.

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
```diff
@@ -84,7 +84,7 @@ public static RegisterInfo GetRegisterInfo(string amiiboId, string nickname)
                 Reserved1       = new Array64<byte>(),
                 Reserved2       = new Array58<byte>()
             };
-            Encoding.ASCII.GetBytes("Ryujinx").CopyTo(registerInfo.Nickname.ToSpan());
+            Encoding.ASCII.GetBytes(amiiboFile.Name).CopyTo(registerInfo.Nickname.ToSpan());
```
jduncanator, line 87: This call will fail when loading existing Amiibo on disk, as their `Name` field will not exist in the JSON, and so be set to `null`. `GetBytes` will throw an `ArgumentNullException`.

jduncanator (changes requested):

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
jduncanator: You can also avoid a copy here by using `Encoding.ASCII.GetEncoder`, and then calling `encoder.Convert(amiiboFile.Name, registerInfo.Nickname.ToSpan(), true, out _, out _, out _);`. This will allow the encoder to write the string directly to the destination memory, rather than allocating an intermediate byte array and then copying that to the Nickname.

jduncanator: You can also avoid a copy here by using the `Span` overload of `GetBytes`:

```
Encoding.ASCII.GetBytes(amiiboFile.Name, registerInfo.Nickname.ToSpan());
```

jozz024:
one question. would it be possible to implement a menu for when the nickname and mii editor is called to set an amiibo nickname instead of crashing? i'm not really sure how to do it so some pointers would be greatly appreciated.

jozz024:
i adressed your first and third points in the latest commit @jduncanator. for your second point, i know amiibo names are stored in utf-16 internally, but i am not sure on how to do any sort of conversion from X encoding to utf-16

Thog:
> one question. would it be possible to implement a menu for when the nickname and mii editor is called to set an amiibo nickname instead of crashing? i'm not really sure how to do it so some pointers would be greatly appreciated.
> 
> this would be used for people who want to change the name in general or want to change the name from "Ryujinx" to something else

Imo you should try to default to the default nickname (Ryujinx) when the value is null

jozz024:
> Imo you should try to default to the default nickname (Ryujinx) when the value is null

oh i already do, thats what that last commit fixes. it writes "Ryujinx" to the amiiboFile when the value doesnt exist, loads amiiboFile.Name, and then saves the file if the value is null

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
```diff
@@ -84,8 +84,16 @@ public static RegisterInfo GetRegisterInfo(string amiiboId, string nickname)
                 Reserved1       = new Array64<byte>(),
                 Reserved2       = new Array58<byte>()
             };
-            Encoding.ASCII.GetBytes("Ryujinx").CopyTo(registerInfo.Nickname.ToSpan());
-
+            if (amiiboFile.Name == null)
```
AcK77, line 87: Add a new empty line before

Ryujinx.HLE/HOS/Services/Nfc/Nfp/VirtualAmiibo.cs
```diff
@@ -84,8 +84,16 @@ public static RegisterInfo GetRegisterInfo(string amiiboId, string nickname)
...
+            {
+                amiiboFile.Name = "Ryujinx";
+                Encoding.ASCII.GetBytes(amiiboFile.Name, registerInfo.Nickname.ToSpan());
+                SaveAmiiboFile(amiiboFile);
+            }
+            else
+            {
+                Encoding.ASCII.GetBytes(amiiboFile.Name, registerInfo.Nickname.ToSpan());
+            }
```
AcK77, line 96: Add a new empty line after

AcK77 (changes requested):

jduncanator:
> i adressed your first and third points in the latest commit @jduncanator. for your second point, i know amiibo names are stored in utf-16 internally, but i am not sure on how to do any sort of conversion from X encoding to utf-16

UTF-16 in C#, confusingly, is handled by the `UnicodeEncoding` class. If it's actually stored internally as UTF-16, then you can simply just do `Encoding.Unicode.GetBytes(amiiboFile.Name, registerInfo.Nickname.ToSpan())` instead, however given ASCII is working correctly at the moment, then I doubt `registerInfo.Nickname` is expecting a UTF-16 string, and may require some more changes elsewhere.

jozz024:
tbf i tried utf-8 at one point and it worked, so it might not care about encoding?

gdkchan:
> tbf i tried utf-8 at one point and it worked, so it might not care about encoding?

ASCII and UTF-8 are compatible. That is, if you decode ASCII text as UTF-8, it will still decode correctly. But the opposite is not true, because the range of characters supported by ASCII is far more limited than UTF-8.

AcK77:
Do you plan to still working on it?

jozz024:
i do plan to, i've just been caught up in some other projects and haven't had a chance to work on it

german77:
All games call the cabinet applet to change the owner of the mii. Including the amiibo name. You either want to add an UI to set the names outside of emulation or implement the cabinet applet. Not sure if it can be ran like the mii editor

jozz024:
i'm well aware @german77, but my issue is that it's only doing that when i'm having it decode as big endian unicode, so it's registering as an invalid name

jozz024:
apologies for never finishing this, i have been busy with other projects and i just haven't had motivation to work on it
if anyone with the authority wants to give me a rundown on what they want before this pull request gets merged, please let me know and i'll get to it asap, as i don't want this to stay stagnant forever

jozz024:
Closing due to lack of interest in amiibo, and now someone has done something better, so I can close this with peace of mind.

PR #2900, opened 2021-12-04
kernel: Improve GetInfo readability and update to 13.0.0

Thog:
This PR updates ``GetInfo`` to make it more readeable (by using an ``InfoType`` enum) and implement missing types.

This is based on my kernel 13.0.0 researches.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
@@ -60,6 +60,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Process
         public KProcessCapabilities Capabilities { get; private set; }
 
         public ulong TitleId { get; private set; }
+        public bool IsApplication { get; private set; }
         public long Pid { get; private set; }
 
         private long _creationTimestamp;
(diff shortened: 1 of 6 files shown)


Ryujinx.HLE/HOS/Kernel/SupervisorCall/InfoType.cs
```diff
@@ -0,0 +1,33 @@
...
+        SystemResourceSizeUsed,
+        ProgramId,
+        // NOTE: Added in 4.0.0, removed in 5.0.0.
+        InitialProcessIdRange,
+        UserExceptionContextAddress,
+        TotalNonSystemMemorySize,
+        UsedNonSystemMemorySize,
+        IsApplication,
+        FreeThreadCount,
+        ThreadTickCount
```
Thog, line 31: This is mapped to 25 instead of 0xF0000002 in purpose as this was changed in 13.x.

AcK77 (approved):

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
```diff
@@ -193,6 +194,7 @@ public KProcess(KernelContext context) : base(context)
             _memRegion = memRegion;
             _contextFactory = contextFactory ?? new ProcessContextFactory();
             _customThreadStart = customThreadStart;
+            IsApplication = creationInfo.Flags.HasFlag(ProcessCreationFlags.IsApplication);
```
gdkchan, line 197: Any reason for adding the auto-property instead of just making the property check if the bit on the flag is set directly?

gdkchan (approved):
lgtm, seems more readable now with the enum!
I haven't checked that the enum values match the previously value though, I assume this has already been checked and tested.

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
Thog: Mostly to follow the original logic here

PR #2904, opened 2021-12-04
misc: Sync Config.json default debug config

Thog:
As title say.

### DIFF
--- a/Ryujinx/Config.json
+++ b/Ryujinx/Config.json
@@ -1,5 +1,5 @@
 {
-  "version": 29,
+  "version": 32,
   "enable_file_log": true,
   "backend_threading": "Auto",
   "res_scale": 1,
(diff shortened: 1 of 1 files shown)


AcK77 (approved):

gdkchan (approved):
Can't we just delete this file? Since the emulator can generate it anyway. Seems like it would be one thing less to worry about.

Thog:
> Can't we just delete this file? Since the emulator can generate it anyway. Seems like it would be one thing less to worry about.

I guess I could do a follow up removing that file and the schema tbh yeah

jduncanator:
Why remove the schema, the schema is there so IDEs (and eventually Ryujinx) can verify the .json file is valid, and can provide type hinting and "Intellisense" when editing the JSON file directly.

Thog:
> Why remove the schema? The schema is there so IDEs (and eventually Ryujinx) can verify the .json file is valid, and can provide type hinting and "Intellisense" when editing the JSON file directly.

Considering that no one update it (or miss to do so), I do not think it's too useful as it.

PR #2242, opened 2021-04-24
Implement some missing 32-bit instructions (VPADAL, VRINTR)

piyachetk:
Implemented instructions
- ~~UHADD8~~
- ~~VCVT (between floating-point and fixed-point)~~
- ~~VMLAL (integer) #3677~~
- ~~VPADAL #6185~~
- ~~VPADDL #3677~~
- ~~VQADD #3677~~
- ~~VQMOVN #3677~~
- ~~VQMOVUN #3677~~
- ~~VQSUB #3677~~
- ~~VRINTR #6185~~

Fixes #1344
Fixes #1644
Fixes #1781
Fixes #1811

Ryujinx/Ryujinx-Games-List#145
- This game will now run (with audio issues but it's playable)

Ryujinx/Ryujinx-Games-List#3506
- This game won't crash but it will show some weird artifacts (unplayable)

Ryujinx/Ryujinx-Games-List#2998
Ryujinx/Ryujinx-Games-List#1951
- These games need the VQMOVN/VQMOVUN instruction and should now be fixed but I haven't tested them yet. Unless they also require some other missing 32-bit instructions, they should be able to boot just fine.

Ryujinx/Ryujinx-Games-List#1943
- This game needs VQMOVN/VQMOVUN, VQSUB, and VMLAL. It still crashes as it need the VQDMULH instruction.

PR #2908, opened 2021-12-06
Implement UHADD8 instruction

### LINKED ISSUE #1644: Unimplemented UHADD8 instruction

Requires the UHADD8 instruction.</br>

```
The instruction at 0x0000000000717db4 (opcode 0xe6700f91) is undefined!
```

### Instruction name
```
uhadd8 r0, r0, r1
```

### Required by: 

https://github.com/Ryujinx/Ryujinx-Games-List/issues/2809

piyachetk:
Implemented instruction
- UHADD8

Fixes #1644

I originally implemented this instruction in #2242 but to make it easier for the maintainers to review, I've decided to split it into a separate pull request.

### DIFF
--- a/ARMeilleure/Decoders/OpCodeTable.cs
+++ b/ARMeilleure/Decoders/OpCodeTable.cs
@@ -776,6 +776,7 @@ namespace ARMeilleure.Decoders
             SetA32("<<<<00010001xxxx0000xxxx0xx1xxxx", InstName.Tst,     InstEmit32.Tst,     OpCode32AluRsReg.Create);
             SetA32("<<<<0111111xxxxxxxxxxxxxx101xxxx", InstName.Ubfx,    InstEmit32.Ubfx,    OpCode32AluBf.Create);
             SetA32("<<<<01110011xxxx1111xxxx0001xxxx", InstName.Udiv,    InstEmit32.Udiv,    OpCode32AluMla.Create);
+            SetA32("<<<<01100111xxxxxxxx11111001xxxx", InstName.Uhadd8,  InstEmit32.Uhadd8,  OpCode32AluReg.Create);
             SetA32("<<<<00000100xxxxxxxxxxxx1001xxxx", InstName.Umaal,   InstEmit32.Umaal,   OpCode32AluUmull.Create);
             SetA32("<<<<0000101xxxxxxxxxxxxx1001xxxx", InstName.Umlal,   InstEmit32.Umlal,   OpCode32AluUmull.Create);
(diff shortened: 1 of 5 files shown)


jduncanator:
This made me think of something. We really should blacklist functions with unimplemented instructions in them from the PTC cache. That would let us implement new instructions without needing to bump the PTC cache version.

A task for another day...

gdkchan:
> This made me think of something. We really should blacklist functions with unimplemented instructions in them from the PTC cache. That would let us implement new instructions without needing to bump the PTC cache version.
> 
> A task for another day...

It's not that simple, some games have permanently undefined (which are also "unimplemented") instructions for certain paths were it wants to force the program to crash because something went wrong. Doing so would prevent those functions from being cached. In addition to that, it is possible that "garbage" is JIT compiled as code. Such data would never be executed, but would be still part of the function, and prevent it from being cached if such a method was used.

gdkchan (approved):
lgtm, pretty clever way to calculate it without 4 additions and 4 shifts!

riperiperi (approved):
Nice one.

CeruleanSky:
@gdkchan After seeing how it was originally implemented, in #2242 , I pointed in that PR to how much simpler it was done in [dynramic](https://github.com/merryhime/dynarmic/blob/8931ee346bd73068af477f34bfda65942c7895ed/src/backend_x64/emit_x64_packed.cpp#L270), as that was what I was using in my private build.  That link also has comments about how it works and also has version that calculates it via `~pavg(~a, ~b)`.  This pull request uses the method that [Subv](https://github.com/merryhime/dynarmic/pull/45/files#diff-d53d841d61d9e190f0c99683ef364b790a82eab3b35c36f08b1486a626e3f022R1307) added dynamic originally, whether Subv came up with it on their own, I do not know.  The pavg method and this are both fairly nifty.

gdkchan:
I guess it's worth leaving the explanation here, since it's not exactly obvious what it is doing, and I had to spend a while staring at it to figure it out (maybe it's worth leaving a comment on the code to explain it?)

This is basically how it would be done in a half-adder circuit.
image
That is, for a binary sum, we have the following results:
0 + 0 = 00 (0)
0 + 1 = 01 (1)
1 + 0 = 01 (1)
1 + 1 = 10 (2)
So the least significant bit which is the sum result is obtained with a XOR operation, while the most significant one that is the carry bit can be obtained with the AND operation. Normally the carry would be added to the next bit, but since the sum was shifted right by one, it can just be added directly. And the `0x7F7F7F7F` is just masking out the bit that is supposed to be shifted out of each byte in the register, because otherwise it would be shifted into the lower byte.

Issue #2791, opened 2021-10-28
[REGRESSION] Luigi's Mansion 3 - Shadow Maps not clearing.


INFINITY99K:
Well, as the title says there is a problem with the shadows rendering in the game, the shadows draw in but when moving they will smear on the ground like in the photo below.
Apart from this issue, the FPS is great very stable but, there's a problem with the res. scaling in the game
![Annotation 2021-10-28 021539](https://user-images.githubusercontent.com/73315805/139165482-f88d7d6a-f92c-4bf6-a513-50da439ca5fe.png)
.
![Annotation 2021-10-28 021607](https://user-images.githubusercontent.com/73315805/139165491-c6076798-d8a9-46aa-a5de-1b98f1bf0089.png)

Linear524:
Latest Master branch shows the same shadow bug:
![ryujinx_capture_2021-11-08_19-24-26](https://user-images.githubusercontent.com/84367301/141059408-2e8cd9f9-4d76-4de9-bbbf-3f8e7ccd7ef8.png)
![ryujinx_capture_2021-11-08_19-24-38](https://user-images.githubusercontent.com/84367301/141059426-9b3834dc-df1a-47c9-a4fd-dff71d841f91.png)
![ryujinx_capture_2021-11-08_19-27-10](https://user-images.githubusercontent.com/84367301/141059480-5c24e633-a684-456a-aea0-194acd41f99c.png)

It looks like shadow buffer is not being updated correctly, or not being flushed from previous frame... (P.S. - on my hardware this issue appears after 2 or 3 minutes after game start. But on other hardware this issue occurs a bit faster. So I think it might depend on VRAM size...)
This issue appears after PR #2534 was merged into Master branch. To be more precise - after AMD fix was applied to tessellation fixup in that PR...

INFINITY99K:
I don't think that the problem appeared after PR was merged into master, I believe that it existed in the PR before it was implemented **(Look at the photos)** Weirdly enough it exhibits the behavior on Yuzu on OpenGL.
And here is sth that is even weirder, before the PR was merged into Master, variant 1.0.7077 suffered from this issue I am going to test variants before 1.0.7077 to make sure that this version didn't break things, and don't worry am not using Nvidia's latest drivers that broke older variants of Ryujinx.

**UPDATE**: Good news, well sorta versions before 1.0.7077 don't have that issue so, that's informative I think? maybe both 1.0.7077 and the PR version have sth wrong in common IDK.

[...]

PR #2911, opened 2021-12-06
Fix for texture pool not being updated when it should + buffer texture fixes

gdkchan:
This PR fixes the following issues:
- Outdated descriptors were used in some cases, because the `SequenceNumber` optimization, intended to avoid checking for memory modifications when not necessary made it miss some updates. To fix this, the GPU methods `InvalidateTextureHeaderCacheNoWfi` and `InvalidateSamplerCacheNoWfi` were added, and now increments the `SequenceNumber`. Said methods *should* be called to invalidate the GPU caches for those pools on the Switch, so we can take advantage of that and note that they were probably modified.
- `CommitTextureBindings` and `CommitImageBindings` would try to bind *buffer textures* before setting the buffer storage. This causes an OpenGL error the first time the texture is used, because it has not been initialized at this point (with a call to `glTexBufferRange`). It is also useless, since `CommitBufferTextureBindings` initializes it with the buffer storage *and* binds it (again). This could potentially lead to a minor performance improvement from the reduced amount of GL calls.
- `TextureBuffer.SetStorage` has an optimization that skips calling `glTexBufferRange` if the offset and size of the buffer are the same. But it doesn't check if the buffer itself changed, which could cause it to use an old buffer if an attempt was made to bind a different buffer with the same offset and size (and without any other buffer being created in the meantime, since it also checks the buffers count). In pratice I didn't find any issue caused by this, but decided to fix anyway. Perhaps there is something that makes this scenario impossible right now and I'm missing it.

First point fixes black vertex explosions on Dragon Quest XI S (and maybe other UE4 games?)
Before:
image
image
After:
image

Would be nice to test a few games (I think the GTA trilogy might also be fixed by this?), and also see if they had any performance regression from the first change.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/ThreedClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/ThreedClass.cs
@@ -34,6 +34,8 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 { nameof(ThreedClassState.LaunchDma), new RwCallback(LaunchDma, null) },
                 { nameof(ThreedClassState.LoadInlineData), new RwCallback(LoadInlineData, null) },
                 { nameof(ThreedClassState.SyncpointAction), new RwCallback(IncrementSyncpoint, null) },
+                { nameof(ThreedClassState.InvalidateSamplerCacheNoWfi), new RwCallback(InvalidateSamplerCacheNoWfi, null) },
+                { nameof(ThreedClassState.InvalidateTextureHeaderCacheNoWfi), new RwCallback(InvalidateTextureHeaderCacheNoWfi, null) },
                 { nameof(ThreedClassState.TextureBarrier), new RwCallback(TextureBarrier, null) },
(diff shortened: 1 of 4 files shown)


meisiek:
Fixes flickering in SpongeBob SquarePants: Battle for Bikini Bottom.

Thog (approved):
lgtm

AcK77 (approved):

Linear524:
I thought this PR will fix regression #2791 , but unfortunately it is not 8P There is something else...

LukeWarnut:
Fixes black textures in Balan Wonderworld Demo
image

LukeWarnut:
Fixes constant vertex explosions in Minecraft (Bedrock).

zombie167:
Fix flickering black textures of SnowRunner
s1
s2

riperiperi (approved):
This looks good to me for now. The non-texture issues with sync lead me to believe that something else should be incrementing the sync number, but this will do for now.

zombie167:
> This looks good to me for now. The non-texture issues with sync lead me to believe that something else should be incrementing the sync number, but this will do for now.
Thanks! Looking forward to further improvements or optimizations

PR #2920, opened 2021-12-14
Remove debug configuration and schema

Thog:
The debug configuration can be created by the emulator and is an
artefact of the past.

I also took the liberty of removing the _schema.json file as most of the
time it is left unused/outdated.


Caian (approved):
lgtm

gdkchan (approved):
Nice, one this less to worry about when we need to add more entries to the config. Having to edit like 10 different files for that is very annoying.

Issue #1524, opened 2020-09-01
Undertale: FPS drops to 0.05FPS in the laboratory level


EmulationFanatic:
### What's the issue you encountered?

Undertale (any version): walk into the laboratory and FPS drops to about 1 frame every 10 seconds.

### How can the issue be reproduced?

Walk right to the laboratory and enter it (save file included here):
[undertale.save.zip](https://github.com/Ryujinx/Ryujinx/files/5156006/undertale.save.zip)

### Log file

[Undertale.log](https://github.com/Ryujinx/Ryujinx/files/5156017/Undertale.log)

### Environment?

- Ryujinx version: 1.0.5266</br>
- Game version: 1.09</br>

- System Specs:
  - OS: Windows 10 2004
  - CPU: i5-6600K
  - GPU: NVIDIA GTX 1080
  - RAM: 16GB
- Applied Mods : [ No ]

### Additional context?

Here is a video showing the issue:
https://streamable.com/srw2jk

riperiperi:
There's a screen in this area that is meant to display the region around the player's head. The performance drop might be related.

gdkchan:
Fixed by #2926.

PR #2926, opened 2021-12-18
Add support for releasing a semaphore to DmaClass

riperiperi:
Fixes freezes in OpenGL games, primarily GameMaker ones such as Undertale (when killing undyne, and alphys lab?).

Before this, it would just wait forever after the DMA completes for a value that would never change. In this particular game, it was copying a portion of the render output so that it could be read by CPU.

Found this by looking at register writes right before it froze. I'm not particularly sure what's meant to happen with a "four word" semaphore - it may write a timestamp similar to the counter report method. Right now, it is skipped and logs a warning,

https://gyazo.com/1b9d637de43e03887055b024e4c23760

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -99,6 +99,30 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
         }
 
         /// <summary>
+        /// Releases a semaphore for a given LaunchDma method call.
+        /// </summary>
+        /// <param name="argument">The LaunchDma call argument</param>
+        private void ReleaseSemaphore(int argument)
+        {
+            LaunchDmaSemaphoreType type = (LaunchDmaSemaphoreType)((argument >> 3) & 0x3);
+            if (type != LaunchDmaSemaphoreType.None)
+            {
+                ulong address = ((ulong)_state.State.SetSemaphoreA << 32) | _state.State.SetSemaphoreB;
+                if (type == LaunchDmaSemaphoreType.ReleaseOneWordSemaphore)
+                {
+                    _channel.MemoryManager.Write(address, _state.State.SetSemaphorePayload);
+                }
+                else /* if (type == LaunchDmaSemaphoreType.ReleaseFourWordSemaphore) */
+                {
+                    // TODO: Check if this includes a timestamp or any other data.
+
+                    _channel.MemoryManager.Write(address, (ulong)_state.State.SetSemaphorePayload);
+                    _channel.MemoryManager.Write(address + 8, 0UL);
+                }
+            }
+        }
+
+        /// <summary>
         /// Performs a buffer to buffer, or buffer to texture copy.
         /// </summary>
         /// <param name="argument">Method call argument</param>
@@ -117,6 +141,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
 
             if (size == 0)
             {
+                ReleaseSemaphore(argument);
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -98,6 +98,30 @@ private static bool IsTextureCopyComplete(DmaTexture tex, bool linear, int bpp,
...
+        {
+            LaunchDmaSemaphoreType type = (LaunchDmaSemaphoreType)((argument >> 3) & 0x3);
+            if (type != LaunchDmaSemaphoreType.None)
+            {
+                ulong address = ((ulong)_state.State.SetSemaphoreA << 32) | _state.State.SetSemaphoreB;
+                if (type == LaunchDmaSemaphoreType.ReleaseOneWordSemaphore)
+                {
+                    _channel.MemoryManager.Write(address, _state.State.SetSemaphorePayload);
+                }
+                else /* if (type == LaunchDmaSemaphoreType.ReleaseFourWordSemaphore) */
```
gdkchan, line 115: is anything using this? it might be better to not implement it at all if you're not sure of how it works.

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -117,6 +141,7 @@ private void LaunchDma(int argument)
 
             if (size == 0)
             {
+                ReleaseSemaphore(argument);
```
gdkchan, line 144: I think it would be better to extract all the copy code to a separate method. Then `LaunchDma` could call `DoDmaCopy(argument);` (this is not a name suggestion btw, just to show what I mean) then call `ReleaseSemaphore(argument);`. That would avoid adding 4 calls at different points here, and avoid potentially missing a call if another early return is added to the method.

gdkchan (approved):
Thumbnails now seems correct on Idol Days and likely the other VN with similar issues, the freeze of a few seconds before entering the save menu was also fixed.
image
Freeze of a few seconds when walking around the city on Yokai Watch 1 fixed.
image
Menu seems to load faster on Digimon Cyber Sleuth, but apparently it had some graphical regressions?
image
This is master:
image
Anyway, approving as I beleive this is the right thing to do, the broken textures is a existing issue potentially made worse by this change.
Nice catch.

gdkchan:
Fixes https://github.com/Ryujinx/Ryujinx/issues/1524

Thog (approved):

GamerzHell9137:
> Thumbnails now seems correct on Idol Days and likely the other VN with similar issues, the freeze of a few seconds before entering the save menu was also fixed. image Freeze of a few seconds when walking around the city on Yokai Watch 1 fixed. image Menu seems to load faster on Digimon Cyber Sleuth, but apparently it had some graphical regressions? image This is master: image Anyway, approving as I beleive this is the right thing to do, the broken textures is a existing issue potentially made worse by this change. Nice catch.

Seems like 1.03 update for the game fixes the texts at least on master
image
The PR messes stuff up but does cut the loading from 30 secs to instant.
In game gets even more messed up but just like you said, most likely something else.
image

GamerzHell9137:
There's a chance that Professor Layton and Digimon are doing something similar with their graphics and the graphics in Layton got even more messed up with the PR but the game is now fullspeed compared to master which is halved, same with loading being cut down from 30secs to instant.
image

lostromb:
I was hopeful that this change would fix the menu screen softlock in Xenoblade 2, but alas, it still seems to be there...

gdkchan:
Xenoblade 2 does not use OpenGL on the Switch.

mooock:
not sure if it helps but

Zaccaria Pinball
The Snack World TreJarers Gold

has similar artifacts

PR #2927, opened 2021-12-19
Implement CSDB instruction

### LINKED ISSUE #2918: Unimplemented CSDB instruction
Requires the CSDB instruction.</br>

```
The instruction at 0x00000000065b89d8 (opcode 0xe320f014) is undefined!
```
```
The instruction at 0x000000000d5229d8 (opcode 0xe320f014) is undefined!
```

#### Instruction name
```
csdb
```

### Required by: 
Monster Rancher 1 & 2 DX

gdkchan:
> Consumption of Speculative Data Barrier is a memory barrier that controls speculative execution and data value
prediction.

Implemented as no-op since that's how it should behave on the Switch CPU. This instruction has been added in new-ish Arm processors to mitigate side-channel attacks that takes advantage of the processors speculative execution. The Switch CPU does not actually support, on an older version of the Arm architecture reference manual, this encoding is documented to be a reserved hint instruction that should execute as NOP.
image
image
This is required by Monster Rancher 1 & 2 DX, which now works.
image
image
image
image
Fixes #2918.


Thog (approved):

riperiperi (approved):
LGTM

PR #2932, opened 2021-12-24
Use bindless for samplers >= 32 and indexed samplers

### LINKED ISSUE #2255: Bravely Default 2 black dots in map
Hi, in the previous build those black dots were not there.
Bravely 2 black dots
[Ryujinx_1.0.6852_2021-04-29_20-13-58.log](https://github.com/Ryujinx/Ryujinx/files/6402269/Ryujinx_1.0.6852_2021-04-29_20-13-58.log)

- Ryujinx version: 1.0.6852</br>
- Game version: 1.0</br>
- System Specs:
  - OS: Windows 10
  - CPU: i5 9600K
  - GPU: NVIDIA GTX 1060 6GB
  - RAM: 16GB DDR4
- Applied Mods : NO

### LINKED ISSUE #2487: TLoZ SSHD particles textures not render properly

Particles textures not render properly

ryujinx_capture_2021-07-18_01-42-16
ryujinx_capture_2021-07-18_01-41-01

### Log file

Not provided

### Environment

- Ryujinx version: 1.0.6970</br>
- Game version: 1.0.0</br>
- System Specs:
  - OS: Windows 10
  - CPU: RYZEN 5 3600X
  - GPU: NVIDIA RTX 2070
  - RAM:16GB
  - Applied Mods : No

gdkchan:
### The problem

This change is mainly adding a way to enable bindless textures to be used on the *host*. In order to understand it, we need to first understand the problems it is trying to solve. Right now there are 3 issues that are more or less related to bindless textures.

**Sampler arrays:**

In GLSL, a Sampler array is declared as `uniform sampler2D multipleTextures[16];` (for example, obviously the sampler type and size varies). On NVIDIA (at least on the Switch) this compiles down to a bindless texture access, where the texture handle is loaded from a constant buffer, using a `LDC` instruction, and a offset on the form of `base + index * 8`. While is is possible to detect that the bindless access is in fact accessing what once was a Sampler array, knowing the size of such array is impossible as that information has been lost in the compilation process.

The current solution does attempt to pattern match those indexed accesses using LDC followed by a bindless texture instruction, but since the size is unknown is simply uses a fixed size of 4. This was done for Luigi's Mansion 3, as it was the first game we found using Sampler arrays, and 4 seemed to be enough for what it was doing. This can and *does* cause problems for other games which uses different sizes, such as The Legend of Zelda: Skyward Sword (black particles when cuting plants with the sword, black squares instead of feathers, etc).

**More than 32 samplers on a shader stage:**

This is a problem that usually arises as a result of the current bindless elimination, which tries to turn the bindless texture access into regular access, creating more samplers in the process. There is a limit to the number of samplers a single stage may have, and on most vendors/GPUs, that limit is 32. When we get past this limit, the shader will simply fail to compile.

We do not have a solution for that right now, but it can be fixed by using *host* bindless texturse, as they are not bound by those limits.

**True bindless:**

Sometimes games simply uses bindless textures on the original GLSL shader, rather than being something produced by the compiler. In those cases, they get the texture handle from the API, and can pass them using any means it wishes, like on vertex buffer, storage buffers, uniform buffers. Bindless elimination can't help here since we can't possibly know from where the handle is coming from to read in advance on all those cases, so we *need* to use host bindless textures to solve that.

### Solution

This solution is mostly using the code from previous attempts (such as #2005), but this time it is being used for other purposes aswell (like overcoming the 32 samplers limit), and most of the logic was moved to the backend, mainly because we will likely do something completely different on Vulkan, as only NVIDIA supports bindless textures there (so far?), and also because Vulkan has much larger limits for the amount of samplers you can use on a shader stage (of course, that will depend on the vendor and GPU).

The change extends the backend with a new method, called `SetBindlessTexture`, that allows a bindless texture to be registered by passing a pair of texture + sampler and texture ID + sampler ID. The texture then becomes accessible on the shader using its *guest* handle. A handle translation process is required for this, where it translate the guest handle to a host one. What the `SetBindlessTexture` method on the backend is making the texture resident, and then writing the host handle on the translation table. Since the handle is a 32-bit value, just using a flat table wouldn't be possible as it would require too much memory, instead, this approach uses a multi-level table that allocates buffer memory on demand. This is implemented with 2 buffers, with the first one having offsets into the second buffer, and the second buffer containing the handle.

So how does this address the 3 issues above?

First, we have the Sampler array issue. We do not know the array size, but there is a limit, and fixed constant buffer and region where the NVN writes those handles. So if we make assumptions about the API, we can simply read all the handles there (up to 128), and make the textures available for bindless access. On the shader side, we can simply remove the bindless to indexed pass and let it be handled by the new bindless implementation that will do the handle lookup and host bindless access. This has the advantage of being limited on the number of textures it might possibly load at once, rather than having to load the entire pool.

For the 32 samplers limit, this is sort of caused by bindless textures too but *not quite*. As mentioned before, all known cases of the number of samplers exceeding 32 are caused by bindless elimination. While disabling the bindless elimination, either fully or partially could be considered a solution for this, we again have the same problem what we do not know which texture is being used and need to load the entire pool. In this case, the bindless elimination information is useful and we should keep them. So instead, the approach used here is making the exceeding samplers bindless. This not only works for samplers generated by the bindless elimination, but also for possible regular ones that also happens to exceed that limit. The limit is also configurable on the code, which allows one to for example, lower it and make *all* the texture accesses bindless on the host, for testing purposes. There is also another problem, the bindless translation tables uses buffer textures, so they also consume texture bindings on the shader, and in this case the amount of bindings drops from 32 to 30. The late bindless "promotion" approach also allows this problem to be solved easily, and we can detect that the amount of samplers is > 32 up-front, and then make anything > 30 bindless.

The last issue, "true bindless", is not something this PR aims to solve, however the changes here also allows it to be supported in the future, potentially without any changes to the backend. One just needs to make sure that the textures the shader might access are "registered", and it should just work. The reason I did not also support it with those changes, is because it is a much more complex problem that will need more testing and more thought on how to fix issues related to texture validity, liveness and whatnot. So this just supports the 2 simpler cases, and lays the groundwork to support the 3rd.

### Current problems

The main problem is the same as the previous bindless attempts. I couldn't get it to work in a stable way with the NVIDIA driver. Mario Tennis Aces seems to always crash after a while with those changes, with what seems to be a memory leak?

### What was fixed

- Black corner on the overworld map on Bravely Default II:
image
- Particles on The Legend of Zelda: Skyward Sword:
image

### Affected games

- Bravely Default II
- The Legend of Zelda: Skyward Sword
- Luigi's Mansion 3
- Mario Tennis Aces
- Any game exceeding the 32 samplers limit (mainly ones heavily using bindless, such as UE4 titles).

Note that being affected here doesn't mean it will actually *improve*. It's mainly an indication on what should be tested, just to be sure the games did not regress before merging.

Closes #2255, closes #2487.


gdkchan:
Closing as I plan to do everything on #3001 now.

PR #2925, opened 2021-12-17
Update to LibHac v0.14.3

Thealexbarney:
- Support reading NCAs with sparse partitions (Fixes #2853)
- Allow deleting read-only files and directories in `LocalFileSystem` (Fixes #2841)
- Disable `AesXtsFileSystem` (Fixes #1947)
- The `IFileSystem` interface has been updated for system version 12.0.0. Now it uses the `Fs.Path` type for paths instead of a `Span<byte>`
- LibHac now uses `SharedRef<T>` and `UniqueRef<T>` types which are similar to the `std::shared_ptr` and `std::unique_ptr` types used in Horizon. These help ensure resources are properly closed. FS IPC interfaces were updated to use these types.
- Fix loading NCAs that don't have a data partition

PR #2933, opened 2021-12-24
Fix wrong title language

gdkchan:
Fixes a regression introduced on #2925 that caused titles to display in the wrong language (for example, Korean being displayed when you select British English). The `TitleLanguage` enum was replaced with a `LibHac.Settings.Language` enum, but the two are actually completely different. So I changed it to use the `TitleLanguage` enum on Ryujinx HLE project that contains the correct values.


meisiek:
I can confirm that "British English" is no longer mapped to Korean. Thank you!

Thog (approved):

emmauss (approved):

PR #2937, opened 2021-12-26
CPU - Implement FCVTMS (Vector)

### LINKED ISSUE #2909: Missing CPU Instruction fcvtms

```
opcode 0x4e21b804
```
### Instruction name
```
fcvtms v4.4s, v0.4s
```

### Required by:
XCOM 2 Collection v1.0.1 [0100d0b00fb74000] [64-bit]: https://github.com/Ryujinx/Ryujinx-Games-List/issues/1780

saldabain:
Fixes #2909 

Add missing instruction that prevents some games from booting.

### DIFF
--- a/ARMeilleure/Decoders/OpCodeTable.cs
+++ b/ARMeilleure/Decoders/OpCodeTable.cs
@@ -285,6 +285,7 @@ namespace ARMeilleure.Decoders
             SetA64("0>1011100<100001110010xxxxxxxxxx", InstName.Fcvtau_V,        InstEmit.Fcvtau_V,        OpCodeSimd.Create);
             SetA64("0x0011100x100001011110xxxxxxxxxx", InstName.Fcvtl_V,         InstEmit.Fcvtl_V,         OpCodeSimd.Create);
             SetA64("x00111100x110000000000xxxxxxxxxx", InstName.Fcvtms_Gp,       InstEmit.Fcvtms_Gp,       OpCodeSimdCvt.Create);
+            SetA64("0x0011100x100001101010xxxxxxxxxx", InstName.Fcvtms_V,        InstEmit.Fcvtms_V,        OpCodeSimdCvt.Create);
             SetA64("x00111100x110001000000xxxxxxxxxx", InstName.Fcvtmu_Gp,       InstEmit.Fcvtmu_Gp,       OpCodeSimdCvt.Create);
             SetA64("0x0011100x100001011010xxxxxxxxxx", InstName.Fcvtn_V,         InstEmit.Fcvtn_V,         OpCodeSimd.Create);
             SetA64("010111100x100001101010xxxxxxxxxx", InstName.Fcvtns_S,        InstEmit.Fcvtns_S,        OpCodeSimd.Create);
--- a/ARMeilleure/Decoders/OpCodeTable.cs
+++ b/ARMeilleure/Decoders/OpCodeTable.cs
@@ -285,7 +285,7 @@ namespace ARMeilleure.Decoders
             SetA64("0>1011100<100001110010xxxxxxxxxx", InstName.Fcvtau_V,        InstEmit.Fcvtau_V,        OpCodeSimd.Create);
             SetA64("0x0011100x100001011110xxxxxxxxxx", InstName.Fcvtl_V,         InstEmit.Fcvtl_V,         OpCodeSimd.Create);
             SetA64("x00111100x110000000000xxxxxxxxxx", InstName.Fcvtms_Gp,       InstEmit.Fcvtms_Gp,       OpCodeSimdCvt.Create);
-            SetA64("0x00111000100001101110xxxxxxxxxx", InstName.Fcvtms_V,        InstEmit.Fcvtms_V,        OpCodeSimd.Create);
+            SetA64("0>0011100<100001101110xxxxxxxxxx", InstName.Fcvtms_V,        InstEmit.Fcvtms_V,        OpCodeSimd.Create);
             SetA64("x00111100x110001000000xxxxxxxxxx", InstName.Fcvtmu_Gp,       InstEmit.Fcvtmu_Gp,       OpCodeSimdCvt.Create);
             SetA64("0x0011100x100001011010xxxxxxxxxx", InstName.Fcvtn_V,         InstEmit.Fcvtn_V,         OpCodeSimd.Create);
             SetA64("010111100x100001101010xxxxxxxxxx", InstName.Fcvtns_S,        InstEmit.Fcvtns_S,        OpCodeSimd.Create);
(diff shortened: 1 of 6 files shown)


meisiek:
It's not working for XCOM 2. The game still crashes with the same Exception:
The instruction at 0x00000000090b75e0 (opcode 0x4e21b804) is undefined!

[Ryujinx_1.0.0+565f9eb_2021-12-26_05-53-17.log](https://github.com/Ryujinx/Ryujinx/files/7776563/Ryujinx_1.0.0%2B565f9eb_2021-12-26_05-53-17.log)

saldabain:
Its still a WIP, its not implemented properly yet.

> It's not working for XCOM 2. The game still crashes with the same Exception: The instruction at 0x00000000090b75e0 (opcode 0x4e21b804) is undefined!
> 
> [Ryujinx_1.0.0+565f9eb_2021-12-26_05-53-17.log](https://github.com/Ryujinx/Ryujinx/files/7776563/Ryujinx_1.0.0%2B565f9eb_2021-12-26_05-53-17.log)

meisiek:
Sorry, I thought after "Add FCVTMS_V Implementation to Armeilleure" the crash would at least be different. I'll check again when this PR is finished.

saldabain:
> Sorry, I thought after "Add FCVTMS_V Implementation to Armeilleure" the crash would at least be different. I'll check again when this PR is finished.

Can you please try this new build? The tests appear to pass

meisiek:
No, same error as before. After that I deleted the shader cache and PPTC cache, but that made no difference.
[Ryujinx_1.0.0+a8581b4_2021-12-26_07-34-28.log](https://github.com/Ryujinx/Ryujinx/files/7776629/Ryujinx_1.0.0%2Ba8581b4_2021-12-26_07-34-28.log)

saldabain:
> No, same error as before. After that I deleted the shader cache and PPTC cache, but that made no difference. [Ryujinx_1.0.0+a8581b4_2021-12-26_07-34-28.log](https://github.com/Ryujinx/Ryujinx/files/7776629/Ryujinx_1.0.0%2Ba8581b4_2021-12-26_07-34-28.log)

Your log file looks like you used the wrong build version -- please use new build that is mentioned in the pr comments. (v 314102c)

meisiek:
You are of course right. Seems I forgot to refresh this page and downloaded not the newest version.
Now the exception is gone, meaning the new opcode gets processed. Good Job!

I started the game 3 times. The first 2 got random crashes, but with the last one I now get into the mission!
grafik

saldabain:
@gdkchan  considering there were random crashes, could you verify that the implementation is correct?

ARMeilleure/Instructions/InstEmitSimdCvt.cs
```diff
@@ -217,6 +217,18 @@ public static void Fcvtms_Gp(ArmEmitterContext context)
...
 
+        public static void Fcvtms_V(ArmEmitterContext context)
+        {
+            if (Optimizations.UseSse41)
+            {
+                EmitSse41FcvtsOpF(context, FPRoundingMode.TowardsMinusInfinity, scalar: false);
+            }
+            else
+            {
+                EmitFcvt(context, (op1) => EmitRoundMathCall(context, MidpointRounding.ToNegativeInfinity, op1), signed: true, scalar: false);
```
gdkchan, line 228: This should call `Math.Floor` not `Math.Round`.

saldabain:
> 

Once the new build is made, please try again -- there was a slight issue with the implementation that may have caused those crashes, let me know if you experience them again :)

ARMeilleure/Instructions/InstEmitSimdCvt.cs
```diff
@@ -225,7 +225,7 @@ public static void Fcvtms_V(ArmEmitterContext context)
             }
             else
             {
-                EmitFcvt(context, (op1) => EmitRoundMathCall(context, MidpointRounding.ToNegativeInfinity, op1), signed: true, scalar: false);
+                EmitFcvt(context, (op1) => EmitFloorMathCall(context, MidpointRounding.ToNegativeInfinity, op1), signed: true, scalar: false);
```
gdkchan, line 228: This is still wrong. The `Math.Floor` method does not take a `MindpointRounding` parameter. See the documentation: https://docs.microsoft.com/en-us/dotnet/api/system.math.floor?view=net-6.0#System_Math_Floor_System_Double_
See also in remarks:
> The behavior of this method follows IEEE Standard 754, section 4. This kind of rounding is sometimes called rounding toward negative infinity. In other words, if d is positive, any fractional component is truncated. If d is negative, the presence of any fractional component causes it to be rounded to the smaller integer. The operation of this method differs from the Ceiling method, which supports rounding toward positive infinity.

The `floor` operation itself *is* rounding towards negative infinity.
This should just be using `EmitUnaryMathCall`. Just see `Fcvtms_Gp`. This method should do the same thing, but writing the result to a vector register instead of a GPR, and also doing a SIMD rather than scalar operation.

gdkchan:
> > 
> 
> Once the new build is made, please try again -- there was a slight issue with the implementation that may have caused those crashes, let me know if you experience them again :)

I don't think so since the change was on the slow path, only taken if SSE 4.1 is not supported, and it's extremely unlikely that this user's CPU does not support it. Plus afaik the crash happens early in the game, way before it even hits this instruction, but I might be wrong, I don't have the game and the user did not provide enough information about it.

ARMeilleure/Instructions/InstEmitSimdCvt.cs
saldabain: Thank you for your guidance, I've addressed as requested

meisiek:
I've tested version 7c14658 and it looks good for me. I've started the game three times and it worked every time. No random crashes.
The only thing is that XCOM 2 seems to be one of this games that can't be closed. Ryujinx hangs every time when I try to close the game. But that has nothing to do with this PR.

ARMeilleure/Decoders/OpCodeTable.cs
```diff
@@ -285,6 +285,7 @@ static OpCodeTable()
             SetA64("0>1011100<100001110010xxxxxxxxxx", InstName.Fcvtau_V,        InstEmit.Fcvtau_V,        OpCodeSimd.Create);
             SetA64("0x0011100x100001011110xxxxxxxxxx", InstName.Fcvtl_V,         InstEmit.Fcvtl_V,         OpCodeSimd.Create);
             SetA64("x00111100x110000000000xxxxxxxxxx", InstName.Fcvtms_Gp,       InstEmit.Fcvtms_Gp,       OpCodeSimdCvt.Create);
+            SetA64("0x00111000100001101110xxxxxxxxxx", InstName.Fcvtms_V,        InstEmit.Fcvtms_V,        OpCodeSimd.Create);
```
gdkchan, line 288:
```suggestion
            SetA64("0>0011100<100001101110xxxxxxxxxx", InstName.Fcvtms_V,        InstEmit.Fcvtms_V,        OpCodeSimd.Create);
```
The mask had the `sz` bit as 0 before, which means it wouldn't match encoding with it set to 1 (used for the .2D double variant of the instruction). Plus as per the manual:
> The encoding sz = 1, Q = 0 is reserved.

This combination is reserved, so the `>` and `<`  ensures that the 01 combination is invalid.
Please also update the tests to include hte `.2D` variant.

saldabain:
Tests have been updated as well as opcode table as suggested

meisiek:
Tried version acfcdbc and had a crash:
[Ryujinx_1.0.0+acfcdbc_2021-12-28_18-27-21.log](https://github.com/Ryujinx/Ryujinx/files/7785791/Ryujinx_1.0.0%2Bacfcdbc_2021-12-28_18-27-21.log)

A short random video plays after the part that you only reach when the new opcode is implemented. I suppose the crash has nothing to do with the opcode implementation but something else is responsible for it.
The next try didn't crash and showed a different video.

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):

PR #2940, opened 2021-12-26
Fix bug causing an audio buffer to be enqueued more than once

gdkchan:
Fixes what I assume to be an oversight that caused buffers to be enqueued (and played) more than once, if the game appended them fast enough.
Fixes audio in the Youtube app. Might potentially fix games with broken audio too.


AcK77 (approved):

riperiperi (approved):
LGTM

Thog (approved):

PR #2941, opened 2021-12-26
Flip scissor box when the YNegate bit is set

gdkchan:
It should be flipped when this bit is set, required by OpenGL games.
Fixes menus being cut off in the Youtube app.
Before:
image
After:
image

Testing to ensure there's no regressions on games is welcome.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -35,6 +35,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
 
         private bool _prevDrawIndexed;
         private bool _prevTfEnable;
+        private bool _prevYNegate;
 
         /// <summary>
         /// Creates a new instance of the state updater.
@@ -426,6 +427,13 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                     int width = scissor.X2 - x;
                     int height = scissor.Y2 - y;
 
+                    if (_state.State.YControl.HasFlag(YControl.NegateY))
+                    {
+                        ref var transform = ref _state.State.ViewportTransform[index];
+                        int vpHeight = (int)MathF.Abs(transform.ScaleY * 2);
+                        y = vpHeight - height - y;
+                    }
+
                     float scale = _channel.TextureManager.RenderTargetScale;
                     if (scale != 1f)
                     {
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Nice catch, wonder if this was affecting anything else.

GamerzHell9137:
Snip snip 
image

LukeWarnut:
Fixes the in-game UI in Bloons TD 5.

Before:
image
After:
image

Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -426,6 +427,13 @@ private void UpdateScissorState()
                     int width = scissor.X2 - x;
                     int height = scissor.Y2 - y;
 
+                    if (_state.State.YControl.HasFlag(YControl.NegateY))
+                    {
+                        ref var transform = ref _state.State.ViewportTransform[index];
+                        int vpHeight = (int)MathF.Abs(transform.ScaleY * 2);
+                        y = vpHeight - height - y;
```
riperiperi, line 434: This makes it possible to have a negative Y for scissor, which I think is invalid in Vulkan (not opengl?). Might want to add Y to height in this case, and then set Y to 0.

riperiperi (commented):
I checked layton and it looks like it's rendering to a 4096x4096 texture, but the viewport does match that (1920x1080 clear, 2880x1620 draws). The 1080p clear is the problem, the viewport height is not set yet at that point... but maybe using the viewport height isn't correct at all. Maybe the scissor region should instead be flipped by the `ScreenScissorState` height? (min size of all render targets)

Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
gdkchan: I thought about that yesterday, but the OpenGL spec doesn't seems to mention anything about a negative Y being invalid, it only says that a negative width or height is invalid. I haven't checked Vulkan, but anyway I will fix it.

gdkchan:
> I checked layton and it looks like it's rendering to a 4096x4096 texture, but the viewport does match that (1920x1080 clear, 2880x1620 draws). The 1080p clear is the problem, the viewport height is not set yet at that point... but maybe using the viewport height isn't correct at all. Maybe the scissor region should instead be flipped by the `ScreenScissorState` height? (min size of all render targets)

Thanks, I have updated it to use screen scissor state, and to account for Y being negative. I don't have Professor Layton (at least, not this one) to test here, so someone else will need to test that.

GamerzHell9137:
The new changes fixed it 
image

riperiperi (approved):
LGTM.

AcK77 (approved):

PR #2942, opened 2021-12-26
Fix DMA copy fast path line size when xCount < stride

gdkchan:
The copy could be out of bounds before if `xCount` was less than stride, as the span size is calculated based on the stride and `xCount` (so it could be missing a few bytes at the end). This change makes it not copy any more than `xCount`. I'm not sure if the fast path should be used in this case though, as technically the data of the few missing pixels would be just 0.

This fixes some random crashes in the Youtube app and maybe some other games that does DMA copy with linear source textures.

Testing to ensure that it did not regress any game is welcome.


riperiperi (approved):
Makes sense.

Thog (approved):

PR #2947, opened 2021-12-27
Implement -p or --profile command line argument

JavidPack:
This PR implements a command line argument for specifying which profile to load, overriding the default behavior of loading the most recently used profile.

This functionality is useful for those who like to, for example, create individualize desktop shortcuts on a shared computer. Each profile user can launch ryujinx directly to their profile without having to visit the manage user profiles window.

Here is an example. The user would add `-p MyUserName` to the target line of the desktop shortcut. They would then do the same but with a different profile name with another shortcut for a different user.
explorer_2021-12-26_23-56-54

I tried to follow existing examples of intended error levels and code styles, but I'm happy to change anything if needed. The somewhat duplicate code parsing the command line arguments was needed since the existing code seems to expect any parameters not captured to be a launch path.

I also feel that the <https://github.com/Ryujinx/Ryujinx/wiki/Ryujinx-Setup-&-Configuration-Guide> page needs a section outlining all existing command line arguments, it took me quite a while to determine from the wiki the available command line arguments. I would like to see if a parameter for launching a game directly would also be possible after this gets merged.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
+++ b/Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
@@ -43,7 +43,28 @@ namespace Ryujinx.HLE.HOS.Services.Account.Acc
             }
             else
             {
-                OpenUser(_accountSaveDataManager.LastOpened);
+                UserId commandLineUserProfileOverride = default; 
+                var args = Environment.GetCommandLineArgs();
+                for (int i = 0; i < args.Length; ++i)
+                {
+                    string arg = args[i];
+
+                    if (arg == "-p" || arg == "--profile")
+                    {
+                        if (i + 1 >= args.Length)
+                        {
+                            Common.Logging.Logger.Error?.Print(Common.Logging.LogClass.Application, $"Invalid option '{arg}'");
+
+                            continue;
+                        }
+
+                        string profileName = args[++i];
+                        commandLineUserProfileOverride = _profiles.FirstOrDefault(x => x.Value.Name == profileName).Value?.UserId ?? default;
+                        if(commandLineUserProfileOverride.IsNull)
+                            Common.Logging.Logger.Warning?.Print(Common.Logging.LogClass.Application, $"The command line specified profile named '{profileName}' was not found");
+                    }
+                }
+                OpenUser(commandLineUserProfileOverride.IsNull ? _accountSaveDataManager.LastOpened : commandLineUserProfileOverride);
             }
         }
 
--- a/Ryujinx/Program.cs
+++ b/Ryujinx/Program.cs
@@ -51,6 +51,16 @@ namespace Ryujinx
 
                     baseDirPathArg = args[++i];
                 }
+                else if (arg == "-p" || arg == "--profile")
+                {
+                    if (i + 1 >= args.Length)
+                    {
+                        Logger.Error?.Print(LogClass.Application, $"Invalid option '{arg}'");
+
+                        continue;
+                    }
+                    i++;
+                }
                 else if (arg == "-f" || arg == "--fullscreen")
                 {
                     startFullscreenArg = true;


Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -43,7 +43,28 @@ public AccountManager(HorizonClient horizonClient)
             }
             else
             {
-                OpenUser(_accountSaveDataManager.LastOpened);
+                UserId commandLineUserProfileOverride = default;
```
gdkchan, line 46: I don't think this is the right place to do this.
The command line arguments should be parsed in the same place as the other ones (`Program.cs` on Ryujinx project), and then it should eventually call `OpenUser` and pass the user ID that was specified with the argument (or user name, in this case it will need to look up the ID from the name).

You'll need to store the IDs somewhere to use when the `AccountManager` instance is created.

gdkchan:
> I would like to see if a parameter for launching a game directly would also be possible after this gets merged.

This should be already possible by passing the ROM path to the emulator.

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
JavidPack: Ok, I wasn't positive how everything was laid out, but I think I implemented this

JavidPack:
Recent commits should address the previous feedback.

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -25,7 +25,7 @@ public class AccountManager
 
         public UserProfile LastOpenedUser { get; private set; }
 
-        public AccountManager(HorizonClient horizonClient)
+        public AccountManager(HorizonClient horizonClient, string CommandLineProfile = null)
```
gdkchan, line 28: Should be called `commandLineProfile` to follow our naming conventions.

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -43,7 +43,14 @@ public AccountManager(HorizonClient horizonClient)
             }
             else
             {
-                OpenUser(_accountSaveDataManager.LastOpened);
+                UserId commandLineUserProfileOverride = default; 
+                if(!string.IsNullOrEmpty(CommandLineProfile))
```
gdkchan, line 47:
```suggestion
                if (!string.IsNullOrEmpty(CommandLineProfile))
```

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -43,7 +43,14 @@ public AccountManager(HorizonClient horizonClient)
...
+                { 
+                    commandLineUserProfileOverride = _profiles.FirstOrDefault(x => x.Value.Name == CommandLineProfile).Value?.UserId ?? default;
+                    if(commandLineUserProfileOverride.IsNull)
```
gdkchan, line 50:
```suggestion
                    if (commandLineUserProfileOverride.IsNull)
```

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -43,7 +43,14 @@ public AccountManager(HorizonClient horizonClient)
...
+                        Common.Logging.Logger.Warning?.Print(Common.Logging.LogClass.Application, $"The command line specified profile named '{CommandLineProfile}' was not found");
```
gdkchan, line 51: Add `using Ryujinx.Common.Logging;` and remove `Common.Logging.`.

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
```diff
@@ -43,7 +43,14 @@ public AccountManager(HorizonClient horizonClient)
             }
             else
             {
-                OpenUser(_accountSaveDataManager.LastOpened);
+                UserId commandLineUserProfileOverride = default; 
+                if(!string.IsNullOrEmpty(CommandLineProfile))
+                { 
+                    commandLineUserProfileOverride = _profiles.FirstOrDefault(x => x.Value.Name == CommandLineProfile).Value?.UserId ?? default;
```
gdkchan, line 49: Can't you do something like:
```cs
_profiles.Values.FirstOrDefault(x => x.Name == commandLineProfile)?.UserId ?? default;
```

Ryujinx.HLE/HOS/Services/Account/Acc/AccountManager.cs
gdkchan: I think something like `initialProfileName` is a better name for this argument.

JavidPack:
The latest commit should resolve the recent review feedback.

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
Nice.

Thog (approved):

PR #2955, opened 2021-12-30
Fix A1B5G5R5 texture format

gdkchan:
The format had the wrong order before, resulting in graphical glitches in at least one title (NSO Nintendo 64 emulator).
This emulator does not work on master due to other components that are not implemented yet, so I have tested it on a private branch, and I'm providing those screenshots only to illustrate what was fixed.

Before:
image
After:
image


riperiperi (approved):
Nice catch.

AcK77 (approved):

Thog (approved):

PR #2956, opened 2021-12-30
Add support for the R4G4 texture format

gdkchan:
This format is used by the NSO N64 emulator. It is supported by Vulkan, but not OpenGL, so I added a compatibility flag to report when its not supported, in which case we do conversion to the R4G4B4A4 on the CPU. The reason I choose this format, is because conversion can be done by simply zero-extending the value from 8-bits to 16-bits.

The N64 emulator does not work on master due to other components that are not implemented yet, so I have tested it on a private branch, and I'm providing those screenshots only to illustrate what was fixed.

Before:
image
(Actually, it would look slightly different on master, as it defaults for RGBA8 format when a format is not implemented, on the above screenshot it simply failed to create the texture).
After:
image

### DIFF
--- a/Ryujinx.Graphics.GAL/Capabilities.cs
+++ b/Ryujinx.Graphics.GAL/Capabilities.cs
@@ -6,6 +6,7 @@ namespace Ryujinx.Graphics.GAL
         public bool HasVectorIndexingBug { get; }
 
         public bool SupportsAstcCompression { get; }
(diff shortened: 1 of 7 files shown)


AcK77 (approved):

Thog (approved):

riperiperi (approved):
LGTM.

PR #2957, opened 2021-12-30
Force crop when presentation cached texture size mismatches

gdkchan:
Right now, the presentation texture size is used to find a matching texture on the cache, but after that is not used anymore, instead it uses the cached texture size. The problem is that due to size alignment, the cached texture might be actually larger than the texture size, leading to gaps when the texture is presented.

This has been fixed here by forcing the extra size to be cropped before presentation, using the crop rectangle that was already being passed from surface flinger.

Before:
image
After:
image

Should fix a similar issue on Hades aswell, but I did not test this one.

### DIFF
--- a/Ryujinx.Graphics.GAL/ImageCrop.cs
+++ b/Ryujinx.Graphics.GAL/ImageCrop.cs
@@ -21,8 +21,7 @@ namespace Ryujinx.Graphics.GAL
             bool  flipY,
             bool  isStretched,
             float aspectRatioX,
-            float aspectRatioY
-            )
+            float aspectRatioY)
         {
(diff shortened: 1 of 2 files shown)


meisiek:
I can confirm that this fixes Hades as well:
grafik

AcK77 (approved):

riperiperi (commented):
This might fix squishing in Super Mario Sunshine as well. Will test later.

riperiperi:
This does fix Super Mario Sunshine.

Before:
image

After:
image

riperiperi (approved):
LGTM

PR #2451, opened 2021-07-08
Implementations of SVCs related to self modifying code

shadowninja108:
Various homebrew (and maybe SM3DAS's SM64 JIT plugin?) abuse these SVCs as part of their JIT implementations. This PR alone doesn't allow for proper JIT, but it's a stepping stone towards it.

(Don't merge until tested thoroughly)

PR #2958, opened 2021-12-30
Implement code memory syscalls

gdkchan:
This implements code memory related syscalls, used by applications that generates and/or modifies code at runtime.
It is required by, for example, emulators that uses a JIT (NSO N64) and mods that patches for function hooking game code on-the-fly (skyline/ARCropolis). Neither will work with those changes alone, but its a step towards it.

Depends on #2954 because code modification requires mapping a RW view of the RX code region.


MutantAura:
supercedes #2451 ?

gdkchan:
> supercedes #2451 ?

Yes.

xerpi:
Tested on Linux x64 and using a 32-bit Switch homebrew.

gdkchan:
This has been rebased and is ready for review. It has already been largely tested with homebrews, mods and the official N64 emulator.

Here's a screenshot of PPSSPP running on it (also requires being launched from the hbmenu/homebrew launcher):
image

riperiperi (approved):
LGTM, glad to see this finally ready for merge.

MelonDS (fast memory disabled) and Citra are working, though due to some GPU issues (regressions?) they seem to have some visual issues right now. Here's some older screenshots from back in January:

image

image

Thog (approved):

riperiperi:
Found out what was wrong with the graphics in those emulators. So here's some up to date screenshots.

- MelonDS: Needs #2237
- Citra: I was using an older version of the branch...

image

image

PR #2965, opened 2022-01-02
Enable CPU JIT cache invalidation

gdkchan:
This PR extends the existing `InvalidateJitCacheRegion` to actually *remove* functions overlapping a given range from the cache. One important thing to note is that the function is not actually removed from memory, instead it is just removed from the tables making it inaccessible. Not removing it from memory is to prevent crashes if it happens to be still being executed when the invalidate method is called. This is something that should be improved in the future.

Since the invalidation requires getting all functions that overlaps the invalidated range, I had to use a different data structure since the `ConcurrentDictionary` is not suited for this. Now it uses a `IntervalTree` that is based on the implementation used to find texture overlaps, with a few differences:
- The texture overlaps one had to support multiple values with the same start address. This doesn't, so I removed the values list and made it a single value.
- The `BSTInsert` and `Delete` methods also had to be adjusted to work with a single value. Now if the value already exists, `BSTInsert` either does nothing or updates the value. `Delete` is simpler and always removes the node (as there's a single value).
- `Get` was changed to return the overlapping keys rather than values, as most of the JIT function removal methods takes an address rather than the item itself. (Maybe it should return a `KeyValuePair`?)
- The `Get` overload that only takes start address was changed to `TryGet`, and now returns a single value and a bool indicating if it exists.

This change will be required to make applications that loads code dynamically (NROs) or has code self modificatoin (like the skyline/ARCropolis mod that calls then stubs functions, plus patches functions for hooking) function properly.

Games that I recommend testing with this change (just to ensure there are no regressions, I don't expect improvements on them).
- Games that use NROs (Super Smash Bros Ultimate).
- Games that map/unmap memory heavily (such as UE4 games in general, just to make sure performance/stuttering is the same).

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSystem.cs
+++ b/ARMeilleure/Instructions/InstEmitSystem.cs
@@ -12,7 +12,8 @@ namespace ARMeilleure.Instructions
 {
     static partial class InstEmit
     {
-        private const int DczSizeLog2 = 4;
+        private const int DczSizeLog2 = 4; // Log2 size in words
+        public const int DczSizeInBytes = 4 << DczSizeLog2;
 
         public static void Hint(ArmEmitterContext context)
         {
@@ -87,7 +88,7 @@ namespace ARMeilleure.Instructions
                     // DC ZVA
                     Operand t = GetIntOrZR(context, op.Rt);
 
-                    for (long offset = 0; offset < (4 << DczSizeLog2); offset += 8)
+                    for (long offset = 0; offset < DczSizeInBytes; offset += 8)
                     {
                         Operand address = context.Add(t, Const(offset));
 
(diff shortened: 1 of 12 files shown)


riperiperi:
Would be nice to get some performance tests on SSBU for this, specifically loading time, stutters when re-entering the menu, etc.

MutantAura:
Can this be rebased for those tests? I'd like to perform the tests with a shader cache etc and forgive me for not wanting to spend 30 mins each time going back and forth :)

MutantAura:
Tested PR and master for about an hour each in SSBU and both felt near enough the same. 
There was the usual menu, css and match intro stutter present but this was on both master and PR. No additional match stutter that I could notice. Master compared to was 1.1.12.

Only thing I did notice was after about 45 mins into my PR playtest the FIFO spiked to the mid-high 70-80s (normally much lower) and disabling vsync resulted in a max of 75fps with FIFO reaching 99. Again normally fps reaches nearer the 140's when unlocked and did so at the start of the play session. I could not reproduce this however so perhaps it was unrelated.

riperiperi (approved):
LGTM.

AcK77 (approved):

PR #2971, opened 2022-01-03
Texture Sync, incompatible overlap handling, data flush improvements.

riperiperi:
This PR aims to solve a bunch of issues caused by texture modification and data flushing, by primarily handling data flush on a per-handle instead of per-view basis, and synchronizing flushes with syncpoint increments.

This introduces a new backend method, so it's not currently compatible with Vulkan (though the benefits are more important there)

# Part 1: Syncing Texture Flush

This one has been a while coming. When texture flush via memory tracking was first added, our CPU and various other components were considerably slower. The GPU was so tight with the CPU as a result of both this and the lack of backend multithreading, so when the game tried to access the data it was almost certain that the draw would have completed and the data would be there, just by chance.

This did not last. Over time people started encountering various issues where textures would flush _before_ their data was fully ready, causing white water in BOTW and rainbow lighting in Splatoon. This is compounded a lot by the addition of Background Multithreading and Vulkan, so much that people thought there was a regression. There wasn't, the texture flush was just behaving as it always did.

The new flushing mechanism is similar to the method we use for buffers, which has proven speedy and always gets the data in place by the time the CPU wants it. There is one key difference - texture data is only synced at syncpoint increment, rather than every WaitForIdle. This is mostly a limitation of buffer flush that can be fixed later, but it means that adding these syncpoints for texture flush should not cost anything at all comparatively.

But things are not as simple as shoving the buffer syncaction logic onto textures, as there is another hurdle that could affect games that write into complex texture view layouts.

## Modified Flags move to the Texture Group
Texture groups are a useful abstraction for keeping track of each subimage's memory tracking within a storage texture, and is shared across all views of a texture. Before now, this was _not used_ for texture flush, and each texture view could have its own modified flag and flush action... these views could also overlap. While this has the benefit of fetching the data for modified views in one chunk, it doesn't bode well for tracking rules that use the gpu modified state of each subimage, as they are decoupled from the texture group and might flush at the wrong time.

Now, the modified flag and flush action have been moved to the texture group, and are tracked for each handle individually. Each handle can also have its own sync numbers for determining which sync object to wait on, or if it should even wait at all (more on this later)

## Flushing by slice, rather than by view
Since modified flags are now owned by texture group handles, rather than texture views, they must be able to flush at a granularity smaller than a view. I've added methods on the backend to allow this, though cases which would flush the whole storage texture will still try to do that. These methods will also use the OGL 4.5 methods for flushing texture sub data, rather than flushing an entire level (all layers) and picking out of that. This has potential to be a little faster or slower depending on the game. Hopefully no drivers complain about this.

## Flushing data that isn't ready
When a game flushes texture data, it's a reasonable assumption that the guest must have waited on some synchronization so that it could be sure that the data was actually there. If not, then they're leaving things up to chance, and they've essentially caused a race. This is undefined behaviour, so if texture data is accessed before its syncpoint is incremented, we flush the data regardless of whether it is ready or not (but wait on the _last_ syncpoint that the guest did acknowledge, if possible).

This is the best of both worlds - we ensure data that's appropriately waited on becomes available, don't waste time waiting on accidental flushes that happen after an unrelated gpu write or other textures that share pages.

When one of these undefined flushes happens, the Modified flag is not cleared. The tracking action is re-added when the sync for the texture _is_ actually reached, in case the game wants to access the up to date data later on. CPU writes during this state are considered to be racing with the GPU, so this is also treated specially.

When the modified flag remains set on a texture handle, data is ignored from the CPU. One possible case is that data is written to one mip level of a texture, but another mip level of the same texture remains untouched and contains data modified by the GPU (and is not flushed due to sync rules). Both are on the same page, so both recieve the same dirty flag from write tracking... but one contains data that is unflushed or flushed out of date. It would be invalid to overwrite that unflushed data from CPU if it truly was not touched, so tracking dirty flags are consumed and ignored in this state. Given a race, we therefore say that the GPU _always wins_.

Finally, it's important to note that accesses from the GPU thread do not follow these rules at all. The GPU's own engines and methods must have the texture data at the latest point in time. Thankfully, flushing from the main GPU thread in our backends guarantees this, so sync is skipped and the data is accessed "immediately". Buffers, conversely, must wait right now.

# Part 2: Flush ordering for incompatible data
As of right now, Ryujinx follows a core assumption with the texture cache, where a use of any given texture should be valid in its own layout+format. If this is the case, then any previously cached textures overlapping the same memory that do _not_ share the layout and format are tharefore invalid - they will contain garbage data. This core assumption keeps important texture data alive while saving time by not flushing or loading data that would be "garbage" in the new format/layout.

This rule wasn't fully established before. Incompatible textures only have the _potential_ to be deleted when a new overlap appears over them, and any checks only happened when the texture was created. If both textures existed in the cache at the same point, they would be disjoint, and could flush separately... in any order. This meant that old texture data could flush over new texture data - not ideal for games with complex streaming behaviour.

This is where the new incompatible overlap rules come into play. To put it simply, if a texture is written to and one or more other incompatible textures overlap its memory, their data is considered _invalid_ and their modified flags are immediately cleared. This means that only one can live at a time, and therefore when data flushes, it will always be the latest. Problem solved! Note that these rules were already in place before, they were just enforced _on creation_, rather than on each use.

This is a large sweeping change that affects every game, so testing would be appreciated to see if everything is still holding up.

# Part 3: Flushing host incompatible formats

Occasionally, a texture format or relationship is used which isn't fully supported on the host. Two such examples are:

- ASTC compressed textures, which are not supported on desktop GPUs. 
- BCn compressed 3D textures, which are not supported by OpenGL (but can be supported by Vulkan)

Ryujinx supports these formats by decompressing them to a supported, uncompressed format on the CPU. But, this means that data cannot be copied to/from these textures on the gpu due to using the replacement format, so they are ineligible for texture copy dependencies and views. This causes problems when certain games draw or copy data to these formats by aliasing them as some other data format:

- Life is Strange: True Colors, and potentially other UE4 games (dungeon defenders?) use ASTC textures for characters and environments. UE4 moves texture data around by aliasing levels of textures as a data format such as R32G32Float, and copying them with a blit. Before, this moved data would be lost, as there was no relationship between the copy target and the ASTC texture.
- BOTW draws into a compressed 3D BCn texture to use for the blue dissolve teleportation animation. The texture contains some perlin noise generated and compressed by the GPU. Before, link would just disappear immediately as ryu could not move the data for this texture.

The changes to add incompatible overlaps allowed me to add an additional case for these textures - the ability to force overlaps of these host incompatible format textures to flush their data back to the CPU before it can be used. This is not the fastest, but it is fully compatible and allows us to cover cases which were completely broken before.

This path is essentially the "slow path" for this kind of relationship, and will allow us to support rendered compressed textures on platforms that don't support BCn like mobile hardware. In time, we can introduce compute shader paths that will allow copy dependencies between data textures and "emulated" compressed textures to be done without any data flushing at all, which will further improve performance.

# Potential issues with buffer/texture flush as it stands
One tracking structure is shared between all users of the same address space - they all see and signal the same tracked handles. However, there might be times where we want to sync on GPU, but the sync should not be visible on CPU yet. On top of this, a sync from CPU could remove tracking without sync flushing, but the GPU might try to use that handle immediately after for something like a UBO/I2M write and expect to get the latest data.

This last case can happen with buffers and textures. Not that it couldn't happen before, but it's important to bring up as potential future work. I have not seen it happen in practice.

# Fixes

**1:**
- Weird issues with BOTW regarding link sinking into the ground and water colour (potentially other minor effects)
  - Before: https://cdn.discordapp.com/attachments/410208534861447170/918516347779026984/unknown.png
  - After: https://cdn.discordapp.com/attachments/728011297857339422/919229695780855848/unknown.png
- Splatoon 2 world lighting being a very solid colour or rainbow after a loading screen.
  - Before: https://cdn.discordapp.com/attachments/728011297857339422/916816387673034752/unknown.png
  - After: https://cdn.discordapp.com/attachments/728011297857339422/919231358981800056/unknown.png
- Saved MK8D replay thumbnails occasionally being from the previous race or black.
- More "random chance" visual issues?

**2:**
- Texture streaming issues in UE3/UE4 games (mostly)
  - There are still some small issues in A Hat in Time, but they are greatly reduced. These are being investigated.

**3:**
- Blue dissolve teleportation animation in BOTW. https://gyazo.com/d57a3827ce137f21b852596fc6101447
- UE4 games that stream ASTC format textures, such as Life is Strange: True Colors. (not perfect, same as UE3/4 in general, but vastly better)
  - Before: https://user-images.githubusercontent.com/2002038/146266301-10db3ce5-9e50-44bc-8bfc-99fd43ef3ccf.PNG
  - After: https://cdn.discordapp.com/attachments/787101806085406724/923630179966918706/unknown.png
 
I also fixed an issue with partial 3D texture DMA/load/flush that caused the gob size in z to be incorrect. Not sure what this fixes, but it might improve some issues with lighting in UE4 games a little.

Some things have been fixed regarding the DmaClass fast path used for videos. It won't try to copy to compressed textures directly now, and is aware of its 1 layer limitation.

Testing would be greately appreciated, to ensure that nothing is broken and that performance is similar.

Specifically, AMD/Intel Windows and Linux should be tested to make sure they don't have any weird issues with `glGetTextureSubImage`, as it's technially a DSA method and these drivers are pretty terrible at that.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -212,7 +212,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                         yCount,
                         dstLinear);
 
-                    if (target != null && !target.Group.HasIncompatibleOverlaps)
+                    if (target != null)
                     {
                         ReadOnlySpan<byte> data;
                         if (srcLinear)
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -212,7 +212,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                         yCount,
                         dstLinear);
 
-                    if (target != null)
+                    if (target != null && !target.Group.HasIncompatibleOverlaps)
                     {
                         ReadOnlySpan<byte> data;
                         if (srcLinear)
@@ -245,6 +245,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                                 srcSpan);
                         }
 
+                        target.SynchronizeMemory();
                         target.SetData(data);
                         target.SignalModified();
 
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -234,7 +234,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                                 src.Height,
                                 src.Depth,
                                 1,
-                                target.Info.Levels,
+                                1,
                                 1,
                                 1,
                                 1,
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -232,6 +232,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                             data = LayoutConverter.ConvertBlockLinearToLinear(
                                 src.Width,
                                 src.Height,
+                                src.Depth,
                                 1,
                                 target.Info.Levels,
                                 1,
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -750,15 +750,17 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// Converts texture data to a format and layout that is supported by the host GPU.
         /// </summary>
         /// <param name="data">Data to be converted</param>
+        /// <param name="level">Mip level to convert</param>
+        /// <param name="single">True to convert a single slice</param>
         /// <returns>Converted data</returns>
         public ReadOnlySpan<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data, int level = 0, bool single = false)
         {
             int width = Info.Width;
             int height = Info.Height;
 
-            int depth = single ? 1 : _depth;
+            int depth = _depth;
             int layers = single ? 1 : _layers;
-            int levels = single ? 1 : Info.Levels;
+            int levels = single ? 1 : (Info.Levels - level);
 
             width = Math.Max(width >> level, 1);
             height = Math.Max(height >> level, 1);
@@ -782,6 +784,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     width,
                     height,
                     depth,
+                    single ? 1 : depth,
                     levels,
                     layers,
                     Info.FormatInfo.BlockWidth,
@@ -834,6 +837,65 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
+        /// Converts texture data from a format and layout that is supported by the host GPU, back into the intended format on the guest GPU.
+        /// </summary>
+        /// <param name="output">Optional output span to convert into</param>
+        /// <param name="data">Data to be converted</param>
+        /// <param name="level">Mip level to convert</param>
+        /// <param name="single">True to convert a single slice</param>
+        /// <returns>Converted data</returns>
+        public ReadOnlySpan<byte> ConvertFromHostCompatibleFormat(Span<byte> output, ReadOnlySpan<byte> data, int level = 0, bool single = false)
+        {
+            if (Target != Target.TextureBuffer)
+            {
+                int width = Info.Width;
+                int height = Info.Height;
+
+                int depth = _depth;
+                int layers = single ? 1 : _layers;
+                int levels = single ? 1 : (Info.Levels - level);
+
+                width = Math.Max(width >> level, 1);
+                height = Math.Max(height >> level, 1);
+                depth = Math.Max(depth >> level, 1);
+
+                if (Info.IsLinear)
+                {
+                    data = LayoutConverter.ConvertLinearToLinearStrided(
+                        output,
+                        Info.Width,
+                        Info.Height,
+                        Info.FormatInfo.BlockWidth,
+                        Info.FormatInfo.BlockHeight,
+                        Info.Stride,
+                        Info.FormatInfo.BytesPerPixel,
+                        data);
+                }
+                else
+                {
+                    data = LayoutConverter.ConvertLinearToBlockLinear(
+                        output,
+                        width,
+                        height,
+                        depth,
+                        single ? 1 : depth,
+                        levels,
+                        layers,
+                        Info.FormatInfo.BlockWidth,
+                        Info.FormatInfo.BlockHeight,
+                        Info.FormatInfo.BytesPerPixel,
+                        Info.GobBlocksInY,
+                        Info.GobBlocksInZ,
+                        Info.GobBlocksInTileX,
+                        _sizeInfo,
+                        data);
+                }
+            }
+
+            return data;
+        }
+
+        /// <summary>
         /// Flushes the texture data.
         /// This causes the texture data to be written back to guest memory.
         /// If the texture was written by the GPU, this includes all modification made by the GPU
@@ -845,7 +907,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <returns>True if data was flushed, false otherwise</returns>
         public bool FlushModified(bool tracked = true)
         {
-            return Group.FlushModified(this, tracked);
+            return TextureCompatibility.CanTextureFlush(Info, _context.Capabilities) && Group.FlushModified(this, tracked);
         }
 
         /// <summary>
@@ -856,45 +918,22 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// Be aware that this is an expensive operation, avoid calling it unless strictly needed.
         /// This may cause data corruption if the memory is already being used for something else on the CPU side.
         /// </summary>
-        /// <param name="offset">The start offset of the range to flush data in bytes</param>
-        /// <param name="size">The size of the range to flush in bytes</param>
         /// <param name="tracked">Whether or not the flush triggers write tracking. If it doesn't, the texture will not be blacklisted for scaling either.</param>
-        public void Flush(int offset, int size, bool tracked)
+        public void Flush(bool tracked)
         {
-            if (TextureCompatibility.IsFormatHostIncompatible(Info, _context.Capabilities))
+            if (TextureCompatibility.CanTextureFlush(Info, _context.Capabilities))
             {
-                return; // Flushing this format is not supported, as it may have been converted to another host format.
+                FlushTextureDataToGuest(tracked);
             }
-
-            if (Info.Target == Target.Texture2DMultisample ||
-                Info.Target == Target.Texture2DMultisampleArray)
-            {
-                return; // Flushing multisample textures is not supported, the host does not allow getting their data.
-            }
-
-            FlushTextureDataRangeToGuest(tracked, offset, size);
         }
-
+        
         /// <summary>
-        /// Flushes the texture data, to be called from an external thread.
-        /// The host backend must ensure that we have shared access to the resource from this thread.
-        /// This is used when flushing from memory access handlers.
+        /// Gets a host texture to use for flushing the texture, at 1x resolution.
+        /// If the HostTexture is already at 1x resolution, it is returned directly.
         /// </summary>
-        /// <param name="offset">The start offset of the range to flush data in bytes</param>
-        /// <param name="size">The size of the range to flush in bytes</param>
-        public void ExternalFlush(int offset, int size)
+        /// <returns>The host texture to flush</returns>
+        public ITexture GetFlushTexture()
         {
-            if (TextureCompatibility.IsFormatHostIncompatible(Info, _context.Capabilities))
-            {
-                return; // Flushing this format is not supported, as it may have been converted to another host format.
-            }
-
-            if (Info.Target == Target.Texture2DMultisample ||
-                Info.Target == Target.Texture2DMultisampleArray)
-            {
-                return; // Flushing multisample textures is not supported, the host does not allow getting their data.
-            }
-
             ITexture texture = HostTexture;
             if (ScaleFactor != 1f)
             {
@@ -902,7 +941,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 texture = _flushHostTexture = GetScaledHostTexture(1f, _flushHostTexture);
             }
 
-            FlushTextureDataRangeToGuest(false, offset, size, texture);
+            return texture;
         }
 
         /// <summary>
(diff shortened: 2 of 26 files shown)


GamerzHell9137:
Life is Strange True Colors

PR
PR1

Master 
image
 
Dungeon Defenders Awakened 

PR
PR2

Master
image

gdkchan:
Fixes texture corruption in cutscenes on The Witcher 3 (gameplay after the cutscenes would also remain corrupted).

Before:
image
image
After:
image
image

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -1489,11 +1535,14 @@ private void DisposeTextures()
         /// Called when the memory for this texture has been unmapped.
         /// Calls are from non-gpu threads.
         /// </summary>
-        public void Unmapped()
+        public void Unmapped(MultiRange unmapRange)
```
gdkchan, line 1538: Missing XML doc for new argument.

Ryujinx.Graphics.Gpu/Image/TextureCache.cs
```diff
@@ -591,78 +614,81 @@ public void Lift(Texture texture)
...
+
+                        viewCompatible++;
 
                         hasLayerViews |= overlap.Info.GetSlices() < texture.Info.GetSlices();
                         hasMipViews |= overlap.Info.Levels < texture.Info.Levels;
                     }
                     else
                     {
+                        bool dataOverlaps = texture.DataOverlaps(overlap);
+                        if (!overlap.IsView && dataOverlaps && incompatibleOverlaps.FindIndex(incompatible => incompatible.Group == overlap.Group) == -1)
```
gdkchan, line 661: Why not use `!Exists` instead of `FindIndex`? It seems to be more suited for this.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -92,34 +93,60 @@ public static FormatInfo ToHostCompatibleFormat(TextureInfo info, Capabilities c
...
-                return lhs.Format == rhs.Format;
+                return lhsFormat.Format == rhsFormat.Format;
             }
 
-            if (lhs.IsCompressed && rhs.IsCompressed)
+            if (lhsFormat.IsCompressed && rhsFormat.IsCompressed)
             {
-                FormatClass lhsClass = GetFormatClass(lhs.Format);
-                FormatClass rhsClass = GetFormatClass(rhs.Format);
+                FormatClass lhsClass = GetFormatClass(lhsFormat.Format);
```
gdkchan, line 142: New format classes needs to be added for ASTC and ETC2, since ASTC is no longer handled earlier.

Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -456,6 +518,62 @@ public static bool ViewLayoutCompatible(TextureInfo lhs, TextureInfo rhs, int le
...
+            {
+                return false;
+            }
+
+            // For linear textures, gob block sizes are ignored.
+            // For block linear textures, the stride is ignored.
+            if (rhs.IsLinear)
+            {
+                int lhsStride = Math.Max(1, lhs.Stride >> lhsLevel);
+                lhsStride = BitUtils.AlignUp(lhsStride, 32);
```
gdkchan, line 543: Would be better to use `Constants.StrideAlignment` here too.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -2,12 +2,26 @@
...
 using Ryujinx.Memory.Range;
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace Ryujinx.Graphics.Gpu.Image
 {
+    struct TextureIncompatibleOverlap
+    {
+        public TextureGroup Group;
```
gdkchan, line 15: Maybe make them readonly if they are not modified.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -953,34 +1234,171 @@ public void CreateCopyDependency(Texture other, int firstLayer, int firstLevel,
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Registers another texture group as an incompatible overlap, if not already registered.
+        /// </summary>
+        /// <param name="other">The texture group to add to the incompatible overlaps list</param>
+        public void RegisterIncompatibleOverlap(TextureIncompatibleOverlap other, bool copy)
```
gdkchan, line 1306: Missing doc for `copy` argument.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -953,34 +1234,171 @@ public void CreateCopyDependency(Texture other, int firstLayer, int firstLevel,
...
+        {
+            if (_incompatibleOverlaps.FindIndex(overlap => overlap.Group == other.Group) == -1)
```
gdkchan, line 1308: Same comment about `FindIndex` vs `!Exists`.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -126,15 +193,17 @@ public void SignalModified()
...
+
+            RegisterSync(context);
         }
 
         /// <summary>
         /// Signal that this handle has either started or ended being modified.
         /// </summary>
         /// <param name="bound">True if this handle is being bound, false if unbound</param>
-        public void SignalModifying(bool bound)
+        public void SignalModifying(bool bound, GpuContext context)
```
gdkchan, line 204: Missing doc for `context`.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -247,73 +373,112 @@ private bool CheckDirty()
...
             {
                 fromHandle = DeferredCopy;
 
-                if (fromHandle != null && fromHandle._bindCount == 0)
+                if (fromHandle != null)
                 {
-                    // Repeat the copy in future if the bind count is greater than 0.
-                    DeferredCopy = null;
+                    // Only copy if the copy texture is still modified.
+                    // It will be set as unmodified if new data is written from cpu, as the data previously in the texture will flush.
```
gdkchan, line 391:
```suggestion
                    // It will be set as unmodified if new data is written from CPU, as the data previously in the texture will flush.
```

Ryujinx.Graphics.Gpu/Memory/MultiRangeWritableBlock.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+
+namespace Ryujinx.Graphics.Gpu.Memory
+{
+    /// <summary>
+    /// A writable block that targets a given MultiRange within a PhysicalMemory instance.
+    /// </summary>
+    internal class MultiRangeWritableBlock : IWritableBlock
+    {
+        private MultiRange _range;
```
gdkchan, line 12: Those can be `readonly`.

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -154,6 +154,24 @@ public unsafe ReadOnlySpan<byte> GetData()
...
+
+            if (HwCapabilities.UsePersistentBufferForFlush)
+            {
+                return _renderer.PersistentBuffers.Default.GetTextureData(this, size, layer, level);
+            }
+            else
+            {
+                IntPtr target = _renderer.PersistentBuffers.Default.GetHostArray(size);
+
+                WriteTo2D(target, layer, level);
```
gdkchan, line 169: It seems to return the offset where the data is actually located. Is it fine to ignore this offset here?

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -182,25 +200,31 @@ private int WriteTo2D(IntPtr data, int layer, int level)
...
+                GL.GetCompressedTextureSubImage(Handle, level, 0, 0, layer, Math.Max(1, Info.Width >> level), Math.Max(1, Info.Height >> level), 1, mipSize, data);
+            }
+            else if (format.PixelFormat != PixelFormat.DepthStencil)
+            {
+                GL.GetTextureSubImage(Handle, level, 0, 0, layer, Math.Max(1, Info.Width >> level), Math.Max(1, Info.Height >> level), 1, pixelFormat, pixelType, mipSize, data);
             }
             else
             {
+                // The GL function returns all layers. Must return the offset of the layer we're interested in.
+                int resultOffset = target switch
```
gdkchan, line 214: This variable has an odd placement since it is not used on the function call below. Could also just do `return target switch` directly (moving it after the call).

Ryujinx.Graphics.OpenGL/PersistentBuffers.cs
```diff
@@ -94,6 +94,21 @@ public unsafe ReadOnlySpan<byte> GetTextureData(TextureView view, int size)
             return new ReadOnlySpan<byte>(_bufferMap.ToPointer(), size);
         }
 
+        public unsafe ReadOnlySpan<byte> GetTextureData(TextureView view, int size, int layer, int level)
+        {
+            EnsureBuffer(size);
+
+            GL.BindBuffer(BufferTarget.PixelPackBuffer, _copyBufferHandle);
+
+            view.WriteToPbo2D(0, layer, level);
```
gdkchan, line 103: This also ignores the result offset.

Djipi71:
Test on BLue Fire : a lot less graphical issues :)
ryujinx_capture_2022-01-04_08-08-55

Djipi71:
Brothers now have perct Graphics with this fix
ryujinx_capture_2022-01-04_08-11-13

Djipi71:
Darksiders 3 Look a lot better
ryujinx_capture_2022-01-04_08-14-46

Djipi71:
Always shadows issues but a lot better too
ryujinx_capture_2022-01-04_08-23-45

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -166,9 +166,9 @@ public unsafe ReadOnlySpan<byte> GetData(int layer, int level)
             {
                 IntPtr target = _renderer.PersistentBuffers.Default.GetHostArray(size);
 
-                WriteTo2D(target, layer, level);
+                int offset = WriteTo2D(target, layer, level);
 
-                return new ReadOnlySpan<byte>(target.ToPointer(), size);
+                return new ReadOnlySpan<byte>(IntPtr.Add(target, offset).ToPointer(), size - offset);
```
gdkchan, line 171: Would probably have been easier to just slice the span here.

gdkchan (approved):
lgtm, great work, nice to see some of those texture issues fixed!

Djipi71:
with the new build Crash4 not starting :(
[Ryujinx_1.0.0+2b8da3f_2022-01-08_05-27-41.log](https://github.com/Ryujinx/Ryujinx/files/7832551/Ryujinx_1.0.0%2B2b8da3f_2022-01-08_05-27-41.log)

LukeWarnut:
> with the new build Crash4 not starting :( [Ryujinx_1.0.0+2b8da3f_2022-01-08_05-27-41.log](https://github.com/Ryujinx/Ryujinx/files/7832551/Ryujinx_1.0.0%2B2b8da3f_2022-01-08_05-27-41.log)

Disable guest internet access.

GamerzHell9137:
> > with the new build Crash4 not starting :( [Ryujinx_1.0.0+2b8da3f_2022-01-08_05-27-41.log](https://github.com/Ryujinx/Ryujinx/files/7832551/Ryujinx_1.0.0%2B2b8da3f_2022-01-08_05-27-41.log)
> 
> Disable guest internet access.

Won't help, there might come a PR that allows the game to boot again tho.

Thog (approved):

Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -232,6 +232,7 @@ private void DmaCopy(int argument)
                             data = LayoutConverter.ConvertBlockLinearToLinear(
                                 src.Width,
                                 src.Height,
+                                src.Depth,
                                 1,
                                 target.Info.Levels,
```
gdkchan, line 237: Can you fix this too?

gdkchan: Don't you need to use the overload of `SetData` that sets a single slice when passing a single layer/level?

riperiperi: The new parameter `sliceDepth` determines that actual depth that is copied. `Depth` just contains the total depth of the texture, so that the gob size is properly calculated.

gdkchan: I was not talking about that. I was talking about the `SetData` that is right after this call. `SetData` expects full texture data for all levels from what I understand, if you are just passing a single level, then the overload that takes a single layer/level should be used, no?

riperiperi: That is true, I didn't notice. I'll pass in `(0, 0)`.

riperiperi: This is a bit annoying since `AlwaysFlushOnOverlap` is set on the other method, and this one is used for partial sync when it is not needed. Also, changing this will remove all uses of the other method.

riperiperi: `memoryManager.Physical.TextureCache.FindTexture` was changed to only accept textures with one layer/level for now. This likely won't make any functional difference right now.

Djipi71:
Seems that the last build before merging have some issues (i hve post some of them on general channel)

PR #2698, opened 2021-10-03
Rewrite shader decoding stage

gdkchan:
**Description**

This changes the way how shader instructions are decoded. Before, it used an approach similar to what is done for the CPU, where instructions are divided into groups of instructions with similar encodings, and each group has an `OpCode*` type that performs the actual decoding process, extracting each field, register index etc. While this approach works rather well for the Arm CPU, it doesn't for shaders because the instruction encodings are very inconsistent. To give an example of this inconsistency, we can look at the `XMAD` instruction. For the register and immediate variants of this instructoin, PSL is encoded on bit 36, while the constant buffer + register variant has it on bit 55, and the register + constant buffer variant doesn't even have it. This type of inconsistency caused several bugs in the past due to fields being decoded from the wrong bits.

This new approach tries to remedy this by using a separate struct per instruction variant. This may be a bit wasteful, but I beleive it's the best way to avoid this kind of error, and also to avoid special handling on the instruction implementation.

The structs and tables are semi auto-generated. It all started with an experiement I did last year, where I tried to automate the process of gathering the GPU ISA information from the official tool (nvdisasm, part of the CUDA toolkit that can be found on NVIDIA's website). Explaining exactly how it works is out of the scope of this PR description, but I will try to give a brief explanation.

First we try to find all instruction mnemonics and variants. We know from open source disassembler and assemblers (such as nouveau) that only a smaller set of bits is responsible for deciding the instruction itself, while the remaining bits are for operands, modifiers, or are unused. So first we can just try to disassemble all possible combinations of those bits to get all existing instruction mnemonics. From there, we can try to find where modifiers, operands, etc are encoded. Since each instruction is a 64-bit value, we can simply set a single bit to 1 and see what changes, then set the next bit, see what changes, etc. Using this approach we can associate bit fields to a given operand or modifier. Once those have been found, we can then try all possible combinations of the modifier bit field to get possible modifier names.

For an example of what the output might look like, see: https://gist.github.com/gdkchan/fb9def32eb1b733b65214ab8697e6db1
Note that this is a very old test and has error, it's not what I ended up using, I'm including it here just for illustration purposes.

**What does this improve?**

Other than what I already mentioned, there are a few other improvements. For example, now we have all the shader instructions on the table (not all of them are actually implemented, of course), this should facilitate future work when we need to actually implement those instructions.

While working on it, I found the following errors on the old decoder that are now fixed.
- One of the double instructions had inexistent fields.
- One of the half instruction had SAT using the FTZ bit.
[...]

PR #2972, opened 2022-01-04
Implement IMUL, PCNT and CONT shader instructions, fix FFMA32I and HFMA32I

gdkchan:
A bunch of shader fixes here:
- Implement the IMUL (integer multiply) shader instruction. Used by MelonDS Switch port and possible other homebrew. Not sure if commercial games uses this shader instruction.
- Implement PCNT and CONT shader instructions. Those are similar to existing PBK/BRK and SSY/SYNC pairs. Also used by MelonDS Switch port and possibly other homebrew, AFAIK the NVIDIA compiler never uses this instruction.
- Fix FMUL32I instruction implementation. Third operand should use the destination register, not "SrcC" as it does not exist for this instruction. (Regression from https://github.com/Ryujinx/Ryujinx/pull/2698).
- Fix a issue similar to the above one for HFMA32I aswell, but this one was also missing from the instruciton table so I have added it too. (Technically not a regression as this instruction was not implemented before).

Testing on games is welcome. MelonDS is tricky to test right now as it requires other changes to work, plus only the newer version uses GPU rendering, the older ones are software rendered and won't be affected by those changes.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -40,7 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 2764;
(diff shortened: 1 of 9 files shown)


riperiperi (approved):
Looking good, nice work.

Thog (approved):

PR #2976, opened 2022-01-05
Stop Emulation UI clean-up + fixes

### LINKED ISSUE #2975: "Close emulation" option does not ask for confirmation, even when "Show confirm exit dialogue" is enabled
When clicking "Close emulation" on Actions tab when in windowed mode, it doesn't trigger the "Do you wanna close emulation" message like it gets triggered when pressing ESC. (It should be triggered because Confirm Exit dialog option is enabled in Ryujinx options)

Another issue is that the message is wrong when pressing ESC to close the emulation, it says "Do you wanna close Ryujinx" instead.

### Environment?

Ryujinx version: 1.0.7159

System Specs:
     OS: Windows 10
     CPU: Ryzen 3600
     GPU: NVIDIA GTX 960
     RAM: 16GB

MutantAura:
Fixes the following bugs/inconsistencies:

- Clicking "Stop Emulation" now brings up the dialog box when the show confirmation setting is ticked. A custom dialog was also created instead of re-using the dialog for exiting the entire emulator as this would be misleading and inconsistent.
- Ending emulation when not in full screen with escape now also uses this new dialog box as before it used the program ending dialog.

Remaining bugs:
- [ ] Ending emulation with Escape when not in full screen will cause the same window to appear when a new game is booted. This is both an annoyance and game-breaking when combined with toggling the confirm prompt. This combination of events causes the launched game to end instantly.

Closes #2975

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1320,14 +1320,13 @@ namespace Ryujinx.Ui
 
         private void StopEmulation_Pressed(object sender, EventArgs args)
         {
-            if (_emulationContext != null)
+            if (!_gameLoaded || !ConfigurationState.Instance.ShowConfirmExit || GtkDialog.CreateEndDialog())
             {
                 UpdateGameMetadata(_emulationContext.Application.TitleIdText);
+                _pauseEmulation.Sensitive = false;
+                _resumeEmulation.Sensitive = false;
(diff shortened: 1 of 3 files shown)


GamerzHell9137:
Fixes https://github.com/Ryujinx/Ryujinx/issues/2975

MutantAura:
Closing as this is fixed in Avalonia #2507

PR #2983, opened 2022-01-08
Return error on DNS resolution when guest internet access is disabled

gdkchan:
The title says it all, if guest internet access is disabled, no DNS lookup is performed and it returns an error immediately. If the guest internet access is disabled, it doesn't make sense that we allow the servers to be reached.

This fixes a regression introduced by #2936 that made Crash Bandicoot 4 crash early on boot after trying to connect to some server. Note that the game will still crash with guest internet access enabled, that should be fixed separately.

The regression likely comes from the fact that before the fix on the linked PR, it was returning a serialized size of 0 which made the applications assume nothing was actually returned. This change should make it behave in a similar way to the old behaviour, except we return an actual error code now.

Thanks to Muffin for finding the change that caused the regression.


AcK77 (approved):

riperiperi (approved):

GamerzHell9137:
Can confirm that Crash is working again 
image

PR #2984, opened 2022-01-09
Fix sampled multisample texture size

gdkchan:
The width/height of render target and copy textures is already pre-multiplied by the driver for multisample textures. For shader sampled textures that are on the pool, they are not pre-multiplied. This changes multiply their size by the multisample size, in order to allow them to match existing textures on the cache, in addition to allowing the texture to have the correct size (as the `TextureCreateInfo` that is passed to the backend has the width and height divided by the amount of samples).

Fixes rendering on Okami HD.
Before (from gamedb):
image
image
PR:
image
image

Might also fix broken fight effects on Rune Factory 5.
Testing on games using multisampling is appreciated.


riperiperi (approved):
Nice.

Thog (approved):

PR #2991, opened 2022-01-10
account: Rework LoadIdTokenCache to auto generate a random JWT token

Thog:
This improve correctness of that call while possibly spoofing possible recognizable patterns.


gdkchan:
Can you elaborate a bit more on the description on how this compares to the official implementation? Just so we know how much of it is done to follow official code, and how much was added/modified by you.

Thog:
> Can you elaborate a bit more on the description on how this compares to the official implementation? Just so we know how much of it is done to follow official code, and how much was added/modified by you.

Based on official format returned by the right server endpoint (see https://github.com/kinnay/NintendoClients/wiki/Switch-Tokens#id-tokens)

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #2992, opened 2022-01-10
sfdnsres: Fix serialization issues

Thog:
Fix a crash on Monster Hunter Rise

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo4.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo4.cs
@@ -29,17 +29,17 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.Types
         {
             Port = IPAddress.HostToNetworkOrder(Port);
 
-            RawIpv4AddressToSwapEndianess(ref Address);
+            RawIpv4AddressNetworkEndianSwap(ref Address);
         }
 
         public void ToHostOrder()
         {
             Port = IPAddress.NetworkToHostOrder(Port);
 
-            RawIpv4AddressToSwapEndianess(ref Address);
+            RawIpv4AddressNetworkEndianSwap(ref Address);
         }
 
-        public static void RawIpv4AddressToSwapEndianess(ref Array4<byte> address)
+        public static void RawIpv4AddressNetworkEndianSwap(ref Array4<byte> address)
         {
             if (BitConverter.IsLittleEndian)
             {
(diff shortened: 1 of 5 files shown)


Ryujinx.HLE/Utilities/StringUtils.cs
```diff
@@ -60,6 +60,24 @@ public static byte[] HexToBytes(string hexString)
             return output;
         }
 
+        public static string ReadUtf8String(ReadOnlySpan<byte> data, out int dataRead)
+        {
+            dataRead = 0;
```
gdkchan, line 65: You could just do `dataRead = data.IndexOf(0) + 1;` here. To account for the case of the span having no null terminator, you would need to additionally add `if (dataRead == 0) dataRead = data.Length`.

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfo4.cs
```diff
@@ -16,14 +17,34 @@ struct AddrInfo4
...
+        }
+
+        public void ToHostOrder()
+        {
+            Port = IPAddress.NetworkToHostOrder(Port);
+
+            RawIpv4AddressToSwapEndianess(ref Address);
+        }
+
+        public static void RawIpv4AddressToSwapEndianess(ref Array4<byte> address)
```
gdkchan, line 42: The name of this function seems a bit odd. `ToSwapEndianess`? I would call it `RawIpv4AddressNetworkEndianSwap`.

Ryujinx.HLE/Utilities/StringUtils.cs
Thog: No null terminator isn't a valid case here as official software use ``strlen``, will adjust anyway.

Ryujinx.HLE/Utilities/StringUtils.cs
```diff
@@ -60,6 +60,19 @@ public static byte[] HexToBytes(string hexString)
             return output;
         }
 
+        public static string ReadUtf8String(ReadOnlySpan<byte> data, out int dataRead)
+        {
+            dataRead = data.IndexOf((byte)0) + 1;
+            int size = data.Length;
```
gdkchan, line 66: Unused?

Ryujinx.HLE/Utilities/StringUtils.cs
```diff
@@ -60,6 +60,19 @@ public static byte[] HexToBytes(string hexString)
...
+
+            if (dataRead <= 1)
```
gdkchan, line 68: Why does it return null for an empty string?

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #2993, opened 2022-01-11
Fix adjacent 3d texture slices being detected as Incompatible Overlaps

riperiperi:
This fixes some regressions caused by #2971 which caused rendered 3D texture data to be lost for most slices. Fixes issues with Xenoblade 2's colour grading, probably a ton of other games.

This also removes the check from TextureCache, making it the tiniest bit smaller (any win is a win here).


gdkchan (approved):
Nice, thanks for the quick fix!

Thog (approved):

PR #2994, opened 2022-01-11
Fix render target clear when sizes mismatch

### LINKED ISSUE #2989: Pathway - 0100701015846800
The game has a glitchy screen and does not scale up to the entire screen.
The user interface is in the correct position, however the game itself is not scaled up, and is in a smaller window causing the rest of the game to be cut off. Also there are visual glitches where some graphics leave a trail that does not disappear.

1. Launch Pathway.
2. You will notice the glitch in the intro screen.
3. The glitch can be seen in the main menu.
4. When launching the game, the glitch can be seen in the game.

### Log file

N/A

### Environment?

- Ryujinx version: 1.0.7164</br>
- Game version: 1.2.5</br>
- System Specs:
  - OS: Windows 10
  - CPU: AMD Ryzen 5 3600
  - GPU: GTX 1070
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Here are some screenshots of the glitch:
screenshot1
screenshot2
screenshot3
screenshot4

gdkchan:
On OpenGL (and Vulkan too iirc), when the bound render targets have different sizes, then it only render on the intersection of all their sizes. On the GPU, this clipping is controlled by the `ScreenScissorState` pair of registers. This register was being mostly ignored before, but for clears, that may cause issues if there are render targets of different sizes bound, and the game is trying to clear one of them, with a screen scissor size that matches the target being cleared. OpenGL would clip it to the smallest size and not clear the entire region.

This issue has been fixed by forcing all other render targets to be unbound, to avoid the host clipping, and then using a custom scissor region, calculated from the screen scissor and user scissor (0). Additionally, I also implemented the clear flags (that can be used to disable user scissor on stencil mask from affecting clear, but everything I tested so far always have both enabled). Also includes other minor improvements.

Fixes Pathway not having the screen entirely cleared.
Before:
image
After:
image
Fixes #2989
Testing is welcome (as always).

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -1,4 +1,3 @@
-using Ryujinx.Graphics.Shader;
 using System;
 
 namespace Ryujinx.Graphics.GAL
@@ -15,9 +14,9 @@ namespace Ryujinx.Graphics.GAL
 
         void ClearRenderTargetDepthStencil(
             float depthValue,
-            bool  depthMask,
-            int   stencilValue,
-            int   stencilMask);
+            bool depthMask,
+            int stencilValue,
+            int stencilMask);
 
         void CommandBufferBarrier();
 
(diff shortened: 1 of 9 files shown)


LukeWarnut:
Improves rendering in Charge Kid.

Before:
image

After:
image

Thog (approved):

riperiperi (approved):
LGTM!

LukeWarnut:
> Improves rendering in Charge Kid.

I'm unable to reproduce this now. It may not have been affected by the change,

PR #2998, opened 2022-01-15
Add korean-enabled fallback font to software keyboard on Windows

Caian:
The fallback fonts were designed primarily for Linux as there's no consensus on what fonts should be available in the system.

But apparently, there are some scenarios on Windows where the font retrieved from GTK can cause issues. One such scenario is with Korean fonts.

This PR adds a font that is compatible with Korean characters (and perhaps other characters as well).

This is a WIP because I believe that Microsoft Sans Serif solves the issue but I don't have a Korean system easily available for testing.


unuruu5760:
Korean Windows uses the Malgun Gothic font. 
Using Malgun Gothic instead of Microsoft Sans Serif completely solves this problem. I've been waiting a long time for this to be merged into mainline and LND.

PR #2999, opened 2022-01-15
kernel: Fix deadlock when pinning in interrupt handler

Thog:
This fix a misplace critical section leave.
This fix a deadlock on DoDonPachi Resurrection when starting a new game and possibly other games.


Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
```diff
@@ -761,12 +761,12 @@ private void InterruptHandler(object sender, EventArgs e)
 
                 currentThread.SetUserInterruptFlag();
 
+                KernelContext.CriticalSection.Leave();
+
                 if (currentThread.IsSchedulable)
```
gdkchan, line 766: If you change the `else if` below to just `if`, then you can remove this `if` block.

gdkchan (approved):
lgtm, thanks.

emmauss (approved):

PR #3002, opened 2022-01-16
Scale scissor used for clears

gdkchan:
Fixes a regression introduced on #2994 that would cause the screen to "flicker" in some games with resolution scale other than native. Affected games includes Luigi's Mansion 3.

Also changes `Math.Ceiling` to `MathF.Ceiling` to avoid an implicit cast to double (plus, float operations are generally faster than double ones).


Thog (approved):

meisiek:
I tested Luigi's Mansion and now the flicker is gone when using resolution scale. Thanks!

riperiperi (approved):

PR #3005, opened 2022-01-17
Use a SSBO for the shader support buffer, if using UBOs would be above limit

### LINKED ISSUE #3003: Beyond a Steel Sky gfx issue.

Some parts of the environment are not rendered.

Screenshot using Ryujinx.
ryujinxbug

Screenshot using Yuzu.
yuzuright

Just use the included save file and load the game.

### Log file

[Ryujinx_1.0.7175_2022-01-17_13-00-09.log](https://github.com/Ryujinx/Ryujinx/files/7881676/Ryujinx_1.0.7175_2022-01-17_13-00-09.log)

### Environment?

- Ryujinx version: 1.0.7175

- Game version: 1.0.29070 (But also happens without the update.)

- System Specs:
  - OS: Windows 10
  - CPU: i7-7700K
  - GPU: NVIDIA GTX 1060
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Save file.
[save.zip](https://github.com/Ryujinx/Ryujinx/files/7881463/save.zip)

gdkchan:
APIs have a limit on the number of uniform buffers one can use on the shader. On NVIDIA OpenGL that is 14. The Switch GPU supports up to 18, but APIs actually exposes less than that to the user. The constant buffers that are not exposed to the user, are instead reserved by the driver and used to store constants and other data used by the shader.

There are some games getting past the host limit. For example, Beyond a Steel Sky has shaders using 14 uniform buffers. When you add the support buffer one, then it becomes 15, which is past the limit. This change addresses this case by instead using a storage buffer for the support buffer if it would be past the limit with a uniform. To make this work, now theres one storage binding reserved for the support buffer too, and it is also bound to the SSBO binding point 0.

Note that this does not address *all* possible cases. For example, if the shader uses 15 or more uniform buffers, it still won't be enough. In those cases, it should try converting some of those uniform buffers to storage buffers.

Also worth noting that Vulkan likely does not have the issue, as the limit on the number of uniform buffers per shaders is higher than OpenGL.

Fixes missing geometry in Beyond a Steel Sky.
Before:
image
After:
image
Closes #3003.

### DIFF
--- a/Ryujinx.Graphics.GAL/Capabilities.cs
+++ b/Ryujinx.Graphics.GAL/Capabilities.cs
@@ -19,7 +19,14 @@ namespace Ryujinx.Graphics.GAL
 
         public int MaximumComputeSharedMemorySize { get; }
         public float MaximumSupportedAnisotropy { get; }
+        public int MaximumSupportedComputeUniforms { get; }
+        public int MaximumSupportedVertexUniforms { get; }
+        public int MaximumSupportedTessControlUniforms { get; }
+        public int MaximumSupportedTessEvaluationUniforms { get; }
+        public int MaximumSupportedGeometryUniforms { get; }
+        public int MaximumSupportedFragmentUniforms { get; }
         public int StorageBufferOffsetAlignment { get; }
+
 
         public Capabilities(
             bool hasFrontFacingBug,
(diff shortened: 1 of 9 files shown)


Thog (approved):

remibrat (approved):

riperiperi (approved):

Thog:
Need rebase and will be mergeable.

Dime333:
Sorry for my ignorance, but is there something that keeps this from getting merged? I'm not sure what a rebase is.

Moonlacer:
> Sorry for my ignorance, but is there something that keeps this from getting merged? I'm not sure what a rebase is.

A rebase is a way of combining the latest Ryujinx version code with this pull request, which may uncover any issues that will stop the merge of this pull request until the mentioned issue(s) is resolved

AcK77:
Needs a rebase

gdkchan:
I just tested the game on OpenGL and the issue doesn't exist anymore, maybe due to some changes we did somewhat recently that reduces the amount of constant buffer that shaders use in some cases.
image
So I will close this as it's no longer needed.

Issue #2357, opened 2021-06-10
Xenoblade Chronicles 2 menu freeze


fehler001:
### What's the issue you encountered?

When click very fast from game main menu to submenu, there is a low chance that game will freeze, the submenu won't show up, while bgm is still there.</br>

### How can the issue be reproduced?

After playing a few minutes in big map, open game menu, click very quickly into either 傭兵団(Mercenary) or SAVE, then back to main menu, then get into either this two menu, there is a low chance to get freeze (other submenu also has this bug but under a very low chance).

### Log file
[Ryujinx_1.0.6909_2021-06-10_12-42-55.log](https://github.com/Ryujinx/Ryujinx/files/6628988/Ryujinx_1.0.6909_2021-06-10_12-42-55.log)

### Environment?

- Ryujinx version: 1.0.6909</br>
- Game version: 2.0.2(JPN)</br>
- System Specs:
  - OS: Windows 10 1909
  - CPU: i7-7700k
  - GPU: NVIDIA GTX 1050 2g
  - RAM: 32GB
- Applied Mods : No

### Additional context?
When playing this game, Dedicated GPU memory is about 1.8/2.0GB, but GPU Memory is above 2g, like around 2.3g.  
I play this game with turning VSync off to get tiny more fps, this bug also happens with VSync on.
</br>

fehler001:
I tested 10 hours on 1.0.6912, encountered freeze once in Mercenary menu, once in directly opening fast travel menu.
It seems the frequency has been eased a lot.

lostromb:
Okay, so I've bumbled around with this issue for a while, having no real experience with this codebase but about 70 hours into the game on emu, and this is the only bug I have ever encountered that is keeping the game from being 100% perfect as far as I can tell. These are my notes so far:
1. My hunch is that it is filesystem related, because it happens not only during the loading transitions between menus (opening the blade or merc menu or, more frustratingly, the save menu) but I have also had it happen a few times during the regular loading screens, usually the initial game load before textures / music / etc. is even present
2. It does not seem to be timing related, at least on file reads - I tried injecting random delay into each file read operation in IStorage and I could still play the game fine despite the huge hitches. I could check if delaying other calls (write? open partition?) have any effect.
3. It does not seem to be caused by an unhandled error code or some kind of exceptional file event, I triggered the bug with log lines on most every error condition I could find and none of them fired. There could be an error happening somewhere else though
[...]

PR #3015, opened 2022-01-21
Add host CPU memory barriers for DMB/DSB and ordered load/store

### LINKED ISSUE #2719: Pokemon Mystery Dungeon: Rescue Team DX inconsistent booting

RTDX seems to struggle to boot on master and consistently seems to hang on startup. There is a chance it does boot but this seems exceedingly random and no variation of settings or versions seems to help this.

Simply attempt to boot the game a few times, you may be able to boot once or twice but as I say it seems fairly inconsistent if the startup process hangs.

### Log file
**All log files are terminated by a Stop emulation command after waiting for around 3-4 mins for change in activity.**

[Ryujinx_1.0.7065_2021-10-08_18-21-35.log](https://github.com/Ryujinx/Ryujinx/files/7313300/Ryujinx_1.0.7065_2021-10-08_18-21-35.log)

Seems to always hang on either:

HLE.OsThread.9 ServiceNv Create: Created map XX with size 0x01000000!
or some form of GPU thread "too slow" message.

Log with all debugging logs enabled: 
[Ryujinx_1.0.7065_2021-10-08_18-20-02.log](https://github.com/Ryujinx/Ryujinx/files/7313306/Ryujinx_1.0.7065_2021-10-08_18-20-02.log)

### Environment?

- Ryujinx version: 1.0.7065
- Game version: 1.0.2
- System Specs:
  - OS: Windows Version	10.0.19043 Build 19043
  - CPU: Ryzen 5 5600X
  - GPU: Nvidia RTX 3070 FE
  - RAM: 32GB DDR4 
- Applied Mods : No

### Additional context?

Works fine in LDN2.3 so I can only imagine some form of regression somewhere between then and now. That doesn't really narrow the issue down though haha

LDN2.3 boot log:
[Ryujinx_1.0.0-ldn2.3_2021-10-08_19-00-33.log](https://github.com/Ryujinx/Ryujinx/files/7313323/Ryujinx_1.0.0-ldn2.3_2021-10-08_19-00-33.log)

gdkchan:
This just adds a memory barrier for the guest barrier instructions on the CPU. Before we had "host mapped memory" (the POWER PR), this was not *as* important because software memory emulation had to do address translation and validation, which would generate a large number of instructions per memory access, which makes memory reordering across separate guest memory accesses less likely. With the change it becomes a bigger problem because guest memory access can compile to a single host instruction.

Should fix softlocks on those games:
- Pokémon Mystery Dungeon DX (would sometimes get stuck on a black screen right before the title screen).
- Pokémon Brilliant Diamond/Shining Pearl (usually entering or exiting buildings, or changing areas in general).
Potentially also affected:
- Pokémon Sword/Shield? I know it has softlocks but I'm not familiarized with this one.
- YUME NIKKI Dream Diary (also before fully showing the title screen and Press Start button).
- Likely others?

Would appreciate testing for some of the games above, since due to the nature of this issue (happening completely randomly or not at all, depending on CPU vendor and model etc). Also would be nice to get confirmation that it's not causing any noticeable performance regression, or other issues that didn't exist before in general.

Closes #2719

### DIFF
--- a/ARMeilleure/CodeGen/X86/Assembler.cs
+++ b/ARMeilleure/CodeGen/X86/Assembler.cs
@@ -358,6 +358,12 @@ namespace ARMeilleure.CodeGen.X86
             WriteInstruction(dest, source, type, X86Instruction.Lea);
         }
 
+        public void LockOr(Operand dest, Operand source, OperandType type)
+        {
+            WriteByte(LockPrefix);
+            WriteInstruction(dest, source, type, X86Instruction.Or);
+        }
+
         public void Mov(Operand dest, Operand source, OperandType type)
         {
             WriteInstruction(dest, source, type, X86Instruction.Mov);
--- a/ARMeilleure/CodeGen/X86/CodeGenerator.cs
+++ b/ARMeilleure/CodeGen/X86/CodeGenerator.cs
@@ -49,6 +49,7 @@ namespace ARMeilleure.CodeGen.X86
             Add(Instruction.Load,                    GenerateLoad);
             Add(Instruction.Load16,                  GenerateLoad16);
             Add(Instruction.Load8,                   GenerateLoad8);
+            Add(Instruction.MemoryBarrier,           GenerateMemoryBarrier);
             Add(Instruction.Multiply,                GenerateMultiply);
             Add(Instruction.Multiply64HighSI,        GenerateMultiply64HighSI);
             Add(Instruction.Multiply64HighUI,        GenerateMultiply64HighUI);
@@ -538,7 +539,7 @@ namespace ARMeilleure.CodeGen.X86
                     context.Assembler.Lea(dest, memOp, dest.Type);
                 }
             }
-            else 
+            else
             {
                 ValidateBinOp(dest, src1, src2);
 
@@ -974,6 +975,11 @@ namespace ARMeilleure.CodeGen.X86
             Debug.Assert(value.Type.IsInteger());
 
             context.Assembler.Movzx8(value, address, value.Type);
+        }
+
+        private static void GenerateMemoryBarrier(CodeGenContext context, Operation operation)
+        {
+            context.Assembler.LockOr(MemoryOp(OperandType.I64, Register(X86Register.Rsp)), Const(0), OperandType.I32);
         }
 
         private static void GenerateMultiply(CodeGenContext context, Operation operation)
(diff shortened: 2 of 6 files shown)


Thog (approved):

MutantAura:
Closes #2719

riperiperi (approved):
Great stuff. Tested BDSP for longer than I'd like to admit and it seems to have solved the issue.

lostromb:
I tried for about 45 minutes to intentionally repro the Xenoblade 2 softlock. Once I got a lockup in the very initial boot screen, before the swirling pattern changed to the crystal loading icon. But I was not able to trigger a lockup in-game where it typically poses a problem. I'll try to do more testing but I will tentatively say this closes #2357

CeruleanSky:
@lostromb While this may mitigate a class of menu softlocks for Xenoblade 2, it will be tough figuring out if this eliminates all causes of them, as Yuzu also suffers from the same problem, though I noticed it doesn't occur as often.  Your hunch about  (possibly slow) filesystem access making the soft lock more likely applies to that emulator as well.  If this pull request makes things more stable, or even better though harder to determine, eliminates the softlocks completely, that would be wonderful. But it might be good to hesitate on closing that ticket.

gdkchan:
> But it might be good to hesitate on closing that ticket.

I will keep it open. If we get a confirmation from lostromb that it has been fixed, and no one else can reproduce it, then we can close it. If someone manages to trigger it again on the future, we can re-open. It's very hard to be 100% sure with this kind of bug, afterall.

lostromb:
Unfortunately I just triggered the Xenoblade menu softlock again on build `Release-1.0.0+b532ddc-win_x64`. So I think #2357 stays open.

DaSk79130569:
I played the game for 5 hours today to test this fix, with the latest version of the emulator up on the main website. My save is midway through the game and in my testing I played sections in multiple major areas.

Prior to this new fix, the game would frequently crash (soft lock) while navigating the menus, and sometimes crash on loading screens.

After this new fix, the game never crashed for me while navigating menus in 5 hours of playtime. So that is a huge improvement! Great work! 

The game did still crash one time while loading an area.

So I conclude that most soft locks have been eliminated or at least the situation is a big improvement, but it is possible there may still be some residual soft locks i.e. from a different mechanism. I assume different user software/hardware configurations could also play a role.

DaSk79130569:
I can confirm that after 15 hours of playtesting, the Xenoblade 2 menu bug seems to have been fixed with new version. I have had zero menu soft locks.

The loading screen soft lock remains.

PR #3026, opened 2022-01-22
Remove Appveyor from Readme and SLN

Mou-Ikkai:
I replaced the Appveyor badge in the Readme with a Github Actions alternative. Along with this, I removed appveyor.yml and removed it from the SLN.


Shahil-Ayato:
Was just about to do this, nice work.

Thog (approved):

Shahil-Ayato (approved):

jduncanator:
It would probably be best to have an alternative setup first before you remove the only CI tool that does a release build.

Actions only builds PR's at the moment, and only generates "dirty" Git tagged builds. We'd need to set it up to generate versioned builds from master.

GitHub actions also isn't a good solution to this as you need to have a GitHub account _and_ be logged in to download artifacts. It's going to be a pretty poor user experience for most users who want to download the build off the website directly.

riperiperi:
Are people still using the appveyor config to run their own builds? Might cause issues just removing the config outright.

yell0wsuit:
Agree with the badge change, but probably should keep appveyor config for the time being like riperiperi said

Thog:
> Are people still using the appveyor config to run their own builds? Might cause issues just removing the config outright.

I mean people could just readd it if they want, it's not like they are going to migrate if we don't make any move here...

AcK77 (approved):

PR #3041, opened 2022-01-25
Fix: Autoupdater doesn't set the correct permissions on Linux/Unix

### LINKED ISSUE #2497: The executable file is not executable after auto-updating on Linux
This started about few weeks ago. Ryujinx performs an auto-update, closes and doesn't re-launch itself. I look into its folder and see that "Ryujinx" file doesn't have its "executable" attribute set on. I have to make it executable manually. Then it works but only until the next update.

edisionnano:
After downloading the latest release from GitHub actions which doesn't have executable rights we call chmod from the binding to libc that we have created. While the chmod command accepts permissions in Symbolic Mode (for example +x for executable rights) or Octal Mode (for example 755), the libc api accepts permissions in Decimal Mode. The 0 preceding 777 in the code suggests that its an octal number in a language like C however such thing does not exist in C# therefore we need to convert it to Decimal. The reason 755 is chosen instead of 777 this time is because these are the default permissions dotnet gives the binary and when converted in decimal the equivalent is 493.

Closes #2497

### DIFF
--- a/Ryujinx/Modules/Updater/Updater.cs
+++ b/Ryujinx/Modules/Updater/Updater.cs
@@ -396,7 +396,7 @@ namespace Ryujinx.Modules
 
             if (!OperatingSystem.IsWindows())
             {
-                chmod(ryuBin, 0777);
+                chmod(ryuBin, 493);
             }
         }
 


gdkchan (approved):
Nice catch.

Thog (approved):

PR #3042, opened 2022-01-26
kernel: A bit of refactoring and fix GetThreadContext3 correctness

Thog:
This PR cherrypick some misc parts of #1504 and give small clean up here and there.

Changes:
- All syscalls output are now before input arguments.
- All syscalls now doesn't use raw memory write to output data structure to userspace.
- All references to ``position`` were renamed to ``address``.
- ``GetThreadContext3`` have been rewritten to be accurate (fixing AArch32 support for it in the process).
- ``KernelTransfer`` now uses ``IsRangeMapped`` and uses generic functions.
- Some fields were switched to unsigned to improve correctness and avoid useless casts.

PR #3043, opened 2022-01-26
misc: Make PID unsigned long instead of long

Thog:
Depends on #3042.

As title say, separated from the other one to avoid a huge unreadable diff.


gdkchan (approved):
lgtm. Would be nice to also rename `PId` and `pId` to `Pid` and `pid` respectively.

riperiperi (approved):
LGTM.

PR #3044, opened 2022-01-27
Add ability to restart emulation in Actions menu.

### LINKED ISSUE #2974: Restart Emulation Action
I would like to request the addition of a "Restart Emulation" option in the action drop down, similar to how there is a pause, unpause and stop emulation function. 

Essentially it would literally just reboot the entire game in just a single click, instead of having you "stop emulation" and then clicking your game in the game's list to restart it. It just saves just a small bit of time essentially.

The main reason I would like to recommend the feature is that simply put... there are some games that once you've loaded your save, don't have a menu where you can access your save files to reload your data if you make a mistake or need to backtrack or whatever the reason. You literally have to exit out of the game, and reboot it to go back to the start menu and access the load screen again. In games where there are multiple choice factors or other things, this can be a common occurrence depending on the circumstance.

Far from it being a priority feature as obviously you can just do this the long way, but it might be a nice function to implement in the future if simple enough to do so.

Thank you. <3

saldabain:
Fixes #2974 

Adds a menu item to restart emulation of the current game.

Needs review as the implementation may not be technically correct (although it appears to work).


pflyly:
This function is very useful in PLA to S/L shiny mons.

Dudice:
Waiting for this function :D

gandalfthewhite19890404:
Can we set hotkey for Restart function?

riperiperi (approved):
Looks good, though might probably fall victim to the emulator freezing when stopping emulation occasionally.

AcK77:
I think you did something wrong during a rebase

PR #3046, opened 2022-01-27
Fix res scale parameters not being updated in vertex shader

riperiperi:
This fixes an issue where the render scale array would not be updated when technically the scales on the flat array were the same, but the start index for the vertex scales was different.


gdkchan (approved):
lgtm, nice catch!

Thog (approved):

PR #3049, opened 2022-01-27
Add timestamp to 16-byte/4-word semaphore releases.

riperiperi:
BOTW was reading a ulong 8 bytes after a semaphore release. Turns out this is the timestamp it was trying to do performance calculation with, so I've made it write when necessary. This fixes BOTW being capped at 20fps all the time. (now it only does this when the game runs too slowly)

image

This mode was also added to the DMA semaphore I added recently, as it is required by a few games. (i think quake?)

The timestamp code has been moved to GPU context. Check other games with an unusually low framerate cap or dynamic resolution to see if they have improved.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -115,8 +115,8 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                 }
                 else /* if (type == LaunchDmaSemaphoreType.ReleaseFourWordSemaphore) */
                 {
-                    _channel.MemoryManager.Write(address, (ulong)_state.State.SetSemaphorePayload);
                     _channel.MemoryManager.Write(address + 8, _context.GetTimestamp());
+                    _channel.MemoryManager.Write(address, (ulong)_state.State.SetSemaphorePayload);
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -115,7 +115,8 @@ private void ReleaseSemaphore(int argument)
                 }
                 else /* if (type == LaunchDmaSemaphoreType.ReleaseFourWordSemaphore) */
                 {
-                    Logger.Warning?.Print(LogClass.Gpu, "DMA semaphore type ReleaseFourWordSemaphore was used, but is not currently implemented.");
+                    _channel.MemoryManager.Write(address, _state.State.SetSemaphorePayload);
```
gdkchan, line 118: Maybe it should be casted to `ulong`? Or maybe adding a write of 0 to offset 4 too like the GPFifoClass one, otherwise there will be 4 bytes that will be unchanged.

riperiperi: Done, totally blanked on that.

riperiperi:
Oh yeah, make sure to test in a bunch of games, in case this breaks something.

MelonSpeedruns:
Fixes Pokémon Legends Arceus having a very low resolution:

Before:
image

After:
image

gdkchan (approved):
lgtm, thanks!
Very nice to see the BOTW issue finally fixed.

Thog (approved):

PR #3054, opened 2022-01-28
Only initialize shader outputs that are actually used on the next stage

gdkchan:
If one shader stage reads an input that is not written by the previous stage, as per the spec, the contents are undefined. This is not a problem on NVIDIA as the contents are actually consistent for those uninitialized values, but it does cause issues for Intel and AMD. So to fix that, the shader translator adds code to explicitly initialize outputs that are used by the next stage, but might not be written on the current one.

Right now it initialize some outputs that don't need to be initialized, for one of those two reasons:
- The specific component is not accessed on the next stage. Right now it only keeps track of the entire vec4 input that is consumed on the next stage. It does not keep track of the individual XYZW float components.
- The output might not be used on the next stage at all. Right now it initializes both outputs written on the current stage *and* inputs consumed on the next one.

This change addresses both of the above cases. First, a new `UInt128` type was added. It is used to keep track of individual components that are used on the next stage, per input. Since there is a maximum of 32, and each one has 4 components (XYZW), 32 * 4 = 128 bits. Now, it will only initialize the specific components that are used on the next stage. Second, it now only initializes the outputs that are consumed on the next stage.

This should make the shader code a bit smaller as it removes those redundant operations, but there is another benefit here. On Pokémon Legends Arceus, the AMD OpenGL driver on Windows complains that some shader uses too many outputs. I'm not sure why it does that, one of the shader where it does this only has 12 outputs on the vertex shader, but anyway, this change also happens to fix this shader compilation error, and allows the models to render on OpenGL:
image

Testing is welcome (to ensure we have no regressions here), especially on Intel and AMD where the initialization is more important.
Here's some games I know that depends on it to render properly:
- Zelda Link's Awakening.
- Monster Hunter Stores 2: Wings of Ruin.
- Hatsune Miku Project DIVA MEGA 39's.
- Luigi's Mansion 3.
- Pokémon Legends Arceus.

I'm also interested in how Pokémon Legends Arceus is running on AMD OpenGL on Windows.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -40,7 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 3132;
+        private const ulong ShaderCodeGenVersion = 3054;
 
         // Progress reporting helpers
         private volatile int _shaderCount;
--- a/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
+++ b/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
@@ -308,7 +308,8 @@ namespace Ryujinx.Graphics.Shader.Decoders
                     int attr = offset + elemIndex * 4;
                     if (attr >= AttributeConsts.UserAttributeBase && attr < AttributeConsts.UserAttributeEnd)
                     {
-                        int index = (attr - AttributeConsts.UserAttributeBase) / 16;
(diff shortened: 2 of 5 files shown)


A-w-x:
does not fix legends arceus on intel opengl (mesa git, i7-1165g7, intel xe)

riperiperi (approved):
LGTM

Z3lcius:
Hi, this will get some uptdate or is the final version? Works just fine whit a old Vulkan version that doesnt works on another Vulkan emulators :)

A-w-x:
linux
intel has been fixed in mesa, the game is rendering correctly now
i tested this pr on latest build and PLA and did not notice regressions but a performance boost

main: village 22fps
pr: village 27-30fps with few drops here and there

there is a perf boost on other games too, tested link's awakening and pokemon sword

AcK77 (approved):

Thog (approved):

PR #3063, opened 2022-01-30
Do not allow render targets not explicitly written by the fragment shader to be modified

gdkchan:
On OpenGL, if you have multiple render targets, but the fragment shader only writes one output, it still writes that output to all bound render targets. This does not match hardware behaviour, which is not writing those render targets. I fixed by forcing the color mask to be disabled for the components that the fragment shader does not write. For this, a new `ShaderInfo` struct is passed to the shader compilation functions on the backend. The struct was added because we will likely need to pass more information in the future (it already does on the Vulkan branch).

The output map handling was simplified on the shader translator. Instead of extracting each bit into a struct with 4 bools, it just stores and use the bitmask directly. This is also more efficient.

Fixes cave in Pokémon Legends Arceus.
Before:
image
After:
image
(Thanks to Muffin for the save).
Also fixes black water on Paper Mario: Origami King. Lines on Pokémon Sword/Shield might be fixed too, but I did not test this one.

Testing is welcome (to ensure nothing regressed etc).

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -1016,7 +1016,7 @@ namespace Ryujinx.Graphics.OpenGL
                 prg.Bind();
             }
 
-            if (prg.HasFragmentShader)
+            if (prg.HasFragmentShader && _fragmentOutputMap != prg.FragmentOutputMap)
             {
                 _fragmentOutputMap = prg.FragmentOutputMap;
 
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -53,6 +53,7 @@ namespace Ryujinx.Graphics.OpenGL
         private ClipOrigin _clipOrigin;
         private ClipDepthMode _clipDepthMode;
 
+        private int _fragmentOutputMap;
         private readonly uint[] _componentMasks;
 
         private uint _scissorEnables;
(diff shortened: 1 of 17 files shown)


Thog (approved):
lgtm

GamerzHell9137:
-Fixed white outlines of objects behind structures in Pokemon Sword and Shield
image
-Fixed boat emblems in Monster Hunter Rise
image

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -1001,18 +1003,30 @@ public void SetPrimitiveTopology(PrimitiveTopology topology)
...
+                prg.Bind();
             }
+
+            if (prg.HasFragmentShader && _fragmentOutputMap != prg.FragmentOutputMap)
+            {
+                _fragmentOutputMap = prg.FragmentOutputMap;
+
+                for (int index = 0; index < Constants.MaxRenderTargets; index++)
+                {
+                    RestoreComponentMask(index);
```
riperiperi, line 1025: Seems a bit wasteful making GL calls to switch color mask for every render target on every program switch. Might benefit from the function remembering the component masks for each render target and skipping calling the GL method if they're the same.

That still doesn't avoid that it's going to _change_ a lot more than before, but I don't think it should cause that much of an issue.

riperiperi (commented):
Looks good, seems to work fine here. Should probably attempt to reduce calls to GL.ColorMask.

riperiperi (approved):
Nice.

PR #3067, opened 2022-01-30
Fix bug that could cause depth buffer to be missing after clear

gdkchan:
Fixes a regression introduced on #2994 that could cause the depth buffer to not be bound after a clear. The issue would happen if the game clears the color buffer after depth, and `clipSizeMismatch` is true. Additionally, make the clip region size take multisampled textures into account. Before, the size was incorrect for multisample textures, which would cause it to detect false mismatches.

Fixes regression where models would not be draw on Sonic Colors Ultimate. Thanks to Muffin for finding the regression.


GamerzHell9137:
Sonic Colors is rendering properly now. Gj gdk!
image

Thog (approved):

riperiperi (approved):

PR #3069, opened 2022-01-31
Add support for alpha to coverage dithering

gdkchan:
This implements alpha to coverage support. This feature can be used with multisampling to discard some samples based on the fragment alpha value, in order to make the pixel partially or fully transparent. In addition to that, NVIDIA also supports alpha to coverage dithering, that improves the perceived transparency levels by discarding more or less pixels depending on the screen coordinates of the pixel, and the alpha value. Even though this is intended to be used with multisampling, there's actually nothing on hardware that prevents this from being used with a regular render target with a single sample. Pokémon Legends Arceus (ab)uses this behavior to create a dithering effect for objects that are close to the camera.

Unfortunately, OpenGL does not support alpha to coverage with non-multisampled framebuffers (it just does nothing in this case), so for those cases, I had to do the implementation on the shader. On top of that, only NVIDIA supports the dithering, so it would not work for other vendors anyway.

This was implemented on a shader, by discarding pixels based on the value of the alpha component of the first output (0), and `FragCoord`. It uses shader specialization to generate a version of the shader with or without the dithering, based on the GPU state. This should prevent it from having shader performance impact for draws that do not use this feature (which should be most of them.

Fixes missing dithering on Pokémon Legends Arceus.
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -59,10 +59,6 @@ namespace Ryujinx.Graphics.OpenGL
 
         private uint _scissorEnables;
 
-        private bool _msAlphaToCoverageEnable;
-        private bool _msAlphaToCoverageDitherEnable;
-        private bool _fbMultisampled;
-
         private bool _tfEnabled;
         private TransformFeedbackPrimitiveType _tfTopology;
 
@@ -924,11 +920,6 @@ namespace Ryujinx.Graphics.OpenGL
 
         public void SetMultisampleState(MultisampleDescriptor multisample)
         {
-            _msAlphaToCoverageEnable = multisample.AlphaToCoverageEnable;
-            _msAlphaToCoverageDitherEnable = multisample.AlphaToCoverageDitherEnable;
-
-            UpdateAlphaToCoverage();
-
             if (multisample.AlphaToCoverageEnable)
             {
                 GL.Enable(EnableCap.SampleAlphaToCoverage);
@@ -1111,7 +1102,6 @@ namespace Ryujinx.Graphics.OpenGL
             EnsureFramebuffer();
 
             bool isBgraChanged = false;
-            bool fbMultisampled = false;
 
             for (int index = 0; index < colors.Length; index++)
             {
@@ -1130,19 +1120,7 @@ namespace Ryujinx.Graphics.OpenGL
 
                         RestoreComponentMask(index);
                     }
-
-                    if (color.Target == Target.Texture2DMultisample ||
-                        color.Target == Target.Texture2DMultisampleArray)
-                    {
-                        fbMultisampled = true;
-                    }
                 }
-            }
-
-            if (_fbMultisampled != fbMultisampled)
-            {
-                _fbMultisampled = fbMultisampled;
-                UpdateAlphaToCoverage();
             }
 
             if (isBgraChanged)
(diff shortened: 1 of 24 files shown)


Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
```diff
@@ -219,6 +222,37 @@ public void PrepareForReturn()
...
+
+            // 11 11 11 10 10 10 10 00
+            // 11 01 01 01 01 00 00 00
+            Operand ditherMask = Const(unchecked((int)0xfbb99110u));
+
+            Operand a2cDitherEndLabel = Label();
+
+            this.BranchIfFalse(a2cDitherEndLabel, Attribute(AttributeConsts.FragmentAlphaToCoverageDither));
+
+            Operand x = this.BitwiseAnd(this.FP32ConvertToU32(Attribute(AttributeConsts.PositionX)), Const(1));
```
riperiperi, line 241: How will this work on a multisampled target? Are we going to rely on host alpha to coverage for multisampling?

gdkchan: Yes, right now it only does this for non-multisampled targets, because alpha to coverage should work with the multisampled ones on OpenGL (although, if the NVIDIA alpha to coverage dithering extension is not supported, it most likely won't have dithering). I'm not sure if that can be emulated on the fragment shader, there is an extension that allows controlling the coverage mask (https://www.khronos.org/registry/OpenGL/extensions/NV/NV_sample_mask_override_coverage.txt), but that too is NVIDIA specific afaik.

gdkchan:
Rebased and ready for review (and re-testing). Now uses shader specialization rather than the support buffer.

Ryujinx.Graphics.OpenGL/Pipeline.cs
```diff
@@ -918,6 +918,34 @@ public void SetLogicOpState(bool enable, LogicalOp op)
...
+                if (multisample.AlphaToOneEnable)
+                {
+                    GL.Enable(EnableCap.SampleAlphaToOne);
+                }
+                else
+                {
+                    GL.Disable(EnableCap.SampleAlphaToOne);
+                }
+
+                if (HwCapabilities.SupportsAlphaToCoverageDitherControl)
```
riperiperi, line 936: How does this interact with the case in PLA where the target is not multisampled, but alpha to coverage is enabled? Will it do the NV behaviour and the shader discard at the same time, or is there something stopping it from doing that?

gdkchan: On OpenGL, alpha-to-coverage with a framebuffer that is not multisampled has no effect.
From the spec:
```
17.3.1 Alpha To Coverage
This step modifies fragment alpha and coverage values based on the values of
SAMPLE_ALPHA_TO_COVERAGE and SAMPLE_ALPHA_TO_ONE. If MULTISAMPLE
is disabled, if the value of SAMPLE_BUFFERS is not one, or if draw buffer zero is
not NONE and the buffer it references has an integer format, these operations are
skipped.
Alpha to coverage and alpha to one operations are enabled or disabled
by calling Enable and Disable with targets SAMPLE_ALPHA_TO_COVERAGE or
SAMPLE_ALPHA_TO_ONE, respectively.
```
Vulkan has no such limitation though so it could conflict with the shader implementation if the same thing was done there.

riperiperi: OK, I guess it will be more of a concern when rebasing Vulkan. I guess technically, if the dithering pattern is identical it will remove the same pixels, so maybe it will just end up punching the same holes twice.

riperiperi (approved):

Ryujinx.Graphics.OpenGL/Pipeline.cs
gdkchan: I'm not sure how it should be handled on Vulkan. AFAIK there's no way to explicitly enable the dithering on Vulkan. It does happen to be enabled by default on NVIDIA, but that's an implementation detail and it wouldn't work for other vendors. Maybe we can expose a capability that would be true if the vendor is NVIDIA, where it would not add the emulation code on the shader. But that's not a safe assumption since NVIDIA could decide to change the defaults, and maybe there are some GPUs that don't support it.

Thog (approved):

PR #3079, opened 2022-02-03
Adjusting how deadzones are calculated

### LINKED ISSUE #3074: controller deadzone feels wrong

controller deadzone implementation isn't working properly. It currently checks x and y independently
https://github.com/Ryujinx/Ryujinx/blob/686757105c83a1485117ff5f5d30b2fad4df39b1/Ryujinx.Input/HLE/NpadController.cs#L392-L399

This has the problem that when inputting one of the cardinal directions it's impossible to do small adjustments on the other axis.

Ideally the implementation should only check if the length is smaller than the deadzone, something like this (not tested)
```c#
private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone) 
{ 
    if (x * x + y * y < deadzone * deadzone)
        return new JoystickPosition 
        { 
            Dx = 0, 
            Dy = 0 
        }; 
    else
        return new JoystickPosition 
        { 
            Dx = ClampAxis(x), 
            Dy = ClampAxis(y) 
        }; 
} 
```

Here is also an article describing the problem in more detail  
https://www.gamedeveloper.com/disciplines/doing-thumbstick-dead-zones-right

skrekhere:
Previously, the way that deadzones were calculated was that if the x axis is smaller than the set value of the deadzone, then X would be 0. same goes for Y. This works for a select few games, but say for example, in some game you're moving forward with the stick pressed all the way forward. if you wanted to turn say 5 degrees to the right, then it wouldn't be possible, as the deadzone would entirely prevent small adjustments on the edge of another axis. With this new implementation, it creates what is essentially a dead circle in the centre, and then smooths the edges from there. This both allows for small adjustments at the extreme ends of a specific axis, but also prevents the problem of a sudden burst of movement once you're past the deadzone, because of there being no "sharp" edges.

Also added .idea/* to the gitignore because I'm sure more than just me uses rider for development, and it makes life easier if it's just there.

Should fix #3074

### DIFF
--- a/Ryujinx.Input/HLE/NpadController.cs
+++ b/Ryujinx.Input/HLE/NpadController.cs
@@ -391,10 +391,15 @@ namespace Ryujinx.Input.HLE
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
         {
-            return new JoystickPosition
+            float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
+            if (magnitudeClamped < deadzone) return new JoystickPosition()
             {
-                Dx = ClampAxis(MathF.Abs(x) > deadzone ? x : 0.0f),
-                Dy = ClampAxis(MathF.Abs(y) > deadzone ? y : 0.0f)
+                Dx = 0, Dy = 0
+            };
+            return new JoystickPosition()
+            {
+                Dx = ClampAxis((x / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
+                Dy = ClampAxis((y / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone)))
             };
         }
 
--- a/Ryujinx.Input/HLE/NpadController.cs
+++ b/Ryujinx.Input/HLE/NpadController.cs
@@ -392,10 +392,12 @@ namespace Ryujinx.Input.HLE
         private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
         {
             float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
-            if (magnitudeClamped <= deadzone) return new JoystickPosition()
+            
+            if (magnitudeClamped <= deadzone)
             {
-                Dx = 0, Dy = 0
-            };
+                return new JoystickPosition() {Dx = 0, Dy = 0};
+            }
+            
             return new JoystickPosition()
             {
                 Dx = ClampAxis((x / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
--- a/Ryujinx.Input/HLE/NpadController.cs
+++ b/Ryujinx.Input/HLE/NpadController.cs
@@ -392,14 +392,14 @@ namespace Ryujinx.Input.HLE
         private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
         {
             float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
-            if (magnitudeClamped < deadzone) return new JoystickPosition()
+            if (magnitudeClamped <= deadzone) return new JoystickPosition()
             {
                 Dx = 0, Dy = 0
             };
             return new JoystickPosition()
             {
-                Dx = ClampAxis((x / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
-                Dy = ClampAxis((y / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone)))
+                Dx = ClampAxis((x / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
+                Dy = ClampAxis((y / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone)))
             };
         }
 
--- a/Ryujinx.Input/HLE/NpadController.cs
+++ b/Ryujinx.Input/HLE/NpadController.cs
@@ -391,7 +391,7 @@ namespace Ryujinx.Input.HLE
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
         {
-            float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
+            float magnitudeClamped = Math.Min(MathF.Sqrt(x * x + y * y), 1f);
             
             if (magnitudeClamped <= deadzone)
             {
@@ -408,14 +408,12 @@ namespace Ryujinx.Input.HLE
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static short ClampAxis(float value)
         {
-            if (value <= -short.MaxValue)
+            if (Math.Sign(value) < 0)
             {
-                return -short.MaxValue;
+                return (short)Math.Max(value * -short.MinValue, short.MinValue);
             }
-            else
-            {
-                return (short)(value * short.MaxValue);
-            }
+
+            return (short)Math.Min(value * short.MaxValue, short.MaxValue);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,9 @@ _TeamCity*
 # DotCover is a Code Coverage Tool
 *.dotCover
 
+# Rider is a Visual Studio alternative
+.idea/*
+
 # NCrunch
 *.ncrunch*
 .*crunch*.local.xml


MutantAura:
I'd suggest a more descriptive PR title and description here.
Something like "Adjustments to deadzone calculation and clamping" and a brief explanation of what you're changing and how it is different/better than the old method. Nothing crazy long but it helps the reviewers and also anyone reading :)

MelonSpeedruns:
Could you push again so the build can be triggered? Looks like something happened to the building system last night and it messed up building.

Vegita2:
I made a deadzone visualizer some time ago and added the current implementation and the one from this pr to it.
https://vigilant-beaver-7e2f51.netlify.app/
(there is a select input at the bottom left)

color ranges from black to white depending on stick input value
black = no input
white =  fully pressed
red circle = deadzone

Basically the difference is this:

Current implementation


Pull request implementation

MutantAura:
Some quick tests performed at deadzone of 0.5 in BoTW

Master:
https://user-images.githubusercontent.com/44103205/152690279-dc835f16-4721-4611-9b6d-59ee457c627c.mp4

PR:
https://user-images.githubusercontent.com/44103205/152690303-45dc9097-67c9-4480-bd0c-f6e1038cc769.mp4

There is much more fine control in the new calculation and removes the almost 8-axis effect a high deadzone would previously apply in the past (See master video). No other abnormalities noticed. Looking good from my perspective.

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,10 +391,15 @@ public GamepadInput GetHLEInputState()
...
+            float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
+            if (magnitudeClamped < deadzone) return new JoystickPosition()
+            {
+                Dx = 0, Dy = 0
+            };
+            return new JoystickPosition()
             {
-                Dx = ClampAxis(MathF.Abs(x) > deadzone ? x : 0.0f),
-                Dy = ClampAxis(MathF.Abs(y) > deadzone ? y : 0.0f)
+                Dx = ClampAxis((x / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
```
gdkchan, line 401:
```suggestion
                Dx = ClampAxis((x / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
```

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,10 +391,15 @@ public GamepadInput GetHLEInputState()
...
+                Dy = ClampAxis((y / (magnitudeClamped)) * ((magnitudeClamped - deadzone) / (1 - deadzone)))
```
gdkchan, line 402:
```suggestion
                Dy = ClampAxis((y / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone)))
```

Ryujinx.Input/HLE/NpadController.cs
gdkchan: Can't magnitude be zero here? Is it fine to divide by zero in this case?

skrekhere: Magnitude should only be zero when x and y are both 0, and in my testing, dividing by zero is fine if it happens. Could add checks for when deadzones are applied to not apply them when both axes are zero, to be safe.

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,10 +391,15 @@ public GamepadInput GetHLEInputState()
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static JoystickPosition ApplyDeadzone(float x, float y, float deadzone)
         {
-            return new JoystickPosition
+            float magnitudeClamped = Math.Clamp(MathF.Sqrt(x * x + y * y), -1f, 1f);
+            if (magnitudeClamped <= deadzone) return new JoystickPosition()
```
AcK77, line 395: Add bracket over the `if` content and put the return on a new line.

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,10 +391,15 @@ public GamepadInput GetHLEInputState()
...
+            {
+                Dx = 0, Dy = 0
+            };
+            return new JoystickPosition()
```
AcK77, line 399: Add a new extra empty line before

AcK77 (changes requested):

Ryujinx.Input/HLE/NpadController.cs
skrekhere: done:)

Ryujinx.Input/HLE/NpadController.cs
skrekhere: and done

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,10 +391,17 @@ public GamepadInput GetHLEInputState()
...
+            if (magnitudeClamped <= deadzone)
+            {
+                return new JoystickPosition() {Dx = 0, Dy = 0};
+            }
+            
+            return new JoystickPosition()
             {
-                Dx = ClampAxis(MathF.Abs(x) > deadzone ? x : 0.0f),
-                Dy = ClampAxis(MathF.Abs(y) > deadzone ? y : 0.0f)
+                Dx = ClampAxis((x / magnitudeClamped) * ((magnitudeClamped - deadzone) / (1 - deadzone))),
```
riperiperi, line 403: I'm not really a fan of the potential divide by zero. ClampAxis seems like it is working by a fluke, as float.NaN casts to 0 as short. If the behaviour of that function changes, this might break things in an unexpected way.

skrekhere: There's no more potential of a divide by zero. that changed when I changed the < to a <= in the first check, before the ClampAxis call entirely. The only potential of a divide by zero before the change was when the magnitude is zero, with the set also being zero, but now a divide by zero is only possible with negative deadzone values, which is a whole other problem if that's allowed to happen.

german77: If deadzone is 1 you also have a divide by zero

german77: Ah nevermind we can't never reach this situation since magnitude is already clamped to 1. Another thing to take in mind is that you don't need to clamp negative values. Since (x*x + y*y) is guaranteed to be positive

gdkchan: Speaking about that, `ClampAxis` appears to be slightly wrong. It would be nice to take a look too, since you're working on this.

riperiperi: Ah, I didn't see the condition above that returns early. It should also trigger if deadzone is >= 1, so no concerns.

And yeah, `ClampAxis` is incorrect right now, it doesn't clamp positive and the negative clamp is wrong.

skrekhere: Latest commit should fix everything up :)

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,24 +391,29 @@ public GamepadInput GetHLEInputState()
...
         {
-            if (value <= -short.MaxValue)
+            if (Math.Sign(value) < 0)
             {
-                return -short.MaxValue;
-            }
-            else
-            {
-                return (short)(value * short.MaxValue);
+                return (short) Math.Max(value * -short.MinValue, short.MinValue);
```
gdkchan, line 413:
```suggestion
                return (short)Math.Max(value * -short.MinValue, short.MinValue);
```

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,24 +391,29 @@ public GamepadInput GetHLEInputState()
...
             }
+
+            return (short) Math.Min(value * short.MaxValue, short.MaxValue);
```
gdkchan, line 416:
```suggestion
            return (short)Math.Min(value * short.MaxValue, short.MaxValue);
```

Ryujinx.Input/HLE/NpadController.cs
gdkchan: Kinda wonder if -1.0 is supposed to map to -32768. That would make the range uneven sinnice short.MaxValue is 32767.

skrekhere:
fixed up that spacing

Ryujinx.Input/HLE/NpadController.cs
```diff
@@ -391,24 +391,29 @@ public GamepadInput GetHLEInputState()
...
         {
-            if (value <= -short.MaxValue)
+            if (Math.Sign(value) < 0)
             {
-                return -short.MaxValue;
-            }
-            else
-            {
-                return (short)(value * short.MaxValue);
+                return (short)Math.Max(value * -short.MinValue, short.MinValue);
```
riperiperi, line 413: This is a bit weird, but it seems to work as intended?
`-short.MinValue` doesn't fit in a short, it will just equal `-32768` again. But here, it seems to negate as int32, then multiply the float, so it's 32768f.

riperiperi (approved):
Nice work, good to see this weird behaviour sorted.

Issue #2287, opened 2021-05-14
[Feature Request] Provide ability to set y/x axis of sticks independently (or add "rotate" pad option)


legend800:
#### Overview:
Provide ability to set y/x axis of sticks independently (with invert) or provide ability to "rotate" pad. 

#### Smaller Details: 
Dolphin example:
![image](https://user-images.githubusercontent.com/20983844/118229571-13068380-b441-11eb-995f-e91bf4fb5707.png)

Related Discord discussion here with devs:
https://discord.com/channels/410208534861447168/814263228926132274/832109061687869450

#### Nature of Request:  
Allow more flexibility to map stick input as the user wishes. Example input from the other emu:
![2021-05-13 23_28_27-yuzu Configuration](https://user-images.githubusercontent.com/20983844/118230718-f4a18780-b442-11eb-822b-b12a1dbd8d27.png)

### Why would this feature be useful?
Games like Super Mario Party need sideways joycon. You could use the Pro Controller to play this game perfectly, with motion, previous to Apr 14th, but now you can't set x/y independently. This prevent you from mapping input to achieve 'sideway joycon'.

Before:
![2](https://user-images.githubusercontent.com/20983844/118230059-e2731980-b441-11eb-90f6-8da2e097d1ef.png)

After:
![1](https://user-images.githubusercontent.com/20983844/118230062-e43cdd00-b441-11eb-9677-a76d8cf0d2b8.png)

jojorne:
I need the ability to set the "speed" that you move the stick buttons. When you use it mapped on a keyboard the current speed is set to 1 and  using it to control things like the camera become quite a challenge since the camera can move quite a lot.

legend800:
Any chance we can see an update to the left stick mapping flexibility soon, so users can use the controllers they're comfortable with, like Pro pad over the cramp-inducing joycon? All we need is a 'sideways' or go back to being able to set each direction like other emus.

Hackerpcs:
+1 on this. I use a cheap PS2-type controller than on the right analog stick, the axis are reverted, the vertical one functions as horizontal and the horizontal as vertical, reverting the left to be right e.g. on the horizontal one doesn't help, it's a different problem. I get can get around with X360CE but it would be nice to have it built in

legend800:
Yep, Nintendo allows for more stick customization than Ryu since it got locked down last April. I really hope they can readd this back or provide some kind of flexibility to the users who do not want to use tiny joycons. 

[...]

PR #3084, opened 2022-02-05
Implement rotate stick 90 degrees clockwise

darko1979:
One solution to feature request #2287
Instead of mapping axes independently I added simple option to swap X/Y sticks.

NOTE: I encountered an issue that probably needs fixing. Inverting axes options work properly if we select Pro Controller or Joycon Pair for controller type, but if we select Joycon Right, invert options for left sticks are actually used for right stick.

### DIFF
--- a/Ryujinx.Common/Configuration/Hid/Controller/JoyconConfigControllerStick.cs
+++ b/Ryujinx.Common/Configuration/Hid/Controller/JoyconConfigControllerStick.cs
@@ -5,7 +5,7 @@
         public Stick Joystick { get; set; }
         public bool InvertStickX { get; set; }
         public bool InvertStickY { get; set; }
-        public bool SwapSticks { get; set; }
+        public bool Rotate90CW { get; set; }
         public Button StickButton { get; set; }
     }
 }
--- a/Ryujinx.Common/Configuration/Hid/Controller/JoyconConfigControllerStick.cs
+++ b/Ryujinx.Common/Configuration/Hid/Controller/JoyconConfigControllerStick.cs
@@ -5,6 +5,7 @@
         public Stick Joystick { get; set; }
         public bool InvertStickX { get; set; }
         public bool InvertStickY { get; set; }
+        public bool SwapSticks { get; set; }
         public Button StickButton { get; set; }
     }
 }
--- a/Ryujinx.Headless.SDL2/Program.cs
+++ b/Ryujinx.Headless.SDL2/Program.cs
@@ -220,7 +220,7 @@ namespace Ryujinx.Headless.SDL2
                             StickButton  = ConfigGamepadInputId.LeftStick,
                             InvertStickX = false,
                             InvertStickY = false,
-                            SwapSticks   = false,
+                            Rotate90CW   = false,
                         },
 
                         RightJoycon = new RightJoyconCommonConfig<ConfigGamepadInputId>
@@ -242,7 +242,7 @@ namespace Ryujinx.Headless.SDL2
                             StickButton  = ConfigGamepadInputId.RightStick,
                             InvertStickX = false,
                             InvertStickY = false,
-                            SwapSticks   = false,
+                            Rotate90CW   = false,
                         },
 
                         Motion = new StandardMotionConfigController
--- a/Ryujinx.Headless.SDL2/Program.cs
+++ b/Ryujinx.Headless.SDL2/Program.cs
@@ -220,6 +220,7 @@ namespace Ryujinx.Headless.SDL2
                             StickButton  = ConfigGamepadInputId.LeftStick,
                             InvertStickX = false,
                             InvertStickY = false,
+                            SwapSticks   = false,
                         },
 
                         RightJoycon = new RightJoyconCommonConfig<ConfigGamepadInputId>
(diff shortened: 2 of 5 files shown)


legend800:
Awesome work! Thanks so much for picking this up. Verified it works well on a Pro Controller. 

2 issues to note:

1) The Swap is properly rotating U/D 90 degrees clockwise, but it does L/R counterclockwise, forcing me to also Invert X in the config to achieve a desire rotation. Both should just be clockwise to be consistent. See pic if it helps.


2) With #1 done, you can now change option to 'Rotate axis 90". You can also use the term "Swap" if you want but use axis instead of sticks as right now it sounds like you're saying you're swapping left and right sticks. :)

darko1979:
@legend800 the option "swap sticks" is just swapping input values from X and Y axis of analog stick bound to joycon stick. There is no rotation happening. The code for that is very simple. To rotate a stick you will have to combine it with invert option for one of the sticks to achieve desired effect. I did not want to create option for rotate because then I would need to create options for rotating 90 clockwise or anti-clockwise.
As for naming the option, because invert options are named "Invert stick X" and "Invert stick Y" I thought naming "swap sticks" would be sufficient to understand what this option does without making it too long as, for example "swap sticks X and Y".

gdkchan:
The name "Swap Sticks" seems misleading since it does not actually swaps the sticks, it swaps the axis of a single stick.

MutantAura:
Overall I'm not convinced I ever like the idea behind the invert X/Y and this PR too much. It feels very much like a bandaid instead of mapping Up, Down, Left and Right separately as it would massively reduce clutter and confusing checkbox names.

However given that is not a super simple job I would instead consider a different name. "Swap sticks" sounds like it literally swaps the L and R stick around which of course is not what is happening. Maybe "Swap Axis" would be a more descriptive name as that is what it is actually doing.

gdkchan:
> Overall I'm not convinced I ever like the idea behind the invert X/Y and this PR too much. It feels very much like a bandaid instead of mapping Up, Down, Left and Right separately as it would massively reduce clutter and confusing checkbox names.

I guess if they were mapped separately (as in, a option to map analog up, down, left and right separately), it would need to auto-fill the other 3 boxes after the first direction is pressed (for example, if you press up on analog up, it would auto-fill right as analog right, down as analog down, etc. Similarly, if you press up as analogic right, it would auto-fill right as analog down, down as analog left and left and analog up). Otherwise, it would allow completely invalid mappings. If it was done like this, it would be just a fancy way to define a rotation of 0, 90, 180 or 270 degrees.

legend800:
Agreed, I would much prefer simply going back to how it was last year and allowing users to map each direction like every other emu out there. 

However, I'll take what I can get and this PR can provide a good solution but it just needs some tweaking. With some feedback addressed, I think it'll work great. 

1) I highly recommend you switch L/R to be clockwise vs. doing them differently. The MAIN use case is to rotate the controller from vertical hold to side hold. There's only one to do that. Everything rotates 90 clockwise. 

2) As explained by me and others, there's 2 sticks. Left analog stick and Right analog stick. We can't say swap sticks. We can say 'Swap Axis' and everyone would understand. After thinking about it, I might also suggest taking the Dolphin approach and go with "Sideways". I'm still a fan of 'Rotate' also. 

2022-02-06 09_30_21-Window

MutantAura:
> I guess if they were mapped separately (as in, a option to map analog up, down, left and right separately), it would need to auto-fill the other 3 boxes after the first direction is pressed (for example, if you press up on analog up, it would auto-fill right as analog right, down as analog down, etc. Similarly, if you press up as analogic right, it would auto-fill right as analog down, down as analog left and left and analog up). Otherwise, it would allow completely invalid mappings. If it was done like this, it would be just a fancy way to define a rotation of 0, 90, 180 or 270 degrees.

Yea this is likely what would need to happen. I also feel like a right click context menu would reduce clutter for invert and rotate options even if the 4 axis idea is dropped. "Invert" and "Rotate" in a context menu with "X" and "Y" in Invert and "0", "90", "180" and "270" in the rotate menu. Then if any modifier were applied it could simply add an asterisks to the end of the "LStick" name to let a user know it was an altered stick.

Stole menu from photoshop but it illustrates the point that I think there are better ways to do this than add more tickboxes.
controller

legend800:
I'll take a couple visible dropdowns over a whole new menu system that users have to "stumble upon" by trying to Right Click with ambiguous asterisks. Options don't = clutter unless it's redundant/unnecessary which this is not (yet). :)

My vote is to keep it standard and visible in UI. Also, to not have this PR get stuck in purgatory, it should be identified what's the minimum v1 to merge that would not require rework for a future v2 to expand to a full perfect set of functionality. Right now, we just need a way to rotate 90 clockwise. Let's focus on that to unblock users to play games with Pro Controllers while using a forward-looking design.

MutantAura:
Hmm you're probably right in that this discussion isn't really within the scope of the PR I was just spitballing for what probably needs to happen in the future. 

Either way to get back on track the name likely needs to be addressed to refer directly to the axis' of the sticks not the sticks themselves.

darko1979:
I changed the swap sticks to rotate 90 clockwise. I hope I got it right :-)

legend800:
Awesome, thanks for the updates! Looks and works great here on several games I've tried. No issues. Well done!

2022-02-08 15_32_00-Ryujinx - Controller Settings - Player1

riperiperi (approved):
Looks good. I guess you could get counter clockwise by inverting stick direction?

legend800:
@riperiperi Correct, with Rotate and Invert checked you get 90 instead of 270. So between all the options, you can finally achieve 0, 90, 180, 270 where only 0 and 180 were possible before this. 

Just need one more reviewer to merge.

Thog (approved):

PR #3096, opened 2022-02-07
Added trace log level

### LINKED ISSUE #1547: Add a Trace log level
## Feature Request

Trace log level.

#### Overview:

We should add a new "Trace" log level that would be used for things that needs a high amount of logs, and would be normally undesirable to enable unless the person is trying to diagnose a specific issue.

Example of things that should use such "Trace" level are the kernel syscall logs and IPC call logs.

This will make the debug log level more useful, as we can print some debug information with it without slowing down the guest applications a lot and without spamming the console and generating a huge log file.

mlgatto:
Added support for trace logs - fixes #1547

### DIFF
--- a/Ryujinx.Common/Logging/Logger.cs
+++ b/Ryujinx.Common/Logging/Logger.cs
@@ -90,6 +90,7 @@ namespace Ryujinx.Common.Logging
         public static Log? Guest     { get; private set; }
         public static Log? AccessLog { get; private set; }
         public static Log? Stub      { get; private set; }
+        public static Log? Trace     { get; private set; }
         public static Log  Notice    { get; } // Always enabled
 
         static Logger()
@@ -117,6 +118,7 @@ namespace Ryujinx.Common.Logging
             Error = new Log(LogLevel.Error);
             Warning = new Log(LogLevel.Warning);
             Info = new Log(LogLevel.Info);
+            Trace = new Log(LogLevel.Trace);
         }
 
         public static void RestartTime()
@@ -172,7 +174,7 @@ namespace Ryujinx.Common.Logging
 
         public static IReadOnlyCollection<LogLevel> GetEnabledLevels()
         {
-            var logs = new Log?[] { Debug, Info, Warning, Error, Guest, AccessLog, Stub };
+            var logs = new Log?[] { Debug, Info, Warning, Error, Guest, AccessLog, Stub, Trace };
             List<LogLevel> levels = new List<LogLevel>(logs.Length);
             foreach (var log in logs)
             {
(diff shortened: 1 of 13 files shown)


gdkchan:
Can you also change:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Kernel/SupervisorCall/SyscallTable.cs#L482
and:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Services/IpcService.cs#L120
To use the new trace log level instead of debug?

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -196,6 +198,7 @@ public static void SetEnable(LogLevel logLevel, bool enabled)
                 case LogLevel.Guest     : Guest     = enabled ? new Log(LogLevel.Guest)    : new Log?(); break;
                 case LogLevel.AccessLog : AccessLog = enabled ? new Log(LogLevel.AccessLog): new Log?(); break;
                 case LogLevel.Stub      : Stub      = enabled ? new Log(LogLevel.Stub)     : new Log?(); break;
+                case LogLevel.Trace     : Trace     = enabled ? new Log(LogLevel.Trace) :    new Log?(); break;
```
AcK77, line 201: Alignment is off here

AcK77 (changes requested):
You have to update the config file to add the new field, even if it’s disabled as default!

Pixxel123:
If it's not going to be enabled by default, I'd probably move the UI toggle to be at the bottom of the list, after Guest Logs and before Fs Access Logs to make it easier for users to have just the defaults on.

mlgatto:
> You have to update the config file to add the new field, even if it’s disabled as default!

@AcK77 Where can I find the config file?

gdkchan:
> > You have to update the config file to add the new field, even if it’s disabled as default!
> 
> @AcK77 Where can I find the config file?

This shouldn't be needed anymore, the default config file was removed some months ago.

mlgatto:
> I mean you still have to update the code which update the config file is needed, see https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx/Configuration/ConfigurationState.cs#L978 as example.

@AcK77 sorry I'm a bit lost, I think I already have updated that section of the code. Can you elaborate a bit more what is still needed? Thanks :)

jduncanator:
> > I mean you still have to update the code which update the config file is needed, see https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx/Configuration/ConfigurationState.cs#L978 as example.
> 
> @AcK77 sorry I'm a bit lost, I think I already have updated that section of the code. Can you elaborate a bit more what is still needed? Thanks :)

You need to add a migration from version 35 to your new version. Even though the migration doesn't really need to do anything explicitly (the default value of a bool is false after all), you still need to do a migration to trigger Ryujinx to reserialize the configuration file on first load, bump the version number, and add the new trace log level option to the configuration file.

gdkchan (approved):
lgtm, thanks.

Thog (approved):

riperiperi (commented):
Nice, will be good having a trace log level. I don't see anything in here to update the dialog that appears when trying to start a game with Debug logs, though. 

The UI has a dialog that pops up and essentially asks the user to disable them due to potential performance impact - but the majority of these logs have been moved to the trace level. I think the warning should be instead for that level, and also the placement of the checkbox in settings might be worth moving next to debug too.

AcK77:
Could you please address the comments of Riperiperi then we can merge it ;)!

riperiperi (approved):

PR #3097, opened 2022-02-07
Add a limit on the number of uses a constant may have

gdkchan:
The `UsesCount` field is limited to 16-bits, so it can overflow if there are more than 65535 uses. The optimization that reuses constant was causing this overflow to happen on a guest function with a lot of memory operations on the same basic block. This change adds a limit on the number of uses a constant may have before it adds another fresh mov, thus avoiding the problem.

The issue was observed on Deathsmiles II. The game is working now with this change.
image


AcK77 (approved):

riperiperi (approved):
LGTM.

PR #3103, opened 2022-02-10
Ship SoundIO library only for the specified runtime

### LINKED ISSUE #2399: Platform specific files for libsoundio are all provided with the build

Since #2211 is merged, we can still cleanup it a bit more, we currently pack libsoundio dependencies in the repo directly, which lead, as example, to get the `.so` and the `.dylib` files in the windows build zip.

Using a nuget package could fixes the issue by return the platform specific file.

edisionnano:
On the project file for the project ```Ryujinx.Audio.Backends.SoundIo``` we specify which SoundIO library will be published depending on the RuntimeIdentifier property; if it isn't linux-x64 or osx-64 it is Windows therefore it gets a .dll, if it isn't linux-x64 or win-x64 then it's OS X so it gets a .dylib and if it isn't win-x64 or osx-64 it has to be Linux so it gets a .so file. And these checks work fine when building the project as an individual project. So if you do for example ```dotnet build Ryujinx.Audio.Backends.SoundIo/Ryujinx.Audio.Backends.SoundIo.csproj -o soundio -r osx-x64``` the generated soundio folder will contain a .dylib but not a .dll or .so file as it should. However, when building Ryujinx as a whole and ```Ryujinx.Audio.Backends.SoundIo``` is built as a subproject the property isn't passed to MSBuild. In order for it to work we have to add the supported RuntimeIdentifiers on the PropertyGroup.

Closes #2399

### DIFF
--- a/Ryujinx.Audio.Backends.SoundIo/Ryujinx.Audio.Backends.SoundIo.csproj
+++ b/Ryujinx.Audio.Backends.SoundIo/Ryujinx.Audio.Backends.SoundIo.csproj
@@ -3,9 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
-    <RuntimeIdentifiers>win-x64</RuntimeIdentifiers>
-    <RuntimeIdentifiers>osx-x64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;linux-x64;osx-x64</RuntimeIdentifiers>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


Ryujinx.Audio.Backends.SoundIo/Ryujinx.Audio.Backends.SoundIo.csproj
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>
```
gdkchan, line 6: I think you can just do `<RuntimeIdentifiers>>win-x64;linux-x64;osx-x64</RuntimeIdentifiers>` on a single line.

edisionnano: > I think you can just do `<RuntimeIdentifiers>>win-x64;linux-x64;osx-x64</RuntimeIdentifiers>` on a single line.

Yeah I know, I thought separating them looked better tho

gdkchan: That's what we do in other projects. Would be nice to do the same here for consistency.

edisionnano: Resolved

gdkchan (approved):
lgtm, thanks.

Thog (approved):
A better way would be to do a proper nugget package but this will do for now

PR #3104, opened 2022-02-10
InstEmitMemory32: Literal loads always have word-aligned PC

merryhime:
There are a few times when reading from R15 is very slightly different from normal and returns a word-aligned PC instead:

1. During a literal load.
2. During a two-word thumb BLX instruction.
3. During an ADR instruction.
4. During a literal store (the only one which isn't UNPREDICTABLE is a VFP store, afaik.)

Thus for the above special cases we need to have a different register read function.

The reason this is an issue is because the value of PC which is exposed to the memory addressing circuitry is always word aligned.

The effect of this change will only be noticeable once thumb support is implemented.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitHelper.cs
@@ -47,6 +47,20 @@ namespace ARMeilleure.Instructions
             }
         }
 
+        public static Operand GetIntA32AlignedPC(ArmEmitterContext context, int regIndex)
+        {
+            if (regIndex == RegisterAlias.Aarch32Pc)
+            {
+                OpCode32 op = (OpCode32)context.CurrOp;
+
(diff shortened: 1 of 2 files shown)


ARMeilleure/Instructions/InstEmitHelper.cs
```diff
@@ -47,6 +47,20 @@ public static Operand GetIntA32(ArmEmitterContext context, int regIndex)
             }
         }
 
+        public static Operand GetIntA32_AlignedPC(ArmEmitterContext context, int regIndex)
```
gdkchan, line 50: I'd call it `GetIntA32AlignedPC` since we don't use underscores on function names.

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
LGTM.

PR #3105, opened 2022-02-10
ARMeilleure: Thumb support (All T16 instructions)

merryhime:
TODO:
- [x] Complete all 16-bit thumb instructions
- [x] ~~Modify reading of instructions to avoid unaligned reads.~~ _Looks like Ryujinx already handles this case reasonably well_.
- [x] Generate some comparative tests with another implementation.

---

Completed implementation of all 16-bit thumb instructions, including some randomized tests.


merryhime:
Ready for r/v when able

ARMeilleure/Decoders/Decoder.cs
```diff
@@ -195,21 +195,31 @@ public static bool BinarySearch(List<Block> blocks, ulong address, out int index
             ulong          limitAddress)
         {
             ulong address = block.Address;
+            int inITBlock = 0;
```
gdkchan, line 198: Maybe it should be called `itBlockSize`, since apparently that's what the variable really is.

ARMeilleure/Decoders/OpCodeT16AddSubSp.cs
```diff
@@ -0,0 +1,21 @@
+namespace ARMeilleure.Decoders
+{
+    class OpCodeT16AddSubSp : OpCodeT16, IOpCode32AluImm
+    {
+        public int Rd => 13;
```
gdkchan, line 5: I suggest using `RegisterAlias.Aarch32Sp`, just to make clear that R13 = SP here.

ARMeilleure/Decoders/OpCodeT16MemLit.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public bool WBack => false;
+        public bool IsLoad => true;
+        public bool Index => true;
+        public bool Add => true;
+
+        public int Immediate { get; }
+
+        public new static OpCode Create(InstDescriptor inst, ulong address, int opCode, bool inITBlock) => new OpCodeT16MemLit(inst, address, opCode, inITBlock);
+        public OpCodeT16MemLit(InstDescriptor inst, ulong address, int opCode, bool inITBlock) : base(inst, address, opCode, inITBlock)
```
gdkchan, line 16: Missing empty line above.

ARMeilleure/Decoders/OpCodeT16MemLit.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ARMeilleure.Decoders
+{
+    class OpCodeT16MemLit : OpCodeT16, IOpCode32Mem
+    {
+        public int Rt { get; }
+        public int Rn => 15;
```
gdkchan, line 6: `RegisterAlias.Aarch32Pc`.

ARMeilleure/Decoders/OpCodeT16MemImm5.cs
```diff
@@ -0,0 +1,57 @@
...
+                case InstName.Strb:
+                case InstName.Ldrb:
+                    Immediate = ((opCode >> 6) & 0x1f);
+                    break;
+                case InstName.Strh:
+                case InstName.Ldrh:
+                    Immediate = ((opCode >> 6) & 0x1f) << 1;
+                    break;
+                default:
+                    throw new System.InvalidOperationException();
```
gdkchan, line 53:
```suggestion
                    throw new InvalidOperationException();
```
And `using System;`.

ARMeilleure/Decoders/OpCodeT16MemMult.cs
```diff
@@ -0,0 +1,41 @@
...
+            RegisterMask = opCode & 0xff;
+            Rn = (opCode >> 8) & 7;
+            
+            int regCount = BitOperations.PopCount((uint)RegisterMask);
+            
+            switch (inst.Name)
+            {
+                case InstName.Stm:
+                    IsLoad = false;
+                    Offset = 0;
```
gdkchan, line 28: `Offset` and `PostOffset` values seems to be the same for LDM and STM, so they could be outside of the case block. Also, `IsLoad` coulld be just `IsLoad = inst.Name == InstName.Ldm`. But I don't mind the switch if you want to keep the validation. You could also use switch expressions:
```cs
IsLoad = inst.Name switch
{
    InstName.Ldm => true,
    InstName.Stm => false,
    _ => throw new InvalidOperationException
};
```

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -54,6 +55,10 @@ public Block CurrBlock
         public bool HighCq { get; }
         public Aarch32Mode Mode { get; }
 
+        public bool IsInIfThenBlock { get { return IfThenBlockState != null && IfThenBlockState.Length > 0; } }
```
gdkchan, line 58:
```suggestion
        public bool IsInIfThenBlock => IfThenBlockState != null && IfThenBlockState.Length > 0;
```

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -54,6 +55,10 @@ public Block CurrBlock
...
+        public Condition CurrentIfThenBlockCond { get { return IfThenBlockState[0]; } }
```
gdkchan, line 59:
```suggestion
        public Condition CurrentIfThenBlockCond => IfThenBlockState[0];
```

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -196,5 +201,13 @@ public Operand TryGetComparisonResult(Condition condition)
 
             return default;
         }
+
+        public void AdvanceIfThenBlockState()
+        {
+            if (IsInIfThenBlock)
+            {
+                IfThenBlockState = IfThenBlockState.Skip(1).ToArray();
```
gdkchan, line 209: It would be more efficient to store the index of the current block state, and use it on `CurrentIfThenBlockCond`. Or use some sort of integer mask instead of an array, and shift the mask here. If you really need to slice the array, you can use `ArraySegment` to do so without creating new copies of the array, but that's no different from just storing a index and using it.

merryhime: The reason I did it this way is because it naturally becomes a zero length array (i.e. out of ITBlock) without having to perform additional checks, but I see your point.

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -9,6 +9,7 @@
 using ARMeilleure.Translation.PTC;
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
merryhime, line 12: No longer required.

gdkchan: Should be removed then.

gdkchan (approved):
lgtm, thanks. Very nice to see the thumb instructions finally being implemented! I tested with Ni no kuni since it's the only Switch game I know to use those instructions, and it gets a bit further now, but requires some 32-bit thumb instructions.

riperiperi (approved):
Nice, checked the instruction encodings and everything looks great. Thanks!

ARMeilleure/Translation/ArmEmitterContext.cs
merryhime: Apologies, was just a reminder for me to do so when I got the time. :p
Resolved.

PR #3106, opened 2022-02-11
Fix missing geometry shader passthrough inputs

gdkchan:
When using geometry shader passthrough, ensure that the inputs consumed by the next stage are passed on the geometry shader stage.
This fixes a regression on Game Builder Garage, but I'm not sure which specific PR caused the regression.
master (solid black characters):
image
PR:
image


Leischii:
Guessing by the file history I would assume the Pr that caused this regression is: https://github.com/Ryujinx/Ryujinx/pull/2538

LukeWarnut:
Before:
image
After:
image

riperiperi (approved):

Thog (approved):

PR #3111, opened 2022-02-12
Use Enum and Delegate.CreateDelegate generic overloads

turbedi:
And remove now unused EnumExtensions.cs


riperiperi (approved):
Cool, didn't know this was added in .NET 5.

PR #3112, opened 2022-02-12
Perform unscaled 2d engine copy on CPU if source texture isn't in cache.

riperiperi:
Unreal Engine games tend to move texture data directly into levels of a texture via DMA. This is all well and good, but if the game is using _texture streaming_, then it will move this data between a pool of storage textures by **blitting** individual levels. This causes a number of problems for us, as People Who Want to Emulate Video Games:

- To copy a single level of a mipmapped texture to another, we need to create a view for the source and destination, which takes time and clutters up the texture cache.
- Blitting compressed texture data is done by aliasing the src/dst textures as another format... which is not supported in OpenGL, so we need to perform additional copies between the "data" representation and the "compressed" one, which is used for rendering.
  - Copy dependencies aren't perfect, so have historically caused bugs when complicated dependencies between textures are present.
- Similar to draws, blits add textures to the auto delete cache, which quickly fills up. When the auto delete cache fills up, this _forces_ the data to be flushed back to the guest memory, which forces a wait on the GPU and lengthy data conversion/copy.
- Blits count as a GPU side modification, so we need to register write tracking on textures. Write tracking costs a little to activate, but the main issue is that the data must be flushed if it is read/written on the CPU side.
  - This is especially bad given how unreal engine loads textures with DMA - the DMA transfer triggers the read tracking for textures, and forces a flush, causing a massive stutter.

This PR adds an additional path for the 2d engine copy (blit) that directly copies the data on the CPU, given a number of conditions:
- The copy regions match. (unscaled copy)
- The two textures are "data compatible" (matching dimensions, layout, bytes per pixel)
- The copy regions start at (0, 0) and align up to the texture dimensions.
- The source texture does not already exist in the cache.
  
This greatly reduces stuttering in A Hat in Time (UE3), as well as fixing any remaining texture issues with it. It will improve stuttering in UE4 games that use texture streaming (for example, if they had random graphical issues fixed by my earlier texture PRs) such as Yoshi's Crafted World. This PR does not fix stuttering in UE4 games due to bindless not working with the shader cache or file access. Really depends on the game.

I'd recommend testing some Unreal Engine games, and other stuff for regressions.

It might be a good idea to also do this copy if the source texture exists, but does not have a gpu modified flag. I'll think about it.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
@@ -1,8 +1,10 @@
-﻿using Ryujinx.Graphics.Device;
+﻿using Ryujinx.Common;
+using Ryujinx.Graphics.Device;
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Engine.Types;
 using Ryujinx.Graphics.Gpu.Image;
 using Ryujinx.Graphics.Texture;
+using Ryujinx.Memory;
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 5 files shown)


riperiperi:
Before:

https://user-images.githubusercontent.com/6294155/153728773-a09720ea-6664-4168-b2e1-a03960ebf996.mp4

After:

https://user-images.githubusercontent.com/6294155/153728776-540fd59d-581e-420a-a4c6-41877504383a.mp4

Ryujinx.Graphics.Gpu/Engine/Twod/TwodClass.cs
```diff
@@ -44,6 +48,180 @@ public TwodClass(GpuChannel channel)
...
+                {
+                    int stride = src.Stride;
+                    int offset = 0;
+                    int lineSize = width * bpp;
+
+                    for (int y = 0; y < height; y++)
+                    {
+                        srcSpan.Slice(offset, lineSize).CopyTo(dstSpan.Slice(offset));
+
+                        offset += src.Stride;
```
gdkchan, line 177:
```suggestion
                        offset += stride;
```

gdkchan (approved):
lgtm, thanks. I tested on SMT V and it was still working fine (maybe a bit smoother, it's hard to tell). A few people reported graphical improvements on other games too, like Fatal Frame (which I guess we should eventually fix on the GPU copy path too. maybe it's related to format compatibility rules).

Thog (approved):

PR #3113, opened 2022-02-12
Ryujinx.Tests: Replace unicorn with dynarmic for A32 tests

merryhime:
Self-explanatory: Unicorn does not implement all userspace A32 instructions available on v8.0.

Also enabled some previously disabled tests.

### DIFF
--- a/Ryujinx.Tests/Cpu/CpuTest32.cs
+++ b/Ryujinx.Tests/Cpu/CpuTest32.cs
@@ -4,10 +4,7 @@ using ARMeilleure.Translation;
 using NUnit.Framework;
 using Ryujinx.Cpu;
 using Ryujinx.Memory;
-using Ryujinx.Tests.Unicorn;
 using System;
-
-using MemoryPermission = Ryujinx.Tests.Unicorn.MemoryPermission;
 
 namespace Ryujinx.Tests.Cpu
 {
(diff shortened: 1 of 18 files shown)


Thog (approved):

CeruleanSky:
Since this is your passion, I'm sure it has been considered, but if by unlikely chance that it hasn't, I was wondering has Unicorn2 been looked at all? 
I saw a post recently that it has entered beta and has updated to qemu 5.0. 
Might be worth a glance to see if it instructions support has been improved. If not things stay the same, but if they did add a few instructions, things can be unit tested against both dynarmic and unicorn where applicable.

https://www.unicorn-engine.org/Unicorn2-beta

merryhime:
@CeruleanSky 

Thank you for the suggestion. I have used Unicorn2 and it remains behind; it is unfortunately based on qemu 5.0 which is rather significantly out-of-date.

Specific points of note:

1. The version of Unicorn used by Ryujinx is a slightly outdated version of [yuzu-emu/unicorn](https://github.com/yuzu-emu/unicorn), the most recent version of which is based on qemu 6.0.0-rc3.

2. Unicorn2 still doesn't support some basic instructions, e.g. `vcvtn.s32.f32` and such-like.

3. Due to my recent lack of time, I've unfortunately don't really tested Unicorn2 that well and am therefore less confident in it's correctness. I enjoy doing comparative testing and do try to upstream fixes when emulators radically disagree. Unfortunately these fixes would not have landed in Unicorn2 due to the use of an out-of-date qemu version.

I would likely reconsider if Unicorn was more up to date.

gdkchan:
The only thing that makes me hesitant about this change is the fact that dynarmic has no Arm (host) support, so we lose the ability to run the tests on Arm CPUs.

merryhime:
I note this changes is only for 32-bit instructions.

With 32-bit instructions, I do have eventual plans to support arm64 as a host with dynarmic.

WIth 64-bit instructions, it's probably preferrable to test against the host directly, if not move to virtualization instead.

merryhime:
As discussed likely going to do this an alternative way.

PR #3118, opened 2022-02-14
Added Contribute Section To Readme

jeb495:


Shahil-Ayato:
Hey, thanks for your suggestion here but this entire change is made redundant on the line right above it, when I rewrote the readme I merged a few of the headings so it would be much cleaner and also much less redundant. I think it’d be better for this not to be here personally.

PR #3119, opened 2022-02-15
Zero DMA region X/Y after copy

gdkchan:
River City Girls Zero is not writing to the destination region X/Y registers, and for this reason, they carry the previous value, however the game expects them to be zero.
This allow the game to get ingame, before it would crash after the manga style cutscene.
image
Still needs confirmation on hardware, so I have made it draft for now.


gdkchan:
Closing, after test on hardware we found that the region registers are simply ignored for linear textures. #3121 has the right fix.

PR #3122, opened 2022-02-16
Added new comments

jeb495:


Ryujinx/Configuration/ConfigurationState.cs
```diff
@@ -990,7 +990,7 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
                 
                 configurationFileUpdated = true;
             }
-            
+            //Output to console if log value changes
```
riperiperi, line 993: This isn't what this does.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -76,7 +76,7 @@ public static void Initialize(string baseDirPath)
             SetupBasePaths();
         }
 
-        private static void SetupBasePaths()
+        private static void SetupBasePaths() //create directories for combined paths
```
riperiperi, line 79: "combined paths" isn't the end goal here...

PR #3129, opened 2022-02-16
Use a basic cubic interpolation for the audren upsampler

riperiperi:
Before, it was selecting the nearest sample, which sounded terrible. This changes it to use a rather simple cubic interpolation, which greatly reduces high frequency noise (could be better tho). This is temporary til the upsampling algorithm used by the switch is reversed.

Improves bad audio in Skyward Sword HD.


Thog (approved):
lgtm

AcK77 (approved):

PR #3132, opened 2022-02-17
Prefer texture over textureSize for sampler type

gdkchan:
Fix one case of shader failing to compile. `textureSize` sampler type is sourced from the texture pool since the GPU instruction does not specify the type. However, the type may still not be ideal in a few cases:
- Texture is sampled with depth compare later on. In this case the type should be `sampler*Shadow` and not just `sampler*`.
- Texture pool type does not match sampled type. For example, texture pool could have Texture2D while it samples as Texture2DArray.

To handle those cases, if there is more than one use of the texture, one by `textureSize` and another by regular `texture` sample operations, it prefers the type from the `texture` sample operation, as this instruction actually specifies the type and it is accurate.

I did find shaders failing to compile on games due to this, but I'm not aware of any game where it causes a visible difference.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -40,7 +40,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
         /// <summary>
         /// Version of the codegen (to be changed when codegen or guest format change).
         /// </summary>
-        private const ulong ShaderCodeGenVersion = 3063;
+        private const ulong ShaderCodeGenVersion = 3132;
 
         // Progress reporting helpers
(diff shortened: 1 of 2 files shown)


LukeWarnut:
Regression found in Nintendo Switch Sports Online Play Test. The opening now consistently runs much slower than master. Tested ~3 times each, same results each time.

Master:
https://user-images.githubusercontent.com/65521430/154571100-27352dd4-1e7c-4082-a1f8-6f48393e41d5.mp4

PR:
https://user-images.githubusercontent.com/65521430/154570730-5e7b9371-465f-4190-ac38-a1309ba2ad83.mp4

LukeWarnut:
Disregard my previous comment. It worked on the github-actions build.

riperiperi (approved):

PR #3101, opened 2022-02-08
[WIP] GDB Stub Prototype

merryhime:
Prototype of a GDB stub for commentary.
Please note this is very incomplete and is primarily here for feedback regarding design and integration.

List of reminders:
- [x] Properly encode binary data on transfer
- [x] Delete single-step blocks
- [x] ~~Gracefully~~ recover from socket errors
- [ ] Consider supporting `Z0`/`Z1` for breakpoints instead
- [ ] Consider memory watchpoints
- [x] Change log level of message
- [x] Support `qThreadExtraInfo,`
- [x] Support `P` for vectors
- [ ] Ensure breakpoints work correctly

PR #3133, opened 2022-02-17
ARMeilleure: Implement single stepping

merryhime:
Allow for single stepping of instructions.

This is forked off #3101 for ease of review. Note however, this PR was implemented [slightly differently](https://github.com/Ryujinx/Ryujinx/pull/3101/files#diff-82cd0b0f0bc66b6bc9dddb35b31c72707554587253f7a5de9f4602766383c6baR99) to allow for the existence of IT blocks introduced by the thumb PR. (i.e. We step over an IT block as a monolithic unit.)

Resolving the above IT block situation is not really considered a priority given it's unlikely to crop up much in the real world use case, thus the additional complexity is unlikely to be worth it. Will revisit if this turns out to be an issue.

### DIFF
--- a/ARMeilleure/Translation/Translator.cs
+++ b/ARMeilleure/Translation/Translator.cs
@@ -266,7 +266,7 @@ namespace ARMeilleure.Translation
 
             Logger.StartPass(PassName.Decoding);
 
-            Block[] blocks = Decoder.Decode(Memory, address, mode, highCq, singleStep ? DecoderMode.SingleInstruction : DecoderMode.Normal);
+            Block[] blocks = Decoder.Decode(Memory, address, mode, highCq, singleStep ? DecoderMode.SingleInstruction : DecoderMode.MultipleBlocks);
 
             Logger.EndPass(PassName.Decoding);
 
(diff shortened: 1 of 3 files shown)


gdkchan (approved):
lgtm, but I would suggest a more descriptive name instead of `DecoderMode.Normal`. Maybe `DecoderMode.MultipleBlocks`?

riperiperi (approved):
Looks good. `MultiBlock` would work instead of normal for me.

merryhime:
that ci failure looks random to me, idk

Thog:
> that ci failure looks random to me, idk

Yeah going to restart it, it's an odd one

ARMeilleure/Translation/Translator.cs
```diff
@@ -209,6 +209,17 @@ private ulong ExecuteSingle(State.ExecutionContext context, ulong address)
             return nextAddr;
         }
 
+        public ulong Step(State.ExecutionContext context, ulong address)
```
gdkchan, line 212: One thing to note here is that some internal state is initialized on the `Execute` method. So eventually, either this should be moved elsewhere as it would prevent `Step` from being called without first calling `Execute`, which is a bit weird.

PR #3137, opened 2022-02-18
Move kernel syscall logs to new trace log level

gdkchan:
As title says.


Thog (approved):

AcK77 (approved):

gdkchan:
nvservices also print the ioctl result codes under debug right now, maybe those should be moved to trace too. Well, at least those  are less frequent.

PR #3153, opened 2022-02-22
Decoder: Exit on trapping instructions, and resume execution at trapping instruction

merryhime:
Exiting when encountering a trapping instruction is important for ensuring correct behavior, because further code beyond the trapping instruction should not execute. This allows for the opportunity for a debugger (e.g. gdbstub) that has taken over the dispatch loop to redirect the program counter appropriately.

If an instruction is a trap, we expect execution to normally resume at the trapping instruction itself. The common usecase for this is when a BRK instruction is inserted as a software breakpoint, and execution is expected to resume at that location once the original instruction has been replaced.

### DIFF
--- a/ARMeilleure/Decoders/Decoder.cs
+++ b/ARMeilleure/Decoders/Decoder.cs
@@ -32,13 +32,13 @@ namespace ARMeilleure.Decoders
 
             int instructionLimit = highCq ? MaxInstsPerFunction : MaxInstsPerFunctionLowCq;
 
-            Block GetBlock(ulong blkAddress, bool mustExit = false)
+            Block GetBlock(ulong blkAddress)
             {
                 if (!visited.TryGetValue(blkAddress, out Block block))
                 {
                     block = new Block(blkAddress);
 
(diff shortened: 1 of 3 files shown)


merryhime:
Issue with this impl is Split will attempt to split the current BB, which just results in badness.

gdkchan (approved):
lgtm, thanks.

ARMeilleure/Decoders/Decoder.cs
```diff
@@ -32,13 +32,13 @@ public static Block[] Decode(IMemoryManager memory, ulong address, ExecutionMode
 
             int instructionLimit = highCq ? MaxInstsPerFunction : MaxInstsPerFunctionLowCq;
 
-            Block GetBlock(ulong blkAddress)
+            Block GetBlock(ulong blkAddress, bool mustExit = false)
```
riperiperi, line 35: In what case would `mustExit` be set? Doesn't seem to be set right now.

merryhime: will remove

riperiperi (approved):
LGTM.

PR #3154, opened 2022-02-22
InstEmitSimdHelper: Add SSE2 impl of saturate float to 32-bit int

merryhime:
Self explanatory. Requires tests for completness.

32-bit version only currently, because writing a 64-bit implementation is [not entirely trivial](https://github.com/merryhime/dynarmic/blob/76ec1afdad3835ad6e87e6b711663a5aaf2c1939/src/dynarmic/backend/x64/emit_x64_floating_point.cpp#L1500-L1528).

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdHelper.cs
@@ -1613,6 +1613,46 @@ namespace ARMeilleure.Instructions
             return context.AddIntrinsic(single ? Intrinsic.X86Andnps : Intrinsic.X86Andnpd, mask, value);
         }
 
+        public static Operand EmitSaturateFloatToInt32(ArmEmitterContext context, Operand value, bool isF64, bool unsigned)
+        {
+            MethodInfo info;
+
+            if (isF64)
+            {
+                info = unsigned
+                    ? typeof(SoftFallback).GetMethod(nameof(SoftFallback.SatF64ToU32))
+                    : typeof(SoftFallback).GetMethod(nameof(SoftFallback.SatF64ToS32));
+            }
+            else
+            {
+                info = unsigned
+                    ? typeof(SoftFallback).GetMethod(nameof(SoftFallback.SatF32ToU32))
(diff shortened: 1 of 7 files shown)


ARMeilleure/Instructions/InstEmitSimdHelper.cs
```diff
@@ -1613,6 +1613,71 @@ public static Operand EmitFloatAbs(ArmEmitterContext context, Operand value, boo
...
+                if (unsigned)
+                {
+                    value = context.AddIntrinsic(Intrinsic.X86Minsd, value, X86GetScalar(context, 0x41efffffffe00000u));
+                    value = context.AddIntrinsic(Intrinsic.X86Maxsd, value, X86GetScalar(context, 0x0000000000000000u));
+                    value = context.AddIntrinsicLong(Intrinsic.X86Cvttsd2si, value);
+                }
+                else
+                {
+                    value = context.AddIntrinsic(Intrinsic.X86Minsd, value, X86GetScalar(context, 0x41dfffffffc00000u));
+                    // Out-of-range value for 32-bit conversion is is 0x80000000, no need for maxsd.
```
gdkchan, line 1634:
```suggestion
                    // Out-of-range value for 32-bit conversion is 0x80000000, no need for maxsd.
```

merryhime:
Can we assume SSE2 as baseline and do away with the soft fallback completely?

gdkchan:
> Can we assume SSE2 as baseline and do away with the soft fallback completely?

This is undesirable as it would make porting to new platforms more difficult (as we are forced to implement fast paths for those instructions). Plus the fallback already exists anyway, so why remove it?

gdkchan:
Just mirrorring what I said on Discord here, I think it is fine to introduce `ConvertFPToSI` and `ConvertFPToUI` IR instructions for this, because float -> int conversion is a "basic" operation. Such an instruction would follow the Arm behaviour, where out of range values are clamped to the maximum allowed range. This would also eliminate the need for the "software fallback" when SSE2 is not supported (non-x86 CPUs).

merryhime:
Revisit later. Integer saturation instructions are higher priority.

PR #3167, opened 2022-02-28
Add X86 support to build.yml

anton2007:
I stuck on Windows 10 X86 without USB stick, but I think Windows 10 X86 will work with this emulator.
I can't understand why there's no X86 support.


Thog:
The Nintendo Switch is a 64 bits console, nothing in the emulator support 32 bits host.

PR #3171, opened 2022-03-03
ThreadedRenderer: Remove check for _dispose

merryhime:
This check is unnecessary and causes command consumption to end too early.

_running is set to false when gpuLoop terminates, which would allow the command loop to terminate.


merryhime:
Brain failure.

PR #3173, opened 2022-03-04
Preparation for initial Flatpack and FlatHub integration

Thog:
This integrate some initial changes required for Flatpack and distribution from FlatHub.

Also added some resources that will be used for packaging on Linux.


gdkchan (approved):
lgtm, thanks.

PR #3176, opened 2022-03-04
Mod loading from atmosphere SD directories

### LINKED ISSUE #3174: [Feature Request] Allow loading mods from SD card
#### Overview:
In addition to the `mods` directory, mods should also be loaded from `sdcard/atmosphere/contents/[title id]/...`. Loading mods from this path is supported by both Atmosphère and Yuzu.

#### Smaller Details: 
Yuzu implementation: https://github.com/yuzu-emu/yuzu/blob/4514325b9cc792b36dbbb05e2f31a62543ab753c/src/core/file_sys/patch_manager.cpp#L348

#### Nature of Request:  
- Addition

There are two major benefits:
1. Distribution of mods would be become easier for mod authors and users could mostly stick to one consistent way of installing mods. If a mod includes the directory structure starting with the `atmosphere` folder, this folder can be copied/merged to the SD card or SD card directory and work with Atmosphère, Yuzu and Ryujinx.
2. More importantly, the planned [support for Skyline](https://github.com/Ryujinx/Ryujinx/pull/2966) allows running some mods/mod loaders like Arcropolis that allow self-updating (see e.g. https://github.com/skyline-rs/skyline-update). This requires the Skyline binary `subsdk9` to be writable, which currently isn't the case since mods can only be accessed via the read-only romfs mount. Loading the exefs from the SD card would allow self-updating without special handling for emulators.

MutantAura:
Implements addition of SD paths into the ModLoader so that standard atmosphere/hardware directory set up mods can be used semi-seamlessly. GUI context menu addition to open the directory is also added.

image
image
image

Closes #3174

### DIFF
--- a/Ryujinx.HLE/HOS/ApplicationLoader.cs
+++ b/Ryujinx.HLE/HOS/ApplicationLoader.cs
@@ -84,7 +84,7 @@ namespace Ryujinx.HLE.HOS
 
             MetaLoader metaData = ReadNpdm(codeFs);
 
-            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(new[] { TitleId }, _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath());
+            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(new[] { TitleId }, _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
 
             if (TitleId != 0)
             {
@@ -388,7 +388,7 @@ namespace Ryujinx.HLE.HOS
 
             MetaLoader metaData = ReadNpdm(codeFs);
 
-            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(_device.Configuration.ContentManager.GetAocTitleIds().Prepend(TitleId), _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath());
+            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(_device.Configuration.ContentManager.GetAocTitleIds().Prepend(TitleId), _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
 
             if (controlNca != null)
             {
--- a/Ryujinx.HLE/HOS/ApplicationLoader.cs
+++ b/Ryujinx.HLE/HOS/ApplicationLoader.cs
@@ -84,7 +84,10 @@ namespace Ryujinx.HLE.HOS
 
             MetaLoader metaData = ReadNpdm(codeFs);
 
-            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(new[] { TitleId }, _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
+            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(
+                new[] { TitleId }, 
+                _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), 
+                _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
 
             if (TitleId != 0)
             {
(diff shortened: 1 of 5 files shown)


Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -84,7 +84,7 @@ public void LoadCart(string exeFsDir, string romFsFile = null)
 
             MetaLoader metaData = ReadNpdm(codeFs);
 
-            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(new[] { TitleId }, _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath());
+            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(new[] { TitleId }, _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
```
gdkchan, line 87: Line is too long, break it down into multiple lines after each `,` and first `(`.

Ryujinx.HLE/HOS/ApplicationLoader.cs
```diff
@@ -388,7 +388,7 @@ private void LoadNca(Nca mainNca, Nca patchNca, Nca controlNca)
 
             MetaLoader metaData = ReadNpdm(codeFs);
 
-            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(_device.Configuration.ContentManager.GetAocTitleIds().Prepend(TitleId), _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath());
+            _device.Configuration.VirtualFileSystem.ModLoader.CollectMods(_device.Configuration.ContentManager.GetAocTitleIds().Prepend(TitleId), _device.Configuration.VirtualFileSystem.ModLoader.GetModsBasePath(), _device.Configuration.VirtualFileSystem.ModLoader.GetSdModsBasePath());
```
gdkchan, line 391: Line is too long, break it down into multiple lines after each `,` and first `(`.

Ryujinx/Ui/Widgets/GameTableContextMenu.Designer.cs
```diff
@@ -88,6 +89,15 @@ private void InitializeComponent()
             };
             _openTitleModDirMenuItem.Activated += OpenTitleModDir_Clicked;
 
+            //
+            // _openTitleSdModDirMenuItem
+            //
+            _openTitleSdModDirMenuItem = new MenuItem("Open SD Atmosphere Mods Directory")
```
gdkchan, line 95: Maybe just `Open Atmosphere Mods Directory`? Unless there's more than one Atmosphere mods directory, and the other one(s) is not located on the SD card, I don't think adding SD here is necessary.

MutantAura: I'm wondering whether to even have this here and maybe instead add it to the "file" context menu. The goal here is to make it easy to drag and drop mod folders set up for the switch. Most of them are just 'atmosphere/contents/ID' when you download them so I guess ideal behaviour should be something that simply opens the SD card.

The issue with this is that people may forget they have a mod installed in the SD directory and then wonder why their game still has it etc. I guess this is hardware behaviour either way...

MutantAura:
Review changes applied

gdkchan (approved):
lgtm, thanks.

tech-ticks:
Thanks for the fast PR, @MutantAura. It seems like there‘s an issue with the folder structure though: In your screenshot, the path is `atmosphere/contents/[title id]/60 FPS/[exefs, romfs]`. Atmosphère only supports one LayeredFS per game however, so it should be `atmosphere/contents/[title id]/[exefs, romfs]` without the intermediate mod directory.

MutantAura:
> Thanks for the fast PR, @MutantAura. It seems like there‘s an issue with the folder structure though: In your screenshot, the path is `atmosphere/contents/[title id]/60 FPS/[exefs, romfs]`. Atmosphère only supports one LayeredFS per game however, so it should be `atmosphere/contents/[title id]/[exefs, romfs]` without the intermediate mod directory.

Ryujinx searches all directories under titleID so it doesn't really matter what the folders beyond that point. The same would work if I removed the "60fps" intermediate.

tech-ticks:
> Ryujinx searches all directories under titleID so it doesn't really matter what the folders beyond that point. The same would work if I removed the "60fps" intermediate.

Ah sorry, I didn‘t know about the recursive search. Thanks, all good then.

Thog (approved):

PR #3178, opened 2022-03-05
T32: Implement Data Processing (Modified Immediate) instructions

merryhime:


ARMeilleure/Decoders/OpCodeT32AluImm.cs
```diff
@@ -0,0 +1,49 @@
...
+                        Immediate = imm8;
+                        break;
+                    case 1:
+                        Immediate = imm8 | (imm8 << 16);
+                        break;
+                    case 2:
+                        Immediate = (imm8 << 8) | (imm8 << 24);
+                        break;
+                    case 3:
+                        Immediate = imm8 | (imm8 << 8) | (imm8 << 16) | (imm8 << 24);
```
gdkchan, line 35: Could do `Immediate = imm8 * 0x01010101;` too. Dunno which one is faster.

gdkchan (approved):
lgtm, thanks.

ARMeilleure/Decoders/OpCodeT32AluImm.cs
merryhime:
```
const int[] factor = new int[4]{1, 0x00010001, 0x01000100, 0x01010101};
Immediate = imm8 * factor[(imm12 >> 8) & 3];
```
?

gdkchan: That's fine too, but I think you can't have const arrays in C#.
```cs
int[] factor = new int[4] { 1, 0x00010001, 0x01000100, 0x01010101 };
```
It would probably be better to make it static and move it out of the constructor, to avoid creating the array every time the constructor runs.
```cs
private static int[] _factor = new int[4] { 1, 0x00010001, 0x01000100, 0x01010101 };
```

gdkchan: Another alternative:
```cs
Vector128<int> factor = Vector128.Create(1, 0x00010001, 0x01000100, 0x01010101);
Immediate = imm8 * v.GetElement((imm12 >> 8) & 3);
```
This one had the best codegen.

gdkchan: As a side note I found that the compiler can recognize this pattern:
```cs
ReadOnlySpan<byte> data = new byte[] { 1, 2, 3, 4 };
```
And avoid the allocation and instead load the data directly as if it was a const array. But sadly that only works for bytes, not for ints, not sure why.

turbedi: > But sadly that only works for bytes, not for ints, not sure why.
[Support for this optimization for other data types is already planned for .NET 7](https://github.com/dotnet/runtime/issues/60948)

gdkchan: Nice, I guess we can revisit it once .NET 7 releases.

Thog (approved):

PR #3179, opened 2022-03-05
A32: Fix ALU immediate instructions

merryhime:
Add a test for some of these arm32 instructions. This reveals an existing bug. Fix this bug.

Bug occurs because carry flag is overwritten before it can be used in the calculation which is incorrect behaviour.


gdkchan:
PPTC version should be changed to force a recompile, since it fixes a bug on existing instructions.

gdkchan (approved):
lgtm, nice catch.

riperiperi (approved):
Nice catch here.

PR #3180, opened 2022-03-06
KThread: Fix GetPsr mask

merryhime:
Very minor PR for ease of review. Forked off from #3101.

Provide helper functions for getting a uint value instead of having to construct from flags.

Also fix GetPsr in KThread while we're here. See [Atmosphere reference](https://github.com/Atmosphere-NX/Atmosphere/blob/96f95b9f95988bd30c3dcac6c2bbabf573a23bb5/libraries/libmesosphere/source/arch/arm64/kern_k_thread_context.cpp#L43).

### DIFF
--- a/ARMeilleure/State/ExecutionContext.cs
+++ b/ARMeilleure/State/ExecutionContext.cs
@@ -103,6 +103,24 @@ namespace ARMeilleure.State
         public bool GetPstateFlag(PState flag)             => _nativeContext.GetPstateFlag(flag);
         public void SetPstateFlag(PState flag, bool value) => _nativeContext.SetPstateFlag(flag, value);
 
+        public uint GetPstate()
+        {
+            uint value = 0;
+            for (int i = 0; i < RegisterConsts.FlagsCount; i++)
+            {
+                value |= GetPstateFlag((PState)i) ? 1u << i : 0u;
+            }
+            return value;
+        }
+        public void SetPstate(uint value)
+        {
+            for (int i = 0; i < RegisterConsts.FlagsCount; i++)
+            {
+                uint flag = 1u << i;
+                SetPstateFlag((PState)i, (value & flag) == flag);
+            }
+        }
+
         public bool GetFPstateFlag(FPState flag) => _nativeContext.GetFPStateFlag(flag);
         public void SetFPstateFlag(FPState flag, bool value) => _nativeContext.SetFPStateFlag(flag, value);
 
--- a/ARMeilleure/State/ExecutionContext.cs
+++ b/ARMeilleure/State/ExecutionContext.cs
@@ -43,6 +43,8 @@ namespace ARMeilleure.State
         public long TpidrEl0 { get; set; }
         public long Tpidr    { get; set; }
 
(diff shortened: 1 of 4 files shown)


Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs
```diff
@@ -658,10 +658,7 @@ public KernelResult GetThreadContext3(out ThreadContext context)
 
         private static uint GetPsr(ARMeilleure.State.ExecutionContext context)
         {
-            return (context.GetPstateFlag(ARMeilleure.State.PState.NFlag) ? (1U << (int)ARMeilleure.State.PState.NFlag) : 0U) |
-                   (context.GetPstateFlag(ARMeilleure.State.PState.ZFlag) ? (1U << (int)ARMeilleure.State.PState.ZFlag) : 0U) |
-                   (context.GetPstateFlag(ARMeilleure.State.PState.CFlag) ? (1U << (int)ARMeilleure.State.PState.CFlag) : 0U) |
-                   (context.GetPstateFlag(ARMeilleure.State.PState.VFlag) ? (1U << (int)ARMeilleure.State.PState.VFlag) : 0U);
+            return context.GetPstate() & 0xF0000000;
```
merryhime, line 661: Was this masking intentional? Notably this would not reflect PSR.T for thumb state. I feel like the full Pstate would be more appropriate here.

gdkchan:
I feel like it should be a property like `Fpsr` and friends for example. In fact maybe the `GetPstateFlag` and `SetPstateFlag` methods should be removed, they are not very useful, and setting or testing a single bit on the uint value is trivial.

Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs
merryhime: As discussed on discord, actual mask should be 0xFF0FFE20.

See also for secondary reference: https://github.com/Atmosphere-NX/Atmosphere/blob/96f95b9f95988bd30c3dcac6c2bbabf573a23bb5/libraries/libmesosphere/source/arch/arm64/kern_k_thread_context.cpp#L43

merryhime:
@gdkchan I feel those are helpful for readability; though yes those methods are used mainly in tests.

I'm fine with making it a property of `ExecutionContext`.

ARMeilleure/State/ExecutionContext.cs
```diff
@@ -43,6 +43,8 @@ public ulong CntpctEl0
         public long TpidrEl0 { get; set; }
         public long Tpidr    { get; set; }
 
+        public uint Pstate { get => _nativeContext.GetPstate(); set => _nativeContext.SetPstate(value); }
```
gdkchan, line 46:
```suggestion
        public uint Pstate
        {
            get => _nativeContext.GetPstate();
            set => _nativeContext.SetPstate(value);
        }
```

gdkchan (approved):
lgtm, minus comment.

Thog (approved):

riperiperi (approved):

PR #3182, opened 2022-03-06
Coalesce Fpsr implementation

merryhime:
There existed an `Fpsr` property in `ExecutionContext` as well as `FpFlags` in `NativeContext`. 

* This is unnecessary duplication of state. Merge the two and have `NativeContext` be ground truth.
* Fix some what-looks-like buggy code related to this. (Notably, fpscr in A32 did not include nzcv!!)
* Remove `NativeInterface.SetFpsrQc` because calling a function for this is kind of silly.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdShift32.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdShift32.cs
@@ -1,5 +1,6 @@
 ﻿using ARMeilleure.Decoders;
 using ARMeilleure.IntermediateRepresentation;
+using ARMeilleure.State;
 using ARMeilleure.Translation;
 using System;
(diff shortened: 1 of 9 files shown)


gdkchan:
I believe it needs a PPTC version increment as one function was removed from `Delegates` class, as if I'm not mistaken the relocation information refers to them by index (and removal of one function means the index of the subsequent ones will change). Also there are 166 test failures.

merryhime:
Ah, I understand now. FP flag state is kept in registers and is not flushed on call to GetFpscr; this would similarly be an issue for any softfp callback.

merryhime:
Closing in lieu of a working PR.

PR #3184, opened 2022-03-08
Lop3Expression: Optimize expressions

merryhime:
Implemented for fun. Feel free to close if not required.

NPN-transforms the truth table to one of the basic 14 classes.

Terminology as per:
image

Sanity check: https://gist.github.com/merryhime/8684e2f6a4fcb7c9fe32ffcc932c1109

### DIFF
--- a/Ryujinx.Graphics.Shader/Instructions/Lop3Expression.cs
+++ b/Ryujinx.Graphics.Shader/Instructions/Lop3Expression.cs
@@ -13,10 +13,14 @@ namespace Ryujinx.Graphics.Shader.Instructions
             True          = 0xff, // true
             In            = 0xf0, // a
             And2          = 0xc0, // a & b
+            Or2           = 0xfc, // a | b
             Xor2          = 0x3c, // a ^ b
             And3          = 0x80, // a & b & c
+            Or3           = 0xfe, // a | b | c
             XorAnd        = 0x60, // a & (b ^ c)
+            XorOr         = 0xf6, // a | (b ^ c)
             OrAnd         = 0xe0, // a & (b | c)
+            AndOr         = 0xf8, // a | (b & c)
             Onehot        = 0x16, // (a & !b & !c) | (!a & b & !c) | (!a & !b & c) - Only one value is true.
             Majority      = 0xe8, // Popcount(a, b, c) >= 2
             Gamble        = 0x81, // (a & b & c) | (!a & !b & !c) - All on or all off
@@ -24,6 +28,7 @@ namespace Ryujinx.Graphics.Shader.Instructions
             Dot           = 0x1a, // a ^ (c | (a & b))
             Mux           = 0xca, // a ? b : c
             AndXor        = 0x78, // a ^ (b & c)
+            OrXor         = 0x1e, // a ^ (b | c)
             Xor3          = 0x96, // a ^ b ^ c
         }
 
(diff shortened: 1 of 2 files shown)


jduncanator:
It'd be nice to define all the "magic values" in there as `private const`s at the top of the class, assuming they have some proper name? Judging by the values in the switch expression, seems like they could?

Ryujinx.Graphics.Shader/Instructions/Lop3Expression.cs
```diff
@@ -9,136 +9,123 @@ static class Lop3Expression
...
-                }
-
-                visited |= mask;
             }
 
-            return expr;
+            return null;
+        }
+
+        private static Operand GetExpr(int imm, EmitterContext context, Operand x, Operand y, Operand z)
```
Thog, line 76: Might be worth mapping ``imm`` to an enum and making it have the size of a byte

gdkchan:
> It'd be nice to define all the "magic values" in there as `private const`s at the top of the class, assuming they have some proper name? Judging by the values in the switch expression, seems like they could? I won't pretend to understand what's happening here, ha!

I think it's worth explaining the problem we're trying to solve here, to put things into context.

The values on the switch expression are the immediate values of the LOP3 instruction. They are effectively the results of the truth table. Since there are 3 inputs, this value is 8-bits, as with 3 1-bit inputs we have effectively 8 possible combinations. For example, for 0x96 (0b10010110) we have:

| A | B | C | Result |
| - | - | - | ------ |
| 0 | 0 | 0 | 0 |
| 0 | 0 | 1 | 1 |
| 0 | 1 | 0 | 1 |
| 0 | 1 | 1 | 0 |
| 1 | 0 | 0 | 1 |
| 1 | 0 | 1 | 0 |
| 1 | 1 | 0 | 0 |
| 1 | 1 | 1 | 1 |

The value on the "Result" column is the immediate value. The problem we're trying to solve here is finding the minimal logical expression that, given the values for inputs A, B and C, can produce the "Result" value. For this specific immediate value, it is `(A ^ B) ^ C`, because if we look at the XOR truth table, we have:

| A | B | Result |
| - | - | ------ |
| 0 | 0 | 0 |
| 0 | 1 | 1 |
| 1 | 0 | 1 |
| 1 | 1 | 0 |

Which is exactly the result of the first 4 rows we had before. But we have an additional input here. We can see that when "A" is 1, we have the inverse results (0b1001 rather than 0b0110). Since XOR has the property inverting the input if the value is 1, and keeping it unchanged if 0, we can simply XOR the result of `(B ^ C)` by `A` to get the results on the first table.

What the instruction does is performing the logical operation for each bit of the 32-bit input values on the register, and producing a 32-bit result.

Given that, I don't think naming them is needed.

Ryujinx.Graphics.Shader/Instructions/Lop3Expression.cs
```diff
@@ -7,138 +7,129 @@ namespace Ryujinx.Graphics.Shader.Instructions
...
 
-            return expr;
+            return null;
+        }
+
+        private static Operand GetExpr(TruthTable imm, EmitterContext context, Operand x, Operand y, Operand z)
+        {
+            return imm switch
+            {
+                TruthTable.False         => Const(IrConsts.False),
```
gdkchan, line 100: I know that was on the previous code, but I think it's better to just use `Const(0)` and `Const(uint.MaxValue)` or `Const(-1)` here, since `IrConsts.False` and `IrConsts.True` is mostly used for predicate results, that only has two possible values, which is not exactly the case here.

Thog (approved):

gdkchan (changes requested):
This is currently causing regressions in at least one game, since it seems to introduce redudant `BitwiseNot` operations.
Previous codegen:
```glsl
    temp_1 = floatBitsToInt(fp_c1_data[0].x) & floatBitsToInt(fp_c0_data[1392].x);
    temp_2 = temp_0 | temp_1;
```
PR:
```glsl
    temp_1 = ~temp_0;
    temp_2 = ~floatBitsToInt(fp_c1_data[0].x);
    temp_3 = ~floatBitsToInt(fp_c0_data[1392].x);
    temp_4 = temp_2 | temp_3;
    temp_5 = temp_1 & temp_4;
    temp_6 = ~temp_5;
```
Disasm:
```
LOP3.LUT R2, R2, c[0x1][0x0], R3, 0xf8 ;
```
This is causing bindless elimination to fail here:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Shader/Translation/Optimizations/BindlessElimination.cs#L56
Which causes Hades to just show a black screen.

merryhime:
Thanks. Likely would be easier to not generate the `~(~a & ~b)` pattern to begin with and instead directly generate `a | b`.

gdkchan (approved):
lgtm, tested a few games and they look file with the change now 👍 
One thing to note is that you need to change the shader codegen version on `Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs` to update the shader cache with the new shaders.
Also it would be nice to have some comparisons of old codegen vs new. I believe it would be mostly improving on XOR expressions that the old one couldn't handle well.

riperiperi (approved):
Nice one, looks good to me.

PR #3185, opened 2022-03-08
Resolution scaling hotkeys

MutantAura:
Pretty simple. 

PageUp = Increase resolution scale by a factor of 1.
PageDown = Decrease resolution scale by a factor of 1.

Scaling up is capped to 4x scaling and scaling down is limited to a lowest value of 1x. Custom scaling values are also affected by this. For example a custom scale of 6x will be able to be stepped down with PageDown but cannot be increased with PageUp. Likewise a scale of 0.2x will scale to 1x with PageUp but will not scale to 0x with PageDown.

This is a runtime change and does not update the configuration file etc.


Niwu34:
Works very well - Thx Mutant!
https://user-images.githubusercontent.com/67392333/157248327-ff7adbdc-b2b9-47fd-b9df-bd92b9777dce.mp4

Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -612,6 +614,36 @@ private bool UpdateFrame()
...
+
+                    // Cap this hotkey at 1x scaling.
+                    if(_currentScale > 1f)
+                    {
+                        // If custom scaling used then first round the float up then decrease.
+                        float _scaleRound = (float)Math.Ceiling(_currentScale);
+                        _scaleRound -= 1f;
+                        Graphics.Gpu.GraphicsConfig.ResScale = _scaleRound;
+                    }
+                }
```
jduncanator, line 646: It would simplify all this if you just make the value wrap around, so if you keep pressing Page Up, it just goes 1, 2, 3, 4, 1, 2, 3, 4.

You can then just change all of this code to:

```suggestion
                const int MaxResolutionScale = 4;

                if (currentHotkeyState.HasFlag(KeyboardHotkeyState.ResScaleUp) &&
                    !_prevHotkeyState.HasFlag(KeyboardHotkeyState.ResScaleUp))
                {
                    Graphics.Gpu.GraphicsConfig.ResScale = Graphics.Gpu.GraphicsConfig.ResScale % MaxResolutionScale + 1;
                }

                if (currentHotkeyState.HasFlag(KeyboardHotkeyState.ResScaleDown) &&
                    !_prevHotkeyState.HasFlag(KeyboardHotkeyState.ResScaleDown))
                {
                    Graphics.Gpu.GraphicsConfig.ResScale = 
                        (MaxResolutionScale + Graphics.Gpu.GraphicsConfig.ResScale - 2)  % MaxResolutionScale + 1;
                }
```

gdkchan:
What if I want to map those hotkeys as regular controller buttons?

MutantAura:
> What if I want to map PageUp/PageDown as regular controller buttons?

Is this not an issue with every hotkey? Function keys are _less_ of a problem but I don't think they're too intuitive for two hotkeys that are so clearly designed to be used together. I'm open to suggestions but I don't think say F9 and F10 would be that great...

Ultimately this is always going to be a problem for specific users until hotkeys themselves are configurable.

gdkchan:
It would be better to make them configurable. I think Fn as hotkeys is generally fine as that's their inteded use. The only exception right now is Tab used to toggle VSync (a poor choice imo but that's another issue). I brought that up because I usually map SL/SR of left and right joycons to PageUp/PageDown and Home/End, and people doesn't seems to take in consideration the ones playing on keyboard when doing those changes, so I felt like it was worth it.

Btw those hotkeys are not really discoverable right now, so one would most likely find them by accident (and then start wondering what happened, why their game is suddenly low res or why it crashed due to it running out of memory with a res that's higher than their hardware supports).

yell0wsuit:
> Btw those hotkeys are not really discoverable right now, so one would most likely find them by accident (and then start wondering what happened, why their game is suddenly low res or why it crashed due to it running out of memory with a res that's higher than their hardware supports).

This. IMO RS is not a general-use function that should be bound to hotkeys (esp. low-end users as they don't want to touch this). If this got merged, I'd expect a whole bunch of issues to disable those hotkeys. I'd rather have configurable hotkeys for this.

MutantAura:
Drafting until I decide what scaling functionality this should have and also until the hotkey issue has been sorted. Likely after hotkeys become configurable as I agree having this set by default could cause more issues than the functionality would add.

MutantAura:
Rebased and updated with Avalonia support.

Now using implementation outlined by @jduncanator above for more compact code and a more intuitive outcome from each keypress. Default hotkey for this is **Unbound** on both Avalonia and GTK but can be bound either through the GUI with Ava or via the config file with GTK (unsure if keeping the GTK Imp was even worth it?).
Locales were added via DeepL: https://www.deepl.com/translator and as such may not be accurate to a native speaker. Feedback is appreciated here! Locales that had no support for hotkey localisation in the first place (French etc.) haven't been localised and should be tackled in a larger locale update for those languages.

Ready for review.

Ryujinx.Ava/AppHost.cs
```diff
@@ -974,6 +976,13 @@ private bool UpdateFrame()
 
                             _parent.ViewModel.Volume = Device.GetVolume();
                             break;
+                        case KeyboardHotkeyState.ResScaleUp:
+                            GraphicsConfig.ResScale = GraphicsConfig.ResScale % MaxResolutionScale + 1;
+                            break;
+                        case KeyboardHotkeyState.ResScaleDown:
+                            Graphics.Gpu.GraphicsConfig.ResScale =
+                            (MaxResolutionScale + GraphicsConfig.ResScale - 2)  % MaxResolutionScale + 1;
```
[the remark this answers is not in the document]
MutantAura: fixed!

Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -1096,6 +1098,22 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
...
+
+                configurationFileFormat.Hotkeys = new KeyboardHotkeys
+                {
+                    ToggleVsync = configurationFileFormat.Hotkeys.ToggleVsync,
+                    Screenshot = configurationFileFormat.Hotkeys.Screenshot,
+                    ShowUi = configurationFileFormat.Hotkeys.ShowUi,
+                    Pause = configurationFileFormat.Hotkeys.Pause,
+                    ToggleMute = configurationFileFormat.Hotkeys.ToggleMute,
+                    ResScaleUp = configurationFileFormat.Hotkeys.ResScaleUp,
+                    ResScaleDown = configurationFileFormat.Hotkeys.ResScaleDown
```
Thog, line 1113: I think this is wrong and should be this
```suggestion
                    ResScaleUp = Key.Unbound,
                    ResScaleDown = Key.Unbound
```

Thog (approved):
lgtm, only one thing

Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
MutantAura: Right you are!

Ryujinx.Ava/Assets/Locales/de_DE.json
```diff
@@ -551,6 +551,8 @@
   "SettingsTabHotkeysShowUiHotkey": "Zeige UI:",
   "SettingsTabHotkeysPauseHotkey": "Pausieren:",
   "SettingsTabHotkeysToggleMuteHotkey": "Stummschalten:",
+  "SettingsTabHotkeysResScaleUpHotkey": "Auflösung erhöhen:",
```
gdkchan, line 554: New entries are usually added to the bottom to make translation of other locales easier.

MutantAura: Moved them down 👍

gdkchan (approved):
lgtm, thanks.

PR #3193, opened 2022-03-13
InstEmitMemoryEx: Barrier after write on ordered store

merryhime:
In theory, mfence should be after a write.

Games that would likely be affected by this and would require testing:

* Pokémon Mystery Dungeon (boot) 
* Pokémon Brilliant Diamond/Shining Pearl (entering/exiting buildings)
* Pokémon Legends Arceus (anywhere)

Please note an alternative to this is to just implement an ordered write as a single instruction (`lock xchg`) which has a built-in memory fence.


gdkchan:
Needs the PPTC version change.

MutantAura:
From my testing of RTDX this doesn't seem to have impacted anything.
Discord tests report that PLA and BDSP are also unchanged 👍

gdkchan (approved):
lgtm, thanks.

Thog (approved):

Issue #2851, opened 2021-11-18
Mario Party Superstars - Eatsa Pizza minigame glitch.


smiglo5:
## Bug Report

Eatsa Pizza minigame glitch.

### What's the issue you encountered?

Sometimes one side of the pizza does not load, the base of the pizza does not load, only the toppings do, at first I thought it was just a visual glitch, but in reality if any player encounters this glitch it will be easier to complete the game because there is less pizza to eat.

### How can the issue be reproduced?

Relaunch the minigame over and over, it will happen eventually, the probability is very high that a glitch will occur

### Log file

[Ryujinx_1.0.7106_2021-11-18_09-03-22.log](https://github.com/Ryujinx/Ryujinx/files/7561104/Ryujinx_1.0.7106_2021-11-18_09-03-22.log)

### Environment?

- Ryujinx version: 1.0.7106</br>

- Game version: 1.1.0</br>

- System Specs:
  - OS: *(Windows 10)*
  - CPU: *(i7-10700)*
  - GPU: *(NVIDIA GTX 1060 6GB)*
  - RAM: *(16GB)*
- Applied Mods : No

And here is the screenshot from this glitch.

![139661038-25bf4ec3-eb13-4eec-8b3d-4887833916e7](https://user-images.githubusercontent.com/30460684/142376296-64a031a2-502e-499e-b414-784a58e83d47.jpg)

Screenshot is from LDN, but glitch is identical on 1.0.7106 mainline.

INFINITY99K:
How was it supposed to like? I don't have the game so, please compare with how it should like.

smiglo5:
> How was it supposed to like? I don't have the game so, please compare with how it should like.

Nope, both sides of pizza should be exactly the same, like this:
![mpc-hc64_QWOzMEEBcQ](https://user-images.githubusercontent.com/30460684/142412065-06db7e77-5973-4c06-ac7f-c741b11e0957.jpg)

Sometimes it renders correctly on Ryujinx, sometimes it's not. That's why I've mentioned opening the game over and over to encounter that glitch in reproduce section of this issue report.

MetrosexualGarbodor:
Reloaded the minigame several times and can confirm this does not happen anymore in master.

https://user-images.githubusercontent.com/79612681/164702964-445ddb0f-8f46-4af3-b035-0e1707c7143a.mp4

Issue #2603, opened 2021-08-29
Fire Emblem Three houses freeze on shader cache boot


ShannonEgoiste:
### What's the issue you encountered?
Ryujinx seem's to freeze when trying to re compil an shader from cache 

### How can the issue be reproduced?
launching the game with a shader cache with the specific shader

or (matter of guesses) do choir on a free days in monasteries (don't deadlock as in booting but the shader seems to came from here)

### Environment
- Ryujinx version: 1.0.7018</br>
- Game version: 1.2.0</br>
- System Specs:
  - OS: Windows 10
  - CPU: i7-7700hq
  - GPU: NVIDIA GTX 1070
  - RAM: 8GB
### Additional context
Some version ago the game was working without any problem...

Aghathis:
[Ryujinx_1.0.7018_2021-08-29_03-56-07.log](https://github.com/Ryujinx/Ryujinx/files/7072471/Ryujinx_1.0.7018_2021-08-29_03-56-07.log)
[e66f540726c0564281dee2819eeae89f.zip](https://github.com/Ryujinx/Ryujinx/files/7072472/e66f540726c0564281dee2819eeae89f.zip)

gdkchan:
I can't reproduce this issue here, I have noticed that a few shaders takes a long time to compile on this game (about 2 minutes), but it does eventually compile.

Aghathis:
not a matter of time... i've tried more than 15 minute and change nothing...
the game was working fine some version ago ! with the same shader cache... 

here at each launch (if i delete this specific shader in cache (e66f...)) the game run... but the shader is recreated (guess it's in the choir in monastery till it's the only place where the game totally freeze more than 10 minute before continuing)

here the full shader cache if it can help :
[shader.zip](https://github.com/Ryujinx/Ryujinx/files/7080147/shader.zip)

Mou-Ikkai:
> not a matter of time... i've tried more than 15 minute and change nothing...
> the game was working fine some version ago ! with the same shader cache...
> 
> here at each launch (if i delete this specific shader in cache (e66f...)) the game run... but the shader is recreated (guess it's in the choir in monastery till it's the only place where the game totally freeze more than 10 minute before continuing)
> 
> here the full shader cache if it can help :
> [shader.zip](https://github.com/Ryujinx/Ryujinx/files/7080147/shader.zip)

You uh... Can't share shaders, bud. It's against the rules and also possibly illegal since it holds game code.

[...]

Issue #1768, opened 2020-12-02
Shader cache: crash on null invalid guest shader files


GamerzHell9137:
5900 is the testing version.
Animal Crossing loaded fine with the shaders from older build and it launched fine.
Games started fine and crashed when a weird square shape object appeared on the character (shader compiled maybe?)
After that the game didn't want to load from the Ryujinx launcher.
By deleting a specific shader (recommended by ripperoni) the issue was fixed.

![image](https://user-images.githubusercontent.com/2002038/100812466-5645da80-343d-11eb-93fc-057c160a7119.png)

Logs:
[Ryujinx_2020-12-02_00-56-55.log](https://github.com/Ryujinx/Ryujinx/files/5626202/Ryujinx_2020-12-02_00-56-55.log)
[Ryujinx_2020-12-02_01-03-14.log](https://github.com/Ryujinx/Ryujinx/files/5626203/Ryujinx_2020-12-02_01-03-14.log)
[Ryujinx_2020-12-02_01-04-23.log](https://github.com/Ryujinx/Ryujinx/files/5626204/Ryujinx_2020-12-02_01-04-23.log)

GamerzHell9137:
Same is happening on Mario Tennis Aces

AcK77:
It's still the case in lastest master build ?

MetrosexualGarbodor:
Closing this as there's been no mention or reports even before the new shader cache was implemented, and as the new cache rebuilds or skips broken shaders, this should not happen anymore.

Issue #864, opened 2020-01-09
Shader cache needs improvements


gdkchan:
The shader cache (in memory not on disk) currently is not very good at keeping track of shaders that moved in memory. Suppose that we have a shader at memory address 0x100. Let's call it shader A. The game draws something with it, then overwrites that code with another shader (let's call it shader B). Since the shader was replaced with a new one, the GPU emulator should keep track of that and translate/compile the new shader. Currently it should work fine, but it is not very efficient. It compares the shader code with the one in memory every time to detect modifications. It's not too bad since shaders are usually small, however it can be improved, using the same memory modification check method we use for other GPU resources like textures and buffers.

Continuing on the example above, suppose the game writes shader A at 0x100 again. Since it's the same shader as before, we can use the same code that was translated/compiled before and avoid a slow re-compilation of the same code. Currently this should also work. It stores a list of shaders per address, and compares one by one until it finds a match. If no match is found, it assumes that the shader was never used before, and translates, compiles and adds it to the list.

Now to the case that currently doesn't work very well. Let's modify the above example slightly. Suppose that shader A is written at 0x100, then shader B is written at 0x100. This works fine. Now shader A is written at 0x200. This currently triggers a re-translation/compilation. It is not smart enough to know that the shader was already translated before, since the address changed.

So, the main problems with the current approach it that it will keep all the shaders used within the same memory address in memory, but if the shader is used again at a different address, it re-compiles and adds it into another list. So this can waste memory with shaders that are never going to be used again, and fail to find shaders that were already compiled before.

A better solution would be calculating a hash from the shader code, and store the shaders using some sort of hash map. This should solve both problems described above, and also help with on-disk shader caches.

EmulationFanatic:
This was accomplished in https://github.com/Ryujinx/Ryujinx/pull/1701

EmulationFanatic:
Reopening as the duplication in memory is still in need of addressing

PR #3194, opened 2022-03-13
New shader cache implementation

### LINKED ISSUE #2045: Bindless textures break the shader cache

### LINKED ISSUE #3004: Yo-Kai Watch 1 for Nintendo Switch: Grafical issue
There's an issue on the game that makes the backgrounds flash with different colours.
 Also when you're pointing a yo kai with the lens it flickes and when you finf a yo -kai whith the lens a glitched texture apears instead of the 3d model of the yo kai, this also happens with the crank-a-kai.

Logs:
[Ryujinx_1.0.7174_2022-01-17_15-33-07.log](https://github.com/Ryujinx/Ryujinx/files/7883598/Ryujinx_1.0.7174_2022-01-17_15-33-07.log)
[Ryujinx_1.0.7175_2022-01-17_15-34-02.log](https://github.com/Ryujinx/Ryujinx/files/7883599/Ryujinx_1.0.7175_2022-01-17_15-34-02.log)
[Ryujinx_1.0.7175_2022-01-17_17-51-01.log](https://github.com/Ryujinx/Ryujinx/files/7883600/Ryujinx_1.0.7175_2022-01-17_17-51-01.log)

Ryujinx version:1.07175
Game version:v1.0,2

System specs:
OS:Windows 10 home, version:21H1
GPU:Intel(R) UHD Graphics 620
CPU:Intel(R)Core(TM)i7_8550U 
RAM:12,8 GB

gdkchan:
This change has the goal of fixing issues related to the shader cache implementation. Here, shader cache refers to two different caches: the memory shader cache and the disk shader cache. Below I divide shader cache problems in categories, and explain how each one is addressed here.

### Shader lists per address are inefficient

The current implementation keeps a list of all shaders that were ever located at a given GPU address. This list keeps growing forever, which is already a problem, but even worse is that every time there is a shader at that address, it will check every single shader in the list to try finding a match. That means performance can *degrade over time* as the list grows larger. It was done this way because at the time there was no disk cache, and I did not want to implement hash based lookup.

The new cache removed the list and now keeps only one entry per address, which means that it always do a single check for shader match. If that fails, it then does a hash based lookup which can find any shader that was already compiled. The implementation on master can do hash based lookup, but only when the shader cache is enabled, while the new one can also do it with shader cache disabled. Furthermore, it no longer needs to know the shader size to do the lookup.

The new lookup uses a sorted list of sizes, with the sizes of all shaders in the cache. For each size, it also has a hash table with all shaders of that size (hash and associated data). It starts by binary searching for the correct size. First, it starts at the middle of the list, and hashes that amount of bytes in memory of the shader GPU address. It then checks on the hash table, if there is any shader with this size whose code matches the hash. If there is a match, it can be either full or partial. Partial means that there is a shader in the cache that matches the hash, but the hash is only for part of the code, and in this case it should keep searching on shaders of greater size, going *right* on the sorted size list. If the match is full, then the entire code matches the hash and it found the shader. If there is no match, then it should try shaders with lower sizes, going *left* on the sorted list.

This approach avoids the need of decoding the shader code to find the size, which can save some time.

### Packing when closing

The old cache uses a ZIP archive where all shader files are packed and compressed. This has a cost that is paid every time the emulation is stopped or the emulator closed, since it has to pack everything back, making the shutdown process slower. The new cache solves this by not using a ZIP archive at all. Now it uses 6 files, which are 3 `.toc` and .`data` file pairs. They are "guest", which stores the guest code and constant buffer data required to translate the shader, "shared", which contains GPU state needed to translate the shader, in addition to metadata produced by the translator, and "host", which contains the host shader binary, retrieved from the driver. The host file pair is also per API and per vendor. The `.toc` file contains offsets into the `.data` file in addition to a small header. The `.data` file contains the actual header. This is because those files are append-only, and each entry has variable length, so that allows easy random access of entries, while only needing to write at the end of both files.

### Sometimes, the GLSL code is kept in memory

On master, the GLSL code of shaders that are compiled (not from the disk cache) is kept in memory. After compilation, it is not really needed so keeping them in memory is just a waste. The new cache solves this by storing the `ShaderProgramInfo` rather `ShaderProgram` in the cahce. The GLSL code is located on `ShaderProgram`. To allow this, some state was moved from `ShaderProgram` to `ShaderProgramInfo`. A copy of the guest code is also kept in memory for comparison purposes. On master, if there are different combinations of the same shader stages, it would need to compile new shaders, and would also create new arrays to store the guest code. The new cache re-uses the existing arrays to avoid wasting memory with the same data.

### Lack of shader specialization

Since shader translation depends on GPU state, the correct thing to do would be creating a new version of the shader if this state changes. But on master, this is not done, and the old cache does not support storing different versions of the same shader. This is addresses here by adding a `SpecializationState` where the state for a version of the shader is stored. In addition to that, it checks if the state recorded there matches the current GPU state for every use. If it doesn't match, a new version of the shader is compiled for the new state.

This fixes graphical glitches on Yokai Watch 1 and maybe other games.
Before:

https://user-images.githubusercontent.com/5624669/158046234-9d5a670a-9487-48ea-8865-0b7460b2f48b.mp4

After:

https://user-images.githubusercontent.com/5624669/158046124-43441d08-78b4-4ca8-9ab7-20794c537081.mp4

Fixes #3004.

### Bindless textures were not supported.

The old cache did not support shaders using bindless textures. This seems to be because it does not save the constant buffer slot where the texture handle is located on the cache. The new cache stores all information required to recompile those shaders, so games using bindless textures (UE4 games and a few others) are now supported.

Fixes #2045.

### Closing while the shader cache loads

It was not possible to close the emulator clealy, or stop emulation while the shader cache was loaded. This is very annoying when one opens a game by mistake, and then finds they now need to wait until the shader cache finishes loading to close. This has been addressed now, the shader cache Initialize method takes a cancellation token that can be used by the UI to cancel the load process at any time.

TODO:
- [x] Ensure that it works with multiple emulator instances.
- [x] Ensure it can handle corrupted cache files (either discard or repair).
- [x] XML docs.
- [x] Migration from old cache format.
- [x] Make purge shader cache option also delete the new cache.
- [x] Investigate driver crash on NVIDIA when rebuilding after migrating from old cache.


Ryujinx.Common/Cache/HashState.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Common.Cache
+{
+    struct HashState
```
Thog, line 6: Might be worth adding a note to say it's based on Fast Hash. Also could be in another namespace than Cache as it could be used for other stuffs?

Ryujinx.Graphics.Gpu/Shader/ShaderSpecializationState.cs
```diff
@@ -0,0 +1,338 @@
...
+{
+    class ShaderSpecializationState
+    {
+        private const uint ComsMagic = (byte)'C' | ((byte)'O' << 8) | ((byte)'M' << 16) | ((byte)'S' << 24);
+        private const uint GfxsMagic = (byte)'G' | ((byte)'F' << 8) | ((byte)'X' << 16) | ((byte)'S' << 24);
+        private const uint TfbdMagic = (byte)'T' | ((byte)'F' << 8) | ((byte)'B' << 16) | ((byte)'D' << 24);
+        private const uint TexkMagic = (byte)'T' | ((byte)'E' << 8) | ((byte)'X' << 16) | ((byte)'K' << 24);
+        private const uint TexsMagic = (byte)'T' | ((byte)'E' << 8) | ((byte)'X' << 16) | ((byte)'S' << 24);
+
+        private enum QueriedStateFlags : byte
```
Thog, line 14: Mark with the Flag attribute?

Thog (commented):
Did a quick dive into it.

Great work this does address the shortcomings of my initial implementation 👍 
I'm not entirely sure if we would be able to carry over the old cache tho...

Ryujinx.Common/Cache/HashState.cs
gdkchan: I plan to change it to the existing XXHash128 implementation, this was just for testing. Will be a bit tricky to do that without duplicating the code since it also has to store internal hash state to allow it to be "continued", but I will try.

jduncanator: I would make the current implementation stateless, passing in a xxHash128 ref struct as the state, then add static methods to the hash implementation that lets you perform one off hashes easily without having to manage that ref struct. Best of both worlds.

MutantAura:
Putting these here from discord. Seems to have 2 core issues so far (error codes below):

On 2nd boot of a game directly after the cache initialising:

`00:00:01.547 |E| GPU.MainThread Application : Unhandled exception caught: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Ryujinx.Graphics.Gpu.Shader.DiskCacheHostStorage.LoadShaders(GpuContext context, ShaderCacheHashTable graphicsCache, ComputeShaderCacheHashTable computeCache, ParallelDiskCacheLoader loader) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Shader\DiskCacheHostStorage.cs:line 126`
   
When running in portable mode I could consistently get this error:

`00:00:10.645 |E| Gpu.BackgroundDiskCacheWriter Application : Unhandled exception caught: System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\stuff\ryujinx-Release-1.0.0+afcc705-win_x64\portable\games\01007ef00011e000\cache\shader\shared.toc'.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String , FileMode , FileAccess , FileShare , FileOptions )`

Other users reported the 2nd error even without portable mode. 
Log error 1: 
[Ryujinx_1.0.0-dirty_2022-03-13_11-47-26.log](https://github.com/Ryujinx/Ryujinx/files/8239783/Ryujinx_1.0.0-dirty_2022-03-13_11-47-26.log)

Log error 2:
[Ryujinx_1.1.0afcc705_2022-03-13_13-18-59.log](https://github.com/Ryujinx/Ryujinx/files/8239785/Ryujinx_1.1.0afcc705_2022-03-13_13-18-59.log)

LukeWarnut:
Crashes with this error if the game has not yet been played on master,

```
00:00:07.861 |E| Gpu.BackgroundDiskCacheWriter Application : Unhandled exception caught: System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\luker\AppData\Roaming\Ryujinx\games\0100f52013a66000\cache\shader\shared.toc'.
   at Microsoft.Win32.SafeHandles.SafeFileHandle.CreateFile(String , FileMode , FileAccess , FileShare , FileOptions )
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String , FileMode , FileAccess , FileShare , FileOptions , Int64 )
   at System.IO.Strategies.OSFileStreamStrategy..ctor(String , FileMode , FileAccess , FileShare , FileOptions , Int64 )
   at Ryujinx.Graphics.Gpu.Shader.DiskCacheHostStorage.AddShader(GpuContext context, CachedShaderProgram program, ReadOnlySpan`1 hostCode) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Shader\DiskCacheHostStorage.cs:line 278
   at Ryujinx.Graphics.Gpu.Shader.BackgroundDiskCacheWriter.ProcessTask(CacheFileOperationTask task) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Shader\BackgroundDiskCacheWriter.cs:line 66
   at Ryujinx.Common.AsyncWorkQueue`1.DoWork() in D:\a\Ryujinx\Ryujinx\Ryujinx.Common\AsyncWorkQueue.cs:line 37
```

LukeWarnut:
> This fixes graphical glitches on Yokai Watch 1, Charge Kid, and maybe other games.

I am not finding any difference in Charge Kid.

Master:
image
image

PR:
image
image

LukeWarnut:
Breaks Super Mario Galaxy (3D All-Stars)
```
00:00:06.124 |E| GPU.MainThread Application : Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: Not mapped: va=0x0000000000005700, size=0x0000000000000004
   at Ryujinx.Cpu.MemoryManagerHostMapped.AssertMapped(UInt64 va, UInt64 size) in D:\a\Ryujinx\Ryujinx\Ryujinx.Cpu\MemoryManagerHostMapped.cs:line 134
   at Ryujinx.Cpu.MemoryManagerHostMapped.Read[T](UInt64 va) in D:\a\Ryujinx\Ryujinx\Ryujinx.Cpu\MemoryManagerHostMapped.cs:line 166
   at Ryujinx.Graphics.Gpu.Image.TextureBindingsManager.ReadPackedId(Int32 stageIndex, Int32 wordOffset, Int32 textureBufferIndex, Int32 samplerBufferIndex) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Image\TextureBindingsManager.cs:line 566
   at Ryujinx.Graphics.Gpu.Shader.ShaderSpecializationState.Matches(GpuChannel channel, GpuChannelPoolState poolState, Boolean isCompute) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Shader\ShaderSpecializationState.cs:line 207
   at Ryujinx.Graphics.Gpu.Engine.Threed.DrawManager.DrawEnd(ThreedClass engine, Int32 firstIndex, Int32 indexCount) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Engine\Threed\DrawManager.cs:line 129
   at Ryujinx.Graphics.Gpu.Engine.Threed.DrawManager.DrawIndexedSmall(ThreedClass engine, Int32 argument, Boolean instanced) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Engine\Threed\DrawManager.cs:line 318
   at Macro(MacroJitContext , IDeviceState , Int32 )
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Send(UInt64 gpuVa, Int32 offset, Int32 argument, Int32 subChannel, Boolean isLastCall) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 264
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Process(UInt64 baseGpuVa, ReadOnlySpan`1 commandBuffer) in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 95
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoDevice.DispatchCalls() in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoDevice.cs:line 188
   at Ryujinx.Ui.RendererWidgetBase.<Render>b__64_0() in D:\a\Ryujinx\Ryujinx\Ryujinx\Ui\RendererWidgetBase.cs:line 393
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.<>c__DisplayClass48_0.<RunLoop>b__0() in D:\a\Ryujinx\Ryujinx\Ryujinx.Graphics.GAL\Multithreading\ThreadedRenderer.cs:line 100
```
[Ryujinx_1.1.0+afcc705_2022-03-13_12-15-17.log](https://github.com/Ryujinx/Ryujinx/files/8240122/Ryujinx_1.1.0%2Bafcc705_2022-03-13_12-15-17.log)

gdkchan:
All reported issues should be fixed now.

gdkchan:
> > This fixes graphical glitches on Yokai Watch 1, Charge Kid, and maybe other games.
> 
> I am not finding any difference in Charge Kid.
> 

The graphics used to look blocky ingame due to it using non-normalized coordinates. But I can't reproduce it now on master, maybe it's random, maybe something else fixed it, who knows. I will just pretend it never happened I guess.

GamerzHell9137:
Atelier Sophie 2 on master was showing
```
00:07:49.861 |E| GUI.RenderLoop Gpu : GL Sync Object 56236 failed to signal within 1000ms. Continuing...
```

and didn't let players progress in the game but with the shader cache rewrite it doesn't freeze in the Near Forest and the game runs properly!

image

riperiperi:
Great to finally have shaders using bindless textures covered by the shader cache. Been doing some early testing and Unreal Engine 4 games are much smoother, even when master has a driver shader cache present. (which inevitably ends up wiped, whereas this does not)

One quick example is in Mario Party Superstars when the character select comes up. In master, it noticably freezes while it translates the shaders for all the characters, whereas with this PR it finds the shaders much faster and probably spends more time loading their textures. 

Master (driver shader cache present):
https://user-images.githubusercontent.com/6294155/158274512-b64eb0f1-d4f2-4762-b152-5af8952d4328.mp4

PR:
https://user-images.githubusercontent.com/6294155/158274550-0661cf9b-0bf3-408c-852f-03935aceaf48.mp4

I'll make some more comparisons shortly.

Shahil-Ayato:
Closes #2045

Shahil-Ayato:
Closes #864

Shahil-Ayato:
Potentially closes #1768

Shahil-Ayato:
Potentially closes #2603

Shahil-Ayato:
Potentially closed #2851

gdkchan:
> Closes #864

This solves the second issue, but not the first one described here. I will think about whenever it should be closed or not.

A-w-x:
fixes minor shield glitch on SSBU on intel (linux)

Ryujinx.Graphics.Gpu/Shader/DiskCache/BackgroundDiskCacheWriter.cs
```diff
@@ -54,13 +72,22 @@ public AddShaderData(CachedShaderProgram program, byte[] hostCode)
         private readonly DiskCacheHostStorage _hostStorage;
         private readonly AsyncWorkQueue<CacheFileOperationTask> _fileWriterWorkerQueue;
 
+        /// <summary>
+        /// Creates an ew background disk cache writer.
```
yell0wsuit, line 76: Spelling error

Ryujinx.Graphics.Gpu/Shader/DiskCache/BackgroundDiskCacheWriter.cs
```diff
@@ -0,0 +1,125 @@
...
+                HostCode = hostCode;
+            }
+        }
+
+        private readonly GpuContext _context;
+        private readonly DiskCacheHostStorage _hostStorage;
+        private readonly AsyncWorkQueue<CacheFileOperationTask> _fileWriterWorkerQueue;
+
+        /// <summary>
+        /// Creates an ew background disk cache writer.
```
yell0wsuit, line 76:
```suggestion
        /// Creates a new background disk cache writer.
```

Ryujinx.Common/Cache/HashState.cs
gdkchan: In the end I just added a comment, as it's easier and I prefer to not need re-testing the hash implementation etc.

gdkchan:
This is ready for review/testing now.
A few things that I changed and are worth noting:
- `CompileShader` was removed from `IRenderer`. A new `ShaderSource` structure was added. It has the shader code (with a string for text shaders and a byte array for binary shaders), the shader stage and language enums. The structure is now passed to `CreateProgram` directly. This was done not only to simplify the GAL interface, but also to allow OpenGL shader objects to be deleted early (they are now deleted after checking the link status). This might allow reducing the driver memory usage.
- The `QueryIsTextureRectangle` method was changed to `QueryTextureCoordNormalized`. It now just returns the coordinate normalized state from the texture pool descritor, rather than "is it a 2D texture and coordinates are not normalized" (rectangle texture), since this is the correct thing to do here (as coordinate normalization is not limited to 2D texture, although NVN only has rectangle textures). This is less problematic now with specialization state, since if for some reason that state changes, it will just recompile the shader.

Things to test:
- Performance. Since it now checks if the shader state matches current GPU state before each draw, it could have some performance impact.
- For regressions in general.

DreadWolf-sketch (approved):

MutantAura:
Works very well on my system. All caches were migrated without issue other than Breath of the Wild which had a lot of invalid texture descriptor errors and then hung for around 10 mins with a full translation bar. It did eventually resolve itself so perhaps some patience will be required for those larger caches. 

Performance was almost identical so I don't think that's an issue. The only game I could consistency measure any kind of loss was PLA and it was minor. May need some lower end system tests here though.

Master: 
image
image

New: 
image
image

Haronee:
Amazing! Since everyone else did their testing on Nvidia GPUs I decided to take a look at AMD ones on Windows just to be sure nothing breaks and I'm happy to report I didn't find any major issue.

Here are some screenshots:
image
image

Tested on an RX570 on latest AMD drivers with an i5 7500 on Windows 10

MutantAura:
Did a quick comparison to riperiperi's earlier videos and I'm no longer seeing those same differences in Mario Party: Superstars.

Intro Screen has a fair amount of delay before rendering:

https://user-images.githubusercontent.com/44103205/162000085-8c9e470f-7f61-47e5-a6ae-a9649225b4bf.mp4

Character select has some general stutter (these characters were selected in a previous run):

https://user-images.githubusercontent.com/44103205/162000203-635fa1cb-77b0-4a56-9a36-1e4a5e7c475b.mp4

Mt Minigame has some general stutter on each boot:

https://user-images.githubusercontent.com/44103205/162000250-3bf2bd84-ff7b-4af9-86ec-24f17ad0ebce.mp4

Seems the new Lego Star Wars game also has similar issues in the intro video (seems partly related to shaders but I still have no idea if this is a video or not lmfao):

https://user-images.githubusercontent.com/44103205/162001596-4def2753-5d3c-40f0-bde7-e1c672467590.mp4

Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
```diff
@@ -0,0 +1,763 @@
...
+
+            /// <summary>
+            /// Reserved space, to be used in the future. Write as zero.
+            /// </summary>
+            public ulong Reserved;
+
+            /// <summary>
+            /// Reserved space, to be used in the future. Write as zero.
+            /// </summary>
+            public ulong Reversed2;
```
riperiperi, line 66: Reserved?

Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheGuestStorage.cs
```diff
@@ -0,0 +1,459 @@
...
+
+            /// <summary>
+            /// Reserved space, to be used in the future. Write as zero.
+            /// </summary>
+            public ulong Reserved;
+
+            /// <summary>
+            /// Reserved space, to be used in the future. Write as zero.
+            /// </summary>
+            public ulong Reversed2;
```
riperiperi, line 58: Reserved?

Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
```diff
@@ -0,0 +1,763 @@
...
+        {
+            using var tocFileStream = DiskCacheCommon.OpenFile(_basePath, SharedTocFileName, writable: true);
+            using var dataFileStream = DiskCacheCommon.OpenFile(_basePath, SharedDataFileName, writable: true);
+
+            tocFileStream.SetLength(0);
+            dataFileStream.SetLength(0);
+        }
+
+        /// <summary>
+        /// Deletes all content from the host caches.
```
riperiperi, line 565: The other two methods say "the ___ cache files".

Ryujinx.Graphics.Gpu/Shader/HashTable/PartitionHashTable.cs
```diff
@@ -0,0 +1,452 @@
...
+                {
+                    return new ReadOnlySpan<byte>(Data).Slice(0, OwnSize);
+                }
+
+                return Data;
+            }
+        }
+
+        /// <summary>
+        /// Hash table bucked.
```
riperiperi, line 98: bucket?

Ryujinx.Graphics.Gpu/Shader/HashTable/PartitionHashTable.cs
```diff
@@ -0,0 +1,452 @@
...
+
+            Entry entry = new Entry(dataHash, data, item);
+
+            AddToBucket(dataHash, ref entry);
+
+            return item;
+        }
+
+        /// <summary>
+        /// Adds an item to the hash table
```
riperiperi, line 151: Full stop

Ryujinx.Graphics.Gpu/Shader/HashTable/PartitionHashTable.cs
```diff
@@ -0,0 +1,452 @@
...
+                        }
+                    }
+                }
+            }
+
+            _buckets = newBuckets;
+        }
+
+        /// <summary>
+        /// Gets the bucked for a given hash.
```
riperiperi, line 429: bucket

Ryujinx.Graphics.Gpu/Shader/HashTable/PartitionHashTable.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        /// <summary>
+        /// Gets a power of two count from a regular count.
+        /// </summary>
+        /// <param name="count">Count</param>
+        /// <returns>Power of two count</returns>
+        private static int GetPow2Count(int count)
+        {
+            // This returns the nearest power of two that is lower than count.
+            // That was done to optimize memory usage rather than performance.
```
riperiperi, line 448: This was done?

Ryujinx.Graphics.Gpu/Shader/DiskCache/ParallelDiskCacheLoader.cs
```diff
@@ -0,0 +1,672 @@
...
+        /// Enqueues a program for validation, which will check if the program was compiled successfully.
+        /// </summary>
+        /// <param name="newEntry">Program entry to be validated</param>
+        private void EnqueueForValidation(ProgramEntry newEntry)
+        {
+            _validationQueue.Enqueue(newEntry);
+
+            // Do not allow more than N shader compilation in-flight, where N is the maximum number of threads
+            // the driver will be using for parallel compilation.
+            // Submiting more seems to cause NVIDIA OpenGL driver to crash.
```
riperiperi, line 496: Submitting

Ryujinx.Graphics.Gpu/Shader/TextureDescriptorCapableGpuAccessor.cs
```diff
@@ -160,10 +160,13 @@ public SamplerType QuerySamplerType(int handle, int cbufSlot = -1)
         /// <param name="handle">Texture handle</param>
         /// <param name="cbufSlot">Constant buffer slot for the texture handle</param>
         /// <returns>True if the texture is a rectangle texture, false otherwise</returns>
-        public bool QueryIsTextureRectangle(int handle, int cbufSlot = -1)
+        public virtual bool QueryTextureCoordNormalized(int handle, int cbufSlot = -1)
         {
-            var descriptor = GetTextureDescriptor(handle, cbufSlot);
+            return QueryIsTextureRectangle(GetTextureDescriptor(handle, cbufSlot));
```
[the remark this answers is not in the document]
gdkchan: This class is not used anymore, I will just remove it.

gdkchan:
Addressed feedback and added a comment on the serialized structs to prevent people from re-ordering or adding fields at the start or middle, which would break compatibility with existing cache files.

jduncanator:
> Did a quick comparison to riperiperi's earlier videos and I'm no longer seeing those same differences in Mario Party: Superstars. _Audio ON is recommended to hear when the sections should start._ Intro Screen has a fair amount of delay before rendering:
> 
>  Intro.mp4 
> Character select has some general stutter (these characters were selected in a previous run):
> 
>  CSS.mp4 
> Mt Minigame has some general stutter on each boot:
> 
>  minigame.mp4 
> Seems the new Lego Star Wars game also has similar issues in the intro video (seems partly related to shaders but I still have no idea if this is a video or not lmfao):
> 
>  Lsw-1.mp4

Sorry, not 100% sure what you're saying here. Are you saying there has been regressions from when it was originally tested? You didn't provide any captions for those videos, are they recorded in master or this PR, on which commit?

riperiperi (approved):
Fantastic work. Great to see specialization and bindless texture shader caching finally make it in. The new behaviour with structs that we can add data to the end of will help greatly with extending the cache in the future, too.

The new parallel shader loader is a huge improvement to having it mixed in with the shader cache code, makes it a lot easier to reason about what's actually going on there.

We talked about the issue @MutantAura had in the discord and were able to pin it down to the driver cache missing the shaders. It seems like when the NVIDIA driver loads a shader binary, but that shader is not in its driver cache, a small penalty is paid when the shader is first used in a draw/compute invocation when it compiles into its actual binary format. This penalty is not as severe as a full GLSL shader compile, but kind of defeats the point of a shader cache.

I checked, and this also happens on master. I don't think this happened on an older version of the NVIDIA driver...

Funny thing is, this penalty is _not_ paid when compiling from GLSL/SPIR-V source, just when loading the binary file. So we can either force the user to compile the GLSL every time, which is either painfully slow when the driver cache is missing, or just moderately slow when the driver cache is present... or load from binary and let the user experience small stutters as the driver performs its last hurdle getting these shaders ready for the GPU.

Since this is how it works on master right now, I'm OK with leaving it like this and coming up with a solution later. This could be:

- A system that does dummy draws using each shader as they are compiled, which should ensure the driver is ready to use them ingame.
- A heuristic that compiles the first N shaders from GLSL, and determines from their compilation times whether they were driver cached or not. If they weren't, it will assume the rest are missing from driver cache and compile the rest from GLSL. If they are, it will load them from binary to save potentially a lot of time.
- A user option to always compile from GLSL (or enable the bullet point above), if the user greatly prefers reducing stuttering over load times)

On master, it's generally unlikely that users trigger the stuttery behaviour, as when they trigger a driver update or shader version mismatch, they will recompile from GLSL. Losing their driver cache for other reasons happens less frequently.

Either way, this PR is great. I would recommend this PR to everyone I know.

Thog (approved):

PR #3207, opened 2022-03-16
De-tile GOB when DMA copying from block linear to pitch kind memory regions

### LINKED ISSUE #2734: Snack World: The Dungeon Crawl unplayable on current version (and previous versions too)

Game opens and goes through the cinematic animation, but when you start a new game FPS drops to 0 and there's some texture corruption.
Problem persists after character creation, opening menus basically freezes the game for a while. FPS is unstable and keeps going to 0.

Just opening the game.

### Log file

[Ryujinx_1.0.7073_2021-10-13_09-05-19.log](https://github.com/Ryujinx/Ryujinx/files/7337962/Ryujinx_1.0.7073_2021-10-13_09-05-19.log)

### Environment?

- Ryujinx version: 1.0.7073
- Game version: 1.0.0
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 5 3600
  - GPU: RX 580
  - RAM: 16GB
- Applied Mods : No

### Additional context?

ryujinx_capture_2021-10-13_09-08-52

gdkchan:
The ordering of GOB blocks on block linear textures is controlled by some parameters like the GOB sizes (`gobBlocksInY`, `gobBlocksInZ`, etc), and theres usually also another one controlling if the texture is linear or block linear. But it seems that this only controls whenever GOBs are used and their ordering, it does not control the ordering of "sectors" inside of a GOB.

First, let's define a GOB. A GOB is a unit of 512 bytes, divided into 8 lines of 64 bytes each. The ordering of the data inside of the GOB seems to be controlled by the memory "kind" (which is part of the page table entry). Usually this doesn't matter as drivers are expected to "do the right thing" and use a block linear kind for block linear texture, and a pitch kind for linear ones. But it turns out that the OpenGL driver does something weird, it copies data from a block linear kind region to a pitch one, using the DMA engine. I beleive this has the effect of "de-tiling" the GOB. This changes implements emulation of this behaviour, by using a new copy function when there is a kind mismatch.

This fixes two types of texture corruption on OpenGL games. The first is the "fuzzy" textures, which is caused by the driver reading the data and expecting "GOB linear" data. The second is general corruption caused by the copy sourcing data from a texture, and was fixed by forcing a flush in those cases.

Before:
image
image
After:
image
image

Needs general testing, but I'm mainly interested on tests on OpenGL games that had corrupted textures. Would be nice if someone could test Snack World aswell, to see if the issue can be closed.
Closes #2734.


GamerzHell9137:
Fixes graphical issues in Professor Layton 
image

LukeWarnut:
Fixes rendering in Cartoon Network: Battle Crashers

PR:
image
image

Master:
image
image

LukeWarnut:
Fixes Putt-Putt Travels Through Time and likely the other Humongous Entertainment games.

PR:
image
image

Master:
image
image

LukeWarnut:
River City Girls Zero has both fixes and a regression.

- The cutscene skip animation is now visible
- The title screen logo is now slightly fixed
- Textures of the models in the image gallery are slightly fixed; now being visible when zooming in.
- Menu options are now rendered improperly.

image
image
image

gdkchan:
I looked at the River City Girls Zero title screen issue. The base level of the texture looks fine, but all other levels are broken. CPU is reading some data at the end of the base level and using it as the data for the second level. Haven't looked at the other levels but I beleive it's a similar issue. Not sure what is going wrong, and the only DMA copy before it copies the incorrect data into the second level is that of the base level.

Anyway, all those changes are caused by it flushing the source texture region on DMA copy now. I also checked the menu selector and it only has one level, but the bottom lines are corrupted. Maybe another case of it loading garbage from memory where it didn't before because it was not flushing the source textures, so it just had no data there. Haven't looked at Zombies ate my neighbour, but it seems to be the same case of "base level is fine, other levels are corrupt".

I think in general this change is an improvement, and the remaining issues should be handled separately.

riperiperi (approved):
Really nice catch here, this should fix a lot of games.

Thog (approved):

PR #3210, opened 2022-03-17
hle: Some cleanup

AcK77:
This PR cleaned up a bit the HLE folder and the VirtualFileSystem one, since we use LibHac, we can use some class of it directly instead of duplicate things. The "Content" of VFS folder is removed since it should be handled in the NCM service directly.
A larger cleanup should be done later since there is still be duplicated code here and there.


Ryujinx.HLE/FileSystem/ContentPath.cs
```diff
@@ -0,0 +1,82 @@
...
+using Path = System.IO.Path;
+
+namespace Ryujinx.HLE.FileSystem
+{
+    internal static class ContentPath
+    {
+        public const string SystemContent    = "@SystemContent";
+        public const string UserContent      = "@UserContent";
+        public const string SdCardContent    = "@SdCardContent";
+        public const string SdCard           = "@Sdcard";
```
emmauss, line 16: I remember there being an issue where `@SdCard` was requested instead. I can't find the issue now as github is currently broken.

Thog: Did you find the issue again?

Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -238,6 +238,8 @@ public Horizon(Switch device)
             SurfaceFlinger = new SurfaceFlinger(device);
 
             InitializeAudioRenderer();
+
```
gdkchan, line 241: I'd remove the separation here.

gdkchan (approved):
I took a quick look, it looks fine in general.

Ryujinx.HLE/FileSystem/ContentPath.cs
emmauss: No. `@SdCard` must have been a mistake on my end, and `@Sdcard` was the actual token.

Thog (approved):

Issue #3021, opened 2022-01-21
OBS Game Capture is capturing incorrect viewport


Migs3:
### What's the issue you encountered?

Attempting to capture the image with OBS Game Capture method causes image to be flipped and capturing an incorrect .

### How can the issue be reproduced?

Launch Ryujinx, Launch OBS, Add Game Capture Method, select Ryujinx, see problem.

Here is everything as I experience it:
- The image is always flipped Vertically.  Regardless of any of the below changes.
- The area captured is affected by Graphics Scaling options and Docked/Handheld modes.
- Fullscreen vs windowed also has an effect on the captured area.
- Here are a few examples:

Again, all of the below are vertically flipped:
1. Native Scaling, Handheld mode, Windowed mode = Correct captured area at minimum window size, changing window size causes it to "shrink" in OBS.
2. Native Scaling, Handheld mode, Fullscreen mode = Capturing full pixels but it's very small size in OBS, there's a very large Black area, looks to be like 1/4 of the screen is where the game is, and 3/4 is no content/black area. (like 720p in a 1440p canvas)
3. Native Scaling, Docked mode, Windowed mode = Limited captured area, changes depending on window size
4. Native Scaling, Docked mode, Fullscreen mode = Similar to 2 above, but slightly bigger, it's like 960p image inside a 1440p canvas.
5. 2X Scaling, Docked mode, Windowed mode = Similar to 3, limited captured area. only "more zoomed"
6. 2X Scaling, Docked mode, Fullscreen mode = Limited captured area, Looks like it's capturing about 1080p of a 1440p image
7. 2X Scaling, Handheld mode, Fullscreen mode = Correct captured area
8. 2X Scaling, Handheld mode, Windowed mode = Similar to 5, limited captured area. only a little less "zoomed"

Most other games/apps capture just fine using the Game Capture method without needing to do anything, so this seems like a problem with how this is getting rendered internally.

Not expecting any kind of response/fix, just pointing out the issue for visibility and maybe helping others set theirs up to work correctly for those that don't know about it and trying to use OBS to stream/record with.

Although the fix could potentially help some incorrect rendering in the future possibly.

### Log file

N/A

### Environment?

- Ryujinx version: 1.0.7181
- Game version: Doesn't matter
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 9 5900X
  - GPU: NVidia RTX 2080
  - RAM: 32GB
- Applied Mods : No

### Additional context?

[...]

PR #3217, opened 2022-03-20
Fix OpenGL issues with RTSS overlays and OBS Game Capture

riperiperi:
OpenGL game overlays and hooks tend to make a lot of assumptions about how games present frames to the screen, since presentation in OpenGL kind of sucks and they would like to have info such as the size of the screen, or if the contents are SRGB rather than linear.

There are two ways of getting this. OBS hooks swap buffers to get a frame for video capture, but it actually checks the bound framebuffer at the time. I made sure that this matches the output framebuffer (the window) so that the output matches the size. RTSS checks the viewport size by default, but this was actually set to the last used viewport by the game, causing the OSD to fly all across the screen depending on how it was used (or res scale). The viewport is now manually set to match the output framebuffer size.

In the case of RTSS, it also loads its resources by destructively setting a pixel pack parameter without regard to what it was set to by the guest application. OpenGL state can be set for a long period of time and is not expected to be set before each call to a method, so randomly changing it isn't great practice. To fix this, I've added a line to set the pixel unpack alignment back to 4 after presentation, which should cover RTSS loading its incredibly ugly font.

- RTSS and overlays that use it should no longer cause certain textures to load incorrectly. (mario kart 8, pokemon legends arceus)
- OBS Game Capture should no longer crop the game output incorrectly, flicker randomly, or capture with incorrect gamma.

This doesn't fix issues with how RTSS reports our frame timings.

Closes https://github.com/Ryujinx/Ryujinx/issues/3021

### DIFF
--- a/Ryujinx.Graphics.OpenGL/Pipeline.cs
+++ b/Ryujinx.Graphics.OpenGL/Pipeline.cs
@@ -1478,6 +1478,11 @@ namespace Ryujinx.Graphics.OpenGL
             _currentComponentMasks |= componentMaskAtIndex;
         }
 
+        public void RestoreClipControl()
+        {
+            GL.ClipControl(_clipOrigin, _clipDepthMode);
+        }
+
         public void RestoreScissor0Enable()
         {
             if ((_scissorEnables & 1u) != 0)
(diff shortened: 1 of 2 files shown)


XyrEmulation:
OBS Game Capture seems to work fine now
Ryujinx - OBS

riperiperi:
Closes #3021

gdkchan (approved):
lgtm, nice to see this finally fixed.

Thog (approved):

gdkchan:
> Closes #3021

You should put it on the PR description to close automatically on merge.

Essasetic:
Also fixes MangoHUD on OpenGL. Using "MANGOHUD_DLSYM=1 mangohud" works as intended now.
unknown
.

AlexUnwinder:
I'll fix pixel unpack state corruption in the next beta of RTSS, thanks for reporting.

PR #3219, opened 2022-03-20
Don't restore Viewport 0 if it hasn't been set yet.

riperiperi:
Fixes a driver crash when starting some games caused by #3217

Also fixes the possibility of this happening with DrawTexture, which existed before.


gdkchan (approved):
Thanks for the quick fix, triangle strategy seems fine now.

PR #3225, opened 2022-03-22
Support NVDEC H264 interlaced video decoding and VIC deinterlacing

gdkchan:
## Background

Video interlacing is the process where a full video frame (called progressive frame) is halved vertically, keeping either the odd or even lines. Whenever the even or odd lines are used is alternated on each frame. That is, if the first interlaced frame consists of even lines, the second one will have only odd lines, and the third will have even lines again, etc.

Example (image not mine):
image

## NVDEC interlacing support

The NVDEC implementation on the emulator had no support for interlaced video decoding. While we use FFMPEG for H264 decoding, and FFMPEG does support interlaced video, the frame that it returns is not exactly what NVDEC would. NVDEC essentially writes 2 frames, each one with half the height of the full progressive frame. The first one contains even fields, while the second one contains odd fields. Meanwhile, FFMPEG returns a full progressive frame. So we need to extract the even fields of the FFMPEG frame, and then write it as the first frame, then extract the odd fields and write it as the second frame.

## VIC deinterlacing support

VIC also supports deinterlacing, but such support was not implemented on the emulator. This PR also implements that part. There are several deinterlacing modes that are supported by VIC:

### BOB

This is a very simple method that simple doubles the lines on each interlaced frame. This way one can re-create the full frame, but some vertical resolution is lost as you have duplicate lines. On static scenes, this creates a very noticeable effect where the lines appear to "jump" up and down, because the fields on each frame are alternating, that is, one is even and the other is the odd line, so they are not from the exact same position of the scene. This makes thing shift up and down by 1 pixel. On high movement scenes, this effect is not noticeable because everything is already moving anyway.

Example of line "jumping" caused by BOB deinterlacing:

https://user-images.githubusercontent.com/5624669/159551909-ecea6672-01f5-4eb0-9f70-61607cbce988.mp4

### Weave

This method fills the missing fields on the frame with the ones from the previous frame. On static scenes, this can re-create a perfect image since nothing is moving. However, on high movement scenes it creates very noticeable artifacts, where objects on every even or odd line appears displaced. That is because the two frame are from different points in time, so the lines that were copied from the previous frame might have objects at slightly different positions than the current frame.

Example of artifacts caused by Weave method:
image

### Motion adaptive (New BOB, DiSi1)

You may have noticed that the two methods above have opposite results. One is good for static scenes but not moving ones, while the other is good for high movement scenes but not static ones. Motion adaptive methods tries to get the "best of both worlds" by choosing which method to use based on the amount of movement for a given pixel. If the pixel did not move, then it uses Weave method, since it results in a perfect image for static scenes, and it does not create the "jump" effect on alternating frames. If the pixel did move, then it uses BOB method, as it does not introduce the artifacts that Weave does on high movement scenes.

Those motion adaptive methods works by first estimating the motion of the pixel between two frames (part and future frames, that are supplied to VIC), and then calculates a movement vector for each pixel based on that. The final output is then an interpolation of the Weave and BOB outputs, were the "weight" is defined by the movement vector magnitude.

The emulator implementation for this mode was simplified, and might not fully match hardware. That is because the exact algorithm that it uses for motion estimation, interpolation, etc, are unknown. So instead I implemented a deinterlacer that just picks BOB or Weave output based on whenever the pixels on past and future frames match or not. From my tests, it produces very good results, even without the more sophisticated approaches.

## What does it affect?

The only game that is know to use interlaced videos is Layton's Mystery Journey (and from what I was told, apparently the Japanese version is not using interlaced videos).

The first problem was that, since there was no deinterlacing, the input frame on VIC only had half the length. This caused a crash on the blend step as it would try to read out of bounds. The second problem was that the output surface is slightly larger than the region of memory mapped. The driver allocates enough memory for a 2048x1080 frame while the output frame size is 2048x1088 (note that the real size is 1920x1080, the 2048 is due to VIC required pitch alignment of 256 bytes, and 1088 is due to H264 required width/height alignment of 16, since the macro block size is 16x16). I tried looking at the VIC structures and documentation to try finding something that could cause the output to be cropped to 1080, but found nothing that would apply in this case, so I have added a new write method that skips unmapped regions, and started using it on VIC. This is making an assumption that this is a "bug" on the driver's end, and that unmapped memory access from VIC does not cause failures.

Of course, even with this fixed, it would be pointless without deinterlace support and NVDEC writing the output correctly, so that's mostly what this PR does.

As a result, videos on this game now works, rather than crashing instantly:
image

### DIFF
--- a/Ryujinx.Graphics.Vic/Image/BufferPool.cs
+++ b/Ryujinx.Graphics.Vic/Image/BufferPool.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Graphics.Vic.Image
         /// If the required buffer is larger than this, it won't be
         /// added to the pool to avoid long term high memory usage.
         /// </summary>
-        private const int MaxBufferSize = 2048 * 1280;
+        private const int MaxBufferSize = 2048 * 2048;
 
         private struct PoolItem
         {
--- a/Ryujinx.Graphics.Gpu/Memory/MemoryManager.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/MemoryManager.cs
@@ -115,6 +115,65 @@ namespace Ryujinx.Graphics.Gpu.Memory
             }
         }
 
+        public ReadOnlySpan<byte> GetSpanMapped(ulong va, int size, bool tracked = false)
+        {
+            bool isContiguous = true;
+            int mappedSize = 0;
+
+            if (ValidateAddress(va) && GetPte(va) != PteUnmapped && Physical.IsMapped(Translate(va)))
+            {
+                ulong endVa = va + (ulong)size;
+                ulong endVaAligned = (endVa + PageMask) & ~PageMask;
+                ulong currentVa = va & ~PageMask;
+
+                int pages = (int)((endVaAligned - currentVa) / PageSize);
+
+                for (int page = 0; page < pages - 1; page++)
+                {
+                    ulong nextVa = currentVa + PageSize;
+                    ulong nextPa = Translate(nextVa);
+
+                    if (!ValidateAddress(nextVa) || GetPte(nextVa) == PteUnmapped || !Physical.IsMapped(nextPa))
+                    {
+                        break;
+                    }
+
+                    if (Translate(currentVa) + PageSize != nextPa)
+                    {
+                        isContiguous = false;
+                    }
+
+                    currentVa += PageSize;
+                }
+
+                currentVa += PageSize;
+
+                if (currentVa > endVa)
+                {
+                    currentVa = endVa;
+                }
+
+                mappedSize = (int)(currentVa - va);
+            }
+            else
+            {
+                return ReadOnlySpan<byte>.Empty;
+            }
+
+            if (isContiguous)
+            {
+                return Physical.GetSpan(Translate(va), mappedSize, tracked);
+            }
+            else
+            {
+                Span<byte> data = new byte[mappedSize];
+
+                ReadImpl(va, data, tracked);
+
+                return data;
+            }
+        }
+
         /// <summary>
         /// Reads data from a possibly non-contiguous region of GPU mapped memory.
         /// </summary>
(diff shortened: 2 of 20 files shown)


Thog (approved):

Ryujinx.Graphics.Vic/Image/BufferPool.cs
```diff
@@ -14,7 +14,7 @@ class BufferPool<T>
         /// If the required buffer is larger than this, it won't be
         /// added to the pool to avoid long term high memory usage.
         /// </summary>
-        private const int MaxBufferSize = 2048 * 1280;
+        private const int MaxBufferSize = 2048 * 2048;
```
gdkchan, line 17: Changed this value as the old one was too low. For example, it wouldn't allow the buffer for a 2048x1088 (after alignment) frame to be pooled. In addition to that the `SurfaceReader` buffers are also pooled. This helped to get rid of some periodic stutter on Layton's Mystery Journey videos. I believe it should probably help other games using 1080p videos with pitch output aswell.

gdkchan, line 17: Changed this value as the old one was too low. For example, it wouldn't allow the buffer for a 2048x1088 (after alignment) frame to be pooled. In addition to that the `SurfaceReader` buffers are also pooled. This helped to get rid of some periodic stutter on Layton's Mystery Journey videos. I believe it should probably help other games using 1080p videos with pitch output aswell.

meisiek:
This fixes every video in STAR WARS Episode I Racer. On master everything was 3 times on the screen. Now it looks correct.

Master:
grafik
grafik

#3225:
grafik
grafik

riperiperi (approved):
Nice work, great to see this video finally working.

Is `WriteMapped` necessary due to the way VIC ends up writing surfaces, or was it added to make things less fatal if unmapped surfaces are written?

gdkchan:
> Is `WriteMapped` necessary due to the way VIC ends up writing surfaces, or was it added to make things less fatal if unmapped surfaces are written?

The latter, but in this case, the surface is not fully unmapped, only the padding region is unmapped. I commented about it on the PR description.

> The second problem was that the output surface is slightly larger than the region of memory mapped. The driver allocates enough memory for a 2048x1080 frame while the output frame size is 2048x1088 (note that the real size is 1920x1080, the 2048 is due to VIC required pitch alignment of 256 bytes, and 1088 is due to H264 required width/height alignment of 16, since the macro block size is 16x16). I tried looking at the VIC structures and documentation to try finding something that could cause the output to be cropped to 1080, but found nothing that would apply in this case, so I have added a new write method that skips unmapped regions, and started using it on VIC. This is making an assumption that this is a "bug" on the driver's end, and that unmapped memory access from VIC does not cause failures.

Although this problem actually happens on the chroma plane (that has half the height, 544, due to chroma subsampling, but driver maps just enough for 540 lines). This problem only happens on 2 frames, the other frames are fine, but I think that's because it maps more memory after so it can just overwrite the newly mapped region instead. It doesn't seems to cause any problem on the actual video output.

PR #3205, opened 2022-03-15
amadeus: Improve and fix delay effect processing

Thog:
This rework the delay effect processing by representing calculation with the appropriate matrix and by unrolling some loops in the code.
This allows better optimization by the JIT while making it more readable.

Also fix a bug in the Surround code path found while looking back at my notes.

PR #3230, opened 2022-03-23
amadeus: Update to REV11

Thog:
This should implement all ABI changes from REV11 on 14.0.0

As Nintendo changed the channel disposition for "legacy" effects (Delay, Reverb and Reverb 3D) to match the standard channel mapping, I took the liberty to just remap to the old disposition for now.
The proper changes will be handled at a later date with a complete rewriting of those 3 effects to be more readable (see https://github.com/Ryujinx/Ryujinx/pull/3205 for the first iteration of it).


AcK77 (approved):

gdkchan (approved):
I wouldn't mind waiting a bit for a proper implementation (comments seems to suggest this is just re-using old code rather than implementing the new functionality?), but I guess theres no harm in a stopgap solution either.

PR #3242, opened 2022-03-29
[README.md] Update the `Building` section

voldemort2826:
- Updated the build command.
- Added a bit info for Linux OS.
- Some wording changes.
- Removed an irrelevant info about Ryujinx system files
(I think it should be a part of `Configuration` in the `Features` section).


README.md
```diff
@@ -59,22 +59,23 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
...
+Either clone the repo by using `git clone https://github.com/Ryujinx/Ryujinx`, or download the repo directly from the web interface by clicking the green <strong>Code</strong> button > <strong>Download ZIP</strong>.
 
 ### Step 3
+To build Ryujinx, open the command prompt or terminal and use the `cd` command to navigate to the Ryujinx project folder. Windows users can also quickly access it by holding <kbd>Shift</kbd> in File Explorer, then right clicking and selecting `Open command window here`.
 
-To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command:
-`dotnet build -c Release -o build`
-the built files will be found in the newly created build directory.
+After that, type the following command:  
+`dotnet publish -c Release -r win-x64 -o build-ryujinx-custom /p:DebugType=embedded Ryujinx --self-contained`
```
MutantAura, line 74: I think I prefer the original `dotnet build -c Release -o build` or `dotnet publish -c Release -o build` but I digress. The system built for is automatically obtained and the new directory created is more intuitively named (the user is already fully aware of what they just built). The other arguments are not required and just complicate the steps.

README.md
```diff
@@ -59,22 +59,23 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
 
 ## Building
 
-If you wish to build the emulator yourself, follow these steps:
+Follow these steps to build the emulator on your computer. Please note that we do not offer support for custom builds.
 
 ### Step 1
-Install the X64 version of [.NET 6.0 (or higher) SDK](https://dotnet.microsoft.com/download/dotnet/6.0).
+Install the x64 version of [.NET 6.0 SDK](https://dotnet.microsoft.com/download/dotnet/6.0) for your operating system.
```
MutantAura, line 65: "Or higher" should likely be kept here. There will undoubtedly be a crossover period in the future with .NET 7 etc. where the project hasn't been migrated but the SDK is released.

README.md
gdkchan: `/p:DebugType=embedded` is required to not create PDB files.

voldemort2826: Well it's just one liner, just copy and paste it :D. It's meant to create less files just like the PR builds (from the release.yml).
And of course they are free to change the output folder to anything they wish, but if needed I might add some info about that.

README.md
voldemort2826: Then `6.0` should also be removed from the link, or when the project is migrated to 7.0 just update the Readme. What's your thoughts?

voldemort2826: Then `6.0` should also be removed from the link (just `https://dotnet.microsoft.com/download/dotnet`), or when the project is migrated to 7.0 just update the Readme. We can't tell if 7.0 is compatible with 6.0-project.

README.md
```diff
@@ -59,22 +59,23 @@ The latest automatic build for Windows, macOS, and Linux can be found on the [Of
...
-To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command:
-`dotnet build -c Release -o build`
-the built files will be found in the newly created build directory.
+After that, type the following command:  
+`dotnet publish -c Release -r win-x64 -o build-ryujinx-custom /p:DebugType=embedded Ryujinx --self-contained`
 
-Ryujinx system files are stored in the `Ryujinx` folder. This folder is located in the user folder, which can be accessed by clicking `Open Ryujinx Folder` under the File menu in the GUI.
+Change `win-x64` to `linux-x64` if you use Linux.
 
+The output build files will be in the `build-ryujinx-custom` folder. Excluding that folder, you can delete the project files to save disk space.
```
MacJim, line 78: Consider adding the (new) output directory to `.gitignore` under the "Build results" section.

voldemort2826: I don't think that's necessary since you can change the output directory's location (Might add info for that). And most importantly you're compiling the source locally.

Shahil-Ayato:
Hey, glad you decided to take a go at the readme, imo people don’t rarely give it the time it deserves in other projects hence why I rewrote it awhile back to make it easy for both new and advanced users to understand the project and how to do certain things like building. So after looking through a lot of the changes you wanted to add in, I can’t say I agree with a few of them. Some just add some unneeded redundancy which makes it more annoying for basic users to build the project. The building part was especially something I looked at. I wrote the readme so that the steps for building are both easy and intuitive for any type of user whether they’re new to something like this or not. Finally, I don’t think you need to bold the stuff you did on how to obtain the source code. Other than that I don’t have any other complaints.

voldemort2826:
> unneeded redundancy which makes it more annoying for basic users to build the project

Is it because of Step 3? The "right-clicking then choose" part is too specific for Windows, so I added a bit for Linux, also the `cd` command can be used for both OSes. Personally, I found the "right-click" part is unclear since it didn't tell users to right-click in the location of Ryujinx project. Perhaps I might rewrite this to make it concise, likely I will switch to the `cd` method.
Also I don't think most normal users would want to build the project, rather they would just download the ready made ones. So it's kinda more towards advanced users who wants to "mess" with stuff.

> bold the stuff you did on how to obtain the source code

I actually adapts the "manual guide" style, but if it's a problem then I can change it back.

Thog:
Closing as stale.

PR #3251, opened 2022-04-02
Implement VMAD shader instruction and improve InvocationInfo and ISBERD handling

### LINKED ISSUE #1037: Implement VMAD shader instruction
The VMAD shader instruction is currently not properly implemented, it is not a big issue because they are usually only used on geometry shaders, and for what it is used for on geometry shaders, what we have right now is fine (just move the third operand), however we should implement it properly in case something else uses it.

gdkchan:
This has 2 shader changes.

The first one is the VMAD implementation. This instruction is usually used on geometry and tessellation shaders, it is used to calculate the ISBE offset for the ISBERD instruction. It does basically `PrimitiveIndex * VerticesPerPrimitive + VertexIndex`. The previous implementation would just pass operand C directly, which is "fine" since we only care about the `VertexIndex` value for the ALD (attribute load) on the geometry and tessellation stages. This PR actually implements the multiply-add operation correctly, which does A * B + C rather than just returning C directly. Like all other video instructions, this one can operate on one of the 4 8-bits values on the 32-bits integer, one of the 2 16-bits value, or the full 32-bits value. It can also optionally shift (by 7 or 15) and saturate the result. All values can be signed or unsigned, and also optionally negated. Since the result can be shifted and saturated, the higher bits of the result are required (and shift will be shifting the higher bits to the low 32-bits result, and saturation will clamp values above the 32-bits range). This is where most of the complexity of the implementation comes from.

Implementing this properly revealed other issues. Mainly that the `gl_PrimitiveID` on the `SR_InvocationInfo` system register was incorrect. First, it does not (and can't) map to `gl_PrimitiveID` since the value on InvocationInfo is just a 8-bits value, as evidenced by the VMAD operation extracting a U8 from it. I believe that this value is some NVIDIA specific primitve index that is just used for indexing into the ISBE memory, and as such, I have decided to hardcode it to 0 instead. As I said before, we only care about the `VertexIndex` value, so `0 * VerticesPerPrimitive + VertexIndex` will still result in `VertexIndex`, which is what we want.

Finally, I'm not aware of anything this change fixes on commercial games. It should fix homebrew that uses Nouveau OpenGL and geometry or tessellation shaders. That's because Mesa shader compiler does not use VMAD, instead it uses a less efficient sequence of BFE/AND/XMAD instructions instead. So to sum it up, that's what the shader translator was producing for Mesa and NVIDIA shaders:
Master:
NVIDIA: `VertexIndex`
Nouveau: `PrimitiveID * VerticesPerPrimitive + VertexIndex`
PR:
NVIDIA: `0 * VerticesPerPrimitive + VertexIndex` == `VertexIndex`
Nouveau: `0 * VerticesPerPrimitive + VertexIndex` == `VertexIndex`

Closes #1037


Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs
```diff
@@ -94,31 +94,19 @@ public static void S2r(EmitterContext context)
                 case SReg.InvocationInfo:
                     if (context.Config.Stage != ShaderStage.Compute && context.Config.Stage != ShaderStage.Fragment)
                     {
-                        Operand primitiveId = Attribute(AttributeConsts.PrimitiveId);
-                        Operand patchVerticesIn;
+                        // Note: Lowest 8-bits seems to contain some primitve index,
```
riperiperi, line 97: primitive

riperiperi (approved):
Great work. Will test this with geometry shader homebrew when I get the chance.

riperiperi (approved):
Seems to be working properly. Great job!

Thog (approved):
lgtm

PR #3252, opened 2022-04-02
Remove save data creation prompt

MutantAura:
This prompt was fairly redundant and was just another box to click. 
Creation of mod directories do not require this extra step and it seems weird to keep this one. If there is a particular reason to keep this around then I'm all ears but I currently can't see a scenario where a user wishes to open a directory... then says "No" when asked to create one.

The logger now simply reads when a directory has been created instead.

### DIFF
--- a/Ryujinx/Ui/Widgets/GameTableContextMenu.cs
+++ b/Ryujinx/Ui/Widgets/GameTableContextMenu.cs
@@ -81,14 +81,14 @@ namespace Ryujinx.Ui.Widgets
         private bool TryFindSaveData(string titleName, ulong titleId, BlitStruct<ApplicationControlProperty> controlHolder, in SaveDataFilter filter, out ulong saveDataId)
         {
             saveDataId = default;
-            
+
             Result result = _horizonClient.Fs.FindSaveDataWithFilter(out SaveDataInfo saveDataInfo, SaveDataSpaceId.User, in filter);
 
             if (ResultFs.TargetNotFound.Includes(result))
             {
                 ref ApplicationControlProperty control = ref controlHolder.Value;
 
-                Logger.Info?.Print(LogClass.Application, $"Creating save dir for Title: {titleName} [{titleId:x16}]");
+                Logger.Info?.Print(LogClass.Application, $"Creating save directory for Title: {titleName} [{titleId:x16}]");
 
                 if (Utilities.IsZeros(controlHolder.ByteSpan))
                 {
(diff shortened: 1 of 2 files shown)


Ryujinx/Ui/Widgets/GameTableContextMenu.cs
```diff
@@ -81,28 +81,15 @@ public GameTableContextMenu(MainWindow parent, VirtualFileSystem virtualFileSyst
         private bool TryFindSaveData(string titleName, ulong titleId, BlitStruct<ApplicationControlProperty> controlHolder, in SaveDataFilter filter, out ulong saveDataId)
         {
             saveDataId = default;
-
+
```
gdkchan, line 84:
```suggestion

```

Ryujinx/Ui/Widgets/GameTableContextMenu.cs
```diff
@@ -81,28 +81,15 @@ public GameTableContextMenu(MainWindow parent, VirtualFileSystem virtualFileSyst
...
-                };
-
-                if (messageDialog.Run() != (int)ResponseType.Yes)
-                {
-                    return false;
-                }
-
                 ref ApplicationControlProperty control = ref controlHolder.Value;
 
+                Logger.Info?.Print(LogClass.Application, $"Creating save dir for Title: {titleName} [{titleId:x16}]");
```
gdkchan, line 91:
```suggestion
                Logger.Info?.Print(LogClass.Application, $"Creating save directory for Title: {titleName} [{titleId:x16}]");
```

Ryujinx/Ui/Widgets/GameTableContextMenu.cs
MutantAura: fixed. VS code auto-indenting again!

Ryujinx/Ui/Widgets/GameTableContextMenu.cs
MutantAura: Applied this change but it's worth mentioning the modloader logger message simply reads "dir" and not "directory". I could switch that here too to make it consistent if that's better?

Ryujinx/Ui/Widgets/GameTableContextMenu.cs
gdkchan: You can configure it to automatically remove trailing white space on save.

Ryujinx/Ui/Widgets/GameTableContextMenu.cs
gdkchan: Yea, it should say directory.

MutantAura: Updated the mod directory creation message also 👍

gdkchan (approved):
lgtm.

Thealexbarney:
The reason for the prompt is because creating a save involves more than simply creating a directory. And because there's no easy option to delete a save once created. You can use a homebrew save manager like JKSV to do it, but most users won't know that

MutantAura:
Do you think there's necessarily an issue with that though? While there is more going on in the backend to log what game is which etc. All that is really happening to the user is a directory creation. It isn't adding multiple MB's worth of data to their install and if they had the game available to create a directory then chances are they will eventually play it and auto create one anyway.

AcK77 (approved):

PR #3254, opened 2022-04-03
Better formatted Date and Time

piplup55:
This is a pretty simple change

Before:
btw_i_use_arch_2022-04-03_08_31

After:
btw_i_use_arch_2022-04-03_08_3

### DIFF
--- a/Ryujinx/Ui/Windows/SettingsWindow.glade
+++ b/Ryujinx/Ui/Windows/SettingsWindow.glade
@@ -1332,9 +1332,9 @@
                                       <object class="GtkLabel">
                                         <property name="visible">True</property>
                                         <property name="can-focus">False</property>
-                                        <property name="tooltip-text" translatable="yes">Change System Time</property>
+                                        <property name="tooltip-text" translatable="yes">Change System Date</property>
                                         <property name="halign">end</property>
-                                        <property name="label" translatable="yes">System Time:</property>
+                                        <property name="label" translatable="yes">System Date:</property>
                                       </object>
                                       <packing>
                                         <property name="expand">False</property>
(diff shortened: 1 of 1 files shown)


Niwu34:
I think that displaying the time first and putting the year at the end of the date makes more sense:
test

Shahil-Ayato:
Imo you should have it be (MM/DD/YYYY)

piplup55:
> I think that displaying the time first and putting the year at the end of the date makes more sense: test

i do think your suggested change is a good idea i just don't know about the date, i moved the system time to the front
image

Shahil-Ayato:
Also should make the format of the DDMMYYYY clear in the naming where it says ```System Date```

piplup55:
> Also should make the format of the DDMMYYYY clear in the naming where it says `System Date`

i'll put it in the hover text otherwise it gets abit long in the ui

piplup55:
after adding the formatting of the date to the hover text
image

voldemort2826:
I think if there's debate on time format, just write the month in 3 letters (JAN/FEB/etc.). no more arguments XD

btw the system time UI is kinda ugly tho, there should be better way to change it without clutters (dunno if this is because of gdk ui). 

Proposal (from Windows lol) -- year should be in "text input" form, time can be 24-hour or whatever:

image

Also needs to have "use OS time" in case users mess up.

piplup55:
> I think if there's debate on time format, just write the month in 3 letters (JAN/FEB/etc.). no more arguments XD
> 
> btw the system time UI is kinda ugly tho, there should be better way to change it without clutters (dunno if this is because of gdk ui).
> 
> Proposal (from Windows lol) -- could use a mix of "text input" and "drop-down" form, time can be 24-hour or whatever:
> 
> image
> 
> Also needs to have "use OS time" in case users mess up.

i don't see the point of rewriting the date and time ui if we are moving to a new ui at some point

voldemort2826:
oh yeah, i forgot about avalonia ui

piplup55:
don't think i will keep it like this but this is DDMMYYYY
image

piplup55:
i think this pr is about done, no point is debating which date format is best (clearly it's DD/MM/YYYY)

Moonlacer:
I second the idea that the DD/MM/YYYY format is the best

piplup55:
no longer needed since the move to avalonia is close

PR #3269, opened 2022-04-10
ForceDpiAware: X11 implementation

merryhime:
Detect dpi based on screen dimensions provided by X server. As is standard expected behavior of a Linux application, this can be overridden by the user's Xft.dpi setting.

### DIFF
--- a/Ryujinx.Common/System/ForceDpiAware.cs
+++ b/Ryujinx.Common/System/ForceDpiAware.cs
@@ -12,20 +12,22 @@ namespace Ryujinx.Common.System
         [DllImport("user32.dll")]
         private static extern bool SetProcessDPIAware();
 
-        [DllImport("libX11.so.6")]
-        public static extern IntPtr XOpenDisplay(string display);
+        private const string X11LibraryName = "libX11.so.6";
 
-        [DllImport("libX11.so.6")]
-        public static extern IntPtr XGetDefault(IntPtr display, string program, string option);
+        [DllImport(X11LibraryName)]
+        private static extern IntPtr XOpenDisplay(string display);
 
-        [DllImport("libX11.so.6")]
-        public static extern int XDisplayWidth(IntPtr display, int screenNumber);
+        [DllImport(X11LibraryName)]
+        private static extern IntPtr XGetDefault(IntPtr display, string program, string option);
 
-        [DllImport("libX11.so.6")]
-        public static extern int XDisplayWidthMM(IntPtr display, int screenNumber);
+        [DllImport(X11LibraryName)]
+        private static extern int XDisplayWidth(IntPtr display, int screenNumber);
 
-        [DllImport("libX11.so.6")]
-        public static extern int XCloseDisplay(IntPtr display);
+        [DllImport(X11LibraryName)]
+        private static extern int XDisplayWidthMM(IntPtr display, int screenNumber);
+
+        [DllImport(X11LibraryName)]
+        private static extern int XCloseDisplay(IntPtr display);
 
         private static readonly double _standardDpiScale = 96.0;
         private static readonly double _maxScaleFactor   = 1.25;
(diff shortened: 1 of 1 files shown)


merryhime:
X11 implementation only. Wayland TBD at a later date (once I have a Wayland environment to test / if another dev with one implements it).

`XDG_SESSION_TYPE` is the best user-configurable way out of several non-ideal ways to detect the display server.

Ryujinx.Common/System/ForceDpiAware.cs
```diff
@@ -36,9 +52,28 @@ public static double GetWindowScaleFactor()
                 {
                     userDpiScale = Graphics.FromHwnd(IntPtr.Zero).DpiX;
                 }
-                else
+                else if (OperatingSystem.IsLinux())
                 {
-                    // TODO: Linux support
+                    string xdgSessionType = Environment.GetEnvironmentVariable("XDG_SESSION_TYPE");
+
+                    if (xdgSessionType == null || xdgSessionType == "x11")
```
Thog, line 59: maybe this could work in upper case (``X11``) so might be worth ignoring case here?

Ryujinx.Common/System/ForceDpiAware.cs
```diff
@@ -36,9 +52,28 @@ public static double GetWindowScaleFactor()
...
+                        string dpiString = Marshal.PtrToStringAnsi(XGetDefault(display, "Xft", "dpi"));
+                        if (dpiString == null || !double.TryParse(dpiString, NumberStyles.Any, CultureInfo.InvariantCulture, out userDpiScale))
+                        {
+                            userDpiScale = (double)XDisplayWidth(display, 0) * 25.4 / (double)XDisplayWidthMM(display, 0);
+                        }
+                        XCloseDisplay(display);
+                    }
+                    else if (xdgSessionType == "wayland")
+                    {
+                        // TODO
```
Thog, line 71: A log here could be nice 👍

Ryujinx.Common/System/ForceDpiAware.cs
```diff
@@ -11,6 +12,21 @@ public static class ForceDpiAware
...
+
+        [DllImport("libX11.so.6")]
+        public static extern int XDisplayWidth(IntPtr display, int screenNumber);
+
+        [DllImport("libX11.so.6")]
+        public static extern int XDisplayWidthMM(IntPtr display, int screenNumber);
+
+        [DllImport("libX11.so.6")]
+        public static extern int XCloseDisplay(IntPtr display);
+
```
Thog, line 29:
```suggestion
        private const string LibraryName = "libX11";

        [DllImport(LibraryName)]
        public static extern IntPtr XOpenDisplay(string display);

        [DllImport(LibraryName)]
        public static extern IntPtr XGetDefault(IntPtr display, string program, string option);

        [DllImport(LibraryName)]
        public static extern int XDisplayWidth(IntPtr display, int screenNumber);

        [DllImport(LibraryName)]
        public static extern int XDisplayWidthMM(IntPtr display, int screenNumber);

        [DllImport(LibraryName)]
        public static extern int XCloseDisplay(IntPtr display);

```

Thog (changes requested):

Thog:
> X11 implementation only. Wayland TBD at a later date (once I have a Wayland environment to test / if another dev with one implements it).
> 
> `XDG_SESSION_TYPE` is probably the best user-configurable way out of several non-ideal ways to detect the display server -- open to other ideas.

For reference, to support wayland with the game view, we will also need integration on the SPB side to create the OpenGL context.

Ryujinx.Common/System/ForceDpiAware.cs
edisionnano: As discussed on the Discord server the name of the so only changes when there are ABI-breaking changes thus we want this to stop functioning if libx11.so.7 comes out

merryhime: My experience with soversions was to specify it where able to ensure ABI breakage doesn't happen down the line; if you have another policy I'm happy to change.

Thog (approved):

AcK77 (approved):

PR #3275, opened 2022-04-13
Implement HwOpus multistream functions

gdkchan:
This PR mainly implements the multistream related Opus decoding functions, that are required by MLB The Show 22 and maybe other games. In addition to that, it does some small refactoring to avoid code duplication:
- Implementation of decoding moved to a single method to avoid code duplication.
- Decoding now uses spans rather than streams when possible. Unfortunately Concentus doesn't have overloads of the decoding functions that takes spans, so we still need to do some copies.
- GetWorkBufferSize functions were updated, before it was only taking the ADSP side into account and only calculating the size of the Opus decoder state. But the service also adds an extra size, so the code was updated to reflect that. This shouldn't have any consequence though as we don't really use the transfer memory passed by the game.

MLB The Show 22 still requires other services implemented to work, so it's not playable even with this change, but it's a step towards it.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Audio/HardwareOpusDecoderManager/Decoder.cs
+++ b/Ryujinx.HLE/HOS/Services/Audio/HardwareOpusDecoderManager/Decoder.cs
@@ -1,0 +1,27 @@
+using Concentus.Structs;
+
+namespace Ryujinx.HLE.HOS.Services.Audio.HardwareOpusDecoderManager
+{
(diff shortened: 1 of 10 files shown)


Thog (approved):

AcK77 (approved):

lostromb:
I'll take the "missing span overloads" as a feature request for Concentus. I mainly haven't implemented it yet because it doesn't seem to offer any performance benefits that I was hoping for.

PR #3276, opened 2022-04-14
Allocate work buffer for audio renderer instead of using guest supplied memory

gdkchan:
Right now we use the memory supplied by the guest when possible (when it is contiguous etc, which should be always the case with host memory manager modes) to store audio renderer state. While this is correct, there are some cases where the game unmaps the memory before the audio renderer is fully disposed. This can cause the audio renderer to try accessing the memory and cause an access violation because it was already unmapped.

This change fixes it by allocating our own memory to be used to store audio renderer state. Fixes an access violation crash on Urban Trial Tricky that started happening after #3267 (since before that change, access violation would cause the thread to just retry forever, which was probably masking the error).

Personally I'm not a fan of this change and I'm open to suggestions. I'm also not sure if pinning the array for a long time is a good idea, I heard it impacts GC negatively.

### DIFF
--- a/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
+++ b/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
@@ -55,7 +55,6 @@ namespace Ryujinx.Audio.Renderer.Server
         private uint _processHandle;
         private ulong _appletResourceId;
 
-        private WritableRegion _workBufferRegion;
         private MemoryHandle _workBufferMemoryPin;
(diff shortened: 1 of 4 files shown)


gdkchan:
Changed to use `GC.AllocateArray` and rebased.

EmulationFanatic:
Also fixes a crash on boot on Mutant Year Zero: Road to Eden. Can this be rebased & reviewed by one more?

Thog (approved):

PR #3283, opened 2022-04-18
Add volume control when scrolling over volume in status bar

guycook:
This augments the volume display in the status bar to allow users to {inc/de}crement the audio volume by scrolling the mousewheel while hovering over the display segment.

In terms of implementation this mirrors the existing click-to-mute functionality in that it's completely transient - any adjustment made will not carry over to subsequent launches or even running a new game, nor will it be reflected in the settings window. As with mute, I think this is reasonable behaviour. Movements are 10% per 'click' of the wheel, and the `else if` is needed because scroll events can also be generated horizontally by trackpads. `Round` is needed as you'll start to see `59%` etc. in the UI if you scroll back and forward enough without it.

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1205,6 +1205,21 @@ namespace Ryujinx.Ui
             }
         }
 
+        private void VolumeStatus_Scroll(object sender, ScrollEventArgs args)
+        {
+            if (_emulationContext != null)
+            {
+                if (args.Event.Direction == Gdk.ScrollDirection.Up)
+                {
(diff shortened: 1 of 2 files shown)


Haronee:
It's a nice QoL feature, I like it.
But I think it would be better if the changes in volume were reflected in the settings window, as they're both the exact same setting.

MutantAura:
The way #2507 does this is also transient so this makes more sense than adjusting settings. 
Anything changed at runtime via the widget should not impact the config file and changing that here would make no sense. Vsync, Docked mode etc. all do not update settings.

How it currently is seems correct.

riperiperi (commented):
For some reason, it doesn't seem to work after clicking to mute/unmute... you have to click somewhere else. Apart from that, this is pretty cool.

TSRBerry:
Could you add this for Avalonia as well?

marysaka:
Closing as stale.

Issue #3298, opened 2022-04-24
NOTICE: Drop support of Windows versions prior to Windows 10 1803


Thog:
Hello there,

This is a notice to indicate support drop for Windows versions prior to Windows 10 1803 on **June 1st 2022**.

### Why are we dropping support?
- Windows 7 EOL was on January 13, 2015
- Windows 8 EOL was on January 12, 2016
- Windows 8.1 EOL was on January 9, 2018
- Graphics driver support also were dropped by all vendors meaning that driver bugs will never get addressed.

voldemort2826:
I think #2954 will also depend on this change, specifically the fast memory manager modes.

gdkchan:
> I think #2954 will also depend on this change, specifically the fast memory manager modes.

The emulator will continue working on Windows 7/8 with this change, but it will always use the software memory manager mode there regardless of what the user chooses.

voldemort2826:
Another important point to make is that dropping support makes the emulator development much easier. Having to include workarounds for old OSes is PITA, and we don't know if those might introduce other chains of bugs \*shrug*

OCRBonk:
As a frequent user of W7 on a dual boot system. I totally get dropping support officially considering the amount of things available to newer OSes. Surprised frankly you supported it at all at any point given the age of the OS.

dio-gh:
Hi,

I'd like to point out that some of the points raised in the ticket are somewhat disingenuous.

- Windows 7 is still receiving [Extended Security Updates](https://docs.microsoft.com/en-us/lifecycle/products/windows-7) until January 10, 2023 for VL (volume licensed) purchases of `Professional`, `Enterprise`, and `Professional for Embedded Systems` versions.
- Windows 8.1 is receiving extended support (not limited to security updates) until [January 10, 2023](https://docs.microsoft.com/en-us/lifecycle/products/windows-81) regardless of license type.
- While it is true that none of the vendors are planning to release functionality related updates for the aforementioned platforms anymore, it is incorrect to say that support has been dropped altogether - [NVIDIA vouched](https://nvidia.custhelp.com/app/answers/detail/a_id/5201/~/support-plan-for-windows-7-and-windows-8%2F8.1) they'll keep critical security updates going until September 2024, and AMD seems to be aiming for [something similar](https://hexus.net/tech/news/software/148003-amds-radeon-driver-moves-windows-7-legacy-support/) with their "legacy support model".

[...]

PR #3299, opened 2022-04-24
infra: Warn about support drop of old Windows versions

Thog:
See #3298.

### DIFF
--- a/Ryujinx/Program.cs
+++ b/Ryujinx/Program.cs
@@ -43,7 +43,7 @@ namespace Ryujinx
 
             if (OperatingSystem.IsWindows() && !OperatingSystem.IsWindowsVersionAtLeast(10, 0, 17134))
             {
-                MessageBoxA(IntPtr.Zero, "You are running an outdated version of Windows.\n\nStarting on June 1st 2022, new versions of Ryujinx will only support Windows 10 1803 and newer.\n", $"Ryujinx {Version}", MB_ICONWARNING);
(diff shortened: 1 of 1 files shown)


gdkchan:
You need to make the warning only show once. It would be very annoying to see it every time you open the emulator.

Thog:
> You need to make the warning only show once. It would be very annoying to see it every time you open the emulator.

The point is that it needs to be seen, if they want it to get away they should update.

voldemort2826:
> The point is that it needs to be seen, if they want it to get away they should update.

A checkbox "Don't show this again" could work.

gdkchan:
Also can't you use a GTK dialog instead of calling Windows API directly?

Ryujinx/Program.cs
```diff
@@ -32,8 +32,20 @@ class Program
...
+        private const uint MB_ICONWARNING = 0x30;
+
         static void Main(string[] args)
-        { 
+        {
+            Version = ReleaseInformations.GetVersion();
+
+            if (OperatingSystem.IsWindows() && !OperatingSystem.IsWindowsVersionAtLeast(10, 0, 17134))
+            {
+                MessageBoxA(IntPtr.Zero, "You are running an outdated version of Windows.\n\nStarting on June 1st 2022, new versions of Ryujinx will only support Windows 10 1803 and newer.\n", $"Ryujinx {Version}", MB_ICONWARNING);
```
jduncanator, line 46: I think the text here could be changed to "will only officially support Windows 10 1803"...

Otherwise we'll have people in Discord asking for support on older versions of Windows because "it still runs!". This makes it clearer that it may still run/work, but we won't be officially supporting it.

AcK77 (approved):

PR #3302, opened 2022-04-27
Fix various issues with texture sync

riperiperi:
A variable called _actionRegistered is used to keep track of whether a tracking action has been registered for a given texture group handle. This variable is set when the action is registered, and should be unset when it is consumed. This is used to skip registering the tracking action if it's already registered, saving some time for render targets that are modified very often.

There were two issues with this. The worst issue was that the tracking action handler exits early if the handle's modified flag is false... which means that it never reset _actionRegistered, as that was done within the Sync() method called later. The second issue was that this variable was set true after the sync action was registered, so it was technically possible for the action to run immediately, set the flag to false, then set it to true.

Both situations would lead to the action never being registered again, as the texture group handle would be sure the action is already registered. This breaks the texture for the remaining runtime, or until it is disposed.

It was also possible for a texture to register sync once, then on future frames the last modified sync number did not update. This may have caused some more minor issues.

I've also used `Interlocked` in a few additional places just to ensure coherent access for things accessed from multiple threads.

Seems to fix the Xenoblade flashing bug. Obviously this needs a lot of testing, since it was random chance. I typically had the most luck getting it to happen by switching time of day on the event theatre screen for a while, then entering the equipment screen by pressing X on an event.

May also fix weird things like random chance air swimming in BOTW, maybe a few texture streaming bugs.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
@@ -172,7 +172,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 _syncActionRegistered = true;
             }
 
-            if (Interlocked.CompareExchange(ref _actionRegistered, 1, 0) == 0)
+            if (Interlocked.Exchange(ref _actionRegistered, 1) == 0)
             {
                 _group.RegisterAction(this);
             }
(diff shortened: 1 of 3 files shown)


gdkchan (approved):
lgtm, nice catch. Nice to see the Xenoblade regression finally fixed.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -171,11 +172,9 @@ private void RegisterSync(GpuContext context)
                 _syncActionRegistered = true;
             }
 
-            if (!_actionRegistered)
+            if (Interlocked.CompareExchange(ref _actionRegistered, 1, 0) == 0)
```
gdkchan, line 175: I believe this could just use `Exchange` since setting the value to 1 when its already 1 is harmless.

Thog (approved):

ivooc:
The problem is back in PLA #3309

Issue #3287, opened 2022-04-20
Sword Art Online: Hollow Realization - Language selection not working


MarkLanders777:
### What's the issue you encountered?

The language displayed is different to the one selected</br>

### How can the issue be reproduced?

Start a new game and select Spanish, the game keeps displaying English (even if I change the language in the settings of Ryujinx) in real console it works fine, spanish is selected and displayed as it should be

### Environment?

- Ryujinx version: 1.1.105</br>
- System Specs:
  - OS: *(Windows 11)*
  - CPU: *(i7 9700 non k)*
  - GPU: *(NVIDIA 1060)*
  - RAM: *(16GB)*

Burzumm:
В какой то определенной игре или нет?

Burzumm:
In some game or not?

liamharper2453:
Hi,

Please try using the 'Latin American Spanish' option in Ryujinx instead of 'Spanish' for this game.

MarkLanders777:
> Hi,
> 
> Please try using the 'Latin American Spanish' option in Ryujinx instead of 'Spanish' for this game.

Yep, that fix the issue

Issue #3232, opened 2022-03-25
Change Language for Individual Games in Ryujinx


nicksmmith:
Hello,

Just want to say that I love you guys' work on Ryujinx, awesome job!!!

So, my feature request is pretty much what it says in the title.

The issue that I found is probably very specific for my situation, but maybe more people can benefit from such Feature in Ryujinx;

-My first language is Brazilian Portuguese
-Some games are only in Portuguese
-if I select Brazilian Portuguese, games in Portuguese language only will be in English

Thank you so much! :)

## Feature Request

[ If any section does not apply, replace its contents with "N/A". ]</br>
[ If you do not have the information needed for a section, replace its contents with "Unknown". ]</br>
[ Lines between [ ] (square brackets) are to be removed before posting. ]</br>

[ Please search for existing [feature requests](https://github.com/Ryujinx/Ryujinx/issues) before you make your own request. ]</br>
[ Duplicate requests will be marked as such and you will be referred to the original request. ]

#### Overview:
- [ Include the basic, high-level concepts for this feature here. ]

#### Smaller Details: 
- [ These may include specific methods of implementation etc. ]

#### Nature of Request:  
[ Remove all that do not apply to your request. ]
- Addition
  - [ Ex: Addition of certain original features or features from other community projects. ]
  - [ If you are suggesting porting features or including features from other projects, include what license they are distributed under and what, if any libraries those project use. ]
- Change
- Removal
  - [Ex: Removal of certain features or implementation due to a specific issue/bug or because of low quality code, etc.]

### Why would this feature be useful?
[ If this is a feature for an end-user, how does it benefit the end-user? ]</br>
[ If this feature is for developers, what does it add to Ryujinx that did not already exist? ]

gdkchan:
I think there are better ways to solve this, like for example changing the language selection "priority" to prefer portuguese if brazilian portuguese is not available. But we'd need to see what the Switch does here first.

Anyway, what you request is basically a subset of https://github.com/Ryujinx/Ryujinx/issues/1796. With per-game settings, it would be possible to change that and more.

PR #3303, opened 2022-04-27
Add supported languages to UI

liamharper2453:
Hi,

First contribution here.

I thought it would be useful if supported languages for games was visible in the UI so I have added that as part of this PR.

image

Background on this is #3287 where the user most likely selected Spanish instead of Latin American Spanish and because Spanish was unsupported the game defaulted back to English.

Sometimes the language shown in-game can be misleading to what is actually needed i.e. in Sword Art Online the select language screen specifies Canadian French. However, it is actually only specifically French that is supported.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -174,7 +174,7 @@ namespace Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.Applicati
 
             return ResultCode.Success;
         }
-        
+
         [CommandHipc(22)]
         // SetTerminateResult(u32)
         public ResultCode SetTerminateResult(ServiceCtx context)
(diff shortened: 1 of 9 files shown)


gdkchan:
Would be better if the bug fix and the UI changes were separate PRs imo.

liamharper2453:
> Would be better if the bug fix and the UI changes were separate PRs imo.

It turns out the bug fix wasn't needed. In that issue raised I think they just selected Spanish instead of Latin American Spanish which the game supports. So taken that out. This is now a UI only change and hopefully this work will make distinctions like that clearer and prevent confusion about that in the future.

gdkchan:
I was starting to wonder what exactly you fixed since `(x & (1 << y)) != (1 << y)` is the same as `(x & (1 << y)) == 0` (since only one bit is set, the result can only be either that bit or 0).

The way to fix this might be better language selection priority (like if Spanish is not available, it can prefer Latin American Spanish rather than English). Something like that would also address #3232, but it requires checking what the service does as we don't want to differ from the console here.

liamharper2453:
> I was starting to wonder what exactly you fixed since `(x & (1 << y)) != (1 << y)` is the same as `(x & (1 << y)) == 0` (since only one bit is set, the result can only be either that bit or 0).
> 
> The way to fix this might be better language selection priority (like if Spanish is not available, it can prefer Latin American Spanish rather than English). Something like that would also address #3232, but it requires checking what the service does as we don't want to differ from the console here.

Yeah that's why I did undid it as I realised it was the same thing. When I was investigating I also got thrown off by the latin spanish/spanish difference and was trying to fix something that wasn't even an issue!

Good shout on language priorities.

MarkLanders777:
> > I was starting to wonder what exactly you fixed since `(x & (1 << y)) != (1 << y)` is the same as `(x & (1 << y)) == 0` (since only one bit is set, the result can only be either that bit or 0).
> > The way to fix this might be better language selection priority (like if Spanish is not available, it can prefer Latin American Spanish rather than English). Something like that would also address #3232, but it requires checking what the service does as we don't want to differ from the console here.
> 
> Yeah that's why I did undid it as I realised it was the same thing. When I was investigating I also got thrown off by the latin spanish/spanish difference and was trying to fix something that wasn't even an issue!
> 
> Good shout on language priorities.

True, Latin American Spanish doesn't work if you select Spanish, Xenoblade 2 works selecting just Spanish and not Latin American but in real hardware does work

Thog:
Need a rebase

marysaka:
Closing as stale.

PR #3304, opened 2022-04-27
Restrict cases where vertex buffer size from index buffer type is used

gdkchan:
This fixes some regression introduced by #3253, like some glitches on Xenoblade Chronicles 2 blade awakening cutscenes. First, we can't guess the vertex buffer size from the index buffer type when the draw is not indexed, as in those cases the index buffer type is completely meaningless. This also does not allow it for instanced vertex buffers as I believe it might be possible to access vertices outside of the range in those cases (due to vertex being incremented per instance, might be wrong though). I have verified that it still works for Perky Little Things and Super Mario 64, which is what the previous PR fixed.


Thog (approved):

riperiperi (approved):
LGTM.

Issue #3309, opened 2022-05-01
Pokemon Legends Arceus GFX regression


MutantAura:
### What's the issue you encountered?

Since #3302 there have been some graphical regressions in the above title (see images below).

### How can the issue be reproduced?

Load into PLA and enter and exit the main hub building (or any building to refresh the area). 

### Log file

[Ryujinx_1.1.108_2022-05-01_17-07-55.log](https://github.com/Ryujinx/Ryujinx/files/8599278/Ryujinx_1.1.108_2022-05-01_17-07-55.log)

### Environment?

- Ryujinx version: 1.1.108
- Game version: 1.1.0
- System Specs:
  - OS: *Windows 10*
  - CPU: *R5 5600X*
  - GPU: *NVIDIA RTX 3070*
  - RAM: *32GB DDR4*

### Additional context?
1.1.106: 
![image](https://user-images.githubusercontent.com/44103205/166154723-ceb328a9-d040-4959-9538-9402e101cc8a.png)
1.1.107: 
![image](https://user-images.githubusercontent.com/44103205/166154729-c450e945-ffaf-4817-9ddd-12fb1df4f66a.png)
![image](https://user-images.githubusercontent.com/44103205/166154740-df9f8f0b-68ff-48be-8210-78fdb8deec02.png)
![image](https://user-images.githubusercontent.com/44103205/166154760-3e626bae-f83a-42f8-ae7b-15333d4fb920.png)

innerspeakerz:
I am experiencing the same problem

ivooc:
This problem is back:
![image](https://user-images.githubusercontent.com/44483972/169920987-5c1fdd9a-6325-4d9c-9e31-919d53677745.png)
![image](https://user-images.githubusercontent.com/44483972/169921043-89f45016-547b-48fe-b056-29f1467fb4e2.png)

riperiperi:
Can you not comment this on every PR? I think there was a very low chance of this happening even before the change to texture sync, the regression was that it became _much higher_.

ivooc:
> Can you not comment this on every PR? I think there was a very low chance of this happening even before the change to texture sync, the regression was that it became _much higher_.

sorry, I thought it would be wise to link the related problems but I guess it was unnecessary. I played around a bit and it is happening far less than when the issue appeared before.

PR #3311, opened 2022-05-01
Fix flush action from multiple threads regression

riperiperi:
If two or more threads encounter a region of memory where a read action has been registered, then they must _both_ wait on the data.

Clearing the action before it completed was causing the null check above to fail, so the action would only be run on the first thread, and the second would end up continuing without waiting. Depending on what the game does, this could be disasterous.

This fixes a regression introduced by #3302 with Pokemon Legends Arceus, and possibly Catherine. There are likely other affected games. What is fixed in that PR should still be fixed.

### DIFF
--- a/Ryujinx.Memory/Tracking/RegionHandle.cs
+++ b/Ryujinx.Memory/Tracking/RegionHandle.cs
@@ -144,9 +144,11 @@ namespace Ryujinx.Memory.Tracking
                 {
                     lock (_preActionLock)
                     {
-                        RegionSignal action = Interlocked.Exchange(ref _preAction, null);
+                        _preAction?.Invoke(address, size);
 
-                        action?.Invoke(address, size);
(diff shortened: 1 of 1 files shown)


MutantAura:
Closes #3309

gdkchan (approved):
Nice, thanks for the quick fix.

MutantAura:
Can confirm this fixes PLA's issues: 
image
image

MarkLanders777:
Catherine seems fixed too 

fixed

Thog (approved):

ivooc:
The problem in PLA came back #3309

VBuuBuu22:
@riperiperi can u help me pls? i don't know how to do to fix this issue. Can you guide me in detail, pls?

MetrosexualGarbodor:
@VBuuBuu22 "merged" means that this is fixed in Ryujinx already. Please go to Discord for help next time.

PR #3316, opened 2022-05-04
Rewrite kernel memory allocator

gdkchan:
This rewrites the kernel memory allocator, with the goal of making the code cleaner and more readable. I'm not expecting any bugfix or performance improvement from this.

Should be tested with games that uses Map/UnmapPhysicalMemory syscalls and that in general does a lot of small allocations, like UE4 games and Super Smash Bros Ultimate.


Ryujinx.HLE/HOS/Kernel/Memory/KPageBitmap.cs
```diff
@@ -0,0 +1,298 @@
...
+
+                    bitsCount /= 2;
+                    mask >>= bitsCount;
+                }
+
+                return selected;
+            }
+        }
+
+        private const int UlongBitSize = sizeof(ulong) * 8;
```
riperiperi, line 80: `ULongBitSize`? Not sure about this one, the only other examples of `ULong` are a SoftFallback method and NUnit's own randomizer, and `Ulong` just has a variable in HashState.

Ryujinx.HLE/HOS/Kernel/Memory/KPageBitmap.cs
```diff
@@ -0,0 +1,298 @@
...
+            }
+
+            private bool GenerateRandomBit()
+            {
+                if (_bitsAvailable == 0)
+                {
+                    RefreshEntropy();
+                }
+
+                bool bit = (_entropy & 1) != 0;
```
riperiperi, line 27: Is entropy ever non-zero?

Ryujinx.HLE/HOS/Kernel/Memory/KPageBitmap.cs
gdkchan: Should it be `UInt64BitSize` then?

Ryujinx.HLE/HOS/Kernel/Memory/KPageBitmap.cs
gdkchan: On the official firmware yes, it generates a random number using MT19973 (Mersenne Twister RNG) with a seed that I think comes from one of the hardware random number registers. On the emulator I have forced it to zero because I believe having deterministic addresses is desirable here, if we had random addresses it would make debugging harder.

So the short answer is, no on the emulator it is always zero right now.

Ryujinx.HLE/HOS/Kernel/Memory/KPageBitmap.cs
riperiperi: That does appear a lot more often, if only because of BinaryReader being used a lot. That would be fine.

riperiperi (approved):

Thog (approved):

PR #3319, opened 2022-05-05
Fix shared memory leak on Windows

gdkchan:
Fixes the following issues:
- Windows requires all view mappings to be unmapped to release the memory. Just closing the file mapping handle is not enough. This was not being done which caused several GBs of memory to leak after the emulation was stopped (or on multi-title games, after switching from one title to the other).
- When using forced 4KB view mapping, if the mapping was done on memory that was already mapped, it would fail. Now it unmaps the memory before mapping in this case. This matches the other implementation that does not work page-by-pagee.
- When using forced 4KB view mappings, if there was a region in memory adjacent with the regular mapping, it would try to coalesce the two which might be invalid as the placeholder manager does not keep track of those 4KB mappings. This was fixed by not reserving the range on the placeholder manager for those regions.
- `UnmapView` was updating the tree outside of the lock, which makes the code not thread safe, it could lead to corruption of the tree if two threads tried to unmap memory at the same time.
- The RO service session incrementss the memory manager reference count, and decrements it when the session is closed. But it was possible for it to not be decremeented if the `ArmProcessContext` was disposed before the session object was disposed. This is because the memory manager instance is set to null. This was fixed by storing the memory manager instance on the RO object.

The first 3 issues were validated with the following test (which is not included as it allocates 1GB of memory in the loop to ensure there's no memory leak, and I think this might be too high for a general test):
```cs
        [Test, Explicit]
        public void Test_AliasLeak()
        {
            for (int i = 0; i < 2000; i++)
            {
                using MemoryBlock backing = new MemoryBlock(0x40000000, MemoryAllocationFlags.Mirrorable);
                using MemoryBlock toAlias = new MemoryBlock(0x10000, MemoryAllocationFlags.Reserve | MemoryAllocationFlags.ViewCompatible);
                using MemoryBlock toAlias2 = new MemoryBlock(0x10000, MemoryAllocationFlags.Reserve | MemoryAllocationFlags.ViewCompatible | MemoryAllocationFlags.ForceWindows4KBViewMapping);

                toAlias.MapView(backing, 0x1000, 0, 0x4000);
                toAlias.MapView(backing, 0x1000, 0x2000, 0x4000);
                toAlias.MapView(backing, 0x1000, 0x4000, 0x4000);
                toAlias.MapView(backing, 0x9000, 0x8000, 0x4000);
                toAlias.UnmapView(backing, 0x3000, 0x1000);

                toAlias2.MapView(backing, 0x1000, 0, 0x4000);
                toAlias2.MapView(backing, 0x1000, 0x2000, 0x4000);
                toAlias2.MapView(backing, 0x1000, 0x4000, 0x4000);
                toAlias2.MapView(backing, 0x9000, 0x8000, 0x4000);
                toAlias2.UnmapView(backing, 0x3000, 0x1000);

                toAlias.Write(4, 0xcafe);
                toAlias2.Write(0, 0xbadc0de);
                Assert.AreEqual(Marshal.ReadInt32(backing.Pointer, 0x1004), 0xcafe);
                Assert.AreEqual(Marshal.ReadInt32(backing.Pointer, 0x1000), 0xbadc0de);
            }
        }
```
And the last fix was validated on Super Mario 64 from Super Mario 3D All Stars.
All these issues were introduced on #2954, although its possible that the memory leak already existed (I did not test if versions prior to that leaked when stopping and restarting emulation).
Testing is welcome as always.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
@@ -30,6 +30,7 @@ namespace Ryujinx.HLE.HOS.Services.Ro
         private List<NroInfo> _nroInfos;
 
         private KProcess _owner;
+        private IVirtualMemoryManager _ownerMm;
 
         private static Random _random = new Random();
 
@@ -38,6 +39,7 @@ namespace Ryujinx.HLE.HOS.Services.Ro
             _nrrInfos = new List<NrrInfo>(MaxNrr);
             _nroInfos = new List<NroInfo>(MaxNro);
             _owner    = null;
+            _ownerMm  = null;
         }
 
         private ResultCode ParseNrr(out NrrInfo nrrInfo, ServiceCtx context, ulong nrrAddress, ulong nrrSize)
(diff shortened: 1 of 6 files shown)


Thog (approved):

riperiperi (approved):
LGTM.

PR #3321, opened 2022-05-05
Add alternative "GL" enum values for StencilOp

riperiperi:
This PR adds the alternative enum values for StencilOp. Similar to the other enums, I added these with the same names but with Gl added to the end. These are used by homebrew using Nouveau, though they might be used by games with the official Vulkan driver.

Reference for the values from envytools:

https://github.com/envytools/envytools/blob/39d90be897f41434d67277ebdf244d6bd419ecd9/rnndb/graph/nv_3ddefs.xml#L77

Fixes some broken graphics in the Citra RetroArch core, such as missing shadows in Mario Kart 7. Likely fixes other homebrew.

Before:
image

After:
image


gdkchan (approved):
lgtm, thanks!

Thog (approved):

PR #3326, opened 2022-05-07
hid: Various fixes and cleanup

AcK77:
This PR implement `am IsGamePlayRecordingSupported` needed by RetroArch. (and fixes `SetGamePlayRecordingState` which uses the same field).

Additionnally to that, RetroArch have an issue related to the number of max players, which is set to 16 as default (fixed here: https://github.com/libretro/RetroArch/commit/aefc0de8d7c2bf30b78002c75eae12f21fb6176f). Since the number of max players on Switch is 8, it throw an exception in our `hid` code because we don't check anything. This is now fixed accordingly to RE.
Now `SetNpadJoyAssignmentModeSingleByDefault`, `SetNpadJoyAssignmentModeSingle`, `SetNpadJoyAssignmentModeDual`  and `SetNpadJoyAssignmentModeSingleWithDestination` are implemented (partially for some)

I've removed the `Hid` suffix on some type because it make no sense and it's not accurate.

RetroArch can be launched now:
image

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
@@ -35,5 +35,10 @@ namespace Ryujinx.HLE.HOS.Services.Hid.HidServer
             PlayerIndex.Unknown  => NpadIdType.Unknown,
             _                    => throw new ArgumentOutOfRangeException(nameof(index))
         };
+
+        public static bool IsValidNpadIdType(NpadIdType npadIdType)
+        {
+            return (npadIdType <= NpadIdType.Player8 || npadIdType == NpadIdType.Handheld || npadIdType == NpadIdType.Unknown);
+        }
     }
 }
--- a/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
@@ -38,7 +38,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid.HidServer
 
         public static bool IsValidNpadIdType(NpadIdType npadIdType)
         {
-            return (npadIdType <= NpadIdType.Player8 || npadIdType == NpadIdType.Handheld || npadIdType == NpadIdType.Unknown);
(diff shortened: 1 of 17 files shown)


Ryujinx.HLE/HOS/Services/Hid/IHidServer.cs
```diff
@@ -909,54 +908,64 @@ public ResultCode GetNpadJoyHoldType(ServiceCtx context)
...
         {
-            long singleJoyId0         = context.RequestData.ReadInt32();
-            long singleJoyId1         = context.RequestData.ReadInt32();
-            long appletResourceUserId = context.RequestData.ReadInt64();
+            NpadIdType npadIdType0          = (NpadIdType)context.RequestData.ReadUInt32();
+            NpadIdType npadIdType1          = (NpadIdType)context.RequestData.ReadUInt32();
+            long       appletResourceUserId = context.RequestData.ReadInt64();
 
-            Logger.Stub?.PrintStub(LogClass.ServiceHid, new { appletResourceUserId, singleJoyId0, singleJoyId1 });
+            if (npadIdType0 <= NpadIdType.Player8 || npadIdType0 == NpadIdType.Handheld || npadIdType0 == NpadIdType.Unknown ||
```
gdkchan, line 964: I think it would be better if you made some `IsValidNpadIdType` function that you can call to check if its valid. Then you can use it in all those functions and reduce the amount of code.

Thog (approved):
lgtm apart from gdkchan comment

Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
```diff
@@ -35,5 +35,10 @@ public static NpadIdType GetNpadIdTypeFromIndex(PlayerIndex index)
             PlayerIndex.Unknown  => NpadIdType.Unknown,
             _                    => throw new ArgumentOutOfRangeException(nameof(index))
         };
+
+        public static bool IsValidNpadIdType(NpadIdType npadIdType)
+        {
+            return (npadIdType <= NpadIdType.Player8 || npadIdType == NpadIdType.Handheld || npadIdType == NpadIdType.Unknown);
```
gdkchan, line 41: nit: No need for the `(` `)` here.

gdkchan (approved):
lgtm other than the comment.

PR #3328, opened 2022-05-08
Implement Viewport Transform Disable

riperiperi:
It seems that the Nouveau OpenGL and official Vulkan guest drivers perform certain blits with a draw, but this draw actually outputs vertexes with the scale 0 to width/height, rather than -1 to 1. For this, the viewport transform is disabled so that the pixel values directly line up. Host APIs don't support doing this, so we need to transform the vertex position in the shader to match a viewport that we provide.

All viewports are set to the minimum clip rectangle for all bound render targets, and the support buffer is updated with a value to rescale the vertex position back to -1 to 1. This is done on the last vertex stage (geometry will do it just before EmitVertex), and only if viewport transform has been disabled. The additional code will not be added to normal shaders.

This fixes the following examples, and possibly more:

#### Before
image

#### After
image

### Thumbnails in NSO N64 emulator

image
Save slots 1 and 4 were created without the fix.

#### Before
image

#### After
image

### Screen copies in Citra retroarch core (only certain games had this issue, also shader compile takes forever in opengl)

~~Retroarch requires a change to HidUtils to run on certain versions.~~ fixed by #3326

#### Before
image

#### After
image

Make sure that this doesn't clear out your shader cache or anything, this is the first time something has been added to the specialization state. It should still rebuild. Also, ideally nothing should change about NVN games.

This will also probably conflict a bit with Vulkan, due to making a few of the same changes it makes for comparing GraphicsState for specialization, and for -1 to 1 conversion also adding the LastInVertexStage variable.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -249,7 +249,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
                         context.AppendLine();
                     }
                 }
-                else if (isFragment)
+                else if (isFragment || context.Config.Stage == ShaderStage.Vertex)
                 {
                     DeclareSupportUniformBlock(context, context.Config.Stage, 0);
                 }
@@ -615,7 +615,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
 
         private static void DeclareSupportUniformBlock(CodeGenContext context, ShaderStage stage, int scaleElements)
         {
-            bool isFragment = stage == ShaderStage.Fragment;
+            bool isFragment = stage == ShaderStage.Fragment || stage == ShaderStage.Vertex;
             if (!isFragment && scaleElements == 0)
             {
                 return;
@@ -630,6 +630,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
                 case ShaderStage.Vertex:
                     context.AppendLine($"uint {DefaultNames.SupportBlockAlphaTestName};");
                     context.AppendLine($"bool {DefaultNames.SupportBlockIsBgraName}[{SupportBuffer.FragmentIsBgraCount}];");
+                    context.AppendLine($"vec4 {DefaultNames.SupportBlockViewportInverse};");
                     context.AppendLine($"int {DefaultNames.SupportBlockFragmentScaleCount};");
                     break;
                 case ShaderStage.Compute:
--- a/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -615,8 +615,10 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
 
         private static void DeclareSupportUniformBlock(CodeGenContext context, ShaderStage stage, int scaleElements)
         {
-            bool isFragment = stage == ShaderStage.Fragment || stage == ShaderStage.Vertex;
-            if (!isFragment && scaleElements == 0)
+            bool needsSupportBlock = stage == ShaderStage.Fragment || 
+                (context.Config.LastInVertexPipeline && context.Config.GpuAccessor.QueryViewportTransformDisable());
+
+            if (!needsSupportBlock && scaleElements == 0)
             {
                 return;
             }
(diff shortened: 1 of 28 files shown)


gdkchan:
> Retroarch requires a change to HidUtils to run on certain versions.

Was that not fixed by #3326?

riperiperi:
Yes, actually. My branch wasn't based on top of it when I was testing.

GamerzHell9137:
Fixes Dragon Quest Builders UI related stuff!

image
image
image

Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
```diff
@@ -206,7 +206,39 @@ private static void EmitOut(EmitterContext context, bool emit, bool cut)
...
+                Operand tempYLocal = null;
+
+                if (context.Config.LastInVertexPipeline && context.Config.GpuAccessor.QueryViewportTransformDisable())
+                {
+                    tempXLocal = Local();
+                    context.Copy(tempXLocal, Attribute(AttributeConsts.PositionX | AttributeConsts.LoadOutputMask));
+                    tempYLocal = Local();
+                    context.Copy(tempYLocal, Attribute(AttributeConsts.PositionY | AttributeConsts.LoadOutputMask));
+
+                    Operand x = Attribute(AttributeConsts.PositionX | AttributeConsts.LoadOutputMask);
```
gdkchan, line 219: I think it's better to have a common "do vertex transformations before emit or return" that could be called before return when last VTG stage is vertex/tessellation, or before emit when its geometry. That would avoid some code duplication here and probably help the depth mode conversion on Vulkan. Atlhough you still need to save and restore Position X/Y here regardless and its only done for geometry.

gdkchan (commented):
I think right now going from a version of the shader cache with a higher struct size to one with a lower size will cause it to be deleted due to it being detected as incompatible. See this code in `BinarySerializer`:
```cs
            // Structs are expected to expand but not shrink between versions.
            if (size > Unsafe.SizeOf<T>())
            {
                throw new DiskCacheLoadException(DiskCacheLoadResult.FileCorruptedInvalidLength);
            }
```
It was just an oversight from my part, but just removing the code should fix the issue. It's not a big deal though and I don't think it needs to be fixed before merging this.

Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
riperiperi: Yeah, though it might be weird returning the temporary locals to use them after the vertex emit.

```csharp
context.PrepareForVertexReturn(saveOld: true, out var oldXLocal,  out var oldYLocal,  out var oldZLocal)
```

Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
```diff
@@ -615,7 +615,7 @@ private static void DeclareOutputAttributePerPatch(CodeGenContext context, int a
 
         private static void DeclareSupportUniformBlock(CodeGenContext context, ShaderStage stage, int scaleElements)
         {
-            bool isFragment = stage == ShaderStage.Fragment;
+            bool isFragment = stage == ShaderStage.Fragment || stage == ShaderStage.Vertex;
```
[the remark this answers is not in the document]
riperiperi: It's also not accurate anymore - it's not required if the gpu state doesn't disable viewport transform. Forgot to change this when adapting to use specialization.

Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
gdkchan: You can have two `PrepareForVertexReturn` functions, one with the out X/Y/Z and another without it. The former would call the latter, and the `saveOld` argument would no longer be needed as that would be implied from the overload that is being used.

gdkchan:
Seems to also fix the title screen on River City Girls Zero:
image
The remaining OpenGL games with issues (some noted on #3207) should probably be re-tested with this change.

gdkchan (approved):
lgtm, thanks! Didn't know that the NVIDIA OpenGL driver also uses it, it's a nice surprise to see those games fixed too.

Thog (approved):

mooock:
Zaccaria Pinball is also fixed from this issue and loads faster without graphic bugs, also tested Zombies Ghoul Patrol that now works perfect. Tested 5 levels.

great job

PR #3338, opened 2022-05-13
Prefetch capabilities before spawning translation threads.

riperiperi:
The Backend Multithreading only expects one thread to submit commands at a time. When compiling shaders, the translator may request the host GPU capabilities from the backend. It's possible for a bunch of translators to do this at the same time.

There's a caching mechanism in place so that the capabilities are only fetched once. By triggering this before spawning the thread, the async translation threads no longer try to queue onto the backend queue all at the same time.

The Capabilities do need to be checked from the GPU thread, due to OpenGL needing a context to check them, so it's not possible to call the underlying backend directly.

### DIFF
--- a/Ryujinx.Graphics.Gpu/GpuContext.cs
+++ b/Ryujinx.Graphics.Gpu/GpuContext.cs
@@ -82,19 +82,7 @@ namespace Ryujinx.Graphics.Gpu
         /// <summary>
         /// Host hardware capabilities.
         /// </summary>
-        internal ref Capabilities Capabilities
-        {
-            get
-            {
-                if (!_capsLoaded)
-                {
-                    _caps = Renderer.GetCapabilities();
-                    _capsLoaded = true;
-                }
-
-                return ref _caps;
-            }
-        }
+        internal ref Capabilities Capabilities => ref _caps;
 
         /// <summary>
         /// Event for signalling shader cache loading progress.
@@ -254,6 +242,8 @@ namespace Ryujinx.Graphics.Gpu
         public void SetGpuThread()
         {
             _gpuThread = Thread.CurrentThread;
+
+            _caps = Renderer.GetCapabilities();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


gdkchan:
Would be better to make it not lazy initialized and instead get it on GPU init somehow (it could be done on `SetGpuThread` or any method that is called from the GPU thread).

Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -82,19 +82,7 @@ public sealed class GpuContext : IDisposable
...
-                if (!_capsLoaded)
-                {
-                    _caps = Renderer.GetCapabilities();
-                    _capsLoaded = true;
-                }
-
-                return ref _caps;
-            }
-        }
+        internal ref Capabilities Capabilities => ref _caps;
```
gdkchan, line 85: This can be changed to just `internal Capabilities Capabilities;` now, then `_caps` and `_capsLoaded`. Also, for some reason it seems that this file has mixed tabs and spaces?

meisiek:
I've deleted the shader cache directory from all my games and restored the version from yesterdays backup.
The crashes that I've reported by starting games that want to compile all shaders seems to be fixed.
Thank you!

Is there a faster way to force Ryujinx to think that all shaders have to be rebuild? For example by modifying the current driver version number?

meisiek:
There is another exception that happens often before shader compilation. I had the exact same error today at least 10 times after forcing all shaders in all games to be recompiled.

meisiek:
[Ryujinx_1.1.0+6063f41_2022-05-13_17-25-52.log](https://github.com/Ryujinx/Ryujinx/files/8689079/Ryujinx_1.1.0%2B6063f41_2022-05-13_17-25-52.log)

gdkchan (approved):
lgtm, thanks for fixing this.
I think using a property there will cause a copy of the capabilities struct though.

Thog (approved):

PR #3347, opened 2022-05-16
Remove freed memory range from tree on memory block disposal

### LINKED ISSUE #3360: Stopping emulation and trying to run a game again sometimes results in an error which makes the emulator window stop responding

The emulator crashes when running any game after stopping emulation once in a random given session, with the following error:
`00:00:00.685 |E| Application : Unhandled exception caught: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> Ryujinx.Memory.WindowsShared.WindowsApiException: VirtualFree returned error code 0x1E7.`
Does not seem to happen on Software memory mode.

Seems to happen randomly on any given instance of Ryujinx, but if it happens, it's always when trying to boot a game after stopping emulation once. I redownloaded the emulator, tried to reproduce it, and couldn't, but upon restarting Ryujinx, it started happening again. Now I can't get it to happen consistently.

### Log file

[Ryujinx_1.1.132_2022-05-25_21-39-26.log.txt](https://github.com/Ryujinx/Ryujinx/files/8774703/Ryujinx_1.1.132_2022-05-25_21-39-26.log.txt)

gdkchan:
The freed memory range was not being removed from the tree, which could cause it to try coalescing the (now inexistent) placeholder with the placeholder of the new region that was allocated. This would usually happen when a new mapping was created on an adjacent memory region, for example when the emulation was stopped for one game and another was launched, there was the change that it would throw the following exception:
```
Ryujinx.Memory.WindowsShared.WindowsApiException: VirtualFree returned error code 0x1E7.
```
This PR should fix the issue making sure the freed range is fully removed from the tree when the memory block is disposed. Additionally, this is limiting the range of memory that it tries to coalesce to just inside of a memory block, so that it won't try to coalesce different memory blocks (which would cause issues on the new `UnreserveRange` function, as it does not try to split range that exceeds the unreserved range. Now that it does not coalesce across blocks, this scenario is impossible).

Testing is welcome.

Note: The issue only affects Windows.
Note 2: This is technically also a regression from #2954.

Fixes #3360.

### DIFF
--- a/Ryujinx.Memory/WindowsShared/PlaceholderManager.cs
+++ b/Ryujinx.Memory/WindowsShared/PlaceholderManager.cs
@@ -50,18 +50,16 @@ namespace Ryujinx.Memory.WindowsShared
         /// <param name="address">Start address of the region to unreserve</param>
         /// <param name="size">Size in bytes of the region to unreserve</param>
         /// <exception cref="WindowsApiException">Thrown when the Windows API returns an error unreserving the memory</exception>
-        public void UnreserveRange(IntPtr location, IntPtr size)
+        public void UnreserveRange(ulong address, ulong size)
         {
-            ulong startAddress = (ulong)location;
-            ulong unmapSize = (ulong)size;
-            ulong endAddress = startAddress + unmapSize;
+            ulong endAddress = address + size;
 
             var overlaps = Array.Empty<IntervalTreeNode<ulong, ulong>>();
             int count;
 
             lock (_mappings)
             {
-                count = _mappings.Get(startAddress, endAddress, ref overlaps);
+                count = _mappings.Get(address, endAddress, ref overlaps);
 
                 for (int index = 0; index < count; index++)
                 {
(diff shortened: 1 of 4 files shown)


Ryujinx.Memory/WindowsShared/PlaceholderManager.cs
```diff
@@ -44,20 +44,67 @@ public void ReserveRange(ulong address, ulong size)
             }
         }
 
+        /// <summary>
+        /// Unreserves a range of memory that has been previously reserved with <see cref="ReserveRange"/>.
+        /// </summary>
+        /// <param name="address">Start address of the region to unreserve</param>
```
riperiperi, line 50: This parameter isn't called address.

gdkchan: Changed the name and type of the arguments for consistency with the `ReserveRange` function.

Thog (approved):

PR #3351, opened 2022-05-19
UI - Avalonia Part 2

### LINKED ISSUE #3395: [Feature Request] Auto load default controller profile when a controller is selected in input devices
#### Overview:
Have the emulator automatically load the default profile instead of having every button unbound by default.

Many people (myself included) don't realize at first that the "Load" button can just skip 99% of the process of binding everything individually. This would make end-users' lives easier and the emulator would require even less setup than it already does.

emmauss:
This part adds the settings window and its child windows and controls.

Closes #3395.


Ryujinx.Common/Configuration/Hid/KeyboardHotkeys.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Ryujinx.Common.Configuration.Hid
 {
-    public struct KeyboardHotkeys
+    public class KeyboardHotkeys
```
Thog, line 3: Any reasons of this change?

gdkchan: It allows the object to be used on a XAML data binding directly. Using a struct would not allow the UI modifications to be propagated back, I assume because it is passed by value and so it modifies a copy of the struct.

emmauss: The xaml bindings don't work with structs, as updates won't affect the original object.

Thog: I see 👍

Ryujinx.Ava/Ui/Windows/ControllerSettingsWindow.axaml.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                    _currentAssigner.ButtonAssigned += (sender, e) =>
+                    {
+                        if (e.IsAssigned)
+                        {
+                            ViewModel.IsModified = true;
+                        }
+                    };
+
+                    _currentAssigner.GetInputAndAssign(assigner, keyboard);
```
AcK77, line 97: Resolve the warning

Ryujinx.Ava/Ui/Windows/ControllerSettingsWindow.axaml.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                shouldUnbind = true;
+            }
+
+            _currentAssigner?.Cancel(shouldUnbind);
+
+            PointerPressed -= MouseClick;
+        }
+
+        private async void PlayerIndexBox_OnSelectionChanged(object? sender, SelectionChangedEventArgs e)
```
AcK77, line 161: Resolve the warning

Ryujinx.Ava/Ui/Windows/SettingsWindow.axaml.cs
```diff
@@ -0,0 +1,298 @@
...
+                    _currentAssigner = new ButtonKeyAssigner(button);
+
+                    FocusManager.Instance.Focus(this, NavigationMethod.Pointer);
+
+                    PointerPressed += MouseClick;
+
+                    IKeyboard keyboard = (IKeyboard)ViewModel.AvaloniaKeyboardDriver.GetGamepad(ViewModel.AvaloniaKeyboardDriver.GamepadsIds[0]);
+                    IButtonAssigner assigner = new KeyboardKeyAssigner(keyboard);
+
+                    _currentAssigner.GetInputAndAssign(assigner);
```
AcK77, line 134: Resolve the warning

AcK77 (changes requested):

Squall-Leonhart:
avalonia builds aren't adding their executable name to the Ryujinx Nvidia Profile, leading to threaded optimization not being disabled, the builds are ryujinx.ava.exe, ryujinx.exe is added

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+            PlayerIndexes.Add(new(PlayerIndex.Player7, LocaleManager.Instance["ControllerSettingsPlayer7"]));
+            PlayerIndexes.Add(new(PlayerIndex.Player8, LocaleManager.Instance["ControllerSettingsPlayer8"]));
+            PlayerIndexes.Add(new(PlayerIndex.Handheld, LocaleManager.Instance["ControllerSettingsHandheld"]));
+        }
+
+        private void LoadConfiguration(InputConfig inputConfig = null)
+        {
+            Config = inputConfig ?? ConfigurationState.Instance.Hid.InputConfig.Value.Find(inputConfig => inputConfig.PlayerIndex == _playerId);
+
+            if (Config is StandardKeyboardInputConfig)
```
gdkchan, line 285:
```suggestion
            if (Config is StandardKeyboardInputConfig keyboardInputConfig)
```

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+            {
+                Configuration = new InputConfiguration<Key, ConfigStickInputId>(Config as StandardKeyboardInputConfig);
```
gdkchan, line 287:
```suggestion
                Configuration = new InputConfiguration<Key, ConfigStickInputId>(keyboardInputConfig);
```

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+            }
+
+            if (Config is StandardControllerInputConfig)
```
gdkchan, line 290: Same as above.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+            {
+                Device = 0;
+            }
+            else
+            {
+                string ident = "";
+
+                if (Config is StandardKeyboardInputConfig)
+                {
+                    ident = "keyboard";
```
gdkchan, line 308: Would be nice to have constants for those too.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+        {
+            if (_device < 0)
+            {
+                return;
+            }
+
+            string id = GetCurrentGamepadId();
+            string selected = Devices[Device].Id;
+
+            if (selected == "disabled")
```
gdkchan, line 348: Same comment about the constants.

gdkchan: On a side note, I'm not sure why strings are being used for this. It seems it would be better to have a struct with device type enum and a integer ID.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+                    Controller = Controllers.ToList().FindIndex(x => x.Type == Config.ControllerType);
+                }
+                else
+                {
+                    Controller = 0;
+                }
+            }
+        }
+
+        private static string GetShrinkedGamepadName(string str)
```
gdkchan, line 432: The word `Shrinked` does not exist. The correct is `Shrunken`.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+        {
+            const string ShrinkChars = "...";
```
gdkchan, line 434: The correct name for this is "Ellipsis".

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
gdkchan: Probably a better name  is `GetShortGamepadName`.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+        {
+            lock (Devices)
+            {
+                Devices.Clear();
+                DeviceList.Clear();
+                Devices.Add(("disabled", LocaleManager.Instance["ControllerSettingsDeviceDisabled"]));
+
+                foreach (string id in _mainWindow.InputManager.KeyboardDriver.GamepadsIds)
+                {
+                    IGamepad gamepad = _mainWindow.InputManager.KeyboardDriver.GetGamepad(id);
```
gdkchan, line 455:
```suggestion
                    using IGamepad gamepad = _mainWindow.InputManager.KeyboardDriver.GetGamepad(id);
```
Then you can remove the dispose call.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+                string id = activeDevice.Split("/")[1];
+
+                config = new StandardKeyboardInputConfig
+                {
+                    Version = Ryujinx.Common.Configuration.Hid.InputConfig.CurrentVersion,
+                    Backend = InputBackendType.WindowKeyboard,
+                    Id = id,
+                    ControllerType = ControllerType.ProController,
+                    LeftJoycon =
+                        new LeftJoyconCommonConfig<Key>
```
gdkchan, line 544: Move this to the line above for consistency with the other. Also needs to fix the indentation.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+                            DpadLeft = Key.Left,
+                            DpadRight = Key.Right,
+                            ButtonMinus = Key.Minus,
+                            ButtonL = Key.E,
+                            ButtonZl = Key.Q,
+                            ButtonSl = Key.Unbound,
+                            ButtonSr = Key.Unbound
+                        },
+                    LeftJoyconStick =
+                        new JoyconConfigKeyboardStick<Key>
```
gdkchan, line 557: Same as above.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+            {
+                ContentDialogHelper.CreateErrorDialog(_owner.GetVisualRoot() as StyleableWindow, LocaleManager.Instance["DialogProfileDefaultProfileOverwriteErrorMessage"]);
+
+                return;
+            }
+            else
+            {
+                bool validFileName = true;
+
+                foreach (char invalidChar in Path.GetInvalidFileNameChars())
```
gdkchan, line 756: You can avoid this loop by doing `bool validFileName = ProfileName.IndexOfAny(Path.GetInvalidFileNameChars()) == -1`.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+                    inputConfig.Id = selected.Split("/")[1].Split(" ")[0];
+                }
+
+                var config = !IsController
+                    ? (Configuration as InputConfiguration<Key, ConfigStickInputId>).GetConfig()
+                    : (Configuration as InputConfiguration<GamepadInputId, ConfigStickInputId>).GetConfig();
+                config.ControllerType = Controllers[_controller].Type;
+                config.PlayerIndex = _playerId;
+
+                int i = newConfig.FindIndex(x => x.PlayerIndex == this.PlayerId);
```
gdkchan, line 857:
```suggestion
                int i = newConfig.FindIndex(x => x.PlayerIndex == PlayerId);
```

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,907 @@
...
+                {
+                    newConfig.Add(config);
+                }
+                else
+                {
+                    newConfig[i] = config;
+                }
+            }
+
+            if (_mainWindow.AppHost != null)
```
gdkchan, line 868: `_mainWindow.AppHost?.NpadManager.ReloadConfiguration(newConfig, ConfigurationState.Instance.Hid.EnableKeyboard, ConfigurationState.Instance.Hid.EnableMouse);`

Ryujinx.Ava/Ui/ViewModels/MainWindowViewModel.cs
```diff
@@ -836,6 +835,22 @@ public async void OpenFolder()
             }
         }
 
+        public void LoadConfigurableHotKeys()
+        {
+            if(AvaloniaMappingHelper.TryGetAvaKey((Ryujinx.Input.Key)ConfigurationState.Instance.Hid.Hotkeys.Value.ShowUi, out var showUiKey))
```
gdkchan, line 840:
```suggestion
            if (AvaloniaMappingHelper.TryGetAvaKey((Ryujinx.Input.Key)ConfigurationState.Instance.Hid.Hotkeys.Value.ShowUi, out var showUiKey))
```

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,403 @@
...
+        public bool EnableError { get; set; }
+        public bool EnableTrace { get; set; }
+        public bool EnableGuest { get; set; }
+        public bool EnableFsAccessLog { get; set; }
+        public bool EnableDebug { get; set; }
+        public bool IsOpenAlEnabled { get; set; }
+        public bool IsSoundIoEnabled { get; set; }
+        public bool IsSDL2Enabled { get; set; }
+        public bool EnableCustomTheme { get; set; }
+        public bool IsResolutionScaleActive => _resolutionScale == 0;
```
gdkchan, line 118: Shouldn't this be `!= 0`?

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,403 @@
...
+
+                OnPropertyChanged(nameof(TimeZone));
+            }
+        }
+
+        public async void BrowseTheme()
+        {
+            var dialog = new OpenFileDialog()
+            {
+                Title = "Select Theme File",
```
gdkchan, line 234: Should be localizable.

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,403 @@
...
+                AllowMultiple = false
+            };
+
+            dialog.Filters.Add(new FileDialogFilter() { Extensions = { "xaml" }, Name = "Xaml Theme File" });
```
gdkchan, line 238: Should be localizable.

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -0,0 +1,403 @@
...
+            Volume = config.System.AudioVolume * 100;
+
+            GraphicsBackendMultithreadingIndex = (int)config.Graphics.BackendThreading.Value;
+
+            OpenglDebugLevel = (int)config.Logger.GraphicsDebugLevel.Value;
+
+            TimeZone = config.System.TimeZone;
+            ShaderDumpPath = config.Graphics.ShadersDumpPath;
+            CustomThemePath = config.Ui.CustomThemePath;
+            BaseStyleIndex = config.Ui.BaseStyle == "Light" ? 0 : 1;
```
gdkchan, line 289: Maybe should be localizable, but then this check would need to change too.

Ryujinx.Ava/Ui/Windows/ControllerSettingsWindow.axaml.cs
```diff
@@ -0,0 +1,199 @@
...
+                assigner = new GamepadButtonAssigner(ViewModel.SelectedGamepad, (config as StandardControllerInputConfig).TriggerThreshold, forStick);
+            }
+            else
+            {
+                throw new Exception("Controller not supported");
+            }
+
+            return assigner;
+        }
+        private void Button_Unchecked(object sender, RoutedEventArgs e)
```
gdkchan, line 141: Missing empty line above.

Ryujinx.Ava/Ui/Windows/SettingsWindow.axaml.cs
```diff
@@ -0,0 +1,298 @@
...
+            }
+
+            _currentAssigner?.Cancel(shouldUnbind);
+
+            PointerPressed -= MouseClick;
+        }
+
+        private void NavPanelOnSelectionChanged(object sender, NavigationViewSelectionChangedEventArgs e)
+        {
+            if (e.SelectedItem is NavigationViewItem navitem)
```
gdkchan, line 172: Is there no way to make the control do this "automatically"?

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
gdkchan: Nevermind, this is correct, but the name is misleading. It should really be `IsCustomResolutionScaleActive`.

Ryujinx.Ava/Ui/Controls/ButtonKeyAssigner.cs
```diff
@@ -0,0 +1,125 @@
...
+            });
+
+            await Dispatcher.UIThread.InvokeAsync(() =>
+            {
+                string pressedButton = assigner.GetPressedButton();
+                if (_shouldUnbind)
+                {
+                    try
+                    {
+                        SetButtonText(ToggledButton, "Unbound");
```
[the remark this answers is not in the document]
emmauss: `Unbound` is an enum value, like all other key text. Localizing it will break binding to the Key enum

Ryujinx.Ava/Ui/Controls/ButtonKeyAssigner.cs
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                    catch { }
```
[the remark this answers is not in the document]
emmauss: No exception currently. It is removed.

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
emmauss: `Light` and `Dark` are used internally for config values, not ui text.

Ryujinx.Ava/Ui/Windows/SettingsWindow.axaml.cs
emmauss: No. The base Navigation Panel control doesn't know how to link pages, without the user handling selection changed.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
emmauss: Strings are simpler. The device list can have as many devices as connected. You would still parse the enums to strings to make comparisons, as we need the name to figure out which type it is.

Ryujinx.Ava/Ui/Controls/ButtonKeyAssigner.cs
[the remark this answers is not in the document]
gdkchan: IMO the other key names should be localizable too. It's an issue that needs to be solved in general.

Ryujinx.Ava/Ui/Models/ControllerModel.cs
```diff
@@ -0,0 +1,6 @@
+﻿using Ryujinx.Common.Configuration.Hid;
+
+namespace Ryujinx.Ava.Ui.Models
+{
+    public record ControllerModel(ControllerType Type, string Name);
```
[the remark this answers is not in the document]
gdkchan: My understanding is that nothing on Ryujinx.Ava needs to be public since no other project will consume it. Is there a reason for making it public?

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
gdkchan: On a lot of places, the code is not checking the device ID, but just the "type", that is, if the input is disabled, keyboard or controller. So it seems beneficial to me to use a struct with one enum for the "type" and a string for the ID. For the cases where you don't care about the actual ID, you have a more efficient comparison (you don't need to compare a string, just an ID), the code becomes less fragile because if someone changes one of the strings for some reason, the code no longer breaks at runtime, instead it would be a compile time error. And one could use the rename feature to change all uses properly.

emmauss: Added DeviceType enum

Ryujinx.Ava/Ui/Models/ControllerModel.cs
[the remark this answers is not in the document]
emmauss: Updated all public classes in Ryujinx.Ava to `internal`

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -0,0 +1,887 @@
...
+        private object _configuration;
+        private string _profileName;
+        private bool _isLoaded;
+        private readonly UserControl _owner;
+
+        public IGamepadDriver AvaloniaKeyboardDriver { get; }
+        public IGamepad SelectedGamepad { get; private set; }
+
+        public ObservableCollection<PlayerModel> PlayerIndexes { get; set; }
+        public ObservableCollection<(DeviceType Type,string Id, string Name)> Devices { get; set; }
```
gdkchan, line 59:
```suggestion
        public ObservableCollection<(DeviceType Type, string Id, string Name)> Devices { get; set; }
```

Ackliez:
The input settings window seems to have an alignment issue with or without optimized scaling

Normalized window
Short
64/175468197-0aebdc1a-2442-493c-89c2-cebdf32d2c0a.png)

gdkchan (approved):
lgtm, other than the fact that the settings window is way too large by default. Not sure why the width was increased again.

Thog (approved):
I agree with @gdkchan on the settings window size being a bit too big but apart from that, some part of the code could use some later refactoring

PR #3361, opened 2022-05-25
time: Make TimeZoneRule blittable and avoid copies

Thog:
This drastically reduce overhead of using TimeZoneRule around the codebase by avoiding to copy TimeZoneRule 16384 bytes all around.

Effect on games is unknown but could improve perf a bit depending on JIT output.

### DIFF
--- a/Ryujinx.Common/Memory/Box.cs
+++ b/Ryujinx.Common/Memory/Box.cs
@@ -1,0 +1,14 @@
+﻿namespace Ryujinx.Common.Memory
+{
+    public class Box<T> where T : unmanaged
+    {
+        private T _data;
+
+        public ref T Data => ref _data;
+
+        public Box()
+        {
+            _data = new T();
+        }
+    }
+}
--- a/Ryujinx.Common/Memory/Box.cs
+++ b/Ryujinx.Common/Memory/Box.cs
@@ -2,13 +2,11 @@
 {
     public class Box<T> where T : unmanaged
     {
-        private T _data;
-
-        public ref T Data => ref _data;
+        public T Data;
 
         public Box()
         {
-            _data = new T();
+            Data = new T();
         }
     }
 }
--- a/Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
+++ b/Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
@@ -202,9 +202,9 @@ namespace Ryujinx.HLE.HOS.Services.Time.StaticService
                 throw new InvalidOperationException();
             }
 
-            ref TimeZoneRule rules = ref context.Memory.GetRef<TimeZoneRule>(bufferPosition);
+            ReadOnlySpan<TimeZoneRule> rules = MemoryMarshal.Cast<byte, TimeZoneRule>(context.Memory.GetSpan(bufferPosition, (int)bufferSize));
 
-            ResultCode resultCode = _timeZoneManager.ToCalendarTime(in rules, posixTime, out CalendarInfo calendar);
+            ResultCode resultCode = _timeZoneManager.ToCalendarTime(in rules[0], posixTime, out CalendarInfo calendar);
 
             if (resultCode == 0)
             {
@@ -247,9 +247,9 @@ namespace Ryujinx.HLE.HOS.Services.Time.StaticService
                 throw new InvalidOperationException();
             }
 
-            ref TimeZoneRule rules = ref context.Memory.GetRef<TimeZoneRule>(inBufferPosition);
+            ReadOnlySpan<TimeZoneRule> rules = MemoryMarshal.Cast<byte, TimeZoneRule>(context.Memory.GetSpan(inBufferPosition, (int)inBufferSize));
 
-            ResultCode resultCode = _timeZoneManager.ToPosixTime(in rules, calendarTime, out long posixTime);
+            ResultCode resultCode = _timeZoneManager.ToPosixTime(in rules[0], calendarTime, out long posixTime);
 
             if (resultCode == ResultCode.Success)
             {
--- a/Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
+++ b/Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
@@ -4,9 +4,12 @@ using Ryujinx.Cpu;
 using Ryujinx.HLE.HOS.Services.Time.Clock;
 using Ryujinx.HLE.HOS.Services.Time.TimeZone;
 using Ryujinx.HLE.Utilities;
+using Ryujinx.Memory;
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 namespace Ryujinx.HLE.HOS.Services.Time.StaticService
 {
@@ -165,11 +168,11 @@ namespace Ryujinx.HLE.HOS.Services.Time.StaticService
 
             using (MemoryStream timeZoneBinaryStream = new MemoryStream(temp))
             {
-                result = _timeZoneManager.ParseTimeZoneRuleBinary(out TimeZoneRule timeZoneRule, timeZoneBinaryStream);
-
-                if (result == ResultCode.Success)
+                using (WritableRegion region = context.Memory.GetWritableRegion(timeZoneRuleBufferPosition, Unsafe.SizeOf<TimeZoneRule>()))
                 {
-                    MemoryHelper.Write(context.Memory, timeZoneRuleBufferPosition, timeZoneRule);
+                    ref TimeZoneRule rule = ref MemoryMarshal.Cast<byte, TimeZoneRule>(region.Memory.Span)[0];
+
+                    result = _timeZoneManager.ParseTimeZoneRuleBinary(ref rule, timeZoneBinaryStream);
                 }
             }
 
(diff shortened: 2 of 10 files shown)


Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
```diff
@@ -200,9 +203,9 @@ public ResultCode ToCalendarTime(ServiceCtx context)
                 throw new InvalidOperationException();
             }
 
-            TimeZoneRule rules = MemoryHelper.Read<TimeZoneRule>(context.Memory, bufferPosition);
+            ref TimeZoneRule rules = ref context.Memory.GetRef<TimeZoneRule>(bufferPosition);
```
gdkchan, line 206: This should use `GetSpan` as `GetRef` will throw if the memory region is not contiguous.

Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZoneContentManager.cs
```diff
@@ -146,7 +146,11 @@ public IEnumerable<(int Offset, string Location, string Abbr)> ParseTzOffsets()
                         continue;
                     }
 
-                    TimeZone.ParseTimeZoneBinary(out TimeZoneRule tzRule, tzif.Get.AsStream());
+                    Common.Memory.Box<TimeZoneRule> tzRuleBox = new Common.Memory.Box<TimeZoneRule>();
```
gdkchan, line 149: I'd add `using Ryujinx.Common.Memory` to avoid fully specifying the namespace here.

Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZoneManager.cs
```diff
@@ -1,14 +1,14 @@
...
 using System.IO;
-using static Ryujinx.HLE.HOS.Services.Time.TimeZone.TimeZoneRule;
 
 namespace Ryujinx.HLE.HOS.Services.Time.TimeZone
 {
     class TimeZoneManager
     {
         private bool                 _isInitialized;
-        private TimeZoneRule         _myRules;
+        private Box<TimeZoneRule>    _myRules;
```
gdkchan, line 11: Why is that necessary? `_myRules` is already on the heap even without the `Box` since `TimeZoneManager` is a reference type.

Ryujinx.HLE/HOS/Services/Time/StaticService/ITimeZoneServiceForPsc.cs
gdkchan: Just to clarify, `GetSpan` does not throw because it returns a `ReadOnlySpan`. It assumes you don't need to modify the data, so it can just return a copy if the memory region is not contiguous. `GetRef` on the other range, does assume that you might modify the data, and so if the memory region is not contiguous, it can't return the ref as it wouldn't point to the whole data.

Ryujinx.Common/Memory/Box.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Ryujinx.Common.Memory
+{
+    public class Box<T> where T : unmanaged
+    {
+        private T _data;
+
+        public ref T Data => ref _data;
```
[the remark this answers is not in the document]
Thog: I just want to avoid the field being accessed in a non ref way and initially I was using an array.

I could try changing that if you want

Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZoneContentManager.cs
Thog: It was conflicting with LibHac Box, will force a using Box =

Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZoneManager.cs
Thog: Easier to replace later on when loading a timezone to replace it without making a full copy

Ryujinx.Common/Memory/Box.cs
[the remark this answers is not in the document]
gdkchan: You can still access 'Data' without 'ref' and it will do a copy of the value in this case.
Example:
```cs
Box<MyStruct> box = new Box<MyStruct>();
DoSomethingWithMyStruct(box.Data); // This does a copy.
DoSomethingWithMyStructByRef(ref box.Data); // This doesn't.
```
Both returning a ref of the field and making the field public and accessing it directly would give the same result here.

Ryujinx.HLE/HOS/Services/Time/TimeZone/TimeZoneManager.cs
gdkchan: You can just replace all instances of `ref _myRules.Data` with `ref _myRules`, and on `SetDeviceLocationNameWithTimeZoneRule` you can do `bool timeZoneConversionSuccess = TimeZone.ParseTimeZoneBinary(ref _myRules, timeZoneBinaryStream);` and remove `Box<TimeZoneRule> rules = new Box<TimeZoneRule>();` and `_myRules            = rules;`. It will achieve the same result and get rid of the `Box<T>`.

Thog: I cannot use _myRules for ParseTimeZoneBinary as this doesn't account for possible parsing issues, that also doesn't account for temporary TimeZoneRule used for posix name parsing

gdkchan: Right, I see what you mean now, you don't want `_myRules` to be left in a invalid state in the case the function fails.

Ryujinx.Common/Memory/Box.cs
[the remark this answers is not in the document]
gdkchan: I think it's fine to leave it like this so marking as resolved.

Ryujinx.HLE/HOS/Services/Time/TimeZone/Types/TimeZoneRule.cs
```diff
@@ -22,17 +24,32 @@ struct TimeZoneRule
...
+        private AtsStorageStruct _ats;
 
-        [MarshalAs(UnmanagedType.ByValArray, SizeConst = TzMaxTypes)]
-        public TimeTypeInfo[] Ttis;
+        public Span<long> Ats => SpanHelpers.AsSpan<AtsStorageStruct, long>(ref _ats);
 
-        [MarshalAs(UnmanagedType.ByValArray, SizeConst = TzCharsArraySize)]
-        public char[] Chars;
+        [StructLayout(LayoutKind.Sequential, Size = sizeof(byte) * TzMaxTimes)]
+        private struct TypesStorageStruct {}
```
gdkchan, line 35:
```suggestion
        private struct TypesStorageStruct { }
```

Ryujinx.HLE/HOS/Services/Time/TimeZone/Types/TimeZoneRule.cs
```diff
@@ -22,17 +24,32 @@ struct TimeZoneRule
...
+
+        [StructLayout(LayoutKind.Sequential, Size = TimeTypeInfo.Size * TzMaxTimes)]
+        private struct TimeTypeInfoStorageStruct { }
+
+        private TimeTypeInfoStorageStruct _ttis;
+
+        public Span<TimeTypeInfo> Ttis => SpanHelpers.AsSpan<TimeTypeInfoStorageStruct, TimeTypeInfo>(ref _ttis);
+
+        [StructLayout(LayoutKind.Sequential, Size = sizeof(byte) * TzCharsArraySize)]
+        private struct CharsStorageStruct {}
```
gdkchan, line 49:
```suggestion
        private struct CharsStorageStruct { }
```

gdkchan (commented):
nit, just for consistency.

Ryujinx.HLE/HOS/Services/Time/TimeZone/Types/TimeZoneRule.cs
```diff
@@ -22,17 +24,32 @@ struct TimeZoneRule
         [MarshalAs(UnmanagedType.I1)]
         public bool GoAhead;
 
-        [MarshalAs(UnmanagedType.ByValArray, SizeConst = TzMaxTimes)]
-        public long[] Ats;
+        [StructLayout(LayoutKind.Sequential, Size = sizeof(long) * TzMaxTimes)]
+        private struct AtsStorageStruct {}
```
gdkchan, line 28:
```suggestion
        private struct AtsStorageStruct { }
```
Sorry, I missed this one on last review.

gdkchan (approved):
Other than the comment, looks good. Would be nice if we had some test coverage for this code.

riperiperi (approved):
LGTM

gdkchan:
Needs a rebase now.

Ryujinx.HLE/HOS/Services/Time/TimeZone/Types/TimeZoneRule.cs
gdkchan: This and the other comments were not addressed (or maybe they were and somehow got reverted with the rebase?)

Thog: huh odd? I was pretty sure I did address that one

gdkchan: If you applied the suggestion from GitHub, then I guess you forgot to pull the changes on your local branch and they got lost on rebase.

PR #3362, opened 2022-05-25
Refactor CPU interface to allow the implementation of other CPU emulators

### LINKED ISSUE #2025: Please move all interaction between ARMeilleure and the rest of the emulator to an interface level in Ryujinx.Cpu
## Feature Request

I believe that ryujinx should be completely independent from the cpu, as in nothing should depend on ARMeilleure. This would make it much easier for contributors to test their own cpu emulators, and would allow changes in the cpu to not affect the rest of the project.

My ideas to implement this:

The only thing that should depend on ARMeilleure should be Ryujinx.Cpu. Ryujinx.Cpu include "Universal Interfaces"

E.g.

The interface would look like
```cs
public delegate void Hook(object Data);

interface CpuContext
{
    public int ID   { get; set; }
    public ulong GetX(int index);
    public void SetX(int index, ulong value);
    public Vector128<float> GetQ(int index);
    public void SetQ(int index,Vector128<float> Value);

    public bool N               { get; set; }
    public bool Z               { get; set; }
    public bool C               { get; set; }
    public bool V               { get; set; }
    public ulong tpidrro_el0    { get; set; }
    public ulong tpidr          { get; set; }
    public ulong dczid_el0      { get; set; }
    public ulong fpsr           { get; set; }
    public ulong fpcr           { get; set; }
    public ulong BaseAddress    { get; set; }
    public Hook SvcHook         { get; set; }

    public void StartExecution();
}
```

Ryujinx.Cpu Should also include actual implementation for cpu emulators.

```cs
public class ARMeilleureCpuContext : CpuContext
{
    public ExecutionContext context;

    //implementation 
}
```

gdkchan:
This refactors the existing CPU related interfaces (and also adds new ones) to allow other CPU emulators to be implemented. This includes not only JIT based emulators, but also hypervisors (for example Apple Hypervisor).

New interfaces:
- ICpuEngine - Can be used to create CPU contexts.
- ICpuContext - Basically "maps" to a guest process. Has the memory manager for the process address space, and can be used to create execution contexts.
- IExecutionContext - Basically "maps" to a guest thread. Has all the CPU state of the thread (the general purpose and FP/SIMD registers, in addition to a few system registers that Switch games uses, which are TPIDR, CNTPCT, FPCR, FPSR and PState). It also exposes functions to allow interrupting and stopping the thread.
- ITickSource - Interface for the internal CPU timer counter. It is also no longer implemented on the ARMeilleure CPU emulator, instead it receives a interface.

Other changes:
- Everything on HLE using `ExecutionContext` directly was changed to use `IExecutionContext` interface instead.
- The way how CPU interrupt are handled was changed a bit, instead of events it now passes callback functions directly. That's because we don't ever need multiple listeners for those events, so just passing the function directly seemed more efficient and appropriate.
- Several functions (most Time service related) that had to access the CPU counter using a few different methods were all changed to instead use the new `ITickSource` interface. Also, several Time related functions that assumed the counter might be inaccessible due to no guest thread having started yet were changed as we now initialize the counter very early, before Time service initialization, so this is no longer needed.
- `Tpidr` was renamed to `TpidrroEl0`, as the previous name was slightly incorrect.
- The missing `Pc` (Program Counter) property was added to the `ExecutionContext` and removed from `KThread`. Note that it does not return a precise PC value, because tracking it precisely with a JIT would be expensive (as we'd need to save the address before each guest instruction or store some information in memory to look it up from host Instruction Pointer).

Closes #2025. We are still using `V128` struct from ARMeilleure, but I believe we should replace it with `Vector128<byte>` or maybe the `UInt128` type that will be included in .NET 7, but this change also requires change to the ARMeilleure helpers and maybe the code generation, so I think it should be done separately. This is also using the `ARMeilleure.IMemoryManager` interface, maybe we should expose a new interface for that too.

Testing is welcome. No user visible changes are expected.


riperiperi (approved):
Looks good, tried on a few games and nothing appears to have regressed.

Thog (approved):
lgtm

PR #3371, opened 2022-05-29
Rewrite SVC handler using source generators rather than IL emit

gdkchan:
Right now we use Reflection.Emit to generate code at runtime that will read SVC function arguments from the CPU emulator `ExecutionContext` registers. It will then be passed to the appropriate function. Right now, it requires a function for 32-bit and 64-bit modes to specify the register where each argument is located using attributes. It looks like so:
```cs
        public KernelResult ConnectToNamedPort64([R(1)] ulong namePtr, [R(1)] out int handle)
        {
            return _syscall.ConnectToNamedPort(out handle, namePtr);
        }
```
Ultimately, it doesn't save much because the registers still needs to be looked up and specified manually. One could aswell just call `context.GetX` and `context.SetX` functions directly and get rid of the runtime generated functions. This process is also very error prone because it's easy to specify the wrong register, and then the syscall won't work properly. This is even more likely on 32-bit syscalls where long arguments are split in 2 registers, which are sometimes not consecutive.

This PR reimplements all that using source generators. Source generators is a new feature introduced on .NET 6 that allows code to be generated at compilation time based on existing code. It is basically a Roslyn compiler extension that one can specify on their projects. The advantage of using source generators is that the compilation cost is paid at compilation time, rather than at runtime, and the code also works in environments where a JIT is not available (such as when compiling in AOT mode).

Changing to source generators was not the only change. The error prone code mentioned before, where one had to manually specify registers where each argument and return value is located is no longer necessary, now it is automatically generated based on the syscall function signature. So it's no longer necessary to have one function for 32-bit mode and another for 64-bit, which ultimately calls the common function, now only the common function is necessary. This also improves logging, which now also prints the value of out parameters, rather than just the result code.

This is removing (AFAIK) one of the last uses of Reflection.Emit in the entire project, the only one that remains now is the Macro Jit, but that can be disabled easily by setting `EnableMacroJit` to false in `GraphicsConfig`. So in theory, it should be "AOT ready" with this change, but I did not try it, so there might be other surprises awaiting.

Note that this is my first time using source generators (it's a fairly new feature), so it's pretty possible that there is a better way to do certain things, so if you see anything that can be improved please comment.

As a consequence of the automatic code generation, this also fixed some bugs in the 32-bit syscalls. To give an example, we can look at the the `GetSystemInfo` syscall.
```cs
        public KernelResult GetSystemInfo32([R(1)] uint subIdLow, [R(2)] uint id, [R(3)] int handle, [R(3)] uint subIdHigh, [R(1)] out int valueLow, [R(2)] out int valueHigh)
        {
            long subId = (long)(subIdLow | ((ulong)subIdHigh << 32));

            KernelResult result = _syscall.GetSystemInfo(out long value, id, handle, subId);

            valueHigh = (int)(value >> 32);
            valueLow = (int)(value & uint.MaxValue);

            return result;
        }
```
We can see that both `handle` and `subIdHigh` are assigned to the same register 3, which is obviously incorrect.
Anyone interested on what the correct version of this function looks like, or what the auto-generated handlers looks like in general, can check it out here: https://gist.github.com/gdkchan/5f860d12fcf166684554e9949452fc4e
Note that this generated code is already a bit outdated, it's just to give an idea.

Those fixes allows more games to run. For example, Genkai Tokki Moero Crystal H currently gets stuck on this black screen on master:
image
With this PR, it's working correctly:
image
image

This change will conflict with #3362, so I prefer to wait until it is merged to merge this. Meanwhile, I will keep it as draft and also do minor improvements here (such as spliting the syscall by group using partial classes and adding stubs for the remaining syscall that will throw `NotImplementedException`). But this should be already fully functional and reviewable.


jduncanator:
Taking a look at the generated code, the kernel result code logging stood out to me immediately as a lot of redundant code duplicated across all of the generated functions. I know that this is autogenerated, and so the benefits of not having redundant code are minimised, but it would be nice to have that code broken out into a log helper (probably a private function of the same class) and used in its place, something like `LogKernelResult`. If it was just one log call, then it probably wouldn't be an issue, but the control flow statement in every generated function is a bit ugly at the moment.

gdkchan:
> Taking a look at the generated code, the kernel result code logging stood out to me immediately as a lot of redundant code duplicated across all of the generated functions. I know that this is autogenerated, and so the benefits of not having redundant code are minimised, but it would be nice to have that code broken out into a log helper (probably a private function of the same class) and used in its place, something like `LogKernelResult`. If it was just one log call, then it probably wouldn't be an issue, but the control flow statement in every generated function is a bit ugly at the moment.

I can do it, but the string formatting cost will be paid for each syscall then, regardless of logs being enabled or not (as I will need to pass the constructed string to the helper function).

jduncanator:
> > Taking a look at the generated code, the kernel result code logging stood out to me immediately as a lot of redundant code duplicated across all of the generated functions. I know that this is autogenerated, and so the benefits of not having redundant code are minimised, but it would be nice to have that code broken out into a log helper (probably a private function of the same class) and used in its place, something like `LogKernelResult`. If it was just one log call, then it probably wouldn't be an issue, but the control flow statement in every generated function is a bit ugly at the moment.
> 
> I can do it, but the string formatting cost will be paid for each syscall then, regardless of logs being enabled or not (as I will need to pass the constructed string to the helper function).

That makes sense, happy to leave it as is.

gdkchan:
I can add a helper function just to check if the result code should be printed as warning or trace. That should not have performance impact and make the ifs more compact.

gdkchan:
Marking as ready for review as this should be complete. I decided to not do the changes I mentioned at the end of the PR description because they are somewhat unrelated to those changes (it's just for organization and can wait). It will conflict with #3362 so let's see what gets merged first, I will just rebase as required.

BTW the gist has been updated with latest codegen.

riperiperi (approved):
LGTM, glad to see automated generation for the syscall register allocations too.

PR #3374, opened 2022-05-31
Extend uses count from ushort to uint on Operand Data structure

gdkchan:
Currently, the operand uses count (and capacity) is limited to a `ushort`, mostly to save memory. The maximum number of uses a variable may have is then 65535, since this is the maximum value you can store on a `ushort`. For 99% of the games that is enough, however there are a few games with very large blocks using mostly the same register over and over again. This pattern can easily reach the limit.

The fix here is simply changing the type used to store the uses count and capacity from `ushort` to `uint`, which now increases the limit to 2147483647 which is the signed 32-bit integer maximum value. The actual amount of memory used didn't actually increase since that, due to alignment, the structure size in memory was 48 bytes. It had 5 unused bytes, so now 4 of them are being occupied by the larger counts, and theres now 1 unused byte. I had to duplicate some code as the existing functions expect a `ushort` type, so I had to add `uint` overloads. With the new generic math feature that will be added on .NET 7, we can eventually de-duplicate this again.

This fixes the overflow exception when booting Taiko Risshiden V DX. I'm pretty sure that there was another game with a similar issue, but now I can't find it.
image
image

### DIFF
--- a/ARMeilleure/IntermediateRepresentation/Operand.cs
+++ b/ARMeilleure/IntermediateRepresentation/Operand.cs
@@ -14,10 +14,11 @@ namespace ARMeilleure.IntermediateRepresentation
             public byte Kind;
             public byte Type;
             public byte SymbolType;
+            public byte Padding; // Unused space.
             public ushort AssignmentsCount;
             public ushort AssignmentsCapacity;
-            public ushort UsesCount;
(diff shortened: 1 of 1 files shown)


Thog (approved):

riperiperi (approved):
LGTM

PR #3375, opened 2022-05-31
infra: Switch to win10-x64 RID and fix PR comment for Avalonia and SDL2 artifact rename

### LINKED ISSUE #3298: NOTICE: Drop support of Windows versions prior to Windows 10 1803

Thog:
Closes #3298.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
 
         - os: windows-latest
           OS_NAME: Windows x64
-          DOTNET_RUNTIME_IDENTIFIER: win-x64
+          DOTNET_RUNTIME_IDENTIFIER: win10-x64
           RELEASE_ZIP_OS_NAME: win_x64
 
       fail-fast: false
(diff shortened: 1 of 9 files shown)


emmauss (approved):

gdkchan:
Why is it necessary to change the RID of all projects? I would expect just the build script change to be enough.

gdkchan:
Most of those projects have no native dependencies and we are not publishing them directly. Maybe the RIDs can be removed from those so that we need to make less changes if we support a new platform in the future. I'm not sure if there's a better way to "unify" it.

PR #3378, opened 2022-06-01
Fix a potential GPFIFO submission race

bylaws:
The syncpoint maximum value represents the maximum possible syncpt value at a given time, however due to PBs being submitted before max was incremented, for a brief moment of time this is not the case which could lead to invalid behaviour if a game waits on the fence at that specific time.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Nv/NvDrvServices/NvHostChannel/NvHostChannelDeviceFile.cs
+++ b/Ryujinx.HLE/HOS/Services/Nv/NvDrvServices/NvHostChannel/NvHostChannelDeviceFile.cs
@@ -429,8 +429,6 @@ namespace Ryujinx.HLE.HOS.Services.Nv.NvDrvServices.NvHostChannel
                 Channel.PushHostCommandBuffer(CreateWaitCommandBuffer(header.Fence));
             }
 
-            Channel.PushEntries(entries);
-
             header.Fence.Id = _channelSyncpoint.Id;
 
             if (header.Flags.HasFlag(SubmitGpfifoFlags.FenceIncrement) || header.Flags.HasFlag(SubmitGpfifoFlags.IncrementWithValue))
@@ -448,6 +446,8 @@ namespace Ryujinx.HLE.HOS.Services.Nv.NvDrvServices.NvHostChannel
             {
                 header.Fence.Value = _device.System.HostSyncpoint.ReadSyncpointMaxValue(header.Fence.Id);
             }
+
+            Channel.PushEntries(entries);
 
             if (header.Flags.HasFlag(SubmitGpfifoFlags.FenceIncrement))
             {


gdkchan:
Can you elaborate on where exactly the race happens? Is it on the case where the game waits for the syncpoiint that was just incremented on submission? If that is the case, I think this scenario might not be possible as the incremented value was technically not returned to the game yet, unless I misunderstood something here.

bylaws:
> Can you elaborate on where exactly the race happens? Is it on the case where the game waits for the syncpoiint that was just incremented on submission? If that is the case, I think this scenario might not be possible as the incremented value was technically not returned to the game yet, unless I misunderstood something here.



> Can you elaborate on where exactly the race happens? Is it on the case where the game waits for the syncpoiint that was just incremented on submission? If that is the case, I think this scenario might not be possible as the incremented value was technically not returned to the game yet, unless I misunderstood something here.



> Can you elaborate on where exactly the race happens? Is it on the case where the game waits for the syncpoiint that was just incremented on submission? If that is the case, I think this scenario might not be possible as the incremented value was technically not returned to the game yet, unless I misunderstood something here.

Ah yeah I think you're actually right there, I just kinda assumed it could cause issues (and it could given you use ReadSyncpointMax and check on cpu buuut games don't do that) and didn't bother to look into it more. The change is still valid though

gdkchan:
If you are interested, there's a race in the syncpoint increment. This is incremented outside of the lock:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Synchronization/Syncpoint.cs#L112
And here we can see code that registers a callback:
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Synchronization/SynchronizationManager.cs#L123
It is possible for the thread to be interrupted before calling the callback. Then the waiting thread runs, gives up due to time out and disposes the event. Then the thread incrementing resumes and tries to set the event. That would throw `ObjectDisposedException` as the event was already disposed. I don't know if it has any other negative effects. I have already hit this once, but its very rare. Fixing is not as simple as calling the callback inside the lock as that would cause deadlocks elsewhere.

gdkchan (approved):
lgtm, thanks.

bylaws:
> If you are interested, there's a race in the syncpoint increment. This is incremented outside of the lock: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Synchronization/Syncpoint.cs#L112 And here we can see code that registers a callback: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Synchronization/SynchronizationManager.cs#L123 It is possible for the thread to be interrupted before calling the callback. Then the waiting thread runs, gives up due to time out and disposes the event. Then the thread incrementing resumes and tries to set the event. That would throw `ObjectDisposedException` as the event was already disposed. I don't know if it has any other negative effects. I have already hit this once, but its very rare. Fixing is not as simple as calling the callback inside the lock as that would cause deadlocks elsewhere.

Oo I might have a look into that, thanks for letting me know

PR #3380, opened 2022-06-02
Fix 3D semaphore counter type 0 handling

bylaws:
Counter type 0 actually releases the semaphore payload rather than a constant zero as was previously thought. This is required by Skyrim.


Thog (approved):
Confirmed on hardware, nice catch 👍

gdkchan (approved):
lgtm, thanks.

PR #2960, opened 2021-12-31
bsd: Revamp API and make socket abstract

Thog:
This part of the code was really ancient and needed some love.

This PR make the following changes:
- The socket implementation was separated from the IClient class. (allowing for possible native implementation of sockets in the future if needed)
- The IPC code of IClient was revamped to use more modern memory APIs for read/write/parsing of arguments.
- Sockets are now per process and refcounted making socket duplication behaves properly.
- The .NET socket is now disposed on close instead of being leaked.
- UDP DontWait was implemented on ReceiveFrom.
- Socket creation flag NonBlocking from BSD was implemented.
- BsdSocket was removed.
- Fds are now reusable.
- EventFds was implemented.
- Poll logic was rewritten to take into account "recent" HLE design changes and to handle EventFds.
- IPv6 socket creation was added (not supported on SendTo, Bind, or Connect still and will need a DNS service revamp)
- Protocol default remapping was added.
- ... Probably more that I missed.

PR #3385, opened 2022-06-05
Allow concurrent BSD EventFd read/write

gdkchan:
The `Read` and `Write` methods are supposed to be called from different threads, but there are two issues that prevents it from working in some cases:
- Currently only one thread is processing BSD IPC requests. So if any of the methods blocks, it will be unable to process requests submitted from other guest threads. This means that if one thread calls blocking `Read` and another calls `Write` after, the `Write` request will never be processed because it will be blocked on `Read`.
- Both `Read` and `Write` methods are taking the same lock. So even if the issue above didn't exist, calling both methods would result in a deadlock.

To fix the first issue I have changed `ServerBase` and added a new optional argument that allows one to specify a thread count, which defaults to 1. This thread count is the number of threads used to service requests for that service. With that implemented, I changed BSD to now use 2 threads for processing requests

For the second issue, I removed the event that it waits on and replaced the event wait and set calls with `Monitor.Wait` and `Monitor.Pulse`. `Monitor.Wait` atomically releases the lock, wait until the object is signaled, and then tries to re-acquire the lock, so it does not cause deadlock because while it is waiting, the lock is released.

This fixes a regression in Diablo 2 where the game would just hang on a black screen on boot. It was likely introduced on #2960 since that's when this code was added, but I did not test it to confirm.
image

### DIFF
--- a/Ryujinx.HLE/HOS/Horizon.cs
+++ b/Ryujinx.HLE/HOS/Horizon.cs
@@ -309,7 +309,7 @@ namespace Ryujinx.HLE.HOS
             // only then doing connections to SM is safe.
             SmServer.InitDone.WaitOne();
 
-            BsdServer = new ServerBase(KernelContext, "BsdServer");
+            BsdServer = new ServerBase(KernelContext, "BsdServer", null, 2);
             AudRenServer = new ServerBase(KernelContext, "AudioRendererServer");
             AudOutServer = new ServerBase(KernelContext, "AudioOutServer");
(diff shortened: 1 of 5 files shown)


Thog (approved):
lgtm, thanks for having taken the time to look into this 👍

riperiperi (approved):
Cool, not sure if 2 threads will always be enough but I can't see games doing much more complicated with this service.

PR #3386, opened 2022-06-05
Fix VIC out of bounds copy

gdkchan:
VIC has parameters that specifies the output rectangle that should be written on the output surface. However, those parameters were ignored. In some cases, this could cause a out of bounds read. For example, we may have a input surface with size 1920x1080 and a output surface with size 1920x1088. The additional lines on the output are due to alignment. In this case, the last 8 lines would be out of bounds of the input as its height is only 1080. When using the vectorized SSE path, this could either "work" if the memory after the input surface data happens to be mapped, or crash otherwise.

When it crashes, it would print a message like the following on console:
```
   at Ryujinx.Graphics.Vic.Blender.BlendOneSse41(Ryujinx.Graphics.Vic.Image.Surface, Ryujinx.Graphics.Vic.Image.Surface, Ryujinx.Graphics.Vic.Types.SlotStruct ByRef)
   at Ryujinx.Graphics.Vic.Blender.BlendOne(Ryujinx.Graphics.Vic.Image.Surface, Ryujinx.Graphics.Vic.Image.Surface, Ryujinx.Graphics.Vic.Types.SlotStruct ByRef)
   at Ryujinx.Graphics.Vic.VicDevice.Execute(Int32)
```
We recently got a report of this error on the game LOOPERS, and had reports of it happening on a few other games in the past.
This change fixes the issue by taking the target rectangle into account. Now it will only copy the region inside that rectangle, which should avoid out of bounds access. In addition to that, I also made it restrict the copy range inside both input and output surfaces, so now a out of bounds copy should be impossible, even if for some reason incorrect values are being passed to VIC. The latter is not required to fix the problem on the observed games, its just to be on the safe side.

Note that the crash mentioned above is random, because it depends on the contents of the memory after the input surface being mapped or not.

### DIFF
--- a/Ryujinx.Graphics.Vic/Blender.cs
+++ b/Ryujinx.Graphics.Vic/Blender.cs
@@ -10,17 +10,22 @@ namespace Ryujinx.Graphics.Vic
 {
     static class Blender
     {
-        public static void BlendOne(Surface dst, Surface src, ref SlotStruct slot)
+        public static void BlendOne(Surface dst, Surface src, ref SlotStruct slot, Rectangle targetRect)
         {
-            if (Sse41.IsSupported && (dst.Width & 3) == 0)
+            int x1 = targetRect.X;
+            int y1 = targetRect.Y;
+            int x2 = Math.Min(src.Width, x1 + targetRect.Width);
+            int y2 = Math.Min(src.Height, y1 + targetRect.Height);
+
+            if (Sse41.IsSupported && ((x1 | x2) & 3) == 0)
             {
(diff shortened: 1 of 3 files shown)


EliEron:
Can confirm that this resolves the crash in LOOPERS, thanks for the quick fix. Though as an aside the game does still have some video issues, like videos not looping when they are meant to. I realize that's a separate issue, but thought I might as well bring it up since it does impact the game quite a bit due to their use of videos for their "noise" effects.

riperiperi (approved):
Nice catch.

Thog (approved):

PR #3388, opened 2022-06-06
Ignore ClipControl on draw texture fallback

gdkchan:
When using the draw textures fallback (used when the NV_draw_texture extension is not supported), we should ignore current ClipControl settings, otherwise it can cause the viewport to be flipped. That was causing some games to be upside down on AMD and Intel, like Moero Chronicles and friends (fox example, see https://github.com/Ryujinx/Ryujinx-Games-List/issues/182#issuecomment-1147521009). NVIDIA is not affected because it supports the NV_draw_texture extension and ClipControl does not seem to affect DrawTexture.


Thog (approved):

riperiperi (approved):
Nice.

Issue #2742, opened 2021-10-16
[Suggestion] Rename vsync label to something else


CapitaineJSparrow:
Hello,

Vsync in ryujinx is really confusing for most people, for pc players it's related to tearing issues or match the framerate of their game to monitor refresh rate. In Ryujinx, This setting is used to match the framerate of the game to the *intended* framerate of the game (dunno the technical details) and when this option is off, games are playing in fast forward most of the time (some games have dynamic framerate). 

I suggest to rename this option to something else (suggestions welcome).

alimbada:
Something like `Use Native Framerate` would be more appropriate.

AcK77:
`Guest VSync` ?

Blargle:
Internal VSync? It is kinda misleading to name it something like a framerate limiter, since while it can act that way it is just a side effect of enabling and disabling it and I remember seeing reports that there are games that also don't unlock with it disabled as well. Plus it has side effects where games can take longer to load or even softlock with it disabled.

MetrosexualGarbodor:
Partially addressed by #2507, as it adds a custom framelimiter.

PR #3391, opened 2022-06-11
Update VSync message

RMED24:
Lets the end user know that the VSync option is controlling the guest VSync, not the host. With a message saying that the game may be sped up if if is disabled. I see this being a common issue with people in the Discord and hopefully at least one person doesn't make the mistake thanks to this change aha

### DIFF
--- a/Ryujinx/Ui/Windows/SettingsWindow.glade
+++ b/Ryujinx/Ui/Windows/SettingsWindow.glade
@@ -1477,7 +1477,7 @@
                                     <property name="visible">True</property>
                                     <property name="can-focus">True</property>
                                     <property name="receives-default">False</property>
-                                    <property name="tooltip-text" translatable="yes">Enables or disables Vertical Sync</property>
+                                    <property name="tooltip-text" translatable="yes">Enables or disables guest Vertical Sync. This may speed up the game beyond what was intended or increase loading times.</property>
                                     <property name="halign">start</property>
                                     <property name="margin-top">5</property>
                                     <property name="margin-bottom">5</property>
--- a/Ryujinx/Ui/Windows/SettingsWindow.glade
+++ b/Ryujinx/Ui/Windows/SettingsWindow.glade
@@ -1477,7 +1477,7 @@
                                     <property name="visible">True</property>
                                     <property name="can-focus">True</property>
                                     <property name="receives-default">False</property>
-                                    <property name="tooltip-text" translatable="yes">Enables or disables guest Vertical Sync. This may speed up the game beyond what was intended or increase loading times.</property>
+                                    <property name="tooltip-text" translatable="yes">Enables or disables guest Vertical Sync. Disabling VSync may speed up the game beyond what was intended or increase loading times.</property>
                                     <property name="halign">start</property>
                                     <property name="margin-top">5</property>
                                     <property name="margin-bottom">5</property>


gdkchan:
You should specify that the side effects will only happen if it is *disabled*. As it is now, it sounds like having it enabled is what causes the issues.

Shahil-Ayato:
Imo should add that it may cause instability if disabled. Other than that lgtm

gdkchan:
Well, does it cause instability? I was not aware of that.

RMED24:
> Imo should add that it may cause instability if disabled. Other than that lgtm

If it causes stability other than increased loading times then that's news to me

CapitaineJSparrow:
There is also desync in ldn parties

Shahil-Ayato:
> Well, does it cause instability? I was not aware of that.

I’ve could’ve sworn I read some reports on the discord that booting games with Vsync off can cause instability?

RMED24:
> There is also desync in ldn parties

Yeah I would have mentioned that if this was the LDN build. I wanna keep it sh

RMED24:
Woops

RMED24:
> You should specify that the side effects will only happen if it is _disabled_. As it is now, it sounds like having it enabled is what causes the issues.

Yeah I made that fix now

Niwu34:
> Well, does it cause instability? I was not aware of that.

Doesn't SMO have a chance to get stuck after the menu (loading screen when Vsync is disabled)

gdkchan:
> > Well, does it cause instability? I was not aware of that.
> 
> Doesn't SMO have a chance to get stuck after the menu (loading screen when Vsync is disabled)

That would be "increase loading times". AFAIK it doesn't get stuck, just very slow.

RMED24:
Yeah I knew about increased loading times which is why I mentioned but I didn't think it caused flat-out stability issues

gdkchan:
As for the message itself, I don't think it will do much since probably most people won't read it (you need to keep the mouse over the option for a while to see the tool tip text). Most people probably associates VSync disabled with more performance so probably just disables it as soon as they see the option.

piplup55:
i think a popup saying what is said in the hover text would be better

CapitaineJSparrow:
Related https://github.com/Ryujinx/Ryujinx/issues/2742

RMED24:
> As for the message itself, I don't think it will do much since probably most people won't read it (you need to keep the mouse over the option for a while to see the tool tip text). Most people probably associates VSync disabled with more performance so probably just disables it as soon as they see the option.

For sure it wouldn't do much to those that just randomly tap. Maybe I could go through all the labels and list the default option on each button? 
Sort of like where it explains each button and then shows the default just in case someone who doesn't know what they're doing messes around with Ryujinx and gets told to turn everything back to default. 

E.G: Enables or disables VSync (Default: Enabled)

gdkchan:
I would just rename it to something like `Enable VSync and frame rate limiter`.

RMED24:
> I would just rename it to something like `Enable VSync and frame rate limiter`.

Fair enough. But I might do the thing where rename this PR and add a listing as to what the default options are like I listed in my previous comment. A lot of people get told to revert the settings to the default because they messed with it without really thinking what the buttons really do. But of course, since they're new they don't really know what the default is now. 

Idk, I just think having a description as to what the default is sort of like how dolphin does it would be useful. See below: 
image

gdkchan:
> > I would just rename it to something like `Enable VSync and frame rate limiter`.
> 
> Fair enough. But I might do the thing where rename this PR and add a listing as to what the default options are like I listed in my previous comment. A lot of people get told to revert the settings to the default because they messed with it without really thinking what the buttons really do. But of course, since they're new they don't really know what the default is now.

That's a different issue. We could just add a button to restore the default settings, it's not necessary to show the default of every option (and making the user manually revert each one would be more work).

RMED24:
> > > I would just rename it to something like `Enable VSync and frame rate limiter`.
> > 
> > 
> > Fair enough. But I might do the thing where rename this PR and add a listing as to what the default options are like I listed in my previous comment. A lot of people get told to revert the settings to the default because they messed with it without really thinking what the buttons really do. But of course, since they're new they don't really know what the default is now.
> 
> That's a different issue. We could just add a button to restore the default settings, it's not necessary to show the default of every option (and making the user manually revert each one would be more work).

Okay. That sounds better to be fair. And yeah it's a different issue but I just thought it'd be the right time to mention whilst it was on my mind. Thanks

AcK77:
I think your rebase is wrong :/

riperiperi:
It looks like you've rebased on top of vulkan, instead of master.

RMED24:
> It looks like you've rebased on top of vulkan, instead of master.

I know this was a long time ago, but yeah I did when I wanted to get vulkan for SMO online back when that came out... Whoops

PR #3393, opened 2022-06-12
Catch Profile.json parse to prevent crash on launch

### LINKED ISSUE #3355: Blank "Profiles.json" in system folder crash Ryujinx
If the file "Profiles.json" contains no text within the "system" folder, Ryujinx will crash just before showing the GUI. I'm honestly not sure why it ended up blank in the first place, but deleting it allowed Ryujinx to create a new file with default information.

Open Profiles.json and remove all the text and save it. Ryujinx will crash after loading from here on out.

### Log file
No log file was created for this issue. I had to figure out why it was crashing myself using the error message.

### Environment?
It doesn't matter, this applies to all versions including the most recent.

### Additional Info
I am using a "portable" folder to keep everything contained within the main Ryujinx folder, but I doubt that will matter.

PhiZero:
Handles the Profile JSON parsing in a try catch so the application does not crash on launch, but instead loads the default profile.

Closes #3355.


Ryujinx.HLE/HOS/Services/Account/Acc/AccountSaveDataManager.cs
```diff
@@ -43,16 +45,24 @@ public AccountSaveDataManager(ConcurrentDictionary<string, UserProfile> profiles
...
+                        profiles.AddOrUpdate(profile.UserId, addedProfile, (key, old) => addedProfile);
+                    }
 
-                LastOpened = new UserId(profilesJson.LastOpened);
+                    LastOpened = new UserId(profilesJson.LastOpened);
+                }
+                catch (Exception e) 
+                {
+                    Logger.Error?.Print(LogClass.Application, $"Failed to parse {_profilesJsonPath}: {e.Message} Loading default profile!");
+                    LastOpened = AccountManager.DefaultUserId;
```
AcK77, line 64: The corrupted JSON file should be cleaned-up, but there will be some orphean save file left since they depends on the user id (only for the other than the default one)

Thog (approved):
lgtm 👍 

I do believe we should try to handle those configuration (user profiles, input profiles, ect) with better care and have common code put in place for this.

PhiZero:
Sorry, this PR completely slipped my mind. 

@AcK77 Is there anything you'd like me to address? I don't think anything needs to be cleaned, the existing Profiles.json will just get populated with a working profile again. The orphaned save files are thing, but at this point the JSON is already corrupted. On top of that, I think the save manager will help users with recovering their saves now.

AcK77:
> @AcK77 Is there anything you'd like me to address? I don't think anything needs to be cleaned, the existing Profiles.json will just get populated with a working profile again. The orphaned save files are an issue, but at this point the JSON is already corrupted. On top of that, I think the save manager will help users with recovering their saves now.

You only assign the default UserId if the parsing failed (`LastOpened = AccountManager.DefaultUserId`). So the `Profiles.json` will not be populated with a working profile except if I miss something.
I agree for the orphaned save files, those could be recovered by the save manager now ;)

PhiZero:
@AcK77 
I just tried it and when I break the json (just {} as content), the catch picks it up and generates a default RyuPlayer profile with:
"user_id": "00000000000000010000000000000000",
"last_opened": "00000000000000010000000000000000"

This is the default UserId as far as I'm aware?

Ryujinx.HLE/HOS/Services/Account/Acc/AccountSaveDataManager.cs
```diff
@@ -43,16 +45,24 @@ public AccountSaveDataManager(ConcurrentDictionary<string, UserProfile> profiles
...
+
+                        profiles.AddOrUpdate(profile.UserId, addedProfile, (key, old) => addedProfile);
+                    }
 
-                LastOpened = new UserId(profilesJson.LastOpened);
+                    LastOpened = new UserId(profilesJson.LastOpened);
+                }
+                catch (Exception e) 
+                {
+                    Logger.Error?.Print(LogClass.Application, $"Failed to parse {_profilesJsonPath}: {e.Message} Loading default profile!");
```
AcK77, line 63:
```suggestion
                    Logger.Error?.Print(LogClass.Application, $"Failed to parse {_profilesJsonPath}: {e.Message} Loading default profile!");

```

AcK77:
> @AcK77 I just tried it and when I break the json (just {} as content), the catch picks it up and generates a default RyuPlayer profile with: "user_id": "00000000000000010000000000000000", "last_opened": "00000000000000010000000000000000"
> 
> This is the default UserId as far as I'm aware?

After a quick check, it's because the user is opened as default, so it's save the JSON to write the account state.
Your change should be enough (with my latest comment applied).

AcK77 (approved):

PR #3398, opened 2022-06-12
Fix doubling of detected gamepads (sometimes the connected event is fired when the app starts even though the pad was connected for some time now).

RhavoX:
The fix rejects the gamepad if one with the same ID is already present.

Here is a screenshot of the issue in action:
image

### DIFF
--- a/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
+++ b/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
@@ -79,6 +79,11 @@ namespace Ryujinx.Input.SDL2
                     return;
                 }
 
+                // sometimes a JoyStick connected event fires after the app starts even though it was connected before
+                // so it is rejected to avoid doubling the entries
+                if (_gamepadsIds.Contains(id))
+                    return;
+
                 if (_gamepadsInstanceIdsMapping.TryAdd(joystickInstanceId, id))
                 {
                     _gamepadsIds.Add(id);


Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
```diff
@@ -79,6 +79,11 @@ private void HandleJoyStickConnected(int joystickDeviceId, int joystickInstanceI
                     return;
                 }
 
+                // sometimes a JoyStick connected event fires after the app starts even though it was connected before
```
riperiperi, line 82: Comments in ryujinx generally start with a capital letter and end with a full stop.

Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
```diff
@@ -79,6 +79,11 @@ private void HandleJoyStickConnected(int joystickDeviceId, int joystickInstanceI
...
+                // so it is rejected to avoid doubling the entries
+                if (_gamepadsIds.Contains(id))
+                    return;
```
riperiperi, line 85: Use the form of `if` with brackets.

riperiperi (commented):
Looks sensible, I've never had this happen myself but it makes sense that it shouldn't dupe IDs.

riperiperi (approved):

Thog (approved):

PR #3399, opened 2022-06-13
Optimize Texture Binding and Shader Specialization Checks

riperiperi:
This PR aims to reverse a performance regression caused by the new shader cache. The main issue is that shader specialization state was being checked for every draw, which was taking a very large amount of time, and essentially ended up iterating the texture bindings for a draw twice. (and calculating offsets, loading handles from cb...)

This changes it so that specialization state for textures is only checked when selecting shader programs (on shader address change), and when the bindings change in the texture bindings manager. If the bindings change and the shader is no longer compatible, it will _force_ shader state to be updated mid-state-update, which should be fine as none of the bindings should change.

Texture binding has been changed significantly to gain more performance in general:
- The pools have been changed to synchronize their memory only once before binding, rather than repeatedly checking on each Get() call.
 - The pools maintain and return a "modified sequence number" that increments each time the pool makes a modification. This is checked against a stored number in users so that they can figure out if the pool was modified since the last time _they_ used it, which in the case of texture bindings forces a rebind.
- The constant buffers containing texture/sampler handles are cached on first access rather than performing individual reads.
  - This speeds up draws with many textures greatly, as reading handles past the first is as easy as indexing into a span, rather than calls to the memory manager.
- Texture/Image state is remembered in terms of [BindingId], rather than [Stage][Index], and is not cleared when programs change. This allows for textures to avoid rebinds even when the program changes.
  
There are some possible downsides, but I think they are mitigated by being infrequent:
- All accesses getting textures will get a reference to the descriptor, as it is needed for checking specialization state. This could slow down a case where the bound texture changes every draw, but in games with lots of draws it's usually batched so that similar things are drawn at the same time.
- When texture and sampler buffer are not the same for each binding, the cache will be refetched each time it changes.
  - I don't think this will happen, but the code can handle it. NVN games that use regular bindings seem to behave the same as ones with eliminated bindless, where the same cb contains all the texture handles.
- When the texture/sampler constant buffer is fragmented, a copy will occur when getting span. I think this is very unlikely to happen, but if it does happen I think it would be the most noticeable cost.

## Xenoblade Chronicles Definitive Edition
This game performs an incredibly large number of draws that should probably be instanced but aren't, so the bound textures tend to stay the same. Thus, the largest improvement.

Screenshots are at FPS high points.

### Before:
image

### After:
image

## Super Mario Odyssey
It's always this game. Note that this game still performs better in Vulkan in certain spots, I think including this one, so benefits may be limited by backend performance. Similar "instanced-adjacent" behaviour where it draws the same thing with a different constant buffer bound.

### Before:
image

### After:
image

## BOTW
This game is limited by other stuff, but there's a good case in the FIFO% lowering, as it has a lot of those similar draws like Xenoblade: (or it could just be this game being weird as usual)

### Before
image

### After
image

## Notes

This changes a lot so it should be tested in a variety of games. It should especially be tested in games that shader specialization fixed in the first place, **_specifically issues involving rectangle textures._** I don't really have any.

I'll leave this draft til more testing is done.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Compute/ComputeClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute/ComputeClass.cs
@@ -237,7 +237,8 @@ namespace Ryujinx.Graphics.Gpu.Engine.Compute
 
             _channel.TextureManager.SetComputeMaxBindings(maxTextureBinding, maxImageBinding);
 
-            _channel.TextureManager.CommitComputeBindings(cs.SpecializationState); // Should be up to date, since the shader was fetched above.
+            // Should never return false for mismatching spec state, since the shader was fetched above.
+            _channel.TextureManager.CommitComputeBindings(cs.SpecializationState); 
             
             _channel.BufferManager.CommitComputeBindings();
 
--- a/Ryujinx.Graphics.Gpu/Engine/Compute/ComputeClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Compute/ComputeClass.cs
@@ -188,6 +188,9 @@ namespace Ryujinx.Graphics.Gpu.Engine.Compute
             _channel.BufferManager.SetComputeStorageBufferBindings(info.SBuffers);
             _channel.BufferManager.SetComputeUniformBufferBindings(info.CBuffers);
 
+            int maxTextureBinding = 0;
+            int maxImageBinding = 0;
+
             TextureBindingInfo[] textureBindings = _channel.TextureManager.RentComputeTextureBindings(info.Textures.Count);
 
             for (int index = 0; index < info.Textures.Count; index++)
@@ -202,6 +205,11 @@ namespace Ryujinx.Graphics.Gpu.Engine.Compute
                     descriptor.CbufSlot,
                     descriptor.HandleIndex,
                     descriptor.Flags);
+
+                if (descriptor.Binding > maxTextureBinding)
+                {
+                    maxTextureBinding = descriptor.Binding;
+                }
             }
 
             TextureBindingInfo[] imageBindings = _channel.TextureManager.RentComputeImageBindings(info.Images.Count);
@@ -220,9 +228,17 @@ namespace Ryujinx.Graphics.Gpu.Engine.Compute
                     descriptor.CbufSlot,
                     descriptor.HandleIndex,
                     descriptor.Flags);
+
+                if (descriptor.Binding > maxImageBinding)
+                {
+                    maxImageBinding = descriptor.Binding;
+                }
             }
 
-            _channel.TextureManager.CommitComputeBindings();
+            _channel.TextureManager.SetComputeMaxBindings(maxTextureBinding, maxImageBinding);
+
+            _channel.TextureManager.CommitComputeBindings(cs.SpecializationState); // Should be up to date, since the shader was fetched above.
+            
             _channel.BufferManager.CommitComputeBindings();
 
             _context.Renderer.Pipeline.DispatchCompute(qmd.CtaRasterWidth, qmd.CtaRasterHeight, qmd.CtaRasterDepth);
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -277,7 +277,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
 
             if (!_channel.TextureManager.CommitGraphicsBindings(_shaderSpecState))
             {
-                Logger.Error?.Print(LogClass.Gpu, "Oh my");
+                // Shader must be reloaded.
                 UpdateShaderState();
             }
 
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -201,7 +201,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
             // of the shader for the new state.
             if (_shaderSpecState != null)
             {
-                if (!_shaderSpecState.MatchesGraphics(_channel, GetPoolState(), GetGraphicsState()))
+                if (!_shaderSpecState.MatchesGraphics(_channel, GetPoolState(), GetGraphicsState(), false))
                 {
                     ForceShaderUpdate();
                 }
@@ -275,7 +275,12 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         {
             UpdateStorageBuffers();
 
-            _channel.TextureManager.CommitGraphicsBindings();
+            if (!_channel.TextureManager.CommitGraphicsBindings(_shaderSpecState))
+            {
+                Logger.Error?.Print(LogClass.Gpu, "Oh my");
+                UpdateShaderState();
+            }
+
             _channel.BufferManager.CommitGraphicsBindings();
         }
 
@@ -1148,6 +1153,9 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 return;
             }
 
+            int maxTextureBinding = 0;
+            int maxImageBinding = 0;
+
             Span<TextureBindingInfo> textureBindings = _channel.TextureManager.RentGraphicsTextureBindings(stage, info.Textures.Count);
 
             if (info.UsesRtLayer)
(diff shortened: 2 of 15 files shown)


gdkchan:
Yokai Watch 1 issue that was fixed by the specializatoin is back:

https://user-images.githubusercontent.com/5624669/173271803-f12a1f3f-4ff6-4e0c-9971-ff5f98f477ec.mp4

Ryujinx.Graphics.Gpu/Shader/ShaderSpecializationState.cs
```diff
@@ -396,36 +446,92 @@ private Box<TextureSpecializationState> GetTextureSpecState(int stageIndex, int
...
+        /// <param name="cachedTextureBufferIndex">The currently cached texture buffer index</param>
+        /// <param name="cachedSamplerBufferIndex">The currently cached sampler buffer index</param>
+        /// <param name="cachedTextureBuffer">The currently cached texture buffer data</param>
+        /// <param name="cachedSamplerBuffer">The currently cached sampler buffer data</param>
+        /// <param name="cachedStageIndex">The currently cached stage</param>
+        /// <param name="textureBufferIndex">The new texture buffer index</param>
+        /// <param name="samplerBufferIndex">The new sampler buffer index</param>
+        /// <param name="stageIndex">Stage index of the constant buffer</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void UpdateCachedBuffer(
```
gdkchan, line 487: I think this method can be static.

Ryujinx.Graphics.Gpu/Shader/ShaderSpecializationState.cs
```diff
@@ -445,55 +551,61 @@ private bool Matches(GpuChannel channel, GpuChannelPoolState poolState, bool isC
...
+            return true;
+        }
 
+        /// <summary>
+        /// Checks if the recorded texture state matches the given texture descriptor.
+        /// </summary>
+        /// <param name="specializationState">Texture specialization state</param>
+        /// <param name="descriptor">Texture descriptor</param>
+        /// <returns>True if the state matches, false otherwise</returns>
+
```
gdkchan, line 603: Odd separation.

Ryujinx.Graphics.Shader/TextureHandle.cs
```diff
@@ -50,5 +51,47 @@ public static (int, int, TextureHandleType) UnpackOffsets(int handle)
         {
             return (handle & 0x3fff, (handle >> 14) & 0x3fff, (TextureHandleType)((uint)handle >> 28));
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int UnpackTextureId(int packedId)
```
gdkchan, line 56: Since this was moved here, I think `UnpackSamplerId` should be moved too.

Ryujinx.Graphics.Shader/TextureHandle.cs
```diff
@@ -50,5 +51,47 @@ public static (int, int, TextureHandleType) UnpackOffsets(int handle)
...
+        {
+            return (packedId >> 0) & 0xfffff;
+        }
+
+        /// <summary>
+        /// Reads a packed texture and sampler ID (basically, the real texture handle)
```
gdkchan, line 62: One day someone should probably get rid of the "real" and "fake" texture handles in comments. It's not a very good way to call it, especially since real might mean different things.

Ryujinx.Graphics.Gpu/Shader/ShaderSpecializationState.cs
```diff
@@ -194,6 +202,48 @@ public ShaderSpecializationState(GpuChannelGraphicsState state, TransformFeedbac
...
+        {
+            _allTextures = _textureSpecialization.ToArray();
+
+            _textureByBinding = new Box<TextureSpecializationState>[stages.Length][];
+            _imageByBinding = new Box<TextureSpecializationState>[stages.Length][];
+
+            for (int i = 0; i < stages.Length; i++)
+            {
+                CachedShaderStage stage = stages[i];
+                if (stage != null && stage.Info != null)
```
gdkchan, line 219: I think it can be shortened to `stage?.Info != null`, but I don't mind how it is now.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -335,25 +360,48 @@ void ConsiderTarget(Texture target)
         /// <summary>
         /// Commits bindings on the compute pipeline.
         /// </summary>
-        public void CommitComputeBindings()
+        /// <param name="specState">Specialization state for the bound shader</param>
+        /// <returns>True if all bound textures match the current shader specialiation state, false otherwise</returns>
```
gdkchan, line 364:
```suggestion
        /// <returns>True if all bound textures match the current shader specialization state, false otherwise</returns>
```

Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -11,7 +15,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class TextureBindingsManager : IDisposable
     {
-        private const int InitialTextureStateSize = 32;
+        private const int InitialTextureStateSize = 32; // Should match binding range for host.
```
gdkchan, line 18: Are those constants no longer just a "hint"?

Ryujinx.Graphics.Gpu/Engine/Compute/ComputeClass.cs
```diff
@@ -188,6 +188,9 @@ private void SendSignalingPcasB(int argument)
             _channel.BufferManager.SetComputeStorageBufferBindings(info.SBuffers);
             _channel.BufferManager.SetComputeUniformBufferBindings(info.CBuffers);
 
+            int maxTextureBinding = 0;
```
gdkchan, line 191: Those should probably be initialized to -1 to avoid creating a 1 element array when textures or images are not used for example. Not a big deal though. Might be worth making sure that all methods receiving those values are prepared to handle a -1 value, but from a quick look it seems to be OK.

Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -1148,6 +1153,9 @@ private void UpdateStageBindings(int stage, ShaderProgramInfo info)
                 return;
             }
 
+            int maxTextureBinding = 0;
```
gdkchan, line 1156: Same comment.

Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
```diff
@@ -323,33 +331,106 @@ private void CommitRenderScale()
...
+        /// </summary>
+        /// <param name="stageIndex">Stage index of the constant buffer</param>
+        /// <param name="cachedTextureBufferIndex">The currently cached texture buffer index</param>
+        /// <param name="cachedSamplerBufferIndex">The currently cached sampler buffer index</param>
+        /// <param name="cachedTextureBuffer">The currently cached texture buffer data</param>
+        /// <param name="cachedSamplerBuffer">The currently cached sampler buffer data</param>
+        /// <param name="textureBufferIndex">The new texture buffer index</param>
+        /// <param name="samplerBufferIndex">The new sampler buffer index</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void UpdateCachedBuffer(
```
gdkchan, line 404: This looks very similar to the `UpdateCachedBuffer` method on shader specialization class. Maybe you can just call it here? Main difference seems to be the additional `cachedStageIndex` argument, but you can probably have a dummy local and set it to stage index just to enable using the existing method and avoid code duplication. I guess it's a little less efficient and might not be worth it if this method is "hot".

Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
riperiperi: No, since they are resized when the max bindings are set below. This comment was temporary. 🥴

Ryujinx.Graphics.Gpu/Shader/ShaderSpecializationState.cs
riperiperi: The `_channel` and `_isCompute` are accessed from the instance, so they would have to become arguments.

gdkchan: Both `channel` and `isCompute` are already arguments here.

riperiperi: Oh, I was looking at the other one.

gdkchan (approved):
lgtm, thanks! Tested a few games that does somewhat unusual things with textures and samplers and they all seem fine. Yokai Watch 1 is fine again too.

Thog (approved):

PR #3400, opened 2022-06-13
Support Array/3D depth-stencil render target, and single layer clears

gdkchan:
This fixes two issues.

First, it adds support for array and 3D textures to depth-stencil render targets. Before, only the color render targets had support for this, and in cases where an Array render target was used with a depth buffer for example, it would cause a framebuffer completness error on OpenGL and not draw anything.

The second and more problematic issue is that render target clears takes a layer parameter, but it was being ignored. What is worse is that it decides if the render target is layered or not based on shader state, and since the shader state is not updated for clears, it might not consider the texture layered at all. This means that often only the first layer would be cleared. To fix this, first I made it force the texture to be considered layered if the clear layer is > 0. The second part of the fix was adding a new `layer` parameter to the backend clear methods, were the layer is passed. Since OpenGL does not have a simple way to clear a single layer, I also had to add another framebuffer and attach the single layer with `glFramebufferTextureLayer` just for the clear.

This fixes missing crowd on Mario Strikers.
Before:
image
After:
image

Testing is welcome, to ensure this or other games did not regress.

### DIFF
--- a/Ryujinx.Graphics.GAL/IPipeline.cs
+++ b/Ryujinx.Graphics.GAL/IPipeline.cs
@@ -10,9 +10,10 @@ namespace Ryujinx.Graphics.GAL
 
         void ClearBuffer(BufferHandle destination, int offset, int size, uint value);
 
-        void ClearRenderTargetColor(int index, uint componentMask, ColorF color);
+        void ClearRenderTargetColor(int index, int layer, uint componentMask, ColorF color);
 
         void ClearRenderTargetDepthStencil(
+            int layer,
             float depthValue,
             bool depthMask,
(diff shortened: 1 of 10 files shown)


riperiperi (approved):
LGTM, nice work. I wonder what else might use this.

Thog (approved):

A-w-x:
this is in build 147 right? because on intel gpu this does not fix missing crowd

gdkchan:
The game might be using geometry shader passthrough (I did not confirm). If that is the case, you can try the Vulkan PR. It should work in both OpenGL and Vulkan there.

PR #3403, opened 2022-06-17
Account for res scale changes when updating bindings

riperiperi:
Fixes a regression introduced by the texture bindings PR.

Also renames TextureStatePerStage, as it's no longer per stage.

Please actually test this one (with res scale and without).


MetrosexualGarbodor:
XenoblaDE works fine now on native, 2x and 4x

image
image
imagen

MutantAura:
Tested just about everything I have on 1x 2x and 3x (BoTW, Smash, Pokemon games, DQXI etc.) and it seems fine, although I didn't have any affected games before. 👍

gdkchan (approved):
Thanks for the quick fix.

Thog (approved):

Issue #3236, opened 2022-03-27
Minecraft [0100d71004694000] Crashing on world load


piplup55:
### What's the issue you encountered?

Minecraft Crashing when loading/Making a world

### How can the issue be reproduced?

Open minecraft and try and load or make a new world

### Log file

[Ryujinx_1.1.85_2022-03-27_18-42-02.log](https://github.com/Ryujinx/Ryujinx/files/8358159/Ryujinx_1.1.85_2022-03-27_18-42-02.log)

### Environment?

- Ryujinx version: 1.1.85
- Game version: 1.0.0

- System Specs:
  - OS: EndeavourOS
  - CPU: Ryzen 7 3700X
  - GPU: NVIDIA GTX 1050
  - RAM: 64GB

- Applied Mods : no

### Additional context?

Last known working version is 1.0.7181

MetrosexualGarbodor:
This and #3237 seem to have the same error.

MutantAura:
Mario Odyssey online mod also seems to have the same error. 
Log:
[Ryujinx_1.1.150_2022-06-18_10-28-14.log](https://github.com/Ryujinx/Ryujinx/files/8934002/Ryujinx_1.1.150_2022-06-18_10-28-14.log)

I'm also updating this issue's title and content to include #3237 as they all seem to have stopped working at the same time so likely same issue.

piplup55:
ever since [SMOO PR 23](https://github.com/CraftyBoss/SuperMarioOdysseyOnline/pull/23) Super Mario Odyssey should work if you build from the dev branch

piplup55:
minecraft now goes in game, other games in this issue needs retesting
![image](https://user-images.githubusercontent.com/100526773/200086240-6d963b9b-a5e6-42fb-8bc2-a17ee9f009dc.png)

TSRBerry:
I believe this was already mentioned in the [progress report](https://blog.ryujinx.org/progress-report-august-2022/) for august.
Minecraft made me aware of the issue that was fixed with #3633

piplup55:
> I believe this was already mentioned in the [progress report](https://blog.ryujinx.org/progress-report-august-2022/) for august. Minecraft made me aware of the issue that was fixed with #3633

this is true, but i never posted a comment on this so i did it for the sake of anyone looking at this who might not know

MarkLanders777:
[...]

PR #3405, opened 2022-06-18
Handle the case where byte size option values are sent to BSD

riperiperi:
Some games and the Super Mario Odyssey Multiplayer Mod do this.

The SMO multiplayer mod also needs you to revert #3394 as it uses a blocking socket to receive (which hangs), and it doesn't seem to like being forced as non-blocking.

May fix some of the other games listed in #3236.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
@@ -304,7 +304,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
                     return LinuxError.EOPNOTSUPP;
                 }
 
-                int value = MemoryMarshal.Read<int>(optionValue);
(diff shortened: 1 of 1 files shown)


riperiperi:
https://user-images.githubusercontent.com/6294155/174457325-eacc8577-6986-4c7a-a2d5-0480289536de.mp4

Here's wha the SMO mod looks like with the other PR reverted.

gdkchan (approved):
lgtm, thanks!

Thog (approved):

gdkchan:
We got a report that the issue on Minecraft still remains. Might be worth addressing that before merging.

riperiperi:
I don't know what the issue is, as I don't have any of the games. Here are some guesses:

- TcpKeepInit is being interpreted as SoLinger because the level is not being checked when determining the game wants SoLinger (unlikely)
- Linger size is actually 5, where it's 1 byte enabled, 4 bytes time.
- Actually just wants to enable without setting time (?)

PR #3410, opened 2022-06-21
Optimize kernel memory block lookup and consolidate RBTree implementations

gdkchan:
Before, a linked list was being used to store the memory block, which is not really efficient when you need to find a block that overlaps a given address. It has O(n) complexity  in the worst case. This replaces it with a intrusive red-black tree which has O(log n) complexity and should make the lookup much faster.

A new `IntrusiveRedBlackTree` was implemented, and the existing red-black tree based collections were updated to use it to avoid code duplication.

Testing is welcome.


Ryujinx.HLE/HOS/Kernel/Memory/KMemoryBlockManager.cs
```diff
@@ -299,31 +282,7 @@ private static bool BlockStateEquals(KMemoryBlock lhs, KMemoryBlock rhs)
...
-                    {
-                        return node;
-                    }
-
-                    node = node.Next;
-                }
-            }
-
-            return null;
+            return _blockTree.GetNode(new KMemoryBlock(address, 1UL, MemoryState.Unmapped, KMemoryPermission.None, MemoryAttribute.None));
```
gdkchan, line 285: This needs to do an allocation for the `KMemoryBlock` per lookup. It might be worth having a "cached" `KMemoryBlock` where we just change the `BaseAddress` for lookup to avoid that. In theory it should be safe as all accesses to this function should be under a lock.

Thog (approved):
Forgot to finish my review for quite a bit oops

lgtm, nice work 👍 

I do believe we should start considering splitting some of our common utils to a separate project to be published on Nuget

AcK77 (approved):

PR #3411, opened 2022-06-22
Avoid lookup of invalid textures if pool did not change

gdkchan:
Currently, the cache returns null for invalid textures. Because we also use null to indicate that the texture is not "cached" on the pool, this results in the `TextureDescriptor` -> `TextureInfo` logic and lookup running over and over if there is an invalid texture on the pool. In some cases, this manifests as a "Invalid texture format" error being printed over and over in the log.

In order to prevent that from happening, I added a bitmap where we set the bit to 1 if the texture at the respective index is invalid. Then on a future call to the `Get` method in the pool, it will not attempt to look it up again and just return the descriptor and null texture directly. Should save a little bit of time in those cases, and prevent the "Invalid texture format" log spam (now the error is only printed once).

This has a little bit of cost (the additional bitmap check, which should be pretty cheap, and I don't believe that the `texture == null` path is common anyway, and a few KB of memory for the bitmap storage), but I believe it's worth it.

I noticed the error being printed on Shin Megami Tensei V. Now it's only printed once.

### DIFF
--- a/Ryujinx.Common/Collections/BitMap.cs
+++ b/Ryujinx.Common/Collections/BitMap.cs
@@ -1,0 +1,226 @@
+namespace Ryujinx.Common.Collections
+{
+    /// <summary>
+    /// Represents a collection that can store 1 bit values.
+    /// </summary>
+    public struct BitMap
+    {
+        /// <summary>
+        /// Size in bits of the integer used internally for the groups of bits.
+        /// </summary>
+        public const int IntSize = 64;
+
+        private const int IntShift = 6;
+        private const int IntMask = IntSize - 1;
+
+        private readonly long[] _masks;
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


gdkchan:
This has one potential issue. If the memory is unmapped at the time the texture is used, and then is later mapped, it will not create the texture until the descriptor is modified. Before the change it would be able to handle that case. It's worth noting that even without this change, there is a case it would also not be able to handle: a texture that was mapped, and then the mapping changed. But this case is even more unlikely. The unmapped case seems reasonable for sparse textures, so it might be worth to address this issue with a different approach (maybe it can just check if the address is valid, but that would be slower as it needs to calculate and add the mipmap offset in some cases etc).

GamerzHell9137:
Kirby doesn't spam it either 
image

Issue #3417, opened 2022-06-24
ACNH Crashes on 1.1.156


ForestSpeeder:
### What's the issue you encountered?

After running the auto-updater to 1.1.156, animal crossing new horizons no longer boots up. This can be fixed by downgrading. The error is 
```
Fatal error. 0x80131506
Repeat 2 times:
--------------------------------
   at System.Buffer.__Memmove(Byte*, Byte*, UIntPtr)
--------------------------------
   at System.Buffer._Memmove(Byte ByRef, Byte ByRef, UIntPtr)
   at Ryujinx.Memory.AddressSpaceManager.ReadImpl(UInt64, System.Span`1<Byte>)
   at Ryujinx.Memory.AddressSpaceManager.GetSpan(UInt64, Int32, Boolean)
   at Ryujinx.Memory.AddressSpaceManager.GetWritableRegion(UInt64, Int32, Boolean)
   at Ryujinx.HLE.HOS.Services.Time.StaticService.ITimeZoneServiceForGlue.LoadTimeZoneRule(Ryujinx.HLE.HOS.ServiceCtx)
   at System.RuntimeMethodHandle.InvokeMethod(System.Object, System.Span`1<System.Object> ByRef, System.Signature, Boolean, Boolean)
   at System.Reflection.RuntimeMethodInfo.Invoke(System.Object, System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
   at Ryujinx.HLE.HOS.Services.IpcService.CallHipcMethod(Ryujinx.HLE.HOS.ServiceCtx)
   at Ryujinx.HLE.HOS.Services.ServerBase.Process(Int32, UInt64)
   at Ryujinx.HLE.HOS.Services.ServerBase.ServerLoop()
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart()
```
Why does ryujinx suggest updating to "pre-release" versions anyway?

### How can the issue be reproduced?

Try to boot up ACNH on 1.1.156?

### Log file

[Ryujinx_1.1.156_2022-06-24_21-29-27.log](https://github.com/Ryujinx/Ryujinx/files/8979804/Ryujinx_1.1.156_2022-06-24_21-29-27.log)

### Environment?

- Ryujinx version: 1.1.156
- Game version: 2.0.5
- System Specs:
  - OS: Windows 10
  - CPU: i7-9700K
  - GPU: GTX 1070
[...]

PR #3418, opened 2022-06-24
timezone: Fix regression caused by #3361

Thog:
Because of that PR, TimeZoneRule was bigger than 0x4000 thanks to a
misuse of a constant.

This PR address this issue and add a new unit test to ensure the size of
TimeZoneRule is 0x4000 bytes.

Also address suggestions that were lost on the original PR.

Close https://github.com/Ryujinx/Ryujinx/issues/3417


gdkchan (approved):
Should probably use `InternalsVisibleTo` attribute rather than public for the tests, but that can be done separately to avoid blocking this important fix.

gdkchan:
Fixes #3417?

Thog:
> Should probably use `InternalsVisibleTo` attribute rather than public for the tests, but that can be done separately to avoid blocking this important fix.

Oh I never used that before, might be worth making an issue for that, that could help reducing visibility of Ryujinx.Audio too

> Fixes #3417?

Added a marking word to the PR text for automatic close.

emmauss (approved):

PR #3424, opened 2022-06-27
Bindless elimination for constant sampler handle

gdkchan:
This is extending the shader bindless texture access elimination to also work on cases where the texture handle is being combined with a constant handle value using bitwise OR operation.

Doing so allows Monster Hunter Rise Sunbreak to render (it was just showing black screen before):
image
Still does not reach the title screen due to other issues.

Testing is welcome. Probably a good idea to test The Witcher 3 and Hades since those two uses the other "odd" handle combinations.

PR #3425, opened 2022-06-27
Support non-contiguous buffers with a virtual buffer cache

gdkchan:
### Description

This an attempt to solve a long standing issue. Currently, the buffer cache only works with CPU virtual addresses. This has the advantage that buffer aliasing "just works". If two buffers at distinct GPU virtual memory regions are mapped at the same physical region, we just use the same buffer. Currently it has one problem, it assumes that all those CPU virtual memory regions where buffers are located are contiguous. In most cases, it really is contiguous, but when it isn't this causes several problems including graphical glitches and crashes.

One "natural" solution for this problem is simply using `MultiRange` for buffers and update the cache to work with buffers that have multiple CPU memory ranges. Afterall, this is what we already do with textures. However, doing so introduces various complications to the cache. When buffer memory is always contiguous, we can always coalesce overlapping buffers in the cache. That way we can always keep the cache without any overlap by coalescing the buffer. But if buffers had multiple ranges, then this is no longer possible. In this case it would be possible for a buffer, for example, share one page of memory, and then have the other pages be completely different memory locations. Coalescing such buffers would be impossible. That would force us to keep overlapping buffers on the cache which creates yet another problem: We now have multiple buffers to choose from when we want to get the buffer for a range of memory that is being used. It also causes a similar issue for buffer flushing, and for tracking handle inheritance when buffers are coalesced, because now we have repeated handles for the same regions when buffers overlap.

To avoid the nightmare described above I decided to use a simple approach that avoids most of those problems. We can simply introduce a virtual buffer cache. It works in pretty much the same way as our current buffer cache, with the main difference that it works with GPU virtual memory regions. The problem does not exist here because a GPU virtual memory range will always be contiguous. What might not be contiguous is the physical memory where it is mapped to. So evidently, this has one problem, aliasing does not work. If two GPU virtual memory regions are mapped to the same physical one, we don't have one but two buffers, which is technically not correct. So in order to get the "best of both worlds", I kept the current buffer cache (let's call it "physical buffer cache"), and added a new "virtual buffer cache". Both works with contiguous ranges, so the complications are avoided. If the physical mappings of a buffer are contiguous, it gets the buffer from the physical cache. Otherwise, it gets it from the virtual cache.

The reason for this choice is simple. It is very unlikely that a game will actively need to alias and *write* buffer memory from different virtual memory regions. If memory is being aliased, most likely it just reused memory and no longer cares about the content in one of the regions. So the priority is keeping the contents in virtual memory consistent. When it can't get a shared physical buffer, it will at least ensure that the contents on the virtual memory region are preserved.

**Deleting buffer after unmap**

Before this change, buffers were kept on the cache forever. This is not a big deal since buffers are usually small, but since this change has the potential to increase the amount of buffers, I felt it was more important to make sure they are deleted. Now they are deleted if all virtual memory that was used to access it is unmapped. Doing so also reduces the amount of aliasing which allows more buffer sharing.

### Games fixed

This fixes a crash in Monster Hunter Rise Sunbreak. Incorrect contents on a storage buffer caused infinite loop on a shader.
Before:
image
After:
image
image

It's possible that it affects other games too. But this is the only one that I'm aware of being fixed.

### What should be tested

This is a somewhat significant change to a "core" component of the GPU emulation, so it has a high chance of breaking things. It needs to be tested extensively. Probably UE4 games (always UE4 games stress testing everything) are good candidates since they like doing memory remapping in addition to other "weird" things.


MutantAura:
Tests:

MK8D: Crashes on entering any race
```
00:00:42.666 |E| GPU.MainThread Application : Unhandled exception caught: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Ryujinx.Memory.Tracking.MultiRegionHandle..ctor(MemoryTracking tracking, UInt64 address, UInt64 size, IEnumerable`1 handles, UInt64 granularity) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Memory\Tracking\MultiRegionHandle.cs:line 54
   at Ryujinx.Memory.Tracking.MemoryTracking.BeginGranularTracking(UInt64 address, UInt64 size, IEnumerable`1 handles, UInt64 granularity) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Memory\Tracking\MemoryTracking.cs:line 145
   at Ryujinx.Cpu.Jit.MemoryManagerHostMapped.BeginGranularTracking(UInt64 address, UInt64 size, IEnumerable`1 handles, UInt64 granularity) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Cpu\Jit\MemoryManagerHostMapped.cs:line 678
   at Ryujinx.Graphics.Gpu.Memory.PhysicalMemory.BeginGranularTracking(MultiRange range, IEnumerable`1 handles, UInt64 granularity) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Memory\PhysicalMemory.cs:line 384
   at Ryujinx.Graphics.Gpu.Memory.Buffer..ctor(GpuContext context, PhysicalMemory physicalMemory, MultiRange range, IEnumerable`1 baseHandles) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Memory\Buffer.cs:line 87
   at Ryujinx.Graphics.Gpu.Memory.VirtualBufferCache.CreateView(UInt64 gpuVa, UInt64 size, BufferView[] overlaps, Int32 overlapCount, Boolean forceVirtual) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Memory\VirtualBufferCache.cs:line 414
   at Ryujinx.Graphics.Gpu.Memory.VirtualBufferCache.CreateBufferAligned(UInt64 gpuVa, UInt64 size) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Memory\VirtualBufferCache.cs:line 370
   at Ryujinx.Graphics.Gpu.Memory.BufferManager.SetGraphicsUniformBuffer(Int32 stage, Int32 index, UInt64 gpuVa, UInt64 size) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Memory\BufferManager.cs:line 277
   at Ryujinx.Graphics.Gpu.Engine.Threed.ConstantBufferUpdater.Bind(Int32 argument, ShaderType type) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Engine\Threed\ConstantBufferUpdater.cs:line 95
   at Macro(MacroJitContext , IDeviceState , Int32 )
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Send(UInt64 gpuVa, Int32 offset, Int32 argument, Int32 subChannel, Boolean isLastCall) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 264
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Process(UInt64 baseGpuVa, ReadOnlySpan`1 commandBuffer) in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 95
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoDevice.DispatchCalls() in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoDevice.cs:line 188
   at Ryujinx.Ui.RendererWidgetBase.<Render>b__70_0() in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx\Ui\RendererWidgetBase.cs:line 392
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.<>c__DisplayClass48_0.<RunLoop>b__0() in C:\Users\domw0\Documents\GitHub\Ryujinx\Ryujinx.Graphics.GAL\Multithreading\ThreadedRenderer.cs:line 100
```
[Ryujinx_1.0.0-dirty_2022-06-27_15-33-03.log](https://github.com/Ryujinx/Ryujinx/files/8992975/Ryujinx_1.0.0-dirty_2022-06-27_15-33-03.log)

Legends Arceus: Visual artefacts are present (did not occur on 2nd boot so maybe random or just after a shader cache rebuild? Ignore if unable to reproduce as neither could I.)
image

Pokemon Shield: Major visual issues
image
Also crashes with the same error as above:
[Ryujinx_1.0.0-dirty_2022-06-27_16-09-14.log](https://github.com/Ryujinx/Ryujinx/files/8993174/Ryujinx_1.0.0-dirty_2022-06-27_16-09-14.log)

SSBU: No issues present.

LoZ:BoTW: Crash on loading into a savefile (same error as above etc.)
[Ryujinx_1.0.0-dirty_2022-06-27_16-18-05.log](https://github.com/Ryujinx/Ryujinx/files/8993278/Ryujinx_1.0.0-dirty_2022-06-27_16-18-05.log)

SSHD: No issues present.

SMO: No issues present.

DQXI: No issues present.

Witcher 3: No issues present.

Error messages and graphical issues above were not seen when testing #3424 alone.

gdkchan:
This is ready for review, but #3424 should be merged first.
All reported issues should be fixed now. It changed quite a bit since last time so it's worth retesting.

Ryujinx.Graphics.Gpu/Memory/VirtualBufferCache.cs
```diff
@@ -0,0 +1,763 @@
...
+            for (int i = 0; i < overlapCount; i++)
+            {
+                BufferView view = overlaps[i];
+                ulong clampedAddress = Math.Max(view.Address, e.Address);
+                ulong clampedEndAddress = Math.Min(view.EndAddress, e.Address + e.Size);
+                ulong clampedSize = clampedEndAddress - clampedAddress;
+
+                if (clampedSize != 0)
+                {
+                    view.Buffer.Unmapped((ulong)view.BaseOffset + (clampedAddress - view.Address), clampedSize);
```
gdkchan, line 78: This will clear the modified ranges, even for buffers that *might* be shared by some other view (a case of the same CPU virtual addresses being mapped in multiple GPU virtual memory locations). However it does that on master too. Shouldn't be hard to only clear if the buffer is mapped on a single GPU memory region, but not sure if it's worth it.

gdkchan:
A note for anyone reviewing it, `BufferRegion` has mostly the same code as the current `Buffer` class, while `Buffer` was almost completely changed. That's why the diff is sort of a mess.

Ryujinx.Graphics.Gpu/Memory/VirtualBufferCache.cs
```diff
@@ -0,0 +1,763 @@
...
+
+                _dirtyCache[gpuVa] = result;
+            }
+
+            result.Buffer.ForceDirty((ulong)result.BufferOffset, result.Size);
+        }
+
+        /// <summary>
+        /// Creates a new buffer for the specified range, if it does not yet exist.
+        /// This can be used to ensure the existance of a buffer.
```
riperiperi, line 251: `existence`, might have been wrong before.

Ryujinx.Graphics.Gpu/Memory/VirtualBufferCache.cs
```diff
@@ -0,0 +1,763 @@
...
+                {
+                    ref BufferView overlapView = ref overlaps[index];
+                    Buffer overlap = overlapView.Buffer;
+
+                    ulong offsetWithinOverlap = overlapView.Address - gpuVa;
+                    bool inheritable = !overlap.HasViews;
+
+                    if (inheritable)
+                    {
+                        // We can only inherit modified ranges if we are also inheriting tracking handles,
```
riperiperi, line 427: Ends with a comma?

Ryujinx.Graphics.Gpu/Memory/VirtualBufferCache.cs
```diff
@@ -0,0 +1,763 @@
...
+        private void ShrinkOverlapsBufferIfNeeded()
+        {
+            if (_viewOverlaps.Length > OverlapsBufferMaxCapacity)
+            {
+                Array.Resize(ref _viewOverlaps, OverlapsBufferMaxCapacity);
+            }
+        }
+
+        /// <summary>
+        /// Copy a buffer data from a given address to another.
```
riperiperi, line 575: `a buffer's data`?

Ryujinx.Graphics.Gpu/Memory/GpuRegionHandle.cs
```diff
@@ -51,6 +51,44 @@ public void Dispose()
...
+        /// </summary>
+        public void ForceDirty()
+        {
+            foreach (var regionHandle in _cpuRegionHandles)
+            {
+                regionHandle.ForceDirty();
+            }
+        }
+
+        public IRegionHandle GetHandle()
```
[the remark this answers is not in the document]
gdkchan: It's required to implement the `IRegionHandle` interface. Although I guess there's no need to implement this interface here. Also all the multi-range `BeginWriteTracking` tracking methods are unused right now, so it can also be removed entirely.

gdkchan:
Addressed PR feedback.

Thog (approved):

Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -194,341 +154,446 @@ public bool FullyContains(ulong address, ulong size)
...
+                if (clampedOffset > region.BaseOffset || clampedEndOffset < regionEndOffset)
                 {
-                    (address, size) = PageAlign(address, size);
-                    ranges.WaitForAndGetRanges(address, size, Flush);
+                    ulong skipSize = clampedOffset - region.BaseOffset;
+
+                    int skipCount = (int)(skipSize / BufferRegion.GranularBufferThreshold);
+                    int takeCount = (int)(clampedSize / BufferRegion.GranularBufferThreshold);
+
+                    regionHandles = regionHandles.Skip(skipCount).Take(takeCount);
```
turbedi, line 481:
```suggestion
                    regionHandles = regionHandles.Take(skipCount..skipCount + takeCount);
```
Take() accepts Range since .NET 6, avoids some allocations compared to Skip().Take()

MetrosexualGarbodor:
This change caused seemingly random crashes, albeit more frequent on LDN, in Pokémon Scarlet and Violet:

```
|E| GPU.MainThread Application : Unhandled exception caught: System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'Ryujinx.Memory.Tracking.RegionHandle'.
   at Ryujinx.Memory.Tracking.RegionHandle.Dispose() in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Memory\Tracking\RegionHandle.cs:line 428
   at Ryujinx.Graphics.Gpu.Memory.Buffer.DisposeTrackingHandles(UInt64 bufferOffset, UInt64 size) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Memory\Buffer.cs:line 557
   at Ryujinx.Graphics.Gpu.Memory.BufferView.GetTrackingHandles(UInt64 newAddress) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Memory\BufferView.cs:line 85
   at Ryujinx.Graphics.Gpu.Memory.VirtualBufferCache.CreateView(UInt64 gpuVa, UInt64 size, BufferView[] overlaps, Int32 overlapCount, Boolean forceVirtual) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Memory\VirtualBufferCache.cs:line 504
   at Ryujinx.Graphics.Gpu.Memory.VirtualBufferCache.CreateBufferAligned(UInt64 gpuVa, UInt64 size) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Memory\VirtualBufferCache.cs:line 327
   at Ryujinx.Graphics.Gpu.Engine.Threed.ConstantBufferUpdater.Bind(Int32 argument, ShaderType type) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Engine\Threed\ConstantBufferUpdater.cs:line 83
   at Macro(MacroJitContext, IDeviceState, Int32)
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Send(UInt64 gpuVa, Int32 offset, Int32 argument, Int32 subChannel, Boolean isLastCall) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 213
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoProcessor.Process(UInt64 baseGpuVa, ReadOnlySpan`1 commandBuffer) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoProcessor.cs:line 82
   at Ryujinx.Graphics.Gpu.Engine.GPFifo.GPFifoDevice.DispatchCalls() in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.Gpu\Engine\GPFifo\GPFifoDevice.cs:line 188
   at Ryujinx.Ui.RendererWidgetBase.<Render>b__75_0() in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx\Ui\RendererWidgetBase.cs:line 403
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.<>c__DisplayClass48_0.<RunLoop>b__0() in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.Graphics.GAL\Multithreading\ThreadedRenderer.cs:line 99
```
[Ryujinx_1.1.0-ldn3.0_2022-11-21_02-41-00.log](https://github.com/Ryujinx/Ryujinx/files/11023169/Ryujinx_1.1.0-ldn3.0_2022-11-21_02-41-00.log)

Posting this now because it was never properly documented and 4 months late is better than never.

gdkchan:
Superseded by #5427.

PR #3431, opened 2022-06-29
Protect against stack overflow caused by deep recursive calls

gdkchan:
On some games, the call stack might get deep, and this is problematic because it can consume a lot of stack space and eventually cause a overflow, which will cause the emulator to crash. This is possible because the CPU emulator implement guest function calls and function calls on the host too, as an optimization. In order to protect against this scenario, I have added a new `CallDepth` counter which keeps track of how deep it is on the call stack. Once it hits a given value, it will stop calling directly and return to the dispatcher. Doing so completely unwinds the stack.

This change might have a minor performance impact since we now need to do more work before each guest call, so I would recommend the following while testing:
- That there was no performance degradation.
- No noticeable increase in PPTC recompilation time.
- No large increase in the PPTC cache file size.

This fixes a crash that happens right after the first screen on Monster Hunter Rise with update 10.0.2. It should also fix a crash on the base game while looking at pictures taken with the camera on the game, but I did not check that one.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitException.cs
+++ b/ARMeilleure/Instructions/InstEmitException.cs
@@ -19,7 +19,7 @@ namespace ARMeilleure.Instructions
 
             context.LoadFromContext();
 
-            context.Return(Const(op.Address));
+            InstEmitFlowHelper.EmitReturn(context, Const(op.Address));
         }
 
         public static void Svc(ArmEmitterContext context)
@@ -49,7 +49,7 @@ namespace ARMeilleure.Instructions
 
             context.LoadFromContext();
 
-            context.Return(Const(op.Address));
+            InstEmitFlowHelper.EmitReturn(context, Const(op.Address));
         }
     }
 }
--- a/ARMeilleure/Instructions/InstEmitException32.cs
+++ b/ARMeilleure/Instructions/InstEmitException32.cs
@@ -35,7 +35,7 @@ namespace ARMeilleure.Instructions
 
             context.LoadFromContext();
 
-            context.Return(Const(context.CurrOp.Address));
+            InstEmitFlowHelper.EmitReturn(context, Const(context.CurrOp.Address));
         }
     }
 }
--- a/ARMeilleure/Instructions/InstEmitFlow.cs
+++ b/ARMeilleure/Instructions/InstEmitFlow.cs
@@ -66,7 +66,7 @@ namespace ARMeilleure.Instructions
         {
             OpCodeBReg op = (OpCodeBReg)context.CurrOp;
 
-            context.Return(GetIntOrZR(context, op.Rn));
+            EmitReturn(context, GetIntOrZR(context, op.Rn));
         }
 
         public static void Tbnz(ArmEmitterContext context) => EmitTb(context, onNotZero: true);
--- a/ARMeilleure/Instructions/InstEmitFlowHelper.cs
+++ b/ARMeilleure/Instructions/InstEmitFlowHelper.cs
@@ -12,6 +12,10 @@ namespace ARMeilleure.Instructions
 {
     static class InstEmitFlowHelper
     {
+        // How many calls we can have in our call stack before we give up and return to the dispatcher.
+        // This prevents stack overflows caused by deep recursive calls.
+        private const int MaxCallDepth = 200;
+
         public static void EmitCondBranch(ArmEmitterContext context, Operand target, Condition cond)
         {
             if (cond != Condition.Al)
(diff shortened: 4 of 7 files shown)


liamwhite:
Something you might consider doing here is to avoid using the host stack at all and simply emulate the CPU RSB yourself -- it probably is a little bit slower, but it avoids many of the headaches associated with recompiling function calls with their own host stackframes.

In the JIT state, you could have something like

```cs
ulong[] rsb_pcs;
ulong[] rsb_host_addresses;
uint rsb_index;
```

On function call,
```c
rsb_index = (rsb_index + 1) % rsb_size;
rsb_pcs[rsb_index] = pc + 4;
rsb_host_addresses[rsb_index] = (native label);
goto function;
```

On function return,
```c
bool pop_hint = rsb_pcs[rsb_index] == lr;
rsb_index = (rsb_index - 1) % rsb_size;
if (pop_hint)
{
    goto rsb_host_addresses[rsb_index];
}
else
{
    goto dispatcher;
}
```

A similar method would apply for A32/T32, but have the additional complication of needing to deal with 16-bit instructions.

gdkchan:
Something like this would require the ability to jump into the middle of a function. Right now the JIT does not support this and I'm not sure how it could be supported, as there's no place to store values that the function is using (it can save the values that directly match guest Arm registers on the `NativeContext`, but any other intermediate value on the function would be lost). Another option would be compiling a new function starting at the return point, where it would jump to on return, but then it would pollute the JIT cache with many smaller chuncks of the same function.

FICTURE7:
I think there might be a correctness issue with this solution. If the return address passes the call guard in the caller then execution will continue in the caller and will not return back to the dispatcher. It seems like a rather slim possibility that would only happen in exotic handcrafted assembly.

If we're going with this solution, then the call depth check could be moved to the entry block of the target translation instead of before each call. This should reduce the amount of code we're emitting. Ideally we would emit the increase on entry and decrease on exit of each translation, but seems like a more involved change.

Side note, I have the suspicion the reason we're stack overflowing here is because of a pathological case I've in the hybrid allocator were it spills excessively and allocates a very large stack frame.

gdkchan:
> I think there might be a correctness issue with this solution. If the return address passes the call guard in the caller then execution will continue in the caller and will not return back to the dispatcher. It seems like a rather slim possibility that would only happen in exotic handcrafted assembly.

Right, but I think it would still produce the correct results in this case, even if it doesn't return to the dispatcher, since it re-loads all guest register from the context after each call. Unless I'm missing something.

> If we're going with this solution, then the call depth check could be moved to the entry block of the target translation instead of before each call. This should reduce the amount of code we're emitting. Ideally we would emit the increase on entry and decrease on exit of each translation, but seems like a more involved change.

I don't think it would be much more complicated than this change, maybe I can give it a try. I guess doing that would also allow only enabling it for some functions (for example, the check is not needed on functions without any calls).

> Side note, I have the suspicion the reason we're stack overflowing here is because of a pathological case I've in the hybrid allocator were it spills excessively and allocates a very large stack frame.

That could be the case. It would be nice to have a register allocator that is as fast and produces better results.

gdkchan:
> I don't think it would be much more complicated than this change, maybe I can give it a try. I guess doing that would also allow only enabling it for some functions (for example, the check is not needed on functions without any calls).

I forgot one important detail, there are also tail calls, it shouldn't increase the call depth value for those. If they were added to the function entry, it would be tricky to tell, I guess it would need to store some value on the context to indicate that or pass it as an argument to the function.

FICTURE7:
> Right, but I think it would still produce the correct results in this case, even if it doesn't return to the dispatcher, since it re-loads all guest register from the context after each call. Unless I'm missing something.

Yeah you're right. After further thought it does not appear to be a problem.

> I forgot one important detail, there are also tail calls, it shouldn't increase the call depth value for those. If they were added to the function entry, it would be tricky to tell, I guess it would need to store some value on the context to indicate that or pass it as an argument to the function.

I think we can decrement the call depth just before a tailcall happens, it should avoid that issue.

> That could be the case. It would be nice to have a register allocator that is as fast and produces better results.

IIRC #2695 is pretty much ready; it is slower than the current allocator (2-5%), but the CQ gains were very good in certain cases. I will see if I can push it to the finish line.

Thog (approved):

gdkchan:
> > I forgot one important detail, there are also tail calls, it shouldn't increase the call depth value for those. If they were added to the function entry, it would be tricky to tell, I guess it would need to store some value on the context to indicate that or pass it as an argument to the function.
> 
> I think we can decrement the call depth just before a tailcall happens, it should avoid that issue.

I changed to increment the call depth at the start of the function, rather than before call. Seems to work. I also took the opportunity to get more information about the problematic calls on Monster Hunter Rise. First, I can confirm that it appears to be indeed a recursive call. There are 2 functions being called in alternation on the call stack:
```
0xE2003C0
0xE200338
0xE2003C0
0xE200338
```
(Those are guest addresses).
The stack overflows after 1596 calls.

AcK77 (approved):

MutantAura:
Performance: Couldn't notice anything user-side in any games. Seemed to be within variance.

CPU tests in bytemark (Master):
image

CPU tests in bytemark (PR):
image

PPTC recompilation tests and filesizes.
1. Run game in master.
2. Run game in PR (rebased) to recompile PPTC and time.
3. Run game in master again to recompile PPTC and time.

Super Mario Odyssey 
Master: 4.25s (63.6 MB)
PR: 4.15s (67.0 MB)

Breath of the Wild
Master: 22.44s (258 MB)
PR: 22.75s (265 MB)

Super Smash Bros: Ultimate
Master: 21.05s (183 MB)
PR: 22.18s (195 MB)

Dragon Quest XI
Master: 17.85s (151 MB)
PR: 18.31s (161 MB)

PR #3432, opened 2022-06-30
Add unsafe memory option for modded games

EwanMac3:
Adds a setting that ignores memory permission checks, allowing certain homebrew and mod menus to function within Ryujinx like on actual hardware.

### DIFF
--- a/Ryujinx.HLE/HLEConfiguration.cs
+++ b/Ryujinx.HLE/HLEConfiguration.cs
@@ -158,13 +158,30 @@ namespace Ryujinx.HLE
         /// </summary>
         public Action RefreshInputConfig { internal get; set; }
 
-        public HLEConfiguration(VirtualFileSystem virtualFileSystem, LibHacHorizonManager libHacHorizonManager,
-            ContentManager contentManager, AccountManager accountManager, UserChannelPersistence userChannelPersistence,
-            IRenderer gpuRenderer, IHardwareDeviceDriver audioDeviceDriver, MemoryConfiguration memoryConfiguration,
-            IHostUiHandler hostUiHandler, SystemLanguage systemLanguage, RegionCode region, bool enableVsync,
-            bool enableDockedMode, bool enablePtc, bool enableInternetAccess, IntegrityCheckLevel fsIntegrityCheckLevel,
-            int fsGlobalAccessLogMode, long systemTimeOffset, string timeZone, MemoryManagerMode memoryManagerMode,
-            bool ignoreMissingServices, AspectRatio aspectRatio, float audioVolume, bool allowUnsafeMem)
+        public HLEConfiguration(VirtualFileSystem      virtualFileSystem,
+                                LibHacHorizonManager   libHacHorizonManager,
+                                ContentManager         contentManager,
+                                AccountManager         accountManager,
+                                UserChannelPersistence userChannelPersistence,
+                                IRenderer              gpuRenderer,
+                                IHardwareDeviceDriver  audioDeviceDriver,
+                                MemoryConfiguration    memoryConfiguration,
+                                IHostUiHandler         hostUiHandler,
+                                SystemLanguage         systemLanguage,
+                                RegionCode             region,
+                                bool                   enableVsync,
+                                bool                   enableDockedMode,
+                                bool                   enablePtc,
+                                bool                   enableInternetAccess,
+                                IntegrityCheckLevel    fsIntegrityCheckLevel,
+                                int                    fsGlobalAccessLogMode,
+                                long                   systemTimeOffset,
+                                string                 timeZone,
+                                MemoryManagerMode      memoryManagerMode,
+                                bool                   ignoreMissingServices,
+                                AspectRatio            aspectRatio,
+                                float                  audioVolume,
+                                bool                   allowUnsafeMem);
         {
             VirtualFileSystem      = virtualFileSystem;
             LibHacHorizonManager   = libHacHorizonManager;
--- a/Ryujinx.HLE/HLEConfiguration.cs
+++ b/Ryujinx.HLE/HLEConfiguration.cs
@@ -90,7 +90,7 @@ namespace Ryujinx.HLE
         /// <summary>
         /// Controls whether code can access normally restricted memory areas.
         /// </summary>
-        internal readonly bool allowJitCodeMem;
+        internal readonly bool AllowJitCodeMem;
         
         /// <summary>
         /// Control the initial state of the docked mode.
@@ -181,7 +181,7 @@ namespace Ryujinx.HLE
                                 bool                   ignoreMissingServices,
                                 AspectRatio            aspectRatio,
                                 float                  audioVolume,
-                                bool                   allowJitCodeMemOpt)
+                                bool                   AllowJitCodeMemOpt)
         {
             VirtualFileSystem      = virtualFileSystem;
             LibHacHorizonManager   = libHacHorizonManager;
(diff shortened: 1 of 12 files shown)


MutantAura:
You may want to make these changes without formatting (or allowing your IDE to auto-format) every file that was changed. Lots of structure has been changed that aren't relevant to the PR changes.

Ryujinx.HLE/HLEConfiguration.cs
```diff
@@ -153,29 +158,13 @@ public class HLEConfiguration
...
-                                bool                   enableInternetAccess,
-                                IntegrityCheckLevel    fsIntegrityCheckLevel,
-                                int                    fsGlobalAccessLogMode,
-                                long                   systemTimeOffset,
-                                string                 timeZone,
-                                MemoryManagerMode      memoryManagerMode,
-                                bool                   ignoreMissingServices,
-                                AspectRatio            aspectRatio,
-                                float                  audioVolume)
+        public HLEConfiguration(VirtualFileSystem virtualFileSystem, LibHacHorizonManager libHacHorizonManager,
```
gdkchan, line 161: Why was this completely reformatted?

Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs
```diff
@@ -8,13 +8,16 @@
 using Ryujinx.HLE.HOS.Kernel.Process;
 using Ryujinx.HLE.HOS.Kernel.Threading;
 using System;
+using System.Management;
 using System.Threading;
 
 namespace Ryujinx.HLE.HOS.Kernel.SupervisorCall
 {
-    [SvcImpl]
+[SvcImpl]
```
gdkchan, line 16: Why was this changed.

Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs
```diff
@@ -8,13 +8,16 @@
...
     class Syscall
     {
+        public static bool allowUnsafeMemAccess;
```
gdkchan, line 19: This should not be using a static field. And it's not following our naming conventions. No changes on this file are necessary to do what you want, you can just force `AllowCodeMemoryForJit` on `KProcess` to true based on the setting.

Ryujinx.Headless.SDL2/Program.cs
```diff
@@ -484,7 +484,7 @@ private static Switch InitializeEmulationContext(WindowBase window, Options opti
                                                                   options.MemoryManagerMode,
                                                                   (bool)options.IgnoreMissingServices,
                                                                   options.AspectRatio,
-                                                                  options.AudioVolume);
+                                                                  options.AudioVolume, options.UnsafeMem);
```
gdkchan, line 487: `options.UnsafeMem` should be on a new line like the other parameters.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -574,7 +574,7 @@ private void InitializeSwitchInstance()
                                                                           ConfigurationState.Instance.System.MemoryManagerMode,
                                                                           ConfigurationState.Instance.System.IgnoreMissingServices,
                                                                           ConfigurationState.Instance.Graphics.AspectRatio,
-                                                                          ConfigurationState.Instance.System.AudioVolume);
+                                                                          ConfigurationState.Instance.System.AudioVolume, ConfigurationState.Instance.System.UnsafeMem.Value);
```
gdkchan, line 577: Should be on new line too.

Ryujinx/Ui/Windows/SettingsWindow.glade
```diff
@@ -1,2696 +1,2714 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.38.2 -->
 <interface>
-  <requires lib="gtk+" version="3.20"/>
```
gdkchan, line 4: Please keep the indentation on the file as before, no need to completely re-format it.

gdkchan (commented):
I don't think `UnsafeMem` is a good name for this feature because it does not correctly describe what it actually does. Maybe something like `Enable code memory kernel patches` and `EnableCodeKernelPatches`.

The problem with exposing something like that is that people will enable it and break some mods. And then they will ask in support or open a issue here saying that it doesn't work.

Right now we already enable it for homebrew, so it should be only a problem for mods. I think if we made the option toggle-able, would make more sense to just enable or disable it for everything.

Ryujinx/Ui/Windows/SettingsWindow.glade
EwanMac3: This one was done by my IDE, I didn't notice until it was already pointed out. Thanks for the tips, I'll revise it.

Ryujinx.HLE/HOS/Kernel/Process/KProcess.cs
```diff
@@ -103,7 +103,7 @@ public KProcess(KernelContext context, bool allowCodeMemoryForJit = false) : bas
 
             Capabilities = new KProcessCapabilities();
 
-            AllowCodeMemoryForJit = allowCodeMemoryForJit;
+            AllowCodeMemoryForJit = context.Device.Configuration.AllowJitCodeMem;
```
gdkchan, line 106: You should pass the setting value on the `allowCodeMemoryForJit` parameter instead, then you don't need to make any change on this file.

Ryujinx.HLE/HLEConfiguration.cs
```diff
@@ -175,7 +180,8 @@ public class HLEConfiguration
                                 MemoryManagerMode      memoryManagerMode,
                                 bool                   ignoreMissingServices,
                                 AspectRatio            aspectRatio,
-                                float                  audioVolume)
+                                float                  audioVolume,
+                                bool                   AllowJitCodeMemOpt)
```
gdkchan, line 184:
```suggestion
                                bool                   allowJitCodeMem)
```

Ryujinx.HLE/HLEConfiguration.cs
```diff
@@ -200,6 +206,7 @@ public class HLEConfiguration
             IgnoreMissingServices  = ignoreMissingServices;
             AspectRatio            = aspectRatio;
             AudioVolume            = audioVolume;
+            AllowJitCodeMem        = AllowJitCodeMemOpt;
```
gdkchan, line 209:
```suggestion
            AllowJitCodeMem        = allowJitCodeMem;
```

gdkchan: I'd rename `AllowJitCodeMem` to `AllowJitCodeMemory` though. Same for the lower case variant.

Ryujinx.Headless.SDL2/Options.cs
```diff
@@ -176,5 +176,8 @@ public class Options
 
         [Value(0, MetaName = "input", HelpText = "Input to load.", Required = true)]
         public string InputPath { get; set; }
+
+        [Option("allow-jit-code-mem", Required = false, Default = false, HelpText = "Allows mods to access areas of memory they normally shouldn't. Don't enable this setting unless a mod is crashing because it needs it.")]
```
gdkchan, line 180: I think the setting needs a better description. Also a mod causing crashes could be due to several reasons, I don't think just enabling this option if they crash is a good advice.

Maybe "Enable kernel code memory patches. This is required by some mods and homebrew that needs to modify code at runtime. Leave it disabled if unsure."

riperiperi:
I'm not sure a configuration option is needed for this at all. If the assumption is that any homebrew or modded games can only be run when the OS is behaving like atmosphere, then wouldn't it be enough to automatically enable this when mods are enabled or homebrew is launched?

marysaka:
Closing as stale.

PR #3438, opened 2022-07-05
Fix Vi managed and stray layers open/close/destroy

gdkchan:
This fixes the Vi implementation of the layer management related functions to match what the service does:
- `CreateStrayLayer` sets the layer state to `Stray`, `CreateManagedLayer` and friends sets the state to `ManagedClosed`.
- `OpenLayer` returns an error if the layer ID does not exist. The previous implementation would create a new layer in this case.
- `OpenLayer` returns an error if the layer state is not `ManagedClosed`. This is the case for layers created by `CreateStrayLayer`, or layers that were already opened. The previous implementation would not return an error and open it again.
- `OpenLayer` sets the state to `ManagedOpened`.
- `CloseLayer` sets the state to `ManagedClosed` if it is `ManagedOpened`. The previous implementation would fully remove the layer, which causes problems for games that closes and opens the same layer again later, like Portal.
- `Destroy*Layer` fully removes the layer. This matches the previous implementation.

This was all checked with RE of the service.

This fixes a crash on Portal caused by the game opening a layer it closed before, the previous implementation would remove the layer so that didn't work. Doesn't go ingame yet due to missing CPU instructions.

Testing is welcome, I'm a bit worried about some game depending on the previous `OpenLayer` behaviour of creating a layer if it doesn't exist to work. But it doesn't seems to be what the service does...


riperiperi (approved):
LGTM, seems to work in all the games I tried.

Thog (approved):

PR #3439, opened 2022-07-05
Implement CPU FCVT Half <-> Double conversion variants

gdkchan:
This implements the half to double and double to half conversion variants of the FCVT instruction. Fast paths does Half -> Single -> Double and Double -> Single -> Half since x86 has no instruction to convert directly between them (except with one of the AVX-512 extensions).

This is required by Portal, when combined with #3438 the game seems to be playable:
image

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdCvt.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdCvt.cs
@@ -105,21 +105,48 @@ namespace ARMeilleure.Instructions
             }
             else if (op.Size == 1 && op.Opc == 3) // Double -> Half.
             {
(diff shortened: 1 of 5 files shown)


ActualMandM:
Portal 2 also boots with the two PRs merged, although it appears to have culling issues.

https://user-images.githubusercontent.com/15317421/177249149-bc342afc-4630-48bd-8217-16e9a6624dd9.mp4

merryhime:
side note: `FPRoundCv` returning a `double` makes me uncomfortable

riperiperi (approved):
LGTM. The SoftFloat diff is pretty terrible but I guess that's because it has been reordered.

Thog (approved):

PR #3442, opened 2022-07-06
Fix buffer texture bindings on buffer resize

gdkchan:
Fixes a potential regression from #3399. Since buffer textures might be resized on the buffer cache, we might need to rebind it even if the bindings did not change otherwise, so that optimization is not safe for those. This change disables it for buffer textures.


Thog (approved):

riperiperi:
Why not have the `hostTexture != null && texture.Target == Target.TextureBuffer` path set `state.CachedTexture` to null? I think that right now, the problem is that this value is stale from a previous assignment, and the `state.CachedTexture != null` check makes it think that a cached texture is bound (but it's the previous one). This might be a bit weird but does skip a check.

Also, is it possible for this to happen with image bindings too?

gdkchan:
Superseded by #3605.

PR #3448, opened 2022-07-08
Avalonia - Ensure mouse cursor is only hidden when mouse is in renderer

emmauss:
Fixes a bug in the hide cursor feature.
The mouse cursor isn't restored  when it's hidden by an option and moves outside the game, but not the window, or if the game exits to menu.


gdkchan (approved):

Thog (approved):

GamerzHell9137:
> mouse cursor hidden is not work in windows

https://github.com/Ryujinx/Ryujinx/issues/3662

PR #3451, opened 2022-07-08
Avalonia - Couple fixes and improvements

emmauss:
Includes;

1.Fix crash with the updater on launch.
2.Reduce size of cheat window
3.Enable Tiered Compilation
4.Remove warnings from the avalonia project

PR #3454, opened 2022-07-09
Avalonia - Reduce cheat window sizes

emmauss:
Reduces the default window size and the space between items.


Niwu34:
looking good now: Thx for the quick fix
image

emmauss:
Added to https://github.com/Ryujinx/Ryujinx/pull/3451

PR #3455, opened 2022-07-09
Avalonia - Use content dialog for user profile manager

emmauss:
This moves the user profile window and related windows to a single content dialog.  The user window is now split to a selector and editor, with the selector only showing the profiles.

image
image
image
image


Niwu34:
When creating a new profile it says `Change Profile Image` which is a bit misleading:
Maybe change that to `Add Profile Image` or make a dialog that comes up when pressing ok without a profile image selected
image

gdkchan:
I like how it looks, the only thing I find weird is the "Close" button. It seems weird to me that it is separate from the other buttons, I think it would look better if they were all in the same row. And in some cases it might be a bit confusing, the second one has a "Cancel" button that one could assume has the same function as "Close", and the last one has two "Close" buttons. I guess the upper "Cancel"/"Close" button actually goes back to the previous dialog, rather than closing everything.

emmauss:
The dialog control has only 3 possible buttons; primary, secondary and cancel. They can all be hidden, but then the only way to close the dialog is pressing ESC, or Enter, which will trigger the default action, like dialogs on windows 10. So at least 1 of those buttons have to be visible. To dismiss the dialog from the content view, it needs the dialog instance to call Close(), but with the navigation in the content, the use will have to go back to the first page, the selector to close if that has the button for closing.

emmauss:
You need to also have the dialog instance to change the actions assigned to the dialog buttons

gdkchan:
I think at the very least, one of the close buttons on "Manage User Profiles" should be renamed. Having two close buttons if confusing. I'd rename the upper one to "Cancel".

Ryujinx.Ava/AppHost.cs
```diff
@@ -450,8 +452,7 @@ public async Task<bool> LoadGuestApplication()
 
                         string message = string.Format(LocaleManager.Instance["DialogFirmwareInstallEmbeddedSuccessMessage"], firmwareVersion.VersionString);
 
-                        await ContentDialogHelper.CreateInfoDialog(_parent,
-                                                             string.Format(LocaleManager.Instance["DialogFirmwareInstalledMessage"], firmwareVersion.VersionString),
+                        await ContentDialogHelper.CreateInfoDialog(string.Format(LocaleManager.Instance["DialogFirmwareInstalledMessage"], firmwareVersion.VersionString),
                                                              message,
```
gdkchan, line 456: Indentation is off.

Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -568,5 +568,8 @@
   "UpdateWindowTitle": "Manage Game Updates",
   "CheatWindowHeading": "Cheats Available for {0} [{1}]",
   "DlcWindowHeading": "DLC Available for {0} [{1}]",
-  "GameUpdateWindowHeading": "DLC Available for {0} [{1}]"
+  "UserProfilesEditProfile": "Edit Selected Profile",
+  "UserProfilesEditProfile": "Edit Selected Profile",
```
gdkchan, line 572: Same entry twice?

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -28,6 +28,7 @@
     <PackageReference Include="jp2masa.Avalonia.Flexbox" Version="0.2.0" />
     <PackageReference Include="DynamicData" Version="7.9.4" />
     <PackageReference Include="FluentAvaloniaUI" Version="1.4.1" />
+    <PackageReference Include="XamlNameReferenceGenerator" Version="1.3.4" />
```
gdkchan, line 31: Just for clarification, what is this dependency needed for?

Ryujinx.Ava/Ui/Controls/NavigatableDialogHost.axaml
```diff
@@ -0,0 +1,10 @@
+<UserControl xmlns="https://github.com/avaloniaui"
```
gdkchan, line 1: Can't comment on file names so adding it here, but the word "Navigatable" doesn't exist. The correct is Navigable.

Ryujinx.Ava/Ui/Controls/NavigatableDialogHost.axaml.cs
```diff
@@ -0,0 +1,85 @@
...
+using Ryujinx.Ava.Common.Locale;
+using Ryujinx.Ava.Ui.ViewModels;
+using Ryujinx.HLE.FileSystem;
+using Ryujinx.HLE.HOS.Services.Account.Acc;
+using System;
+using System.Threading.Tasks;
+
+namespace Ryujinx.Ava.Ui.Controls
+{
+    public partial class NavigatableDialogHost : UserControl
```
gdkchan, line 13: Same as above.

Ryujinx.Ava/Ui/Controls/ProfileImageSelectionDialog.axaml.cs
```diff
@@ -58,36 +64,30 @@ private async void Import_OnClick(object sender, RoutedEventArgs e)
 
             dialog.AllowMultiple = false;
 
-            string[] image = await dialog.ShowAsync(this);
+            string[] image = await dialog.ShowAsync(((TopLevel)_parent.GetVisualRoot()) as Window);
```
gdkchan, line 67: Is the double casting really needed here?

Ryujinx.Ava/Ui/Windows/AvatarWindow.axaml
```diff
@@ -1,32 +1,31 @@
-<Window xmlns="https://github.com/avaloniaui"
+<UserControl xmlns="https://github.com/avaloniaui"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:ui="clr-namespace:FluentAvalonia.UI.Controls;assembly=FluentAvalonia"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         mc:Ignorable="d" d:DesignWidth="400" d:DesignHeight="350"
         x:Class="Ryujinx.Ava.Ui.Windows.AvatarWindow"
-        CanResize="False"
+		Margin="0"
```
gdkchan, line 8: This is using tabs rather than spaces.

Ryujinx.Ava/Ui/Windows/AvatarWindow.axaml.cs
```diff
@@ -1,70 +1,78 @@
...
+                NavigatedTo(e);
+            }, RoutingStrategies.Direct);
+        }
+
+        private void NavigatedTo(NavigationEventArgs arg)
+        {
             if (Program.PreviewerDetached)
             {
-                Title = $"Ryujinx {Program.Version} - " + LocaleManager.Instance["AvatarWindowTitle"];
+                switch (arg.NavigationMode)
```
gdkchan, line 40: This can just use an if since there's only a single case.

Ryujinx.Ava/Ui/Windows/DlcManagerWindow.axaml.cs
```diff
@@ -125,8 +115,7 @@ private Nca TryCreateNca(IStorage ncaStorage, string containerPath)
             }
             catch (Exception ex)
             {
-                ContentDialogHelper.CreateErrorDialog(this,
-                    string.Format(LocaleManager.Instance[
+                ContentDialogHelper.CreateErrorDialog(string.Format(LocaleManager.Instance[
                         "DialogDlcLoadNcaErrorMessage"], ex.Message, containerPath));
```
gdkchan, line 119: Can you move it to the line above? Generally we only break down function call arguments at `,`.

Ryujinx.Ava/Ryujinx.Ava.csproj
emmauss: This generates properties for named elements in a xaml linked class, similar to a designer file auto generated for winforms or wpf

Ryujinx.Ava/Ui/Controls/NavigatableDialogHost.axaml
emmauss: WIll change the name to a less confusing one.

Ryujinx.Ava/Ui/Controls/ProfileImageSelectionDialog.axaml.cs
emmauss: Yeah. VisualRoot can't be casted directly to Window. I don't know why they made it that way.

emmauss:
Addressed Review

gdkchan:
On the avatar selection dialog, it seems that the list is being cut off at the right side, as you can see from the selection:
image
The scroll bar is also not showing, might be because the list if being cut.

The User Id that it shows when you click on "Add new profile" is not correct, its just 0s.

Trying to click in "Save" without setting a profile image does nothing. I think it should at least show an error message. Or maybe it should have a default image. I guess having no name written could cause the same issue.

Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -568,5 +568,11 @@
...
   "CheatWindowHeading": "Cheats Available for {0} [{1}]",
   "DlcWindowHeading": "DLC Available for {0} [{1}]",
-  "GameUpdateWindowHeading": "DLC Available for {0} [{1}]"
+  "UserProfilesEditProfile": "Edit Selected Profile",
+  "Cancel": "Cancel",
+  "Save": "Save",
+  "Discard": "Discard",
+  "UserProfilesSetProfileImage": "Set Profile Image",
+  "UserProfileEmptyNameError": "Name is required",
+  "UserProfileEmptyNoImageError":  "Profile image must be set"
```
gdkchan, line 577: Maybe that's supposed to be `UserProfileNoImageError` rather than `UserProfileEmptyNoImageError`?

Ryujinx.Ava/Ui/Controls/UserEditor.axaml.cs
```diff
@@ -0,0 +1,123 @@
...
+                ChangePictureButton.IsVisible = !_isNewUser;
+            }
+        }
+
+        private void CloseButton_Click(object sender, RoutedEventArgs e)
+        {
+            _parent?.GoBack();
+        }
+
+        private async void SaveButton_Click(object sender, RoutedEventArgs e)
```
gdkchan, line 66: Compiler is complaining about this not being awaited.

gdkchan (approved):
lgtm other than comments.

Ryujinx.Ava/Ui/Controls/UserEditor.axaml.cs
emmauss: Resolved

AcK77:
Needs rebase

emmauss:
Rebased

emmauss:
The following labels on the account selector was updated to be shorter;
UserProfilesAddNewProfile
UserProfilesDeleteSelectedProfile

Thog (approved):
lgtm

Squall-Leonhart:
changes to the DLC Manager snuck into this change and introduced issue #3491

PR #3457, opened 2022-07-10
Propagate Shader phi nodes with the same source value from all blocks

gdkchan:
In order to ensure that all variables only have a single assignment to keep optimizations simple, we keep the IR in [SSA form](https://en.wikipedia.org/wiki/Static_single-assignment_form). It is possible for a variable to be set to the same value in multiple blocks, and currently the shader translator will insert a copy on each one of them. Since it does not know that the values are actually the same, it will insert a Phi on blocks with multiple predecessors where the value is used. But the Phi is not actually needed in this case, if the value is actually the same on all of them, it doesn't matter from where it comes from.

This change extends the optimizer to also propagate the Phi source values if the value is always the same, regardless of the incoming block. This makes the generated shader code smaller and better.

The main noticeable benefit here is that bindless elimination now works on more cases, which fixes the flickering that Monster Hunter Rise had with latest update and Sunbreak expansion.

**Note:** The game still needs the other fixes (#3431 and #3425) to actually run, of course.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -21,7 +21,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 1;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 3069;
+        private const uint CodeGenVersion = 3457;
 
         private const string SharedTocFileName = "shared.toc";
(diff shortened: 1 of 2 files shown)


Thog (approved):

grayexoc (approved):

zombie167:
Almost perfect ! Thank you very much.
Except for the occasional color error of the textures on Palico & Palamute

c1
c2

Alyasa42:
not work stuck and crash on
HLE.OsThread.16 ServicePctl CheckFreeCommunicationPermission: Stubbed

MetrosexualGarbodor:
@Alyasa42 read the PR 

> Note: The game still needs the other fixes (https://github.com/Ryujinx/Ryujinx/pull/3431 and https://github.com/Ryujinx/Ryujinx/pull/3425) to actually run, of course.

GodwynMiles:
> Almost perfect ! Thank you very much. Except for the occasional color error of the textures on Palico & Palamute
> 
> c1 c2

zombie167 how do you install MH Rise v10.0.3 ?

PR #3462, opened 2022-07-12
BSD: Allow use of DontWait flag in Receive

fruityloops1:
This allows using the DontWait flag in Receive, instead of in ReceiveFrom only.


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

Issue #3377, opened 2022-06-01
latest ffmpeg is 5, yet ryu looks for the old one (4.4)


Harsh534:
### What's the issue you encountered?

Any game that (I believe) uses mpeg cut-scenes instantly gets CTD, for ex. Monster hunter rise</br>

### How can the issue be reproduced?

On fedora linux by installing ryujinx using pinejinx script from terminal, set it up and launch a game with MPEG intro cut scenes (Monster hunter rise in my case) and the emulator will CTD </br>

### Log file

[Ryujinx_1.1.136_2022-06-01_19-00-34.log](https://github.com/Ryujinx/Ryujinx/files/8814761/Ryujinx_1.1.136_2022-06-01_19-00-34.log)

### Environment?

- Ryujinx version: 1.1.136</br>
- Game version: 3.9.1</br>
- System Specs:
  - OS: Fedora Linux 36
  - CPU: Ryzen 5 5600X
  - GPU: RX 580 
  - RAM: 16GB
- Applied Mods : No

AcK77:
As the error said:
> System.DllNotFoundException: Unable to load DLL 'avutil.56 under /home/virums/.local/share/Ryujinx/': The specified module could not be found.

So I don’t think it’s related to the ffmpeg version but maybe to the way we search the library (https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Nvdec.FFmpeg/FFmpegContext.cs#L67)

edisionnano:
The abi has changed on ffmpeg 5, it should be looking for avutil.57.

ActualMandM:
Can confirm that the emulator crashes whenever a video tries to be played.
![image](https://user-images.githubusercontent.com/15317421/178592965-0827a3e8-e5b4-406b-9414-b49bfb6ca998.png)
[Ryujinx_1.1.178_2022-07-12_13-52-27.log](https://github.com/Ryujinx/Ryujinx/files/9096772/Ryujinx_1.1.178_2022-07-12_13-52-27.log)

Issue #2762, opened 2021-10-20
FFmpeg crash: Value cannot be null. (Parameter 'ptr')


Mou-Ikkai:
### What's the issue you encountered?

Any video decoding requiring FFMpeg (H264), seems to be recently crashing on AMD linux with 
```Value cannot be null. (Parameter 'ptr')```

### How can the issue be reproduced?

AMD GPU, then load a game that uses H264 video (not vp9!) For perfect 1:1 testing, use the distro I specify below in Environment.
(Project Diva's SEGA video usually crashes, so does Deemo's starting video, along with Nickelodeon All-Star Brawl's main menu)

### Log file

[Ryujinx_1.0.7082_2021-10-20_00-26-40.log](https://github.com/Ryujinx/Ryujinx/files/7378202/Ryujinx_1.0.7082_2021-10-20_00-26-40.log)
[Ryujinx_1.0.7082_2021-10-20_00-28-53.log](https://github.com/Ryujinx/Ryujinx/files/7378203/Ryujinx_1.0.7082_2021-10-20_00-28-53.log)

### Environment?

- Ryujinx version: 1.0.7082</br>
- System Specs:
  - OS: *Linux Mint 20.2 - Kernel 5.11.0-37-generic*
  - CPU: *Ryzen 9 3900X*
  - GPU: *5700XT Mesa 21.2.4*
  - RAM: *32GB DDR4 + 16GB NVMe Swap*

### Additional context?

https://pastebin.com/jYFBSt0F

PR #3466, opened 2022-07-13
Remove dependency for FFmpeg.AutoGen and Update FFmpeg to 5.0.1 for Windows

Thog:
This remove FFmpeg.AutoGen dep entirely and provide custom bindings for our usage.

Reasons for this removal:
- FFmpeg.AutoGen manually import library causing multiple issues (LDPATH being ignored is one example)
- FFmpeg.AutoGen uses ancient delegate handling making it store **every** pointers it known at static construction.
- FFmpeg.AutoGen was updated to 5.0.0 but never updated AVCodec hearder causing issues for Ryujinx migration

Because we do not currently distribute any FFmpeg libraries on Linux (except for FlatHub), and because of the fragmentation of the FFmpeg version across LTS distros, I implemented support for FFmpeg 4.4, 5.0 and 5.1 (the upcoming release)

Also updated FFmpeg Windows build to 5.0.1

I need testers on ArchLinux to confirm the ldd hack isn't necessary anymore.

Close https://github.com/Ryujinx/Ryujinx/issues/3377


Ryujinx.Graphics.Nvdec.FFmpeg/FFmpegContext.cs
```diff
@@ -1,99 +1,84 @@
...
         private readonly AVCodec* _codec;
         private AVPacket* _packet;
         private AVCodecContext* _context;
 
         public FFmpegContext(AVCodecID codecId)
         {
-            _codec = ffmpeg.avcodec_find_decoder(codecId);
-            if (_codec == null)
+            _codec = FFmpegApi.avcodec_find_decoder(codecId);
+            if ((IntPtr)_codec == IntPtr.Zero)
```
gdkchan, line 19: Why not keep checking if its null to avoid the IntPtr cast?

Ryujinx.Graphics.Nvdec.FFmpeg/FFmpegContext.cs
```diff
@@ -1,99 +1,84 @@
...
 
-            _decodeFrame = Marshal.GetDelegateForFunctionPointer<AVCodec_decode>(_codec->decode.Pointer);
+            int avCodecRawVersion = FFmpegApi.avcodec_version();
+            int avCodecMajorVersion = avCodecRawVersion >> 16;
+            int avCodecMinorVersion = (avCodecRawVersion >> 8) & 0xFF;
+
+            // libavcodec 59.24 changed AvCodec to move its private API and also move the codec function to an union.
+            if (avCodecMajorVersion > 59 || (avCodecMajorVersion == 59 && avCodecMinorVersion > 24))
+            {
+                _decodeFrame = Marshal.GetDelegateForFunctionPointer<FFCodec.AVCodec_decode>(((FFCodec*)_codec)->CodecCallback);
```
gdkchan, line 56: I wonder how this will behave if there's no decode function and instead the pointer on the union is the pointer to the other function. Currently this causes #2762, but I guess with newer FFmpeg it might cause different problems since the pointer will no longer be null?

Ryujinx.Graphics.Nvdec.FFmpeg/Native/FFmpegApi.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                return $"lib{libraryName}.so.{version}";
+            }
+            else if (OperatingSystem.IsMacOS())
+            {
+                return $"lib{libraryName}.{version}.dylib";
+            }
+            else
+            {
+                throw new NotImplementedException($"Unsupported OS for ffmpeg {RuntimeInformation.RuntimeIdentifier}");
```
gdkchan, line 35:
```suggestion
                throw new NotImplementedException($"Unsupported OS for FFmpeg: {RuntimeInformation.RuntimeIdentifier}.");
```

Ryujinx.Graphics.Nvdec.FFmpeg/Native/FFmpegApi.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        [DllImport(AvCodecLibraryName, CallingConvention = CallingConvention.Cdecl)]
+        internal static unsafe extern int avcodec_open2(AVCodecContext* avctx, AVCodec* codec, void **options);
+
+        [DllImport(AvCodecLibraryName, CallingConvention = CallingConvention.Cdecl)]
+        internal static unsafe extern int avcodec_close(AVCodecContext* avctx);
+
+        [DllImport(AvCodecLibraryName, CallingConvention = CallingConvention.Cdecl)]
+        internal static unsafe extern void avcodec_free_context(AVCodecContext** avctx);
+
```
gdkchan, line 116: Is the gap here intentional?

Ryujinx.Graphics.Nvdec.FFmpeg/FFmpegContext.cs
Thog: It's odd if that happens because, if you ask for a decoder you are supposed to only get a decoder...

So this will no longer be null yes but I don't see why it would be null in the first place.

If the AVCodec doesn't report as a decoder tho, we could detect that I think (I saw some meta information in AVClass I think)

gdkchan:
I guess it closes #3377?

Ryujinx.Graphics.Nvdec.FFmpeg/FFmpegContext.cs
Thog: oh that's a leftover from my migration stuffs will revert that change

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #3472, opened 2022-07-15
Fix decoding of block after shader BRA.CC instructions without predicate

### LINKED ISSUE #2760: Green light out of bounds on Jump Force

While playing the game JUMP FORCE Deluxe Edition, green lights are shown out of the bounds:
image
image
image

No errors on the logs.

Just start a new game and enter the tutorial.

### Log file

[Ryujinx_1.0.7081_2021-10-18_22-58-58.log](https://github.com/Ryujinx/Ryujinx/files/7369624/Ryujinx_1.0.7081_2021-10-18_22-58-58.log)

### Environment?

- Ryujinx version: 1.0.7081</br>

- Game version: 1.0.0</br>

- System Specs:
  - OS: Windows 10
  - CPU:  Ryzen 5 3600
  - GPU: NVIDIA GTX 670
  - RAM: 16GB
- Applied Mods : No 

### Additional context?

No

gdkchan:
Because the decoder did not consider branch instructions that uses the CC register as conditional, it would skip blocks right after the branch instruction and only visit the branch target address. This causes some code to be missing from the shader.

Fixes "green lights" on JUMP FORCE Deluxe.
Before:
image
After:
image
Closes #2760.

Note that the shader is still not entirely correct on this game. It has a bindless texture instruction using a handle that is the output from a shuffle instruction. The bindless elimination pass can't handle this. This is likely some compiler generated pattern that we will need to recognize and transform into the proper texture operation.

This also fixes a minor issue introduced by #3194, bindless textures were considered to have non-normalized coordinates, which caused it to generate the code to normalize them and would also add a bogus texture descriptor with a handle of 0, which later would cause the texture to be accessed from the pool and generate an error spam in the log about invalid texture formats. The fix is considering bindless textures as always being normalized, since we can't read the texture pool as the handle is unknown at compile time.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -21,7 +21,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 1;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 3478;
+        private const uint CodeGenVersion = 3472;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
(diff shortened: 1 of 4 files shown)


MetrosexualGarbodor:
Closes #2760.

Thog (approved):

AcK77 (approved):

PR #3473, opened 2022-07-15
fix: Ensure to load latest version of ffmpeg libraries first

Thog:
Fix a possible crash related to older version of ffmpeg being loaded
instead of the one shipped with the emulator.


gdkchan:
Are you sure that this fixes the issue? I think it would be better to enforce  it to use the version shipped with the emulator, regardless of whats on the system (not sure why it's not doing that already, it should always search on the application folder first, I think?)

Thog:
> Are you sure that this fixes the issue? I think it would be better to enforce it to use the version shipped with the emulator, regardless of whats on the system (not sure why it's not doing that already, it should always search on the application folder first, I think?)

We only ship ffmpeg on Windows, as already detailed on the other PR, this is to support older Linux distro still supported like Ubuntu 20.04.

AcK77 (approved):

gdkchan (approved):

PR #3481, opened 2022-07-22
Prevent textures from being flushed twice without modifications

gdkchan:
There are two cases where a flush action is registered, first when the texture is modified (`SignalModified`) is called, and the second one is when `CreateHostSyncIfNeeded` is called *after* a sync action has been registered (this is also done when the flush action is registered). The problem is that if the texture data is guest memory is accessed before the sync is created, it will trigger a flush. Then when the sync is created, the flush action is registered again, and a future access from the  GPU thread will trigger another flush even if data did not change. This is inefficient, but the main problem is that it causes glitches in some cases. On OpenGL games, a texture is first modified by 2D engine (it does a blit on GPU, that calls `SignalModified` which will make the texture flushable). I2M then pushes some data on the region at the end of the texture. This is done from GPU thread on CPU, so it triggers a flush, but no syncs were created yet. After that, a sync is created then DMA copies the texture data to another location to be read from guest on CPU. This triggers another flush, which overwrites the data from I2M with garbage data from the texture.

I fixed it by making `ClearActionRegistered` effectively "unregister" the future sync action. I also removed the flush action registration on the sync action, it seems unnecessary since it is already registered when `SignalModified` is called on the texture, unless I'm missing something (quite possible, this is not my code).

Fixes some, but not all instances of the bottom lines being corrupted on some textures on OpenGL games.
Example:
Digimon Story Cyber Sleuth (floor):
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
@@ -10,7 +10,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// A tracking handle for a texture group, which represents a range of views in a storage texture.
     /// Retains a list of overlapping texture views, a modified flag, and tracking for each
     /// CPU VA range that the views cover.
-    /// Also tracks copy dependencies for the handle - references to other handles that must be kept 
+    /// Also tracks copy dependencies for the handle - references to other handles that must be kept
     /// in sync with this one before use.
     /// </summary>
     class TextureGroupHandle : IDisposable
@@ -251,7 +251,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                     Modified = false;
                 }
-                
+
                 // If the difference is <= 0, no data is not ready yet. Flush any data we can without waiting or removing modified flag.
             }
             else
@@ -267,6 +267,14 @@ namespace Ryujinx.Graphics.Gpu.Image
         public void ClearActionRegistered()
         {
             Interlocked.Exchange(ref _actionRegistered, 0);
+
+            // If we have a pending SyncAction, make sure that it won't do anything as data was already flushed.
+            if (_syncActionRegistered)
+            {
+                SignalModifiedDirty();
+
+                _syncActionRegistered = false;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Thog (approved):

riperiperi:
There is a reason the sync action is registered twice. There are two timelines:

- Immediately after draw, a GPU method may flush the texture without waiting for sync.
- After the next syncpoint, a CPU method may read and flush the texture.

Imagine a case where the modified region is accessed almost immediately from guest CPU after it is written, but before sync. The tracking action will run, but it will detect that it is trying to flush before any sync is created, so it will skip doing it entirely (or wait for and flush an older version). Then after this point, the sync action is gone, as they can only fire once, so it must be re-registered when the sync is created, as the data is now available for guest access.

This _should_ be guarded against using the `Modified` boolean only, which is _shared_ between both timelines. This should be unset when flushing data the first time (on gpu thread this forcibly happens inline, so Modified flag should _always_ be removed), then when the action runs again, it should be a no-op.

riperiperi:
Note that there is an issue with this, but it's unrelated to the issue in this PR and involves valid accesses from both timelines:

- Modification from draw
- Attempts to trigger flush from CPU (read or write), skipped due to sync not being created yet.
- Overlapping access from GPU, but tracking action was already consumed.
  - Should have performed blocking flush on GPU thread, but action was consumed so CPU could continue.
- Sync action registers tracking again - too late.

The solution to _this_ problem would be separate tracking status for GPU/CPU, but it is very complicated and only fixes this incredibly specific scenario.

gdkchan:
> The tracking action will run, but it will detect that it is trying to flush before any sync is created, so it will skip doing it entirely (or wait for and flush an older version). Then after this point, the sync action is gone, as they can only fire once, so it must be re-registered when the sync is created, as the data is now available for guest access.

So the idea here is making multiple CPU access (some before the syncpoint, which presumably is incorrect since the GPU may not have finished writing the data, and some after) work? I assumed that the game would only attempt to access it from CPU after waiting on the syncpoint, but I guess thats a reasonable scenario aswell.

gdkchan:
I will close it since it breaks the case riperiperi described above, and it doesn't fully fix the issue anyway. Might experiment with setting texture data from I2M directly similar to what is done on DMA, but ideally we should get the flush working for this case too, since such an optimization might not catch all cases.

PR #3483, opened 2022-07-23
Avalonia - Couple fixes and improvements to vulkan

emmauss:
1. Fix up crash that occurs when you toggle fullscreen
2. Add fallback to OpenGl if vulkan is not available.
3. Add swapchain present mode control to gtk
4. Fix screenshot on avalonia vulkan 
5. Fix favorite not being saved on avalonia

### DIFF
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -94,7 +94,6 @@ namespace Ryujinx.Ava
                 .With(new Ui.Vulkan.VulkanOptions()
                 {
                     ApplicationName = "Ryujinx.Graphics.Vulkan",
-                    VulkanVersion = new Version(1, 2),
                     MaxQueueCount = 2,
                     PreferDiscreteGpu = true,
                     PreferredDevice = !PreviewerDetached ? "" : ConfigurationState.Instance.Graphics.PreferredGpu.Value,
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -188,7 +188,7 @@ namespace Ryujinx.Ava
 
                     ConfigurationState.Instance.Graphics.GraphicsBackend.Value = GraphicsBackend.OpenGl;
 
-                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable vulkan physical device is not available. Falling back to OpenGl");
+                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable Vulkan physical device is not available. Falling back to OpenGL");
                 }
             }
 
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -188,7 +188,7 @@ namespace Ryujinx.Ava
 
                     ConfigurationState.Instance.Graphics.GraphicsBackend.Value = GraphicsBackend.OpenGl;
 
-                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable Vulkan physical device is not available. Falling back to OpenGL");
+                    Logger.Warning?.PrintMsg(LogClass.Application, "A suitable Vulkan physical device is not available. Falling back to OpenGL");
                 }
             }
 
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -182,6 +182,18 @@ namespace Ryujinx.Ava
 
             if (UseVulkan)
             {
+                if (VulkanRenderer.GetPhysicalDevices().Length == 0)
+                {
+                    UseVulkan = false;
+
+                    ConfigurationState.Instance.Graphics.GraphicsBackend.Value = GraphicsBackend.OpenGl;
+
+                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable vulkan physical device is not available. Falling back to OpenGl");
+                }
+            }
+
+            if (UseVulkan)
+            {
                 // With a custom gpu backend, avalonia doesn't enable dpi awareness, so the backend must handle it. This isn't so for the opengl backed,
                 // as that uses avalonia's gpu backend and it's enabled there.
                 ForceDpiAware.Windows();
--- a/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
+++ b/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
@@ -24,7 +24,6 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
             Surface = surface;
 
             // Skia seems to only create surfaces from images with unorm format
-
             IsRgba = Display.SurfaceFormat.Format >= Format.R8G8B8A8Unorm &&
                      Display.SurfaceFormat.Format <= Format.R8G8B8A8Srgb;
 
@@ -87,6 +86,7 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
         {
             _platformInterface.Device.WaitIdle();
             Image?.Dispose();
+            Image = null;
         }
     }
 }
--- a/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
+++ b/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
@@ -22,15 +22,18 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
             _platformInterface = platformInterface;
 
             var device = VulkanInitialization.CreateDevice(platformInterface.Api,
-                                                           platformInterface.PhysicalDevice.InternalHandle,
-                                                           platformInterface.PhysicalDevice.QueueFamilyIndex,
-                                                           VulkanInitialization.GetSupportedExtensions(platformInterface.Api, platformInterface.PhysicalDevice.InternalHandle),
-                                                           platformInterface.PhysicalDevice.QueueCount);
+                platformInterface.PhysicalDevice.InternalHandle,
+                platformInterface.PhysicalDevice.QueueFamilyIndex,
+                VulkanInitialization.GetSupportedExtensions(platformInterface.Api, platformInterface.PhysicalDevice.InternalHandle),
+                platformInterface.PhysicalDevice.QueueCount);
 
             Device = new VulkanDevice(device, platformInterface.PhysicalDevice, platformInterface.Api);
 
-            Display = VulkanDisplay.CreateDisplay(platformInterface.Instance, Device,
-                platformInterface.PhysicalDevice, surface);
+            Display = VulkanDisplay.CreateDisplay(
+                platformInterface.Instance,
+                Device,
+                platformInterface.PhysicalDevice,
+                surface);
             Surface = surface;
 
             // Skia seems to only create surfaces from images with unorm format
--- a/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
+++ b/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
@@ -1,5 +1,6 @@
 using System;
 using Avalonia;
+using Ryujinx.Graphics.Vulkan;
 using Silk.NET.Vulkan;
 
 namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
@@ -14,12 +15,21 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
         private object _lock = new object();
 
         public uint MipLevels => Image.MipLevels;
+        public VulkanDevice Device { get; }
 
         public VulkanSurfaceRenderTarget(VulkanPlatformInterface platformInterface, VulkanSurface surface)
         {
             _platformInterface = platformInterface;
 
-            Display = VulkanDisplay.CreateDisplay(platformInterface.Instance, platformInterface.Device,
+            var device = VulkanInitialization.CreateDevice(platformInterface.Api,
+                                                           platformInterface.PhysicalDevice.InternalHandle,
+                                                           platformInterface.PhysicalDevice.QueueFamilyIndex,
+                                                           VulkanInitialization.GetSupportedExtensions(platformInterface.Api, platformInterface.PhysicalDevice.InternalHandle),
+                                                           platformInterface.PhysicalDevice.QueueCount);
+
+            Device = new VulkanDevice(device, platformInterface.PhysicalDevice, platformInterface.Api);
+
+            Display = VulkanDisplay.CreateDisplay(platformInterface.Instance, Device,
                 platformInterface.PhysicalDevice, surface);
             Surface = surface;
 
@@ -51,6 +61,7 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
                 DestroyImage();
                 Display?.Dispose();
                 Surface?.Dispose();
+                Device?.Dispose();
 
                 Display = null;
                 Surface = null;
@@ -67,7 +78,7 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
             _commandBuffer?.WaitForFence();
             _commandBuffer = null;
 
-            var session = new VulkanSurfaceRenderingSession(Display, _platformInterface.Device, this, scaling);
+            var session = new VulkanSurfaceRenderingSession(Display, Device, this, scaling);
 
             Image.TransitionLayout(ImageLayout.ColorAttachmentOptimal, AccessFlags.AccessNoneKhr);
 
@@ -84,7 +95,7 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
         {
             Size = Display.Size;
 
-            Image = new VulkanImage(_platformInterface.Device, _platformInterface.PhysicalDevice, Display.CommandBufferPool, ImageFormat, Size);
+            Image = new VulkanImage(Device, _platformInterface.PhysicalDevice, Display.CommandBufferPool, ImageFormat, Size);
         }
 
         private void DestroyImage()
--- a/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
+++ b/Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
@@ -7,11 +7,11 @@ namespace Ryujinx.Ava.Ui.Vulkan.Surfaces
     internal class VulkanSurfaceRenderTarget : IDisposable
     {
         private readonly VulkanPlatformInterface _platformInterface;
-
         private readonly Format _format;
 
-        public VulkanImage Image { get; private set; }
-        public bool IsCorrupted { get; private set; } = true;
+        private VulkanCommandBufferPool.VulkanCommandBuffer _commandBuffer;
+        private VulkanImage Image { get; set; }
+        private object _lock = new object();
 
         public uint MipLevels => Image.MipLevels;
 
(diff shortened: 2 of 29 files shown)


TSRBerry:
Both the settings and the main window have this sort of transparent effect at the bottom and right side.
image

Clicking the left side of the controller type dropdown will actually click the reset button for the input device.
image

Running a game in fullscreen looks like this:
image

Ryujinx.Ava/Program.cs
```diff
@@ -181,6 +180,18 @@ private static void Initialize(string[] args)
...
 
+            if (UseVulkan)
+            {
+                if (VulkanRenderer.GetPhysicalDevices().Length == 0)
+                {
+                    UseVulkan = false;
+
+                    ConfigurationState.Instance.Graphics.GraphicsBackend.Value = GraphicsBackend.OpenGl;
+
+                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable vulkan physical device is not available. Falling back to OpenGl");
```
gdkchan, line 191:
```suggestion
                    Logger.Warning?.PrintMsg(LogClass.Application, $"A suitable Vulkan physical device is not available. Falling back to OpenGL");
```

Ryujinx.Ava/Ui/Backend/Vulkan/VulkanCommandBufferPool.cs
```diff
@@ -101,6 +110,20 @@ internal unsafe VulkanCommandBuffer(VulkanDevice device, VulkanCommandBufferPool
                 device.Api.CreateFence(device.InternalHandle, fenceCreateInfo, null, out _fence);
             }
 
+            public void WaitForFence()
+            {
+                if (!_isDisposed)
```
gdkchan, line 115: You could change it to:
```cs
if (_isDisposed)
{
    return;
}
```
To avoid so much nesting.

Ryujinx.Ava/Ui/Backend/Vulkan/VulkanCommandBufferPool.cs
```diff
@@ -171,11 +194,19 @@ public void Submit()
                 _commandBufferPool.DisposeCommandBuffer(this);
             }
 
-            public void Dispose()
+            public unsafe void Dispose()
             {
-                _device.Api.WaitForFences(_device.InternalHandle, 1, _fence, true, ulong.MaxValue);
-                _device.Api.FreeCommandBuffers(_device.InternalHandle, _commandBufferPool._commandPool, 1, InternalHandle);
-                _device.Api.DestroyFence(_device.InternalHandle, _fence, Span<AllocationCallbacks>.Empty);
```
gdkchan, line 178: Why was this changed? Passing the empty span allows removing the `unsafe` keyword from the function.

Ryujinx.Ava/Ui/Backend/Vulkan/VulkanImage.cs
```diff
@@ -148,20 +148,18 @@ internal void TransitionLayout(ImageLayout destinationLayout, AccessFlags destin
...
-        }
-
         public unsafe void Dispose()
         {
-            _device.Api.DestroyImageView(_device.InternalHandle, _imageView.Value, null);
-            _device.Api.DestroyImage(_device.InternalHandle, InternalHandle.Value, null);
-            _device.Api.FreeMemory(_device.InternalHandle, _imageMemory, null);
+            if (InternalHandle != null)
+            {
+                _device.Api.DestroyImageView(_device.InternalHandle, _imageView.Value, null);
```
gdkchan, line 155: Could probably be made "safe" by passing empty span rather than null as the allocator argument.

Ryujinx.Ava/Ui/Controls/VulkanRendererControl.cs
```diff
@@ -51,19 +86,42 @@ internal override void MakeCurrent(SwappableNativeWindowBase window)
 
         internal override void Present(object image)
         {
-            Dispatcher.UIThread.InvokeAsync(() =>
+            Image = image;
+
+            lock (_imagesInFlight)
```
gdkchan, line 91: If you need to lock it anyway, why are you using a `ConcurrentQueue`? The idea of the `ConcurrentQueue` is that you don't need to use locks since its already thread safe. If you need to lock might aswell just use a regular `Queue`.

Ryujinx.Ava/Ui/ViewModels/MainWindowViewModel.cs
```diff
@@ -929,6 +930,13 @@ public static void OpenLogsFolder()
 
         public void ToggleFullscreen()
         {
+            if (Environment.TickCount64 - _lastFullscreenToggle < 500)
```
gdkchan, line 933: In which unit the value 500 is? Ticks? Might be worth adding a constant and maybe a comment here.

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -233,20 +243,31 @@ private unsafe void LoadAvailableGpus()
...
-                    names.Add($"{device.Name} {(device.IsDiscrete ? "(dGpu)" : "")}");
+                    IsVulkanAvailable = false;
+                    GraphicsBackendIndex = 1;
+                }
+                else
+                {
+                    foreach (var device in devices)
+                    {
+                        _gpuIds.Add(device.Id);
+                        names.Add($"{device.Name} {(device.IsDiscrete ? "(dGpu)" : "")}");
```
gdkchan, line 257:
```suggestion
                        names.Add($"{device.Name} {(device.IsDiscrete ? "(dGPU)" : "")}");
```

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -233,20 +243,31 @@ private unsafe void LoadAvailableGpus()
...
                 foreach (var device in VulkanPhysicalDevice.SuitableDevices)
                 {
-                    _gpuIds.Add(VulkanInitialization.StringFromIdPair(device.Value.VendorID, device.Value.DeviceID));
+                    _gpuIds.Add(
+                        VulkanInitialization.StringFromIdPair(device.Value.VendorID, device.Value.DeviceID));
                     var value = device.Value;
                     var name = value.DeviceName;
-                    names.Add($"{Marshal.PtrToStringAnsi((IntPtr)name)} {(device.Value.DeviceType == PhysicalDeviceType.DiscreteGpu ? "(dGpu)" : "")}");
+                    names.Add(
+                        $"{Marshal.PtrToStringAnsi((IntPtr)name)} {(device.Value.DeviceType == PhysicalDeviceType.DiscreteGpu ? "(dGpu)" : "")}");
```
gdkchan, line 270:
```suggestion
                        $"{Marshal.PtrToStringAnsi((IntPtr)name)} {(device.Value.DeviceType == PhysicalDeviceType.DiscreteGpu ? "(dGPU)" : "")}");
```

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -60,45 +62,57 @@ private void RecreateImages()
...
-                    SharingMode = SharingMode.Exclusive,
-                    InitialLayout = ImageLayout.Undefined,
-                    Flags = ImageCreateFlags.ImageCreateMutableFormatBit
-                };
+                SType = StructureType.ImageCreateInfo,
+                ImageType = ImageType.ImageType2D,
+                Format = Format,
+                Extent =
+                    new Extent3D((uint?)_width,
+                        (uint?)_height, 1),
```
gdkchan, line 96: That can be moved to the line above.

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -244,20 +265,42 @@ public override unsafe void Present(ITexture texture, ImageCrop crop, Action<obj
...
 
             _gd.CommandBufferPool.Return(
                 cbs,
                 null,
                 stackalloc[] { PipelineStageFlags.PipelineStageColorAttachmentOutputBit },
                 null);
 
-            var memory = _imageAllocationAuto[_nextImage].GetUnsafe().Memory;
-            var presentInfo = new PresentImageInfo(image.GetUnsafe().Value, memory, _imageSizes[_nextImage], _imageOffsets[_nextImage], _renderFinishedSemaphore, _imageAvailableSemaphore);
+            var j = _nextImage;
```
gdkchan, line 276: Why is that called `j`?

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -339,23 +385,89 @@ public override void Dispose()
...
+        private bool _isActive;
+        private readonly Auto<DisposableImage> _image;
+        private readonly Auto<MemoryAllocation> _memory;
+
+        public Image Image
+        {
+            get
+            {
+                _isActive = true;
+                _image.IncrementReferenceCount();
```
gdkchan, line 417: This seems fragile. Any access of the property would increment the reference count, and the code accessing would have no way to decrement it after. Personally I would do something that enforces the caller to decrement the reference count the same number of times after its done with it, to prevent memory leaks. For example, you can wrap it on a disposable struct and the called can put a `using` to ensure it will be disposed on function or scope exit, and the `Dispose` method can decrement the reference count. I did not look at the caller so it might not be suitable in this case though, but the general idea is that the caller must be aware it has to decrement the reference count.

Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
```diff
@@ -56,7 +56,7 @@ private MemoryAllocation Allocate(int memoryTypeIndex, ulong size, ulong alignme
             return newBl.Allocate(size, alignment, map);
         }
 
-        private static int FindSuitableMemoryTypeIndex(Vk api, PhysicalDevice physicalDevice, uint memoryTypeBits, MemoryPropertyFlags flags)
+        internal static int FindSuitableMemoryTypeIndex(Vk api, PhysicalDevice physicalDevice, uint memoryTypeBits, MemoryPropertyFlags flags)
```
gdkchan, line 59: Why was this changed? On a quick search, I can't find any of the new code calling it.

Ryujinx.Graphics.Vulkan/ImageWindow.cs
emmauss: The image isn't used and released in one method/scope. Skia holds the image until it flushes, and even then it's not guaranteed that all commands are completed on canvas.Flush. The reference s decremented when the swapchain presents or when the vulkan control is removed from view.

gdkchan:
The bug that I mentioned before still exists. If you play a game that plays an NVDEC video, and start hovering the menus very fast, first you will notice that the video starts lagging a lot. Then eventually, either the game freezes and memory starts leaking, or it just crashes immediately. This only seems to happen when it is playing NVDEC videos (maybe because usually when it does that, it is constantly uploading texture data?).

I tested with validation layers and usually there's nothing out of the ordinary before the crash, but I once got this:
```
00:00:39.306 |E| RenderTimerTickThread Gpu DebugReport: Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkBeginCommandBuffer(): object of type VkCommandPool is simultaneously used in thread 0x4eb4 and thread 0x2848
00:00:39.306 |E| Gpu DebugReport: Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkBeginCommandBuffer(): object of type VkCommandPool is simultaneously used in thread 0x2848 and thread 0x4eb4
```
To reproduce:
Play a game with a long video. I used Shantae and the Seven Sirens. When the video starts playing, click on one of the menus ("File" or whatever) and then keep moving the mouse left and right  on all the menus. Eventually it will crash, usually takes about 10-20 seconds.

jduncanator:
> The bug that I mentioned before still exists. If you play a game that plays an NVDEC video, and start hovering the menus very fast, first you will notice that the video starts lagging a lot. Then eventually, either the game freezes and memory starts leaking, or it just crashes immediately. This only seems to happen when it is playing NVDEC videos (maybe because usually when it does that, it is constantly uploading texture data?).
> 
> 
> 
> I tested with validation layers and usually there's nothing out of the ordinary before the crash, but I once got this:
> 
> ```
> 
> 00:00:39.306 |E| RenderTimerTickThread Gpu DebugReport: Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkBeginCommandBuffer(): object of type VkCommandPool is simultaneously used in thread 0x4eb4 and thread 0x2848
> 
> 00:00:39.306 |E| Gpu DebugReport: Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkBeginCommandBuffer(): object of type VkCommandPool is simultaneously used in thread 0x2848 and thread 0x4eb4
> 
> ```
> 
> To reproduce:
> 
> Play a game with a long video. I used Shantae and the Seven Sirens. When the video starts playing, click on one of the menus ("File" or whatever) and then keep moving the mouse left and right  on all the menus. Eventually it will crash, usually takes about 10-20 seconds.

For me, you don't even need to interact with the menu. If you launch a game that plays a video whilst you're idle on the menu, if you leave it long enough, it will gradually slow down and then just crash. You don't even have to touch the mouse or keyboard.

I can't see anything out of the ordinary except VRAM slowly going up, but the crash doesn't seem to be an OOM - at least from what I can tell.

gdkchan:
As of latest commit, the crash while videos are being played has been fixed here, but now there's a NullReferenceException when closing the window.

Ryujinx.Ava/Ui/Backend/Vulkan/VulkanDisplay.cs
```diff
@@ -15,16 +14,30 @@ internal class VulkanDisplay : IDisposable
...
-        private VulkanDevice _device { get; }
-        private ImageView[] _swapchainImageViews = new ImageView[0];
+        private ImageView[] _swapchainImageViews = Array.Empty<ImageView>();
         private bool _vsyncStateChanged;
         private bool _vsyncEnabled;
+        private bool _surfaceChanged;
+
+        public event EventHandler Presented;
+
+        public bool NotifySurfaceChanged
```
gdkchan, line 31: This name seems odd for what it does. It is called `NotifySurfaceChanged` but it actually sets `_surfaceChanged` to false. I think this should probably be a regular method, IMO proprties should not do much more than just getting or setting internal state, and I would not expect accessing a property to have "side effects".

Ryujinx.Ava/Ui/Controls/RendererControl.cs
```diff
@@ -13,22 +13,10 @@ internal abstract class RendererControl : Control
...
-        }
-
-        public readonly static StyledProperty<object> ImageProperty =
-            AvaloniaProperty.Register<RendererControl, object>(
-                nameof(Image),
-                0,
-                inherits: true,
-                defaultBindingMode: BindingMode.TwoWay);
-
         protected object Image
```
gdkchan, line 16: Maybe this could become an auto-property now?

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -339,23 +365,81 @@ public override void Dispose()
...
-        public ulong MemorySize { get; set; }
-        public ulong MemoryOffset { get; set; }
-        public Semaphore ReadySemaphore { get; }
-        public Semaphore AvailableSemaphore { get; }
-
-        public PresentImageInfo(Image image, DeviceMemory memory, ulong memorySize, ulong memoryOffset, Semaphore readySemaphore, Semaphore availableSemaphore)
+        private readonly Auto<DisposableImage> _image;
+        private readonly Auto<MemoryAllocation> _memory;
+
+        public Image Image
```
gdkchan, line 391: Those two can be just:
```cs
public Image Image => _image.GetUnsafe().Value;
```

Ryujinx.Graphics.Vulkan/WindowBase.cs
```diff
@@ -10,5 +10,6 @@ internal abstract class WindowBase: IWindow
         public abstract void Dispose();
         public abstract void Present(ITexture texture, ImageCrop crop, Action<object> swapBuffersCallback);
         public abstract void SetSize(int width, int height);
+        public virtual void ChangeVSyncMode(bool vsyncEnabled) { }
```
gdkchan, line 13: Why is that not abstract like the others?

Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -402,6 +402,8 @@ public void Render()
                 Device.Gpu.InitializeShaderCache(_gpuCancellationTokenSource.Token);
                 Translator.IsReadyForTranslation.Set();
 
+                Renderer?.Window?.ChangeVSyncMode(Device.EnableDeviceVsync);
```
gdkchan, line 405: Can Window really be null here?

gdkchan (commented):
Does this close https://github.com/Ryujinx/Ryujinx/issues/3534 ?

Ryujinx/Ui/RendererWidgetBase.cs
emmauss: No

gdkchan (approved):
lgtm, thanks.

Ryujinx.Ava/Ui/Backend/Vulkan/Surfaces/VulkanSurfaceRenderTarget.cs
```diff
@@ -1,30 +1,39 @@
...
-            Display = VulkanDisplay.CreateDisplay(platformInterface.Instance, platformInterface.Device,
+            var device = VulkanInitialization.CreateDevice(platformInterface.Api,
+                                                           platformInterface.PhysicalDevice.InternalHandle,
+                                                           platformInterface.PhysicalDevice.QueueFamilyIndex,
+                                                           VulkanInitialization.GetSupportedExtensions(platformInterface.Api, platformInterface.PhysicalDevice.InternalHandle),
+                                                           platformInterface.PhysicalDevice.QueueCount);
+
+            Device = new VulkanDevice(device, platformInterface.PhysicalDevice, platformInterface.Api);
+
+            Display = VulkanDisplay.CreateDisplay(platformInterface.Instance, Device,
```
AcK77, line 32: It should be fine on one line

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -18,87 +20,86 @@ class ImageWindow : WindowBase, IWindow, IDisposable
...
+                SType = StructureType.ImageCreateInfo,
+                ImageType = ImageType.ImageType2D,
+                Format = Format,
+                Extent = new Extent3D((uint?)_width, (uint?)_height, 1),
+                MipLevels = 1,
+                ArrayLayers = 1,
+                Samples = SampleCountFlags.SampleCount1Bit,
+                Tiling = ImageTiling.Optimal,
+                Usage = ImageUsageFlags.ImageUsageColorAttachmentBit | ImageUsageFlags.ImageUsageTransferSrcBit |
+                    ImageUsageFlags.ImageUsageTransferDstBit,
```
AcK77, line 90: Should be on the above line

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -205,13 +208,18 @@ public override unsafe void Present(ITexture texture, ImageCrop crop, Action<obj
 
             if (ScreenCaptureRequested)
             {
-                CaptureFrame(view, srcX0, srcY0, srcX1 - srcX0, srcY1 - srcY0, view.Info.Format.IsBgr(), crop.FlipX, crop.FlipY);
+                CaptureFrame(view, srcX0, srcY0, srcX1 - srcX0, srcY1 - srcY0, view.Info.Format.IsBgr(), crop.FlipX,
+                    crop.FlipY);
```
AcK77, line 212: Should be on the above line

Ryujinx.Graphics.Vulkan/ImageWindow.cs
```diff
@@ -205,13 +208,18 @@ public override unsafe void Present(ITexture texture, ImageCrop crop, Action<obj
...
 
                 ScreenCaptureRequested = false;
             }
 
-            float ratioX = crop.IsStretched ? 1.0f : MathF.Min(1.0f, _height * crop.AspectRatioX / (_width * crop.AspectRatioY));
-            float ratioY = crop.IsStretched ? 1.0f : MathF.Min(1.0f, _width * crop.AspectRatioY / (_height * crop.AspectRatioX));
+            float ratioX = crop.IsStretched
```
AcK77, line 217: The formatting was better before, on one line (same for ratioY)

AcK77 (approved):

PR #3487, opened 2022-07-25
Add a sampler pool cache and improve texture pool cache

gdkchan:
Currently, the texture pool cache has a hard limit of 4 pools. If a game is actively using more than that, then the least recently used one is removed. This is not good for performance as we will need to re-create the pool (and possibly the textures) again later. In addition to that, there is no sampler pool cache, so the emulator needs to re-create the pool and all samplers every time the sampler pool changes (which may or may not happen frequently, most games only uses one pool, so there's no problem there).

To address those issues, this change takes a new factor into account before deleting pools. It only deletes pools that have not been used for some time. In order to avoid introducing non-deterministic behavior, it does not actually use time, rather it uses a "timestamp" that is incremented on every draw/dispatch. The pool capacity has been reduced from 4 to 2, since they are only removed if it is old enough now, so for actively used pools, the cache is capacity is basically unlimited.

A sampler pool cache has been added, it behaves exactly the same as the texture pool cache.

This improves performance on Super Zangyura.
Before:

https://user-images.githubusercontent.com/5624669/180701274-aac03efc-cd5a-44b7-8913-afb13cd160d0.mp4

After:

https://user-images.githubusercontent.com/5624669/180701320-74a02299-d86f-429a-a0ba-6eebcbf1028a.mp4

Note that most games won't show any performance improvement because they only use one pool. Only games that uses multiple pools are improved.
Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureManager.cs
@@ -16,6 +16,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         private readonly TextureBindingsManager _cpBindingsManager;
         private readonly TextureBindingsManager _gpBindingsManager;
         private readonly TexturePoolCache _texturePoolCache;
+        private readonly SamplerPoolCache _samplerPoolCache;
 
         private readonly Texture[] _rtColors;
         private readonly ITexture[] _rtHostColors;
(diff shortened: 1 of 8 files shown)


Ryujinx.Graphics.Gpu/Image/TextureManager.cs
```diff
@@ -382,6 +389,9 @@ public bool CommitComputeBindings(ShaderSpecializationState specState)
         /// <returns>True if all bound textures match the current shader specialization state, false otherwise</returns>
         public bool CommitGraphicsBindings(ShaderSpecializationState specState)
         {
+            _texturePoolCache.Tick();
```
riperiperi, line 392: The current number is 20000, but this method is called for every draw. Some games might spam a _lot_ of draws - maybe it could invalidate a pool before the frame is completed?

riperiperi (approved):
LGTM, can't really test right now though.

Ryujinx.Graphics.Gpu/Image/TextureManager.cs
gdkchan: Yes, but 20K is quite a lot. Are you suggesting increasing it then? Or using a different method to increment this counter?

AcK77 (approved):

PR #2518, opened 2021-07-30
Vulkan backend

gdkchan:
### Overview

This is a WIP implementation of a Vulkan graphics backend. The main goal here is improving the performance for AMD users, however there are other advantages of using Vulkan for all vendors.

Main advantages of Vulkan (for all vendors):
- Support for compressed 3D textures: Fixes graphical glitches on Tales of Vesperia, Ghost & Goblins returns, Monster Hunter Rise and more.
- Support for textures views between compressed and uncompressed formats. Avoids copies, we do not yet take advantage of this, but will in the future.
- Supports for 2D array views from 3D textures. Avoids copies, we do not yet take advantage of this, but will in the future.
- Allows transform feedback buffers to have other uses while the transform feedback is active. Avoids copies.

### Pending work

Those items should be completed before the Vulkan implementation is merged.
- [x] Multisampling support.
- [x] Alpha test support.
- [x] Custom image formats at bind. Or maybe change how it works to play nicely with Vulkan? (Required to render ground etc correctly on Xenoblade Chronicles Definitive Edition, fire/ice etc effects on Kirby Star Allies).
- [ ] Make sure it plays nicely with the AMD driver (there are currently some known bugs with AMD drivers, more on that later).
- [x] Integrate SPIR-V backend and make sure it is functionally equivalent to the GLSL one.
- [ ] Write basic tests for shaders and the backend, to ensure they behave as expected.

Other nice-to-haves, not essential for merge:
- [ ] Index buffer conversion for quads. Should improve the speed of games using quads.
- [x] Better Uint8 index buffer conversion. ~~Should have some very minor performance improvement for AMD.~~ AMD now supports Uint8 index buffers.
- [ ] Experiment with other descriptor set update methods. May improve performance. I already made quite a lot of tests in this area however, and it doesn't seems like there's much we can do to improve it further.
- [x] ~~Consider using a separate thread for command submission. Might benefit Intel due to the workaround that makes it submit commands a lot.~~ No longer an issue, the workaround was removed as the bug was fixed on their driver.
- [x] Pipeline cache. Effectively a on-disk shader cache on Vulkan. (Note: We're not using pipeline caches, instead it caches the SPIR-V binary).

### Known issues

[...]

PR #3514, opened 2022-08-01
Fix resolution scale values not being updated

gdkchan:
#3399 introduced some regressions on resolution scaling that #3403 attempted to fix, however even with this fix some issues remained. I attempted to fix some of the remaining issues on #2518 because it was reported that resolution scaling was broken on Atelier Ryza and other games from the series. However, this introduced yet another issue on Xenoblade Chronicles Definitive Edition.

The problem seems to be caused by the interaction between textures and images. The scales for both is stored on  the same buffer and same array, yet on the texture bindings manager their state is separate. Right now, the only things that forces an update is the scale index, usage flags or the texture itself changing. One simple scenario where this could break is a game using a image that is scaled on a single draw. None of that state would change, yet any other texture using the same scale index on a past or future draw would overwrite the scale value, which would require an update when the image is used, but that doesn't happen right now.

I have fixed this by removing the optimization. In order to avoid calling the `UpdateScale` function for every single texture, I have added a check to call it only if a scale is needed.

Fixes regression when scaling Xenoblade Chronicles Definitive Edition.
Before:
image
After:
image

Testing is welcome, would be also nice to get confirmation that performance was not affected. Also worth testing on other games since I think there were some weird resolution scale issues on other games that didn't happen before?

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -50,8 +50,6 @@ namespace Ryujinx.Graphics.Gpu.Image
             public int InvalidatedSequence;
             public Texture CachedTexture;
             public Sampler CachedSampler;
-            public int ScaleIndex;
-            public TextureUsageFlags UsageFlags;
         }
 
         private TextureState[] _textureState;
@@ -535,14 +533,12 @@ namespace Ryujinx.Graphics.Gpu.Image
                     // The texture is already bound.
                     state.CachedTexture.SynchronizeMemory();
 
-                    if ((state.ScaleIndex != index || state.UsageFlags != usageFlags) &&
+                    if ((usageFlags & TextureUsageFlags.NeedsScaleValue) != 0 &&
                         UpdateScale(state.CachedTexture, usageFlags, index, stage))
                     {
                         ITexture hostTextureRebind = state.CachedTexture.GetTargetTexture(bindingInfo.Target);
 
                         state.Texture = hostTextureRebind;
-                        state.ScaleIndex = index;
-                        state.UsageFlags = usageFlags;
 
                         _context.Renderer.Pipeline.SetTextureAndSampler(stage, bindingInfo.Binding, hostTextureRebind, state.Sampler);
                     }
@@ -573,7 +569,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 {
                     bool textureOrSamplerChanged = state.Texture != hostTexture || state.Sampler != hostSampler;
 
-                    if ((state.ScaleIndex != index || state.UsageFlags != usageFlags || textureOrSamplerChanged) &&
+                    if ((usageFlags & TextureUsageFlags.NeedsScaleValue) != 0 &&
                         UpdateScale(texture, usageFlags, index, stage))
                     {
                         hostTexture = texture?.GetTargetTexture(bindingInfo.Target);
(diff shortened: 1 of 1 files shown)


sakuka47:
For mario odyssey, just a few slowdowns for the creation of vulkan backend shaders)

Kirby does not work under vulkan, due to an "invalid textures format 0X7ffff( sRGB false)" message.

Under openGL I have the same message in red but the game works with upscale in 2x

GamerzHell9137:
> For mario odyssey, just a few slowdowns for the creation of vulkan backend shaders)
> 
> Kirby does not work under vulkan, due to an "invalid textures format 0X7ffff( sRGB false)" message.
> 
> Under openGL I have the same message in red but the game works with upscale in 2x

Which Kirby? Forgotten Land is fine.
image

sakuka47:
> 

I had to create shaders in openGL to be able to start the game in vulkan

GamerzHell9137:
> Not for me I have same error message on Vulkan after ryujinx shutdonw
> 
> 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Invalid instruction at 0x006DF8 (0x0000000000000000). 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Shader instruction Votevtg is not implemented.

Not a bug, its showing for everyone.

sakuka47:
> > Not for me I have same error message on Vulkan after ryujinx shutdonw
> > 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Invalid instruction at 0x006DF8 (0x0000000000000000). 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Shader instruction Votevtg is not implemented.
> 
> Not a bug, its showing for everyone.

certainly but at home ryujinx cuts in any case on vulkan, on openGL I have the message but the game is running.

GamerzHell9137:
> > > Not for me I have same error message on Vulkan after ryujinx shutdonw
> > > 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Invalid instruction at 0x006DF8 (0x0000000000000000). 00:00:07.720 |W| GPU.MainThread Gpu Log: Shader translator: Shader instruction Votevtg is not implemented.
> > 
> > 
> > Not a bug, its showing for everyone.
> 
> certainly but at home ryujinx cuts in any case on vulkan, on openGL I have the message but the game is running.

That's unrelated to this issue and PR tho.

PhiZero:
This also fixes the same(?) issue in XBC3. I was certain there were similar noticeable issues on XBC2, and I think I showed a screenshot of this in Vulkan Report channel on discord, but I cannot reproduce this anymore and see no issues at a glance.

1.1.201 @ 2x:
11201_2x

This PR @ 2x:
pr3514_2x

On this PR at 2x (and on master, barring the artifacting that this PR addresses), the image is blurrier than on 1x. I'm not sure if this is in the scope of this PR however.

1.1.201 @ 1x:
11201_1x

AcK77 (approved):

sakuka47:
Also works for mario kart 8

riperiperi:
Was anything wrong in Mario Kart 8? It doesn't bind images.

sakuka47:
No worries, I test the games I own to see if there is a regression on any of them.

Arnauw:
Nice fix !
I can confirm it also fixes the same issue happening with trees in the Marsh.
This also the lights that were "flickering" when moving and using upscaling (like the trees lights in the Marsh at night for example)

OldManKain:
**Confirmed** Fixes the res scale bugs in Xenoblade Chronicles Definitive Edition.

riperiperi (approved):
I think this works for now. Only draws that use previously rendered inputs (that need scale) should trigger this path, so it won't happen often.

PR #3520, opened 2022-08-01
Standardise GPU naming convention on the UI

MutantAura:
Currently OpenGL and Vulkan return different strings for the GPU vendor section of the UI. 
image
image

This is more annoying for AMD GPU's as their OpenGL designation is actually still "ATI technologies inc." which is a very legacy name and has confused some users in the past. Intel is the same in both APIs so no issues there.
This change makes the vendor title for OpenGL equal to the Vulkan titles. Also took the chance to remove a couple of redundant strings in the GTK render bar, calling the function directly is what Avalonia does and seems more intuitive.

image

### DIFF
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -196,7 +196,7 @@ namespace Ryujinx.Ava
                 return (AmdVendor);
 
                 default:
-                return _renderer.GetHardwareInfo().GpuVendor;
+                return vendor;
             }
         }
 
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -78,8 +78,6 @@ namespace Ryujinx.Ava
         private bool _isActive;
         private long _lastCursorMoveTime;
 
-        private string _gpuVendorName;
-
         private KeyboardHotkeyState _prevHotkeyState;
 
         private IRenderer _renderer;
@@ -895,7 +893,6 @@ namespace Ryujinx.Ava
             // Run a status update only when a frame is to be drawn. This prevents from updating the ui and wasting a render when no frame is queued
             string dockedMode = ConfigurationState.Instance.System.EnableDockedMode ? LocaleManager.Instance["Docked"] : LocaleManager.Instance["Handheld"];
             float scale = GraphicsConfig.ResScale;
-            _gpuVendorName = GetGpuVendorName();
 
             if (scale != 1)
             {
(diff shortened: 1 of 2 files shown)


EmulationFanatic:
How, if at all, will this affect GPU vendor-specific workarounds (like the view format AMD workaround) ?

MutantAura:
> How, if at all, will this affect GPU vendor-specific workarounds (like the view format AMD workaround) ?

Not at all. This calls the exact same functions to get the raw string, such as "NVIDIA corporation" and simply intercepts and changes it before it reaches the UI. Nothing here touches any of the backend detection.

gdkchan:
Personally I prefer how it was before since it's something returned from the driver, it's not exactly something we control. On Vulkan, we get a vendor ID instead and map the names to the vendor ID. From what I recall, mesa has different strings too, sometimes returning "AMD", but i have also seen "X.Org" (not the full name, I just remember that it starts with X.Org).

MutantAura:
While it isn't something we control I don't think the GUI is necessarily a place to be displaying purely functional information. Personally I think if we can make the whole experience more consistent and intuitive for general users then it's worth it. 
The function defaults to just the driver name should a match not be found (you mentioned the mesa scenario) so it should be the same as usual for those edge cases.

Given OpenGL is _somewhat_ of an option for AMD going forward it seems unfriendly for their vendor name to be "ATI Technologies inc." as many people have questioned in the past.

Entirely up to you though. Nvidia's name is less annoying and so it doesn't affect me either way :)

Mixaill:
For AMD Vulkan, it may be useful to display the driver flavor (Adrenalin, Mesa, AMDVLK, AMDGPU-Pro)

Ryujinx.Ava/AppHost.cs
```diff
@@ -180,6 +182,24 @@ private void SetRendererWindowSize(Size size)
...
+            {
+                case ("nvidia corporation"):
+                return (NvidiaVendor);
+
+                case ("ati technologies inc."):
+                case ("advanced micro devices, inc."):
+                return (AmdVendor);
+
+                default:
+                return _renderer.GetHardwareInfo().GpuVendor;
```
ExE-Boss, line 199: Can’t this just do:
```suggestion
                return vendor;
```

Ryujinx/Ui/RendererWidgetBase.cs
```diff
@@ -125,7 +126,21 @@ public RendererWidgetBase(InputManager inputManager, GraphicsDebugLevel glLogLev
...
+            {
+                case ("nvidia corporation"):
+                return (NvidiaVendor);
+
+                case ("ati technologies inc."):
+                case ("advanced micro devices, inc."):
+                return (AmdVendor);
+
+                default:
+                return Renderer.GetHardwareInfo().GpuVendor;
```
ExE-Boss, line 141:
```suggestion
                return vendor;
```

PR #3525, opened 2022-08-01
Fix geometry shader passthrough fallback being used when feature is supported

gdkchan:
#2518 introduced a regression on GPUs that supports geometry shader passthrough. The fallback was being used even if the feature is supported, that is invalid because you can't assign to passthrough inputs and causes compilation error on OpenGL. Might be causing issues on Vulkan too, but likely not as bad as OpenGL.

Fixes regression on Marvel Ultimate Alliance 3 on Maxwell and newer NVIDIA GPUs.
image


riperiperi (approved):
LGTM.

GamerzHell9137:
Fixes the regression on OpenGL, Vulkan plays same.

emmauss (approved):

Thog:
Rebased on top of https://github.com/Ryujinx/Ryujinx/commit/f92650fcffa69cc45899b73e14d3dc7316b05d44

PR #3526, opened 2022-08-01
SPIR-V: Initialize undefined variables with 0

riperiperi:
Changes undefined values on spir-v shaders (caused by phi nodes) to be initialized instead of truly undefined.

Fixes an issue with NVIDIA gpus seemingly not liking when a variable is _potentially_ undefined. Not sure about the details at the moment.

Fixes:
- Tilt shift blur effect in Link's Awakening (bottom of the screen)
- Potentially block flickering on newer NVIDIA gpus in Splatoon 2? Needs testing.

Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Spirv/CodeGenContext.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Spirv/CodeGenContext.cs
@@ -234,7 +234,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Spirv
                     IrOperandType.Constant => GetConstant(type, operand),
                     IrOperandType.ConstantBuffer => GetConstantBuffer(type, operand),
                     IrOperandType.LocalVariable => GetLocal(type, operand),
-                    IrOperandType.Undefined => Undef(GetType(type)),
+                    IrOperandType.Undefined => Constant(GetType(type), operand.Value),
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Shader/CodeGen/Spirv/CodeGenContext.cs
```diff
@@ -234,7 +234,7 @@ public Instruction Get(AggregateType type, IAstNode node)
                     IrOperandType.Constant => GetConstant(type, operand),
                     IrOperandType.ConstantBuffer => GetConstantBuffer(type, operand),
                     IrOperandType.LocalVariable => GetLocal(type, operand),
-                    IrOperandType.Undefined => Undef(GetType(type)),
+                    IrOperandType.Undefined => Constant(GetType(type), operand.Value),
```
gdkchan, line 237:
```suggestion
                    IrOperandType.Undefined => Constant(GetType(type), 0),
```
`Undef` does not really carry any value, so `Value` will always be 0 anyway.

Thog (approved):

gdkchan (approved):
Thanks, seems reasonable, I think it can generate slightly worse code since it explicit initializes the variable to 0, where presumably with 'Undef' it wouldn't need to do so.

PR #3540, opened 2022-08-03
Fix Multithreaded Compilation of Shader Cache on OpenGL

riperiperi:
This was broken by the Vulkan changes - OpenGL was building host caches at boot on one thread, which is very notably slower than when it is multithreaded.

This was caused by trying to get the program binary immediately after compilation started, which blocks. Now it does it after compilation has completed.


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

Thog (approved):

PR #3558, opened 2022-08-07
Ava procon fix

Shahil-Ayato:
Fixes the how the pro controller image looks **on Avalonia** 

Note this shouldn't be merged yet until Avalonia is fully done and set as the default UI in master, as this new image doesn't work with the GTK UI that well.

Marking as a draft until then

Before:
image
After:
Screenshot 2022-08-07 025826


gdkchan:
The L and R buttons text is misaligned on the "After" image, and the "Before" image looks incorrect (I don't think it ever looked like this?). Also why would the change *need* to be done on GTK? The two UIs don't really need to match.

It would be also nice to know what exactly it "fixes" since I did not notice anything wrong with it before.

Squall-Leonhart:
before image is how it looks on avalonia, right down to the aliased lines.

image

TSRBerry:
Superseded by #4990

PR #3569, opened 2022-08-11
Avalonia - Add Polish Translation

LunaDook:
Everything here should be good to go, we gave it a once over and will double check tomorrow for anything we didn't catch before submitting.


gdkchan (approved):
Can't speak about the quality of the translation since I don't speak Polish, but otherwise lgtm.

LunaDook:
Me and @MoonbowJelly noticed a few issues we didn't catch now that the artifacts are up and we can see some of the changes more visually (providing context that we missed that changes the words to be used or declension, mostly), please do not merge this yet. We'll push fixes shortly.

LunaDook:
We noted that we thought we might have missed a line to do with the about dialogue box but it seems this line is missing from the locales file?
image

LunaDook:
Also noted that in the options drop-down that the last character of the translated string is getting slightly cut off. It's supposed to be an m.
image

LunaDook:
Alright, other then that one window title that seems untranslatable at the moment everything should now be fixed and good to go.

AcK77 (approved):

PR #1523, opened 2020-09-01
Improve Controller Applet UX

mageven:
In my last PR about this (#1453) I said-
> Note: While I'm sure there's going to be plenty of good suggestions and changes that can be made, I'm unable to touch more code (like wrestling with GTK) atm unless there's a bug/regression.

But, alas, I noticed quite a few users are confused about what the message dialog tells them to do. So, in order to improve the user experience, now controller applet directly launches a modified version of Input Settings.

Since there are more UI elements to play with, I added some missing fields and now the whole process is more visual and interactive.

Also, I noticed there was some talk about a crash in Splatoon 2 related to multiple configured controllers while playing with the new local wireless feature. This one crash is because controller applet was being silently spammed. The underlying reason was that the applet didn't consider `EnableSingleMode` and so thought the configuration was valid (it wasn't). This has also been fixed. Added a few more checks too while we're at it.

Controller applet now errs on the side of caution and is _less_ silent now to prevent issues like above cropping up.

I'll wait for testers to post screenshots of the different ways it looks with different games.

Also moved some logs to Debug which closes #1452

SettingsWindow was edited with glade, hence the line count.

**Note:** There are still some invalid configurations related to handheld config (bottom-most), so I suggest using it only when necessary. Docked is superior anyway. I'm not really satisfied with these changes but it'll have to do until a full HID RE.

Issue #3559, opened 2022-08-07
[ Xenoblade Chronicles 3 ] Char movements and audio broken


Jeffslash666:
### What's the issue you encountered?

As the title mention the movements of the main caracter are weird and audio is broken (missing), tried change audio backend and nothing changes.
### How can the issue be reproduced?

It happens instantly when i go in game.
### Log file

[ Logs files can be found under ``Logs`` folder in Ryujinx program folder. ]</br>
[ If you don't include a crash report in instances of crash related issues, we will ask you one to provide one. ]

### Environment?

- Ryujinx version:  1.1.208

- System Specs:
  - OS: Windows 10
  - CPU: i5 9600kf (stock)
  - GPU: Nvidia 1060 6gb
  - RAM: 16GB ddr4
- Applied Mods : No

### Additional context?

Sent you a video and the logs.

https://user-images.githubusercontent.com/83434355/183288077-14a91aa5-f273-47cb-bd71-eeaa52f2e398.mp4
[Ryujinx_1.1.208_2022-08-07_07-55-34.log](https://github.com/Ryujinx/Ryujinx/files/9276655/Ryujinx_1.1.208_2022-08-07_07-55-34.log)

GamerzHell9137:
I see nothing wrong with the character movement. Also looks like your F2 button is stuck and its muting and unmuting the emulator, the volume icon is flashing.

MutantAura:
For support please use our discord and not the issue tracker: https://discord.gg/ryujinx

PR #3571, opened 2022-08-11
ControllerApplet: Override player counts when SingleMode is set

mageven:
This strictly bandaid fix (as in it'll work until HID is properly reversed) stops ControllerApplet spam seen in Splatoon 2, Xenoblade Chronicles 3 and other single-player games which don't play well with multiple inputs configured/enabled. Instead the usual "Controller Applet" message dialog pops up requesting "exactly 1 player". 

The spam eventually caused crashes like `InvalidOperationException: Out of handles` or `InvalidMemoryAccess` in Splatoon 2 or very weird behavior such as bad/no audio, erratic character movement and eventual crashes(?) in Xenoblade 3. 

This oversight was originally fixed in the now abandoned #1523. Relevant excerpt-
> Also, I noticed there was some talk about a crash in Splatoon 2 related to multiple configured controllers while playing with the new local wireless feature. This one crash is because controller applet was being silently spammed. The underlying reason was that the applet didn't consider EnableSingleMode and so thought the configuration was valid (it wasn't). This has also been fixed.

**Note:** This change is only relevant if you see ControllerApplet spam in the logs and is ineffective otherwise. 
Apologies to the team for such long-term inconvenience. Better late than never I guess.

### DIFF
--- a/Ryujinx.HLE/HOS/Applets/Controller/ControllerApplet.cs
+++ b/Ryujinx.HLE/HOS/Applets/Controller/ControllerApplet.cs
@@ -72,8 +72,15 @@ namespace Ryujinx.HLE.HOS.Applets
 
             int playerMin = argHeader.PlayerCountMin;
             int playerMax = argHeader.PlayerCountMax;
+            bool singleMode = argHeader.EnableSingleMode != 0;
 
             Logger.Stub?.PrintStub(LogClass.ServiceHid, $"ControllerApplet Arg {playerMin} {playerMax} {argHeader.EnableTakeOverConnection} {argHeader.EnableSingleMode}");
+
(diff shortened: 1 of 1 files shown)


Thog (approved):

AcK77 (approved):

jduncanator:
It would be good to get issues opened for those crashes you observed if there aren't any already. We certainly shouldn't be leaking handles regardless, so that's probably something we should investigate and fix.

mageven:
Maybe I should've phrased it better. I didn't experience any of them (especially crashes) myself. I just mentioned the errors in the logs peppered around in Discord and the Issues page. Like #3559, that was closed recently.

PR #3573, opened 2022-08-11
fatal: Implement Service

### LINKED ISSUE #83: fatal:u is not implemented thus far.
Somehow, fatal:u is not implemented. I can understand that fatal:u may not be needed to be implemented, because, it's an actual crash on the target hardware system.

I am hoping that if fatal:u is implemented, there may be some way to see how it is triggering the fatal:u service on the Switch.

The hint I got for seeing a fatal:u is because I'm allocating too much memory on the Switch. But it shouldn't take up too much (25MB at most).

If you have a way to implement fatal:u, that's all I'm reporting about.

AcK77:
This PR adds a basic implementation of fatal service, guest processes call it when there is something wrong. But since we can already have all informations by debugging it's not really useful.
In any case, that's avoid an unimplemented service exception. Structs/Enum are based on Atmosphère source code.

After logs the error report, I call SvcBreak. Feedbacks are welcome on this, since some guests calls it right after fatal service so I can remove it if needed.

(Closes #83 ~finally~)

image

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Fatal/IService.cs
+++ b/Ryujinx.HLE/HOS/Services/Fatal/IService.cs
@@ -56,7 +56,7 @@ namespace Ryujinx.HLE.HOS.Services.Fatal
             errorReport.AppendLine("ErrorReport log:");
 
             errorReport.AppendLine($"\tTitleId: {context.Device.Application.TitleId:x16}");
-            errorReport.AppendLine($"\tPId: {pid}");
+            errorReport.AppendLine($"\tPid: {pid}");
             errorReport.AppendLine($"\tResultCode: {((int)resultCode & 0x1FF) + 2000}-{((int)resultCode >> 9) & 0x3FFF:d4}");
             errorReport.AppendLine($"\tFatalPolicy: {fatalPolicy}");
 
@@ -68,7 +68,7 @@ namespace Ryujinx.HLE.HOS.Services.Fatal
                 {
                     CpuContext64 cpuContext64 = MemoryMarshal.Cast<byte, CpuContext64>(cpuContext)[0];
 
-                    errorReport.AppendLine($"\tStartAddress: 0x{cpuContext64.StartAddress:16}");
+                    errorReport.AppendLine($"\tStartAddress: 0x{cpuContext64.StartAddress:x16}");
                     errorReport.AppendLine($"\tRegisterSetFlags: {cpuContext64.RegisterSetFlags}");
 
                     if (cpuContext64.StackTraceSize > 0)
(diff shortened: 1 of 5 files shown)


Thog (approved):

Ryujinx.HLE/HOS/Services/Fatal/IService.cs
```diff
@@ -1,8 +1,147 @@
...
+
+            if (cpuContext != null)
+            {
+                errorReport.AppendLine("CPU Context:");
+
+                if (context.Device.Application.TitleIs64Bit)
+                {
+                    CpuContext64 cpuContext64 = MemoryMarshal.Cast<byte, CpuContext64>(cpuContext)[0];
+
+                    errorReport.AppendLine($"\tStartAddress: 0x{cpuContext64.StartAddress:16}");
```
gdkchan, line 71:
```suggestion
                    errorReport.AppendLine($"\tStartAddress: 0x{cpuContext64.StartAddress:x16}");
```

Ryujinx.HLE/HOS/Services/Fatal/IService.cs
```diff
@@ -1,8 +1,147 @@
...
+
+        private ResultCode ThrowFatalWithCpuContextImpl(ServiceCtx context, ResultCode resultCode, ulong pid, FatalPolicy fatalPolicy, ReadOnlySpan<byte> cpuContext)
+        {
+            StringBuilder errorReport = new StringBuilder();
+
+            errorReport.AppendLine();
+            errorReport.AppendLine("ErrorReport log:");
+
+            errorReport.AppendLine($"\tTitleId: {context.Device.Application.TitleId:x16}");
+            errorReport.AppendLine($"\tPId: {pid}");
```
gdkchan, line 59:
```suggestion
            errorReport.AppendLine($"\tPid: {pid}");
```
Looks better imo.

Ryujinx.HLE/HOS/Services/Fatal/IService.cs
```diff
@@ -1,8 +1,147 @@
...
+                        {
+                            errorReport.AppendLine($"\t\t0x{cpuContext64.StackTrace[i]:x16}");
+                        }
+                    }
+
+                    errorReport.AppendLine("\tRegisters:");
+
+                    for (int i = 0; i < cpuContext64.X.Length; i++)
+                    {
+                        errorReport.AppendLine($"\t\tX[{i:d2}]:\t{cpuContext64.X[i]:x16}");
```
gdkchan, line 88: Maybe this and the others below should have a 0x prefix to indicate its a hex value?

gdkchan:
Where are the CPU context values supposed to come from? Seems odd that they are all 0s, and some of those CPU registers can only be read by kernel.

riperiperi (approved):

riperiperi:
Code looks OK, but it is weird that the example is passing a cpu context that is all 0s.

PR #3576, opened 2022-08-11
Vulkan: Add ETC2 texture formats

gdkchan:
Due to an oversight, those formats were never added to Vulkan format table. This change fixes this and also adds the remaining ETC2 formats (currently unused but it's there for when we need it in the future).

Unfortunately NVIDIA does not support ETC2 on Vulkan, so games using this format remains broken there until we add a software decompressor. Very few games uses it, mostly mobile ports.

On Intel and AMD GPUs that supports the extension, the games using the format should work now on Vulkan (Vegas Party, Radiation Island, etc).


Squall-Leonhart:
Its probably possible to use the interop extensions to unpack the ETC2 texture with opengl on hardware instead of a software decompressor.

Thog (approved):

gdkchan:
> Its probably possible to use the interop extensions to unpack the ETC2 texture with opengl on hardware instead of a software decompressor.

I think the question is whenever the GPU actually supports this format. It's possible that the OpenGL driver is doing software decompression.

Squall-Leonhart:
Nvidia Kepler and later support hardware ETC2
AMD Polaris and earlier do not.

AcK77 (approved):

PR #3583, opened 2022-08-14
Fix texture bindings using wrong sampler pool in some cases

### LINKED ISSUE #3561: Animal Crossing New Horizons: Flickering overworld regression

Animal Crossing New Horizons has a regression that causes the overworld to flickering frequently. The bug was introduced with the merge of the Vulkan PR and happens with both backends ever since.

There is no particular trigger for the bug, it randomly appears. Moving or standing still doesn't make a difference either. It feels like something moving through a shadow increases how frequently it appears, but that is just a feeling and not proven. It doesn't seem to happen in buildings, only outside on your island or ones you can travel to.

### Log file

Logs do not mention anything special when it happens, I also checked the performance tab in the task manager when it happens and nothing like spikes or drops on anything there either.

### Environment?

- Ryujinx version: 1.0.208</br>
- Game version: 2.0.5</br>
- System Specs:
  - OS: *Windows 10*
  - CPU: *AMD Ryzen 5 3600*
  - GPU: *NVIDIA GTX 950*
  - RAM: *16GB*
- Applied Mods : No

### Additional context?

The regression must have happened in the later stages of the Vulkan PR since I remember doing tests in an earlier version of the PR where the performance was worse but there was no flickering. 

Video showing a glipse of this on some bushes:
https://i.gyazo.com/92fe1d92eeed74cdf4cdd175880bfcaa.mp4

Video trying to show crossing shadows makes this bug appear more:
https://i.gyazo.com/55f416cc550425790d0bc0116bd26ca9.mp4

Video going over frame by frame in a case where this bug appeared on one of the ground textures:
https://i.gyazo.com/98a889daaab20fb5bff360fd0d2f164f.mp4

gdkchan:
This was an oversight during Vulkan rebase. Since #3487 added a sampler pool cache, it should use the sampler pool passed as argument rather than the one on `_samplerPool` field.

Thanks to Leischii for finding this issue.

Closes #3561

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -553,7 +553,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 specStateMatches &= specState.MatchesTexture(stage, index, descriptor);
 
-                Sampler sampler = _samplerPool?.Get(samplerId);
+                Sampler sampler = samplerPool?.Get(samplerId);
 
                 ITexture hostTexture = texture?.GetTargetTexture(bindingInfo.Target);
                 ISampler hostSampler = sampler?.GetHostSampler(texture);


Leischii:
Fixes the flickering regression on Animal Crossing New Horizons.

GamerzHell9137:
The flickering issue with Atelier Ryza when Vsync is unlocked is fixed too.

GamerzHell9137:
Fixes flickering issue in No More Heroes 3.

AcK77 (approved):

riperiperi (approved):

PR #3585, opened 2022-08-14
ARMeilleure: Hardware accelerate SHA256

merryhime:

PR #3587, opened 2022-08-14
PreAllocator: Check if instruction supports a Vex prefix in IsVexSameOperandDestSrc1

merryhime:
Don't assume all vector instructions have a vex encoding.

Not checking this causes invalid allocations on vector instructions that don't have a vex prefixed variant.


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

gdkchan:
A note for people writing the changelog, I don't believe this fixes issues with any of the existing instructions as all the instructions we use right now does support the VEX prefix, it's mostly for #3585 where the instructions does not support VEX prefix.

PR #3590, opened 2022-08-15
feat: package main release as appimage on Linux

cat-master21:
This PR aims to package Ryujinx as an appimage so some dependencies can be lowered on GNU Linux. Currently Yuzu already does this.

Some things that need to be done:

- [X] Create intial things
- [x] complete AppImageBuilder.yml
- [ ] Make sure it all works (basicly test it all)


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-dotnet@v1
+      - uses: actions/setup-dotnet@v2.1.0
```
Thog, line 29: This should be done in a separate PR.

Also I don't like precise version pinning here, please prefer v2 here for example.

.github/workflows/release.yml
```diff
@@ -77,21 +78,36 @@ jobs:
...
         shell: bash
+      
+      - uses: AppImageCrafters/build-appimage@master
+        with:
+          recipe: "../AppImageBuilder.yml"
+          args: "--skip-test"
 
       - name: Pushing new release
-        uses: ncipollo/release-action@v1
+        uses: ncipollo/release-action@v1.10.0
```
Thog, line 110: This should be done in a separate PR.

Also I don't like precise version pinning here, please prefer v1 here for example. (so no changes)

.github/workflows/release.yml
```diff
@@ -77,21 +78,36 @@ jobs:
...
 
+         # Most users won't need this and these are mainly for debug so not packaging as an appimage
           pushd publish_linux_sdl2_headless
           tar -czvf ../release_output/sdl2-ryujinx-headless-${{ steps.version_info.outputs.build_version }}-linux_x64.tar.gz publish
           popd
 
           pushd publish_linux_ava
           tar -czvf ../release_output/test-ava-ryujinx-${{ steps.version_info.outputs.build_version }}-linux_x64.tar.gz publish
           popd
+          cd ./publish
```
Thog, line 101: Should be separated from that step.

.github/workflows/release.yml
```diff
@@ -77,21 +78,36 @@ jobs:
...
         shell: bash
+      
+      - uses: AppImageCrafters/build-appimage@master
+        with:
+          recipe: "../AppImageBuilder.yml"
+          args: "--skip-test"
```
Thog, line 107: We have a reduced amount of allowed actions on the org.

I'm not a fan of 1) docker based actions 2) the fact it refer to the master branch.

Please prefer on host installation.


```suggestion
      - name: Packing AppImage (Linux)
        run: |
         sudo pip3 install appimage-builder==1.0.0
         appimage-builder --recipe AppImageBuilder.yml --skip-test
        shell: bash
```

.github/workflows/release.yml
```diff
@@ -77,21 +78,36 @@ jobs:
...
+         #sudo mv appimage-builder-x86_64.AppImage /usr/bin/appimage-builder
+         
+         #appimage-builder --recipe AppImageBuilder.yml --skip-test
+         mv /home/runner/work/Ryujinx/Ryujinx/Ryujinx-*-x86_64.AppImage publish
+         
+         # pushd publish_linux
+         # tar -czvf ../release_output/ryujinx-${{ steps.version_info.outputs.build_version }}-linux_x64.tar.gz publish
+         # popd
 
+         # Most users won't need this and these are mainly for debug so not packaging as an appimage
```
Thog, line 93: This **shouldn't be commented**.

This will remain the main way to install Ryujinx on Linux.

AppImageBuilder.yml
```diff
@@ -0,0 +1,55 @@
...
+    exec: usr/bin/ryujinx
+    
+    # Set the application main executable as argument. Use '$@' to forward CLI parameters
+    exec_args: "$@"
+
+  apt:
+    arch: [amd64, i386]
+    sources:
+      - sourceline: 'deb http://archive.ubuntu.com/ubuntu/ kinetic main restricted universe multiverse'
+        key_url: 'http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x871920D1991BC93C'
```
Thog, line 28: 1) We don't support i386, no need of this arch here
2) You shouldn't pack version as recent as Ubuntu 22.04 **as we support Ubuntu 18.04 as a minimal requirement**

AppImageBuilder.yml
```diff
@@ -0,0 +1,55 @@
...
+    #- 
+
+    exclude:
+    - humanity-icon-theme
+
+  after_runtime: |
+    # rm -f $APPDIR/.bundle.yml
+    
+  runtime:
+    arch: [ x86_64, i386 ]
```
Thog, line 42: Same comment as before i386 isn't supported.

AppImageBuilder.yml
```diff
@@ -0,0 +1,55 @@
...
+    #Make Directories and link them together
+
+AppDir:
+  path: ./AppDir
+
+  app_info:
+    id: org.ryujinx.Ryujinx
+    name: Ryujinx
+    icon: ryujinx-logo
+    version: 1.1
```
Thog, line 17: The version should be changed dynamically here.

AppImageBuilder.yml
```diff
@@ -0,0 +1,55 @@
+version: 1
```
Thog, line 1: This doesn't seems to account for the updater?

If you need to disable it, some changes need to happens on it.

Thog (changes requested):

.github/workflows/release.yml
cat-master21: Yeah I didn't want to use this either. Unfortunately appimages cannot be built in windows or WSL.

AppImageBuilder.yml
cat-master21: I have sed run in one of the previous steps along with all the other sed commands to change this on [line 49](https://github.com/Ryujinx/Ryujinx/blob/c828def3f0a120aabb55bf1705309ae6eaa7f86b/.github/workflows/release.yml#L49) in release.yml

.github/workflows/release.yml
cat-master21: Meaning that AppImages will be or just without appimages? Either way I will remove the comment.

cat-master21:
@Thog can I change the runner to `ubuntu-latest`? I wish to do this because WSL does not support building AppImages because of libfuse. I just don't know if you would not want to do that.

AppImageBuilder.yml
cat-master21: Also I don't think it is possible for it to be "dynamic" so it uses `sed` as it does to update other components as explained in previous comment.

AppImageBuilder.yml
cat-master21: I think this version number is used for what AppImage specification and/or appimage-builder version.

cat-master21: I think [line 21](https://github.com/cat-master21/Ryujinx/blob/0382356d43b94d73ccd711c5742a423f2da525be/AppImageBuilder.yml#L21) is what your looking for. But `sed` in release.yml updates this just like the rest of the components.

cat-master21:
Something cool that I think could be done is packaging a newer version of mesa considering older Linux distrobutions package old ones https://appimage-builder.readthedocs.io/en/latest/advanced/helpers.html#libgl

Thog:
> @Thog can I change the runner to `ubuntu-latest`? I wish to do this because WSL does not support building AppImages because of libfuse. I just don't know if you would not want to do that.

Impossible at present. building the Windows release require Windows because of broken packaging since .NET Core 3.1 I think (the icon of the executable is missing for example)

I'm not a huge fan of splitting every build in its own workflow because that imply more lost time for release.
Maybe the appimage could be generated in a job depending on the release one?

.github/workflows/release.yml
Thog: The tar compressed version will be the default and official way by default.
That's also to avoid breaking compat for older installation.

AppImageBuilder.yml
Thog: The real issue is that I don't think the updater in the current state will be able to update appimage builds.
We handle tar/zip manually

qurious-pixel:
@cat-master21 how did you handle the logs folder being created and crashing the appimage?

cat-master21:
@qurious-pixel I built the appimage manually and it failed on the header compatibility check no matter what I did. But I will try your suggestion👍.

cat-master21:
So @qurious-pixel mkappimage works and removes the issue with building without FUSE. I can't really debug the error because this is an issue with dotnet itself. Output: 
```
bash-5.1$ /home/pumkinking/Downloads/Ryujinx-0d889ad-x86_64.AppImage
Failure processing application bundle.
Bundle header version compatibility check failed. Header version: 18875693.69800449
A fatal error occured while processing application bundle
bash-5.1$
```
I think this may have to do with appimage-builder modifying magic bytes and patching to make it run from a specific PATH and what not I guess. For now all I want to include is `libx11.so` and possibly mesa soon enough.

qurious-pixel:
> I think this may have to do with appimage-builder modifying magic bytes and patching to make it run from a specific PATH and what not I guess. For now all I want to include is `libx11.so` and possibly mesa soon enough.

Make sure you check the exclude list <https://github.com/appimage-packages/AppImages/blob/master/excludelist>
Sometimes adding a lib will greatly expand the dependencies required for bundling. 

I keep the libs from the publish folder together, and everything else in `./usr/lib`

cat-master21:
S a issue with the appimage is 1. it can't boot because Ryujinx requires logs to be created in the Ryujinx folder, and 2. this has issues with Ryujinx's updater.

cat-master21:
I think I will close this because 1. Ryujinx doesn't any dependencies 2. can't really happen 3. does not seem to provide any benefits really.

PR #3595, opened 2022-08-16
Skipped over the last "Count" key explicitly

CloneDeath:
Skipped over the last "Count" key explicitly, instead of relying on an exception.

Previously, the code included all values, and just blindly swallowed exceptions.
Now, we just skip the last (Count) key.


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

riperiperi (approved):

PR #3596, opened 2022-08-16
OpenGL: Fix blit from non-multisample to multisample texture

### LINKED ISSUE #3592: Regression: Fate/EXTELLA missing graphics since 1.1.139

[Fate/EXTELLA](https://github.com/Ryujinx/Ryujinx-Games-List/issues/882) no longer renders most graphics. Only the UI and menu items are showing now.
image

Launch Fate/EXTELLA on any version of the emulator from 1.1.139 or newer.

### Log file

### Environment?

- Ryujinx version: 1.1.139 --> current build
- Game version: 1.0.0
- System Specs:
  - OS: Windows 11
  - CPU: Ryzen 7 5800X
  - GPU: NVIDIA RTX 3080 Ti
  - RAM: 32GB
- Applied Mods : No

### Additional context?

Graphics were last working on this game in build 1.1.138

gdkchan:
This fixes a regression from #3382. Since a copy dependency is created, it will try to copy both ways, including from a non-multisample texture to a multisample one if the former was modified. Due to the fact that the intermediate texture used for the copy had a multisample target, the blit operation was failing since if both textures are multisample, then the sample count must match. This change fixes it by doing the copy using a compute shader, where it reads or writes each sample from the source texture into the destination texture. This approach preserves all samples, so there should be no quality loss and the output should match the console. It also improves what I did originally on #3382, as this approach will preserve the samples on the larger image before the "resolve".

Fixes regression on Fate/EXTELLA.
image

Fixes #3592.


ExE-Boss:
This fixes <https://github.com/Ryujinx/Ryujinx/issues/3592>

EmulationFanatic:
The blur introduced with this approach is quite noticeable on Fate/EXTELLA.
The difference is more noticeable if you download the pictures and compare maximized.

Last time it was rendering:
2022-08-16 04_34_53-Ryujinx 1 1 138 - Fate_EXTELLA v1 0 0 (010053E002EA2000) (64-bit)
2022-08-16 04_37_43-Ryujinx 1 1 138 - Fate_EXTELLA v1 0 0 (010053E002EA2000) (64-bit)


With this PR:
2022-08-16 04_31_39-Ryujinx 1 1 0+c229558 - Fate_EXTELLA v1 0 0 (010053E002EA2000) (64-bit)
2022-08-16 04_33_09-Ryujinx 1 1 0+c229558 - Fate_EXTELLA v1 0 0 (010053E002EA2000) (64-bit)

gdkchan:
PR has been updated with a new approach that does not cause and quality loss and should produce accurate results.

riperiperi (approved):
LGTM, nice work on the new approach!

gdkchan:
It's worth noting that this will not work for depth/stencil texture copy. I don't think there's much we can do about those since afaik it's not possible to write to a multisample depth texture at all (outside of regular render and setting gl_FragDepth on the fragment shader). So maybe we should just not allow copy between multisample and non-multisample depth textures. I'm not aware of any games that needs this right now, so I guess we can worry about it later.

gdkchan:
I also found that binding a sRGB texture view as image does not work at all on NVIDIA OpenGL, even if you use a non-sRGB format with same BPP. So I introduced a workaround for this (creating a non-sRGB view). This is likely causing issues on some games like Kirby, so if we want to fix them on NVIDIA OpenGL we will need to use this workaround for all image bindings.

PR #3601, opened 2022-08-18
Change 'Purge PPTC Cache' label & tooltip to reflect function behavior

EmulationFanatic:
This PR modifies the "Purge PPTC Cache" function label to instead be "Queue PPTC Rebuild", and modifies the associated tooltip to be "Trigger PPTC to rebuild at boot time on the next game launch."
The reason for this change is that executing this function only deletes the PPTC .cache file, leaving the .info file intact. This means all functions cached so far are still intact in a list in the .info file, and the .cache file will be rebuilt from said file on next boot of the game. 
In lieu of changing the function to actually delete both the .cache and .info files, we should at least be descriptive about the function's true behavior.

Preview of GTK changes (also changed in Avalonia)
Before:
image

After:
image


EmulationFanatic:
> Deleting .info files is useful in some cases. For example updating Ryujinx from an very old version could cause PPTC related bugs.

Yes, I agree.
This PR does not change any functions, it just updates the description of what the function actually does when clicked.

Thog (approved):

AcK77 (approved):

Issue #3435, opened 2022-07-01
Regression: Flickering in Densha de Go introduced in 1.1.149


frituurpan:
### What's the issue you encountered?

Flickering scenery when playing Densha de GO!! Hashirou Yamanote-Sen.  
During gameplay the textures flickers.

### How can the issue be reproduced?

- Run the game.
- Run the route as show in the video (2nd route of 'story mode').
- It also happens on other routes, but not so much on some.

### Log file

I don't know which logs to include. Please let me know if I should add some.

### Environment?

- Ryujinx version: 1.1.149 and also confirmed in 1.1.160</br>
- The flickering does NOT occur in 1.1.148.
- Game version: 1.0.0</br>
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 7 5700X
  - GPU: 3080 12GB
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Additional info about your environment:</br>
I tried messing with the settings, but those do not influence the flickering. (docked, vsync, resolution, etc)

Video
See ending for flickering
https://user-images.githubusercontent.com/7461848/176957305-844f706f-46c3-43b0-ae85-73167f6b3f52.mp4

GamerzHell9137:
Can confirm that its happening too.

gdkchan:
Can you check if #3442 fixes it?

GamerzHell9137:
> Can you check if #3442 fixes it?

Nope, still happening.

frituurpan:
Tested: It is still happening in #3442

noigeaR:
The flickering is fixed for me with 1.1.201 and Vulkan backend.
It is still broken with OpenGL unfortunately.

atahanacar:
Flickering is still happening on 1.1.338, Linux, AMD RX 570 with RADV (amdvlk has completely black 3D graphics)

GamerzHell9137:
Can someone check the game again, seems like its fine now at least on my end.

frituurpan:
It was resolved for me when switching to vulkan when it released. 
I haven't checked opengl yet.

atahanacar:
The issue is still happening.

GamerzHell9137:
> It was resolved for me when switching to vulkan when it released. I haven't checked opengl yet.

Ohh so it's fixed? Guess this can be closed then.

riperiperi:
It’s possible that it never worked on AMD, so it wouldn’t be a regression in that case.

[...]

PR #3605, opened 2022-08-18
Rebind textures if format changes or they're buffer textures

riperiperi:
Fixes a regression from #3399 in the Mario Party Superstars spotlight minigame where an image would not rebind when its binding format was changed. Some other games may be affected.

Should also obsolete #3442, as this fixes it in a different way. Check the games that were meant to be fixed by that.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -47,6 +47,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             public int TextureHandle;
             public int SamplerHandle;
+            public Format ImageFormat;
             public int InvalidatedSequence;
             public Texture CachedTexture;
             public Sampler CachedSampler;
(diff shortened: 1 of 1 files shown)


GamerzHell9137:
Doen't fix Densha De Go 
https://github.com/Ryujinx/Ryujinx/issues/3435

riperiperi:
> Doen't fix Densha De Go that was caused by the PR #3435

Neither did the other PR.

GamerzHell9137:
> > Doen't fix Densha De Go that was caused by the PR #3435
> 
> Neither did the other PR. I don't think this particular oversight would be causing that issue.

Ohh, my bad. There's no known games that would get fixed by the other PR, it was just thought it would fix Densha which is why I tested.

gdkchan:
The issue I was trying to fix is a weird black vertex explosion on OpenGL that was happening for only 1 frame or so on Shin Megami Tensei V. I tested both this and my PR on it again, and I still get it on both. I'm not sure of what else to test as I'm not aware of any other issue that might be caused by the bindings change.

gdkchan (approved):
lgtm, thanks.

Thog (approved):

PR #3606, opened 2022-08-18
Removed redundant field initializers. 

CloneDeath:
Values were being assigned to their default values.
`bool` => `false`
`object` => `null`
`int` => `0`
etc...

Note: `NativeCtxStorage` is a `struct`, so it's already going to be initialized with the default constructor. It isn't `null`, don't worry :)

### DIFF
--- a/ARMeilleure/CodeGen/RegisterAllocators/CopyResolver.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/CopyResolver.cs
@@ -147,9 +147,9 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
             }
         }
 
-        private Queue<Operation> _fillQueue = null;
-        private Queue<Operation> _spillQueue = null;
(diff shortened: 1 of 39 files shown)


jduncanator:
I feel like being explicit is never a bad thing, especially when it's something like this, which usually comes down to personal preference.

I'd vote to leave these as is.

Thog:
> I feel like being explicit is never a bad thing, especially when it's something like this which usually comes down to personal preference. Explicitly declaring the default value also makes refactoring much easier in the future, it enables you to catch errors that would otherwise go unnoticed.
> 
> For example, changing a class to a struct, by removing all of the null assignments, the default declaration meaning will change, without any sign, and introduce behavioural difference. By explicitly defining the expected value, when you refactor, you'll get warnings/errors in places that need attention.
> 
> I'd vote to leave these as is.

I do believe we need to leave it as it, closing

PR #3607, opened 2022-08-19
Check if game directories have been updated before refreshing GUI

MutantAura:
Previously no matter what setting was changed the GUI directory page would refresh and reload the entire game list. For larger libraries this can take an extremely long time and can make changing any settings quite painful for those users. 

This change adds a new variable that tracks if the game directories have been modified and will only force a refresh on save/apply if there have actually been valid changes made within the functions.


EmulationFanatic:
This is a most welcome change! Tested for a few minutes and it seems to refresh when (and only when) needed.

riperiperi (approved):
Thank you 👼

AcK77 (approved):

Issue #3604, opened 2022-08-18
Regression: Blossom Tales II displays rainbow/mosaic pixels under NPC characters since 1.1.107


EmulationFanatic:
### What's the issue you encountered?

As reported by Discord user kakasita#0149, Blossom Tales 2 displays rainbow/mosaic pixels under NPC characters since version 1.1.107 (PR #3302).

1.1.106:
![image](https://user-images.githubusercontent.com/62343878/185471376-dcb98ccc-c0ff-4fb4-a455-0238426ea9e5.png)

1.1.107:
![image](https://user-images.githubusercontent.com/62343878/185471674-2610e5aa-05fb-4b38-ac21-522ae1fd90ea.png)

1.1.222 (Vulkan):
![image](https://user-images.githubusercontent.com/62343878/185472291-04a0c7fe-3cfd-412d-acc1-9b3662759e49.png)

### How can the issue be reproduced?

Launch Blossom Tales II on Ryujinx version 1.1.107 or newer. 1.1.106 and earlier are not affected.

### Environment?
Tested on OpenGL and Vulkan.
Ryzen 7 5800X
NVIDIA RTX 3080 Ti

F1ReB4LL:
Also broke Disjunction - https://github.com/Ryujinx/Ryujinx-Games-List/issues/3323

![ryujinx_disj11106](https://user-images.githubusercontent.com/39680140/185744249-db2b28bf-deac-49f8-9372-0ba2bc167505.png)
![ryujinx_disj11107](https://user-images.githubusercontent.com/39680140/185744254-6d82d792-0b85-44dc-b132-49688c440812.png)

F1ReB4LL:
Also broke Katana ZERO v1.0.5's loading screen text - https://github.com/Ryujinx/Ryujinx-Games-List/issues/1173

![ryujinx_ktz11106](https://user-images.githubusercontent.com/39680140/185745090-34eb5fe9-11f4-453b-9414-2578ce12d92b.png)
![ryujinx_ktz11107](https://user-images.githubusercontent.com/39680140/185745092-b2c33c4c-49d0-4ded-bcc1-a50bf2d19b66.png)

F1ReB4LL:
Also broke Aggelos - https://github.com/Ryujinx/Ryujinx-Games-List/issues/1946

![ryujinx_agg11106](https://user-images.githubusercontent.com/39680140/185746423-d0f0afce-5380-4c3d-854e-810384d0f171.png)
[...]

PR #3610, opened 2022-08-20
Fast path for Inline-to-Memory texture data transfers

gdkchan:
This adds a fast path for Inline-to-memory texture data transfers, similar to what's done on DMA engine. This one has the difference that it can transfer to the sub-region of a texture. Normally, since OpenGL games transfer part of the texture using 2D engine and the remaining using Inline-to-memory, we do a GPU side blit, and then immediately after flush the texture to write the remaining data on CPU. This is not only very inefficient, but also a source of bugs. Due to the flush issue that I attempted to fix in #3481, it can actually break those textures.

This approach simply submits the remaining data to GPU directly if the texture is already in the cache, which does not require flush and is much more efficient, and also fixes some texture corruption on OpenGL games.

For example, on Digimon Cyber Sleuth.
Before:
image
After:
image

Still, I believe we should get the flush working properly because the fast path but not be taken in all cases (for example, when there are multiple matches).

Testing in welcome, should only affect OpenGL *games* (it affects both the Vulkan and OpenGL backend on the emulator).

### DIFF
--- a/Ryujinx.Graphics.GAL/ITexture.cs
+++ b/Ryujinx.Graphics.GAL/ITexture.cs
@@ -19,6 +19,7 @@ namespace Ryujinx.Graphics.GAL
 
         void SetData(ReadOnlySpan<byte> data);
         void SetData(ReadOnlySpan<byte> data, int layer, int level);
+        void SetData(ReadOnlySpan<byte> data, int layer, int level, Rectangle<int> region);
         void SetStorage(BufferRange buffer);
         void Release();
     }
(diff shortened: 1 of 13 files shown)


LukeWarnut:
Seems to fix some texture issues in GameMaker Studio games. Fixes the top-right icons in HoloCure when modded over Super Perils of Baking.

Before:
image

After:
image

LukeWarnut:
Fixes lines in River City Girls Zero's menu:
Before:
image

After:
image

GamerzHell9137:
Fixes the dotted lines under Case announcements in Professor Layton
Before:
image
After:
image

EmulationFanatic:
Fixes https://github.com/Ryujinx/Ryujinx/issues/3604
image

Thog (approved):

riperiperi (approved):
Nice. I wonder why this was happening in the first place.

F1ReB4LL:
Disjunction and Aggelos are still regressed since 1.1.107, apparently broken by the same PR https://github.com/Ryujinx/Ryujinx/pull/3302 as others.

Disjunction (1.1.106):
ryujinx_disj11106
Disjunction (1.1.107):
ryujinx_disj11107
Disjunction (1.1.233 OpenGL):
ryujinx_disj11233ogl
Disjunction (1.1.233 Vulkan):
ryujinx_disj11233vlk

Aggelos (1.1.106):
ryujinx_agg11106
Aggelos (1.1.107):
ryujinx_agg11107
Aggelos (1.1.233 OpenGL):
ryujinx_agg11233ogl
Aggelos (1.1.233 Vulkan):
ryujinx_agg11233vlk

PR #3614, opened 2022-08-21
Implement some 32-bit Thumb instructions

gdkchan:
This PR implements the following 32-bit Thumb instructions:
- LDM/STM.
- LDAEX/STLEX.
- LDR/STR (with register offset shifted by immediate).
- LDRD/STRD.

Fixes the following issues:
- CBZ/CBNZ implementation was using `GetIntOrZR` function rather than `GetIntA32`, which was causing asserts as the register size was 64-bit rather than 32-bit.
- `OpCodeT16BImmCmp` did not implement `IOpCode32BImm` interface which was causing the target block to be missed by the decoder, which later on caused exceptions.


Thog (approved):
lgtm

LDj3SNuD (approved):

riperiperi (approved):
LGTM!

PR #3618, opened 2022-08-22
Update PPTC dialog text to match label and tooltip

RKennedy9064:
This update is to have the dialog text match the text changes from #3601.

Before:
image

After:
image


EmulationFanatic:
Good catch - only thing I would maybe recommend is making it say "You are about to queue a PPTC rebuild on the next boot of:"

RKennedy9064:
@EmulationFanatic Sounds good, just update the text. Looks like my IDE also automatically formatted out some empty spaces. Let me know if you want me to revert that.

EmulationFanatic:
> @EmulationFanatic Sounds good, just update the text. Looks like my IDE also automatically formatted out some empty spaces. Let me know if you want me to revert that.

I would revert the extra space deletions, simply because they are outside the scope of this PR and could have unintended consequences.

RKennedy9064:
Okay I think I got the formatting back to how it was before this change.

EmulationFanatic (approved):
LGTM

Thog (approved):

PR #3630, opened 2022-08-25
bsd: Fix Poll writting in input buffer

Thog:
This is a very old oversight on our Poll implementation.
This worked so far reliably because games and homebrews pass the same buffer as input and output.
Also fix ``IsUnexpectedLinuxError``


AcK77 (approved):

gdkchan (approved):
Nice catch.

PR #3570, opened 2022-08-11
gpu: Improve Format Table

AcK77:
This is just a rebase with all requested changes fixed of https://github.com/Ryujinx/Ryujinx/pull/1890

Co-Authored-By: Bruno Macabeus <macabeus@users.noreply.github.com>

Closes https://github.com/Ryujinx/Ryujinx/issues/1889

PR #3631, opened 2022-08-26
Replace image format magic numbers with enums

### LINKED ISSUE #1889: Given a meaningful name for texture format values
## Feature Request

On [`FormatTable.cs` file](https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Graphics.Gpu/Image/FormatTable.cs#L13-L111) there are a lot of magic numbers on the directory `_textureFormats`.
We should refac it into a meaningful name using the names from [`g80_defs.xml`](https://github.com/envytools/envytools/blob/master/rnndb/g80_defs.xml) and [`gm200_texture.xml`](https://github.com/envytools/envytools/blob/master/rnndb/graph/gm200_texture.xml).

It's a suggestion from @gdkchan ([message link](https://discord.com/channels/410208534861447168/410432399500115968/797557576727265330))

mageven:
Third time's the charm right? This PR uses the recently released NV official headers to fill up the enums. As a small bonus, also replaces the vertex attribute formats with another enum. Almost none of the work here is manual so there's very little chance of anything breaking. Manual whitespace formatting has been separated to its own commit (will be lost with the eventual auto-format).

As in the previous PR(?), base enums are exhaustive but a combination is generated only if it is used in the table. I noticed a couple of formats would have duplicate names if we assume the same type for all components, so the name is verbose if they don't match.

Closes #1889, Supersedes #3570


Ryujinx.Graphics.Gpu/Image/FormatTable.cs
```diff
@@ -8,195 +8,542 @@ namespace Ryujinx.Graphics.Gpu.Image
...
+            Astc2D10X5UnormSrgb              = Astc2D10X5        | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4956
+            Astc2D10X6UnormSrgb              = Astc2D10X6        | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4957
+            Astc2D10X8UnormSrgb              = Astc2D10X8        | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4953
+            Astc2D10X10UnormSrgb             = Astc2D10X10       | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4945
+            Astc2D12X10UnormSrgb             = Astc2D12X10       | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4954
+            Astc2D12X12UnormSrgb             = Astc2D12X12       | RUnorm | GUnorm | BUnorm | AUnorm | Srgb, // 0xa4946
+            A5B5G5R1Unorm                    = A5B5G5R1          | RUnorm | GUnorm | BUnorm | AUnorm,        // 0x24913
+        }
+
+        private enum VertexAttributeFormat : ulong
```
mageven, line 245: This could be `uint` but I kept the original dictionary definition.

AcK77 (approved):

Ryujinx.Graphics.Gpu/Image/FormatTable.cs
```diff
@@ -8,195 +8,542 @@ namespace Ryujinx.Graphics.Gpu.Image
...
+            Zf32X20V4S8Cov8R8V = 0x36,
+            X8Z24X16V8S8Cov4R12V = 0x37,
+            Zf32X16V8X8Cov4R12V = 0x38,
+            Zf32X16V8S8Cov4R12V = 0x39,
+            Z16 = 0x3a,
+            V8Z24Cov8R24V = 0x3b,
+            X8Z24X16V8S8Cov8R24V = 0x3c,
+            Zf32X16V8X8Cov8R24V = 0x3d,
+            Zf32X16V8S8Cov8R24V = 0x3e,
+            Astc2D4X4 = 0x40,
```
gdkchan, line 76: I think it would be better if X was lower case for those ASTC block dimensions.

Ryujinx.Graphics.Gpu/Image/FormatTable.cs
```diff
@@ -8,195 +8,542 @@ namespace Ryujinx.Graphics.Gpu.Image
...
+
+            RFloat = Float << 7,
+            GFloat = Float << 10,
+            BFloat = Float << 13,
+            AFloat = Float << 16,
+
+            Srgb = 0x1 << 19, // Custom encoding
+
+            // Combinations
+            R8Unorm                          = R8                | RUnorm | GUnorm | BUnorm | AUnorm,        // 0x2491d
```
gdkchan, line 139: Not sure if we should have the actual value as a comment here, because someone manually updating the table will need to include it too.

mageven: Thought it'd be easier at-a-glance. Do you think there's many such cases left?

gdkchan: Probably not. I have added all formats exposed on NVN (but that was in 2019, it's possible that they exposed more formats since then), and then there's the formats used by OpenGL and Vulkan that NVN doesn't expose that we have been adding manually.

gdkchan: Let's just keep the comment with the hex value for now. We don't need to update this frequently so it's probably not going to be an issue.

Ryujinx.Graphics.Gpu/Image/FormatTable.cs
mageven: I've changed it to `uint`. Function accessing it had `uint` anyway.

gdkchan (approved):
lgtm, thanks!

PR #3633, opened 2022-08-27
Bsd: Fix ArgumentOutOfRangeException in SetSocketOption

TSRBerry:
This PR adds a length check for `optionValue` before trying to read bytes that could possibly be out of range for this array.

Adding this check allows Minecraft to boot and play properly when "Enable guest internet access" is left unchecked.
Enabling that option will still cause the game to hang on boot and even disabled it will still hang when trying to stop emulation.
Although I have no idea why that happens, my blind guess would be that this is related to blocking sockets somehow?

Related issue: #3236

---

Also I want to add that I'm a bit confused about why these changes fix this issue.
When logging the values I noticed that nothing should have fulfilled [this if-condition](https://github.com/Ryujinx/Ryujinx/blob/6922862db8677fb8067a3e35d2433b1f12f8329c/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs#L326), since the option `BsdSocketOption.SoLinger` was never logged. 

Here is my log file: [Ryujinx_1.0.0-dirty_2022-08-27_02-04-36.log](https://github.com/Ryujinx/Ryujinx/files/9437001/Ryujinx_1.0.0-dirty_2022-08-27_02-04-36.log)

As you can see, the option that fulfills the if-condition in this case is `BsdSocketOption.TcpKeepInit`.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
@@ -325,7 +325,12 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
                 if (option == BsdSocketOption.SoLinger)
                 {
-                    int value2 = MemoryMarshal.Read<int>(optionValue[4..]);
+                    int value2 = 0;
+                    
+                    if (optionValue.Length >= 8)
+                    {
+                        value2 = MemoryMarshal.Read<int>(optionValue[4..]);
+                    }
 
                     Socket.SetSocketOption(level, SocketOptionName.Linger, new LingerOption(value != 0, value2));
                 }
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
@@ -323,10 +323,10 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
                 int value = optionValue.Length >= 4 ? MemoryMarshal.Read<int>(optionValue) : MemoryMarshal.Read<byte>(optionValue);
 
-                if (option == BsdSocketOption.SoLinger)
+                if (level == SocketOptionLevel.Socket && option == BsdSocketOption.SoLinger)
                 {
                     int value2 = 0;
-                    
+
                     if (optionValue.Length >= 8)
                     {
                         value2 = MemoryMarshal.Read<int>(optionValue[4..]);


riperiperi:
> When logging the values I noticed that nothing should have fulfilled [this if-condition](https://github.com/Ryujinx/Ryujinx/blob/6922862db8677fb8067a3e35d2433b1f12f8329c/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs#L326), since the option BsdSocketOption.SoLinger was never logged.

Look at the values in the enum `BsdSocketOption`.

`SoLinger = 0x80,`
`TcpKeepInit = 128,`

Option IDs are only unique within the same socket option level. It's very bad for debugging, and also for this check since it really should not be triggering this path if `level` is not socket. Unfortunately for us that's how the values work out. 🥴

TSRBerry:
> Option IDs are only unique within the same socket option level. It's very bad for debugging, and also for this check since it really should not be triggering this path if level is not socket.

Doesn't my log file say `level` is socket when it's triggering that path? If not, should I instead add a check for `level`?

Theslees:
```00:00:00.605 |E| Application : Unhandled exception caught: System.NullReferenceException: Object reference not set to an instance of an object.
   at Ryujinx.Ui.MainWindow..ctor(Builder builder) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Ui/MainWindow.cs:line 153
   at Ryujinx.Ui.MainWindow..ctor() in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Ui/MainWindow.cs:line 143
   at Ryujinx.Program.Main(String[] args) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Program.cs:line 195
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Ryujinx.Ui.MainWindow..ctor(Builder builder) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Ui/MainWindow.cs:line 153
   at Ryujinx.Ui.MainWindow..ctor() in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Ui/MainWindow.cs:line 143
   at Ryujinx.Program.Main(String[] args) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx/Program.cs:line 195
Aborted (core dumped)
```

[Ryujinx_1.1.0+dfed9b0_2022-08-26_22-37-45.log](https://github.com/Ryujinx/Ryujinx/files/9437187/Ryujinx_1.1.0%2Bdfed9b0_2022-08-26_22-37-45.log)

Attempted to run on arch linux via `./Ryujinx`

TSRBerry:
> Option IDs are only unique within the same socket option level. It's very bad for debugging, and also for this check since it really should not be triggering this path if `level` is not socket.

Nvm, I understand what you mean now.

Wouldn't it be better to use multiple enums then?

riperiperi:
No, this check (and any other checks for the option type) just needs to also check the level.

AcK77 (approved):

Thog (approved):

Issue #3509, opened 2022-07-31
Run on linux fails


albfan:
### What's the issue you encountered?

.Net do not find libx11 on linux

### How can the issue be reproduced?

running ryujinx

### Log file

```
     39478:	find library=libX11.so [0]; searching
     39478:	 search cache=/etc/ld.so.cache
     39478:	 search path=/lib64:/usr/lib64		(system search path)
     39478:	  trying file=/lib64/libX11.so
     39478:	  trying file=/usr/lib64/libX11.so
     39478:	
     39478:	find library=liblibX11.so [0]; searching
     39478:	 search cache=/etc/ld.so.cache
     39478:	 search path=/lib64:/usr/lib64		(system search path)
     39478:	  trying file=/lib64/liblibX11.so
     39478:	  trying file=/usr/lib64/liblibX11.so
     39478:	
     39478:	find library=libX11 [0]; searching
     39478:	 search cache=/etc/ld.so.cache
     39478:	 search path=/lib64:/usr/lib64		(system search path)
     39478:	  trying file=/lib64/libX11
     39478:	  trying file=/usr/lib64/libX11
     39478:	
     39478:	find library=liblibX11 [0]; searching
     39478:	 search cache=/etc/ld.so.cache
     39478:	 search path=/lib64:/usr/lib64		(system search path)
     39478:	  trying file=/lib64/liblibX11
     39478:	  trying file=/usr/lib64/liblibX11
     39478:	
Unhandled exception. 00:00:00.006 |W| Application GetWindowScaleFactor: Couldn't determine monitor DPI: Wayland not yet supported
System.DllNotFoundException: Unable to load shared library 'libX11' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblibX11: cannot open shared object file: No such file or directory
   at Ryujinx.Program.XInitThreads()
   at Ryujinx.Program.Main(String[] args) in D:\a\Ryujinx\Ryujinx\Ryujinx\Program.cs:line 104
Abortado (`core' generado)
```

### Environment?

- Ryujinx version: master</br>
- System Specs:
  - OS: Fedora 36
  - CPU: i7
  - GPU: Nvidia
  - RAM: 32GB
- Applied Mods : No
- 
### Additional context?

Can be fixed with:

    sudo ln -s /usr/lib64/libX11.so.6 /usr/lib64/libX11.so

Additional info about your environment:</br>

[...]

PR #3638, opened 2022-08-27
fix: package libX11.so in Linux builds

cat-master21:
I found out that by copying the library to Ryujinx's folder, you do not need libX11-dev/libX11-devel GNU Linux platforms.
You can try this by copying `usr/lib/x86_64-linux-gnu/libX11.so.6` or `usr/lib64/libX11.so.6` to Ryujinx's executable folder and it works.
This PR bundles [libX11 from Ubuntu Bionic](https://packages.ubuntu.com/bionic/libx11-6). I wanted originally to link it however libX11's place varies per linux distro however.

Closes: #3509

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,6 +75,28 @@ jobs:
           dotnet publish -c Release -r linux-x64 -o ./publish_linux_sdl2_headless/publish /p:Version="${{ steps.version_info.outputs.build_version }}" /p:SourceRevisionId="${{ steps.version_info.outputs.git_short_hash }}" /p:DebugType=embedded Ryujinx.Headless.SDL2 --self-contained
           dotnet publish -c Release -r linux-x64 -o ./publish_linux_ava/publish /p:Version="${{ steps.version_info.outputs.build_version }}" /p:SourceRevisionId="${{ steps.version_info.outputs.git_short_hash }}" /p:DebugType=embedded Ryujinx.Ava --self-contained
 
+      - name: Package libX11.so in Linux builds
+        run: |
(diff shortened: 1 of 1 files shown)


jduncanator:
This is definitely not the correct fix. This will break Ryujinx on any distribution that distributes a different version of X11, or uses a different/incompatible libc version.

Like every other application on Linux, we expect the user to install the required dependencies as part of installing Linux. The user should be installing the correctly built x11 from their distro repo.

cat-master21:
Well that's why I was thinking of symlinking it from the user's libX11 however on Fedora this is in `/usr/lib64` while Ubuntu has it in `usr/lib/x86_64-linux-gnu` oddly.

cat-master21:
I'm thinking maybe a script that would link it to Ryujin's folder if found perhaps.

cat-master21:
@jduncanator how about a static lib then. the libX11 from Ubuntu worked for me but I think a static lib should maintain distro compatability.

jduncanator:
A static lib would still require us to build and maintain the dependency, and make sure we're back-porting security fixes. It would also still have issues with distro that have older or newer versions of X11 installed that is incompatible with the specific X11 version we would ship. That's not a responsibility we realistically want to take onboard, when the user can just get the library from their distro repository.

What's the problem you're actually trying to solve here? It seems like you're trying to avoid the user installing a single dependency, yet we have many other dependencies on Linux the user needs to have installed. What makes X11 special that we should be redistributing it?

cat-master21:
Well I wanted to make sure Ryujinx is dependcy free mostly and solving this would mean that users would not have to install libX11. Also FYI if X11 has incompatable versions, why does appimage bundle libX11 on purpose?

Issue #2914, opened 2021-12-07
Tokyo Mirage Sessions #FE: Desyncs in scripted cutscenes


Oschowa:
### What's the issue you encountered?

In many scripted cutscenes in the game Tokyo Mirage Sessions FE Encore [0100A9400C9C2000] there are desyncs between the character motion, the subtitles and the audio occurring. I tried all 3 Memory Manager Modes, but they don't affect the issue.
  

### How can the issue be reproduced?

Load the following savegame and walk a few step forward, the following cutscene will desync in obvious ways.

[TokyoMirageSave.zip](https://github.com/Ryujinx/Ryujinx/files/7668163/TokyoMirageSave.zip)
  
### Log file

[Ryujinx_1.0.7116_2021-12-07_13-01-13.log](https://github.com/Ryujinx/Ryujinx/files/7668170/Ryujinx_1.0.7116_2021-12-07_13-01-13.log)
  

### Environment?

- Ryujinx version: 1.0.7116</br>
- Game version:1.0.0</br>
- System Specs:
  - OS: *Arch LInux*
  - CPU: *Ryzen 3700x*
  - GPU: *RX 6700 XT (Mesa/radeonsi git)*
  - RAM: *32GB*
- Applied Mods : No

EmulationFanatic:
Resolved by https://github.com/Ryujinx/Ryujinx/pull/3642

PR #3642, opened 2022-08-28
Change vsync signal to happen at 60hz, regardless of swap interval

riperiperi:
The vsync signal was signalling out at the swap interval, rather than every 16.667ms. Most games don't care about this, but there are a select few that do read the signal. This PR changes it to signal at 60hz regardless, and now swap interval is handled by setting a delay to the next frame when a present happens.

Fixes cutscene issues in [Tokyo Mirage Sessions #FE](https://github.com/Ryujinx/Ryujinx/issues/2914). The events within the cutscene relied on the vsync timer being at 60hz, whereas the animations relied on the actual presentation rate being 30hz. Having both run at the same speed caused interesting issues, but mostly just the voice lines playing at increasingly long delays.

This might also fix some weird game speed issues in The Legend of Zelda: Link's Awakening, and Breath of the Wild.

Games that run at 30fps should be tested to make sure that nothing breaks.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
+++ b/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
@@ -328,7 +328,7 @@ namespace Ryujinx.HLE.HOS.Services.SurfaceFlinger
                             Compose();
 
                             // When a frame is presented, delay the next one by its swap interval value.
-                            _swapIntervalDelay = _swapInterval - 1;
+                            _swapIntervalDelay = Math.Max(0, _swapInterval - 1);
                         }
 
                         _device.System?.SignalVsync();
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -91,7 +92,7 @@ private void UpdateSwapInterval(int swapInterval)
             }
             else
             {
-                _ticksPerFrame = Stopwatch.Frequency / (TargetFps / _swapInterval);
+                _ticksPerFrame = Stopwatch.Frequency / (TargetFps);
```
gdkchan, line 95:
```suggestion
                _ticksPerFrame = Stopwatch.Frequency / TargetFps;
```

Thog (approved):

gdkchan (approved):
lgtm, thanks.

Squall-Leonhart:
Fixes Skyward Sword performance weirdness with opengl on 50hz LCD tv's.
VK was never affected.

PR #3643, opened 2022-08-29
Avalonia - Misc changes to UX

MutantAura:
A few changes bundled together here as they all somewhat overlapped.

- Settings navbar is now not compact which, at least to me, seems far nicer to use and falls more in-line with new WinUI tools. Some borders have been added to make it less floaty too.
image

- All text now wraps with an overflow globally as this removes the need to do this for individual blocks when changing locales etc. Some are quite long and overall this prioritises the information always being displayed which is (again in my opinion) usually more important than forcing a static UI. (see navbar again)
image

- Right stick control border is now aligned properly, profile buttons now match refresh size (see above image).

- Controller ID has been removed in favour of a numbering system for "same" controller types. ID is a fairly meaningless string for 99% of users and realistically added nothing to the UX. It is now printed in the log whenever the LoadDevices() method is called instead so users who still need this information can get it. Multiple controllers of the same type will now number themselves from 0 upwards in the order they were connected in.
image
image

- Volume readout on the bottom widget has been re-aligned. Border no longer shows when muted.

### DIFF
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -879,7 +879,7 @@ namespace Ryujinx.Ava
 
             StatusUpdatedEvent?.Invoke(this, new StatusUpdatedEventArgs(
                 Device.EnableDeviceVsync,
-                Device.GetVolume(),
+                LocaleManager.Instance["VolumeShort"] + $": {(int)(Device.GetVolume() * 100)}%",
                 Renderer.IsVulkan ? "Vulkan" : "OpenGL",
                 dockedMode,
                 ConfigurationState.Instance.Graphics.AspectRatio.Value.ToText(),
--- a/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -588,5 +588,6 @@
   "SettingsTabGraphicsPreferredGpuTooltip": "Select the graphics card that will be used with the Vulkan graphics backend.\n\nDoes not affect the GPU that OpenGL will use.\n\nSet to the GPU flagged as \"dGPU\" if unsure. If there isn't one, leave untouched.",
   "SettingsAppRequiredRestartMessage": "Ryujinx Restart Required",
   "SettingsGpuBackendRestartMessage": "Graphics Backend or Gpu settings have been modified. This will require a restart to be applied",
-  "SettingsGpuBackendRestartSubMessage": "Do you want to restart now?"
+  "SettingsGpuBackendRestartSubMessage": "Do you want to restart now?",
+  "VolumeShort": "Vol"
 }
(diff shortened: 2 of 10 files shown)


gdkchan:
Features & Enhancement is not aligned like the other headings, on your first screenshot.

gdkchan:
image
Profile load/save/delete buttons are too wide. I think they should have the same size as the refresh button.

Squall-Leonhart:
> Features & Enhancement is not aligned like the other headings, on your first screenshot.

neither is the line above it.

MutantAura:
I dont think they ever have been 👀:
image
(master image ^)

 I'll roll all of the above into this change then:

MutantAura:
Should all be resolved? I also took the liberty of adding a couple of classes for textblock (headers) and borders (settings). 
The results look something like this: 
image

emmauss (approved):

emmauss:
> Window aspect ratio: 16:10 OS: Windows 11
> 
> I don't know if this affects other aspect ratios but all other Avalonia builds I have tried have a weird scaling issue in full screen and tabs don't work when clicked
> 
> image

Does this occur only on vulkan?

Ackliez:
H

> > Window aspect ratio: 16:10 OS: Windows 11
> > I don't know if this affects other aspect ratios but all other Avalonia builds I have tried have a weird scaling issue in full screen and tabs don't work when clicked
> > image
> 
> Does this occur only on vulkan?

Haven't checked

Ackliez:
> H
> 
> > > Window aspect ratio: 16:10 OS: Windows 11
> > > I don't know if this affects other aspect ratios but all other Avalonia builds I have tried have a weird scaling issue in full screen and tabs don't work when clicked
> > > image
> > 
> > 
> > Does this occur only on vulkan?
> 
> Haven't checked

Yes just on vulkan

merryhime:
Nice PR. As requested previously, here's a screenshot of the volume control in the status bar on 8dadecb6b4498b1f1d14774a24de2ae167fa9cfc, on my Linux machine:
image

MutantAura:
> Nice PR. As requested previously, here's a screenshot of the volume control in the status bar on [8dadecb](https://github.com/Ryujinx/Ryujinx/commit/8dadecb6b4498b1f1d14774a24de2ae167fa9cfc), on my Linux machine: image

Alright I guess it's still somewhat misaligned outside of Windows. Very weird why this varies per OS. I guess the icon itself isn't consistent? 
Would it be better for me to incorporate your changes here or to remove my edits to the volume bar and just merge your PR separately?

MutantAura:
Alright let's just go with your original solution from: #3698 then. It seems the most consistent across all platforms which is the aim here. 
I can cherry-pick that commit into this branch if that's cool with you?

gdkchan:
We might need to make the "Vol:" text localizable, I'm guessing that in some languages it is not called Volume.

riperiperi (approved):
Generally looks better here.

gdkchan:
Needs a rebase.

MutantAura:
Rebased

gdkchan (approved):
There's a few things I'd change, but I think it's good in general. Thanks for fixing those long standing issues on Avalonia.

PR #3646, opened 2022-08-30
account: Implement LoadNetworkServiceLicenseKindAsync

Thog:
This is needed to run Pokemon Legends Arceus 1.1.1 with guest internet enabled.

The game still get stuck at loading screen.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Account/Acc/IAsyncContext.cs
+++ b/Ryujinx.HLE/HOS/Services/Account/Acc/IAsyncContext.cs
@@ -7,18 +7,18 @@ namespace Ryujinx.HLE.HOS.Services.Account.Acc
 {
     class IAsyncContext : IpcService
     {
-        protected AsyncExecution _asyncExecution;
+        protected AsyncExecution AsyncExecution;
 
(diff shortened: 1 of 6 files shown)


AcK77 (approved):

Ryujinx.HLE/HOS/Services/Account/Acc/IAsyncContext.cs
```diff
@@ -7,7 +7,7 @@ namespace Ryujinx.HLE.HOS.Services.Account.Acc
 {
     class IAsyncContext : IpcService
     {
-        AsyncExecution _asyncExecution;
+        protected AsyncExecution _asyncExecution;
```
gdkchan, line 10: I think since that was changed to protected, it should be renamed to `AsyncExecution`? From what I recall, that's how we have been naming protected fields.

Thog: I think you are right, going to fix that now 👍

gdkchan (approved):
lgtm, thanks.

PR #3650, opened 2022-08-31
Fix mapping leaks caused by UnmapView not working on Linux

### LINKED ISSUE #3372: Multiple games crash upon launch when using Host memory management + Linux on 1.1.110 and above
## Bug Report

Repost of #3356 as per request from @MutantAura.

Shin Megami Tensei V crashes upon launch when running under Linux and with memory management set to Host or Host Unchecked.  
Setting memory management to Software allows the game to launch.

Additional game crashes reported by @HerrIrrtum:

- Triangle Strategy
- Bravely Default II
- Oceanhorn 2
- Tetris Effect Connected

Possibly related to #3349 

Ryujinx closes with a segfault or an "Out of memory" left in the terminal while launching a problematic game.

All games have been functional previously.

1. Run any recent Linux distro (Arch, Ubuntu 22.04, etc.)
2. Ensure memory management is set to "Host" or "Host (Unchecked)"
3. Launch a problematic game in Ryujinx (for example SMT V)
4. Ryujinx crashes and closes during game launch

### Log file

[ryujinx-smtv.log](https://github.com/Ryujinx/Ryujinx/files/8798860/ryujinx-smtv.log)

### Environment?

- Ryujinx version: 1.1.132
- Game version: 1.0.0, 1.0.1 and 1.0.2  
- System Specs:
  - OS: *Arch Linux x86_64*
  - CPU: *Ryzen 5 5600X*
  - GPU: *AMD Radeon RX 6700 XT*
  - RAM: *32GB*
- Applied Mods : No

### Additional context?

Tested on Arch Linux, Chimera OS and Ubuntu.  
Tested with Intel Iris Xe and UHD 605 graphics, as well as RX 6700 XT.  
All machines have 16GB ram or more.  
Same crash happens on all machines.

SMT V is the only game I have personally tested.

gdkchan:
This fixes an issue where UnmapView was failing on Linux because the flags combination being passed was invalid.
Needs to be tested on Linux, would be nice to know if it fixes #3372.

### DIFF
--- a/Ryujinx.Memory.Tests/Tests.cs
+++ b/Ryujinx.Memory.Tests/Tests.cs
@@ -92,5 +92,31 @@ namespace Ryujinx.Memory.Tests
                 }
             }
         }
+
+        [Test]
+        public void Test_AliasMapLeak()
+        {
+            if (OperatingSystem.IsMacOS())
+            {
+                // Memory aliasing tests fail on CI at the moment.
+                return;
+            }
+
+            ulong pageSize = 4096;
+            ulong size = 100000 * pageSize; // The mappings limit on Linux is usually around 65K, so let's make sure we are above that.
+
+            using MemoryBlock backing = new MemoryBlock(pageSize, MemoryAllocationFlags.Mirrorable);
+            using MemoryBlock toAlias = new MemoryBlock(size, MemoryAllocationFlags.Reserve | MemoryAllocationFlags.ViewCompatible);
+
+            for (ulong offset = 0; offset < size; offset += pageSize)
+            {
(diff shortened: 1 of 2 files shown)


gdkchan:
The test did not fail in CI which is odd, it only failed when I added a throw for mmap returning -1.

Thog (approved):
nice catch

riperiperi (approved):

MutantAura:
Unfortunately does not fix the out of memory errors in https://github.com/Ryujinx/Ryujinx/issues/3372 
SMTV: 
image

PR #3652, opened 2022-08-31
Bsd: Fix NullReferenceException in BsdSockAddr.FromIPEndPoint()

TSRBerry:
This PR adds another check to prevent games from crashing with guest internet access enabled.

@EmulationFanatic sent me a log of "Victor Vran Overkill Edition" crashing on boot because a `System.NullReferenceException` was thrown inside of `BsdSockAddr.FromIPEndPoint()`. 
Following the stack trace I found a missing check for `socket.RemoteEndPoint` to ensure it's not null.

With this change "Victor Vran Overkill Edition" boots with guest internet access enabled, although it will still deadlock shortly after, but that's a different issue.

Thanks to @EmulationFanatic for confirming my fix works so quickly!

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -568,7 +568,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
             LinuxError errno  = LinuxError.EBADF;
             ISocket    socket = _context.RetrieveSocket(socketFd);
 
-            if (socket != null)
+            if (socket != null && socket.RemoteEndPoint != null)
             {
                 errno = LinuxError.SUCCESS;
 
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -567,7 +567,6 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
             LinuxError errno  = LinuxError.EBADF;
             ISocket    socket = _context.RetrieveSocket(socketFd);
-            
             if (socket != null)
             {
                 errno = LinuxError.ENOTCONN;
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -568,7 +568,7 @@ public ResultCode GetPeerName(ServiceCtx context)
             LinuxError errno  = LinuxError.EBADF;
             ISocket    socket = _context.RetrieveSocket(socketFd);
 
-            if (socket != null)
```
Thog, line 571: I do believe that if ``RemoteEndPoint`` is null we should return ``ENOTCONN`` as per standard [getpeername(2)](https://man7.org/linux/man-pages/man2/getpeername.2.html)

Thog (changes requested):

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
TSRBerry: That makes sense, will add that shortly!

Thog (approved):

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -567,14 +567,19 @@ public ResultCode GetPeerName(ServiceCtx context)
 
             LinuxError errno  = LinuxError.EBADF;
             ISocket    socket = _context.RetrieveSocket(socketFd);
-
+
```
gdkchan, line 570:
```suggestion

```

riperiperi: Did it remove this empty line entirely? The next line is showing as 570 now.

TSRBerry: I just clicked on 'commit suggestion' since I'm currently not home, which makes everything involving git a little bit harder :D

AcK77 (approved):

PR #3659, opened 2022-09-01
bsd: Fix Poll(0) returning ETIMEDOUT instead of SUCCESS

Thog:
This was an oversight of the implementation.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -336,6 +336,12 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
                 context.Memory.Write(outputBufferPosition + (ulong)(i * Unsafe.SizeOf<PollEventData>()), events[i].Data);
             }
 
+            // In case of non blocking call timeout should not be returned.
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Nice catch.

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -336,6 +336,12 @@ bool IsUnexpectedLinuxError(LinuxError error)
                 context.Memory.Write(outputBufferPosition + (ulong)(i * Unsafe.SizeOf<PollEventData>()), events[i].Data);
             }
 
+            // In case of non blocking call timeout should not be returned.
```
gdkchan, line 339: It doesn't look like this is checking if the socket is blocking or not.

Thog: non blocking in poll case isn't the same as socket blocking

gdkchan: I guess a "non blocking call" is a call where the timeout is 0 then?

Thog: exactly yeah

PR #3663, opened 2022-09-02
ARMeilleure: Add initial support for AVX512(EVEX encoding)

Wunkolo:
The first sample use-case being a single-instruction alternative to the vector-`NOT` instruction by utilizing the [vpternlogd instruction](https://www.felixcloutier.com/x86/vpternlogd:vpternlogq).

Implements enough of the EVEX encoding features to utilize AVX512 instructions for a 128-bit register use-case.

Does not include:
* Embedded rounding controls
* Embedded exception controls

`Not_V_*` unit tests all pass on my `i9-11900k` and on my `i9-7900x`. Would be good to get some additional verifications on additional hardware or possibly utilizing the [Intel Software Development Emulator](https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html).

### DIFF
--- a/ARMeilleure/CodeGen/X86/Assembler.cs
+++ b/ARMeilleure/CodeGen/X86/Assembler.cs
@@ -1033,7 +1033,13 @@ namespace ARMeilleure.CodeGen.X86
 
             Debug.Assert(opCode != BadOp, "Invalid opcode value.");
 
-            if ((flags & InstructionFlags.Vex) != 0 && HardwareCapabilities.SupportsVexEncoding)
+            if ((flags & InstructionFlags.Evex) != 0 && HardwareCapabilities.SupportsEvexEncoding)
+            {
+                WriteEvexInst(dest, src1, src2, type, flags, opCode);
+
+                opCode &= 0xff;
+            }
+            else if ((flags & InstructionFlags.Vex) != 0 && HardwareCapabilities.SupportsVexEncoding)
             {
                 // In a vex encoding, only one prefix can be active at a time. The active prefix is encoded in the second byte using two bits.
 
@@ -1150,6 +1156,103 @@ namespace ARMeilleure.CodeGen.X86
                     }
                 }
             }
+        }
+
+        private void WriteEvexInst(
+            Operand dest,
+            Operand src1,
+            Operand src2,
+            OperandType type,
+            InstructionFlags flags,
+            int opCode,
+            bool broadcast = false,
+            int registerWidth = 128,
+            int maskRegisterIdx = 0,
+            bool zeroElements = false)
+        {
+            int destIdx = dest.GetRegister().Index;
+            int src1Idx = src1.GetRegister().Index;
+            int src2Idx = src2.GetRegister().Index;
+
+            WriteByte(0x62);
+
+            // P0
+            // Extend dest register
+            bool r = (destIdx & 8) == 0;
+            // Extend src register
+            bool x = (src1Idx & 16) == 0;
+            // Extend src register
+            bool b = (src1Idx & 8) == 0;
+            // Extend dest register
+            bool rp = (destIdx & 16) == 0;
+            // Escape code index
+            byte mm = 0b00;
+
+            switch ((ushort)(opCode >> 8))
+            {
+                case 0xf00: mm = 0b01; break;
+                case 0xf38: mm = 0b10; break;
+                case 0xf3a: mm = 0b11; break;
+
+                default: Debug.Assert(false, $"Failed to EVEX encode opcode 0x{opCode:X}."); break;
+            }
+
+            WriteByte(
+                (byte)(
+                    (r ? 0x80 : 0) |
+                    (x ? 0x40 : 0) |
+                    (b ? 0x20 : 0) |
+                    (rp ? 0x10 : 0) |
+                    mm));
+
+            // P1
+            // Specify 64-bit lane mode
+            bool w = Is64Bits(type);
+            // Src2 register index
+            byte vvvv = (byte)(src2Idx & 0b1111);
+            // Opcode prefix
+            byte pp = (flags & InstructionFlags.PrefixMask) switch
+            {
+                InstructionFlags.Prefix66 => 0b01,
+                InstructionFlags.PrefixF3 => 0b10,
+                InstructionFlags.PrefixF2 => 0b11,
+                _ => 0
+            };
+            WriteByte(
+                (byte)(
+                    (w ? 0x80 : 0) |
+                    (vvvv << 3) |
+                    0b100 |
+                    pp));
+
+            // P2
+            // Mask register determines what elements to zero, rather than what elements to merge
+            bool z = zeroElements;
+            // Specifies register-width
+            byte ll = 0b00;
+            switch (registerWidth)
+            {
+                case 128: ll = 0b00; break;
+                case 256: ll = 0b01; break;
+                case 512: ll = 0b10; break;
+
+                default: Debug.Assert(false, $"Invalid EVEX vector register width {registerWidth}."); break;
+            }
+            // Embedded broadcast in the case of a memory operand
+            bool bcast = broadcast;
+            // Extend src2 register
+            bool vp = (src2Idx & 16) == 0;
+            // Mask register index
+            Debug.Assert(maskRegisterIdx < 8, $"Invalid mask register index {maskRegisterIdx}.");
+            byte aaa = (byte)(maskRegisterIdx & 0b111);
+
+            WriteByte(
+                (byte)(
+                    (z ? 0x80 : 0) |
+                    (ll << 5) |
+                    (bcast ? 0x10 : 0) |
+                    (vp ? 8 : 0) |
+                    aaa));
         }
 
         private void WriteCompactInst(Operand operand, int opCode)
--- a/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
+++ b/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
@@ -29,6 +29,11 @@ namespace ARMeilleure.CodeGen.X86
                 FeatureInfo7Ecx = (FeatureFlags7Ecx)ecx7;
             }
 
+            Xcr0InfoEax = (Xcr0FlagsEax)GetXcr0Eax();
+        }
+
+        private static uint GetXcr0Eax()
+        {
             byte[] asmGetXcr0 = {
                 0x31, 0xc9, // xor ecx, ecx
                 0xf, 0x01, 0xd0, // xgetbv
@@ -43,7 +48,7 @@ namespace ARMeilleure.CodeGen.X86
 
             var fGetXcr0 = Marshal.GetDelegateForFunctionPointer<GetXcr0>(memGetXcr0.Pointer);
 
-            Xcr0InfoEax = (Xcr0FlagsEax)fGetXcr0();
+            return fGetXcr0();
         }
 
         [Flags]
--- a/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
+++ b/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
@@ -44,6 +44,7 @@ namespace ARMeilleure.CodeGen.X86
             Sse42 = 1 << 20,
             Popcnt = 1 << 23,
             Aes = 1 << 25,
+            Osxsave = 1 << 27,
             Avx = 1 << 28,
             F16c = 1 << 29
         }
@@ -80,12 +81,12 @@ namespace ARMeilleure.CodeGen.X86
         public static bool SupportsSse42 => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Sse42);
         public static bool SupportsPopcnt => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Popcnt);
         public static bool SupportsAesni => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Aes);
-        public static bool SupportsAvx => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Avx);
+        public static bool SupportsAvx => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Avx | FeatureFlags1Ecx.Osxsave);
         public static bool SupportsAvx2 => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx2) && SupportsAvx;
-        public static bool SupportsAvx512F => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512f);
-        public static bool SupportsAvx512Vl => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512vl);
-        public static bool SupportsAvx512Bw => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512bw);
-        public static bool SupportsAvx512Dq => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512dq);
+        public static bool SupportsAvx512F => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512f) && FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Osxsave);
+        public static bool SupportsAvx512Vl => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512vl) && SupportsAvx512F;
+        public static bool SupportsAvx512Bw => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512bw) && SupportsAvx512F;
+        public static bool SupportsAvx512Dq => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512dq) && SupportsAvx512F;
         public static bool SupportsF16c => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.F16c);
         public static bool SupportsSha => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Sha);
         public static bool SupportsGfni => FeatureInfo7Ecx.HasFlag(FeatureFlags7Ecx.Gfni);
--- a/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
+++ b/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
@@ -34,7 +34,8 @@ namespace ARMeilleure.CodeGen.X86
 
         private static uint GetXcr0Eax()
         {
-            ReadOnlySpan<byte> asmGetXcr0 = new byte[]{
+            ReadOnlySpan<byte> asmGetXcr0 = new byte[]
+            {
                 0x31, 0xc9, // xor ecx, ecx
                 0xf, 0x01, 0xd0, // xgetbv
                 0xc3, // ret
@@ -98,7 +99,7 @@ namespace ARMeilleure.CodeGen.X86
             YmmHi128 = 1 << 2,
             Opmask = 1 << 5,
             ZmmHi256 = 1 << 6,
-            Hi16Zmm = 1 << 7,
+            Hi16Zmm = 1 << 7
         }
 
         public static FeatureFlags1Edx FeatureInfo1Edx { get; }
--- a/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
+++ b/ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
@@ -1,10 +1,13 @@
+using Ryujinx.Memory;
 using System;
+using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics.X86;
 
 namespace ARMeilleure.CodeGen.X86
 {
     static class HardwareCapabilities
     {
+        private delegate uint GetXcr0();
         static HardwareCapabilities()
         {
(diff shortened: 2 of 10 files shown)


ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1033,7 +1033,90 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
             Debug.Assert(opCode != BadOp, "Invalid opcode value.");
 
-            if ((flags & InstructionFlags.Vex) != 0 && HardwareCapabilities.SupportsVexEncoding)
+            if ((flags & InstructionFlags.Evex) != 0 && HardwareCapabilities.SupportsEVexEncoding)
+            {
+                WriteByte(0x62);
+
+                // P0
+                // Extend dest register
+                bool R = !((dest.GetRegister().Index & 8) != 0);
```
gdkchan, line 1042: Local variable names should start in lower case. Also those checks would be simpler by doing `(x & mask) == 0` rather than `!((x & mask) != 0)`.

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1033,7 +1033,90 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+                switch ((ushort)(opCode >> 8))
+                {
+                    case 0xf00: mm = 0b01; break;
+                    case 0xf38: mm = 0b10; break;
+                    case 0xf3a: mm = 0b11; break;
+
+                    default: Debug.Assert(false, $"Failed to EVEX encode opcode 0x{opCode:X}."); break;
+                }
+
+                WriteByte(
```
gdkchan, line 1061: I think it could be better to move writing those 3 bytes into a separate function, it could also take enums for each one of those fields, which would avoid the need for some of those comments, as the enum would have a name, along with its contents.

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1033,7 +1033,90 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+                // Merge or Zero
+                bool z = false;
+                // Specifies register-width
+                // 00: 128 | 01: 256 | 10: 512
+                byte LL = 0b00;
+                // Embedded broadcast
+                bool b = false;
+                // Extend src2 register
+                bool Vp = !((src2.GetRegister().Index & 16) != 0);
+                // Mask register index(not utilized)
```
gdkchan, line 1104:
```suggestion
                // Mask register index (not utilized)
```

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -70,11 +74,13 @@ public enum FeatureFlags7Ebx
...
         public static bool SupportsAvx2 => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx2) && SupportsAvx;
+        public static bool SupportsAvx512Ortho => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512f | FeatureFlags7Ebx.Avx512vl | FeatureFlags7Ebx.Avx512dq);
         public static bool SupportsF16c => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.F16c);
         public static bool SupportsSha => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Sha);
 
         public static bool ForceLegacySse { get; set; }
 
-        public static bool SupportsVexEncoding => SupportsAvx && !ForceLegacySse;
+        public static bool SupportsVexEncoding => SupportsAvx512Ortho && SupportsAvx && !ForceLegacySse;
+        public static bool SupportsEVexEncoding => SupportsAvx512Ortho && !ForceLegacySse;
```
gdkchan, line 84:
```suggestion
        public static bool SupportsEvexEncoding => SupportsAvx512Ortho && !ForceLegacySse;
```
For consistency.

ARMeilleure/Instructions/InstEmitSimdLogical.cs
```diff
@@ -224,7 +224,22 @@ public static void Eor_V(ArmEmitterContext context)
 
         public static void Not_V(ArmEmitterContext context)
         {
-            if (Optimizations.UseSse2)
+            if (Optimizations.UseAvx512)
```
gdkchan, line 227: Don't we need one of those for each subset of AVX-512 that is supported? Since some CPUs are not going to support all of them, and we don't want to disable all AVX-512 optimizations if some of them are not available (not a problem right now since there's only one used, but I'd imagine it would be in the future).

Wunkolo: Maybe a better name for this is `SupportsAvx512Ortho`. Which is what I am naming the combination of `AVX512F+AVX512VL+AVX512DQ` under the hood which is the minimum amount of AVX512 extensions to be able to work with 128-bit registers _and_ interface with floats/doubles and 64-bit integers(Technically I would like to include `AVX512BW` as well to operate on 8-bit and 16-bit elements but didn't want the qualification of Orthongonal-AVX512). Currently AVX512 support is defined by AVX512-Ortho.

Maybe I can break up all the subsets and then have more higher-order groupings for things like `AVX512Ortho=AVX512F+AVX512VL`, `AVX512OrthoFloat=AVX512Ortho+AVX512DQ` and such similar to what I did in Dynarmic:
https://github.com/merryhime/dynarmic/blob/e49fee0ca11517a39d3da65e5dc09e4d0cc3d020/src/dynarmic/backend/x64/host_feature.h#L12-L44

Wunkolo: I broke up each of the subsets into their own flags within `HardwareCapabilities`, and added the more super-set groupings into `Optimizations`. 👍

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1152,6 +1158,99 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
             }
         }
 
+        private void WriteEvexInst(Operand dest, Operand src1, Operand src2, OperandType type, InstructionFlags flags, int opCode, bool broadcast = false, int registerWidth = 128, int maskRegisterIdx = 0, bool zeroElements = false)
```
gdkchan, line 1161:
```suggestion
        private void WriteEvexInst(
            Operand dest,
            Operand src1,
            Operand src2,
            OperandType type,
            InstructionFlags flags,
            int opCode,
            bool broadcast = false,
            int registerWidth = 128,
            int maskRegisterIdx = 0,
            bool zeroElements = false)
```

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1152,6 +1158,99 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+                InstructionFlags.PrefixF3 => 0b10,
+                InstructionFlags.PrefixF2 => 0b11,
+                _ => 0
+            };
+            WriteByte(
+                (byte)(
+                    (w ? 0x80 : 0) |
+                    (vvvv << 3) |
+                    0b100 |
+                    (pp)
```
gdkchan, line 1218:
```suggestion
                    pp
```

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1152,6 +1158,99 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+            // Extend src2 register
+            bool vp = (src2Idx & 16) == 0;
+            // Mask register index
+            Debug.Assert(maskRegisterIdx < 8, $"Invalid mask register index {maskRegisterIdx}.");
+            byte aaa = (byte)(maskRegisterIdx & 0b111);
+
+            WriteByte(
+                (byte)(
+                    (z ? 0x80 : 0) |
+                    ((ll & 3) << 5) |
```
gdkchan, line 1246: `ll` can never be out of range, so the `& 3` seems unnecessary here, unless more fields are going to be added to `ll`?

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1152,6 +1158,99 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+                    (bcast ? 0x10 : 0) |
+                    (vp ? 8 : 0) |
+                    (aaa & 7)
```
gdkchan, line 1249: `aaa` is already masked above, so doing that again here seems unnecessary.

ARMeilleure/Instructions/InstEmitSimdLogical.cs
```diff
@@ -176,7 +176,7 @@ public static void Bsl_V(ArmEmitterContext context)
                 Operand m = GetVec(op.Rm);
 
                 Operand res = context.AddIntrinsic(Intrinsic.X86Pxor, n, m);
-
+
```
gdkchan, line 179: Whitespace on empty line.

ARMeilleure/CodeGen/X86/Assembler.cs
```diff
@@ -1152,6 +1158,99 @@ private void WriteInstruction(Operand dest, Operand source, OperandType type, X8
...
+                default: Debug.Assert(false, $"Failed to EVEX encode opcode 0x{opCode:X}."); break;
+            }
+
+            WriteByte(
+                (byte)(
+                    (r ? 0x80 : 0) |
+                    (x ? 0x40 : 0) |
+                    (b ? 0x20 : 0) |
+                    (rp ? 0x10 : 0) |
+                    mm
```
gdkchan, line 1196:
```suggestion
                    mm));
```
That's how we have been formatting the code, same for the calls below.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -70,11 +74,16 @@ public enum FeatureFlags7Ebx
         public static bool SupportsAesni => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Aes);
         public static bool SupportsAvx => FeatureInfo1Ecx.HasFlag(FeatureFlags1Ecx.Avx);
         public static bool SupportsAvx2 => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx2) && SupportsAvx;
+        public static bool SupportsAvx512F => FeatureInfo7Ebx.HasFlag(FeatureFlags7Ebx.Avx512f);
```
merryhime, line 77: One has to verify that the OS kernel has AVX512 support enabled before using AVX512 instructions, otherwise this will SIGILL.

See Figure 15-2 in Volume 1 of the Intel 64 Software Developer’s Manual, but a summary is:

1. Verify that OSXSAVE is supported (bit 27 of CPUID.1H:ECX == 1).
2. Check enabled state in XCR0 via XGETBV (XCR0[7:5] = 0b111 && XCR0[2:1] == 0b11).
3. Then if bit 16 of CPUID.0x7.0:EBX is enabled should you use AVX512 instructions.

Wunkolo: I've added the `OSXSAVE` check for AVX and AVX512, though the problem I am having now is finding a way to invoke `XGETBV` to check if the operating system supports preserving YMM and ZMM state. We currently do not use 256 or 512-bit registers, but this will be a problem that needs to be solved if we decide to.

[Avx.IsSupported](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.x86.avx.issupported?view=net-6.0) _might_ be enough to verify ymm support since [I believe it checks XCR0 under the hood](https://github.com/dotnet/runtime/blob/243cf9f617a0bb99c1189de54c1bbd8d26cdba5a/src/coreclr/nativeaot/Runtime/amd64/GC.asm#L7-L12).

But there is no such equivalent for zmm though.

merryhime: One has an emitter so one could emit `xgetbv`.

Wunkolo: Just capturing what was said on the discord: At the moment all code only touches the SSE-level state and does not interface with the upper `ymm`,`zmm`, opmask, or `{x,y,z}mm16-31`-registers. It would be safe in its current state since only xmm is ever utilized but the instant any of those other registers are touched then we need to explicitly start checking for xgetbv to know if the operating system is preserving those additional registers.
image

merryhime: Just to document what I stated on discord:

> fairly sure AVX-512 is an XSAVE-enabled feature too (i.e. will #UD unless enabled via XCR0); this isn't just about state

Thus on a system which is `disable_avx512`, I suspect not checking xcr0 will crash. (I was able to verify state-upgrade on an AVX512 macOS system in this case.)

Wunkolo: The mainline linux kernel doesn't seem to support `disable_avx512` so I'm unable to test with a kernel that supports that option at the moment:
`Unknown kernel command line parameters "disable_avx512", will be passed to user space.`

The closest flags on the mainline kernel are `noxsave` and `noxsaveopt` which controls the entirely of xsave context switching and not just AVX or AVX512 in particular.
`noxsave` gives me illegal instruction on _all_ AVX instructions.
`noxsaveopt` seems to let me use xmm,ymm,zmm AVX instructions.

Windows seems to have an equivalent feature by utilizing `bcdedit /set xsavedisable 1`:
https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/bcdedit--set#additional-settings

Disabling AVX512 in the motherboard bios disables the `cpuid` flags entirely.

Checking `XCR0` would be the safest bet as I am unable to test kernel-disabled AVX512 at the moment.

merryhime: Thanks for looking into it. :+1:

Wunkolo: Added XCR0 checks for AVX and AVX512 👍

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -23,6 +26,26 @@ static HardwareCapabilities()
...
+
+            byte[] asmGetXcr0 = {
+                0x31, 0xc9, // xor ecx, ecx
+                0xf, 0x01, 0xd0, // xgetbv
+                0xc3, // ret
+            };
+
+            MemoryBlock memGetXcr0 = new MemoryBlock((ulong)asmGetXcr0.Length);
+
+            memGetXcr0.Reprotect(0, (ulong)asmGetXcr0.Length, MemoryPermission.ReadWriteExecute);
```
gdkchan, line 38: This reprotect is not necessary as initially the block is already RW (and you don't need executable permission just to write the code).

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -23,6 +26,26 @@ static HardwareCapabilities()
...
+
+            memGetXcr0.Write(0, asmGetXcr0);
+
+            memGetXcr0.Reprotect(0, (ulong)asmGetXcr0.Length, MemoryPermission.ReadAndExecute);
+
+            var fGetXcr0 = Marshal.GetDelegateForFunctionPointer<GetXcr0>(memGetXcr0.Pointer);
+
+            Xcr0InfoEax = (Xcr0FlagsEax)fGetXcr0();
+
+            memGetXcr0.Dispose();
```
gdkchan, line 48: If you change the MemoryBlock line to `using MemoryBlock memGetXcr0 = new MemoryBlock((ulong)asmGetXcr0.Length);` you can remove this dispose call.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -1,10 +1,13 @@
+using Ryujinx.Memory;
 using System;
+using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics.X86;
 
 namespace ARMeilleure.CodeGen.X86
 {
     static class HardwareCapabilities
     {
+        private delegate uint GetXcr0();
```
gdkchan, line 10: Add an empty line below to separate this from the static constructor.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
jduncanator: I think it would be better to move this entire functionality out of the constructor into a static function on the class,  it seems misplaced just shoved into the constructor like this.

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -23,6 +26,26 @@ static HardwareCapabilities()
...
+
+            MemoryBlock memGetXcr0 = new MemoryBlock((ulong)asmGetXcr0.Length);
+
+            memGetXcr0.Reprotect(0, (ulong)asmGetXcr0.Length, MemoryPermission.ReadWriteExecute);
+
+            memGetXcr0.Write(0, asmGetXcr0);
+
+            memGetXcr0.Reprotect(0, (ulong)asmGetXcr0.Length, MemoryPermission.ReadAndExecute);
+
+            var fGetXcr0 = Marshal.GetDelegateForFunctionPointer<GetXcr0>(memGetXcr0.Pointer);
```
jduncanator, line 43: You can use  `Func<uint>` here instead of declaring a new delegate type.

jduncanator: Is that a runtime exception? Weird...

jduncanator: Happy to leave this as is then!

gdkchan:
Needs a rebase.

Wunkolo:
Rebased and added `XCR0` information to the `FeatureInfo` structure. 👍

marysaka (approved):

marysaka:
Need another rebase and please review again @gdkchan

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -24,6 +28,27 @@ static HardwareCapabilities()
                 FeatureInfo7Ebx = (FeatureFlags7Ebx)ebx7;
                 FeatureInfo7Ecx = (FeatureFlags7Ecx)ecx7;
             }
+
+            Xcr0InfoEax = (Xcr0FlagsEax)GetXcr0Eax();
+        }
+
+        private static uint GetXcr0Eax()
+        {
+            byte[] asmGetXcr0 = {
```
turbedi, line 37:
```suggestion
            ReadOnlySpan<byte> asmGetXcr1 = new byte[] {
```
Compiler optimization to avoid the allocation

jduncanator: This still allocates, you'd need to stackalloc it to avoid the allocation. This optimisation only works when the variable is a statically declared member of a static class.

jduncanator: I stand corrected. It seems this optimisation has been expanded to support lifting function scoped variables declared `ReadOnlySpan` into static members.

marysaka:
Still waiting for a review from @gdkchan

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -24,6 +28,27 @@ static HardwareCapabilities()
                 FeatureInfo7Ebx = (FeatureFlags7Ebx)ebx7;
                 FeatureInfo7Ecx = (FeatureFlags7Ecx)ecx7;
             }
+
+            Xcr0InfoEax = (Xcr0FlagsEax)GetXcr0Eax();
+        }
+
+        private static uint GetXcr0Eax()
+        {
+            ReadOnlySpan<byte> asmGetXcr0 = new byte[]{
```
gdkchan, line 37:
```suggestion
            ReadOnlySpan<byte> asmGetXcr0 = new byte[]
            {
```

ARMeilleure/CodeGen/X86/HardwareCapabilities.cs
```diff
@@ -61,10 +91,21 @@ public enum FeatureFlags7Ecx
...
         }
 
+        [Flags]
+        public enum Xcr0FlagsEax
+        {
+            Sse = 1 << 1,
+            YmmHi128 = 1 << 2,
+            Opmask = 1 << 5,
+            ZmmHi256 = 1 << 6,
+            Hi16Zmm = 1 << 7,
```
gdkchan, line 101:
```suggestion
            Hi16Zmm = 1 << 7
```

gdkchan (commented):
Sorry for the delay, I will merge after those are addressed.

gdkchan (approved):
lgtm, thanks.

blottto:
broken emulation from master 1.1.478 black screen on mario paper and crashes on pokemon scarlet these problems with Intel I7-12700KF (AVX512 Enable) 
[Ryujinx_1.1.478_2022-12-18_18-54-06.log](https://github.com/Ryujinx/Ryujinx/files/10254720/Ryujinx_1.1.478_2022-12-18_18-54-06.log)
[Ryujinx_1.1.478_2022-12-18_18-57-35.log](https://github.com/Ryujinx/Ryujinx/files/10254721/Ryujinx_1.1.478_2022-12-18_18-57-35.log)
 and RTX2080
Black

PR #3670, opened 2022-09-04
Avalonia - Fixes updater

emmauss:
Fixes updater on the avalonia ui. It was incomplete previously. The window has also been replaced for a task dialog.


Narugakuruga:
The en_US.json locales changelog is unreadable. Could you make the spacings like original?

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -228,25 +242,57 @@ public static void UpdateRyujinx(UpdaterWindow updateDialog, string downloadUrl)
...
+            if (UpdateSuccessful)
+            {
+
+                var shouldRestart = await ContentDialogHelper.CreateChoiceDialog(LocaleManager.Instance["RyujinxUpdater"],
+                    LocaleManager.Instance["DialogUpdaterCompleteMessage"],
+                    LocaleManager.Instance["DialogUpdaterRestartMessage"]);
+
+                if (shouldRestart)
+                {
+                    string ryuName = OperatingSystem.IsWindows() ? "Ryujinx.Ava.exe" : "Ryujinx.Ava";
```
gdkchan, line 279: Is there no way to get the currently running executable name? It would be less fragile than the current approach I think. This code will break if we change the executable name as is.

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -228,25 +242,57 @@ public static void UpdateRyujinx(UpdaterWindow updateDialog, string downloadUrl)
...
+                    string ryuExe = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ryuName);
+                    string ryuArg = string.Join(" ", Environment.GetCommandLineArgs().AsEnumerable().Skip(1).ToArray());
```
gdkchan, line 281: If the first arg that is skipped here is the executable name, it seems like it could be a replacement for the code above.

gdkchan: Also not sure why `AsEnumerable()` is needed since arrays should be enumerable already.

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -384,19 +432,20 @@ private static void DoUpdateWithSingleThreadWorker(UpdaterWindow updateDialog, s
...
                     }
                 }
 
-                InstallUpdate(updateDialog, updateFile);
+                InstallUpdate(taskDialog, updateFile);
             }
         }
 
-        private static void DoUpdateWithSingleThread(UpdaterWindow updateDialog, string downloadUrl, string updateFile)
+        private static void DoUpdateWithSingleThread(TaskDialog taskDIalog, string downloadUrl, string updateFile)
```
gdkchan, line 446:
```suggestion
        private static void DoUpdateWithSingleThread(TaskDialog taskDialog, string downloadUrl, string updateFile)
```

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -384,19 +432,20 @@ private static void DoUpdateWithSingleThreadWorker(UpdaterWindow updateDialog, s
 
                             byteWritten += readSize;
 
-                            updateDialog.ProgressBar.Value = ((double)byteWritten / totalBytes) * 100;
+                            taskDialog.SetProgressBarState((double)byteWritten / totalBytes * 100, TaskDialogProgressState.Normal);
```
gdkchan, line 435: Would be nice to move this logic to calculate the percentage into a small static helper function. It should probably also handle the case where `totalBytes` is zero to avoid a division by zero. Same for the other functions with similar code, but if you use the helper function on all of them, then they should be all handled properly.

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -228,25 +242,57 @@ public static void UpdateRyujinx(UpdaterWindow updateDialog, string downloadUrl)
...
+
+                if (shouldRestart)
+                {
+                    string ryuName = OperatingSystem.IsWindows() ? "Ryujinx.Ava.exe" : "Ryujinx.Ava";
+                    string ryuExe = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ryuName);
+                    string ryuArg = string.Join(" ", Environment.GetCommandLineArgs().AsEnumerable().Skip(1).ToArray());
+
+                    if (!OperatingSystem.IsWindows())
+                    {
+                        chmod(ryuExe, 0777);
```
gdkchan, line 285: We had a problem with this before because the value 777 is supposed to be octal. But here it is decimal, I believe it is wrong.

emmauss: Doesn't a 0 prefix make a numeric literal octal?

gdkchan (approved):
lgtm, thanks.

Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -589,5 +589,6 @@
   "SettingsAppRequiredRestartMessage": "Ryujinx Restart Required",
   "SettingsGpuBackendRestartMessage": "Graphics Backend or Gpu settings have been modified. This will require a restart to be applied",
   "SettingsGpuBackendRestartSubMessage": "Do you want to restart now?",
+  "RyujinxUpdaterMessage": "Do you want to update Ryujinx to the latest version?"
```
LunaDook, line 592: You forgot a comma here, the current build made by github actions crashes because of it.

LunaDook, line 592: You forgot a comma here.

LunaDook (changes requested):
The current GA build is broken, easy fix:

Ryujinx.Ava/Assets/Locales/en_US.json
emmauss: Thanks

LunaDook (approved):

marysaka (approved):
lgtm

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -16,16 +16,16 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Avalonia" Version="0.10.15" />
-    <PackageReference Include="Avalonia.Desktop" Version="0.10.15" />
-    <PackageReference Include="Avalonia.Diagnostics" Version="0.10.15" />
-    <PackageReference Include="Avalonia.Controls.DataGrid" Version="0.10.15" />
-    <PackageReference Include="Avalonia.Markup.Xaml.Loader" Version="0.10.15" />
-    <PackageReference Include="Avalonia.Svg" Version="0.10.14" />
-    <PackageReference Include="Avalonia.Svg.Skia" Version="0.10.14" />
+    <PackageReference Include="Avalonia" Version="0.10.18" />
```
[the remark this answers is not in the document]
emmauss: It's been reverted.

PR #3672, opened 2022-09-05
Transform shader LDC into constant buffer access if offset is constant

gdkchan:
The `LoadConstant` IR operation is used for constant buffer loads with non-constant offset or constant buffer index. If both are constant, we can just use the regular "Cbuf" operand. Doing so allow bindless elimination to kick in on more cases (since it does not expect a `LoadConstant` operation, as those are usually used when the offset is not constant).

Improves rendering on Ys VIII: Lacrimosa of DANA:
Before:
image
After:
image


Thog (approved):

riperiperi (approved):
Nice catch.

PR #3680, opened 2022-09-06
Scale SamplesPassed counter by RT scale on report

riperiperi:
Adds a scale factor for samples passed counter report based on the render target scale at the time. This ensures that when a game reads this counter, it appears similar to the result at 1x.

This doesn't cover cases where the the render target scale changes during the queried draws, though that might be better to handle along with other scope related issues in a future rework of counters. Games generally don't count for occlusion queries over render target changes anyways.

Fixes an issue in Splatoon 2 and the splatfest demo where the special charge would scale too quickly at high res, points at the end of the game would be broken (but still provide a correct winner), and playing at a low res would make it impossible to swim in ink.

Before (0.25x):
image

After (0.25x):
image

Not that I'd recommend playing at this resolution, just demonstrates that it goes both ways. It might be better not to scale the ink surface, but this bug with samples passed is more apparent.

May also affect LOD scaling in The Witcher 3.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
@@ -1,4 +1,5 @@
 ﻿using Ryujinx.Graphics.GAL;
+using System;
 
 namespace Ryujinx.Graphics.Gpu.Engine.Threed
 {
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
```diff
@@ -151,10 +152,21 @@ private void ReportCounter(ReportCounterType type)
...
+                _ => 1f
+            };
+
             ICounterEvent counter = null;
 
             void resultHandler(object evt, ulong result)
             {
+                if (divisor != 1f)
+                {
+                    result = (ulong)Math.Ceiling(result / divisor);
```
gdkchan, line 167:
```suggestion
                    result = (ulong)MathF.Ceiling(result / divisor);
```

gdkchan (approved):
Seems reasonable.

Thog (approved):

PR #3681, opened 2022-09-08
Local Amiibo.json should be used if connection failed

AxesP:
Currently Ryujinx is not loading any Amiibo if connection fails, even if the Amiibo.json exists. This fix will use the local file and show a Dialog if connection fails.


Ryujinx/Ui/Windows/AmiiboWindow.cs
```diff
@@ -225,7 +232,7 @@ private async Task<bool> NeedsUpdate(DateTime oldLastModified)
 
                 if (response.IsSuccessStatusCode)
                 {
-                    return response.Content.Headers.LastModified != oldLastModified;
+                    return response.Content.Headers.LastModified.Value.TimeOfDay != oldLastModified.AddTicks(-oldLastModified.Ticks % TimeSpan.TicksPerSecond).TimeOfDay;
```
Thog, line 235: why was this changed?

Ryujinx.Ava/Ui/ViewModels/AmiiboWindowViewModel.cs
```diff
@@ -361,7 +368,7 @@ private async Task<bool> NeedsUpdate(DateTime oldLastModified)
 
                 if (response.IsSuccessStatusCode)
                 {
-                    return response.Content.Headers.LastModified != oldLastModified;
+                    return response.Content.Headers.LastModified?.TimeOfDay != oldLastModified.AddTicks(-oldLastModified.Ticks % TimeSpan.TicksPerSecond).TimeOfDay;
```
Thog, line 371: why was this changed?

Thog (changes requested):
Some initial comments

Also the ``JsonSerializer.Deserialize`` should be checked to be valid data, there is always the possibility the json is corrupted.

Ryujinx/Ui/Windows/AmiiboWindow.cs
AxesP: That was changed because the LastModified header has no milliseconds on it, check https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified . And the current JSON in https://amiibo.ryujinx.org/ has milliseconds on it. That will cause an error when comparing the two dates, having the same year, month, day, hour and minutes but different milliseconds, finishing on a returning "false" when the file is the same as the one in the server. This is one possible fix. The other options are:

1. Instead of comparing the date of the header, we can download the file and get the "lastmodified" property in the JSON and comparing that with oldLastModified instead of response.Content.Headers.LastModified. (This will likely never fail).
2. Compare the date in the header with the oldLastModified but without the milliseconds on it (The change I made, it's more like a workaround than a fix).
3. Correct the date on the JSON file uploaded to https://amiibo.ryujinx.org/ and allways put 0 in the milliseconds.
Example: "2022-09-09T22:00:20.776625" is the current value, the correct one should be: "2022-09-09T22:00:20.000000".
4. Or another suggestion...

AxesP:
Thanks for reviewing the code, Thog.
Okay, I updated the JSON deserialization to be more secure. Now it will use the remote file if the local is oudated or invalid,
and if deserialization fails at one point, it will use the last file that was able to deserialize or no Amiibo if neither the local or remote are valid JSON.
 Also I was thinking if those functionalities are likely to be on another project? Like Ryujinx.Ui.Common?

Thog:
> Thanks for reviewing the code, Thog. Okay, I updated the JSON deserialization to be more secure. Now it will use the remote file if the local is oudated or invalid, and if deserialization fails at one point, it will use the last file that was able to deserialize or no Amiibo if neither the local or remote are valid JSON. Also I was thinking if those functionalities are likely to be on another project? Like Ryujinx.Ui.Common?

I think it would be nice to have it in [Ryujinx.Ui.Common](https://github.com/Ryujinx/Ryujinx/tree/master/Ryujinx.Ui.Common) to avoid duplication yeah but that can be addressed in another PR if you want

Thog (approved):

AxesP:
Sorry @Thog , requested your review again by mistake. Not an expert on GitHub pull requests!

Ryujinx/Ui/Windows/AmiiboWindow.cs
marysaka: > That was changed because the LastModified header has no milliseconds on it

We aren't expecting any update with a milliseconds granularity as those data are statically generated daily if any update is present in the database.

marysaka: As such I would like that this change and [the other one](https://github.com/Ryujinx/Ryujinx/pull/3681/files/de2723e5cfa7f7962bf4d01bd78fd20e8311a8c3#r967648145) be reverted.

marysaka:
> Sorry @Thog , requested your review again by mistake. Not an expert on GitHub pull requests!

No worries, I kind of was away for the past weeks 😅

Ryujinx/Ui/Windows/AmiiboWindow.cs
AxesP: Okay, changes reverted.
Anyways, this problem needs more explanation by my side and since it goes beyond the goal of this pull request, maybe it should be addressed on another pull request.

AxesP:
> > Sorry @Thog , requested your review again by mistake. Not an expert on GitHub pull requests!
> 
> No worries, I kind of was away for the past weeks 😅

Me too... I'm having time to review this again after like 4 weeks.
Anyways, after reverting that date comparison thing, I think we are finally done with this pull request and I should only wait for approval to start working on moving that Amiibo code to Ryujinx.Ui.Common.

Thanks!

marysaka (approved):

Ryujinx.Ava/Ui/ViewModels/AmiiboWindowViewModel.cs
```diff
@@ -181,33 +182,62 @@ public void Dispose()
             _httpClient.Dispose();
         }
 
-        private async Task LoadContentAsync()
+        private bool TryGetAmiiboJson(string json, out Amiibo.AmiiboJson amiiboJson)
+        {
+            try
+            {
+                amiiboJson = JsonHelper.Deserialize<Amiibo.AmiiboJson>(json);
+                return true;
```
AcK77, line 190:
```suggestion

                return true;
```

Ryujinx.Ava/Ui/ViewModels/AmiiboWindowViewModel.cs
```diff
@@ -181,33 +182,62 @@ public void Dispose()
...
+            }
+            catch
+            {
+                amiiboJson = JsonHelper.Deserialize<Amiibo.AmiiboJson>(DefaultJson);
+                return false;
```
AcK77, line 195:
```suggestion

                return false;
```

Ryujinx/Ui/Windows/AmiiboWindow.cs
```diff
@@ -121,35 +121,62 @@ public AmiiboWindow() : base($"Ryujinx {Program.Version} - Amiibo")
             _ = LoadContentAsync();
         }
 
-        private async Task LoadContentAsync()
+        private bool TryGetAmiiboJson(string json, out AmiiboJson amiiboJson)
+        {
+            try
+            {
+                amiiboJson = JsonHelper.Deserialize<AmiiboJson>(json);
+                return true;
```
AcK77, line 129:
```suggestion

                return true;
```

Ryujinx/Ui/Windows/AmiiboWindow.cs
```diff
@@ -121,35 +121,62 @@ public AmiiboWindow() : base($"Ryujinx {Program.Version} - Amiibo")
...
+            }
+            catch
+            {
+                amiiboJson = JsonHelper.Deserialize<AmiiboJson>(DEFAULT_JSON);
+                return false;
```
AcK77, line 134:
```suggestion

                return false;
```

AcK77 (changes requested):

AcK77 (approved):

AcK77:
Needs rebase

AxesP:
Rebase done.

AcK77:
Format have to be fixed (you can look at the CI logs to known the issues)

AxesP:
@AcK77 Code reformatted.
My bad. I was trying to get Visual Studio Code to use the editorconfig but with no success...

AxesP:
@AcK77 One reference to the logging was removed in the merge by mistake, I readded the reference and ran dotnet build to double check (My bad again). The project should compile now.

AcK77 (approved):

PR #3683, opened 2022-09-08
Add ADD (zx imm12), NOP, MOV (rs), LDA, TBB, TBH, MOV (zx imm16) and CLZ thumb instructions

gdkchan:
Implements the following T32 instructions:
- ADD (zero-extended 12-bit immediate variant)
- NOP
- MOV (register shifted by register variant)
- MOV (zero-extended 16-bit immediate variant)
- TBB/TBH (Table Branch Byte/Halfword, those *only* exists on T32)
- LDA (Load Acquire)
- CLZ (Count Leading Zeros)

Additionally fixes some existing implementations:
- Fix T16 BLX using `Blx` rather than `Blxr`. It was causing `InvalidCastException` because `Blx` expects an immediate address, not a register.
- Fix T16 CBZ/CBNZ instructions control flow. `IsAarch32UncoditionalBranch` was returning true for those instructions, which would cause the code immediately following it (branch not taken path) to be skipped and was silently generating incorrect code.
- Fix T32 LDRD/STRD encoding. It was overlapping with LDA/STL in the `OpCodeTable`.
- Fix T32 LDRD/STRD immediate decoding. It was not being multiplied by 4.
- Fix T32 LDRD/STRD Rt2 register. Unlike the A32 variant, on Thumb the Rt2 register can be explicitly specified on the instruction, instead of being the register immediately after Rt. The `InstEmit` function and OpCode interfaces were changed to account for this.

This allows Vita2HOS to launch again, it goes as far as the old version that did not use as many Thumb instructions. I planned to get more work done here, but since this contains some important fixes, I decided to PR it sooner rather than later.


Thog (approved):

riperiperi (approved):
LGTM.

xerpi:
Thank you very much for the PR, much appreciated!

PR #3687, opened 2022-09-10
Implement Thumb (32-bit) memory (ordered), multiply, extension and bitfield instructions

gdkchan:
Implements the following instructions:
- All remaining Load-Acquire, Load-Acquire (exclusive), Store-Release and Store-Release (exclusive) instructions.
- Multiply instructions that are implemented on A32.
- SDIV and UDIV.
- Bitfield instructions implemented on A32 (BFC, BFI, SBFX and UBFX).
- Zero and sign-extension instructions (UXTB, SXTB, etc).
- Remaining register shifted by immediate byte/hword/word memory instructions.

Change the existing implementations:
- Of T32 memory instructions with 8-bit immediate offset to allow Rt to be PC. Previously it was not allowed, not sure why, this is valid according to the manual and applications do make use of it.
- Of T32 BL instruction that had an incorrect immediate value.

Allows Vita2HOS to go a bit further when launching applications. I will probably work on Thumb NEON instructions next.


Thog (approved):

ARMeilleure/Decoders/OpCodeT32AluMla.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public OpCodeT32AluMla(InstDescriptor inst, ulong address, int opCode) : base(inst, address, opCode)
+        {
+            Rm = (opCode >> 0) & 0xf;
+            Rd = (opCode >> 8) & 0xf;
+            Ra = (opCode >> 12) & 0xf;
+            Rn = (opCode >> 16) & 0xf;
+            R = (opCode & (1 << 4)) != 0;
+
+            MHigh = ((opCode >> 4) & 0x1) == 1;
```
riperiperi, line 25: This and R are the same? Bit weird that it checks the same thing two different ways.

gdkchan: Yes, it also clashes on `OpCode32AluMla`:
```cs
            R = (opCode & (1 << 5)) != 0;

            NHigh = ((opCode >> 5) & 0x1) == 1;
            MHigh = ((opCode >> 6) & 0x1) == 1;
```
They are used by different instructions. The one that has R does not have NHigh/MHigh, and the one that has NHigh/MHigh does not have R. Maybe they should use different types, but I was just following what was done on the A32 implementation.

riperiperi (approved):

PR #3689, opened 2022-09-10
Periodically Flush Commands for Vulkan

riperiperi:
NVIDIA's OpenGL driver has a built-in mechanism to automatically flush commands to GPU when a lot have been queued. It's also pretty inconsistent, but we'll ignore that for now.

Our Vulkan implementation only submits a command buffer (flush equivalent) when it needs to. This is typically when another command buffer needs to be sequenced after it, presenting a frame, or an edge case where we flush around GPU queries to get results sooner.

This difference in flush behaviour causes a notable difference between Vulkan and OpenGL when we have to wait for commands. In the worst case, we will wait for a sync point that has just been created. In Vulkan, this sync point is created by flushing the command buffer, and storing a waitable fence that signals its completion. Our command buffer contains _every command that we queued since the last submit_, which could be an entire frame's worth of draws.

This has a huge effect on CPU <-> GPU latency. The more commands in a command buffer, the longer we have to wait for it to complete, which results in wasted time. Because we don't know when the guest will force us to wait, we always want the smallest possible latency.

By periodically flushing commands, we ensure that each command buffer takes a more consistent, smaller amount of time to execute, and that the back of the GPU queue isn't as far away when we need to wait for something to happen. This also might reduce time that the GPU is left inactive while commands are being built.

This PR adds a class to keep track of time spent writing commands since the last flush, and flush commands on attachment change (when render pass will be ended). Flushing for query results has been integrated into this new class and tweaked a bit.

## Improvements:

The main affected game is Pokemon Sword, which got significantly faster in overworld areas due to reduced waiting time when it flushes a shadow map from the main GPU thread.

These screenshots have additional performance counters. GpuWait is time spent waiting on the GPU from guest threads, GpuInvoke is time spent waiting on the GPU from the main GPU thread.

### Before:
image

### After
image

This one is weird in that a lot of its cost lies in layout conversion rather than waiting for texture flush. This can be improved in the future. On a weaker system (steam deck) this goes from around 22FPS -> 25FPS, though the layout conversion is likely hitting hardest here.

Another affected game is BOTW, which gets faster depending on the area. This game flushes textures/buffers from its game thread, which is the bottleneck. The effect is smaller, and this could be improved further by pre-emptively flushing textures that are flushed often.

### Before:
image

### After:
image

## Testing:

Flush latency and throughput may be improved on other games that are inexplicably slower than OpenGL. It's possible that certain games could have their performance _decreased_ slightly due to command buffer submit not being free, but it is unlikely.

Also, flushing to get query results sooner has been tweaked to improve the number of full draw skips that can be done. (tested in SMO)

This does not affect Xenoblade DE being slower in some areas.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
+++ b/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
@@ -1,0 +1,78 @@
+﻿using System.Diagnostics;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    internal class AutoFlushCounter
+    {
+        /** How often to flush on framebuffer change. */
+        private readonly static long FramebufferFlushTimer = Stopwatch.Frequency / 1000;
+
+        private const int MinDrawCountForFlush = 10;
+        private const int InitialQueryCountForFlush = 32;
+
+        private long _lastFlush;
+        private ulong _lastDrawCount;
+        private bool _hasPendingQuery;
+        private int _queryCount;
+
+        public void RegisterFlush(ulong drawCount)
+        {
+            _lastFlush = Stopwatch.GetTimestamp();
+            _lastDrawCount = drawCount;
+
+            _hasPendingQuery = false;
+        }
+
+        public void RegisterPendingQuery()
+        {
+            _hasPendingQuery = true;
+        }
+
+        public bool ShouldFlushQuery()
+        {
+            return _hasPendingQuery;
+        }
+
+        public bool ShouldFlushDraw()
+        {
+            // Interrupt render passes to flush queries, so that early results arrive sooner.
+            if (_hasPendingQuery)
+            {
+                if (++_queryCount == InitialQueryCountForFlush)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        public bool ShouldFlush(ulong drawCount)
+        {
+            _queryCount = 0;
+
+            if (_hasPendingQuery)
+            {
+                return true;
+            }
+
+            long draws = (long)(drawCount - _lastDrawCount);
+
+            if (draws < MinDrawCountForFlush)
+            {
+                if (draws == 0)
+                {
+                    _lastFlush = Stopwatch.GetTimestamp();
+                }
+
+                return false;
+            }
+
+            long flushTimeout = FramebufferFlushTimer;
+
+            long now = Stopwatch.GetTimestamp();
+
+            return now > _lastFlush + flushTimeout;
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System.Diagnostics;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    internal class AutoFlushCounter
+    {
+        /** How often to flush on framebuffer change. */
```
gdkchan, line 7:
```suggestion
        /* How often to flush on framebuffer change. */
```
Could also use `// ` comments which is what we usually use for single line comments.

Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
```diff
@@ -0,0 +1,78 @@
...
+        private readonly static long FramebufferFlushTimer = Stopwatch.Frequency / 1000;
```
gdkchan, line 8: I think private static fields are named like private instance fields, `_framebufferFlushTimer`, although I don't think the naming for those is consistent in the code right now.

gdkchan (commented):
It's worth noting that the fact that the command submit is partially time based could cause intermittent issues, in case we have a bug that only manifests for a given sequences of commands on a command buffer for example. This is just an hypothetical scenario, I'm not saying that such an issue exists right now, just something to keep in mind.

Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
riperiperi: I was kind of torn on this because it's essentially a `const`, but Stopwatch.Frequency is only available at runtime.

LukeWarnut:
LIVE A LIVE is crashes at launch for me with this PR. It runs fine on master.

[Ryujinx_1.1.0+f39b918_2022-09-11_13-24-36.log](https://github.com/Ryujinx/Ryujinx/files/9543452/Ryujinx_1.1.0%2Bf39b918_2022-09-11_13-24-36.log)

riperiperi:
I don't think this is related, this error happens when your shader cache is corrupt.

LukeWarnut:
> I don't think this is related, this error happens when your shader cache is corrupt.

You're right. Now it crashes on master. Not sure why it didn't happen before.

gdkchan (approved):
lgtm, nice to see performance improved on Vulkan! I think the private static field should be renamed with underscore prefix and lower case first character to match existing code style though.

Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
SomeoneIsWorking: Sorry to be intruding but I'm also in favor of naming  `readonly static` as same as `const` as `PascalCase` regardless of access modifier. Different case without `readonly`.

gdkchan: Is that what other projects are doing? We are just trying to follow standard C# naming here.

Thog (approved):

PR #3690, opened 2022-09-10
T32: Add Vfp instructions

merryhime:
T32 has identical encoding to A32 except cond is always 1110.


gdkchan:
When combined with #3687, it allows the triangle homebrew to work on Vita2HOS:
image

gdkchan (approved):
LGTM, thanks. Didn't know that the encoding of some T32 instructions is actually identical to the A32 ones.

riperiperi (approved):
Nice, cool way of dealing with the encodings being similar in the OpCodeTable.

PR #3691, opened 2022-09-10
Implement VRINT (vector) Arm32 NEON instructions

gdkchan:
This implements the NEON VRINTA/N/M/P instructions (Advanced SIMD variants). The operations match the scalar variants.
With #3687 and #3690, allows Ni no Kuni to go ingame, but only with a save (since otherwise it requires the web applet).

image
image
Something seems wrong with voice audio, but I do not believe to be caused by this implementation.


gdkchan:
For convenience, here's the save that I used:
[config.zip](https://github.com/Ryujinx/Ryujinx/files/9541608/config.zip)

MarkLanders777:
#3690 conflicts with 3691, so they can't be merged together with 3687

merryhime (approved):

riperiperi (approved):

AcK77 (approved):

PR #3692, opened 2022-09-11
T32: Implement Asimd instructions

merryhime:
Hopefully self explanatory; makes use of T32 encoding being very similar to the A32 encoding for these instructions.

Really there are only two transforms here:
```
11110100 -> 11111001
1111001x -> 111x1111
```

Instruction classes affected by bit 24->28 shift:
 * `OpCode32SimdImm`
 * `OpCode32SimdLong`
 * `OpCode32SimdReg`
 * `OpCode32SimdRegElem`
 * `OpCode32SimdRegElemLong`
 * `OpCode32SimdRegLong`
 * `OpCode32SimdRegWide`
 * `OpCode32SimdShImm`
 * `OpCode32SimdShImmLong`
 * `OpCode32SimdShImmNarrow`

Note that this is usually handled in a base class; because this is usually the common `U` bit.

### DIFF
--- a/ARMeilleure/Decoders/OpCode32Simd.cs
+++ b/ARMeilleure/Decoders/OpCode32Simd.cs
@@ -7,14 +7,15 @@
         public bool F { get; protected set; }
         public bool U { get; }
 
-        public new static OpCode Create(InstDescriptor inst, ulong address, int opCode) => new OpCode32Simd(inst, address, opCode);
+        public new static OpCode Create(InstDescriptor inst, ulong address, int opCode) => new OpCode32Simd(inst, address, opCode, false);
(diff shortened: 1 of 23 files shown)


gdkchan:
When combined with #3693, allows VITA-8 to work on Vita2HOS.
image
image

gdkchan (approved):
LGTM, thanks! Nice to see that they have almost identical encodings too, sharing them will make our job a bit easier if we need to change/fix one of those encodings in the future.

riperiperi (approved):
Great, nice to see so many T32 encodings down at once.

Thog:
Need a rebase

merryhime:
done

riperiperi (approved):

PR #3694, opened 2022-09-11
A32/T32/A64: Implement Hint instructions (CSDB, SEV, SEVL, WFE, WFI, YIELD)

### LINKED ISSUE #3686: Unimplemented SEV instruction

```
The instruction at 0x00000000003357f4 (opcode 0xe320f004) is undefined!
```

### Instruction name
```
0x0000000000000000:  04 F0 20 E3    sev 
```

### Required by: 
Hanayaka Nari Waga Ichizoku Modern Nostalgie [01008DE00C022000] [32-bit] (no game issue exists)
Original issue here: https://github.com/Ryujinx/Ryujinx/issues/2717

merryhime:
Implement as nops.

Closes #3686.


gdkchan:
You should probably remove the existing A64 catch-all hint instruction: https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Decoders/OpCodeTable.cs#L110

merryhime:
Would you prefer the other way around? (adding a catch all to all instruction sets?)

gdkchan:
> Would you prefer the other way around? (adding a catch all to all instruction sets?)

Personally I prefer having them separate.
And eventually we should probably implement WFE/SEV/SEVL since otherwise the CPU will be spin waiting (which is technically still correct, but not optimal).

merryhime:
> And eventually we should probably implement WFE/SEV/SEVL since otherwise the CPU will be spin waiting (which is technically still correct, but not optimal).

My fear is introducing unintentional deadlocks due to beibng overenthusiastic here. I think a simple thing to do is to do a host OS thread yield on WFE/WFI when required (since afaik Ryujinx is 1:1 with host:emu threads).

ARMeilleure/Decoders/OpCodeTable.cs
```diff
@@ -107,7 +108,6 @@ static OpCodeTable()
             SetA64("11001010xx0xxxxxxxxxxxxxxxxxxxxx", InstName.Eor,             InstEmit.Eor,             OpCodeAluRs.Create);
             SetA64("00010011100xxxxx0xxxxxxxxxxxxxxx", InstName.Extr,            InstEmit.Extr,            OpCodeAluRs.Create);
             SetA64("10010011110xxxxxxxxxxxxxxxxxxxxx", InstName.Extr,            InstEmit.Extr,            OpCodeAluRs.Create);
-            SetA64("11010101000000110010xxxxxxx11111", InstName.Hint,            InstEmit.Hint,            OpCodeSystem.Create);
```
gdkchan, line 110: The enum entry and the Hint function can be removed too.

gdkchan (approved):
lgtm, thanks!

ARMeilleure/Decoders/OpCodeTable.cs
```diff
@@ -1060,7 +1070,14 @@ static OpCodeTable()
...
             SetT16("101110x1xxxxxxxx", InstName.Cbnz,   InstEmit32.Cbnz,    OpCodeT16BImmCmp.Create);
             SetT16("1011110xxxxxxxxx", InstName.Pop,    InstEmit32.Ldm,     OpCodeT16MemStack.Create);
-            SetT16("10111111xxxx0000", InstName.Nop,    InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111100000000", InstName.Nop,    InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111100010000", InstName.Yield,  InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111100100000", InstName.Wfe,    InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111100110000", InstName.Wfi,    InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111101000000", InstName.Sev,    InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("1011111101010000", InstName.Sevl,   InstEmit32.Nop,     OpCodeT16.Create);
+            SetT16("10111111011x0000", InstName.Hint,   InstEmit32.Nop,     OpCodeT16.Create); // Hint instruction
```
gdkchan, line 1079: Are there no unallocated hint instructions on A32 and A64?

merryhime: They do exist, I could do the same.

Thog (approved):

gdkchan:
Needs a rebase.

merryhime:
rebased

PR #3695, opened 2022-09-12
Fix increment on Arm32 NEON VLDn/VSTn instructions with regs > 1

gdkchan:
Due to what appears to be an oversight, not all "modes" were being tested on the VLDn/VSTn (pair) tests. Only VLD1 and VST1 was being tested. Fixing that revealed several failures caused by an incorrect register increment value, which caused some of those instructions to get or put the value on the wrong register.

This change fixes the above issues, and additionally updates the `OpCodeTable` to account for several "Undefined" encoding, which were considered valid before.

Fixes black screen on No More Heroes 2.
Before:
image
After:
image
Fixes several missing effects on Pikmin 3 Deluxe.
Before (pay attention to the antenna above the main character head):
image
After:
image
Fixes the main character falling below the ground on Sphinx the Cursed Mummy.
Before:

https://user-images.githubusercontent.com/5624669/189557314-6fc81fed-efc4-49a2-993f-eac3016088d5.mp4

After:

https://user-images.githubusercontent.com/5624669/189557325-64246fcb-bc37-4846-81c7-5d60811ac66c.mp4

This also fixes bad voice audio quality on Ni no Kuni.

### DIFF
--- a/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
@@ -67,7 +67,7 @@ namespace ARMeilleure.Instructions
 
                 for (int i = 0; i < count; i++)
                 {
-                    // Read an element from a double simd register.
+                    // Accesses an element from a double simd register.
                     Operand address = context.Add(n, Const(offset));
                     if (eBytes == 8)
                     {
@@ -145,7 +145,7 @@ namespace ARMeilleure.Instructions
                         int elemD = d + reg;
                         for (int i = 0; i < count; i++)
                         {
-                            // Write an element from a double simd register,
+                            // Accesses an element from a double simd register,
                             // add ebytes for each element.
                             Operand address = context.Add(n, Const(offset));
                             int index = ((elemD & 1) << (3 - op.Size)) + elem;
--- a/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
+++ b/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
@@ -67,7 +67,7 @@ namespace ARMeilleure.Instructions
 
                 for (int i = 0; i < count; i++)
                 {
-                    // Write an element from a double simd register.
+                    // Read an element from a double simd register.
                     Operand address = context.Add(n, Const(offset));
                     if (eBytes == 8)
                     {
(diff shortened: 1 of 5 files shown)


GamerzHell9137:
No More Heroes 1 is working too now! 
image

LukeWarnut:
(Fittingly) Fixes in-game sound effects in Double Dragon Neon.

ARMeilleure/Instructions/InstEmitSimdMemory32.cs
```diff
@@ -67,7 +67,7 @@ public static void EmitVStoreOrLoadN(ArmEmitterContext context, int count, bool
 
                 for (int i = 0; i < count; i++)
                 {
-                    // Write an element from a double simd register.
+                    // Read an element from a double simd register.
```
riperiperi, line 70: I'm not sure this is much better, since it either loads or stores. Maybe "access"?

ARMeilleure/Instructions/InstEmitSimdMemory32.cs
```diff
@@ -144,7 +145,7 @@ public static void EmitVStoreOrLoadN(ArmEmitterContext context, int count, bool
                         int elemD = d + reg;
                         for (int i = 0; i < count; i++)
                         {
-                            // Write an element from a double simd register
+                            // Write an element from a double simd register,
```
riperiperi, line 148: This comment should get any changes the above one gets.

riperiperi (approved):
Nice catch, don't know how I added 11 cases to the table to test but only ran 4 of them. 🥴 

Can confirm that this fixes most of the missing effects in Pikmin 3 Deluxe (demo), which is good news for the guy totally losing it over on the games list repo.

Before:
image

After:
image

EmulationFanatic:
Fixes bad audio on Sky Gamblers: Storm Raiders

Before:
https://cdn.discordapp.com/attachments/694459861190311957/1018851213015527435/Storm_Raiders_BadAudio.mp4

After:
https://cdn.discordapp.com/attachments/694459861190311957/1018851319009787974/StormRaiders_Fixed.mp4

ARMeilleure/Instructions/InstEmitSimdMemory32.cs
gdkchan: For some reason I assumed that one path handled loads and the other stores when I wrote this. But they both handle loads and stores. I have changed both to "Accesses".

Thog (approved):

PR #3697, opened 2022-09-12
Fix bindless 1D textures having a buffer type on the shader

gdkchan:
Because 1D and Buffer textures uses the same type on most texture instructions on the shader, it must get the actual texture type from GPU state. However, it was not doing that for bindless textures and assuming a buffer type, even if bindless elimination was successful. This change allows it to fix the type if bindless elimination is successful.

Fixes black screen on Prinny 2.
Master:
image
This PR:
image
This PR + #3695:
image

Testing is welcome.


LukeWarnut:
Also fixes Prinny 1
image

riperiperi (approved):

Thog (approved):

PR #3698, opened 2022-09-12
Avalonia: Line up volume status bar item

merryhime:
image

Issue with alignment had to do with use of iconography. The simplest fix was to remove this.


MutantAura:
Is alignment with the icon not possible on all systems? I fixed it's alignment in: https://github.com/Ryujinx/Ryujinx/pull/3643 but haven't checked on other OS's etc.
See below:
image

merryhime:
I tried doing similar on *nix. I don't remember it working. I'll verify this later.

merryhime:
Closed in favour of whatever will happen in #3643.

Issue #2917, opened 2021-12-09
CPU JIT should do inline saturation for saturating instructions


gdkchan:
#### Overview:
Currently, there's a significant amount of saturation functions implemented as C# "helper" function in the `SoftFallback` class. Some of those functions are very simple (like the SatXXXToXXX functions), and in those cases, the costs of calling the function alone, including the native <-> managed transition cost is far greater than that of the function itself, and so doing those calls is bad for performance. Instead, we should have efficient inline implementations using IR and avoid the calls entirely.

There are also several other "SatQ" functions in this same class that is used by other instructions. Some of those is responsible for the slow video decoding in some games, like Astral Chain for example, which uses software decoded Bink 2 videos. We should also inline those to improve performance on those titles.

#### Smaller Details: 
I think most of them are simple enough to just emit inline, but if there's a concern that they might be too big, we might aswell generate a native function and call it, it's still faster than a native -> managed call.

#### Nature of Request:  
This is a performance enhancement.

### Why would this feature be useful?
Brings the performance up for some software decoded videos and potentially other cases.

saldabain:
If you could provide an example on how to implement using IR, I can try to tackle this issue.

PR #3701, opened 2022-09-14
Fpsr and Fpcr freed.

LDj3SNuD:
Handling/isolation of Fpsr and Fpcr via register for IRs and via memory for Tests and KThreads, with implicit synchronization to context exchanges (explicit for SoftFloat calls), without having to call the NativeInterface managed methods (Get/Set Fpsr/Fpcr/Fpscr but above all SetFpsrQc and GetFpcrFz); also due to the inlining work of the previous two PRs (Saturating and ShlReg regions) and this PR (Rounding region).

Tests performed locally in both release and debug modes, in both lowcq and highcq, with FastFP to true and false (explicit FP tests included). Tested with the title Tony Hawk's PS.


riperiperi (approved):
Nice! This has been a long time coming. 😌

ARMeilleure/Instructions/InstEmitSimdHelper.cs
```diff
@@ -361,6 +361,56 @@ public static Operand EmitRoundMathCall(ArmEmitterContext context, MidpointRound
...
+            context.Copy(res, EmitUnaryMathCall(context, nameof(Math.Ceiling), op));
+            context.Branch(lblEnd);
+
+            context.MarkLabel(lbl2);
+            context.BranchIf(lbl3, rMode, rM, Comparison.NotEqual);
+            context.Copy(res, EmitUnaryMathCall(context, nameof(Math.Floor), op));
+            context.Branch(lblEnd);
+
+            context.MarkLabel(lbl3);
+            context.BranchIf(lblEnd, rMode, rZ, Comparison.NotEqual);
```
gdkchan, line 405: Is that case really possible? If RMode only has 2, then there are only 4 possible values.

ARMeilleure/Instructions/SoftFloat.cs
```diff
@@ -12,8 +12,8 @@ static SoftFloat()
             RecipSqrtEstimateTable = BuildRecipSqrtEstimateTable();
         }
 
-        internal static readonly byte[] RecipEstimateTable;
-        internal static readonly byte[] RecipSqrtEstimateTable;
+        public static readonly byte[] RecipEstimateTable;
```
gdkchan, line 15: Why are those public now?

gdkchan: Oh I see, they were internal before. I think they should be private? Does anything outside of that class needs to access it?

gdkchan:
I guess it can finally close #2917.

ARMeilleure/Instructions/InstEmitSimdHelper.cs
LDj3SNuD: That is in the hypothetical case rMode has a spurious value, to preserve op and not truncate it; as the default case, for completeness.

ARMeilleure/Instructions/SoftFloat.cs
LDj3SNuD: The class is already internal, needless to reiterate for its members; references in SoftFloat32/64 therefore cannot be private.

ARMeilleure/Instructions/InstEmitSimdHelper.cs
gdkchan: Removing it would improve the generated code since it would eliminate one conditional branch. It should not be possible for the value to not be 0, 1, 2 or 3, since it only reads 2 bits. It would also match the old managed implementation that does not account for invalid values (it ends with an `else` rather than `else if`).

LDj3SNuD:
> ~I guess it can finally close #2917.~ Nevermind, it also mentions the inline saturation for float to integer conversion, which was not done yet.

Thanks to #3712 and this PR it is no longer necessary to inline the Saturation region methods, because there are no more A64/A32 instructions that have only slow paths; except for the A32 Vcvt_FI instruction (with roundWithFpscr to true).
I think I can optimize it with a new commit in this PR.

ARMeilleure/Instructions/InstEmitSimdHelper.cs
LDj3SNuD: Done.

gdkchan (approved):
Very nice, thanks for looking into this! I guess now we only need to call managed code for instructions that really needs it, such as SVC and access to system registers. I tested this on Tony Hawk's Pro Skater and noticed some improvement aswell compared to master, it's pretty smooth now.

PR #3702, opened 2022-09-14
Fix partial unmap reprotection on Windows

gdkchan:
This fixes a regression from #3648 that caused punctuation to be missing from text on Super Smash Bros Ultimate, among other issues. Because `MapViewInterval` updates the protections on the tree (to RW), and it is called on partial unmap, restoring protections was not working properly and it would always try to reprotect as RW, which causes the write tracking to be lost. In addition to that, I also fixed the protections tree initialization, which was passing size rather than end address, this was likely causing other issues.

Only affects Windows, and since most games don't trigger partial unmap, the number of affected games is likely low.


Thog (approved):

riperiperi (approved):
Nice catch, wonder if this might affect other games.

PR #3707, opened 2022-09-16
UI: Allow overriding graphics backend + Move command line parser into a new class

TSRBerry:
This PR adds a new command line option `-g`/`--graphics-backend` which allows to override the previously configured value on launch. This is especially useful when launching Ryujinx with a file path to auto start, since some games work better on Vulkan while others prefer OpenGL.

Additionally command line arguments are now kept when Avalonia restarts (except for the overridden graphics backend).

In the process I also moved some shared logic between GTK and Avalonia into a new static class `ProgramHelper`, which reduces the amount of duplicated code for these projects.


riperiperi (commented):
Nice to see this cleaned up a little.

Ryujinx.Ava/Program.cs
```diff
@@ -112,69 +102,17 @@ public static AppBuilder BuildAvaloniaApp()
...
-                    startFullscreenArg = true;
-                }
-                else
-                {
-                    launchPathArg = arg;
-                }
-            }
-
-            // Delete backup files after updating.
-            Task.Run(Updater.CleanupUpdate);
```
riperiperi, line 157: What happened to this?

jduncanator:
Don't worry about it until it merges. If theirs merges first, you'll just have to resolve the merge conflict here. If the other way around, they'll resolve it in their PR.

Yohoki:
> Just saw #3744 and I'm not sure how to approach this. Since they are fixing something that I'm moving to another file, should I add these changes or should it be the other way around and they would need to adjust their PR?
> 
> https://github.com/TSRBerry/Ryujinx/blob/288c904767c1815367eff7fa0d78d0b74cc7eb1f/Ryujinx.Ui.Common/Helper/ProgramHelper.cs#L81

Mine's a very minimal change and I've been asked to look at a different change anyway. I'll figure mine out and make a change after your PR goes through.

Like the addition, btw. I'll be using this.

TSRBerry:
I thought it would be rude to just 'steal' your PR like that. :D
I just saw it and thought it could cause some issues that we may want to discuss beforehand.

Yohoki:
No big issue for me. You changed the code a bit, so I'll need to retest it and things, But looking at it here, it looks like the bug I was addressing is still present, even though you're handling the strings a bit differently. But it should still be an easy fix. I have no problem redoing it. I need the practice anyway.

TSRBerry:
Rebased onto master.

Ryujinx.Ui.Common/Helper/ProgramHelper.cs
```diff
@@ -0,0 +1,195 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Ui.Common.Helper
+{
+    public static class ProgramHelper
+    {
+        private static string _overrideGraphicsBackend = null;
+        private static string _baseDirPathArg          = null;
+        public  static string CommandLineProfile       = null;
```
gdkchan, line 17: I don't think we need the alignment on everything here, it looks a bit weird.

Ryujinx.Ui.Common/Helper/ProgramHelper.cs
```diff
@@ -0,0 +1,195 @@
...
+using Ryujinx.Common.SystemInfo;
+using Ryujinx.Ui.Common.Configuration;
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Ui.Common.Helper
+{
+    public static class ProgramHelper
```
gdkchan, line 13: Would be nice to have a more descriptive name for this class. Maybe something like `CommandLineParser` or `CommandLineState`. It seems that some shared initialization logic was moved here aswell. My personal preference is only handling the command line parsing logic here.

Ryujinx.Ui.Common/Helper/ProgramHelper.cs
```diff
@@ -0,0 +1,195 @@
...
+                }
+            }
+        }
+
+        // Return command line args without config overrides
+        public static string[] GetCommandLineArgs()
+        {
+            var args = Environment.GetCommandLineArgs().AsEnumerable().Skip(1).ToList();
+
+            if (args.Contains("-g") || args.Contains("--graphics-backend"))
```
gdkchan, line 83: Would be better to not need to check those args again here. If we need to add more args later, then we will potentially need to ignore them here aswell, which means that there's at least 2 methods that needs to be updated. I would prefer if all that logic was contained in a single method, ideally on `ParseArguments`. I think in general a better approach is not using `Environment.GetCommandLineArgs()` at all, and instead just store the arguments that this method will need on `ParseArguments`. Then this method would just return the array that `ParseArguments` built.

I think it might be cleaner to make this a non-static class and do the `ParseArguments` logic on the constructor, since that way you can enforce that the methods can't be called without the state being initialized beforehand, but then you need to pass it around.

Ryujinx.Ui.Common/Helper/ProgramHelper.cs
TSRBerry: Not sure if I understand you correctly.
Should I move the command line specific stuff into a new static class `CommandLineState` and keep the other shared logic in here? Or would you prefer to only move the command line logic into a separate class without keeping everything else?

gdkchan: My suggestion was moving the command line logic into a separate class, and "unsharing" the remaining initialization code that was moved here. So the `ProgramHelper` class would not exist anymore. I agree in sharing as much code as possible and reducing duplication, but in this case it feels out of place, and since one of the UIs will go away eventually, I don't think it will be a big deal. I think in the future we should look into refactoring this in a way that makes implementing new UIs easier and move the initialization logic somewhere else, but that's outside the scope of this change.

TSRBerry:
(It seems I can't properly re-request reviews for some reason..)

Ryujinx.Ui.Common/Helper/CommandLineState.cs
```diff
@@ -0,0 +1,82 @@
+using Ryujinx.Common.Logging;
+using System.Collections.Generic;
+using System.Linq;
```
gdkchan, line 3: Is that really needed? From a quick look it doesn't look like anything is using LINQ here.

Ryujinx.Ui.Common/Helper/CommandLineState.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace Ryujinx.Ui.Common.Helper
+{
+    public static class CommandLineState
+    {
+        public static string[] Arguments { get; private set; }
+
+        public static string OverrideGraphicsBackend { get; private set; }
+        public static string BaseDirPathArg          { get; private set; }
+        public static string CommandLineProfile      { get; private set; }
```
gdkchan, line 13: nit: This can be called just `Profile` since the class is already called `CommandLineState`.

gdkchan (approved):
lgtm other than a few nits, thanks.

Ryujinx.Ui.Common/Helper/CommandLineState.cs
TSRBerry: In L79 I'm using `args.ToArray()` which is a LINQ function.

Should I try to avoid using it?

gdkchan: 'List<T>' has a 'ToArray' method (see: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.toarray?view=net-7.0), it's not a LINQ extension (although there is a `ToArray` extension method for other collections that does not explicitly implement it). Unless I'm missing something here?

TSRBerry: Heh, I figured it out, I typo'd it :D
I got compiler errors without the using, but only because I was trying to do `args.ToArray()` instead of `arguments.ToArray()`

marysaka (approved):

PR #3715, opened 2022-09-19
Convert Quads to Triangles in Vulkan

riperiperi:
This PR adds a mechanism for topology conversion, and uses it to convert quads to triangles before drawing. This is an alternative to our previous approach of queuing one draw per quad, and converts quads to Triangles instead of Triangle Fan.

This is done using an `IndexBufferPattern`, which is a simple way to define a repeating pattern of indices that can be used for topology conversion.

- Non-Indexed draws become indexed, using a shared index buffer that indexes the quad vertices so that they can be drawn as a triangle list. This buffer is stored in the IndexBufferPattern and expanded when needed.
- Indexed draws have their index buffers converted using the `IndexBufferPattern` into one that can be drawn as a triangle list. The rules are quite similar to the vertex buffer restride - the buffer is recalculated whenever the parent buffer is written to.

I8 -> I16 conversion has also been updated to respond to writes to vertex buffers that are currently bound.

Affected games include Skyward Sword, FAST RMX (still has some fun texture cache issues, but this uses the most quads) and Xenoblade 1/2/3. Generally you should check to make sure games don't blow up - I don't expect them to become considerably faster.

It's possible to implement topology conversion with a compute shader, but it's not common enough that I think it's needed for a first impl. It's also possible to port this to OpenGL in the future, but host quads generally work better when available.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
+++ b/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
@@ -79,13 +79,13 @@ namespace Ryujinx.Graphics.Vulkan
 
             autoBuffer = gd.BufferManager.GetBufferTopologyConversion(cbs, _handle, _offset + firstIndexOffset, indexCount * indexSize, pattern, indexSize);
 
-            int size = convertedCount * indexSize;
+            int size = convertedCount * 4;
 
             _buffer = autoBuffer;
 
             if (autoBuffer != null)
             {
-                gd.Api.CmdBindIndexBuffer(cbs.CommandBuffer, autoBuffer.Get(cbs, 0, size).Value, 0, _type);
+                gd.Api.CmdBindIndexBuffer(cbs.CommandBuffer, autoBuffer.Get(cbs, 0, size).Value, 0, IndexType.Uint32);
             }
         }
 
--- a/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
+++ b/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
@@ -93,9 +93,5 @@ namespace Ryujinx.Graphics.Vulkan
         {
             return _buffer == buffer;
         }
-
-        public void Dispose()
-        {
-        }
     }
 }
--- a/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
+++ b/Ryujinx.Graphics.Vulkan/IndexBufferState.cs
@@ -1,0 +1,101 @@
+﻿using Silk.NET.Vulkan;
+using System;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    internal struct IndexBufferState
+    {
+        public static IndexBufferState Null => new IndexBufferState(GAL.BufferHandle.Null, 0, 0);
+
+        private readonly int _offset;
+        private readonly int _size;
+        private readonly IndexType _type;
+
+        private readonly GAL.BufferHandle _handle;
+        private Auto<DisposableBuffer> _buffer;
+
+        public IndexBufferState(GAL.BufferHandle handle, int offset, int size, IndexType type)
+        {
+            _handle = handle;
+            _offset = offset;
+            _size = size;
+            _type = type;
+            _buffer = null;
+        }
+
+        public IndexBufferState(GAL.BufferHandle handle, int offset, int size)
+        {
+            _handle = handle;
+            _offset = offset;
+            _size = size;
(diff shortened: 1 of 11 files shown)


Ryujinx.Graphics.Vulkan/PipelineBase.cs
```diff
@@ -505,29 +567,15 @@ public void SetIndexBuffer(BufferRange buffer, GAL.IndexType type)
             if (buffer.Handle != BufferHandle.Null)
             {
                 Auto<DisposableBuffer> ib = null;
```
gdkchan, line 569: Seems unused now.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -517,6 +517,24 @@ private unsafe void PrintGpuInformation()
             Logger.Notice.Print(LogClass.Gpu, $"{GpuVendor} {GpuRenderer} ({GpuVersion})");
         }
 
+        public GAL.PrimitiveTopology TopologyRemap(GAL.PrimitiveTopology topology)
```
gdkchan, line 520: I'm a little confused with this, it seems that the function on `EnumConversion` that returns a VkPrimitiveTopology will already return Triangles for Quads. I guess `EnumConversion` does not need to do that anymore since it already swaps the topology here? Also it seems that QuadStrip is not handled here.

Ryujinx.Graphics.Vulkan/IndexBufferState.cs
```diff
@@ -0,0 +1,101 @@
...
+                gd.Api.CmdBindIndexBuffer(cbs.CommandBuffer, autoBuffer.Get(cbs, 0, size).Value, 0, IndexType.Uint32);
+            }
+        }
+
+        public bool BoundEquals(Auto<DisposableBuffer> buffer)
+        {
+            return _buffer == buffer;
+        }
+
+        public void Dispose()
```
gdkchan, line 97: Why is that needed? It doesn't looks like this implements `IDisposable` and the function is just empty.

riperiperi: Leftover from `BufferState`. This one only stores a handle, so it doesn't need it anymore. I just forgot to remove it.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
riperiperi: The main difference is that this one is able to check device capabilities, which could prove useful. Should `GAL.PrimitiveTopology.Quads` and `GAL.PrimitiveTopology.QuadStrip` throw if this function is used to replace it?

riperiperi: I guess it could also fall through to `LogInvalidAndReturn`.

gdkchan: > Should GAL.PrimitiveTopology.Quads and GAL.PrimitiveTopology.QuadStrip throw in EnumConversion if this function is used to replace it?

I guess so, it's "cleaner" to handle it on a single place. It probably makes sense for those conversion functions to always return the equivalent Vulkan function, rather than making assumptions about what the caller will do for unsupported types. I think right now the default behavior for unhandled values is returning some default, rather than throwing, so I guess this one would be a bit different from the others, but I do think that throwing here for quads makes sense.

gdkchan: > I guess it could also fall through to LogInvalidAndReturn.

Yea, this is fine.

gdkchan:
It does seems to be a little bit faster on Intel.
Master:
image
PR:
image

gdkchan (approved):
lgtm, thanks. Nice to see this optimization, I didn't notice a difference on Fast RMX with NVIDIA, but Intel seems to benefit from it and maybe AMD will too.

Thog (approved):

MetrosexualGarbodor:
Tested all Xenoblades for some time each and found no discernible differences. Performance is also the same.

MarkLanders777:
> Tested all Xenoblades for some time each (on Nvidia) and found no discernible differences. Performance is also the same.

It was a little more smooth experience for me in Xenoblade 3, some animations are a little faster for example when you are using arts (the ones with soulhacker) it feels more faster than before (RTX NVIDIA 2060)

PR #3719, opened 2022-09-21
Vulkan: Zero blend state when disabled or write mask is 0

riperiperi:
Any difference in the blend state when blend is disabled is meaningless, but Ryujinx would compare different disabled blends and compile them as separate pipelines. This change ensures that all pipelines where blend state is meaningless record it as zero, which avoids compiling a bunch of pipelines that are essentially identical.

The NVIDIA driver is pretty forgiving when it comes to silly pipeline misses like this, but other drivers don't offer the same level of kindness.

This should reduce stuttering on those drivers, and might improve overall performance very slightly due to less pipeline variants being in the hash table.

Just needs tests to ensure nothing totally broke. Performance difference will probably be impossible to measure on NVIDIA.


gdkchan (approved):
lgtm, thanks. Nice to see the amount of VkPipelines reduced. Maybe we should look into other cases of state changes being redundant in the future.

Thog (approved):

PR #2194, opened 2021-04-10
Miria: The Death of OpenTK 3

Thog:
As a preparation for the new UI system and for Vulkan, this PR removes OpenTK 3 dependency entirely.

To do so, a new input system was implemented for keyboard and controller handling. On the controller side, SDL2 was chosen and handling of controllers was completely revisited.

With this new input system the following new features are now available:
- Controllers can now be hot-plugged without any constraints. No need to click any "refresh" buttons.
- Motion is supported without the need of external tools on supported controllers. (Legacy Cemuhook integration is still present if desired)
- Rumble API is exposed **while still not in use for now** (outside the scope of this PR)
- DS4, DS5, Pro Controller and Single Joycons are natively supported. (Joycon pairs are not yet natively supported and outside the scope of this PR; Cemuhook integration can be used for joycon pairs in the meantime.)
- GTK3 keyboard introduces support for different regional layouts.
- When a gamepad isn't supported by SDL2, a user can extend compatibility and mappings via the addition of a ``SDL_GameControllerDB.txt`` file in the same directory as ``Ryujinx.exe``. See [SDL_GameControllerDB](https://github.com/gabomdq/SDL_GameControllerDB) for community efforts.

As the nature of these changes are highly incompatible with (and impossible to migrate from) the old system, this will reset input configurations.

Changelog:
- Switch to OpenTK 4 for OpenAL implementation and mark OpenAL implementation as obsolete.
- Switch to OpenTK 4 for the Ryujinx.Graphics.OpenGL project.
- Use [SPB](https://github.com/Thog/SPB) for context creation and manipulations instead of OpenTK 3.
   - Currently only WGL and GLX backend are supported. (no regressions compared to old implementation)
- Switch to SDL2 for controller input.
- Switch to GTK3 for keyboard input.
- Individual keyboard support was removed.

Fixes #1465, Fixes #1098, Fixes #2071, Fixes #1978

PR #3720, opened 2022-09-22
Fix phantom configured controllers

mageven:
`ReloadConfiguration` and `RefreshInputConfigForHLE` pass the unfiltered InputConfig down the chain even when the mapped host controller is invalid (missing, etc..). Bug probably introduced with #2194 and #2269. 

Admittedly I don't understand Miria much, so I've opted for a shallow fix. Let me know if there's a better way or I missed something.

### DIFF
--- a/Ryujinx.Input/HLE/NpadManager.cs
+++ b/Ryujinx.Input/HLE/NpadManager.cs
@@ -51,7 +51,16 @@ namespace Ryujinx.Input.HLE
         {
             lock (_lock)
             {
-                _device.Hid.RefreshInputConfig(_inputConfig);
+                List<InputConfig> validInputs = new List<InputConfig>();
+                foreach (var inputConfigEntry in _inputConfig)
+                {
+                    if (_controllers[(int)inputConfigEntry.PlayerIndex] != null)
+                    {
+                        validInputs.Add(inputConfigEntry);
+                    }
+                }
+
+                _device.Hid.RefreshInputConfig(validInputs);
             }
         }
 
@@ -103,6 +112,8 @@ namespace Ryujinx.Input.HLE
                     _controllers[i] = null;
                 }
 
+                List<InputConfig> validInputs = new List<InputConfig>();
+
                 foreach (InputConfig inputConfigEntry in inputConfig)
                 {
                     NpadController controller = new NpadController(_cemuHookClient);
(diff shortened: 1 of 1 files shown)


mageven:
Thinking about it more, this change is slightly wider reaching than I intended, although I think for the better. 

Now physically connecting/disconnecting controllers will immediately reconfigure the assigned player on the guest side. So, essentially a "P2 has joined" when the corresponding controller for P2 is connected and vice versa. I think this flow is closer to how it works natively. Only difference being right now the player index is fixed (in the input settings) but on hardware it is dynamically assigned with help from controller applet.

A few users might be annoyed if their controller has connectivity issues (bad signal or flimsy USB connector), the game or controller applet will pester them to fix it. Before this PR, controller disconnects would be invisible to the game as the player is never removed. This would've resulted in lost inputs without any notice so I think it's better this way.

**Note:** This fix only effects emulation side. Ryu's input configuration is persistent (for other good reasons). The GUI will still show once-configured-but-now-disconnected controllers as "disabled", just with this PR it won't effect the guest. Not sure if there's an issue already open for UI inconsistency.

MutantAura:
By "phantom configured controllers" here are does this fix the bug where the controller applet will pop-up seemingly with no controllers configured, or is this fixing the GUI bug where multiple of the same controller can sometimes appear? I think it's the former? I've added the GUI tag for now just to be safe.

mageven:
I've not touched SDL2 controller detection or input configuration GUI so it's definitely not the latter. Also, I wasn't aware of multiple copies of a controller appearing. Is that not an SDL2 specific quirk? Not sure how to filter them out.

> where the controller applet will pop-up seemingly with no controllers configured

As a side-effect, yes. Without this change, it could pop-up when the stale host controllers would still be mapped to players that the game doesn't like. Then the usual support advice given was to delete the entire `config.json`.

The title's "configured controllers" refers to log lines with the same name-
```
GUI.WindowThread Hid Configure: Configured Controller ProController to Player1
GUI.WindowThread Hid Configure: Configured Controller JoyconPair to Player2
```

marysaka (approved):

riperiperi (approved):
Nice one. I think the original service has all controllers attached, but then the controller applet decides which are actually "bound" via manual selection with L+R.

PR #3721, opened 2022-09-25
Do not clear the rejit queue when overlaps count is equal to 0.

LDj3SNuD:
From the log it is clear that there are about 700 calls in about 200 seconds to `Translator.InvalidateJitCacheRegion()` via `KPageTableBase.UnmapPhysicalMemory()` without overlapping; this could cause one or more guest functions to miss the chance to be retranslated in high quality:
[clear_rejit_queue_no_thanks.log](https://github.com/Ryujinx/Ryujinx/files/9639893/clear_rejit_queue_no_thanks.log)


gdkchan (approved):
Nice catch, but it would be nice to also include an explanation of why PPTC version needs to be changed (I assume that it is only saving information for HighCq functions and existing profiles might be missing some functions permanently due to this?)

LDj3SNuD:
> 

Both Ptc and the Profiler are able to replace an LQ function with an HQ, but any overlapping LQs (some will still be needed) would be dead weight. Nothing nonfunctional if I reverse the last commit.

LDj3SNuD:
Considering that we are also invalidating the profiling .info files, I would recommend tweaking the value of this constant https://github.com/Ryujinx/Ryujinx/blob/master/ARMeilleure/Decoders/Decoder.cs#L16, which was lowered to speed up PPTC https://github.com/Ryujinx/Ryujinx/commit/5e724cf24e3d696b95be859c055a617e5d37bf80#diff-82cd0b0f0bc66b6bc9dddb35b31c72707554587253f7a5de9f4602766383c6ba (there was no POWER PR yet), to an intermediate value of `3750`; the rebuilding times of the .cache files will be longer, but you will get all the other benefits.

AcK77 (approved):

PR #3722, opened 2022-09-25
 Avalonia: Update Polish Translation

LunaDook:
Adding the new shortened "volume" string from the recent updates to Avalonia.

You need the period there otherwise it could be read as "Głoś" -> "Preach".


LunaDook:
Also changing the name of the language in the dropdown to be in Polish, in line with the others listed.

LunaDook:
Realizing that period isn't necessary considering the string's usage (which to be fair, I should have checked when I added it). Removed it to allow things to look better.

LunaDook:
Adding the new strings relating to the volume hotkeys.

gdkchan (approved):
Seems reasonable.

LunaDook:
Also adding the string added as part of the updater fixes.

AcK77 (approved):

PR #3723, opened 2022-09-25
Vulkan: Implement multisample <-> non-multisample copies and depth-stencil resolve

gdkchan:
This has two changes. The first one "ports" the change on #3596 to Vulkan. It did not support copying non-multisample to multisample textures before, and would crash with a device lost error on games that does that (like Fate Extella: The Umbral Star). The second one adds support for depth-stencil resolve. `vkCmdResolveImage` does not support depth-stencil textures for some reason, so we need to create and begin/end a render pass for that. Before this change, games doing depth-stencil resolve would crash on Intel (and maybe other GPUs/drivers, but on NVIDIA it was working) before getting ingame. With this change, the game works.

I also changed the texture compatibility rules to forbid making depth-stencil multisample <-> non-multisample textures copy compatible, since neither Vulkan nor OpenGL supports this. Fate Extella was triggering it, so it's something to keep in mind.

Fate Extella: The Umbral Star now works on Vulkan (tested on NVIDIA and Intel):
image
image
Sonic Colors Ultimate also works on Intel Vulkan (NVIDIA was already working so I didn't test there):
image

~~One thing that I don't like on this change is that I had to include 5 copies of the compute shader for the multisample copy, where the only thing that changes is the image format. I'm not sure if there's a way around this, maybe SPIR-V shader specialization could help here, but i haven't looked into it. Other options are finding and patching the relevant byte(s) on the SPIR-V binary, or using the image without format capability to omit the format entirely, but that's not supported by all drivers.~~ No longer an issue thanks to mageven.


riperiperi (approved):
LGTM! The specialization stuff could definitely come in handy for future builtin shaders.

PR #3728, opened 2022-09-29
Fix incorrect tessellation inputs/outputs

gdkchan:
This is mainly fixing a wrong assumption that I made about tessellation per-patch input/output buffer. I assumed that it was the same as the regular attribute buffer, but it isn't really. The only offsets from the per-patch buffer that have a special meaning are the ones < 0x18. Those map to the TessOuterLevel and TessInnerLevel built-ins. The remaining space can be used for anything, so it's just generic storage. So this PR is changing the code including the GLSL and SPIR-V backends to account for this.

In addition to that, it fixes some existing issues that affects SPIR-V/Vulkan:
- On SPIR-V, the `LoadOutputMask` flag was being ignored for `LoadAttribute`, which could cause it to read from a input rather than output variable.
- `gl_TessCoord` was being declared with a vec2 type on SPIR-V, rather than a vec3 type.
- SPIR-V was using locations for per-patch inputs/outputs that overlaps with the regular inputs/outputs. This was not causing validation errors but it seems to cause issues on some drivers. Both GLSL and SPIR-V were changed to "allocate" locations from the unused locations, so that they never overlap.
- The winding order on the tessellation evaluation shader is now inverted when targeting Vulkan. This is required because it is also inverted on the backend (see `EnumConversion.Convert(this GAL.FrontFace frontFace)` on `Ryujinx.Graphics.Vulkan`).

This fixes rendering on The Legend of Heroes: Trails from Zero.
Before:
image
image
After:
image
image
This affects both OpenGL and Vulkan, but the "before" images would be different from OpenGL (just a black background with text). That's because the shader fails to compile, while on Vulkan the behavior depends on the driver.

The change has also been tested on Intel (Windows). While it no longer crashes on Intel, tessellation is still not working correctly. I also tested GLSL + Vulkan on Intel, and it was also rendering incorrectly, so I don't think the SPIR-V code generation is the issue.

It might potentially affect The Witcher 3 on Vulkan, which I believe had missing ground or something like that on Vulkan? I don't have a save on that area to test.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 3697;
+        private const uint CodeGenVersion = 3728;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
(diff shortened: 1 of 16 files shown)


Thog (approved):

riperiperi (approved):
LGTM, nice one!

PR #3730, opened 2022-10-01
Avoid allocating unmanaged string per shader

mageven:
Implemented in a fun way for zero allocations (like C string literals). If this is too hacky, I'll replace it with a static `AllocHGlobal`, it should clean up properly without a free. 

Not sure why there's a condition for disposing but I kept it (afaik it's not even referenced outside of `ShaderCollection`). Let me know if it can be removed too. 

Also added missing `IDisposable` for clarity.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/Shader.cs
+++ b/Ryujinx.Graphics.Vulkan/Shader.cs
@@ -15,16 +15,7 @@ namespace Ryujinx.Graphics.Vulkan
         // Take this lock when using them.
         private static object _shaderOptionsLock = new object();
 
-        private static ReadOnlySpan<byte> MainEntryPointName => new byte[] { (byte)'m', (byte)'a', (byte)'i', (byte)'n', 0 };
-        private static readonly IntPtr PtrMainEntryPointName;
-
-        unsafe static Shader()
-        {
-            fixed (byte* ptr = MainEntryPointName)
-            {
-                PtrMainEntryPointName = (IntPtr)ptr;
-            }
-        }
+        private static readonly IntPtr _ptrMainEntryPointName = Marshal.StringToHGlobalAnsi("main");
 
         private readonly Vk _api;
         private readonly Device _device;
@@ -156,7 +147,7 @@ namespace Ryujinx.Graphics.Vulkan
                 SType = StructureType.PipelineShaderStageCreateInfo,
                 Stage = _stage,
                 Module = _module,
-                PName = (byte*)PtrMainEntryPointName
+                PName = (byte*)_ptrMainEntryPointName
             };
         }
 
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Vulkan/Shader.cs
```diff
@@ -9,17 +9,28 @@
...
 namespace Ryujinx.Graphics.Vulkan
 {
-    class Shader
+    class Shader : IDisposable
     {
         // The shaderc.net dependency's Options constructor and dispose are not thread safe.
         // Take this lock when using them.
         private static object _shaderOptionsLock = new object();
 
+        private static ReadOnlySpan<byte> MainEntryPointName => new byte[] { (byte)'m', (byte)'a', (byte)'i', (byte)'n', 0 };
```
gdkchan, line 18: Private static fields are named starting with `_` and first character is lower case, like the name above.

Ryujinx.Graphics.Vulkan/Shader.cs
```diff
@@ -9,17 +9,28 @@
...
+        private static readonly IntPtr PtrMainEntryPointName;
+
+        unsafe static Shader()
+        {
+            fixed (byte* ptr = MainEntryPointName)
```
gdkchan, line 23: This feels a bit dangerous since the pointer is used before the pinned region. I think it would be better to use the safer method.

mageven: Ideally the Span should just point to static data in the executable. Ok, I will use the safe method.

mageven: What about the condition in dispose? Was it a leftover?

gdkchan: > What about the condition in dispose? Was it a leftover?

I'm not sure if I understand the question. The old check on dispose was to avoid double free in case dispose is called more than once.

gdkchan (approved):
I think it would be better if `_active` was called `_disposed` instead (and the logic would be the inverse, `if (!_disposed)` and `_disposed = true;`). Other than that, LGTM, thanks!

marysaka (approved):

Ryujinx.Graphics.Vulkan/Shader.cs
```diff
@@ -9,17 +9,19 @@
...
 namespace Ryujinx.Graphics.Vulkan
 {
-    class Shader
+    class Shader : IDisposable
     {
         // The shaderc.net dependency's Options constructor and dispose are not thread safe.
         // Take this lock when using them.
         private static object _shaderOptionsLock = new object();
 
+        private static readonly IntPtr _ptrMainEntryPointName = Marshal.StringToHGlobalAnsi("main");
```
marysaka, line 18: Technically the way it is written, this will leak... I don't think this have any impact considering we need it at any point in time

marysaka (approved):
lgtm

Ryujinx.Graphics.Vulkan/Shader.cs
gdkchan: Should not leak since its a static field, it will be allocated only once for the process lifetime.

PR #3731, opened 2022-10-01
Vulkan: Fix buffer texture storage not being updated on buffer handle reuse

gdkchan:
Because buffer handles can be re-used (when the old buffer is deleted and a new one created), it is possible that the same handle will actually refer to different buffers. So skipping the update if the handle is the same as the old one is not correct. To fix this, it also forces an update if new buffers were created since the last update. We're currently doing the same thing on OpenGL. This was not always a problem on Vulkan, as buffer handles were not reused, but at some point I added the `IdList` which allows reuse of handles.

This is a tentative fix for the weird issue were models constantly swap back to a very old animation frame on UE4 games. It's hard to confirm if this change actually fixes the issue due to its random nature though. Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/BufferManager.cs
+++ b/Ryujinx.Graphics.Vulkan/BufferManager.cs
@@ -38,6 +38,8 @@ namespace Ryujinx.Graphics.Vulkan
 
         private readonly IdList<BufferHolder> _buffers;
 
+        public int BufferCount { get; private set; }
+
         public StagingBuffer StagingBuffer { get; }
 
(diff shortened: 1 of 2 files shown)


marysaka (approved):

riperiperi (approved):
Nice catch! Weird that this was even manifesting through the Threaded GAL, since that has its own set of buffer handles.

gdkchan:
> Nice catch! Weird that this was even manifesting through the Threaded GAL, since that has its own set of buffer handles.

I don't think this changes anything as the underlying handles from the backend can still have the same value (due to reuse on the backend), so the problem would still exist.

PR #3732, opened 2022-10-01
Fix shader SULD (bindless) instruction using wrong register as handle

gdkchan:
If we don't have textures, we don't need the scale helpers. This was causing some shaders to fail compilation because in those cases it does not define the support block, so `s_render_scale` would be undefined. This some shaders failing to compile on Sea of Solitude on OpenGL. Bindless elimination fails on this case, which is what causes it to have no textures but having the `IntegerSampling` flag set. ~~The fact that bindless elimination fails is a separate issue, and is likely causing other glitches.~~

I found that the bindless elimination failing was due to another bug, it was always using register 0 as Rc for the SULD (bindless) instruction rather than the actual Rc value encoded on the instruction. I also found that the encoding for all the surface instruction variants was slightly incorrect (it has some bits set as "don't care" when they should have a value set. I think this was not causing issues in pratice though), so I fixed that aswell.

This fixes the regression of Sea of Solitude which might have been from #2698.

Before (vertex explosions):
image
After:
image

Also fixes Shadowrun Returns.

Before:
image
After:
image


marysaka (approved):
nice catch!

riperiperi (approved):

PR #3736, opened 2022-10-02
Support use of buffer ranges with size 0

### LINKED ISSUE #3734: IndexOutOfRangeException from Vulkan in Fire Emblem Warriors Three Hopes
### What's the issue you have encountered?

Game ID: 010071F0143EA000
Game Title: Fire Emblem Warriors: Three Hopes
The game just crash after the video of Prologue of Blue House is played / skipped, with IndexOutOfRangeException from Vulkan.

Normal playthrough the story using with Blue Fraction, and playthrough to Prologue.

### Environment?

- Ryujinx version: 1.1.283
- Game version: 1.0.1
- System Specs:
  - OS: Windows 11
  - CPU: Intel i5-12600K
  - GPU: Nvidia RTX 2070 Super
  - RAM: 32 GB
- Applied Mods: _N/A_

[Ryujinx_1.1.283_2022-10-02_08-05-47.log](https://github.com/Ryujinx/Ryujinx/files/9691434/Ryujinx_1.1.283_2022-10-02_08-05-47.log)

gdkchan:
`BufferUsageBitmap` calculates a inclusive end address as `offset + size - 1`, however, if size is 0, it will result in a invalid offset that is less than the start. `BitMap.SetRange` is not prepared to handle this and will throw `IndexOutOfRangeException` if such ranges are passed. This started happening after #3715 because some games draws quads with a vertex count of 0. The fix here explicitly returns in `BufferUsageBitmap` when the size is 0 since there's nothing to be done in this case.

Fixes #3734.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/BufferUsageBitmap.cs
+++ b/Ryujinx.Graphics.Vulkan/BufferUsageBitmap.cs
@@ -24,6 +24,11 @@
 
         public void Add(int cbIndex, int offset, int size)
         {
+            if (size == 0)
+            {
+                return;
+            }
+
             // Some usages can be out of bounds (vertex buffer on amd), so bound if necessary.
             if (offset + size > _size)
             {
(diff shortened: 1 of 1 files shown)


marysaka (approved):

riperiperi (approved):

gdkchan:
I think there is another possible bug here, if size is 0 on the first quads draw, then presumably `expectedSize <= _currentSize` will be true, and it might return `BufferHandle.Null` since it will not have been initialized. I will give that case a try later and see if it breaks.

PR #3738, opened 2022-10-03
Vulkan: Dispose TextureStorage when views hit 0 instead of immediately

riperiperi:
Due to the `using` statement being scoped to the `CreateTextureView` method, `TextureStorage` would be disposed as soon as the view was returned.

This was largely fine as the `TextureStorage` resources were being kept alive by the views holding their own references to them, but it also meant that dispose is only called as soon as the texture is created.

Aliased Storages are `TextureStorages` created with the same allocation as another `TextureStorage`, if they have to be aliased as another format. We keep track of a TextureStorage's `_aliasedStorages` as they are created, and dispose them when the `TextureStorage` is disposed...

...except it is disposed immediately, before any aliased storages are even created. The aliased storages added after this will never be disposed, and the texture's allocation will never be freed.

---

This PR attempts to fix this by disposing `TextureStorage` when its view count reaches 0. The other use of texture storage - the D32S8 blit - still manually disposes the storage, but regular uses created via the GAL are now disposed by the view count.

I think this makes the most sense, as otherwise in the future this behaviour might be forgotton and more things could be added to the `Dispose()` method that don't work due to it not actually calling at the right time.

This should improve memory leaks in Super Mario Odyssey, most noticeable when resolution scaling. The memory usage of the game is still wildly unpredictable due to how it interacts with the texture cache, but now it shouldn't get considerably worse as you play... I hope. I've seen it typically recover back to the same level often, though it can spike significantly.

Might improve memory usage on other games, or GPUs that need to alias textures more often.

Please test a bunch of games on multiple GPUs to make sure this doesn't break anything.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/TextureStorage.cs
+++ b/Ryujinx.Graphics.Vulkan/TextureStorage.cs
@@ -480,6 +480,8 @@ namespace Ryujinx.Graphics.Vulkan
             if (--_viewsCount == 0)
             {
                 _gd.PipelineInternal?.FlushCommandsIfWeightExceeding(_imageAuto, _size);
+
+                Dispose();
             }
         }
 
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
Nice catch, making the dispose behave as expected should avoid someone tripping up in the future.

AcK77 (approved):

PR #3739, opened 2022-10-04
Fix memory corruption in BCAT and FS Read methods when buffer is larger than needed

gdkchan:
Right now, the FS Read methods allocates an array without any regard for existing memory contents, reads the data from the file (or directory entries for `IDirectory.Read`) and then writes the buffer in memory. In most cases this does not cause issues because the application  usually doesn't care about the existing memory contents, but in some cases, applications might pass a buffer that is large and exposes other data to the service. In those cases, since it allocates a new array for the buffer and does not read existing memory contents, such values would be zeroed.

This was causing a crash on the title screen of SWORD ART ONLINE: Alicization Lycoris, because part of the buffer used on `IDirectory.Read` contained data for a `FiberType` struct, and since the method would zero out the struct, the game would skip a step where it frees the fiber and removes it from a internal list. Since the fiber was not removed from the list, it would later try to use it and cause a crash because it had already exited and was not usable anymore.

The title mentioned above can now go ingame:
image

This change uses `GetWritableRegion`, which will automatically copy the existing memory contents if needed, and just return a span of the memory directly if possible, avoiding the issue.

The change on `IFile.Read` is just to be on the safe side, but SWORD ART ONLINE seems to work even without that change.

Testing is welcome.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IDirectory.cs
+++ b/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IDirectory.cs
@@ -17,17 +17,17 @@ namespace Ryujinx.HLE.HOS.Services.Fs.FileSystemProxy
         // Read() -> (u64 count, buffer<nn::fssrv::sf::IDirectoryEntry, 6, 0> entries)
         public ResultCode Read(ServiceCtx context)
         {
-            ulong bufferPosition = context.Request.ReceiveBuff[0].Position;
+            ulong bufferAddress = context.Request.ReceiveBuff[0].Position;
             ulong bufferLen = context.Request.ReceiveBuff[0].Size;
 
-            byte[] entryBuffer = new byte[bufferLen];
+            using (var region = context.Memory.GetWritableRegion(bufferAddress, (int)bufferLen))
+            {
+                Result result = _baseDirectory.Get.Read(out long entriesRead, new OutBuffer(region.Memory.Span));
 
-            Result result = _baseDirectory.Get.Read(out long entriesRead, new OutBuffer(entryBuffer));
+                context.ResponseData.Write(entriesRead);
 
-            context.Memory.Write(bufferPosition, entryBuffer);
-            context.ResponseData.Write(entriesRead);
-
-            return (ResultCode)result.Value;
+                return (ResultCode)result.Value;
+            }
         }
 
         [CommandHipc(1)]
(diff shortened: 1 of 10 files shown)


marysaka (approved):

AcK77 (approved):

Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IDirectory.cs
```diff
@@ -17,17 +17,17 @@ public IDirectory(ref SharedRef<LibHac.FsSrv.Sf.IDirectory> directory)
...
         public ResultCode Read(ServiceCtx context)
         {
-            ulong bufferPosition = context.Request.ReceiveBuff[0].Position;
+            ulong bufferAddress = context.Request.ReceiveBuff[0].Position;
             ulong bufferLen = context.Request.ReceiveBuff[0].Size;
 
-            byte[] entryBuffer = new byte[bufferLen];
-
-            Result result = _baseDirectory.Get.Read(out long entriesRead, new OutBuffer(entryBuffer));
+            using (var region = context.Memory.GetWritableRegion(bufferAddress, (int)bufferLen))
```
riperiperi, line 23: This writable region needs to be tracked, as the Write is is replacing was tracked.

Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFile.cs
```diff
@@ -19,23 +19,23 @@ public IFile(ref SharedRef<LibHac.FsSrv.Sf.IFile> baseFile)
...
 
             long offset = context.RequestData.ReadInt64();
             long size   = context.RequestData.ReadInt64();
 
-            byte[] data = new byte[context.Request.ReceiveBuff[0].Size];
-
-            Result result = _baseFile.Get.Read(out long bytesRead, offset, new OutBuffer(data), size, readOption);
-
-            context.Memory.Write(position, data);
+            using (var region = context.Memory.GetWritableRegion(bufferAddress, (int)bufferLen))
```
riperiperi, line 31: Same as above.

riperiperi:
Is IStorage immune to this issue? The code seems similar to IFile and IDirectory.

Thealexbarney:
Yeah, I was gonna say this should also probably be applied there

Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IDirectory.cs
gdkchan: This write happens on the HLE service `AddressSpaceManager` which does not support write tracking right now, but I will change it.

PR #3741, opened 2022-10-05
Change NvMap ID allocation to match nvservices

gdkchan:
This changes the NvMap ID allocation to match what the official service does. The official service uses an ever incrementing ID (which seems very odd to me, since it eventually can overflow, but it would probably take long enough to happen that no one will experience it in practice unless they leave a game running for months/years and it constantly create nvmaps). It also increments the ID by 4 rather than 1 which is also weird (and would make it overflow even faster...)

This fixes a random crash that could happen when starting a new game on Animal Crossing: New Horizons. A similar issue on other titles might be also fixed.

Testing is welcome.


marysaka (approved):

MutantAura:
Seems to also resolve the random crashes exhibited in Breath of the Wild. 
Horse auto-riding for over an hour and nothing yet: 
image

Leischii:
Fixes the following things in Animal Crossing New Horizons:
- All crashes! All previous ways of forcing a crash don't work anymore even after multiple attemps
- Texture corruptions. And this is not only talking about the ones happening with the force crash. This also includes:
-- "Black void" corruption
-- Randomly misplaced dust particles while moving
-- Particle corruption when placing/destroying dirt with the builder tool
-- Island previews missing/corrupted square-shaped parts when choosing your island 

TLDR: This fixes afaik every single bug left in the game. This is just amazing <3

MetrosexualGarbodor:
After playing 40 mind-numbing minutes of Pokémon Sword and going through pokecenter doors for 5 of them, it's possible this might have fixed random crashes in Sword/Shield as well. Here's a Sword save in case anyone is willing to test further.

[pkmn sword before entering town crash.zip](https://github.com/Ryujinx/Ryujinx/files/9718355/pkmn.sword.before.entering.town.crash.zip)

riperiperi:
I can confirm that all of these are now fixed for me too. Excellent sleuthing!

riperiperi (approved):

PR #3743, opened 2022-10-06
Vulkan: Fix some issues with CacheByRange

riperiperi:
- Cache now clears under more circumstances, the most important being the fast path write.
- Cache supports partial clear which should help when more buffers join.
- Fixed an issue with I8->I16 conversion where it wouldn't register the buffer for use on dispose.

Should hopefully fix issues with https://github.com/Ryujinx/Ryujinx-Games-List/issues/4010 and maybe others.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/CacheByRange.cs
+++ b/Ryujinx.Graphics.Vulkan/CacheByRange.cs
@@ -25,6 +25,11 @@ namespace Ryujinx.Graphics.Vulkan
             return other is I8ToI16CacheKey;
         }
 
+        public void SetBuffer(Auto<DisposableBuffer> buffer)
+        {
+            _buffer = buffer;
+        }
+
         public void Dispose()
         {
             _gd.PipelineInternal.DirtyIndexBuffer(_buffer);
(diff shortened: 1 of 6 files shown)


Ryujinx.Graphics.Vulkan/CacheByRange.cs
```diff
@@ -160,6 +165,34 @@ public void Clear()
             }
         }
 
+        public void ClearRange(int offset, int size)
+        {
+            if (_ranges != null && _ranges.Count > 0)
+            {
+                int end = offset + size;
+
+                foreach (ulong range in _ranges.Keys)
```
gdkchan, line 174: Might be more efficient to iterate both key and values here rather than just the keys to avoid the lookup on `var entries = _ranges[range];`.

Ryujinx.Graphics.Vulkan/CacheByRange.cs
```diff
@@ -184,6 +217,11 @@ private static ulong PackRange(int offset, int size)
             return (uint)offset | ((ulong)size << 32);
```
gdkchan, line 217: Maybe this should be on a new `PackRange` methods that takes 2 ints and returns a ulong, since there's one for unpack.

riperiperi: What do you mean? This is already a method with those input/outputs.

gdkchan: Ah right, sorry. GitHub collapsed the method so I thought that code was part of `GetEntries`.

gdkchan:
The problem is that `_gd` is null on the struct, not the buffer itself. It's using the parameterless struct constructor rather than the one were you can pass the renderer instance.

gdkchan:
Fixing the above issue prevents the NRE, but for some reason the background is upside down on XC3 title screen with the index type uint8 conversion:
image

With native uint8 support:
image

---

Aslo tested eBASEBALL and it renders properly again.
image

riperiperi:
Xenoblade 3 should be fixed. It does reconvert a bunch of buffers every frame now, though.

marysaka (approved):

gdkchan (approved):
LGTM, thanks.

PR #3745, opened 2022-10-07
GPU: Pass SpanOrArray for Texture SetData to avoid copy

riperiperi:
Texture data is often converted before upload, meaning that an array was allocated to perform the conversion into. However, the backend SetData methods were being passed a Span of that data, and the Multithreaded layer does `ToArray()` on it so that it can be stored for later! This method can't extract the original array, so it creates a copy.

This PR changes the type passed for textures to a new ref struct called SpanOrArray, which is backed by either a ReadOnlySpan or an array. The benefit here is that we can have a ToArray method that doesn't copy if it is originally backed by an array.

This will also avoid a copy when running the ASTC decoder.

On NieR this was taking 38% of texture upload time, which it does a _lot_ of when you move between areas, so there should be a 1.6x performance boost when strictly uploading textures. No doubt this will also improve texture streaming performance in UE4 games, and maybe a small improvement with video playback.

Note that NieR is slow for some other reasons in general, the texture uploads just add to that.

From the numbers, it's probably possible to improve the upload rate by a further 1.6x by performing layout conversion on GPU. I'm not sure if we could improve it further than that - multithreading conversion on CPU would probably result in memory bottleneck.

This doesn't extend to buffers, since we don't convert their data on the GPU emulator side.

### DIFF
--- a/Ryujinx.Common/Memory/SpanOrArray.cs
+++ b/Ryujinx.Common/Memory/SpanOrArray.cs
@@ -1,0 +1,98 @@
+﻿using System;
+
+namespace Ryujinx.Common.Memory
+{
+    /// <summary>
+    /// A struct that can represent both a Span and Array.
+    /// This is useful to keep the Array representation when possible to avoid copies.
+    /// </summary>
+    /// <typeparam name="T">Element Type</typeparam>
+    public ref struct SpanOrArray<T> where T : unmanaged
+    {
+        public readonly T[] Array;
+        public readonly ReadOnlySpan<T> Span;
+
+        /// <summary>
+        /// Create a new SpanOrArray from an array.
+        /// </summary>
+        /// <param name="array">Array to store</param>
+        public SpanOrArray(T[] array)
+        {
+            Array = array;
+            Span = ReadOnlySpan<T>.Empty;
+        }
+
+        /// <summary>
+        /// Create a new SpanOrArray from a readonly span.
+        /// </summary>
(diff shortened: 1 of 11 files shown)


Ryujinx.Common/Memory/SpanOrArray.cs
```diff
@@ -0,0 +1,98 @@
...
+        public static implicit operator SpanOrArray<T>(Span<T> span)
+        {
+            return new SpanOrArray<T>(span);
+        }
+
+        /// <summary>
+        /// Cast a SpanOrArray to an array.
+        /// </summary>
+        /// <param name="spanOrArray">Source SpanOrArray</param>
+        public static implicit operator T[](SpanOrArray<T> spanOrArray)
```
gdkchan, line 84: This one is odd since it does not actually perform a cast in some cases.

gdkchan (approved):
LGTM except for a single comment, thanks.

Ryujinx.Common/Memory/SpanOrArray.cs
riperiperi: Should I remove it entirely in favor of explicitly using `ToArray` for this?

gdkchan: Yea, that makes more sense imo, I wouldn't expect a implicit cast to do an allocation.

marysaka (approved):

PR #3754, opened 2022-10-11
Add short duration texture cache

riperiperi:
Time for obscure edge case performance improvements.

This texture cache takes textures that lose their last pool reference and keeps them alive until the next frame, or until an incompatible overlap removes it. This is done since under certain circumstances, a texture's reference can be wiped from a pool despite it still being in use - though typically the reference will return when rendering the next frame.

This texture cache maintains a dictionary of TextureDescriptor to texture so that textures can be easily rediscovered if they are added to the pool shortly after.

While this may slightly increase texture memory usage when quickly going through a bunch of temporary textures, it's still bounded due to the overlap removal rule. This is an alternative to replacing the pool method of texture liveness checks entirely, as that could cause larger issues.

This greatly increases performance in Hyrule Warriors: Age of Calamity, which would slow to a crawl when too much happened at once. It may positively affect some UE4 games which dip framerate severely under certain circumstances.

### Hyrule Warriors: Age of Calamity

The game no longer becomes unplayable when dropping frames. This game used to run better, but that was before we started doing things correctly. This change does not reduce the accuracy of what is going on with the compute shaders and buffer flushes behind the scenes, it just brute forces past it.

#### Before (OGL)
https://cdn.discordapp.com/attachments/694459861190311957/1029424662523941024/before_ogl.mp4

#### After (OGL)
https://cdn.discordapp.com/attachments/694459861190311957/1029424661550878850/after_ogl.mp4

#### After (VK)
https://cdn.discordapp.com/attachments/694459861190311957/1029425850862534696/after_vk.mp4

This game typically performs better in Vulkan, but depending on your hardware you might need this branch to mitigate hardware issues (this branch's changes are already present):

https://github.com/riperiperi/Ryujinx/commits/feat/vk-buffer-migration

Maybe more comparisons to come. I would suggest trying areas in UE4 games that plummet FPS for no reason.

## Testing

Please test as many games as you can to verify that:
- No other games have worse performance
- No other games have texture issues
- No other games use an excessive amount of memory (at least, not more than they already do... looking at you FAST RMX)

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -169,18 +169,15 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <returns>The texture if found, null otherwise</returns>
         public Texture FindShortCache(in TextureDescriptor descriptor)
         {
-            if (_shortCacheLookup.Count > 0)
+            if (_shortCacheLookup.Count > 0 && _shortCacheLookup.TryGetValue(descriptor, out var entry))
             {
-                if (_shortCacheLookup.TryGetValue(descriptor, out var entry))
+                if (entry.InvalidatedSequence == entry.Texture.InvalidatedSequence)
                 {
-                    if (entry.InvalidatedSequence == entry.Texture.InvalidatedSequence)
-                    {
-                        return entry.Texture;
-                    }
-                    else
-                    {
-                        _shortCacheLookup.Remove(descriptor);
-                    }
+                    return entry.Texture;
+                }
+                else
+                {
+                    _shortCacheLookup.Remove(descriptor);
                 }
             }
 
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -16,6 +16,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         private readonly LinkedList<Texture> _textures;
         private readonly ConcurrentQueue<Texture> _deferredRemovals;
 
+        private HashSet<Texture> _shortCacheBuilder;
+        private HashSet<Texture> _shortCache;
+
         /// <summary>
         /// Creates a new instance of the automatic deletion cache.
         /// </summary>
@@ -23,6 +26,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             _textures = new LinkedList<Texture>();
             _deferredRemovals = new ConcurrentQueue<Texture>();
+
+            _shortCacheBuilder = new HashSet<Texture>();
+            _shortCache = new HashSet<Texture>();
         }
 
         /// <summary>
@@ -128,6 +134,57 @@ namespace Ryujinx.Graphics.Gpu.Image
         public void RemoveDeferred(Texture texture)
         {
             _deferredRemovals.Enqueue(texture);
+        }
+
+        /// <summary>
+        /// Removes a texture from the short duration cache.
+        /// </summary>
+        /// <param name="texture">Texture to remove from the short cache</param>
+        public void RemoveShortCache(Texture texture)
+        {
+            bool removed = _shortCache.Remove(texture);
+            removed |= _shortCacheBuilder.Remove(texture);
+
+            if (removed)
+            {
+                texture.DecrementReferenceCount();
+
+                texture.IsShortCached = false;
+            }
+        }
+
+        /// <summary>
+        /// Adds a texture to the short duration cache.
+        /// It starts in the builder set, and it is moved into the deletion set on next process.
+        /// </summary>
+        /// <param name="texture">Texture to add to the short cache</param>
+        public void AddShortCache(Texture texture)
+        {
+            _shortCacheBuilder.Add(texture);
+
+            texture.IsShortCached = true;
+
+            texture.IncrementReferenceCount();
+        }
+
+        /// <summary>
+        /// Delete textures from the short duration cache.
+        /// Moves the builder set to be deleted on next process.
+        /// </summary>
+        public void ProcessShortCache()
+        {
+            HashSet<Texture> toRemove = _shortCache;
+
+            foreach (var texture in toRemove)
+            {
+                texture.DecrementReferenceCount();
+
+                texture.IsShortCached = false;
+            }
+
+            toRemove.Clear();
+            _shortCache = _shortCacheBuilder;
+            _shortCacheBuilder = toRemove;
         }
 
         public IEnumerator<Texture> GetEnumerator()
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -5,6 +5,28 @@ using System.Collections.Generic;
 namespace Ryujinx.Graphics.Gpu.Image
 {
     /// <summary>
+    /// An entry on the short duration texture cache.
+    /// </summary>
+    class ShortTextureCacheEntry
+    {
+        public readonly TextureDescriptor Descriptor;
+        public readonly int InvalidatedSequence;
+        public readonly Texture Texture;
+
+        /// <summary>
+        /// Create a new entry on the short duration texture cache.
+        /// </summary>
+        /// <param name="descriptor">Last descriptor that referenced the texture</param>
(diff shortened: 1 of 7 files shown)


Ryujinx.Graphics.Gpu/Image/Pool.cs
```diff
@@ -94,6 +94,16 @@ public T2 GetDescriptor(int id)
             return ref MemoryMarshal.Cast<byte, T2>(PhysicalMemory.GetSpan(Address + (ulong)id * DescriptorSize, DescriptorSize))[0];
```
gdkchan, line 94: Maybe this can start calling the new method below to simplify it a bit? You'd only have to add the id offset to the address.

Ryujinx.Graphics.Gpu/Image/TextureDescriptor.cs
```diff
@@ -249,5 +250,24 @@ public bool Equals(ref TextureDescriptor other)
...
         }
+
+        /// <summary>
+        /// Check if two descriptors are equal.
+        /// </summary>
+        /// <param name="other">The descriptor to compare against</param>
+        /// <returns>True if they are equal, false otherwise</returns>
+        public bool Equals(TextureDescriptor other)
+        {
+            return Unsafe.As<TextureDescriptor, Vector256<byte>>(ref this).Equals(Unsafe.As<TextureDescriptor, Vector256<byte>>(ref other));
```
gdkchan, line 261: I think this could call the ref overload to avoid repeating this code here.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -50,6 +50,20 @@ public TexturePool(GpuContext context, GpuChannel channel, ulong address, int ma
 
             ref readonly TextureDescriptor descriptor = ref GetDescriptorRef(id);
 
+            if (texture == null)
```
gdkchan, line 53: This can be simplified by moving it on the null check below like so:
```cs
            if (texture == null)
            {
                texture = PhysicalMemory.TextureCache.FindShortCache(descriptor);

                if (texture == null)
                {
                    TextureInfo info = GetInfo(descriptor, out int layerSize);

                    ProcessDereferenceQueue();

                    texture = PhysicalMemory.TextureCache.FindOrCreateTexture(_channel.MemoryManager, TextureSearchFlags.ForSampler, info, layerSize);
                }
```
Would be cool to be able to do something like:
```cs
texture = PhysicalMemory.TextureCache.FindShortCache(descriptor) ?? PhysicalMemory.TextureCache.FindOrCreateTexture(_channel.MemoryManager, TextureSearchFlags.ForSampler, info, layerSize);
```
But might not be possible if `ProcessDereferenceQueue` *must* be called before the find method (which I assume to be the case).

ger96rh29:
image
the performance in my case is increased, I don't know if it should be better but compared to what I had before it is better, I don't know if the problem with my nvidia gpu and my amd cpu also influences let me know if it also affects opengl since with vulkan the performance is terrible my specifications are ryzen 5 3550H, gtx 1650 and 16G ram

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
riperiperi: It does need to be called. I've addressed the other feedback.

riperiperi:
Maybe needs retesting to make sure nothing blew up.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
gdkchan: On my first suggestion, it is still called so I think it's still valid.

riperiperi: That’s what I went with, I meant “other feedback” including that.

Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
```diff
@@ -104,6 +136,88 @@ public bool Remove(Texture texture, bool flush)
...
         }
 
+        /// <summary>
+        /// Attempt to find a texture on the short duration cache.
+        /// </summary>
+        /// <param name="descriptor">The texture descriptor</param>
+        /// <returns>The texture if found, null otherwise</returns>
+        public Texture FindShortCache(in TextureDescriptor descriptor)
+        {
+            if (_shortCacheLookup.Count > 0)
```
gdkchan, line 146:
```suggestion
            if (_shortCacheLookup.Count > 0 && _shortCacheLookup.TryGetValue(descriptor, out var entry))
```
To avoid the nesting. Although it seems that the count > 0 check could be just removed, if the count is 0 then `TryGetValue` will not return true. Unless the count check is some kind of optimization.

Ryujinx.Graphics.Gpu/Image/TextureDescriptor.cs
```diff
@@ -249,5 +250,24 @@ public bool Equals(ref TextureDescriptor other)
...
+        /// </summary>
+        /// <param name="other">The descriptor to compare against</param>
+        /// <returns>True if they are equal, false otherwise</returns>
+        public bool Equals(TextureDescriptor other)
+        {
+            return Equals(ref other);
+        }
+
+        /// <summary>
+        /// Gets a hashcode for this descriptor.
```
gdkchan, line 265:
```suggestion
        /// Gets a hash code for this descriptor.
```

Ryujinx.Graphics.Gpu/Window.cs
```diff
@@ -204,6 +204,8 @@ public void Present(Action swapBuffersCallback)
 
                 Texture texture = pt.Cache.FindOrCreateTexture(null, TextureSearchFlags.WithUpscale, pt.Info, 0, null, pt.Range);
 
+                pt.Cache.Tick();
```
gdkchan, line 207: Does the fact that this is only done here means that if a texture belongs to a cache where none of the textures are ever presented, it will never be freed?

Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
```diff
@@ -14,12 +36,22 @@ class AutoDeleteCache : IEnumerable<Texture>
 
         private readonly LinkedList<Texture> _textures;
 
+        private HashSet<ShortTextureCacheEntry> _shortCacheBuilder;
```
gdkchan, line 39: Do we need to force remove those on dispose, or are textures freed when the `GpuContext` is disposed right now regardless of the reference count?

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
gdkchan: Ah ok, GitHub did not mark it as "outdated" so I assumed it was not changed.

Ryujinx.Graphics.Gpu/Window.cs
riperiperi: Yes.

gdkchan: I think it would be nice to make sure the textures will eventually be freed even if the program never presents them. I don't think games will hit that case, but I can see it happening with homebrew tests that only do offscreen rendering for example.

Bjorn29512:
**Improves performance of FE3H in specific scenarios.**

Mainline: 
image

PR:
image

jochuan:
Can this be rebased ?

Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
riperiperi: It's an optimisation to avoid any costs for the dictionary method, it looks like it does a bit more work than the count check and the large majority of the time it will be 0.

Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
riperiperi: Yes, it disposes regardless of reference count.

Ryujinx.Graphics.Gpu/Window.cs
riperiperi: I'm not sure where this would be done if not present. The goal of the cache is to keep textures alive that are used on back-to-back frames. I guess they will always be deleted on overlap regardless of this?

gdkchan (approved):
lgtm, thanks.

marysaka:
Need a rebase @riperiperi

riperiperi:
Rebased.

PR #3760, opened 2022-10-13
Avoid allocations in .Parse methods

turbedi:
Use the Span overloads of the Parse methods when possible to avoid string allocations and remove one unnecessarry array allocation


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #3761, opened 2022-10-14
bsd: Check if socket is bound before calling RecvFrom()

TSRBerry:
This small PR fixes the `InvalidOperationException` that would be thrown if `RecvFrom()` gets called without calling `Bind()` first.
I'm not really sure about the return value, but couldn't find a better fit. Since I don't have bsdsockets RE'd enough to look this up, I decided to leave it like that and hope someone knows the correct return value to use here.

---

This fixes the crash Overpass experienced when selecting the career mode.

Thanks to @EmulationFanatic for sending me the crash log and testing my patch.


marysaka (approved):
nice catch

AcK77 (approved):

PR #3762, opened 2022-10-15
Fix various issues caused by Vertex/Index buffer conversions

riperiperi:
- The arguments for the 0th dummy vertex buffer were incorrect - it was given an offset of 16 rather than a size of 16. This was causing a lot of validation errors and out of bounds access.
- The wrong size was used when doing `autoBuffer.Get` on a converted vertex buffer.
- The possibility of a vertex buffer being disposed and then rebound can rebindings to find a different buffer where the current range is out of bounds. Avoid binding when out of range to prevent validation errors.
- The above also affects generation of converted buffers, which was a bit more fatal. Conversion functions now attempt to bound input offset/size.

Fixes some issues introduced in #3679 ,  #3715.


marysaka (approved):

gdkchan (approved):
lgtm, thanks.

PR #3763, opened 2022-10-16
Fix primitive count calculation for topology conversion

riperiperi:
Luigi's Mansion 3 performs a few non-index quads draws with 6 vertices for the mini map. It's meant to ignore the last two, but the index pattern's primitive count calculation was rounding up and drawing two quads instead of one.

https://user-images.githubusercontent.com/6294155/196014080-7f74a3be-265a-4f1e-a2e7-45ea30ee72f9.mp4

No idea why the game does this but this should fix random triangles in the map.

This is a recent regression from #3715. Feel free to make sure other stuff hasn't broken.


marysaka (approved):

gdkchan (approved):
lgtm, nice catch!

PR #3764, opened 2022-10-16
TamperMachine: Fix input mask check

mageven:
I noticed a report that cheats with conditional inputs don't work and scanning through just code, found this logic error. Please test if this fixes the problem as I pretty much never use cheats.

I would've preferred `(_input.Value ^ _mask) == 0` but original implementation uses this check-
https://github.com/Atmosphere-NX/Atmosphere/blob/master/stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp#L976

### DIFF
--- a/Ryujinx.HLE/HOS/Tamper/Conditions/InputMask.cs
+++ b/Ryujinx.HLE/HOS/Tamper/Conditions/InputMask.cs
@@ -13,7 +13,7 @@ namespace Ryujinx.HLE.HOS.Tamper.Conditions
 
         public bool Evaluate()
         {
-            return (_input.Value & _mask) != 0;
+            return (_input.Value & _mask) == _mask;
         }
     }
 }


marysaka (approved):

OldManKain:
Just tested this, Conditional Button Cheats work properly now.
I tested 10 different cheats with multiple button combinations without Issue.
Button Combinations I tested.
A + B, Y, X L3, R3
B + Y, X, L3, R3
Y + X, L3, R3
X + L3, R3
L + A, B, X, Y, L, R, L3, R3
R + A, B, X, Y, L, R, L3, R3
ZL + A, B, X, Y, L, R, L3, R3
ZR + A, B, X, Y, L, R, L3, R3
Also Cheats with automatic conditions still work like my 60fps 30fps cutscenes fix for Xenoblade Chronicles 2.

gdkchan (approved):
lgtm, nice catch!

PR #3765, opened 2022-10-17
Implement OpenDataStorageWithProgramIndex partially

### LINKED ISSUE #1596: fsp-srv IFileSystemProxy:205 (OpenDataStorageWithProgramIndex) is not implemented

fsp-srv IFileSystemProxy: 205 (OpenDataStorageWithProgramIndex) is not implemented

[Switchbrew](https://switchbrew.org/wiki/Filesystem_services#OpenDataStorageWithProgramIndex) | 

### Service description
```
Takes an input u8, returns an #IStorage.
```

### Required by: 

https://github.com/Ryujinx/Ryujinx-Games-List/issues/2578
https://github.com/Ryujinx/Ryujinx-Games-List/issues/2921

gdkchan:
This implements `OpenDataStorageWithProgramIndex`. This implementation only supports accessing the storage if the program index is the same as the current program index. It doesn't look like the emulator has any infrastructure right now to access data from other NCAs from the game. The current content manager only "installs" firmware NCAs, so only those are accessible.

If an attempt is made to access the storage of another program, it will throw `NotImplementedException`.
This is enough for all games that I have that requires this functionality.

RollerCoaster Tycoon 3:
image
MLB The Show 22:
image
This one has quite a few issues. It runs very slow and crashes on Vulkan due the local size being above the limit of 1024 on my graphics card (although it seems that Pascal and older have a higher limit of 1536, so maybe it will work on those cards on Vulkan?)

I believe it fixes #1596, but we might need to revisit this if a game has multi-program and needs to access the storage of the other programs.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
+++ b/Ryujinx.HLE/HOS/Services/Am/AppletOE/ApplicationProxyService/ApplicationProxy/IApplicationFunctions.cs
@@ -550,7 +550,7 @@ namespace Ryujinx.HLE.HOS.Services.Am.AppletOE.ApplicationProxyService.Applicati
 
         [CommandHipc(123)] // 5.0.0+
         // GetPreviousProgramIndex() -> s32 program_index
-        public ResultCode GetPreviousProgramIndex(ServiceCtx context)
+        public ResultCode GetPrevious(ServiceCtx context)
         {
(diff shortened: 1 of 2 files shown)


AcK77 (approved):

marysaka (approved):

EmulationFanatic:
Immortals Fenyx Rising now boots:
image

However it took several boot attempts to get to a working title screen, and a significant amount of menu graphics are missing. FPS is low in certain parts of the menu. 
image


If you keep pushing through to try to go in-game it eventually crashes before actual gameplay (with either a missing TRAP instruction error or generic fatal memory access error).

PR #3771, opened 2022-10-18
Use upstream unicorn for Ryujinx.Tests.Unicorn

TSRBerry:
This PR adds a modified version of `const_generator.py` used to generate official unicron bindings, which allows Ryujinx to get rid of patched unicorn versions for its tests.

I'm preparing multiple versions of this script and would like to receive feedback which one works the best. After that I'm going to revert/drop the commits of the undesired versions.

The constants I committed were generated from this commit: https://github.com/unicorn-engine/unicorn/commit/df3aa0fccbce9e1420e82110cbae5951755a0698, which is currently the latest commit in the dev branch.

~~I'll mark this PR ready for review as soon as I added the github workflow changes.~~


TSRBerry:
Do I need to use a specific cmake generator for the dll to make CI work?

TSRBerry:
> I'm preparing multiple versions of this script and would like to receive feedback which one works the best.

- Version 1: 070f296359c9ae38297d7eaf68b472e228888f93
- Version 2: dd895c9083dac7f792f9a28b22210677c27d69ce
- Version 3: 054c97fcef841f251879fd2ce4d2b404856a667d

TSRBerry:
Okay, now I have really no clue why the windows CI keeps failing. This new dll was working inside my VM and all tests completed successfully.
Does anyone know what's happening here?

TSRBerry:
As a sanity check I tried using the official dotnet bindings from unicorn instead of the Ryujinx ones.
Locally on Linux  everything runs fine again.

Using a completely statically linked dll the Windows VM fails with an `AccessViolationException` again. Unicorn tests execute fine, but the dotnet samples also crash, which means something is wrong with the official dotnet bindings too.

Ryujinx.Tests - `AccessViolationException` occurs at the exact same point:
```
The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at UnicornManaged.Binding.NativeBinding+Imported.uc_close(UIntPtr)
--------------------------------
   at UnicornManaged.Binding.NativeBinding+instance@72-1.UnicornManaged.Binding.IBinding.Close(UIntPtr)
   at UnicornManaged.Unicorn.Close()
   at Ryujinx.Tests.Unicorn.UnicornAArch64.Dispose()
   at Ryujinx.Tests.Cpu.CpuTest.Teardown()
```

UnicornSamples (part of unicorn dotnet bindings) executes for a bit and then crashes too:
```
*** Start: X86Code32
Unicorn version: (0, 0, 33554687)
Available memory: 3334500352
Calling VirtualAlloc with size: 1073741824
>>> Start tracing code
[+] 0x1000000: 41                  inc ecx
[+] 0x1000001: 4A                  dec edx

The program '[6996] UnicornSamples.exe' has exited with code 3221226505 (0xc0000409).
```

I believe the exit code indicates a stack overflow, but I'm not sure.

Not sure why this happens, which is why I'm documenting it and hope somebody has an idea what the issue could be.

---

Using the dll that succeeds with Ryujinx bindings in my VM, also leads to an error with the unicorn bindings, because it seems to check the dependencies here. The missing module here is `libwinpthread-1.dll`
```
Failed Adds_32bit(31,31,2147483648,uint.MinValue,1) [< 1 ms]
  Error Message:
   OneTimeSetUp: System.TypeInitializationException : The type initializer for 'Ryujinx.Tests.Cpu.CpuTest' threw an exception.
  ----> System.DllNotFoundException : Unable to load DLL 'unicorn' or one of its dependencies: The specified module could not be found. (0x8007007E)
```

---

I compile the completely statically linked dll with the following options:
```
link_libraries("-static-libgcc -static-libstdc++ -static")
```

The other one I usually test (because the tests succeed in my VM with it) is compiled using these options:
```
link_libraries("-static-libgcc -static-libstdc++")
```

TSRBerry:
Local `Ryujinx.Tests` results for bb09c6de1ec18064486c1ca9f756ee4ed9a4f15f:

- Linux:
  ```
  Passed!  - Failed:     0, Passed: 41454, Skipped:   189, Total: 41643, Duration: 23 h 39 m
  ```
- Windows VM:
  ```
  SystemInfo:
    - Windows 10 (19044.2130)
    - .NET version: 6.0.402
    - Test tool version: Microsoft (R) Test Execution Command Line Tool Version 17.3.1 (x64)

  Test result:
  Passed!  - Failed:     0, Passed: 41454, Skipped:   189, Total: 41643, Duration: 23 h 48 m
  ```

TSRBerry:
This is the patch I apply on top of [unicorn's dev branch](https://github.com/unicorn-engine/unicorn/tree/dev) just to figure things out.
[unicorn-dev.patch.txt](https://github.com/Ryujinx/Ryujinx/files/9838853/unicorn-dev.patch.txt)

The `mingw-w64.cmake` patch is actually required to make it work in my Windows VM, but the other changes are just present to make it easier to figure out the issue.

merryhime:
One could reduce qemu's tb-size to fix this error.

TSRBerry:
> One could reduce qemu's tb-size to fix this error. Alternatively, one could properly dispose of unicorn so that we don't rely on GC to clean up the 1GiB allocations of virtual memory space.

Could you explain this a bit further? I'm not sure if I understand what you mean.
How would reducing qemu's tb-size fix this error? I tried to reduce `DEFAULT_CODE_GEN_BUFFER_SIZE_1` to 256MB before, but that just made the test fail for every OS and even locally.

merryhime:
This PR is fixed after two changes:

1. With #3794 so that we don't overcommit memory

2. Not using a mingw-compiled dll (a normal dll would not require `libwinpthread-1.dll`, and would not have unusual breakage).

You can obtain a compatible precompiled `unicorn.dll` here: https://github.com/unicorn-engine/unicorn/suites/8892486340/artifacts/406750015

Successful CI pass after the above changes:
* https://github.com/merryhime/Ryujinx/actions/runs/3308374838/jobs/5460703597
* https://github.com/merryhime/Ryujinx/actions/runs/3308374838/jobs/5460703566

TSRBerry:
That's awesome! Thank you so much!

TSRBerry:
One last thing needs to be figured out: How do we load the unicorn library for the linux CI?

my last attempt didn't seem to work: https://github.com/Ryujinx/Ryujinx/actions/runs/3309740031/jobs/5463224358#step:8:23

Should I write code to modify the search path using NativeLibrary?

TSRBerry:
*Sigh*

Locally if `unicorn.so` is present in the output directory it will be loaded correctly now and if it's not present we get an error message and all cpu tests are skipped. 

I have no clue why CI can't successfully load the library even tho the right file is there.
Sanity check inside a docker container also worked fine, it found the file executed the tests and passed.

TSRBerry:
Since the windows tests are still working fine for CI, I'd say this is "done" and can be reviewed.

Ryujinx.Tests/Ryujinx.Tests.csproj
```diff
@@ -34,7 +34,17 @@
   </ItemGroup>
 
   <Target Name="CopyUnicorn" AfterTargets="Build">
-    <Copy SourceFiles="..\Ryujinx.Tests.Unicorn\libs\$(TargetOS)\unicorn.dll" DestinationFolder="$(OutputPath)" ContinueOnError="true" />
+    <ItemGroup>
+      <UnicornLib Include="..\Ryujinx.Tests.Unicorn\libs\$(TargetOS)\unicorn.*"/>
```
merryhime, line 38: Consider:
```suggestion
      <UnicornLib Include="..\Ryujinx.Tests.Unicorn\libs\$(TargetOS)\*unicorn.*"/>
```

merryhime:
Would probably prefer the linux library to be named libunicorn.so for consistency with expected naming.

TSRBerry:
Oh yeah that makes sense, I'll add that!

TSRBerry:
(Why does me requesting a new review, remove the other pending review requests....
I don't even have the permission to request reviews from people that haven't reviewed my PR yet.)

marysaka (approved):

gdkchan:
This should probably be a separate nuget package which includes the pre-built unicorn libraries. We shoulld probably not be keeping those in the tree here.

Also, will we get CPU testing on Linux now with those changes? From what I recall it was ignored on anything but Windows before. It would probably be nice to get it working on MacOS too since that's also a current build target.

TSRBerry:
> Also, will we get CPU testing on Linux now with those changes?

For some reason it doesn't work in CI, but locally it works correctly and the tests get executed.

> This should probably be a separate nuget package which includes the pre-built unicorn libraries. We shoulld probably not be keeping those in the tree here.

Does that mean we should wait until https://github.com/unicorn-engine/unicorn/pull/1728 is merged?

gdkchan:
> Does that mean we should wait until [unicorn-engine/unicorn#1728](https://github.com/unicorn-engine/unicorn/pull/1728) is merged?

Depends on how long it will take. Since we have no control over that, I will review those changes so that we can merge it. Once they have a nuget package, I think we should just use.

TSRBerry:
Awesome, thank you!

gdkchan (approved):
lgtm, thanks. Hopefully we can switch to the nuget package soon.

PR #2203, opened 2021-04-13
Add security and versioning dependency alerts

aleks-ivanov:
- add `dependabot.yml` which automatically enables Dependabot's dependency versioning scanner and dependency update PRs bot by declaring dependency ecosystems and sources in the project. For dependency security vulnerabilities scanner and vulnerable dependency update PRs bot, [enable "Dependabot alerts" and "Dependabot security updates"](https://docs.github.com/en/github/managing-security-vulnerabilities/configuring-dependabot-security-updates)

Resolves #2202

PR #3774, opened 2022-10-19
CI: Update workflows

TSRBerry:
As I spent a lot of time staring at the workflows, I noticed that most actions used within them aren't on their latest version.

Aside from that I also updated the workflows to replace the deprecated `set-output` workflow command with the new one and added `--no-build` to the test action, as Ryujinx was built right before.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,15 +48,15 @@ jobs:
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
       RYUJINX_BASE_VERSION: "1.1.0"
     steps:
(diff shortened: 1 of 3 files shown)


EmulationFanatic:
Just for visibility... https://github.com/Ryujinx/Ryujinx/pull/2203

TSRBerry:
Oh I didn't see that. In that case I'll remove the `dependabot.yml` again.

TSRBerry:
Dependabot stuff will be done in a new separate PR, I'll rebase this one in a few minutes.

marysaka (approved):

PR #3775, opened 2022-10-19
GPU: Use a bitmap to track buffer modified flags.

riperiperi:
We use a `MultiRegionHandle` to handle tracking buffer data. This has one tracking handle per guest page (4kb) which is signalled whenever data is written from the emulated CPU, a service, or gpu method (on cpu). When a handle is signalled, its `Dirty` flag is set, which lets it remember it has been modifed. Before we use a buffer for anything, we want to check to see if it has been modified, so we iterate each handle in the `MultiRegionHandle` and check each one's `Dirty` flag, queuing up a data upload to the backend whenever anything changes.

Most games bind small amounts of buffer data, which have us checking just a few handles. However, there are games that provide HUGE sizes for their bindings (eg. 10MB), intentionally or not, since there is no cost to doing this on hardware. We have to check buffer memory before draws... Which means a lot of iterating buffer handles to check dirty flags (and sync numbers). For 10 MB, that's 2560 handles, per draw.

With a bitmap, the dirty flag for each page of the buffer is represented as a single bit. All these bits are in sequence (excellent memory locality) and _64_ can be fetched at a time. This means that 256kb of data can be checked for modification at a time. That's a 64x improvement! Nice.

There is an additional optimisation that can skip checking a buffer entirely if all pages have been previously checked within the active sequence number. This is useful if an entire buffer is checked for some reason, and skips all the logic for calculating masks.

## Improved Games
There are likely a bunch of improved games I'm not aware of. Typically, games which have a FIFO% which seems to increase unnaturally as more things appear onscreen (compared to a "normal" game like SMO) 

## Greatly improved (>4x)
- Mario + Rabbids Kingdom Battle (the one that's been out for years)
- Super Mario 3D All-Stars (sunshine and galaxy)
  - Sunshine is slow due to JIT shortcomings. Mostly that the source PPC->A64 code is a nightmare.
- Zombie Army 4: Dead War (??? i don't know either)

#### Before:
image

#### After:
image
(framerate is the same between backends in this situation)

## Somewhat improved
- Monster Hunter Rise (replaces LDN perf hack, this also runs faster with indirect draw hle)

Most games are not likely to see an improvement (SMO), or may see some small improvement due to their use of large Vertex/Index buffers. Please test games to see if anything gets better.

## TODO
- ~Would it be faster to remove RegionHandleBase, and simply use the bitmap handle even when doing single handles? (in the name of reducing virtual calls)~
- ~XML docs, though these kind of suck in MultiRegionHandle anyways.~
- ~Full removal of MultiRegionHandle and updating tests?~

Please test on a bunch of games. This does affect everything so hopefully it doesn't blow anything up.


Ryujinx.Memory/Tracking/MultithreadedBitmap.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            int wordIndex = bit >> IntShift;
+            int wordBit = bit & IntMask;
+
+            long wordMask = 1L << wordBit;
+
+            long existing;
+            long newValue;
+
+            do
```
gdkchan, line 103: .NET 7 is adding `Interlocked.And` and `Interlocked.Or` APIs. So depending on when this is merged, it might be worth switching to the new APIs here.

riperiperi:
Now ready for review.

I was testing with .NET 7 to see what the performance was like devirtualizing the tracking handle methods.

.NET 7 master: 74 fps max
.NET 7 with RegionHandleBase: 69 fps max (nice... well, not really)
.NET 7 with RegionHandle: 75 fps max (within error?)

Conclusion is that it's definitely worth it. Downside is that individual handles for things like textures are backed by single bit multithreaded bitmaps, but it's not necessarily a deal breaker as these are much cheaper to check than buffers.

riperiperi:
Games that are also improved:
- Mario + Rabbids Sparks of Hope (images from @butzman#5685 on discord)
image
to
image

(the camera angle is slightly different, but the reported FIFO utilization is so huge that the bottleneck is likely elsewhere)

- Bayonetta 3 (images from @Gengar#6742 on discord)
image
to
image

Ryujinx.Memory/Tracking/MultiRegionHandle.cs
```diff
@@ -40,75 +53,89 @@ internal MultiRegionHandle(MemoryTracking tracking, ulong address, ulong size, I
...
 
-        public void SignalWrite()
+        public void ForceDirty(ulong address, ulong size)
         {
             Dirty = true;
+
+            int startHandle = (int)((address - Address) / Granularity);
+            int lastHandle = (int)((address + (size - 1) - Address) / Granularity);
+
+            // TODO: speed up
```
gdkchan, line 128:
```suggestion
            // TODO: Speed up.
```

gdkchan: Did you plan to do that here and forgot, or is it a future work item?

Ryujinx.Memory/Tracking/MultithreadedBitmap.cs
```diff
@@ -0,0 +1,161 @@
+﻿using System;
+using System.Threading;
+
+namespace Ryujinx.Memory.Tracking
+{
+    /// <summary>
+    /// A bitmap that can be safely modified from other threads.
+    /// </summary>
+    internal class MultithreadedBitmap
```
gdkchan, line 9: I think `ConcurrentBitmap` is a better name for this, to match the name of .NET thread safe collections. Or maybe just `ThreadSafeBitmap`.

Ryujinx.Memory/Tracking/RegionHandle.cs
```diff
@@ -108,7 +176,7 @@ private void ClearVolatile()
         public bool DirtyOrVolatile()
         {
             _checkCount++;
-            return Dirty || _volatile;
+            return _volatile || Dirty;
```
gdkchan, line 179: Why is the access order (and `Dirty` getter potentially not being called) important here?

Ryujinx.Memory/Tracking/MultiRegionHandle.cs
riperiperi: It’s a weird one. The original plan was to calculate the range of bits that would need to be set, and batch them all at once. However, now this increments _checkCount on the handle to trigger the handle to become “volatile”, so they kind of need to be set one at a time.

Ryujinx.Memory/Tracking/RegionHandle.cs
riperiperi: Dirty now has a cost, since it’s not a simple boolean field and instead a bitmap lookup. Doing _volatile first allows it to skip doing that when it’s set.

gdkchan: Ok, makes sense.

Ryujinx.Memory/Tracking/MultithreadedBitmap.cs
riperiperi: Done.

riperiperi:
Oh no, I forgot to rename a file. One sec.

gdkchan (approved):
LGTM, thanks! Always nice to see 500% performance improvements. I tested Super Mario Galaxy and Mario + Rabidds Kingdom Battle here, and they are both much faster than before. Both still have some performance issues here (Super Mario Galaxy reaches about 45 FPS inside the first level, it was around 20 before, and Mario + Rabbids is now full speed most of the time, although it still drops in some specific cases).

I wonder some of the cases that improved were due to index buffer modification checks. Right now it always checks the whole buffer, but we only really need to start the range starting at `firstIndex * indexTypeSize`.

Anyway, very nice, this should unblock a few other things like better handling of the index buffer for indirect draws.

marysaka (approved):

Linear524:
Holy $#1t !!! Astral Chain now runs fullspeed all the time and sudden framedrops have been fixed ! :)
Awesome work guys !!! )))

marwcel:
> Download the artifacts for this pull request:
> 
> * [ryujinx-Release-1.1.0+ea37552-linux_x64](https://nightly.link/Ryujinx/Ryujinx/actions/artifacts/416211506.zip)
> * [ryujinx-Release-1.1.0+ea37552-osx_x64](https://nightly.link/Ryujinx/Ryujinx/actions/artifacts/416211507.zip)
> * [ryujinx-Release-1.1.0+-win_x64](https://nightly.link/Ryujinx/Ryujinx/actions/artifacts/416211508.zip)
> 
> Experimental GUI (Avalonia)
> GUI-less (SDL2)
> Only for Developers

how can i download this for mac? the link says its expired

MutantAura:
> how can i download this for mac? the link says its expired

This is already merged and part of the `macos1` build: https://ryujinx.org/download

PR #3781, opened 2022-10-21
Vulkan: Fix vertex position Z conversion with geometry shader passthrough

gdkchan:
Right now we do the Z value conversion for the -1 to 1 depth mode in the last vertex pipeline stage shader that is used. But because shaders using geometry shader passthrough don't (and can't) modify `gl_Position`, when those shaders are used, it wouldn't do the conversion at all. This fixes the issue by not considering geometry shader with passthrough as `LastInVertexPipeline`, so it will do the conversion in the previous stage instead, which can be vertex or tessellation control.

Fixes black screen in Game Builder Garage on Vulkan.
Before:
image
After:
image


riperiperi (approved):
Nice catch!

AcK77 (approved):

PR #3792, opened 2022-10-21
Add InfoType.MesosphereCurrentProcess

Slattz:
Add's InfoType.MesosphereCurrentProcess to enum InfoType. 
It allows exefs replacements and homebrew to easily get their own process handle for use with certain SVCs, such as MapProcessMemory.

This isn't vanilla HOS behaviour but rather an extension added by Atmosphere, so I am not sure if this is fitting for Ryujinx. It is possible to get the process handle with a hacky method involving an IPC trick, but this is not exactly desirable.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs
+++ b/Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs
@@ -2107,6 +2107,34 @@ namespace Ryujinx.HLE.HOS.Kernel.SupervisorCall
                         break;
                     }
 
+                case InfoType.MesosphereCurrentProcess:
+                    {
+                        if (handle != 0)
+                        {
+                            return KernelResult.InvalidHandle;
+                        }
+
+                        if ((ulong)subId != 0)
+                        {
+                            return KernelResult.InvalidCombination;
+                        }
+
+                        KProcess currentProcess = KernelStatic.GetCurrentProcess();
+                        KHandleTable handleTable = currentProcess.HandleTable;
+                        int outHandle = 0;
+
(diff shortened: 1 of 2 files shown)


gdkchan:
We already added an atmosphere extension function to PM IDebugMonitorInterface that allows homebrew to get a handle for its own process. Why does that need to be added to the kernel aswell?

Slattz:
Honestly, just for the sake of simplicity. For example, a custom rtld that interacts with nnsdk, rather than libnx, doesn't have any pm:dmnt functionality implemented. It is much more simple just to call the SVC than to implement all the required IPC functionality to interact with pm:dmnt.

This is probably a niche use case, but it is functionality I would extremely appreciate and also provides parity with both Atmosphere and Yuzu.

marysaka (approved):

Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs
```diff
@@ -2107,6 +2107,34 @@ public KernelResult GetInfo(out ulong value, InfoType id, int handle, long subId
...
+                        if ((ulong)subId != 0)
+                        {
+                            return KernelResult.InvalidCombination;
+                        }
+
+                        KProcess currentProcess = KernelStatic.GetCurrentProcess();
+                        KHandleTable handleTable = currentProcess.HandleTable;
+                        int outHandle = 0;
+
+                        KernelResult result = handleTable.GenerateHandle(currentProcess, out outHandle);
```
TSRBerry, line 2126:
```suggestion

                        KernelResult result = handleTable.GenerateHandle(currentProcess, out int outHandle);
```
Could you define `outHandle` inline?

TSRBerry (approved):
Lgtm!

PR #3793, opened 2022-10-23
Vulkan: Use dynamic state for blend constants

riperiperi:
Prevents Mario Kart 8 Deluxe from creating an unbounded number of graphics pipelines when using blend constants (it seems to blend between colours depending on progress through certain course sections). Of course, it's possible for other games to do this.

Should generally use less memory and speed up drivers where compiling pipeline variants is slower.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -535,10 +535,11 @@ namespace Ryujinx.Graphics.Vulkan
                 vkBlend = new PipelineColorBlendAttachmentState();
             }
 
-            _newState.BlendConstantR = blend.BlendConstant.Red;
-            _newState.BlendConstantG = blend.BlendConstant.Green;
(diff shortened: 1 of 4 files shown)


GamerzHell9137:
Master:
image
PR:
image

gdkchan:
Could be worth to also remove `Id6` and `Id7` from `PipelineUid` (and then renumbering the subsequent fields), which should allow the comparison and hashing to be a bit faster, since the state size is reduced by 128 bits.

gdkchan (approved):
Nice catch, I remember that I couldn't use this initially due to a bug in Silk.NET (which I reported and has been fixed long ago). Looks like I forgot to ever change  to dynamic state after it was fixed.

As a side note, it would be nice to pass the blend constant as a single value, rather than one per blend entry (which doesn't make sense since there's only one blend constant). We can't change it without breaking the shader cache because that state is serialized to the shader cache...

marysaka (approved):

PR #3748, opened 2022-10-08
Implement HLE macro for DrawElementsIndirect

gdkchan:
This implements an HLE macro for the `DrawElementsIndirect` NVN macro, improving performance on games that uses this type of draw. This is mostly extracted from #3306.

**Affected games:**

Nier Automata: The End of YoRHa Edition. Lower FIFO % and slightly better performance in some areas.
Before:
image
After:
image

Monster Hunter Rise. Lower FIFO % and better performance in some areas (hard to tell how much better without unlocking the frame rate).
Before:
image
After:
image

Marvel Ultimate Alliance 3. Unfortunately this one won't really be improved since it also uses conditional rendering with GPU written buffers, so it still needs to flush data for that which nullifies all benefits from this change. I'm including it here just to indicate that it needs to be tested to make sure its not *worse*.

Nintendo Switch Sports, but this one is not playable right now due to other issues.

**Problems:**

- Conditional rendering comparison currently does not flush buffers before reading values. Marvel Ultimate Alliance 3 needs this, but since the counter values are in the same buffer as the indirect data, it was being flushed with the indirect data before. Since this change makes it skip this flush, I had to make the counter reads trigger flush to avoid regressing this game.
- We have no way to know the index buffer size with indirect draws, since the sub-range that is used is part of the indirect data (`firstIndex` and `indexCount`). This creates various problems and complications.
  - We don't know how much buffer data we need to update for the draw. Right now it just assumes a maximum size and updates all that.
  - ~~Can't convert index buffers on Vulkan, so any games using indirect draws with quads won't work properly. This can be fixed by doing the conversion in compute while reading the range from the indirect data there too.~~ This is now supported on 711ce42d176948a6e93127033ea6bccc09b7cd31.

**Other changes:**

- `MultiDrawIndirectCount` and `DrawIndirectCount` methods were combined to avoid code duplication.
- None of the indirect draw methods reads the CPU written indirect data anymore, and instead just always assume a fixed index buffer size. That's because if the count is written from GPU, then the CPU side data might contain garbage.
- The `MultiDraw*` methods were renamed removing the `Multi` prefix. This matches the Vulkan API names and allows all draw methods to be grouped together when sorted alphabetically, which is pretty neat.

Testing is welcome.

PR #3798, opened 2022-10-25
Vulkan: Fix indirect buffer barrier

gdkchan:
This is basically extracting a fix from #3748 since I don't know when that PR is going to be merged. Right now, we are missing the `IndirectCommandRead` bit on buffer barriers, which might potentially cause parameters to be fetched before write to the buffer is complete and cause the draw to use incorrect parameters. This was causing some DeviceLost crashes on Monster Hunter Rise with latest NVIDIA drivers, and maybe the older ones aswell.


marysaka (approved):

riperiperi (approved):
This makes sense since it's possible for data to be set from CPU then used in an indirect draw, which in our case could be a staging buffer copy, and the guest wouldn't emit a barrier for that. Note that directly writing a mapping doesn't generate any barriers, so I wonder if that would have any issues.

PR #3805, opened 2022-10-28
Reduce usage of Marshal.PtrToStructure and Marshal.StructureToPtr

marysaka:
This changes most codepaths of the emulator to not rely on ``Marshal.PtrToStructure`` and ``Marshal.StructureToPtr``.

Some structures were changed to make them unmanaged as a result.

I kept ``ReadStruct`` in ``SoftwareKeyboardApplet`` as it would probably be more of a burden to manipulate data safely there (and the overhead is acceptable)

Heavy testing is advised but structure size matched so it should technically be fine.


gdkchan (approved):
Nice, we just need some testing to make sure the structs didn't change, as historically we had regressions from this kind of change.

AcK77 (approved):

PR #3807, opened 2022-10-29
SPIR-V: Fix tessellation control shader output types

gdkchan:
GLSL requires that tessellation control shaders are indexed with `gl_InvocationID` built-in input. You can't actually use anything else as index, so this is effectively useless. On SPIR-V, this was not being done. This was not causing any validation errors, and most compilers did not complain about it, except AMD on Windows, where it would cause the compiler to crash. So to fix this, I have changed it to index outputs using `gl_InvocationID` on tessellation control shaders, which matches the GLSL compiler, and fixes the AMD crash.

Fixes crash on the following games on AMD Vulkan (Windows):
- Bayonetta 3 (chapter 1, after the cutscene).
- Luigi's Mansion 3 (right before the title screen).
- And likely any other game using tessellation (see list below, I did not test all of them).

Testing is welcome. Below is a list of games I know that uses tessellation (if you don't have AMD, they should be tested just to make sure they didn't break):
- Bayonetta 3 (only some areas, like the island (?) on chapter 1).
- Luigi's Mansion 3 (uses it right on the title screen, most visible on that sand room).
- The Witcher 3.
- The Legend of Heroes: Trails from Zero.
- Mario Golf: Super Rush.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 3781;
+        private const uint CodeGenVersion = 3807;
 
         private const string SharedTocFileName = "shared.toc";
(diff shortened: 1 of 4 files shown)


marysaka (approved):

Iifriet:
[Ryujinx_1.1.0_2022-10-29_06-50-47.log](https://github.com/Ryujinx/Ryujinx/files/9893493/Ryujinx_1.1.0_2022-10-29_06-50-47.log)

Sucessfully gets past crash at the start of chapter 1 on a 6800 XT using driver 22.10.3.  God rays flooding the scene are still present.

riperiperi (approved):
Nice catch! Good to see this working on AMD now.

Metrichi:
Confirming. Completed whole Chapter 1(the Island and Shibuya) on rx 580 4gb. No crashes at this moment.

PR #3815, opened 2022-10-31
Fix Fedora support

marysaka:
For some reasons, my fresh installation of Fedora 36 (KDE) doesn't have a symlink for libX11.so.

This PR fixes this by trying to import the library with its major version or fallback to the normal way.

SPB was updated in consequence.


AcK77 (approved):

edisionnano:
Someone should check Ubuntu too since it also needs the dev package of this lib

riperiperi (approved):
LGTM. Maybe a bit weird replacing the global import resolver.

PR #3818, opened 2022-11-01
SDL2Driver: Invoke dispatcher on main thread

merryhime:
Executes `SDL_PumpEvents` on main thread.

I note the SDL2 frontend already manually calls `SDL_PumpEvents` in its main loop.

macOS enforces the need for events pumped on the main thread.

### DIFF
--- a/Ryujinx.SDL2.Common/SDL2Driver.cs
+++ b/Ryujinx.SDL2.Common/SDL2Driver.cs
@@ -28,6 +28,8 @@ namespace Ryujinx.SDL2.Common
             }
         }
 
+        public static Action<Action> MainThreadDispatcher { get; set; }
+
         private const uint SdlInitFlags = SDL_INIT_EVENTS | SDL_INIT_GAMECONTROLLER | SDL_INIT_JOYSTICK | SDL_INIT_AUDIO | SDL_INIT_VIDEO;
 
         private bool _isRunning;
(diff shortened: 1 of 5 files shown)


young40 (approved):

marysaka (approved):

Ryujinx.SDL2.Common/SDL2Driver.cs
```diff
@@ -141,18 +144,19 @@ private void HandleSDLEvent(ref SDL_Event evnt)
...
+
+                SDL_Event[] events = new SDL_Event[1];
+                while (SDL_PeepEvents(events, 1, SDL_eventaction.SDL_GETEVENT, SDL_EventType.SDL_FIRSTEVENT, SDL_EventType.SDL_LASTEVENT) != 0)
                 {
-                    HandleSDLEvent(ref evnt);
+                    HandleSDLEvent(ref events[0]);
                 }
 
-                waitHandle.Wait(WaitTimeMs);
+                waitHandle.Wait(WaitTime.Milliseconds);
```
marysaka, line 159: I would actually prefer all of this to be in the MainThreadDispatcher callback.
So having the old code in the callback.

marysaka (changes requested):

merryhime:
Stuck the poll loop onto the main thread.

riperiperi (commented):
The changes look good to me, for MacOS. Does it make sense to do this on other platforms?

marysaka (approved):

gdkchan (approved):
lgtm, just not sure if we want to do that for all platforms (since this only a problem on MacOS). But not having different paths per platform is nice.

EmulationFanatic:
Tested some controller config/unplugging & replugging during runtime and while the input window was open. Everything works as it should.

PR #3822, opened 2022-11-03
Ensure all pending draws are done before compute dispatch

gdkchan:
Since instanced draws does one call per instance, we currently use an approach were the draw is deferred until we get the total instance count. In some rare circumstances, it was possible for a compute dispatch to be done while the draw was still pending, but then later it would do the pending draw, but at this point the state is already wrong, and since a compute shader is bound for the compute dispatch, it would try to draw with a compute shader which is invalid. This was causing crashes on Vulkan on the game Nights of Azure 2: Bride of the New Moon.

This change is basically just making sure all pending draws are done before the compute dispatch operation.
With this change, the aforementioned game now works on Vulkan.
image


riperiperi (approved):
Nice catch.

marysaka (approved):

PR #3828, opened 2022-11-09
GPU: Don't trigger uploads for redundant buffer updates

riperiperi:
Benefits:
- Reduces number of buffer uploads greatly in certain games (4096 byte cost each time)
- Avoids splitting render pass on Vulkan (cost depends on GPU driver)
- Batches adjacent ConstantBufferUpdater (CBU) writes into one access, reduces number of memory manager calls.

Downsides:
- Check/triggers read tracking for upload regions (slightly slower)
- Can be slow in software memory if range is non-contigous (forces redundant update anyways)
  - The redundancy check method says that it can return true if checking would be too slow, and that's what it does in this case. Allocating and filling an array with the non-contiguous regions would be a lot slower than just saying the write happened.
- Upload must be flushed by potential users (could be buggy if a case is missed)

This has a lot of potential to break things since it moves the actual data upload to the flush method. This is to avoid performing the redundancy check per-integer for certain uploads. Hopefully we should flush before everything that can reasonably read CBU output.

Improves performance in Xenoblade Chronicles: Definitive Edition (Vulkan, note there is another issue here) and Link's Awakening. May very slightly reduce performance in games that do CBU that is not redundant, and are pegged at 99% FIFO.

This one is dangerous, so check with as many games and different engines as you can.

### DIFF
--- a/Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
+++ b/Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
@@ -308,6 +308,34 @@ namespace Ryujinx.Cpu.Jit
         }
 
         /// <inheritdoc/>
+        public bool WriteWithRedundancyCheck(ulong va, ReadOnlySpan<byte> data)
+        {
+            try
+            {
+                SignalMemoryTracking(va, (ulong)data.Length, false);
+
+                Span<byte> target = _addressSpaceMirror.GetSpan(va, data.Length);
+                bool changed = !data.SequenceEqual(target);
+
+                if (changed)
+                {
+                    data.CopyTo(target);
+                }
+
+                return changed;
+            }
+            catch (InvalidMemoryRegionException)
+            {
+                if (_invalidAccessHandler == null || !_invalidAccessHandler(va))
+                {
+                    throw;
+                }
+
+                return true;
+            }
+        }
+
+        /// <inheritdoc/>
         public ReadOnlySpan<byte> GetSpan(ulong va, int size, bool tracked = false)
         {
             if (tracked)
--- a/Ryujinx.Cpu/Jit/MemoryManager.cs
+++ b/Ryujinx.Cpu/Jit/MemoryManager.cs
@@ -180,6 +180,37 @@ namespace Ryujinx.Cpu.Jit
             WriteImpl(va, data);
         }
 
+        /// <inheritdoc/>
+        public bool WriteWithRedundancyCheck(ulong va, ReadOnlySpan<byte> data)
+        {
+            if (data.Length == 0)
+            {
+                return false;
+            }
+
+            SignalMemoryTracking(va, (ulong)data.Length, false);
+
+            if (IsContiguousAndMapped(va, data.Length))
+            {
+                var target = _backingMemory.GetSpan(GetPhysicalAddressInternal(va), data.Length);
+
+                bool changed = !data.SequenceEqual(target);
+
+                if (changed)
+                {
+                    data.CopyTo(target);
+                }
+
+                return changed;
+            }
+            else
+            {
+                WriteImpl(va, data);
+
+                return true;
+            }
+        }
+
         /// <summary>
         /// Writes data to CPU mapped memory.
         /// </summary>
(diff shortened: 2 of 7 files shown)


LukeWarnut:
Tested around 15 different games using different engines. No regressions found. Nice work!

riperiperi:
Before:
image

After:
image

Note: 76 fps on opengl. The remaining slowdown is due to the game creating an excessive amount of sync when drawing characters.

Bjorn29512:
Tested about 8 games and played each for a bit to make sure nothing broke, seems perfectly fine so far.

Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
```diff
@@ -307,6 +307,34 @@ public void WriteUntracked(ulong va, ReadOnlySpan<byte> data)
             }
         }
 
+        /// <inheritdoc/>
+        public bool WriteWithRedundancyCheck(ulong va, ReadOnlySpan<byte> data)
```
gdkchan, line 311: If we're going to introduce a new method on the memory manager, I think it's more valuable to add something like a `GetWritableSpan` method and then implement the remaining logic on the GPU project. Maybe the existing `GetSpan` method should be renamed to `GetReadOnlySpan` and the new one would be called `GetSpan`. It would need to throw if the region is not contiguous on the software memory manager.

riperiperi: There's a bunch of weird specific stuff here that made me move it to the GPU. The read tracking before write, the rule to do a regular write when non-contiguous (writable span would force a copy), the redundancy check without looking up the memory address again... I think it's easier to keep it as its own thing.

lostromb:
Tested the following games with no noticeable regressions. Didn't observe any noticeable performance differences (though all games except Hyrule Warriors ran at full speed on both branches). Windows 11, Ryzen 3800, Nvidia 1660 Super, Vulkan

- Torchlight 3
- Tricky Towers
- Cadance of Hyrule
- Xenoblade 2
- A Short Hike
- Deltarune
- Lumines Remastered
- Dragon Ball FighterZ
- Valkyria Chronicles 4
- Dead or Alive Xtreme 3 Scarlet
- Alien Isolation
- Shantae and the Seven Sirens
- Sonic Mania
- Metroid Dread
- Hyrule Warriors
- Picross S4
- Kirby and the Forgotten Land
- Untitled Goose Game

lostromb:
I don't have things set up right now for more accurate perf numbers, but anecdotally this appears to speed up Smash and Pokemon Sword by about 4% (again, with the same config as in the previous comment)

Jayzee2008:
> 

hi lostromb, i'm having the same specs like you except with i5-9500F, does this PR also significantly improve Bayonetta 3 performance?

lostromb:
I don't have Bayonetta 3 unfortunately. But I did want to revisit this change with detailed numbers for other games just to check my assumptions. And yeah..... there's not much change here, positive or negative. So obviously the perf difference depends on how the game engine behaves.

image
image

Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
gdkchan: On the host memory manager, that would be basically just exposing the `SignalMemoryTracking` + `_addressSpaceMirror.GetSpan` since the region is always contiguous there unless it is completely invalid. For the software memory manager, it would throw if the region is not contiguous since returning a (writable) span in this case is just not possible. So it would catch the non-contiguous exception and call the regular write method in this case. Alternatively the method could be `TryGetWritableSpan` and it would call the regular write method if the returned value is false. The downside is that it needs to do two calls to the memory manager in this case which is less efficient, but only for the software memory manager, unless there's something else I'm missing.

gdkchan (approved):
LGTM, but I still think if a new method is going to be added to the memory manager, it would be better to add one that returns a writable span since that's also useful for other cases (like for example, memory copy and fill implementations. `IVirtualMemoryManager.Fill` currently uses `GetWritableRegion` with a fixed chunk size, which will probably do useless copies in some cases where the chunk is not contiguous. If we had a method to get a writable span and get the contiguous region size, it could be implemented with zero copies and zero allocations).

AcK77 (approved):

PR #3829, opened 2022-11-10
Minor improvement to Vulkan pipeline state and bindings management

gdkchan:
This improves the management of pipeline state and bindings management on the Vulkan backend:
- #3793 started using dynamic state for blend constants, which reduces  the amount of pipelines that we need to create, however the state was still there, and was still being compared/hashed which is a bit wasteful. This PR removes the blend constant from the pipeline state since it is not used anymore. Should make pipeline state lookup a little bit faster.
- Changed the descriptor set updater to stop ignoring null textures and clean up the bindings, instead of keeping stale bindings. This is to avoid bugs where we keep potentially already freed resources bounds, and also the potential for "garbage" to be sampled on the shader. The new behaviour should be closer to the hardware behaviour, and OpenGL was changed too.
- Avoid one array allocation/copy on the ASTC redundant update check path. This is something that I noticed a while ago and was pointed out by one of our contributors recently. Decided to do the change here because I felt like it was not necessary to open a separate PR for such a minor change.

I wanted to remove `BlendConstant` from the `BlendDescriptor` too. The `BlendDescriptor` is passed for each blend, but this is not really necessary because there's only one blend constant shared by all blends. However since this is currently stored in the shader cache, changing the struct would break compatibility with existing shader caches so I just kept it as-is.

Testing is welcome to make sure nothing broke, but I think chances are pretty low with this one.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -708,11 +708,12 @@ namespace Ryujinx.Graphics.Gpu.Image
                 else
                 {
                     bool dataMatches = _currentData != null && data.SequenceEqual(_currentData);
-                    _currentData = data.ToArray();
                     if (dataMatches)
                     {
                         return;
                     }
+
+                    _currentData = data.ToArray();
                 }
             }
 
--- a/Ryujinx.Graphics.OpenGL/Image/TextureBase.cs
+++ b/Ryujinx.Graphics.OpenGL/Image/TextureBase.cs
@@ -34,5 +34,22 @@ namespace Ryujinx.Graphics.OpenGL.Image
             GL.ActiveTexture(TextureUnit.Texture0 + unit);
             GL.BindTexture(target, Handle);
         }
+
+        public static void ClearBinding(int unit)
+        {
+            GL.ActiveTexture(TextureUnit.Texture0 + unit);
+
+            // Clear all possible targets since we don't know which one the shader will access.
+            GL.BindTexture(TextureTarget.Texture1D, 0);
+            GL.BindTexture(TextureTarget.Texture2D, 0);
+            GL.BindTexture(TextureTarget.Texture3D, 0);
+            GL.BindTexture(TextureTarget.Texture1DArray, 0);
+            GL.BindTexture(TextureTarget.Texture2DArray, 0);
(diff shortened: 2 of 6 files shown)


Ryujinx.Graphics.OpenGL/Image/TextureBase.cs
```diff
@@ -34,5 +34,22 @@ protected void Bind(TextureTarget target, int unit)
             GL.ActiveTexture(TextureUnit.Texture0 + unit);
             GL.BindTexture(target, Handle);
         }
+
+        public static void ClearBinding(int unit)
+        {
+            GL.ActiveTexture(TextureUnit.Texture0 + unit);
+
+            // Clear all possible targets since we don't know which one the shader will access.
+            GL.BindTexture(TextureTarget.Texture1D, 0);
```
riperiperi, line 43: What about this?
https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBindTextureUnit.xhtml

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -708,11 +708,12 @@ public void SynchronizeFull()
                 else
                 {
                     bool dataMatches = _currentData != null && data.SequenceEqual(_currentData);
-                    _currentData = data.ToArray();
                     if (dataMatches)
                     {
                         return;
                     }
+
+                    _currentData = data.ToArray();
```
riperiperi, line 716: Ouch, nice catch here.

Thorbenl:
FYI: this fixed lots of issue sofr Pokemon Scarlet and Violet :)

Ryujinx.Graphics.OpenGL/Image/TextureBase.cs
gdkchan: Nice, I didn't know this was a thing, updated.

Frizzacazzi:
> AMD still does

Experiencing the same.

riperiperi:
Experiencing the same what?

Frizzacazzi:
> Experiencing the same what?

AMD Radeon Pro 450 still yellowscreens. Sorry for not being clear before.

help-14 (approved):

marysaka (approved):

Issue #3518, opened 2022-08-01
Xenoblade Chronicles: Definitive Edition [0100FF500E34A000]: Flickering / Missing grass on title screen


SomeoneIsWorking:
### What's the issue you encountered?

Flickering / Missing grass on title screen on the ground.
https://youtu.be/sWfVh_Iq4ec
This glitch was only happening in Vulkan until 22.7.1 driver.
Now it happens with both Vulkan and OpenGL

### How can the issue be reproduced?

Boot the game then wait for title screen to load.

### Log file

[Ryujinx_1.1.201_2022-08-01_17-14-37.log](https://github.com/Ryujinx/Ryujinx/files/9234647/Ryujinx_1.1.201_2022-08-01_17-14-37.log)

### Environment?

- Ryujinx version: 1.1.201</br>
- Game version: 1.1.2</br>
- System Specs:
  - OS: Windows 10
  - CPU: Ryzen 5 2600
  - GPU: Radeon RX 570
  - RAM: 16GB
- Applied Mods : Speed increased (x1.5)

### Additional Context?

I previously reported this on Vulkan PR (when it was exclusive to Vulkan renderer) and decided to move it here since
- The PR has been merged
- The glitch is not exclusive to Vulkan renderer anymore

How the missing grass is supposed to look like
![image](https://user-images.githubusercontent.com/9450683/182168689-026743b4-e5be-47c7-9f99-cc5cf57adcce.png)

Monoken:
Same problem here..

### Environment

- Ryujinx version: 1.1.206
- Game version: 1.1.2

###  System Specs:
> OS: Windows 10

> CPU: Ryzen 5 5900x
> GPU: Radeon 6900XT
> RAM: 64GB
> Applied Mods from Emusak: 

![image](https://user-images.githubusercontent.com/110509296/182736983-b3387460-645b-44f4-b499-672ab8fff2e4.png)


### **Example of grass not showing up**

![image](https://user-images.githubusercontent.com/110509296/182736589-33a9bd1e-e224-421d-b1b9-6331b7264bf8.png)

Log: [Ryujinx_1.1.206_2022-08-03_17-12-26.log](https://github.com/Ryujinx/Ryujinx/files/9255740/Ryujinx_1.1.206_2022-08-03_17-12-26.log)

MetrosexualGarbodor:
Please do not mention piracy software here.

Monoken:
> Please do not mention piracy software here.

[...]

PR #3832, opened 2022-11-11
Use vector transform feedback outputs if possible

gdkchan:
Currently, shaders with transform feedback use individual, per-component outputs. That's because each component might potentially have a different, non-sequential offset, and in those cases we might need to specify different offsets per component. This works fine on NVIDIA, but causes an issue where some outputs are not written, or written at the wrong offset on Intel. To work around this issue, this change now attempts to use vector outputs in some cases. This is done when the offsets are sequential, and there's no fragment shader stage (because in this case, we don't need to worry about the VTG output not matching the fragment shader inputs).

Fixes grass on Xenoblade Chronicles Definitive Edition on Intel.
Before:
image
After:
image

This might improve things for AMD aswell, so it would be nice to get some AMD card owners to test the problematic games (Xenoblade, Pokkén Tournament, etc).
This affects all vendors, but on NVIDIA the previous approach was already looking fine, so we're just looking for regressions there.
Games to test:
- Xenoblade (all of them uses it for the grass, 3 seems to use it for particles too?)
- Pokkén Tournament
- Pokémon Legends Arceus
- SNK Heroines Tag Team Frenzy

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 3833;
+        private const uint CodeGenVersion = 3831;
 
         private const string SharedTocFileName = "shared.toc";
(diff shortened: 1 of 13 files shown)


Bjorn29512:
**XB 3 :**

Particles on Nvidia are still broken, they do render initially but every time there is a camera shift they disappear there is a chance its an entirely different issue. As far as performance is concerned i didnt spot any regression so on that end this game is fine.

image

**XBDE :** 

Seems fine no perf or visual regression noticed for it.

Mainline : 
image

PR:
image

riperiperi (approved):
Nice work.

Wes0x97:
awesome work, i'm getting about 28-30fps on the whole map... sometimes it drops to 28 in cities.
currently using Ryzen 9/ RTX 3070 TI with 2x resolution.

marysaka (approved):

SomeoneIsWorking:
image
Resolves #3518

PR #3833, opened 2022-11-11
Fix VertexId and InstanceId on Vulkan

gdkchan:
`gl_VertexId`  and `gl_InstanceID` are only valid on OpenGL. On Vulkan, `gl_InstanceIndex` should be used instead. However, there's a difference between the two. `gl_InstanceIndex` includes the `gl_BaseInstance` while `gl_InstanceID` doesn't. So this code replaces `gl_InstanceID` with `gl_InstanceID - gl_BaseInstance`, and `gl_VertexID` with `gl_VertexIndex` (the two should be equivalent).

The old approach was already working fine on NVIDIA, but not Intel.
Fixes incorrect rendering on games using `gl_InstanceID` on the shader and instanced draw with non-zero base instance on Intel.
Before:
image
After:
image

Other games that are likely affected are Sniper Elite 3, Metro 2033 Redux, and others, but I didn't test those.
I don't know if the issue affects AMD.
Testing is welcome.


marysaka (approved):

Lekyii:
This seems to fix the model texture on Pokemon Scarlet on AMD RX 6700

riperiperi (approved):
Cool. I look forward to RenderDoc not complaining about this all the time. 😌

PR #3845, opened 2022-11-13
Prevent raw Unicode control codes from showing on software keyboard applet.

lostromb:
There seems to be an internal formatting protocol with the real software keyboard involving Unicode control codes such as 0x13 (Device Control 3), which is used to influence the formatting somehow. However, just copying this string straight across to our software keyboard applets ends up showing ugly Unicode blocks with the raw control code.

Since we currently don't bother to handle these codes correctly (pending a more complete software keyboard applet), let's just strip these codes away before displaying them. This applies to any character under the "C0 Controls" [table on the Unicode spec](https://www.unicode.org/charts/PDF/U0000.pdf), including newlines, tabs, form feeds, etc.

Before
image

After
image

On Avalonia
image


Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardApplet.cs
```diff
@@ -205,11 +205,14 @@ private void ExecuteForegroundKeyboard()
             {
                 // Call the configured GUI handler to get user's input.
 
+                Logger.Info?.Print(LogClass.Application, $"Software keyboard initializing, header {_keyboardForegroundConfig.HeaderText}, sub {_keyboardForegroundConfig.SubtitleText}, guide {_keyboardForegroundConfig.GuideText}");
```
lostromb, line 208: whoops

lostromb: Removed.

Ryujinx.Tests/HLE/SoftwareKeyboardTests.cs
```diff
@@ -0,0 +1,83 @@
+﻿using NUnit.Framework;
+using Ryujinx.HLE.HOS.Applets;
+using System;
+using System.Collections.Generic;
```
gdkchan, line 4: I think some of those imports are not necessary.

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardApplet.cs
```diff
@@ -764,6 +763,37 @@ private void PushForegroundResponse(bool interactive)
             }
         }
 
+        /// <summary>
+        /// Some games send special control codes (such as 0x13) as part of the string (as a special formatting protocol with the keyboard?).
```
gdkchan, line 767: This sounds more something for the `remarks` section. Our method summaries usually explains what the method does, rather than *why* it does or the expected use case.

lostromb: Sure, I can add to remarks. I did want to keep this note around as an implementation detail for whenever someone does a software keyboard overhaul

Ryujinx.Tests/HLE/SoftwareKeyboardTests.cs
lostromb: Removed.

Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardApplet.cs
lostromb: Updated.

Ryujinx.HLE/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System.Runtime.CompilerServices;
```
marysaka, line 1: oh that's neat... I wasn't aware of that trick.

I should refacto some stuffs for Ryujinx.Audio tests I guess

riperiperi: Yeah, this is pretty cool.

marysaka (approved):

Ryujinx.Tests/HLE/SoftwareKeyboardTests.cs
```diff
@@ -0,0 +1,71 @@
...
+        public void StripUnicodeControlCodes_StripsToEmptyString()
+        {
+            string specialInput = Encoding.UTF8.GetString(new byte[] { 17, 18, 19, 20 }); // DC1 - DC4 special codes
+            Assert.AreEqual(string.Empty, SoftwareKeyboardApplet.StripUnicodeControlCodes(specialInput));
+        }
+
+        [Test]
+        public void StripUnicodeControlCodes_PreservesMultiCodePoints()
+        {
+            // Turtles are a good example of multi-codepoint Unicode chars
```
riperiperi, line 66: Iconic comment

riperiperi (approved):
LGTM

Ryujinx.Tests/HLE/SoftwareKeyboardTests.cs
lostromb: Yes, they're very handy for tests like this

Ryujinx.HLE/AssemblyInfo.cs
lostromb: I considered making `StripUnicodeControlCodes` part of a public static class somewhere so this wouldn't be necessary, but as its functionality is currently just coupled with the keyboard applet (and as a workaround for lack of functionality in that applet) it didn't seem worth it.

lostromb:
[Apparently this causes a regression on Steam Deck, now investigating...](https://cdn.discordapp.com/attachments/412730623405457408/1042582920134660187/20221117_003154.mp4)

PR #3850, opened 2022-11-15
GPU: VRAM Enhancements

Fliperworld:
This update adds settings so the user can manage VRAM consumption, and gives a significant gain in performance during texture recompression.

-**Texture Maximum Dimension** : 
Available values:  Unlimited, 128,256,512,1024.
Attempts to resize an 2D Astc compressed texture while maintaining  aspect ratio.

ie: Limited to **512**
2048x2048 ->   512x512
2048x1024 ->   **1024x512**
1024x1024 ->    512x512
1024x512   ->   Ignored
900x600     ->   Ignored
256x256     ->   Ignored  

Uses SkiaSharp for image scaling, and supports mipmaps and texture recompression.

-**VRAM cache maximum**:
Adds support to decrease texture cache size limit (5 to 2048). 

Bumped configuration version to 42

Ava settings
ryu settings

Testing environment:
S.O :Windows 11 22h2.
Memory :  32 GB 
CPU: Xeon E5 2440 V3 unlocked
GPU: Amd RX 460 unlocked **2GB** Vram
Game: Bayonetta 3

Ryujinx Settings :
Vulkan Backend
Enable Shader Cache checked
Texture recompression checked
Texture limited to 512 and cache to 80
Docked mode
ss1
ss2

With these settings, was able to play Bayonetta 3 without ever maxing out the 2GB of VRAM, even with firefox running in the background.The tradeoff was more texture loadings, and and some loss in quality...

### DIFF
--- a/Ryujinx/Ui/Windows/AboutWindow.Designer.cs
+++ b/Ryujinx/Ui/Windows/AboutWindow.Designer.cs
@@ -113,7 +113,7 @@ namespace Ryujinx.Ui.Windows
             //
             // _ryujinxPhoneticLabel
             //
-            _ryujinxPhoneticLabel = new Label("(REE-YOU-JI-NX)")
+            _ryujinxPhoneticLabel = new Label("(REE-YOU-JINX)")
             {
                 Justify = Justification.Center
             };


gdkchan:
I assume that this game goes past the 2GB of VRAM usage even with texture recompression, and the goal with this change is making the game work on GPUs with only 2GB of VRAM right?

If that's the case, have you considered using BC1 compression instead? It would likely be faster than BC7 and the output would have half of the size (BC1 only uses 8 bytes per block rather than 16). Quality would be a little bit worse but probably still better than halving the texture resolution. The main limitation on BC1 is that you can only have 1 bit of alpha, so if the texture has semi transparency it wouldn't be usable.

jduncanator:
I also feel like this is technical debt in quite a hot path that we really don't want to incur. The Switch itself has 4GB of RAM, users have no place trying to emulate a game on a GPU with 2GB of VRAM.

If we really must keep some form of down scaling, I'd rather restrict it to scale in powers of two, and just discard every second, fourth, eighth (etc) pixel, rather than depending on some complex image scaling algorithm from an external dependency..

Fliperworld:
@gdkchan : I hadn't considered BC1 yet for two reasons: I have no idea how much semi-transparent textures games would use, and mainly because i intend to keep Ryujinx's default recompressions. But it might be interesting to implement a boolean to use BC1 instead of BC7 in the emulator.

As for the quality...halving all textures resolutions would be terrible, so there is a filter to ignore certain resolutions/limits.
Let's assume the user set the maximum to 256 , and a 1024 x 512 texture needs to be resized.

1024 to 256 would have a ratio of 4 , so a direct resize would be to 256x64 . This is not a good idea, so the filter calculates the ratio of the smallest dimension and uses it , generating a final texture of 512*128.
This system ends up ignoring some limitations of user-defined dimensions, but ends up being much more superior than
the native resizing mode (which by the way is excellent for upscaling),because it doesn't downscale all textures.

In order not to have too many graphics quality losses, there is the option to limit the amount of textures in the VRAM, which does not affect the quality, but avoids crashes due to lack of VRAM when creating textures.

@jduncanator  : Not all textures are scaled down, as explained above. 
As for using SkiaSharp....I only used this dependency because Ryujinx.ava uses Avalonia.SVG.Skia which uses this same dependency to manipulate bitmaps, so instead of using System.Drawning.Common which has become a windows exclusive library, I chose one that is cross platform and is already in use in the solution.

jduncanator:
> @gdkchan : I hadn't considered BC1 yet for two reasons: I have no idea how much semi-transparent textures games would use, and mainly because i intend to keep Ryujinx's default recompressions. But it might be interesting to implement a boolean to use BC1 instead of BC7 in the emulator.
> 
> As for the quality...halving all textures resolutions would be terrible, so there is a filter to ignore certain resolutions/limits. Let's assume the user set the maximum to 256 , and a 1024 x 512 texture needs to be resized.
> 
> 1024 to 256 would have a ratio of 4 , so a direct resize would be to 256x64 . This is not a good idea, so the filter calculates the ratio of the smallest dimension and uses it , generating a final texture of 512*128. This system ends up ignoring some limitations of user-defined dimensions, but ends up being much more superior than the native resizing mode (which by the way is excellent for upscaling),because it doesn't downscale all textures.
> 
> In order not to have too many graphics quality losses, there is the option to limit the amount of textures in the VRAM, which does not affect the quality, but avoids crashes due to lack of VRAM when creating textures.
> 
> @jduncanator : Not all textures are scaled down, as explained above. As for using SkiaSharp....I only used this dependency because Ryujinx.ava uses Avalonia.SVG.Skia which uses this same dependency to manipulate bitmaps, so instead of using System.Drawning.Common which has become a windows exclusive library, I chose one that is cross platform and is already in use in the solution.

Yes, except Ryujinx.GPU is designed to be as abstracted as possible. It makes sense for the Avalonia UI to have that dependency, but not the Ryujinx.GPU project. You can still achieve everything you want to without needing to resort to an external dependency. The power of two suggestion still works perfectly in your examples if you include a threshold to prevent it from shrinking textures that don't need shrinking.

I think gdk's comment about halving the size when using BC1 was not referring to spatial size, but binary size. Put another way, BC1 has half the `bpp` as BC7 after compression. The dimensions remain the same, but the actual data is half the size in memory.

sakuka47:
I find the PR 3850 for interesting, however, is it useful for those who have more than 4GB of Vram?

An option with a drop-down menu, where the user selects the amount of Vram to use so that the emulator automatically optimizes the texture compression.

Fliperworld:
Yes, it actually helps more who has more VRAM, as the user will not need to apply limitations to the extreme.I assume that just setting cache texture to about 1.5k and maybe just limiting textures to 1024 solves almost all VRAM problems for users with 4GB or more vram.

riperiperi:
Unrelated to other comments, the method of reducing the texture resolution this way (manually downsampling) is unnecessary work. Nearly all of these textures come with mipmaps - just decode starting at a lower level.

I'm also generally always against adding settings for things, or game specific hacks, or I guess hacks in general. I agree that a faster BC1 recompression mode could be helpful, and maybe also fixing the issue that causes cutscene stutters (texture deletion?) instead of trying to work around it by reducing texture resolution.

sakuka47:
The less knowledgeable user could get lost in the settings and create more problems in the end

Fliperworld:
> Unrelated to other comments, the method of reducing the texture resolution this way (manually downsampling) is unnecessary work. Nearly all of these textures come with mipmaps - just decode starting at a lower level.
> 
> I'm also generally always against adding settings for things, or game specific hacks, or I guess hacks in general. I agree that a faster BC1 recompression mode could be helpful (most of these textures are opaque), and maybe also fixing the issue that causes cutscene stutters (texture deletion?) instead of trying to work around it by reducing texture resolution.
> 
> The AutoDeleteCache can't contain ASTC textures, so I don't think adding a limit is going to have an impact in this game. I also don't like the idea of this being configurable, as users can easily screw it up and cause textures to be flushed and recreated every frame.

My local copy is heavily focused on dumping assets and textures to png, and loading a image file to be used(Texture modding tools...).So I assumed that an implementation of a class focused on manipulating images could be useful for some future feature.

Limit on AutoDeleteCache , i aggree that a user could mess up setting this with values bellow 512. I don't know about others games, but Bayoneta 3sometimes  creates a bunch of 1600x900  black textures at start and end of chapters/verse ...and forgets it.... Thoose textures are add often to AutoDeleteCache, I suppose that a limit of 1024 is "secure" and avoids a unexpected huge VRam increase.
About the stuttering in cutscenes ...i din't notice any texture deletion, but i noticed that this happen at start of video game assets, thoose are  720x405 Avi and webm encoded files...i don't know if this helps in anything...
cutscenes

Anyway, what do you guys think of:

1- Changing the boolean "Enable Texture Recompression"  in settings to a label "Texture Recompression" with values : OFF, High, Low (off, bc7, bc1);

2?- A user configurable boolean "Use Small Texture Cache" ,default off uses all 2048 limit or on using 1024.

3?-A boolean that limits Texture Size to 1024 and gets thoose data from mimaps as suggested by @riperiperi

EmulationFanatic:
> Anyway, what do you guys think of:
> 
> 1- Changing the boolean "Enable Texture Recompression" in settings to a label "Texture Recompression" with values : OFF, High, Low (off, bc7, bc1);
> 
> 2?- A user configurable boolean "Use Small Texture Cache" ,default off uses all 2048 limit or on using 1024.
> 
> 3?-A boolean that limits Texture Size to 1024 and gets thoose data from mimaps as suggested by riperiperi

I would go with number 1 and remove the other two options entirely.

lostromb:
On one hand, it would nice to support users with low VRAM GPUs enabling them to play games that they couldn't before at the cost of some texture quality.

On the other hand, this is nothing more than a hack which will have to be maintained forever and which people will complain about if it doesn't work.... I'm thinking of speedhacks in Dolphin that have been implemented for years and years and which still generate bogus error reports from users turning them on by default

MutantAura:
I'll leave the merits of such a feature up to the experts but I do think a UX decision would need to be made about something like this. Laying out the settings as they are I think would only realistically cause confusion and create a lot of issues for us going forward. 

Can this not be behind a single toggle, perhaps combined with texture recompression, that would:

1. Query the users VRAM.
2. Use a defined preset or heuristic to determine the "Maximum Dimensions" and "Cache Maximum" values.

This strips a lot of the heavy lifting and failure points away from the users and into something that is relatively easy to maintain.
Users would only realistically use texture recompression if they were low on VRAM anyway so making all these VRAM saving features accessible at the push of a single button seems preferable to me at least.
Purely user and support perspective though. Can't offer any insight on the core feature itself :)

Fliperworld:
> Can this not be behind a single toggle, perhaps combined with texture recompression, that would:
> 
>     1. Query the users VRAM.
> 
>     2. Use a defined preset or heuristic to determine the "Maximum Dimensions" and "Cache Maximum" values.
> 
Querying available Vram can be trick , depends on OS, GPu Vendor, even driver, and just keeping track of how many bytes can't give a confiable result. 
Maybe a pre defined resolution scale "Low Res" that acts like "Maximum Dimensions" set at 512 and   "Cache Maximum" set at something like 1024 ?

jduncanator:
> > Anyway, what do you guys think of:
> > 1- Changing the boolean "Enable Texture Recompression" in settings to a label "Texture Recompression" with values : OFF, High, Low (off, bc7, bc1);
> > 2?- A user configurable boolean "Use Small Texture Cache" ,default off uses all 2048 limit or on using 1024.
> > 3?-A boolean that limits Texture Size to 1024 and gets thoose data from mimaps as suggested by riperiperi
> 
> I would go with number 1 and remove the other two options/changes entirely.

I agree with this approach. My position is the same as @riperiperi's on this. We've always gone to great lengths to avoid hacks like this in the past, and I think it's something we can be proud about that makes Ryujinx unique from other emulators.

MarkLanders777:
> > > Anyway, what do you guys think of:
> > > 1- Changing the boolean "Enable Texture Recompression" in settings to a label "Texture Recompression" with values : OFF, High, Low (off, bc7, bc1);
> > > 2?- A user configurable boolean "Use Small Texture Cache" ,default off uses all 2048 limit or on using 1024.
> > > 3?-A boolean that limits Texture Size to 1024 and gets thoose data from mimaps as suggested by riperiperi
> > 
> > 
> > I would go with number 1 and remove the other two options/changes entirely.
> 
> I agree with this approach. My position is the same as @riperiperi's on this. We've always gone to great lengths to avoid hacks like this in the past, and I think it's something we can be proud about that makes Ryujinx unique from other emulators.
> 
> I would tweak the wording perhaps from "High" and "Low" to "Prefer Quality", and "Prefer Size". High and Low don't really mean much to an end user.

"Quality", and "Speed" could be better?

EmulationFanatic:
> > > > Anyway, what do you guys think of:
> > > > 1- Changing the boolean "Enable Texture Recompression" in settings to a label "Texture Recompression" with values : OFF, High, Low (off, bc7, bc1);
> > > > 2?- A user configurable boolean "Use Small Texture Cache" ,default off uses all 2048 limit or on using 1024.
> > > > 3?-A boolean that limits Texture Size to 1024 and gets thoose data from mimaps as suggested by riperiperi
> > > 
> > > 
> > > I would go with number 1 and remove the other two options/changes entirely.
> > 
> > 
> > I agree with this approach. My position is the same as @riperiperi's on this. We've always gone to great lengths to avoid hacks like this in the past, and I think it's something we can be proud about that makes Ryujinx unique from other emulators.
> > I would tweak the wording perhaps from "High" and "Low" to "Prefer Quality", and "Prefer Size". High and Low don't really mean much to an end user.
> 
> "Quality", and "Speed" could be better?

It's not quality vs. speed though; the lower quality does not offer better speed. It offers more VRAM savings.
Whatever the option names would be, here's what they'd actually do:
Option 1: Better quality, more VRAM used/less VRAM savings.
Option 2: Worse quality, less VRAM used/more VRAM savings.

Plvy1n:
Guys I have this build takes Baynetta 3 instantly I wanted to check because I have a video card has only 2 gig of video memory maybe this build only works who has more than 2 gig do not know what the problem can not be explain?

Plvy1n:
Cool feature for me as the owner of gtx1050 2gb very helpful in Bayonetta 3 passed without crashes with this feature would be cool if this feature was added to the master build but video will not add soon)))).

Fliperworld:
My appologies, I accidentally closed this PR when i  was forcing my visual studio to resync.Anyway... I did some tests with BC1 recompression... in fast mode(using default textures sizes) it gives me worse results than resizing textures to a limit os 512 , and so many textures ignores transparency and shows a black square(i also checked, descompressed it was 0 on alpha)...i also tryed some external libs to be sure that its not was a mistake in my recompression code... maybe some shader/GPU vendor issue?!. BC1 in "balanced" mode(about 50) tries,gives a meh result, and is so slow, even using implementations instricts.

In my personal experience, BC1/DXT1 worth nothing, it's much better and easy just create big Astc texture/texture view with sizes from n miplevel and decoding/recrompressing from it. This increases performance and saves VRAM.
For me, the only issue with this approach is find out where perform/set this downscaled sizes...maybe on GetCreateInfo?! 
Unfortunately(or not) i have so many work to do for the next 15 days, so i can't try it for now.

Plvy1n:
> My appologies, I accidentally closed this PR when i was forcing my visual studio to resync.Anyway... I did some tests with BC1 recompression... in fast mode(using default textures sizes) it gives me worse results than resizing textures to a limit os 512 , and so many textures ignores transparency and shows a black square(i also checked, descompressed it was 0 on alpha)...i also tryed some external libs to be sure that its not was a mistake in my recompression code... maybe some shader/GPU vendor issue?!. BC1 in "balanced" mode(about 50) tries,gives a meh result, and is so slow, even using implementations instricts.
> 
> In my personal experience, BC1/DXT1 worth nothing, it's much better and easy just create big Astc texture/texture view with sizes from n miplevel and decoding/recrompressing from it. This increases performance and saves VRAM. For me, the only issue with this approach is find out where perform/set this downscaled sizes...maybe on GetCreateInfo?! Unfortunately(or not) i have so many work to do for the next 15 days, so i can't try it for now.

Good luck in your work and do not overload yourself heavily I'm glad you released this build with it I was able to fully pass Bayonetta3 on his feeble hardware.

DniweTamp:
you can add a check if image has alpha or not
then use dxt1/bc1 if it doesn't, and dxt5/bc3 if it does (for proper alpha)

Fliperworld:
@DniweTamp  I also did it in my tests...to be more precise, i tryed a "texture quality implementation" ....
ultra = no recompression
high = bc7
medium = smart (start at bc1, check for alpha on data before recompression...if there is any alpha, sets a flag to use bc7 forever on those textures with this flag).
low and/(or not) above = not implementd, after tests on 'medium".....

bc3~bc6 ....i ignored because this implementation was to save vram....

The "medium" on bayonneta 3 was a failure in my tests for those reasons:
Of 136 textures loaded in one chapter , about 28 has no alpha....the total performance has decreased (since after those checks, almost all textures still using bc7).
One of those "no alpha" textures was bayonetta face...well her face has so many gradients, and bc1 is to block for gradients, contrasting directly with her classes(bc7 for alpha on lens....). Even her face decoded/recompressed on mipmap level 3 has better quality(using bc7 of course). 
The amount of vram saved has...well... about 200mb....

So...ATM im doing some tests with a few lines of code...
On TextureInfo creation :

```
int limit = 512;//hardcoded for tests

            int _removedLevels = 0;
            if (formatInfo.Format.IsAstc())
            {
                while (width > limit && height > limit && levels > 1)
                {
                    width >>= 1;
                    height >>= 1;
                    levels--;
                    _removedLevels++;
                }
            }
            RemovedLevels = _removedLevels;
```
  On **CalculateSizeInfo**, width and height need to be left shiffted by RemovedLevels before returning a value;
  
Texture.cs:
```
///ConvertToHostCompatibleFormat
//To aling Host data with correct sizes

int width = Info.Width << Info.RemovedLevels;
            int height = Info.Height << Info.RemovedLevels;
. . .
int levels = single ? 1 : ((Info.Levels + Info.RemovedLevels) - level);

.  .  .
//
if (GraphicsConfig.EnableTextureRecompression)
                {
                   //Encode with current texture width,height miplevels
                    if (Info.RemovedLevels > 0)
                    {
                        width = Math.Max(Info.Width >> level, 1);
                        height = Math.Max(Info.Height >> level, 1);
                        levels -= Info.RemovedLevels;
                    }
                    result = BCnEncoder.EncodeBC7(decoded, width, height, depth, levels, layers);
                    return result;
                }
```
Also a **optional argument on AstcDecoder.TryDecodeToRgba8P** , where i pass  Info.RemovedLevels, to skip some levels if necessary and only decode what texture needs.(im not sharing it now because is so ugly ATM, im just setting a width -1 on AstcLevel thats need to be ignored,also not updating currentOutputOffset on ignored levels, and ignoring it on processes block).

That was a experimental implementation that's is giving about 50% on vram reduction, a notable fix on cutscenes sync. Its easy to maintaining , and gives a nice VRam saving + faster encoding/decoding.
But yes. ..there is a unepected "bug" thats i'm trying to solve: Resizing the emulator window makes VRAM consumption increases to a normal consumption...

jduncanator:
Before moving forward with any more changes, and potentially wasting any more time, I think it would be wise to have a discussion with the maintainers about your desired approach, as most of them here have expressed that the current approach is unlikely to get merged into the emulator.

I think we're all in agreeance that hacking something like this into the texture creation/decode pipeline is not something we would like to pursue in Ryujinx, so to come up with something viable that could be merged, we'd probably need to discuss a less hacky solution than what's currently already been suggested.

Fliperworld:
@jduncanator That's why i'm not making any pull request or opening a issue. I'm just sharing some experience/workflow of what i did to solve a issue for me that's can be a start point or even a solution for those who would suffer a similar problem in a future. 
Since this pr got so many sugestions, it is a good search/discussion point.
If someone wanna use BC1 recompression for performance, i suggest a filter to use BC1 only on rgb textures with compression astc 8x8 or above..., since even 8x6 is way better than BC1 fast recompression.BC5? not tested by me....maybe above 6x6.

For a "legit" real vram saving, maybe some changes on "custom resolution" ; currently any value below 1,0 causes some Textures to never show up, and always uses more vram than native 1x scale. Even my suggestion on [https://github.com/Ryujinx/Ryujinx/pull/3850#issuecomment-1321280865](url) never be used, it can be a hint to limit 'downscaling' on custom resolution and to not downscale textures w/o miplevels, like GUI and Fonts.I initially started there(prebuild scale codes),but for my lack of  knowledge, i found nothing that decrease VRAM without messing around so much.

gdkchan:
I think it would be useful to see a comparison with the BC1, BC7 and the original ASTC textures so that we can have a better idea of the quality loss we're talking about here.

Fliperworld:
@gdkchan  Sure, there as some samples on Original ASTC (only at a point where my vram explode) , BC7, BC1 , and my hack at 'limit 512"(the lowelest one):

MBs = Total VRAM used at this point, not the real amount of each texture.
All samples(except the hack one) uses Default Texture Sizes. 
Samples on BC1 are not filtered to ignore Alphas! , for a overall comparation, just ignore parts where alpha are nedeed.

title screen:
[ASTC Uncompressed- 700MB]
ASTC1

[BC7 - 614MB ]
BC71

[BC1 - 476MB]
BC1

[Hack one BC7- 475MB]


Chapter Select Screen:
[ASTC Uncompressed - 1337MB]
ASTC2

[BC7 - 713MB]
BC72

[BC1 -599MB]
BC1-2

[Hack - 555MB]


Chapter 3

[ASTC - 2GB]
BC73

[BC7 i forgot here]

[BC1 - 1285MB]
BC1-3

[Hack - 1048MB]


--------------------

[BC7 -2002MB]
BC74

[BC1 -1303MB]
BC1-4

[Hack - 1055MB]


-------------------

[BC7 - 2016MB]
BC75

[BC1 -1427MB ]
BC1-5

[Hack -1214MB ]


I also can record something if needed, raw rgba...

Fliperworld:
To be fair, if i use something like [BCnEncoder.NET](https://github.com/Nominom/BCnEncoder.NET), just for tests, i got same results on fast recompress, and a little less block on Slow mode, at a cost of about 10~20X more time than BC7 fast....

DniweTamp:
could you share a latest build for testing purposes

Fliperworld:
> could you share a latest build for testing purposes

https://github.com/Fliperworld/Ryujinx/tree/Testing-ForceBC1  <--needs cleanup , i almost finished a implementation(missing  alpha support), before start testing some existent codes...theoretically with some C# optmizations(i'm a c++ developer) ["CompressBlockFirstIMP"](https://github.com/Fliperworld/Ryujinx/blob/9c2fa5f49f0defccc8d7a96ea0b57cf6b0a00e86/Ryujinx.Graphics.Texture/Encoders/BC1Encoder.cs#L58) is "ok".

https://github.com/Fliperworld/Ryujinx/tree/Testing-ForceMipLevels-Limit512

nyxo100:
how i can install this ? links are not working

Plvy1n:
> how i can install this ? links are not working

I have on google disc left this version can share only I do not know whether you can leave a link here or not.

nyxo100:
> > how i can install this ? links are not working
> 
> I have on google disc left this version can share only I do not know whether you can leave a link here or not.

can you pass me the link pls?

Plvy1n:
> > > how i can install this ? links are not working
> > 
> > 
> > I have on google disc left this version can share only I do not know whether you can leave a link here or not.
> 
> can you pass me the link pls?

https://drive.google.com/drive/folders/1p54g_vURcaFIEAwFAnq6mhTObW7KsG3q

MetrosexualGarbodor:
Please don't post links to custom builds, if you'd like to use these changes then build them locally by yourself. Downloading custom builds from others entails a security risk.

PETONPOG:
Can someone give me a link pls?

PR #3853, opened 2022-11-15
Use ReadOnlySpan<byte> compiler optimization in more places

turbedi:


lostromb:
I question how much this may actually speed things up, but it doesn't look like it will hurt either

Ryujinx.Graphics.Vulkan/Shaders/ShaderBinaries.cs
```diff
@@ -1,3 +1,5 @@
+using System;
```
marysaka, line 1: Seems you forgot to make the changes in this file?

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #3862, opened 2022-11-18
Prune ForceDirty and CheckModified caches on unmap

riperiperi:
Since we're now using this for modified checks on the HLE indirect draw method, I'm worried that leaving these to forever gather cache entries isn't the best idea for performance in the long term, and it could keep old buffer objects alive for longer than they should be.

This PR adds the ability to prune invalid entries before checking these caches, and queues it whenever gpu memory is unmapped. It also aligns modified checks to the page size, as I figured it would be possible for a huge number of overlapping over a game's runtime. Ideally this should reduce time spent doing buffer modified lookups after a while of playing a game.

This prevents Super Mario Odyssey from having 10s of thousands of entries in the modified cache in Metro Kingdom, and them duplicating when entering and leaving a building (should be cleared, as they were unmapped).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
@@ -493,6 +493,9 @@ namespace Ryujinx.Graphics.Gpu.Memory
             _pruneCaches = false;
         }
 
+        /// <summary>
+        /// Queues a prune of invalid entries the next time a dictionary cache is accessed.
+        /// </summary>
         public void QueuePrune()
         {
             _pruneCaches = true;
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
```diff
@@ -435,6 +453,51 @@ public void SynchronizeBufferRange(ulong address, ulong size)
...
+            Prune(_dirtyCache, ref toDelete);
+
+            toDelete?.Clear();
+
+            Prune(_modifiedCache, ref toDelete);
+
+            _pruneCaches = false;
+        }
+
+        public void QueuePrune()
```
gdkchan, line 496: This method is missing XML docs.

gdkchan (approved):
lgtm other than the comment.

marysaka (approved):

PR #3865, opened 2022-11-18
Gpu: Fix thread safety of ReregisterRanges

riperiperi:
A quick fix to prevent reading the wrong value of Count when reregistering ranges for a new target buffer. Buffer flushes from another thread can modify the range list when the lock isn't active, which can change the count.

This prevents some crashes in Pokemon Scarlet/Violet and some other games. It's probably likely that buffer migration during flush is causing some other issues in this game, but this at least prevents the crashing.


gdkchan (approved):
Seems reasonable.

marysaka (approved):

PR #3866, opened 2022-11-19
Move gl_Layer from vertex to geometry if GPU does not support it on vertex

gdkchan:
Older GPUs (like 1st generation Maxwell and older NVIDIA GPUs) does not support writing `Layer` from the vertex shader. This causes issues on Pokémon Scarlet/Violet (black screen on OpenGL and a crash on Vulkan). This PR attempts to work around the issue by moving it to the geometry shader, if writing it on vertex is not supported.

Unfortunately I don't own such an old graphics card to test this, so this is depending on someone that owns one of those to test.


lostromb:
Dug an old GT 640 out of the scrap pile. Downgraded driver version to NVidia 456.71. On master, the behavior matches what you describe.

On PR (OpenGL):

image

Vulkan:

image

Looks good to me.

gdkchan:
Thanks for testing.

marysaka (approved):

Ryujinx.Graphics.Shader/Translation/TranslatorContext.cs
```diff
@@ -149,5 +155,104 @@ public ShaderProgram Translate(TranslatorContext other = null)
...
+            context.EndPrimitive();
+
+            var operations = context.GetOperations();
+            var cfg = ControlFlowGraph.Create(operations);
+            var function = new Function(cfg.Blocks, "main", false, 0, 0);
+
+            var sInfo = StructuredProgram.MakeStructuredProgram(new[] { function }, config);
+
+            var info = new ShaderProgramInfo(
+                config.GetConstantBufferDescriptors(),
```
riperiperi, line 239: Maybe there should be an overload that takes a ShaderConfig and runs all these functions itself, so that this doesn't have to be duplicated each time a shader is created from one.

riperiperi (approved):
Great to see, much nicer than I expected it would turn out. Using the existing fixed function stuff for this is a really nice surprise.

2Tsk2:
image
outside looks ok too, tested on a 940MX

Ryujinx.Graphics.Shader/Translation/TranslatorContext.cs
gdkchan: I moved it to `ShaderConfig`.

XKurotsunoX:
Working fine in my gtx 570 now
image

PR #3867, opened 2022-11-19
Vulkan: Clear dummy texture to (0,0,0,0) on creation

riperiperi:
This might fix an issue with AMD gpus on linux where the data could contain random garbage. On the switch, it always samples as 0.

Or so I think. Needs testing.


riperiperi:
Hearing reports that this works.

marysaka (approved):

gdkchan (approved):
lgtm, thanks.

Issue #3314, opened 2022-05-03
Expand DRAM setting hangs PLA from 1.1.110 onwards.


MutantAura:
### What's the issue you encountered?

From 1.1.110: https://github.com/Ryujinx/Ryujinx/pull/2954 enabling the expand DRAM size in Pokemon Legends Arceus is causing a hang on game boot. While this setting shouldn't be turned on normally and isn't required for any mods this is technically a regression.

### How can the issue be reproduced?

Enable the expanded DRAM option and boot PLA.

### Log file

[Ryujinx_1.1.113_2022-05-04_00-13-40.log](https://github.com/Ryujinx/Ryujinx/files/8616845/Ryujinx_1.1.113_2022-05-04_00-13-40.log)

### Environment?

- Ryujinx version: 1.1.113
- Game version: 1.1.1
- System Specs:
  - OS: *Windows 10)
  - CPU: *R5 5600X*
  - GPU: *NVIDIA RTX 3070*
  - RAM: *32GB DDR4*

### Additional context?

Regression report.

gdkchan:
I took a look and it also happens with software memory manager mode. In fact it crashes on the software memory manager mode with Expand DRAM even before the memory aliasing PR was merged, so it seems to be an issue with the memory allocation done by the HLE kernel (possibly a bug) that did not manifest before because we used to ignore the backing memory allocation with the fast memory manager mode, and instead just allocate the memory on the emulator address space directly. Needs more investigation.

Kidakakash:
thank you

Termasarus:
Still crashing on 1.1.338

BloodGozilla:
still crashing on 1.1.343

FeyrisTan:
Still happening. Cant use some mods without the option :(.

PR #3869, opened 2022-11-19
Add DRAM expansion launch warning

MutantAura:
Self-explanatory. Setting currently causes some titles to crash on launch, causing a lot of un-needed support requests.
Can potentially be reverted when #3314 is resolved.
image

### DIFF
--- a/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -596,5 +596,5 @@
   "SettingsEnableMacroHLE": "Enable Macro HLE",
   "SettingsEnableMacroHLETooltip": "High-level emulation of GPU Macro code.\n\nImproves performance, but may cause graphical glitches in some games.\n\nLeave ON if unsure.",
   "DialogPerformanceCheckExpandDramEnabledMessage": "You have DRAM expansion enabled, which is designed to be used for 4K game-mods ONLY.",
-  "DialogPerformanceCheckExpandDramEnabledConfirmMessage": "This setting is known to cause crashes and instability. If you are unsure if you need this, then you don't. Would you like to disable DRAM expansion now?"
+  "DialogPerformanceCheckExpandDramEnabledConfirmMessage": "This setting, when enabled, is currently known to cause crashes and instability in some games. If you are unsure whether you need this option, leave it disabled. Disable DRAM expansion now?"
 }
--- a/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -594,5 +594,7 @@
   "SettingsTabHotkeysVolumeDownHotkey": "Decrease Volume:",
   "VolumeShort": "Vol",
   "SettingsEnableMacroHLE": "Enable Macro HLE",
-  "SettingsEnableMacroHLETooltip": "High-level emulation of GPU Macro code.\n\nImproves performance, but may cause graphical glitches in some games.\n\nLeave ON if unsure."
+  "SettingsEnableMacroHLETooltip": "High-level emulation of GPU Macro code.\n\nImproves performance, but may cause graphical glitches in some games.\n\nLeave ON if unsure.",
+  "DialogPerformanceCheckExpandDramEnabledMessage": "You have DRAM expansion enabled, which is designed to be used for 4K game-mods ONLY.",
+  "DialogPerformanceCheckExpandDramEnabledConfirmMessage": "This setting is known to cause crashes and instability. If you are unsure if you need this, then you don't. Would you like to disable DRAM expansion now?"
 }
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -684,7 +684,7 @@ namespace Ryujinx.Ui
                 shadersDumpWarningDialog.Dispose();
             }
 
-            if (ConfigurationState.Instance.System.ExpandRam.Value)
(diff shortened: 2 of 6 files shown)


MetrosexualGarbodor:
IMO, people will simply ignore the warning, so it would be best to either detect the game IDs (Pokémon Scarlet/Violet and Arceus) and automatically disable the option, or if we don't want to do game IDs, disable it when no mods are loaded. 

If neither of those are possible this is good too, just saying there's a high probability of people skipping it.

MutantAura:
> IMO, people will simply ignore the warning, so it would be best to either detect the game IDs (Pokémon Scarlet/Violet and Arceus) and automatically disable the option, or if we don't want to do game IDs, disable it when no mods are loaded.
> 
> If neither of those are possible this is good too, just saying there's a high probability of people skipping it.

The aim here, even if they skip it, is to trigger some neurons. If they receive a warning, ignore it and then their game crashes, they're far more likely to actually read it the next time. Which is why I haven't decided to allow the option selection to be saved.

WiseHorror:
My only suggestion with this warning would be to also mention that this setting does not impact performance.

EmulationFanatic:
This approach seems a bit like using a sledgehammer to drive a pushpin. How many games are known to crash or have instability with the DRAM expansion enabled?  The most obvious is Pokemon Legends Arceus, so would it be possible to start with the pop-up for only that title ID?

I'll separately review with my verbiage suggestions :)

Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -594,5 +594,7 @@
   "SettingsTabHotkeysVolumeDownHotkey": "Decrease Volume:",
   "VolumeShort": "Vol",
   "SettingsEnableMacroHLE": "Enable Macro HLE",
-  "SettingsEnableMacroHLETooltip": "High-level emulation of GPU Macro code.\n\nImproves performance, but may cause graphical glitches in some games.\n\nLeave ON if unsure."
+  "SettingsEnableMacroHLETooltip": "High-level emulation of GPU Macro code.\n\nImproves performance, but may cause graphical glitches in some games.\n\nLeave ON if unsure.",
+  "DialogPerformanceCheckExpandDramEnabledMessage": "You have DRAM expansion enabled, which is designed to be used for 4K game-mods ONLY.",
+  "DialogPerformanceCheckExpandDramEnabledConfirmMessage": "This setting is known to cause crashes and instability. If you are unsure if you need this, then you don't. Would you like to disable DRAM expansion now?"
```
EmulationFanatic, line 599: Suggestion: "This setting, when enabled, is currently known to cause crashes and instability in some games including Pokemon Legends Arceus. If you are unsure whether you need this option, **leave it disabled**. Disable DRAM expansion now?"

Ryujinx/Ui/MainWindow.cs
```diff
@@ -683,6 +683,24 @@ public void PerformanceCheck()
...
                 shadersDumpWarningDialog.Dispose();
             }
+
+            if (ConfigurationState.Instance.System.ExpandRam.Value)
+            {
+                MessageDialog dramWarningDialog = new MessageDialog(this, DialogFlags.Modal, MessageType.Warning, ButtonsType.YesNo, null)
+                {
+                    Title         = "Ryujinx - Warning",
+                    Text          = "You have DRAM expansion enabled, which is designed to be used for 4K game-mods ONLY.",
+                    SecondaryText = "This setting is known to cause crashes and instability. If you are unsure if you need this, then you don't. Would you like to disable DRAM expansion now?"
```
EmulationFanatic, line 693: Suggestion: "This setting, when enabled, is currently known to cause crashes and instability in some games including Pokemon Legends Arceus. If you are unsure whether you need this option, **leave it disabled**. Disable DRAM expansion now?"

EmulationFanatic (commented):
Added some verbiage suggestions for clarity/conciseness.

Serfrost:
@EmulationFanatic Due to recent circumstances, I've seen more crashes with this setting with Scarlet \ Violet. I'd simply encourage not naming any specific games unless the dialogue changes automatically based on the title ID being launched.

On the contrary I don't really think an entire pop-up is necessary, rather, turning off the option as an override for some games is more or less going to be the least enduser-involved choice.

However, preventing whatever is causing the crashes outright with a patch for those specific titles may be the most sensible option (to avoid the ire of mod devs,) even if it isn't within spec.

MutantAura:
If the user selects "No" to disabling the expansion then the prompt will now not appear again. I've also taken the liberty of resetting the Expand RAM setting in general as well. We're seeing far too many people with it active. 🗡️

Ryujinx.Ava/Assets/Locales/en_US.json
MutantAura: Not a huge fan of mentioning specific titles, especially when that crash is technically a regression. I'd rather leave it vague. Perhaps just: `This setting, when enabled, is currently known to cause crashes and instability in some games. If you are unsure whether you need this option, leave it disabled. Disable DRAM expansion now?".

EmulationFanatic: Sure.

Serfrost:
> @MutantAura I've also taken the liberty of resetting the Expand RAM setting in general as well. We're seeing far too many people with it active.

Wholeheartedly agree with this decision, otherwise those who currently have it enabled would likely outweigh those who decide to read it and turn it off willingly after the change.

gdkchan:
Is that still necessary now that #3870 was merged?

MutantAura:
Possibly not. I wanted to give it a few days to see if #3870 was having the desired effect. 
Happy to close for now and re-assess if we ever hit similar issues in the future.

PR #3870, opened 2022-11-19
Reword the description of the 6GB expand DRAM hack to be less tantalizing

EmulationFanatic:
We've seen and continue to see hapless end users having crashes in some games (read: Pokémon Legends Arceus) because they've enabled what appears to be the best option in the world: "Expand DRAM to 6GiB". Who _wouldn't_ want that?
Despite warnings, tooltips, being in the **Hacks (may cause instability)** section and common sense, this problem is sure to continue manifesting unless the idea of expanding DRAM is expunged from the description entirely.
This PR is kissing cousins with https://github.com/Ryujinx/Ryujinx/pull/3869, and replaces the 6GB DRAM option's label (and some tooltip text) with something more boring.

Suggestions on other text are welcome.

Before:
image

After:
image


Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -119,7 +119,7 @@
   "SettingsTabSystemAudioBackendSDL2": "SDL2",
   "SettingsTabSystemHacks": "Hacks",
   "SettingsTabSystemHacksNote": " (may cause instability)",
-  "SettingsTabSystemExpandDramSize": "Expand DRAM Size to 6GiB",
+  "SettingsTabSystemExpandDramSize": "Use alternative MemoryMode layout (specific 4k mods only)",
```
MutantAura, line 122: This is a fairly long string to localize and I also think using "4K" or "mods" at all in this is a bad idea. Could possibly just become:
`Use alternative memory layout (Developers)`

4K mods are already mentioned in the tooltip and I don't see a reason to confuse any users thinking they may need it when resolution scaling or using mods in general.

Ryujinx/Ui/Windows/SettingsWindow.glade
```diff
@@ -1749,11 +1749,11 @@
                                 <property name="orientation">vertical</property>
                                 <child>
                                   <object class="GtkCheckButton" id="_expandRamToggle">
-                                    <property name="label" translatable="yes">Expand DRAM Size to 6GiB</property>
+                                    <property name="label" translatable="yes">Use alternative MemoryMode layout (specific 4K mods only - causes crashes otherwise)</property>
```
MutantAura, line 1752: Same as above.

Ryujinx.Ava/Assets/Locales/en_US.json
EmulationFanatic: That's fair. I'm all for keeping it simple.

Ryujinx/Ui/Windows/SettingsWindow.glade
EmulationFanatic: Addressed in another commit.

MutantAura (approved):

lostromb:
maybe name the setting something like "Development Kit Mode", "Emulate Switch Devkit", "Simulate Switch Debug Hardware"

OldManKain:
> maybe name the setting something like "Development Kit Mode", "Emulate Switch Devkit", "Simulate Switch Debug Hardware"

I like this Idea of a rename from Expand Dram Size to 6GiB to something that doesn't seem like a performance boots/hack to the layman.

marysaka (approved):

PR #3877, opened 2022-11-20
Avalonia - Fix controller insertion crash

emmauss:
Fixes crash that occurs when you connect a new controller while the settings window is opened on Avalonia.


Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
```diff
@@ -460,7 +460,6 @@ public void LoadDevices()
                 {
                     using IGamepad gamepad = _mainWindow.InputManager.KeyboardDriver.GetGamepad(id);
 
-                    Logger.Info?.Print(LogClass.Configuration, $"{GetShortGamepadName(gamepad.Name)} has been connected with ID: {gamepad.Id}");
```
gdkchan, line 463: Remove this empty line too?

gdkchan (commented):
Is the log not useful? Another way of fixing it without removing the logs would be just moving it inside the `if` I guess (assuming that the crash was a NRE). If the log is not useful though I think it's fine to just remove it.

Ryujinx.Ava/Ui/ViewModels/ControllerSettingsViewModel.cs
emmauss: Removed

TSRBerry (approved):

AcK77:
I’d prefere to keep the logs and move it in the condition instead of removing it.

emmauss:
> I’d prefere to keep the logs and move it in the condition instead of removing it.

The reason I removed the log is that, there's no similar log on the gtk side, so I thought it was redundant.

gdkchan (approved):
LGTM, thanks.

riperiperi (approved):

MutantAura:
Bit late to the party, the reason it exists here and not GTK is because the controller ID was removed from the GUI on avalonia. I felt it was still useful to tell the user what it was somewhere though. It's not super important but it does have a use Avalonia side.

PR #3880, opened 2022-11-21
Rename Hipc to Cmif where appropriate

Thealexbarney:


marysaka (approved):
lgtm, I do believe we should probably someday rename `Ipc` to `IPC` everywhere

Thealexbarney:
> lgtm, I do believe we should probably someday rename `Ipc` to `IPC` everywhere

`Ipc` already follows the standard naming style while `IPC` doesn't

TSRBerry:
I rebased this PR, hope this was okay. (Still have the original here, so I can revert if it wasn't.)

When this was created we didn't have the Ryujinx.Horizon projects, so these aren't adjusted here.
I wasn't 100% sure if they need to be adjusted, so I didn't do it.

@Thealexbarney let me know if you can take a look again or if you want me to do this.

Thealexbarney:
I thought this was fixed in the recent IPC PRs. Is that not the case?

TSRBerry:
Uhm, at least not that I know of :sweat_smile:

Thealexbarney:
Apparently I was thinking of `CmifCommandAttribute` which is only used in a few places

TSRBerry:
@Thealexbarney Could you answer this question I asked a few comments above?

> let me know if you can take a look again or if you want me to do this.

Thealexbarney:
The rebase looks fine to me. Looking at Ryujinx.Horizon, it appears that naming between HIPC/CMIF was done properly there. I don't have an IDE open right now, and GitHub won't let you search PRs, so I can't easily check if there were any other incorrectly named HIPC symbols that were added recently. If there weren't, then the rebase looks fine

gdkchan (approved):
This is already correct on the new IPC system, but I guess it doesn't hurt to rename it on the old implementation too.

PR #3616, opened 2022-08-22
Add Support for Post Processing Effects

emmauss:
Adds post processing effects using shaders.
Currently only FXAA and SMAA shaders are supported.
2x No AA
2-no-fxaa

2x FXAA
2-fxaa

Fixes #2729 as scaling can be done in post processing

PR #3881, opened 2022-11-21
GAL: Send all buffer assignments at once rather than individually

riperiperi:
The GAL Multithreading layer is good at reducing the cost of calls that would take longer on the backend, but the downside is an extra cost for loading commands into the command queue, and resource references/spans into circular queues.

The `(int first, BufferRange[] ranges)` method call has very significant performance implications when the bindings are spread out, which they generally always are in Vulkan, as the method ends up being called a lot of times every buffer update. This change makes it so that these methods are only called a maximum of one time per draw.

Significantly improves GPU thread performance in SMO, Pokemon Scarlet/Violet, and any other games that tend to bind many constant buffers at once. Probably not as much in OpenGL.

This change is already in LDN 3.0.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -973,9 +973,9 @@ namespace Ryujinx.Graphics.Vulkan
             SignalStateChange();
         }
 
-        public void SetStorageBuffers(int first, ReadOnlySpan<BufferRange> buffers)
+        public void SetStorageBuffers(ReadOnlySpan<BufferAssignment> buffers)
         {
-            _descriptorSetUpdater.SetStorageBuffers(CommandBuffer, first, buffers);
+            _descriptorSetUpdater.SetStorageBuffers(CommandBuffer, buffers);
         }
 
         public void SetStorageBuffers(int first, ReadOnlySpan<Auto<DisposableBuffer>> buffers)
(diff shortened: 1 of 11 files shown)


riperiperi:
Super Mario Odyssey (best case FPS)

Before:
image

After:
image

Maybe I should get actual metrics.

Littl3Guy:
Fixes the visual glitch in Mario Golf: Super Rush
Before:
mg

After:
mg2

marysaka (approved):

Ryujinx.Graphics.Vulkan/PipelineBase.cs
```diff
@@ -1015,7 +1015,19 @@ public void SetTransformFeedbackBuffers(ReadOnlySpan<BufferRange> buffers)
 
         public void SetUniformBuffers(int first, ReadOnlySpan<BufferRange> buffers)
```
gdkchan, line 1016: Can't the callers be modified to use the new method so that we can remove this one?

WiseHorror:
Latest commit seems to create a conflict with #3616

AcK77 (approved):

lostromb:
Breath of the Wild seems to regress about 7% with this change, from 39 down to 36 FPS.
I'm still trying to get more accurate numbers across different games...

gdkchan:
> Breath of the Wild seems to regress about 7% with this change, from 39 down to 36 FPS. I'm still trying to get more accurate numbers across different games...

Where exactly in the game are you testing this? SInce performance may vary a lot depending on the area and view.

lostromb:
Just on the first vista looking at the horizon from the plateau. But right now I'm also checking Mario Odyssey to validate that things run faster over there like we assume. And other games too if I have time tonight

image

gdkchan (approved):
LGTM, thanks. This did  make me wonder if instead of the buffer handle, having a `IBuffer` interface would be better, and that would allow the handle -> buffer lookup to be removed from the Vulkan backend (and from the GAL layer). My plan when I first introduced this was just exposing the handles from the driver directly, but turns out that would be difficult on Vulkan due to all the bookkeeping that we need to do for buffers and other resources, with requires a managed object to put the state on.

lostromb:
Actual numbers for different games. And I ran the BOTW numbers again since I was right to doubt them the first time.

Mario Odyssey - **66.9 -> 77.0 FPS (15% faster)**
Capturing the pole overlooking Metro Kingdom
image

Smash Ultimate - **42.9 -> 50.9 FPS (18% faster)**
8 CPU Marios in Fountain of Dreams
image

Breath of the Wild - **39.3 -> 40.6 FPS** (3% faster)
Great plateau, first vista
image

Pokemon Arceus - **36.0 -> 35.5 FPS (negligible decrease)**
Standing on the bridge of town looking at HQ
image

PR #3883, opened 2022-11-23
GPU: Relax locking on Buffer Cache

riperiperi:
I did this on ncbuffer2 when we were using it for LDN 3, but I noticed that it can apply to the current buffer manager too, and it's an easy performance win.

The only buffer access that can come from another thread is the overlap search for buffers that have been unmapped. Everything else, including all modifications, come from the main GPU thread. That means we only need to lock the range list when it's being modified, as that's the only time where we'll cause a race with the unmapped handler.

This has a significant performance improvements in situations where FIFO is high, like the other two PRs. Joined together they give a nice boost (73.6 master -> 79 -> 83 fps in SMO).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
+++ b/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
@@ -22,6 +22,10 @@ namespace Ryujinx.Graphics.Gpu.Memory
         private readonly GpuContext _context;
         private readonly PhysicalMemory _physicalMemory;
 
+        /// <remarks>
+        /// Only modified from the GPU thread. Must lock for add/remove.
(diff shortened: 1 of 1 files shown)


riperiperi:
Followup comparison from https://github.com/Ryujinx/Ryujinx/pull/3881#issuecomment-1322807806 with #3881, #3882 and this:

image

riperiperi:
I think it's more checking the lock multiple times every draw, rather than any lock contention. It surprised me too.

GamerzHell9137:
Awesome stuff, with the PRs above commented the FPS goes from 65 to 75 on a Ryzen 3600

image

image

AcK77 (approved):

PR #3884, opened 2022-11-23
ava: Fix JsonSerializer warnings

AcK77:
Since we move to .NET7, JsonSerializer now needs to have explicit options as arguments, which leads to some warnings in Avalonia project. This is fixed by using our `JsonHelper` class.


TSRBerry (approved):
Lgtm!

marysaka (approved):

Issue #3873, opened 2022-11-20
Opening a game without installed firmware crashes Ryujinx


Thealexbarney:
Opening a game without installed firmware and then selecting the "OK" or "Open the Setup Guide" options crashes Ryujinx. This stack trace is from 1.1.336
```
System.NullReferenceException: Object reference not set to an instance of an object.
  at Ryujinx.HLE.HOS.Horizon.Dispose(Boolean disposing) in C:\...\Ryujinx\Ryujinx.HLE\HOS\Horizon.cs:line 482
  at Ryujinx.HLE.HOS.Horizon.Dispose() in C:\...\Ryujinx\Ryujinx.HLE\HOS\Horizon.cs:line 409
  at Ryujinx.HLE.Switch.Dispose(Boolean disposing) in C:\...\Ryujinx\Ryujinx.HLE\Switch.cs:line 159
  at Ryujinx.HLE.Switch.Dispose() in C:\...\Ryujinx\Ryujinx.HLE\Switch.cs:line 152
  at Ryujinx.Ui.MainWindow.LoadApplication(String path, Boolean startFullscreen) in C:\...\Ryujinx\Ryujinx\Ui\MainWindow.cs:line 771
  ```

When disposing the emulation context's `Switch` object, `Horizon.Dispose()` tries to run `Fs.UnregisterProgram` using the process ID from the application's Horizon client. The problem is the client was never created because the application was never actually launched, so trying to get the process ID results in a NullReferenceException.

Skipping the `UnregisterProgram` call results in a deadlock where the main thread's waiting on the render widget to exit, even though it was never started.

PR #3885, opened 2022-11-23
ui: Fixes disposing on GTK/Avalonia and Firmware Messages on Avalonia

AcK77:
This fixes a dispose issue under Horizon/GTK/Avalonia, we don't check if the ApplicationClient is null so it throw NRE. We don't check if the main loop is active and waiting an event which is set in the main loop... So that could lead to a freeze.
Avalonia code is running on a main thread and waiting for the popup result, popup are running on a UI thread for the main thread never get the result and freeze.

Everything works fine in GTK/Avalonia (Thanks to @TSRBerry and @emmauss for their help) now.

Fixes https://github.com/Ryujinx/Ryujinx/issues/3873

### DIFF
--- a/Ryujinx/Ui/VKRenderer.cs
+++ b/Ryujinx/Ui/VKRenderer.cs
@@ -72,7 +72,11 @@ namespace Ryujinx.Ui
 
         protected override void Dispose(bool disposing)
         {
-            Device.DisposeGpu();
+            if (Device != null)
+            {
+                Device.DisposeGpu();
+            }
+
             NpadManager.Dispose();
         }
     }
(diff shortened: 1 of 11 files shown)


Ryujinx/Ui/VKRenderer.cs
```diff
@@ -72,7 +72,11 @@ protected override string GetGpuBackendName()
 
         protected override void Dispose(bool disposing)
         {
-            Device.DisposeGpu();
+            if (Device != null)
+            {
+                Device.DisposeGpu();
+            }
```
TSRBerry, line 78:
```suggestion
                Device?.DisposeGpu();
```
Could this be written like that?

AcK77: True!

TSRBerry:
This should fix everything on the Avalonia side. I tested this on Linux.

AcK77:
Still freeze on windows for me using Avalonia :(

AcK77:
Everything should works as expected now.

marysaka (approved):

emmauss (approved):

PR #3886, opened 2022-11-23
ui: Disallow checking for updates while emulation active

### LINKED ISSUE #3801: Prevent auto-updater from executing during runtime
#### Overview:
Currently it is possible to use the auto-updater functions while a game or application is actively running. This causes unexpected behavior if the update is then applied, leading to complete failure or only partial execution of the update process, leaving the user to have to manually clean up the Ryujinx folder & re-download the latest emulator package from the website.

- During runtime, the Help > Check for Updates menu option should be disabled.
- If the auto-updater check has already found an updated build and popped up the "Do you want to update Ryujinx?" dialog box, that box should be automatically terminated if emulation is then initiated or found to be active (i.e. if the user launched a game while the auto-update check was still in progress).

#### Nature of Request:  
[ Remove all that do not apply to your request. ]
- Change

To prevent the untimely deaths of Ryujinx installations and generally improve the UX.

ethanriverpage:
Related issue: #3801 

Prevents user from checking for updates while emulation is active; helps to prevent Ryujinx installation destroying itself by updating while game is running. Behaves as expected.

Only partially fixes #3801, does not kill auto-update prompt if emulation is active during check. 

Might be redundant under GTK with the switch to Avalonia, but should be easily implemented if not already.


marysaka:
Could you look at Avalonia side to see if a check is needed and add it if missing?

ethanriverpage:
@marysaka Check was needed, added check in Avalonia. Functions as expected.

ethanriverpage:
Either me or Git went a little crazy, but now the staged changes are correct and ready to be merged upon review.

marysaka (approved):

PR #3889, opened 2022-11-24
Vulkan: Destroy old swapchain on swapchain recreation

Fliperworld:
Currently, after a window resize, the old swapchain is passed to SwapchainCreateInfoKHR, but it is never destroyed after creating a new swapchain, causing a huge "vram leak", this can be easily reproduced, just resize the game window(preferably with a game that uses a lot of VRAM) ,and monitor the growth of VRAM usage until it runs out.
This PR adrress  this issue and also fixes a random screen freeze that occurs sometimes when switching from FullScreen mode to window mode.
This affect AMD gpus.

*Edit : Tested on Windows 11 with RX460 , RX 6500 XT and RTX 3070.


marysaka (approved):

riperiperi:
> this can be easily reproduced

Not really, it doesn't happen on my GTX 1070, so until recently I had no idea this was here. But I have reproduced it on another device.

You need to be careful with destroying the swapchain here - it could still be in use. For me destroying it immediately crashed the driver randomly when resizing, I had to do `vkDeviceWaitIdle` to ensure that the swapchain wasn't in use to avoid this, though this does add some delay when the swapchain is recreated. Another way would be to queue the swapchain for removal in a number of frames that matches the swap chain image count, though that's a bit more complicated.

Fliperworld:
In c++ i usually do a vkDeviceWaitIdle, vkDestroyFramebuffer,vkDestroyImageView and then vkDestroySwapchainKHR, but since i had no crashs,i wrongly guessed that  vkDeviceWaitIdle as a overkill.

Added a call to vkDeviceWaitIdle before vkDestroySwapchainKHR,

Serfrost:
Can confirm this has fixed some user-reported problems when using fullscreen. Prior behavior would result in a total system deadlock (black screen, unresponsive, unable to operate their PC.) This would result in them needing to forcefully shut down their PC after trying to enter/exist Fullscreen mode.

I would encourage this be added to LDN sooner than later to prevent the same problems.

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -57,6 +58,13 @@ private void RecreateSwapchain()
             }
 
             CreateSwapchain();
+
+            unsafe
+            {
+                //Destroy old Swapchain
```
gdkchan, line 64:
```suggestion
                // Destroy old Swapchain.
```

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -57,6 +58,13 @@ private void RecreateSwapchain()
             }
 
             CreateSwapchain();
+
+            unsafe
```
jduncanator, line 62: We don’t usually use inline unsafe blocks like this any where else in the Vulkan source code. Mark the entire function as unsafe, rather than just this block.

gdkchan: Can't unsafe be avoided by using the Span overload?

jduncanator: Yes it can. Sorry, wrote the comment on my phone without looking too closely. This should be able to use the Span.Empty overload.

Fliperworld: Done, unsafe code avoided.

gdkchan:
This is causing crashes on NVIDIA as soon as games start presenting on Vulkan.
```
   at Silk.NET.Vulkan.Extensions.KHR.KhrSwapchain.DestroySwapchain(Silk.NET.Vulkan.Device, Silk.NET.Vulkan.SwapchainKHR, Silk.NET.Vulkan.AllocationCallbacks ByRef)
   at Silk.NET.Vulkan.Extensions.KHR.KhrSwapchainOverloads.DestroySwapchain(Silk.NET.Vulkan.Extensions.KHR.KhrSwapchain, Silk.NET.Vulkan.Device, Silk.NET.Vulkan.SwapchainKHR, System.ReadOnlySpan`1<Silk.NET.Vulkan.AllocationCallbacks>)
   at Ryujinx.Graphics.Vulkan.Window.RecreateSwapchain()
   at Ryujinx.Graphics.Vulkan.Window.Present(Ryujinx.Graphics.GAL.ITexture, Ryujinx.Graphics.GAL.ImageCrop, System.Action)
   at Ryujinx.Graphics.GAL.Multithreading.Commands.Window.WindowPresentCommand.Run(Ryujinx.Graphics.GAL.Multithreading.Commands.Window.WindowPresentCommand ByRef, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper+<>c.<InitLookup>b__5_31(System.Span`1<Byte>, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper.RunCommand(System.Span`1<Byte>, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.RunLoop(System.Action)
   at Ryujinx.Ui.RendererWidgetBase.Render()
```
Driver version 516.40.

Fliperworld:
> This is causing crashes on NVIDIA as soon as games start presenting on Vulkan.

@gdkchan ,I implemented a Queue to destroy the old Swapchain. Could you please check for me if this fixes crashes on NVIDIA Gpu?

Unfortunately I only have  AMD GPUs, and I didn't get crashes even without vkDeviceWaitIdle, queues or any other protection.

After the queue implementation, when entering FullScreen mode, a strange behavior appears on the RX 6500. The game immediately enters fullscreen, blinks on window mode, and then goes FullScreen. Doesn't happen when exiting FullScreen mode, window resize or on RX 460.

jduncanator:
Personally, I feel like implementing a queue like this is a bit of a convoluted workaround to this issue. If disposing of the existing swap chain is a legitimate issue, then some refactoring would probably be preferred over trying to manipulate the current code to work.

Littl3Guy:
Can confirm this fixes the game freezing when exiting fullscreen mode.
Master:


https://user-images.githubusercontent.com/115218507/204122261-96c10519-3149-46d1-84b6-07ce8349c84c.mp4


(The black screen is just OBS issues and can be ignored)

Fliperworld:
I agree that a queue is a bit "overkill", and i was avoiding it , theoretically just a vkDeviceWaitIdle before Destroy Swapchain should be enough. I read somewhere that after vkAcquireNextImageKHR , vkQueuePresentKHR with that swapchain and image index should be called before swap chain disposal, so i did it. Well ... everything worked fine, until i got a random screen freeze when exiting FullScreen.

riperiperi:
It doesn't seem right that the screen just freezes if you don't dispose a swapchain, that seems more like a driver bug. As for the queue, I don't think refactoring the swapchain significantly would help much. Refactor it to do what? I'm not aware of a better way to dispose the swapchain.

mageven:
If these problems arise because we lose any info about a retired swapchain, how about going the other way?

DeviceWaitIdle -> Destroy the current swapchain first -> Create a new one without passing the old one.

It probably won't be as smooth but handling is much simpler. It's not like recreating swapchains is done frequently. The reuse part doesn't really affect emulators much as the swapchain queue would be very small because we'd almost always use mailbox or immediate presentation modes.

Fliperworld:
> It doesn't seem right that the screen just freezes if you don't dispose a swapchain, that seems more like a driver bug. 

I suppose that is related on how Windows handles the desktop. When the "screen freezes" , the system appears to be locked (no alt+tab, no alt+f4,no ctrl+shift+esc etc...) but, CTRL+Alt+Del works, and even if we just cancel/press esc on the 'panic screen' , everything goes back to normal(including the game), and there is no new entries on "Windows event log" related to any driver issue. It's as if the system was waiting for some signal that the swap chain of any full screen application was destroyed in order to repaint the desktop and start listening windows events again.

>    at Silk.NET.Vulkan.Extensions.KHR.KhrSwapchain.DestroySwapchain(Silk.NET.Vulkan.Device, Silk.NET.Vulkan.SwapchainKHR,

That part intrigued me ...a crash on Destroy after vkDeviceWaitIdle. If some requered info was lost, wasn't it supposed to crash in QueuePresent? That's why i went straight to the safest way I know to dispose a swapchain.

gdkchan:
> > This is causing crashes on NVIDIA as soon as games start presenting on Vulkan.
> 
> @gdkchan ,I implemented a Queue to destroy the old Swapchain. Could you please check for me if this fixes crashes on NVIDIA Gpu?
> 
> Unfortunately I only have AMD GPUs, and I didn't get crashes even without vkDeviceWaitIdle, queues or any other protection.
> 
> After the queue implementation, when entering FullScreen mode, a strange behavior appears on the RX 6500. The game immediately enters fullscreen, blinks on window mode, and then goes FullScreen. Doesn't happen when exiting FullScreen mode, window resize or on RX 460.

Sorry for the delay. I tested it now, still crashes, although in a different place now:
```
   at Ryujinx.Graphics.Vulkan.Window.Present(Ryujinx.Graphics.GAL.ITexture, Ryujinx.Graphics.GAL.ImageCrop, System.Action)
   at Ryujinx.Graphics.GAL.Multithreading.Commands.Window.WindowPresentCommand.Run(Ryujinx.Graphics.GAL.Multithreading.Commands.Window.WindowPresentCommand ByRef, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper+<>c.<InitLookup>b__5_31(System.Span`1<Byte>, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper.RunCommand(System.Span`1<Byte>, Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer, Ryujinx.Graphics.GAL.IRenderer)
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.RunLoop(System.Action)
   at Ryujinx.Ui.RendererWidgetBase.Render()
```

Fliperworld:
> DeviceWaitIdle -> Destroy the current swapchain first -> Create a new one without passing the old one.

@mageven  Thanks, this was the only way that i managed to fix crashes on Nvidia GPUs.  Apparently Nvidia GPUs "reuse/locks" the old swapchain forever.

@gdkchan  No problem, i was a little busy those days . I reverted the changes leaving the code more like the original PR, and also tested on RTX 3070 with drivers 527.56.

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -28,7 +29,7 @@ class Window : WindowBase, IDisposable
         private bool _vsyncEnabled;
         private bool _vsyncModeChanged;
         private VkFormat _format;
-
+
```
gdkchan, line 32:
```suggestion

```

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -49,14 +50,21 @@ public unsafe Window(VulkanRenderer gd, SurfaceKHR surface, PhysicalDevice physi
...
             {
                 _swapchainImageViews[i].Dispose();
             }
 
+            // Destroy old Swapchain.
+            _gd.Api.DeviceWaitIdle(_device);
+            _gd.SwapchainApi.DestroySwapchain(_device, oldSwapchain, Span<AllocationCallbacks>.Empty);
+
             CreateSwapchain();
+
```
gdkchan, line 67:
```suggestion
```

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -219,6 +226,7 @@ public static Extent2D ChooseSwapExtent(SurfaceCapabilitiesKHR capabilities)
 
         public unsafe override void Present(ITexture texture, ImageCrop crop, Action swapBuffersCallback)
         {
+
```
gdkchan, line 229:
```suggestion
```

Ryujinx.Graphics.Vulkan/Window.cs
gdkchan: This suggestion was supposed to be removing the whitespace, not the entire line.

gdkchan (approved):
lgtm, thanks. Sorry for the delay. Would be nice to add back the line that was removed by mistake due to the whitespace removal suggestion.

Issue #3662, opened 2022-09-01
Avalonia - Issue tracker


MutantAura:
## Bug Report

The Avalonia GUI currently has a few bugs and issues ranging from minor to major. This issue serves as a tracker and easily extendable way to follow progress on specific limitations. The tracker will be broken into "Major", "Moderate" and "Minor" which range from bugs that need to be resolved before Avalonia can become default all the way to more 'nice-to-have' features.

Leave a comment if there is anything **reproducible** that has been missed. This is also not an issue to add feature requests to. Comments asking for new GUI features will be marked as off-topic.

### Major

- [x] Updater does not function. Creates a blank dialog box with no content. (Fixed in https://github.com/Ryujinx/Ryujinx/pull/3670)
- [x] Using Vulkan on Windows (have checked Linux on Nvidia and Intel, does not occur there) the program is extremely unresponsive and takes disproportionate time to load GUI elements compared to OpenGL (Menus, tooltips, changing between windows etc.). Appears to have started after the merge of https://github.com/Ryujinx/Ryujinx/pull/3483. (Fixed by https://github.com/Ryujinx/Ryujinx/pull/3674)
- [x] Overlays cause serious issues especially when entering fullscreen. The Nvidia overlay will glitch for multiple seconds and shadowplay becomes nearly unusable. RTSS and similar monitoring software will also monitor the UI itself rather than the game window, making them also functionless. (Fixed in https://github.com/Ryujinx/Ryujinx/pull/3674)
- [x] Some Linux distros show a transparent render window using Vulkan. This also extends to any dialog boxes.
- [x] ContentDialog aren't visible in fullscreen on macOs (https://github.com/Ryujinx/Ryujinx/issues/4939)
### Moderate

- [x] Type boxes do not lose focus when clicking emptyspace and will remain active until an element that has a focus property is selected. Whitespace does not appear to have any focus properties. (Fixed in https://github.com/Ryujinx/Ryujinx/pull/3991)
![image](https://user-images.githubusercontent.com/44103205/194761090-1a68ba5b-794d-492b-ab73-e96113d63f29.png)
![image](https://user-images.githubusercontent.com/44103205/194761283-ece7b65b-c238-4d26-8471-bea716f7b3c4.png)
- [x] "Hide Cursor on Idle" setting does not work when hovering the render window on Windows 10/11. (Pending: https://github.com/Ryujinx/Ryujinx/pull/4266)
[...]

PR #3896, opened 2022-11-25
ava: Rework DLC Manager, Add various fixes and cleanup

AcK77:
In a way to continue the polishing of Avalonia UI, I've made some changes:

- The DLC Manager window is reworked and redesigned to be more friendly (Fixes one issue in https://github.com/Ryujinx/Ryujinx/issues/3662), related code have been cleaned up aswell.

image

Only left issue is a white rectangle when you click on cells (excepting checkboxes), if anyone (@emmauss ?) know how to remove it, you're welcome:
image

- Currently, if you input something in the search textbox, the gamelist is refreshed on each keystroke. I've added a timer of 1 second so it let you time to finish writing the whole keyword and then the gamelist is refreshed. Fixes an issue refreshing large gamelists too much, which takes too much times between keystrokes.

- In the Title Update Manager. When you add a new title update, you still have to select it by clicking on it and then save, now the update is selected automatically.

- In the settings windows, when you delete a game directory it doesn't select the next one right after, so you have to click on the next entry to delete it again. This is fixed by selecting the last entry available. Height of game directories list is increased too.

- I've done some cleanup here and there too.


EmulationFanatic:
These are all great improvements!

marysaka (approved):

emmauss (approved):

PR #3902, opened 2022-11-26
Avalonia: Fix OpenGL crashing on Linux

TSRBerry:
Fixes a regression from #3901

Thanks to @piplup55 for notifying me about the issue!

(This time Avalonia works on Linux works for both OpenGL and Vulkan. We only need to fix invisible dialogs after that, before it's usable again!)

### DIFF
--- a/Ryujinx.Ava/Ui/Controls/VulkanEmbeddedWindow.cs
+++ b/Ryujinx.Ava/Ui/Controls/VulkanEmbeddedWindow.cs
@@ -18,11 +18,9 @@ namespace Ryujinx.Ava.Ui
         [SupportedOSPlatform("linux")]
         protected override IPlatformHandle CreateLinux(IPlatformHandle parent)
         {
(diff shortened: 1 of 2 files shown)


Ryujinx.Ava/Ui/Controls/VulkanEmbeddedWindow.cs
```diff
@@ -1,17 +1,34 @@
...
     {
         private NativeWindowBase _window;
 
+        [SupportedOSPlatform("linux")]
+        protected override IPlatformHandle CreateLinux(IPlatformHandle parent)
+        {
+            X11Window = new GLXWindow(new NativeHandle(X11.DefaultDisplay), new NativeHandle(parent.Handle));
+
+            WindowHandle = X11Window.WindowHandle.RawHandle;
+
```
AcK77, line 24: Remove extra empty line (same for OpenGL file)

AcK77 (approved):

marysaka (approved):

PR #3915, opened 2022-11-27
Unbreak bug_report.md

andOlga:
It would appear that GitHub does not allow the "about" section of an issue template to exceed 200 characters.
image
Because of this, the template is missing from the issue type selector when opening a new issue, meaning that there is a bug that stops me from reporting bugs 😆
image
I have shortened the extra information (originally added in 51bb8707efbbb1af9ca959e68c8ab7d6bd26dc07) so that it fits under this limit, which should make the template usable again.


EmulationFanatic (approved):
LGTM! Thanks.

AcK77 (approved):

Issue #3922, opened 2022-11-27
Touchscreen X/Y values appear to be scaled incorrectly (Paper Bad)


lostromb:
## Bug Report

Game: Paper Bad - 01002A0019914000

### What's the issue you encountered?

While testing the game Paper Bad, I found that the way it registered touch input coordinates did not match where I was actually clicking on the emulator screen. The way that the game reads touchscreen X/Y coordinates provided by Ryujinx is somehow scaled incorrectly, which leads to the target marker not actually lining up with where you touch.

Clicking on the very TOP RIGHT of the screen puts the pointer a little northeast of center:
![ryujinx_capture_2022-11-27_13-27-24](https://user-images.githubusercontent.com/918053/204161810-60a5f207-8c25-478a-8e9c-d7d5297edefb.jpg)

Clicking on the very BOTTOM LEFT of the screen puts the pointer not quite at the lower left corner:
![ryujinx_capture_2022-11-27_13-27-31](https://user-images.githubusercontent.com/918053/204161813-8e77e844-aa4c-4c0c-9edf-431af05212de.jpg)

I had originally thought this was caused by DPI scaling in Windows causing a mismatch between monitor coordinates and in-game canvas coordinates. But that doesn't seem to be the case as it still reproes with DPI scaling set to 100%, as well as on a separate Linux machine with no DPI scaling applied.

### How can the issue be reproduced?

Load up Paper Bad and click anywhere

### Log file

[paper bad.zip](https://github.com/Ryujinx/Ryujinx/files/10099265/paper.bad.zip)

### Environment?

- Ryujinx version: 1.0.392 (though the log shows 1.0.0-dirty, it's a local build of the latest master. And I've reproed the bug on both 1.0.392 and 1.0.351, on separate systems)</br>
- Game version: 1.0.0</br>

- System Specs:
  - OS: Windows 11 / Linux Mint
  - CPU: AMD Ryzen 3800X / Intel Xeon W-2145
  - GPU: Nvidia GTX 1660 Super / Nvidia Quadro P400
  - RAM: 32GiB / 64GiB
- Applied Mods : No

### Additional context?

Additional info about your environment:</br>
Tested on both Windows 11 and Linux Mint, with Windows DPI scaling set to either 100% or 150%. Fullscreen / windowed, OpenGL / Vulkan. Behavior was the same across all platforms.

gdkchan:
[...]

PR #3923, opened 2022-11-27
Log a warning if users are trying to play a touchscreen game in docked mode

lostromb:
This is mostly to address cases of "user error" and to avoid bug reports like this #3922.

If the console is set to docked mode and the user is trying to interact with the touchscreen, show a warning on the log that this is physically impossible on real hardware and may not work.


Ryujinx.Input/HLE/TouchScreenManager.cs
```diff
@@ -1,13 +1,17 @@
...
 using Ryujinx.HLE.HOS.Services.Hid;
 using Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.TouchScreen;
+using Ryujinx.HLE.HOS.SystemState;
 using System;
 
 namespace Ryujinx.Input.HLE
 {
     public class TouchScreenManager : IDisposable
     {
+        private static readonly TimeSpan _invalidTouchWarningInterval = TimeSpan.FromSeconds(5);
```
lostromb, line 12: maybe need to reformat the name of this var since it's a const

lostromb: actually nah I guess the style is all right, based on consts in other files...

german77:
You might want to cleanup the commits to just one

lostromb:
> You might want to cleanup the commits to just one

yeah my master branch is garbage. I can rebase I guess...

lostromb:
barf!

Issue #3918, opened 2022-11-27
homebrew app store crash before loading the store list


CrazyDani:
## Missing Service Call

Hey, my problem is that I want to use the homebrew app store, but when I try to start it it just starts loading and then the emulator crashes (guest internet access is already enabled and i am using the ldn version 

00:00:11.133 |E| HLE.OsThread.17 Application : Unhandled exception caught: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'length')
   at Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.Types.AddrInfoSerialized.Read(ReadOnlySpan`1 buffer, ReadOnlySpan`1& rest) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.HLE\HOS\Services\Sockets\Sfdnsres\Types\AddrInfoSerialized.cs:line 55
   at Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.IResolver.DeserializeAddrInfos(IVirtualMemoryManager memory, UInt64 address, UInt64 size) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.HLE\HOS\Services\Sockets\Sfdnsres\IResolver.cs:line 575
   at Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.IResolver.GetAddrInfoRequestImpl(ServiceCtx context, UInt64 responseBufferPosition, UInt64 responseBufferSize, Boolean withOptions, UInt64 optionsBufferPosition, UInt64 optionsBufferSize) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.HLE\HOS\Services\Sockets\Sfdnsres\IResolver.cs:line 499
   at Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.IResolver.GetAddrInfoRequest(ServiceCtx context) in D:\a\Ryujinx_staging\Ryujinx_staging\Ryujinx.HLE\HOS\Services\Sockets\Sfdnsres\IResolver.cs:line 159
   at System.RuntimeMethodHandle.InvokeMethod(Object, Void**, Signature, Boolean)
   at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)
   --- End of inner exception stack trace ---
[...]

PR #3924, opened 2022-11-27
sfdnsres: Fix deserializer of AddrInfoSerialized when addresses are empty

AcK77:
We currently doesn't check the `AddressLength` of the `AddrInfoSerializedHeader` and this could lead to an oob since we still trying to parse addresses for the (unused) passed `hints` in `GetAddrInfoRequestImpl`.

I've cleaned up a bit the code too, and added a missing padding which could help to understand the size of `AddrInfo4`.

Fixes https://github.com/Ryujinx/Ryujinx/issues/3918

HB App Store seems to works (at least run, I haven't tried it more):
image

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfoSerialized.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfoSerialized.cs
@@ -35,7 +35,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Sfdnsres.Types
 
             AddrInfo4? socketAddress = null;
             Array4<byte>? rawIPv4Address = null;
-            string canonicalName = null;
+            string canonicalName;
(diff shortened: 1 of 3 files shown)


marysaka (approved):

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/Types/AddrInfoSerialized.cs
```diff
@@ -50,6 +50,13 @@ public static AddrInfoSerialized Read(ReadOnlySpan<byte> buffer, out ReadOnlySpa
 
             Debug.Assert(header.Magic == SfdnsresContants.AddrInfoMagic);
 
+            if (header.AddressLength == 0)
```
marysaka, line 53: Nice catch, I totally forgot that check in my implementation oops

TSRBerry (approved):

PR #3925, opened 2022-11-28
Log a warning if users are trying to play a touchscreen game in docked mode

lostromb:
This is mostly to address cases of "user error" and to avoid bug reports like this https://github.com/Ryujinx/Ryujinx/issues/3922.

If the console is set to docked mode and the user is trying to interact with the touchscreen, show a warning on the log that this is physically impossible on real hardware and may not work.

### DIFF
--- a/Ryujinx.Audio/Renderer/Dsp/AdpcmHelper.cs
+++ b/Ryujinx.Audio/Renderer/Dsp/AdpcmHelper.cs
@@ -1,4 +1,5 @@
 using Ryujinx.Audio.Renderer.Dsp.State;
+using Ryujinx.Common.Logging;
 using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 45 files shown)


TSRBerry:
Not sure if a user would see this since the console often gets disabled.

Also there a few games like Mario Kart 8 that allow touch input while in docked mode even if it's impossible on real hardware.

lostromb:
Surely this is..... better than nothing, right? The biggest downside I might see is log spam for people using their mouse on Mario Maker 2 in docked mode since that's probably the most common use case that would hit this scenario.

Yes, some games may still work correctly but that's not a guarantee. Some games may act like the touch coordinates are scaled incorrectly because they change resolution in docked mode and never account for what happens if touch inputs still happen in that mode.

It seems like if the user is trying to do something that is ergonomically implausible on real hardware, we should try to hint at this fact somehow before they waste developer time with false bug reports.

gdkchan:
I don't like the fact that this is spammed on the log every 10 seconds. Also I don't think it would be very effective since most users either disables the console, or disable all logs in the settings, or both, and they generally don't pay attention to the log. The best option here in my opinion if we *need* to make the user aware of this is adding an option, "Allow touch controls in docked mode". When enabled it could show a warning message that some games won't behave properly, or maybe just the tool tip is enough. Maybe such option would be under hacks. But it's possible that this would just replace one problem with another, as people would start reporting issues with touch not working because they are in docked mode...

lostromb:
Alright then. How about a single warning message.

I figured debating about making this configurable or having some extra UI to handle it would go nowhere which is why I'm just adding it as a log message. Again, my intent is mainly to save developers' time by quickly calling out behavior that games aren't necessarily designed for.

gdkchan:
Since I was requested to review again, I generally don't like this change. I don't think spamming a warning message is a good solution. I think the best solution would probably be adding a setting to enable touch in docked mode under hacks. It could warn the user when enabled, and be disabled by default.

Issue #3858, opened 2022-11-16
.NET 6/7 improvements


gdkchan:
#### Overview:
This is a short list of improvements that have been enabled by .NET 6 or .NET 7, with the introduction of new features and APIs.

#### Smaller Details: 
- [x] `Interlocked.And`/`Interlocked.Or` (#3937)

`Ryujinx.Memory/Tracking/ConcurrentBitmap` currently uses a compare and swap look to set and clear bits on the bitmap array. Those news methods can be used to update it atomically in a more efficient way.

- [x] Generic math (#3929)

We have some methods on `Ryujinx.Common/Utilities/BitUtils` with overloads for all the common integer types. With the new generic math feature, we can replace them with a single method that operates on integers. There are other cases that can potentially benefit too like the `Add` methods on `ARMeilleure/IntermediateRepresentation/Operand`.

- [x] System.Half (#4106)

This was actually introduced in .NET 6. We can remove `Ryujinx.Graphics.Shader/Translation/Optimizations/HalfConversion` and replace it uses with a reinterpretation of the ushort value to the Half type followed by a cast to float.

turbedi:
Another one: [P/Invoke source generation](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation)

luca-bertero:
Seal internal types: https://github.com/dotnet/runtime/issues/49944. 
It show really big improvement in the performance

gdkchan:
Another one that I forgot is that C# not has ref fields. I think it can replace some (or all?) usages of the `Ptr<T>` type on the VP9 decoder (`Ryujinx.Graphics.Nvdec.Vp9`). Probably a bit more complicated than the other changes.

mageven:
> - `Interlocked.And`/`Interlocked.Or`

> `Ryujinx.Memory/Tracking/ConcurrentBitmap` currently uses a compare and swap look to set and clear bits on the bitmap array. Those news methods can be used to update it atomically in a more efficient way.

This part was resolved by #3937

turbedi:
Another set of new APIs introduced in .NET 6/7: Throw helper APIs like
[ArgumentNullException.ThrowIfNull](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.throwifnull?view=net-7.0#system-argumentnullexception-throwifnull(system-object-system-string))
[...]

PR #3929, opened 2022-11-28
Added Generic Math to BitUtils

HunterBarney:
Based on issue #3858. 
Replaces several BitUtils overloads with 1 generic method that accepts several types.
I will work on more of the .net 6/7 improvements when I have the time.

### DIFF
--- a/Ryujinx.Common/Utilities/BitUtils.cs
+++ b/Ryujinx.Common/Utilities/BitUtils.cs
@@ -3,11 +3,8 @@ using System.Numerics;
 
 namespace Ryujinx.Common
 {
-
     public static class BitUtils
     {
-      
-
         public static T AlignUp<T>(T value, T size)
             where T : IBinaryInteger<T>
         {
@@ -25,7 +22,6 @@ namespace Ryujinx.Common
         {
             return (value + dividend--) / dividend;
         }
-
 
         public static int Pow2RoundUp(int value)
         {
--- a/Ryujinx.Common/Utilities/BitUtils.cs
+++ b/Ryujinx.Common/Utilities/BitUtils.cs
@@ -1,83 +1,31 @@
+using System;
 using System.Numerics;
 
 namespace Ryujinx.Common
 {
+
     public static class BitUtils
     {
-        public static uint AlignUp(uint value, int size)
+      
+
+        public static T AlignUp<T>(T value, T size)
+            where T : IBinaryInteger<T>
         {
-            return (uint)AlignUp((int)value, size);
+            return (value + (size--)) & -size;
         }
 
-        public static int AlignUp(int value, int size)
-        {
-            return (value + (size - 1)) & -size;
-        }
-
-        public static ulong AlignUp(ulong value, int size)
-        {
-            return (ulong)AlignUp((long)value, size);
-        }
-
-        public static long AlignUp(long value, int size)
-        {
-            return AlignUp(value, (long)size);
-        }
-
-        public static ulong AlignUp(ulong value, ulong size)
-        {
-            return (ulong)AlignUp((long)value, (long)size);
-        }
-
-        public static long AlignUp(long value, long size)
-        {
-            return (value + (size - 1)) & -size;
-        }
-
-        public static uint AlignDown(uint value, int size)
-        {
-            return (uint)AlignDown((int)value, size);
-        }
-
-        public static int AlignDown(int value, int size)
+        public static T AlignDown<T>(T value, T size)
+            where T : IBinaryInteger<T>
         {
             return value & -size;
         }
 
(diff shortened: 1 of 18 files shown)


gdkchan:
There's some odd spacing between the methods now, please remove the extra lines.
image
Generally we only have one line separating methods in the code.

HunterBarney:
Sorry about that, nasty habit of mine

Ryujinx.Common/Utilities/BitUtils.cs
```diff
@@ -1,82 +1,26 @@
...
 namespace Ryujinx.Common
 {
     public static class BitUtils
     {
-        public static uint AlignUp(uint value, int size)
+        public static T AlignUp<T>(T value, T size)
+            where T : IBinaryInteger<T>
         {
-            return (uint)AlignUp((int)value, size);
+            return (value + (size--)) & -size;
```
gdkchan, line 11: I don't think this code is equivalent to the previous code. This seems to be doing `(value + size) & -(size -1)` while the previous one was `(value + (size - 1)) & -size`. Generally I think modifying method parameter values is a bad practice, because one can very easily see a parameter being used in the code, without noticing something else in the code changed its value and its no longer what one might expect.

The same comments also applies to the `DivRoundUp` method which was modified in a similar way.

HunterBarney: This is still doing `(value + (size -1)) & -size`
I am not able to do `(values + (size - 1))` in the return as 1 is counted as an integer and is not accepted in a generic statement, so i replaced the `size-1` with `size--` which decrements the value by one effectively doing the same thing.

gdkchan: `size--` is a post-decrement not a pre-decrement. It will use the original value and then decrement the variable.
You can try it yourself, create a new console application and do:
```cs
    static void Main(string[] args)
    {
        Console.WriteLine(AlignUp(17, 8) + " " + AlignUp2(17, 8));
    }
    
    public static int AlignUp(int value, int size)
    {
        return (value + (size--)) & -size;
    }
    
    public static int AlignUp2(int value, int size)
    {
        return (value + (size - 1)) & -size;
    }
```
It will print `25 24` which means that the results are different.

HunterBarney: Oh wow i just saw my mistake, It should be `--size` instead of `size--`, will fix now.

gdkchan: That would be sill wrong since it will be doing `-(size - 1)` rather than `-size`.

gdkchan: Yes, that sounds like the right way of doing it.

HunterBarney: Alright thank you, sorry for the trouble.

Ryujinx.Common/Utilities/BitUtils.cs
```diff
@@ -1,82 +1,26 @@
...
-        }
-
-        public static ulong DivRoundUp(ulong value, uint dividend)
-        {
-            return (value + dividend - 1) / dividend;
-        }
-
-        public static long DivRoundUp(long value, int dividend)
+        public static T DivRoundUp<T>(T value, T dividend)
+            where T: IBinaryInteger<T>
```
gdkchan, line 21:
```suggestion
            where T : IBinaryInteger<T>
```

Ryujinx.HLE/HOS/Kernel/Memory/KCodeMemory.cs
```diff
@@ -44,7 +44,7 @@ public KernelResult Initialize(ulong address, ulong size)
 
         public KernelResult Map(ulong address, ulong size, KMemoryPermission perm)
         {
-            if (_pageList.GetPagesCount() != BitUtils.DivRoundUp(size, KPageTableBase.PageSize))
+            if (_pageList.GetPagesCount() != BitUtils.AlignDown<ulong>(size, (ulong)KPageTableBase.PageSize))
```
gdkchan, line 47: I just noticed that this was changed from `DivRoundUp` to `AlignDown`. There are several other calls that were changed aswell making the code incorrect. You'll likely need to revise all those changes to make sure that the correct methods are called.

HunterBarney: This should be fixed now. I think this happened due to me relying too much on intellisense and it suggested bad edits.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

AcK77 (approved):

PR #3932, opened 2022-11-28
GPU: Always draw polygon topology as triangle fan

riperiperi:
Polygon topology wasn't really supported and would only work on OpenGL on drivers that haven't removed it. As an alternative, this PR makes all cases of polygon topology use triangle fan. The topology type and transform feedback type have not been changed, as I don't think geo shader/tfb should be used with polygons.

The OpenGL spec states:
> Only convex polygons are guaranteed to be drawn correctly by the GL.

For convex polygons, triangle fan is equivalent to polygon. I imagine this is probably how it works on device, as this get-out-of-jail-free card is too enticing to pass up.

This fixes the stat display in Pokemon S/V in Vulkan, or on OpenGL on drivers that don't support GL_POLYGON in compatibility mode.


marysaka (approved):

GlitchGamingX:
Will this update be added to the LDN build?

PR #3934, opened 2022-11-28
infra: Add distribution files for macOS

marysaka:
This upstream macOS packing and distribution files.

With this Avalonia UI currently crash on SDL2 (need #3818)


distribution/macos/create_app_bundle.sh
```diff
@@ -0,0 +1,51 @@
...
+    if ! [ -x "$(command -v rcodesign)" ];
+    then
+        echo "Cannot find rcodesign on your system, please install rcodesign."
+        exit 1
+    fi
+
+    # NOTE: Currently require https://github.com/indygreg/apple-platform-rs/pull/44 to work on other OSes.
+    # cargo install --git "https://github.com/marysaka/apple-platform-rs" --branch "fix/adhoc-app-bundle" apple-codesign --bin "rcodesign"
+    echo "Usign rcodesign for ad-hoc signing"
+    rcodesign sign --entitlements-xml-path $ENTITLEMENTS_FILE_PATH $APP_BUNDLE_DIRECTORY
```
kabiroberai, line 46: rcodesign has a few other issues with frameworks atm, see also https://github.com/indygreg/apple-platform-rs/pull/46. Consider supporting/recommending ldid on non-Darwin platforms: https://github.com/ProcursusTeam/ldid.

distribution/macos/bundle_fix_up.py
```diff
@@ -0,0 +1,609 @@
...
+            symtab_offset = offset
+        elif cmd == LC_CODE_SIGNATURE:
+            codesign_offset = offset
+
+        offset += cmdsize
+        pass
+
+    assert linkedit_offset is not None and symtab_offset is not None
+
+    # If there is a codesign section, clean it up.
```
kabiroberai, line 424: How come this is necessary? Afaik `codesign` replaces the existing signature if one exists (and if rcodesign doesn't then replacing it with ldid might help.)

distribution/macos/create_app_bundle.sh
marysaka: I have worked with rcodesign on iOS and macOS for a long time without any issues.

This bundle issue I fixed is only apparent with ADHOC codesigning of universal binaries.

distribution/macos/bundle_fix_up.py
marysaka: This code is here to handle **.NET bundle** being destroyed by vanilla LLVM version of install_name_tool.

I am explicitly reading .NET bundle to the file, and following what .NET logic that you can find [here](https://github.com/dotnet/runtime/blob/main/src/installer/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs#L202) and [here](https://github.com/dotnet/runtime/blob/main/src/installer/managed/Microsoft.NET.HostModel/AppHost/MachOUtils.cs#L368).

Please make sure to read the code fully before making this kind of comments.

I don't care how good idid could be, rcodesign is plenty enough for our usecase.

distribution/macos/create_app_bundle.sh
kabiroberai: yeah rcodesign is probably sufficient for this use case, since there aren't any framework bundles + the only Darwin platform being targeted is macOS. It isn't as well supported on other platforms, which makes sense given its scope and primary use case as a python application packager. I primarily wanted to flag this because those other use cases tend to become relevant once we're out of single-binary land and bring in things like plugins/nested frameworks.

distribution/macos/bundle_fix_up.py
kabiroberai: ah if upstream install_name_tool is the issue then consider using the build from apple's cctools instead — this fork cross-compiles to Linux: https://github.com/tpoechtrager/cctools-port. If you're interested in giving this a spin, I have a minimal toolchain available here: https://github.com/kabiroberai/darwin-tools-linux/releases/tag/v2.2.0.

And to be clear, this isn't criticism at all; I recognize that Linux->Darwin tooling is pretty sparse and unorthodox atm, so as someone who's also worked extensively in that area (though specializing slightly differently on the iOS, [Objective]C[++], and Swift side of things) I thought I'd share some feedback based on my own experiences in that field and community.

PR #3935, opened 2022-11-28
Avalonia: Clean up leftover RenderTimer & Fix minimum and initial window size

TSRBerry:
This PR removes the custom render timer which was needed when the game was still rendered in the GUI and not in an embedded window.

Additionally, the `RendererHost` doesn't use a `ContentControl` anymore and assigns the `EmbeddedWindow` directly to its `Content` property.

Aside from that the minimum and initial window size have been changed, so no black bars will be displayed when these are used.

---

This cleanup is all basically all we should do to address #3931, since it shouldn't be up to us to fix Avalonia's default behavior. Instead Avalonia should default to the refresh rate of the target monitor which would resolve this issue.


Ryujinx.Ava/Ui/Windows/MainWindow.axaml.cs
```diff
@@ -90,7 +90,9 @@ public MainWindow()
 
             Title = $"Ryujinx {Program.Version}";
 
-            Height /= Program.WindowScaleFactor;
+            // NOTE: Height is not usable here, since it would still be 0 at this point.
+            var barHeight = MenuBar.MinHeight + StatusBar.MinHeight;
```
gdkchan, line 94: Wonder if this would be still correct if we use the `NativeMenuBar` control and the bar is at the top on macOS.

Ryujinx.Ava/Program.cs
```diff
@@ -24,17 +21,14 @@ namespace Ryujinx.Ava
     internal class Program
     {
         public static double      WindowScaleFactor { get; set; }
```
gdkchan, line 23: This should be either re-aligned or the alignment should be removed entirely.

Ryujinx.Ava/Ui/Windows/MainWindow.axaml.cs
TSRBerry: I have no idea, but I can't really test this, since I don't have a mac.

AcK77 (approved):

marysaka (approved):

PR #3940, opened 2022-11-29
GPU: Periodically clear volatile flags

riperiperi:
Buffer tracking handles have an optimisation where they can be set as _volatile, which means that we assume the data is always modified. This avoids the cost of setting and clearing memory protection on this frequently modified data, as generally the game will have changed it every frame anyways.

This works fine in games that always put their commonly modified constant buffers in the same place (say, for model world matrices might be updated every use), but if the memory is freed and reused by buffer data that is modified less often, the _volatile flag could remain and force data to upload every frame that hasn't changed.

The way these flags are set is determining when multiple consecutive sequence number checks say that the buffer is dirty. If they do, then the tracking handle's _volatile flag is set and it is always assumed to be dirty.

Unfortunately, this means we lose that original source of information - we aren't tracking memory writes to the buffer, so we don't know when it _stops_ updating every frame. The only solution is to unset some of these periodically, and have them correct their state in the frame or two afterwards.

The strategy for doing this is by keeping track of the number of elapsed frames, so that we can reset volatile flags periodically. We only check this when buffer data is _uploaded_, as that is an indicator the tracking handle might be _volatile.

After a set number of frames has passed since the last check, on upload we manually set. This is done with a period of 30 frames right now - each 30 frames clears 1/64 buffer volatile flags. The flag a period clears is based on its frame number, and the buffer's page number. If a buffer isn't synchronized in a long time, it will clear all the flags for the clearing periods that it missed (for example, after 1920 frames a check will clear all flags in the buffer).

All this is attempting to minimize iteration of buffers, unnecessary checks, expensively resetting all flags at the same time... but it can still be less than ideal clearing these flags. Notably there's no feedback whether any part of a multiregionhandle is set as volatile, so we end up iterating handles even if they were never volatile. The infrequency and splitting the work into only the buffers that are uploaded, and only 1/64 of the flags every 30 frames hopefully mitigate all of this.

Fixes performance lowering dramatically over time in Pokemon Scarlet/Violet. May fix similar issues in other games, even if they are more subtle.

### DIFF
--- a/Ryujinx.Cpu/Tracking/CpuMultiRegionHandle.cs
+++ b/Ryujinx.Cpu/Tracking/CpuMultiRegionHandle.cs
@@ -15,6 +15,7 @@ namespace Ryujinx.Cpu.Tracking
             _impl = impl;
         }
 
+        public void ClearVolatile(ulong mask) => _impl.ClearVolatile(mask);
         public void Dispose() => _impl.Dispose();
         public void ForceDirty(ulong address, ulong size) => _impl.ForceDirty(address, size);
         public IEnumerable<IRegionHandle> GetHandles() => _impl.GetHandles();
--- a/Ryujinx.Cpu/Tracking/CpuRegionHandle.cs
+++ b/Ryujinx.Cpu/Tracking/CpuRegionHandle.cs
@@ -18,6 +18,7 @@ namespace Ryujinx.Cpu.Tracking
             _impl = impl;
         }
 
+        public void ClearVolatile() => _impl.ClearVolatile();
         public void Dispose() => _impl.Dispose();
         public bool DirtyOrVolatile() => _impl.DirtyOrVolatile();
         public void ForceDirty() => _impl.ForceDirty();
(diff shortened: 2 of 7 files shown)


riperiperi:
This PR is draft because things can be a bit complicated. Running around mesagoza, in  and out of loading zones for about 4 minutes in master causes the average framerate to dip significantly:

image

When forcing all buffer volatile flags to true, the issue resolves:

image

So it's definitely a big issue.

The problem is that this game changes how the buffers are used so often that some performance is still lost after moving around. With a cycle time of around a minute, you can still notice a drop in performance that gets better as you wait around. (around 5fps instead of like 20, that gradually recovers) In a less extreme area, this will probably be lower.

I'm not sure if there's a better way of detecting when buffer uses are doing this and accellerating the process if it is. We obviously don't want to affect games where this does not change at all, hence the very gradual clearing of flags I've got right now.

riperiperi:
Actually, the above result may have been caused by an accidental change made to my testing branch (with the clear volatile ui button) where registering tracking actions wasn't clearing volatile flags. There is much less of an impact when this change is not there...

It's also worth noting that some areas vary in performance wildly. Maybe this is due to cpu modified data being so close to gpu modified data, which would explain the difference with that ClearVolatile removed. I'm not getting many extra frames when clearing volatile (more like 1), but who knows if this would help in the long term for memory that isn't shared.

riperiperi:
I'm not convinced this has a meaningful effect right now, so I'm going to close it.

PR #3942, opened 2022-11-29
Ensure that vertex attribute buffer index is valid on GPU

gdkchan:
Some games will have vertex attributes with a invalid buffer index (the vertex buffer doesn't exist). This is invalid and may cause crashes on both on OpenGL and Vulkan depending on the driver. The OpenGL backend was already validating this and disabling the attribute if the buffer index was invalid. However, Vulkan wasn't. This changes does the validation on the GPU project, and removes the validation from the OpenGL backend. So now both backends should be covered.

Fixes some crashes on specific games on Vulkan with some drivers. I don't think this issue affects NVIDIA.


riperiperi (approved):
Nice catch.

marysaka (approved):

edisionnano:
Apparently this causes NVIDIA OpenGL on Linux to crash when playing Super Mario Odyssey (#4126)

PR #3943, opened 2022-11-29
Remove shader dependency on SPV_KHR_shader_ballot and SPV_KHR_subgroup_vote extensions

gdkchan:
This replaces the use of those two extensions with equivalent functionality that is supported on "base" SPIR-V and does not require any extension. This is more widely supported and required on MoltenVK as it does not support those extensions.


riperiperi (approved):
Nice, UE4 and Mario Party uses seem fine.

marysaka (approved):

PR #3953, opened 2022-11-29
chore: Update Silk.NET to 2.16.0

marysaka:
Also fixed obsolete warnings.


Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -168,12 +168,12 @@ private static SurfaceFormatKHR ChooseSwapSurfaceFormat(SurfaceFormatKHR[] avail
         {
             if (availableFormats.Length == 1 && availableFormats[0].Format == VkFormat.Undefined)
             {
-                return new SurfaceFormatKHR(VkFormat.B8G8R8A8Unorm, ColorSpaceKHR.ColorspaceSrgbNonlinearKhr);
+                return new SurfaceFormatKHR(VkFormat.B8G8R8A8Unorm, ColorSpaceKHR.PaceSrgbNonlinearKhr);
```
gdkchan, line 171: Not an issue with this PR, but maybe something to report to Silk.NET. I don't think this name is supposed to start with "Pace".

gdkchan (approved):
Nice, didn't know they did this kind of breaking change on minor version updates.

Ryujinx.Graphics.Vulkan/Window.cs
marysaka: Yeah need to open an issue on their repo about that

AcK77 (approved):

PR #3955, opened 2022-11-30
Ava GUI: Make Dialogue More Intuitive

IsaacMarovitz:
Adjusts several aspects of the settings dialogue to make it more intuitive and standard.

- `Save` renamed to `OK`
- `Close` renamed to `Cancel`
- Layout of buttons adjusted to establish a clearer visual hierarchy (Adapts to match OS)
- Made `OK` default action (shortcut with `Enter`) and accented it
- Added `Escape` shortcut to `Cancel` button

Screenshot 2022-11-30 at 11 33 10


EmulationFanatic (approved):
Verbiage and appearance LGTM. I'll leave it to another to review the method :)
Thanks!

Ryujinx.Ava/Ui/ViewModels/SettingsViewModel.cs
```diff
@@ -474,11 +479,40 @@ public void SaveSettings()
             MainWindow.UpdateGraphicsConfig();
 
             _previousVolumeLevel = Volume;
+
+            if (_owner is SettingsWindow owner)
```
AcK77, line 483: Isn’t `_owner` type always `SettingsWindow` since it’s the model class?

IsaacMarovitz: `_owner` is type `StyleableWindow`. I can just do `_owner as SettingsWindow`, but that is a possible `NullReferenceException`/

AcK77 (approved):

marysaka (approved):

Squall-Leonhart:
Introduced a regression on the Controller Applet

"SettingsButtonClose"
image

PR #3956, opened 2022-11-30
hle: Massive cleanup of everything except services

AcK77:
This PR cleanup the full HLE project, services will cleaned up later:

- header `using` cleaned and sorted.
- `Class name = new Class()` -> `Class name = new()`.
- `var name = new Class()` -> `Class name = new()`
- `using` are inlined when possible.
- `Slice` use now `[]`.
- New line at EOF have been removed for consistencies.
- Some littles changes here and there.


gdkchan:
> var name = new Class() -> Class name = new()

I prefer `var name = new Class()` because we use var in other cases aswell (like for assigning the return value from a method call for example), and it's nice to keep it consistent.

AcK77:
Well when you have :
```
ClassA name1 = new ClassA();
var name2 = new ClassB();
```

It's more consistent to have :
```
ClassA name1 = new();
ClassB name2 = new();
```
than :
```
ClassA name1 = new();
var name2 = new ClassB();
```

That's what I've changed. I haven't remove all `var`.

gdkchan:
```cs
var name1 = new ClassA();
var name2 = new ClassB();
```
Is also an option.

marysaka:
> ```cs
> var name1 = new ClassA();
> var name2 = new ClassB();
> ```
> 
> Is also an option.

If the object creation is deferred later in the code this is terrible to read imo.

I should remind you that because of my eyes condition, I would really prefer having type stated at declaration time, otherwise this will just make me unable to work on HLE anymore.

marysaka (changes requested):
We need to put line return on every files instead of removing it because it makes it hard to `cat` or `grep` the project

PR #3958, opened 2022-11-30
Implement non-MS to MS copies with draws

gdkchan:
This implements the non-multisample to multisample copies using draws on Vulkan. This is required because the Apple Silicon GPU does not support image load or store to multisample images. The multisample to non-multisample copy was also simplified and no longer uses specialization constants.

On top of that, this has a change that reduces the sample count on texture creation to something that the GPU supports. This is necessary because the Apple Silicon GPU only supports a sample count of up to 4 and some games uses more than that.


riperiperi (approved):
LGTM, I don't have any of the games that were affected by this, I guess it's probably worth checking them to make sure they aren't blurrier?

marysaka (approved):

PR #3968, opened 2022-11-30
Sanitise Configuration State for OpenGL on macOS

IsaacMarovitz:
+ Check when loading the configuration file if the graphics backend has been set to OpenGL and force switch to Vulkan
+ Change default graphics backend on macOS to Vulkan
+ If graphics backend is changed to OpenGL at runtime, force back to Vulkan


Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -1191,6 +1191,13 @@ public ConfigurationLoadResult Load(ConfigurationFileFormat configurationFileFor
 
                 configurationFileFormat.EnableMacroHLE = true;
             }
+            
+            if (OperatingSystem.IsMacOS() && configurationFileFormat.GraphicsBackend == GraphicsBackend.OpenGl)
```
AcK77, line 1195: This should be checked In `LogValueChange` by checking the type and then sets the right value if it’s macOS.

You could add a `using` for the Logging class too.

AcK77:
Instead of put a logging, adding a check in LoadDefault and when the backend is changed, it could be probably better to force the GraphicBackend field to Vulkan everytime if it's on macOS ?

gdkchan:
I think it would be better if we had a way to check if each backend is supported on the current platform. Then we could take action from here, I think showing an error message saying that OpenGL or Vulkan is not supported would be enough.

IsaacMarovitz:
> I think it would be better if we had a way to check if each backend is supported on the current platform. Then we could take action from here, I think showing an error message saying that OpenGL or Vulkan is not supported would be enough.

I will generalise the approach in this PR slightly in order to make it more customisable in future.

PR #3969, opened 2022-12-01
amadeus: Fix wrong SendCommands logic

marysaka:
A very dumb mistake oops.

Might help with audio desync, might cause audio stutters, will see!

Need heavy testing on **every games that use the audio renderer**

### DIFF
--- a/Ryujinx.Audio/Renderer/Dsp/AudioProcessor.cs
+++ b/Ryujinx.Audio/Renderer/Dsp/AudioProcessor.cs
@@ -116,7 +116,7 @@ namespace Ryujinx.Audio.Renderer.Dsp
             };
         }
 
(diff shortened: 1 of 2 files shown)


AcK77 (approved):

Ryujinx.Audio/Renderer/Dsp/AudioProcessor.cs
```diff
@@ -116,6 +116,11 @@ public void Send(int sessionId, CommandList commands, int renderingLimit, ulong
             };
         }
 
+        public bool HasRemainCommands(int sessionId)
```
gdkchan, line 119: `HasRemainingCommands` would be a better name.

gdkchan (approved):
Code lgtm, but I did not test audren games (I believe we already had some people test it).

PR #3974, opened 2022-12-01
Refactor codebase

tygyh:
A few refactorings focusing on reducing nesting and making the un-nesting easier. There is more to come if this is accepted.

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
@@ -748,7 +748,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Spirv
                 pCoords = Src(AggregateType.S32);
             }
 
-            pCoords = ScalingHelpers.ApplyScaling(context, texOp, pCoords, intCoords: true, isBindless, isIndexed, isArray, pCount);
+            pCoords = ScalingHelpers.ApplyScaling(context, texOp, pCoords, isBindless, isIndexed, isArray, pCount);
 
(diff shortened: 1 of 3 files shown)


marysaka:
You seems to have messed up your rebase with master here

marysaka (changes requested):
I don't see the benefits of inverting all the if here, this makes it harder to read or understand (Not to mention that this doesn't follow the project codestyle overall).

The changes made in aa73a9429a0f8661e6b378f649f7dca514781f83 complexity the usage of ApplyScaling so I’m also against it.

Otherwise the changes in 559858405d42c1382a7be6023920d43ac43da717, b81383c286f0560e96b1b17417581aeefb40d52c, and c2da63bed14e69bae206cf9036d79b28ef47c758 are welcome.

tygyh:
I solved it! Now back to the discussion

marysaka (changes requested):
Not a fan of the ApplyScaling changes as it just increase the complexity the caller side.

tygyh:
In my opinion the complexity of figuring out what the method does is slightly lowered by removing one of the parameters

marysaka:
> In my opinion the complexity of figuring out what the method does is slightly lowered by removing one of the parameters

Well it apply scaling depending of the parameters in input? I don't see why moving the condition were it will actually do the scaling, improve anything here?

tygyh:
I would calculate the complixity of the parameters as n! where n represents the number of parameters. By moving the intCoords out of the method we reduce the complexity of the ordering from 6! (720) to 5! (120). I would also like to reduce it further by moving the calculation of index out of the method and sending it in as a parameter thus reducing the complixity to 3! (6), but that's a discussion for later.

PR #3995, opened 2022-12-03
Change default Vsync toggle hotkey to F1 instead of Tab

MetrosexualGarbodor:
People are constantly inadvertently uncapping their framerates with the hotkey being set to Tab due to Alt Tabbing. Furthermore, Avalonia uses Tab for UI navigation, so it doesn't even work there. As more people switch to Avalonia, it becomes more apparent that the default needs to change.

Chose F1 specifically because it's the closest to Tab. Was considering F9 instead as that has less chances of misclicking F2 (mute audio) or Esc (quit emulation), though I leave that up for debate. Regardless, it can be customized on Avalonia, so not too big of a deal.


marysaka (approved):

AcK77 (approved):

Littl3Guy:
The default is still Tab after updating to 1.1.433:
Untitled

PR #3996, opened 2022-12-03
Update 'OpenGL Log Level' to 'Graphics Backend Log Level'

Candygoblen123:
This PR updates the setting "OpenGL Log Level" to "Graphics Backend Log Level" to better reflect that the setting effects both OpenGL logging and Vulkan Logging.

Screenshot 2022-12-02 at 8 38 15 PM


edisionnano:
What about GTK?

Candygoblen123:
> What about GTK?

GTK already seems to say Graphics Backend Log Level. 

See https://github.com/Ryujinx/Ryujinx/blob/ab0491817e87b9dd134c41764ba213f8c8559e9b/Ryujinx/Ui/Windows/SettingsWindow.glade#L2650 and this screenshot: 
image

gdkchan:
I think it's better to also change they key name so that it doesn't include "Opengl", and then remove it from all the other locale files so that they can be re-translated.

Candygoblen123:
I've changed the keys from `"SettingsTabLoggingOpenglLogLevel"` to `"SettingsTabLoggingGraphicsBackendLogLevel"` including the keys for the drop-down. I also removed the `"SettingsTabLoggingOpenglLogLevel"` from all other Locals, but just renamed the keys for the drop-down, keeping the translation intact.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4003, opened 2022-12-03
GPU: Swap bindings array instead of copying

riperiperi:
Reduces work on UpdateShaderState. Now the cost is a few reference moves for arrays (for stages with bindings), rather than copying data. Cached bindings arrays are created when shaders are loaded.

1-2 fps here on the usual spot in SMO.

Upside: Removes a bunch of redundant code from ComputeClass, cheaper UpdateShaderState.
Downside: bindings arrays are no longer readonly (though it was only the top level array that was readonly, not the contained ones).


riperiperi:
I'm not sure what to do about the default arrays for `textureBindings` etc that are created. If the, it could still be possible for these arrays to be called without a shader being set, I guess? Right now they are left in.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4004, opened 2022-12-03
GPU: Use lazy checks for specialization state

riperiperi:
This PR adds a new class, the SpecializationStateUpdater, that allows elements of specialization state to be updated individually, and signal the state is checked when it changes between draws, instead of building and checking it on every draw. This also avoids building spec state switching shaders.

Most state updates have been moved behind the shader state update, so that their specialization state updates make it in before shaders are fetched.

Downside: Fields in GpuChannelGraphicsState are no longer readonly. To counteract copies that might be caused this I pass it as `ref` when possible, though maybe `in` would be better? Not really sure about the quirks of `in` and the difference probably won't show on a benchmark.

The result is around 2 extra FPS on SMO here in the usual spot. Not much right now, but it will remove costs when we're doing more expensive specialization checks, such as fragment output type specialization for macos. It may also help more on other games with more draws.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs
@@ -116,9 +116,10 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         }
 
         /// <summary>
-        /// Indicates whether alpha-to-coverage is enabled.
+        /// Updates alpha-to-coverage state, and sets it as changed.
         /// </summary>
-        /// <param name="value">The new value</param>
+        /// <param name="enable">Whether alpha-to-coverage is enabled</param>
+        /// <param name="ditherEnable">Whether alpha-to-coverage dithering is enabled</param>
         public void SetAlphaToCoverageEnable(bool enable, bool ditherEnable)
         {
             _graphics.AlphaToCoverageEnable = enable;
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs
@@ -1,0 +1,279 @@
+﻿using Ryujinx.Common.Memory;
+using Ryujinx.Graphics.GAL;
+using Ryujinx.Graphics.Gpu.Shader;
+using Ryujinx.Graphics.Shader;
+
+namespace Ryujinx.Graphics.Gpu.Engine.Threed
+{
+    /// <summary>
+    /// Maintains a "current" specialiation state, and provides a flag to check if it has changed meaningfully.
+    /// </summary>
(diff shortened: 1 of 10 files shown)


Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs
```diff
@@ -0,0 +1,279 @@
...
+                _graphics.TessellationMode = value;
+
+                Signal();
+            }
+        }
+
+        /// <summary>
+        /// Indicates whether alpha-to-coverage is enabled.
+        /// </summary>
+        /// <param name="value">The new value</param>
```
gdkchan, line 121: Wrong XML docs (parameters are `enable` and `ditherEnable`).

Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -16,9 +17,9 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
     /// </summary>
     class StateUpdater
     {
-        public const int ShaderStateIndex = 16;
+        public const int ShaderStateIndex = 26;
```
gdkchan, line 20: I think we should expose properties with those values and query the correct value by name on the class constructor. Thee current approach is too error prone. Doesn't need to be done right now though.

gdkchan (commented):
`GetGraphicsState` and `GetPoolState` on `StateUpdater` seems to be unused  now, so I guess they can be removed?

Also, it seems that the state on `SpecializationStateUpdater` is updated when each state update method is called. Since that happens after the the specialization checks that would invalidate the current shader, isn't it just comparing state from the previous draw?

riperiperi:
I assumed since the spec check was done there before, the state update had already happened, but that wouldn't make sense as it sets the shader dirty flag. I've updated it to reload after (UpdateShaderState rather than set the flag), and added a comment about the behaviour should not change _vtgWritesRtLayer.

gdkchan (approved):
lgtm, but it's a bit unfortunate that there's now 2 places that might cause the shader to be updated (one for state spec and other for textures spec).

riperiperi:
It kind of sucks, but I can at least explain how it works now:

- UpdateShaderState is up first. If the shader state is dirty, it does the full check with textures and everything.
- Check before bindings to see if graphics state is dirty after processing updates, and reloads the shader if it is different. If the shader was changed, this will always do nothing, as the state dirty flag was reset as it used the current state fo find the shader earlier.
  - I wasn't comfortable combining this one with the texture bindings check, as that would mean that none of the bindings are allowed to change due to spec state. I'm not sure that would always be true for graphics spec state now or in the future.
- Texture bindings check as they are fetched, and reload the shader if different. If the shader was changed by _either_ of the above two things, the texture state will already match.

At least it should only ever reload once, it's just a case of when.

marysaka (approved):

PR #4005, opened 2022-12-03
Use source generated regular expressions

SukharevAndrey:
This PR replaces all regular expressions with [source generated ones](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators). Small change, but it (at least in theory) should enable better application trimming and NativeAOT friendliness. The only downside is that classes need to be made partial.


gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #4008, opened 2022-12-04
Add support for installing firmware from an NSP file.

Steveice10:
Adds support for installing firmware from an NSP file containing all of the firmware NCAs. Since code already exists for installing from an XCI update partition, most of it was just reused but with the NSP's PartitionFileSystem instead.

Tested on both Avalonia and GTK UIs with my own 15.0.1 NSP dump.

### DIFF
--- a/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -23,7 +23,7 @@
   "MenuBarActionsScanAmiibo": "Scan An Amiibo",
   "MenuBarTools": "_Tools",
   "MenuBarToolsInstallFirmware": "Install Firmware",
-  "MenuBarFileToolsInstallFirmwareFromFile": "Install a firmware from XCI or ZIP",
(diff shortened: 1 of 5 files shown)


marysaka:
I don't see what could generate NSP for dumped firmware here...?

Steveice10:
NSPs can be a convenient format for holding arbitrary sets of Switch contents that is consistent with how game contents are stored. I'm not sure if it's a widely supported dumping option but I've been using it myself for my own firmware dumps.

Since it re-uses the same code-paths as the existing XCI update partition support, it didn't seem like a big leap to also just support installing firmware from an NSP as another option.

jduncanator:
I don't think the problem is necessarily with implementing the functionality technically, but more so with the user experience of such option

Given NSP dumps are so common, yet NSP dumps with Update partitions are incredibly rare (essentially non-existent), adding this option is likely to increase confusion with end users when they attempt to install Firmware from a standard NSP eShop dump, and then wonder why it isn't working.

A standard XCI will _always_ have an update partition (I don't include trimmed XCI as standard), and so is a much better user experience.

Steveice10:
Fair enough, that's something I hadn't considered. Will just close this out then to avoid potential confusion.

PR #4011, opened 2022-12-04
Restrict shader storage buffer search when match fails

gdkchan:
Right now, when the match for storage buffer fails, we generate code to check the current address against the address for all possible 16 storage buffers until we find a match. However it's very unlikely that a shader will use all 16, so this PR instead modifies the check so that only the storage buffers that might be accessed by the shader are checked. This is done based on the constant buffer (0) ranges that are accessed on the shader.

The advantages of doing this is that the shader code is smaller/a bit more efficient in the cases where the match fails. This should also help to avoid an error on macOS where it gets out of storage buffer bindings when taking unused storage buffers for other purposes.

Potential downside is that the cases where the base address comes from something other than the constant buffer 0 would not work, but I don't think this is actually possible since those accesses are compiler generated.

Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Shader/Translation/Optimizations/GlobalToStorage.cs
+++ b/Ryujinx.Graphics.Shader/Translation/Optimizations/GlobalToStorage.cs
@@ -8,14 +8,25 @@ namespace Ryujinx.Graphics.Shader.Translation.Optimizations
 {
     static class GlobalToStorage
     {
-        public static void RunPass(BasicBlock block, ShaderConfig config)
+        public static void RunPass(BasicBlock block, ShaderConfig config, ref int sbUseMask)
         {
             int sbStart = GetStorageBaseCbOffset(config.Stage);
-
             int sbEnd = sbStart + StorageDescsSize;
 
             for (LinkedListNode<INode> node = block.Operations.First; node != null; node = node.Next)
             {
+                for (int index = 0; index < node.Value.SourcesCount; index++)
+                {
+                    Operand src = node.Value.GetSource(index);
+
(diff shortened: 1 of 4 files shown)


marysaka (approved):

Ryujinx.Graphics.Shader/Translation/Optimizations/GlobalToStorage.cs
```diff
@@ -8,14 +8,25 @@ namespace Ryujinx.Graphics.Shader.Translation.Optimizations
...
-
             int sbEnd = sbStart + StorageDescsSize;
 
             for (LinkedListNode<INode> node = block.Operations.First; node != null; node = node.Next)
             {
+                for (int index = 0; index < node.Value.SourcesCount; index++)
+                {
+                    Operand src = node.Value.GetSource(index);
+
+                    int storageIndex = GetStorageIndex(src, sbStart, sbEnd);
```
riperiperi, line 22: Is this called on every single source in a shader? I feel this could become very, very expensive, and even worse if the phi node lookup is improved.

riperiperi: Ah, I thought it was one of the fuller versions that checked for the addition, rather than just the constant buffer. This is actually fine.

riperiperi (approved):
Looks good, I suppose this will also avoid accidentally syncing or setting storage buffers as written that shouldn't even be bound.

The phi node traversal is pretty limited right now, it seems to break a lot for LDG on compute shaders. I might look into improving it.

PR #4013, opened 2022-12-04
Include a start.sh file with correct launch options

RMED24:
On quite a few occasions I've seen people launch Ryujinx on linux with different methods. Most people seem to be fine, but I have noticed the occasional issue from users that have occurred at least once in the past.

One of these issues would be people trying to launch Ryujinx in wayland and having it crash on them, which is rather annoying...

Another issue is to do with the missing libraries on some distros. Whilst this on its own wouldn't solve the issue. It could potentially be further looked into to make sure the correct dependencies are installed on each distro (such as ffmpeg or libx11-dev)

This isn't final and stuff is likely to be changed. But I want reviews to be done first however...


RMED24:
I think it's just best I wait for the build job to complete first. I don't think I'll get it on first try, but hey it is worth trying

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -66,6 +66,13 @@
       <TargetPath>THIRDPARTY.md</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+   
+   <ItemGroup>
+     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' != 'win10-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'">
```
TSRBerry:
```suggestion
     <ContentWithTargetPath Include="..\distribution\linux\Ryujinx.sh" Condition="'$(RuntimeIdentifier)' != 'win10-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'">
```
How about renaming this to `Ryujinx.sh`?

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -66,6 +66,13 @@
...
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+        <TargetPath>start.sh</TargetPath>
```
TSRBerry, line 73:
```suggestion
        <TargetPath>Ryujinx.sh</TargetPath>
```
Same here

Ryujinx/Ryujinx.csproj
```diff
@@ -55,6 +55,13 @@
       <TargetPath>THIRDPARTY.md</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+   
+   <ItemGroup>
+     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' != 'win10-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+        <TargetPath>start.sh</TargetPath>
```
TSRBerry, line 62: Here too

distribution/linux/start.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+# We recommend using Pine-jinx if you want the game to be optimised and installed for your system. This is mostly for PR builds and people who want a portable install.
```
marysaka, line 2:
```suggestion
```
No we recommend FlatHub release.

Ryujinx/Ryujinx.csproj
```diff
@@ -55,6 +55,13 @@
       <TargetPath>THIRDPARTY.md</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+   
+   <ItemGroup>
+     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' != 'win10-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'">
```
marysaka, line 60:
```suggestion
     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' == 'linux-x64'">
```

Ryujinx.Ava/Ryujinx.Ava.csproj
marysaka, line 71:
```suggestion
     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' == 'linux-x64'">
```

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -66,6 +66,13 @@
       <TargetPath>THIRDPARTY.md</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+   
+   <ItemGroup>
+     <ContentWithTargetPath Include="..\distribution\linux\start.sh" Condition="'$(RuntimeIdentifier)' != 'win10-x64' AND '$(RuntimeIdentifier)' != 'osx-x64'">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
TSRBerry, line 72: Currently the output binary for Avalonia is called `Ryujinx.Ava`. To make this script work you'll need to set the `AssemblyName` in this csproj to `Ryujinx` too.

RMED24:
I got it right first try... Re-basing the PR and opening it up for a merge... Adding checks for dependencies would be nice although not required for a merge...

distribution/linux/start.sh
```diff
@@ -0,0 +1,9 @@
...
+# It is also recommended to run start.sh in the terminal, although this isn't required in order for it to operate
+SCRIPT_DIR=$(dirname $0)
+chmod +x "$SCRIPT_DIR/Ryujinx"
+
+DOTNET_EnableAlternateStackCheck=1 GDK_BACKEND=x11 "$SCRIPT_DIR/Ryujinx" "$@"
```
marysaka, line 7: I would prefer ``GDK_BACKEND`` to be set in Ryujinx's Program.cs for GTK3 frontend.

marysaka (changes requested):

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -66,6 +66,13 @@
       <TargetPath>THIRDPARTY.md</TargetPath>
     </ContentWithTargetPath>
   </ItemGroup>
+   
+   <ItemGroup>
```
gdkchan, line 70: Indentation is off on all lines you added since you use 3 spaces, everything else uses 2.

distribution/linux/start.sh
TSRBerry:
```suggestion

# We recommend using Pine-jinx if you want the game to be optimized and installed for your system. This is mostly for PR builds and people who want a portable install.
```
One empty line after the shebang line and a small typo.

If we are shipping a start script we should recommend using that over everything else. But since Pine-jinx already exists, I'm not sure what the best way to handle this is.

distribution/linux/start.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+# We recommend using Pine-jinx if you want the game to be optimised and installed for your system. This is mostly for PR builds and people who want a portable install.
+# It is also recommended to run start.sh in the terminal, although this isn't required in order for it to operate
+SCRIPT_DIR=$(dirname $0)
+chmod +x "$SCRIPT_DIR/Ryujinx"
```
marysaka, line 5: Instead of this, make the CI use [add_tar_exec.py](https://github.com/Ryujinx/Ryujinx/blob/master/distribution/misc/add_tar_exec.py) for the release job.

gdkchan (commented):
For FFmpeg we should probably just include a pre-built library like we do on Windows, to make sure it will use a library that works.

Seems weird that you need to launch the emulator through a script rather than directly. But I don't know a better way to do this so...

distribution/linux/start.sh
TSRBerry, line 4: Is this still necessary currently? If so, then we should check if the binary already has execute permissions.

distribution/linux/start.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
+# We recommend using Pine-jinx if you want the game to be optimised and installed for your system. This is mostly for PR builds and people who want a portable install.
+# It is also recommended to run start.sh in the terminal, although this isn't required in order for it to operate
```
TSRBerry, line 2:
```suggestion
```
I think this comment is not needed. The people testing PRs will probably already know to run the script in the terminal.

distribution/linux/start.sh
```diff
@@ -0,0 +1,9 @@
...
+SCRIPT_DIR=$(dirname $0)
```
TSRBerry, line 3:
```suggestion
SCRIPT_DIR=$(dirname $(realpath $0))

```
Just in case someone starts to mess with symlinks.

TSRBerry (dismissed):
I like this idea! There are already a lot of Linux specific things that need to be done, so we might as well include a script for that.

distribution/linux/start.sh
RMED24: > Is this still necessary currently? If so, then we should check if the binary already has execute permissions.

I wanted to be sure, because my worst nightmare is always those who include files that I have to chmod myself. It's also confusing for beginners coming from windows

RMED24:
> Seems weird that you need to launch the emulator through a script rather than directly. But I don't know a better way to do this so...

I don't disagree, but I can't think of another way either

Ryujinx.Ava/Ryujinx.Ava.csproj
RMED24: I forgot about that. I will look into this soon

Ryujinx.Ava/Ryujinx.Ava.csproj
RMED24: Done

distribution/linux/start.sh
RMED24: This line was completely removed, and is no longer relevant now

distribution/linux/start.sh
TSRBerry: Look at this comment, I think it's the better solution:
https://github.com/Ryujinx/Ryujinx/pull/4013#discussion_r1039019263

Ryujinx.Ava/Ryujinx.Ava.csproj
RMED24: 7037c075affbdf3fe0044dc0e9a3ee3cd31a3092 should close this issue

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -6,7 +6,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Version>1.0.0-dirty</Version>
     <DefineConstants Condition=" '$(ExtraDefineConstants)' != '' ">$(DefineConstants);$(ExtraDefineConstants)</DefineConstants>
-    <RootNamespace>Ryujinx.Ava</RootNamespace>
+    <RootNamespace>Ryujinx</RootNamespace>
```
TSRBerry, line 9:
```suggestion
    <RootNamespace>Ryujinx.Ava</RootNamespace>
    <AssemblyName>Ryujinx</AssemblyName>
```

RMED24:
I need 7037c075affbdf3fe0044dc0e9a3ee3cd31a3092 to be double checked in order to make sure this is right and doesn't break anything. If it does, _please_ let me know

Ryujinx.Ava/Ryujinx.Ava.csproj
RMED24: Legend

Ryujinx.Ava/Ryujinx.Ava.csproj
```diff
@@ -7,6 +7,7 @@
     <Version>1.0.0-dirty</Version>
     <DefineConstants Condition=" '$(ExtraDefineConstants)' != '' ">$(DefineConstants);$(ExtraDefineConstants)</DefineConstants>
     <RootNamespace>Ryujinx.Ava</RootNamespace>
+    <AssemblyName>Ryujinx</AssemblyName>
```
marysaka, line 10: You shouldn't do that considering the updater on Avalonia take current executable name and try to rerun itself that way.

This will make it bug at relaunch

marysaka (changes requested):

Ryujinx.Ava/Ryujinx.Ava.csproj
RMED24: > You shouldn't do that considering the updater on Avalonia take current executable name and try to rerun itself that way.
> 
> This will make it bug at relaunch

The only other solutions I can see to this is either packing 2 seperate scripts for the avalonia and GTK build, or just flat out renaming Ryujinx.Ava. If there's another solution or we want to go ahead with one of these changes, please let me know

RMED24:
This GDK_BACKEND variable is something I will compile and test now on wayland myself. As Mary said it wouldn't work and TSR said it did. I will compile this and confirm it

RMED24:
This GDK_BACKEND variable is something I will compile and test now on wayland myself. As Mary said it wouldn't work and TSR said it did. I will compile this and confirm it

RMED24:
> This GDK_BACKEND variable is something I will compile and test now on wayland myself. As Mary said it wouldn't work and TSR said it did. I will compile this and confirm it

@TSRBerry looks like this still causes a crash and hasn't fixed the issue at hand. There needs to be more done it seems...

RMED24:
From launching the Ryujinx binary directly... We have liftoff! 

image

Ryujinx.Ava/Ryujinx.Ava.csproj
TSRBerry: > You shouldn't do that considering the updater on Avalonia take current executable name and try to rerun itself that way.

@marysaka what if we push an update (before this PR gets merged) that uses `Ryujinx` as the next executable if `Ryujinx.Ava` is not available?

TSRBerry:
Needs a rebase

RMED24:
I promise I'll get this all finished up in the next few days. I've had to deal with work and college which has given me little time to finish this... I will create a quick to-do list for myself right now so I can keep on top of it.
- Solve the conflict and rebase my PR to the latest version
- Make sure Ryujinx.sh is executable
Once these are solved, I will ping the testers role in order to ask people on Sway, GNOME and KDE plasma-wayland-session to test this works properly when they have xwayland installed... Things are looking good so far however

RMED24:
> SDL is switching back to defaulting to wayland soon.
> 
> I'm not sure if that also breaks Ryujinx but probably worth investigating and setting that to x11 too if it causes issues while you're at it.

I'll have a look online and keep an eye out just to be sure. Thanks for letting me know

marysaka:
Need a rebase

RMED24:
> Need a rebase

Resolved

RMED24:
Wtf when did I request a review I never did that

TSRBerry:
I changed a few file paths in distribution and renamed `ReleaseInformation.cs`.
That's why I also opened a PR on the flathub repo, as it would break after merging this PR.

marysaka (approved):

Melechtna:
This is bad, and we shouldn't be doing this, instead we should be working on getting Wayland working properly, rather than just throwing it out.

The issue seems to stem from Gdk not being configured to accept wayland windows

00:00:00.027 |E| .NET TP Worker (Ryujinx:3347901): Gdk-CRITICAL **: 13:45:14.007: gdk_x11_window_get_xid: assertion 'GDK_IS_X11_WINDOW (window)' failed

This part here, I've gone in and reverted what this commit does, so I could trouble shoot the actual issue instead of bandaid it.

While I don't have a precise solution to offer, as I typically work with Qt over GDK, likely you'd need to add something to detect when running in a wayland session, or if I'm not mistaken, make its draw calls agnostic to x11 or wayland, and let the DE handle when to use either as with Qt. All that need be done to test things, is to remove or comment out lines 107 and 108 in Ryujinx/program.cs as of the 1117 release. I'm not sure where these "fixes" were shoved into ava, but I'm unable to find them and may be else where, so I can't really confirm if ava might just be fine without forcing everything to run in xWayland.

I am unable to give a more proper log, as of right now this crashes so hard that it starts making other things crash within the system, so besides this error, the logger never has a chance to print anything else out at the moment, assuming it would even if GDK was configured correctly.

TSRBerry:
I'm sorry but this PR has nothing to do with wayland.

RMED24:
> I'm sorry but this PR has nothing to do with wayland.

No Melechtna is right in some regard. This PR was the one where ryujinx always launched in xwayland mode no matter what, as the backend was set to x11 in the ``Program.cs`` file. If you remember, ryujinx got native wayland support before this but it ONLY worked in the menus. It would crash as soon as a game was open. 

This was a bandaid solution as nobody could really suggest better at the time, and I wanted it to be a one click solution for people. I wanted this solution because some ryujinx users were having it crash, and having to set an env environment isn't an ideal solution. Especially to someone who is a beginner.

Maybe a new PR should be opened if it can be fully figured out, and ryujinx can become a native wayland application. After all, GTK and (I believe) avalonia both support wayland.

Melechtna:
> > I'm sorry but this PR has nothing to do with wayland.
> 
> No Melechtna is right in some regard. This PR was the one where ryujinx always launched in xwayland mode no matter what, as the backend was set to x11 in the `Program.cs` file. If you remember, ryujinx got native wayland support before this but it ONLY worked in the menus. It would crash as soon as a game was open.
> 
> This was a bandaid solution as nobody could really suggest better at the time, and I wanted it to be a one click solution for people. I wanted this solution because some ryujinx users were having it crash, and having to set an env environment isn't an ideal solution. Especially to someone who is a beginner.
> 
> Maybe a new PR should be opened if it can be fully figured out, and ryujinx can become a native wayland application. After all, GTK and (I believe) avalonia both support wayland.

The trouble being, and if you look what keeps getting linked back here, is this bandaid solution is being used as a "we fixed it" answer, when the opposite is true. I'm having trouble locating the code responsible for this xid method, but likely, from what I'm reading, we'd want to swap this out for process or window IDs instead, and it should make the code x11 and wayland agnostic, so it draws to the screen regardless of which environment its in, but I'd need to know what code handles this action to try and implement a proper fix. I'd open a PR, but as I said, it'd likely just get shut down and linked back here anyway.

PR #4017, opened 2022-12-05
Bsd: Implement Select

TSRBerry:
A while ago @gdkchan already implemented `Select`, but never created a PR for it.
This PR uses this implementation and adapts it to the current Ryujinx version. 

With this implementation @EmulationFanatic was able to start a LAN game in the Saints Row games.

---

Note: The implementation was tested before I cleaned it up. I'm sure I didn't break anything by cleaning it up, but just in case, let's wait for confirmation that everything still works as expected.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -1,5 +1,7 @@
 ﻿using Ryujinx.Common;
 using Ryujinx.Common.Logging;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Types;
 using Ryujinx.Memory;
 using System;
 using System.Collections.Generic;
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -29,39 +29,6 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
             _isPrivileged = isPrivileged;
         }
 
-        private void BuildMask(List<Socket> sockets, Span<byte> mask)
-        {
-            foreach (Socket socket in sockets)
-            {
-                int fd = _context.RetrieveFileDescriptor(socket);
-
-                mask[fd >> 3] |= (byte)(1 << (fd & 7));
-            }
-        }
-
-        private List<Socket> FillFromFdMask(ReadOnlySpan<byte> mask)
-        {
-            List<Socket> sockets = new();
-
-            for (int i = 0; i < mask.Length; i++)
-            {
-                byte current = mask[i];
-
-                while (current != 0)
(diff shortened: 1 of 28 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -202,12 +236,43 @@ public ResultCode Open(ServiceCtx context)
...
+
+            System.Net.Sockets.Socket.Select(readSockets, writeSockets, errorSockets, timeout);
+
+            using var readFdsOut  = context.Memory.GetWritableRegion(readFdsOutBufferPosition, (int)readFdsOutBufferSize);
+            using var writeFdsOut = context.Memory.GetWritableRegion(writeFdsOutBufferPosition, (int)writeFdsOutBufferSize);
+            using var errorFdsOut = context.Memory.GetWritableRegion(errorFdsOutBufferPosition, (int)errorFdsOutBufferSize);
+
+            BuildMask(readSockets, readFdsOut.Memory.Span);
+            BuildMask(writeSockets, writeFdsOut.Memory.Span);
+            BuildMask(errorSockets, errorFdsOut.Memory.Span);
```
marysaka, line 273: Not a huge fan to have the socket logic in here, we have abstraction around for Poll...

Move to ManagedSocketPollManager and extend IPollManager? (might be worth renaming it)

I also think it is possible to Select an eventfd (EventFileDescriptor), might be worth adding for completeness

marysaka (changes requested):

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
TSRBerry: When I first implemented this method I created a new struct for FileDescriptorSets and tried to use Poll as a reference to implement Select. But I couldn't really understand Poll, so I never finished my implementation.

> Move to ManagedSocketPollManager and extend IPollManager? (might be worth renaming it)

I'm not quite sure how to go about that, since Poll looks very different from Select. I locally added Select to IPollManager with this signature:
```cs
LinuxError Select(List<ISocket> readSockets, List<ISocket> writeSockets, List<ISocket> errorSockets, int timeoutMilliseconds, out int updatedCount)
```
But this won't be compatible with Poll. How should I go about this?

TSRBerry: Hmm, I attempt to move Select to IPollManager, but I feel like I just made things a lot worse than before.

https://github.com/Ryujinx/Ryujinx/pull/4017/commits/281eea265939364912bf965b66c06190b39a09bc

TSRBerry: Redid this again and this time I feel like I understood what you meant. Hopefully it's better this way!

marysaka (approved):

EmulationFanatic:
Also enables [Just Die Already](https://github.com/Ryujinx/Ryujinx-Games-List/issues/4131) to boot and play normally (hangs forever at a black screen otherwise).
image

AcK77 (approved):

PR #4029, opened 2022-12-05
Fix shaders with global memory access from unknown locations

gdkchan:
Fixes a regression from #4011 where shaders with global memory access from unknown location would do the global -> storage replacement, but since there are no storage buffers declared, it would fail to generate the SPIR-V, resulting in crashes. Affected game is Mario Golf Super Rush, and maybe a few more (but generally those games are not working properly anyway, with the exception of Mario Golf, since the global memory access will be wrong in those cases).


marysaka (approved):

AcK77 (approved):

riperiperi (approved):
LGTM, I suppose this will complicate things with the ube lookup in the other PR potentially running solo.

PR #3905, opened 2022-11-26
Sanitize paths in stack traces

mageven:
Some users are confusing build machine paths in stack traces with their own install paths. This change replaces the project directory paths with just the project names.

Uses a `PathMap` in `Directory.Build.props` to define a common property for all projects. Let me know if you prefer a per-project property or conversely, gathering more properties into it as well (`TargetFramework` seems like a good candidate).

Example-
```
00:04:42.739 |E| HLE.GuestThread.20 Application : Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: Attempted to access an invalid memory region.
   at Ryujinx.Memory.Tracking.MemoryTracking.VirtualMemoryEvent(UInt64 address, UInt64 size, Boolean write, Boolean precise) in D:\a\Ryujinx\Ryujinx\Ryujinx.Memory\Tracking\MemoryTracking.cs:line 275
   at ARMeilleure.Translation.Translator.Execute(ExecutionContext context, UInt64 address) in D:\a\Ryujinx\Ryujinx\ARMeilleure\Translation\Translator.cs:line 130
   at Ryujinx.Cpu.Jit.JitCpuContext.Execute(IExecutionContext context, UInt64 address) in D:\a\Ryujinx\Ryujinx\Ryujinx.Cpu\Jit\JitCpuContext.cs:line 32
   at Ryujinx.HLE.HOS.ArmProcessContext`1.Execute(IExecutionContext context, UInt64 codeAddress) in D:\a\Ryujinx\Ryujinx\Ryujinx.HLE\HOS\ArmProcessContext.cs:line 40
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart() in D:\a\Ryujinx\Ryujinx\Ryujinx.HLE\HOS\Kernel\Threading\KThread.cs:line 1265
```
will become
```
00:04:42.739 |E| HLE.GuestThread.20 Application : Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: Attempted to access an invalid memory region.
   at Ryujinx.Memory.Tracking.MemoryTracking.VirtualMemoryEvent(UInt64 address, UInt64 size, Boolean write, Boolean precise) in Ryujinx.Memory\Tracking\MemoryTracking.cs:line 275
   at ARMeilleure.Translation.Translator.Execute(ExecutionContext context, UInt64 address) in ARMeilleure\Translation\Translator.cs:line 130
[...]

PR #4038, opened 2022-12-06
Enable `ContinuousIntegrationBuild` in workflows

mageven:
Supersedes #3905 via `DeterministicSourcePaths` enabled by `ContinuousIntegrationBuild`.

Firstly, a bugfix: I noticed `build.yml` artifacts (PR builds) were missing debug metadata when not passing `-p:DebugType=embedded` to `dotnet build` so I fixed it.

Secondly, I pushed two approaches for enabling `ContinuousIntegrationBuild`-
ea71676f sets the flag in workflow scripts [as suggested in #3905]
b0fc6699 uses `DotNet.ReproducibleBuilds` pkg to automate this (https://github.com/dotnet/reproducible-builds) [thanks to marysaka for showing me this]

The package way carries extra baggage (most notably SourceLink) and I'm not sure about all the flags it sets so I'll let you decide.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,8 +47,8 @@ jobs:
       POWERSHELL_TELEMETRY_OPTOUT: 1
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
       RYUJINX_BASE_VERSION: "1.1.0"
-      RYUJINX_BUILD_COMMON_ARGS: "-p:DebugType=embedded -p:ExtraDefineConstants=DISABLE_UPDATER -p:ContinuousIntegrationBuild=true"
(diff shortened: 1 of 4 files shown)


TSRBerry:
Why is https://github.com/Ryujinx/Ryujinx/commit/ea71676f5d1ab7b8e44db2fb832f7e59ed930ee3 not using this approach outlined in the devlogs [here](https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/#deterministic-builds)?

---

I think I'd prefer the approach taken in https://github.com/Ryujinx/Ryujinx/commit/b0fc6699bbca54bd4fd288c2b61467b4815236d7, since it takes care of the configuration, is simple and doesn't mess with anything we care about (afaik).

marysaka:
Closing as stale.

PR #4039, opened 2022-12-06
macOS: Fix Struct Layout Packing

IsaacMarovitz:
Fix struct layout packing so more test can run on macOS


gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
LGTM, I didn't realise that this would affect its base alignment.

PR #4048, opened 2022-12-06
ava: Restyle the Status Bar

AcK77:
This PR change 3 things:

- The Menu font size is reduced to follow what any other application do (We could do it as a settings later).
- Status Bar margin/size is reduced to get something not too large for nothing and well aligned:

Before:
image
After:
image

- The selector height when you click on a game in the game list view is increased:
Before:
image
After:
image

### DIFF
--- a/Ryujinx.Ava/Assets/Styles/Styles.xaml
+++ b/Ryujinx.Ava/Assets/Styles/Styles.xaml
@@ -161,6 +161,7 @@
     <Style Selector="MenuItem">
         <Setter Property="Height" Value="{DynamicResource MenuItemHeight}" />
         <Setter Property="Padding" Value="{DynamicResource MenuItemPadding}" />
+        <Setter Property="FontSize" Value="12" />
     </Style>
     <Style Selector="MenuItem:selected /template/ Border#root">
         <Setter Property="Background" Value="{DynamicResource ThemeControlBorderColor}" />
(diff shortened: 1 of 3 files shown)


gdkchan:
I think it would look better if the menu bar and status bar had the same height, right now it looks a little weird IMO since the status bar is so much thinner than the menu bar.

AcK77:
> I think it would look better if the menu bar and status bar had the same height, right now it looks a little weird IMO since the status bar is so much thinner than the menu bar.

image
Well in Visual Studio the status bar doesn't have the same height as the menu...

marysaka (approved):

gdkchan:
> > I think it would look better if the menu bar and status bar had the same height, right now it looks a little weird IMO since the status bar is so much thinner than the menu bar.
> 
> image Well in Visual Studio the status bar doesn't have the same height as the menu...

This is not really comparable since the menu on Visual Studio has 2 rows, not just one, and it also doesn't have a separate window title bar.

AcK77:
> This is not really comparable since the menu on Visual Studio has 2 rows, not just one, and it also doesn't have a separate window title bar.

image

In red, the menu bar, with the margin. In blue, the status bar. I only compare on row here. I don't see what's the issue with the title bar in both case.
In any case, I'll not redo the menu, it could be done later.

TSRBerry (approved):
Maybe we can also shrink the MenuBar a tiny bit to make the proportions look better with the new StatusBar :D

PR #4051, opened 2022-12-07
Filter hidden game files from the Game List

Candygoblen123:
`._` files are automatically generated by macOS to store certain metadata about files when you transfer them to an SD card. These files are hidden in finder (even if you enable “show hidden files”) and can only be deleted through terminal. Some users have reported that these files appeared in the game list, so this is a PR that filters them out of the game list.

### DIFF
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -18,6 +18,7 @@ using Ryujinx.Ui.Common.Configuration.System;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.Json;
@@ -104,7 +105,18 @@ namespace Ryujinx.Ui.App.Common
                         continue;
                     }
 
-                    foreach (string app in Directory.EnumerateFiles(appDir, "*", SearchOption.AllDirectories))
+                    IEnumerable<string> content = Enumerable.Empty<string>();
+
+                    try
+                    {
+                        content = Directory.EnumerateFiles(appDir, "*", SearchOption.AllDirectories);
+                    }
+                    catch (UnauthorizedAccessException)
+                    {
+                        Logger.Warning?.Print(LogClass.Application, $"Failed to get access to directory: \"{appDir}\"");
+                    }
+
+                    foreach (string app in content)
                     {
                         if (_cancellationToken.Token.IsCancellationRequested)
                         {
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -160,12 +160,13 @@ namespace Ryujinx.Ui.App.Common
 
                         string extension = Path.GetExtension(app).ToLower();
 
-                        if ((extension == ".nsp")  ||
+                        if (((extension == ".nsp") ||
                             (extension == ".pfs0") ||
                             (extension == ".xci")  ||
                             (extension == ".nca")  ||
                             (extension == ".nro")  ||
-                            (extension == ".nso"))
+                            (extension == ".nso")) &&
+                            (Path.GetFileName(app).Substring(0, 2) != "._"))
                         {
                             applications.Add(app);
                             numApplicationsFound++;
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -77,43 +77,43 @@ namespace Ryujinx.Ui.App.Common
             while (stack.Count > 0)
             {
                 string   dir     = stack.Pop();
-                string[] content = Array.Empty<string>();
+                IEnumerable<string> content = Array.Empty<string>();
+                IEnumerator<string> contentEnum = content.GetEnumerator();
 
                 try
                 {
-                    content = Directory.GetFiles(dir, "*");
+                    content = Directory.EnumerateFiles(dir);
+                    contentEnum = content.GetEnumerator();
                 }
                 catch (UnauthorizedAccessException)
                 {
                     Logger.Warning?.Print(LogClass.Application, $"Failed to get access to directory: \"{dir}\"");
                 }
 
-                if (content.Length > 0)
+                while (contentEnum.MoveNext())
                 {
-                    foreach (string file in content)
+                    string file = contentEnum.Current;
+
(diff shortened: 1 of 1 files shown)


Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -160,12 +160,13 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
 
-                        if ((extension == ".nsp")  ||
+                        if (((extension == ".nsp") ||
                             (extension == ".pfs0") ||
                             (extension == ".xci")  ||
                             (extension == ".nca")  ||
                             (extension == ".nro")  ||
-                            (extension == ".nso"))
+                            (extension == ".nso")) &&
+                            (Path.GetFileName(app).Substring(0, 2) != "._"))
```
TSRBerry, line 169: Move this into `GetFilesInDirectory()` and use `StartWith()` instead of `Substring()`.

I think it might also be a good idea to filter all files starting with a `.`, so other hidden files won't be listed either.

TSRBerry (changes requested):

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
AcK77: Or maybe don't list files when this condition `File.GetAttributes(path).HasFlag(FileAttributes.Hidden)` is `true` ?
(I don't know if files starting with `._` or `.` have it or not)

TSRBerry: Not sure if that works for Linux and/or macOS

gdkchan: I think checking for hidden files is a better solution too (assuming it solves the problem).

TSRBerry: Yup! We can use that, so no filtering of the filename is needed.

Source: https://learn.microsoft.com/en-us/dotnet/api/system.io.fileattributes?view=net-7.0
image

Candygoblen123: Yeah, I just implemented it and it seems to work just fine.

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -160,12 +160,13 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
                         string extension = Path.GetExtension(app).ToLower();
 
-                        if ((extension == ".nsp")  ||
+                        if (((extension == ".nsp") ||
                             (extension == ".pfs0") ||
                             (extension == ".xci")  ||
                             (extension == ".nca")  ||
                             (extension == ".nro")  ||
-                            (extension == ".nso"))
+                            (extension == ".nso")) &&
```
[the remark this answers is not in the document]
Candygoblen123: Im sorry, but I don't have much experience in C# and im not sure what you mean by that.

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -92,7 +92,9 @@ public IEnumerable<string> GetFilesInDirectory(string directory)
                 {
                     foreach (string file in content)
                     {
-                        yield return file;
+                        if (!File.GetAttributes(file).HasFlag(FileAttributes.Hidden)) {
```
gdkchan, line 95:
```suggestion
                        if (!File.GetAttributes(file).HasFlag(FileAttributes.Hidden))
                        {
```

TSRBerry:
I'd like to ask you for one more thing. Could you also change this?

```diff
- content = Directory.GetFiles(dir, "*");
+ Directory.EnumerateFiles(dir);
```

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -77,40 +77,43 @@ public IEnumerable<string> GetFilesInDirectory(string directory)
             while (stack.Count > 0)
             {
                 string   dir     = stack.Pop();
-                string[] content = Array.Empty<string>();
+                IEnumerable<string> content = Array.Empty<string>();
```
TSRBerry, line 80: Sorry I didn't see that before, but this method can be completely removed.

Instead you should use `if (!File.GetAttributes(file).HasFlag(FileAttributes.Hidden))` within the foreach-loop in `LoadApplication()`.

TSRBerry (approved):
Thank you!

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -151,7 +104,7 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
                         continue;
                     }
 
-                    foreach (string app in GetFilesInDirectory(appDir))
+                    foreach (string app in Directory.EnumerateFiles(appDir, "*", SearchOption.AllDirectories))
```
TSRBerry, line 107: Wrap this into a `try` block, like we discussed on discord

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -18,6 +18,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
```
TSRBerry, line 21:
```suggestion
```
See other comment below.

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -151,7 +105,18 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
+                    try
+                    {
+                        content = Directory.EnumerateFiles(appDir, "*", SearchOption.AllDirectories);
+                    }
+                    catch (UnauthorizedAccessException)
+                    {
+                        Logger.Warning?.Print(LogClass.Application, $"Failed to get access to directory: \"{appDir}\"");
+                    }
+
+                    foreach (string app in content)
```
TSRBerry, line 119:
```suggestion
                    foreach (string app in Directory.EnumerateFiles(appDir, "*", SearchOption.AllDirectories))
```
Github doesn't let me create a comment further down, so I can put everything into the suggestion block.

You should move the foreach-loop into the try block. This allows you to get rid of the `content` variable and you won't need to import `System.Linq`.

TSRBerry (changes requested):

TSRBerry (approved):
Lgtm, thanks!

marysaka (approved):

PR #4069, opened 2022-12-08
Fix shader FSWZADD instruction

gdkchan:
Fixes a mistake on the helper function for this shader instruciton that was causing the LUT index to be out of bounds.

Fixes text rendering on The Stanley Parable: Ultra Deluxe, and maybe other games.
Before:
image
After:
image


GamerzHell9137:
Fixes text in Just Dance 2023 Edition 
PR:
image
Master:
image

marysaka (approved):
nice catch 👍

GamerzHell9137:
Fixes layering issues on OlliOlli World which is the last graphical issue the game had on Ryu.

PR:
PR
PR2

Master:
image

image

PR #4082, opened 2022-12-09
misc: Remove dependency on System.Drawing.Common

marysaka:
We only used it in one spot for DPI scaling factor.

This implements the same behaviour using gdiplus.

This remove 700KB of dependency to download and around 170KB unpacked.

### DIFF
--- a/Ryujinx.Common/System/GdiPlusHelper.cs
+++ b/Ryujinx.Common/System/GdiPlusHelper.cs
@@ -1,0 +1,76 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+namespace Ryujinx.Common.System
+{
+    [SupportedOSPlatform("windows")]
+    public static class GdiPlusHelper
+    {
+        private const string LibraryName = "gdiplus.dll";
+
+        private static readonly IntPtr _initToken;
+
+        static GdiPlusHelper()
+        {
+            CheckStatus(GdiplusStartup(out _initToken, StartupInputEx.Default, out _));
+        }
+
+        private static void CheckStatus(int gdiStatus)
+        {
+            if (gdiStatus != 0)
+            {
+                throw new Exception($"GDI Status Error: {gdiStatus}");
+            }
+        }
+
+        private struct StartupInputEx
+        {
+            public int GdiplusVersion;
+
+#pragma warning disable CS0649
+            public IntPtr DebugEventCallback;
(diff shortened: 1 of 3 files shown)


Ryujinx.Common/System/GdiPlusHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            };
+        }
+
+        private struct StartupOutput
+        {
+            public IntPtr NotificationHook;
+            public IntPtr NotificationUnhook;
+        }
+
+        [DllImport(LibraryName)]
```
TSRBerry, line 55:
```suggestion
        [LibraryImport(LibraryName)]
```
We should switch to using `LibraryImport` with .NET 7.

> The P/Invoke source generator, included with the .NET 7 SDK and enabled by default, looks for [LibraryImportAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.libraryimportattribute) on a static and partial method to trigger compile-time source generation of marshalling code, removing the need for the generation of an IL stub at run time and allowing the P/Invoke to be inlined. [Analyzers](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1050-1069) and code fixers are also included to help with migration from the built-in system to the source generator and with usage in general.

Source: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation

marysaka: > We should switch to using `LibraryImport` with .NET 7.
> 
> > The P/Invoke source generator, included with the .NET 7 SDK and enabled by default, looks for [LibraryImportAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.libraryimportattribute) on a static and partial method to trigger compile-time source generation of marshalling code, removing the need for the generation of an IL stub at run time and allowing the P/Invoke to be inlined. [Analyzers](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1050-1069) and code fixers are also included to help with migration from the built-in system to the source generator and with usage in general.
> 
> Source: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation
> 
> Edit: Removed the suggestion because I forgot to make the function `static` and `partial` instead of `extern`

I think we should make an issue and do that across all project in another PR

TSRBerry (approved):

PR #4083, opened 2022-12-09
Use method overloads that support trimming. Mark some types to be trimming friendly

SukharevAndrey:
This PR fixes some easy trimming warnings by using trimming friendly overloads. Some types were attributed too (it looks ugly though, but this is the only way to fix).

Other trimming warnings such as using dynamic code during JSON serialization/deserialization and reflection usages will be fixed in subsequent PRs.


gdkchan (approved):
lgtm, thanks.

marysaka (approved):

AcK77 (approved):

PR #4087, opened 2022-12-09
Vulkan: Don't flush commands when creating most sync

riperiperi:
When the WaitForIdle method is called, we create sync as some internal GPU method may read back written buffer data. Some games randomly intersperse compute dispatch into their render passes, which result in this happening an unbounded number of times depending on how many times they run compute or insert memory barriers.

Creating sync in Vulkan is expensive, as we need to flush the current command buffer so that it can be waited on. We have a limited number of active command buffers due to how we track resource usage, so submitting too many command buffers will force us to wait for them to return to the pool.

This PR allows less "important" sync (things which are less likely to be waited on) to wait on a command buffer's result without submitting it, instead relying on AutoFlush or another more important sync to flush it later on.

Because of the possibility of us waiting for a command buffer that hasn't submitted yet, any thread needs to be able to force the active command buffer to submit. The ability to do this has been added to the backend multithreading via an "Interrupt", though it is not currently supported without multithreading.

OpenGL drivers should already be doing something similar so they don't blow up when creating lots of sync, which is why this hasn't been a problem for these games over there.

Improves Vulkan performance on Xenoblade DE/3, Pokemon Scarlet/Violet, and Zelda BOTW (still another large issue here). Needs some testing to make sure nothing broke.

### DIFF
--- a/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
+++ b/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
@@ -1,5 +1,6 @@
 ﻿using Ryujinx.Common;
 using Ryujinx.Common.Configuration;
+using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.GAL.Multithreading.Commands;
 using Ryujinx.Graphics.GAL.Multithreading.Commands.Buffer;
 using Ryujinx.Graphics.GAL.Multithreading.Commands.Renderer;
@@ -39,7 +40,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
         private CircularSpanPool _spanPool;
 
         private ManualResetEventSlim _invokeRun;
-        private ManualResetEventSlim _interruptRun;
+        private AutoResetEvent _interruptRun;
 
         private bool _lastSampleCounterClear = true;
 
@@ -87,7 +88,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
             _galWorkAvailable = new ManualResetEventSlim(false);
             _invokeRun = new ManualResetEventSlim();
-            _interruptRun = new ManualResetEventSlim();
+            _interruptRun = new AutoResetEvent(false);
             _spanPool = new CircularSpanPool(this, SpanPoolBytes);
             SpanPool = _spanPool;
 
@@ -452,7 +453,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
                 _galWorkAvailable.Set();
 
-                _interruptRun.Wait();
+                _interruptRun.WaitOne();
             }
         }
 
--- a/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
+++ b/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
@@ -1,6 +1,5 @@
 ﻿using Ryujinx.Common;
 using Ryujinx.Common.Configuration;
-using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.GAL.Multithreading.Commands;
 using Ryujinx.Graphics.GAL.Multithreading.Commands.Buffer;
 using Ryujinx.Graphics.GAL.Multithreading.Commands.Renderer;
--- a/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
+++ b/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
@@ -48,7 +48,6 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
         private int _consumerPtr;
         private int _commandCount;
-        private int _interrupted;
 
         private int _producerPtr;
         private int _lastProducedPtr;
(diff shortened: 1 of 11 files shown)


riperiperi:
Xenoblade 3 Before:
image

After:
image

(with 60 fps mod)

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -46,6 +48,7 @@ public class ThreadedRenderer : IRenderer
 
         private int _consumerPtr;
         private int _commandCount;
+        private int _interrupted;
```
gdkchan, line 51: Why is that needed? Can't you just check if `_interruptAction` is not null as indicator there is a interrupt action to be called?

gdkchan (commented):
> Because of the possibility of us waiting for a command buffer that hasn't submitted yet, any thread needs to be able to force the active command buffer to submit. The ability to do this has been added to the backend multithreading via an "Interrupt", though it is not currently supported without multithreading.

Isn't better to just disable this optimization entirely if the backend multithreading is disabled then? As far I can tell it would also allow `SetInterruptAction` to be removed since it doesn't look like it does anything when the multithreading is enabled.

MetrosexualGarbodor:
With this PR, Vulkan now performs better than OpenGL on Xenoblade.
master: 45-50fps on OpenGL, 43-47fps on Vulkan
#4087:  45-50fps on OpenGL, 50-54fps on Vulkan
image
image


Amazing job 👍

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
riperiperi: It was a performance consideration, but I did a benchmark run and Volatile.Read(ref int) != 0 and Volatile.Read(ref object) != null are actually the same. I thought that checking the object might cost more, and this loop is obviously super hot, but since it isn't i'll change it.

riperiperi:
> Isn't better to just disable this optimization entirely if the backend multithreading is disabled then? As far I can tell it would also allow SetInterruptAction to be removed since it doesn't look like it does anything when the multithreading is enabled.

The optimization is disabled if backend multithreading is disabled, as SetInterruptAction would never be called. The issue is that the real backend needs to be able to tell the multithreading backend about the interrupt action.

yuwenhua653:
Smaller boost on low-end devices
And send a lot of error logs in the background
`|E|Gpu : VK Sync Object 184458 failed to signal within 1000ms. Continuing..`
`|E| HLE.GuestThread.19 Gpu : VK Sync Object 184466 failed to signal within 1000ms. Continuing...`

riperiperi:
> And send a lot of error logs in the background
Can you explain what game is doing this?

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -125,21 +124,18 @@ public void RenderLoop()
...
 
-                if (Volatile.Read(ref _interrupted) != 0)
+                if (Volatile.Read(ref _interruptAction) != null)
                 {
-                    while (Volatile.Read(ref _interruptAction) == null) { }
-
                     _interruptAction();
                     _interruptRun.Set();
 
                     Interlocked.Exchange(ref _interruptAction, null);
```
gdkchan, line 132: I think it would be safer to do something like `Action interruptAction = Interlocked.Exchange(ref _interruptAction, null);` and then call `interruptAction` if non-null. Probably not going to be a problem here since there is a single consumer, but it's simple enough that it might be worth doing.

gdkchan: I'm assuming that there is also no need for the thread setting the `_interruptAction` to wait until the previous action completes before setting the new one and continuing. If that's not the case then just ignore this suggestion.

yuwenhua653:
> > And send a lot of error logs in the background
> 
> Can you explain what game is doing this?
Xenoblade 3 Pokemon Scarlet/Violet Both have the same error, by the way I use AMD GPU

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
riperiperi: It currently assumes that only one can run at a time and it needs to invoke. That's not necessarily needed right now, but since AMD doesn't seem to like what I'm doing maybe I need to change it to flush, then register the waitable afterwards (which would need it to wait).

riperiperi:
The issue with AMD GPUs should be fixed now. Might fix a few other inaccuracies when using other drivers (not sure how they all react to queuing a fence signal after waiting for it)

Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -38,6 +39,7 @@ public class ThreadedRenderer : IRenderer
         private CircularSpanPool _spanPool;
 
         private ManualResetEventSlim _invokeRun;
+        private AutoResetEvent _interruptRun;
```
gdkchan, line 42: This is not being disposed.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -83,6 +83,7 @@ public sealed class VulkanRenderer : IRenderer
         public bool PreferThreading => true;
 
         public event EventHandler<ScreenCaptureImageInfo> ScreenCaptured;
+        public Action<Action> InterruptAction;
```
gdkchan, line 86: This should probably not be public.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4088, opened 2022-12-10
audio: Rewrite SoundIo bindings

marysaka:
Introduce with https://github.com/Ryujinx/Ryujinx/pull/406, our SoundIo bindings were always a bit pedantic...

This rewrite SoundIo bindings to be safer and not a pedantic autogenerated mess.

Next step would be to move natives out of this repository and autobuild them for all supported OSes.

PR #4035, opened 2022-12-06
Solution Wide: Fix Warnings

IsaacMarovitz:
+ Fixes several warnings in `Ryujinx.Ava`
+ Removes unnecessary casts and type specifiers solution-wide (majority of edits)
+ Sets enums used with bitwise operators as `[Flags]`

 Warnings changed as suggested by JetBrains Rider. This may cause unintended errors as Rider does not completely understand Avalonia.

Results in an overall reduction in warnings of ~1-2K

**Please test thoroughly**

Should not affect functionality except when occasional oversights/bugs are fixed.

PR #4091, opened 2022-12-10
Fix Redundant Qualifer Warnings

IsaacMarovitz:
Supersedes https://github.com/Ryujinx/Ryujinx/pull/4035

The majority of redundant qualifier warnings are present in SoundIO bindings. I've ignored those as they will be fixed by #4088.

It should not affect functionality.


Ryujinx.Graphics.Shader/CodeGen/Spirv/Declarations.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Ryujinx.Graphics.Shader.CodeGen.Spirv
 {
-    using SpvInstruction = Spv.Generator.Instruction;
+    using SpvInstruction = Instruction;
```
gdkchan, line 13: Can't `SpvInstruction` be replaced by `Instruction` in this class? Then we could remove this using.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

Issue #4030, opened 2022-12-05
[Feature Request] Avalonia - Update/DLC auto-selection / auto-detection


vesper8:
It takes quite a lot of clicks matching up DLCs and Updates with each game especially when those DLCs/Updates are in folders containing many dozen files. I just think it would be really cool if, at the very least, the 'Select File' would auto-focus on the files/folders matching the clicked-on game's name. Usually the updates/dlcs always share the same filename beginning so maybe this would be really easy to implement.

Or.. going the extra mile would be to just intelligently attempt to match updates and dlcs with games based on name or game ids. So one could supply Ryujinx with a folder containing game updates and a folder containing game DLCs as a general setting, and newly added (or existing too) games would attempt to load the latest update and all DLCs automatically.

Is this too ambitious I wonder? Would be quite a QoL addition IMO!

Thanks for considering it!

Issue #3265, opened 2022-04-08
Feature Request : Configure to auto add DLC and Updates to game from folder.


TomTurbine:
#### Overview:

Basically, having the ability to configure Ryujinx so when it detects a game, it will automatically scan predefined areas for potential DLC and Updates to it.

#### Smaller Details: 

Any method to auto scan would be greatly appreciated by many. But some potential ways I can think of would be:

1) Set it up where when it detects a game, it checks any subfolders of the same name for any potential updates/dlc and adds them or..

2) Have Update/DLC Folders setup where when it detects a game, it will automatically generate a folder with that titles ID number and name where people can drag and drop the files to or even right click the game and have them moved that way if need be.

### Why would this feature be useful?

When a game update comes out, that way you no longer had to manually delete an older update and then add the newer one from the menu, just delete the old one and add the new one and it would be handled automatically.

When it comes to people who like to keep backups or update a system with a large library, instead of having to manually add all of them from the menus again, just have the emulator check the stuff when the games are detected or checked.

In both of those situation, just delete any outdated updates and the emulator will automatically replace it with the newer one you put in the folder, any time you added a DLC, the emulator would automatically add it to the game so long as you have it in the correct folder.

Neipas09:
Damn this has been requested for over a year :(

TomTurbine:
Any word on if this feature will ever be added?

minhnhattonthat:
For the time being, I have created a script to scan a specified folder and add the updates/dlcs for ryujinx

https://github.com/minhnhattonthat/ryujinx-tool

KruyKaze:
> For the time being, I have created a script to scan a specified folder and add the updates/dlcs for ryujinx
> 
> https://github.com/minhnhattonthat/ryujinx-tool

Referring to step 2 in your link, is this a windows only solution?
thanks

minhnhattonthat:
> > For the time being, I have created a script to scan a specified folder and add the updates/dlcs for ryujinx
> > https://github.com/minhnhattonthat/ryujinx-tool
> 
> Referring to step 2 in your link, is this a windows only solution? thanks

Yup it's only for windows unfortunately, since I haven't found any multiplatform hactool alternative. Hopefully the Ryujinx team can intergrate this feature in the near future

KruyKaze:
> > > For the time being, I have created a script to scan a specified folder and add the updates/dlcs for ryujinx
[...]

PR #4099, opened 2022-12-11
New approach for a automatic import of dlcs and title updates from a …

GodAkisuke:
[Issue #4030](https://github.com/Ryujinx/Ryujinx/issues/4030)

Added 2x Global Folders in the Ryujinx base directory called GlobalDlcs and GlobalUpdates.
The user can put all game dlcs in the GlobalDlcs folder (for exmple: smash dlcs, hyrule warrior dlcs, xenoblade dlcs...) same with updates for the GlobalUpdates folder.
He can open the browser location of the folders with 2x new added buttons under file. (Open Global Updates Folder, Open Dlcs Folder).
After doing that the user can press the 2x new added buttons (Load Global Updates, Load Global Dlcs) and the updates and dlcs of the folder should be installed to the according games loaded from the default games folder.

Enhancement Ideas:
Let the user choose the Global Update and Global Dlc folder.
Make a new Menu tab like "File" for automatiion loading settings.
Maybe some improvement ideas for the linking between games and updates/dlcs. (currently through name of the path and the game with a filter for sonder symbols because they can't appear in a path)

### DIFF
--- a/Ryujinx.Ava/Ui/ViewModels/MainWindowViewModel.cs
+++ b/Ryujinx.Ava/Ui/ViewModels/MainWindowViewModel.cs
@@ -30,6 +30,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Path = System.IO.Path;
@@ -941,6 +942,58 @@ namespace Ryujinx.Ava.Ui.ViewModels
             new DirectoryInfo(logPath).Create();
 
             OpenHelper.OpenFolder(logPath);
+        }
+
+        public static void OpenGlobalDlcsFolder()
+        {
+            string globalDlcsPath = Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalDlcs");
+
+            new DirectoryInfo(globalDlcsPath).Create();
+
+            OpenHelper.OpenFolder(globalDlcsPath);
+        }
+
+        public static void OpenGlobalUpdatesFolder()
+        {
+            string golbalUpdatesPath = Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalUpdates");
+
+            new DirectoryInfo(golbalUpdatesPath).Create();
+
+            OpenHelper.OpenFolder(golbalUpdatesPath);
+        }
+
+        public async void LoadGlobalDlcs()
+        {
+            AutoDownloadableContentLoader downloadableContentManager = new AutoDownloadableContentLoader(Applications, _owner.VirtualFileSystem);
+
+            //Searches for the files in the Global Dlcs folder and puts their path and titlename (from folder) in a dictionary.
+            Dictionary<string, string> dlcsAndUpdatesPathWithGameName = Directory.GetFiles(Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalDlcs"))
+                .ToDictionary(x => x, y => Path.GetFileName(y).Split(new[] { "[" }, StringSplitOptions.RemoveEmptyEntries)
+                .First()
+                .Trim());
+
+
+            foreach (ApplicationData application in Applications)
+            {
+                await downloadableContentManager.AutoLoadDlcsAsync(application, dlcsAndUpdatesPathWithGameName);
+            }
+        }
+
+        public async void LoadGlobalUpdates()
+        {
+            AutoTitleUpdateLoader titleUpdateManager = new AutoTitleUpdateLoader(Applications, _owner.VirtualFileSystem);
+
+            //Searches for the files in the Global Updates folder and puts their path and titlename (from folder) in a dictionary.
+            Dictionary<string, string> dlcsAndUpdatesPathWithGameName = Directory.GetFiles(Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalUpdates"))
+                .ToDictionary(x => x, y => Path.GetFileName(y).Split(new[] { "[" }, StringSplitOptions.RemoveEmptyEntries)
+                .First()
+                .Trim());
+
+
+            foreach (ApplicationData application in Applications)
+            {
+                await titleUpdateManager.AutoLoadUpdatesAsync(application, dlcsAndUpdatesPathWithGameName);
+            }
         }
 
         public void ToggleFullscreen()
--- a/Ryujinx.Ava/Ui/Windows/MainWindow.axaml
+++ b/Ryujinx.Ava/Ui/Windows/MainWindow.axaml
@@ -103,7 +103,25 @@
                                 Command="{ReflectionBinding OpenLogsFolder}"
                                 Header="{locale:Locale MenuBarFileOpenLogsFolder}"
                                 ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
-                            <Separator />
+							<Separator />
+							<MenuItem
+							Command="{ReflectionBinding OpenGlobalUpdatesFolder}"
+							Header="Open Global Updates Folder"
+							ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<MenuItem
+                                Command="{ReflectionBinding LoadGlobalUpdates}"
+                                Header="Load Gobal Updates Folder"
+                                ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<Separator />
+							<MenuItem
+							Command="{ReflectionBinding OpenGlobalDlcsFolder}"
+							Header="Open Global Dlc Folder"
+							ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<MenuItem
+							Command="{ReflectionBinding LoadGlobalDlcs}"
+							Header="Load Gobal Dlcs"
+							ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<Separator />
                             <MenuItem
                                 Command="{ReflectionBinding CloseWindow}"
                                 Header="{locale:Locale MenuBarFileExit}"
--- a/Ryujinx.Ava/Common/AutoDownloadableContentLoader.cs
+++ b/Ryujinx.Ava/Common/AutoDownloadableContentLoader.cs
@@ -1,0 +1,188 @@
+﻿using Avalonia.Collections;
+using Avalonia.Controls;
+using Avalonia.Threading;
+using LibHac.Common;
+using LibHac.Fs;
+using LibHac.Fs.Fsa;
+using LibHac.FsSystem;
+using LibHac.Tools.Fs;
+using LibHac.Tools.FsSystem;
+using LibHac.Tools.FsSystem.NcaUtils;
+using Ryujinx.Ava.Common.Locale;
+using Ryujinx.Ava.Ui.Controls;
+using Ryujinx.Ava.Ui.Models;
+using Ryujinx.Common.Configuration;
+using Ryujinx.Common.Utilities;
+using Ryujinx.HLE.FileSystem;
+using Ryujinx.Ui.App.Common;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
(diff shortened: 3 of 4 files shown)


EmulationFanatic:
Just a couple of thoughts:
1. Given that the auto-updater deletes all of the contents of the Ryujinx program folder, how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved?
2. Title is misspelled "Tilte" in the new applicable .cs file.

GodAkisuke:
> Just a couple of thoughts:
> 
> 1. Given that the auto-updater deletes all of the contents of the Ryujinx program folder, how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved?
> 2. Title is misspelled "Tilte" in the new applicable .cs file.

To Point 1.
Which files get deleted and what exactly do you mean by "how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved"?

Point 2. - Thank you  :)

MutantAura:
> > Just a couple of thoughts:
> > 
> > 1. Given that the auto-updater deletes all of the contents of the Ryujinx program folder, how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved?
> > 2. Title is misspelled "Tilte" in the new applicable .cs file.
> 
> To Point 1. Which files get deleted and what exactly do you mean by "how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved"?
> 
> Point 2. - Thank you :)

The auto-updater deletes all files and directories inside the executable directory and _then_ adds the new files. Anything at all inside that directory is going to be deletes unless it is manually excluded such as when the portable configuration is used. 
I'd recommend using the configuration directory instead of the base directory because of this.

Ryujinx.Ava/Ui/Windows/MainWindow.axaml
```diff
@@ -103,7 +103,25 @@
...
+                                Header="Load Gobal Updates Folder"
+                                ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<Separator />
+							<MenuItem
+							Command="{ReflectionBinding OpenGlobalDlcsFolder}"
+							Header="Open Global Dlc Folder"
+							ToolTip.Tip="{locale:Locale OpenRyujinxLogsTooltip}" />
+							<MenuItem
+							Command="{ReflectionBinding LoadGlobalDlcs}"
+							Header="Load Gobal Dlcs"
```
MutantAura, line 122: Spelling mistake but we use localizable text for these that pull from the locale json files. Should be something like:
`Header="{locale:Locale MenuBarFileOpenGlobalDLCFolder}"`

Same for all other new menu items.

Ryujinx.Ava/Ui/ViewModels/MainWindowViewModel.cs
```diff
@@ -943,6 +944,58 @@ public static void OpenLogsFolder()
...
+            string globalDlcsPath = Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalDlcs");
+
+            new DirectoryInfo(globalDlcsPath).Create();
+
+            OpenHelper.OpenFolder(globalDlcsPath);
+        }
+
+        public static void OpenGlobalUpdatesFolder()
+        {
+            string golbalUpdatesPath = Path.Combine(ReleaseInformations.GetBaseApplicationDirectory(), "GlobalUpdates");
```
MutantAura, line 958: Spelling mistake.

gdkchan:
IMO the best way to solve this is some "Add from folder" or something button on the Update manager and DLC manager. Then it can find all the updates/DLCs for the given game on the select folder and sub-folders. This can be extended further by adding some option to search all updates/DLCs for all the games on the games list on a given folder. I don't think users will want to have to place their updates and DLCs at a fixed location.

AcK77:
> IMO the best way to solve this is some "Add from folder" or something button on the Update manager and DLC manager. Then it can find all the updates/DLCs for the given game on the select folder and sub-folders. This can be extended further by adding some option to search all updates/DLCs for all the games on the games list on a given folder. I don't think users will want to have to place their updates and DLCs at a fixed location.

I agree, a fixed location isn't a good solution, you should follow what gdkchan suggested!

GhostofReachCashGrab:
a Better way would be adding where you keep your updates and DLC folders in settings then they get added in updates and dlc configs by default instead of a single global folder for all that will be wiped away since most of us have updates and dlc on a separate drive.

GodAkisuke:
> a Better way would be adding where you keep your updates and DLC folders in settings then they get added in updates and dlc configs by default instead of a single global folder for all that will be wiped away since most of us have updates and dlc on a separate drive.

I like both ideas and think will do both. 
The first if the user wants to update it manually (maybe if something failed) . 
The other for the automation process. 
Tanks for the input.

lostromb:
I store my games + update files on a network share so customizable paths will be a necessity for me. I assume there are other users who do the same.

GodAkisuke:
I added global strings to system settings for the paths to a Global Title Update folder and a Global Dlcs Folder. 
I moved the buttons for auto import to the according manager (dlc, titleupdate manager) and they will load all dlcs/updates for all games in Applications (Game List).
I also used local strings and translated them to all languages.

GodAkisuke:
I added global strings to system settings for the paths to a Global Title Update folder and a Global Dlcs Folder.
I moved the buttons for auto import to the according manager (dlc, titleupdate manager) and they will load all dlcs/updates for all games in Applications (Game List).
I also used local strings and translated them to all languages.

TomTurbine:
This looks a lot like what I was thinking since I made a similar suggestion.
https://github.com/Ryujinx/Ryujinx/issues/3265

Honestly,  a good why to handle it, I have mine setup where each game has a folder with an "Update" and "DLC" subfolder and the game importer sees them just fine if I add a directly, I can tell because the counter if way higher than the game count while it is adding them all before it goes back down at the end.

You could just have it setup so you can add them with the regular old game "Game Directory" folders. Just have the scan first go through and add all the actual game files and then assign the others to those games in a second pass since they have a required game ID to match.

I mean the emulator already sees them with that scan, just having them assign afterwards automatically based on game ID might be possible without too much I would think/hope.

GodAkisuke:
> This looks a lot like what I was thinking since I made a similar suggestion. #3265
> 
> Honestly, a good why to handle it, I have mine setup where each game has a folder with an "Update" and "DLC" subfolder and the game importer sees them just fine if I add a directly, I can tell because the counter if way higher than the game count while it is adding them all before it goes back down at the end.
> 
> You could just have it setup so you can add them with the regular old game "Game Directory" folders. Just have the scan first go through and add all the actual game files and then assign the others to those games in a second pass since they have a required game ID to match.
> 
> I mean the emulator already sees them with that scan, just having them assign afterwards automatically based on game ID might be possible without too much I would think/hope.

I think "maybe" a auto scan at the import of a game would be a nice feature but I don't think a second hidden ongoing behaviour that loads some things, without the user knowing or needing in some cases, is what we want.
I would do the auto scan on game import only suggest with a checkbox in settings for this feature.
Is your Issue still required after this feature?

TomTurbine:
> I think "maybe" a auto scan at the import of a game would be a nice feature but I don't think a second hidden ongoing behaviour that loads some things, without the user knowing or needing in some cases, is what we want. I would do the auto scan on game import only suggest with a checkbox in settings for this feature. Is your Issue still required after this feature?

That would be fine, just thinking of if you ever have to wipe your computer or replace a hard drive and restoring your games from backup and having to manually assign the updates and dlc for countless games can get really tedious.

Just having it show up in the game scan and given an option of "Would you like to automatically add the detected updates and dlc for the selected game(s) too?" Would greatly speed it up.

Went tried the separate folder route and having 1 folder for Games, 1 for updates, and multiple subfolders for DLC just gets rough when you are trying to keep things situated while keeping the updates and DLC with the games makes it much easier to organize.

One other suggestion to potentially add to this, when you are at the menu with the detailed view of your list, right below the version number, including the number of DLC added as well could potentially help a few people when dealing with games with lots of DLC.

GodAkisuke:
> That would be fine, just thinking of if you ever have to wipe your computer or replace a hard drive and restoring your games from backup and having to manually assign the updates and dlc for countless games can get really tedious.

> Actually i reset my computer every quarter of a year so I thought, hey this would be great for me let's start doing it.

> Just having it show up in the game scan and given an option of "Would you like to automatically add the detected updates and dlc for the selected game(s) too?" Would greatly speed it up.

> I don't actually think we need this because you can just place all files where they belong set the paths and press 2 buttons.
> That should be speedy enough for all dlcs and title updates. But maybe it's worth another issue if this one's released.

> Went and tried the separate folder route and having 1 folder for Games, 1 for updates, and multiple subfolders for DLC just gets rough when you are trying to keep things situated while keeping the updates and DLC with the games makes it much easier to organize.

>You can now place all dlcs/titleUpdate inc subDirs in each one folder set the path and it handles installation for you.

> One other suggestion to potentially add to this, when you are at the menu with the detailed view of your list, right below the version number, including the number of DLC added as well could potentially help a few people when dealing with games with lots of DLC.

> Good idea! But I think it would be even better to show the titleVersion and titleDlcs in the listed view next to the game.
I think I'll add this in the next few days

image

lostromb:
> Just having it show up in the game scan and given an option of "Would you like to automatically add the detected updates and dlc for the selected game(s) too?" Would greatly speed it up.

Yeah, why not this? I just have .xci and .nsp files all jumbled up in one big folder. If Ryujinx has the ability to scan and cache what each file contains semantically (I assume at least it can differentiate between update and DLC .nsps), it should be smart enough to just automatically choose the latest update and enable all available DLCs for each given title ID without user intervention.

Have a dialog like this pop up whenever a new game directory is added or when new files are detected within the directory.

image

AcK77 (commented):
You don't need to edit all locale files. Only `en_US.json` is needed, other translations are done in crowdin right now (even if current files on the repo aren't up-to-date).
Instead of using `GlobalDlcPath` and `GlobalTitleUpdatePath` I'd just use `GameDirs` to sort content when the folders are listed. Then if it's an application it goes to App, if it's an update, it goes to updates, etc... It do things easier without add some settings fields for "nothing".

Squall-Leonhart:
> 1. Given that the auto-updater deletes all of the contents of the Ryujinx program folder, how does this PR address there now being important files ( the updates and dlc) in this location that should be persistent & preserved?

It doesn't delete all of the contents, else Portable would be wiped along with it.

J-Swift:
Anyone know the status of this one? Its something I've wanted, mainly to have more of an automated bootstrapping process for a new system. It looks like the author went MIA to me, or am I misreading the situation? I'd be willing to take a look at pushing it forward.

GodAkisuke:
> Anyone know the status of this one? Its something I've wanted, mainly to have more of an automated bootstrapping process for a new system. It looks like the author went MIA to me, or am I misreading the situation? I'd be willing to take a look at pushing it forward.

Hey, I had many difficulties to solve in rl so I had to abondone this for quite some time. 
If I can help you finish this one just let me know.

J-Swift:
@GodAkisuke no worries. I think the easiest path forward since things have been restructured so much is for me to manually copy over the changes to the new directory structure. It looks like theres some "other" stuff (e.g. savedata backup/restore). Is that part of the necessary changes, or just extra?

J-Swift:
@AcK77 / @MutantAura (pinging you two just because you were involved as reviewers previously, I'm not really sure who better to tag): I tried to find a breakdown but couldnt in any wiki or whatever. My understanding of the codebase is that Avalonia is now the 'default' UI and GTK is legacy? Is that accurate? If so, is there any soft or hard requirement on adding features to both?

TSRBerry:
There is no longer a requirement to implement GUI features for both GTK and Avalonia. You can completely ignore GTK now.

GodAkisuke:
> @GodAkisuke no worries. I think the easiest path forward since things have been restructured so much is for me to manually copy over the changes to the new directory structure. It looks like theres some "other" stuff (e.g. savedata backup/restore). Is that part of the necessary changes, or just extra?

Tbh i cant remember to have changed some of the other stuff. Do you have some examples?

J-Swift:
For anyone following this, I did end up getting to a testable state here: https://github.com/Ryujinx/Ryujinx/pull/7205

As long as you are comfortable running the code directly to test you can check it out 😅

PR #4106, opened 2022-12-12
Remove Half Conversion

IsaacMarovitz:
As suggested in #3858


marysaka (approved):

TSRBerry (approved):

gdkchan:
Because this might potentially change shader code, the shader cache version should be changed on `Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs` changing `CodeGenVersion` to the number of this PR which is 4106. This will force the shader cache to rebuild on next launch. That when if something goes wrong due to this we can find out sooner rather than later.

gdkchan (approved):
lgtm, thanks.

PR #4108, opened 2022-12-12
GPU: Add fallback when 16-bit formats are not supported

### LINKED ISSUE #3951: [macOS: GPU] R4G4B4A4 texture format not supported on macOS

When attempting to load a R4G4B4A4 on the macOS build, Ryujinx will crash with a ErrorFormatNotSupported error.
I noticed the error in Super Kirby Clash, but there should be similar issues in Vroom in the night sky, Ni no Kuni 1, and Ys 8.
```
00:03:11.769 |E| GUI.RenderThread Application : Unhandled exception caught: Ryujinx.Graphics.Vulkan.VulkanException: Unexpected API error "ErrorFormatNotSupported".
   at Ryujinx.Graphics.Vulkan.CommandBufferPool.Return(CommandBufferScoped cbs, ReadOnlySpan`1 waitSemaphores, ReadOnlySpan`1 waitDstStageMask, ReadOnlySpan`1 signalSemaphores) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.Vulkan/CommandBufferPool.cs:line 299
   at Ryujinx.Graphics.Vulkan.CommandBufferPool.Return(CommandBufferScoped cbs) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.Vulkan/CommandBufferPool.cs:line 248
   at Ryujinx.Graphics.Vulkan.PipelineFull.FlushCommandsImpl() in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.Vulkan/PipelineFull.cs:line 204
   at Ryujinx.Graphics.GAL.Multithreading.Commands.SetRenderTargetsCommand.Run(SetRenderTargetsCommand& command, ThreadedRenderer threaded, IRenderer renderer) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.GAL/Multithreading/Commands/SetRenderTargetsCommand.cs:line 21
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper.<>c.<InitLookup>b__5_74(Span`1 memory, ThreadedRenderer threaded, IRenderer renderer) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs:line 209
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.RunLoop(Action gpuLoop) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs:line 107
   at Ryujinx.Ava.AppHost.RenderLoop() in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Ava/AppHost.cs:line 860
```

**Super Kirby Clash**:

- Press up on the mission board
- Select any mission
- Select your character
- "Change role"

Attempting to recreate any screenshots from this PR will also likely have the same effect: https://github.com/Ryujinx/Ryujinx/pull/3696

### Log file

[R4G4B4A4.log](https://github.com/Ryujinx/Ryujinx/files/10116935/R4G4B4A4.log)

### Environment?

- Ryujinx version: 1.1.0-macos1
- Game version: 1.0.1
- System Specs:
  - OS: macOS Monterey (12.6)
  - CPU: Intel Core i7-10700K
  - GPU: AMD Radeon Pro 5700 XT 16 GB
  - RAM: 32 GiB
- Applied Mods : No

### Additional context?

Ryujinx is not stable on the machine I tested on, but this is likely an issue regardless of Mac hardware.

riperiperi:
MoltenVK doesn't support these formats on Rosetta/x86, but curiously they are supported on Arm64. This PR adds a conversion to RGBA8 when 16 bit formats are not supported by the host GPU. Generally when one of them is supported, all of them are, so I don't expect this to trigger on existing desktop GPUs.

I'm not too much of a fan of the floating point conversion for 5bit to 8bit, but it works to get 0-31 into 0-255.

Should fix #3951 , once it makes it into a macos build.

Similar to other converted fallbacks, these textures currently can't be flushed back to memory.


marysaka (approved):

riperiperi:
There's an issue with textures that don't have their width aligned to 4 bytes that needs to be resolved.

riperiperi:
I fixed the issue. Now it takes into account the number of padding pixels (remainder) are present on each line and skips them. It also affects the existing R4G4 to R4G4B4A4 conversion, where right now any remainder will just make it not do the vectorized path.

Ryujinx.Graphics.GAL/Format.cs
```diff
@@ -448,6 +448,27 @@ public static bool IsRtColorCompatible(this Format format)
...
         }
 
+        /// <summary>
+        /// Checks if the texture format is 16 bit RGBA.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the texture format is 16 bit RGBA format, false otherwise</returns>
+        public static bool Is16BitRGBA(this Format format)
+        {
+            switch (format)
```
gdkchan, line 458: `A1B5G5R5Unorm` seems to be missing.

Ryujinx.Graphics.Texture/PixelConverter.cs
```diff
@@ -7,30 +8,161 @@ namespace Ryujinx.Graphics.Texture
...
             }
+            else
+            {
+                int offset = 0;
+                int outOffset = 0;
 
-            Span<ushort> outputSpan = MemoryMarshal.Cast<byte, ushort>(output);
+                for (int y = 0; y < height; y++)
+                {
+                    for (int i = 0; i < width; i++)
```
gdkchan, line 60: I think calling it `x` makes more sense than calling it `i`.

Ryujinx.Graphics.Texture/PixelConverter.cs
```diff
@@ -7,30 +8,161 @@ namespace Ryujinx.Graphics.Texture
...
+            byte[] output = new byte[data.Length * 2];
+            int offset = 0;
+            int outOffset = 0;
+
+            (int remainder, int outRemainder, int height) = GetLineRemainders(data.Length, width, 2, 4);
+
+            ReadOnlySpan<ushort> inputSpan = MemoryMarshal.Cast<byte, ushort>(data);
+            Span<uint> outputSpan = MemoryMarshal.Cast<byte, uint>(output);
+
+            float factor5Bit = 255.499f/31f;
```
gdkchan, line 84:
```suggestion
            float factor5Bit = 255.499f / 31f;
```

Ryujinx.Graphics.Texture/PixelConverter.cs
```diff
@@ -7,30 +8,161 @@ namespace Ryujinx.Graphics.Texture
...
+            float factor6Bit = 255.499f/63f;
```
gdkchan, line 85:
```suggestion
            float factor6Bit = 255.499f / 63f;
```

Ryujinx.Graphics.Texture/PixelConverter.cs
```diff
@@ -7,30 +8,161 @@ namespace Ryujinx.Graphics.Texture
...
+
+            for (int y = 0; y < height; y++)
+            {
+                for (int i = 0; i < width; i++)
+                {
+                    ushort packed = inputSpan[offset++];
+                    uint r = (uint)((packed & 31) * factor5Bit);
```
gdkchan, line 92: Isn't more efficient to just replicate the bits? Like by doing `((packed << 3) & 0xf8) | ((packed >> 2) & 7)`. I haven't checked if it matches the float version, but that's usually how I see this conversion being done.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -373,6 +373,13 @@ public unsafe Capabilities GetCapabilities()
                 GAL.Format.Bc7Srgb,
                 GAL.Format.Bc7Unorm);
 
+            bool supports16BitRGBAFormat = FormatCapabilities.OptimalFormatsSupport(compressedFormatFeatureFlags,
+                GAL.Format.R5G6B5Unorm,
+                GAL.Format.R5G5B5A1Unorm,
+                GAL.Format.R5G5B5X1Unorm,
+                GAL.Format.B5G6R5Unorm,
+                GAL.Format.B5G5R5A1Unorm,
+                GAL.Format.R4G4B4A4Unorm);
```
gdkchan, line 382: `R4G4B4A4Unorm` maps to `A4B4G4R4UnormPack16Ext`, which is not supported on Apple Silicon. I believe this check might prevent all 16-bit formats from being used there. Maybe more granularity for those capabilities will be required. On a side note, the fact that this format is not supported there might be causing issues on N64 Online, so it might be better to do the conversion to RGBA8 there (although not sure if that would fix the issue).

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -859,13 +859,35 @@ public SpanOrArray<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data,
...
+                    case Format.R4G4B4A4Unorm:
+                        result = PixelConverter.ConvertRGBA4ToRGBA8(result, width);
+                        break;
+                }
             }
             else if (!TextureCompatibility.HostSupportsBcFormat(Format, Target, _context.Capabilities))
             {
                 switch (Format)
                 {
-                    case Format.Bc1RgbaSrgb:
```
gdkchan, line 868: Why was that removed?

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -859,13 +859,35 @@ public SpanOrArray<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data,
             }
             else if (!_context.Capabilities.SupportsR4G4Format && Format == Format.R4G4Unorm)
             {
-                result = PixelConverter.ConvertR4G4ToR4G4B4A4(result);
+                result = PixelConverter.ConvertR4G4ToR4G4B4A4(result, width);
```
gdkchan, line 862: Move that to the `else` path? Seems wasteful to do this conversion if the host doesn't support the format.

Ryujinx.Graphics.GAL/Capabilities.cs
```diff
@@ -18,6 +18,7 @@ namespace Ryujinx.Graphics.GAL
         public readonly bool SupportsBgraFormat;
         public readonly bool SupportsR4G4Format;
         public readonly bool SupportsSnormBufferTextureFormat;
+        public readonly bool Supports16BitRGBAFormat;
```
gdkchan, line 21: I think it should be called `Supports16BitRgbaFormat` for consistency with the other names. Although some of those formats are RGB formats rather than RGBA, so maybe there's a better name for this capability.

Ryujinx.Graphics.GAL/Format.cs
riperiperi: It doesn't have a conversion.

Ryujinx.Graphics.Gpu/Image/Texture.cs
riperiperi: I don't know how this happened. It was cherry-picked from the mac branch, so maybe something went wrong there.

Ryujinx.Graphics.Texture/PixelConverter.cs
riperiperi: That does sound reasonable.

Ryujinx.Graphics.Gpu/Image/Texture.cs
riperiperi: It is converted from R4G4 to R4G4B4A4, then to R8G8B8A8. There's no straight conversion to RGBA8.

Ryujinx.Graphics.Texture/PixelConverter.cs
```diff
@@ -7,30 +8,170 @@ namespace Ryujinx.Graphics.Texture
...
+                    uint outputPacked =  0xff000000;
+                         outputPacked |= (packed << 3) & 0x000000f8;
+                         outputPacked |= (packed << 6) & 0x0000f800;
+                         outputPacked |= (packed << 9) & 0x00f80000;
+
+                    // Replicate 5 bit components.
+                    outputPacked |= (outputPacked >> 5) & 0x00070707;
+
+                    uint a = forceAlpha ? 1 : (packed >> 15);
+                    outputPacked &= a * 0xffffffffu;
```
gdkchan, line 136: Why does this force RGB to 0 when alpha is 0? I think it's possible that something will have alpha set to 0 but the shader might still access the RGB information.

riperiperi:
I've updated the PR with separated R4G4 capability, and added A1B5G5R5 conversion. I've also changed the conversions to remove FP math. Conversions appear to work (when forced) in Mario Kart 8, NSO N64 emulator. I'm not sure what other games use these formats.

The naming is a bit weird. 16-bit on its own would include R8G8, 16-bit packed would include R4G4... so the only common factor between these formats is that they have 5 bit components. That's how I've named the capability.

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -873,6 +873,9 @@ public SpanOrArray<byte> ConvertToHostCompatibleFormat(ReadOnlySpan<byte> data,
                     case Format.Etc2RgbUnorm:
                         result = ETC2Decoder.DecodeRgb(result, width, height, depth, levels, layers);
                         break;
+                    case Format.A1B5G5R5Unorm:
```
gdkchan, line 876: Why is that grouped with ETC2?

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -396,6 +396,17 @@ public unsafe Capabilities GetCapabilities()
                 GAL.Format.Etc2RgbSrgb,
                 GAL.Format.Etc2RgbUnorm);
 
+            bool supports5BitComponentFormat = FormatCapabilities.OptimalFormatsSupport(compressedFormatFeatureFlags,
```
gdkchan, line 399: Maybe we need different flags here. `compressedFormatFeatureFlags` assumes that the format won't need to be rendered to or blitted to (as you can't do that with compressed formats), but those restrictions doesn't apply to those packed 16-bit formats. Ultimately it wouldn't make much difference right now as we don't have a fallback path for render when the render target format is not supported, but I think it might be worth taking this into account for capability check.

Ryujinx.Graphics.GAL/Capabilities.cs
```diff
@@ -18,7 +18,9 @@ namespace Ryujinx.Graphics.GAL
         public readonly bool Supports3DTextureCompression;
         public readonly bool SupportsBgraFormat;
         public readonly bool SupportsR4G4Format;
+        public readonly bool SupportsRGBA4Format;
```
gdkchan, line 21: For consistency, I think the name should be changed to `SupportsR4G4B4A4Format` or the one above should be changed to `SupportsRg4Format`. Same for the conversion method names.

Ryujinx.Graphics.Gpu/Image/Texture.cs
riperiperi: Because this has been rebased like 15 times.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
riperiperi: I think it's possible to render to these textures right now, but it will be stored internally as RGBA8 and can't be flushed.

Ryujinx.Graphics.GAL/Capabilities.cs
riperiperi: I've renamed everything to specify bits for each component. I also renamed Is16BitRGBA to is16BitPacked.

gdkchan (approved):
lgtm, thanks.

PR #4113, opened 2022-12-13
Save manager backup/restore-backup 4006

GodAkisuke:
Issue 4006 - Save savedata as backup and restore it
-----------------------------------------------------------
So i added 2 buttons in the save manager from the usesr profile menu.
One button compress the savedata folder into a zip located to a user specified path.
The other button extracts the backup and tries to replace the current savedata, with the same name and parent folder, with the backup save files.
All games i have had different savedata filenames so i assumed this could be a good approach because i got told theres no way to figure out which savedata folders and games are linked.

PS:
sadly i forgott to change branch so this feature comes when my recent pr is done.
The only files changed with this pr are ->
---------------------------------------------------------
Ryujinx.Ava/Ui/Controls/BackupSavedataCommand.cs
Ryujinx.Ava/Ui/Controls/RestoreSavedataCommand.cs
Ryujinx.Ava/Ui/ViewModels/UserProfileViewModel.cs

sry for the mess

### DIFF
--- a/Ryujinx.Ava/Common/AutoDownloadableContentLoader.cs
+++ b/Ryujinx.Ava/Common/AutoDownloadableContentLoader.cs
@@ -1,0 +1,188 @@
+﻿using Avalonia.Collections;
+using Avalonia.Controls;
+using Avalonia.Threading;
+using LibHac.Common;
+using LibHac.Fs;
+using LibHac.Fs.Fsa;
+using LibHac.FsSystem;
+using LibHac.Tools.Fs;
+using LibHac.Tools.FsSystem;
+using LibHac.Tools.FsSystem.NcaUtils;
+using Ryujinx.Ava.Common.Locale;
+using Ryujinx.Ava.Ui.Controls;
+using Ryujinx.Ava.Ui.Models;
+using Ryujinx.Common.Configuration;
+using Ryujinx.Common.Utilities;
+using Ryujinx.HLE.FileSystem;
+using Ryujinx.Ui.App.Common;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Drawing.Printing;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Path = System.IO.Path;
+
+namespace Ryujinx.Ava.Ui.Controls
+{
+    internal class AutoDownloadableContentLoader
+    {
+        public readonly List<ApplicationData> Applications;
+        public AvaloniaList<DownloadableContentModel> DownloadableContents { get; private set; }
+        private readonly VirtualFileSystem FileSystem;
+
+        public AutoDownloadableContentLoader(Collection<ApplicationData> applications, VirtualFileSystem virtualFileSystem)
+        {
+            Applications = applications.ToList();
+            FileSystem = virtualFileSystem;
+        }
+
+        public async Task AutoLoadDlcsAsync(ApplicationData application, Dictionary<string, string> dlcPathAndGameNames)
+        {
+            DownloadableContents = new AvaloniaList<DownloadableContentModel>();
+
+            char[] bannedSymbols = { '.', ',', ':', ';', '>', '<', '\'', '\"', };
+            string gameTitle = string.Join("", application.TitleName.Split(bannedSymbols)).ToLower().Trim();
+
+
+            //Loops through the Dlcs to the given gameTitle and adds them to the downloadableContent List
+            dlcPathAndGameNames.Where(titleDlc => titleDlc.Value.ToLower() == gameTitle)
+                .ToList()
+                .ForEach(async dlc => await AddDownloadableContent(dlc.Key, application));
+
+
+            List<DownloadableContentContainer> downloadableContentContainers = new List<DownloadableContentContainer>();
+            string jsonPath = LoadJsonFromTitle(application, downloadableContentContainers);
+            Save(jsonPath, downloadableContentContainers);
+        }
+
+
+        private async Task AddDownloadableContent(string path, ApplicationData applicationData)
+        {
+            if (!File.Exists(path) || DownloadableContents.FirstOrDefault(x => x.ContainerPath == path) != null)
+            {
+                return;
+            }
+
+            using FileStream containerFile = File.OpenRead(path);
+
+            PartitionFileSystem partitionFileSystem = new(containerFile.AsStorage());
+            bool containsDownloadableContent = false;
+
+            FileSystem.ImportTickets(partitionFileSystem);
+
+            foreach (DirectoryEntryEx fileEntry in partitionFileSystem.EnumerateEntries("/", "*.nca"))
+            {
+                using var ncaFile = new UniqueRef<IFile>();
+
+                partitionFileSystem.OpenFile(ref ncaFile.Ref(), fileEntry.FullPath.ToU8Span(), OpenMode.Read).ThrowIfFailure();
+
+                Nca nca = TryOpenNca(ncaFile.Get.AsStorage(), path);
+                if (nca == null)
+                {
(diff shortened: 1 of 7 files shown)


CeruleanSky:
I thought the ExtraData files were used to map folders to games, for example my save for Voice of Cards Beasts of Burden with  ID 010022D016030000 and here is a hex view of the ExtraData file:

image

But I'm not familiar with the system or the codepaths, and I suppose there might be exceptions with apps and homebrew.

GodAkisuke:
> I thought the ExtraData files were used to map folders to games, for example my save for Voice of Cards Beasts of Burden with ID 010022D016030000 and here is a hex view of the ExtraData file:
> 
> image
> 
> But I'm not familiar with the system or the codepaths, and I suppose there might be exceptions with apps and homebrew.

Just checked it with super smash and it seems you are right. I'll dig into this.

AcK77:
This is not the right way to go, I suggest you to take a look at https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/Ui/Controls/SaveManager.axaml.cs and the UserProfile code if you want to find a proper way to backup and restore savedata. You should take a look at the LibHac code too.

You removed ExtraData which could be used by some games. You store saves in a Zip at a hardcoded place instead of letting the users choose the destination. You don’t use locale fields for the new added texts.

GodAkisuke:
> This is not the right way to go, I suggest you to take a look at https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/Ui/Controls/SaveManager.axaml.cs and the UserProfile code if you want to find a proper way to backup and restore savedata. You should take a look at the LibHac code too.
> 
> You removed ExtraData which could be used by some games. You store saves in a Zip at a hardcoded place instead of letting the users choose the destination. You don’t use locale fields for the new added texts.

I understand my way wasn't the right way to go but the extra files truly seem to save the TitleId.
So I could just compare the extrafiles titleId and replace it. This seems a good way imo but I m new in Ryujinx dev.

emmauss:
Well, there are a lot of things that should be done differently,
UI
1. Make the export buttons part of the content view, and not the content dialog. In this case, they should be in `SaveManager.xaml`
2. You don't have to create an ICommand interface to bind a command to a button. You can bind delegates and methods to buttons straight from the viewmodel. you need to do the above first.
3. To enhance UX, you can make the saves in the save list checkable, and then allow exporting the selected or all saves.

Export
1. You export the whole save directory, which includes saves for all users, even though the options are only accessible when you open the saves for 1 user
2. Switch saves have a specific way they are created an managed. Use the libhac save client to actually access and export the actual user's save directory.
3. The output of the export should be in a format that you can reverse the export when restoring. That is, if the game already has a save, it should replace the save, if it doesn't, it should create a new save for that user. In both cases, Libhac will handle save directory creation.
@GodAkisuke are you on the discord server?

GodAkisuke:
> Well, there are a lot of things that should be done differently, UI
> 
> 1. Make the export buttons part of the content view, and not the content dialog. In this case, they should be in `SaveManager.axaml`
> 2. You don't have to create an ICommand interface to bind a command to a button. You can bind delegates and methods to buttons straight from the viewmodel. you need to do the above first.
> 3. To enhance UX, you can make the saves in the save list checkable, and then allow exporting the selected or all saves.
> 
> Export
> 
> 1. You export the whole save directory, which includes saves for all users, even though the options are only accessible when you open the saves for 1 user
> 2. Switch saves have a specific way they are created an managed. Use the libhac save client to actually access and export the actual user's save directory.
> 3. The output of the export should be in a format that you can reverse the export when restoring. That is, if the game already has a save, it should replace the save, if it doesn't, it should create a new save for that user. In both cases, Libhac will handle save directory creation.
>    @GodAkisuke are you on the discord server?

yes I am

GodAkisuke:
> > Well, there are a lot of things that should be done differently, UI
> > 
> > 1. Make the export buttons part of the content view, and not the content dialog. In this case, they should be in `SaveManager.axaml`
> > 2. You don't have to create an ICommand interface to bind a command to a button. You can bind delegates and methods to buttons straight from the viewmodel. you need to do the above first.
> > 3. To enhance UX, you can make the saves in the save list checkable, and then allow exporting the selected or all saves.
> > 
> > Export
> > 
> > 1. You export the whole save directory, which includes saves for all users, even though the options are only accessible when you open the saves for 1 user
> > 2. Switch saves have a specific way they are created an managed. Use the libhac save client to actually access and export the actual user's save directory.
> > 3. The output of the export should be in a format that you can reverse the export when restoring. That is, if the game already has a save, it should replace the save, if it doesn't, it should create a new save for that user. In both cases, Libhac will handle save directory creation.
> >    @GodAkisuke are you on the discord server?
> 
> yes I a

> > Well, there are a lot of things that should be done differently, UI
> > 
> > 1. Make the export buttons part of the content view, and not the content dialog. In this case, they should be in `SaveManager.axaml`
> > 2. You don't have to create an ICommand interface to bind a command to a button. You can bind delegates and methods to buttons straight from the viewmodel. you need to do the above first.
> > 3. To enhance UX, you can make the saves in the save list checkable, and then allow exporting the selected or all saves.
> > 
> > Export
> > 
> > 1. You export the whole save directory, which includes saves for all users, even though the options are only accessible when you open the saves for 1 user
> > 2. Switch saves have a specific way they are created an managed. Use the libhac save client to actually access and export the actual user's save directory.
> > 3. The output of the export should be in a format that you can reverse the export when restoring. That is, if the game already has a save, it should replace the save, if it doesn't, it should create a new save for that user. In both cases, Libhac will handle save directory creation.
> >    @GodAkisuke are you on the discord server?
> 
> yes I am

I try to use libhac but I did already and couldn't find a good documentation or the right things to use. 
I got to get the game saves but i wasn't able to replace them with the backup files.
You got any Idea where i could start looking or have a doc for me?

GodAkisuke:
https://github.com/Ryujinx/Ryujinx/commit/94a2c9de75920b0f0faf317afb74d2804ee960f7

Moved buttons to saveManager.
Removed ICommand interface.
Exporter working so far using libhac library.
Importer probably comming tomorrow (since its 2.am today)
Selection comes after export and import working.

CeruleanSky:
This may or may not be out of scope of your intentions as I haven't looked over what you are doing, and perhaps could be more trouble than you want to put working on management of saves; but if not and if you haven't already considered it, furthering your management work to also be able to import Yuza saves could be useful. If it is too much out of scope or trouble and you do decide to do so, it might be better to put in another pull request.

GodAkisuke:
So everythings working now as far as i tested it.

Feature
--------
Added Import/Export buttons to the savemanager.
Also added a new SaveDataFileManager who handles restoring/saving backups through libhac horizon os.
Language support for all ryujinx supported languages.
Saves the backups in folders with their game ids because its some random number in ryujinx save folder.
With this said the backups can also be used in yuzu and vice versa if you zip the saves from there (don't pin me on that).

jojodunet:
Could we please have a link in the menu/tools for Save Manager tool ? (with Install Firmware tool)
CE8E7792-9AB9-4705-8188-8DA6A43F2969

jojodunet:
@GodAkisuke Could you please check the issues (branch conflicts) and rebase to test the PR?

Ryujinx.Ava/Common/SaveDataFileManager.cs
```diff
@@ -0,0 +1,357 @@
...
+                Result commitFilesResult = _horizonClient.Fs.CommitSaveData(mountName);
+                if (commitFilesResult.IsFailure()) return commitFilesResult;
+
+                _horizonClient.Fs.Unmount(mountName);
+            }
+
+            return Result.Success;
+        }
+
+        private bool CreateSaveData(ApplicationData app)
```
patrick-hovsepian, line 226: is this a copy of the method in `ApplicationHelper`? It shouldn't be duplicated

PR #4114, opened 2022-12-13
Implement JIT Arm64 backend

gdkchan:
This adds Arm64 target support to the JIT, allowing it to work on Arm CPUs.

Most Arm64 instructions map 1:1 to host instructions. Some don't right now due to lack of time or limitations that needs to be addressed (for example, TBL, TBX and LDn/STn instructions requires consecutive registers to be allocated, this requires changes to the register allocators). Some doesn't require anything special, it was just not done yet. Some Arm32 instruction map to Arm64 instructions right now too, but a lot of them were not optimized yet since they are more complicated to map to Arm64 instructions.

I ran the tests on Arm64 Linux. Unicorn seems to crash midway with a SIGILL exception for some reasons, but all tests up to the point that it did run passed. Floating point operations are a bit tricky to test right now because some of them are expected to fail. Maybe we can address that in the future by also mapping guest FPCR/FPSR to the host register for more accurate floating point emulation on Arm.

Host mapped memory manager is not working on macOS or Linux right now for different reasons. On Linux it simply fails to reserve the range since it usually only has 39 bits of address space, and then it tries to reserve 39 bits from that... macOS has 48 bits of address space so that's not a problem there, but it does not work due to the 16 KB pages, other changes in the emulator are required to work around that limitation. I have not tested this on Windows Arm, so I don't know how it will behave there.

The JIT also needs a native library on macOS to change the JIT region protection from RW to RX and back. This can't be done from C# because the functions that changes it will change all JIT memory regions in the process, not just ours, and since the .NET code is JIT generated, it wouldn't be able to execute anymore causing a crash. Linux does not require this as it allows RWX.

There are a few changes here that affects x86 too, so it needs to be tested there, on both Windows and Linux. Some of them are minor optimizations that I made to improve the Arm codegen.


Ryujinx.Memory/MemoryManagerUnixHelper.cs
```diff
@@ -151,10 +151,9 @@ private static int MmapFlagsToSystemFlags(MmapFlags flags)
                 }
             }
 
-            if (OperatingSystem.IsMacOSVersionAtLeast(10, 14))
+            if (flags.HasFlag(MmapFlags.MAP_JIT_DARWIN) && OperatingSystem.IsMacOS())
```
marysaka, line 154: Considering other path checks version, might be worth keeping it that way.

```suggestion
            if (flags.HasFlag(MmapFlags.MAP_JIT_DARWIN) && OperatingSystem.IsMacOSVersionAtLeast(10, 14))
```

ARMeilleure/Translation/Cache/JitCache.cs
```diff
@@ -36,10 +36,11 @@ public static void Initialize(IJitMemoryAllocator allocator)
                 if (_initialized) return;
 
                 _jitRegion = new ReservedRegion(allocator, CacheSize);
+                _jitCacheInvalidator = new JitCacheInvalidation(allocator);
 
                 _cacheAllocator = new CacheMemoryAllocator(CacheSize);
 
-                if (OperatingSystem.IsWindows())
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
marysaka, line 43: Why was that changed?

ARMeilleure/Optimizations.cs
```diff
@@ -30,6 +34,8 @@ public static bool ForceLegacySse
             set => HardwareCapabilities.ForceLegacySse = value;
         }
 
+        internal static bool UseAdvSimd => UseAdvSimdIfAvailable && RuntimeInformation.ProcessArchitecture == Architecture.Arm64;
```
marysaka, line 37:
```suggestion
        internal static bool UseAdvSimd => UseAdvSimdIfAvailable && AdvSimd.IsSupported;
```

ARMeilleure/Native/macos_jit_support/support.c
```diff
@@ -0,0 +1,14 @@
...
+
+#include <libkern/OSCacheControl.h>
+
+void armeilleure_jit_memcpy(void *dst, const void *src, size_t n) {
+    pthread_jit_write_protect_np(0);
+    memcpy(dst, src, n);
+    pthread_jit_write_protect_np(1);
+
+    // Ensure that the instruction cache for this range is invalidated.
+    sys_icache_invalidate(dst, n);
```
marysaka, line 13: Might be worth trying moving ``sys_icache_invalidate`` to ``JitCacheInvalidation.Invalidate`` to make it more uniform and update ``JitCache.Map``.

ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            if (operand.Type == OperandType.I32)
+            {
+                value |= value << 32;
+            }
+
+            return TryEncodeBitMask(value, out immN, out immS, out immR);
+        }
+
+        public static bool TryEncodeBitMask(ulong value, out int immN, out int immS, out int immR)
```
merryhime, line 45: There are O(1) methods of calculating this if you're interested.

ARMeilleure/CodeGen/Arm64/CodeGenContext.cs
```diff
@@ -0,0 +1,286 @@
...
+
+            _constantPoolOffset = constantPoolBaseOffset;
+            _constantPoolWritten = true;
+
+            return constantPoolBaseOffset;
+        }
+
+        public (byte[], RelocInfo) GetCode()
+        {
+            long constantPoolBaseOffset = WriteConstantPool();
```
merryhime, line 232: Is there a guarantee that the stream is 8-byte aligned at this point?

ARMeilleure/CodeGen/Arm64/CallingConvention.cs
```diff
@@ -0,0 +1,90 @@
+using System;
+
+namespace ARMeilleure.CodeGen.Arm64
+{
+    static class CallingConvention
+    {
+        private const int RegistersMask = unchecked((int)0xffffffff);
+        private const int ReservedRegsMask = (1 << CodeGenCommon.ReservedRegister) | (1 << 29) | (1 << 30) | (1 << 31);
```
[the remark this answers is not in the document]
marysaka: Windows on ARM64 also use x18 for TLS.

ARMeilleure/Translation/Cache/JitCacheInvalidation.cs
```diff
@@ -0,0 +1,79 @@
+using ARMeilleure.Memory;
+using System;
+using System.Runtime.InteropServices;
+
+namespace ARMeilleure.Translation.Cache
+{
+    class JitCacheInvalidation
+    {
+        private static int[] _invalidationCode = new int[]
```
merryhime, line 9: Perhaps consider checking CTR_EL0.DIC/CTR_EL0.IDC to avoid unnecessary cache invalidations.

ARMeilleure/Instructions/InstEmitSystem.cs
```diff
@@ -150,6 +150,8 @@ private static void EmitGetFpsr(ArmEmitterContext context)
         {
             OpCodeSystem op = (OpCodeSystem)context.CurrOp;
 
+            context.SyncQcFlag();
```
merryhime, line 153: Would also recommend SetFpsr to perhaps `ClearQcFlag`, otherwise pollution with previous Qc state would likely occur. Would also recommend similar Qc handling around each `Call` if not already done.

Ryujinx.Memory/MemoryBlock.cs
```diff
@@ -454,6 +434,21 @@ public static bool SupportsFlags(MemoryAllocationFlags flags)
...
+        {
+            if (OperatingSystem.IsMacOS() && RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
+            {
+                return 1UL << 14;
+            }
+
+            return 1UL << 12;
+        }
+
+        public static bool IsPageAligned(ulong address, ulong size)
```
merryhime, line 447: Unused function.

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -140,6 +142,49 @@ private void ResetBlockState()
...
+                    Operand fpsr = AddIntrinsicInt(Intrinsic.Arm64MrsFpsr);
+
+                    uint qcFlagMask = (uint)FPSR.Qc;
+
+                    Operand qcClearLabel = Label();
+
+                    BranchIfFalse(qcClearLabel, BitwiseAnd(fpsr, Const(qcFlagMask)));
+
+                    fpsr = BitwiseAnd(fpsr, Const(~qcFlagMask));
+                    AddIntrinsicNoRet(Intrinsic.Arm64MsrFpsr, fpsr);
```
merryhime, line 165: performing `fpsr &= ~(1 << 27)` is unnecessary; you can just set `fpsr = 0`

(If an when you implement fpsr saving, you would want to reset the whole thing anyway, and they all behave as sticky bits, so this will not need to be changed.)

ARMeilleure/Translation/ArmEmitterContext.cs
```diff
@@ -140,6 +142,49 @@ private void ResetBlockState()
...
+        public void ClearQcFlag()
+        {
+            if (Optimizations.UseAdvSimd)
+            {
+                Operand fpsr = AddIntrinsicInt(Intrinsic.Arm64MrsFpsr);
+
+                uint qcFlagMask = (uint)FPSR.Qc;
+
+                fpsr = BitwiseAnd(fpsr, Const(~qcFlagMask));
+                AddIntrinsicNoRet(Intrinsic.Arm64MsrFpsr, fpsr);
```
merryhime, line 184: This can just be `MSR FPSR, XZR`, there's no need for masking (same reasoning as above).

ARMeilleure/CodeGen/Arm64/CodeGenContext.cs
gdkchan: I don't think so, but `LDR (literal)` offset is always 4 bytes aligned, even when loading 8 bytes values, so I don't think it should be a problem? Maybe there is a performance advantage in making it 8 byte aligned? But it should be correct in both cases.

ARMeilleure/Native/macos_jit_support/support.c
gdkchan: Worth noting that the implementation on `sys_icache_invalidate` is probably a bit different from `JitCacheInvalidation.Invalidate` though. I think it's better to follow the recommended practices for each platform.

ARMeilleure/CodeGen/Arm64/CodeGenContext.cs
merryhime: It won't be a problem with behaviour (since aarch64 supports misaligned loads!), it's just a very minor performance point.

ARMeilleure/Translation/ArmEmitterContext.cs
gdkchan: Wouldn't that also clear the other bits? That seems undesirable since only Qc is being consumed here (although I don't think the other bits will be accessed right now anyway).

I think in the future, would be better to just map guest FPSR to host FPSR directly instead of doing this. Maybe the same can be done for FPCR, but I haven't checked if there's anything in there that could possibly break things if changed (I guess not, and in any case we should restore the previous FPSR/FPCR state on managed <-> native transitions if we do that).

ARMeilleure/Optimizations.cs
gdkchan: It's worth noting that ties the emulator JIT to whatever the runtime supports, which is an artificial limitation.

gdkchan: It doesn't look like .NET has any equivalent to `CpuId` for Arm64 right now, so I made this change. It will be a problem if we need to use a group of instructions for which .NET has no intrinsic though (like polynomial multiply).

ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
gdkchan: Sure.

merryhime (approved):

marysaka (approved):
lgtm, will need a rebase

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
```diff
@@ -786,7 +883,11 @@ public static void Vmlal_I(ArmEmitterContext context)
 
         public static void Vmls_S(ArmEmitterContext context)
         {
-            if (Optimizations.FastFP && Optimizations.UseSse2)
+            if (Optimizations.FastFP && Optimizations.UseAdvSimd)
+            {
+                InstEmitSimdHelper32Arm64.EmitScalarTernaryOpF32(context, Intrinsic.Arm64FmlsV);
```
riperiperi, line 888: No scalar intrinsic for this?

ARMeilleure/Optimizations.cs
```diff
@@ -8,6 +9,9 @@ public static class Optimizations
 
         public static bool AllowLcqInFunctionTable  { get; set; } = true;
         public static bool UseUnmanagedDispatchLoop { get; set; } = true;
+        public static bool UseDirectTableCalls      { get; set; } = true;
```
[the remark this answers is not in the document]
gdkchan: It was introduced as a way around stack overflows caused by deep call chains, but it's not being used right now (always true) so I will remove it.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
```diff
@@ -2110,7 +2557,11 @@ public static void Pmull_V(ArmEmitterContext context)
         {
             OpCodeSimdReg op = (OpCodeSimdReg)context.CurrOp;
 
-            if (Optimizations.UsePclmulqdq && op.Size == 3)
+            if (Optimizations.UseAdvSimd && false) // Not supported by all Arm CPUs.
```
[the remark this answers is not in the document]
gdkchan: I'm not aware of any way to check in .NET directly. We will probably need to write some assembly or call some system function to get the CPU capabilities.

ARMeilleure/CodeGen/Arm64/CodeGenerator.cs
```diff
@@ -0,0 +1,1581 @@
...
+using ARMeilleure.CodeGen.RegisterAllocators;
+using ARMeilleure.CodeGen.Unwinding;
+using ARMeilleure.Common;
+using ARMeilleure.Diagnostics;
+using ARMeilleure.IntermediateRepresentation;
+using ARMeilleure.Translation;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Numerics;
```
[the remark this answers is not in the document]
gdkchan: Will do but this is not consistent across the project iirc.

ARMeilleure/Instructions/InstEmitSimdArithmetic32.cs
gdkchan: The only scalar variant of this instruction is a "by element" one. I would need to also specify the element index to use it. Can be done but it would make the implementation more complicated. Given that for Arm32 it needs to extract/insert scalar values at the right place, I think using vector operations should still produce correct results.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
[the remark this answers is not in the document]
marysaka: On Linux at least you can check it via [getauxval(3)](https://man7.org/linux/man-pages/man3/getauxval.3.html) with AT_HWCAP2 passed.

https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h#L30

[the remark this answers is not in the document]
merryhime: macOS you can check this via [sysctlbyname](https://developer.apple.com/documentation/kernel/1387446-sysctlbyname/determining_instruction_set_characteristics) with hw.optional.arm.FEAT_PMULL passed.

on Windows you'll need to parse the appropriate registry key.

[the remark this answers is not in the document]
gdkchan: Should II do it on this PR then? I was planning to keep it for the future like the other checks that will require different code per platform (such as the one to get page size).

riperiperi (approved):
LGTM, nice work!

PR #4116, opened 2022-12-13
 Vulkan: enable VK_EXT_custom_border_color features

### LINKED ISSUE #3993: [Regression]Super Smash Bros Ultimate crashes RADV
I don't know if this is a genuine regression or a side effect but since 1.1.299 this game crashes RADV after joining a match
We have a mesa issue open however this is probably not a mesa issue
https://gitlab.freedesktop.org/mesa/mesa/-/issues/7611

It would be nice if we at least got a hotfix to disable border color on RADV until the issue is investigated

Scarlet/Violet also crash on boot however it could be for another reason

### LINKED ISSUE #4072: Metroid Dread crashes RADV driver on some shaders, massively fails Vulkan validation layers

I experience an issue where certain shaders, such as the one which first gets cached when Samus is submerged in lava without the Gravity Suit acquired. I also took the liberty of running the game at my current save with the Vulkan Validation Layers inserted into the process, and it spewed out a huge log of nonstop validation errors, which I've included in a stdout log from the emulator.

1. Install Metroid Dread base game
2. Install patch 2.1.0
3. Install the attached save data
4. Run the emulator with `vulkan-validation-layers` installed and `VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation` in the environment, logging stdout to a log file
5. Load the save game in the game with a sufficient shader cache
6. Move around assuming it loads

### Log file

Plain run: [ryujinx.log.gz](https://github.com/Ryujinx/Ryujinx/files/10185210/ryujinx.log.gz)
Run which leads to a room which crashes RADV, but doesn't crash AMDGPU Pro: 
[ryujinx.2.log.gz](https://github.com/Ryujinx/Ryujinx/files/10185298/ryujinx.2.log.gz)

### Environment?

- Ryujinx version: 1.1.450-1
- Game version: 2.1.0
- System Specs:
  - OS: Arch Linux, linux-tkg-cfs 6.1.rc8-272
  - CPU: Ryzen R7 2700
  - GPU: Radeon RX 480 8GB
  - RAM: 32GB
- Applied Mods : No

### Additional context?

I am running `vulkan-validation-layers` 1.3.231.1-1, and the game only runs past the crashing shaders if I use AMDGPU Pro Vulkan driver, not if I use the open source RADV, regardless of whether I use the default ACO or switch to LLVM.

Attaching save file: 
[metroid_dread_save.tar.gz](https://github.com/Ryujinx/Ryujinx/files/10185306/metroid_dread_save.tar.gz)

Crash from this save occurs by loading, dropping down into the vertical room below the elevator, blasting the wall to the right, navigating downward, and to the left, and entering the previously entered Network Room. Crash on RADV occurs the instant the room transition begins, whereas AMDGPU-Pro loads the room instantly.

DadSchoorse:
radv only create the border color bo if this feature is enabled, so it crashed when creating samplers with custom border colors 
Fixes #4072
Fixes #3993

### DIFF
--- a/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
@@ -374,7 +374,24 @@ namespace Ryujinx.Graphics.Vulkan
             api.GetPhysicalDeviceProperties(physicalDevice, out var properties);
             bool useRobustBufferAccess = VendorUtils.FromId(properties.VendorID) == Vendor.Nvidia;
 
-            var supportedFeatures = api.GetPhysicalDeviceFeature(physicalDevice);
+            PhysicalDeviceFeatures2 features2 = new PhysicalDeviceFeatures2()
+            {
+                SType = StructureType.PhysicalDeviceFeatures2
+            };
+
+            PhysicalDeviceCustomBorderColorFeaturesEXT featuresCustomBorderColorSupported = new PhysicalDeviceCustomBorderColorFeaturesEXT()
+            {
+                SType = StructureType.PhysicalDeviceCustomBorderColorFeaturesExt
+            };
+
+            if (supportedExtensions.Contains("VK_EXT_custom_border_color"))
+            {
+                features2.PNext = &featuresCustomBorderColorSupported;
+            }
+
+            api.GetPhysicalDeviceFeatures2(physicalDevice, &features2);
+
+            var supportedFeatures = features2.Features;
 
             var features = new PhysicalDeviceFeatures()
             {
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
```diff
@@ -490,6 +490,21 @@ public static Device CreateDevice(Vk api, PhysicalDevice physicalDevice, uint qu
...
+            PhysicalDeviceCustomBorderColorFeaturesEXT featuresCustomBorderColor;
+
+            if (supportedExtensions.Contains("VK_EXT_custom_border_color"))
+            {
+                featuresCustomBorderColor = new PhysicalDeviceCustomBorderColorFeaturesEXT()
+                {
+                    SType = StructureType.PhysicalDeviceCustomBorderColorFeaturesExt,
+                    PNext = pExtendedFeatures,
+                    CustomBorderColors = true,
+                    CustomBorderColorWithoutFormat = true,
```
gdkchan, line 503: Shouldn't it check if this is actually supported before enabling it? AFAIK `VK_EXT_custom_border_color` being supported doesn't imply the `CustomBorderColorWithoutFormat` feature is supported. If it's not supported we should probably behave as if custom border color is not supported at all as we don't pass the format right now.

DadSchoorse: Right. I guess this is a weird edge case, afaict only the raspberry pi supports border colors but not without format.

DadSchoorse: If you want more extensive feature checks, it might be faster if you do it yourself. I'm not really familiar with this code base and C# in general.

marysaka: WIll push the check later then I suppose 👍

marysaka: Should be good

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
LGTM, nice catch.

PR #4121, opened 2022-12-14
Implement a software ETC2 texture decoder

gdkchan:
Desktop GPUs don't support the ETC2 format on Vulkan, which was causing graphical issues and crashes in a few games that uses those formats. This change implements a software decoder that will decompress those textures on CPU when the format is not supported on the GPU.

Fixes crashes on Vegas Party, Paradigm Paradox and any other game with the Etc2 format error. Mainly affects NVIDIA and AMD since Intel supports those formats, and Apple Silicon also supports it.

image
image

I also added the 1-bit alpha ETC2 format (ETC2 punchthrough alpha) which was missing from the format table, but I'm not aware of any game that uses it.

It's worth noting that NVN does not expose the ETC2 format on the Switch, and Vulkan probably doesn't either, so games using that are those using OpenGL and GLES APIs.

Now the only texture compression format that is missing is EAC, but I'm not aware of any game that uses that.

### DIFF
--- a/Ryujinx.Graphics.Texture/ETC2Decoder.cs
+++ b/Ryujinx.Graphics.Texture/ETC2Decoder.cs
@@ -8,6 +8,9 @@ namespace Ryujinx.Graphics.Texture
     {
         private const uint AlphaMask = 0xff000000u;
 
+        private const int BlockWidth = 4;
+        private const int BlockHeight = 4;
+
         private static readonly int[][] _etc1Lut =
         {
             new int[] { 2, 8, -2, -8 },
@@ -51,17 +54,17 @@ namespace Ryujinx.Graphics.Texture
 
             int inputOffset = 0;
 
-            byte[] output = new byte[width * height * 4];
+            byte[] output = new byte[CalculateOutputSize(width, height, depth, levels, layers)];
 
             Span<uint> outputUint = MemoryMarshal.Cast<byte, uint>(output);
-            Span<uint> tile = stackalloc uint[16];
+            Span<uint> tile = stackalloc uint[BlockWidth * BlockHeight];
 
             int imageBaseOOffs = 0;
 
             for (int l = 0; l < levels; l++)
             {
-                int wInBlocks = (width + 3) / 4;
-                int hInBlocks = (height + 3) / 4;
+                int wInBlocks = (width + BlockWidth - 1) / BlockWidth;
+                int hInBlocks = (height + BlockHeight - 1) / BlockHeight;
 
                 for (int l2 = 0; l2 < layers; l2++)
                 {
@@ -69,13 +72,13 @@ namespace Ryujinx.Graphics.Texture
                     {
                         for (int y = 0; y < hInBlocks; y++)
                         {
-                            int ty = y * 4;
-                            int bh = Math.Min(4, height - ty);
+                            int ty = y * BlockHeight;
+                            int bh = Math.Min(BlockHeight, height - ty);
 
                             for (int x = 0; x < wInBlocks; x++)
                             {
-                                int tx = x * 4;
-                                int bw = Math.Min(4, width - tx);
+                                int tx = x * BlockWidth;
+                                int bw = Math.Min(BlockWidth, width - tx);
 
                                 ulong colorBlock = dataUlong[inputOffset++];
 
@@ -89,7 +92,7 @@ namespace Ryujinx.Graphics.Texture
                                     {
                                         int oOffs = oOffsBase + px;
 
-                                        outputUint[oOffs] = tile[py * 4 + px] | AlphaMask;
+                                        outputUint[oOffs] = tile[py * BlockHeight + px] | AlphaMask;
                                     }
                                 }
                             }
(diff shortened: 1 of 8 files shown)


Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+
+                                for (int py = 0; py < bh; py++)
+                                {
+                                    int oOffsBase = imageBaseOOffs + ((ty + py) * width) + tx;
+
+                                    for (int px = 0; px < bw; px++)
+                                    {
+                                        int oOffs = oOffsBase + px;
+
+                                        outputUint[oOffs] = tile[py * BlockHeight + px] | AlphaMask;
```
riperiperi, line 95: Should be `py * BlockWidth`? Since the purpose is skipping rows.

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+                                if (alphaMultiplier != 0)
+                                {
+                                    for (int py = 0; py < bh; py++)
+                                    {
+                                        int oOffsBase = imageBaseOOffs + ((ty + py) * width) + tx;
+
+                                        for (int px = 0; px < bw; px++)
+                                        {
+                                            int oOffs = oOffsBase + px;
+                                            int alphaIndex = (int)((alphaIndices >> (((px * 4 + py) ^ 0xf) * 3)) & 7);
```
riperiperi, line 223: This seems a bit weird, in other examples it's `(y * Block____ + x)` as a collection of rows, but here it is the opposite with a collection of columns. I'm not familiar with the ETC format so I don't know if this is the right ordering. Also, is `4` meant to be one of `BlockWidth` or `BlockHeight`?

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+            Span<uint> palette = stackalloc uint[4];
+
+            palette[0] = Pack(r1, g1, b1);
+            palette[1] = Pack(r2, g2, b2, dist);
+            palette[2] = Pack(r2, g2, b2);
+            palette[3] = Pack(r2, g2, b2, -dist);
+
+            blockHigh = BinaryPrimitives.ReverseEndianness(blockHigh);
+
+            for (int y = 0; y < 4; y++)
```
riperiperi, line 440: BlockWidth and BlockHeight constants?

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+            Span<uint> palette = stackalloc uint[4];
+
+            palette[0] = Pack(r1, g1, b1, dist);
+            palette[1] = Pack(r1, g1, b1, -dist);
+            palette[2] = Pack(r2, g2, b2, dist);
+            palette[3] = Pack(r2, g2, b2, -dist);
+
+            blockHigh = BinaryPrimitives.ReverseEndianness(blockHigh);
+
+            for (int y = 0; y < 4; y++)
```
riperiperi, line 498: Could use the constants here too.

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+
+            rh = (rh << 2) | (rh >> 4);
+            gh = (gh << 1) | (gh >> 6);
+            bh = (bh << 2) | (bh >> 4);
+
+            rv = (rv << 2) | (rv >> 4);
+            gv = (gv << 1) | (gv >> 6);
+            bv = (bv << 2) | (bv >> 4);
+
+            for (int y = 0; y < 4; y++)
```
riperiperi, line 550: Constants here too?

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+            int[] table1 = _etc1Lut[(blockLow >> 29) & 7];
+            int[] table2 = _etc1Lut[(blockLow >> 26) & 7];
+
+            bool flip = (blockLow & 0x1000000) != 0;
+
+            if (!flip)
+            {
+                for (int y = 0; y < 4; y++)
+                {
+                    for (int x = 0; x < 2; x++)
```
riperiperi, line 586: Not sure about constants here since this is a half block.

Ryujinx.Graphics.Texture/ETC2Decoder.cs
```diff
@@ -0,0 +1,681 @@
...
+                        tile[offset1] = color1;
+                        tile[offset2] = color2;
+                    }
+                }
+            }
+        }
+
+        private static uint CalculatePixel(uint r, uint g, uint b, int x, int y, uint block, int[] table, uint alphaMask)
+        {
+            int index = (x * 4) + y;
```
riperiperi, line 620: BlockHeight?

Ryujinx.Graphics.Texture/ETC2Decoder.cs
gdkchan: On ETC2, the indices for pixels are encoded per column rather than per row, that's why it's different when calculating bit positions to extract the data from the block. It's also big endian and starts from the most significant bit which makes it very awkward to decode.

riperiperi: It looks like this is done later in CalculatePixel, since the textures look correct in the screenshot I'm guessing that's how the format orders the pixels.

gdkchan: I will use `BlockHeight` since it's skipping columns in this case.

gdkchan:
Replaced the remaining `4`s with the `BlockWidth`/`BlockHeight` constants and started using the `DivRoundUp` helper method instead of doing the subtraction/division inline.

AcK77 (approved):

PR #4123, opened 2022-12-15
Implement another non-indexed draw method on GPU

gdkchan:
This implements one missing method on the GPU that is used to do non-indexed draws in some cases. They are similar to the indexed ones that the Vulkan driver uses, with the main difference being that they are for non-indexed draws instead.

I also changed the name of a few other draw methods to match the official documentation, and fixed the primitive type override behaviour (before it would always override for those draws, instead of only doing so if the override is actually enabled). There is another thing that is probably wrong, there is one method per index type, so I guess the method itself defines the index type, rather than it using the actual index type used on regular draws. I did not change it though since it still needs to be tested on hardware to confirm, I just left a TODO comment.

Fixes rendering on Ikaruga.
Before:
image
After:
image


marysaka (approved):

mdsmendes94:
thank you gdkchan

PR #4128, opened 2022-12-15
Replace StructArrayHelpers with custom source generation task

TSRBerry:
This was really hard to get right, since documentation for custom MSBuild tasks really isn't the best.

The approach via a custom MSBuild task avoids the following issues we would have using source generation:

- We are able to scan the whole solution during the compilation of one project.
- We can use `LibraryImport` without causing compiler errors.
  - Since `LibraryImport` is also a source generator it wouldn't know about the generated sources and wouldn't be able to generate the P/Invoke methods.

We currently only have this (`GenerateArrays`) custom task, but using the [Ryujinx.CustomTasks](https://github.com/Ryujinx/Ryujinx.CustomTasks) Nuget package, we could add even more of them in the future.

---

Relevant documentation:

- [MSBuild: Create a custom task for code generation](https://learn.microsoft.com/en-us/visualstudio/msbuild/tutorial-custom-task-code-generation)
- [MSBuild: Create items during execution](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-items#create-items-during-execution)
- [C# Source Generators](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview)
- [Code Analysis: IdentifierNameSyntax](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.identifiernamesyntax)
- [Code Analysis: GenericNameSyntax](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.genericnamesyntax)


gdkchan:
Why is it on a separate nuget package? Personally I'd prefer if it was on this repo since it's a pretty basic component required to allow the project to build.

jduncanator:
> Why is it on a separate nuget package? Personally I'd prefer if it was on this repo since it's a pretty basic component required to allow the project to build.

I agree with this.

I'm also not a fan of adding a dependency on MSBuild. Currently Ryu can be built fine without requiring MSBuild.

This is a good change, but I think it's only worthwhile if we can achieve it without requiring a MSBuild task.

TSRBerry:
> This is a good change, but I think it's only worthwhile if we can achieve it without requiring a MSBuild task.

That's not possible, since other source generation can't use the output of other source generators, so it would lead to a lot of issues with `LibraryImport`. Also a source generator can only see the current project, not the whole solution.

> I'm also not a fan of adding a dependency on MSBuild. Currently Ryu can be built fine without requiring MSBuild.

I'm not sure why this is a bad thing. MSBuild is shipped with dotnet anyways, right? Why not use it?

> Why is it on a separate nuget package? Personally I'd prefer if it was on this repo since it's a pretty basic component required to allow the project to build.

Microsoft recommends to put custom tasks in it's own package and this allows us to add more custom tasks if we need/want to.

> It's important to avoid generating a MSBuild custom task in the same MSBuild process which is going to consume it. The new project should be in a complete different Visual Studio solution, or the new project use a dll pre-generated and re-located from the standard output.

Source: [Custom Task: Code Generation](https://learn.microsoft.com/en-us/visualstudio/msbuild/tutorial-custom-task-code-generation?view=vs-2022#generate-a-console-app-and-use-the-custom-task)

TSRBerry:
Wait we have a dependency on MSBuild anyway, since .NET projects are always using MSBuild.

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build#msbuild

gdkchan:
I suppose the `StructByteArrayHelpers` removal was left for another PR?

TSRBerry:
Oh no, I just forgot about that. Is there any reason we need them instead of using `Array512<byte>` for example?

gdkchan:
See this comment https://github.com/Ryujinx/Ryujinx/pull/2789#issue-1037314976

Apparently it's a runtime limitation caused by the excessive nesting of structs with the approach currently used on `master`.

TSRBerry:
Let me see if we can replace them now. If we can't I'll add another source generator for that.

TSRBerry:
Okay seems like replacing them works just fine. I ran all unit tests and played a grand prix in Mario Kart 8 without any issues.

So I guess it works?

Ryujinx.HLE/HOS/Applets/Browser/WebCommonReturnValue.cs
```diff
@@ -6,7 +6,7 @@ public struct WebCommonReturnValue
     {
         public WebExitReason ExitReason;
         public uint          Padding;
-        public ByteArray4096 LastUrl;
+        public Array4096<byte> LastUrl;
```
gdkchan, line 9: This is misaligned now.

Ryujinx.HLE/HOS/Applets/Error/ApplicationErrorArg.cs
```diff
@@ -8,7 +8,7 @@ struct ApplicationErrorArg
     {
         public uint          ErrorNumber;
         public ulong         LanguageCode;
-        public ByteArray2048 MessageText;
-        public ByteArray2048 DetailsText;
+        public Array2048<byte> MessageText;
```
gdkchan, line 11: This is misaligned now.

Ryujinx.HLE/HOS/Applets/Browser/WebCommonReturnValue.cs
TSRBerry: Ah sorry about that, totally forgot search and replace could cause that :D

gdkchan: Yea, that's why I'd prefer if those files were auto-formatted with alignment removed. It makes things easier when using those tools.

TSRBerry:
I'll mark this as draft until [semantic analysis](https://github.com/Ryujinx/Ryujinx.CustomTasks/pull/2) gets merged into [Ryujinx.CustomTasks](https://github.com/Ryujinx/Ryujinx.CustomTasks).

PR #4129, opened 2022-12-15
GPU: Force rebind when pool changes

gdkchan:
There was an issue after the optimization that tries to avoid updating textures and samplers when the state didn't change. The issue was that the checks weren't enough if the texture or sampler pool changed, but the handles remained the same. I fixed that by forcing it to update if the current texture pool or sampler pool changed, but that only fixed the issue partially. The scenario below is still possible for example:
- Draw is done with sampler pool A, uses samplers 1 and 2 from pool A.
- Current sampler pool is changed to sampler pool B.
- Draw is done with sampler pool B, uses sampler 1 from pool B (correct).
- Draw is done with sampler pool B, uses sampler 1 from pool B (correct), and sampler 2 from pool A (wrong).

To fix this issue, I changed it to just clear all cached state when those pools change, also forcing bindings to be updated on future draws with the new pool.

This fixes graphical issues on "The New Prince of Tennis: LET'S GO!! ~Daily Life~ from RisingBeat".
Before:
image
After:
image
(Pay attention to the right side character arms, bottom of the screen, behind the dialog box).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureBindingsManager.cs
@@ -320,10 +320,15 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             // Check if the texture pool has been modified since bindings were last committed.
             // If it wasn't, then it's possible to avoid looking up textures again when the handle remains the same.
-            bool poolModified = _cachedTexturePool != texturePool || _cachedSamplerPool != samplerPool;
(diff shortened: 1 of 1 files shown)


marysaka (approved):

lostromb:
I was hoping this might fix a similar-looking texture cache (?) bug with the font rendering in 7 Days of Rose, but apparently not.

image

AcK77 (approved):

Issue #2393, opened 2021-06-23
BSD service: Recv buffer size 0


StephenCWills:
### What's the issue you encountered?

In calls to `recv()` from a homebrew app, the BSD service in Ryujinx is attempting to write to a zero-length buffer.

### Logging in the homebrew app:

Here is the relevant code from the homebrew app (C++).
```C++
fprintf(stderr, "recv1(%d)\n", static_cast<int>(size));
signed_size_type result = ::recv(s, static_cast<char*>(data), size, flags);
```

The `fprintf()` call above produced the following message in Ryujinx.
```
00:00:46.897 |W| HLE.GuestThread.15 KernelSvc OutputDebugString: recv1(65535)
```

### Screenshots from debugging Ryujinx:

This is a screenshot from [IClient.cs](https://github.com/Ryujinx/Ryujinx/blob/50ba233ac628ffb9661ea2bda2cda216d18e81c1/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs#L470) that shows the size of the buffer after it's been retrieved via `GetBufferType0x22()`.
> ![image](https://user-images.githubusercontent.com/9203145/123018340-5c30e800-d39c-11eb-9919-8969ef499b7b.png)

Here is a screenshot from when I stepped into `GetBufferType0x22()` showing which branch of the logic it took.
> ![image](https://user-images.githubusercontent.com/9203145/123017844-6bfbfc80-d39b-11eb-8207-f33c9391b479.png)

I also checked the size of `ReceiveBuff`, and it seems to match the size of the buffer passed into `::recv()`.
> ![image](https://user-images.githubusercontent.com/9203145/123017848-6ef6ed00-d39b-11eb-85ad-e2d520291d1c.png)

### How can the issue be reproduced?

The steps I've used to reproduce this are fairly involved since it requires setup of a multiplayer game in [DevilutionX](https://github.com/diasurgical/devilutionX). To reproduce, you'll need to use a test build of the game linked to this PR: https://github.com/diasurgical/devilutionX/pull/2199. I had to host a game on my PC using the Windows x64 build and then attempt to join it using the Switch build via Ryujinx.

### Additional context?

[...]

Issue #3458, opened 2022-07-11
bsd::RecvFrom: Ryujinx does not verify output buffer size before writing socket address


jduncanator:
When writing the endpoint address of the packet received from the socket, we do not verify the output buffer is allocated, and has enough space to hold the endpoint address. We should verify  `sockAddrOutSize` is non-zero at a minimum. Additionally, IPEndpoint can represent either an IPv4 address or an IPv6 address. If the socket receives a packet from an IPv6 address, but the client has only  allocated  a buffer with size enough for an IPv4 address, we will attempt to do an OOB write. We should verify the output buffer has enough space to represent the IPEndpoint.

https://github.com/Ryujinx/Ryujinx/blob/3af42d6c7e9e71c504b87a7b0f7f960fe83418fb/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs#L408

marysaka:
Fixed by https://github.com/Ryujinx/Ryujinx/pull/4135

PR #4135, opened 2022-12-16
bsd::RecvFrom: verify output buffer size before writing socket address

gnisman:
bsd::RecvFrom: verify output buffer size (sockAddrOutSize) is non zero before writing socket address
fix for #3458

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -530,7 +530,9 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
                     receiveRegion.Dispose();
 
-                    if (sockAddrOutSize != 0)
+                    var bsdSockAddr = BsdSockAddr.FromIPEndPoint(endPoint);
+
+                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) bsdSockAddr.Size())
                     {
                         context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
                     }
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -536,6 +536,10 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
                     {
                         context.Memory.Write(sockAddrOutPosition, bsdSockAddr);
                     }
+                    else
+                    {
+                        errno = LinuxError.ENOMEM;
+                    }
                 }
             }
 
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -532,7 +532,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
                     var bsdSockAddr = BsdSockAddr.FromIPEndPoint(endPoint);
 
-                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) bsdSockAddr.Size())
+                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) Unsafe.SizeOf<BsdSockAddr>())
                     {
                         context.Memory.Write(sockAddrOutPosition, bsdSockAddr);
                     }
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
@@ -530,7 +530,10 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd
 
                     receiveRegion.Dispose();
 
-                    context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
+                    if (sockAddrOutSize != 0)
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -530,7 +530,10 @@ public ResultCode RecvFrom(ServiceCtx context)
 
                     receiveRegion.Dispose();
 
-                    context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
+                    if (sockAddrOutSize != 0)
+                    {
+                        context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
```
TSRBerry, line 535: Missing a check if `BsdSockAddr.FromIPEndPoint(endPoint)` fits into the output buffer.

gnisman: Done

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -530,7 +530,12 @@ public ResultCode RecvFrom(ServiceCtx context)
 
                     receiveRegion.Dispose();
 
-                    context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
+                    var bsdSockAddr = BsdSockAddr.FromIPEndPoint(endPoint);
+
+                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) bsdSockAddr.Size())
```
marysaka, line 535: From my understanding here you are only implementing BsdSockAddr.Size to get the size of a single element size, you can just use [Unsafe.SizeOf API](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.sizeof?view=net-7.0).
```suggestion
                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) Unsafe.SizeOf<BsdSockAddr>())
```

Ryujinx.HLE/HOS/Services/Sockets/Bsd/Types/BsdSockAddr.cs
```diff
@@ -35,5 +35,10 @@ public static BsdSockAddr FromIPEndPoint(IPEndPoint endpoint)
 
             return result;
         }
+        
+        public int Size()
+        {
+            return MemoryMarshal.Cast<BsdSockAddr, byte>(MemoryMarshal.CreateSpan(ref this, 1)).Length;
+        }
```
marysaka, line 42: See previous suggestion
```suggestion
```

marysaka (changes requested):

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -530,7 +530,12 @@ public ResultCode RecvFrom(ServiceCtx context)
 
                     receiveRegion.Dispose();
 
-                    context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
+                    var bsdSockAddr = BsdSockAddr.FromIPEndPoint(endPoint);
+
+                    if (sockAddrOutSize != 0 && sockAddrOutSize >= (ulong) Unsafe.SizeOf<BsdSockAddr>())
```
TSRBerry, line 535: You should probably return an error if we can't write the address to memory.

gnisman: Do you suggest to change the errno which is LinuxError enum? if so what should be the error?

TSRBerry: `LinuxError.ENOMEM` sounds fitting, but I should add that I currently don't know if that's the way HOS deals with it.

TSRBerry (approved):
Lgtm, but another review should probably check if HOS deals with this at all and if so is this implementation correct?

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -530,7 +530,16 @@ public ResultCode RecvFrom(ServiceCtx context)
...
+                    {
+                        context.Memory.Write(sockAddrOutPosition, bsdSockAddr);
```
AcK77, line 537:
```suggestion
                        context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
```

`bsdSockAddr` is a bit useless here since you don't use it elsewhere.

gnisman: Agree, its a left over from before when i had bsdSockAddr.Size()

marysaka (approved):

Ryujinx.HLE/HOS/Services/Sockets/Bsd/IClient.cs
```diff
@@ -530,7 +530,16 @@ public ResultCode RecvFrom(ServiceCtx context)
 
                     receiveRegion.Dispose();
 
-                    context.Memory.Write(sockAddrOutPosition, BsdSockAddr.FromIPEndPoint(endPoint));
+                    var bsdSockAddr = BsdSockAddr.FromIPEndPoint(endPoint);
+
```
marysaka, line 534:
```suggestion
```

AcK77 (approved):

gdkchan:
Could it fix #2393 too? It looks like the same problem but on a different function...

PR #4155, opened 2022-12-20
make UI display correct content in Chinese

### LINKED ISSUE #2510: The language of some game icons is inconsistent with the language of the title
The language of some game icons is inconsistent with the language of the title
image
When selecting Taiwanese Chinese, the icon is Korean
image
When selecting traditional Chinese, the icon is Chinese, but the title is Korean

Open the emulator to switch language display
- Ryujinx version: 1.0.<6976/br>
- Game version: 1.0.0</br>
- System Specs:
  - OS: *Windows 11*
  - CPU: *R7 5700U*
  - GPU: *AMD Radeon(TM) Graphics *
  - RAM: *16GB*
- Applied Mods : [  No ]

chanchancl:
Due to enum.Parse can't transform SystemLanguage to TitleLanguage correctly, it makes 
```
private void GetGameInformation(ref ApplicationControlProperty controlData, out string titleName, out string titleId, out string publisher, out string version)
{
     _ = Enum.TryParse(_desiredTitleLanguage.ToString(), out TitleLanguage desiredTitleLanguage);
    ...
}
```
below transform always get AmericanEnglish in desiredTitleLanguage which is not right, and UI will always show English Title in game table even if you chose SimplifiedChinese as Title Language .

Here are some compartions.

Before (SimplifiedChinese, TraditionalChinese)
image
image

After (SimplifiedChinese, TraditionalChinese)
image
image

Fixes #2510.

### DIFF
--- a/Ryujinx.HLE/HOS/SystemState/SystemStateMgr.cs
+++ b/Ryujinx.HLE/HOS/SystemState/SystemStateMgr.cs
@@ -57,9 +57,9 @@ namespace Ryujinx.HLE.HOS.SystemState
             DesiredTitleLanguage = language switch
             {
                 SystemLanguage.Taiwanese or
-                SystemLanguage.TraditionalChinese => TitleLanguage.Taiwanese,
+                SystemLanguage.TraditionalChinese => TitleLanguage.TraditionalChinese,
                 SystemLanguage.Chinese or
-                SystemLanguage.SimplifiedChinese  => TitleLanguage.Chinese,
+                SystemLanguage.SimplifiedChinese  => TitleLanguage.SimplifiedChinese,
                 _                                 => Enum.Parse<TitleLanguage>(Enum.GetName<SystemLanguage>(language)),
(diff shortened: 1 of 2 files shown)


AcK77 (approved):
I don't think `TitleLanguage` is a thing in HOS, I always wonder why we used that. Thanks for the fix.

TSRBerry (approved):

gdkchan:
> I don't think `TitleLanguage` is a thing in HOS, I always wonder why we used that. Thanks for the fix.

`TitleLanguage` matches the ordering of the entries on the control.nacp file https://switchbrew.org/wiki/NACP#ApplicationTitle (which is different from the actual `SystemLanguage` enum). Maybe that's why it was added.

gdkchan (approved):
I think the best solution here is having a switch to transform from one enum to another. Depending on names matching exactly for the parse to work is very fragile. But we can take that solution for now.

PR #4156, opened 2022-12-20
ARMeilleure: Hash _data pointer instead of value for Operand

riperiperi:
I noticed a weirdly high cost for dictionary accesses from MarkLabel etc. Turns out that the hash code was always the same for labels, so the whole point of having a dictionary was missed and it was putting everything in the same bucket. I made it always hash the _data pointer as that's a good source of identifiable and "random" data.

image

Will likely affect lowCq (first boot time) more than highCq (pptc, generally).

### DIFF
--- a/ARMeilleure/IntermediateRepresentation/Operand.cs
+++ b/ARMeilleure/IntermediateRepresentation/Operand.cs
@@ -378,14 +378,7 @@ namespace ARMeilleure.IntermediateRepresentation
 
         public override int GetHashCode()
         {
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
Nice, I like those simple changes.

TSRBerry:
Since this a change to ARMeilleure, does this not require a change to the internal version?

riperiperi:
This doesn't meaningfully change any generated code, it will just speed up the compilation.

TSRBerry (approved):

PR #4168, opened 2022-12-23
Frequently reset "Enable logging" setting and remove always required logging options

TSRBerry:
This PR removes the GUI options to disable the default logging settings.
Instead they are now always enabled and the option to disable logging will only persist for the current run.

This was done since we are getting a lot of new users asking for support who either upload almost empty logs, or aren't able to find them, because they were disabled.

Since these options don't have an impact on performance, we can safely remove them and encourage users to not disable logging (especially if they want to get support).


gdkchan:
I don't think this should be removed, having a way to filter what gets logged is useful.

TSRBerry:
For users or for developers?

gdkchan:
Mostly for developers. For users it's not that useful, but I still think it's better to have the option. It's possible that some game is generating a lot of logs for some reason, and this gives the user the option to work around the issue at least. In general I don't think someone misusing some option in the emulator justifies removing it entirely, otherwise we might as well remove most of the options in the emulator. People disables V-Sync and complain that the game is too fast, they disable shader cache and complain that the game stutters, they set memory manager mode to software and complain that the game is too slow, they enable the expand DRAM option and complain that Pokémon crashes...

TSRBerry:
Hmm, I think I agree with that. Do you have something else in mind that we could do to reduce the number of support requests the discord gets with no or only minimal logs?

EmulationFanatic:
> Hmm, I think I agree with that. Do you have something else in mind that we could do to reduce the number of support requests the discord gets with no or only minimal logs?

Telemetry :/

gdkchan:
What we have been doing for something that should usually not be enabled/disabled/modified by the average user is showing a warning message when they are changed saying that the option is intended for developers only or something along those lines. In this case it could be something like "Keeping logs enabled is important for troubleshooting issues. It's recommended that you keep X log settings enabled. Disabling those options does not increase performance", with "X" being replaced with the log options that we consider important for troubleshooting.

gdkchan:
Another option is some Advanced settings tab that is hidden by default. It would be made visible by checking some box on regular settings that would show a warning message. We can "hide" all the settings we don't want the average user messing with there.

gdkchan:
> Another option is some Advanced settings tab that is hidden by default. It would be made visible by checking some box on regular settings that would show a warning message. We can "hide" all the settings we don't want the average user messing with there.

I think I prefer the first option though.

TSRBerry:
Both options seem good to me, but I'm not sure which one would be more effective.

I can see users dismissing the developer settings warning without giving it any thought. But I suppose the same argument could be made for the logging warning.
Also the current settings already tell you it doesn't impact performance and people still disable them for some reason. :/


> I think I prefer the first option though (just showing the warning).

This should at least some people to leave them on. Maybe we should also consider some changes to Ryuko and change the process of getting help on Discord a little...? But I can't think of a good way to handle this better..

gdkchan:
AFAIK the information that is most useful for troubleshooting is the current settings, and the exception message on cases where the game crashes. Maybe we can make sure those are always logged regardless of the settings. Is there anything else?

TSRBerry:
Some prepo logs contain useful information like error codes which could directly explain certain errors.
For LDN problems I mostly look for P2P warnings and "GetIpv4Address" (should be log level INFO) getting spammed in the log. But other than that, I'm not sure.

PR #4174, opened 2022-12-24
Some minor cleanups and optimizations

turbedi:
1st commit: Replace a Array.Clear(x, 0, x.Length) call with Array.Clear(x) (more performant)

2nd commit: Use DateTime.UnixEpoch field instead of manually initializing a static field in some places.

3rd commit: Replace SHA256.ComputeHash calls with new static SHA256.HashData method (More performant and avoids the need to initialize a SHA256 instance)

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
@@ -142,14 +142,9 @@ namespace Ryujinx.HLE.HOS.Services.Ro
 
             _owner.CpuMemory.Read(nroAddress, nroData);
 
-            byte[] nroHash = null;
-
             MemoryStream stream = new MemoryStream(nroData);
 
-            using (SHA256 hasher = SHA256.Create())
-            {
-                nroHash = hasher.ComputeHash(stream);
-            }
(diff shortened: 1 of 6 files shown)


AcK77 (approved):

gdkchan (approved):
lgtm, thanks.

Ryujinx.HLE/HOS/Services/Ro/IRoInterface.cs
```diff
@@ -142,14 +142,9 @@ public ResultCode ParseNro(out NroInfo res, ServiceCtx context, ulong nroAddress
...
 
-            byte[] nroHash = null;
-
             MemoryStream stream = new MemoryStream(nroData);
 
-            using (SHA256 hasher = SHA256.Create())
-            {
-                nroHash = hasher.ComputeHash(stream);
-            }
+            byte[] nroHash = SHA256.HashData(stream);
```
gdkchan, line 147: Can't this pass `nroData` directly to get rid of the `MemoryStream` (which BTW doesn't seems to be disposed here).

gdkchan: Oh, the stream is used in other places, so it's not that simple. We should fix the disposal issue though.

PR #4210, opened 2023-01-05
Implement missing service calls in `pm`

### LINKED ISSUE #2516: EdiZon NRO (homebrew) crashes on startup
## Bug Report

Exception when running EdiZon snapshot NRO: `System.NotImplementedException: pm:info`
Exception when running EdiZon release NRO: `Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: Attempted to access an invalid memory region.`

Upon loading either homebrew EdiZonNRO, Ryujinx crashes.

Load the latest EdiZon NRO in Ryujinx.

Snapshot: https://github.com/WerWolv/EdiZon/releases/tag/snapshot
Release: https://github.com/WerWolv/EdiZon/releases/tag/v3.0.1 (inside SD.zip > Switch > EdiZon)

### Log file

[release.log](https://github.com/Ryujinx/Ryujinx/files/6907589/release.log)
[snapshot.log](https://github.com/Ryujinx/Ryujinx/files/6907590/snapshot.log)

### Environment?

Ryujinx LDN 2.3 - Windows 10

### Additional context?

Not sure if this homebrew even works with Ryujinx saves, but it should at least load, right?

IsaacMarovitz:
Implement following missing service calls in `pm`

**`pm:info`:**
- `GetProgramId`

**`pm:dmnt`:**
- `GetApplicationProcessId`

Fixes #2516

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Pm/IDebugMonitorInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Pm/IDebugMonitorInterface.cs
@@ -15,13 +15,14 @@ namespace Ryujinx.HLE.HOS.Services.Pm
         public ResultCode GetApplicationProcessId(ServiceCtx context)
         {
             // TODO: Not correct as it shouldn't be directly using kernel objects here
-            KProcess process = KernelStatic.GetApplicationProcess();
-
-            if (process != null)
+            foreach (KProcess process in context.Device.System.KernelContext.Processes.Values)
             {
-                context.ResponseData.Write(process.Pid);
+                if (process.IsApplication)
+                {
+                    context.ResponseData.Write(process.Pid);
 
-                return ResultCode.Success;
+                    return ResultCode.Success;
+                }
             }
 
             return ResultCode.ProcessNotFound;
--- a/Ryujinx.HLE/HOS/Services/Pm/IDebugMonitorInterface.cs
+++ b/Ryujinx.HLE/HOS/Services/Pm/IDebugMonitorInterface.cs
@@ -14,6 +14,7 @@ namespace Ryujinx.HLE.HOS.Services.Pm
         // GetProgramId() -> sf::Out<ncm::ProgramId> out_process_id
         public ResultCode GetApplicationProcessId(ServiceCtx context)
         {
+            // TODO: Not correct as it shouldn't be directly using kernel objects here
             KProcess process = KernelStatic.GetApplicationProcess();
 
             if (process != null)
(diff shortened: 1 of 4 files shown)


marysaka (approved):
Not technically correct but I think that will do for now.

TSRBerry:
> Not technically correct but I think that will do for now.

Hm if it's not a stub do we want to have stuff in master that's not correct?

marysaka:
> > Not technically correct but I think that will do for now.
> 
> Hm if it's not a stub do we want to have stuff in master that's not correct?

It is not correct as it shouldn't be directly using kernel objects here.
Making it an accurate implementation would require implementing the register interface part and ensure we register every processes started with it.

Ryujinx.HLE/HOS/Services/Pm/IInformationInterface.cs
```diff
@@ -1,8 +1,30 @@
...
+        [CommandHipc(0)]
+        // GetProgramId(os::ProcessId process_id) -> sf::Out<ncm::ProgramId> out
+        public ResultCode GetProgramId(ServiceCtx context)
+        {
+            ulong pid = context.RequestData.ReadUInt64();
+
+            // TODO: Not correct as it shouldn't be directly using kernel objects here
+            KProcess process = KernelStatic.GetProcessByPid(pid);
+
+            if (process != null)
```
AcK77, line 20:
```suggestion
            // TODO: Not correct as it shouldn't be directly using kernel objects here
            if (context.Device.System.KernelContext.Processes.TryGetValue(pid, out KProcess process))
```

Ryujinx.HLE/HOS/Services/Pm/IDebugMonitorInterface.cs
```diff
@@ -10,6 +10,23 @@ class IDebugMonitorInterface : IpcService
...
+        {
+            // TODO: Not correct as it shouldn't be directly using kernel objects here
+            KProcess process = KernelStatic.GetApplicationProcess();
+
+            if (process != null)
+            {
+                context.ResponseData.Write(process.Pid);
+
+                return ResultCode.Success;
+            }
```
AcK77, line 25:
```suggestion
            // TODO: Not correct as it shouldn't be directly using kernel objects here
            foreach (KProcess process in context.Device.System.KernelContext.Processes.Values)
            {
                if (process.IsApplication)
                {
                    context.ResponseData.Write(process.Pid);

                    return ResultCode.Success;
                }
            }
```

AcK77 (changes requested):
If you apply my 2 suggestions then you can remove the change in KernelStatic.

AcK77 (approved):

PR #4211, opened 2023-01-06
Remove textures from cache on unmap if not mapped and modified

gdkchan:
Currently we have a `AutoDeleteCache` that keeps the mostly frequently used textures on the cache, while deleting the least used ones. It has a hard limit of 2048 entries right now, so if the count gets past that limit, it will start deleting textures. In addition to that we remove data when the storage for different textures overlaps, and that works well for most games. However there is a rare scenario where this does not work well, which is the game frequently remapping textures at different memory regions. The fixed limit on the number of entries does not work well when each texture is very large. We can use a very high amount of VRAM with a low amount of textures. And the overlap removal doesn't help because the textures are located in different memory regions.

To fix this issue, this change forces the texture to be removed from the `AutoDeleteCache` if it has been unmapped, and there's no GPU modified region left that is still on a mapped sub-range. Since we already force remove the texture from the pools on unmap, this effectively causes the texture to be deleted since nothing will be holding a reference to it.

This fixes a VRAM "leak" on the game Witch`s Garden (which is technically not a "leak" because it would eventually stop growing, but the usage is so high that probably anyone would run out of memory after a while, so we might as well call it a leak).

VRAM usage on task manager:
master:
image
(Eventually would run out of memory and crash).
PR:
image
Scene:
image

Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -1,4 +1,5 @@
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Ryujinx.Graphics.Gpu.Image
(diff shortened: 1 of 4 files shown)


riperiperi (approved):
LGTM. 

Unrelated, rules for removing overlaps here are still a little weird, though I think that's to avoid removing textures used in rotation.

AcK77 (approved):

PR #4215, opened 2023-01-06
[hipc] Fix 'Unexpected result code Success returned' in Reply()

TSRBerry:
I recently noticed a crash I was getting when minimizing Ryujinx.Headless.SDL2 while it was loading a game.

This PR fixes this crash by adding a new method `AbortOnFailureUnless` and replacing the call to `AbortUnless` with it inside the `Api.Reply` method.

This results in `AbortUnless` being unused currently, should it be removed?


gdkchan (approved):
NIce, thanks for fixing this.

AcK77 (approved):

PR #4220, opened 2023-01-07
Horizon: Impl Prepo, Fixes bugs, Clean things

AcK77:
Since #4188 services have to be ported to our new IPC handling, to beginning, I've ported prepo which is an easy one, I've done some changes too:

- `sm` service in original impl use a `Info` logging when a service is registered, I've changed it into `Debug` to avoid spamming when new services are registered.
- There is a bug in `HipcGenerator` when a IPC command have no args, this is now fixed.
- When a IPC command is unimplemented our new handling currently return the accurate result code, but if the guest doesn't crash, then we can't know (there is currently a warning log) if there is an unimplement command. So I've added a `ThrowOnInvalidCommandIds` options, and then throw an exception when we reach an unimplemented command. This is the default behavior now.
- New `lm` service is cleaned, some classes have been renamed to reflect the real service names. Interfaces are added.
- New `sm` service have been cleaned too. Interfaces are added. Nothing more. 
- New `prepo` service in implemented, everything needed was checked by RE, even the ~annoying~ const values. 
  - `RequestImmediateTransmission` and `GetTransmissionStatus` have been updated to reflect real service a bit.
  - `IsUserAgreementCheckEnabled` and `SetUserAgreementCheckEnabled` have been added because why not.
  - `ProcessPlayReport` have been a bit improved.
  - `prepo:d` have been added, even if it's only available in FW1.0.0, just in case.
- Add code to waiting all services registeration before running guest (@gdkchan code).
- Some cleanup here and there.


Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,219 @@
...
+        [CmifCommand(20100)]
+        // SaveSystemReport(u64, pid, buffer<u8, 9>, buffer<bytes, 5>)
+        public Result SaveSystemReport([Buffer(HipcBufferFlags.In | HipcBufferFlags.Pointer)] ReadOnlySpan<byte> gameRoomBuffer, [Buffer(HipcBufferFlags.In | HipcBufferFlags.MapAlias)] ReadOnlySpan<byte> reportBuffer, [ClientProcessId] ulong pid)
+        {
+            if ((_permissionLevel & PrepoServicePermissionLevel.System) != 0)
+            {
+                return PrepoResult.PermissionDenied;
+            }
+
+            // We don't care about the differences since we don't use the play report.
```
gdkchan, line 121: Which "differences" is this comment talking about?

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,219 @@
...
+        // IsUserAgreementCheckEnabled() -> u8
+        public Result IsUserAgreementCheckEnabled(out bool enabled)
+        {
+            enabled = false;
+
+            if (_permissionLevel == PrepoServicePermissionLevel.User || _permissionLevel == PrepoServicePermissionLevel.System)
+            {
+                enabled = _userAgreementCheckEnabled;
+
+                // If "enabled" is false it's sets some internal fields to 0.
```
gdkchan, line 148:
```suggestion
                // If "enabled" is false, it sets some internal fields to 0.
```

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,219 @@
...
+                // Then it mount "prepo-sys:/is_user_agreement_check_enabled.bin" and return the contained bool.
```
gdkchan, line 149:
```suggestion
                // Then, it mounts "prepo-sys:/is_user_agreement_check_enabled.bin" and returns the contained bool.
```

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        [CmifCommand(40101)] // 2.0.0+
+        // SetUserAgreementCheckEnabled(u8)
+        public Result SetUserAgreementCheckEnabled(bool enabled)
+        {
+            if (_permissionLevel == PrepoServicePermissionLevel.User || _permissionLevel == PrepoServicePermissionLevel.System)
+            {
+                _userAgreementCheckEnabled = enabled;
+
+                // If "enabled" is false it's sets some internal fields to 0.
```
gdkchan, line 166: More or less the same suggestions as the function above, since the comment is mostly the same.

Ryujinx.Horizon/Sdk/Account/UserId.cs
```diff
@@ -0,0 +1,63 @@
+﻿using LibHac.Account;
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Horizon.Sdk.Account
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly record struct UserId
```
gdkchan, line 10: Can we call it `Uid` to match the official name?

Ryujinx.Horizon/Sdk/Sf/Cmif/ServiceDispatchTableBase.cs
```diff
@@ -39,17 +39,19 @@ protected Result ProcessMessageImpl(ref ServiceDispatchContext context, ReadOnly
 
             if (!entries.TryGetValue((int)commandId, out var commandHandler))
             {
-                Logger.Warning?.Print(LogClass.KernelIpc, $"{objectName} command ID 0x{commandId:X} is not implemented");
```
gdkchan, line 42: Keeping the warning is still useful for when ignore missing services is enabled.

Ryujinx.Horizon/Sdk/Sf/Hipc/ServerManagerBase.cs
```diff
@@ -198,15 +198,12 @@ private void ProcessWaitList()
...
-                case UserDataTag.Server:
-                    return ProcessForServer(holder);
-                case UserDataTag.Session:
-                    return ProcessForSession(holder);
-                default:
-                    throw new NotImplementedException(((UserDataTag)holder.UserData).ToString());
-            }
+                UserDataTag.Server => ProcessForServer(holder),
+                UserDataTag.Session => ProcessForSession(holder),
+                _ => throw new NotImplementedException(((UserDataTag)holder.UserData).ToString()),
```
gdkchan, line 205:
```suggestion
                _ => throw new NotImplementedException(((UserDataTag)holder.UserData).ToString())
```

Ryujinx.Horizon/Sdk/Sf/Hipc/ServerSessionManager.cs
```diff
@@ -187,17 +190,12 @@ private Result ProcessRequestImpl(ServerSession session, Span<byte> inMessage, S
...
-                    return DispatchRequest(session.ServiceObjectHolder, session, inMessage, outMessage);
-                case CommandType.Control:
-                case CommandType.ControlWithContext:
-                    return DispatchManagerRequest(session, inMessage, outMessage);
-                default:
-                    return HipcResult.UnknownCommandType;
-            }
+                CommandType.Request or CommandType.RequestWithContext => DispatchRequest(session.ServiceObjectHolder, session, inMessage, outMessage),
+                CommandType.Control or CommandType.ControlWithContext => DispatchManagerRequest(session, inMessage, outMessage),
+                _ => HipcResult.UnknownCommandType,
```
gdkchan, line 197:
```suggestion
                _ => HipcResult.UnknownCommandType
```

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                if (userId.IsNull)
+                {
+                    return PrepoResult.InvalidArgument;
+                }
+            }
+
+            string gameRoom = Encoding.UTF8.GetString(gameRoomBuffer).TrimEnd();
+
+            if (gameRoom == string.Empty)
```
gdkchan, line 188: Is the service really checking the "string", and not the buffer length?

Ryujinx.Horizon/Prepo/PrepoService.cs
AcK77: Only difference I noticed by a quick RE is just the report Kind, I remove the comment then.

Ryujinx.Horizon/Sdk/Account/UserId.cs
AcK77: True, I change it to reflect `nn::account::Uid`

Ryujinx.Horizon/LogManager/ILogService.cs
```diff
@@ -4,16 +4,17 @@
...
 namespace Ryujinx.Horizon.LogManager
 {
-    partial class LmLog : IServiceObject
+    partial class ILogService : IServiceObject
     {
         public LogDestination LogDestination { get; set; } = LogDestination.TargetManager;
 
         [CmifCommand(0)]
-        public Result OpenLogger(out LmLogger logger, [ClientProcessId] ulong clientProcessId)
+        // OpenLogger(u64, pid) -> object<nn::lm::ILogger>
```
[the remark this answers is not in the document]
AcK77: I think it's still nice to keep them, as example for `SetDestination(LogDestination destination)`, `LogDestination` size is u32, we can know that (for later) with the commented symbol, if we remove them it could be hard to point out issue later (same thing available for pid and the u64 placeholder).
It's just one line for each call, and for what it cost to keep them, it's still nice to have the "real" information about the call.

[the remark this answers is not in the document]
gdkchan: The cost is that everytime someone implements a new service function, someone will need to add this comment and figure out the proper naming and proper formatting for it. A lot of them doesn't even reflect the latest information as they come from very old auto-generated SwIPC definitions.

For example on LM, we have `Log(buffer<unknown, 0x21>)` when we already know that this buffer contains the message and we even parse it and display., so it's certainly not unknow. And the function itself already has all this information in a much nicer format:
```cs
Result Log([Buffer(HipcBufferFlags.In | HipcBufferFlags.AutoSelect)] Span<byte> message)
```
It is telling that the buffer is a input buffer with "auto-select" mode, rather than just the value 0x21 which at a first glance doesn't really say anything.

To give another example, on prepo there's `SaveReport(u64, pid, buffer<u8, 9>, buffer<bytes, 5>)`. The type `u8` seems wrong, the correct type is `char` since it is a string. Also what does "bytes" mean? Certainly all buffers contain "bytes" in then. How should I decide if a buffer has "bytes" or "u8" if I'm manually writing the implementation for a function and need to include this comment?

And then we have things like `IsUserAgreementCheckEnabled() -> u8`. This function returns a bool, but here it's called `u8`. I have seen some cases of it called `bool` and some of it being called `b8`. So which one should be used?

Ryujinx.Horizon/HorizonStatic.cs
```diff
@@ -21,24 +21,24 @@ static class HorizonStatic
         [ThreadStatic]
         private static int _threadHandle;
 
-        public static HorizonOptions Options => _options;
-        public static ISyscallApi Syscall => _syscall;
-        public static IVirtualMemoryManager AddressSpace => _addressSpace;
-        public static IThreadContext ThreadContext => _threadContext;
-        public static int CurrentThreadHandle => _threadHandle;
+        public static HorizonOptions        Options             => _options;
```
gdkchan, line 24: FWIW I'm not sure what's the point of manually aligning everything like this. I get that in some cases it improves readability a little bit, but then every time we rename something, we need to manually inspect each line where it is used an re-align it, which is rather counter-productive. Plus by default Visual Studio will remove this alignment if you try to format the document or copy a block of code, so one has to constantly fight against it to maintain the alignment...

Ryujinx.Horizon/LogManager/LogService.cs
```diff
@@ -4,16 +4,18 @@
...
     {
         public LogDestination LogDestination { get; set; } = LogDestination.TargetManager;
 
         [CmifCommand(0)]
-        public Result OpenLogger(out LmLogger logger, [ClientProcessId] ulong clientProcessId)
+        // OpenLogger(u64, pid) -> object<nn::lm::ILogger>
+        public Result OpenLogger(out LmLogger logger, [ClientProcessId] ulong pid)
         {
-            logger = new LmLogger(this, clientProcessId);
+            // NOTE: Internal name is ILogger, but we rename it LmLogger to avoid name clash with Ryujinx.Common.Logging logger.
```
gdkchan, line 15: I think the official name is just `Logger` no? `ILogger` is the interface.

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        [CmifCommand(10200)]
+        // RequestImmediateTransmission()
+        public Result RequestImmediateTransmission()
+        {
+            _immediateTransmissionEnabled = true;
+
+            // It signals an event of nn::prepo::detail::service::core::TransmissionStatusManager that requests the transmission of the report.
+            // Since we don't use reports it's fine to do nothing.
```
gdkchan, line 72:
```suggestion
            // Since we don't use reports, it's fine to do nothing.
```

Ryujinx.Horizon/Prepo/PrepoService.cs
gdkchan: There was no answer here, I guess you checked and the implementation is correct?

Ryujinx.Horizon/Sdk/Sf/Cmif/DomainServiceObjectDispatchTable.cs
```diff
@@ -21,7 +21,7 @@ private Result ProcessMessageImpl(ref ServiceDispatchContext context, ServerDoma
 
             var inHeader = MemoryMarshal.Cast<byte, CmifDomainInHeader>(inRawData)[0];
 
-            ReadOnlySpan<byte> inDomainRawData = inRawData.Slice(Unsafe.SizeOf<CmifDomainInHeader>());
+            ReadOnlySpan<byte> inDomainRawData = inRawData[Unsafe.SizeOf<CmifDomainInHeader>()..];
```
gdkchan, line 24: FWIW I prefer how it was before. The new syntax is confusing for me. I guess it's just because I'm not used to it yet...

Ryujinx.Horizon/Sdk/Sf/Hipc/ServerSessionManager.cs
```diff
@@ -187,17 +192,12 @@ private Result ProcessRequestImpl(ServerSession session, Span<byte> inMessage, S
...
-                case CommandType.Request:
-                case CommandType.RequestWithContext:
-                    return DispatchRequest(session.ServiceObjectHolder, session, inMessage, outMessage);
-                case CommandType.Control:
-                case CommandType.ControlWithContext:
-                    return DispatchManagerRequest(session, inMessage, outMessage);
-                default:
-                    return HipcResult.UnknownCommandType;
-            }
+                CommandType.Request or CommandType.RequestWithContext => DispatchRequest(session.ServiceObjectHolder, session, inMessage, outMessage),
```
gdkchan, line 197: This is another case where I think the previous version was easier to understand.

Ryujinx.Horizon/Sdk/Sm/SmApi.cs
```diff
@@ -7,16 +7,18 @@ namespace Ryujinx.Horizon.Sdk.Sm
 {
     class SmApi
     {
+        private const string smName = "sm:";
```
gdkchan, line 10:
```suggestion
        private const string SmName = "sm:";
```

gdkchan (commented):
Refactoring should have been probably left for another PR since it's not related to the original goal of this change at all.

Ryujinx.Horizon/Prepo/PrepoService.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                return PrepoResult.InvalidState;
+            }
+
+            if (reportBuffer.Length == 0)
+            {
+                return PrepoResult.InvalidBufferSize;
+            }
+
+            // NOTE: Service call arp:r using the pid to get the application id, if it fails PrepoResult.InvalidPid is returned.
```
gdkchan, line 201:
```suggestion
            // NOTE: The service calls arp:r using the pid to get the application id, if it fails PrepoResult.InvalidPid is returned.
```

Ryujinx.Horizon/LogManager/LogService.cs
AcK77: `nn::lm::ILogger`

Ryujinx.Horizon/HorizonStatic.cs
AcK77: If people doesn't want to realign things it's fine. I just think it improves readability so I changed it. And well, it's not like the code will changed a lot in a near future too...

Ryujinx.Horizon/LogManager/ILogService.cs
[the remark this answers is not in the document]
AcK77: I've removed all symbols then

gdkchan (approved):
lgtm, thanks. Very nice to see more services migrated.
We will probably need to decide what the namespace will be when we implement BCAT and friends though, since BCAT, Prepo and a few others are hosted by the same process (and will share the same Main function).

marysaka (approved):

PR #4222, opened 2023-01-07
ava: Fix regression caused by #4013

marysaka:
Avalonia seems to not like when the artifact doesns't match the root namespace...

Address that by moving the binary to "Ryujinx" like we do on macOS build.


.github/workflows/build.yml
```diff
@@ -69,6 +69,12 @@ jobs:
       - name: Publish Ryujinx.Ava
         run: dotnet publish -c "${{ matrix.configuration }}" -r "${{ matrix.DOTNET_RUNTIME_IDENTIFIER }}" -o ./publish_ava -p:Version="${{ env.RYUJINX_BASE_VERSION }}" -p:DebugType=embedded -p:SourceRevisionId="${{ steps.git_short_hash.outputs.result }}" -p:ExtraDefineConstants=DISABLE_UPDATER Ryujinx.Ava --self-contained true
         if: github.event_name == 'pull_request'
+      - name: Rename Avalonia (Windows)
+        run: mv ./publish_ava/Ryujinx.Ava.exe ./publish_ava/Ryujinx.exe
+        if: runner.os == 'Windows' && github.event_name == 'pull_request'
+      - name: Rename Avalonia (Unix)
+        run: mv ./publish_ava/Ryujinx ./publish_ava/Ryujinx
```
merryhime, line 76: Is this correct.

marysaka: I got that wrong, again :upside_down_face:

PR #4227, opened 2023-01-07
Ava UI: Implement built-in console window

merryhime:
image

Auto-scroll keeps scroll locked to end.
Open Logs open the logs folder for easy copying of logs.

Linked to Show Console configuration option:
image

Note: I attempted to also capture stdout for mvk, but unfortunately the simple implementation using Console.SetOut/Console.SetError didn't work.

### DIFF
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -123,8 +123,6 @@ namespace Ryujinx.Ava
             // Delete backup files after updating.
             Task.Run(Updater.CleanupUpdate);
 
-            Console.Title = $"Ryujinx Console {Version}";
-
             // Hook unhandled exception and process exit events.
             AppDomain.CurrentDomain.UnhandledException += (sender, e) => ProcessUnhandledException(e.ExceptionObject as Exception, e.IsTerminating);
             AppDomain.CurrentDomain.ProcessExit        += (sender, e) => Exit();
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -1,5 +1,6 @@
 using Avalonia;
 using Avalonia.Threading;
+using Ryujinx.Ava.UI.Models;
 using Ryujinx.Ava.UI.Windows;
 using Ryujinx.Common;
 using Ryujinx.Common.Configuration;
@@ -136,6 +137,7 @@ namespace Ryujinx.Ava
 
             // Initialize the logger system.
             LoggerModule.Initialize();
+            InMemoryLogTarget.Register();
 
             // Register mime types on linux.
             if (OperatingSystem.IsLinux())
--- a/Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
+++ b/Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
@@ -4,20 +4,8 @@ using System.Collections.ObjectModel;
 
 namespace Ryujinx.Ava.UI.Models
 {
-    internal class InMemoryLogTarget : ILogTarget
+    class InMemoryLogTarget : ILogTarget
     {
-        internal struct Entry
-        {
-            public Entry(Color color, string text)
-            {
-                Color = color;
-                Text = text;
-            }
-
-            public Color Color { get; }
-            public string Text { get; }
-        }
-
         private static Color GetLogColor(LogLevel level)
         {
             return level switch
@@ -40,7 +28,7 @@ namespace Ryujinx.Ava.UI.Models
 
         private const int MaximumSize = 1000;
 
-        public readonly ObservableCollection<Entry> Entries;
+        public readonly ObservableCollection<InMemoryLogTargetEntry> Entries;
 
         string ILogTarget.Name => _name;
 
@@ -50,7 +38,7 @@ namespace Ryujinx.Ava.UI.Models
         {
             _formatter = new DefaultLogFormatter();
             _name      = name;
-            Entries    = new ObservableCollection<Entry>();
+            Entries    = new ObservableCollection<InMemoryLogTargetEntry>();
         }
 
         public static void Register()
@@ -65,11 +53,11 @@ namespace Ryujinx.Ava.UI.Models
 
         private void AddEntry(Color color, string text)
         {
-            Entries.Insert(0, new Entry(color, text));
+            Entries.Add(new InMemoryLogTargetEntry(color, text));
 
             if (Entries.Count > MaximumSize)
             {
-                Entries.RemoveAt(Entries.Count - 1);
+                Entries.RemoveAt(0);
             }
         }
 
--- a/Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
+++ b/Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
@@ -4,41 +4,47 @@ using System.Collections.ObjectModel;
 
 namespace Ryujinx.Ava.UI.Models
 {
-    public class InMemoryLogTarget : ILogTarget
+    internal class InMemoryLogTarget : ILogTarget
     {
-        public class Entry
+        internal struct Entry
         {
-            public Color Color { get; set; }
-            public string Text { get; set; }
+            public Entry(Color color, string text)
+            {
+                Color = color;
+                Text = text;
+            }
+
+            public Color Color { get; }
+            public string Text { get; }
         }
 
-        private static Color GetLogColor(LogLevel level) => level switch {
-            LogLevel.Info    => Colors.White,
-            LogLevel.Warning => Colors.Yellow,
-            LogLevel.Error   => Colors.Red,
-            LogLevel.Stub    => Colors.DarkGray,
-            LogLevel.Notice  => Colors.Cyan,
-            LogLevel.Trace   => Colors.DarkCyan,
-            LogLevel.StdErr  => Colors.Gray,
-            _                => Colors.Gray,
-        };
+        private static Color GetLogColor(LogLevel level)
+        {
+            return level switch
+            {
+                LogLevel.Info    => Colors.White,
+                LogLevel.Warning => Colors.Yellow,
+                LogLevel.Error   => Colors.Red,
+                LogLevel.Stub    => Colors.DarkGray,
+                LogLevel.Notice  => Colors.Cyan,
+                LogLevel.Trace   => Colors.DarkCyan,
+                LogLevel.StdErr  => Colors.Gray,
+                _                => Colors.Gray,
+            };
+        }
 
         private static readonly InMemoryLogTarget _instance = new InMemoryLogTarget("inMemory");
 
         private readonly ILogFormatter _formatter;
         private readonly string _name;
 
-        private const int _maximumSize = 20000;
+        private const int MaximumSize = 20000;
 
-        public ObservableCollection<Entry> Entries;
+        public readonly ObservableCollection<Entry> Entries;
 
-        string ILogTarget.Name { get => _name; }
+        string ILogTarget.Name => _name;
 
         public static InMemoryLogTarget Instance => _instance;
-
-        static InMemoryLogTarget()
-        {
-        }
 
         private InMemoryLogTarget(string name)
         {
@@ -59,13 +65,9 @@ namespace Ryujinx.Ava.UI.Models
 
         private void AddEntry(Color color, string text)
         {
-            Entries.Add(new Entry()
-            {
-                Color = color,
-                Text = text,
-            });
+            Entries.Add(new Entry(color, text));
 
-            if (Entries.Count > _maximumSize)
+            if (Entries.Count > MaximumSize)
             {
                 Entries.RemoveAt(0);
             }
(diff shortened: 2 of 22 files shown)


Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml
```diff
@@ -0,0 +1,65 @@
...
+            Grid.Row="0"
+            Grid.Column="0"
+            VerticalAlignment="Top"
+            HorizontalAlignment="Right"
+            Background="{DynamicResource ThemeContentBackgroundColor}">
+            <CheckBox
+                HorizontalAlignment="Center"
+                Name="AutoScrollCheckBox"
+                IsChecked="{Binding AutoScroll}"
+                Content="Auto-scroll" />
```
AcK77, line 58: This should be added in Locale

Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml
```diff
@@ -0,0 +1,65 @@
...
+            <Button
+                HorizontalAlignment="Center"
+                Command="{ReflectionBinding OpenLogsFolder}"
+                Content="Open Logs" />
```
AcK77, line 62: This should be added in Locale

AcK77 (changes requested):
About the design:
- The console title window should include the Ryujinx version as the original one.
- If `Auto-scroll` is checked, the scrollbar should be still active (even if it auto-scroll down, because if you stop the emulation then you can't scroll up to read the log without unchecking `Auto-scroll`.
- Maybe we should be able to select the "native" console or the Avalonia console as output in the settings or at least add `<OutputType>WinExe</OutputType>` in the csproj ?

I really like the idea, that could lead to more possibilities like some debug tools or anything related to debug.

merryhime:
Added in stderr redirection: stderr output from external code now goes through the logging system. Re-enable stdout output because not thinking of doing stdout redirection currently.

This allows for easier mvk/mesa debugging stderr will now show up in log files.

image

merryhime:
@AcK77 

Good points.

* Added version to title as requested
* Enabled user scrolling when `Auto-scroll` is active. Originally the scrolling was just locked to always be at the bottom
* `WinExe` added
* Added `stderr` redirection
* I've also reenabled logging via stdout so running via the command line is always a fallback option if required

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+            public string Text { get; set; }
+        }
+
+        private static Color GetLogColor(LogLevel level) => level switch {
+            LogLevel.Info    => Colors.White,
+            LogLevel.Warning => Colors.Yellow,
+            LogLevel.Error   => Colors.Red,
+            LogLevel.Stub    => Colors.DarkGray,
+            LogLevel.Notice  => Colors.Cyan,
+            LogLevel.Trace   => Colors.DarkCyan,
```
AcK77, line 21:
```suggestion
            LogLevel.Warning => Color.FromRgb(235, 205, 85),
            LogLevel.Error   => Color.FromRgb(255, 95, 85),
            LogLevel.Stub    => Colors.DarkGray,
            LogLevel.Notice  => Color.FromRgb(2, 197, 229),
            LogLevel.Trace   => Color.FromRgb(200, 146, 245),
```

Just a little cosmetic suggestion, you can ignore it if you don't like.

image

gdkchan: Personally I prefer the current colors, they are slightly brighter which makes it easier to read on a dark background.

Ryujinx.Ava/Program.cs
```diff
@@ -1,5 +1,6 @@
 using Avalonia;
```
AcK77, line 1: With "WinExe" added L126 `Console.Title = $"Ryujinx Console {Version}";` have to be removed because there is no Console and the emulator can't start!

AcK77 (changes requested):

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
AcK77: It's just following some of the Switch colors scheme

GamerzHell9137:
Idk if its known but there's 2 console windows now on Windows.

Ryujinx.Common/Logging/Logger.cs
```diff
@@ -14,6 +15,8 @@ public static class Logger
 
         private static readonly List<ILogTarget> m_LogTargets;
 
+        private static readonly StdErrAdapter m_stdErrAdapter;
```
gdkchan, line 18: The naming here seems a bit weird. We usually don't use `m_` prefix, just `_`. And I think `m_` would not be correct here anyway since those are static fields? Those are usually prefixed with `s_`.

Ryujinx.Ava/Program.cs
merryhime: Hah. Good catch.

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    public class InMemoryLogTarget : ILogTarget
+    {
+        public class Entry
+        {
+            public Color Color { get; set; }
+            public string Text { get; set; }
+        }
+
+        private static Color GetLogColor(LogLevel level) => level switch {
```
gdkchan, line 15:
```suggestion
        private static Color GetLogColor(LogLevel level)
        {
            return level switch
            {
```
Basically the same suggestion as the other PR.

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        private const int _maximumSize = 20000;
+
+        public ObservableCollection<Entry> Entries;
+
+        string ILogTarget.Name { get => _name; }
+
+        public static InMemoryLogTarget Instance => _instance;
+
+        static InMemoryLogTarget()
```
gdkchan, line 39: Why does it need the empty static constructor?

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+            LogLevel.StdErr  => Colors.Gray,
+            _                => Colors.Gray,
+        };
+
+        private static readonly InMemoryLogTarget _instance = new InMemoryLogTarget("inMemory");
+
+        private readonly ILogFormatter _formatter;
+        private readonly string _name;
+
+        private const int _maximumSize = 20000;
```
gdkchan, line 31: Since it's `const` it should be named `MaximumSize`.

Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        private void LogEntriesOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            changeCount++;
+        }
+
+        private void AutoScrollCheckBoxOnChecked(object sender, RoutedEventArgs e)
+        {
+            if (AutoScrollCheckBox.IsChecked == true)
```
gdkchan, line 39:
```suggestion
            if (AutoScrollCheckBox.IsChecked)
```

Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                ConsoleScrollViewer.ScrollToEnd();
+            }
+        }
+
+        private void ConsoleItemsControlOnLayoutUpdated(object sender, EventArgs e)
+        {
+            if (AutoScrollCheckBox.IsChecked == true && autoScrollChangeCount != changeCount)
```
gdkchan, line 47:
```suggestion
            if (AutoScrollCheckBox.IsChecked && autoScrollChangeCount != changeCount)
```

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public ObservableCollection<Entry> Entries;
```
gdkchan, line 33: Can this be private? Ideally it should not be possible to modify internal state outside of the `InMemoryLogTarget` class.

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
+using Avalonia.Media;
+using Ryujinx.Common.Logging;
+using System.Collections.ObjectModel;
+
+namespace Ryujinx.Ava.UI.Models
+{
+    public class InMemoryLogTarget : ILogTarget
+    {
+        public class Entry
```
gdkchan, line 9: If the list is made private, I think this can become private too.

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            public Color Color { get; set; }
```
gdkchan, line 11: Unless there's a need to modify those values after instantiation, I think it would be better to remove the setter and initialize it on the constructor. Maybe you could make it a struct too to avoid the object allocation cost etc.

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
```diff
@@ -0,0 +1,78 @@
...
+        private static readonly InMemoryLogTarget _instance = new InMemoryLogTarget("inMemory");
+
+        private readonly ILogFormatter _formatter;
+        private readonly string _name;
+
+        private const int _maximumSize = 20000;
+
+        public ObservableCollection<Entry> Entries;
+
+        string ILogTarget.Name { get => _name; }
```
gdkchan, line 35:
```suggestion
        string ILogTarget.Name => _name;
```
I think this is valid too?

Ryujinx.Common/Logging/LogLevel.cs
```diff
@@ -10,6 +10,7 @@ public enum LogLevel
         Guest,
         AccessLog,
         Notice,
-        Trace
+        Trace,
+        StdErr,
```
gdkchan, line 14: Do we need a new log level for that? Can't we use the existing `Error` level?

merryhime: I wasn't entirely sure what level to put it at; some of it is informational, some of it is errors.

Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml.cs
merryhime: IsChecked is `bool?`

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
merryhime: This is exposed and bound to UI. Not sure how to achieve that otherwise.

Ryujinx.Ava/UI/Windows/ConsoleWindow.axaml.cs
gdkchan: I see.

TSRBerry:
I noticed the new console window can cause some performance issues:

When a lot of stuff gets logged really fast, the window can't keep up and freezes at some point. This also freezes the main window.
image
I'm currently hearing the smash title song in the screenshot, but can't see anything because both windows are frozen.
Meanwhile the native console window still logs more messages.

Probably also related to very fast logging:
As long as the console window is open I drop some fps which causes noticeable audio stutters. 
image

As soon as I close the console window the stutters are gone and my fps are back to normal:
image

Ryujinx.Ava/UI/Models/InMemoryLogTarget.cs
TSRBerry: Suggested colors look great!
During testing I didn't notice any readability issues, but if they make messages harder to read, we should probably adjust them a bit.

merryhime:
I suspect this is because ItemsPresenter doesn’t do virtualisation but will look into it.

merryhime:
I have also notice a disadvantage of this approach is that since this relies on the emulator being functional in order to have working output, sometimes things can be missed. For example, in a recent instance when the emulator threw a `System.InvalidOperationException`, this was shown on stdout and in file logs but not in this window, because the emulator had crashed.

MutantAura:
> I have also noticed a disadvantage of this approach is that since this relies on the emulator being functional in order to have working output, sometimes things can be missed. For example, in a recent instance when the emulator threw a `System.InvalidOperationException`, this was shown on stdout and in file logs but not in this window, because the emulator had crashed.

Is this a huge issue as long as it's in the file log? Realistically if the program has crashed then the console window would usually close shortly after anyway (at least on Windows this is the behaviour). It may also push more people to actually send us actual log files instead of console screenshots...

merryhime:
I personally feel there are reliability and inherent performance issues with a built in console window, so closing this.

PR #4228, opened 2023-01-07
Relax Vulkan requirements

gdkchan:
This is relaxing the Vulkan requirements, basically optionally requesting more features, to allow it to run on more GPUs.
This mainly targets the Raspberry Pi 4, but it can benefit other mobile GPUs too.

Changes:
- More features are now only requested if the drivers supports it, instead of being always requested.
- Transform feedback is only enabled if the extension is supported and the feature is supported.
- Choose composite alpha flags based on what the driver supports. Not all drivers supports the "Opaque" mode.
- The number of color attachment on a sub pass was always 8 due to the way how `MaxAttachmentIndex` was set. However the Raspberry Pi GPU only supports up to 4. `MaxAttachmentIndex` was changed to actually reflect the maximum index used. This allows games that uses 4 attachments or less to work on those GPUs.
- Buffers were only created on memory pools with the `HostCached` flag. However not all driver exposes a `HostCached` memory pool, and that would make the memory allocation fail on those drivers. It was now changed to take "alternative" flags. If no memory pool is found with the requested flags, it returns one with the alternative flags. The new functionality is used to request an allocation without the "HostCached" flag.

This along with #4114 allows some games to work on the Raspberry Pi 4. Host mapped mode does not work due to the issue mentioned on the Arm64 PR. There are other things to note:
- Performance is too low right now for any game to be considered playable.
  - I'm also using an outdated version of the Vulkan driver. It's possible that the latest version has better performance.
  - I'm also running with stock clocks and don't have any kind of cooling on the board. I don't know how much that would impact performance.
  - I'm also using VNC to stream the video to my computer since I don't have an actual display connected to it. I don't know if or how much the VNC server impacts performance.
- It was only tested on the 8 GB model. I don't think it would work with the 4 GB or lower models.

I'm not sure how we will maintain support for the Raspberry Pi going forward, but I don't think it's worth working around Pi limitations given it likely can't run most games with playable performance anyway. So the fixes it will receive will likely be to issues that also affects more capable GPUs.

Screenshot of some games running:
image
image
image
image
image

### DIFF
--- a/Ryujinx.Graphics.Vulkan/BufferManager.cs
+++ b/Ryujinx.Graphics.Vulkan/BufferManager.cs
@@ -14,6 +14,12 @@ namespace Ryujinx.Graphics.Vulkan
             MemoryPropertyFlags.HostCoherentBit |
             MemoryPropertyFlags.HostCachedBit;
 
+        // Some drivers don't expose a "HostCached" memory type,
+        // so we need those alternative flags for the allocation to succeed there.
+        private const MemoryPropertyFlags DefaultBufferMemoryAltFlags =
+            MemoryPropertyFlags.HostVisibleBit |
+            MemoryPropertyFlags.HostCoherentBit;
+
         private const MemoryPropertyFlags DeviceLocalBufferMemoryFlags =
             MemoryPropertyFlags.DeviceLocalBit;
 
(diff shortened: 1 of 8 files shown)


marysaka (approved):

AcK77:
Needs a rebase

gdkchan:
Rebased.

AcK77 (approved):

PR #4231, opened 2023-01-08
Ava: Make Avalonia use our logging system

merryhime:
Instead of outputting to `System.Diagnostics.Trace`, capture Avalonia logging events and adapt to our logging system.

### DIFF
--- a/Ryujinx.Ava/Helper/LoggerAdapter.cs
+++ b/Ryujinx.Ava/Helper/LoggerAdapter.cs
@@ -1,0 +1,163 @@
+using Avalonia.Utilities;
+using System;
+using System.Text;
+
+namespace Ryujinx.Ava.UI.Helper
+{
+    using AvaLogger = Avalonia.Logging.Logger;
+    using AvaLogLevel = Avalonia.Logging.LogEventLevel;
+    using RyuLogger = Ryujinx.Common.Logging.Logger;
+    using RyuLogClass = Ryujinx.Common.Logging.LogClass;
+
+    public class LoggerAdapter : Avalonia.Logging.ILogSink
+    {
+        public static void Register()
+        {
+            AvaLogger.Sink = new LoggerAdapter();
+        }
+
+        private static RyuLogger.Log? GetLog(AvaLogLevel level) => level switch {
+            AvaLogLevel.Verbose => RyuLogger.Trace,
+            AvaLogLevel.Debug => RyuLogger.Debug,
+            AvaLogLevel.Information => RyuLogger.Info,
+            AvaLogLevel.Warning => RyuLogger.Warning,
+            AvaLogLevel.Error => RyuLogger.Error,
+            AvaLogLevel.Fatal => RyuLogger.Notice,
+            _ => throw new ArgumentOutOfRangeException(nameof(level), level, null)
+        };
+
+        public bool IsEnabled(AvaLogLevel level, string area)
+        {
+            return GetLog(level) != null;
+        }
+
+        public void Log(AvaLogLevel level, string area, object source, string messageTemplate)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, null));
+        }
+
+        public void Log<T0>(AvaLogLevel level, string area, object source, string messageTemplate, T0 propertyValue0)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] {propertyValue0}));
+        }
+
+        public void Log<T0, T1>(AvaLogLevel level, string area, object source, string messageTemplate, T0 propertyValue0,  T1 propertyValue1)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] {propertyValue0, propertyValue1}));
+        }
+
+        public void Log<T0, T1, T2>(AvaLogLevel level, string area, object source, string messageTemplate, T0 propertyValue0, T1 propertyValue1, T2 propertyValue2)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] {propertyValue0, propertyValue1, propertyValue2}));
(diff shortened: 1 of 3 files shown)


Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] {propertyValue0, propertyValue1, propertyValue2}));
+        }
+
+        public void Log(AvaLogLevel level, string area, object source, string messageTemplate, params object[] propertyValues)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, propertyValues));
+        }
+
+        private static string Format<T0, T1, T2>(
```
gdkchan, line 59: `T0`, `T1` and `T2` seems to be unused here. Was the intention here having multiple values with those generic types rather than `object[] values`?

Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -0,0 +1,163 @@
...
+    using RyuLogClass = Ryujinx.Common.Logging.LogClass;
+
+    public class LoggerAdapter : Avalonia.Logging.ILogSink
+    {
+        public static void Register()
+        {
+            AvaLogger.Sink = new LoggerAdapter();
+        }
+
+        private static RyuLogger.Log? GetLog(AvaLogLevel level) => level switch {
```
gdkchan, line 19:
```suggestion
        private static RyuLogger.Log? GetLog(AvaLogLevel level)
        {
            return level switch
            {
```
With the appropriate indentation changes.

Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Text;
+
+namespace Ryujinx.Ava.UI.Helper
+{
+    using AvaLogger = Avalonia.Logging.Logger;
+    using AvaLogLevel = Avalonia.Logging.LogEventLevel;
+    using RyuLogger = Ryujinx.Common.Logging.Logger;
+    using RyuLogClass = Ryujinx.Common.Logging.LogClass;
+
+    public class LoggerAdapter : Avalonia.Logging.ILogSink
```
gdkchan, line 12: Does it need to be `public`?

Ryujinx.Common/Logging/LogClass.cs
```diff
@@ -5,11 +5,12 @@ public enum LogClass
         Application,
         Audio,
         AudioRenderer,
+        Avalonia,
```
gdkchan, line 8: Maybe it should be `GUI` or something along those lines, to allow it to be used with other UIs. For example, if we wanted to add similar logging to GTK too (I know it is going away eventually, but I don't think we should only consider one UI).

Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -0,0 +1,163 @@
...
+                    }
+                    else
+                    {
+                        result.Append('{');
+                        r.Take();
+                    }
+                }
+            }
+
+            if (source is object)
```
gdkchan, line 151: Can source *not* be an `object`? This check seems redundant.

Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        public void Log(AvaLogLevel level, string area, object source, string messageTemplate)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, null));
+        }
+
+        public void Log<T0>(AvaLogLevel level, string area, object source, string messageTemplate, T0 propertyValue0)
+        {
+            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] {propertyValue0}));
```
gdkchan, line 41:
```suggestion
            GetLog(level)?.PrintMsg(RyuLogClass.Avalonia, Format<object, object, object>(area, messageTemplate, source, new object[] { propertyValue0 }));
```
Same for the functions below.

Ryujinx.Ava/Helper/LoggerAdapter.cs
merryhime: Unfortunately they're a required part of the ILogSink interface in 0.10.18, see https://github.com/AvaloniaUI/Avalonia/blob/0.10.18/src/Avalonia.Base/Logging/ILogSink.cs

These functions will be removed in 0.11.0.

Ryujinx.Ava/Helper/LoggerAdapter.cs
merryhime: `source` can be `null`, for which `source is object` becomes `false`.
I can change this if required, but I do note that I am copying this from Avalonia here.
https://github.com/AvaloniaUI/Avalonia/blob/0.10.18/src/Avalonia.Base/Logging/TraceLogSink.cs#L115

Ryujinx.Ava/Helper/LoggerAdapter.cs
merryhime: Noted, I'm not sure what I was thinking, I'll remove the generic types.

Ryujinx.Ava/Helper/LoggerAdapter.cs
gdkchan: Ok, it is fine to keep it then.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

Ryujinx.Common/Logging/LogClass.cs
```diff
@@ -66,6 +66,7 @@ public enum LogClass
         ServiceVi,
         SurfaceFlinger,
         TamperMachine,
-        Vic
+        Ui,
+        Vic,
```
AcK77, line 70:
```suggestion
        Vic
```

TSRBerry (approved):
Lgtm!

PR #4196, opened 2023-01-01
Ava GUI: `AboutWindow` Refactor

IsaacMarovitz:
**Changes:**
- Moves `AboutWindow` to a `ContentDialog`
- Update logos and fix light theme
- Redesign layout slightly
- Add a `ViewModel` and switch `AboutWindow` to compiled bindings

Screenshot 2023-01-01 at 19 03 08

PR #4178, opened 2022-12-25
Ava GUI: `MainWindow` Refactor

IsaacMarovitz:
**Changes:**
- `MainWindow` Refactored

Superceeds part of https://github.com/Ryujinx/Ryujinx/pull/3997

PR #4237, opened 2023-01-09
ava: Fixes regressions from refactoring

### LINKED ISSUE #4235: Right klick kontext menu if you right klick on a game is not working properly. (latest Ava Build)
Since avalonia build 1.1.520 the right klick context menu if you right klick on a game is not working properly.,
The menue is showing up but if you coose an option nothing happens afterwards.
So it's impossible to set titleupdates oder dlc's atm.

Till avalonia version 1.1.519 this was working fine.

AcK77:
This PR fixes some regressions introduced in #4178 and #4196:

- The About window is now used as a `ContentDialog`, which have to be used as an overlay to be able to show it over the rendering control. Without that, while a game is running, the UI just freeze because you can't quit the window which is under the renderer. It's now fixed using the code done in #3938. In the same way, I've made a new static class called `ContentDialogOverlay` which check if the overlay is needed or not, and do the right thing, to avoid duplicate code. I've cleaned `ContentDialog` a bit too. Autoformatted `AboutWindow.axaml` too.

- Right click context menu on game unusable because `*SelectedApplication` were not sets, it's now fixed. Closes #4235.

- `Stop emulation` menu item not working anymore. Because the async method was run in an inlined Task, removing this make thing working back.


gdkchan (approved):
Code lgtm, thanks for the quick fix.

Ryujinx.Ava/UI/Helpers/ContentDialogOverlay.cs
```diff
@@ -0,0 +1,106 @@
...
+using Avalonia.Controls;
+using Avalonia.Controls.ApplicationLifetimes;
+using FluentAvalonia.UI.Controls;
+using Ryujinx.Ava.UI.Windows;
+using System;
+using System.Threading.Tasks;
+
+namespace Ryujinx.Ava.UI.Helpers
+{
+    public static class ContentDialogOverlay
```
TSRBerry, line 11: Why are these methods not part of `ContentDialogHelper`?

AcK77: True, done!

TSRBerry (approved):
Lgtm and can confirm it works on Linux!

PR #4240, opened 2023-01-09
ava: Cleanup AppHost

AcK77:
This PR cleaned up the AppHost file a bit (adding the infamous extra spaces to improve readability), resorting private vars, removing useless vars, improving the code here and there, like the AudioBackend check.

### DIFF
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -41,6 +41,7 @@ using SixLabors.ImageSharp.PixelFormats;
 using SixLabors.ImageSharp.Processing;
 using SPB.Graphics.Vulkan;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Threading;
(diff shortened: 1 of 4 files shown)


Ryujinx.Ava/AppHost.cs
```diff
@@ -527,52 +521,42 @@ public async Task<bool> LoadGuestApplication()
...
-                            Logger.Info?.Print(LogClass.Application, "Loading as XCI.");
-
-                            Device.LoadXci(ApplicationPath);
-
-                            break;
-                        }
-                    case ".nca":
-                        {
-                            Logger.Info?.Print(LogClass.Application, "Loading as NCA.");
+                string extension = Path.GetExtension(ApplicationPath).ToLowerInvariant();
```
TSRBerry, line 524: Why was the switch-case removed?

AcK77: Just a preference, since we use `if else` for other loading, I keep using the same thing for the other ones.

gdkchan: `switch` is more efficient than a `if`/`else if`/`else` chain.

AcK77: Do we really have to be "efficient" for just an extension check ?

gdkchan: No, but I think it looks better too. I don't think all the `if`/`else if` are an improvement.

Ryujinx.Ava/AppHost.cs
```diff
@@ -201,12 +206,12 @@ private unsafe void Renderer_ScreenCaptured(object sender, ScreenCaptureImageInf
                 {
                     lock (_lockObject)
                     {
-                        var currentTime = DateTime.Now;
-                        string filename = $"ryujinx_capture_{currentTime.Year}-{currentTime.Month:D2}-{currentTime.Day:D2}_{currentTime.Hour:D2}-{currentTime.Minute:D2}-{currentTime.Second:D2}.png";
+                        string filename = $"ryujinx_capture_{DateTime.Now.Year}-{DateTime.Now.Month:D2}-{DateTime.Now.Day:D2}_{DateTime.Now.Hour:D2}-{DateTime.Now.Minute:D2}-{DateTime.Now.Second:D2}.png";
```
gdkchan, line 209: This change looks a bit bogus. Accessing `DateTime.Now` several will get the date/time at different points in time, which might cause it to use incorrect time on those file names. For example if someone takes a screenshot right before midnight (let's say 23:59:59) of day 1, it might use that time but then on the next access it might be already the next day (day 2). So the date will be 23:59 of day 2 while it was taken on 23:59 of day 1 which is way off.

Ryujinx.Ava/AppHost.cs
```diff
@@ -1046,7 +968,7 @@ private bool UpdateFrame()
                 }
             }
 
-            // Touchscreen
+            // Touchscreen.
```
gdkchan, line 971:
```suggestion
            // Touch Screen.
```
I'm not sure about this one but I don't think we generally write touch screen as if it was a single word?

Ryujinx.Ava/AppHost.cs
marysaka: Yeah I think the switch case was better here

Ryujinx.Ava/AppHost.cs
TSRBerry: I think it should be touch screen, but I'm not so sure about the dot at the end.

Ryujinx.Ava/AppHost.cs
AcK77: Tbh, I prefer: https://github.com/AcK77/Ryujinx/blob/apphos-refacto/Ryujinx.Ava/AppHost.cs#L494-L569
Instead of: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/AppHost.cs#L500-L585
But if you 3 still prefer the old one, I can revert this change :)

Ryujinx.Ava/AppHost.cs
AcK77: True! I'll revert this change ;)

Ryujinx.Ava/AppHost.cs
gdkchan: If other places are calling it "Touchscreen", then just keep it as "Touchscreen".

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4243, opened 2023-01-09
ava: Generate Locale menu automatically

AcK77:
Currently in `MenuMainBarView.axaml` we list all available languages and hardcode the language name with the language key. It's a bit bad beause if we want to add a new language, we have to edit the `csproj` and the `axaml` with the translated language name and the language code. I've put all translations in their respective locale files, add code into `MainMenuBarView` constructor to generate the menu automatically. Now we just have to edit the `csproj` if we want to add a new language.


TSRBerry (commented):
Since we recently added a source generator to generate a locale enum, could this also be done like that? This way we wouldn't need to do this during runtime

marysaka:
> Since we recently added a source generator to generate a locale enum, could this also be done like that? This way we wouldn't need to do this during runtime

Technically this isn't reflection here but yeah it would probably be better to have cdodegen for that.

marysaka (approved):

Ryujinx.Ava/Assets/Locales/de_DE.json
```diff
@@ -609,5 +609,6 @@
   "Search": "Suche",
   "UserProfilesRecoverLostAccounts": "Konto wiederherstellen",
   "Recover": "Wiederherstellen",
-  "UserProfilesRecoverHeading": "Speicherstände wurden für die folgenden Konten gefunden"
+  "UserProfilesRecoverHeading": "Speicherstände wurden für die folgenden Konten gefunden",
+  "Language": "Deutsch"
```
gdkchan, line 613: Rather minor thing, but I'd prefer if those were the first entry on those files since it's pretty "basic" information about the language file.

gdkchan (approved):
lgtm other than the comment.

Ryujinx.Ava/Assets/Locales/de_DE.json
AcK77: Fixed

PR #4248, opened 2023-01-09
Ava: Add missing null check to ContentDialogHelper.ShowAsync()

TSRBerry:
Another small PR to fix a regression introduced with #4237.

This PR adds a missing null check to `ContentDialogHelper.ShowAsync()`. This is needed since `GetMainWindow()` can also return null, which is the case on first startup.

Example log from someone on Discord: 
[Ryujinx_1.1.526_2023-01-09_23-10-17.log](https://github.com/Ryujinx/Ryujinx/files/10377344/Ryujinx_1.1.526_2023-01-09_23-10-17.log)

### DIFF
--- a/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
+++ b/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
@@ -349,7 +349,7 @@ namespace Ryujinx.Ava.UI.Helpers
 
             Window parent = GetMainWindow();
 
-            if (parent is not null && parent.IsActive && parent is MainWindow window && window.ViewModel.IsGameRunning)
+            if (parent != null && parent.IsActive && parent is MainWindow window && window.ViewModel.IsGameRunning)
             {
                 contentDialogOverlayWindow = new()
                 {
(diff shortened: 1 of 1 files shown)


Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
```diff
@@ -349,7 +349,7 @@ public static async Task<ContentDialogResult> ShowAsync(ContentDialog contentDia
 
             Window parent = GetMainWindow();
 
-            if (parent.IsActive && parent is MainWindow window && window.ViewModel.IsGameRunning)
+            if (parent is not null && parent.IsActive && parent is MainWindow window && window.ViewModel.IsGameRunning)
```
gdkchan, line 352:
```suggestion
            if (parent != null && parent.IsActive && parent is MainWindow window && window.ViewModel.IsGameRunning)
```
Any reason for not doing that

TSRBerry: not really, I'm just used to type it like that. I'll change it!

TSRBerry: Might also be influence from Python :D

gdkchan (approved):
Thanks.

Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
mattzink: FYI: `is not null` generates better IL than `!=`, since the latter has to do a virtual call to `object.Equals()` while the former checks null directly

TSRBerry: That's good to know, thanks!

PR #4249, opened 2023-01-09
Fix linux packaging step for CI & Add headless build support to Ryujinx.sh

TSRBerry:
Minor *emergency* PR, since I broke CI with #4224 and didn't include the headless build in #4013.

This PR fixes the packaging step in our release workflow and also adds `Ryujinx.sh` to the headless build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,22 +75,25 @@ jobs:
       - name: Packing Linux builds
         run: |
           pushd publish_linux
-          tar --exclude "publish/Ryujinx" -cvf ../release_output/ryujinx-${{ steps.version_info.outputs.build_version }}-linux_x64.tar publish
+          tar --exclude "publish/Ryujinx" --exclude "publish/Ryujinx.sh" -cvf ../release_output/ryujinx-${{ steps.version_info.outputs.build_version }}-linux_x64.tar publish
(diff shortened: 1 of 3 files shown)


Ryujinx.Headless.SDL2/Program.cs
```diff
@@ -422,6 +467,12 @@ void LoadPlayerConfiguration(string inputProfileName, string inputId, PlayerInde
                 ));
             }
 
+            // Register mime types on linux.
```
gdkchan, line 470: Why is that done by default? What if the user doesn't want it?

TSRBerry: Hmm that's a good question. Now that I think about it they aren't even useful for the headless build.

But how do we deal with the other GUIs? Do I add a prompt?

marysaka (approved):

Ryujinx.Headless.SDL2/Program.cs
TSRBerry: Want me to do this here or in a separate PR?

gdkchan: Separate PR, it's completely unrelated to this change.

gdkchan (approved):

PR #4250, opened 2023-01-10
GUI: Add option to register file types

TSRBerry:
As discussed here https://github.com/Ryujinx/Ryujinx/pull/4249#discussion_r1065194888, this PR stops Ryujinx from automatically registering file types for the user.

Instead a new menu button was added which allows the user to manually register them.

Note: This is currently missing macOS support.

---

GTK:
image

Ava:
image


Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+    public static class FileAssociationHelper
+    {
+        public static bool IsTypeAssociationSupported => OperatingSystem.IsLinux();
+
+        [SupportedOSPlatform("linux")]
+        private static bool RegisterLinuxMimeTypes()
+        {
+            if (ReleaseInformation.IsFlatHubBuild())
+            {
+                return false;
```
gdkchan, line 19: I guess that will return an error message if one tries to do that on FlatHub builds. Is that fine? It will also return an error message if one attempts to use the feature on Windows or macOS (which will likely lead users to assume it is broken rather than unsupported and file bug reports).

Also if you decide to add an error message about the unsupported platform, it's probably better to do something generic like "This feature is not currently supported on your platform." so that it can be used in other places too, rather than it being something we will add then remove from the locale files later.

TSRBerry: It's probably possible to add this feature for FlatHub builds too, but I don't know how.

The menu button is not visible on unsupported platforms.

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml
```diff
@@ -197,6 +197,10 @@
                     <MenuItem Command="{ReflectionBinding InstallFirmwareFromFile}" Header="{locale:Locale MenuBarFileToolsInstallFirmwareFromFile}" />
                     <MenuItem Command="{ReflectionBinding InstallFirmwareFromFolder}" Header="{locale:Locale MenuBarFileToolsInstallFirmwareFromDirectory}" />
                 </MenuItem>
+                <MenuItem
```
AcK77, line 200: Since it's something you want to set, I think this should be a button in the settings window instead of a new menu item.

TSRBerry: How about making two menu buttons? Since we also want to add the option to remove them again.

As a option in the settings window it would feel a bit out of place to use a checkbox for that. Especially because we usually save settings to the config file.

IsaacMarovitz:
On Avalonia 11, having registered types is required for filtered file picker dialogues to work correctly. Perhaps it would be better to have them registered by default and have the option to unregister them instead as for most users I think this will be perceived as a bug/annoyance rather than a wanted feature.

gdkchan:
> On Avalonia 11, having registered types is required for filtered file picker dialogues to work correctly. Perhaps it would be better to have them registered by default and have the option to unregister them instead as for most users I think this will be perceived as a bug/annoyance rather than a wanted feature.

They are native dialogs, why would Avalonia version change anything on them?

TSRBerry:
Before I'll mark this as ready for review again, I'd like to add an option to unregister file types as well.

Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
```diff
@@ -21,24 +21,26 @@ public static partial class FileAssociationHelper
                                                          !ReleaseInformation.IsFlatHubBuild();
 
         [SupportedOSPlatform("linux")]
-        private static bool RegisterLinuxMimeTypes()
+        private static bool InstallLinuxMimeTypes(bool uninstall=false)
```
gdkchan:
```suggestion
        private static bool InstallLinuxMimeTypes(bool uninstall = false)
```

Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
```diff
@@ -21,24 +21,26 @@ public static partial class FileAssociationHelper
...
         {
             string mimeDbPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share", "mime");
+            string installKeyword = !uninstall ? "install" : "uninstall";
```
gdkchan, line 26:
```suggestion
            string installKeyword = uninstall ? "uninstall" : "install";
```

gdkchan (commented):
A screenshot of the new options on the UI would be nice (and useful for the progress report).

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml
gdkchan: > Since it's something you want to set, I think this should be a button in the settings window instead of a new menu item.

Maybe that would be better. Usually this is done on the installer, however I have seen a few programs with an option to register file extensions as part of the settings. I have never seen it on the file menu.

TSRBerry:
Added the screenshots for GTK and Avalonia.

@AcK77 suggested to change this submenu to a checkbox menu item instead. If I do this I'll update the screenshots again.

gdkchan:
> Added the screenshots for GTK and Avalonia.
> 
> @AcK77 suggested to change this submenu to a checkbox menu item instead. If I do this I'll update the screenshots again.

I think a checkbox is weird for something like this (since you're not just enabling or disabling a feature, it is sort of doing a system wide change). If you're going to keep it a menu I think it's fine how it is.

TSRBerry:
Hmm in that case, I think it's fine the way it is right now! :D

Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
marysaka: menu association is handled by flatpak directly at install.

AcK77 (approved):

Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
```diff
@@ -65,7 +70,7 @@ private static bool InstallLinuxMimeTypes(bool uninstall = false)
         }
 
        [SupportedOSPlatform("windows")]
```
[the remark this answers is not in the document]
TSRBerry: Oh, thank you!

marysaka (approved):

PR #4252, opened 2023-01-10
Implement support for page sizes > 4KB

gdkchan:
This is implementing support for page sizes that are larger than 4KB, which is the case on Apple Silicon and some other platforms. The approach used here is similar to what we used to work around the 64KB granularity limitation on Windows 7/8, where we do page aligned "private" memory allocations, unless the memory region absolutely needs to be shared, such as *shared memory* regions.

For IPC buffers that needs to be shared with other guest processes and are often not aligned, a `MapForeign` method was added which takes a raw pointer, and is only supported by the HLE services memory manager. It is used to map "private" allocations on the service "process". A `GetHostRegions` method was added to get the raw pointers from the guest process for mapping on the service. A `Supports4KBPages` property was added to allow all the above to be disabled if 4KB pages are supported, and then it continues using the old and more correct approach where it actually aliases the memory.

The regular map method now takes a flag that indicates if the memory mapping can be "private" (the `pa` address can be ignored and it can map it anywhere). If the memory can't be private it will try the best it can to align and map it with aliasing anyway. If that is not possible, it will throw an exception. HLE kernel can catch those exceptions and return an error to the guest to force it to map at a different address.

It also includes minor adjustments made by @riperiperi to allow memory tracking to work with page sizes larger than 4KB.


Ryujinx.Memory/MemoryMapFlags.cs
```diff
@@ -0,0 +1,20 @@
+namespace Ryujinx.Memory
+{
+    /// <summary>
+    /// Flags that indicate how the host memory should be mapped.
+    /// </summary>
+    public enum MemoryMapFlags
```
marysaka, line 6:
```suggestion
    [Flags]
    public enum MemoryMapFlags
```

marysaka (approved):

riperiperi (approved):
LGTM, appears to work well.

The texture flush hack isn't great, but the performance suffers a huge amount from (modified) textures sharing pages otherwise.

PR #4255, opened 2023-01-10
Ava: Move Ava logging to Logger.Debug

AcK77:
Since #4231 we currently redirect Avalonia logs to our Logger, which is pretty nice. But since it uses our Logging level too, it now leads to a massive flood in our Log files. To avoid that, I've included all `AvaLogLevel` to the log message, and make all Ava Logs using `Logger.Debug`.


Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -16,56 +16,46 @@ public static void Register()
             AvaLogger.Sink = new LoggerAdapter();
         }
 
-        private static RyuLogger.Log? GetLog(AvaLogLevel level)
```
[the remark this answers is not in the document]
AcK77: This method isn't needed anymore since I redirect everything to `RyuLogger.Debug` I don't see why I should kept it.
I can't just redirect `AvaLogLevel.Information` to `RyuLogger.Debug` because that could leads to misinterpretation. I prefer to add the `AvaLogLevel` to the logged message now.

TSRBerry (commented):
We might want to log errors and/or fatals to the corresponding Ryu log levels, but lgtm otherwise.

Ryujinx.Ava/Helper/LoggerAdapter.cs
```diff
@@ -16,56 +16,46 @@ public static void Register()
...
-        {
-            return level switch
-            {
-                AvaLogLevel.Verbose => RyuLogger.Trace,
-                AvaLogLevel.Debug => RyuLogger.Debug,
-                AvaLogLevel.Information => RyuLogger.Info,
```
IsaacMarovitz, line 25: Would it not be better to just change this line to `RyuLogger.Trace` or `RyuLogger.Debug`?

IsaacMarovitz (changes requested):

Ryujinx.Ava/Helper/LoggerAdapter.cs
gdkchan: I don't think we should add a new log level just for that.

TSRBerry: I think the way you did it is good.
I'd just log errors and fatals to the corresponding Ryu log levels.

AcK77:
In a way to keep some of useful things, I've moved:
- `AvaLogLevel.Verbose`, `AvaLogLevel.Debug`, `AvaLogLevel.Information` and `AvaLogLevel.Warning` to `RyuLogger.Debug`
- `AvaLogLevel.Error` and `AvaLogLevel.Fatal` to `RyuLogger.Error`,

gdkchan (approved):
lgtm, thanks.

Issue #4259, opened 2023-01-10
Arm64: Detect CPU capabilities


gdkchan:
#4114 added JIT Arm64 support, however right now we can't use some instructions directly because we have no way to detect if the CPU actually supports them. Right now, the only fast path that is disabled due to this is the one for the PMULL instruction, but in the future we will also hit that issue for crypto and hash instructions that are optional.

We should implement a function to get CPU capabilities for each platform, and expose properties on the `ARMeilleure.Optimizations` class that will be used by the instruction implementations to check if a given feature is supported.

MutantAura:
Complete?

gdkchan:
Yes.

PR #4264, opened 2023-01-11
Arm64: Cpu feature detection

merryhime:
Proposal for #4259.

To be tested.

### DIFF
--- a/ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
+++ b/ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
@@ -25,17 +25,17 @@ namespace ARMeilleure.CodeGen.Arm64
 
             if (OperatingSystem.IsMacOS())
             {
-                foreach ((string flagName, MacOsFeatureFlags flag) in Enumerable.Zip(Enum.GetNames<MacOsFeatureFlags>(), Enum.GetValues<MacOsFeatureFlags>()))
+                for (int i = 0; i < _sysctlNames.Length; i++)
                 {
-                    string sysctlName = typeof(MacOsFeatureFlags).GetField(flagName)!.GetCustomAttributes(false).OfType<SysctlName>().Single().Name;
-
-                    if (CheckSysctlName(sysctlName))
+                    if (CheckSysctlName(_sysctlNames[i]))
                     {
-                        MacOsFeatureInfo |= flag;
+                        MacOsFeatureInfo |= (MacOsFeatureFlags)(1 << i);
                     }
                 }
             }
         }
+
+#region Linux
 
         private const ulong AT_HWCAP = 16;
         private const ulong AT_HWCAP2 = 26;
@@ -125,6 +125,10 @@ namespace ARMeilleure.CodeGen.Arm64
         public static LinuxFeatureFlagsHwCap LinuxFeatureInfoHwCap { get; } = 0;
         public static LinuxFeatureFlagsHwCap2 LinuxFeatureInfoHwCap2 { get; } = 0;
 
+#endregion
+
+#region macOS
+
         [LibraryImport("libSystem.dylib", SetLastError = true)]
         private static unsafe partial int sysctlbyname([MarshalAs(UnmanagedType.LPStr)] string name, int* oldValue, ref ulong oldSize, IntPtr newValue, ulong newValueSize);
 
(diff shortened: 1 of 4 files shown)


merryhime:
Tested on macOS.

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -0,0 +1,226 @@
...
+            {
+                LinuxFeatureInfoHwCap = (LinuxFeatureFlagsHwCap)getauxval(AT_HWCAP);
+                LinuxFeatureInfoHwCap2 = (LinuxFeatureFlagsHwCap2)getauxval(AT_HWCAP2);
+            }
+
+            if (OperatingSystem.IsMacOS())
+            {
+                foreach ((string flagName, MacOsFeatureFlags flag) in Enumerable.Zip(Enum.GetNames<MacOsFeatureFlags>(), Enum.GetValues<MacOsFeatureFlags>()))
+                {
+                    string sysctlName = typeof(MacOsFeatureFlags).GetField(flagName)!.GetCustomAttributes(false).OfType<SysctlName>().Single().Name;
```
gdkchan, line 30: I would prefer if this did not make use of reflection to enable us to use NativeAOT in the future. For example, those names could be on a string array, and it could do something like:
```cs
for (int i = 0; i < _sysctlNames.Length; i++)
{
    string name = _sysctlNames[i];
    if (CheckSysctlName(sysctlName))
    {
        MacOsFeatureInfo |= (MacOsFeatureFlags)(1 << i);
    }
}
```
Then the `SysctlNameAttribute` and the use of reflection would not be necessary.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
```diff
@@ -2556,7 +2556,7 @@ public static void Pmull_V(ArmEmitterContext context)
         {
             OpCodeSimdReg op = (OpCodeSimdReg)context.CurrOp;
 
-            if (Optimizations.UseAdvSimd && false) // Not supported by all Arm CPUs.
+            if (Optimizations.UseArm64Pmull) // Not supported by all Arm CPUs.
```
gdkchan, line 2559: The comment can be removed, it was added mainly to explain why the fast path was always disabled.

ARMeilleure/Optimizations.cs
```diff
@@ -29,25 +32,26 @@ public static class Optimizations
...
-        internal static bool UseSse42     => UseSse42IfAvailable     && HardwareCapabilities.SupportsSse42;
-        internal static bool UsePopCnt    => UsePopCntIfAvailable    && HardwareCapabilities.SupportsPopcnt;
-        internal static bool UseAvx       => UseAvxIfAvailable       && HardwareCapabilities.SupportsAvx && !ForceLegacySse;
-        internal static bool UseF16c      => UseF16cIfAvailable      && HardwareCapabilities.SupportsF16c;
-        internal static bool UseFma       => UseFmaIfAvailable       && HardwareCapabilities.SupportsFma;
-        internal static bool UseAesni     => UseAesniIfAvailable     && HardwareCapabilities.SupportsAesni;
-        internal static bool UsePclmulqdq => UsePclmulqdqIfAvailable && HardwareCapabilities.SupportsPclmulqdq;
-        internal static bool UseSha       => UseShaIfAvailable       && HardwareCapabilities.SupportsSha;
-        internal static bool UseGfni      => UseGfniIfAvailable      && HardwareCapabilities.SupportsGfni;
+        internal static bool UseAdvSimd    => UseAdvSimdIfAvailable    && AdvSimd.IsSupported;
```
gdkchan, line 39: Can be replace `AdvSimd.IsSupported` with a check for SIMD support using `Arm64HardwareCapabilities` for consistency?

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -146,81 +150,30 @@ private static bool CheckSysctlName(string name)
...
-            Lrcpc2       = 1 << 24,
-            [SysctlName("hw.optional.arm.FEAT_FRINTTS")]
-            Frintts      = 1 << 25,
-            [SysctlName("hw.optional.arm.FEAT_SB")]
-            Sb           = 1 << 26,
-            [SysctlName("hw.optional.arm.FEAT_BF16")]
-            Bf16         = 1 << 27,
-            [SysctlName("hw.optional.arm.FEAT_I8MM")]
-            I8mm         = 1 << 28
+            Fp      = 1 << 0,
```
gdkchan, line 159: Can you keep `Lse`, `Sha1`, `Sha256` and `Crc32`? Those are the ones we will need to check I beleive, since the Switch CPU does not support the other extensions (it also doesn't support `Lse` but we can use it in some cases).

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -0,0 +1,186 @@
...
+            Mte3        = 1 << 22,
+            Sme         = 1 << 23,
+            Sme_i16i64  = 1 << 24,
+            Sme_f64f64  = 1 << 25,
+            Sme_i8i32   = 1 << 26,
+            Sme_f16f32  = 1 << 27,
+            Sme_b16f32  = 1 << 28,
+            Sme_f32f32  = 1 << 29,
+            Sme_fa64    = 1 << 30,
+            Wfxt        = (ulong)1 << 31,
```
gdkchan, line 117:
```suggestion
            Wfxt        = 1UL << 31,
```

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -0,0 +1,186 @@
...
+        [Flags]
+        public enum LinuxFeatureFlagsHwCap : ulong
+        {
+            Fp        = 1 << 0,
+            Asimd     = 1 << 1,
+            Evtstrm   = 1 << 2,
+            Aes       = 1 << 3,
+            Pmull     = 1 << 4,
+            Sha1      = 1 << 5,
+            Sha2      = 1 << 6,
```
gdkchan, line 55: A bit odd that it's not called "Sha256" on the Linux enum. But if that's how they call it it's fine.

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        public static LinuxFeatureFlagsHwCap LinuxFeatureInfoHwCap { get; } = 0;
+        public static LinuxFeatureFlagsHwCap2 LinuxFeatureInfoHwCap2 { get; } = 0;
+
+#endregion
+
+#region macOS
+
+        [LibraryImport("libSystem.dylib", SetLastError = true)]
+        private static unsafe partial int sysctlbyname([MarshalAs(UnmanagedType.LPStr)] string name, ref int oldValue, ref ulong oldSize, IntPtr newValue, ulong newValueSize);
```
gdkchan, line 133: `oldValue` can be `out` rather than `ref` if you want to declare the varialbe on the function call directly like:
```cs
if (sysctlbyname(name, out int val, ref size, IntPtr.Zero, 0) == 0 && size == (ulong)Unsafe.SizeOf<int>())
```

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        [SupportedOSPlatform("macos")]
+        private static bool CheckSysctlName(string name)
+        {
+            int val = 0;
+            ulong size = (ulong)Unsafe.SizeOf<int>();
```
gdkchan, line 139: `sizeof(int)` should be valid too.

gdkchan (approved):
Looks good other than comments, thanks for doing this.

ARMeilleure/CodeGen/Arm64/HardwareCapabilities.cs
merryhime: *nods* ARM is similarly inconsistent with naming (ID_AA64ISAR0_EL1.SHA2 vs FEAT_SHA256).

marysaka (approved):

PR #4266, opened 2023-01-11
Ava UI: Fixes "Hide Cursor on Idle" for Windows

AcK77:
There is currently a known issue where `Hide Cursor on Idle` setting doesn't works on Windows.

It's due to the way we handle the renderer embedded window on Avalonia: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/UI/Helpers/EmbeddedWindow.cs#L136
We create our own W32 window, so even if we want to hide/show the cursor in an Avalonia control, it doesn't work. It seems to work on Linux/macOS because we handle that probably differently.
I fixed it by reworking things a bit and using some `LibraryImport`.

During my test I've discovered another issue, when `Direct Mouse Access` is enabled, it should hide the cursor, it currently doesn't work at all because of the condition used, this compare the wrong object... and so the wrong bounds. I've fixed it, but for an unknown reason the cursor blinking on windows (I think we can't do better for now), it should work as expected on Linux/macOS.

Closes one item in https://github.com/Ryujinx/Ryujinx/issues/3662

It could be nice to get some tests for Windows/Linux/macOS to know if everything works correctly.

### DIFF
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -66,9 +66,8 @@ namespace Ryujinx.Ava
         private const float VolumeDelta        = 0.05f;
 
         private static readonly Cursor InvisibleCursor = new(StandardCursorType.None);
-
-        [SupportedOSPlatform("windows")] private static readonly IntPtr InvisibleCursorWin = CreateEmptyCursor();
-        [SupportedOSPlatform("windows")] private static readonly IntPtr DefaultCursorWin = CreateArrowCursor();
+        private readonly IntPtr InvisibleCursorWin;
+        private readonly IntPtr DefaultCursorWin;
 
         private readonly long      _ticksPerFrame;
         private readonly Stopwatch _chrono;
@@ -163,6 +162,12 @@ namespace Ryujinx.Ava
             ConfigurationState.Instance.HideCursorOnIdle.Event += HideCursorState_Changed;
 
             _topLevel.PointerMoved += TopLevel_PointerMoved;
+
+            if (OperatingSystem.IsWindows())
+            {
+                InvisibleCursorWin = CreateEmptyCursor();
+                DefaultCursorWin   = CreateArrowCursor();
+            }
 
             ConfigurationState.Instance.System.IgnoreMissingServices.Event += UpdateIgnoreMissingServicesState;
             ConfigurationState.Instance.Graphics.AspectRatio.Event         += UpdateAspectRatioState;
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -60,7 +60,7 @@ namespace Ryujinx.Ava
 {
     internal class AppHost
     {
-        private const int   CursorHideIdleTime = 8;    // Hide Cursor seconds.
+        private const int   CursorHideIdleTime = 5;    // Hide Cursor seconds.
         private const float MaxResolutionScale = 4.0f; // Max resolution hotkeys can scale to before wrapping.
         private const int   TargetFps          = 60;
         private const float VolumeDelta        = 0.05f;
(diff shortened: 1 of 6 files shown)


MutantAura:
Works fine on Windows but my Linux install isn't liking it:
```
/home/mutant/Downloads/ava-ryujinx-Release-1.1.0+5014dfa-linux_x64/libuser32.dll: cannot open shared object file: No such file or directory

   at Ryujinx.Ava.UI.Helpers.Win32NativeInterop.<CreateCursor>g____PInvoke|9_0(IntPtr hInst, Int32 xHotSpot, Int32 yHotSpot, Int32 nWidth, Int32 nHeight, Byte* pvANDPlane, Byte* pvXORPlane)
   at Ryujinx.Ava.UI.Helpers.Win32NativeInterop.CreateCursor(IntPtr hInst, Int32 xHotSpot, Int32 yHotSpot, Int32 nWidth, Int32 nHeight, Byte[] pvANDPlane, Byte[] pvXORPlane) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx.Ava/Microsoft.Interop.LibraryImportGenerator/Microsoft.Interop.LibraryImportGenerator/LibraryImports.g.cs:line 133
   at Ryujinx.Ava.UI.Helpers.Win32NativeInterop.CreateEmptyCursor() in /home/runner/work/Ryujinx/Ryujinx/Ryujinx.Ava/UI/Helpers/Win32NativeInterop.cs:line 75
   at Ryujinx.Ava.AppHost..cctor() in /home/runner/work/Ryujinx/Ryujinx/Ryujinx.Ava/AppHost.cs:line 70
   at Ryujinx.Ava.AppHost.<HideCursor>b__75_0() in /home/runner/work/Ryujinx/Ryujinx/Ryujinx.Ava/AppHost.cs:line 211
   at Avalonia.X11.X11PlatformThreading.CheckSignaled() in /_/src/Avalonia.X11/X11PlatformThreading.cs:line 164
   at Avalonia.X11.X11PlatformThreading.RunLoop(CancellationToken cancellationToken) in /_/src/Avalonia.X11/X11PlatformThreading.cs:line 205
   at Avalonia.Threading.Dispatcher.MainLoop(CancellationToken cancellationToken) in /_/src/Avalonia.Base/Threading/Dispatcher.cs:line 65
   at Avalonia.Controls.ApplicationLifetimes.ClassicDesktopStyleApplicationLifetime.Start(String[] args) in /_/src/Avalonia.Controls/ApplicationLifetimes/ClassicDesktopStyleApplicationLifetime.cs:line 120
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime[T](T builder, String[] args, ShutdownMode shutdownMode) in /_/src/Avalonia.Controls/ApplicationLifetimes/ClassicDesktopStyleApplicationLifetime.cs:line 209
   at Ryujinx.Ava.Program.Main(String[] args) in /home/runner/work/Ryujinx/Ryujinx/Ryujinx.Ava/Program.cs:line 95
Aborted (core dumped)
   ```

**Comment by AcK77** [comment]
> Works fine on Windows but my Linux install isn't liking it:

Should be fixed now

**Comment by MutantAura** [comment]
> Should be fixed now

Can confirm! All working as expected :)

**Review by marysaka — APPROVED** [review]

File: Ryujinx.Ava/AppHost.cs [hunk 1]
```diff
@@ -58,12 +60,14 @@ namespace Ryujinx.Ava
 {
     internal class AppHost
     {
-        private const int   CursorHideIdleTime = 8;    // Hide Cursor seconds.
+        private const int   CursorHideIdleTime = 5;    // Hide Cursor seconds.
```
TSRBerry, line 63: If this gets changed don't you need to adjust this for Gtk and Headless as well?

TSRBerry (approved):

PR #4270, opened 2023-01-11
Ava UI: Reorder settings of Resolution Scaler

AcK77:
This PR re-order the settings of the Resolution Scaler in Ava UI to follow what we currently have in GTK.

Before:
image

After:
image

I've sorted the configurations fields by their categories in `LoadCurrentConfiguration` and `SaveSettings` too. Some cleanup have to be done in a near future as those names suggest.

(Closes some items of https://github.com/Ryujinx/Ryujinx/issues/3662)


marysaka (approved):

gdkchan (approved):
If possible, using `Length - 1` instead of `4` is preferable. Then if we ad more resolution scale presets (which will make the custom entry move down), we don't need to change that code.

AcK77:
> If possible, using Length - 1 instead of 4 is preferable. Then if we add more resolution scale presets (which will make the custom entry move down), we don't need to change that code.
Sadly it's not possible since we avoid to access of the view from the ViewModel, The value is a long and not an enum too, so I can't know the Length...

PR #4272, opened 2023-01-11
Ptc: Check process architecture

merryhime:
Add an `Architecture` field to PTC `OuterHeader`.


gdkchan (approved):
Makes sense, I can see people trying to move caches created on x86 machines to Arm ones. Although the difference in the capabilities would probably force it to rebuild anyway.

gdkchan:
I wonder if the values in the `Architecture` enum are guaranteed to remain the same between .NET versions.

marysaka (approved):

PR #4274, opened 2023-01-11
lm: Handle Tail flag in LogPacket

AcK77:
When guest use `lm` service to print some logging, sent packets have a flag field. We currently doesn't take care of the flags, that could lead to strange logging like that:

image

This is now fixed:

image

It's not useful at all, but... Why not 😄 !

### DIFF
--- a/Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
+++ b/Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
@@ -1,6 +1,7 @@
 using Ryujinx.Common.Logging;
 using Ryujinx.Common.Memory;
 using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.LogManager.Types;
 using Ryujinx.Horizon.Sdk.Lm;
 using Ryujinx.Horizon.Sdk.Sf;
 using Ryujinx.Horizon.Sdk.Sf.Hipc;
@@ -16,10 +17,14 @@ namespace Ryujinx.Horizon.LogManager.Ipc
         private readonly LogService _log;
         private readonly ulong      _pid;
 
+        private LogPacket _logPacket;
+
         public LmLogger(LogService log, ulong pid)
         {
             _log = log;
             _pid = pid;
+
+            _logPacket = new LogPacket();
         }
 
         [CmifCommand(0)]
@@ -30,7 +35,12 @@ namespace Ryujinx.Horizon.LogManager.Ipc
                 return Result.Success;
             }
 
-            Logger.Guest?.Print(LogClass.ServiceLm, LogImpl(message));
+            if (LogImpl(message))
+            {
+                Logger.Guest?.Print(LogClass.ServiceLm, _logPacket.ToString());
+
+                _logPacket = new LogPacket();
+            }
 
             return Result.Success;
         }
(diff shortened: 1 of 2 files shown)


Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
```diff
@@ -60,58 +70,87 @@ private static bool SetProcessId(Span<byte> message, ulong processId)
             return true;
         }
 
-        private static string LogImpl(ReadOnlySpan<byte> message)
+        private bool LogImpl(ReadOnlySpan<byte> buffer)
```
gdkchan, line 73: Why was it renamed from message to buffer?

Ryujinx.Horizon/LogManager/Types/LogPacket.cs
```diff
@@ -0,0 +1,72 @@
+﻿using Ryujinx.Horizon.Sdk.Diag;
+using System.Text;
+
+namespace Ryujinx.Horizon.LogManager.Types
+{
+    class LogPacket
```
gdkchan, line 6: Can it be a `struct` rather than a `class`?

Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
```diff
@@ -60,58 +70,87 @@ private static bool SetProcessId(Span<byte> message, ulong processId)
...
-                    fieldStr = $"{field}: {reader.Read<long>()}s";
+                    _logPacket.Time = reader.Read<long>();
                 }
-                else if (field < LogDataChunkKey.Count)
+                else if (key == LogDataChunkKey.Message)
                 {
-                    fieldStr = $"{field}: '{Encoding.UTF8.GetString(reader.GetSpan(size)).TrimEnd()}'";
+                    string message = Encoding.UTF8.GetString(reader.GetSpan(size)).TrimEnd();
+
+                    if (!isTailPacket)
```
gdkchan, line 115: This check seems unnecessary. I think it could always do
```cs
                        if (string.IsNullOrEmpty(_logPacket.Message))
                        {
                            _logPacket.Message = message;
                        }
                        else
                        {
                            _logPacket.Message += message;
                        }
```

Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
```diff
@@ -60,58 +70,87 @@ private static bool SetProcessId(Span<byte> message, ulong processId)
...
+                    {
+                        _logPacket.Message += message;
+                    }
+                    else
+                    {
+                        if (string.IsNullOrEmpty(_logPacket.Message))
```
gdkchan, line 121: Maybe it should do `_logPacket.Message = message;` if the `Head` flag is set too? Since that's supposed to be the start of the message.

gdkchan (commented):
In general I'm just a bit worried that this can cause issues if for some reason an application is not setting the head or tail flags properly. It could lead to messages not being displayed and high memory usage for the messages it is keeping in memory to eventually display.

Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
AcK77: I reverted this change and renamed the reading text: `text`

Ryujinx.Horizon/LogManager/Ipc/LmLogger.cs
AcK77: Check is reworked

AcK77:
> In general I'm just a bit worried that this can cause issues if for some reason an application is not setting the head or tail flags properly. It could lead to messages not being displayed and high memory usage for the messages it is keeping in memory to eventually display.

I've added a Length check and simulate a `isTailFlag` if the message length is more than 5000 chars

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4275, opened 2023-01-12
Ava UI: Add Notifications and Cleanup

AcK77:
This PR bring a new feature "Notifications", we have now the ability to show some in-app notifications:
(Screenshot is here just to show the feature)
image

The notifications as they are, have some limitations:
- They can only contains text.
- They can't be deleted programmatically.
- It can only display 4 notifications at a time (can be changed in code).
- It displays each notification within a 5 second window.

I've modified and cleaned ApplicationHelper to show some example of the usage of this new functionnality:
(This not fixes the issue about the dialog staying open after you extracted a NCA section)
IUmqNUySIA

While I'm working on it, I've found some inconsistencies, like the context menu when you right click on a game, this is now fixed, `MainWindowViewModel` have been cleaned a bit too.

Before:
image
After:
image
image


gdkchan:
I think the error should remain as dialogs, because they require the user exclusive attention. Toast notifications are much easier to miss.

AcK77:
> I think the error should remain as dialogs, because they require the user exclusive attention. Toast notifications are much easier to miss.

Reverted

marysaka (approved):

Ryujinx.Ava/UI/Helpers/NotificationHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+                _templateAppliedEvent.Set();
+            };
+
+            Task.Run(async () =>
+            {
+                _templateAppliedEvent.WaitOne();
+
+                foreach (var notification in _notifications.GetConsumingEnumerable())
+                {
+                    Dispatcher.UIThread.Post(() =>
```
TSRBerry, line 43: Any reason you replaced ever `Dispatcher.UIThread.Post()` with `Dispatcher.UIThread.InvokeAsync()`, but not this one?

Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -733,39 +740,29 @@ public int ListItemSelectorSize
...
+                };
             }
         }
 
         public int GridItemSelectorSize
         {
             get
             {
-                switch (ConfigurationState.Instance.Ui.GridSize)
+                return (int)ConfigurationState.Instance.Ui.GridSize switch
```
TSRBerry, line 758: Is the cast needed here?

TSRBerry (commented):
Lgtm! I just have two questions, before I'll approve.

Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
AcK77: Fixed using `ConfigurationState.Instance.Ui.GridSize.Value` instead. Good catch!

Ryujinx.Ava/UI/Helpers/NotificationHelper.cs
AcK77: Because it dispatch `_notificationManager.Show(notification);` which isn't async. So it's not needed to call it inside an async context!

TSRBerry (approved):
Awesome, lgtm now!

PR #4278, opened 2023-01-12
Prepo: Fix SaveSystemReport* IPC definitions

AcK77:
This PR fixes a little regression I've made in #4220, `SaveSystemReport*` IPCs should use an ApplicationId instead of a pid.
I've added an `ApplicationId` struct to do that more cleaner as just use an `ulong`. 

During testing, an issue of wrong indexes args in HipcGenerator was found and fixed (thanks to @gdkchan )

### DIFF
--- a/Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
+++ b/Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
@@ -42,7 +42,7 @@ namespace Ryujinx.Horizon.Prepo.Ipc
                 return PrepoResult.PermissionDenied;
             }
 
-            ProcessPlayReport(PlayReportKind.Normal, pid, gameRoomBuffer, reportBuffer, Uid.Null);
+            ProcessPlayReport(PlayReportKind.Normal, gameRoomBuffer, reportBuffer, pid, Uid.Null);
 
             return Result.Success;
         }
@@ -57,7 +57,7 @@ namespace Ryujinx.Horizon.Prepo.Ipc
                 return PrepoResult.PermissionDenied;
             }
 
-            ProcessPlayReport(PlayReportKind.Normal, pid, gameRoomBuffer, reportBuffer, userId, true);
+            ProcessPlayReport(PlayReportKind.Normal, gameRoomBuffer, reportBuffer, pid, userId, true);
 
             return Result.Success;
         }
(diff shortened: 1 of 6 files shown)


Ryujinx.Horizon/Sdk/Ncm/ApplicationId.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public static int Length => sizeof(ulong);
+
+        public static ApplicationId First => new(0x0100000000010000);
+
+        public static ApplicationId Last => new(0x01FFFFFFFFFFFFFF);
+
+        public static ApplicationId Invalid => new(0);
+
+        public bool IsInvalid => Id < First.Id && Id > Last.Id;
```
gdkchan, line 15:
```suggestion
        public bool IsInvalid => Id < First.Id || Id > Last.Id;
```

gdkchan: Also having a `IsValid` property seems more in-line with other functions that we have. Generally checking if something is valid seems more natural to me than checking if it's invalid.

Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
```diff
@@ -191,16 +191,23 @@ private static Result ProcessPlayReport(PlayReportKind playReportKind, ulong pid
...
             MessagePackObject deserializedReport = MessagePackSerializer.UnpackMessagePackObject(reportBuffer.ToArray());
 
             builder.AppendLine();
             builder.AppendLine("PlayReport log:");
             builder.AppendLine($" Kind: {playReportKind}");
-            builder.AppendLine($" Pid: {pid}");
+
+            // NOTE: The service calls arp:r using the pid to get the application id, if it fails PrepoResult.InvalidPid is returned.
+            //       Reports are stored internally and an event is signaled to transmit them.
+            if (pid != 0 )
```
gdkchan, line 203:
```suggestion
            if (pid != 0)
```

Ryujinx.Horizon/Sdk/Ncm/ApplicationId.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public static int Length => sizeof(ulong);
+
+        public static ApplicationId First => new(0x0100000000010000);
+
+        public static ApplicationId Last => new(0x01FFFFFFFFFFFFFF);
+
+        public static ApplicationId Invalid => new(0);
+
+        public bool IsValid => Id >= First.Id || Id <= Last.Id;
```
gdkchan, line 15:
```suggestion
        public bool IsValid => Id >= First.Id && Id <= Last.Id;
```

gdkchan (approved):

PR #4282, opened 2023-01-13
Relax Vulkan requirements

gdkchan:
Same as #4228, but now I also fixed `MaxColorAttachmentIndex` which was wrong and was breaking a lot of games. `AttachmentIndices` also had an additional index at the end for depth-stencil which was always 0. Due to that, if a depth-stencil texture was bound, `MaxColorAttachmentIndex` would be always 0. I have removed the useless 0 index at the end which should make the value of `MaxColorAttachmentIndex` correct, as it will be the index of the last color attachment as it should be.

I recommend actually testing it this time.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
@@ -591,7 +591,11 @@ namespace Ryujinx.Graphics.Vulkan
 
             IsAmdWindows = Vendor == Vendor.Amd && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
             IsIntelWindows = Vendor == Vendor.Intel && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-            IsTBDR = IsMoltenVk || Vendor == Vendor.Qualcomm || Vendor == Vendor.ARM || Vendor == Vendor.ImgTec;
+            IsTBDR = IsMoltenVk ||
+                Vendor == Vendor.Qualcomm ||
+                Vendor == Vendor.ARM ||
+                Vendor == Vendor.Broadcom ||
+                Vendor == Vendor.ImgTec;
 
             GpuVendor = vendorName;
             GpuRenderer = Marshal.PtrToStringAnsi((IntPtr)properties.DeviceName);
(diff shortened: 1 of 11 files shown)


Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -590,7 +590,11 @@ private unsafe void PrintGpuInformation()
 
             IsAmdWindows = Vendor == Vendor.Amd && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
             IsIntelWindows = Vendor == Vendor.Intel && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
marysaka, line 592: Random note here: We should be using ``OperatingSystem.IsWindows()``.

marysaka (approved):

riperiperi:
This seems to be breaking multithreaded compilation - the background compiled pipeline is different from the one requested by the first draw. Lots of logs like this:

`00:00:09.033 |W| GUI.RenderLoop Gpu TryGetGraphicsPipeline: Background pipeline compile missed on draw - incorrect pipeline state?`

Happens most notably in clubhouse games. The visuals across games seem fine.

gdkchan:
> This seems to be breaking multithreaded compilation - the background compiled pipeline is different from the one requested by the first draw. Lots of logs like this:
> 
> `00:00:09.033 |W| GUI.RenderLoop Gpu TryGetGraphicsPipeline: Background pipeline compile missed on draw - incorrect pipeline state?`
> 
> Happens most notably in clubhouse games. The visuals across games seem fine.

I fixed the `ColorBlendAttachmentStateCount` on `PipelineConverter` that was using a constant value before. It's the only thing I changed that could affect this I believe. Can you check if the issue is fixed?

I do still get `Background pipeline compile missed on draw - incorrect pipeline state?` warnings on some games still, but I also get them on master for at least a few months now, so it's hard to notice new warnings of this type caused by this change.

riperiperi (approved):
You're right, Pokemon Scarlet/Violet do seem to be triggering this on Master, will need to look into that at some point. The games I was getting this error on with this PR are fixed tho, and I'm not getting any other issues.

TonyMegaPro200:
hi, how could i download Ryujinx for Raspbian (Raspberry Pi 4) please

gdkchan:
We are not releasing any Arm64 Linux build right now, you will have to build it yourself. Also it has been a few months, I don't know what is the current state of Ryujinx on the Raspberry Pi 4.

TonyMegaPro200:
And do you know were can i find information for trying to build it on the raspberry pi?

gdkchan:
It's the same process as building on any other platform. You can just follow the guide on the README.

PR #4284, opened 2023-01-13
Fix texture modified on CPU from GPU thread after being modified on GPU not being updated

gdkchan:
Because the `Modified` flag was not being cleared for those textures, the data written from CPU (on GPU thread) was never uploaded to GPU on the modified texture. The texture would keep stale, incorrect data. The fix here is clearing the modified flag if the data is modified from the CPU on the GPU thread.

This fixes missing textures on EVE ghost enemies, and possibly other games.

Before:
image
After:
image

Note that on Vulkan, the issue only happens with backend multithreading set to On or Auto. On OpenGL it should happen regardless of this setting. That's because the flag was already being cleared if the `Sync` call happens from the GPU thread (which was always the case on Vulkan with backend threading disabled).

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
@@ -1431,10 +1431,10 @@ namespace Ryujinx.Graphics.Gpu.Image
                 return;
             }
 
+            handle.Sync(_context);
+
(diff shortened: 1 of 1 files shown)


marysaka (approved):

riperiperi (approved):
Looks good. I'm not sure why the `Sync` call ended up in there.

PR #4287, opened 2023-01-14
Remove use of reflection on GAL multithreading

gdkchan:
As the title says. Another step towards NativeAOT.

The `IGalCommand<T>` interface would not be necessary if `Run` was a instance method and did not take `ref T` as the first argument. I'm not sure why it's static, but I kept it like that in case there was some reason I was not aware.


riperiperi (commented):
LGTM. I think the use of the generic method should serve the same purpose of avoiding virtual calls, and in testing there doesn't seem to be a perf impact.

riperiperi (approved):

AcK77 (approved):

Issue #4062, opened 2022-12-07
MacOS upstreaming roadmap


riperiperi:
The MacOS build was made using a private tree that we need to upstream in pieces, as there is too much to review all-in-one, and a few things that were a bit messier than they should be on macos.

## GPU/UI

- [x] MacOS UI support (#3980)
- [x] Portability Subset, MVK configuration and properly handling limitations (#4202)
  - Many changes to handle portability subset changes and moltenvk specific behaviours. Configuration is required so that environment variables don't need to be set to use argument buffers.
- [x] 16-bit texture conversion (#4108)
  - Metal on intel macs (and rosetta) does not support 16-bit texture formats. Most have a workaround in the version we released, but there are a few missing still. In A64 mode these formats are supported.
- [x] Attachment format specialization (#4202)
  - Attachment output type (float/int) must match the bound attachments on Metal. Technically this is true of Vulkan too, but drivers are typically able to reinterpret the data and get the proper result. With #4004 this change is less expensive, though it will be weird to add as render target state must be checked after shader loading. 
- [x] Blend State Fix (#5122)
  - There are a few issues with blend state that need sorting.
  - [x] (#4404)
- [x] Precise Occlusion Queries (#4292)
  - This flag needs to be set to guarantee that occlusion query result is always the number of samples passed. On desktop GPUs it tends to be that regardless, but MVK default is 1/0.
- [x] Buffer Mirrors (#4899, WIP)
  - Required for good performance under TBDR renderers. Avoids render pass splitting to update buffers as much as possible by uploading to a temporary buffer and generating mirrors in the staging buffer when bindings are committed. 
- [x] TFB-on-SB (https://github.com/Ryujinx/Ryujinx/pull/5080)
  - TFB emulation via storage writes. Required for games that use transform feedback. (pokemon, xenoblade, etc)
- [x] GS-Layer-To-VS (#4368)
  - Required for UE4 games to render correctly.
- [x] GS-on-VS (#5551)
  - Geometry shader emulation. If official support were to come via MoltenVK, that would be preferable.
- [x] **_MoltenVK Argument Buffers bug workarounds_**
[...]

PR #4292, opened 2023-01-14
Vulkan: Explicitly enable precise occlusion queries

riperiperi:
The only guarantee of the occlusion query type in Vulkan is that it will be zero when no samples pass, and non-zero when any samples pass. Of course, most GPUs implement this by just placing the # of samples in the result and calling it a day. However, this lax restriction means that GPUs could just report a boolean (1/0) or report a value after one is recorded, but before all samples have been counted.

MoltenVK falls in the first category - by default it only reports 1/0 for occlusion queries. Thankfully, there is a feature and flag that you can use to force compatible drivers to provide a "precise" query result, that being the real # of samples passed.

Should fix ink collision in Splatoon 2/3 on MoltenVK. Mentioned on (#4062).


gdkchan (approved):
Looks good.

AcK77 (approved):

PR #4294, opened 2023-01-15
UI: Fix applications times

AcK77:
This PR improve a bit the way applications times are displayed.

Before:
image

After:
image

Since GTK and Ava both uses the same code. I can't use locale for now. In any case things will have to be refactored when GTK will be dropped.
I have cleaned the file a bit too.

### DIFF
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -681,11 +681,25 @@ namespace Ryujinx.Ui.App.Common
         {
             System.TimeSpan time = System.TimeSpan.FromSeconds(seconds);
 
-            string timeString = (time.Days    != 0) ? "{0}d"     : "";
-            timeString       += (time.Hours   != 0) ? " {1:D2}h" : "";
-            timeString       += (time.Minutes != 0) ? " {2:D2}m" : "";
+            string timeString;
+            if (time.Days != 0)
+            {
+                timeString = $"{time.Days}d {time.Hours:D2}h {time.Minutes:D2}m";
+            }
+            else if (time.Hours != 0)
+            {
+                timeString = $"{time.Hours:D2}h {time.Minutes:D2}m";
+            }
+            else if (time.Minutes != 0)
+            {
+                timeString = $"{time.Minutes:D2}m";
+            }
+            else
+            {
+                timeString = "Never";
+            }
 
-            return timeString == "" ? "Never" : string.Format(timeString, time.Days, time.Hours, time.Minutes);
+            return timeString;
         }
 
         private void GetGameInformation(ref ApplicationControlProperty controlData, out string titleName, out string titleId, out string publisher, out string version)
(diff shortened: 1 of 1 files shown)


IsaacMarovitz:
I don't know how I feel about this format. Maybe it would be better to let the user choose between the Steam style format that we have currently or time in just minutes. At the very least I think this format could benefit from some spaces between each number.

TSRBerry:
Could you add some spaces in the format? So it's like "2d 12h 23min"

AcK77:
> I don't know how I feel about this format. Maybe it would be better to let the user choose between the Steam style format that we have currently or time in just minutes.

We aren't currently in an great UI state to add setting for that, maybe later. In master the time just print "minutes", "hours" or "days" without too much precision, which is fixed right now.

> - At the very least I think this format could benefit from some spaces between each number.
> - Could you add some spaces in the format? So it's like "2d 12h 23min"

Fixed

TSRBerry (approved):

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -529,192 +529,163 @@ public byte[] GetApplicationIcon(string applicationPath)
...
-            }
-            else if (seconds < secondsPerDay)
-            {
-                readableString = $"{Math.Round(seconds / secondsPerHour, 1, MidpointRounding.AwayFromZero)} hours";
-            }
-            else
-            {
-                readableString = $"{Math.Round(seconds / secondsPerDay, 1, MidpointRounding.AwayFromZero)} days";
-            }
+            string timeString = (time.Days    != 0) ? "{0}d"     : "";
```
gdkchan, line 684: Why not something like:
```cs
if (time.Days != 0)
{
    timeString = $"{time.Days}d {time.Hours:D2}h {time.Minutes:D2}m";
}
else if (time.Hours != 0)
{
    timeString = $"{time.Hours:D2}h {time.Minutes:D2}m";
}
else if (time.Minutes != 0)
{
    timeString = $"{time.Minutes:D2}m";
}
else
{
    timeString = "Never";
}

gdkchan: BTW this should be localizable.

gdkchan: Also the current code is probably not correct since if hours or minutes are 0, it will just omit then.

marysaka: > BTW this should be localizable.

@AcK77 mentioned to me in DMs before that PR that the common UI code cannot support localization as it's tied to Avalonia frontend.

AcK77: As @marysaka said, it's not possible to get it localizable for now, because GTK depend on this code too.
I've changed the condition by yours, I thought omit hours and minutes when they are 0 could be nice, but it's better to get the full time if days or hours exists.

gdkchan (approved):
lgtm, thanks. In the future it would be nice if you made changes that are not supposed to change functionality but otherwise causes a lot of diffs into separate commits (like the using changes for example). That would make the code easier to review as I could look into the individual commits and skip the trivial changes.

PR #4296, opened 2023-01-15
Move Ryujinx Folder from ~/.config to ~/Library/Application Support on macOS

Candygoblen123:
This change makes Ryujinx more consistent with other apps on macOS. This PR also implements automatic migration of Ryujinx data from the old path to the new one, whenever Ryujinx detects the old data directory exists, but the new one does not.

### DIFF
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -100,7 +100,6 @@ namespace Ryujinx.Common.Configuration
                 {
                     CopyDirectory(oldConfigPath, BaseDirPath);
                     Directory.Delete(oldConfigPath, true);
-                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
                 }
             }
 
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -100,6 +100,7 @@ namespace Ryujinx.Common.Configuration
                 {
                     CopyDirectory(oldConfigPath, BaseDirPath);
                     Directory.Delete(oldConfigPath, true);
+                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
                 }
             }
 
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -45,10 +45,10 @@ namespace Ryujinx.Common.Configuration
 
         public static void Initialize(string baseDirPath)
         {
-            string appDataPath = null;
-            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX))
+            string appDataPath;
+            if (OperatingSystem.IsMacOS())
             {
-                appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "Library/Application Support");
+                appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "Library", "Application Support");
             }
             else
             {
@@ -89,12 +89,20 @@ namespace Ryujinx.Common.Configuration
 
             BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
-            // Copy .config dir to application support dir
-            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX) && Mode == LaunchMode.UserProfile)
+            // NOTE: Copies the Ryujinx folder in `~/.config` to `~/Library/Application Support` if one is found
+            // and a Ryujinx folder does not already exist in Application Support.
+            // Also creates a symlink from `~/.config/Ryujinx` to `~/Library/Application Support/Ryujinx` to preserve backwards compatibility.
+            // This should be removed in the future.
+            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
             {
-                var oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
+                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
                 if (Path.Exists(oldConfigPath) && !Path.Exists(BaseDirPath))
+                {
                     CopyDirectory(oldConfigPath, BaseDirPath);
+                    Directory.Delete(oldConfigPath, true);
+                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
+                }
+
             }
 
             SetupBasePaths();
@@ -113,7 +121,9 @@ namespace Ryujinx.Common.Configuration
             var dir = new DirectoryInfo(sourceDir);
 
             if (!dir.Exists)
+            {
                 throw new DirectoryNotFoundException($"Source directory not found: {dir.FullName}");
+            }
 
             DirectoryInfo[] subDirs = dir.GetDirectories();
             Directory.CreateDirectory(destinationDir);
@@ -121,15 +131,15 @@ namespace Ryujinx.Common.Configuration
             foreach (FileInfo file in dir.GetFiles())
             {
                 if (file.Name == ".DS_Store")
+                {
                     continue;
-                string destFile = Path.Combine(destinationDir, file.Name);
-                file.CopyTo(destFile);
+                }
+                file.CopyTo(Path.Combine(destinationDir, file.Name));
             }
 
             foreach (DirectoryInfo subDir in subDirs)
             {
-                string destDir = Path.Combine(destinationDir, subDir.Name);
-                CopyDirectory(subDir.FullName, destDir);
+                CopyDirectory(subDir.FullName, Path.Combine(destinationDir, subDir.Name));
             }
         }
 
--- a/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -45,7 +45,15 @@ namespace Ryujinx.Common.Configuration
 
         public static void Initialize(string baseDirPath)
         {
-            string appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            string appDataPath = null;
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX))
+            {
+                appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "Library/Application Support");
+            }
+            else
+            {
+                appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            }
 
             if (appDataPath.Length == 0)
             {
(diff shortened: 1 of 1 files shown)


merryhime:
It would be helpful to somehow mark the old directory as having been copied from, otherwise intentionally deleted files will be magically recreated from older copies (e.g. by renaming the directory to `Ryujinx.old` or similar).

Candygoblen123:
Right now it checks if the path `~/Library/Application Support/Ryujinx` exists, and if it does exist it does nothing. That should be sufficient in preventing intentionally deleted files from being magically recreated. However I can rename the directory as well (or it might be better to create a hidden file, called `.migrated` or something, that marks if the dir has been migrated, to ensure that older versions can still find the directory). Or alternatively, it might be a good idea to just create a symlink from `~/.config/Ryujinx` to `~/Library/Application Support/Ryujinx`.

merryhime:
I feel a symlink may be the most elegant solution here, given that one may wish to bisect on Ryujinx versions prior to migration.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -45,7 +45,15 @@ static AppDataManager()
 
         public static void Initialize(string baseDirPath)
         {
-            string appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
+            string appDataPath = null;
```
AcK77, line 48:
```suggestion
            string appDataPath;
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -45,7 +45,15 @@ static AppDataManager()
...
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX))
```
AcK77, line 49:
```suggestion
            if (OperatingSystem.IsMacOS())
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -45,7 +45,15 @@ static AppDataManager()
...
+            {
+                appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "Library/Application Support");
```
AcK77, line 51:
```suggestion
                appDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), "Library", "Application Support");
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,14 @@ public static void Initialize(string baseDirPath)
 
             BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
+            // Copy .config dir to application support dir
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX) && Mode == LaunchMode.UserProfile)
```
AcK77, line 93:
```suggestion
            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,14 @@ public static void Initialize(string baseDirPath)
 
             BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
+            // Copy .config dir to application support dir
```
AcK77, line 92: Start your comment with `// NOTE: ` and add a more explicit comment, don't forget to tell it should be removed after some point. There is no need to keep the migration in the future when most of users have already done it.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,14 @@ public static void Initialize(string baseDirPath)
...
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.OSX) && Mode == LaunchMode.UserProfile)
+            {
+                var oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
```
AcK77, line 95:
```suggestion
                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,14 @@ public static void Initialize(string baseDirPath)
...
+                if (Path.Exists(oldConfigPath) && !Path.Exists(BaseDirPath))
+                    CopyDirectory(oldConfigPath, BaseDirPath);
```
AcK77, line 97:
```suggestion
                if (Path.Exists(oldConfigPath) && !Path.Exists(BaseDirPath))
                {
                    CopyDirectory(oldConfigPath, BaseDirPath);
                }
```
You didn't delete or rename `oldConfigPath` which could be nice too.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -92,6 +108,31 @@ private static void SetupBasePaths()
             Directory.CreateDirectory(KeysDirPath = Path.Combine(BaseDirPath, KeysDir));
         }
 
+        private static void CopyDirectory(string sourceDir, string destinationDir)
+        {
+            var dir = new DirectoryInfo(sourceDir);
+
+            if (!dir.Exists)
+                throw new DirectoryNotFoundException($"Source directory not found: {dir.FullName}");
```
AcK77, line 116:
```suggestion
            if (!dir.Exists)
            {
                throw new DirectoryNotFoundException($"Source directory not found: {dir.FullName}");
            }
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -92,6 +108,31 @@ private static void SetupBasePaths()
...
+
+            DirectoryInfo[] subDirs = dir.GetDirectories();
+            Directory.CreateDirectory(destinationDir);
+
+            foreach (FileInfo file in dir.GetFiles())
+            {
+                if (file.Name == ".DS_Store")
+                    continue;
```
AcK77, line 124:
```suggestion
                if (file.Name == ".DS_Store")
                {
                    continue;
                }
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -92,6 +108,31 @@ private static void SetupBasePaths()
...
+                string destFile = Path.Combine(destinationDir, file.Name);
+                file.CopyTo(destFile);
```
AcK77, line 126:
```suggestion

                file.CopyTo(Path.Combine(destinationDir, file.Name));
```

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -92,6 +108,31 @@ private static void SetupBasePaths()
...
+            }
+
+            foreach (DirectoryInfo subDir in subDirs)
+            {
+                string destDir = Path.Combine(destinationDir, subDir.Name);
+                CopyDirectory(subDir.FullName, destDir);
```
AcK77, line 132:
```suggestion
                CopyDirectory(subDir.FullName, Path.Combine(destinationDir, subDir.Name));
```

AcK77 (changes requested):

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,22 @@ public static void Initialize(string baseDirPath)
 
             BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
+            // NOTE: Copies the Ryujinx folder in `~/.config` to `~/Library/Application Support` if one is found
```
UltraHDR, line 92:
```suggestion
            // NOTE: Moves the Ryujinx folder in `~/.config` to `~/Library/Application Support` if one is found
```

UltraHDR:
> I feel a symlink may be the most elegant solution here, given that one may wish to bisect on Ryujinx versions prior to migration.

I'm curious if a bisect would be helpful as macOS support is very basic/not working?

I would prefer if an app didn't create symlinks

but you know best

Just my 2 cents

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,22 @@ public static void Initialize(string baseDirPath)
...
+            // Also creates a symlink from `~/.config/Ryujinx` to `~/Library/Application Support/Ryujinx` to preserve backwards compatibility.
+            // This should be removed in the future.
+            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
+            {
+                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
+                if (Path.Exists(oldConfigPath) && !Path.Exists(BaseDirPath))
+                {
+                    CopyDirectory(oldConfigPath, BaseDirPath);
+                    Directory.Delete(oldConfigPath, true);
+                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
```
AcK77, line 103: I'm not sure if created a symlink is a good thing to do here.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,22 @@ public static void Initialize(string baseDirPath)
...
+                }
+
```
AcK77, line 105: Remove extra empty line

Ryujinx.Common/Configuration/AppDataManager.cs
Candygoblen123: As per the previous comment, the symlink was created to maintain backwards compatibility. If you don't care about it, I can remove it.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -92,6 +114,33 @@ private static void SetupBasePaths()
...
+            DirectoryInfo[] subDirs = dir.GetDirectories();
+            Directory.CreateDirectory(destinationDir);
+
+            foreach (FileInfo file in dir.GetFiles())
+            {
+                if (file.Name == ".DS_Store")
+                {
+                    continue;
+                }
+                file.CopyTo(Path.Combine(destinationDir, file.Name));
```
AcK77, line 135:
```suggestion

                file.CopyTo(Path.Combine(destinationDir, file.Name));
```

AcK77 (approved):
LGTM! Thanks!

seivan:
If I could add one tweak here, is to allow an ENV variable to override it. I like my settings in dotfiles, but preferably not at `~/`. Don't mind adding the change myself, if it's ok.

Candygoblen123:
> If I could add one tweak here, is to allow an ENV variable to override it. I like my settings in dotfiles, but preferably not at `~/`. Don't mind adding the change myself, if it's ok.

You can already do this using the `-r` command line argument.

Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -81,6 +89,20 @@ public static void Initialize(string baseDirPath)
...
+            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
+            {
+                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
+                if (Path.Exists(oldConfigPath) && !Path.Exists(BaseDirPath))
+                {
+                    CopyDirectory(oldConfigPath, BaseDirPath);
+                    Directory.Delete(oldConfigPath, true);
+                }
+            }
+
```
marysaka, line 105: The symlink creation should be added back tbh (discussed that with @AcK77 as it didn't understood the idea around it initially)

marysaka (changes requested):
Only one thing otherwise lgtm

Ryujinx.Common/Configuration/AppDataManager.cs
AcK77: Sorry I’ve misunderstood how things works here, it should be better to keep the symlink yeah :/

Ryujinx.Common/Configuration/AppDataManager.cs
AcK77: This is still not addressed

marysaka (approved):

seivan:
> > If I could add one tweak here, is to allow an ENV variable to override it. I like my settings in dotfiles, but preferably not at `~/`. Don't mind adding the change myself, if it's ok.
> 
> You can already do this using the `-r` command line argument.

That's not the same thing. Reading the ENV is by default and unobtrusive. Asking to pass a command line argument each time means you have to run said binary with the argument.

Candygoblen123:
As a reminder, user documentation should be updated to reflect this change before the next macOS update is released.

PR #4297, opened 2023-01-15
Ava UI: `Renderer` refactoring

AcK77:
This PR refactor and cleanup everything related with Renderer hosting, I've renamed some classes for consistencies and moved everything in a new `Renderer` folder.

There are still `Win32NativeInterop` and `MetalHelper` to clean but that could be done later.


marysaka (approved):

Ryujinx.Ava/AppHost.cs
```diff
@@ -430,11 +429,11 @@ public void DisposeGpu()
                 _windowsMultimediaTimerResolution = null;
             }
 
-            Renderer?.MakeCurrent();
+            (_rendererHost.EmbeddedWindow as EmbeddedWindowOpenGL)?.MakeCurrent();
 
             Device.DisposeGpu();
 
-            Renderer?.MakeCurrent(null);
+            (_rendererHost.EmbeddedWindow as EmbeddedWindowOpenGL)?.MakeCurrent(null);
```
TSRBerry, line 436: Why is this specific to OpenGL now?

Ryujinx.Ava/AppHost.cs
```diff
@@ -853,7 +848,7 @@ public void UpdateStatus()
             StatusUpdatedEvent?.Invoke(this, new StatusUpdatedEventArgs(
                 Device.EnableDeviceVsync,
                 LocaleManager.Instance[LocaleKeys.VolumeShort] + $": {(int)(Device.GetVolume() * 100)}%",
-                Renderer.IsVulkan ? "Vulkan" : "OpenGL",
+                ConfigurationState.Instance.Graphics.GraphicsBackend.Value == GraphicsBackend.Vulkan ? "Vulkan" : "OpenGL",
```
TSRBerry, line 851:
```suggestion
                ConfigurationState.Instance.Graphics.GraphicsBackend.Value,
```

Ryujinx.Ava/UI/Renderer/EmbeddedWindowOpenGL.cs
```diff
@@ -48,19 +45,20 @@ public override void OnWindowCreated()
...
 
             var flags = OpenGLContextFlags.Compat;
-            if (_graphicsDebugLevel != GraphicsDebugLevel.None)
+            if (ConfigurationState.Instance.Logger.GraphicsDebugLevel != GraphicsDebugLevel.None)
             {
                 flags |= OpenGLContextFlags.Debug;
             }
 
-            Context = PlatformHelper.CreateOpenGLContext(FramebufferFormat.Default, _major, _minor, flags);
+            var graphicsMode = Environment.OSVersion.Platform == PlatformID.Unix ? new FramebufferFormat(new ColorFormat(8, 8, 8, 0), 16, 0, ColorFormat.Zero, 0, 2, false) : FramebufferFormat.Default;
```
TSRBerry, line 53: Why was the framebuffer format changed?

Ryujinx.Ava/UI/Renderer/RendererHost.axaml.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (EmbeddedWindow != null)
+            {
+                EmbeddedWindow.WindowCreated -= CurrentWindow_WindowCreated;
+                EmbeddedWindow.SizeChanged   -= CurrentWindow_SizeChanged;
+            }
+        }
+
+        public SurfaceKHR CreateVulkanSurface(Instance instance, Vk api)
+        {
+            return (EmbeddedWindow is EmbeddedWindowVulkan vulkanEmbeddedWindow) ? vulkanEmbeddedWindow.CreateSurface(instance) : default;
```
TSRBerry, line 54: Might want to throw an exception here if the window is not EmbeddedWindowVulkan.

Ryujinx.Ava/AppHost.cs
```diff
@@ -837,7 +832,7 @@ private unsafe void RenderLoop()
                 }
             });
 
-            Renderer?.MakeCurrent(null);
+            (_rendererHost.EmbeddedWindow as EmbeddedWindowOpenGL)?.MakeCurrent(null);
```
TSRBerry, line 835: Should this be specific to OpenGL now?

TSRBerry (changes requested):
A few questions and one suggestion, but otherwise fine!

Ryujinx.Ava/AppHost.cs
AcK77: Switching context is OpenGL only because of the background context, the check was done in Renderer class before, which make no sense since it’s an OpenGL feature.

Ryujinx.Ava/AppHost.cs
AcK77: In case of OpenGL, it will print `OpenGl` In the status bar, if I rename the enum value, it will invalid the configuration. That’s why I’ve just replaced the variable.

Ryujinx.Ava/UI/Renderer/EmbeddedWindowOpenGL.cs
AcK77: The same thing is used on GTK and it seems not be ported, so I added it back just in case.

Ryujinx.Ava/AppHost.cs
AcK77: Same as the other question about `MakeCurrent`

Ryujinx.Ava/UI/Renderer/RendererHost.axaml.cs
AcK77: Fixed

Ryujinx.Ava/AppHost.cs
TSRBerry: Are we not storing 0 or 1, the enum value, in the config file?

TSRBerry (approved):

Ryujinx.Ava/AppHost.cs
AcK77: `"graphics_backend": "Vulkan"` sadly no :(

PR #4299, opened 2023-01-15
ConfigurationState: Default to Vulkan on macOS

merryhime:
Starting on macOS and having it immediately crash when loading something in is not a great first-experience.

Set graphics backend to Vulkan on macOS in default settings.

(Let me know if you want something more substantial than this, or if this is something you're alreadly planning, and I'll close this.)


merryhime:
Should Vulkan just become default?

IsaacMarovitz:
I initially started this in #3968 and I think it's worth adding a bit more sanitising like I started doing. IMO we should have a general way of checking is the graphics backend is supported on that given device, same with OpenGL.

gdkchan:
Yea, the ideal solution here is doing something similar to what we do with audio backend, where if a backend is not supported it switches to another one automatically. And if none of them is supported it switches to a "dummy" one. We currently don't have a dummy graphics backend (maybe we should? but that's a separate issue).

gdkchan (approved):
I think it's fine to accept something like that for now, given that the ideal solution is much more complicated, and that is making the emulator crash on macOS with the default configuration.

PR #4301, opened 2023-01-16
HOS: Load RomFs by pid

AcK77:
We currently loading only one RomFs at a time, which could be wrong if one day we want to load more than one guest at time. This PR fixes that by loading romfs by pid.


gdkchan (approved):
This seems reasonable for now, but would be nice to know how the RomFS registration is supposed to work on the Switch and implement it more like that.

TSRBerry (approved):

PR #4305, opened 2023-01-16
Ava UI: Fix `string.Format` issues in Locale

AcK77:
We currently use a bunch of `string.Format` when locale uses dynamic fields (`{0}`). We already have a method for that in the LocaleManager, so I changed all call to `string.Format` to this method.

This is needed because the `string.Format` could failed because of an updated locale in `en_US` which don't have the same args as the previous one.

But it could crash in the `LocaleManager` instead since our current implementation doesn't handle that too. I have rewrited it to support that and returns the default locale instead, or the locale key name.

If `string.Format` failed, then the UI does nothing, so we currently can't open the Title Update Manager in other languages than `en_US`. It's now fixed.

During my change I've found `TitleUpdate` using a locale named `DialogDlcLoadNcaErrorMessage`, I renamed it `DialogLoadNcaErrorMessage` to be more generic.

### DIFF
--- a/Ryujinx.Ava/Common/Locale/LocaleManager.cs
+++ b/Ryujinx.Ava/Common/Locale/LocaleManager.cs
@@ -59,7 +59,7 @@ namespace Ryujinx.Ava.Common.Locale
                 // Check if the locale contains the key.
                 if (_localeStrings.TryGetValue(key, out string value))
                 {
-                    // Check if the locale needs to be formatted.
+                    // Check if the localized string needs to be formatted.
                     if (_dynamicValues.TryGetValue(key, out var dynamicValue))
                     {
                         try
@@ -68,7 +68,7 @@ namespace Ryujinx.Ava.Common.Locale
                         }
                         catch (Exception)
                         {
-                            // If formatting failed, then use the default locale text.
+                            // If formatting failed use the default text instead.
                             if (_localeDefaultStrings.TryGetValue(key, out value))
                             {
                                 try
@@ -77,7 +77,7 @@ namespace Ryujinx.Ava.Common.Locale
                                 }
                                 catch (Exception)
                                 {
-                                    // If formatting locale default text failed, then returns the key.
+                                    // If formatting the default text failed return the key.
                                     return key.ToString();
                                 }
                             }
@@ -87,13 +87,13 @@ namespace Ryujinx.Ava.Common.Locale
                     return value;
                 }
 
-                // If the locale doesn't contain the key, then returns the default one.
+                // If the locale doesn't contain the key return the default one.
                 if (_localeDefaultStrings.TryGetValue(key, out string defaultValue))
                 {
                     return defaultValue;
                 }
 
-                // If the locale text doesn't exist, then returns the key.
+                // If the locale text doesn't exist return the key.
                 return key.ToString();
             }
             set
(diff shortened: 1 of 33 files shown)


Ryujinx.Ava/Common/Locale/LocaleManager.cs
```diff
@@ -13,56 +13,87 @@ class LocaleManager : BaseModel
...
                     if (_dynamicValues.TryGetValue(key, out var dynamicValue))
                     {
-                        return string.Format(value, dynamicValue);
+                        try
+                        {
+                            return string.Format(value, dynamicValue);
+                        }
+                        catch (Exception)
+                        {
+                            // If formatting failed, then use the default locale text.
```
TSRBerry, line 71:
```suggestion
                            // If formatting failed use the default text instead.
```

Ryujinx.Ava/Common/Locale/LocaleManager.cs
```diff
@@ -13,56 +13,87 @@ class LocaleManager : BaseModel
...
+                                    return key.ToString();
+                                }
+                            }
+                        }
                     }
 
                     return value;
                 }
 
+                // If the locale doesn't contain the key, then returns the default one.
```
TSRBerry, line 90:
```suggestion
                // If the locale doesn't contain the key return the default one.
```

Ryujinx.Ava/Common/Locale/LocaleManager.cs
```diff
@@ -13,56 +13,87 @@ class LocaleManager : BaseModel
...
+                if (_localeDefaultStrings.TryGetValue(key, out string defaultValue))
+                {
+                    return defaultValue;
+                }
+
+                // If the locale text doesn't exist, then returns the key.
```
TSRBerry, line 96:
```suggestion
                // If the locale text doesn't exist return the key.
```

TSRBerry (commented):
Couldn't you return the localized string from `UpdateDynamicValue()`? So you don't need to add another lookup after formatting the string.

marysaka:
> Couldn't you return the localized string from `UpdateDynamicValue()`? So you don't need to add another lookup after formatting the string.

I think that would break Avalonia preview of views if we do that.

marysaka (approved):
In the future it would be nice to look into trying to enforce this kind of fmt at compile time so we could have type checking in place.

gdkchan:
> Couldn't you return the localized string from `UpdateDynamicValue()`? So you don't need to add another lookup after formatting the string.

Yes, I think it would be better if it was done this way. And also rename `UpdateDynamicValue` to something like `GetDynamicValue` if you do that.

AcK77:
> > Couldn't you return the localized string from `UpdateDynamicValue()`? So you don't need to add another lookup after formatting the string.
> 
> Yes, I think it would be better if it was done this way. And also rename `UpdateDynamicValue` to something like `GetDynamicValue` if you do that.

This cannot be done that way without larger changes, because we use LocaleKey in the AXAML using LocaleExtension, and the dynamic locales value could be changed anywhere else in the code, as example: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs#L1645

TSRBerry:
Isn't gdkchan only talking about returning from `UpdateDynamicValue()` here?
That should be possible to address, right?

TSRBerry (approved):

PR #4306, opened 2023-01-17
NativeSignalHandler: Fix write flag

merryhime:
AArch64 linux requires testing.

Offsets reference: https://gist.github.com/merryhime/bb1b41f0c8c55e413e36b7ec55e389a8

Refer to the following documentation regarding AArch64 Linux to lookup ESR: https://github.com/torvalds/linux/blob/6e50979a9c87371fdb85d16058f9b5cb40751501/arch/arm64/include/uapi/asm/sigcontext.h#L39

### DIFF
--- a/ARMeilleure/Signal/NativeSignalHandler.cs
+++ b/ARMeilleure/Signal/NativeSignalHandler.cs
@@ -258,24 +258,23 @@ namespace ARMeilleure.Signal
                 const ulong mcontextOffset = 48; // uc_mcontext
                 Operand ctxPtr = context.Load(OperandType.I64, context.Add(ucontextPtr, Const(mcontextOffset)));
 
-                if (RuntimeInformation.OSArchitecture == Architecture.Arm64)
+                if (RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
                 {
                     const ulong esrOffset = 8; // __es.__esr
                     Operand esr = context.Load(OperandType.I64, context.Add(ctxPtr, Const(esrOffset)));
                     return context.BitwiseAnd(esr, Const(0x40ul));
                 }
 
-                if (RuntimeInformation.OSArchitecture == Architecture.X64)
+                if (RuntimeInformation.ProcessArchitecture == Architecture.X64)
                 {
                     const ulong errOffset = 4; // __es.__err
                     Operand err = context.Load(OperandType.I64, context.Add(ctxPtr, Const(errOffset)));
                     return context.BitwiseAnd(err, Const(2ul));
                 }
             }
-
-            if (OperatingSystem.IsLinux())
+            else if (OperatingSystem.IsLinux())
             {
-                if (RuntimeInformation.OSArchitecture == Architecture.Arm64)
+                if (RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
                 {
                     Operand auxPtr = context.AllocateLocal(OperandType.I64);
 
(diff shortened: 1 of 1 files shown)


ARMeilleure/Signal/NativeSignalHandler.cs
```diff
@@ -251,18 +245,89 @@ private static Operand EmitGenericRegionCheck(EmitterContext context, IntPtr sig
...
+                {
+                    const ulong errOffset = 4; // __es.__err
+                    Operand err = context.Load(OperandType.I64, context.Add(ctxPtr, Const(errOffset)));
+                    return context.BitwiseAnd(err, Const(2ul));
+                }
+            }
+
+            if (OperatingSystem.IsLinux())
+            {
+                if (RuntimeInformation.OSArchitecture == Architecture.Arm64)
```
gdkchan, line 278: We're currently using `RuntimeInformation.ProcessArchitecture` on ARMeilleure, maybe this should use it too? Not sure when `OSArchitecture` and `ProcessArchitecture` would be different (maybe `ProcessArchitecture` can be X86 when running on 32-bit mode on a "X64" CPU?)

ARMeilleure/Signal/NativeSignalHandler.cs
```diff
@@ -251,18 +245,89 @@ private static Operand EmitGenericRegionCheck(EmitterContext context, IntPtr sig
...
+
+                if (RuntimeInformation.OSArchitecture == Architecture.X64)
+                {
+                    const int errOffset = 192; // uc_mcontext.gregs[REG_ERR]
+                    Operand err = context.Load(OperandType.I64, context.Add(ucontextPtr, Const(errOffset)));
+                    return context.BitwiseAnd(err, Const(2ul));
+                }
+            }
+
+            throw new NotImplementedException();
```
gdkchan, line 318: Maybe it should throw `PlatformNotSupportedException`?

ARMeilleure/Signal/NativeSignalHandler.cs
```diff
@@ -251,18 +245,89 @@ private static Operand EmitGenericRegionCheck(EmitterContext context, IntPtr sig
...
+
+                if (RuntimeInformation.OSArchitecture == Architecture.X64)
+                {
+                    const ulong errOffset = 4; // __es.__err
+                    Operand err = context.Load(OperandType.I64, context.Add(ctxPtr, Const(errOffset)));
+                    return context.BitwiseAnd(err, Const(2ul));
+                }
+            }
+
+            if (OperatingSystem.IsLinux())
```
gdkchan, line 276: Could use `else if` since both can't be true at the same time.

ARMeilleure/Signal/NativeSignalHandler.cs
merryhime: Ah, you're correct, it should be `ProcessArchitecture`, not sure what I was thinking when I wrote that.

riperiperi (approved):
Looks good. This allows Mario Kart 8 Deluxe to use res scaling on ARM64 MacOS, as it avoids write flags being triggered on output textures as the access is a read. Unfortunately, that means it's still not possible to remove the workaround that disables flush of block linear textures.

Hopefully this should also help some weird performance issues on Linux too.

gdkchan (approved):
Code lgtm, might benefit from some testing on Linux.

TSRBerry:
Don't have any arm64 hardware with me, so I only tested on my x86_64 laptop.
For my tests I also needed to rebase this branch, since it still had the AddressSpace NRE, so I wasn't able to stop emulation at first.

I played one MK8 cup without issues and was able to play ACNH for about 10 minutes without any issues as well, so everything seems to be working fine!

<details>
  <summary>System information</summary>
  
  ```
  OS: Arch Linux
  Kernel: x86_64 Linux 5.15.88-2-lts
  DE: KDE 5.102.0 / Plasma 5.26.5
  WM: KWin
  CPU: AMD Ryzen 5 4500U with Radeon Graphics @ 6x 2.375GHz
  GPU: AMD Radeon Graphics (renoir, LLVM 15.0.7, DRM 3.42, 5.15.88-2-lts)
  RAM: 2835MiB / 7311MiB
  ```
</details>

merryhime:
❤️

PR #4309, opened 2023-01-18
Handle parsing of corrupt Config.json and prevent crash on launch

gnisman:
For some reason if Config.json is corrupted and contains only { }
JsonSerializer.Deserialize is not validating json against a schema.
ConfigurationFileFormat.TryLoad will return true but ConfigurationState.Instance.Load will throw an exception

It will prevent crash on launch of Ryujinx and improve the user experience

### DIFF
--- a/Ryujinx.Ava/Program.cs
+++ b/Ryujinx.Ava/Program.cs
@@ -205,7 +205,17 @@ namespace Ryujinx.Ava
             {
                 if (ConfigurationFileFormat.TryLoad(ConfigurationPath, out ConfigurationFileFormat configurationFileFormat))
                 {
-                    ConfigurationState.Instance.Load(configurationFileFormat, ConfigurationPath);
+                    try
+                    {
+                        ConfigurationState.Instance.Load(configurationFileFormat, ConfigurationPath);
+                    }
+                    catch
+                    {
+                        ConfigurationState.Instance.LoadDefault();
+                        ConfigurationState.Instance.ToFileFormat().SaveConfig(ConfigurationPath);
+                        
+                        Logger.Warning?.PrintMsg(LogClass.Application, $"Failed to load config! Loading the default config instead.");
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


TSRBerry:
I think a better fix for this would be to change this

https://github.com/Ryujinx/Ryujinx/blob/410be95ab68ff1246eb24d6157c18d4c8c2dcf5a/Ryujinx.Ui.Common/Configuration/ConfigurationFileFormat.cs#L343

to

```cs
return configurationFileFormat.Version != 0;
```

instead.

gnisman:
@TSRBerry  it wont work
If for example your json looks like this `{
  "version": 42
}`

configurationFileFormat.Version != 0 will return true but both the json file and ConfigurationFileFormat are corrupted
and it will crash the tool because of null ref in some of the fields

TSRBerry:
Why would it look like this?

It's either empty, doesn't exist or is stored correctly.

I was thinking about this before I commented my suggestion and found two ways to address this issue in a "clean" way:

1. The way above, since the file shouldn't contain only some data. It either contains everything, nothing or an empty object (afaik).

2. We could initialize https://github.com/Ryujinx/Ryujinx/blob/410be95ab68ff1246eb24d6157c18d4c8c2dcf5a/Ryujinx.Ui.Common/Configuration/ConfigurationFileFormat.cs#L102 with an empty array which would also fix the crash, but wouldn't load default settings. We could address that by adding the defaults to `ConfigurationFileFormat` instead, so the defaults would already be present in case the json file is corrupted and only loaded partially.

I suggested the first way, because I think it's the easiest solution to this problem and shouldn't have any issues. That said, both ways work and since these are just suggestions, so you are free to do add  either one of them or just ignore them completely.

I should have explained my reasoning in the first comment, sorry about that.

gnisman:
Yes you are right the file either contains everything, nothing or an empty object
so we only need to handle an empty json object {}
in that case `return configurationFileFormat.Version != 0;` would be the best solution.
Will fix it now, Thank you

TSRBerry (approved):
Thank you for fixing this and bearing with me! :D

marysaka (approved):

PR #4314, opened 2023-01-19
Handle parsing of corrupt Profiles.json and prevent crash on launch

gnisman:
In case of corrupted json JsonHelper.DeserializeFromFile<ProfilesJson>(_profilesJsonPath) throws an exception.
Solution is to catch the parsing error and use default profile.
It will prevent crash on launch of Ryujinx and improve the user experience


gnisman:
ss_1

gnisman:
On real HW the saves are coupled to the user profile 
not sure what will happen to the old saves if Profiles.json is corrupt

PhiZero:
I already addressed this in #3393. I kinda just totally forgot about it though, sorry about that.

TSRBerry:
Closing as duplicate of #3393.

gnisman:
@PhiZero sorry didn't know you have a fix for this problem

PR #4269, opened 2023-01-11
Ava UI: Fixes and cleanup Updater

AcK77:
Since https://github.com/Ryujinx/Ryujinx/pull/4178 is merged, users reported the Updater behavior is a bit strange:

- If you deny the update on startup, then the MenuItem is disabled.
- Running a game and stop it makes the MenuItem enabled back, but if you click, nothing appear.

(Preview of them can be shown at https://github.com/Ryujinx/Ryujinx/issues/3662#issuecomment-1375512394)

This PR should fixes that, the logic using a bool was broken. I've cleaned up some parts of the updater too but not too much since it's not the point here.

(Closes some items of https://github.com/Ryujinx/Ryujinx/issues/3662)

PR #4315, opened 2023-01-19
GUI: Small Updater refactor & Set correct permissions on Linux when extracting files

TSRBerry:
As I was checking for Linux exclusive Updater issues in Avalonia, I noticed a lot of suggestions in the files.
Even after this small refactor they still have a lot of remaining suggestions/warnings, but since we plan on replacing the Updater anyway, I didn't want to bother fixing everything.

I mostly cared about setting the correct permissions on the extracted files on Linux and it turns out we never used the file permissions present in the archive.

Aside from that I also made small typo corrections.

- [x] ~~Will be rebased once #4269 is merged.~~

(The diff of this PR looks horrible, sorry about that!)

---

Thanks to @piplup55 for telling me about this issue!

### DIFF
--- a/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -300,8 +300,6 @@ namespace Ryujinx.Modules
                         ryuExe = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, OperatingSystem.IsWindows() ? "Ryujinx.exe" : "Ryujinx");
                     }
 
-                    SetFileExecutable(ryuExe);
-
                     Process.Start(ryuExe, CommandLineState.Arguments);
 
                     Environment.Exit(0);
@@ -472,22 +470,6 @@ namespace Ryujinx.Modules
             worker.Start();
         }
 
-        private static void SetFileExecutable(string path)
-        {
-            const UnixFileMode ExecutableFileMode = UnixFileMode.UserExecute |
-                                                    UnixFileMode.UserWrite |
-                                                    UnixFileMode.UserRead |
-                                                    UnixFileMode.GroupRead |
-                                                    UnixFileMode.GroupWrite |
-                                                    UnixFileMode.OtherRead |
-                                                    UnixFileMode.OtherWrite;
-
-            if (!OperatingSystem.IsWindows() && File.Exists(path))
-            {
-                File.SetUnixFileMode(path, ExecutableFileMode);
-            }
-        }
-
         private static async void InstallUpdate(TaskDialog taskDialog, string updateFile)
         {
             // Extract Update
@@ -605,8 +587,6 @@ namespace Ryujinx.Modules
             });
 
             Directory.Delete(UpdateDir, true);
-
-            SetFileExecutable(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx"));
 
             _updateSuccessful = true;
 
--- a/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -478,9 +478,9 @@ namespace Ryujinx.Modules
 
             if (OperatingSystem.IsLinux())
             {
-                await using Stream          inStream   = File.OpenRead(updateFile);
-                await using GZipInputStream gzipStream = new(inStream);
-                await using TarInputStream  tarStream  = new(gzipStream, Encoding.ASCII);
+                using Stream          inStream   = File.OpenRead(updateFile);
+                using GZipInputStream gzipStream = new(inStream);
+                using TarInputStream  tarStream  = new(gzipStream, Encoding.ASCII);
 
                 await Task.Run(() =>
                 {
@@ -516,7 +516,7 @@ namespace Ryujinx.Modules
             }
             else
             {
-                await using Stream  inStream = File.OpenRead(updateFile);
+                using Stream  inStream = File.OpenRead(updateFile);
                 using ZipFile zipFile  = new(inStream);
 
                 await Task.Run(() =>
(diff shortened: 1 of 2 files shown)


Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -516,13 +502,12 @@ private static async void InstallUpdate(TaskDialog taskDialog, string updateFile
                             tarStream.CopyEntryContents(outStream);
                         }
 
+                        File.SetUnixFileMode(outPath, (UnixFileMode)tarEntry.TarHeader.Mode);
```
AcK77, line 505:
```suggestion
                        if (OperatingSystem.IsLinux())
                        {
                            File.SetUnixFileMode(outPath, (UnixFileMode)tarEntry.TarHeader.Mode);
                        }
```

Why not add this in a condition and remove the `Debug.Assert` above ?

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -496,14 +478,18 @@ private static async void InstallUpdate(TaskDialog taskDialog, string updateFile
 
             if (OperatingSystem.IsLinux())
             {
-                using Stream          inStream   = File.OpenRead(updateFile);
-                using GZipInputStream gzipStream = new(inStream);
-                using TarInputStream  tarStream  = new(gzipStream, Encoding.ASCII);
+                await using Stream          inStream   = File.OpenRead(updateFile);
+                await using GZipInputStream gzipStream = new(inStream);
+                await using TarInputStream  tarStream  = new(gzipStream, Encoding.ASCII);
```
AcK77, line 483: Are you sure about those `await` ?

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -297,11 +297,9 @@ public static async void UpdateRyujinx(Window parent, string downloadUrl)
 
                     if (!Path.Exists(ryuExe))
                     {
-                        ryuExe = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, OperatingSystem.IsWindows() ? "Ryujinx.exe" : "Ryujinx");
+                        ryuExe = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, OperatingSystem.IsWindows() ? "Ryujinx.Ava.exe" : "Ryujinx.Ava");
```
AcK77, line 300: Nice catch!

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -603,8 +588,6 @@ private static async void InstallUpdate(TaskDialog taskDialog, string updateFile
 
             Directory.Delete(UpdateDir, true);
 
-            SetFileExecutable(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx"));
```
AcK77, line 606: It's not needed anymore ?

Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -531,7 +516,7 @@ private static async void InstallUpdate(TaskDialog taskDialog, string updateFile
             }
             else
             {
-                using Stream  inStream = File.OpenRead(updateFile);
+                await using Stream  inStream = File.OpenRead(updateFile);
```
AcK77, line 519: same as other `await`

AcK77 (commented):
Comments are the same for GTK Updater

Ryujinx.Ava/Modules/Updater/Updater.cs
TSRBerry: `Debug.Assert()` won't be added to release binaries and this method only gets called on Linux anyway. So we could skip the check entirely and execute the call to `SetUnixFileMode` directly.

Ryujinx.Ava/Modules/Updater/Updater.cs
TSRBerry: IDE suggested adding them, so I did. Should just allow other threads to run while these are yielding I believe.

Ryujinx.Ava/Modules/Updater/Updater.cs
TSRBerry: Yup, since we already set the correct permissions when extracting the update now!

Ryujinx.Ava/Modules/Updater/Updater.cs
AcK77: Well those methods aren't `async` so I don't see why we should `await` them

Ryujinx.Ava/Modules/Updater/Updater.cs
AcK77: I'd prefer to get it under a condition and remove the assert to get the code more obvious than just add code to remove warning and assume something isn't called under specifics OS

Ryujinx.Ava/Modules/Updater/Updater.cs
TSRBerry: Okay I think I figured it out. These classes implement the `IAsyncDisposable` interface, which allows us to call `DisposeAsync()` in an async context instead of the regular `Dispose()`. Since these usings are inside a async method the IDE recommends switching to `await using`.

I can remove it again if you want, but it should be safe to leave them like that.

AcK77: I prefer to add `async` only when it's needed, so it's probably better to remove them :)

AcK77 (approved):

Ryujinx.Ava/Modules/Updater/Updater.cs
marysaka: Not a nice catch, this is actually wanted for when we migrate from GTK to Avalonia as the default UI

AcK77: But this is the Ava Updater not the GTK one

marysaka (approved):

PR #4318, opened 2023-01-19
Implement CSET and CSETP shader instructions

gdkchan:
Implements 2 shader instructions, CSET and CSETP. CSETP was already implemented, but it was a very incomplete and partial implementation, which was good enough for most games but not all of them. This change implements a more complete and more correct implementation.

Fixes textures on Persona 4 Golden.
Before:
image
After:
image

Testing is welcome.


yuwenhua653:
Looks like there is no problem

marysaka (approved):

riperiperi (approved):
Nice! 😌

PR #4320, opened 2023-01-20
AvaloniaKeyboardDriver: Swallow TextInput events to avoid bell on macOS

merryhime:
On macOS, one often has the bell noise when using a keyboard as a controller.

Here we propose a fix to this by swallowing the TextInputEvent when it bubbles back up to MainWindow if Handled wasn't set during tunnelling.

We don't do this on KeyDownEvent because we want TextInputEvent to occur.

Comments very welcome.


riperiperi (approved):
Nice work, seems to work fine. I think the workaround included in macos1 broke text input in a few places, this one works properly.

marysaka (approved):

PR #4276, opened 2023-01-12
Ava UI: `TitleUpdateWindow` Refactor

IsaacMarovitz:
- Switches to using compiled bindings
- Moves `TitleUpdateWindow` to a content dialogue
- Updates layout to match rest of Ava UI

Folder opener should be done programmatically on macOS when #4286 is merged

Screenshot 2023-01-12 at 22 06 30

PR #4326, opened 2023-01-21
Ava UI: Various Fixes

AcK77:
This PR fixes :
- A screenshot path issue introduced in a recent PR (#4240), which currently doesn't allow users to take screenshots because of a malformed path.
- After the refactoring of `TitleUpdateManager` (#4276), if the JSON is malformed, it reset the internal field but not the file itself. If an update file doesn't exist anymore, it doesn't appear in the UI but still exist in the JSON file. This is now fixed.
- After the refactoring of `Renderer` (#4297), in `_wndProcDelegate` some fields are tested without taking care of the message send to the window, this is now fixed.

### DIFF
--- a/Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
+++ b/Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
@@ -21,8 +21,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using SpanHelpers = LibHac.Common.SpanHelpers;
+using System.Text;
 using Path = System.IO.Path;
+using SpanHelpers = LibHac.Common.SpanHelpers;
 
 namespace Ryujinx.Ava.UI.ViewModels;
 
@@ -90,6 +91,8 @@ public class TitleUpdateViewModel : BaseModel
                 Selected = "",
                 Paths    = new List<string>()
             };
+
+            Save();
         }
 
         LoadUpdates();
@@ -101,6 +104,9 @@ public class TitleUpdateViewModel : BaseModel
         {
             AddUpdate(path);
         }
+
+        // NOTE: Save the list again to remove leftovers.
+        Save();
 
         TitleUpdateModel selected = TitleUpdates.FirstOrDefault(x => x.Path == _titleUpdateWindowData.Selected, null);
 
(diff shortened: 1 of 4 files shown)


gdkchan (approved):

Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
```diff
@@ -223,4 +229,24 @@ public async void Add()
...
+
+            if (update == SelectedUpdate)
+            {
+                _titleUpdateWindowData.Selected = update.Path;
+            }
+        }
+
+        using FileStream titleUpdateJsonStream = File.Create(_titleUpdateJsonPath, 4096, FileOptions.WriteThrough);
+
+        titleUpdateJsonStream.Write(Encoding.UTF8.GetBytes(JsonHelper.Serialize(_titleUpdateWindowData, true)));
```
TSRBerry, line 250:
```suggestion
        File.WriteAllBytes(_titleUpdateJsonPath, Encoding.UTF8.GetBytes(JsonHelper.Serialize(_titleUpdateWindowData, true)));
```
Could this be shortened like this?

Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
```diff
@@ -102,6 +105,9 @@ private void LoadUpdates()
             AddUpdate(path);
         }
 
+        // NOTE: Save the list again to remove leftovers.
+        Save();
```
TSRBerry, line 109: It's probably not easier to remove the paths that don't exist in `TitleUpdates` from `_titleUpdateWindowData.Paths` instead, right?

TSRBerry (commented):
Lgtm, nice to see these issues getting fixed. :D
Just some questions again, before I'll approve.

Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
AcK77: The using is here to use `WriteThrough` idk why it was added, maybe to be sure the real data will be written. I just kept the original code as-is.

Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
AcK77: That will not update the JSON file if a file doesn't exists and missing files path will stack in the JSON files.

Ryujinx.Ava/UI/ViewModels/TitleUpdateViewModel.cs
TSRBerry: Hmm, it looks fine to me to replace. `WriteThrough` just ensures cache is ignored and stuff is written to the disk directly. `WriteAllBytes` will open the file and close it after everything was written, so the changes should also be written to the disk at that point.

AcK77: That should be fine yeah. Fixed!

TSRBerry (approved):

PR #4328, opened 2023-01-21
Arm64: Simplify TryEncodeBitMask and use for constants

merryhime:
* O(1) algorithm for determining bit immediates.
* Detect when bit immediates are usable for materalizing constants.
* Very minor improvement in performance on arm64.

### DIFF
--- a/ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
+++ b/ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
@@ -1,5 +1,4 @@
 using ARMeilleure.IntermediateRepresentation;
-using System;
 using System.Numerics;
 
 namespace ARMeilleure.CodeGen.Arm64
@@ -50,7 +49,7 @@ namespace ARMeilleure.CodeGen.Arm64
             // Any value AND all ones will be equal itself, so it's effectively a no-op.
             // Any value OR all ones will be equal all ones, so one can just use MOV.
             // Any value XOR all ones will be equal its inverse, so one can just use MVN.
-            if (value == ulong.MaxValue)
+            if (value == 0 || value == ulong.MaxValue)
             {
                 immN = 0;
                 immS = 0;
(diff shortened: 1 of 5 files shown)


ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
```diff
@@ -140,34 +78,11 @@ public static bool TryEncodeBitMask(ulong value, out int immN, out int immS, out
...
-            }
-            else
-            {
-                immN = 0;
-                immS = (0xf80 >> BitOperations.Log2((uint)repetitions)) & 0x3f;
-            }
-
-            immS |= onesCount - 1;
+            immN = (elementSize >> 6) & 1;
+            immS = (((~elementSize + 1) << 1) | (onesInElement - 1)) & 0x3f;
```
merryhime, line 82: I'm happy to switch this to `(0xf80 >> BitOperations.Log2((uint)repetitions))` if that makes more sense to you.

gdkchan:
Might be worth changing `CodeGenerator.GenerateCopy` to also try using the bitmask mov if possible, since that's just one instruction. That was not done before because the `TryEncodeBitMask` function was pretty expensive, so  checking it for every copy could be a bit costly.

ARMeilleure/CodeGen/Arm64/CodeGenCommon.cs
gdkchan: `(~elementSize + 1)` could be `-elementSize` I guess?

merryhime:
> Might be worth changing CodeGenerator.GenerateCopy to also try using the bitmask mov if possible, since that's just one instruction.

Done I think.

gdkchan (approved):
lgtm, thanks. Tested a bunch of games on my m1 macbook and they are working fine.

marysaka (approved):

PR #4329, opened 2023-01-21
Vulkan: Reset queries on same command buffer

riperiperi:
Vulkan seems to complain when the queries are reset on another command buffer. No idea why, the spec really could be written better in this regard. This fixes complaints logged by the validation layers, and hopefully any implementations that care extensively about them.

This change _guesses_ how many queries need to be reset and resets as many as possible at the same time to avoid splitting render passes. If it resets too many queries, we didn't waste too much time - if it runs out of resets it will batch reset 10 more.

The number of queries reset is the maximum number of queries in the last 3 frames. This has been worked into the AutoFlushCounter so that it only resets up to 32 if it is yet to force a command buffer submission in this attachment.

This is only done for samples passed queries right now, as they have by far the most resets. Maybe in future, shared query pools with multiple entries can be used so that they can be reset in bulk rather than by a ton of sequential calls.

This fixes occlusion queries having bogus results on macOS, mostly visible in Super Mario Odyssey and A Hat in Time. This is also already in `macos1`.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
+++ b/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Linq;
 
 namespace Ryujinx.Graphics.Vulkan
 {
@@ -16,6 +17,10 @@ namespace Ryujinx.Graphics.Vulkan
         private bool _hasPendingQuery;
         private int _queryCount;
 
+        private int[] _queryCountHistory = new int[3];
+        private int _queryCountHistoryPtr;
+        private int _remainingQueries;
+
         public void RegisterFlush(ulong drawCount)
         {
             _lastFlush = Stopwatch.GetTimestamp();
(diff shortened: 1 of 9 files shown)


Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
```diff
@@ -16,6 +17,10 @@ internal class AutoFlushCounter
         private bool _hasPendingQuery;
         private int _queryCount;
 
+        private int[] _queryCountHistory = new int[3];
+        private int _queryCountHistoryPtr;
```
gdkchan, line 21: Maybe call it '_queryCountHistoryIndex' rather than '_queryCountHistoryPtr' since it's not really a pointer.

Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -226,6 +226,7 @@ public static Extent2D ChooseSwapExtent(SurfaceCapabilitiesKHR capabilities)
         public unsafe override void Present(ITexture texture, ImageCrop crop, Action swapBuffersCallback)
         {
             uint nextImage = 0;
+            _gd.PipelineInternal.AutoFlush.Present();
```
gdkchan, line 229: Nit: Do the call before `uint nextImage = 0;`?

gdkchan (approved):
lgtm, thanks.

Squall-Leonhart:
Theres no global state tracking for Command Buffers, so theres probably no way for one command buffer to know the state of another in a way that wouldn't rely on undefined behavior and race condition, as of 2021 though, the spec does include a way to Reset a Pool of Command buffers, but some drivers just relay this as resetting the individual command buffers in a loop.

marysaka (approved):

PR #4337, opened 2023-01-23
Remove use of GetFunctionPointerForDelegate to get JIT cache function pointer

gdkchan:
This allows us to get rid of a call to `Marshal.GetFunctionPointerForDelegate`. Should make translating new functions a little bit faster on the JIT, but it's unlikely to give any noticeable improvements in practice.


marysaka (approved):

riperiperi (approved):
Makes a lot more sense like this.

PR #4343, opened 2023-01-26
Check KeyboardMode in GUI

WilliamWsyHK:
This PR will allow the GUI to check against the `KeyboardMode` before accepting the text blindly.

This will probably resolve some funny moments such as entering free-text, and then crashed, which suppose to accept only numbers, such as in Pokemon the center asks "month of birth".

Ref. https://discord.com/channels/410208534861447168/410208534861447170/1068159781640093756


Ryujinx/Ui/Applet/SwkbdAppletDialog.cs
```diff
@@ -81,9 +87,72 @@ private void OnInputActivated(object sender, EventArgs e)
             }
         }
 
+        private bool CheckInputTextAgainstKeyboardMode()
+        {
+            bool isTextAgreeWithKeyboardMode = false;
```
Pythonic-Rainbow, line 92: I think this should be true? I tested your PR with this line being true and it works as expected in GTK

Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
```diff
@@ -93,8 +93,67 @@ void OverlayOnPositionChanged(object sender, PixelPointEventArgs e)
             {
                 if (eventArgs.Result == ContentDialogResult.Primary)
                 {
-                    result = UserResult.Ok;
-                    input = content.Input.Text;
+                    bool isTextAgreeWithKeyboardMode = false;
```
Pythonic-Rainbow, line 96: I think this should be true?  But for some reason I can't compile ava

Pythonic-Rainbow (changes requested):
I think the default value of `isTextAgreeWithKeyboardMode` should be true?

WilliamWsyHK:
Yes, that is the thing I discover just after I shutdown my computer, I will change it once available.

Ryujinx/Ui/Applet/SwkbdAppletDialog.cs
WilliamWsyHK: Updated.

Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
WilliamWsyHK: Updated.

Pythonic-Rainbow (approved):

Pythonic-Rainbow:
The check for number works. I have a few suggestions to further improve this PR:

1. Add checks for other KeyboardModes
2. Maybe we can display more than 1 constraints in the dialog. In the case of registering birthday in Pokemon Scarlet, the dialog will pop up with "Must be 1-2 characters". The line will change to "Must be numbers only" when I input an alphabet. Maybe we can display both constraints at the same time?

Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
```diff
@@ -125,11 +127,24 @@ public void SetInputLengthValidation(int min, int max)
...
 
+        private void SetInputValidation(KeyboardMode mode)
+        {
+            _checkInput = mode switch
+            {
+                KeyboardMode.NumbersOnly => text => text.All(char.IsDigit),
+                KeyboardMode.Alphabet => text => text.All(char.IsAsciiLetter),
+                KeyboardMode.ASCII => text => text.All(char.IsAscii),
+                _ => _ => true
+            };
```
TSRBerry, line 138: Could you add the validation info messages here as well?

TSRBerry (changes requested):
I restructured your code a little bit, I hope that was okay.

I'm not sure about the current message, they feel a bit weird, but I'm not sure what to write instead.

Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
WilliamWsyHK: Added. Please review.

src/Ryujinx/Ui/Applet/SwkbdAppletDialog.cs
```diff
@@ -53,23 +63,49 @@ public void SetInputLengthValidation(int min, int max)
...
+                    _checkInput = text => text.All(char.IsDigit);
+                    break;
+                case KeyboardMode.Alphabet:
+                    _validationInfoText += "<i>Must be alphabets only.</i>";
+                    _checkInput = text => text.All(char.IsAsciiLetter);
+                    break;
+                case KeyboardMode.ASCII:
+                    _validationInfoText += "<i>Must be ASCII text only.</i>";
+                    _checkInput = text => text.All(char.IsAscii);
+                    break;
```
TSRBerry, line 102: @Ryujinx/moderators Could you take a look at the text here and check if that's fine? Maybe there is a way to write this better, but I'm not sure.

src/Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -542,6 +542,9 @@
   "SwkbdMinCharacters": "Must be at least {0} characters long",
   "SwkbdMinRangeCharacters": "Must be {0}-{1} characters long",
   "SoftwareKeyboard": "Software Keyboard",
+  "SoftwareKeyboardModeNumbersOnly": "Must be numbers only",
+  "SoftwareKeyboardModeAlphabet": "Must be alphabets only",
+  "SoftwareKeyboardModeASCII": "Must be ASCII text only",
```
TSRBerry, line 547: Same for the text in Avalonia

src/Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
```diff
@@ -9,18 +9,23 @@
...
     internal partial class SwkbdAppletDialog : UserControl
     {
-        private Predicate<int> _checkLength;
+        private Predicate<int> _checkLength = _ => true;
+        private Predicate<string> _checkInput = _ => true;
         private int _inputMax;
         private int _inputMin;
         private string _placeholder;
 
+        private string _validationInfoText = "";
```
TSRBerry, line 27: This is not needed for Avalonia, since `Error.Text` has a getter you can use. For Gtk I only added it because the `Markup` property doesn't have a getter.

src/Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
```diff
@@ -107,29 +117,51 @@ public void SetInputLengthValidation(int min, int max)
...
+            ApplyValidationInfo();
+            Message_TextInput(this, new TextInputEventArgs());
+        }
+
+        private void SetInputValidation(KeyboardMode mode)
+        {
+            switch (mode)
+            {
+                case KeyboardMode.NumbersOnly:
+                    _validationInfoText = string.IsNullOrEmpty(_validationInfoText) ? LocaleManager.Instance.UpdateAndGetDynamicValue(LocaleKeys.SoftwareKeyboardModeNumbersOnly) : string.Join("\n", _validationInfoText, LocaleManager.Instance.UpdateAndGetDynamicValue(LocaleKeys.SoftwareKeyboardModeNumbersOnly));
```
TSRBerry, line 140: Here it might be better to get the text from `LocaleManager` first and then do the check if the error text is empty.

TSRBerry (changes requested):
Thank you for the adjustments! I pinged the mods to take a look at the messages here and added a few comments, where I think we could improve the Avalonia part a little.

src/Ryujinx.Ava/UI/Applet/SwkbdAppletDialog.axaml.cs
WilliamWsyHK: Updated.

TSRBerry (approved):
Lgtm, thank you!

(Still not sure about the messages, but since I don't know better either I think it's fine to approve.)

riperiperi (commented):
This makes sense, though I'm not sure if IsAsciiLetter is identical to the Alphabet mode. Ascii won't have a bunch of accented characters, for instance.

AcK77 (approved):

src/Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -545,6 +545,9 @@
...
```
yell0wsuit, line 550:
```suggestion
  "SoftwareKeyboardModeNumbersOnly": "Must be in numbers (0-9) only",
  "SoftwareKeyboardModeAlphabet": "Must be in English alphabets only",
  "SoftwareKeyboardModeASCII": "Must be in numbers (0-9) and/or English alphabets only",
```

``Must be ASCII text only`` may confuse some users because they may not know about ASCII. While ASCII has the usual punctuation marks such as ``. , ! @ # $ % ^ & * ( )`` and so on (haven't tested which games use these), it also has the control characters (nul, del, etc.) -- see https://en.wikipedia.org/wiki/ASCII#Character_set. So we need to test which games that accepts those characters and change the ``SoftwareKeyboardModeASCII`` message accordingly so that users can understand easily.

TSRBerry: This should also be changed for Gtk

marysaka: This can also be address separately imo.

marysaka (approved):

WilliamWsyHK:
Thanks, I will rebase one more time and force-push it.

PR #4345, opened 2023-01-26
Fixed some multiple enumerations

goremykin:


jduncanator:
Why? Another random bot PR?

gdkchan:
> Why? Another random bot PR?

This looks like a valid PR.

jduncanator:
> > Why? Another random bot PR?
> 
> 
> 
> This looks like a valid PR.

Half the changes don't even make sense. Like calling ToArray, but then assigning it back as an IEnumerable anyway, or assigning the first element to a variable, but then not using it.

I'm not sure what the purpose of this PR is? Everything in this PR is fine as is as an IEnumerable.

gdkchan:
I'm not sure either. It would be better if the author provided description. Creating an array for the `IEnumerable` *might* be more efficient if it is being enumerated multiple times. It probably would not apply to all the cases changed here though.

jduncanator:
> I'm not sure either. It would be better if the author provided description. Creating an array for the `IEnumerable` *might* be more efficient if it is being enumerated multiple times. It probably would not apply to all the cases changed here though.

There's only one or two instances changed here that it _might_ provide a performance improvement, and neither are in performance sensitive areas.

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -1568,7 +1568,7 @@ public bool DataOverlaps(Texture texture, TextureViewCompatibility compatibility
             MultiRange otherRange = texture.Range;
 
             IEnumerable<MultiRange> regions = _sizeInfo.AllRegions().Select((region) => Range.GetSlice((ulong)region.Offset, (ulong)region.Size));
-            IEnumerable<MultiRange> otherRegions = texture._sizeInfo.AllRegions().Select((region) => otherRange.GetSlice((ulong)region.Offset, (ulong)region.Size));
+            MultiRange[] otherRegions = texture._sizeInfo.AllRegions().Select((region) => otherRange.GetSlice((ulong)region.Offset, (ulong)region.Size)).ToArray();
```
jduncanator, line 1571: This change is useless, as the array is only iterated over once, and is iterated over as an IEnumerable in a foreach loop anyway.

It actually results in all of the values being iterated twice. Once first to build the array, and then a second time to iterate over the built array. The original code would be more performant.

Ryujinx.Graphics.Shader/Instructions/InstEmitFlowControl.cs
```diff
@@ -41,7 +41,7 @@ public static void Brx(EmitterContext context)
 
             Operand address = context.IAdd(Register(op.SrcA, RegisterType.Gpr), Const(offset));
 
-            var targets = context.CurrBlock.Successors.Skip(startIndex);
+            var targets = context.CurrBlock.Successors.Skip(startIndex).ToArray();
```
jduncanator, line 44: Ditto for this change. The enumerable is only iterated over once, and is iterated over as an IEnumerable in a foreach loop anyway.

Again, it results in all of the values being iterated twice. Once first to build the array, and then a second time to iterate over the built array. The original code would be more performant.

Ryujinx.HLE/HOS/Services/Sdb/Pdm/QueryService/QueryPlayStatisticsManager.cs
```diff
@@ -74,12 +74,14 @@ internal static ResultCode GetPlayStatistics(ServiceCtx context, bool byUserId =
...
+            var filteredApplicationPlayStatisticsArray = filteredApplicationPlayStatistics.ToArray();
+            
+            for (int i = 0; i < filteredApplicationPlayStatisticsArray.Length; i++)
             {
-                MemoryHelper.Write(context.Memory, outputPosition + (ulong)(i * Unsafe.SizeOf<ApplicationPlayStatistics>()), filteredApplicationPlayStatistics.ElementAt(i).Value);
+                MemoryHelper.Write(context.Memory, outputPosition + (ulong)(i * Unsafe.SizeOf<ApplicationPlayStatistics>()), filteredApplicationPlayStatisticsArray[i].Value);
             }
 
-            context.ResponseData.Write(filteredApplicationPlayStatistics.Count());
+            context.ResponseData.Write(filteredApplicationPlayStatisticsArray.Length);
```
jduncanator, line 84: This change could provide a slight performance benefit, as the old code had to enumerate twice, once for the loop and then once for the ElementAt call, however this code is not in a performance sensitive area, so I'd prefer the readability of the original code over this.

jduncanator (changes requested):

goremykin:
Wow, didn't expect so quick response. Give me some time to explain each change

Ryujinx.Memory.Tests/MultiRegionTrackingTests.cs
```diff
@@ -389,7 +389,7 @@ public void InheritHandles()
                 Assert.IsTrue(throws);
             }
 
-            IEnumerable<IRegionHandle> combinedHandles = combined.GetHandles();
+            IEnumerable<IRegionHandle> combinedHandles = combined.GetHandles().ToArray();
```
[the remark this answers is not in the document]
goremykin: This is just a test, but I decided to add .ToArray here just to suppress the warning. Yes, I agree that indexer would be better

Ryujinx.Input/Assigner/GamepadButtonAssigner.cs
```diff
@@ -61,11 +61,12 @@ public bool ShouldCancel()
 
         public string GetPressedButton()
         {
-            IEnumerable<GamepadButtonInputId> pressedButtons = _detector.GetPressedButtons();
+            var pressedButtons = _detector.GetPressedButtons().ToArray();
```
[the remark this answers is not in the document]
goremykin: `GetPressedButtons` returns `_stats.Where(kvp => CheckButtonPressed(kvp.Value)).Select(kvp => kvp.Key)`

So previously we executed the whole chain until we meet a suitable element when called Any. And again when we called First.
I just noticed, that my variant is bad, as I iterate the whole chain ones, but to the end. So probably the best solution is:
`_detector.GetPressedButtons().Take(1).ToArray();`

So in this case this chain `_stats.Where(kvp => CheckButtonPressed(kvp.Value)).Select(kvp => kvp.Key)` will be exetuted once until we meet the element.

What do you think?

Ryujinx.HLE/HOS/Services/Sdb/Pdm/QueryService/QueryPlayStatisticsManager.cs
goremykin: Yes, as you noticed we perform filtering twice. But make attention on the filtering on the line 64:
`filteredApplicationPlayStatistics = filteredApplicationPlayStatistics.Where(kv => titleIds.Contains(kv.Value.TitleId));`

`titleIds` is `List<ulong>`

This place in itself is not the most optimal, but among other things, it is also performed twice...
In any case it doesn't matter as `applicationPlayStatistics` is not used. So let's revert if you like the previous one

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
```diff
@@ -350,7 +350,7 @@ public ResultCode AtmosphereReloadHostsFile(ServiceCtx context)
 
             if (hostEntry != null)
             {
-                IEnumerable<IPAddress> addresses = GetIpv4Addresses(hostEntry);
+                IPAddress[] addresses = GetIpv4Addresses(hostEntry);
```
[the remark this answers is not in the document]
jduncanator: The general rule of thumb in the code base is for API's to return IEnumerable everywhere, unless there is a good reason not to. I'd prefer leaving GetIpv4Addresses returning an IEnumerable, and then serialising it to an array here if needed.

Ryujinx.Graphics.Gpu/Image/Texture.cs
goremykin: Hm, I am not sure. So what is `otherRegions`?
It is actually a lazy iteration over _mipOffsets or AllOffsets:
```
public IEnumerable<Region> AllRegions()
{
    if (_is3D)
    {
        for (int i = 0; i < _mipOffsets.Length; i++)
        {
            int maxSize = TotalSize - _mipOffsets[i];
            yield return new Region(_mipOffsets[i], Math.Min(maxSize, LevelSizes[i]));
        }
    }
    else
    {
        for (int i = 0; i < AllOffsets.Length; i++)
        {
            yield return new Region(AllOffsets[i], SliceSizes[i % _levels]);
        }
    }
}
```

plus select:
`texture._sizeInfo.AllRegions().Select((region) => otherRange.GetSlice((ulong)region.Offset, (ulong)region.Size))`

If `foreach (MultiRange region in regions)` will make more than one iteration, we will execute all this chain above from the start over and over.

Please correct me if I am wrong

goremykin: Just put a breakpoint in the `AllRegions`, make sure `regions` has more than 2 elements and run.
The breakpoint must fire on every iteration of the outer loop

jduncanator: This is how `yield` works. It's to be expected that the AllRegions function is re-entrant, however it will re-enter _inside_ the for loop. It's not iterating the entire `_mipOffsets` array each time. If you place a breakpoint at the start of the AllRegions function, you'll see it's only hit once, however a breakpoint inside the for loop will be hit for each entry of _mipOffsets. This is expected, as it's generating the IEnumerable on the fly.

Changing this to use `.ToArray` won't change any of that, keep your breakpoint there and you'll see that the breakpoint in AllRegions is hit the same amount of times.

jduncanator: Ahh I see what you mean now. It wasn't clear originally that you were referring just to `otherRegions` being iterated multiple times. Yes, if `regions` has multiple items in it, then `otherRegions` will be iterated multiple times. This is probably a valid change in that circumstance.

If you're going to change this, I'd be changing both. It looks wrong to leave one as an IEnumerable, and the other as an array.

Ryujinx.Graphics.Shader/Instructions/InstEmitFlowControl.cs
jduncanator, line 44: Rather than make this change here, it's probably better to refactor this entire code. Even after making this an array, `targets` is sorted and then iterated over multiple times further down, which is likely going to be a bigger performance impact than just changing this here.

jduncanator (changes requested):

Ryujinx.Graphics.Shader/Instructions/InstEmitFlowControl.cs
goremykin: Initially I wanted to disagree as `Skip` actually iterates the source. So if you run this code:
```
var tmp = Generate().Skip(2);

foreach (var _ in tmp)
{
}

foreach (var _ in tmp)
{
}

IEnumerable<int> Generate()
{
    Console.Write(1);
    yield return 1;
    Console.Write(2);
    yield return 2;
    Console.Write(3);
    yield return 3;
    Console.Write(4);
    yield return 4;
    Console.Write(5);
    yield return 5;
}
```
You will see "1234512345" in the output instead of "345345".

**But**, I didn't notice that Successors is actually a List. Usually linq has a lot of checks and optimizations for such cases, I need to perform some benchmarks...

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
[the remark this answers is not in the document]
goremykin: Do you mean return array as IEnumerable from the `GetIpv4Addresses` and cast after call or leave `GetIpv4Addresses` as it was before, but call GetIpv4Addresses().ToArray() where we need to iterate it twice?

Honestly I prefer the second option as in the first scenario when we return a list / array as IEnumerable, code analyzers are not able to detect that this is actually a list / array and show warnings. In this case rider / resharper will warn that multiple enumeration is possible.

But your project - your rules, will do as you say

Ryujinx.Graphics.Shader/Instructions/InstEmitFlowControl.cs
goremykin: Actually you are right. When we call .Skip from a List an then iterate over the result it is smart and really skips this block unlike an example above.

How do you feel about comments that suppress warnings? Something like this:
```
// ReSharper disable once PossibleMultipleEnumeration
// No multiple enumeration as Skip called from the List
```

Additionally I noticed that we sort our enumeration twice in some cases:
```
foreach (var target in targets.OrderBy(x => x.Address))
...
var sortedTargets = targets.OrderBy(x => x.Address);
foreach (Block nextTarget in sortedTargets)
```

Maybe tomorrow or day after tomorrow will check can we do anything or not. Also Array.Sort may be much more performant. So just give me some time to check some ideas :)

Ryujinx.Graphics.Gpu/Image/Texture.cs
goremykin: From the readability perspective probably you are right, but now the outer loop is lazy and it gives some benefits:
```
IEnumerable<MultiRange> regions = _sizeInfo.AllRegions().Select((region) => Range.GetSlice((ulong)region.Offset, (ulong)region.Size));
foreach (MultiRange region in regions)
{
    ...
    if (region.OverlapsWith(otherRegion))
    {
        return true;
    }
    ...
}
```

If this condition in the loop will be true on the, let's say, first iteration, our _mipOffsets / AllOffsets will be iterated only once and Select will be applied to the first item only

Ryujinx.HLE/HOS/Services/Sockets/Sfdnsres/IResolver.cs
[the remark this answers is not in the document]
lostromb: > The general rule of thumb in the code base is for public API's (and functions in general) to return IEnumerable everywhere, unless there is a good reason not to.

Would you perhaps consider `IReadOnlyCollection` instead for return types that we expect might be iterated on more than once?

I can already see though that it's kind of a false tradeoff... For example: methods that return the `IEnumerable` results of a LINQ query - a caller that wanted to iterate twice might call `ToList()` on the enumerable first, like how this PR is doing. But then, if you changed the method to return an `IReadOnlyCollection` or `IReadOnlyList`, it would likely just do the call to `ToList()` internally anyways, so there's no actual performance difference unless the data being enumerated was already a list to begin with.

If it were me and I wanted to overengineer a solution to this, I would probably make an extension method like `IEnumerable.ToSegmentedList()` which would use a pooled segmented list type to use as the temporary buffer which would also amortize the memory allocations.... But I think that's overkill here.

goremykin:
Closed as no feedback

PR #4351, opened 2023-01-29
Accelerate NVDEC VIC surface read/write and colorspace conversion with Arm64 HW intrinsics

gdkchan:
As the title says. Additionally improves `ReadNv12` on x86 a bit (4 OR instructions instead of 8).

Contributes to #4062 even though this one is not listed there.


marysaka (approved):
Seems fine to me :+1:

AcK77 (approved):

PR #4356, opened 2023-01-31
Implement safe depth-stencil blit using stencil export extension

gdkchan:
This implements a different way of doing blit operations for formats where blit is not supported on the GPU. It makes use of the `VK_EXT_shader_stencil_export` extension to write the stencil value of stencil and depth-stencil textures, which notably NVIDIA does not support, but it also does not use the "safe" blit path, so this is not a problem. This change should only impact AMD.

It needs to be tested on AMD GPUs, on Windows and Linux. I want to know if it fixes games that crashes right off the bat with the resolutoin scaling above native, such as Pokémon Scarlet/Violet and Fire Emblem Engage (?). It would be nice to test it on games that blits D24S8 textures too like Pokémon Sword/Shield.


gdkchan:
Marked this as ready for review as a few people tested and confirmed it fixes the depth-stencil blit related crashes on AMD. Pokémon Scarlet/Violet was still not tested however.

MutantAura:
Confirmed to fixing SV booting with resolution scaling. 
Also resolves anomalously low performance for AMD cards it seems.

Master: 
image

Branch: 
image

marysaka (approved):

riperiperi (approved):
Nice work. It's weird how the stencil export extension is not supported on NVIDIA, but I guess it doesn't need to use it?

gdkchan:
> Nice work. It's weird how the stencil export extension is not supported on NVIDIA, but I guess it doesn't need to use it?

Correct, it doesn't because NVIDIA supports D32FloatS8Uint texture blit, and even if it didn't it aliases the depth format as a colour format to and do a colour blit on NVIDIA and Intel (this doesn't work on AMD, that's why it has this "safe" path).

PR #4358, opened 2023-02-01
hle: Stub a bunch of service calls required to boot qlaunch.

Steveice10:
Stubs a bunch of service calls made by qlaunch (Home Menu) during boot. This allows it to get a lot further into initialization, up until it crashes with a bad 0x0 memory access.


gdkchan:
Do you know why it crashes? Is it not due to one of those stubs?

Steveice10:
> Do you know why it crashes? Is it not due to one of those stubs?

I have not been able to trace back the root cause yet, it seems to be an attempt to load some function pointer in what looks like scene initialization code so I'm not sure if that would be related to any of these stubs.

PR #4364, opened 2023-02-03
Handle mismatching texture size with copy dependencies

### LINKED ISSUE #1629: The Longest Five Minutes: Title screen/menus no longer render

There is a regression introduced in 1.0.4921 [PR 1198](https://github.com/Ryujinx/Ryujinx/pull/1198) that stops title screen and menu graphics from being rendered in the game [The Longest Five Minutes](https://github.com/Ryujinx/Ryujinx-Games-List/issues/1591), making the game unplayable.

Fire up The Longest Five Minutes on Ryujinx 1.0.4921 or newer. Earlier versions of Ryujinx render the game fine.

### Log file

[TLFM.log](https://github.com/Ryujinx/Ryujinx/files/5408676/TLFM.log)

### Environment?

- Ryujinx version: 1.0.4921

- Game version: 1.0.0

- System Specs:
  - OS: Windows 10
  - CPU: i5-6600K
  - GPU: NVIDIA GTX 1080
  - RAM: 16GB
- Applied Mods : No

### Additional context?

Before PR 1198:

2020-06-03 15_00_14-Ryujinx 1 0 4675 - The Longest Five Minutes v1 0 0 (0100CE1004E72000) (64-bit)

After PR 1198:

2020-10-20 06_13_19-Ryujinx 1 0 5545 - The Longest Five Minutes v1 0 0 (0100CE1004E72000) (64-bit)

### LINKED ISSUE #4325: [Bug Report] master build -  rendering  color issues etc. with ryujinx
## Bug Report

rendering  color issues 
the problem of game save files

#rendering  color issues
 When running games in vulkan mode (or OpenGL), game rendering is likely to have abnormal color issues, the entire rendering will be overexposed and brightened.In the zelda, especially if the 'link' has an extra energy bar, If the extra energy is expended, the game will render normally.  (This seems to exist in other games, e.g. fire emblem engage,Breath of the Wild)
#the problem of game save files
The files of ryujinx are stored by two directories ' \0 ' and ' \1 ' under  '\save '  folder. The archives seem to be consistent, but the impact of reading and writing is ultimately based on the data in  ' \0 ' directory . If every time you save the data in zelda, only  data in the ' \1 ' directory will be rewritten, and  ' \0 ' will not be rewritten by mistake, which leads to the game data not being saved. (This seems to exist in other games, e.g. Persona 5 Royal Edition, kirby and the forgotten land,Breath of the Wild)

### Environment?

- Ryujinx version: 1.1.570
- vulkan mode

- Game version: 1.6.0
- the legend of Zelda Breath of the Wild

- Game version: 1.0.2
- Persona 5 Royal Edition

- System Specs:
  - OS:     win11 home consumer 21h2 22000.556
  - CPU:  Intel i7 12700h
  - GPU:  NVIDIA RTX 3060 6GB
  - RAM: 16GiB
- Applied Mods : No

### Additional context?
IMG_20230121_035538
IMG_20230121_035739

gdkchan:
Because NVN aligns the width of the textures for copy and render, matching textures by size exactly on the texture cache does not always work. The texture entry on the texture pool always have the correct size, but render and copy textures do not, so cases where a texture was rendered or copied, and then later accessed on a shader would not work if we simply checked if the width of the texture matches with the one on the cache.

To make this case work we currently change the texture size when it is accessed on the shader. This is done with a `ChangeSize` method that under the hood creates a new texture, copies the old texture data to the new one, then handles re-creating all child views as needed. This not only adds unneeded complexity, but it's also a source of bugs. The main goal of this PR is getting rid of the `ChangeSize` method and instead handle that using existing methods (in this case copy dependency).

Using copy dependency is safer because there's no risk of data being lost. If for some reason the texture accessed on the shader is smaller than the one in the cache, but they overlap, copy dependency will ensure that the sub-region that is accessed is copied on use, but the original texture still exists, so the data is still there.

The first commit is making it use copy dependencies when the size does not match, but the aligned size does. Subsequent commits does other changes to ensure that we do copies on the minimal possible amount of cases, since copies are not exactly free. Those are some of the changes intended to reduce copies:
- We only assume that the width is aligned now. Height is always expected to match exactly. Since NVN does not align the height, that should be fine. It reduces the amount of "false" overlaps and useless copies that we do. The downside is higher VRAM usage, and that cases of height mismatch but aligned height being the same would no longer work properly.
- The width of the texture that is created for render or copy is now the minimum of the size hint and the lower bound of the aligned width. Since the size hint is *usually* equal to the real texture size, this allows the texture to be created with the correct size from the start, which avoids the need of a copy dependency for shader access later on.
- When finding textures for copy or render on the cache, the width if allowed to be greater than or equal to the requested width, as long the aligned width is equal. Because render and copy can only modify a sub-region of the texture, the texture being larger does not really matter as the additional space can't be accessed anyway.

This fixes a few games that were not rendering properly.
void tRrLM();
master:
image
PR:
image
The Longest Five Minutes:
master:
image
PR:
image
Closes #1629, closes #4325.

It might also fix the "bike nuke" issue on Zelda BOTW, but this one is harder to test since the issue is somewhat random. But at least the cause of this issue that I was able to find should be fixed on this PR.

Might also fix random texture corruption in other games.

This should be tested extensively to ensure there are no graphical or performance regressions on games. Probably UE4 games are good candidates since they copy textures a lot.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
@@ -803,14 +803,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 _context.Renderer.Pipeline.SetScissors(scissors);
             }
 
-            if (clipMismatch)
-            {
-                _channel.TextureManager.UpdateRenderTarget(index);
-            }
-            else
-            {
-                _channel.TextureManager.UpdateRenderTargets();
-            }
+            _channel.TextureManager.UpdateRenderTargets();
 
             if (componentMask != 0)
             {
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/DrawManager.cs
@@ -725,10 +725,25 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 return;
             }
 
+            bool clearDepth = (argument & 1) != 0;
+            bool clearStencil = (argument & 2) != 0;
+            uint componentMask = (uint)((argument >> 2) & 0xf);
             int index = (argument >> 6) & 0xf;
             int layer = (argument >> 10) & 0x3ff;
 
-            engine.UpdateRenderTargetState(useControl: false, layered: layer != 0 || layerCount > 1, singleUse: index);
+            RenderTargetUpdateFlags updateFlags = RenderTargetUpdateFlags.SingleColor;
+
+            if (layer != 0 || layerCount > 1)
+            {
+                updateFlags |= RenderTargetUpdateFlags.Layered;
+            }
+
+            if (clearDepth || clearStencil)
+            {
+                updateFlags |= RenderTargetUpdateFlags.UpdateDepthStencil;
+            }
+
+            engine.UpdateRenderTargetState(updateFlags, singleUse: componentMask != 0 ? index : -1);
 
             // If there is a mismatch on the host clip region and the one explicitly defined by the guest
             // on the screen scissor state, then we need to force only one texture to be bound to avoid
@@ -797,10 +812,6 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 _channel.TextureManager.UpdateRenderTargets();
             }
 
-            bool clearDepth = (argument & 1) != 0;
-            bool clearStencil = (argument & 2) != 0;
-            uint componentMask = (uint)((argument >> 2) & 0xf);
-
             if (componentMask != 0)
             {
                 var clearColor = _state.State.ClearColors;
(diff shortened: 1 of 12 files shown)


nullavisus:
Just wanted to add that while the "bike nuke" issue seems to be fixed in the overworld in BotW, its still an issue while using the menu. Getting off the bike removes the graphical bugs in the menu it seems, so there does indeed seem to be two issues happening at once here. Regardless, at least the game isn't completely unplayable anymore. 
Screenshot (47)

gdkchan:
Which version did you test that on? Because my latest commit is supposed to fix that issue.

GamerzHell9137:
Fixes overlayed elements on Disgea 6

image
image

MutantAura:
> Which version did you test that on? Because my latest commit is supposed to fix that issue.

Mentioned issue is fixed as of latest commit. BoTW rainbow terrain bug is also eliminated on my end.
Seems to partially fix some of the issues AMD GCN cards had in Mario Kart 8 Deluxe too (used to have a green tint).
image

riperiperi (approved):
LGTM. I tested a few games to make sure there weren't any regressions, including A Hat in Time with the fast copies disabled, and I also checked to see if copy frequency changed on a few games. It actually seems to copy _less_ often on Animal Crossing: New Horizons.

LM3 also seems OK with how it's using memory despite constant resizes, but I didn't look too far into it.

It's good to see these resize methods (and the size change edge case when getting a texture from the pool) removed, since they were pretty finnicky.

marysaka (approved):

PR #4365, opened 2023-02-04
Implement copy dependency for depth and color textures

### LINKED ISSUE #2791: [REGRESSION] Luigi's Mansion 3 - Shadow Maps not clearing.

gdkchan:
Luigi's Mansion 3 renders a texture using the R16Unorm format, and samples it using the D16Unorm format. We did not support that case before since those formats are not view compatible. This change adds support for this case using copies (copy dependency).

Fixes shadow issues on Luigi's Mansion 3.
Before:

https://user-images.githubusercontent.com/5624669/216788492-7c69012d-6641-467a-8c17-cb321faf0d2b.mp4

After:

https://user-images.githubusercontent.com/5624669/216788486-e85b3e8a-43e4-4713-ae9b-7205bd4dc92d.mp4

Thanks to @riperiperi for finding the issue.

Fixes #2791.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -735,9 +735,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                 {
                     if (overlap.IsView)
                     {
-                        overlapCompatibility = overlapCompatibility == TextureViewCompatibility.FormatAlias ?
-                            TextureViewCompatibility.Incompatible :
-                            TextureViewCompatibility.CopyOnly;
+                        overlapCompatibility = TextureViewCompatibility.CopyOnly;
                     }
                     else
                     {
@@ -815,7 +813,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     Texture overlap = _textureOverlaps[index];
                     OverlapInfo oInfo = _overlapInfo[index];
 
-                    if (oInfo.Compatibility <= TextureViewCompatibility.LayoutIncompatible || oInfo.Compatibility == TextureViewCompatibility.FormatAlias)
+                    if (oInfo.Compatibility <= TextureViewCompatibility.LayoutIncompatible)
                     {
                         if (!overlap.IsView && texture.DataOverlaps(overlap, oInfo.Compatibility))
                         {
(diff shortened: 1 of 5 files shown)


gdkchan:
Now that the OpenGL regression was fixed and the other PR was merged, I was able to test it on OpenGL. Seems to be working correctly, so ready for review. It is using PBO copies on OpenGL which are usually slow, so I'm a bit worried about the performance there. But it's hard to tell since Luigi's Mansion 3 performance on OpenGL is already extremely poor here, at least without the mod to disable dynamic res that I don't have.

marysaka (approved):

Bjorn29512:
No noticeable regressions in botw, bayo 3 or smo with this PR, will update as i test more titles.

riperiperi:
This PR causes BOTW to copy a depth texture every frame:

image

Normally:

image

It could definitely be a lot worse, but this could hurt weaker GPUs and might cause issues when scaling. I think the rules need to be changed to prefer the most recently modified texture instead of a "perfect" match. One weird case with that would be 2D copy to depth texture (as R32), I guess you'd still want the recently modified match as it will probably be D32 due to being used as a depth attachment.

gdkchan:
I rebased it, but BOTW still copies between a R32Float and a D32Float texture on every frame. Not sure what was the plan to mitigate this, maybe keeping the `FormatAlias` compatibility and let it sample as the other format?

riperiperi:
The texture lookup should prefer the last used format when depth aliasing is possible. There's a possibility it selects the wrong one when the texture has a cached entry on the texture pool (it used it as R32 in the past, but the recent modifications were all D32)... But I'd imagine that should be random and happen after a while, not happen immediately.

gdkchan:
This should be ready for review again, #5719 stopped BOTW from copying depth textures every frame, so I think there's nothing blocking it now.

marysaka (commented):
lgtm

Linear524:
Can't wait for this to merge into master branch :)

riperiperi (approved):
LGTM, tested on a few games and everything looks good. Nice work!

PR #4367, opened 2023-02-05
Implement VP9 loop filtering

gdkchan:
The main goal of this change is porting the loop filtering from libvpx, which should fix the block artifacts on some VP9 videos on games using NVDEC to decode them. In addition to that, there are two other changes:

- The remaining decoder code required to decode a VP9 video (with headers included) has been added. That was done because it's much better to test the decoder standalone with a video file. I decided to keep that code on the emulator, even if some of it is unused, since it makes standalone testing easier in the future too, and we can include unit tests with video files.
- Large refactoring of both new and existing code to conform with our conding styles, done by @TSRBerry (thanks!) Some of it has been automated.

Since we had no loop filtering before, this change will make video decoding slower. That may cause frame drop etc if the decoder is not fast enough in some games. I plan to optimize the decoder more in the future to make up for that, but if possible I'd prefer to not do it as part of this PR, but if the perf loss is too severe I might consider.

This will need to be tested on games that had the block artifacts, it would be nice to confirm if they match hardware now, and get some before/after screenshots etc.


Ryujinx.Graphics.Nvdec.Vp9/ReadBitBuffer.cs
```diff
@@ -0,0 +1,96 @@
...
+        private static int GetMsb(uint n)
+        {
+            Debug.Assert(n != 0);
+            return 31 ^ BitOperations.LeadingZeroCount(n);
+        }
+
+        private static int GetUnsignedBits(uint numValues)
+        {
+            return numValues > 0 ? GetMsb(numValues) + 1 : 0;
+        }
```
turbedi, line 23: These functions already exist in [Ryujinx.Graphics.Nvdec.Vp9/Common/BitUtils.cs](https://github.com/Ryujinx/Ryujinx/blob/43081c16c48d73dfd585d951281fcbd60bebccbc/Ryujinx.Graphics.Nvdec.Vp9/Common/BitUtils.cs#L46)

Bjorn29512:
Significantly improves the block artifacts in FE: Engage. 

Before: 
image

After:
image

Ryujinx.Graphics.Nvdec.Vp9/ReadBitBuffer.cs
TSRBerry: Oh oops, good catch!

Exhigh:
A rebase is required

PR #4368, opened 2023-02-05
Move gl_Layer to vertex shader if geometry is not supported

gdkchan:
Metal and MoltenVK does not support geometry shaders. All UE4 games have a geometry shader that is only used the output layer to render on a 3D LUT texture. This does not really require a geometry shader and can be done on the vertex shader. This change does tne layer assignment on the vertex shader if possible, when geometry shaders are not supported, which allows those games to render correctly on macOS.

master:
Captura de Tela 2023-02-04 às 22 20 45
PR:
Captura de Tela 2023-02-04 às 22 23 06

Contributes to #4062. It's worth noting that the method used here to detect the geometry shaders is different to the one used on macos1 release, because the macos1 release had geometry shader emulation which made the pattern matching not work as the IR  it was producing for the geometry shaders was different. It should have the same end result though.


marysaka (approved):

Ryujinx.Graphics.Shader/Translation/ShaderConfig.cs
```diff
@@ -245,6 +247,22 @@ public void SetLayerOutputAttribute(int attr)
             LayerOutputAttribute = attr;
         }
 
+        public void SetGeomtryShaderLayerInputAttribute(int attr)
```
pierr3, line 250: Would this not be a small typo in Geometry?

gdkchan: Yes, thanks. I fixed it.

Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
```diff
@@ -421,6 +426,39 @@ private void ShaderCacheStateUpdate(ShaderCacheState state, int current, int tot
...
+                if (info.Identification == ShaderIdentification.GeometryLayerPassthrough)
+                {
+                    // We managed to identify that this geometry shader is only used to set the output Layer value,
+                    // we can set the Layer on the previous stage instead (usually the vertex stage) and eliminate it.
+
+                    for (int i = 3; i >= 1; i--)
+                    {
+                        if (translatorContexts[i] != null)
+                        {
+                            translatorContexts[i].SetGeomtryShaderLayerInputAttribute(info.GpLayerInputAttribute);
```
riperiperi, line 451:
```suggestion
                            translatorContexts[i].SetGeometryShaderLayerInputAttribute(info.GpLayerInputAttribute);
```

Ryujinx.Graphics.Shader/Translation/TranslatorContext.cs
```diff
@@ -138,6 +138,16 @@ public void SetNextStage(TranslatorContext nextStage)
             _config.MergeFromtNextStage(nextStage._config);
         }
 
+        public void SetGeomtryShaderLayerInputAttribute(int attr)
```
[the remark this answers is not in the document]
gdkchan: Fixed, somehow I missed those even when the other person commented about the error on `ShaderConfig`.

AcK77 (approved):

PR #4202, opened 2023-01-03
Vulkan: Add workarounds for MoltenVK

riperiperi:
This PR does the following things:

- Adds MoltenVK configuration class and provides helpful configurations without manually specifying environment variables.
- Portability subset workarounds for MoltenVK, based on properties/features:
  - Enables vertex attribute restriding (always 4 byte for now)
  - Enables triangle fan retopology conversion.
  - Enables 3D texture view workaround.
- Add shader specialization for output attribute types (float/int/uint)
  - This should also prevent validation errors relating to this, though all desktop GPUs were OK with it before.
  - This has been rewritten compared to what is in the macos build so it needs some testing.
- Adds special treatement for duplicate render attachments when the GPU is detected as TBDR.
  - Fixes some graphical issues in Mario Kart 8 and Zelda BOTW. The error is caused by two copies of the render target existing in tiled memory, where one that was not written may take priority and erase any changes made during the render pass. Generally when this happens, only one render target is enabled/written, so that target is selected and the other is masked out.
- Add alternate versions of the clear shader for int and uint type targets.
- Disable null descriptors if they are not available, but Secretly Use Them Anyways on MoltenVK as a Funny Joke.
  - In all seriousness, using null descriptors avoids issues with stale resource usage.
  - Rosetta doesn't seem to care. Nobody is allowed to gaslight me by saying rosetta doesn't work without this.
- Actually check for individual ASTC formats before saying it is supported. MoltenVK on Rosetta supports the extension, but not the individual formats.
- Minor additional changes/reordering.
 
This PR does not include other things mentioned on #4062, such as transform feedback or geometry shader emulation. It also doesn't include A64 JIT, hypervisor or any related workarounds, so things will only really work under Rosetta. The performance enhancement "buffer mirrors" also is not present, which may negatively impact some games. This may be lessened by recent changes to reduce inline updates.

One important thing this doesn't include is any sort of workaround for resource usage errors on MoltenVK. These issues are better explained on #4062, but I really have not had time to make some kind of reproduction for the issue to report it to them. I'd rather we didn't have to add some weird resource tracking to our descriptor set updater just for MoltenVK.

PR #4369, opened 2023-02-05
Insert bitcast for assignment of fragment integer outputs on GLSL

gdkchan:
Fixes a regression caused by #4202 which caused GLSL shaders to fail to compile because with this change, fragment outputs might now have integer types, however when assigning from a float variable, it would not insert the variable because internally the code generator assumed it had a float type. This change updates the code to use the correct fragment output type.

Fixes games rendering to integer textures on OpenGL.
Before:
image
After:
image


marysaka (approved):

riperiperi (approved):
Nice catch. I thought I handled this, maybe it got lost in a rebase.

PR #4374, opened 2023-02-06
Support safe blit on non-2D textures

### LINKED ISSUE #4372: 1.1.601 Causes Fire Emblem Engage to crash when playing above native res
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description of Issue

A regression in #4356 ironically causes crashes on map start in FE:E for AMD on linux when the resolution is set to above native. Note that this is exactly the crash that PR was speculated to fix on windows, just on linux now too.

### Reproduction Steps

- Set resolution to above native
- Begin any map (I've only tested mid-game maps but all have had this issue)
- Game should crash as soon as a charictar prortrait is rendered with a dialog box

### Log File

[Ryujinx_1.1.601_2023-02-06_01-46-09.log](https://github.com/Ryujinx/Ryujinx/files/10660435/Ryujinx_1.1.601_2023-02-06_01-46-09.log)

### OS

Ubuntu 20.04

### Ryujinx version

1.1.601

### Game version

1.1

### Environment?

- ##### CPU: R5 3600
- ##### GPU: RX 5700 XT
- ##### RAM: 32 GB
- ##### Mesa: 22.3.4 and 21.x (both have issue)

### Additional context?

There was some speculation that 20.04s newest Mesa was the issue, since it doesn't support Vk 1.3. I updated to the latest Mesa manually and the issue persists.

I've also tested this on 1.1.600 and had no issues at all, so this seems to be the fix in 1.1.601 backfiring

gdkchan:
After #4356 we had some reports of crashes on AMD + Linux because we're hitting the "unsafe" path for non-2D textures. This change allows the safe blit path to be used for all non-2D texture targets.

Fixes #4372. We will need testing on AMD (preferably on both Windows and Linux), for Fire Emblem Engage.


andyblarblar:
This PR fixes the crash in FE:E described in https://github.com/Ryujinx/Ryujinx/issues/4372 for me.

marysaka (approved):

gdkchan:
> There is one thing, I don't think `gl_SampleID` on multisampled copies will be correct for the texel fetch when the copy is scaled (though not sure how you'd query the sample pattern to select the correct sample). ColorBlitMsFragment also doesn't support linear filtering, but that's probably not too bad as it's an edge case for res scale swaps.

Right, I was actually thinking on copies for textures of the same size when I implemented this. I forgot that it can happen for resolution scale copies which will have different sizes (which is kind of silly since resolution scale is what was causing issues with the previous implementation). I guess the correct way to handle this is just pretending the texture has W*SamplesInX width and H*SamplesInY height and then blitting as if it was a 2D texture, but on a shader and doing the interpolation ourselves. It would also require having a different shader for integer textures as those should not be interpolated etc.

TehNomad:
Fire Emblem Engage has been working great on Ryujinx 1.1.604. I haven't had any crashes upscaled 2X on my AMD Vega 56 running on Windows. Thanks for fixing this issue @gdkchan.

PR #4377, opened 2023-02-07
Better filtering of invalid NPadIds when games request unknown supported players

lostromb:
While testing the game Kingdom Rush [0100A280121F6000], I noticed Ryujinx would crash in HidUtils.GetIndexFromNpadIdType() because the game was requesting an unknown NPad ID. The data the game actually sends is:
[0] 0x1d635b53
[1] 0x00000000 (Player1)

This PR adds extra `IsValidNpadIdType()` checks to ignore these unknown values. It also improves the logic of `IsValidNpadIdType()` a bit and makes it more centralized to other functions within IHidServer (for example, where the check used to be just `nPadId >= 0`

With this change the game now boots fine

image

I assume the game is not expecting HID service to return `ResultCode.InvalidNpadIdType` because if I do then it breaks execution

image

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/HidServer/HidUtils.cs
@@ -38,7 +38,9 @@ namespace Ryujinx.HLE.HOS.Services.Hid.HidServer
 
         public static bool IsValidNpadIdType(NpadIdType npadIdType)
         {
-            return npadIdType <= NpadIdType.Player8 || npadIdType == NpadIdType.Handheld || npadIdType == NpadIdType.Unknown;
+            return (npadIdType >= NpadIdType.Player1 && npadIdType <= NpadIdType.Player8) ||
+                npadIdType == NpadIdType.Handheld ||
+                npadIdType == NpadIdType.Unknown;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
I think there's probably something else wrong with HID that is causing the game to pass invalid values. But given we already do validation of the inputs, and our validation was apparently incorrect (it didn't account for negative integer values), I think this is fine.

lostromb:
> I think there's probably something else wrong with HID that is causing the game to pass invalid values

should we file a bug with this game as a repro just so we don't forget about it?

lostromb:
I definitely agree there is some other bug in HID service that is causing out-of-bounds memory reads. If I extend the array bounds to look into memory adjacent to what was passed in, I see what looks like valid data
image

gdkchan:
It's possible some function is not writing the data where it should, and the game is reading memory garbage.
You can try clearing the memory before that line: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.HLE/HOS/Services/ServerBase.cs#L315
by writing zeros to `messagePtr` region up until the end (the total size is 0x100 bytes, so you would write 0x100 - messagePtr bytes of zeros).
If that makes the invalid value, that would confirm the theory. Then you could find the function by getting creative and writing some special value that is unique per call to trace it.

marysaka (approved):

PR #4379, opened 2023-02-07
Ava: Add ChangeVSyncMode() call to render loop

TSRBerry:
This change makes the render loop behave the same as the one in GTK.


gdkchan (approved):
Can confirm is fixes tearing here on NVIDIA + Windows.

riperiperi (approved):
Nice catch, I didn't notice this one since I force driver vsync. Maybe this was causing the weird differences between screens on macos. (perhaps some force vsync)

PR #4125, opened 2022-12-15
Clear CPU side data on GPU buffer clears

gdkchan:
Some games (namely Mario + Rabbids Sparks of Hope) uses buffer clears to clear texture data. This PR change buffer clears to clear the data on both guest and host GPU memory. That way, when the texture data is loaded from guest memory, it loads the correct data rather than memory garbage. In addition to that, memory tracking is signalled to indicate to the textures that the data has been modified. The memory tracking region handles was modified to store a "ID", which represents resource kinds. This is used to enable us to signal all resources *except* buffers, since in this case we clear the data on both GPU and guest memory, there's no need to reload buffer data.

Fixes black screen on Mario + Rabbids Sparks of Hope.
Before:
image
After:
image

PR #4380, opened 2023-02-07
Add copy dependency for some incompatible texture formats

gdkchan:
This enables copy dependency for some formats that would normally be considered incompatible (as the host API does not allow copy between them). Mario + Rabbids Sparks of Hope makes use of that aliasing a R8Unorm texture as RGBA8Unorm. This is only allowed when Width * BytesPerPixel is the same for both textures, since otherwise it wouldn't make sense.

Fixes vertical stripes on Mario + Rabbids Sparks of Hope.
#4125:
image
This PR:
image

Testing is welcome.


Vaskiemaia:
The PR makes the game work perfectly, long gone are the days of vertical lines.

Previous to this PR, the game would only show the vertical lines in certain areas, but not during battles.

To make sure it was solved, I have tested a brand new save and reached the first world and it worked great.

Video - https://www.youtube.com/watch?v=27udDqzPQms

gdkchan:
Ready for review.

riperiperi (approved):
Looks good, nice catch. Not so sure about the ratio stuff, but it won't be used in the current case.

gdkchan:
> Not so sure about the ratio stuff, but it won't be used in the current case.

Not sure about what exactly? It is currently limited to a 4x difference in the formats BPP at most, since you can only write 4 components at most (RGBA) on the shader using image stores. It is possible to do more than that using multiple store operations, but it would require more shaders to handle that case (which we might never need to handle in practice).

riperiperi:
It's mostly because the ratio implies the difference in component count.
A ratio of 2 will cause an out of bounds store in the shortening shader for b and a components, and an out of bounds load in the widening shader. It will also only work if the lowest bpp texture is single component, rg to rgba won't work (ratio 2), as it won't alternate `r` and `g` between two pixels to build the final result.

If all of this is intentional that's fine.

gdkchan:
AFAIK out of bounds stores will just discard the values if they don't exist in the format. So if the image format is RG the BA values doesn't really matter and it will not store them. A ratio of 2 should work as long the BPP is <= 4. For example, if the format is RG16, it will use a R32 format (that has a single component). And if the destination is RGBA16, it will use RG32 format (that has two components). So it will read R and G from offsets 0 and 1, and store them on the destination. It will read B and A from offsets 2 and 3, but since the destination format doesn't have B and A those values doesn't really matter and they will not be actually stored anywhere. For RG32 to RGBA32, yea it will not work right now. I think this is about the best it can get with a single branch-less "catch-all" shader.

Djipi71:
Make many test on 4 World :) For me it work perfectly in 4X 3X and 2 X
ryujinx_capture_2023-02-20_11-35-12
ryujinx_capture_2023-02-20_12-06-07
ryujinx_capture_2023-02-21_10-16-03
ryujinx_capture_2023-02-21_10-50-51

marysaka (approved):

PR #4382, opened 2023-02-08
Replace BitConverter.ToString(bytes).Replace("-", "") with Convert.ToHexString(bytes)

turbedi:
Less code and allocations


AcK77 (approved):

TSRBerry (approved):
Thank you!

PR #4391, opened 2023-02-09
ui/discord: Implement per-game image icons

jduncanator:
This makes use of an [undocumented Discord RPC](https://github.com/Mastermindzh/tidal-hifi/pull/84) feature that allows specifying a full URL to an image for the large image key, rather than a pre-defined key from a limited set of uploaded images.

In this instance, we use a custom image redirection service that redirects to a games icon served by the Nintendo eShop directly. This service acts as a dumb redirector, looking up a mapping between title IDs and Nintendo eShop icon images. This avoids the need to host any potentially copyrighted images directly on servers operated by the Ryujinx project. The title ID to eShop mapping data is sourced from a yet unreleased game compatibility DB I'm working on, and is currently updated ~weekly~ daily by a CI process, but this can be run more frequently if needed.

The title ID database is currently populated with all titles from the US, EU, and JPN regions. There may be the occasional title that isn't present if it was launched exclusively in another region.

The source code for the redirection service is available here: https://github.com/jduncanator/cf-eshop-image-redirector

An example URL can be found here: https://img-eshop.worker.ryujinx.dev/titleid/01006A800016E000, which redirects to https://img-eshop.cdn.nintendo.net/i/08af58551a19df2a73ccb36f720388434a1965776b34675c6f69af3f93280330.jpg.

 

The Discord CDN proxies all of these requests, so the Nintendo servers are never hit directly by an end user: https://media.discordapp.net/external/bFQjtVycdy95Irpy1wB9Y24JlHyxGIlqoHYwkKZv9B8/https/img-eshop.worker.ryujinx.dev/titleid/010012101468C000

### DIFF
--- a/Ryujinx.Ui.Common/DiscordIntegrationModule.cs
+++ b/Ryujinx.Ui.Common/DiscordIntegrationModule.cs
@@ -66,7 +66,7 @@ namespace Ryujinx.Ui.Common
             {
                 Assets = new Assets
                 {
-                    LargeImageKey  = "game",
+                    LargeImageKey  = $"https://img-eshop.worker.ryujinx.dev/titleid/${titleId}",
                     LargeImageText = titleName,
                     SmallImageKey  = "ryujinx",
                     SmallImageText = Description,


IsaacMarovitz:
This is probably out of scope for this PR, but the Ryujinx logo displayed in Discord looks off to me. The purple background makes it hard to see and its lacking padding around the edge so it gets clipped weirdly. Maybe it would be worth changing the icon source for the logo too?

jduncanator:
> This is probably out of scope for this PR, but the Ryujinx logo displayed in Discord looks off to me. The purple background makes it hard to see and its lacking padding around the edge so it gets clipped weirdly. Maybe it would be worth changing the icon source for the logo too?

We should be able to do that directly in Discord by replacing the image for that image key. I think you'll find the Ryujinx logo in there is the very original one from ages ago, which we don't use any more 😅

IsaacMarovitz:
> We should be able to do that directly in Discord by replacing the image for that image key. I think you'll find the Ryujinx logo in there is the very original one from ages ago, which we don't use any more 😅 I reckon the purple comes from the fact the logo is transparent, and purple is the background color Discord use for behind the image.

Rather than using an image key we should use a URL. Just like the large icon, the small icon can be directed to a URL rather than a set uploaded one. We could point it to a logo on Git repo which would make updating it a breeze.

jduncanator:
> > We should be able to do that directly in Discord by replacing the image for that image key. I think you'll find the Ryujinx logo in there is the very original one from ages ago, which we don't use any more 😅 I reckon the purple comes from the fact the logo is transparent, and purple is the background color Discord use for behind the image.
> 
> Rather than using an image key we should use a URL. Just like the large icon, the small icon can be directed to a URL rather than a set uploaded one. We could point it to a logo on Git repo which would make updating it a breeze.

Why? The Ryujinx logo "never" changes, and we'd then need to host it somewhere else. What's the problem with keeping the Ryujinx logo using the official method, and using a image key?

TSRBerry:
It doesn't seem to be an undocumented feature:

> To use an external image via media proxy, specify the URL as the field's value when sending. You will only receive the `mp:` prefix via the gateway.

Source: https://discord.com/developers/docs/topics/gateway-events#activity-object-activity-asset-image

The description isn't very clear imo, but to me that sounds like external URLs are also expected to work here.

---

I think your PR is awesome, but I don't like we rely on an external (and closed source!) service in order for this to work correctly.

jduncanator:
The service is open-source, and can be hosted by anyone. I'm just providing a copy of it with data pre-populated to ease migration. More than happy for someone to suggest a viable alternative.

TSRBerry:
Afaik the mapping logic is not open source, only the redirect service is. Am I wrong?

jduncanator:
The mapping logic _is_ the redirect service. The only thing that isn't public is the process with which the Nintendo eShop is scraped for data, which I don't feel comfortable publishing. 

I'm more than happy to provide a feed of this data populating an instance of this service running under the control of the Ryujinx project.

TSRBerry:
We could also open a port on `localhost`, expose the game image from the game itself like that and send it to discord.
What do you think about that?

jduncanator:
> We could also open a port on `localhost`, expose the game image from the game itself like that and send it to discord.
> 
> What do you think about that?

This wouldn't work, as the images are proxied by Discord's CDN, so localhost wouldn't be accessible. 

Besides, even if it wasn't proxied, this would only work for Discord users who were actively running Ryujinx and had the game in their library, everyone else would just see a broken image, as localhost would refer to their local machine.

TSRBerry:
Ah right, my bad. The JSON or any other mapping method still sounds pretty good to me, what do you think?

Foul-Tarnished:
Could you make it so that it says : 

"The Legend of Zelda: Tears of Kingdom" (title)
"Playing on Ryujinx" (description) ?

So that we see the name of the game instead of ryu, in the contact list ?
Because discord only show the title, unless you actually click on the user account..

symsym-1629:
i'd love this ! btw it's just 1 line

TSRBerry:
well sure it's just one line, but you did read the discussion here, before commenting right?

jduncanator:
Not sure why this is waiting on author. If someone wants to PR a working alternative that fits better with everyone's expectations, more than happy for someone to do that.

I don't plan on making any further changes to this approach, as I personally feel like it strikes the correct balance of privacy preservation, ease of maintenance and frequency of data updates.

TSRBerry:
What do you dislike about the JSON approach?

jduncanator:
The JSON file would be several MB's (the current data stored in the KB is about 10MB total), it would need to be fetched from some central location in Ryujinx (yet another web request to an API to fetch _more_ data), and would need to be cached/stored locally on the users system somewhere, with some process to update this daily.

It's not that I dislike the JSON approach, it's just that it's so much more inconvenient, and requires much more overhead to do effectively the same thing as this does - the current approach is entirely stateless and does not require Ryujinx to store or request anything locally.

marysaka:
> The JSON file would be several MB's (the current data stored in KV is about 10MB total), it would need to be fetched from some central location in Ryujinx (yet another web request to an API to fetch _more_ data), and would need to be cached/stored locally on the users system somewhere, with some process to update this daily.
> 
> It's not that I dislike the JSON approach, it's just that it's so much more inconvenient, and requires more overhead and code to do effectively the same thing as this does - whilst the current approach is entirely stateless and does not require Ryujinx to store or request anything locally. Both would work fine, but I don't really see the downsides to the current approach either. I'm happy to address whatever your concerns are.
> 
> I spoke with gdkchan, and he's happy for the service to be hosted by the project, rather than a third party, if that's what the hang-up with the current approach is. I always intended for the project to run their own instance of the redirector, which is why I made sure the source was published for it.

We aren't using Cloudflare for Ryujinx's infra anymore and I don't plan that to change.

As a result, I would prefer the backend to be handled by us and feed from a JSON that could be available publicly as this could be used by other future services (like a game compat list)

jduncanator:
The code is all open source, happy for you guys to implement your own version and host it yourself.

TSRBerry:
> The code is all open source, happy for you guys to implement your own version and host it yourself.

That was the first point I raised with this PR. Not everything is open source for this.

PR #4401, opened 2023-02-10
Fix partial updates for textures.

riperiperi:
I was forcing some types of texture to partially update when investigating performance with games that stream in data, and noticed that partially loading texture data was really broken on both backends.

Fixes Vulkan partial texture upload by getting the correct expected size for the texture. Fixes partial texture upload on both backends for both Texture 2D Array and Cubemap using the wrong offset and always uploading to the first layer/level for a handle. 3D might also be affected.

This might fix textures randomly having incorrect data in games that render to it - jumbled in the case of OpenGL, and outdated/black in the case of Vulkan. This case typically happens in UE4 games.


gdkchan (approved):
lgtm, thanks. Always nice to see issues related to texture corruption fixed.

marysaka (approved):

PR #4404, opened 2023-02-10
Use index fragment shader output when dual source blend is enabled

gdkchan:
This was causing crashes on some games on macOS, but I don't think any other driver is affected.

Contributes to #4062 even though this one is not listed there.


nsklaus:
i have a branch where i experiment adding all your PRs regarding to upstreaming macos support.
i import your PRs before they are merged and so i get to test the progress being made early on.
after i fetched #4368 i had no particular problems, but now, adding this #4404 on top of it, i notice heavy corruption in fonts in skyrim. i'm sorry i cannot give more technical details. i attach a screenshot below. this wasn't happening before fetching #4404 
Screenshot 2023-02-11 at 21 41 50

marysaka (approved):
lgtm

Ryujinx.Graphics.Gpu/Shader/GpuChannelGraphicsState.cs
```diff
@@ -92,6 +92,11 @@ struct GpuChannelGraphicsState
         /// </summary>
         public Array8<AttributeType> FragmentOutputTypes;
 
+        /// <summary>
+        /// Indicates whether dual source blend is enabled.
+        /// </summary>
+        public readonly bool DualSourceBlendEnable;
```
riperiperi, line 98: Where is this set from right now? The other properties here are set by the SpecializationStateUpdater (they are also not readonly anymore), but that isn't modified in this PR.

The constructor for this class with all the attributes actually isn't called at all.

gdkchan: I think the code that would pass the value on the constructor was lost in the rebase (since it was removed on master)  and nothing was added to replace it. I added some code to set it to true on `UpdateBlendState` now.

It's a bit hard to test right now since the games that uses dual-source blend are also hitting other issues (Skyward Sword uses logical operations + blend enabled with integer targets, Metroid Prime Remastered has invalid resource errors).

iMonZ:
Hey, anything new here?

AcK77 (approved):

PR #4406, opened 2023-02-12
vulkan: Do not call vkCmdSetViewport when viewportCount is 0

marysaka:
This fix validation error [VUID-vkCmdSetViewport-viewportCount-arraylength](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdSetViewport-viewportCount-arraylength).

Testing is advised to check for any regressions.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/PipelineDynamicState.cs
+++ b/Ryujinx.Graphics.Vulkan/PipelineDynamicState.cs
@@ -21,7 +21,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         private Array4<float> _blendConstants;
 
-        public int ViewportsCount;
+        public uint ViewportsCount;
         public Array16<Viewport> Viewports;
 
         private enum DirtyFlags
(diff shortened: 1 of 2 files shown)


Ryujinx.Graphics.Vulkan/PipelineDynamicState.cs
```diff
@@ -21,7 +21,7 @@ struct PipelineDynamicState
 
         private Array4<float> _blendConstants;
 
-        public int ViewportsCount;
+        public uint ViewportsCount;
         public Array16<Viewport> Viewports;
```
gdkchan, line 25: Can this become private now? Like `private Array16<Viewport> _viewports;`

gdkchan (approved):
lgtm, thanks.

Ryujinx.Graphics.Vulkan/PipelineDynamicState.cs
marysaka: It's technically still read outside PipelineDynamicState but not written, so I guess I should make this a property?

gdkchan: If it's still used outside `PipelineDynamicState`, then just leave it as is.

riperiperi (approved):

PR #4408, opened 2023-02-12
vulkan: Enforce Vulkan 1.2+ at instance API level and 1.1+ at device level

marysaka:
This ensure we don't end up trying to initialize with anything currently incompatible.
We could drop the hard requirement on Vulkan 1.2 by not using ``VkPhysicalDeviceVulkan11Features`` or ``VkPhysicalDeviceVulkan12Features`` when not supported by the instance API.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
@@ -14,6 +14,9 @@ namespace Ryujinx.Graphics.Vulkan
     public unsafe static class VulkanInitialization
     {
         private const uint InvalidIndex = uint.MaxValue;
+        private static uint MinimalVulkanVersion = Vk.Version11.Value;
+        private static uint MinimalInstanceVulkanVersion = Vk.Version12.Value;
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
lgtm, thanks.

Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
```diff
@@ -239,6 +242,27 @@ internal static DeviceInfo[] GetSuitablePhysicalDevices(Vk api)
...
+            // We ensure that vkEnumerateInstanceVersion is present (added in 1.1).
+            // If the instance doesn't support it, no device is going to be 1.1 compatible.
+            if (api.GetInstanceProcAddr(instance, "vkEnumerateInstanceVersion") == IntPtr.Zero)
+            {
+                api.DestroyInstance(instance, null);
+
+                return Array.Empty<DeviceInfo>();
+            }
+
+            // We currently assumes that the instance is compatible with Vulkan 1.2
```
riperiperi, line 254:
```suggestion
            // We currently assume that the instance is compatible with Vulkan 1.2
```

riperiperi (approved):

PR #4410, opened 2023-02-13
Use SIMD acceleration for audio upsampler

lostromb:
Gives about a 2x speed boost for audio upsampling on x64 hardware. Might cause a slight speed regression on platforms where .Net doesn't support vectorization yet because it no longer force-inlines the filter kernel (though the JIT is smart enough to not even emit the SIMD IL when Vector.IsHardwareAccelerated is false, so there's no added code size in that case)

image

|       Method | InputSampleCount |     Mean |     Error |    StdDev |
|------------- |----------------- |---------:|----------:|----------:|
| UpsampleBaseline |               40 | 3.755 ms | 0.0749 ms | 0.0919 ms |
| UpsampleSIMD |               40 | 1.499 ms | 0.0291 ms | 0.0323 ms |
| UpsampleBaseline |               80 | 3.024 ms | 0.0578 ms | 0.0710 ms |
| UpsampleSIMD |               80 | 1.103 ms | 0.0172 ms | 0.0287 ms |
| UpsampleBaseline |              160 | 3.198 ms | 0.0619 ms | 0.1100 ms |
| UpsampleSIMD |              160 | 1.461 ms | 0.0286 ms | 0.0268 ms |


lostromb:
There's also a resampler available to us in `Concentus.Common`, though I haven't had the time to port SIMD acceleration to the public package yet so it's missing the faster sinc filter, but if we need higher-quality resampling it's there

Ryujinx.Audio/Renderer/Dsp/UpsamplerHelper.cs
```diff
@@ -70,16 +71,32 @@ public static void Upsample(Span<float> outputBuffer, ReadOnlySpan<float> inputB
...
                 Debug.Assert(bank.Length == FilterBankLength);
-                for (int j = 0; j < FilterBankLength; j++)
+
+                int curIdx = 0;
+                int stopIdx = FilterBankLength - (FilterBankLength % Vector<float>.Count);
+                if (Vector.IsHardwareAccelerated)
+                {
+                    // Do SIMD-accelerated block operations where possible.
+                    // Only about a 2x speedup since filter bank length is short
+                    while(curIdx < stopIdx)
```
gdkchan, line 87:
```suggestion
                    while (curIdx < stopIdx)
```

Ryujinx.Audio/Renderer/Dsp/UpsamplerHelper.cs
```diff
@@ -70,16 +71,32 @@ public static void Upsample(Span<float> outputBuffer, ReadOnlySpan<float> inputB
...
+                    {
+                        result += Vector.Dot(
+                            new Vector<float>(bank.AsSpan().Slice(curIdx, Vector<float>.Count)),
+                            new Vector<float>(state.History.AsSpan().Slice(curIdx, Vector<float>.Count)));
+                        curIdx += Vector<float>.Count;
+                    }
+                }
+
+                while(curIdx < FilterBankLength)
```
gdkchan, line 96:
```suggestion
                while (curIdx < FilterBankLength)
```

gdkchan (commented):
Wonder if it's possible to change the vectorized paths on the audio resamplers to also use `Vector<T>` or the new cross-platform 128-bit SIMD API to benefit both x64 and arm64.

lostromb:
If there are other audio kernels like this which use a basic loop over a buffer (FIR, Sinc, etc.) I'd be happy to add them to this PR

lostromb:
> Wonder if it's possible to change the vectorized paths on the audio resamplers to also use `Vector<T>` or the new cross-platform 128-bit SIMD API to benefit both x64 and arm64.

I was going to say it gets a bit complicated with the `Widen()` operation splitting things across two vectors when you convert int16 to float32, but it's not impossible. Doing so would automatically give you 256-bit support on x64-AVX2, and give you acceleration automatically in the future for things like SIMD on ARM. I will look into it some more...

Ryujinx.Audio/Renderer/Dsp/ResamplerHelper.cs
```diff
@@ -394,35 +395,33 @@ private unsafe static void ResampleDefaultQuality(Span<float> outputBuffer, Read
...
 
-                            for (; i < (sampleCount & ~3); i += 4)
-                            {
-                                Vector128<int> intInput0 = Sse41.ConvertToVector128Int32(pInput + (uint)i);
-                                Vector128<int> intInput1 = Sse41.ConvertToVector128Int32(pInput + (uint)i + 1);
-                                Vector128<int> intInput2 = Sse41.ConvertToVector128Int32(pInput + (uint)i + 2);
-                                Vector128<int> intInput3 = Sse41.ConvertToVector128Int32(pInput + (uint)i + 3);
+                        for (; i < (sampleCount & ~3); i += 4)
+                        {
+                            Vector128<int> intInput0 = Sse41.ConvertToVector128Int32(pInput + (uint)i);
```
lostromb, line 404: this is unchanged; just removed the indent from the multiple `fixed` statements.

marysaka (approved):
lgtm, thanks!

gdkchan (approved):
Code lgtm. Would be nice to get confirmation that the output was verified to match the scalar version.

lostromb:
I verified it manually but I should add a unit test too...

lostromb:
The new upsampler output matches the original code's output, but that output was
image

Investigating some more, maybe I'm just calling the function wrong...

Ryujinx.Tests/Audio/Renderer/Dsp/UpsamplerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+                // 440 hz sine wave with amplitude = 0.5f at input sample rate
+                inputBuffer[sample] = MathF.Sin((440 / inputSampleRate) * (float)sample * MathF.PI * 2f) * 0.5f;
+            }
+
+            int inputIdx = 0;
+            int outputIdx = 0;
+            while (inputIdx + inputBlockSize < numInputSamples)
+            {
+                int outputBufLength = (int)Math.Round((float)(inputIdx + inputBlockSize) * outputSampleRate / inputSampleRate) - outputIdx;
+                UpsamplerHelper.Upsample(
```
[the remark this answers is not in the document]
lostromb: fixed.

lostromb:
So.... there's definitely no regression, and I added test cases for the resampler and upsampler. Should be fine to merge now.

PR #4414, opened 2023-02-13
Update LibHac to 0.18.0

### LINKED ISSUE #4149: Dialogue box/bcat save directory creation regression
## Bug Report

Dialogue box/bcat save directory creation regression

If I'm missing anything or formatted improperly please let me know, this is my first bug report, thanks

when attempting to open the bcat directory of a game without one in the latest version of ryujinx, a dialogue box should open that facilitates the creation of that save directory. Instead, this does not occur, and a ResultFsTargetNotFound (2002-1002) error is thrown. This functionality works in older versions (such as LDN 2.4) and can carry over to the latest version. 

Right click a game without a bcat directory, choose "open bcat directory". Notably, If the user has no games with bcat directories, the first will succeed, though without opening a dialogue. Any other game will throw the error described when trying to create the directory. Any directories created by previous versions of ryujinx will open properly.

Ryujinx 1.1.480:
https://gfycat.com/grandunitedamphiuma

Ryujinx LDN 2.4:
https://gfycat.com/fr/coolfortunatefennecfox

### Log file

1.1480:
[Ryujinx_1.1.480_2022-12-19_13-17-53.log](https://github.com/Ryujinx/Ryujinx/files/10261962/Ryujinx_1.1.480_2022-12-19_13-17-53.log)

### Environment?

- Ryujinx version: 1.1.480</br>
- Game version: N/A</br>

- System Specs:
  - OS: *Windows 10*
  - CPU: *r7 3700x*
  - GPU: *gtx 1660 super 6gb*
  - RAM: *32GB*
- Applied Mods : N/A

### Additional context?

I believe another dialogue box for missing controllers is no longer working as well but I was not able to show this in LDN 2.4. I may be misremembering on this count, though the log does complain about it.

Steveice10:
Updates LibHac to the 0.18.0 release. One notable feature is that loading personalized ticket title keys is now supported with the right console keys dumped.

Main changes:
* Changes uses of ```.Ref()``` to ```.Ref```.
* Changes two uses of LibHac SpanHelpers to Ryujinx SpanHelpers due to ref compile errors.
* Updates ticket loading to check for null title key instead of gating to only common keys, as LibHac will now return null when the key cannot be read.
* Fix a bug in `EnsureApplicationSaveData` that wouldn't create a bcat save if any other bcat save already existed. (closes #4149)

Tested with a few games I have and seems to be working as expected.


Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
```diff
@@ -78,7 +76,7 @@ private ushort CalculateDeviceCrc()
```
Thealexbarney, line 76: These should be annotated with `UnscopedRef` because `this` escapes the scope of the method.
```suggestion
        [UnscopedRef]
```

Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
```diff
@@ -78,7 +76,7 @@ private ushort CalculateDeviceCrc()
 
         private ReadOnlySpan<byte> AsSpan()
         {
-            return MemoryMarshal.AsBytes(SpanHelpers.CreateReadOnlySpan(in this, 1));
+            return MemoryMarshal.AsBytes(SpanHelpers.CreateReadOnlySpan(ref this, 1));
```
Thealexbarney, line 79: Might as well simplify this a bit
```suggestion
            return SpanHelpers.AsReadOnlyByteSpan(in this);
```

Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
```diff
@@ -78,7 +76,7 @@ private ushort CalculateDeviceCrc()
...
         }
```
Thealexbarney, line 81:
```suggestion
        [UnscopedRef]
```

Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
```diff
@@ -1,7 +1,5 @@
-﻿using LibHac.Common;
-using Ryujinx.HLE.Utilities;
+﻿using Ryujinx.Common.Utilities;
```
Thealexbarney, line 1: Switch back to the other `SpanHelpers` class to get proper lifetime tracking
```suggestion
using LibHac.Common;
```

Thealexbarney (changes requested):
Looks mostly okay to me at a glance without actually running the code. The only issue is switching to the Ryujinx copy of the `SpanHelpers` class. The `UnscopedRef` attribute should be used instead of using the Ryujinx copy.

Observation outside the scope of this PR:
I noticed that the `scoped` keyword was added to [SpanHelpers](https://github.com/Ryujinx/Ryujinx/blob/e4f68592c3a6e51414e5f78eef096f21bf735eb1/Ryujinx.Common/Utilities/SpanHelpers.cs) when updating to .NET 7. This isn't correct because the input struct ref *does* escape the scope of the method, so the lifetime of the struct isn't properly tracked. The `scoped` keyword should be removed from these methods, and any callers should be annotated with `scoped` or `UnscopedRef` as necessary.

[Here's](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createreadonlyspan?view=net-7.0#remarks) an explanation in the .NET docs under the remarks section.

Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
Steveice10: Ryujinx does the exact same with other struct members elsewhere in the code; this just matches these two cases to the rest of the codebase.

Since these two uses aren’t really specifically related to LibHac at all but just happened to use it’s helper I would argue if there’s an issue there would need to be a larger fix of the Ryujinx SpanHelpers outside of this PR, and aligning these uses onto the same SpanHelpers would suffice for now.

gdkchan: In general we don't want to add a dependency to Libhac in the code if we don't need to, as if we want to use the code in a separate project or standalone we don't want to depend on Libhac just for minor utility methods.

Thealexbarney: The `SpanHelpers` class in Ryujinx is (or was) just a copy of the class in Libhac, so it matter too much which one is used outside of possible naming collisions. I only suggested using the one in Libhac because it doesn't have the incorrect `scoped` parameters, and fixing the one in Ryujinx would probably require properly annotating other functions throughout Ryujinx

gdkchan: I think the right thing to do here is fixing the one on Ryujinx, or removing the one on Ryujinx and using the Libhac method everywhere. In both cases it would require all existing uses of those methods. My preference would be of course keeping and fixing the one on Ryujinx due to the reasons I mentioned before. In any case, if this is a existing issue and the code is working as is, I think it would be fine to fix in a separate PR.

Thealexbarney: The code in question in the PR is safe as it currently is. It's just bypassing the compiler's lifetime tracking so it's up to the developer to make sure that the code's safe. It should be fine to commit it as-is and do the fix in a separate PR.

Steveice10:
Rebased via GitHub tool

Ryujinx.HLE/HOS/Services/Mii/Types/StoreData.cs
marysaka: I would really prefer that those helpers be from one dependency.

However, I do not want to bring LibHac dependency on non HLE parts, this part would need to be spitted out of it.

Not to mention it could be beneficial for the .NET ecosystem in general.

In the meantime, @Steveice10 can you revert the changes here and add a TODO comment stating the lifetime issue?

Steveice10: > I would really prefer that those helpers be from one dependency.
> 
> However, I do not want to bring LibHac dependency on non HLE parts, this part would need to be spitted out of it.
> 
> Not to mention it could be beneficial for the .NET ecosystem in general.
> 
> In the meantime, @Steveice10 can you revert the changes here and add a TODO comment stating the lifetime issue?

This change is actually removing the LibHac dependency. Without this change it won't compile because of LibHac changes.

marysaka: urgh right

> > I would really prefer that those helpers be from one dependency.
> > However, I do not want to bring LibHac dependency on non HLE parts, this part would need to be spitted out of it.
> > Not to mention it could be beneficial for the .NET ecosystem in general.
> > In the meantime, @Steveice10 can you revert the changes here and add a TODO comment stating the lifetime issue?
> 
> This change is actually removing the LibHac dependency. Without this change it won't compile because of LibHac changes.

Okay, I guess it's fine then.

It would be nice to open an issue for that @Thealexbarney, I will try to handle it when I get some times (or someone can look into that).

PR #4418, opened 2023-02-13
vulkan: Clean up MemoryAllocator

marysaka:
This started as an attempt to remove `vkGetPhysicalDeviceMemoryProperties` in `FindSuitableMemoryTypeIndex` (As this could have some overhead and shouldn't change at runtime) and turned in a little bigger cleanup.


gdkchan (approved):
lgtm, thanks.

marysaka:
> LGTM, I wonder if any drivers had a really slow implementation of this. woozy_face

Probably no drivers but this avoid querying something static every time...

I do wonder if we could improve the MemoryAllocator a bit with some newer extensions (for example memory budget), in that case we will likely need to query some stuffs at allocation time again I guess.

PR #4419, opened 2023-02-14
vulkan: Respect VK_KHR_portability_subset vertex stride alignment

marysaka:
We were hardcoding alignment to 4, but by specs it can be any values that is a power of 2.

This also enable VK_KHR_portability_subset if present as per specs requirements. ([VUID-VkDeviceCreateInfo-pProperties-04451](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451))

### DIFF
--- a/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
@@ -639,7 +639,7 @@ namespace Ryujinx.Graphics.Vulkan
             PrintGpuInformation();
         }
 
-        public bool NeedsVertexBufferAlignment(int attrScalarAlignment, out int alignment)
+        internal bool NeedsVertexBufferAlignment(int attrScalarAlignment, out int alignment)
         {
             if (Capabilities.VertexBufferAlignment > 1)
             {
--- a/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
@@ -234,10 +234,15 @@ namespace Ryujinx.Graphics.Vulkan
             Api.GetPhysicalDeviceFeatures2(_physicalDevice, &features2);
 
             var portabilityFlags = PortabilitySubsetFlags.None;
+            uint vertexBufferAlignment = 0;
 
             if (usePortability)
             {
-                portabilityFlags |= propertiesPortabilitySubset.MinVertexInputBindingStrideAlignment > 1 ? PortabilitySubsetFlags.VertexBufferAlignment4B : 0;
+                if ((propertiesPortabilitySubset.MinVertexInputBindingStrideAlignment % 2) != 0)
+                {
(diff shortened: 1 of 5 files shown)


Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -234,10 +234,15 @@ private unsafe void LoadFeatures(string[] supportedExtensions, uint maxQueueCoun
             Api.GetPhysicalDeviceFeatures2(_physicalDevice, &features2);
 
             var portabilityFlags = PortabilitySubsetFlags.None;
+            uint vertexBufferAlignment = 0;
 
             if (usePortability)
             {
-                portabilityFlags |= propertiesPortabilitySubset.MinVertexInputBindingStrideAlignment > 1 ? PortabilitySubsetFlags.VertexBufferAlignment4B : 0;
+                if ((propertiesPortabilitySubset.MinVertexInputBindingStrideAlignment % 2) != 0)
```
gdkchan, line 241: Why is that check needed?

AcK77 (approved):

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
marysaka: Per spec it should be a power of 2, I can remove the check if you want tho

gdkchan: This is just checking if it's a multiple of 2, not a power of 2. Also I don't think we need to handle cases where the driver is passing a invalid value.

Also I think it would make more sense for the default alignment to be 1, not 0. An alignment of 0 bytes doesn't make sense.

marysaka: Oops that's right, my bad here.

For alignment sure I can change it to 1 that would make more sense yeah.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -638,9 +641,9 @@ public void Initialize(GraphicsDebugLevel logLevel)
 
         public bool NeedsVertexBufferAlignment(int attrScalarAlignment, out int alignment)
```
gdkchan, line 642: Existing issue, but this method should be internal rather than public.

gdkchan (approved):
lgtm minus comment.

Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
gdkchan: There's a few others too like `TopologyRemap` and `TopologyUnsupported`.

marysaka: Yeah there is also ton of refactoring that could be done in VulkanInitialization too, I think I will tackles the others later.

PR #4427, opened 2023-02-15
Logging: Redirect StdErr into logging system

merryhime:
Redirect stderr output into logging system.

This allows for easier mvk/mesa debugging since stderr will now show up in log files.

Extracted from #4227.

### DIFF
--- a/Ryujinx.Common/SystemInterop/StdErrAdapter.cs
+++ b/Ryujinx.Common/SystemInterop/StdErrAdapter.cs
@@ -68,7 +68,7 @@ namespace Ryujinx.Common.SystemInterop
         }
 
         [LibraryImport("libc", SetLastError = true)]
-        private static partial int dup2 (int fd, int fd2);
+        private static partial int dup2(int fd, int fd2);
 
         [LibraryImport("libc", SetLastError = true)]
         private static unsafe partial int pipe(int* pipefd);
@@ -84,7 +84,7 @@ namespace Ryujinx.Common.SystemInterop
                 }
                 else
                 {
-                    throw new ();
+                    throw new();
                 }
             }
         }
--- a/Ryujinx.Common/SystemInterop/StdErrAdapter.cs
+++ b/Ryujinx.Common/SystemInterop/StdErrAdapter.cs
@@ -55,8 +55,11 @@ namespace Ryujinx.Common.SystemInterop
         {
             if (_disposable)
             {
-                _pipeReader?.Close();
-                _pipeWriter?.Close();
+                if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
+                {
+                    _pipeReader?.Close();
+                    _pipeWriter?.Close();
+                }
 
                 _disposable = false;
             }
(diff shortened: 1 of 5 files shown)


marysaka (changes requested):
I would prefer not importing Mono.Unix again as this force us to have another native dependency again :sweat_smile:

merryhime:
@marysaka removed dependency on Mono.Unix

marysaka (approved):

Ryujinx.Common/SystemInterop/StdErrAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        [LibraryImport("libc", SetLastError = true)]
+        private static partial int dup2 (int fd, int fd2);
```
riperiperi, line 71:
```suggestion
        private static partial int dup2(int fd, int fd2);
```

Ryujinx.Common/SystemInterop/StdErrAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+            Span<int> pipefd = stackalloc int[2];
+            fixed (int* ptr = pipefd)
+            {
+                if (pipe(ptr) == 0)
+                {
+                    return (pipefd[0], pipefd[1]);
+                }
+                else
+                {
+                    throw new ();
```
riperiperi, line 87:
```suggestion
                    throw new();
```

Ryujinx.Common/SystemInterop/UnixStream.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Common.SystemInterop
+{
+    public partial class UnixStream : Stream, IDisposable
```
TSRBerry, line 7:
```suggestion
    [UnsupportedOSPlatform("windows")]
    public partial class UnixStream : Stream, IDisposable
```
Just to make sure we never use it for Windows.

Ryujinx.Common/SystemInterop/StdErrAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+        public StdErrAdapter()
+        {
+            if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
+            {
+                RegisterPosix();
+            }
+        }
+
+        [SupportedOSPlatform("linux")]
+        [SupportedOSPlatform("macos")]
```
TSRBerry, line 26: Could you move these to the class?

TSRBerry (commented):
Just have a few suggestions, but lgtm otherwise!

Ryujinx.Common/SystemInterop/UnixStream.cs
merryhime: sgtm

Ryujinx.Common/SystemInterop/StdErrAdapter.cs
merryhime: Would rather leave this be, and have this check be encapsulated in this class.

TSRBerry: Fine for me as well!

AcK77 (approved):

TSRBerry (approved):
Thank you!

PR #4394, opened 2023-02-09
Allow partially mapped textures with unmapped start

gdkchan:
Currently, we allow textures to be partially mapped (that is, not all pages of the texture are mapped). This is required to support sparse textures. However, it did not support having the start of the texture unmapped. For this case, it just assumed the texture didn't exist. This PR updates the code to allow partially mapped textures with a unmapped start region. For the lookup, it uses the address of the first page that is mapped.

Fixes broken lighting on Metroid Prime Remastered.

Before:
image
After:
image

Testing is welcome. Would be nice to know if there's any performance regression (or increase of stuttering/memory usage) since it is creating a large texture now, which was not being created before.

PR #4430, opened 2023-02-16
Validate dimensions before creating texture

gdkchan:
There are some cases where texture pool entries are completely invalid and would contain garbage. Those cases would usually have a invalid GPU virtual address too which would prevent the texture from being created. But since #4394, an attempt is made to create the texture even if the start address is unmapped ("invalid"), to allow sparse textures to work. This change is adds additional validation to prevent an attempt of calculating the texture size with invalid dimensions, which causes exception.

This fixes a regression caused by the linked PR on SSBU while using ARCropolis mods.


marysaka (approved):
lgtm

riperiperi (approved):
Fair enough.

Issue #4428, opened 2023-02-15
[Feature Request] Missing instructions on how to build on macos


lucicam:
### Overview

There are no instructions on the particularities of building on macos (especially for M1/M2)

### Smaller details

Building with just the command in the readme fails on macos.

### Nature of request

Documentation request

### Why would this feature be useful?

Developers wanting to help with the emulator from a mac

gdkchan:
Related: #4371.

lucicam:
looked more into in in the meantime, seems like it doesn't build if there are spaces in the path where the project is cloned, maybe it could be specified in the documentation. I'll close the issue, thank you.

gdkchan:
It might be better to leave it open to track doing something about this, since it's not the first time we get this question.

lucicam:
Indeed. I made a quick pull request to specify this in the readme https://github.com/Ryujinx/Ryujinx/pull/4434

PR #4434, opened 2023-02-16
Update README.md

lucicam:
Added note that the path should not contain spaces in order for the build to succeed. Seems like more people had this issue so it should at least be written in the readme. Related to #4428

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Either use `git clone https://github.com/Ryujinx/Ryujinx` on the command line to
 
 ### Step 3
 
-To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command (the path must not contain spaces):
+To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command:
 `dotnet build -c Release -o build`
 the built files will be found in the newly created build directory.
 
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Either use `git clone https://github.com/Ryujinx/Ryujinx` on the command line to
 
 ### Step 3
 
-To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command:
+To build Ryujinx, open a command prompt inside the project directory. You can quickly access it on Windows by holding shift in File Explorer, then right clicking and selecting `Open command window here`. Then type the following command (the path must not contain spaces):
 `dotnet build -c Release -o build`
 the built files will be found in the newly created build directory.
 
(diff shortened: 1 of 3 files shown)


gdkchan:
Is having spaces really a problem when using `dotnet build`? I thought it was an issue with the macOS build script.

siegmund-heiss-ich:
There is a problem at command execution `codesign --entitlements <Path>`. I think it would be better to implement a solution for this directly in the code rather than telling the user how a working path should look like.

I think a path with spaces would work if we had a backslash in front of every space.
We could check the path before command execution.

EliEron:
> Is having spaces really a problem when using `dotnet build`? I thought it was an issue with the macOS build script.

Building with spaces works fine on Windows & Linux, however while verifying this I did discover that there is one issue with the command in the readme. The ```-o``` option is no longer supported for building or publishing in the newest SDK [(#15607)](https://github.com/dotnet/sdk/issues/15607) so that option should be removed from the readme.

lucicam:
Actually, it's not the shell script which fails (at least not initially)

```
  Ryujinx.Ava -> /private/tmp/test build/Ryujinx/build/Ryujinx.Ava.dll
  /private/tmp/test: No such file or directory
/private/tmp/test build/Ryujinx/Ryujinx.Ava/Ryujinx.Ava.csproj(16,5): error MSB3073: The command "codesign --entitlements '/private/tmp/test build/Ryujinx/Ryujinx.Ava/../distribution/macos/entitlements.xml/' -f --deep -s - /private/tmp/test build/Ryujinx/build/Ryujinx.Ava" exited with code 1.

Build FAILED.
```

It fails because the command replaces the path but it contains spaces. I tried adding "&quot;" before and after the path, but doesn't seem to work

lucicam:
> > Is having spaces really a problem when using `dotnet build`? I thought it was an issue with the macOS build script.
> 
> Building with spaces works fine on Windows & Linux, however while verifying this I did discover that there is one issue with the command in the readme. The `-o` option is no longer supported for building or publishing in the newest SDK [(#15607)](https://github.com/dotnet/sdk/issues/15607) so that option should be removed from the readme.

This should be a problem in linux also, but probably it's not because there is no command which has spaces in the .csproj files

lucicam:
I looked a bit into it and it seems to be some problem on linux/macos with msbuild, because I tried replacing the spaces with '\ ' (escaping the space) but it replaces it with '/' ... https://github.com/dotnet/msbuild/issues/3468. I also tried to replace with "%5C" but still it replaces them with '/' and from that link, seems to be a general issue.
I'm not too experienced with msbuild but it also seems like I can't really run shell scripts or command with pipe in them, so I'm not sure how we can fix this unless this issue is actually fixed on msbuild side.

TimeZlicer:
```
<Exec Command="codesign --entitlements '$(ProjectDir)..\distribution\macos\entitlements.xml' -f --deep -s $(SigningCertificate) '$(TargetDir)$(TargetName)'" />
```
In `Ryujinx.Ava.csproj`, both paths need to be in single quote(')
This should resolve the error for `dotnet build -c Release`

---

`create_app_bundle.sh`
`create_macos_release.sh`
The paths in these scripts should be in double quote("), for example
```
dotnet publish -c Release -r osx-arm64 -o "$TEMP_DIRECTORY"/publish_arm64 $DOTNET_COMMON_ARGS Ryujinx.Ava  
dotnet publish -c Release -r osx-x64 -o "$TEMP_DIRECTORY"/publish_x64 $DOTNET_COMMON_ARGS Ryujinx.Ava
```

---

Hope the above helps

lucicam:
Closed because TimeZlicer made a PR fixing in multiple places. I can confirm it works with single quotes... I could have sworn I tried this exact thing but I probably put them in the wrong places -_-

PR #4438, opened 2023-02-16
Perform bounds checking before list indexer to avoid frequent exceptions

jhorv:
IdList's TryGetValue() method was using the _list indexer without checking bounds first. This caused very frequent ArgumentOutOfRangeExceptions in the VS Output window, as well as a performance hit. This adds a range check so that most calls will succeed without an exception being thrown.


riperiperi (commented):
If you do this, I don't think you need the try/catch anymore. I do agree that this is probably worth it, as null bindings can easily trigger this.

gdkchan:
I believe this method can be called from multiple threads, so it must be thread safe. The reason I made it catch both exceptions is that I was afraid it could access out of bounds after the index was already checked internally, if the internal array was changed from another thread after the check has already been done, I think it could throw `IndexOutOfRangeException`. This is not exactly ideal, maybe there's a better thread safe way to check this.

Also some of those invalid handles might be caused by an attempt to get the index buffer for a non-indexed draw, which is not really necessary, so we should probably avoid that too.

riperiperi:
Maybe catching the only the index exception will be a good compromise. The fact that this is a list means that something might go wrong when changing size too, but that would affect it right now.

jhorv:
I'm trying to think of a safer/simpler way of doing this... I see `IdList.Add()` allocates an `int32` ID and returns it, and that gets stored as a `ulong` `BufferHandle`. What if we changed `IdList` to use a `ConcurrentDictionary<ulong, T>`, where the key is assigned by incrementing a "next ID" counter via `Interocked.Increment()`? I don't know much about low-level graphics programming, but if my hasty math is right, allocating one ID every nanosecond would still take 585 years to roll over.

It seems possible the perf negatives of `ConcurrentDictionary` would be offset by no longer having to scan the `List<T>` for empty slots.

I'd be happy to re-do the PR with that approach, just say the word.

gdkchan:
If a dictionary was used it could just use the Vulkan handle as a key. The point of using a list is that access is O(1).

jhorv:
I did some benchmarking before I saw your last comment. If I had seen what you wrote about lookups being more important, I probably wouldn't have bothered. But it's too late - I benchmarked, and the results are below. :) The benchmark code is also attached as a LINQPad file. The test did adds, removes, then lookups. The ConcurrentDictionary-based version was faster but allocated almost 10x more memory.

I don't see much point in benchmarking lookups only, hard to get faster than a List<T>/array index.

|                           Method | RngSeed | ItemCount | AddRemoveRounds |     Mean |   Error |  StdDev |    Gen0 |   Gen1 | Allocated |
|--------------------------------- |-------- |---------- |---------------- |---------:|--------:|--------:|--------:|-------:|----------:|
|              Test_IdList_ListOfT |       7 |      1000 |               3 | 367.3 μs | 2.62 μs | 2.32 μs |  2.9297 | 0.4883 |  48.81 KB |
| Test_IdList_ConcurrentDictionary |       7 |      1000 |               3 | 285.7 μs | 1.07 μs | 1.00 μs | 26.8555 | 8.3008 | 441.24 KB |

[List_vs_ConcurrentDictionary.linq.txt](https://github.com/Ryujinx/Ryujinx/files/10762861/List_vs_ConcurrentDictionary.linq.txt)

Ryujinx.Graphics.Vulkan/IdList.cs
```diff
@@ -80,8 +80,16 @@ public bool TryGetValue(int id, out T value)
 
             try
             {
-                value = _list[id];
-                return value != null;
+                if (id >= 0 && id < _list.Count)
```
gdkchan, line 83:
```suggestion
                if ((uint)id < (uint)_list.Count)
```
Should be a little bit faster since it does one comparison rather than 2.

jhorv: Sorry I didn't know I could just "Commit suggestion".

gdkchan (approved):
lgtm, thanks. Not sure if the JIT would be able to eliminate the bounds check on the list indexer when a explicit bounds check is done before accessing it... If it can't it could be a minor performance regression on cases where the `id` is valid.

riperiperi (approved):
LGTM. I think the largest impact I saw from this was when running in debug or with the profiler.

Issue #3671, opened 2022-09-05
[Feature Request] Allow the Main Window to remember its sizing and position upon closing


Daisouji:
The title. I think it is a good feature to keep consistent w/ other platform programs.

theboy181:
BUMP!

HaizenTrist:
Hi, I'm new to open-source contribution but I think I can handle this. Can I take the issue?

HaizenTrist:
Made a pull request, though not sure if I did it correctly as this is my first issue.

Daisouji:
addressed in #4657

MutantAura:
We'll close this after it's been merged :)

MutantAura:
Closed by https://github.com/Ryujinx/Ryujinx/pull/4657

PR #4439, opened 2023-02-16
Remember main window position and size on exit

HaizenTrist:
https://github.com/Ryujinx/Ryujinx/issues/3671


HaizenTrist:
https://github.com/Ryujinx/Ryujinx/issues/3671

Daisouji:
It remember position and size. It doesn't remember windowstate (Maximized)

Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -146,6 +171,11 @@ public UiSection()
                 LanguageCode      = new ReactiveObject<string>();
                 ShowConsole       = new ReactiveObject<bool>();
                 ShowConsole.Event += static (s, e) => { ConsoleHelper.SetConsoleWindowState(e.NewValue); };
+                WindowSizeWidth   = new ReactiveObject<int>();
+                WindowSizeHeight   = new ReactiveObject<int>();
```
gdkchan, line 175:
```suggestion
                WindowSizeHeight  = new ReactiveObject<int>();
```

HaizenTrist:
Good point I forgot about that, the new commits should check for maximize state as well now.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -152,8 +152,10 @@ private MainWindow(Builder builder) : base(builder.GetRawOwnedObject("_mainWin")
             int monitorWidth  = monitor.Geometry.Width  * monitor.ScaleFactor;
             int monitorHeight = monitor.Geometry.Height * monitor.ScaleFactor;
 
-            DefaultWidth  = monitorWidth  < 1280 ? monitorWidth  : 1280;
-            DefaultHeight = monitorHeight < 760  ? monitorHeight : 760;
+            DefaultWidth  = monitorWidth  < 1280 ? monitorWidth  : ConfigurationState.Instance.Ui.WindowSizeWidth;
+            DefaultHeight = monitorHeight < 760  ? monitorHeight : ConfigurationState.Instance.Ui.WindowSizeHeight;
+            Move(ConfigurationState.Instance.Ui.WindowPositionX,ConfigurationState.Instance.Ui.WindowPositionY);
+            if (ConfigurationState.Instance.Ui.WindowMaximized == true) { Maximize(); };
```
gdkchan, line 158:
```suggestion
            if (ConfigurationState.Instance.Ui.WindowMaximized == true)
            {
                Maximize();
            }
```

gdkchan: Not sure if the `== true` is needed either.

HaizenTrist: My bad, it does work without, I just wrote `== true` out of habit. Made the change.

Ryujinx/Ui/MainWindow.cs
```diff
@@ -152,8 +152,13 @@ private MainWindow(Builder builder) : base(builder.GetRawOwnedObject("_mainWin")
             int monitorWidth  = monitor.Geometry.Width  * monitor.ScaleFactor;
             int monitorHeight = monitor.Geometry.Height * monitor.ScaleFactor;
 
-            DefaultWidth  = monitorWidth  < 1280 ? monitorWidth  : 1280;
-            DefaultHeight = monitorHeight < 760  ? monitorHeight : 760;
+            DefaultWidth  = monitorWidth  < 1280 ? monitorWidth  : ConfigurationState.Instance.Ui.WindowSizeWidth;
+            DefaultHeight = monitorHeight < 760  ? monitorHeight : ConfigurationState.Instance.Ui.WindowSizeHeight;
+            Move(ConfigurationState.Instance.Ui.WindowPositionX,ConfigurationState.Instance.Ui.WindowPositionY);
```
gdkchan, line 157:
```suggestion
            Move(ConfigurationState.Instance.Ui.WindowPositionX, ConfigurationState.Instance.Ui.WindowPositionY);
```
Sorry, I missed that one before.

gdkchan:
Would be nice to implement this on Avalonia UI too (Ryujinx.Ava) since the plan is migrating to that one in the future.

CeruleanSky:
If not already doing so somewhere in the code base, might want to double check to ensure that the win pos and size coordinates are not out of bounds of the current resolution.

I haven't tested this, but I often see that when a person plugs a second display(usually a laptop), open and closes a program on the extended screen, then unplugs their display, that window positions may be saved off screen or sizes that are applicable to multi monitor are no longer valid when having just one screen.

a234106261 (approved):

a234106261:
Simple, but useful change to make Ryujinx easier to use

Daisouji (approved):
It's working well,  we could do a port to Avalonia in another PR.

HaizenTrist:
Got a bit busy, but I'd be willing to add checks like a second monitor and try a port after taking some time to learn Avalonia.

Got a bit busy, but I'd be willing to add checks like a second monitor and try a port after taking some time to learn Avalonia.

Also, sorry if this is a silly question but is there something I need to do to merge or does this just need approval? Since this is my first time using GitHub I'm not familiar with the merging process.

MetrosexualGarbodor:
It depends on the project. Here, PRs require 2 approvals from maintainers in order for them to be merged. If you're done making changes to this one, just say so and the developers will review whenever they have some time to do it.

HaizenTrist:
Added checks to bring window size and position within bounds (like in the case of missing monitors), but I only have a smaller second monitor to test with. Would appreciate testing from someone with a multi-monitor setup.

MutantAura:
> Added checks to bring window size and position within bounds (like in the case of missing monitors), but I only have a smaller second monitor to test with. Would appreciate testing from someone with a multi-monitor setup. Also would like to merge this if there are no bugs/other features to add.

Do you want to cherry pick my avalonia and GTK method commits from: https://github.com/Ryujinx/Ryujinx/pull/4657 as I dont think you've added support for Ryujinx.Ava. Or would you prefer I merge your commits into my branch. Don't mind which, just seeing which we want to keep open.

Daisouji:
I think a good idea is to combine both into 1 PR.

HaizenTrist:
I don't have a preference on which PR to keep open but do think it'd be a good idea to merge into one. Maybe keep this open as its the first? But I'm not sure how a merging between PRs works.

MutantAura:
You should just be able to grab my branch, merge it into this branch and then push.

HaizenTrist:
I added some checks to GTK earlier to ensure the window wouldn't be drawn larger than the monitor or out of bounds, I don't think Ava has similar checks yet. I've never worked with Ava, is this something you could add?

AcK77:
Superseded by #4657

Issue #4420, opened 2023-02-14
[Regression] Performance Regression in Super Mario 3D All Stars : Super Mario Galaxy 


Bjorn29512:
### Description of Issue

Super Mario 3D All Stars : Super Mario Galaxy experienced significant performance regression in November 2022 with ryujinx version 1.1.357. (Vulkan)
![image](https://user-images.githubusercontent.com/110204265/218707960-6625c13f-6feb-4fd0-9331-02fbc01d056d.png)

Attachments: 

v1.1.356
![image](https://user-images.githubusercontent.com/110204265/218707855-160a215f-8230-4746-9930-4d3c21d0807a.png)

v1.1.357
![image](https://user-images.githubusercontent.com/110204265/218707906-ec4c1f13-fc14-499a-9d95-83092cbfa5c2.png)

### Reproduction Steps

Run SM galaxy and load into the game.

### Log File

[Ryujinx_1.1.619_2023-02-14_15-56-13.log](https://github.com/Ryujinx/Ryujinx/files/10731408/Ryujinx_1.1.619_2023-02-14_15-56-13.log)

### OS

Windows 11

### Ryujinx version

1.1.619

### CPU

i5-12400f

### GPU

GTX 1660 Super

### RAM

32 GB

### Applied Mods?

- [ ] Yes

Squall-Leonhart:
As the star bits land and turn into their spinning bouncing forms, Fifo goes up and fps goes down.

PR #4442, opened 2023-02-18
Update range for remapped sparse textures instead of recreating them

riperiperi:
Sparsely mapped textures are useful when a game wants texture data to cover a large area (say, an entire level), but loading the whole thing would take up too much memory. You can load and unload parts of the texture by mapping physical pages into the texture's virtual memory allocation in GPU address space, and unmapping them when you're finished. This way, you only pay for the pages you use.

Ryujinx's way of handling sparsely mapped textures was to delete their pool references entirely when _any_ mapping change happened, as the texture cache would simply recreate it the next time it was used. This works, but is slow, especially if your game has mapped several 48MB textures in this way, and changes mappings mid-game.

This PR adds the ability to migrate a texture to a new set of physical ranges. The existing texture data will remain intact, and any regions where the mapping has changed will be marked as dirty and synchronized on use. Note that it will only happen for the earliest GPU VA pool reference on the texture, any other pool references will be deleted as before.

Greatly improves stuttering going through doors on Metroid Prime: Remastered. Does not eliminate it entirely, because the game uses ASTC textures, which must be decoded before uploading on desktop, and are loaded around the same time a door is opened.

## Other fixes

This PR also fixes a few issues with partial updates and subresource tracking:
- Fixed an issue where handles for 3d texture slices would have a very small size, as it would only count to the beginning of the next slice (which may have been interleaved with this one)
- Partial updates are now enabled when using a texture and _any_ subresource is not modified, rather than only when a handle is unmapped. May improve performance in games with texture streaming.
- Subresource tracking (layers/levels) are now enabled on sufficiently large 3D and array textures to avoid impacts from synchronizing the whole thing. They are also enabled if the texture is partially mapped. This reduces stuttering from texture updates.
  - Greatly reduces FIFO% in NieR: Automata when travelling between areas. Framerate still dips due to Vulkan buffer memory stuff and cb0 accesses that can't be eliminated.
  - May affect texture streaming performance in Xenoblade. Who knows in what way.
- Gets blocks of slices from the source texture rather than one slice at a time, so that 3D textures don't copy the same interleaved data range each slice (potentially up to 16x the data)
- Partial uploads of 3D textures no longer allocate arrays for the full depth of the texture instead of one slice.

## Areas for improvement

A few things could still be improved:
- Handles are fully recreated every time a sparse mapping is changed, not just the handles that overlapped the ranges that changed mapping. Improving this could be very difficult.
- InheritHandles and ReplaceHandles are `O(n*m)` _best case_. This was the case for InheritHandles before, but it is more of a problem with sparse textures as they could be remapped whenever and the game assumes it's free.
- Texture ranges can't be updated if they have any views, are a view themselves, or have any form of copy dependency. This is because the texture on the other side could have a pool reference with a different GPU VA. I'm not sure if there's a way to handle this, or if there's any need to. It's very unlikely that a sparsely mapped texture will be rendered to, and that's typically what creates views. There's always swizzle and format alias, though...
- The backend could be made aware of layers/levels that are fully unmapped and do sparse mapping for real, though this could get quite complicated.

NOTE: draft for some small self-review cleanup and hopefully game testing.

This changes quite a lot regarding unmapped textures and data uploads, so it would be a good idea to test this in a variety of games.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -36,6 +36,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             public TexturePool Pool;
             public int ID;
+            public ulong GpuAddress;
         }
 
         private GpuContext _context;
@@ -1502,11 +1503,12 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         /// <param name="pool">The texture pool this texture has been added to</param>
         /// <param name="id">The ID of the reference to this texture in the pool</param>
-        public void IncrementReferenceCount(TexturePool pool, int id)
+        /// <param name="gpuVa">GPU VA of the pool reference</param>
+        public void IncrementReferenceCount(TexturePool pool, int id, ulong gpuVa)
         {
             lock (_poolOwners)
             {
-                _poolOwners.Add(new TexturePoolOwner { Pool = pool, ID = id });
+                _poolOwners.Add(new TexturePoolOwner { Pool = pool, ID = id, GpuAddress = gpuVa });
             }
             _referenceCount++;
 
@@ -1610,9 +1612,21 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             lock (_poolOwners)
             {
+                ulong address = 0;
+
                 foreach (var owner in _poolOwners)
                 {
-                    owner.Pool.QueueUpdateMapping(this, owner.ID);
+                    if (address == 0 || address == owner.GpuAddress)
+                    {
+                        address = owner.GpuAddress;
+
+                        owner.Pool.QueueUpdateMapping(this, owner.ID);
+                    }
+                    else
+                    {
+                        // If there is a different GPU va mapping, prefer the first and delete the others.
+                        owner.Pool.ForceRemove(this, owner.ID, true);
+                    }
                 }
 
                 _poolOwners.Clear();
--- a/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -731,6 +731,8 @@ namespace Ryujinx.Graphics.Gpu.Image
             height = Math.Max(height >> level, 1);
             depth = Math.Max(depth >> level, 1);
 
+            int sliceDepth = single ? 1 : depth;
+
             SpanOrArray<byte> result;
 
             if (Info.IsLinear)
@@ -751,7 +753,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     width,
                     height,
                     depth,
-                    single ? 1 : depth,
+                    sliceDepth,
                     levels,
                     layers,
                     Info.FormatInfo.BlockWidth,
@@ -775,7 +777,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     Info.FormatInfo.BlockHeight,
                     width,
                     height,
-                    depth,
+                    sliceDepth,
                     levels,
                     layers,
                     out byte[] decoded))
@@ -787,7 +789,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                 if (GraphicsConfig.EnableTextureRecompression)
                 {
-                    decoded = BCnEncoder.EncodeBC7(decoded, width, height, depth, levels, layers);
+                    decoded = BCnEncoder.EncodeBC7(decoded, width, height, sliceDepth, levels, layers);
                 }
 
                 result = decoded;
@@ -798,15 +800,15 @@ namespace Ryujinx.Graphics.Gpu.Image
                 {
                     case Format.Etc2RgbaSrgb:
                     case Format.Etc2RgbaUnorm:
-                        result = ETC2Decoder.DecodeRgba(result, width, height, depth, levels, layers);
+                        result = ETC2Decoder.DecodeRgba(result, width, height, sliceDepth, levels, layers);
                         break;
                     case Format.Etc2RgbPtaSrgb:
                     case Format.Etc2RgbPtaUnorm:
-                        result = ETC2Decoder.DecodePta(result, width, height, depth, levels, layers);
+                        result = ETC2Decoder.DecodePta(result, width, height, sliceDepth, levels, layers);
                         break;
                     case Format.Etc2RgbSrgb:
                     case Format.Etc2RgbUnorm:
-                        result = ETC2Decoder.DecodeRgb(result, width, height, depth, levels, layers);
+                        result = ETC2Decoder.DecodeRgb(result, width, height, sliceDepth, levels, layers);
                         break;
                 }
             }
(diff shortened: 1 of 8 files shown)


Bjorn29512:
Tested 20+ titles and invested a couple minutes of playtime in each, noticed no performance or any visual difference compared to master, cant speak for specific scenarios that would break any visuals or any crashes introduced since that is outside the scope of a limited testing like this. Metroid prime did get a significant improvement when it comes stuttering as per the objective of the PR.

Attachments showing the titles tested : 
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image

Squall-Leonhart:
This significantly improves the issue reported in #4420

Bjorn29512:
> This improves the issue reported in #4420 but Fifo still eventually hits 99%

No its still the same, no such effect you are talking about.

image

Ryujinx.Graphics.Gpu/Image/TextureCache.cs
```diff
@@ -194,6 +195,39 @@ public void Lift(Texture texture)
...
+        {
+            // There cannot be an existing texture compatible with this mapping in the texture cache already.
+            int overlapCount = _textures.FindOverlaps(range, ref _textureOverlaps);
+
+            for (int i = 0; i < overlapCount; i++)
+            {
+                var other = _textureOverlaps[i];
+                
+                if (texture != other &&
+                    (texture.IsViewCompatible(other.Info, other.Range, true, other.LayerSize, _context.Capabilities, out int _, out int _) != TextureViewCompatibility.Incompatible ||
```
gdkchan, line 215:
```suggestion
                    (texture.IsViewCompatible(other.Info, other.Range, true, other.LayerSize, _context.Capabilities, out _, out _) != TextureViewCompatibility.Incompatible ||
```

Ryujinx.Graphics.Gpu/Image/TextureCache.cs
```diff
@@ -194,6 +195,39 @@ public void Lift(Texture texture)
...
+                    other.IsViewCompatible(texture.Info, texture.Range, true, texture.LayerSize, _context.Capabilities, out int _, out int _) != TextureViewCompatibility.Incompatible))
```
gdkchan, line 216:
```suggestion
                    other.IsViewCompatible(texture.Info, texture.Range, true, texture.LayerSize, _context.Capabilities, out _, out _) != TextureViewCompatibility.Incompatible))
```

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -39,6 +39,11 @@ public TextureIncompatibleOverlap(TextureGroup group, TextureViewCompatibility c
     /// </summary>
     class TextureGroup : IDisposable
     {
+        /// <summary>
+        /// Threshold of layers to force granular handles (and thus partial loading) on array/3d textures.
```
gdkchan, line 43:
```suggestion
        /// Threshold of layers to force granular handles (and thus partial loading) on array/3D textures.
```

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -331,24 +359,55 @@ public void SynchronizeMemory(Texture texture)
...
                 {
                     var info = GetHandleInformation(baseHandle + i);
 
+                    if (spanEndIndex <= i - 1)
+                    {
+                        spanEndIndex = i;
+
+                        if (_is3D)
+                        {
+                            // Look ahead to see how many handles need loaded.
```
gdkchan, line 378:
```suggestion
                            // Look ahead to see how many handles need loaded.
```
"need loaded" sounds bit weird. Maybe "needs to be loaded"?

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -1,19 +1,45 @@
...
+    /// <summary>
+    /// A request to dereference a texture from a pool.
+    /// </summary>
+    struct DereferenceRequest
+    {
+        public readonly bool IsUnmapped;
+        public readonly Texture Texture;
+        public readonly int ID;
+
+        public DereferenceRequest(Texture texture)
```
gdkchan, line 21: Missing XML docs for the constructors and fields.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -159,24 +189,95 @@ public void ForceRemove(Texture texture, int id, bool deferred)
...
+
+                    Items[request.ID] = texture;
+
+                    // Create a new pool reference, as the last one was removed on unmap.
+
+                    texture.IncrementReferenceCount(this, request.ID, address);
+                    texture.DecrementReferenceCount();
+
+                    // Refetch the range. Changes since the last check could have been lost
+                    // as the cache entry was not restored (required to queue mapping change)
```
gdkchan, line 253:
```suggestion
                    // as the cache entry was not restored (required to queue mapping change).
```

ActualMandM:
No noticeable difference with Groove Coaster and Project DIVA between master and this PR.

riperiperi:
I think I'm done tweaking this, should probably retest to make sure everything's still working.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -7,6 +7,7 @@
 using Ryujinx.Memory.Range;
 using System;
 using System.Collections.Generic;
+using System.Reflection.Metadata;
```
gdkchan, line 10: Why is that needed?

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -1057,17 +1121,55 @@ public void Inherit(TextureGroup other)
...
                 {
-                    foreach (CpuRegionHandle handle in groupHandle.Handles)
+                    if (rangeChanged)
                     {
-                        handle.Reprotect();
+                        // When the storage range changes, this becomes a little different.
+                        // If a range does not match one in the original, treat it as modified.
+                        // It has been newly mapped and its data must be synchronized.
+
+                        if (groupHandle.Handles.Length != 0)
```
gdkchan, line 1139: Would be nice to change it to something like:
```cs
if (groupHandle.Handles.Length == 0)
{
    continue;
}
```
To avoid all this nesting.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -1057,17 +1121,55 @@ public void Inherit(TextureGroup other)
...
+                        {
+                            foreach (var oldGroup in _handles)
+                            {
+                                if (groupHandle.OverlapsWith(oldGroup.Offset, oldGroup.Size))
```
gdkchan, line 1143: Similar suggestion to the above to avoid nesting.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -12,8 +15,61 @@ namespace Ryujinx.Graphics.Gpu.Image
...
+            {
+                IsRemapped = isRemapped;
+                Texture = texture;
+                ID = id;
+            }
+
+            /// <summary>
+            /// Create a dereference request for a texture removal.
+            /// </summary>
+            /// <param name="texture">The texture being removed</param>
```
gdkchan, line 54: Missing `<returns>`.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -12,8 +15,61 @@ namespace Ryujinx.Graphics.Gpu.Image
...
+            public static DereferenceRequest Remove(Texture texture)
+            {
+                return new DereferenceRequest(false, texture, 0);
+            }
+
+            /// <summary>
+            /// Create a dereference request for a texture remapping with a specific pool ID.
+            /// </summary>
+            /// <param name="texture">The texture being remapped</param>
+            /// <param name="id">The ID of the pool entry, used to restore remapped textures</param>
```
gdkchan, line 64: Missing `<returns>`.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -155,28 +215,106 @@ public int CheckModified()
...
 
+        /// <summary>
+        /// Queues a request to update a texture's mapping. 
+        /// Mapping is updated later to avoid deleting the texture if it is still sparsely mapped.
+        /// </summary>
+        /// <param name="texture">Texture with potential mapping change</param>
+        /// <param name="id">ID in cache of texture with potential mapping change</param>
+        public void QueueUpdateMapping(Texture texture, int id)
+        {
+            if (Interlocked.Exchange(ref Items[id], null) != null)
```
gdkchan, line 241: Am I correct assuming that `Items[id]` is expected to be the same as `texture` if not null? If so, I think it's possible to remove the texture parameter and use the previous value on `Items[id]`.

Ryujinx.Graphics.Gpu/Image/Texture.cs
```diff
@@ -1585,6 +1605,36 @@ public void RemoveFromPools(bool deferred)
...
+                {
+                    if (address == 0 || address == owner.GpuAddress)
+                    {
+                        address = owner.GpuAddress;
+
+                        owner.Pool.QueueUpdateMapping(this, owner.ID);
+                    }
+                    else
+                    {
+                        // If there is a different GPU va mapping, prefer the first and delete the others.
```
gdkchan, line 1627:
```suggestion
                        // If there is a different GPU VA mapping, prefer the first and delete the others.
```

gdkchan (approved):
lgtm, thanks. Would be nice to address the remaining comment but it's not a big deal.

Ryujinx.Graphics.Gpu/Image/TexturePool.cs
riperiperi: My initial worry is that `Items[id]` could be swapped for another non-null texture, but I've spent a long time figuring out a case where that might happen, and cannot find one (due to the combination of `lock` and interlocked exchange). This method is always called within `_poolOwners` lock, which greatly limits how much can change. 

I can change it, though maybe to be extra cautious, it should instead be an exchange check to see if the old value is equal to `texture`.

Khobztajine:
Hi am new on github and I want to thank you for this build, On xenoblade chronicle definitive edition, I had 35/55 fps with a lot of stutter with the main version of ryujinx but with this build my game run at 60 fps with no drop and no stutter, I dont know what magic it is but its work amazingly.

Disorientator:
Sorry in advance if this is a stupid question, but does the ASTC texture decoding take place on the CPU or GPU?

GamerzHell9137:
> Sorry in advance if this is a stupid question, but does the ASTC texture decoding take place on the CPU or GPU?

CPU, and no, changing to GPU wouldn't help much.

Disorientator:
Yeap, that was my next question. Thanks for clarifying.

PR #4451, opened 2023-02-19
Memory: Faster Split for NonOverlappingRangeList

riperiperi:
I noticed that in Xenoblade 2, the game can end up spending a lot of time adding and removing tracking handles. One of the main causes of this is actually splitting existing VirtualRegions, which does the following:

- Remove existing VirtualRegion from list
- Update existing VirtualRegion to end at split address, create new VirtualRegion starting at split address
- Add updated VirtualRegion (left) to list
- Add new VirtualRegion (right) to list

This costs 1 deletion and 2 insertions. When there are more VirtualRegions, this gets a lot more expensive, as insertions are done by copying all values to the right, and deletions by copying values to the left.

This PR simply allows it to look up the VirtualRegion being split, and replace its entry with the new end address without insertion or deletion. This makes a split only cost 1 insertion and a binary search lookup (very cheap). This isn't all of the cost on Xenoblade 2, but it does significantly reduce it.

There might be something else to this - we could find a way to reduce the handle count for the game (merging on deletion? buffer deletion?), we could use a different structure for virtual regions, as the current one is optimal for buffer lookups which nearly always read, memory tracking has more of a balance between read/write. That's for a later date though, this was an easy improvment.

### DIFF
--- a/Ryujinx.Memory/Range/NonOverlappingRangeList.cs
+++ b/Ryujinx.Memory/Range/NonOverlappingRangeList.cs
@@ -97,10 +97,8 @@ namespace Ryujinx.Memory.Range
         /// <returns>The new region (high part)</returns>
         private T Split(T region, ulong splitAddress)
         {
-            Remove(region);
-
             T newRegion = (T)region.Split(splitAddress);
-            Add(region);
+            Update(region);
             Add(newRegion);
             return newRegion;
         }
--- a/Ryujinx.Memory/Range/RangeList.cs
+++ b/Ryujinx.Memory/Range/RangeList.cs
@@ -67,6 +67,43 @@ namespace Ryujinx.Memory.Range
             Insert(index, new RangeItem<T>(item));
         }
 
+        /// <summary>
+        /// Updates an item's end address on the list. Address must be the same.
+        /// </summary>
+        /// <param name="item">The item to be updated</param>
+        /// <returns>True if the item was located and updated, false otherwise</returns>
+        public bool Update(T item)
+        {
+            int index = BinarySearch(item.Address);
+
+            if (index >= 0)
+            {
+                while (index > 0 && _items[index - 1].Address == item.Address)
+                {
+                    index--;
+                }
+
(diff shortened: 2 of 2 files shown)


gdkchan (approved):
lgtm, but I think it's possible to do only one lookup here. AFAIK those items are sorted by address, so if you split one, the right split will be always located *after* the existing item that has been split, so it can just insert directly at that index rather than doing another lookup to find where to insert it.

riperiperi:
I didn’t go that far as it would be more use specific than a replace method, but it would be a bit faster. The expensive psrt right now is the copy on insertion.

marysaka (approved):

lostromb:
Can't really notice a perf difference in other games. Oh well

Pokemon Scarlet
Branch | FPS Mean | Median | p99 Low% | Frametime StdDev ms | Stutter %
------|---------|--------|------|------------------|---------
master | 47.11 | 48.37 | 34.60 | 2.89 | 0.09
range-list-split | 46.55 | 47.61 | 35.07 | 2.89 | 0.13

Atelier Ryza 2
Branch | FPS Mean | Median | p99 Low% | Frametime StdDev ms | Stutter %
------|---------|--------|------|------------------|---------
master | 53.35 | 54.74 | 35.42 | 2.69 | 0.00
range-list-split | 54.75 | 55.84 | 42.50 | 1.96 | 0.00

Breath of the Wild
Branch | FPS Mean | Median | p99 Low% | Frametime StdDev ms | Stutter %
------|---------|--------|------|------------------|---------
master | 43.85 | 44.61 | 31.79 | 2.63 | 0.00
range-list-split | 43.85 | 44.62 | 30.78 | 2.75 | 0.00

SMT V
Branch | FPS Mean | Median | p99 Low% | Frametime StdDev ms | Stutter %
------|---------|--------|------|------------------|---------
master | 84.46 | 86.91 | 56.21 | 2.04 | 0.00
range-list-split | 79.54 | 82.28 | 48.62 | 2.62 | 0.08

PR #4460, opened 2023-02-21
misc: changes base application directory behaviour

marysaka:
This allows changing base application directory behavior at build time via FORCE_EXTERNAL_BASE_DIR.

This is intended to be used by nixpkgs and flathub builds.

I also added the missing patch for macOS that we have on macos1 to avoid invalidating code signature.

### DIFF
--- a/Ryujinx.Common/ReleaseInformation.cs
+++ b/Ryujinx.Common/ReleaseInformation.cs
@@ -40,14 +40,21 @@ namespace Ryujinx.Common
             }
         }
 
+#if FORCE_EXTERNAL_BASE_DIR
         public static string GetBaseApplicationDirectory()
         {
-            if (IsFlatHubBuild())
+            return AppDataManager.BaseDirPath;
+        }
+#else
+        public static string GetBaseApplicationDirectory()
+        {
+            if (IsFlatHubBuild() || OperatingSystem.IsMacOS())
             {
                 return AppDataManager.BaseDirPath;
             }
 
             return AppDomain.CurrentDomain.BaseDirectory;
         }
+#endif
     }
 }


TSRBerry (approved):

riperiperi (approved):

devusb:
Just want to check my understanding of the proper way to use this flag -- right now I am setting `DISABLE_UPDATER` as 
```
"/p:ExtraDefineConstants=DISABLE_UPDATER"
```
which works, but doing the same with
```
"/p:ExtraDefineConstants=FORCE_EXTERNAL_BASE_DIR"
```
doesn't seem to be respected (just trying it by itself first, not trying to combine with `DISABLE_UPDATER` yet intentionally), however
```
"/p:DefineConstants=FORCE_EXTERNAL_BASE_DIR"
```
does seem to be picked up (but may cause other problems). 
Do I need to set this flag differently?
Thanks.

marysaka:
> Just want to check my understanding of the proper way to use this flag -- right now I am setting `DISABLE_UPDATER` as
> 
> ```
> "/p:ExtraDefineConstants=DISABLE_UPDATER"
> ```
> 
> which works, but doing the same with
> 
> ```
> "/p:ExtraDefineConstants=FORCE_EXTERNAL_BASE_DIR"
> ```
> 
> doesn't seem to be respected (just trying it by itself first, not trying to combine with `DISABLE_UPDATER` yet intentionally), however
> 
> ```
> "/p:DefineConstants=FORCE_EXTERNAL_BASE_DIR"
> ```
> 
> does seem to be picked up (but may cause other problems?). Do I need to set this flag differently? Thanks.

you should do:

```
"/p:ExtraDefineConstants=DISABLE_UPDATER;FORCE_EXTERNAL_BASE_DIR"
```

devusb:
@marysaka thanks -- thought I was going crazy for a second.

Issue #4463, opened 2023-02-21
[Regression] #3680 breaking upscaling in games


GamerzHell9137:
### Description of the issue

After the merge of #3680 on couple of games upscaling got broken.

So far the games found that break are:

Wreckfest
WarioWare: Get it Together

There might be more games but so far those games were reporting being broken.

https://user-images.githubusercontent.com/2002038/220464805-2e98af8f-eac2-488c-97d0-aa07abc0be92.mp4

https://user-images.githubusercontent.com/2002038/220464851-a6ed0c9b-1582-4889-a4ff-532821941976.mp4

### Reproduction steps

Just enabling the upscaler

### Log file

N/A

### OS

Windows 10

### Ryujinx version

1.1.630

### Game version

v1.0.0

### CPU

Ryzen 3600

### GPU

NVIDIA RTX 3060 Ti

### RAM

16 GB

PR #4471, opened 2023-02-23
GPU: Scale counter results before addition

riperiperi:
Counter results were being scaled on ReportCounter, which meant that the _total_ value of the counter was being scaled. Not only could this result in very large numbers and weird overflows if the game doesn't clear the counter, but it also caused the result to change drastically.

This PR changes scaling to be done when the value is added to the counter on the backend. This should evaluate the scale at the same time as before, on report counter, but avoiding the issue with scaling the total.

I've also included a fix for the 64-bit result partially being written. Drivers tend to write the low half first, then the high half. Now, we retry if the high half is 0xFFFFFFFF. This fixes counter results reading as 0 randomly (game was interpreting 0xFFFFFFFF00000000 as 32-bit uint), which turned into a very large value when rescaled, as the low half was no longer 0. This might fix some random chance issues with counter results. It would be nice to write the availability as a separate value _after_ the result copy, but I don't think any of our current counter results will be negative, so this should be fine for now.

Fixes scaling in WarioWare: Get It Together, at least in the demo, where it seems to compare old/new counters and broke down when scaling was enabled. Verified to still work in Splatoon 3 with scaling on/off in opengl/vulkan.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Threed/SemaphoreUpdater.cs
@@ -152,21 +152,10 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
 
             ulong ticks = _context.GetTimestamp();
 
-            float divisor = type switch
-            {
-                ReportCounterType.SamplesPassed => _channel.TextureManager.RenderTargetScale * _channel.TextureManager.RenderTargetScale,
-                _ => 1f
-            };
-
             ICounterEvent counter = null;
 
             void resultHandler(object evt, ulong result)
             {
-                if (divisor != 1f)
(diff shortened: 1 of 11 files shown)


GamerzHell9137:
Fixes Wreckfest upscaling too 
image

With the merge of this https://github.com/Ryujinx/Ryujinx/issues/4463 can be closed.

gdkchan:
Doesn't it need to end the query and start a new one if the resolution scale of the render target changes, to divide the accumulated count by the correct scale?

riperiperi:
Yes, but it didn't do that before.

gdkchan (approved):
Looks reasonable, but would be nice if it handled scale changes between render targets since its doing it on the backend now.

marysaka (approved):

PR #4484, opened 2023-02-26
CPU: Avoid argument value copies on the JIT

gdkchan:
This has two changes. The first commit is just minor refactoring renaming some functions with names that better match what they do. It also moves OS-specific functions to separate files on x86. For Arm64, moving them is not necessary as all operating systems follows the same ABI there.

The second commit is an optimization. It avoids the insertion of copies to copy the function argument values into a separate register. Instead it just propagates the argument value into each use site to avoid copies. This was already being done for Arm64, so this change is just doing the same for x86.

Sample diff on Windows.
Before:
```nasm
0:  53                      push   rbx
1:  48 83 ec 40             sub    rsp,0x40
5:  48 89 cb                mov    rbx,rcx
8:  48 89 d9                mov    rcx,rbx
b:  8b 91 00 04 00 00       mov    edx,DWORD PTR [rcx+0x400]
11: 85 d2                   test   edx,edx
13: 74 4c                   je     0x61
15: 83 ea 01                sub    edx,0x1
18: 89 91 00 04 00 00       mov    DWORD PTR [rcx+0x400],edx
1e: 48 b9 48 20 e9 15 01    movabs rcx,0x20115e92048
25: 02 00 00
28: 48 8b 09                mov    rcx,QWORD PTR [rcx]
2b: 48 89 da                mov    rdx,rbx
2e: 48 c7 82 80 00 00 00    mov    QWORD PTR [rdx+0x80],0x2b832048
35: 48 20 83 2b
39: 48 89 8a 88 00 00 00    mov    QWORD PTR [rdx+0x88],rcx
40: 48 89 da                mov    rdx,rbx
43: 48 89 8a 08 04 00 00    mov    QWORD PTR [rdx+0x408],rcx
```
After:
```nasm
0:  53                      push   rbx
1:  48 83 ec 40             sub    rsp,0x40
5:  48 89 cb                mov    rbx,rcx
8:  8b 8b 00 04 00 00       mov    ecx,DWORD PTR [rbx+0x400]
e:  85 c9                   test   ecx,ecx
10: 74 46                   je     0x58
12: 83 e9 01                sub    ecx,0x1
15: 89 8b 00 04 00 00       mov    DWORD PTR [rbx+0x400],ecx
1b: 48 b9 48 20 61 14 7d    movabs rcx,0x27d14612048
22: 02 00 00
25: 48 8b 09                mov    rcx,QWORD PTR [rcx]
28: 48 c7 83 80 00 00 00    mov    QWORD PTR [rbx+0x80],0x2b832048
2f: 48 20 83 2b
33: 48 89 8b 88 00 00 00    mov    QWORD PTR [rbx+0x88],rcx
3a: 48 89 8b 08 04 00 00    mov    QWORD PTR [rbx+0x408],rcx
```
On this example, 3 mov instructions were removed, saving 9 bytes. On larger functions, it leads to the removal of more movs. It can result on a minor performance improvement.

Needs testing on Windows and Linux.

### DIFF
--- a/ARMeilleure/CodeGen/Arm64/PreAllocator.cs
+++ b/ARMeilleure/CodeGen/Arm64/PreAllocator.cs
@@ -54,8 +54,8 @@ namespace ARMeilleure.CodeGen.Arm64
                         continue;
                     }
 
-                    HandleConstantRegCopy(constants, block.Operations, node);
-                    HandleDestructiveRegCopy(block.Operations, node);
+                    InsertConstantRegCopies(constants, block.Operations, node);
+                    InsertDestructiveRegCopies(block.Operations, node);
 
                     switch (node.Instruction)
                     {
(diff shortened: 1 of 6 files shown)


CeruleanSky:
Probably out of scope and you probably already considered it, but if not, if rax can be checked to see if it is available I believe it can do mov rax, [64bit const] knocking out another move.

```movabs rcx, 64bit const
mov    rcx,QWORD PTR [rcx]
```

`mov rax, qword ptr [64bit const]`

gdkchan:
It is not possible to encode a 64-bit constant as part of the memory operand on x86, only a sign extended 32-bit constant (at least from what I recall).

CeruleanSky:
`48 a1 48 20 61 14 7d 02 00 00 mov rax, qword ptr [0x27d14612048]`

image
Also tried it out in winrepl_x64 , the memory contained nops.
image

gdkchan:
Interesting, I didn't know about this encoding. The fact that it can only be used with RAX means that it would likely need to insert another mov after the load though (from RAX to the register that the allocator selected for that operation). If it did happen to also select RAX, then the additional mov could be eliminated, but it would require doing some preferencing on the register allocator to incentive it to select the fixed register and the mov destination register, if available. Since RAX is not callee saved, it would probably not be worth using it in most cases as it would need to be preserved across calls, it would only be worth if the value loaded is consumed right after (before any calls or operations that could trash the RAX register).

marysaka (approved):

riperiperi (approved):
LGTM.

PR #4491, opened 2023-02-27
Sockets: Properly convert error codes on MacOS

riperiperi:
The error codes for MacOS are very different to how they are on windows or linux. An alternate mapping is used when the host operating system is MacOS.

This PR also defaults IsDhcpEnabled to true when interfaceProperties.DhcpServerAddresses is not available.

This change was already in `macos1`.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/WinSockHelper.cs
+++ b/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/WinSockHelper.cs
@@ -10,133 +10,133 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
         private static readonly Dictionary<WsaError, LinuxError> _errorMap = new()
         {
             // WSAEINTR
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/WinSockHelper.cs
```diff
@@ -90,6 +91,54 @@ static class WinSockHelper
             {0, 0}
         };
 
+        private static readonly Dictionary<int, LinuxError> _errorMapMacOs = new()
+        {
+            {35, LinuxError.EAGAIN},
```
gdkchan, line 96:
```suggestion
            { 35, LinuxError.EAGAIN },
```
It should be formatted like this for consistency, although `_errorMap` is also missing the spaces. Would be nice to fix both.

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #4503, opened 2023-03-03
Add missing hotkeys, allow modifiers and controller button mappings on Avalonia

gdkchan:
This is adding a few features that have been requested related to hotkeys on Avalonia:
- Allow the few missing options that does have hotkeys to be configured (toggle fullscreen, toggle docked mode and exit).
- Allow key modifiers (control, alt, shift and command on macOS) to be part of hotkey mappings (rather than just a single key).
- Allow single controller buttons and button combinations to be used as hotkey. Currently it is possible to have one keyboard hotkey and one controller hotkey for each configurable hotkey.

I do not plan to implement this on GTK.

TODO:
- Config file migration?
- Remove old hotkey management from Avalonia.
- Update hotkeys on menus.
- Investigate a few bugs.
- Add new strings to locale files.


TomTurbine:
Any way I could request the status of this change?
This is one feature, that while small, is actually a bigger deal than you think to many users.

jgkings:
Llegué a esta solicitud ya que me encontré con esta limitación en la asignación de las teclas de acceso rápido, al parecer aún está en regreso.

TomTurbine:
Any word on the progress of this one yet?

Tsury:
Please implement this

Kupo91:
I would also very much appreciate information on what is blocking this pull request given that it is already 16 months old. Obviously many users are interested in this feature

gdkchan:
> I would also very much appreciate information on what is blocking this pull request given that it is already 16 months old. Obviously many users are interested in this feature

I had an issue with the way how hotkeys and controller buttons are handled with Avalonia UI. There are currently 2 different ways to handle hotkeys, one of them is not configurable and only works with keyboard, and IIRC is implemented by Avalonia itself. The other can support controllers easily, but it checks the pressed buttons in the render loop, so it only works when a game is running. So to sum it up, one method works everywhere but only supports keyboard, the other supports controller buttons but only works when there's a game running. At the time, I did not want to rewrite the hotkey management to support keyboard and controller hotkeys both in and outisde of a game.

In general, I think a better solution overall is an UI that has been designed from the ground up to work with a controller. Even if this is implemented, someone playing exclusively with a controller will likely run into other limitations.

TomTurbine:
Question, would it be possible to run a low resource applet in the background when a game isn't running just to have that loop active and be able to read from them? That way, it always has a way to read from the controllers, depending on how light they could make that applet actually be.

If the applet had to require the same resources as a full game or even the Mii Maker applet, I could see issues, but just having some token process running at as low a resource as possible to function would be an interesting workaround.

PR #4507, opened 2023-03-04
misc: Some dependencies cleanup

marysaka:
We had some leftovers from Vulkan support release that were around.

I also took that opportunity to use System.IO.Hashing instead of Crc32.NET in Ryujinx.Input.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,10 +45,7 @@
     <PackageVersion Include="SPB" Version="0.0.4-build28" />
     <PackageVersion Include="System.Drawing.Common" Version="7.0.0" />
     <PackageVersion Include="System.IdentityModel.Tokens.Jwt" Version="6.27.0" />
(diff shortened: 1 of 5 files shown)


AcK77 (approved):

gdkchan:
Removing `shaderc` means that we can't use GLSL on Vulkan. This is not something we allow doing on current builds, but it's possible to manually enable it by changing `GraphicsConfig.EnableSpirvCompilationOnVulkan` to `false`, which might be useful for debugging. I don't like the `shaderc` dependency, but I don't know a better option right now (other than making our own binding, which might not be too hard actually since we only need to call one function on the library). There's also the fact that we don't ship the shaderc library, so anyone willing to use this for testing needs to build or get it themselves and put it somewhere the emulator can load.

marysaka:
Rebased and dropped the commit removing shaderc.net dependency.

gdkchan (approved):

PR #4509, opened 2023-03-04
Misc performance tweaks

jhorv:
Miscellaneous performance tweaks:
- remove the use of LINQ in a hot path
- use `Array.Empty<T>()` to avoid allocating 0-byte arrays
- adjust some for loops so that the JIT will elide bounds checking

### DIFF
--- a/ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
+++ b/ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
@@ -433,16 +433,11 @@ namespace ARMeilleure.CodeGen.RegisterAllocators
 
         private static int GetHighestValueIndex(Span<int> span)
         {
-            int highest = span[0];
-
-            if (highest == int.MaxValue)
-            {
-                return 0;
-            }
+            int highest = int.MinValue;
 
             int selected = 0;
 
-            for (int index = 1; index < span.Length; index++)
+            for (int index = 0; index < span.Length; index++)
             {
                 int current = span[index];
 
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -152,6 +152,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
                 if (current.TimePoint < lowestTimePoint)
                 {
                     selected = current;
+                    lowestTimePoint = current.TimePoint;
                 }
             }
 
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -149,7 +149,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
             {
                 WaitingObject current = _waitingObjects[index];
 
-                if (current.TimePoint < lowestTimePoint)
+                if (current.TimePoint <= lowestTimePoint)
                 {
                     selected = current;
                     lowestTimePoint = current.TimePoint;
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -145,7 +145,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Common
             
             long lowestTimePoint = long.MaxValue;
 
-            for (int index = 0; index < _waitingObjects.Count; index++)
+            for (int index = _waitingObjects.Count - 1; index >= 0; index--)
             {
                 WaitingObject current = _waitingObjects[index];
 
--- a/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
@@ -1,7 +1,6 @@
 using Ryujinx.Common;
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 
 namespace Ryujinx.HLE.HOS.Kernel.Common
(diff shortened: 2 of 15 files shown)


Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -140,6 +139,25 @@ private void WaitAndCheckScheduledObjects()
             }
         }
 
+        private WaitingObject GetNextWaitingObject()
+        {
+            WaitingObject selected = null;
```
gdkchan, line 144: Maybe that should be initialized to `_waitingObjects[0]` (or null if the array length is 0) to avoid the extremely unlike case where all time points are equal to `long.MaxValue`. I don't think this can happen in practice, so it's more making sure that the method work properly for every possible input.

ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
```diff
@@ -433,16 +433,11 @@ private void AllocateRegWithSpill(AllocationContext context, LiveInterval curren
 
         private static int GetHighestValueIndex(Span<int> span)
         {
-            int highest = span[0];
-
-            if (highest == int.MaxValue)
```
gdkchan, line 438: Why was this fast path removed? Does it not happen in practice?

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
jhorv: I set it to null because the LINQ version's FirstOrDefault() would also have set it to null.

gdkchan: the LINQ version wouldn't set it to null if the TimePoint value of all elements is `long.MaxValue`.

ARMeilleure/CodeGen/RegisterAllocators/LinearScanAllocator.cs
jhorv: I removed that because with the loop being adjusted to start at index 0, and the loop body already having a test for int.MaxValue, that condition will still result in an early return.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
gdkchan: You also never set the `lowestTimePoint` value so I think it will always return the last element.

jhorv: The LINQ version will set it to null if `_waitingObjects` is empty.

gdkchan: Yes. That's not what I'm talking about. I'm talking about the case where `_waitingObjects` is **not** empty and has a single (or more) elements with a `TimePoint` value equal to `long.MaxValue`. In that case `if (current.TimePoint < lowestTimePoint)` would be false and your `GetNextWaitingObject` would return null, while the LINQ version would return the first element on `_waitingObjects`.

jhorv: I added a commit that fixes the error of not setting lowestTimePoint anywhere.  What do you think the best way to handle the scenario you're describing - change the comparison to `current.TimePoint <= lowestTimePoint`? With that, the behavior will be that in the event of a tie, the last one in the list with the same TimePoint because the one chosen.

gdkchan: You can do what I suggested on the first comment. `WaitingObject selected = _waitingObjects.Length > 0 ? _waitingObjects[0].Value : null;` or something like that.

jhorv: I added a commit that compares with <=. This is probably different from what the LINQ version is doing, since I think the behavior of OrderBy() for ties is undefined. Done this way, _waitingObjects with ties on TimePoint will act LIFO - would it be better to loop through the list backward to give it "FIFO for ties" behavior?

jhorv: I added a commit to loop backward so it'll be FIFO for ties.

gdkchan (approved):
lgtm, thanks.

lostromb:
Huh. I learned something today.
```
byte[] buf = new byte[count];
for (int c = 0; c < buf.Length; c++)
{
    buf[c] = 10;
}
```

```
       sub       rsp,28
       movsxd    rdx,ecx
       mov       rcx,offset MT_System.Byte[]
       call      CORINFO_HELP_NEWARR_1_VC
       xor       edx,edx
       mov       ecx,[rax+8]
       test      ecx,ecx
       jle       short M01_L01
       nop
M01_L00:
       movsxd    r8,edx
       mov       byte ptr [rax+r8+10],0A
       inc       edx
       cmp       ecx,edx
       jg        short M01_L00
M01_L01:
       add       rsp,28
       ret
```

If you use `buf.Length` in the `for` loop, there's no CORINFO_HELP_RNGCHKFAIL path on the generated asm, and the loop overhead gets reduced by 5-10%. Good to know!

jhorv:
> If you use `buf.Length` in the `for` loop, there's no CORINFO_HELP_RNGCHKFAIL path on the generated asm, and the loop overhead gets reduced by 5-10%. Good to know!

Yes! Also note that the compiler has special-case handling on `foreach` over an array, automatically emitting the fast form:  `for (int index = 0; index < array.Length; index++)`

jhorv:
Check this out - older, but still applicable!

[RyuJIT Just-in-Time Compiler Optimization Enhancements](https://devblogs.microsoft.com/dotnet/ryujit-just-in-time-compiler-optimization-enhancements/)

riperiperi (approved):

PR #4526, opened 2023-03-09
Added check for eventual symlink when displaying game files.

Shalito01:
Addresses issue #4525. Checks if the file is a symlink it gets the true full path of the file so it shows the correct size.

### DIFF
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -907,3 +907,4 @@ namespace Ryujinx.Ui.App.Common
         }
     }
 }
+
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -132,11 +132,19 @@ namespace Ryujinx.Ui.App.Common
                 }
 
                 // Loops through applications list, creating a struct and then firing an event containing the struct for each application
-                foreach (string applicationPath in applications)
+                foreach (string appPath in applications)
                 {
                     if (_cancellationToken.Token.IsCancellationRequested)
                     {
                         return;
+                    }
+
+                    string applicationPath = appPath;
+
+                    // Follows eventual symlink
+                    var tempFile = Directory.ResolveLinkTarget(appPath, true);
+                    if(tempFile != null) {
+                        applicationPath = tempFile.FullName;
                     }
 
                     double fileSize = new FileInfo(applicationPath).Length * 0.000000000931;
--- a/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -122,7 +122,7 @@ namespace Ryujinx.Ui.App.Common
                             if (!fileInfo.Attributes.HasFlag(FileAttributes.Hidden) && extension is ".nsp" or ".pfs0" or ".xci" or ".nca" or ".nro" or ".nso")
                             {
                                 var fullPath = fileInfo.ResolveLinkTarget(true)?.FullName ?? fileInfo.FullName;
-                                applications.Add(app);
+                                applications.Add(fullPath);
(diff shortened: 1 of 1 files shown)


TSRBerry:
Could you please add a description and explain why you made this change?

Also you might want to check that commit, it looks like you are using two Github accounts and probably committed with the wrong identity?

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -129,13 +129,21 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
                     }
 
+                    string applicationPath = appPath;
+
+                    // Follows eventual symlink
+                    var tempFile = Directory.ResolveLinkTarget(appPath, true);
+                    if(tempFile != null) {
+                        applicationPath = tempFile.FullName;
+                    }
+
```
TSRBerry, line 146: I think we could solve this in a better way if we combine this step with the other operations when we add application paths to the list.

```suggestion
```

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -129,13 +129,21 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
                 }
 
                 // Loops through applications list, creating a struct and then firing an event containing the struct for each application
-                foreach (string applicationPath in applications)
+                foreach (string appPath in applications)
```
TSRBerry, line 132:
```suggestion
                foreach (string applicationPath in applications)
```

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -112,9 +112,9 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
                             {
                                 return;
                             }
-                        
+
```
TSRBerry, line 115:
```suggestion

                            var fileInfo = new FileInfo(app);
```
If we create an instance of `FileInfo` here, we are able to simplify the steps below and check for symlinks as well.

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -112,9 +112,9 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
                             string extension = Path.GetExtension(app).ToLower();
```
TSRBerry, line 116:
```suggestion
                            string extension = fileInfo.Extension.ToLower();
```

Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -112,9 +112,9 @@ public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage
...
-                        
+
                             if (!File.GetAttributes(app).HasFlag(FileAttributes.Hidden) && extension is ".nsp" or ".pfs0" or ".xci" or ".nca" or ".nro" or ".nso")
                             {
                                 applications.Add(app);
```
TSRBerry, line 120:
```suggestion
                            if (!fileInfo.Attributes.HasFlag(FileAttributes.Hidden) && extension is ".nsp" or ".pfs0" or ".xci" or ".nca" or ".nro" or ".nso")
                            {
                                var fullPath = fileInfo.ResolveLinkTarget(true)?.FullName ?? fileInfo.FullName;

                                applications.Add(fullPath);
```

Now we only need to add a check for symlinks here and resolve the actual path before we add it to the list of applications.

TSRBerry (changes requested):
I can think we might be able to solve this issue in a better/simpler way.

Let me know what you think about that.

Shalito01:
Yeah, I think it's a lot better in this way.

riperiperi (approved):
I think this makes sense.

TSRBerry (approved):
I was waiting for a new review request here, but seems like everything is already done.
Make sure to re-request reviewers once you are done adding changes, so we get a notification to review again.

Lgtm!

PR #4528, opened 2023-03-09
GPU: Fast path for adding one texture view to a group

riperiperi:
Texture group handles must store a list of their overlapping views, so they can be properly notified when a write is detected, and a few other things relating to texture readback. This is generally created when the group is established, with each handle looping over all views to find its overlaps. This whole process was also done when only a single view was added (and no handles were changed), however...

Sonic Frontiers had a huge cubemap array with 7350 faces (175 cubemaps * 6 faces * 7 levels), so iterating over both handles and existing views when adding a new one added up very fast (potentially 50 million iterations to add the final views). Since we are only adding a single view, we only need to _add_ that view to the existing overlaps, rather than recalculate them all.

This greatly improves performance during loading screens and a few seconds into gameplay on the "open zone" sections of Sonic Frontiers and its demo. May improve loading times or stutters on some other games.

Note that the current texture cache rules will cause these views to fall out of the cache, as there are more than the hard cap, so the cost will be repaid when reloading the open zone.

I also added some code to properly remove overlaps when texture views are removed, since it seems that was missing.

This can be improved further by only iterating handles that overlap the view (filter by range), but so can a few places in TextureGroup, so better to do all at once. The full generation of overlaps could probably be improved in a similar way.

I recommend testing a few games to make sure nothing breaks.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
@@ -168,9 +168,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             // Overlaps can be accessed from the memory tracking signal handler, so access must be atomic.
 
-            int endOffset = Offset + Size;
-
-            if (offset < endOffset && Offset < offset + (int)view.Size)
+            if (OverlapsWith(offset, (int)view.Size))
             {
                 lock (Overlaps)
                 {
@@ -188,9 +186,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         {
             // Overlaps can be accessed from the memory tracking signal handler, so access must be atomic.
 
-            int endOffset = Offset + Size;
-
-            if (offset < endOffset && Offset < offset + (int)view.Size)
+            if (OverlapsWith(offset, (int)view.Size))
             {
                 lock (Overlaps)
                 {
--- a/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
+++ b/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
@@ -160,6 +160,46 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
+        /// Adds a single texture view as an overlap if its range overlaps.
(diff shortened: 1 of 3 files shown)


Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -159,6 +159,46 @@ public void RecalculateOverlaps(TextureGroup group, List<Texture> views)
...
+        /// </summary>
+        /// <param name="offset">The offset of the view in the group</param>
+        /// <param name="view">The texture to add as an overlap</param>
+        public void AddOverlap(int offset, Texture view)
+        {
+            // Overlaps can be accessed from the memory tracking signal handler, so access must be atomic.
+
+            int endOffset = Offset + Size;
+
+            if (offset < endOffset && Offset < offset + (int)view.Size)
```
gdkchan, line 173:
```suggestion
            if (OverlapsWith(offset, (int)view.Size))
```
And then `int endOffset = Offset + Size;` can be removed.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -159,6 +159,46 @@ public void RecalculateOverlaps(TextureGroup group, List<Texture> views)
...
+        /// </summary>
+        /// <param name="offset">The offset of the view in the group</param>
+        /// <param name="view">The texture to add as an overlap</param>
+        public void RemoveOverlap(int offset, Texture view)
+        {
+            // Overlaps can be accessed from the memory tracking signal handler, so access must be atomic.
+
+            int endOffset = Offset + Size;
+
+            if (offset < endOffset && Offset < offset + (int)view.Size)
```
gdkchan, line 193: Same comment as above.

Vaskiemaia:
I have done some testing and the game loads a significant amount faster from the main menu, but also upon reloading the world (when falling off the map for example). The experience is far more enjoyable than in the past.

The game still has performance issues but that is beyond the scope of this PR.

Video:

- https://streamable.com/r2kpl3

- https://streamable.com/pu7eye

PR #4529, opened 2023-03-10
[Flatpak] Add release github workflow

TSRBerry:
This PR ports the python scripts @marysaka previously used to automatically update the Ryujinx flatpak to GitHub Actions.

This should help to keep dependencies better in sync and will hopefully help to break flatpak builds a bit less.

@marysaka's scripts did most of the work here and I only needed to do the little work to port this.

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -1,0 +1,161 @@
+name: Flatpak release job
+
+on:
+  workflow_call:
+    inputs:
+      ryujinx_version:
+        required: true
+        type: string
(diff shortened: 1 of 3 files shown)


marysaka (approved):

iMonZ:
Is it possible to add arm64 for flatpak Linux?

marysaka:
> Is it possible to add arm64 for flatpak Linux?

See https://github.com/flathub/org.ryujinx.Ryujinx/issues/39#issuecomment-1432831324

PR #4530, opened 2023-03-10
Update syscall capabilites to include SVCs from FW 15.0.0

TSRBerry:
This PR introduces a new KernelConstant `SupervisorCallCount` which gets used to make sure we are not setting SVCs that are out of range.

Since the amount of SVCs increased with firmware version 15.0.0, I also increased the size of `SvcAccessMask` to make room for them.

Additionally an enum (`CapabilityType`) was introduced along with some common operations in `CapabilityExtensions` to make the code a little bit more readable.

As a result of this PR the homebrew menu (using `hbl.nsp`) can now be launched again!

---

~~I'm planning on adding the logic for `MapRegion` capabilities, which is why this PR is currently marked as a draft.~~
Nvm, it seems `MapRegion` is mostly unused and is probably not worth the effort required to support it for now.
If you still want me to add it before accepting this PR feel free to tell me.

### DIFF
--- a/Ryujinx.HLE/HOS/ProgramLoader.cs
+++ b/Ryujinx.HLE/HOS/ProgramLoader.cs
@@ -318,7 +318,7 @@ namespace Ryujinx.HLE.HOS
 
             result = process.Initialize(
                 creationInfo,
-                MemoryMarshal.Cast<byte, int>(npdm.KernelCapabilityData).ToArray(),
+                MemoryMarshal.Cast<byte, uint>(npdm.KernelCapabilityData).ToArray(),
                 resourceLimit,
                 memoryRegion,
                 processContextFactory);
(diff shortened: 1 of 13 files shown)


Ryujinx.HLE/HOS/ProgramLoader.cs
```diff
@@ -318,7 +318,7 @@ public static bool LoadKip(KernelContext context, KipExecutable kip)
 
             result = process.Initialize(
                 creationInfo,
-                MemoryMarshal.Cast<byte, int>(npdm.KernelCapabilityData).ToArray(),
+                MemoryMarshal.Cast<byte, uint>(npdm.KernelCapabilityData).ToArray(),
```
gdkchan, line 321: Why is the `ToArray` necessary? Can't we pass the span directly here?

gdkchan (approved):
lgtm, thanks.

Ryujinx.HLE/HOS/ProgramLoader.cs
TSRBerry: I'll take a look! It should be possible.

marysaka (approved):

PR #4537, opened 2023-03-12
Reducing memory allocations

jhorv:
Used the profiler to focus on memory allocation hot paths. Introduced a new dependency on `Microsoft.IO.RecyclableMemoryStream` and use it as a drop-in replacement everywhere possible. Also focused on:

- avoiding array creation
- reusing Stream, BinaryReader, and BinaryWriter instances in thread-safe situations
- avoiding creation of BinaryWriter in some hot paths
- passing `Span`, `ReadOnlySpan`, `ReadOnlySequence` instead of `byte[]` when possible
- reducing LINQ use by `KScheduler`

### DIFF
--- a/Ryujinx.Common/Extensions/StreamExtensions.cs
+++ b/Ryujinx.Common/Extensions/StreamExtensions.cs
@@ -105,7 +105,7 @@ namespace Ryujinx.Common
             if (count <= 16)
             {
                 Span<byte> span = stackalloc byte[count];
-                @this.Write(span);
+                span.Fill(value);
                 stream.Write(span);
             }
             else
--- a/Ryujinx.Common/Extensions/StreamExtensions.cs
+++ b/Ryujinx.Common/Extensions/StreamExtensions.cs
@@ -7,17 +7,19 @@ namespace Ryujinx.Common
     public static class StreamExtensions
     {
         /// <summary>
-        /// Writes a Span of int to this stream.
+        /// Writes a <cref="ReadOnlySpan<int>" /> to this stream.
         ///
-        /// This default implementation calls the Write(Object, int, int)
-        /// method to write the byte array.
+        /// This default implementation converts each buffer value to a stack-allocated
+        /// byte array, then writes it to the Stream using <cref="System.Stream.Write(byte[])" />.
         /// </summary>
         /// <param name="stream">The stream to be written to</param>
         /// <param name="buffer">The buffer of values to be written</param>
         public static void Write(this Stream stream, ReadOnlySpan<int> buffer)
         {
             if (buffer.Length == 0)
+            {
                 return;
+            }
 
             Span<byte> byteBuffer = stackalloc byte[sizeof(int)];
 
@@ -100,21 +102,27 @@ namespace Ryujinx.Common
         public static void WriteByte(this Stream stream, byte value, int count)
         {
             if (count <= 0)
+            {
                 return;
-
-            if (count <= 16)
-            {
-                Span<byte> span = stackalloc byte[count];
-                span.Fill(value);
-                stream.Write(span);
             }
-            else
+
+            const int BlockSize = 16;
+
+            int blockCount = count / BlockSize;
+            if (blockCount > 0)
             {
-                // TODO make this better - maybe stackalloc a small Span<byte> and write it in a loop, then handle remainder
-                for (int x = 0; x < count; x++)
+                Span<byte> span = stackalloc byte[BlockSize];
+                span.Fill(value);
+                for (int x = 0; x < blockCount; x++)
                 {
-                    stream.WriteByte(value);
+                    stream.Write(span);
                 }
+            }
+
+            int nonBlockBytes = count % BlockSize;
+            for (int x = 0; x < nonBlockBytes; x++)
+            {
+                stream.WriteByte(value);
             }
         }
     }
--- a/Ryujinx.Common/Extensions/StreamExtensions.cs
+++ b/Ryujinx.Common/Extensions/StreamExtensions.cs
@@ -103,7 +103,7 @@ namespace Ryujinx.Common
         }
 
         /// <summary>
-        /// Writes a sequence of zero bytes to the Stream.
+        /// Writes a sequence of bytes to the Stream.
         /// </summary>
         /// <param name="stream">The stream to be written to.</param>
         /// <param name="value">The byte to be written</param>
--- a/Ryujinx.Common/Extensions/StreamExtensions.cs
+++ b/Ryujinx.Common/Extensions/StreamExtensions.cs
@@ -1,0 +1,119 @@
+﻿using System;
+using System.Buffers.Binary;
+using System.IO;
+
+namespace Ryujinx.Common
+{
+    public static class StreamExtensions
+    {
+        /// <summary>
+        /// Writes a Span of int to this stream.
+        ///
+        /// This default implementation calls the Write(Object, int, int)
+        /// method to write the byte array.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="buffer"></param>
+        public static void Write(this Stream @this, ReadOnlySpan<int> buffer)
+        {
+            if (buffer.Length == 0)
+                return;
+
+            Span<byte> byteBuffer = stackalloc byte[sizeof(int)];
+
+            foreach (int value in buffer)
+            {
+                BinaryPrimitives.WriteInt32LittleEndian(byteBuffer, value);
+                @this.Write(byteBuffer);
+            }
+        }
+
+        /// <summary>
+        /// Writes a four-byte signed integer to this stream. The current position
+        /// of the stream is advanced by four.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="value"></param>
+        public static void Write(this Stream @this, int value)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(int)];
+            BinaryPrimitives.WriteInt32LittleEndian(buffer, value);
+            @this.Write(buffer);
+        }
+
+        /// <summary>
+        /// Writes an eight-byte signed integer to this stream. The current position
+        /// of the stream is advanced by eight.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="value"></param>
+        public static void Write(this Stream @this, long value)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(long)];
+            BinaryPrimitives.WriteInt64LittleEndian(buffer, value);
+            @this.Write(buffer);
+        }
+
+        /// <summary>
+        // Writes a four-byte unsigned integer to this stream. The current position
+        // of the stream is advanced by four.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="value"></param>
+        public static void Write(this Stream @this, uint value)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(uint)];
+            BinaryPrimitives.WriteUInt32LittleEndian(buffer, value);
+            @this.Write(buffer);
+        }
+
+        /// <summary>
+        /// Writes an eight-byte unsigned integer to this stream. The current
+        /// position of the stream is advanced by eight.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="value"></param>
+        public static void Write(this Stream @this, ulong value)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(ulong)];
+            BinaryPrimitives.WriteUInt64LittleEndian(buffer, value);
+            @this.Write(buffer);
+        }
+
+        /// <summary>
+        /// Writes the contents of stream to this, by calling stream.CopyTo(@this).
+        /// Exists to provide consistency with other MemoryStream.Write methods.
+        /// </summary>
+        /// <param name="this">the stream that will be written to</param>
+        /// <param name="source">the stream that will be read from</param>
+        public static void Write(this Stream @this, Stream source)
+        {
+            source.CopyTo(@this);
+        }
+
+        /// <summary>
+        /// Writes a sequence of zero bytes to the Stream.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="length"></param>
+        public static void WriteByte(this Stream @this, byte value, int count)
+        {
+            if (count <= 0)
+                return;
+
+            if (count <= 16)
+            {
+                Span<byte> span = stackalloc byte[count];
+                @this.Write(span);
+            }
+            else
+            {
+                // TODO make this better - maybe stackalloc a small Span<byte> and write it in a loop, then handle remainder
+                for (int x = 0; x < count; x++)
+                {
+                    @this.WriteByte(value);
+                }
+            }
+        }
+    }
+}
--- a/Ryujinx.Common/Extensions/StreamExtensions.cs
+++ b/Ryujinx.Common/Extensions/StreamExtensions.cs
@@ -12,9 +12,9 @@ namespace Ryujinx.Common
         /// This default implementation calls the Write(Object, int, int)
         /// method to write the byte array.
         /// </summary>
-        /// <param name="this"></param>
-        /// <param name="buffer"></param>
-        public static void Write(this Stream @this, ReadOnlySpan<int> buffer)
+        /// <param name="stream">The stream to be written to</param>
+        /// <param name="buffer">The buffer of values to be written</param>
+        public static void Write(this Stream stream, ReadOnlySpan<int> buffer)
         {
             if (buffer.Length == 0)
                 return;
@@ -24,7 +24,7 @@ namespace Ryujinx.Common
             foreach (int value in buffer)
             {
                 BinaryPrimitives.WriteInt32LittleEndian(byteBuffer, value);
-                @this.Write(byteBuffer);
+                stream.Write(byteBuffer);
             }
         }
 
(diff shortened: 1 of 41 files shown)


Ryujinx.Common/Memory/MemoryStreamManager.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace Ryujinx.Common.Memory
+{
+    public static class MemoryStreamManager
+    {
+        private static readonly RecyclableMemoryStreamManager _shared = new RecyclableMemoryStreamManager();
+
+        //public static RecyclableMemoryStreamManager Shared => _shared;
+
+        /// <summary>
+        /// Exists until RecyclableMemoryStream version 3x is released, which has API improvements.
```
gdkchan, line 14: Maybe that can be worded better.

Ryujinx.Common/Memory/MemoryStreamManager.cs
```diff
@@ -0,0 +1,84 @@
...
+                => GetStream(Guid.NewGuid(), null, buffer, 0, buffer.Length);
+
+            public static RecyclableMemoryStream GetStream(ReadOnlySpan<byte> buffer)
+                => GetStream(Guid.NewGuid(), null, buffer);
+
+            /// <summary>
+            /// Retrieve a new <c>RecyclableMemoryStream</c> object with the given tag and with contents copied from the provided
+            /// buffer. The provided buffer is not wrapped or used after construction.
+            /// </summary>
+            /// <remarks>The new stream's position is set to the beginning of the stream when returned.</remarks>
```
gdkchan, line 32: FWIW we usually don't have `.` at the end of parameter XML docs, only the summary.

Ryujinx.Common/Memory/MemoryStreamManager.cs
```diff
@@ -0,0 +1,84 @@
...
+                    stream.Position = 0;
+                    return stream;
+                }
+                catch
+                {
+                    stream?.Dispose();
+                    throw;
+                }
+            }
+
```
gdkchan, line 81:
```suggestion
```

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
```diff
@@ -331,6 +345,51 @@ private void WriteImpl(ulong va, ReadOnlySpan<byte> data)
...
+            catch (InvalidMemoryRegionException)
+            {
+                if (_invalidAccessHandler == null || !_invalidAccessHandler(va))
+                {
+                    throw;
+                }
+            }
+        }
+
+
```
gdkchan, line 392:
```suggestion
```

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
```diff
@@ -245,6 +246,19 @@ public void Write(ulong va, ReadOnlySpan<byte> data)
             WriteImpl(va, data);
         }
 
+        /// <inheritdoc/>
+        public void Write(ulong va, ReadOnlySequence<byte> data)
```
gdkchan, line 250: What's the purpose of those new methods?

Ryujinx.HLE/HOS/Kernel/Threading/KPriorityQueue.cs
```diff
@@ -37,12 +37,85 @@ public IEnumerable<KThread> SuggestedThreads(int core)
...
+            int count = 0;
+            while (prio < KScheduler.PrioritiesCount)
+            {
+                LinkedList<KThread> list = listPerPrioPerCore[prio][core];
+
+                LinkedListNode<KThread> node = list.First;
+
+                if (node != null)
+                {
+                    if (count == index)
```
gdkchan, line 77: Missing `{ }`.

Ryujinx.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            source.CopyTo(@this);
+        }
+
+        /// <summary>
+        /// Writes a sequence of zero bytes to the Stream.
+        /// </summary>
+        /// <param name="this"></param>
+        /// <param name="length"></param>
+        public static void WriteByte(this Stream @this, byte value, int count)
```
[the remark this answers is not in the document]
jhorv: I'll fix value not being used. I named it WriteByte to be an overload of the existing Stream.WriteByte(byte) method. Do you want it renamed to Memset?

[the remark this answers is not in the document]
gdkchan: I don't think it makes sense to call it Memset in this case, since it is operating on a stream which might not even be accessing memory data directly. Maybe something like `Fill` or `Replicate`. I think `WriteByte` is fine too as long you fix it to actually use the `value` parameter.

Ryujinx.Common/Memory/MemoryStreamManager.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.IO;
+
+namespace Ryujinx.Common.Memory
+{
+    public static class MemoryStreamManager
+    {
+        private static readonly RecyclableMemoryStreamManager _shared = new RecyclableMemoryStreamManager();
+
+        //public static RecyclableMemoryStreamManager Shared => _shared;
```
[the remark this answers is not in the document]
jhorv: Line 9: passed to the constructor of new RecyclableMemoryStreamManager() objects
Line 10-11: will remove

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
jhorv: These new methods allow writing the content of a `RecyclableMemoryStream` without making a copy.
`RecyclableMemoryStream` has a `GetReadOnlySequence()` method, which is more efficient than calling `Stream.ToArray()` because `ToArray()` creates a new large `byte[]` to contain everything int the stream, then copies the stream contents to it.

Ryujinx.Common/Extensions/StreamExtensions.cs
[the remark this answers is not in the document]
jhorv: Just pushed 3 commits, including one that fixes value being unused. I left the name as WriteByte.

jhorv:
@gdkchan I started marking your comments "resolved" as I fixed them, but then I thought you might prefer to mark them resolved when _you_ think they're resolved? Let me know if you have a preference.

gdkchan:
Doesn't matter who marks them as resolved for me, as long they are actually resolved. I think the PR author marking it as resolved is probably better, since then other reviewing the PR doesn't have to be distracted with a lot of "noise" for issues that are already resolved.

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
gdkchan: It's unfortunate that we need to implement new methods for that. It would be better if the stream could return a span with the stream contents. This is essentially duplicating write methods on every memory manager, only changing the type from span to sequence.

Ryujinx.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            if (count <= 16)
+            {
+                Span<byte> span = stackalloc byte[count];
+                span.Fill(value);
+                stream.Write(span);
+            }
+            else
+            {
+                // TODO make this better - maybe stackalloc a small Span<byte> and write it in a loop, then handle remainder
```
gdkchan, line 113: It doesn't seems too hard to this, so might as well do it now.
Something like:
```cs
Span<byte> span = stackalloc byte[16];
span.Fill(value);
int x;
for (x = 0; x < count & ~(16 - 1); x += 16)
{
    stream.Write(span);
}

for (; x < count; x++)
{
    stream.WriteByte(value);
}
```
It is probably better to add a constant for the value "16" and use it in the code to make it easy to change, if we decide to tune this value later.

Ryujinx.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.IO;
+
+namespace Ryujinx.Common
+{
+    public static class StreamExtensions
+    {
+        /// <summary>
+        /// Writes a Span of int to this stream.
+        ///
+        /// This default implementation calls the Write(Object, int, int)
```
gdkchan, line 12: Ideally it should use `see cref` to point to the Write overload here, so that it is displayed with syntax highlight on the IDE etc.

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
jhorv: Agreed, but unfortunately part of the `RecyclableMemoryStream` efficiency comes by it using multiple blocks instead of resizing one big array. Those blocks are wrapped as a `ReadOnlySequence` without copying, whereas if we used `RecyclableMemoryStream`'s GetSpan or GetMemory, it has to make a copy consolidate the blocks.

gdkchan: If it had some way to expose those blocks, and then some `AsSpan` method to get a span from them, then we could use the existing `Write` methods that take span. Writing in chunks is fine for us. But I didn't look at which API it exposes yet so not sure if that would be possible.

Ryujinx.Common/Extensions/StreamExtensions.cs
jhorv: Okay, I'll implement soon. I'll probably avoid bit manipulation for clarity, and as long as we use a ^2 const int for the BlockSize (like we are) then the JIT will turn / and % into bit shifts.

Ryujinx.Common/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            if (buffer.Length == 0)
+                return;
+
+            Span<byte> byteBuffer = stackalloc byte[sizeof(int)];
+
+            foreach (int value in buffer)
+            {
+                BinaryPrimitives.WriteInt32LittleEndian(byteBuffer, value);
+                stream.Write(byteBuffer);
```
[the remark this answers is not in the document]
jhorv: You can't directly cast ReadOnlySpan<int> to ReadOnlySpan<byte>, is there a low-level conversion method somewhere that would do that? Even if there is, wouldn't that do the wrong thing on big-endian platforms?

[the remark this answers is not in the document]
gdkchan: There is `MemoryMarshal.Cast<byte, int>(byteBuffer)`.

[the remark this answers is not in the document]
jhorv: Sorry, I left out the ReadOnlySpan types in my last comment, but you know what I mean. :)
Cool about MemoryMarshal.Cast, I'll do that when the platform is little-endian, and otherwise do the current way.

Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
lostromb: ReadOnlySequence can expose each of its underlying blocks using `ReadOnlySequence<byte>.TryGet(ref SequencePosition position, out ReadOnlyMemory<T> memory)`

jhorv: `ReadOnlySequence<T>` doesn't implement `IEnumerable` but does implement `GetEnumerator()`, so it can be enumerated via foreach as though it were a `IEnumerable<ReadOnlyMemory<T>>`.  It's possible to change every callsite of `Write(ulong, ReadOnlySequence<T>)` to something like:

```
                foreach (ReadOnlyMemory<byte> segment in responseStream.GetReadOnlySequence())
                    process.CpuMemory.Write(messagePtr, segment.Span);
```

The downsides of that:

- Most implementations of `IVirtualMemoryManager.Write(ulong, ReadOnlySpan<byte>)` do a non-trivial amount of work in each call - address checking, mapping, all that - and that work would be repeated for each segment unnecessarily.
- The use of `Write(ulong, ReadOnlySegment<T>)` has 2 call sites in this PR, but IMO should be getting more over time as we reduce the number of buffer copies made throughout the codebase.

gdkchan: > Most implementations of IVirtualMemoryManager.Write(ulong, ReadOnlySpan<byte>) do a non-trivial amount of work in each call - address checking, mapping, all that - and that work would be repeated for each segment unnecessarily.

The one on `MemoryManagerHostMapped` (which is what most people will be using) only checks if the address and size are valid, and signals memory tracking depending on which one you call. The other ones will also need to check if the memory is contiguous. The cost depends on the size of the data being written. If it's fairly small I think it would be fine.

> The use of Write(ulong, ReadOnlySegment<T>) has 2 call sites in this PR, but IMO should be getting more over time as we reduce the number of buffer copies made throughout the codebase.

I think we should just have a helper method to write a `ReadOnlySegment<byte>` to a `IVirtualMemoryManager`, calling the span overload. I'm not sure what would be the best place to put it. Maybe it could be a extension method.

jhorv: I just added a commit that reverts adding all the `Write(ulong, ReadOnlySegment<byte>)` implementations, and also added a default implementation on IVirtualMemoryManager that loops and delegates to `Write(ulong, ReadOnlySpan<byte>)`.

TSRBerry (approved):
Lgtm, thank you!

I just added some stylistic change requests, nothing major.

jhorv:
Apologies for this late commit [6d67848](https://github.com/Ryujinx/Ryujinx/pull/4537/commits/6d67848d39a9d866dad5e9c1a1228658b36467e8) which completes the work of removing LINQ from KScheduler and KPriorityQueue.

Previously this PR removed LINQ for enumerating the KPriorityQueue's `ScheduledThreads`, but not `SuggestedThreads`.  This was because the way I removed LINQ wasn't very elegant - it was mostly just custom methods for the searches KScheduler needed, and they weren't reusable between scheduled and suggested threads.

This new commit has a nice solution I found for low-allocation enumerating `SuggestedThreads` - a custom struct-based enumerator type. As a value type it doesn't cause allocations. It does not implement `IEnumerator<KThread>`, but it does have Current and MoveNext() members, which allow C# to use it in a foreach. With this approach I was able to remove the LINQ usage for `SuggestedThreads`, and re-work the LINQ-less enumeration of `ScheduledThreads` to use this cleaner solution.

jhorv:
I'm preparing some before/after numbers on the effects of this PR, but here's a quick one:

**Test Scenario:**  Start Ryujinx.Ava w/ launch options to run Metroid Prime Remastered. Run until to "Press A to Start", then close.

| branch/commit | Allocations |
| ---- | ---- |
master d56d335c | 18.753 million
this PR 6d67848 | 10.413 million

jhorv:
**Test Scenario:**  Start Ryujinx.Ava w/ launch options to run Metroid Prime Remastered. Run until to "Press A to Start", then close.  Time to run scenario: ~ 24 seconds

| | master d56d335c | this PR 6d67848 | Delta |
| --- | ---- | ---- | ---- |
Allocations | 18.753 million | 10.413 million | -8.34 million (-44%)
Total GC | 110 | 82 | -28 (-25%)
Total Pause Duration | 13.77 seconds | 9.3 seconds | -4.47 seconds (-32%)

Excel file w/ additional details: 
[RyujinxComparison_master-perfMemory.xlsx](https://github.com/Ryujinx/Ryujinx/files/10997151/RyujinxComparison_master-perfMemory.xlsx)

@lostromb I was admiring the performance comparison tables in [PR# 4540](https://github.com/Ryujinx/Ryujinx/pull/4540), would you please tell me what you use to get those (FPS, frametime, stutter) stats?

lostromb:
Those numbers look great to me. Reducing allocations in the future is going to get harder and harder because I know there's a lot of places in memory manager, texture cache and DMA which just allocate huge arrays willy-nilly, but we can whittle those down in the future.

My performance statistics I normally keep as a locally stashed change so I can apply it to any given branch that I want to test, but the commit is [here](https://github.com/Ryujinx/Ryujinx/compare/master...lostromb:Ryujinx:perf_metrics) if you want to pull it yourself

riperiperi:
> Those numbers look great to me. Reducing allocations in the future is going to get harder and harder because I know there's a lot of places in memory manager, texture cache and DMA which just allocate huge arrays willy-nilly, but we can whittle those down in the future. Maybe better use of `ArrayPool` in the case of transient arrays, like I know there's a scratch array allocation on the Opus decoder path and that's been bugging me a lot because I wrote that Opus library.
> 
> My performance statistics I normally keep as a locally stashed change so I can apply it to any given branch that I want to test, but the commit is [here](https://github.com/Ryujinx/Ryujinx/compare/master...lostromb:Ryujinx:perf_metrics) if you want to pull it yourself

It's important to note the frequency of allocations, and if they're necessary. For example, DMA is typically infrequent (unless displaying a video), MemoryManager allocations are also infrequent as they generally only occur when reading back a non-contiguous region of memory, which doesn't happen in the host mapped mode. Texture Cache allocations are usually due to data conversions, which should not overwrite the source data. It's never done "willy-nilly", though DMA could use improvement.

lostromb:
Starting to see this error nondeterministically in master when stopping emulation
```
00:08:51.668 |E| HLE.OsThread.20 Application : Unhandled exception caught: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at Microsoft.IO.RecyclableMemoryStreamManager.ReturnBlocks(List`1 blocks, Guid id, String tag) in /_/src/RecyclableMemoryStreamManager.cs:line 570
   at Microsoft.IO.RecyclableMemoryStream.Dispose(Boolean disposing) in /_/src/RecyclableMemoryStream.cs:line 351
   at Ryujinx.HLE.HOS.Services.ServerBase.Dispose(Boolean disposing)
   at Ryujinx.HLE.HOS.Services.ServerBase.ServerLoop()
   at Ryujinx.HLE.HOS.Services.ServerBase.Main()
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart()
```

jhorv:
> Starting to see this error nondeterministically in master when stopping emulation
> 
> ```
> 00:08:51.668 |E| HLE.OsThread.20 Application : Unhandled exception caught: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
> ```

I finally got this myself, I'm confident it is fixed by PR #4577 .

PR #4177, opened 2022-12-25
Ava GUI: `SettingsWindow` Refactor

IsaacMarovitz:
**Changes:**
- `SettingsWindow` Refactored

Superceeds part of #3997

PR #4538, opened 2023-03-12
Increase access permissions for AvaloniaList<Timezone>

MutantAura:
Fixes a regression from https://github.com/Ryujinx/Ryujinx/pull/4177 where the window couldn't access the timezone list in the ViewModel.

image


TSRBerry (approved):
Thank you!

gdkchan (approved):
Seems reasonable.

PR #4555, opened 2023-03-18
Ability to hide file types in Game List

### LINKED ISSUE #4281: [Feature Request] Ability to hide file types in Game List
#### Overview: 
- I'm requesting the ability to hide certain file types (for example, `nca` files) from the Game List.

#### Nature of Request:  

- Addition
  - The Wii/Gamecube emulator Dolphin has a menu option to hide files from the game list (`dol`, `elf`, `wad`, etc), which inspired this request here.

It would remove stray entries from the Game List, especially in cases where one may be datamining in a directory that Ryujinx is normally set to scan.

NitroTears:
This PR closes #4281 ([Feature Request] Ability to hide file types in Game List)

The config file format version has been bumped to 45
 - Also fixed the version 44 check (with antialiasing/scaling filter accidentally being logged as 42)
 - This includes a HiddenFilesTypes struct, HiddenFileTypesSettings class ect.

A dropdown of checkboxes identical to Enable GUI Columns has been added to check which file types should not be shown on the game list.

[Here is a gif of it in action](https://cdn.discordapp.com/attachments/410432399500115968/1086464759727132762/gifnamehere.gif) (as posted on the Ryujinx discord)

 Happy to discuss If you want me to change anything or the like.


NitroTears:
The Ava version has been added. Just waiting on minor change suggestions / review before finalising.

NitroTears:
The logic to `hide` is now to `show` as per feedback. That should be it!

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml.cs
```diff
@@ -29,6 +30,55 @@ public MainMenuBarView()
         {
             InitializeComponent();
 
+            HideFileTypesMenuItem.Items = GenerateHideFileTypeItems();
+            ChangeLanguageMenuItem.Items = GenerateLanguageMenuItems();
+        }
+
+        private CheckBox[] GenerateHideFileTypeItems()
```
TSRBerry, line 37:
```suggestion
        private CheckBox[] GenerateToggleFileTypeItems()
```

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml.cs
```diff
@@ -29,6 +30,55 @@ public MainMenuBarView()
         {
             InitializeComponent();
 
+            HideFileTypesMenuItem.Items = GenerateHideFileTypeItems();
```
TSRBerry, line 33:
```suggestion
            ToggleFileTypesMenuItem.Items = GenerateToggleFileTypeItems();
```

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml.cs
```diff
@@ -29,6 +30,55 @@ public MainMenuBarView()
...
+                    Content = ".NRO",
+                    IsChecked = ConfigurationState.Instance.Ui.ShownFileTypes.NRO.Value,
+                    Command = MiniCommand.Create(() => ViewModel.HideFileType("nro"))
+                },
+                new CheckBox()
+                {
+                    Content = ".NSO",
+                    IsChecked = ConfigurationState.Instance.Ui.ShownFileTypes.NSO.Value,
+                    Command = MiniCommand.Create(() => ViewModel.HideFileType("nso"))
+                }
```
TSRBerry, line 76: Could you add an enum for the file types and change this to a for-loop?

The file type enum can then also be used in a few other cases.

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml
```diff
@@ -78,6 +78,7 @@
                 </MenuItem>
                 <Separator />
                 <MenuItem Name="ChangeLanguageMenuItem" Header="{locale:Locale MenuBarOptionsChangeLanguage}" />
+				<MenuItem Name="HideFileTypesMenuItem" Header="{locale:Locale MenuBarShownFileTypes}" />
```
TSRBerry, line 81:
```suggestion
				<MenuItem Name="ToggleFileTypesMenuItem" Header="{locale:Locale MenuBarShownFileTypes}" />
```

Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -429,6 +429,7 @@
   "DlcManagerEnableAllButton": "Enable All",
   "DlcManagerDisableAllButton": "Disable All",
   "MenuBarOptionsChangeLanguage": "Change Language",
+  "MenuBarShownFileTypes": "Show File Types",
```
TSRBerry, line 432:
```suggestion
  "MenuBarShowFileTypes": "Show File Types",
```

TSRBerry (changes requested):
I think we might be able to simplify this a bit more. Could you take a look at the comments?

Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml.cs
NitroTears: I think I *mostly* can at a quick look.

```cs
List<CheckBox> checkBoxes = new List<CheckBox>();

foreach (string type in Enum.GetNames(typeof (FileTypes)))
{
    checkBoxes.Add(new CheckBox()
    {
        Content = $".{type}",
        IsChecked = ConfigurationState.Instance.Ui.ShownFileTypes.NSP.Value,
        Command = MiniCommand.Create(() => ViewModel.ToggleFileType(type))
    });
}
```

With the `IsChecked` property, is there I way I can dynamically change the value that is being looked at based on the `type` string. I know I can't use `Ui.ShownFileTypes[type].Value` like I can in JS. What do you think?

TSRBerry: The only way I know to access `ShownFileTypes` like that would be using Reflection.

But I think it might be easier to write an [extension method](https://learn.microsoft.com/en-us/dotnet/api/system.enum?view=net-7.0#adding-enumeration-methods) for this enum, which returns the proper value.
You could then use that to (sort of) dynamically get the value you need here.

NitroTears: I've justed added the commits with the review feedback changes and my try at that extension method. 
Here is the code snippet for quick reference.

`MainMenuBarView.axaml.cs`
```cs
checkBoxes.Add(new CheckBox()
{
    Content = $".{fileName}",
    IsChecked = ((FileTypes)item).GetConfigValue(ConfigurationState.Instance.Ui.ShownFileTypes),
    Command = MiniCommand.Create(() => ViewModel.ToggleFileType(fileName))
});
```

`FileTypes.cs`
```cs
public static bool GetConfigValue(this FileTypes type, ShownFileTypeSettings config)
{
    return ((ReactiveObject<bool>) typeof(ShownFileTypeSettings).GetProperty(Enum.GetName(type))
        .GetValue(config, null)).Value;
}
```

It does look a tad messy, in exchange for it being more consice. I have added some documentation to the function as well. Let me know what you think.

TSRBerry: This is what I had in mind:

```cs
public static bool GetConfigValue(this FileTypes type, ShownFileTypeSettings config) => type switch
{
    FileTypes.NSP => config.NSP.Value,
    FileTypes.PFS0 => config.PFS0.Value,
    FileTypes.XCI => config.XCI.Value,
    FileTypes.NCA => config.NCA.Value,
    FileTypes.NRO => config.NRO.Value,
    FileTypes.NSO => config.NSO.Value,
    _ => throw new ArgumentOutOfRangeException(nameof(type), type, null)
};
```
This also avoids the use of reflection.

Could you move the extensions into a new file? I think that would be more in line with the rest of the extensions we added in this project.

NitroTears: That's all done now 👍🏻

TSRBerry (approved):
Lgtm, thank you!

Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1336,6 +1336,22 @@ public void ChangeLanguage(object languageCode)
             }
         }
 
+        public void ToggleFileType(string fileType)
+        {
+            _ = fileType switch
```
marysaka, line 1341: Can you fix the warning here?

marysaka (changes requested):
Only one thing.

Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
NitroTears: Added the suggested default case, and an argument out of range exception for it.

NitroTears:
Just checking if there is anything else? @marysaka

marysaka (approved):

TSRBerry:
> Just checking if there is anything else? @/marysaka

Make sure to re-request reviews if you addressed everything, otherwise it will be hard to notice you are done.

PR #4556, opened 2023-03-18
Eliminate boxing allocations caused by ISampledData structs

jhorv:
There are 7 structs used to hold input device data. These structs implemented interface `ISampledData` so that `AtomicStorage<T>` could access a `ulong SamplingNumber` field. C# implicitly boxes value types any time they are accessed through an interface, resulting in an allocation. These input device structs are so heavily used that on this developer's machine it results in between 30,000-40,000 boxing memory allocations _per second_ while a game is running.

This PR eliminates all of these allocations by:

- replacing the `ISampledData` interface with `ISampledDataStruct`
    - `ISampledDataStruct` contains no instance members - it is a "marker interface" only.
    - `ISampledDataStruct` is used to constrain static method `ISampledDataStruct.GetSamplingNumber(ref T sampledDataStruct)`, which casts a struct to a ReadOnlySpan<byte> and reads the first 8 bytes as the `SamplingNumber` field.
    - Thus this is a convention-based interface - implementers of `ISampledDataStruct` must follow the convention that the first 8 bytes of the struct point to a `ulong SamplingNumber`.
- AtomicStorage<T> and RingLifo<T> are now constrained to `unmanaged, ISampledDataStruct`
- Changed AtomicStorage<T> to use `ISampledDataStruct.GetSamplingNumber(ref T sampledDataStruct)` for getting a value's SamplingNumber where it was using the `ISampledData` interface before.

### DIFF
--- a/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
@@ -6,18 +6,27 @@ namespace Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Common
 {
     /// <summary>
     /// This is a "marker interface" to add some compile-time safety to a convention-based optimization.
-    /// Any struct implementing this interface must use its first 8 bytes as a "Sampling Number".
-    /// This is most often accomplished by using <c>StructLayoutAttribute</c> and careful member ordering,
-    /// such that the first member is a <c>ulong</c> (.NET type: <c>System.UInt64</c>) field.
+    ///
+    /// Any struct implementing this interface should:
+    ///   - use <c>StructLayoutAttribute</c> (and related attribtues) to explicity control how the struct is laid out in memory.
+    ///   - ensure that the method <c>ISampledDataStruct.GetSamplingNumberFieldOffset()</c> correctly returns the offset, in bytes,
+    ///     to the ulong "Sampling Number" field within the struct. Most types have it as the first field, so the default offset is 0.
     /// 
     /// Example:
     /// 
     /// <c>
-    ///         [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    ///         [StructLayout(LayoutKind.Sequential, Pack = 8)]
     ///         struct DebugPadState : ISampledDataStruct
     ///         {
-    ///             public ulong SamplingNumber;
-    ///             
+    ///             public ulong SamplingNumber;    // 1st field, so no need to add special handling to GetSamplingNumberFieldOffset()
+    ///             // other members...
+    ///         }
+    ///
+    ///         [StructLayout(LayoutKind.Sequential, Pack = 8)]
+    ///         struct SixAxisSensorState : ISampledDataStruct
+    ///         {
+    ///             public ulong DeltaTime;
+    ///             public ulong SamplingNumber;    // Not the first field - needs special handling in GetSamplingNumberFieldOffset()
     ///             // other members...
     ///         }
     /// </c>
@@ -32,9 +41,25 @@ namespace Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Common
 
             ReadOnlySpan<byte> byteSpan = MemoryMarshal.Cast<T, byte>(structSpan);
 
+            int fieldOffset = GetSamplingNumberFieldOffset(ref sampledDataStruct);
+
+            if (fieldOffset > 0)
+            {
+                byteSpan = byteSpan.Slice(fieldOffset);
+            }
+
             ulong value = BinaryPrimitives.ReadUInt64LittleEndian(byteSpan);
 
             return value;
+        }
+
+        private static int GetSamplingNumberFieldOffset<T>(ref T sampledDataStruct) where T : unmanaged, ISampledDataStruct
+        {
+            return sampledDataStruct switch
+            {
+                Npad.SixAxisSensorState _ => sizeof(ulong),
+                _ => 0
+            };
         }
     }
 }
--- a/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Buffers.Binary;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Common
+{
+    /// <summary>
+    /// This is a "marker interface" to add some compile-time safety to a convention-based optimization.
+    /// Any struct implementing this interface must use its first 8 bytes as a "Sampling Number".
+    /// This is most often accomplished by using <c>StructLayoutAttribute</c> and careful member ordering,
+    /// such that the first member is a <c>ulong</c> (.NET type: <c>System.UInt64</c>) field.
+    /// 
+    /// Example:
+    /// 
+    /// <c>
+    ///         [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    ///         struct DebugPadState : ISampledDataStruct
+    ///         {
+    ///             public ulong SamplingNumber;
+    ///             
+    ///             // other members...
+    ///         }
+    /// </c>
+    /// </summary>
+    internal interface ISampledDataStruct
+    {
+        // No Instance Members - marker interface only
+
+        public static ulong GetSamplingNumber<T>(ref T sampledDataStruct) where T : unmanaged, ISampledDataStruct
+        {
+            ReadOnlySpan<T> structSpan = MemoryMarshal.CreateReadOnlySpan(ref sampledDataStruct, 1);
+
+            ReadOnlySpan<byte> byteSpan = MemoryMarshal.Cast<T, byte>(structSpan);
+
+            ulong value = BitConverter.IsLittleEndian
+                ? BinaryPrimitives.ReadUInt64LittleEndian(byteSpan)
+                : BinaryPrimitives.ReadUInt64BigEndian(byteSpan);
+
+            return value;
+        }
+    }
+}
--- a/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
+++ b/Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
@@ -32,9 +32,7 @@ namespace Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Common
 
             ReadOnlySpan<byte> byteSpan = MemoryMarshal.Cast<T, byte>(structSpan);
 
-            ulong value = BitConverter.IsLittleEndian
-                ? BinaryPrimitives.ReadUInt64LittleEndian(byteSpan)
-                : BinaryPrimitives.ReadUInt64BigEndian(byteSpan);
+            ulong value = BinaryPrimitives.ReadUInt64LittleEndian(byteSpan);
 
             return value;
         }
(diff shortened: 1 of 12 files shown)


Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Common/ISampledDataStruct.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        // No Instance Members - marker interface only
+
+        public static ulong GetSamplingNumber<T>(ref T sampledDataStruct) where T : unmanaged, ISampledDataStruct
+        {
+            ReadOnlySpan<T> structSpan = MemoryMarshal.CreateReadOnlySpan(ref sampledDataStruct, 1);
+
+            ReadOnlySpan<byte> byteSpan = MemoryMarshal.Cast<T, byte>(structSpan);
+
+            ulong value = BitConverter.IsLittleEndian
```
gdkchan, line 35: Since this value is visible to the game, and the Switch is little endian, I don't think that value could be possibly stored in big endian form in memory. If it was running on a big endian architecture, we'd probably need to make sure all those are written as little endian.

marysaka (approved):

lostromb:
This is interesting, though I can't seem to reproduce the boxing behavior that you're describing. It seems to me like it's quite possible to just return the value through the interface without boxing.... at least, that's what my test code always does. Is the compiler optimizing something away? What am I missing?
```
interface IProducer
{
    ulong Value { get; }
}

struct Producer : IProducer
{
    public ulong Value => (ulong)Random.Shared.Next();
}

IProducer producer = new Producer();
ulong sum = 0;
while (true) sum += producer.Value;
```

jhorv:
> This is interesting, though I can't seem to reproduce the boxing behavior that you're describing. It seems to me like it's quite possible to just return the value through the interface without boxing.... at least, that's what my test code always does. Is the compiler optimizing something away? What am I missing?

In your test case I believe the compiler is smart enough to know the real type is Producer, because it's being created right there.

If you look at [the spreadsheet](https://github.com/Ryujinx/Ryujinx/files/10997151/RyujinxComparison_master-perfMemory.xlsx) I attached to my previous PR #4556 , tab "master-d56d335c SOH" has 5 rows demonstrating the boxing. Copying below. These are all induced by the before-PR's AtomicStorage.cs lines 17-19.

| Type | Allocations|
| ---------------------------------------------------------------------------- | --------- |
| Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.DebugPad.DebugPadState       | 23,367    |
| Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Mouse.MouseState             | 23,367    |
| Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Npad.NpadCommonState         | 1,425,608 |
| Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.Npad.SixAxisSensorState      | 1,402,309 |
| Ryujinx.HLE.HOS.Services.Hid.Types.SharedMemory.TouchScreen.TouchScreenState | 46,203    |

lostromb:
I've been trying a bunch of things, including setting the same `IProducer` reference to a value the compiler can't possibly predict, and my code is still allocation-free.
```
IProducer producerByIface;
while (true)
{
    producerByIface = CreateRandomProducer(rand);
    producerByIface.Value.GetHashCode();
}
```

The only thing that triggers boxing is if I change the interface to return `object` instead of the value type. I just want to try to understand what's going on here; I don't doubt your fix solves the problem, but it's also a rather unintuitive pattern for anyone looking at the code to try and understand.

lostromb:
Okay, I think I understand now. I was looking for allocations in the `while (true) sum += producer.Value;` line in my test code, when the allocation actually happened at `IProducer producer = new Producer();`. In other words, casting a struct which implements an interface to a reference of that interface type. Which, of course, is covered by the first sentence of the [documentation of boxing](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing).

jhorv:
> Okay, I think I understand now. I was looking for allocations in the `while (true) sum += producer.Value;` line in my test code, when the allocation actually happened at `IProducer producer = new Producer();`. In other words, casting a struct which implements an interface to a reference of that interface type. Which, of course, is covered by the first sentence of the [documentation of boxing](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing).

You will see a boxing allocation in the `while (true) sum += producer.Value;` line if you move it into a separate function, where I presume the compiler will no longer have immediate visibility of the `new Producer()` line.

Like this:

```
static void AccessValue(IProducer p)
{
    while (true) sum += producer.Value;       // This will box if the actual type of p is a struct
}
```

jhorv:
Basically what it comes down to is: accessing a type's member through an interface requires virtual dispatch - the same mechanism that allows virtual methods work. Virtual dispatch works on reference types, not value types. So, a struct will be boxed so that it satisfies that requirement.

For similar reasons, a struct will be boxed if you call any of the built-in virtual methods like Equals(object), GetHashCode(), and ToString()  - _if_ your struct doesn't override them. Those methods come from System.Object, and if your struct doesn't provide overrides then virtual dispatch gets involved to find out which base class provides them. Virtual dispatch needs it as a reference type, so it gets boxed.

Structs have many subtle caveats. That's why ValueTuples and record structs are so nice - they implement a lot of the best practices for you automatically.

riperiperi:
I have no idea why, but when I start Breath of the Wild or Mario Kart 8 with this branch, the emulator just gets stuck and it doesn't draw any frames. The last stuff it prints is HID related, so I guess something has gone wrong. (BOTW)

```
00:01:08.197 |S| HLE.OsThread.9 ServiceHid ActiveNpadImpl: Stubbed. {appletResourceUserId: 1 ; revision: 2}
00:01:08.200 |S| HLE.OsThread.9 ServiceHid SetSupportedNpadIdType: Stubbed. 9 Players: Player1,Player2,Player3,Player4,Player5,Player6,Player7,Player8,Handheld
00:01:08.201 |S| HLE.OsThread.9 ServiceHid SetSupportedNpadStyleSet: Stubbed. {pid: 96 ; appletResourceUserId: 1 ; type: ProController, Handheld, JoyconPair, JoyconLeft, JoyconRight}
00:01:08.217 |S| HLE.OsThread.9 ServiceHid StartSixAxisSensor: Stubbed. {appletResourceUserId: 1 ; sixAxisSensorHandle: 5}
00:01:08.218 |S| HLE.OsThread.9 ServiceHid StartSixAxisSensor: Stubbed. {appletResourceUserId: 1 ; sixAxisSensorHandle: 65541}
```

Ryujinx.HLE/HOS/Services/Hid/Types/SharedMemory/Npad/SixAxisSensorState.cs
```diff
@@ -1,19 +1,21 @@
...
 {
-    struct SixAxisSensorState : ISampledData
+    [StructLayout(LayoutKind.Sequential, Pack = 1)]
+    struct SixAxisSensorState : ISampledDataStruct
     {
-        public ulong DeltaTime;
+        // MUST BE THE 1st MEMBER
         public ulong SamplingNumber;
+
+        public ulong DeltaTime;
```
riperiperi, line 13: You can't change the layout of this struct, it needs to be the same layout as on the guest.

Changing this causes games that use sixaxis sensor state (mario kart 8, botw) to freeze on boot.

riperiperi (changes requested):

jhorv:
Commit [1a4c5ef](https://github.com/Ryujinx/Ryujinx/pull/4556/commits/1a4c5ef79c2dd1b1017371731a4360594f492c2a) makes significant changes to this PR:

- Restored the original field order (V1 of this PR changed the field order of the `SixAxisSensorState` struct)
- Removed comments in each of the 7 ISampledDataStruct types that warn `// MUST BE THE 1st MEMBER` above the `SamplingNumber` field. The presence of the `StructLayoutAttribute` on the type implies the field order is important.
- On the 7 struct types, changed Pack = 1 to Pack = 8. This is because I found that at Pack=1, the `KeyboardState` struct was 44 bytes in size, vs. 48 when StructLayout wasn't set at all. With Pack=8, all 7 structs are the same size as they were before this PR, and it seems reasonable that the native hardware might align same as Pack=8 (though this is still an assumption on my part! It could be that the native hardware expects Pack=1 or Pack=4 - `KeyboardState` is the only 1 of the 7 structs that has a different size depending on the Pack value. Do we have any strange issues reported where KeyboardState is involved?)
- Changed the `ISampledDataStruct` xml docs to describe how an implementor should comply with this convention.
- Changed `ISampledDataStruct.GetSamplingNumber()` to get a given type's `SamplingNumber` field offset from a private method, `GetSamplingNumberFieldOffset()`.

I've benchmarked the original implementation, the broken v1 of this PR, and this revised PR:


|                                         Method | Iterations |     Mean |   Error |  StdDev |     Gen0 |   Gen1 | Allocated |
|----------------------------------------------- |----------- |---------:|--------:|--------:|---------:|-------:|----------:|
| Control_StructFieldAccessedThroughISampledData |     100000 | 382.6 μs | 1.10 μs | 0.92 μs | 190.9180 | 0.4883 | 3200001 B |
|                PRVersion1_AssumeFieldAtOffset0 |     100000 | 145.1 μs | 0.22 μs | 0.20 μs |        - |      - |         - |
|             PRVersion2_LookupFieldOffsetByType |     100000 | 159.3 μs | 0.84 μs | 0.79 μs |        - |      - |         - |


This shows that this revised PR:
- works with the original struct field order
- eliminates the boxing allocations
- is more than twice as fast as the pre-PR version

marysaka (approved):

riperiperi:
>  Do we have any strange issues reported where KeyboardState is involved?)

Keyboard state does not work properly in most games. It's also hard to test since it only works when a controller's input source is set to keyboard, which is counterintuitive as you only really want to use it when the keyboard _isn't_ being used as a controller.

riperiperi (approved):
LGTM. I'm not sure if these should be using Pack = 8. The keyboard stuff should probably be looked into and fixed separately.

jhorv:
> LGTM. I'm not sure if these should be using Pack = 8. The keyboard stuff should probably be looked into and fixed separately.

What should we do with the KeyboardState pack? Here is a handy table of what the size of KeyboardState ends up being with different Pack values.

Pack | Size of KeyboardState
| -- | -- |
0 (as before this PR) | 48
1 | 44 bytes
4 |  44 bytes
8 | 48 bytes

The default of 0 means "default for the current platform", I'm sure we don't want that. Pack = 8 is the only one that results in it being the same size as before this PR. I can set Pack = 1 (as I've seen elsewhere in the codebase), but that seems likely to break something. With all that said, please tell me what you'd like it to be and I'll make the change.

jhorv:
I'll do some more digging and see where the problem is with KeyboardState. I set Pack = 8 to all the types because it seemed logical, but mostly because it was required to make KeyboardState stay the same size. If I can figure out that KeyboardState is missing 4 bytes worth of something that would make it work with Pack = 1, then I'll adjust this PR to make them all Pack = 1.

jhorv:
I pushed a commit with the following changes:

- Changed all the ISampledDataStruct types to Pack = 1
- added a `uint Unknown` field to `KeyboardState`

I added the `KeyboardState` field based on reasoning that with `Pack = 0` the compiler adds a 32-bit pad between `Modifiers` and `Keys`. To make it work with `Pack = 1`, I explicitly added a `uint Unknown` field between `Modifiers` and `Keys` and based a small bit of testing, this appears correct. BotW and Metroid Prime Remastered both start/run fine, with input set to either a PS5 controller or a physical keyboard.

I think I don't understand how this `KeyboardState` struct is used though - I tried running with the debugger attached to inspect that `uint`-sized space, but I couldn't get any breakpoints to hit. I'm sure someone else knows why, but for now the field is just `Unknown` to me. If it contains data, it apparently isn't that important.

riperiperi:
One game that uses the keyboard is Hypnospace Outlaw. Animal Crossing never worked so you can't verify if anything changed with that.

jhorv:
I've manually verified that the pre-PR and post-PR `KeyboardState` struct layout is identical. I don't have any games that use `KeyboardState` to see if the padding space is actually a field, but I think that can be tackled in a future PR if/when the  need arises.

gdkchan:
The correct size of `KeyboadState` is 48 bytes according to switchbrew: https://switchbrew.org/wiki/HID_services#KeyboardState
The keyboard state is updated at https://github.com/Ryujinx/Ryujinx/blob/51b3953cfc3b440aa0b7b1b06d33626426e54556/Ryujinx.Input/HLE/NpadManager.cs#L234

jhorv:
Just pushed a commit that:

- makes `KeyboardModifier` a ulong enum instead of uint
- removes the `uint Unknown` I added to `KeyboardState` for padding

riperiperi (approved):

PR #4564, opened 2023-03-18
Avoid copying more handles than we have space for

TSRBerry:
This small PR is supposed to fix a regression introduced by #4537.

@riperiperi noticed this:
> _sessionHandles can be modified from other threads, before it would be converted to array then copied, but now it will create an array with the current size, then error out if it copies and it changed to be larger.

At first I used two locks for `_sessionHandles` and `_portHandles`, but thought it might be better to just ensure we are only copying the handles we reserved space for even if the size of the lists change.

I'm not quite sure which approach is better, so feel free to tell me to change it to locks (or something else I'm not thinking of).

### DIFF
--- a/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -32,6 +32,8 @@ namespace Ryujinx.HLE.HOS.Services
             0x01007FFF
         };
 
+        private readonly object _handleLock = new();
+
         private readonly KernelContext _context;
         private KProcess _selfProcess;
 
@@ -77,7 +79,7 @@ namespace Ryujinx.HLE.HOS.Services
 
         private void AddPort(int serverPortHandle, Func<IpcService> objectFactory)
         {
-            lock (_portHandles)
+            lock (_handleLock)
             {
                 _portHandles.Add(serverPortHandle);
             }
@@ -95,7 +97,7 @@ namespace Ryujinx.HLE.HOS.Services
 
         public void AddSessionObj(int serverSessionHandle, IpcService obj)
         {
-            lock (_sessionHandles)
+            lock (_handleLock)
             {
                 _sessionHandles.Add(serverSessionHandle);
             }
@@ -132,8 +134,7 @@ namespace Ryujinx.HLE.HOS.Services
 
             while (true)
             {
-                lock (_portHandles)
-                lock (_sessionHandles)
+                lock (_handleLock)
                 {
                     int handleCount = _portHandles.Count + _sessionHandles.Count;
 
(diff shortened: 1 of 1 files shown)


gdkchan:
> but thought it might be better to just ensure we are only copying the handles we reserved space for even if the size of the lists change.

I don't think that would be correct either since the number of handles on the list might be lower by the time it does the copy, which would it cause it to fail too. Also I don't know if the list `CopyTo` method is thread safe, which could also cause it to fail internally if the list is modified while it's inside that method.

Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -126,54 +126,60 @@ private void ServerLoop()
 
             while (true)
             {
-                int handleCount = _portHandles.Count + _sessionHandles.Count;
+                lock (_portHandles)
+                {
+                    lock(_sessionHandles)
```
TSRBerry, line 131:
```suggestion
                lock (_portHandles)
                lock(_sessionHandles)
```
Would it also work if I write it like this? Just to reduce nesting

TSRBerry: To answer my own question: the IL looks the same, so I don't see any reason why it wouldn't work.

gdkchan:
The code is currently still wrong since the place that actually modifies those lists don't have a lock, so the lock is not actually protecting anything. Also it could be better to use the same lock for both lists to avoid the need to lock twice.

TSRBerry:
> The code is currently still wrong since the place that actually modifies those lists don't have a lock, so the lock is not actually protecting anything. Also it could be better to use the same lock for both lists to avoid the need to lock twice.

Just read this now, there aren't any other locks missing, right? I double checked and I'm pretty sure I surrounded everything properly with locks. 

Oh and I'll replace the locks with one lock object, so we don't need to lock twice here.

gdkchan (approved):
Looks good now. There's also the dictionary (`_ports` and `_sessions`) access that also is (and always has been) wrong, if sessions can be added from different threads. I'm not sure which case causes it to do that from different threads (the code was clearly written with the assumption it can't happen).

Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -126,54 +134,57 @@ private void ServerLoop()
...
+                    int[] handles = ArrayPool<int>.Shared.Rent(handleCount);
 
-                // We still need a timeout here to allow the service to pick up and listen new sessions...
-                var rc = _context.Syscall.ReplyAndReceive(out int signaledIndex, handles.AsSpan(0, handleCount), replyTargetHandle, 1000000L);
+                    _portHandles.CopyTo(handles, 0);
+                    _sessionHandles.CopyTo(handles, _portHandles.Count);
 
-                thread.HandlePostSyscall();
+                    // We still need a timeout here to allow the service to pick up and listen new sessions...
+                    var rc = _context.Syscall.ReplyAndReceive(out int signaledIndex, handles.AsSpan(0, handleCount), replyTargetHandle, 1000000L);
```
gdkchan, line 147: This call will block execution. Blocking execution while holding a lock is bad and can easily cause deadlocks. Ideally you should release the lock as soon as possible.

marysaka (approved):

PR #4565, opened 2023-03-18
Refactor attribute handling on the shader generator

gdkchan:
This change attempts to address some shortcomings in the way how "attributes" are handled on the shader translator.
Some of the issues are:
- There are two ways to represent "attributes" on the IR. Either using the Attribute operand types, or using the `LoadAttribute` and `StoreAttribute` instructions. Those instructions must exist because in some cases, loading the value (or storing) requires multiple parameters. On geometry or tessellation shader, you need to know which vertex to load (since it has one invocation per primitive, not per vertex). In some cases you might be accessing an array attribute where the index is not constant, etc. Having two ways to represent the same thing can be a source of bugs, since sometimes you might check one but forget to check the other, or have a way to "canonicalise" how it is represented in the IR to avoid having to check both in several places.
- It doesn't really match any shading language. The GPU has an attribute buffer that stores per-vertex data. It also has system registers that usually stores per invocation (or some higher scope) data. On the IR, all those use `Attribute` operands. Since since registers don't really have an "offset", we invent offsets that are stored on the `Attribute` operand. There's nothing inherently wrong with that, it's just somewhat ugly.

Main improvements
---

This change removes the `Attribute` operand types. Now all attributes access are done using the new `Load` and `Store` IR instructions. This also a cleaner way to load from a output. Initially I assumed that you could only load from inputs, and store to outputs, so all the code made that assumption. But it is actually possible to load from outputs, so I added a flag on the attribute value to indicate it should actually load from the output, not the input. This was also a source of bugs in the past because not all the code was aware of the existence of that flag, and failing to mask it off would cause it to try accessing invalid attributes. The change on this PR removes the assumption entirely, now the actual instruction is entirely irrelevant for the storage it will access. If the storage type is output, it will load (or store) from the output.

Previously, the storage that was accessed was stored within the instruction itself, as a separate bit field on the value. This masking and shifting is required every time one needs to access the storage type from the instruction. It also means care must be taken when comparing instructions, as it's possible that you don't care about the storage type, so you would need to mask it out in those cases. Now, a separate field has been added on the `Operation` for the storage type. This eliminates the need to mask and shift to access the value. In the future, I plan to do something similar with the operation type that is also stored within the instruction itself (like FP32 and FP64).

The new `Load` and `Store` instructions will have all indices used to access the scalar value on the attribute. This is a closer match to the actual shading languages, where we don't have raw byte offsets but actual types we can index into. For example, to access a given user input or output, we at least 2 indices, the first one being the location, the second one being the vector element index. Geometry shaders will have an additional one in the middle for inputs, which is the vertex # in the current primitive. This also matches the logical order they would be accessed, which means it's possible to feed those indices into `OpAccessChain` operations almost directly on SPIR-V for example.

Other improvements
---

- GLSL now supports `gl_ViewportIndex`. It will be used as long the host GPU supports it. Before it was ignored, and was only used on SPIR-V.
- Both GLSL and SPIR-V now supports `gl_ViewportMask`, as long the `NV_viewport_array2` extension is supported.
- Invalid or unsupported attributes will now log a warning. Before they would be just silently ignored.
- Invalid attribute usage for a given shader stage will now log a warning, and read as zero/discard writes. Before it would produce a invalid shaders, which would fail to compile and potentially crash. Note that those errors are not actually expected, if a shader is invalid there's likely something else wrong, but I believe this is a nicer way to handle it.
- Valid attributes that are not supported by the host also produces a warning now. Before they would read as zero/discard writes, but wouldn't produce any warning.
- Added partial support for `gl_FogFragCoord`. Before it was just ignored, now it is converted to a "general" attribute. This is legacy functionality that is no longer supported in modern APIs, so we will likely need to emulate it.

Fixes
---

- A compilation failures for tessellation shaders on MoltenVK, caused by use of subgroup inputs, even when the generated shader code doesn't actually use it, has been fixed. (See screenshots below).
- An error that would cause fixed function inputs/outputs to not work on older GPUs that don't support setting gl_Layer on the vertex shader was fixed. (See @lostromb comment).

Before:
Captura de Tela 2023-03-19 às 12 22 17
After:
Captura de Tela 2023-03-19 às 12 20 19

Future plans
---

I have plans for other similar changes. If desirable, I can make some kind of umbrella issue detailing them. It might be a good way to gather feedback before I start working on it (well, I should probably have done that before working on *this* change).

- Use new `Load`/`Store` instructions for constant buffer. In fact, they suffer from more or less the same issue as attributes. There is a constant buffer operand type, and a `LoadConstant` IR instruction which is used to load with non-constant offsets.
  - Changing this has the advantage of allowing us to access the support buffer (which is defined as an actual structure, not a raw block of data, like `vec4[4096]`). The way how it would work is that each structure field has a index, and the `Load` instruction would specify the index of the field to access, rather than a raw byte offset. That's how it works on SPIR-V when we use `OpAccessChain`. On GLSL we just need to resolve the field name from the index, which is also rather simple.
- Use new `Load`/`Store` instructions for the remaining storage types (global memory, that becomes storage buffer, shared memory and local memory).
- Stop using guest buffer slot numbers, and use the binding number directly on the IR. This would allow us to access other emulator reserved buffers from IR code directly in a clean way. Currently they need special handling on the backend (we have special attribute offsets that maps to specific fields that we need to access on the support buffer, for example).
  - This will also be required for macOS, where we need additional emulator reserved buffers for various tasks, like transform feedback emulation. I had to do something very ugly to make it work there, using special slot numbers which later down the line take unused storage buffer slots. This was causing crashes on games where all storage buffers slots were in use.
  - In combination with the first constant buffer change, this will allow us to stop accessing constant buffer fields with attributes. This also means we get access to the entire support buffer without having to add handling for each field on both GLSL and SPIR-V (or any other language we may implement in the future). This in turn allow us to move some helper functions, such as the scaling helpers, out of the backend and generate them in the IR instead, further reducing the amount of backend specific code and unblocking other optimisations.

Testing
---

Testing is welcome to ensure there are no regressions. I'm not really expecting any improvement from this other than the few things I mentioned, this is mostly just refactoring to make the code cleaner and hopefully easier to maintain.


lostromb:
> An error that would cause fixed function inputs/outputs to not work on older GPUs that don't support setting gl_Layer on the vertex shader should be fixed. (To be tested).

I have an nvidia GT 640, would that help in testing this? (I assume it's the same "pokemon violet doesn't render at all" bug that was fixed earlier)

gdkchan:
> > An error that would cause fixed function inputs/outputs to not work on older GPUs that don't support setting gl_Layer on the vertex shader should be fixed. (To be tested).
> 
> I have an nvidia GT 640, would that help in testing this? (I assume it's the same "pokemon violet doesn't render at all" bug that was fixed earlier)

This issue was aused by that PR yes, but it's a different issue. That graphics card should be suitable to test this, but you'll need to test a game that uses one of the legacy attributes, like Dragon Quest 3.

lostromb:
Seems fine.
ryujinx_capture_2023-03-20_17-26-50
[Ryujinx_1.1.0+fcc635e_2023-03-20_17-28-56-opengl.log](https://github.com/Ryujinx/Ryujinx/files/11024034/Ryujinx_1.1.0%2Bfcc635e_2023-03-20_17-28-56-opengl.log)

Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
```diff
@@ -441,8 +464,11 @@ private void GenerateAlphaToCoverageDitherDiscard()
             // 11 01 01 01 01 00 00 00
             Operand ditherMask = Const(unchecked((int)0xfbb99110u));
 
-            Operand x = this.BitwiseAnd(this.FP32ConvertToU32(Attribute(AttributeConsts.PositionX)), Const(1));
-            Operand y = this.BitwiseAnd(this.FP32ConvertToU32(Attribute(AttributeConsts.PositionY)), Const(1));
+            Operand fragCoordX = this.Load(StorageKind.Input, IoVariable.FragmentCoord, null, Const(0));
+            Operand fragCoordY = this.Load(StorageKind.Input, IoVariable.FragmentCoord, null, Const(0));
```
riperiperi, line 468: Are these meant to have a different component?

Ryujinx.Graphics.Shader/StructuredIr/StructuredProgramContext.cs
```diff
@@ -317,35 +300,11 @@ public AstOperand NewTemp(AggregateType type)
...
-            }
-            else if (operand.Type == OperandType.AttributePerPatch)
-            {
-                Info.OutputsPerPatch.Add(operand.Value & AttributeConsts.Mask);
-            }
-
             return GetOperand(operand);
         }
 
         public AstOperand GetOperandUse(Operand operand)
```
riperiperi, line 306: These two methods are now the same, is there any reason to keep them separate? (apart from avoiding modifying every use)

Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
```diff
@@ -643,30 +647,28 @@ private static void DeclareOutputAttribute(CodeGenContext context, int attr)
...
                     context.AppendLine($"layout (location = {attr}{xfb}) out {type} {name};");
                 }
-                else
-                {
-                    for (int c = 0; c < 4; c++)
-                    {
-                        char swzMask = "xyzw"[c];
 
-                        string xfb = string.Empty;
+                for (int c = components > 1 ? components : 0; c < 4; c++)
```
riperiperi, line 659: Is this right? If there is one component, this will start on 0 (0, 1, 2, 3), but if there are 2 it will start on 2 (2, 3).

riperiperi (commented):
I think this is a much nicer way to do things. I've looked over everything and left comments, though I think this needs a lot of testing and I haven't done that yet.

Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
gdkchan: Yes, if components is 1, then all outputs are float with a explicit component number (0, 1, 2, 3). If components is greater than 1, then the first output is a vector with that number of components. The remaining outputs are floats with a explicit components. For example, if components is 2, first output is vec2, 3rd and 4th ones are floats with components 2 and 3. If components is 4, then the first output is vec4 and there are no remaining components for that vector.

Ryujinx.Graphics.Shader/StructuredIr/StructuredProgramContext.cs
gdkchan: Not really. I just kept them in case we need the distinction of which operand is being assigned to and which one is just a use in the future. But I don't foresee it being needed so I can remove it.

Ryujinx.Graphics.Shader/IGpuAccessor.cs
```diff
@@ -335,7 +335,16 @@ bool QueryHostSupportsTextureShadowLod()
         /// Queries host GPU shader viewport index output support.
```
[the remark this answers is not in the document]
gdkchan: I'm not sure if I understand the question, but this is a situation similar to `gl_Layer` where the core spec supports it on the geometry (and output) and fragment (as input) stages. Later a extension has been added that also allows modifying it on the vertex and tessellation stages. Not all GPUs supports it, so this additional capability check is needed.

I will change the comment to clarify it's just on vertex and tessellation.

marysaka (approved):
lgtm but need a rebase.

PR #4569, opened 2023-03-19
Ava: Fix nca extraction window never closing & minor cleanup

TSRBerry:
Thank you to @Niwu34 for reporting this issue!

This PR replaces the confirmation dialog which was used for nca extraction with `UpdateWaitWindow`.
I adjusted this window a little bit, so it's now able to cancel stuff using a `CancellationTokenSource`.

I also tried to make the swkbd dialog auto focus the input box, but there seems to be a bug with Avalonia's `DialogHost`, so it never worked. Let's hope the next Avalonia version also fixes this. Regardless I cleaned up some leftover code I found in there as well and decided to include that in this PR.


marysaka (approved):

riperiperi (approved):
Looks good.

AcK77 (approved):

PR #4577, opened 2023-03-22
ServerBase thread safety

jhorv:
@lostromb [commented](https://github.com/Ryujinx/Ryujinx/pull/4537#issuecomment-1477130450) on closed PR #4537 about an intermittent exception when stopping emulation. This PR fixes some issues I found while investigating, and though I was unable to reproduce @lostromb's reported exception, I believe this will fix it.

- ServerBase.Dispose() could be (and often was) called multiple times.
- ServerBase.Dispose() had no checks to ensure the ServerLoop() had stopped first.
- Reworked the port and session storage to use a Dictionary only, instead of a Dictionary and List
- Reworked the locking to use a ReaderWriterLockSlim, and made every access of the port and session stores thread-safe.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -201,9 +201,9 @@ namespace Ryujinx.HLE.HOS.Services
                     handleLockTaken = _handleLock.TryEnterReadLock(Timeout.Infinite);
 
                     portHandleCount = _ports.Count;
-                    
+
                     handleCount = portHandleCount + _sessions.Count;
-                    
+
                     handles = ArrayPool<int>.Shared.Rent(handleCount);
 
                     _ports.Keys.CopyTo(handles, 0);
@@ -336,7 +336,7 @@ namespace Ryujinx.HLE.HOS.Services
                     _requestDataReader,
                     _responseDataWriter);
 
-                _sessions[serverSessionHandle].CallCmifMethod(context);
+                GetSessionObj(serverSessionHandle).CallCmifMethod(context);
 
                 response.RawData = _responseDataStream.ToArray();
             }
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -488,7 +488,7 @@ namespace Ryujinx.HLE.HOS.Services
             {
                 if (!_threadStopped.Wait(ThreadStoppedWaitTimeout))
                 {
-                    Logger.Warning.Value.PrintRawMsg($"the ServerBase thread didn't signal as stopped within {ThreadStoppedWaitTimeout:g}, resources will be leaked!");
+                    Logger.Warning?.Print(LogClass.Service, $"the ServerBase thread didn't signal it has stopped within {ThreadStoppedWaitTimeout:g}, resources may be leaked!");
                 }
                 else if (Interlocked.Exchange(ref _isDisposed, 1) == 0)
                 {
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -33,11 +33,12 @@ namespace Ryujinx.HLE.HOS.Services
             0x01007FFF
         };
 
-        // The amount of time that Dispose() will wait on the _threadStopped wait handle.
-        private static readonly TimeSpan ThreadStoppedWaitTimeout = TimeSpan.FromSeconds(3);
+        // The amount of time Dispose() will wait to Join() the thread executing the ServerLoop()
+        private static readonly TimeSpan ThreadJoinTimeout = TimeSpan.FromSeconds(3);
 
         private readonly KernelContext _context;
         private KProcess _selfProcess;
+        private KThread _selfThread;
 
         private readonly ReaderWriterLockSlim _handleLock = new ReaderWriterLockSlim();
         private readonly Dictionary<int, IpcService> _sessions = new Dictionary<int, IpcService>();
@@ -49,7 +50,6 @@ namespace Ryujinx.HLE.HOS.Services
         private readonly MemoryStream _responseDataStream;
         private readonly BinaryWriter _responseDataWriter;
 
-        private readonly ManualResetEventSlim _threadStopped;
         private int _isDisposed = 0;
 
         public ManualResetEvent InitDone { get; }
@@ -69,8 +69,6 @@ namespace Ryujinx.HLE.HOS.Services
             InitDone = new ManualResetEvent(false);
             Name = name;
             SmObjectFactory = smObjectFactory;
-
-            _threadStopped = new ManualResetEventSlim(false);
 
             const ProcessCreationFlags flags =
                 ProcessCreationFlags.EnableAslr |
@@ -167,8 +165,8 @@ namespace Ryujinx.HLE.HOS.Services
 
         private void ServerLoop()
         {
-            _threadStopped.Reset();
             _selfProcess = KernelStatic.GetCurrentProcess();
+            _selfThread = KernelStatic.GetCurrentThread();
 
             if (SmObjectFactory != null)
             {
@@ -179,8 +177,7 @@ namespace Ryujinx.HLE.HOS.Services
 
             InitDone.Set();
 
-            KThread thread = KernelStatic.GetCurrentThread();
-            ulong messagePtr = thread.TlsAddress;
+            ulong messagePtr = _selfThread.TlsAddress;
             _context.Syscall.SetHeapSize(out ulong heapAddr, 0x200000);
 
             _selfProcess.CpuMemory.Write(messagePtr + 0x0, 0);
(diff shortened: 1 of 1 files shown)


Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -48,6 +49,9 @@ class ServerBase : IDisposable
         private readonly MemoryStream _responseDataStream;
         private readonly BinaryWriter _responseDataWriter;
 
+        private readonly ManualResetEventSlim _threadStopped;
+        private int _isDisposed = 0;
```
lostromb, line 53: Is this used?
In my own code I typically adopt this pattern at the top of every Dispose() function to enforce single-dispose semantics using an integer flag like this:
`if (Interlocked.CompareExchange(ref _isDisposed, 1, 0) != 0) return;`
I assume you were going for a similar design.

lostromb:
I know in the past when I looked at the shutdown sequence, I found it was not a simple case of just making sure everything was disposed. There were cases such as queuing a FIFO message to the GPU thread to start disposing of textures, handles, etc. but then also we were trying to dispose of the FIFO queue itself at the same time. I imagine there could be similar cases where we're trying to dispose of host resources and IPC channels and such while they are already half-disposed, which might lead to thread contention or deadlocks if you don't carefully control which mutexes get held during disposal routines. Does someone more familiar with the code know if this is a concern?

Ryujinx.HLE/HOS/Services/ServerBase.cs
jhorv: Yes, it's at the top of the Dispose(bool).

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -394,24 +486,31 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                foreach (IpcService service in _sessions.Values)
+                if (!_threadStopped.Wait(ThreadStoppedWaitTimeout))
                 {
-                    if (service is IDisposable disposableObj)
+                    Logger.Warning?.Print(LogClass.Service, $"the ServerBase thread didn't signal it has stopped within {ThreadStoppedWaitTimeout:g}, resources may be leaked!");
```
riperiperi, line 491:
```suggestion
                    Logger.Warning?.Print(LogClass.Service, $"The ServerBase thread didn't signal it has stopped within {ThreadStoppedWaitTimeout:g}, resources may be leaked!");
```

Also, it's not really "may be leaked" if you're skipping the clear and dispose of everything below.

riperiperi:
I encountered this error when closing BOTW with the latest CI build:
image

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -394,24 +486,31 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                foreach (IpcService service in _sessions.Values)
+                if (!_threadStopped.Wait(ThreadStoppedWaitTimeout))
```
riperiperi, line 489: This can be used after it's disposed.

riperiperi (changes requested):

riperiperi:
It still hung on close a bunch of times, but I think there are a few other reasons it can do that (mostly GAL related). I wasn't able to get it to crash.
image

riperiperi (approved):
Looks good. Honestly I'm not sure why these are being disposed multiple times in the first place...

jhorv:
> It still hung on close a bunch of times, but I think there are a few other reasons it can do that (mostly GAL related). I wasn't able to get it to crash.

I've noticed master without this PR hanging on close a bunch lately. Haven't looked into it yet.

lostromb:
I've considered in the past adding some kind of feedback signal to say "send dispose commands to gpu thread, wait for gpu thread to signal finished, then dispose of queue." But I think that's outside the scope of this review

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -394,24 +486,33 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                foreach (IpcService service in _sessions.Values)
+                if (Interlocked.Exchange(ref _isDisposed, 1) == 0)
                 {
-                    if (service is IDisposable disposableObj)
+                    if (!_threadStopped.Wait(ThreadStoppedWaitTimeout))
```
gdkchan, line 491: Can't this just call `Thread.Join` instead of waiting this event, since it wants to wait for the thread exit? You would need to add something like `_selfThread = KernelStatic.GetCurrentThread();` on the `ServerLoop` method, and then call `_selfThread.HostThread.Join();`.  It is of course, not needed if dispose is called from the same thread if would wait for (and might be even invalid to call `Join` in that case). Also it would be invalid if dispose is called before the thread starts, so it could be worth to add a null check too.

jhorv: Great idea, fixed in #5331f94

marysaka (approved):
lgtm

IsaacMarovitz (approved):
Fixed hang when stopping emulation in TOTK for me 👍

noxifoxi:
> Fixed hang when stopping emulation in TOTK for me +1

I merged this PR to the master branch on my test fork, and it still hangs for me

lostromb:
> > Fixed hang when stopping emulation in TOTK for me +1
> 
> I merged this PR to the master branch on my test fork, and it still hangs for me

Again, there are other known issues that could cause deadlock on shutdown, this is just addressing one of them

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -392,26 +473,35 @@ private void FillHipcResponse(IpcMessage response, long result, ReadOnlySpan<byt
 
         protected virtual void Dispose(bool disposing)
         {
-            if (disposing)
+            if (disposing && _selfThread != null)
             {
-                foreach (IpcService service in _sessions.Values)
+                if (_selfThread.HostThread.ManagedThreadId != Environment.CurrentManagedThreadId
+                    && _selfThread.HostThread.Join(ThreadJoinTimeout) == false)
```
gdkchan, line 479: Why is the time out necessary?

jhorv: With the timeout it will log a message instead of possibly hanging on shutdown.

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -97,11 +109,53 @@ public void AddSessionObj(KServerSession serverSession, IpcService obj)
...
+            }
+        }
+
+        private IpcService GetSessionObj(int serverSessionHandle)
+        {
+            bool lockTaken = false;
+            try
+            {
+                lockTaken = _handleLock.TryEnterReadLock(Timeout.Infinite);
+                return _sessions[serverSessionHandle];
```
AcK77, line 133:
```suggestion

                return _sessions[serverSessionHandle];
```

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
gdkchan: It doesn't feel right to do this. If it needs that to exit properly, then there's still something wrong with the code.

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -392,26 +473,35 @@ private void FillHipcResponse(IpcMessage response, long result, ReadOnlySpan<byt
...
         {
-            if (disposing)
+            if (disposing && _selfThread != null)
             {
-                foreach (IpcService service in _sessions.Values)
+                if (_selfThread.HostThread.ManagedThreadId != Environment.CurrentManagedThreadId
+                    && _selfThread.HostThread.Join(ThreadJoinTimeout) == false)
+                {
+                    Logger.Warning?.Print(LogClass.Service, $"The ServerBase thread didn't terminate within {ThreadJoinTimeout:g}, resources will not have Dispose() called to avoid a race condition.");
+                    return;
```
AcK77, line 482:
```suggestion

                    return;
```

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
AcK77, line 479:
```suggestion
                if (_selfThread.HostThread.ManagedThreadId != Environment.CurrentManagedThreadId && _selfThread.HostThread.Join(ThreadJoinTimeout) == false)
```

jhorv: I'm not sure there even is an issue here, but if there is, at least it writes info to the log. The alternative is to hang silently, providing no useful info until someone attaches a debugger.

jhorv: What if after logging, I add a call to `_selfThread.HostThread.Join(Timeout.Infinite)`? I'd adjust the log message to say something like "about to try waiting indefinitely for this thread to finish".

jhorv: Also, I don't believe this thread is known to hang - this PR doesn't fix hangs on stop, it fixes exceptions on stop due to a race condition. I added a timeout here because I think timeouts are better than infinite waits, if only to log information about what happened.

gdkchan: The sentence has a double negative that makes it more confusing now. You said that you "don't believe" that it "is not known to hang", which means  that it is known to hang I guess? But I assume that you mean the opposite.

Anyway my original comment is because such approaches feel `On Error Resume Next` levels of hacky. If the logic is correct and we don't expect it to ever hang, then there's no need for a time out. I guess it is fair putting it as a fail safe, but we usually don't do that. Also having a time out to ignore the hang would make it harder to debug since you can't just attach a debugger to inspect the invalid state anymore. I guess my main question is, why do we need this? Just to know if a hang is caused by `ServerBase` or something else when it happens, without needing to have a debugger attached?

gdkchan (approved):
lgtm, thanks.

PR #4578, opened 2023-03-22
Eliminate redundant multiplications by gl_FragCoord.w on the shader

gdkchan:
This change removes a somewhat redundant pattern on the shader, cause by some code produced by the compiler for perspective correction on fragment shaders. Basically the compiler uses the "Multiply" mode on the IPA instruction for those inputs. The multiplier is `1.0 / gl_FragCoord.w`. To counter-act this, the emulator insert a multiplication by `gl_FragCoord.w` for perspective inputs, and the end result is a `(x * gl_FragCoord.w) * (1.0 / gl_FragCoord.w)`. This is adding an additional step that can recognize an eliminate this pattern.

Sample diff:
Before:
```glsl
    // 0x000010: 0xE003FF87CFF7FF09 Ipa
    // 0x000018: 0x508000000047090D Mufu
    temp_0 = 1.0 / gl_FragCoord.w;
    // 0x000028: 0xE043FF8A00D7FF04 Ipa
    temp_1 = in_attr2.x * gl_FragCoord.w;
    temp_2 = temp_1 * temp_0;
```
After:
```glsl
    // 0x000010: 0xE003FF87CFF7FF09 Ipa
    // 0x000018: 0x508000000047090D Mufu
    // 0x000028: 0xE043FF8A00D7FF04 Ipa
    temp_0 = in_attr2.x;
```
This was actually causing glitches in a few games, like Demon Slayer, likely due to floating point error making the value slightly different after those operations. So this change fixes this game as well.
Before:
image
image
After:
image
image
The only thing I was worried about is the optimisation being applied to cases that are not compiler generated, so I limited it as much as possible. It will only affect multiplication by any fragment shader *inputs* by `gl_FragCoord.w` followed by a multiplication by `1.0 / gl_FragCoord.w`, everything else is untouched.

Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 4892;
+        private const uint CodeGenVersion = 4578;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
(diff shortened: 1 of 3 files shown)


marysaka (approved):

riperiperi (approved):
Looks good. I guess any worries with this would be around translation performance for this kind of  optimization.

- Iterating the block for each potential optimisation seems wasteful, though I guess this can be necessary to take into account previous optimisation results.
- Minimizing the number of operations per check is important. There will be a lot of FP32 multiply in any shader, so this one might end up doing most of the check quite often. You'd probably need to profile to check if that's significant.

I don't think the shader translation speed will be significantly impacted, but it wouldn't hurt to test (driver cache enabled, delete ryu host cache).

Shrimbop:
How do you apply this/build this to your Ryujinx program?

Neristee:
Quick test with a locally built build using this PR. Including commit 1d3b9d1.

BOTW (1.1.781) : 45-48 fps.
BOTW (1.1.781 + PR) : 43-47 fps.

Close to margin of error tbh.

Bjorn29512:
I could not replicate the reported regression (now a deleted comment) on botw at either native or with scaling but ig for sanity sake its better to get samples from more ppl.

image

riperiperi:
People have reported that this doesn't really affect shader compile time, so should be good to go. 👍

PR #4094, opened 2022-12-10
Use source generated json serializers in order to improve code trimming

SukharevAndrey:
This PR introduces source generated json serializers instead of regular ones. This fixes a lot of trimming warnings in the solution by analyzing types used in serialization in compile time.

Additional changes:
* Fix improper `JsonSerializerOptions` usage as they must be used statically to avoid performance issues. 
* Unify deserializing/serializing from/to files.
* Default settings are now small and do not include converters for concrete types.
* Mark all used enums with converter attribute using generic `TypedStringEnumConverter` . Standard `JsonStringEnumConverter` is not trimming friendly so it can't be used globally.
* Write JSONs indented by default.
* Amiibo json models are extracted to a common library.
* Get rid of two NewtonsoftJson usages in updater. Use typed model instead of parsing fields manually.

PR #4582, opened 2023-03-22
Source generated json serializers

SukharevAndrey:
Fixed version of #4094


gdkchan:
I tested it here, and it still fails to load the configuration json from a previous version
> Failed to load config! Loading the default config instead.

TSRBerry:
Could you share your config file? It loads fine for me.

TSRBerry:
Nvm, found them! Now it works correctly.

marysaka (approved):
Seems fine to me, will need another rebase

PR #4584, opened 2023-03-22
ARMeilleure: Check for `XSAVE` cpuid flag for AVX{2,512}

Wunkolo:
Protection for the `xgetbv` instruction for systems that do not support `xcr0` such as nehalem processors.

The `XSAVE` cpuid indicates support for `XSAVE`, `XRESTOR`, `XSETBV`, `XGETBV` while `OSXSAVE` indicates if the operating system itself has `XSAVE` turned on. Both must be checked at the same time.

Addresses issues like #4583


DahniMae:
Fixes Issue #4583

gdkchan (approved):

riperiperi (approved):

PR #4596, opened 2023-03-24
Vulkan: Insert barriers before clears

riperiperi:
Newer NVIDIA GPUs seem to be able to start clearing render targets before the last rasterization task is completed, which can cause it to clear a texture while it is being sampled.

This change adds a barrier from read to write when doing a clear, assuming it has been sampled in the past. It could be possible for this to be needed for sample into draw by some GPU, but it's not right now afaik.

This should fix visual artifacts on newer NVIDIA GPUs and driver combos. Contrary to popular belief, Tetris® Effect: Connected is not affected. Testing welcome, hopefully should fix most cases of this and not cost too much performance.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
+++ b/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
@@ -2,6 +2,7 @@
 using Silk.NET.Vulkan;
 using System;
 using System.Linq;
+using System.Reflection;
 using VkFormat = Silk.NET.Vulkan.Format;
 
 namespace Ryujinx.Graphics.Vulkan
@@ -217,6 +218,24 @@ namespace Ryujinx.Graphics.Vulkan
             _depthStencil?.Storage.SetModification(
                 AccessFlags.DepthStencilAttachmentWriteBit,
                 PipelineStageFlags.ColorAttachmentOutputBit);
+        }
+
+        public void InsertClearBarrier(CommandBufferScoped cbs, int index)
+        {
+            if (_colors != null)
+            {
+                int realIndex = Array.IndexOf(AttachmentIndices, index);
+
+                if (realIndex != -1)
+                {
+                    _colors[realIndex].Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.ColorAttachmentWriteBit, PipelineStageFlags.ColorAttachmentOutputBit);
+                }
+            }
+        }
+
+        public void InsertClearBarrierDS(CommandBufferScoped cbs)
+        {
+            _depthStencil?.Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.ColorAttachmentWriteBit, PipelineStageFlags.ColorAttachmentOutputBit);
         }
     }
 }
--- a/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
+++ b/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
@@ -234,7 +234,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         public void InsertClearBarrierDS(CommandBufferScoped cbs)
         {
-            _depthStencil?.Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.ColorAttachmentWriteBit, PipelineStageFlags.ColorAttachmentOutputBit);
+            _depthStencil?.Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.DepthStencilAttachmentWriteBit, PipelineStageFlags.EarlyFragmentTestsBit);
         }
     }
 }
--- a/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
+++ b/Ryujinx.Graphics.Vulkan/FramebufferParams.cs
@@ -2,7 +2,6 @@
 using Silk.NET.Vulkan;
 using System;
 using System.Linq;
-using System.Reflection;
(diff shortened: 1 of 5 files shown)


OldManKain:
Confirmed to fix multiple games that had rendering artifacts on 30 series Nvidia GPU's with driver versions 522.25 and newer.

### GPU: 3070

### XBDE
Before
ryujinx_capture_2023-03-24_17-07-30
After
ryujinx_capture_2023-03-24_17-28-27
</br>

### XB2
Before
ryujinx_capture_2023-03-24_17-03-31
After
ryujinx_capture_2023-03-24_17-25-01
</br>

### XB3
Before
ryujinx_capture_2023-03-24_17-04-57
After
ryujinx_capture_2023-03-24_17-26-32
</br>

### MK8 Deluxe
Before
ryujinx_capture_2023-03-13_09-49-02
After
ryujinx_capture_2023-03-24_17-29-56

Ryujinx.Graphics.Vulkan/FramebufferParams.cs
```diff
@@ -218,5 +218,23 @@ public void UpdateModifications()
...
+                if (realIndex != -1)
+                {
+                    _colors[realIndex].Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.ColorAttachmentWriteBit, PipelineStageFlags.ColorAttachmentOutputBit);
+                }
+            }
+        }
+
+        public void InsertClearBarrierDS(CommandBufferScoped cbs)
+        {
+            _depthStencil?.Storage?.InsertReadToWriteBarrier(cbs, AccessFlags.ColorAttachmentWriteBit, PipelineStageFlags.ColorAttachmentOutputBit);
```
gdkchan, line 237: Should the flag here be `DepthStencilAttachmentWriteBit`?

Ryujinx.Graphics.Vulkan/TextureStorage.cs
```diff
@@ -419,6 +421,33 @@ public TextureView CreateView(TextureCreateInfo info, int firstLayer, int firstL
             }
         }
 
+        private ImageAspectFlags GetAspectFlags()
```
gdkchan, line 424: There is a extension method on `EnumConversion` called `ConvertAspectFlags` that does the same thing, so I think it can be used instead (and this can be removed. BTW while you are at it, would be nice to fix this spacing issue on said method:
```cs
GAL.Format.S8UintD24Unorm    => ImageAspectFlags.DepthBit | ImageAspectFlags.StencilBit,
```

gdkchan (commented):
I know it is a existing issue, but there are some validation errors that are caused by those barriers. It would be nice to fix them since it causes a lot of noise when the validation logging is enabled. I'm not sure how easy it would be to fix it, and since it's a existing issue I don't thin it *needs* to be fixed as part of this PR, but it would still be nice to make the debugging experience a bit better.

Ryujinx.Graphics.Vulkan/FramebufferParams.cs
riperiperi: Yeah, the stage is meant to be early fragmebt test too I think. I used the right ones when testing then copy pasted the color barrier 🥴

MetrosexualGarbodor:
Also seems to fix the issue on RTX 4000 series GPUs 
Screenshot_20230325-104955_Discord

GamerzHell9137:
Even tho all other games were fixed, Bayonetta Origins is one of the games on which the fix didn't apply, its doing something else than the games above

image

zurce:
Fixed a weird graphical issue with Hyper Light Drifter too


Before :
Screenshot 2023-03-25 112341

With this build:
Screenshot 2023-03-25 114244

zurce:
Cloud in XC3 also seem to be rendering correctly, before they had a trail of artifacts on the corner of the screen, now it's clean 

Before:
Screenshot 2023-03-25 113620

With this build:
Screenshot 2023-03-25 114649

Tested with a 4090 and a i7-13700k

Ryujinx.Graphics.Vulkan/TextureStorage.cs
```diff
@@ -440,31 +442,39 @@ public void SetModification(AccessFlags accessFlags, PipelineStageFlags stage)
...
         }
 
-        public void InsertBarrier(CommandBufferScoped cbs, AccessFlags dstAccessFlags, PipelineStageFlags dstStageFlags)
+        public void InsertReadToWriteBarrier(CommandBufferScoped cbs, AccessFlags dstAccessFlags, PipelineStageFlags dstStageFlags)
         {
-            if (_lastModificationAccess != AccessFlags.NoneKhr)
+            if (_lastReadAccess != AccessFlags.NoneKhr)
             {
-                ImageAspectFlags aspectFlags;
+                ImageAspectFlags aspectFlags = EnumConversion.ConvertAspectFlags(Info.Format);
```
gdkchan, line 449:
```suggestion
                ImageAspectFlags aspectFlags = Info.Format.ConvertAspectFlags();
```
Since it's a extension method, I think you can just do that (same for the other one).
And would be nice to fix that spacing on that method as I commented before:
```cs
GAL.Format.S8UintD24Unorm    => ImageAspectFlags.DepthBit | ImageAspectFlags.StencilBit,
```

marysaka (approved):

gdkchan (approved):
lgtm, thanks. Very nice to see the ampere issue fixed.

PR #4601, opened 2023-03-25
Vulkan: Add barriers for image uses

riperiperi:
Texture+Sampler uses already have these, just adding them to images. This may fix some more issues with newer NVIDIA drivers, or possibly block artifacts on AMD (that aren't subgroup related, so appear on RDNA).

This is experimental to see if it actually affects anything. Based on #4596 .


Exhigh:
Needs a rebase

Issue #4229, opened 2023-01-08
SDL_AUDIODRIVER=dummy crashes with 'OpenStream failed with error: "Audio device already open"'


zhuowei:
### What's the issue you encountered?

Ryujinx crashes when running with SDL2's Dummy audio driver and the game requests more than one audio session.

(Yes, this is a pointless usecase, I know, since if I don't want sound I could just use the dummy backend in Ryujinx instead of SDL2's dummy backend. I think this may also happen with other SDL audio backends with OnlyHasDefaultOutputDevice set, though: https://github.com/libsdl-org/SDL/blob/f070c83a6059c604cbd098680ddaee391b0a7341/src/audio/SDL_audio.c#L1374 I think the Jack backend is the only major non-dummy one that users may use?)

### How can the issue be reproduced?

- Launch Ryujinx with SDL_AUDIODRIVER=dummy:
  `SDL_AUDIODRIVER=dummy ./Ryujinx.app/Contents/MacOS/Ryujinx`
- Load a game that opens more than one audio session. (I tested with Pokémon Brilliant Diamond; I'm not sure if other games does the same)
- Ryujinx crashes with:

```
00:01:39.508 |I| HLE.OsThread.7 AudioRenderer AcquireSessionId: Registered new output (1)
00:01:39.543 |E| HLE.OsThread.7 Application : Unhandled exception caught: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: OpenStream failed with error: "Audio device already open"
   at Ryujinx.Audio.Backends.SDL2.SDL2HardwareDeviceSession.EnsureAudioStreamSetup(AudioBuffer buffer) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Audio.Backends.SDL2/SDL2HardwareDeviceSession.cs:line 44
   at Ryujinx.Audio.Common.AudioDeviceSession.FlushToHardware() in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Audio/Common/AudioDeviceSession.cs:line 154
   at Ryujinx.Audio.Output.AudioOutputSystem.AppendBuffer(UInt64 bufferTag, AudioUserBuffer& userBuffer) in /runner/_work/Ryujinx_Eve/Ryujinx_Eve/Ryujinx.Audio/Output/AudioOutputSystem.cs:line 202
[...]

PR #4605, opened 2023-03-26
audout: Fix a possible crash with SDL2 when the SDL2 audio backend is dummy

marysaka:
This PR makes audio device error not fatal.

In case of error, the SDL2 audio backend will behave like the dummy backend.

Close https://github.com/Ryujinx/Ryujinx/issues/4229.


gdkchan (approved):
lgtm, but it's worth noting that some games won't work with that approach, or will run very slowly because the audio thread will be over working since it will always need more audio data all the time.

marysaka:
> lgtm, but it's worth noting that some games won't work with that approach, or will run very slowly because the audio thread will be over working since it will always need more audio data all the time.

Yeah I know... I really need to revisit the overall dummy audio approach so it does notify updates at the correct time.

Issue #4547, opened 2023-03-14
Ryujinx ava/regular process does not shut down upon File>Exit in SteamOS(Arch)


m3talslug:
### Description of the issue

Running the latest Stable build of SteamOS(Arch Linux).  Installed RyuJinx and am able to run games etc.  When I close the app, the process Ryujinx remains in my task manager and has to be killed.  This happens on all the builds of Ryujinx at the time of writing this.  

### Reproduction steps

1. Run Ryujinx from either the executable or the .sh
2. Open the task maanger so you can see the process Ryujinx
3. In Ryujinx click File > Exit
4. The Ryujinx window should close but the process Ryujinx still remains in the task manager

### Log file

[Ryujinx_1.1.662_2023-03-14_13-11-21.log](https://github.com/Ryujinx/Ryujinx/files/10971834/Ryujinx_1.1.662_2023-03-14_13-11-21.log)

### OS

SteamOS

### Ryujinx version

1.1.662

### CPU

Zen 2 4c/8t

### GPU

8 RDNA 2 CUs

### RAM

16

hrzhu:
I can confirm this. If I open and close Ryujinx multiple times. Multiple instances of Ryujinx process remain running in the background. And all memory consumed by Ryujinx are not released. My current workaround is set Terminal=True in Ryujinx.desktop and manually close the terminal every time. This could be tedious for a usual user.

dylanwulf:
I can confirm this is also happening on Fedora (KDE Plasma/Wayland). Both the flathub version and the version downloaded from the Ryujinx website do not shut down the process when you close the window.

PR #4617, opened 2023-03-29
Fix Linux hang on shutdown

jhorv:
This PR addresses issue #4547 (at least, it works on my Arch install) by adjusting the redirection of stderr to the Ryujinx log. Before this PR, process shutdown would wait indefinitely inside a `read` call on the stderr file descriptor. This PR avoids  the hang by doing the read asynchronously, and with a cancellation token.

This PR also removes the `UnixStream` class, as the built-in `FileStream` can already work with Unix file descriptors.


TSRBerry (approved):
Thank you very much!

marysaka (approved):

PR #1775, opened 2020-12-03
Fix Vnmls_S fast path (F64: losing input d value). Fix Vnmla_S & Vnmls_S slow paths (using fused inst.s). Fix Vfma_V slow path not using StandardFPSCRValue().

LDj3SNuD:
Add Vfms_V inst. with Fma fast path and slow path.

Add Vfma_S & Vfms_S Fma fast paths.
Add Vfnma_S & Vfnms_S Sse fast paths.

Add Tests for all affected inst.s.

Nits.

PR #4618, opened 2023-03-30
ARMeilleure: Respect FZ/RM flags for all floating point operations

riperiperi:
This is a change in strategy for emulating the FZ FPCR flag (and others). Before, it was set before instructions that "needed it" and reset after. However, this missed a few hot instructions like the multiplication instructions, and the entirety of A32.

The new strategy is to set the FZ flag only in the following circumstances:

- Set to match FPCR before translated functions/loop are executed.
- Reset when calling SoftFloat methods, set when returning.
- Reset when exiting execution.

This allows us to remove the code around the existing FZ aware instructions, and get the accuracy benefits on all floating point instructions executed while in translated code. This also allows us to set the processor's default round mode to match the mode set in FPCR (RM), which should improve accuracy for normal numbers when results are imprecise.

I've also modified instructions to use FMA when their ARM reference implementations use fused operations. This should improve performance when games use these instructions (less instructions, less register pressure), while also making the operations more accurate with FZ active (the intermediate result is not flushed to zero or treated in any way). This includes the A64 variants of rsqrt and reciprocal step - the A32 variants explicitly perform the operations separately for some reason and can flush to zero in the middle, so were already fine.

Single step executions now need to be called with a context wrapper - right now it just contains the FZ/RM flag initializations, and won't actually do anything on ARM.

This fixes a bug in Breath of the Wild where some physics interactions could randomly crash the game due to subnormal values not flushing to zero. https://github.com/Ryujinx/Ryujinx-Games-List/issues/86#issuecomment-1488968394 Thanks to @OpaqueReptile and @MelonSpeedruns for pointing this out and initial testing. This was also the "Lynel crash", as being around a lynel could trigger a crash for a similar reason.

I recommend testing a bunch of games, especially games affected when this was originally added, such as #1611.

### Notes & Future
The x86 handling only resets/sets the control flags when doing SoftFloat operations, which explicitly need to handle subnormal results without flushing to zero. This is because I've verified that managed methods should not be able to change the current thread's Ftz mode. I've added a test that verifies this, which will alert us if it changes in future.

Status flags on the ARM backend are something to look at in future, when I have more time or someone else checks it out. They might be a bit of a different story as we will want to store them before _any_ managed method, and we may want to also use the host flags when possible rather than flushing back to the FpFlags state mid execution.


gdkchan:
I think we should have the same treatment for Arm too (there's already some code to save and restore the Qc flag, if all that could be handled at the same place, it would be better). There's also the rounding mode that right now doesn't match the guest rounding mode.

I think the potential issue there is floating point operations that are emulated with multiple operations. Flushing to zero after some intermediate operation could lead to some errors.

riperiperi:
I had a look at all instructions that do multiple floating point operations and found the following instructions. None of the negates will flush to zero as I think they are all bitwise ops.
```
Fabd: sub then abs (bitwise, unaffected by ftz on previous result?)
Fmadd: mul then add (d = n*m + a)
Fmla: mul then add (d = n*m + d)
Fmsub: mul then subtract (d = a - n*m)
Fmls: mul then subtract (d = d - n*m)

Fnmadd: neg (bitwise) then mul then sub (d = -a - n*m)
Fnmsub: neg (bitwise) then mul then add (d = -a + n*m)

Frecps: Reciprocal step, 2 - n*m. Not sure if ftz will have an effect on result. Could probably use fused mla.
Frsqrts: Sqrt estimation step, (3 - n*m) * 0.5. Not sure if ftz will have an effect on result. Cannot use mla for the whole thing, but last multiplier always goes down a power of 2 so might be fine with ftz.

Fnmul: mul then negate. (bitwise, unaffected by ftz on previous result?)
```

In all cases, these instructions can currently produce final results with subnormals on master. So it's trading not ever flushing subnormal results, to potentially having slightly _different_ results on these instructions (but never subnormal). I think the ones that are just one floating point op and sign flipping are going to be fine.

In every remaining case it could be avoided by using FMA3 instructions. Mla, Mls etc are self explanatory. Frecps is (2 - n*m). For Frsqrts, only the (3 - n*m) portion can become an VFNMADD, but the following multiplication by 0.5 means a subnormal flushed to zero should make no difference, as it would only get smaller (which would also flush to zero).

FMA instructions are used... On A32 only, for some reason, so I guess these instructions are fine there. It was added back in #1775.

I also noticed that denormal-as-zero (`DAZ`) is enabled on the flag combination we were using for instructions supported for `FZ`, and all instructions now... While this might have been helpful without other instuctions flushing to zero, I think it's inaccurate as input denormals don't appear to be treated in our slow paths. The flag `FIZ` is used in ARM to control flushing input to zero, which should be the same as `DAZ`, but it's only available if you have `FEAT_AFP` (alternate floating point handling), which is an ARMv8.7 feature, so the switch can't use it.

riperiperi:
I'm leaving this PR as x86 only, improving the FPCR/FPSR flags for A64 will probably be a big task on its own and will require a lot of testing.

I've done the following since opening the PR:
- FMA instructions use host FMA instructions when possible to avoid intermediate round/ftz.
- Round mode is now respected.
- Mxscr instructions are now just load/store rather than mask. This allows multiple bits to be cleared/set at the same time, and mirrors how it would be done on the ARM backend.
- Daz flag is not set.
- Added test to ensure flush to zero flag is not reset in managed methods, either by pinvoke, floating point math or by exception handling. This allows the FP control flags to be set/reset less often.

I recommend retesting this PR with LM3 and BOTW, as well as sanity checking a few other games.

gdkchan:
> I'm leaving this PR as x86 only, improving the FPCR/FPSR flags for A64 will probably be a big task on its own and will require a lot of testing.

What's special about Arm? Doesn't it just need to do the same thing as x86?

riperiperi:
The problem is changing how the status flag Qc works. The control flags would be largely similar.

gdkchan:
Can't you implement the FPCR change and leave the Qc flag alone then? At least it would fix the main issue this is trying to address on Arm as well.

gdkchan:
BTW, setting rounding mode on Arm would allow us to use instructions that depends on the current rounding mode for rounding or float <-> int conversion directly, like FRINTX for example:
```cs
        public static void Frintx_S(ArmEmitterContext context)
        {
            // TODO Arm64: Fast path. Should we set host FPCR?
            EmitScalarUnaryOpF(context, (op1) =>
            {
                return EmitRoundByRMode(context, op1);
            });
        }
```
It would allow those `TODO`s to be removed. Of course, it doesn't need to be done as part of this PR, but implementing the rounding mode for Arm would enable this change to be made.

gdkchan:
> Added test to ensure flush to zero flag is not reset in managed methods, either by pinvoke, floating point math or by exception handling. This allows the FP control flags to be set/reset less often.

I would still be worried about managed methods doing floating point operations that are affected by the FTZ flag. It could introduce subtle bugs depending on the guest changing those flags. If possible I think it would be better to reset/restore those flags for all managed method calls, but if there's a significant cost in doing that I think the risk can be justified.

riperiperi:
I've added just the A64 FPCR flags, and now FRINTX calls the host method as a result. I haven't tested this yet.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
```diff
@@ -1146,6 +1135,15 @@ public static void Fminv_V(ArmEmitterContext context)
             {
                 InstEmitSimdHelperArm64.EmitScalarTernaryOpFRdByElem(context, Intrinsic.Arm64FmlaSe);
             }
+            else if (Optimizations.UseFma)
+            {
+                EmitScalarTernaryOpByElemF(context, (op1, op2, op3) =>
+                {
+                    OpCodeSimdRegElemF op = (OpCodeSimdRegElemF)context.CurrOp;
+
+                    return context.AddIntrinsic((op.Size & 1) != 0 ? Intrinsic.X86Vfmadd231sd : Intrinsic.X86Vfmadd231ss, op1, op2, op3);
```
gdkchan, line 1144: This seems somewhat invalid since the operands are of `FP32` or `FP64` type. Those intrinsic operations usually expect `V128`. Also it is not as efficient as it could be since it extracts all elements (even those at index 0). I think the best thing to do here is adding a instruction to move the element of the third operand to the correct position, then doing the operation. I think that can be done with a shuffle instruction, or for the double case, just moving from high to low. I'm sure there's prior art for that somewhere in other "by element" instruction, maybe it can be used as reference.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
```diff
@@ -1261,6 +1287,15 @@ public static void Fminv_V(ArmEmitterContext context)
             {
                 InstEmitSimdHelperArm64.EmitScalarTernaryOpFRdByElem(context, Intrinsic.Arm64FmlsSe);
             }
+            else if (Optimizations.UseFma)
+            {
+                EmitScalarTernaryOpByElemF(context, (op1, op2, op3) =>
+                {
+                    OpCodeSimdRegElemF op = (OpCodeSimdRegElemF)context.CurrOp;
+
+                    return context.AddIntrinsic((op.Size & 1) != 0 ? Intrinsic.X86Vfnmadd231sd : Intrinsic.X86Vfnmadd231ss, op1, op2, op3);
```
gdkchan, line 1296: Same comment as above.

riperiperi:
Appears to work properly on ARM, including the fp flags test, minus BOTW crashing after a while for GPU reasons.

ARMeilleure/Instructions/InstEmitSimdArithmetic.cs
```diff
@@ -2113,20 +2296,32 @@ public static void Frintp_V(ArmEmitterContext context)
 
         public static void Frintx_S(ArmEmitterContext context)
         {
-            // TODO Arm64: Fast path. Should we set host FPCR?
-            EmitScalarUnaryOpF(context, (op1) =>
+            if (Optimizations.UseAdvSimd)
+            {
+                InstEmitSimdHelperArm64.EmitVectorUnaryOpF(context, Intrinsic.Arm64FrintxS);
```
gdkchan, line 2301: Should probably be `EmitScalarUnaryOpF`? Since that's a scalar operation.

gdkchan (approved):
lgtm minus comment.

It's nice to see this issue finally fixed. I didn't expect it to be caused by floating point operation inaccuracy since those usually causes weird glitches but not crashes. Nice catch.

I did test a few games here, but I didn't really test it extensively. I think one game worth testing is the LEGO Harry Potter Collection where there is a documented logic bug. Unfortunately I don't have that game to test though.

I still think it would be better to to restore/set the FP flags before/after every managed call as I said before, but if there's significant cost in doing that then I guess it's fine to leave as is.

ARMeilleure/Instructions/InstEmitSimdCvt32.cs
```diff
@@ -372,9 +374,12 @@ public static void Vcvt_TB(ArmEmitterContext context)
                         ? typeof(SoftFloat16_64).GetMethod(nameof(SoftFloat16_64.FPConvert))
                         : typeof(SoftFloat16_32).GetMethod(nameof(SoftFloat16_32.FPConvert));
 
+
```
gdkchan, line 377: Double empty line.

Bjorn29512:
Tested about a dozen + titles and didn't notice any immediate regressions either performance or visual related, retested BOTW and its fine as well and lynel crashes fixed, dont have LM 3 so cant speak for that title.

marysaka (approved):

PR #4624, opened 2023-03-31
Reducing Memory Allocations 202303

jhorv:
Another PR focused on reducing memory allocations by:

- replacing new allocations with use of `ArrayPool` and `MemoryPool` when ownership is clear
- reducing LINQ use in hot paths
- using dynamically generated delegates instead of reflection to call `IpcService` methods

In addition, this PR:

- tweaks `KTimeManager` so that it doesn't `Thread.Yield()` more than what appears to be intended
- goes after large object heap (LOH) allocations. Much more work to be done here.

Here is a summary of the difference these changes make to an approximately 24 second run, wherein Ryujinx.Ava is started with launch options to immediately run Metroid Prime Remastered.

|                              | master-5131b71 | This PR | Delta       |
|------------------------------|----------------|---------------|-------------|
| Realtime Run Time (seconds)  | 24             | 24            |             |
| Profile Run Time (seconds)   | 72             | 66            |             |
| SOH Allocations              | 14,646,765     | 10,814,886    | -3,831,879   |
| SOH Allocated Bytes          | 1,250,253,413  | 1,035,782,495 | -214,470,918 |
| LOH Allocations              | 6,351          | 6,192         | -159         |
| LOH Allocated Bytes          | 3,043,422,313  | 2,966,044,293 | -77,378,020  |
| Total GC                     | 89             | 76            | -13          |
| Total GC Pause Duration (ms) | 10,295         | 8,647         | -1,648       |
| % GC Pause to Run Time       | 14.29902778    | 13.10154545   |  -1.2       |


Ryujinx.Audio/Renderer/Server/Voice/VoiceState.cs
```diff
@@ -378,7 +378,7 @@ public void UpdatePlayState(PlayState userPlayState)
         /// <param name="outStatus">The given user output.</param>
         /// <param name="parameter">The user parameter.</param>
         /// <param name="voiceUpdateStates">The voice states associated to the <see cref="VoiceState"/>.</param>
-        public void WriteOutStatus(ref VoiceOutStatus outStatus, ref VoiceInParameter parameter, Memory<VoiceUpdateState>[] voiceUpdateStates)
+        public void WriteOutStatus(ref VoiceOutStatus outStatus, ref VoiceInParameter parameter, Span<Memory<VoiceUpdateState>> voiceUpdateStates)
```
gdkchan, line 381: Can't those be `ReadOnlySpan`s? Presumably it doesn't need to change the elements.

Ryujinx.HLE/HOS/Ipc/IpcMessage.cs
```diff
@@ -27,101 +27,105 @@ class IpcMessage
...
-                RecvListBuff.Add(new IpcRecvListBuffDesc(reader.ReadUInt64()));
+                for (int index = 0; index < recvListCount; index++)
+                {
+                    RecvListBuff.Add(new IpcRecvListBuffDesc(reader.ReadUInt64()));
+                }
+
+                ObjectIds = new List<int>(0);
             }
         }
-
```
gdkchan, line 124: Why was the separation removed here?

Ryujinx.HLE/HOS/Services/IpcService.cs
```diff
@@ -278,5 +281,22 @@ public virtual void DestroyAtExit()
 
             _domainObjects.Clear();
         }
+
+        private static IpcServiceMethodDelegate CreateIpcServiceMethodDelegate(MethodInfo methodInfo)
+        {
+            var dynamicMethod = new DynamicMethod(methodInfo.Name, methodInfo.ReturnType, new[] { typeof(IpcService), typeof(ServiceCtx) });
```
gdkchan, line 287: This will certainly break AOT. The current use of reflection already gives it some trouble.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -71,7 +71,13 @@ public void UnscheduleFutureInvocation(IKFutureSchedulerObject schedulerObj)
         {
             lock (_context.CriticalSection.Lock)
             {
-                _waitingObjects.RemoveAll(x => x.Object == schedulerObj);
+                for (int index = _waitingObjects.Count - 1; index >= 0; index--)
```
gdkchan, line 74: The ideal data structure for this would probably be a binary tree, since it needs to be sorted by time and allow fast insertion and removal.

Ryujinx.Audio/Renderer/Server/Voice/VoiceState.cs
jhorv: Fixed in commit 07c5214

Ryujinx.HLE/HOS/Ipc/IpcMessage.cs
jhorv: Accident, fixed in commit 438daa7

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
jhorv: The intent here is a drop-in replacement for `RemoveAll`. This specific call shows up unusually high in the memory profiler, so replacing it with an equivalent `for` loop seemed the best low-risk solution. I agree there's certainly a better data structure for `_waitingObjects`, but a) I don't see the `List<>` as an outlier when profiling CPU time and b) I didn't want to add more risk to this PR.

Ryujinx.HLE/HOS/Services/IpcService.cs
jhorv: Yes, but will it prevent AOT more than reflection already does? Until a more AOT-friendly approach is implemented, this change is low-risk, performs better, and it eliminates the array allocations from passing arguments to `MethodInfo.Invoke()`.

jhorv: As discussed in Discord, I reverted this commit.

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -105,15 +111,8 @@ private void WaitAndCheckScheduledObjects()
                             }
                             else
                             {
-                                while (Interlocked.Read(ref _enforceWakeupFromSpinWait) != 1 && PerformanceCounter.ElapsedTicks <= next.TimePoint)
+                                while (Interlocked.Read(ref _enforceWakeupFromSpinWait) != 1 && PerformanceCounter.ElapsedTicks < next.TimePoint)
                                 {
-                                    if (spinWait.NextSpinWillYield)
```
riperiperi, line 110: Why was this removed? The goal here was to avoid the case that does Sleep(1) and always yield.

jhorv: Yeah, on further reflection - this is kernel timing code, it makes sense to not let `SpinLock` go off and do its own thing. Especially after re-reading the comments in SpinLock.cs and seeing why it considers all these possibilities of yielding and sleeping - it's because `SpinLock` is a generic primitive that expects we might be spinning because we're waiting for a locked resource to be freed. But `KTimeManager` isn't doing that - it's just waiting for a very small amount of time to pass.

I think the right thing is to put the `spinWait.SpinOnce()` in an `else {}`. I'll commit that shortly, and hopefully you agree.

riperiperi: This code only runs on the path where the ms wait is less than 1ms. Sleeping for 1ms is the exact opposite thing we want to do... If we were gonna do that, we wouldn't have this path in the first place. Removing this sub-ms timing is not a good idea, it will break timing on very short sleeps which can mess with a bunch of games, including BOTW.

It's more important to look at the reason this case is constantly happening. That reason is the current implementation of ServerBase constantly polls every 1ms for new sessions created by SM. It waits for 1ms on the guest scheduler, which has _very high precision_, so even though we don't need precision for polling the sessions, it causes the TimeManager to spin waiting for the timings anyways (because this is needed for other types of sleep). Because the time waited is since the call to ReplyAndReceive, the resume timepoints for each server quickly drift out of sync and soon enough there are timepoints every .1ms that need to be spun on.

It doesn't need to do this, the service can instead be registered on SM, but only the newer implementations of services do that. (i think there are only like 2 right now?) I don't really support breaking the kernel to work around this, rather than fixing the root cause of the issue.

riperiperi: It's worth noting that the CPU use on the TimeManager isn't "real", because it is constantly yielding, it will immediately yield to a thread that that needs to run if all other cores are busy. It does mean that clocks may be pushed higher if the OS isn't aware that the spinning is just waiting.

jhorv: @riperiperi What are your thoughts on how it is now - with the spinWait in an `else {}` block, so that `Thread.Yield()` and `spinWait.SpinOnce()` aren't called consecutively?

riperiperi: I think that makes sense.

marysaka (approved):

Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
riperiperi: The reason for this showing up so high in the profile is probably `schedulerObj` being captured by this anonymous function.

Ryujinx.Common/Memory/MemoryBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace Ryujinx.Common.Memory
+{
+    /// <summary>
+    /// This class simplifies use of <c>System.IO.MemoryPool<byte></c>. That memory pool often returns more
+    /// memory than requested, which poses a challenge to code that uses the Length of a given <c>Memory</c>
+    /// or <c>Span</c> to know the size of the subject. This class helps by a) simplifying the API for Rent(),
+    /// b) implementing the disposable <c>IMemoryOwner</c> interface, and c) tracking the originally requested
+    /// length, and only exposing a slice of that length.
+    /// </summary>
+    public sealed class MemoryBuffer : IMemoryOwner<byte>, IDisposable
```
riperiperi, line 14: Using a class like this will cause a heap allocation. Is there no way to do this with a struct?

jhorv: It could be a struct, but I think a class make more sense in this case:

- The type's primary functionality is exposed through interfaces. If it is a struct and its functionality is used through an interface, the struct will get boxed, i.e. a memory allocation. If that happens even once, it negates the potential savings of it being a struct. In general, its best to avoid interfaces on structs where performance could matter.
- The type's purpose is to represent ownership of something (the memory allocation), which IMO is better represented by a reference type. Passing around a reference to a single class instance, vs a struct getting copied everywhere it goes. If we were to, for example, add a `bool _isDisposed` field, that field would be of limited value in a `struct MemoryBuffer` where everyone is looking at their own copy of the data structure, vs. the one instance of a reference type.

IMO it's reasonable to treat the relatively small class allocation as tolerable overhead of the (probably much larger) allocation it helps to manage.

Ryujinx.Common/Memory/MemoryBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+                if (clear)
+                {
+                    memory.Span.Clear();
+                }
+
+                return new MemoryBuffer(memoryOwner, memory, length);
+            }
+        }
+
+        public Memory<byte> Memory { get; }
```
riperiperi, line 86: Why are these at the bottom of the class?

Ryujinx.Common/Memory/MemoryBuffer.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            return Rent(checked((int)length), clear);
+        }
+
+        /// <summary>
+        /// Rents memory from the shared memory pool and returns it in new <c>MemoryBuffer</c> instance.
+        /// </summary>
+        /// <param name="length">the buffer's required length in bytes</param>
+        /// <param name="clear"><c>true</c> to clear the memory after it is rented from the pool. Only the desired length is cleared.</param>
+        /// <returns>a new MemoryBuffer instance containing memory rented from <c>MemoryPool<byte>.Shared</c></returns>
```
riperiperi, line 56: Other docs typically have the first word of params/returns capitalized, though there is no full stop.

Ryujinx.Common/Memory/MemoryBuffer.cs
riperiperi: I'm still not sure about this. This class internally uses `MemoryPool<byte>.Shared.Rent`, which _also_ creates an object that implements IMemoryOwner. It seems weird to have another allocation for what is essentially just a wrapper around ArrayMemoryPoolBuffer.

What future use would there be for this class as `IMemoryOwner<T>`, instead of just a disposable wrapper?

jhorv: Just to be sure we're on the same page I'm gonna quick describe what problem `MemoryBuffer` solves: all these *Pools return blocks in sizes that are a power of 2. Ask to `Rent(400)` bytes and it'll return a `IMemoryOwner` wrapping 512 bytes, with no information that you only wanted 400 to begin with. This makes the `IMemoryOwner` returned by a `Rent()` method hard to use directly in a lot of scenarios - so much code looks at the `Length` to know how much to process. So, `MemoryBuffer` is a `IMemoryOwner` that only exposes a slice of the requested size.

Future uses are anywhere we want to:

1. use pooled memory without having to separately manage the "amount requested vs. amount received" problem
2. pass pooled memory to a recipient where the recipient is responsible for freeing it, again without having to pass separately the "usable length".

I have an unfinished PR targeting LOH allocations and a quick glance in VS shows 30 locations using `MemoryBuffer`. There are many more opportunities than those, I just slowed up on that work because it depends on `MemoryBuffer` from this PR.

Would you prefer that instead of `MemoryBuffer` wrapping use of `MemoryPool`, we make our own implementation like `ArrayMemoryPool<T>` that uses `ArrayPool<T>.Shared.Rent()` directly, while still providing the `MemoryBuffer` benefit of only exposing a `Memory<byte>` sliced to the requested size? I kinda like that idea. Something like.... `ByteMemoryPool`?

Ryujinx.Common/Memory/MemoryBuffer.cs
jhorv: I tend to order members like:

- fields
- constructors
- constructor-like static methods
- properties
- methods

The only method is non-static, non-constructor-like method is `Dispose()`, so these properties are very near the bottom of the class.

Ryujinx.Common/Memory/MemoryBuffer.cs
jhorv: xmldocs in the latest commit follow this pattern.

Ryujinx.Common/Memory/MemoryBuffer.cs
jhorv: > Would you prefer that instead of `MemoryBuffer` wrapping use of `MemoryPool`, we make our own implementation like `ArrayMemoryPool<T>` that uses `ArrayPool<T>.Shared.Rent()` directly, while still providing the `MemoryBuffer` benefit of only exposing a `Memory<byte>` sliced to the requested size? I kinda like that idea. Something like.... `ByteMemoryPool`?

I hope you like this idea :) I just pushed 2 commits that remove `MemoryBuffer` and replace it with the new `ByteMemoryPool` having the following characteristics:

- it has the same interface as `System.Buffers.MemoryPool<T>`. Maybe unnecessary, but consistency is nice.
- it rents memory from `ArrayPool<T>.Shared` directly
- it returns its own `IMemoryOwner<T>` implementation (`ByteMemoryPoolBuffer`) with the same benefits as `MemoryBuffer`. Meaning, it's an `IMemoryOwner` that exposes a `Memory<byte>` with a `Length` matching the length requested, not the length actually rented.

And if you don't like this idea, that's okay, let's keep talking. :)

riperiperi (approved):
Good, I think this makes a lot more sense. I think the official memory pool probably should have worked like this in the first place.

There's probably a bit of debate whether `output` and `performanceOutput` need to be cleared, but I'm not an expert on audren. I think they probably don't.

riperiperi:
You need to rebase this for it to be merged, btw.

TSRBerry:
Is this supposed to have 77 commits instead of 39 as shown on your fork?

jhorv:
> You need to rebase this for it to be merged, btw.

I rebased, but is this the right result? Now it looks like this PR includes all the changes merged since when this PR was initially created.

TSRBerry:
Did you force push your changes? If not you might have combined your old state with the new one.

riperiperi:
Something has definitely went wrong. You could try using `git rebase --onto` to place your commits on top of origin master, but I'm not sure how you got in this state in the first place.

jhorv:
> 



> Did you force push your changes? If not you might have combined your old state with the new one.

No, didn't force push.

Issue #4549, opened 2023-03-15
[Bug] Ryujinx autoupdate deleting desktop.ini


munt-kaicho:
### Description of the issue

So as the title of the report says, whenever I use the autoupdate feature the desktop.ini file for the Ryujinx installation folder gets deleted along with the old files.
I tried setting it as a system file but still the same result.

Would be much appreciated if it could be excluded from the update process because I use an InfoTip and custom icon for the folder and I have to redo those after every update.

### Reproduction steps

Every time I autoupdate Ryujinx is deletes the desktop.ini file in the installation folder.

### Log file

No logs to submit

### OS

Windows 11

### Ryujinx version

1.1.665

### List of applied mods

No mods

PR #4626, opened 2023-04-01
Ensure the updater doesn't delete hidden or system files

SpadVN:
Ensure the updater doesn't delete hidden or system files when applying an update.

Closes: #4549


gdkchan:
A better solution would be just not deleting what is not needed during the update. I don't know why the updater currently deletes it instead of just replacing the files with the ones on the update folder.

AcK77:
It’s probably better to replace: 
https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/Modules/Updater/Updater.cs#L737
with: 
`return files.Where(f => !new FileInfo(f).Attributes.HasFlag(FileAttributes.Hidden | FileAttributes.System));`
Since desktop.ini is a hidden file and deleting a system file is wrong too. The change have to be done in GTK UI (Ryujinx project) too.

SpadVN:
> It’s probably better to replace: https://github.com/Ryujinx/Ryujinx/blob/master/Ryujinx.Ava/Modules/Updater/Updater.cs#L737 with: `return files.Where(f => !new FileInfo(f).Attributes.HasFlag(FileAttributes.Hidden | FileAttributes.System));` Since desktop.ini is a hidden file and deleting a system file is wrong too. The change have to be done in GTK UI (Ryujinx project) too.

For sure a better solution, I implemented it, thank you!

gdkchan:
> A better solution would be just not deleting what is not needed during the update. I don't know why the updater currently deletes it instead of just replacing the files with the ones on the update folder.

Since I was requested to review, my comment still stands. This is addressing one very specific issue, but I think it should address the more broad issue that the emulator should not be deleting anything that the user or the system has placed on the emulator folder.

SpadVN:
> > A better solution would be just not deleting what is not needed during the update. I don't know why the updater currently deletes it instead of just replacing the files with the ones on the update folder.
> 
> Since I was requested to review, my comment still stands. This is addressing one very specific issue, but I think it should address the more broad issue that the emulator should not be deleting anything that the user or the system has placed on the emulator folder.

I will look into this.

AcK77 (approved):

marysaka:
> > A better solution would be just not deleting what is not needed during the update. I don't know why the updater currently deletes it instead of just replacing the files with the ones on the update folder.
> 
> Since I was requested to review, my comment still stands. This is addressing one very specific issue, but I think it should address the more broad issue that the emulator should not be deleting anything that the user or the system has placed on the emulator folder.

I think that for the time being this is a satisfying solution.

Not deleting user data mean that we have no way to know what should be delete.
If we follow a logic that we only replace files that are in the update archive, that will possibly bite us if a dynamic library we wanted to remove is still present and cause issues for example.

marysaka (approved):

gdkchan:
> > > A better solution would be just not deleting what is not needed during the update. I don't know why the updater currently deletes it instead of just replacing the files with the ones on the update folder.
> > 
> > 
> > Since I was requested to review, my comment still stands. This is addressing one very specific issue, but I think it should address the more broad issue that the emulator should not be deleting anything that the user or the system has placed on the emulator folder.
> 
> I think that for the time being this is a satisfying solution.
> 
> Not deleting user data mean that we have no way to know what should be delete. If we follow a logic that we only replace files that are in the update archive, that will possibly bite us if a dynamic library we wanted to remove is still present and cause issues for example.

It's not really because it's not a solution. It just solves a very specific case and doesn't solve all of the other cases that are actually much worse than a missing `desktop.ini` file (like the person that for some reason extracted the emulator  in the root of their hard drive, and then the update process deleted all the hard drive contents).

The issue of potentially not deleting a library on update is pretty much harmless in comparison. If a new version doesn't need a native library anymore and we keep the library around, the only harm done is some wasted disk space, unless for some reason we stop shipping a library and decide to depend on a system one that seems extremely unlikely. Even if that was a big deal, we could just have the updater clear this specific library or all of them, it's still a better solution than this (IMO).

gdkchan:
I see now that this was changed to not delete hidden or system files rather than not deleting `desktop.ini`, which is a better solution IMO. While my opinion that it would be better to fix the broader issue still stands, I think it is fine as is (not that it matters too much now that it has been merged, anyway).

PR #4632, opened 2023-04-02
vulkan: Cleanup PhysicalDevice and Instance querying

marysaka:
That clean up a bit of duplicate logic and avoid querying for stuffs everywhere.

Also move to use an HashSet to store extensions string and use Span instead of unsafe pointers when possible.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
@@ -179,7 +179,6 @@ namespace Ryujinx.Graphics.Vulkan
                 EnabledLayerCount = 0
             };
 
-
             Result result = VulkanInstance.Create(api, ref instanceCreateInfo, out var rawInstance);
 
             Marshal.FreeHGlobal(appName);
--- a/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
+++ b/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
@@ -47,35 +47,23 @@ namespace Ryujinx.Graphics.Vulkan
             KhrSwapchain.ExtensionName
         };
 
-        internal static Instance CreateInstance(Vk api, GraphicsDebugLevel logLevel, string[] requiredExtensions)
+        internal static VulkanInstance CreateInstance(Vk api, GraphicsDebugLevel logLevel, string[] requiredExtensions)
         {
             var enabledLayers = new List<string>();
 
+            var instanceExtensions = VulkanInstance.GetInstanceExtensions(api);
+            var instanceLayers = VulkanInstance.GetInstanceLayers(api);
+
             void AddAvailableLayer(string layerName)
             {
-                uint layerPropertiesCount;
-
-                api.EnumerateInstanceLayerProperties(&layerPropertiesCount, null).ThrowOnError();
-
-                LayerProperties[] layerProperties = new LayerProperties[layerPropertiesCount];
-
-                fixed (LayerProperties* pLayerProperties = layerProperties)
+                if (instanceLayers.Contains(layerName))
                 {
-                    api.EnumerateInstanceLayerProperties(&layerPropertiesCount, layerProperties).ThrowOnError();
-
-                    for (int i = 0; i < layerPropertiesCount; i++)
-                    {
-                        string currentLayerName = Marshal.PtrToStringAnsi((IntPtr)pLayerProperties[i].LayerName);
-
-                        if (currentLayerName == layerName)
-                        {
-                            enabledLayers.Add(layerName);
-                            return;
-                        }
-                    }
+                    enabledLayers.Add(layerName);
                 }
-
-                Logger.Warning?.Print(LogClass.Gpu, $"Missing layer {layerName}");
+                else
+                {
+                    Logger.Warning?.Print(LogClass.Gpu, $"Missing layer {layerName}");
+                }
             }
 
             if (logLevel != GraphicsDebugLevel.None)
@@ -85,7 +73,7 @@ namespace Ryujinx.Graphics.Vulkan
 
             var enabledExtensions = requiredExtensions;
 
-            if (api.IsInstanceExtensionPresent("VK_EXT_debug_utils"))
+            if (instanceExtensions.Contains("VK_EXT_debug_utils"))
             {
                 enabledExtensions = enabledExtensions.Append(ExtDebugUtils.ExtensionName).ToArray();
             }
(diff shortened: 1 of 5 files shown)


Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
```diff
@@ -198,145 +179,90 @@ internal static DeviceInfo[] GetSuitablePhysicalDevices(Vk api)
                 EnabledLayerCount = 0
             };
```
gdkchan, line 181: Double empty line.

Ryujinx.Graphics.Vulkan/VulkanInstance.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                InstanceVersion = Vk.Version10;
+            }
+            else
+            {
+                uint rawInstanceVersion = 0;
+
+                if (api.EnumerateInstanceVersion(ref rawInstanceVersion) != Result.Success)
+                {
+                    rawInstanceVersion = Vk.Version11.Value;
```
gdkchan, line 34: Why does it assume 1.1 in case of error, and not 1.0?

Ryujinx.Graphics.Vulkan/VulkanInstance.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            api.EnumerateInstanceExtensionProperties(ReadOnlySpan<byte>.Empty, SpanHelpers.AsSpan(ref propertiesCount), extensionProperties).ThrowOnError();
+
+            unsafe
+            {
+                return extensionProperties.Select(x => Marshal.PtrToStringAnsi((IntPtr)x.ExtensionName)).ToHashSet();
+            }
+        }
+
+        public static HashSet<string> GetInstanceLayers(Vk api)
```
gdkchan, line 100: Maybe that should return `IReadOnlySet<string>`? `HashSet` implements that interface so the code wouldn't need to change, only the return type. The advantage is that we would be returning an immutable collection, which I think makes more sense in this case since the set of supported extensions is not supposed to be modified outside of this method.

Ryujinx.Graphics.Vulkan/VulkanInstance.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            if (!_disposed)
+            {
+                unsafe
+                {
+                    _api.DestroyInstance(Instance, null);
```
gdkchan, line 122: Maybe you can get rid of the unsafe context by using a span overload?

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    readonly struct VulkanPhysicalDevice
+    {
+        public readonly PhysicalDevice PhysicalDevice;
+        public readonly PhysicalDeviceFeatures PhysicalDeviceFeatures;
+        public readonly PhysicalDeviceProperties PhysicalDeviceProperties;
+        public readonly PhysicalDeviceMemoryProperties PhysicalDeviceMemoryProperties;
+        public readonly QueueFamilyProperties[] QueueFamilyProperties;
+        public readonly string DeviceName;
+        public readonly HashSet<string> DeviceExtensions;
```
gdkchan, line 19: If that method is changed to `IReadOnlySet<string>`, this field will need to change too.

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            api.EnumerateDeviceExtensionProperties(PhysicalDevice, Span<byte>.Empty, SpanHelpers.AsSpan(ref propertiesCount), extensionProperties).ThrowOnError();
+
+            unsafe
+            {
+                DeviceExtensions = extensionProperties.Select(x => Marshal.PtrToStringAnsi((IntPtr)x.ExtensionName)).ToHashSet();
+            }
+        }
+
+        public string Id => $"0x{PhysicalDeviceProperties.VendorID:X}_0x{PhysicalDeviceProperties.DeviceID:X}";
```
gdkchan, line 56: I suppose this will need to change eventually due to #3547.

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public bool IsDeviceExtensionPresent(string extension) => DeviceExtensions.Contains(extension);
+
+        public DeviceInfo ToDeviceInfo()
+        {
+            return new DeviceInfo(
+                    Id,
```
gdkchan, line 63: Excessive indentation here.

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
gdkchan: Also not sure why it is public if you're already exposing a `IsDeviceExtensionPresent` method. Presumably all checks should be done using that method instead of accessing the `HashSet`.

Ryujinx.Graphics.Vulkan/VulkanInstance.cs
marysaka: Because vkEnumerateInstanceVersion is 1.1 only so if the symbol is defined, the Vulkan instance (not device) is assumed compatible with 1.1

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
marysaka: > Also not sure why it is public if you're already exposing a IsDeviceExtensionPresent

It is also used to select enabled extension [here](https://github.com/Ryujinx/Ryujinx/pull/4632/files#diff-3a827561fb68d197a13803562577589e452e0f71fa39704a94960c4d6632cfa1R509).

gdkchan (approved):
lgtm, thanks.

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -0,0 +1,70 @@
...
+        public readonly string DeviceName;
+        public readonly IReadOnlySet<string> DeviceExtensions;
+
+        public VulkanPhysicalDevice(Vk api, PhysicalDevice physicalDevice)
+        {
+            PhysicalDevice = physicalDevice;
+            PhysicalDeviceFeatures = api.GetPhysicalDeviceFeature(PhysicalDevice);
+
+            api.GetPhysicalDeviceProperties(PhysicalDevice, out var physicalDeviceProperties);
+            PhysicalDeviceProperties = physicalDeviceProperties;
```
riperiperi, line 28: Is there a reason this one is copied from a local variable? The one below isn't.

riperiperi (approved):
This makes a lot more sense. LGTM!

Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
marysaka: the unsafe block right after would then need a fixed expression for `DeviceName`.

PR #4633, opened 2023-04-02
hle: Set ProcessResult name from NACP

TSRBerry:
This PR reads the `Name` for ProcessResult from the NACP, so `Name` contains the title name which is more a lot more useful to know when we receive logs from users.

The title name will be logged in the language chosen by the user.

### DIFF
--- a/Ryujinx.HLE/Loaders/Processes/ProcessLoaderHelper.cs
+++ b/Ryujinx.HLE/Loaders/Processes/ProcessLoaderHelper.cs
@@ -405,7 +405,16 @@ namespace Ryujinx.HLE.Loaders.Processes
             // Once everything is loaded, we can load cheats.
             device.Configuration.VirtualFileSystem.ModLoader.LoadCheats(programId, tamperInfo, device.TamperMachine);
 
-            return new ProcessResult(metaLoader, applicationControlProperties, diskCacheEnabled, allowCodeMemoryForJit, processContextFactory.DiskCacheLoadState, process.Pid, meta.MainThreadPriority, meta.MainThreadStackSize, device.System.State.DesiredTitleLanguage);
+            return new ProcessResult(
+                metaLoader,
+                applicationControlProperties,
+                diskCacheEnabled,
+                allowCodeMemoryForJit,
+                processContextFactory.DiskCacheLoadState,
+                process.Pid,
+                meta.MainThreadPriority,
+                meta.MainThreadStackSize,
+                device.System.State.DesiredTitleLanguage);
         }
 
         public static Result LoadIntoMemory(KProcess process, IExecutable image, ulong baseAddress)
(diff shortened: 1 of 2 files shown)


Ryujinx.HLE/Loaders/Processes/ProcessLoaderHelper.cs
```diff
@@ -405,7 +405,7 @@ public static bool LoadKip(KernelContext context, KipExecutable kip)
             // Once everything is loaded, we can load cheats.
             device.Configuration.VirtualFileSystem.ModLoader.LoadCheats(programId, tamperInfo, device.TamperMachine);
 
-            return new ProcessResult(metaLoader, applicationControlProperties, diskCacheEnabled, allowCodeMemoryForJit, processContextFactory.DiskCacheLoadState, process.Pid, meta.MainThreadPriority, meta.MainThreadStackSize);
+            return new ProcessResult(metaLoader, applicationControlProperties, diskCacheEnabled, allowCodeMemoryForJit, processContextFactory.DiskCacheLoadState, process.Pid, meta.MainThreadPriority, meta.MainThreadStackSize, device.System.State.DesiredTitleLanguage);
```
gdkchan, line 408: Can you split that into multiple lines?

Ryujinx.HLE/Loaders/Processes/ProcessResult.cs
```diff
@@ -35,7 +36,9 @@ public struct ProcessResult
             IDiskCacheLoadState        diskCacheLoadState,
             ulong                      pid,
             byte                       mainThreadPriority,
-            uint                       mainThreadStackSize)
+            uint                       mainThreadStackSize,
+            TitleLanguage              titleLanguage
+            )
```
gdkchan, line 41: `)` should be on the line above.

Ryujinx.HLE/Loaders/Processes/ProcessResult.cs
```diff
@@ -50,7 +53,7 @@ public struct ProcessResult
             {
                 ulong programId = metaLoader.GetProgramId();
 
-                Name          = metaLoader.GetProgramName();
+                Name          = ApplicationControlProperties.Title[(int)titleLanguage].NameString.ToString();
```
gdkchan, line 56: Is the `titleLanguage` passed here checked to ensure the name for that language actually exists for the game?

TSRBerry: I didn't think it could be possible for a game to just not have a title for a language, but it doesn't hurt to add a check here. Will do that!

gdkchan: They don't need to have a title for every possible language. Some games don't even have an "AmericanEnglish" title.

marysaka (approved):

WilliamWsyHK:
Tested there were regression from 1.1.689
image
(`Mario Party` becomes `buffet`)

while this PR seems to fix it
image
(No more `buffet`, let's go Party!)

WilliamWsyHK (approved):

AcK77 (approved):

PR #4480, opened 2023-02-25
HLE: Refactoring of ApplicationLoader

AcK77:
This PR attempt to refactor `ApplicationLoader` in a better way for avoid duplicated code and helping in a potential multi-processes support in the future.

In that way, some things are a bit "wrong" but should works as expected since we can only run one Application at a time.
I've done some loading testing but it could be nice to test different loading too, like Homebrew, NSP Homebrew, games using `jit` services, and more...

PR #4643, opened 2023-04-05
HLE: Deal with empty title names properly

TSRBerry:
This restores the behavior we had before #4480 to read title names in case the name of the desired language is empty.

### DIFF
--- a/Ryujinx.Ava/AppHost.cs
+++ b/Ryujinx.Ava/AppHost.cs
@@ -321,13 +321,11 @@ namespace Ryujinx.Ava
             _viewModel.IsGameRunning = true;
 
             var activeProcess   = Device.Processes.ActiveApplication;
-            var nacp            = activeProcess.ApplicationControlProperties;
-            int desiredLanguage = (int)Device.System.State.DesiredTitleLanguage;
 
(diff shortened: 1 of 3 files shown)


TSRBerry:
I thought it wouldn't even have array entries if they are empty, that's why I added these checks:
https://github.com/Ryujinx/Ryujinx/blob/63dedbda862533342fd8fac148f0c1d6427c9d84/Ryujinx.HLE/Loaders/Processes/ProcessResult.cs#L55-L57

But clearly they weren't enough.

gdkchan:
Actually I think they are since some languages were only added later, so on older games they probably don't exist and the array is smaller.

gdkchan:
> The array should always have the same size, not sure how it behaves with languages which were added later. https://github.com/Thealexbarney/LibHac/blob/6f6836c302aace7b37d4957746b17c2152aaa7b5/src/LibHac/Ns/ApplicationControlProperty.cs#L10
> 
> Do I keep the checks, or should I remove them?

Yea, I just checked and it seems that this size was always fixed to 16 entries (so they were smart enough to keep some "reserved" entries from the beginning that they used later). So I believe you can remove the bounds check.

marysaka (approved):

gdkchan (approved):
lgtm, thanks.

PR #4662, opened 2023-04-10
GPU: Keep rendered textures without any pool references alive

riperiperi:
Occasionally games are very wasteful and clear/write to a texture without ever sampling it. As rendered textures in NVN games seem to mostly have overlapping memory ranges (they are temporary), the texture will eventually get overwritten.

Normally, this would trigger a removal from the auto delete cache, but a pool entry would keep the texture alive. This is important so that games don't recreate all of their render targets every frame. However, with these textures that are never used, they will get deleted immediately and recreated on the next frame. This might also happen with textures that are rendered then copied somewhere else, as that also won't use a pool entry.

Because of the incompatible overlap, there won't be a data upload, but both OpenGL and Vulkan have costs when creating new texture storage. It really depends on your driver.

This change makes it so the ShortTextureCache can keep textures that have never had a pool reference alive for a few frames, so they're not constantly being created and deleted.

This improves performance in Zelda BOTW v1.6 a little bit, as it was recreating a number of textures every frame. v1.0 did not exhibit this behaviour. Example of a frame:

```
00:00:48.128 |E| GPU.MainThread Gpu : New texture: 320x180x1 1lv Texture2D R11G11B10Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 96x96x1 1lv Texture2D R16G16B16A16Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 48x48x1 1lv Texture2D R16G16B16A16Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 24x24x1 1lv Texture2D R16G16B16A16Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 12x12x1 1lv Texture2D R16G16B16A16Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 6x6x1 1lv Texture2D R16G16B16A16Float
00:00:48.129 |E| GPU.MainThread Gpu : New texture: 696x42x1 1lv Texture2D R8G8B8A8Unorm
00:00:48.130 |E| GPU.MainThread Gpu : New texture: 2400x1140x1 1lv Texture2D R8Unorm
```
The last texture is pretty big. If your driver decides to clear/process the memory on creation, then it could be wasting a lot of time. Note that none of these are views, they all create storage.

It's possible that other games do this, it just depends how careless they are when it comes to rendering stuff that is never used. I'd recommend testing a bunch of games to ensure this doesn't break anything.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/AutoDeleteCache.cs
@@ -1,5 +1,4 @@
 using System.Collections;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace Ryujinx.Graphics.Gpu.Image
@@ -9,6 +8,7 @@ namespace Ryujinx.Graphics.Gpu.Image
     /// </summary>
     class ShortTextureCacheEntry
     {
+        public bool IsAutoDelete;
         public readonly TextureDescriptor Descriptor;
         public readonly int InvalidatedSequence;
         public readonly Texture Texture;
@@ -21,6 +21,17 @@ namespace Ryujinx.Graphics.Gpu.Image
         public ShortTextureCacheEntry(TextureDescriptor descriptor, Texture texture)
         {
             Descriptor = descriptor;
+            InvalidatedSequence = texture.InvalidatedSequence;
+            Texture = texture;
+        }
+
+        /// <summary>
+        /// Create a new entry on the short duration texture cache from the auto delete cache.
+        /// </summary>
+        /// <param name="texture">The texture</param>
+        public ShortTextureCacheEntry(Texture texture)
+        {
+            IsAutoDelete = true;
             InvalidatedSequence = texture.InvalidatedSequence;
             Texture = texture;
         }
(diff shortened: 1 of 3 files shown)


riperiperi:
I forgot to mention - one important thing about this using the short duration cache is that it gets removed from it when a pool reference is added. That's useful for this case too, as the reason it was added in the first place was not having a pool reference to keep it alive.

gdkchan:
Is it not simpler to just add an additional rule to the auto delete cache that a texture can only be deleted if it has not been used for N frames?

riperiperi:
I still think this fits best on the short texture cache. It has the same removal trigger (new pool reference) and the timed component. The AutoDeleteCache is entirely concerned with keeping gpu modified data alive, and flushing it back to memory on deletion.

Doing this for all AutoDeleteCache textures could make a bad situation worse with regards to games like Fast RMX which create and destroy a lot of textures and end up almost filling VRAM with them. Maybe in future we can make overlapped textures release their storage textures and then regain them, in which case we'd want to keep overlapping texture objects alive without storage, but as of right now it's not a great idea.

marysaka (approved):

gdkchan (approved):
Seems reasonable.

PR #4663, opened 2023-04-10
Avalonia - Force activate parent window before dialog is shown

MutantAura:
Fixes an issue in https://github.com/Ryujinx/Ryujinx/issues/3662 where content dialogs would not be spawned if the window was minimized or otherwise lost focus by the user.


TSRBerry (approved):
Thank you! I wasn't sure how to tackle this issue, since I wanted to figure out the cause of it. But since it's probably some kind of Avalonia issue this fix is probably enough imo!

PR #4667, opened 2023-04-10
Vulkan: add situational "Fast Flush" mode

riperiperi:
The AutoFlushCounter class was added to periodically flush Vulkan command buffers throughout a frame, which reduces latency to the GPU as commands are submitted and processed much sooner. This was done by allowing command buffers to flush when framebuffer attachments changed.

However, some games have incredibly long render passes with a large number of draws, and really aggressive data access that forces GPU sync.

The Vulkan backend could potentially end up building a single command buffer for 4-5ms if a pass has enough draws, such as in BOTW. In the scenario where sync is waited on immediately after submission, this would have to wait for the completion of a much longer command buffer than usual.

The solution is to force command buffer submission periodically in a "fast flush" mode. This will end up splitting render passes, but it will only enable if sync is aggressive enough.

This should improve performance in GPU limited scenarios, or in games that aggressively wait on synchronization. In some games, it may only kick in when res scaling. It won't trigger in games like SMO where sync is not an issue, or games like Scarlet/Violet at 1x when GPU performance is not the bottleneck.

Improves performance in Pokemon Scarlet/Violet (res scaled) and BOTW (in general).

### DIFF
--- a/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
+++ b/Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Ryujinx.Common.Logging;
+using System;
 using System.Diagnostics;
 using System.Linq;
 
@@ -8,6 +9,9 @@ namespace Ryujinx.Graphics.Vulkan
     {
         // How often to flush on framebuffer change.
         private readonly static long FramebufferFlushTimer = Stopwatch.Frequency / 1000;
+
+        // How often to flush on draw.
+        private readonly static long DrawFlushTimer = Stopwatch.Frequency / 666;
 
         private const int MinDrawCountForFlush = 10;
         private const int MinConsecutiveQueryForFlush = 10;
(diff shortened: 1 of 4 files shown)


Ryujinx.Graphics.Vulkan/AutoFlushCounter.cs
```diff
@@ -9,10 +10,24 @@ internal class AutoFlushCounter
         // How often to flush on framebuffer change.
         private readonly static long FramebufferFlushTimer = Stopwatch.Frequency / 1000;
 
+        // How often to flush on draw when fast flush mode is enabled.
+        private readonly static long DrawFlushTimer = Stopwatch.Frequency / 666;
```
gdkchan, line 14: It would be nice to have `const`s for those values with a name that reflects their meaning.

riperiperi: These need to be static as Stopwatch.Frequency is evaluated at runtime. Do you mean the divisor? I could add `// 1.5ms` after it to explain why it's being done

gdkchan: Yea, I was talking about the divisor.

gdkchan (approved):
lgtm, seems reasonable.

PR #4671, opened 2023-04-11
Avalonia - Move swkbd message null check into constructor

MutantAura:
Fixes an issue in https://github.com/Ryujinx/Ryujinx/issues/3662 where the software keyboard in Avalonia does not populate any example text the game provides such as default character names etc.

Before:
image

After: 
image


TSRBerry (approved):

MutantAura:
> lgtm, thanks. I suppose making the old approach work would require calling `OnPropertyChanged` when the `Message` is modified?

I'd assume so, either way I prefer assigning everything together in a block syntax-wise.

TSRBerry:
True! It doesn't really matter for this, but that would have been cleaner yeah.

PR #4673, opened 2023-04-13
Fix incorrect fragment origin when YNegate is enabled

gdkchan:
The `YNegate` register value seems to affect the origin of `gl_FragCoord` on shaders. Since it did not take this into account, games that had `YNegate` enabled and that used `gl_FragCoord` on the fragment shader to calculate coords to sample a framebuffer texture for example, would be upside down.

This uses different approaches per renderer to flip `gl_FragCoord.y`. On OpenGL it just adds `layout (origin_upper_left) in vec4 gl_FragCoord;` on the fragment shader which changes the origin for those coords. Vulkan, however, does not support this (it only supports a upper left origin), so for Vulkan I added a new viewport size field on the support buffer, where the viewport width and height is written when `YNegate` is enabled. The shader then does `viewportSize.y - gl_FragCoord.y` for every `gl_FragCoord.y` access.

This depends on #4646 since I had to do changes to the support buffer, and those changes make this process much simpler (no need to update code on both backends for the new values or the attribute mess). One thing that is still pretty bad is having to pass the value to the backends to update the support buffer. This is duplicate code that shouldn't exist, and I don't thin including those case specific parameters on the backend is very good in general, but I haven't looked at how it can be improved yet. Ideally the support buffer would be updated entirely from the GPU project and the backends wouldn't need to care about it, but there are probably some obstacles to making that work. Maybe a better way forward is updating some values on the GPU project, and some on the backend.

Fixes upside down rendering on a few OpenGL games, like 20XX and Go Rally.
Before:
image
image
After:
image
image
Only OpenGL games should be affected. It is probably a good idea to test other OpenGL games that were fine before to ensure they are still fine.

Thanks to Furoryan for telling me about this fix.

### DIFF
--- a/src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
+++ b/src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
@@ -188,10 +188,19 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Glsl
                 context.AppendLine();
             }
 
-            if (context.Config.Stage == ShaderStage.Fragment && context.Config.GpuAccessor.QueryEarlyZForce())
+            if (context.Config.Stage == ShaderStage.Fragment)
             {
-                context.AppendLine("layout(early_fragment_tests) in;");
-                context.AppendLine();
+                if (context.Config.GpuAccessor.QueryEarlyZForce())
+                {
+                    context.AppendLine("layout (early_fragment_tests) in;");
+                    context.AppendLine();
+                }
+
+                if (context.Config.Properties.OriginUpperLeft)
+                {
+                    context.AppendLine("layout (origin_upper_left) in vec4 gl_FragCoord;");
+                    context.AppendLine();
+                }
             }
 
             if ((info.HelperFunctionsMask & HelperFunctionsMask.MultiplyHighS32) != 0)
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
@@ -161,6 +161,18 @@ namespace Ryujinx.Graphics.Shader.Instructions
                     // FragCoord X/Y must be divided by the render target scale, if resolution scaling is active,
                     // because the shader code is not expecting scaled values.
                     res = context.FPDivide(res, context.Load(StorageKind.ConstantBuffer, SupportBuffer.Binding, Const((int)SupportBufferField.RenderScale), Const(0)));
+
+                    if (op.Imm10 == AttributeConsts.PositionY && context.Config.Options.TargetApi != TargetApi.OpenGL)
+                    {
+                        // If YNegate is enabled, we need to flip the fragment coordinates vertically, unless
+                        // the API supports changing the origin (only OpenGL does).
+                        if (context.Config.GpuAccessor.QueryYNegateEnabled())
+                        {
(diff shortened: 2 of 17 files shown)


marysaka (approved):
lgtm

riperiperi:
This will update the support buffer on every attachment change when yNegate is enabled. I'm not sure how much this will affect games, or the full extent of games that use yNegate (probably mostly opengl guest). Maybe mirrors could help this, though on desktop they should probably only be enabled on the support buffer. I don't think there's any pretty way of doing this only when the shader needs it.

gdkchan:
It only needs to update it when the viewport size changes (in addition to YNegate being enabled). For most draws I'd expect the viewport size to be the same.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
```diff
@@ -243,10 +243,19 @@ public static void Declare(CodeGenContext context, StructuredProgramInfo info)
...
             {
-                if (isFragment && context.Config.GpuAccessor.QueryEarlyZForce())
+                if (isFragment)
                 {
-                    context.AppendLine("layout(early_fragment_tests) in;");
-                    context.AppendLine();
+                    if (context.Config.GpuAccessor.QueryEarlyZForce())
+                    {
+                        context.AppendLine("layout (early_fragment_tests) in;");
+                        context.AppendLine();
```
patrick-hovsepian, line 251: is it possible/does it make sense to collapse the line append in the previous statement? i.e. `"..in;\n");`

AcK77:
Needs a rebase

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Declarations.cs
gdkchan: I think it should work, but for consistency I prefer to call `context.AppendLine();`.

marysaka (approved):

src/Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
```diff
@@ -161,6 +161,18 @@ public static void Ipa(EmitterContext context)
...
                     // because the shader code is not expecting scaled values.
                     res = context.FPDivide(res, context.Load(StorageKind.ConstantBuffer, SupportBuffer.Binding, Const((int)SupportBufferField.RenderScale), Const(0)));
+
+                    if (op.Imm10 == AttributeConsts.PositionY && context.Config.Options.TargetApi != TargetApi.OpenGL)
+                    {
+                        // If YNegate is enabled, we need to flip the fragment coordinates vertically, unless
+                        // the API supports changing the origin (only OpenGL does).
+                        if (context.Config.GpuAccessor.QueryYNegateEnabled())
+                        {
+                            Operand viewportHeight = context.Load(StorageKind.ConstantBuffer, 0, Const((int)SupportBufferField.ViewportSize), Const(1));
```
riperiperi, line 171: Here, the viewport Y is being inverted _after_ the frag coord has been divided by the scale, so it's at 1x. However, the viewport size is set _after_ scale is multiplied at `StateUpdater.cs:747` and `StateUpdater.cs:1448`, meaning it's not at 1x. I think either this needs to happen before scale, or the viewport size needs to be transformed to 1x scale before upload.

gdkchan: Yep, nice catch. I fixed it now and used the same method to calculate the viewport size from the viewport transform update method, and the shader update method, and removed the scale multiplication. It has been tested with 1x and 2x scale on an affected game.

riperiperi (approved):
LGTM, nice work!

PR #4657, opened 2023-04-08
Allow window to remember its size, position and state (GTK + Avalonia)

MutantAura:
Based on #4439 with support for Avalonia. 

As the two frameworks fight over data types and the like, GTK is given priority for now. It will be a quick job to edit and expand upon once deprecated e.g. WindowState could start in fullscreen instead of just maximised etc.

CC @HaizenTrist

Issue #4599, opened 2023-03-25
[Feature Request] Pick windows size (similar to yuzu) instead of resizing manually


MCPabs:
### Overview

Implement an option to choose between regular window sizes (720/1080/1440 etc) and make it persistent (unless we change it).

Extra: Option to apply resolution scale filter automatically based on your windows size (2x; 3x etc).

### Smaller details

N/A

### Nature of request

N/A

### Why would this feature be useful?

If well implemented, we would avoid weird window sizes due to manual resizing and also save it, we would avoid doing it every single time we start Ryunjix.

This is a picky feature, I am enjoying Ryunjix a lot.

Revan20911:
Somewhat new to OSS but I think I can handle this one!

MutantAura:
> Somewhat new to OSS but I think I can handle this one!

Do you want this assigned?
May want to springboard off: https://github.com/Ryujinx/Ryujinx/pull/4657 as once this it will implement one of the requests here and also adds mechanisms to save and set window states.

Revan20911:
Yea that sounds like a good idea. Go ahead and assign it to me and I will start looking at it.

PR #4681, opened 2023-04-15
Added dropdown menu for resolutions

Revan20911:
#4599 

Added a dropdown menu that lets users select a resolution. Upon hitting save the window is resized. 

This incorporates changes made in #4657 which also has the size saved upon exiting the application.

### DIFF
--- a/Ryujinx/Ui/MainWindow.cs
+++ b/Ryujinx/Ui/MainWindow.cs
@@ -1011,6 +1011,14 @@ namespace Ryujinx.Ui
             }
         }
 
+        public void UpdateWindowResolution(){
+            
+            int width = ConfigurationState.Instance.Ui.WindowSizeWidth.Value;
+            int height = ConfigurationState.Instance.Ui.WindowSizeHeight.Value;
+
+            Resize(width, height);
+        }
+
         public void UpdateGraphicsConfig()
         {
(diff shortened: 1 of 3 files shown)


IsaacMarovitz:
It is more crucial this is added to Avalonia, as GTK will be removed in the not-too-distant future.

TSRBerry:
> It is more crucial this is added to Avalonia

Currently everything needs to be implemented for both UIs, it doesn't really matter which UI they begin with.

>  as GTK will be removed in the not-too-distant future.

We are still pretty far away from that, so as long as both UIs are covered everything is fine. Ava still has issues and we are still in the process of cleaning it up.

Revan20911:
I'll work on getting it added into the Avalonia version. I started with GTK because there was already another request somewhat related to this that was done for GTK so it was easy to spring off of that.

riperiperi:
Needs rebased now that #4657 has merged.

AcK77:
> Needs rebased now that https://github.com/Ryujinx/Ryujinx/pull/4657 has merged.

As stated by riperiperi more than a month ago, can you rebase it please ?

GamerzHell9137:
Not sure if intentional or not but the chosen resolution once the emulator is reopened isn't the one chosen.

PR #4685, opened 2023-04-16
Change SMAA filter texture clear method

gdkchan:
This changes the method used for clear for the SMAA filter textures. Instead of using the helper shader, it calls the clear method on `PipelineBase` directly, which uses `vkCmdClearAttachments`. This fixes a crash on Intel GPUs when using the SMAA filter. I still need to investigate why `HelperSHadeer.Clear` crashes on Intel, since some games will use that.

I also cleaned up the code a bit and removed some things that doesn't make sense (like setting viewport and scissor before a compute dispatch ???).

Testing is welcome to ensure the SMAA filter is still working as before.


gdkchan:
This also fixes SMAA on macOS. The reason it was not working before is because the component type was set to `ComponentType.UnsignedInteger` which is wrong, since unorm formats are read/written as float.

marysaka (approved):

riperiperi (approved):
Cool, seems to be working fine.

PR #4688, opened 2023-04-16
Vulkan: HashTableSlim lookup optimization

riperiperi:
Really small optimization: since we've already calculated the hash for the input pipeline state, we can use it to eliminate mismatching pipelines within the same hashtable bucket a lot faster.

This mostly helps when there are a lot of pipeline variants for a shader collection.

### DIFF
--- a/Ryujinx.Graphics.Vulkan/HashTableSlim.cs
+++ b/Ryujinx.Graphics.Vulkan/HashTableSlim.cs
@@ -15,6 +15,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         private struct Entry
         {
+            public int Hash;
             public K Key;
             public V Value;
         }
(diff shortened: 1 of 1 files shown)


gdkchan (approved):
lgtm, thanks.
On the subject of pipeline hashes, the pipeline hashing on `GetHashCode` can be improved. The simple and fast way that is usually used to hash values is:
```cs
x *= SomePrimeValue;
x ^= valueToHash;
```
But instead I did:
```cs
x ^= valueToHash;
x *= SomePrimeValue;
```
Which I guess could produce somewhat worse hashes? From what I recall I did measure the amount of collisions and the hash table distribution when working on this and I think it was fine though.

marysaka (approved):

PR #4692, opened 2023-04-17
Fix Avalonia shader dumping confirmation dialog text

tech-ticks:
Clicking "No" disables shader dumping, but the dialog says that it's the other way around.


marysaka (approved):

TSRBerry:
Wouldn't it be better and less confusing to flip the condition in the code?

MutantAura:
Yea I think just flip the dialog condition. The meaning is clear but the dialog isn't behaving accordingly, flipping the wording just feels awkward.

tech-ticks:
I guess the "enabled" part of the localization key `DialogPerformanceCheckShaderDumpEnabledConfirmMessage` is a bit misleading, but yeah, flipping the condition makes sense

tech-ticks:
I've just noticed that I was on an old state and this has already been fixed in https://github.com/Ryujinx/Ryujinx/commit/4b495f3333c26b530710bf5c4d95870afd1f9570... Sorry for that!

PR #4702, opened 2023-04-20
Linux: Automatically increase vm.max_map_count if it's too low

### LINKED ISSUE #4941: Ryujinx Crashes During TotK Gameplay
### Description of the issue

Ryujinx crashes seemingly randomly during TotK. This did not happen at all during the initial sky islands gameplay.

Info about the first crash (1/2)
It only started after I dropped to the ground and started exploring. The first time it crashed when I was going to meet Hoz during the "Crisis at Hyrule Castle" mission. I was attempting to climb on the wall shown in the pictures when the game crashed.
image
image

Info about the second crash (2/2)
The second crash occurred after I talked to Hoz. I fast traveled to Kyononis Shrine after talking to Hoz and was walking back to talk to Purah for the second time when it crashed.

Note about logs: I pasted all three of the three log files in the Ryujinx logs folder. I think only the first two should be useful (Ryujinx_1.1.799_2023-05-14_00-06-01.log  and  Ryujinx_1.1.799_2023-05-14_00-20-45.log) but I added all three just in case.

### Reproduction steps

The crashes do not seem reproducible because I reloaded my save after both crashes and tried not to change what I did in order to see if the crashes would happen again and they did not.

### Log file

[Ryujinx_1.1.799_2023-05-14_00-06-01.log](https://github.com/Ryujinx/Ryujinx/files/11471412/Ryujinx_1.1.799_2023-05-14_00-06-01.log)
[Ryujinx_1.1.799_2023-05-14_00-20-45.log](https://github.com/Ryujinx/Ryujinx/files/11471413/Ryujinx_1.1.799_2023-05-14_00-20-45.log)
[Ryujinx_1.1.799_2023-05-14_00-54-57.log](https://github.com/Ryujinx/Ryujinx/files/11471421/Ryujinx_1.1.799_2023-05-14_00-54-57.log)

### OS

Fedora 38 Linux

### Ryujinx version

1.1.799

### Game version

1.0.0

### CPU

AMD Ryzen 7 3700X (16) @ 3.600GHz

### GPU

AMD Radeon RX 5600 XT

### RAM

32GB

### List of applied mods

_No response_

### Additional context?

_No response_

TSRBerry:
This PR adds a workaround for #3372 and result checking for memory related pinvoke calls.

If a game adds too many memory map areas Ryujinx will crash with a segmentation fault, which makes these issues difficult to debug.
But since @gdkchan was able to find the cause of this type of crash, we are able to temporarily increase `vm.max_map_count` while Ryujinx is running and reset it to the original value once Ryujinx was closed.

The threshold value in the script is just the default value that the [kernel documentation](https://kernel.org/doc/Documentation/sysctl/vm.txt) for `vm.max_map_count` suggested. This check is just supposed to make sure we aren't messing with the settings if the user has modified them already.

---

Closes #4941


Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -76,17 +77,33 @@ public static bool Commit(IntPtr address, ulong size, bool forJit)
...
         {
             // Must be writable for madvise to work properly.
-            mprotect(address, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE);
+            if (mprotect(address, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE) != 0)
+            {
+                throw new SystemException(Marshal.GetLastPInvokeErrorMessage());
+            }
 
-            madvise(address, size, MADV_REMOVE);
+            if (madvise(address, size, MADV_REMOVE) != 0)
```
riperiperi, line 96: I hope nothing else races to change permissions, otherwise these could crash the emulator where otherwise they did not. I suppose Decommit isn't used right now, so it's more a worry for another time...

riperiperi (approved):
Looks sensible, though I'm not an expert on whether `262144` is high enough for all games. It would be nice if this stuff could be set when opening the main Ryujinx executable, too bad it can't.

TSRBerry:
I mean it could be, but we might not be able to revert to defaults if Ryujinx crashes.

Ryujinx.Memory/MemoryManagementUnix.cs
marysaka: Shouldn't we just return false instead of throwing everywhere here?

that kind of change the behavior of the function entirely... and the return value doesn't meaning anything anymore.

distribution/linux/Ryujinx.sh
```diff
@@ -11,4 +11,16 @@ if [ -f "$SCRIPT_DIR/Ryujinx.Headless.SDL2" ]; then
     RYUJINX_BIN="Ryujinx.Headless.SDL2"
 fi
 
+ORIG_MAX_MAP_COUNT=$(sysctl vm.max_map_count | awk -F' ' '{print $3}')
+
+if [ $ORIG_MAX_MAP_COUNT -le 65530 ]; then
+  echo "Increasing max amount of memory map areas..."
+  pkexec sh -c 'sysctl -w vm.max_map_count=262144'
```
marysaka, line 18: This required PolKit, we cannot assume that someone have it installed.

Ryujinx.Memory/WindowsShared/WindowsApi.cs
```diff
@@ -1,8 +1,10 @@
 using System;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 
 namespace Ryujinx.Memory.WindowsShared
 {
+    [SupportedOSPlatform("windows")]
```
marysaka, line 7: Cause warnings on build (see CI report)

marysaka (changes requested):

distribution/linux/Ryujinx.sh
TSRBerry: Do you have a suggestion on how to deal with that?

Ryujinx.Memory/MemoryManagementUnix.cs
TSRBerry: Afaik Ryujinx would crash either way even if we return false here.

TSRBerry:
Requesting both reviews, since I essentially changed the whole thing again.

Ryujinx.Ui.Common/Helper/LinuxHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.IO;
+using System.Runtime.Versioning;
+
+namespace Ryujinx.Ui.Common.Helper
+{
+    [SupportedOSPlatform("linux")]
+    public static class LinuxHelper
+    {
+        // NOTE: This value was determined by manual tests and might need to be increased again.
+        public const int RecommendedVmMaxMapCount = 65530;
```
TSRBerry, line 12: I'm pretty sure we can go lower here, but I'll need community testing to figure this out.
I don't have any of the game in the referenced issue and can't trigger the issue on my side.

Ryujinx.Ui.Common/Helper/LinuxHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.IO;
+using System.Runtime.Versioning;
+
+namespace Ryujinx.Ui.Common.Helper
+{
+    [SupportedOSPlatform("linux")]
+    public static class LinuxHelper
+    {
+        // NOTE: This value was determined by manual tests and might need to be increased again.
+        public const int RecommendedVmMaxMapCount = 524288;
```
TSRBerry, line 12: I'm pretty sure we can go lower here, but I'll need community testing to figure this out.
I don't have any of the game in the referenced issue and can't trigger the issue on my side.

riperiperi (commented):
How will this behave when Ryujinx is installed as a flatpak? I doubt that it'll be able to write sysctl.d config, not sure about the vm map count.

TSRBerry:
The flatpak probably wouldn't be able to do any of these things. :thinking: 

I'll need to think of something for that. :/

TSRBerry:
@riperiperi Currently flatpak users would see this warning (even if pkexec is installed, because it doesn't have access to it).
Would this be okay?

image

Ideally we should use dbus to talk to polkitd, but I have no idea how to do that yet and it doesn't seem to be easy.

marysaka (approved):

src/Ryujinx.Ava/UI/Windows/MainWindow.axaml.cs
```diff
@@ -258,7 +259,64 @@ private void Initialize()
...
+                LocaleManager.Instance[LocaleKeys.LinuxVmMaxMapCountDialogTextSecondary],
+                LocaleManager.Instance[LocaleKeys.LinuxVmMaxMapCountDialogButtonUntilRestart],
+                LocaleManager.Instance[LocaleKeys.LinuxVmMaxMapCountDialogButtonPersistent],
+                LocaleManager.Instance[LocaleKeys.InputDialogNo],
+                (int)Symbol.Help
+            );
+
+            int rc;
+
+            switch (response)
```
AcK77, line 292: I'm not a huge fan of getting the whole `switch` duplicated between both UI just because of the `response` cast.
Even more because the `LinuxHelper` class is in `Ryujinx.Ui.Common`, everything should goes there.

src/Ryujinx/Program.cs
```diff
@@ -264,6 +264,71 @@ void SetEnvironmentVariableNoCaching(string key, string value)
...
+                            "Ryujinx - Low limit for memory mappings detected",
+                            $"Would you like to increase the value of vm.max_map_count to {LinuxHelper.RecommendedVmMaxMapCount}?",
+                            "Some games might try to create more memory mappings than currently allowed. " +
+                            "Ryujinx will crash as soon as this limit gets exceeded.",
+                            buttonTexts,
+                            MessageType.Question);
+
+                        int rc;
+
+                        switch ((int)response)
```
AcK77, line 294: Sams comment as the Ava project.

src/Ryujinx.Ava/UI/Windows/MainWindow.axaml.cs
TSRBerry: Well that would be the case for the updater as well, but since we use 2 different frameworks to display the dialogs I don't think it makes sense to combine them.

AcK77: Dialogs are before this part of the code. Here it's just dialog responses parsing. That's why I said that. But I'm agree for the updater too, There should be only one class...

TSRBerry: The dialog response codes are also different per framework here, unfortunately.

AcK77: Maybe you could change them into the same int values.

TSRBerry: Meh, I kind of prefer to use the enum values here. Would you mind if I leave them as is?

AcK77 (approved):

PR #4703, opened 2023-04-21
Shader: Bias textureGather instructions on AMD/Intel

riperiperi:
There seems to be a bit of a disagreement on how textureGather should select texels when it is exactly half way between them. On BOTW, this tie breaks incorrectly for AMD and Intel on windows on both Vulkan and OpenGL, but correctly on NVIDIA, Apple and mesa drivers.

This is most noticeable in BOTW, where the difference in texel choice breaks the shadows in the game with a weird looking misalignment. A small positive bias causes the instruction to return the correct texels, so I decided to make it determine and add the smallest bias it possibly could.

This PR adds a capability that reveals both the sub texel precision and whether it should be used to bias textureGather. In OpenGL this is assumed, because I don't think it's reported.

This needs to apply after scaling (and with unscaled texture sizes) so this is currently performed in the shader backend. The offset is `1 / (textureSize() * float(1 << (texelPrecisionBits + 1)))`

Some testing with both AMD and Intel GPUs to ensure I didn't blow up any other games would be nice, and also that it fixes other AMD GPUs (I only tested steam deck).

### DIFF
--- a/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
+++ b/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
@@ -4,6 +4,7 @@ using Ryujinx.Graphics.Shader.Translation;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Numerics;
 using static Spv.Specification;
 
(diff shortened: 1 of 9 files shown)


Ryujinx.Graphics.Shader/IGpuAccessor.cs
```diff
@@ -196,6 +196,15 @@ bool QueryHasUnalignedStorageBuffer()
             return false;
         }
 
+        /// <summary>
+        /// Queries host's gather operation precision bits for biasing their coordinates. Zero means no bias.
+        /// </summary>
+        /// <returns>Bits of gater operation precision to use for coordinate bias</returns>
```
gdkchan, line 202:
```suggestion
        /// <returns>Bits of gather operation precision to use for coordinate bias</returns>
```

riperiperi:
### AMD

#### Master:
image

#### PR:
image

riperiperi:
### Intel

#### Master* (not really, but yes for the purposes of this comparison)
image

#### PR
image

gdkchan (approved):
lgtm, thanks.

Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
```diff
@@ -1556,6 +1557,33 @@ SpvInstruction AssemblePVector(int count)
...
+                    // GPU requires texture gather to be slightly offset to match NVIDIA behaviour when point is exactly between two texels.
+                    // Offset by the gather precision divided by 2 to correct for rounding.
+                    var sizeType = pCount == 1 ? context.TypeS32() : context.TypeVector(context.TypeS32(), pCount);
+                    var pVectorType = pCount == 1 ? context.TypeFP32() : context.TypeVector(context.TypeFP32(), pCount);
+
+                    var bias = context.Constant(context.TypeFP32(), (float)(1 << (gatherBiasPrecision + 1)));
+                    var biasVector = context.CompositeConstruct(pVectorType, Enumerable.Repeat(bias, pCount).ToArray());
+
+                    var one = context.Constant(context.TypeFP32(), 1f);
+                    var oneVector = context.CompositeConstruct(pVectorType, Enumerable.Repeat(one, pCount).ToArray());
```
TSRBerry, line 1574: Is using LINQ fine here? I guess it probably is, but looking at the PRs aiming to reduce allocations they always reduce the usage of LINQ as well, so I'm wondering why it's fine here.

TSRBerry (approved):
Lgtm! I'm curious about one thing tho, so I left a comment with my question.

Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
gdkchan: I think the idea is removing it on hot paths (code that runs often), were the allocations and lower performance that LINQ has could have a significant impact. This is not really a hot path, so I don't think it would have a significant impact.

bunnei:
@riperiperi Unless I am missing something here, @Wollnashorn should be credited for this great fix that benefits both projects!

riperiperi:
What are you talking about? Was this user a contributor to Cemu?

riperiperi:
image
For reference, this was 6 months ago.

Shahil-Ayato:
> @riperiperi Unless I am missing something here, @Wollnashorn should be credited for this great fix that benefits both projects!

Instead of trying to do a big “gotcha moment” on our GitHub, next time reach out to us on discord to ask us about something like this, we’re not hard to get in touch with.

PR #4705, opened 2023-04-22
Removed MotionInput Calibration

### LINKED ISSUE #4366: Motion controls are reset involuntarily on the Pro Controller.
### Description of Issue

When aiming in Splatoon 2 and 3 using a Switch Pro Controller, it is common to have unintentional abrupt camera movements.

Even when the controller is left resting on a flat surface, there is a constant shaking of the camera that (to my knowledge) should not occur.

In all my tests, I have only encountered this problem in Ryujinx, both in Master and LDN builds (with their respective GTK and Avalonia GUI variants).

In addition to that:

 · I do not use software such as DS4Windows or Betterjoy.

 · I have tried changing the Gyro Deadzone settings, but to no avail.
 
 · I connect the controller to the computer via a cable, due to the lack of bluetooth support on my computer.

### Reproduction Steps

1) Connect an original Switch Pro controller to your computer.

2) Run any version of Ryujinx of your choice.

3) Enable motion controls both in the game and on the Ryujinx input screen.

4) Tilt the controller and point it at a target to verify that the camera does indeed reset unintentionally.

### Log File

[Ryujinx_1.1.0-ldn3.0.1_2023-02-04_22-58-26.log](https://github.com/Ryujinx/Ryujinx/files/10609777/Ryujinx_1.1.0-ldn3.0.1_2023-02-04_22-58-26.log)

### OS

Windows 10 21H2

### Ryujinx version

Any release to date (eg: Master: 1.1.599 | LDN: 1.1.0-ldn3.0.1)

### Game version

Splatoon 2: 5.5.1 | Splatoon 3: 2.1.0

### Environment?

- CPU: Intel(R) Core(TM) i7-4790K
- GPU: NVIDIA GTX 960
- RAM: 16GB
- Applied Mods: No

### Additional context?

https://user-images.githubusercontent.com/124469126/216792608-921c6088-3238-411e-be63-721c3ebab857.mp4

SpicerXD:
Don't know why this is here.
It just seems to set the filter to an identity. Which then quickly returns to where it's supposed to be anyways.
Fixes #4366

### DIFF
--- a/Ryujinx.Input/Motion/MotionInput.cs
+++ b/Ryujinx.Input/Motion/MotionInput.cs
@@ -12,7 +12,6 @@ namespace Ryujinx.Input
         public Vector3 Rotation      { get; set; }
 
         private readonly MotionSensorFilter _filter;
-        private int _calibrationFrame = 0;
 
         public MotionInput()
         {
@@ -29,26 +28,6 @@ namespace Ryujinx.Input
         {
             if (TimeStamp != 0)
             {
-                if (gyro.Length() <= 1f && accel.Length() >= 0.8f && accel.Z >= 0.8f)
-                {
-                    _calibrationFrame++;
-
-                    if (_calibrationFrame >= 90)
-                    {
-                        gyro = Vector3.Zero;
-
-                        Rotation = Vector3.Zero;
-
-                        _filter.Reset();
-
-                        _calibrationFrame = 0;
-                    }
-                }
-                else
-                {
-                    _calibrationFrame = 0;
-                }
-
                 Accelerometer = -accel;
 
                 if (gyro.Length() < deadzone)


Reonu:
This seems to fix the gyro controls re-centering themselves constantly in Hyrule Warriors AoC 👍

riperiperi:
I think it's meant to avoid drifting, but it clearly doesn't behave properly.

MelonSpeedruns:
Can confirm it fixes the bow aiming and motion shrines in Breath of the Wild!

riperiperi (approved):
I think this class does need to be improved in future, but this is good for avoiding bigger issues issue right now.

TSRBerry (approved):

sachahjkl:
NB : I still get the drift using a Dualsense controller via USB. To make sure my controller didn't misbehave, I checked the metrics for my gyro in DS4 and tested the controller on Cemu on BOTW -> no drift.
I'm on Ryujinx 1.1.762.

critkitten:
> 

same using a 8bit controller in switch mode.

P-G19:
I don't think it's right to send it here ig since PR is already merged. maybe make a new issue?

RoaminGrape:
Using CemuHook on a Dualsense with DS4Windows on 1.1.826 with Metroid Prime Remastered has the pointer setting immediately confirm its controller resting message. Afterward it will drift with a fast hard right and bring up continuous re-centering prompts. It performs correctly on other switch emulators.

OldManKain:
> Using CemuHook on a Dualsense with DS4Windows on 1.1.826 with Metroid Prime Remastered has the pointer setting immediately confirm its controller resting message. Afterward it will drift with a fast hard right and bring up continuous re-centering prompts. It performs correctly on other switch emulators.

Why use all these wrappers when Ryujinx has native DualSense support.

RoaminGrape:
> Why use all these wrappers when Ryujinx has native DualSense support.

In my case it's just DS4Windows. Getting games to recognize non-corded 360 controllers can be difficult across the board. Even without the wrapper, it behaves the same it seems.

PR #4707, opened 2023-04-22
Make config filename changable for releases & Log to Ryujinx directory if application directory is not writable

### LINKED ISSUE #1526: [Feature Request] Change File Log location directory

### LINKED ISSUE #2887: [Feature Request] Add Separate Configuration files for GitHub Artifacts builds
### What feature are you suggesting?
- Have separate Configuration files for GitHub artifacts

#### Overview:
- Make PR builds (GitHub Artifacts) have their own separate Configuration file instead of using the same one as the master build 

####Smaller Details: 
- This should only be created for users who open a PR build and it shouldn’t create a GitHub artifacts configuration file is the user has never opened one at all. 

- I’m not too sure on this one but, it should be noted that the person who does manage to PR this should be ready to import this change to Avalonia if possible or needed as having it be done for GTK and then not long after it’s not there for when Avalonia gets merged would be a big shame. 

#### Nature of Request:  
  - This would be very similar to what we current do with LDN builds as they create configuration files called `LDNConfig.json` Made specifically for the LDN builds. 

This would make it much less of a hassle to test PR builds as I know many people (including myself) have had their configuration file for master be deleted many times because we test PR’s.

This would help developers test their new pull requests much easier as they don’t need to worry about their configuration file for master being reset.

TSRBerry:
This PR ~~moves logging from the application directory to the Ryujinx directory and~~ moves the config filename to `ReleaseInformation.cs`.

~~The logging change was attempted a few times before, see #1530 and #2027.~~ However I think we shouldn't assume the application directory is writable, since it's usually not the case. It's better to write log files to locations we expect to be writable anyway.

~~As a user I also don't need to keep track of two different directories, as everything I need to be aware of is handled in one directory.~~

As outlined in #2887 the config file change allows users to test PR builds without needing to worry about the config file they use for our release builds.

---

Flatpak PR: flathub/org.ryujinx.Ryujinx#60

Closes #2887


marysaka (approved):

TSRBerry:
Rebased!

gdkchan:
> As outlined in https://github.com/Ryujinx/Ryujinx/issues/2887 the config file change allows users to test PR builds without needing to worry about the config file they use for our release builds.

Won't this cause people testing PRs to have a lot of stray config files in the Ryujinx folder?

TSRBerry:
There will just be one new `PRConfig.json`. The idea is that this config is only used by PR builds and testers won't need to backup and restore their main config file if they switch between PR and release builds.

The `PRConfig.json` could still get reset by different PRs, but that's at least a bit better than resetting the main config file.

src/Ryujinx.Common/Logging/Logger.cs
```diff
@@ -10,13 +10,13 @@ namespace Ryujinx.Common.Logging
 {
     public static class Logger
     {
-        private static readonly Stopwatch m_Time;
+        private static readonly Stopwatch Time;
```
AcK77, line 13:
```suggestion
        private static readonly Stopwatch _time;
```

src/Ryujinx.Common/Logging/Logger.cs
```diff
@@ -10,13 +10,13 @@ namespace Ryujinx.Common.Logging
...
 
-        private static readonly bool[] m_EnabledClasses;
+        private static readonly bool[] EnabledClasses;
```
AcK77, line 15:
```suggestion
        private static readonly bool[] _enabledClasses;
```

src/Ryujinx.Common/Logging/Logger.cs
```diff
@@ -10,13 +10,13 @@ namespace Ryujinx.Common.Logging
...
 
-        private static readonly List<ILogTarget> m_LogTargets;
+        private static readonly List<ILogTarget> LogTargets;
```
AcK77, line 17:
```suggestion
        private static readonly List<ILogTarget> _logTargets;
```

src/Ryujinx.Common/Logging/Logger.cs
```diff
@@ -10,13 +10,13 @@ namespace Ryujinx.Common.Logging
...
 
-        private static readonly StdErrAdapter _stdErrAdapter;
+        private static readonly StdErrAdapter StdErrAdapter;
```
AcK77, line 19:
```suggestion
        private static readonly StdErrAdapter _stdErrAdapter;
```

src/Ryujinx.Common/Utilities/FileSystemUtils.cs
```diff
@@ -0,0 +1,48 @@
+using System.IO;
+
+namespace Ryujinx.Common.Utilities
+{
+    public static class FileSystemUtils
+    {
+        public static void CopyDirectory(string sourceDir, string destinationDir, bool recursive)
```
AcK77, line 7: I think we already Copy/Move directories elsewhere in the code, maybe it's could be nice to use this one instead and remove duplicated code (if it's possible) ?

src/Ryujinx.Ui.Common/Helper/FileAssociationHelper.cs
```diff
@@ -11,21 +11,21 @@ namespace Ryujinx.Ui.Common.Helper
 {
     public static partial class FileAssociationHelper
     {
-        private static readonly string[] _fileExtensions = { ".nca", ".nro", ".nso", ".nsp", ".xci" };
+        private static readonly string[] FileExtensions = { ".nca", ".nro", ".nso", ".nsp", ".xci" };
```
AcK77, line 14: I'd prefer to keep private field with `_` and camelCase. (Same for other vars)

distribution/linux/Ryujinx.desktop
```diff
@@ -3,12 +3,12 @@ Version=1.0
 Name=Ryujinx
 Type=Application
 Icon=Ryujinx
-Exec=env DOTNET_EnableAlternateStackCheck=1 Ryujinx %f
+Exec=Ryujinx.sh %f
```
AcK77, line 6: Nice catch!

AcK77 (changes requested):

src/Ryujinx.Common/Utilities/FileSystemUtils.cs
TSRBerry: Done!

AcK77:
> However I think we shouldn't assume the application directory is writable, since it's usually not the case. It's better to write log files to locations we expect to be writable anyway.

Maybe we should check if the Ryujinx folder is writable and if not we can write log files to another folder as your PR suggest ?

TSRBerry:
Sure, although I'd prefer not writing to the application directory at all even if it's writable.
If you prefer the check I'll implement it in a bit.

TSRBerry:
The current state of this PR doesn't really hit the target anymore, but I guess these changes still offer slight improvements.

src/Ryujinx.Headless.SDL2/Program.cs
```diff
@@ -427,11 +427,27 @@ static void LoadPlayerConfiguration(string inputProfileName, string inputId, Pla
...
+                FileStream logFile = FileLogTarget.PrepareLogFile(AppDomain.CurrentDomain.BaseDirectory);
+
+                if (logFile == null)
+                {
+                    logFile = FileLogTarget.PrepareLogFile(AppDataManager.BaseDirPath);
+
+                    if (logFile == null)
+                    {
+                        Logger.Error?.Print(LogClass.Application, "No writable log directory available. Make sure either the application directory or the Ryujinx directory is writable.");
+                        ;
```
AcK77, line 439:
```suggestion

```

AcK77 (approved):

marysaka (approved):

PR #4708, opened 2023-04-23
Use vector transform feedback outputs with fragment shaders

### LINKED ISSUE #4546: Characters don't appear on Vulkan and OpenGl appear sideways[Bug] [Vulkan] [Gpu]
### Description of the issue

When I open Pokemon Legends Arceus or Pokemon Scarlett everything seems to work as normal but when the actual gameplay starts all characters are invisible
My gpu is a Rx66000 I was running on 23.3.1 driver but I made a downgrade trying to solve this problem and wasn't  fixed

and error always appear on the cmd:
.NET ThreadPool Worker Gpu : Background Compilation failed: Unexpected API error "ErrorUnknown".

### Reproduction steps

Configure to vulkan open the game and pass the intruduction

### Log file

[Ryujinx_1.1.0+c12233c_2023-03-14_02-47-06.log](https://github.com/Ryujinx/Ryujinx/files/10969849/Ryujinx_1.1.0%2Bc12233c_2023-03-14_02-47-06.log)

### OS

Windows 11

### Ryujinx version

lastest and I tried on 1.1.0+c12233c

### Game version

_No response_

### CPU

_No response_

### GPU

Rx6600

### RAM

_No response_

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
Transform feedback is a feature that writes the VTG stages output into a buffer. It is possible to write different output components to different offsets into that buffer. To make that work, instead of declaring a `vec4` as output, the emulator breaks that down into 4 individual `float` outputs. The problem is that drivers (other than NVIDIA) usually have problems with this for some reason, resulting in weird bugs, so I made #3832 to address this.

When there are consecutive outputs that are supposed to be written into consecutive buffer offsets, it just "joins" them and makes a vector. The problem is that right now, this is only done when the draw has no fragment shader, because for the cases where it does have a fragment shader, the VTG outputs would not match the fragment shader inputs. It turns out that this is not a problem. I tested games that uses transform feedback and have a fragment shader, and had no errors with OpenGL or Vulkan, with both NVIDIA and Intel on Windows, and @riperiperi tested it on AMD.

This fixes characters not rendering on Pokémon Legends Arceus with newer AMD driver.
Fixes #4546.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 4703;
+        private const uint CodeGenVersion = 4708;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
--- a/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 4708;
+        private const uint CodeGenVersion = 4707;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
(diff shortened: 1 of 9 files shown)


riperiperi (commented):
I can confirm this works on AMD+Vulkan+Windows, but it doesn't seem to work on OpenGL. The console prints "Program Link Failed for unknown reason.", which isn't particularly helpful. I'm not sure if this worked on OpenGL in the past.

Master doesn't do this, but it looks like... this.....
image

riperiperi:
This error appears when compiling GLSL with shaderc:
```
00:00:44.132 |E| .NET ThreadPool Worker Gpu GlslToSpirv: Shader compilation error: CompilationError Ryu:7: warning: '#extension' : extension not supported: GL_EXT_texture_shadow_lod
Ryu:334: error: 'out_attr0_w' : undeclared identifier
```

gdkchan:
> This error appears when compiling GLSL with shaderc:
> 
> ```
> 00:00:44.132 |E| .NET ThreadPool Worker Gpu GlslToSpirv: Shader compilation error: CompilationError Ryu:7: warning: '#extension' : extension not supported: GL_EXT_texture_shadow_lod
> Ryu:334: error: 'out_attr0_w' : undeclared identifier
> ```

This is fixed now. I actually had fixed it already on #4565. When vector transform feedback outputs were used, it would not declare the remaining components that are not written into the TFB. Although fixing this revealed another issue, the compiler indeed doesn't like when the vertex output and fragment inputs don't match exactly, so I had to use vector inputs on the fragment shader too to make them match.

Might be worth re-testing Vulkan too.

riperiperi:
image
Seems to work on OpenGL now. I also tested Vulkan and that still looks good while performing a lot better.

riperiperi (approved):

marysaka (approved):

PR #4711, opened 2023-04-23
GPU: Pre-emptively flush textures that are flushed often (to imported memory when available)

riperiperi:
Games occasionally read back data from the GPU to use on the CPU, usually when that information would be a lot harder to calculate on the CPU (needs expensive raycast, rasterization). Most games tend to do this by copying data to a buffer, but some games are less well behaved and do this using textures, usually with a linear memory layout. For us, this can also happen accidentally, as textures can be close in proximity to other data that's modified/read often.

A large offender is The Legend of Zelda: Breath of the Wild, which reads back all sorts of things every frame from linear textures. Foot placements on terrain, water level at an object's/link's position, even some information used to colour underwater terrain and used to populate grass (in v1.6.0). The game basically breaks if you don't do it properly.

Currently, flushing texture data on demand is done by queuing up a copy to an intermediate buffer on the GPU, waiting on that copy, then reading that data back. The data that's read back is then copied again into guest memory, in the correct layout.

Now, it determines whether a texture is being flushed often, and performs this copy directly to a buffer that is mapped on CPU accessible memory. This skips the additional wait for the copy to be submitted+completed (other draws may be in progress), which can save a lot of time. On AMD GPUs, this wait happened on the same Vulkan queue as the main rendering, so the GPU scheduler couldn't interrupt longer work to get a faster result, making this wait even longer.

When the texture layout is linear, it's even able to skip the CPU copy and perform the copy directly on the GPU, by importing the memory directly as a buffer and copying to that instead of a GPU managed buffer that needs another copy to guest memory. This has other benefits (flush on write sync _could_ be relaxed) though that's a lot harder to take advantage of.

To do these things, both a method to copy from texture to buffer (with optional stride) and to create a buffer that always has persistently mapped storage were added. Texture to buffer with stride is only required when the memory is host imported, so it is not implemented on OpenGL.

Host memory imports are tracked by a range list that attempts to reuse existing allocations where possible. This reduces allocations, but also might help if a GPU doesn't like importing the same pages multiple times. I haven't found this last thing to be an issue, though.

Reserving host memory skips the multithreading layer and happens immediately. Host memory must be reserved before it is imported, and reserving host memory must be followed by some kind of use (that will decrement the use count on dispose).

## Platforms Tested:
- NVIDIA GTX 1070 Windows
  - Works properly.
- AMD Windows (steam deck)
  - Works properly.
- AMD RADV Linux (steam deck)
  - Host importing shared memory fails gracefully in host mapped mode. Emulator uses persistent buffer and copies out of it instead.
- M1 MoltenVK (1.2.3) MacOS
  - Works properly in Hypervisor and JIT.

## Notes

- Host imported memory does not function on Linux+AMD, as it seems to fail importing any pointer into shared memory. You _can_ use host imported memory in Software Memory mode, but you might have bigger problems at that point.
  - Linux+AMD will still use the new persistent buffer flush, so its extra cost just boils down to a data copy.
  - This might work if `mremap` were used instead of shared memory. Testing/changing that is outside of the scope of this PR though.
- Right now, syncs created on texture readbacks that do this copy are _hard_, meaning that they will always wait on the most recent copy, regardless of whether a guest syncpoint has appeared and been waited on after it. This is also an existing problem for buffers, and could cause more sync waiting than necessary.
  - This could be mitigated in future by only waiting for the writes before nearest guest syncpoint, but likely only for read only accesses, as flush-on-write needs to be tight to avoid data loss when more data is bound than necessary. I might try this out in future.
- 1x1 non-linear textures are sometimes read back (xenoblade DE). These are still compatible with host import, though, as with one pixel there can be no reordering of data.
- Texture flush from GPU can't use this right now. This rules out DMA copies, though they could be improved with buffer copies instead of syncing on GPU.
  - This could also be done by making these cases copy to a destination texture rather than flushing on copy, then the destination texture could handle flush-on-read.

I'd recommend testing this across a bunch of games and setups to ensure nothing breaks.


riperiperi:
Skyward Sword uses texture readback to control whether the sun is obscured by something, so it can dim/hide the lens flare. This is read even when not looking at the sun. The texture is pretty large (1920x1080 RGBA8), so not having to do a CPU copy is really helpful.

Master:
image

PR:
image

Ryujinx.Graphics.GAL/ITexture.cs
```diff
@@ -17,6 +17,7 @@ public interface ITexture
 
         PinnedSpan<byte> GetData();
         PinnedSpan<byte> GetData(int layer, int level);
+        void GetData(BufferRange range, int layer, int level, int stride);
```
gdkchan, line 20: Shouldn't that be called `CopyTo`? Since it copies from the texture to a buffer. `Get` gives the idea that it must return something, which it doesn't.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -57,6 +57,8 @@ class TextureGroup : IDisposable
         /// </summary>
         public bool HasCopyDependencies { get; set; }
 
+        public bool HasFlushBuffer => _flushBuffer != BufferHandle.Null;
```
gdkchan, line 60: Missing XML docs.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
```diff
@@ -565,6 +582,58 @@ public bool FlushModified(Texture texture, bool tracked)
...
+            }
+
+            int sliceStart = handle.BaseSlice;
+            int sliceEnd = sliceStart + handle.SliceCount;
+
+            for (int i = sliceStart; i < sliceEnd; i++)
+            {
+                (int layer, int level) = GetLayerLevelForView(i);
+
+                Storage.GetFlushTexture().GetData(new BufferRange(_flushBuffer, _allOffsets[i], 1), layer, level, _flushBufferImported ? Storage.Info.Stride : 0);
```
gdkchan, line 633: Why is the buffer size always 1 here?

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -241,6 +292,13 @@ public void SignalModifying(bool bound, GpuContext context)
         {
             SignalModified(context);
 
+            if (!bound && _syncActionRegistered && NextSyncCopies())
+            {
+                // On unbind, textures that flush often should immediately create sync so their result can be obtained as soon as possible.
+
+                context.CreateHostSyncIfNeeded(false, false, true);
```
gdkchan, line 299: Maybe a flags enum would work better for that method, since it has multiple bool parameters now. It is hard to tell at a first glance what each bool value means on the call site. Another option would be using explicit parameter names on the call site to specify what each one is.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -296,15 +364,40 @@ public void ClearActionRegistered()
             Interlocked.Exchange(ref _actionRegistered, 0);
         }
 
+        /// <summary>
+        /// Action to perform before a sync number is registered after modification.
+        /// This action will copy the texture data to the flush buffer if this texture
+        /// flushes often enough, which is determined by the flush balance.
+        /// </summary>
+        /// <inheritdoc/>
```
gdkchan, line 372: Does `inheritdoc` include the additional XML docs from the interface in addition to the summary that you provided here?

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -296,15 +364,40 @@ public void ClearActionRegistered()
...
+        public void SyncPreAction(bool syncpoint)
+        {
+            if (syncpoint || NextSyncCopies())
+            {
+                // TODO: only copy if the copy has not been performed since the last modification
```
gdkchan, line 377:
```suggestion
                // TODO: Only copy if the copy has not been performed since the last modification.
```
Does that need to be done before merge?

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
```diff
@@ -1,8 +1,11 @@
-﻿using Ryujinx.Cpu.Tracking;
+﻿using Ryujinx.Common.Logging;
+using Ryujinx.Cpu.Tracking;
+using Ryujinx.Graphics.Gpu.Synchronization;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Timers;
```
gdkchan, line 8: What is that for? It doesn't look like this class is using any timer.

Ryujinx.Graphics.Gpu/Synchronization/ISyncActionHandler.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    interface ISyncActionHandler
+    {
+        /// <summary>
+        /// Action to be performed when some synchronizing action is reached after modification.
+        /// Generally used to register read/write tracking to flush resources from GPU when their memory is used.
+        /// </summary>
+        /// <param name="syncpoint">True if the action is a guest syncpoint</param>
+        /// <returns>True if the action is to be removed, false otherwise</returns>
+        public bool SyncAction(bool syncpoint);
```
gdkchan, line 14:
```suggestion
        bool SyncAction(bool syncpoint);
```

Ryujinx.Graphics.Gpu/Synchronization/ISyncActionHandler.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        /// <summary>
+        /// Action to be performed immediately before sync is created.
+        /// </summary>
+        /// <param name="syncpoint">True if the action is a guest syncpoint</param>
+        public void SyncPreAction(bool syncpoint) { }
```
gdkchan, line 20:
```suggestion
        void SyncPreAction(bool syncpoint) { }
```

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -287,6 +287,23 @@ public unsafe PinnedSpan<byte> GetData(int layer, int level)
...
+            if (stride != 0)
+            {
+                throw new NotSupportedException("Stride conversion for texture copy to buffer not supported.");
+            }
+
+            GL.BindBuffer(BufferTarget.PixelPackBuffer, range.Handle.ToInt32());
+
+            int offset = WriteToPbo2D(range.Offset, layer, level);
+
+            if (offset != 0)
```
gdkchan, line 301: Should it not check if the format is a depth or stencil format before throwing that? Assuming that the problem is just with those formats.

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -287,6 +287,23 @@ public unsafe PinnedSpan<byte> GetData(int layer, int level)
             }
         }
 
+        public void GetData(BufferRange range, int layer, int level, int stride)
+        {
+            if (stride != 0)
```
gdkchan, line 292: It could check if the stride is exactly Width * BytesPerPixel which should be fine too I believe?

Ryujinx.Graphics.OpenGL/PersistentBuffers.cs
```diff
@@ -13,13 +14,39 @@ class PersistentBuffers : IDisposable
...
+            if (_maps.ContainsKey(handle))
+            {
+                GL.BindBuffer(BufferTarget.CopyWriteBuffer, handle.ToInt32());
+                GL.UnmapBuffer(BufferTarget.CopyWriteBuffer);
+
+                _maps.Remove(handle);
+            }
+        }
+
+        public bool TryGetPersist(BufferHandle handle, out IntPtr ptr)
```
gdkchan, line 46: Should it not be called `TryGetPersistent`? Although I think just `TryGet` would be fine too since the class name makes it clear that it's a persistent buffer.

Ryujinx.Graphics.Vulkan/BufferManager.cs
```diff
@@ -75,6 +115,32 @@ public BufferHandle CreateWithHandle(VulkanRenderer gd, int size, out BufferHold
...
+
+            if (gd.Capabilities.SupportsIndirectParameters)
+            {
+                usage |= BufferUsageFlags.IndirectBufferBit;
+            }
+
+            var bufferCreateInfo = new BufferCreateInfo()
+            {
+                SType = StructureType.BufferCreateInfo,
+                Size = (ulong)4096,
```
gdkchan, line 130:
```suggestion
                Size = 4096UL,
```
Maybe it's better to add a constant to make clear what the value means?

Ryujinx.Graphics.Vulkan/HostMemoryAllocator.cs
```diff
@@ -0,0 +1,188 @@
...
+                {
+                    HostMemoryAllocation existing = allocations[i];
+
+                    if (start >= existing.Start && end <= existing.End)
+                    {
+                        return (existing.Allocation, start - existing.Start);
+                    }
+                }
+
+                throw new InvalidOperationException($"No host allocation was prepared for requested range {pointer:x16}:{size:x16}.");
```
gdkchan, line 165: Might be worth adding a `0x` prefix to make clear they are hex values.

Ryujinx.Graphics.Vulkan/HostMemoryAllocator.cs
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+
+        public unsafe void Free(DeviceMemory memory, ulong offset, ulong size)
```
gdkchan, line 169: Does that need the `unsafe` due to `FreeMemory`? You might be able to get rid of it using a span overload.

Ryujinx.Graphics.Vulkan/Shaders/ConvertD32S8ToD24S8ShaderSource.comp
```diff
@@ -0,0 +1,57 @@
...
+
+    // Finally, get the starting offset. Make sure to count extra copies.
+
+    int startCopy = allInvocationCopies * index + min(copiesRequired % invocations, index);
+
+    int srcOffset = startCopy * 2;
+    int dstOffset = dstStartOffset + startCopy;
+
+    // Perform the conversion for this region.
+    for (int i=0; i<copyCount; i++) {
```
gdkchan, line 49:
```suggestion
    for (int i = 0; i < copyCount; i++)
    {
```

Ryujinx.Graphics.Vulkan/Shaders/ConvertD32S8ToD24S8ShaderSource.comp
```diff
@@ -0,0 +1,57 @@
...
+
+    int copiesRequired = pixelCount;
+
+    // Find the copies that this invocation should perform.
+    
+    // - Copies that all invocations perform.
+    int allInvocationCopies = copiesRequired / invocations;
+
+    // - Extra remainder copy that this invocation performs.
+    int index = int(gl_LocalInvocationID.x);
```
gdkchan, line 36: Shouldn't it ensure that the index is within bounds? Size the local size is always 64 which might be more than needed.

Ryujinx.Graphics.GAL/IRenderer.cs
```diff
@@ -21,11 +21,14 @@ BufferHandle CreateBuffer(int size)
...
             return CreateBuffer(size, BufferHandle.Null);
         }
+        BufferHandle CreateBuffer(nint pointer, int size);
+        BufferHandle CreateBuffer(int size, BufferAccess access);
 
         IProgram CreateProgram(ShaderSource[] shaders, ShaderInfo info);
 
         ISampler CreateSampler(SamplerCreateInfo info);
         ITexture CreateTexture(TextureCreateInfo info, float scale);
+        bool PrepareHostMapping(nint address, ulong size);
```
gdkchan, line 31: Why is that method needed? It seems that the GPU emulator could just check if the capability (`SupportsHostImportedMemory`) is supported and call the `CreateBuffer` overload that takes a pointer. If for some reason the import failed, it could just return `BufferHandle.Null`, in which case the GPU emulator would call the regular `CreateBuffer` method with `BufferAccess.FlushPersistent` flag.

riperiperi: This would force the multuthreading layer to flush all commands.

gdkchan: I think I get it now. Essentially `PrepareHostMapping` can be called from any thread, but `CreateBuffer` can't. So it can just call `PrepareHostMapping` directly instead of putting it on the queue.

riperiperi:
I changed the SyncManager to only lock the handle when it's about to wait, so that the lock isn't held during the time it interrupts the threaded renderer. This should avoid some random deadlocks that might have also happened in master in some circumstances.

Ryujinx.Graphics.Vulkan/HostMemoryAllocator.cs
riperiperi: I can, but sending over an empty span of allocation callbacks does seem a bit weird.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
riperiperi: More visual studio autocompletes

Ryujinx.Graphics.Vulkan/BufferManager.cs
riperiperi: This is more just to see what the requirements are for a buffer, so there's not much meaning for the size.

Ryujinx.Graphics.Vulkan/Shaders/ConvertD32S8ToD24S8ShaderSource.comp
riperiperi: The math here is a bit funny. When the total count is less than the local size, `allInvocationCopies` will be 0, meaning that the only copies that are performed are the "remainder" ones, and `copyCount` will just be 0 for a bunch of them. The stride conversion shader does the same thing.

Ryujinx.Graphics.Vulkan/HostMemoryAllocator.cs
gdkchan: We're already doing that in several other places. FWIW those overloads are a bit weird since I don't think you can pass more than one allocation callback, so passing a span there doesn't make sense. I think ideally Silk.NET would have a overload where this parameter wouldn't exist since it's supposed to be optional.

Bjorn29512:
Tested about 20+ titles :

Looked for any performance changes compared to master or any visual oddities that can be reliably spotted in 15 mins of playtime for each of the following titles 

Key observations :

- Performance improvements compared to master in several titles :

  - The Legend of Zelda: Breath of The Wild : Performance improved in this PR compared to master by 5 frames avg and the game is              finally able to hit 60 fps in Korok Forest region and other spots more consistently on my setup.
  
     Master:
     image
     
     PR:
     image

  - The Legend of Zelda: Skyward Sword HD : Epic 30+ fps improvement compared to master.

     Master: 
     image
     
     PR: 
     image
  
  - Xenoblade Chronicles: Definitive Edition : +7-8 fps compared to master.
  
    Master:
    image

    PR: 
    image

  - Xenoblade Chronicles 2: Xeno games keep eating good for no reason, another improvement of around 5fps avg.
    
    Master:
    image
    
    PR: 
    image
    
- Other tested titles where no Change/Regression/Issue was noticed : 
  - Bayonetta 3
  - FE: Three Hopes
  - FE: Engage
  - FE: Three Houses
  - Mario + Rabbids: Sparks of Hope
  - Metroid Dread 
  - Metroid Prime Remastered 
  - Pokemon Violet
  - Pokemon Sword
  - SMT 5
  - Super Mario Odyssey
  - TloZ: Link's Awakening
  - Xenoblade Chronicles 3
  - Astral Chain
  - Yo Kai Watch 4
  - Monster Hunter Generations Ultimate
  - Marvel Ultimate Alliance 3
  - Kirby and the Forgotten Land

Nerboruto:
an option to put all textures on shared system gpu memory and skip copies wouldn't be feasible? the bandwidth of the pci-ex gen3 is 12gb/s... this would also avoid problem of out of vram on 2gb gpu.
Wouldn't that also benefit to the cpu texture decoder?

riperiperi:
I don’t know what that proposal has to do with this PR, but here’s why that’s not a great idea.

Putting ALL astc textures on system memory could result in nearly every texture in a game being loaded in from system memory, uncompressed. They’re upwards of 4x the size and desktop GPUs aren’t designed to be used like this. I’ve done experiments with mapping regular textures on system memory (not even astc) and the impact was huge, not really acceptable. Remember that 12gb/s is raw copy throughput, it doesn’t account for that data being accessed like a texture.

A better idea is selectively moving the least used textures to and from system memory as it approaches the max. That also doesn’t really have much to do with this PR.

Nerboruto:
> I don’t know what that proposal has to do with this PR, but here’s why that’s not a great idea.
> 
> Putting ALL astc textures on system memory could result in nearly every texture in a game being loaded in from system memory, uncompressed. They’re upwards of 4x the size and desktop GPUs aren’t designed to be used like this. I’ve done experiments with mapping regular textures on system memory (not even astc) and the impact was huge, not really acceptable. Remember that 12gb/s is raw copy throughput, it doesn’t account for that data being accessed like a texture.
> 
> A better idea is selectively moving the least used textures to and from system memory as it approaches the max. That also doesn’t really have much to do with this PR.

yes sorry, it's not inherent to this pr... I'm puzzling over this thing.

amayra:
> Host imported memory does not function on Linux+AMD, as it seems to fail importing any pointer into shared memory. 

is this Security Measure in Unix-like operating system or mesa/AMD driver to blame for this ?

does it work in under linux+nvidia

riperiperi:
> > Host imported memory does not function on Linux+AMD, as it seems to fail importing any pointer into shared memory.
> 
> is this Security Measure or limitation in Unix-like operating system or mesa/AMD driver to blame for this ?
> 
> does it work in under linux+nvidia

I'm not sure, I didn't test it. Bylaws has reported a similar restriction on Android+Qualcomm that isn't present when using mremap, so maybe it is linux related rather than driver.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
riperiperi: Yeah, I just use it to keep the existing descriptions for arguments.

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
riperiperi: The stride also needs to be aligned up to 4 bytes, due to the pixel pack options. Adding this check won't do anything right now as stride will always be 0 when copying to a non-host-imported buffer, but I can add it anyways.

Ryujinx.Graphics.OpenGL/Image/TextureView.cs
riperiperi: This error is caused by a copy not being able to place the data at offset 0, which can only happen for depth stencil textures... But it's not really specific to them, it's more of an error that the data couldn't be put in the proper location.

I'll add an exception to stop it from performing the copy in the first place, rather than reacting to the offset.

Ryujinx.Graphics.Vulkan/BufferManager.cs
riperiperi: I've changed this to the system page size.

Ryujinx.Graphics.GAL/ITexture.cs
riperiperi: Renamed to CopyTo.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
riperiperi: Added docs.

Ryujinx.Graphics.Gpu/Image/TextureGroup.cs
riperiperi: Just lazy, should be `_sliceSizes[level]`.

Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
riperiperi: I've added a flags enum.

src/Ryujinx.Graphics.Gpu/Synchronization/HostSyncFlags.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
```
gdkchan, line 1:
```suggestion
﻿using System;
﻿
```

gdkchan (approved):
lgtm other than the nit, thanks.

riperiperi:
I've improved the sync rules so that they can wait for a further back syncpoint when possible, and also avoid some potential errors that might have been introduced with regular (not through a buffer) texture flushing. This was all part of the original plan, I just never got around to getting it working.

marysaka (approved):

PR #4717, opened 2023-04-24
Implement DMA texture copy component shuffle

gdkchan:
This implements the remap functionality that the DMA engine has. It is more or less like a shuffle, and can be used to re-order components on the pixels of a texture for example. The OpenGL driver uses it to perform BGRA -> RGBA texture data conversion in a few cases.

Fixes inverted red and blue on 20XX and Dragon Quest Builders, maybe other OpenGL games too.

**Dragon Quest Builders:**
Before:
image
After:
image
**20XX (title screen):**
Before:
image
After:
image
**20XX (in-game):**
Before:
image
After:
image

NVN uses remap in a few cases too, but not as a shuffle, it only enables it to increase the copy size. In any case, it might be worth to test NVN games just to ensure there's no regression.

Testing is welcome.

### DIFF
--- a/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -6,6 +6,7 @@ using Ryujinx.Graphics.Texture;
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
 
 namespace Ryujinx.Graphics.Gpu.Engine.Dma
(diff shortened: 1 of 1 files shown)


Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -32,6 +33,69 @@ private enum CopyFlags
...
+                RegionX = regionX;
+                RegionY = regionY;
+                BaseOffset = baseOffset;
+                Bpp = bpp;
+                Linear = linear;
+                Calculator = calculator;
+            }
+        }
+
+        [StructLayout(LayoutKind.Sequential, Size = 3)]
```
marysaka, line 91: Shouldn't the align explicitly be set to 1?

marysaka (approved):

riperiperi (approved):
Nice one.

PR #4720, opened 2023-04-25
ModLoader: Fix case sensitivy issues

TSRBerry:
This PR fixes an issue where the mod subdirectories had to be lowercase to be recognized by Ryujinx.

Before, this wouldn't be loaded without renaming `exeFS` to `exefs`.
image

(Screenshot by @piplup55)

These changes also allow for more nesting of the mod directories, but that's just a side-effect of the way it works now.

Aside from that some refactoring for the `ModLoader` class was done.

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1551,8 +1551,8 @@ namespace Ryujinx.Ava.UI.ViewModels
         {
             if (SelectedApplication != null)
             {
-                string modsBasePath  = VirtualFileSystem.ModLoader.GetModsBasePath();
-                string titleModsPath = VirtualFileSystem.ModLoader.GetTitleDir(modsBasePath, SelectedApplication.TitleId);
+                string modsBasePath  = ModLoader.GetModsBasePath();
+                string titleModsPath = ModLoader.GetTitleDir(modsBasePath, SelectedApplication.TitleId);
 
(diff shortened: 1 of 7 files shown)


marysaka (approved):

gdkchan:
Is Atmosphere case sensitive? I think we should make sure the emulator behaves the same as the console to avoid surprises when people tests mods on emulators.

gdkchan:
If that's the case, we should fix case sensitivity issues related to romfs mods them. I had a few reports of people changing the casing on file names and the mod not working (like for example, writing `Filename.bin` when the game has `filename.bin` would not work).

TSRBerry:
Did I miss something in my PR? I didn't adjust it for exefs specifically, but for all affected instances in ModLoader.

TSRBerry:
Also Windows systems are entirely unaffected by case sensitivity issues, since Windows filesystems are case insensitive by default.

gdkchan:
> Did I miss something in my PR? I didn't adjust it for exefs specifically, but for all affected instances in ModLoader.

I only took a quick look at the code, and was going from the PR description (which only mentions the `exeFS` case), so if romfs is already handled properly too, then nevermind the comment.

> Also Windows systems are entirely unaffected by case sensitivity issues, since Windows filesystems are case insensitive by default.

It does affect Windows, NTFS is case sensitive. At least, the mod issue that I described above was happening on Windows.

TSRBerry:
>  I probably need to take a look at VFS to figure out how we handle that currently.

Seems like something we probably want to change on the LibHac side.

AcK77 (approved):

PR #4722, opened 2023-04-25
Add missing check for thread termination on ArbitrateLock

gdkchan:
Because `ArbitrateLock` was not checking `TerminationRequested` for the current thread, it would wait for the lock even if the thread was already terminated, which would cause the thread to never exit (since the thread that owns the lock is most likely already terminated as well and has exited).

This should fix one of the instances of the emulator freezing when you stop emulation or close, it will not fix *all* of them since they are caused by different issues. Also it's hard to test this since this specific only happens if you close exactly when a guest thread call `ArbitrateLock`, but before the function gets a chance to run.

I also replaced all usages of the pattern `ShallBeTerminated || SchedFlags == ThreadSchedState.TerminationPending` with the `TerminationRequested ` property, and did other minor improvements that should not affect functionality.

### DIFF
--- a/Ryujinx.HLE/HOS/Kernel/Threading/KAddressArbiter.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Threading/KAddressArbiter.cs
@@ -31,6 +31,14 @@ namespace Ryujinx.HLE.HOS.Kernel.Threading
 
             _context.CriticalSection.Enter();
 
+            if (currentThread.ShallBeTerminated ||
+                currentThread.SchedFlags == ThreadSchedState.TerminationPending)
+            {
+                _context.CriticalSection.Leave();
+
+                return KernelResult.ThreadTerminating;
+            }
+
             currentThread.SignaledObj   = null;
             currentThread.ObjSyncResult = Result.Success;
 
--- a/Ryujinx.HLE/HOS/Kernel/Threading/KAddressArbiter.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Threading/KAddressArbiter.cs
@@ -31,8 +31,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Threading
 
             _context.CriticalSection.Enter();
 
-            if (currentThread.ShallBeTerminated ||
-                currentThread.SchedFlags == ThreadSchedState.TerminationPending)
+            if (currentThread.TerminationRequested)
             {
                 _context.CriticalSection.Leave();
 
@@ -122,8 +121,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Threading
             currentThread.SignaledObj   = null;
             currentThread.ObjSyncResult = KernelResult.TimedOut;
 
-            if (currentThread.ShallBeTerminated ||
-                currentThread.SchedFlags == ThreadSchedState.TerminationPending)
+            if (currentThread.TerminationRequested)
             {
                 _context.CriticalSection.Leave();
 
@@ -288,8 +286,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Threading
 
             _context.CriticalSection.Enter();
 
-            if (currentThread.ShallBeTerminated ||
-                currentThread.SchedFlags == ThreadSchedState.TerminationPending)
+            if (currentThread.TerminationRequested)
             {
                 _context.CriticalSection.Leave();
 
@@ -359,8 +356,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Threading
 
             _context.CriticalSection.Enter();
 
-            if (currentThread.ShallBeTerminated ||
-                currentThread.SchedFlags == ThreadSchedState.TerminationPending)
+            if (currentThread.TerminationRequested)
             {
                 _context.CriticalSection.Leave();
 
--- a/Ryujinx.HLE/HOS/Kernel/Ipc/KServerSession.cs
+++ b/Ryujinx.HLE/HOS/Kernel/Ipc/KServerSession.cs
@@ -188,8 +188,7 @@ namespace Ryujinx.HLE.HOS.Kernel.Ipc
 
             if (request.AsyncEvent == null)
             {
-                if (request.ClientThread.ShallBeTerminated ||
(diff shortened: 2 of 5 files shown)


riperiperi (approved):
Seems reasonable, won’t be able to test myself for a bit tho.

Ryujinx.HLE/HOS/Kernel/Threading/KAddressArbiter.cs
```diff
@@ -31,6 +31,13 @@ public Result ArbitrateLock(int ownerHandle, ulong mutexAddress, int requesterHa
 
             _context.CriticalSection.Enter();
 
+            if (currentThread.TerminationRequested)
```
TSRBerry, line 34: Shouldn't this check be done, before entering the critical section?
As far as I understand the thread would already be blocked at this point since it wouldn't be able to enter the critical section. Where am I wrong? :D

gdkchan: Doing it before entering the critical section would be incorrect.
It is easier to understand with a super simplified version of the code. Let's consider a simplified version of the code before this change:
```cs
Result ArbitrateLock()
{
    KThread currentThread = GetCurrentThread();

    lock (CriticalSectionLock)
    {
        currentThread.SchedFlags = ThreadSchedFlags.Paused;
        currentThread.SchedulerWaitEvent.Wait();
    }

    return Result.Success;
}

void TerminateThread(KThread threadToTerminate)
{
    lock (CriticalSectionLock)
    {
        threadToTerminate.TerminationRequested = true;

        if (threadToTerminate.SchedFlags == ThreadSchedFlags.Paused)
        {
            threadToTerminate.SchedulerWaitEvent.Signal();
        }
    }
}
```
Let's say we have two threads, A and B. Thread A is a guest thread that is calling `ArbitrateLock`, while thread B is a thread responsible for terminating all guest thread, and it calls `TerminateThread` passing each guest thread as parameter to do so.

So let's analyse this code. First we can see that before the thread waits for the event to be signalled on `ArbitrateLock`, it sets `SchedFlags` to `ThreadSchedFlags.Paused`. This indicates that the thread is suspended, in wait state, and thus the event can be signalled to wake it up. Now if we look at the `TerminateThread` method, first it sets `TerminationRequested` to true, which indicates *the thread should terminate itself*. Then, when `SchedFlags` of the thread being terminated is `ThreadSchedFlags.Paused`, which means that the thread is suspended, it will simply signal the event to wake it up.

The problem here is that, since `ArbitrateLock` did not check `TerminationRequested`, it was possible that, if the thread was terminated right when `ArbitrateLock` was called, and before `currentThread.SchedFlags = ThreadSchedFlags.Paused;` runs, it would wait for  `SchedulerWaitEvent` forever, since nothing else would signal it. The path that is supposed to signal that event on `TerminateThread` would not be taken because when it was called, the thread was not suspended *yet*.

So let's take a look at the fix you're proposing here:
```cs
void ArbitrateLock()
{
    KThread currentThread = GetCurrentThread();

    if (currentThread.TerminationRequested)
    {
        return Result.Terminating;
    }

    lock (CriticalSectionLock)
    {
        currentThread.SchedFlags = ThreadSchedFlags.Paused;
        currentThread.SchedulerWaitEvent.Wait();
    }

    return Result.Success;
}
```
It only partially fixes the issue. Since we have two different threads, it is still possible that thread B will see a state where `SchedFlags` is not `ThreadSchedFlags.Paused` yet, so it will just set `TerminationRequested` to true. But if thread A is just after the `if (currentThread.TerminationRequested)` check and before the `currentThread.SchedFlags = ThreadSchedFlags.Paused;` line, the same problem can still occur. It will wait forever since thread B did not signal the event and the thread was not suspended yet at that point.

Now if we move the check inside the lock, we solve the problem:
```cs
void ArbitrateLock()
{
    KThread currentThread = GetCurrentThread();

    lock (CriticalSectionLock)
    {
        if (currentThread.TerminationRequested)
        {
            return Result.Terminating;
        }

        currentThread.SchedFlags = ThreadSchedFlags.Paused;
        currentThread.SchedulerWaitEvent.Wait();
    }

    return Result.Success;
}
```
The lock ensures that only one thread will run at a time. So it's not possible for the code inside `ArbitrateLock` and `TerminateThread` to run at the same time.

Now for the `CriticalSection` `Enter` and `Leave` methods, which might have been the source of the confusion: They act like a regular lock, for the most part. `Enter` acquires the lock, and blocks if another thread already owns the lock. `Leave` releases the lock and allow the next waiting thread to run. However, one thing special about the `Leave` implementation here is that is also handles the guest context switching. It will pass control to the next thread that is supposed to run on the next core, and wait until its time for the current thread to run again. You can see most of that logic on: https://github.com/Ryujinx/Ryujinx/blob/5131b71437b36f9b876046a2fddd5465652e0f10/Ryujinx.HLE/HOS/Kernel/Threading/KScheduler.cs#L276

So this is the method we need to worry about. The `Leave` method might block for a long time, and in the case of this bug, it would block forever waiting for `SchedulerWaitEvent` to be signalled, something that would never happen since all the other guest threads had already exited at that point, and the terminate method would not signal it due to the reasons explained earlier.

TSRBerry: Thank you very much for this explanation! It's much clearer now, however I have one question left.

If `currentThread.SchedulerWaitEvent.Wait()` is executed isn't the thread still holding the `CriticalSectionLock` lock? Or does the `Wait()` call release it until the thread receives the signal to wake up?

gdkchan: That was an oversight from my part on this simplified code. The actual wait happens when `KCriticalSection.Leave` is called, more specifically on the `Schedule` method that I linked above. At that point, the lock has already been released. In any case, I edited the simplified code above to fix that mistake.

TSRBerry: I see, thank you very much this helped a lot!

PR #4723, opened 2023-04-25
Fix some invalid blits involving depth textures

gdkchan:
This is a change to prevent some cases of invalid blits involving depth-stencil textures. Both OpenGL and Vulkan only allows blits with depth textures with the *exact same* format. And of course, they don't allow blits between depth and color textures.

From the Vulkan spec:
>  If either of srcImage or dstImage was created with a depth/stencil format, the other must have exactly the same format

From the OpenGL spec:
> GL_INVALID_OPERATION is generated if mask contains GL_DEPTH_BUFFER_BIT or GL_STENCIL_BUFFER_BIT and the source and destination depth and stencil formats do not match. 

So first, this disables the additional depth aliasing rules when searching for the destination texture in the cache. If the source texture is a depth texture, destination *must* also be depth. If it's color, it *must* also be color. So we should not allow any aliasing of color and depth formats when looking for the destination texture. There's also an additional check where it will only allow the depth aliasing if the source and destination texture formats are the same, due to what I mentioned above: The host API does not support blits between different depth formats.

I originally made this change in an attempt to fix the reported error on Twelve Minutes on macOS, but I can't repro it anymore on `master`, so maybe the root cause of that problem was fixed already. Still, I think it's a good change since we should not allow invalid blits in any case.


marysaka (approved):

riperiperi (approved):
Makes sense, I think.

PR #4670, opened 2023-04-11
Add workflow to automatically check code style issues for PRs

TSRBerry:
This PR was originally planned to contain a lot more things, but formatting already took a while and I didn't expect our codebase to report so many issues (via dotnet-format).

A new workflow will now be triggered for every PR which will run dotnet-format to check for style issues.
If the workflow doesn't succeed the author of the PR will need to make some style changes, before we are able to merge the PR. The issues dotnet-format found will be uploaded as an artifact but also printed to the log of the workflow run.

This allows reviewers to focus more on actual code review instead of adding a lot of comments about minor styling issues. This will hopefully make reviews a little bit easier and authors still get feedback about code styling with the workflow.

I was able to automate silencing some of the issues dotnet-format reported with this [script](https://gist.github.com/TSRBerry/5460e3eaf8ca259211e08742ee10b08d).

---

Since this PR required a few other code changes to make dotnet-format happy, feel free to suggest improvements/changes to those as well.

Feedback about this approach would also be appreciated!

---

List of code style/quality rules we need to discuss:

- [x] [IDE0044](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0044): Add readonly modifier
- [ ] [IDE0051](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0051): Remove unused private member
- [ ] [IDE0052](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0052): Remove unread private member
- [ ] [IDE0059](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0059): Remove unnecessary value assignment
- [ ] [IDE0060](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0060): Remove unused parameter
- [x] [IDE0066](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0066): Use switch expression
- [x] [IDE1006](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules#rule-id-ide1006-naming-rule-violation):  Naming rule violation

- [ ] [CS0169](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0169): The private field 'class member' is never used
[...]

PR #4725, opened 2023-04-26
Move Ryujinx projects to a new src directory

TSRBerry:
This PR contains only the move related commits which were previously also part of #4670. 

All Ryujinx projects including the solution file were moved into a new `src` directory, since @gdkchan wanted to do that for a while now.

I'm actually not quite sure if the solution file should stay in the repository root, since most IDEs will treat the directory with the solution file as the root directory. This would make accessing `.github` or `distribution` harder than before.

---

Make sure you **aren't** squash merging this PR!


gdkchan:
> I'm actually not quite sure if the solution file should stay in the repository root, since most IDEs will treat the directory with the solution file as the root directory. This would make accessing `.github` or `distribution` harder than before.

FWIW other projects have the solution file at the root. I think it's fine if it makes things easier.

gdkchan (approved):
Thanks, I just hope that the rebase will really go smoothly.

TSRBerry:
Kept the solution file (and related files which need to stay at the solution root) in the repository root and moved everything else.

Just requested the review again, since I also needed to adjust the github workflows again.

marysaka (approved):

gdkchan (approved):
lgtm, thanks.

PR #4726, opened 2023-04-26
Draft: Add config option for logging

jardon:
This is a PR I threw together for supporting the configuration of the logging directory.  I am by no means a C# developer so if you think this is garbage, you won't hurt my feelings.  If this is something that would be useful to the project, I can iterate on it to clean it up.

Also, if there's a better way to do this, I am all ears.  I spent a little bit of time working on this because I was playing around with trying to package this differently and dumping logs into the same directory as the binary was becoming problematic.

### DIFF
--- a/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
+++ b/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
+using Microsoft.Extensions.Configuration;
+using System.Runtime.InteropServices;
 
 namespace Ryujinx.Common.Logging
 {
@@ -12,6 +14,17 @@ namespace Ryujinx.Common.Logging
 
         string ILogTarget.Name { get => _name; }
 
+        private static string _config = GetConfigName();
+
+        // Build a config object, using env vars and JSON providers.
+        private static IConfiguration config = new ConfigurationBuilder()
+            .AddJsonFile(_config)
+            .AddEnvironmentVariables()
+            .Build();
+
+        // Get values from the config given their key and their target type.
+        public static Settings settings = config.GetRequiredSection("Settings").Get<Settings>();
+
         public FileLogTarget(string path, string name)
             : this(path, name, FileShare.Read, FileMode.Append)
         { }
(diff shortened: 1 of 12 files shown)


TSRBerry:
> dumping logs into the same directory as the binary was becoming problematic

Related: #4707

Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -12,14 +14,25 @@ public class FileLogTarget : ILogTarget
...
+        private static string _config = GetConfigName();
+
+        // Build a config object, using env vars and JSON providers.
+        private static IConfiguration config = new ConfigurationBuilder()
+            .AddJsonFile(_config)
+            .AddEnvironmentVariables()
+            .Build();
+
+        // Get values from the config given their key and their target type.
+        public static Settings settings = config.GetRequiredSection("Settings").Get<Settings>();
```
WilliamWsyHK, line 26: `public static Settings Settings` (notice the `S` is capital letter in the Property name)

Ryujinx.Common/Settings.cs
```diff
@@ -0,0 +1,4 @@
+public sealed class Settings
+{
+    public required string logDir { get; set; }
```
WilliamWsyHK, line 3: `public required string LogDirectory` (`l` to `L` and I just tried to make it more readable for the property name.)

Ryujinx/linux.json
```diff
@@ -0,0 +1,5 @@
+{
+    "Settings": {
+        "logDir": "/home/jardon/.local/ryujinx"
```
WilliamWsyHK, line 3: Please do not hard-code the user name. Other *nix platform as well.

Ryujinx/windows.json
```diff
@@ -0,0 +1,5 @@
+{
+    "Settings": {
+        "logDir": "C:"
```
WilliamWsyHK, line 3: I suppose it is better not to write to C: drive directly.

WilliamWsyHK (commented):
I believe for the path, we might use the current `logs` folder under the current Ryujinx config folder for this purpose.
If it is portable, `logs` folder should be created under the portable folder.

Ryujinx/linux.json
jardon: This is a draft :) Those were just some values to prove that it worked.

PR #4744, opened 2023-04-29
Signal interrupt event to improve back-end frame presentation

### LINKED ISSUE #3906: (All operating systems) - V-Sync/framerate synchronization not working correctly
#### Overview:
- At the moment, it appears that the presentation of rendered frames is not properly synced to the display's refresh rate.
- Therefore, a game that outputs a stable 60 FPS will still look noticably stuttery, as if dropping frames.

#### Smaller Details: 
- This issue occurs on all tested refresh rates: 59.94/60/120 Hz (internal "ProMotion" display); or 60 Hz (external display)
- It occurs on both tested games (Fast RMX or Super Mario Odyssey). Both were tested in situations where a stable 60 FPS was effectively guaranteed. The amount of (I presume) skipped frames appeared to be very similar

#### Nature of Request:  
- Change
  - if Ryujinx has no "real" support for this at the moment, it could count as Addition

- especially 60 FPS games would benefit a lot, since at the moment they will appear quite choppy.
- Potentially, unstable framerates or 30 FPS titles could also benefit from improved framerate → refresh rate synchronization.

al81-ru:
Fixes #3906

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Engine/GPFifo/GPFifoDevice.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/GPFifo/GPFifoDevice.cs
@@ -252,6 +252,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.GPFifo
         public void Interrupt()
         {
             _interrupt = true;
+            _event.Set();
         }
 
         /// <summary>


gdkchan:
Why is the 8ms wait a problem? The event is signalled as soon as the thread starts receiving commands IIRC.

riperiperi:
There is an issue with this, but it's not the 8ms. The `Interrupt` method in this class is called when there's a frame ready to present, but it currently isn't signalling this event so it is waiting way longer than it should be.

Add `_event.Set();` to the end of `Interrupt()`.

al81-ru:
> Add `_event.Set();` to the end of `Interrupt()`.

You are right, it solves the problem

riperiperi:
You should probably change this PR to just do that one change. Thanks for figuring this out.

al81-ru:
Done. Glad to be able to contribute

gdkchan:
If it fixes the issue #3906 you should add `Fixes #3906` or `Closes #3906` on the PR description, to have it close automatically on merge.

gdkchan (approved):
lgtm, thanks. I assumed that the event was already being signalled on all cases where the thread might have work to do, but I guess not.

riperiperi (approved):
LGTM. Hopefully this should help some weird behaviour when using freesync/gsync and bouts of duplicate frames.

Issue #4598, opened 2023-03-25
[Regression] Link's Awakening bomb texture glitch introduced in 1.1.664


JENOVAAbsolute:
### Description of the issue

API: Vulkan
Resolution Scale: 2x
Post Processing Effect: SMAA Ultra
Upscale: FSR 80.0
All other settings: Default

Visual bug with bomb texture in Link's Awakening. This bug is oddly specific. It's only happened when the miniboss Master Stalfos in Catfish's Maw is present on screen. It has not happened anywhere in the game prior to this. It can happen in all four rooms you fight him in, as well as custom dungeons you create with him in it. It stops happening as soon as he is no longer present.

UPDATE: This is a regression that started with build 1.1.664, discovered by [Bjorn29512](https://github.com/Bjorn29512). Tested on 1.1.663, and there is no issue there.

Image showing the bug is below.
![bomb](https://user-images.githubusercontent.com/60951301/227704079-51d4d7a5-48a3-4636-a5a1-02ea9bb00619.png)

### Reproduction steps

- Load the save file provided in the comment section
- Speak to Dampe
- Choose to Adventure into a dungeon
- Go to the Free tab and enter the only dungeon there
- Enter the room to the right as soon as you spawn
- Place a bunch of bombs once battle starts

### Log file

[Ryujinx_1.1.681_2023-03-25_03-06-42.log](https://github.com/Ryujinx/Ryujinx/files/11068320/Ryujinx_1.1.681_2023-03-25_03-06-42.log)

### OS

Windows 10

### Ryujinx version

1.1.681

### Game version

1.0.1

### CPU

Ryzen 5 5600X

### GPU

Nvidia RTX 3060Ti

### RAM

16Gb.

Bjorn29512:
can you upload a save ?

JENOVAAbsolute:
> can you upload a save ?

Oh Lord. I should have. I completed the dungeon already. I apologize for that.

JENOVAAbsolute:
Thankfully I was able to recreated a dungeon with the miniboss, and the bug happens there too. Save file is attached. You will start in Dampe's shack. The custom dungeon is in the Free tab, and the miniboss will be in the room to the right of the entrance.

[[SAVE].zip](https://github.com/Ryujinx/Ryujinx/files/11069827/SAVE.zip)

Bjorn29512:
[...]

PR #4745, opened 2023-04-30
GPU: Fix errors handling texture remapping

### LINKED ISSUE #4698: [Regression] Xenoblade Chronicles 3 : Screen goes white on initiating the boss fight from build 1.1.664 onwards
### Description of the issue

Build 1.1.664 #4442  introduced a regression in XB 3 where the screen would go white after initiating a boss battle in Keves Castle region.

### Reproduction steps

- MAKE SURE ALL MODS ARE DISABLED 
- Select load game and open the save in slot no 3/ second save in order : Keves Castle region.
- Once you enter the game approach the checkpoint in front and skip the cut scene. 
[0.zip](https://github.com/Ryujinx/Ryujinx/files/11276731/0.zip)

### Log file

[Ryujinx_1.1.663_2023-04-20_00-44-12.log](https://github.com/Ryujinx/Ryujinx/files/11276807/Ryujinx_1.1.663_2023-04-20_00-44-12.log)
[Ryujinx_1.1.664_2023-04-20_00-32-14.log](https://github.com/Ryujinx/Ryujinx/files/11276808/Ryujinx_1.1.664_2023-04-20_00-32-14.log)

### OS

Windows 11

### Ryujinx version

1.1.723

### Game version

1.3.0

### CPU

i5-12400f

### GPU

GTX 1660 Super (driver 531.41)

### RAM

32 GB

### List of applied mods

NONE

### Additional context?

Attachments showcasing the issue : 

1.1.663
image

1.1.664
image

riperiperi:
This PR fixes a few issues introduced by the sparse texture PR #4442, and also some others that might not have been caught previously.

- Fixes an error where a pool entry and memory mapping changing at the same time could cause a texture to rebind its data from the wrong GPU VA (data swaps).
  - This should fix most issues cause by that PR.
- Introduces "remap actions" which can be performed after a gpu mapping has completed. These can be added to when handling the event before the mapping occurs.
- Fixes an error where the texture pool could act on a mapping change before the mapping has actually been changed. ("Unmapped" event happens before change, we need to signal it changed _after_ it completes)
  - "Unmapped" event adds to the partial mapping list, "remap action" signals pool mapping change on all partially mapped textures.
  - When the pool mappings are processed, the textures remove themselves from the partial mapping list if they don't have any unmapped regions.
- Fixes an issue where a texture that became sparsely mapped after creation wouldn't be notified when new pages were mapped.

I also added a `ToString()` method for `MultiRange`, which makes debugging anything to do with memory ranges a bit easier.

Closes #4698.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -120,27 +120,6 @@ namespace Ryujinx.Graphics.Gpu.Image
         }
 
         /// <summary>
-        /// Handles removal of textures written to a memory region being remapped.
-        /// </summary>
-        /// <param name="sender">Sender object</param>
-        /// <param name="e">Event arguments</param>
-        public void MemoryRemappedHandler(object sender, UnmapEventArgs e)
-        {
-            // Any texture that has been unmapped at any point or is partially unmapped
-            // should update their pool references after the remap completes.
-
-            MultiRange unmapped = ((MemoryManager)sender).GetPhysicalRegions(e.Address, e.Size);
-
-            lock (_partiallyMappedTextures)
-            {
-                foreach (var texture in _partiallyMappedTextures)
-                {
-                    texture.UpdatePoolMappings();
-                }
-            }
-        }
-
-        /// <summary>
         /// Determines if a given texture is eligible for upscaling from its info.
         /// </summary>
         /// <param name="info">The texture info to check</param>
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -1171,6 +1171,7 @@ namespace Ryujinx.Graphics.Gpu.Image
                     if (range.GetSubRange(i).Address == MemoryManager.PteUnmapped)
                     {
                         partiallyMapped = true;
+                        break;
                     }
                 }
 
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -82,13 +82,39 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (overlapCount > 0)
             {
-                lock (_partiallyMappedTextures)
+                for (int i = 0; i < overlapCount; i++)
                 {
-                    for (int i = 0; i < overlapCount; i++)
+                    overlaps[i].Unmapped(unmapped);
+                }
+            }
+
+            lock (_partiallyMappedTextures)
+            {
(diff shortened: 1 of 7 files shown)


riperiperi:
Fixes https://github.com/Ryujinx/Ryujinx/issues/4598.

src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
```diff
@@ -1135,6 +1150,43 @@ public void RemoveTextureFromCache(Texture texture)
...
+            lock (_partiallyMappedTextures)
+            {
+                range = memoryManager.GetPhysicalRegions(address, texture.Size);
+                bool partiallyMapped = false;
+
+                for (int i = 0; i < range.Count; i++)
+                {
+                    if (range.GetSubRange(i).Address == MemoryManager.PteUnmapped)
+                    {
+                        partiallyMapped = true;
```
gdkchan, line 1173: This could also break out of the loop here, as a minor optimisation.

src/Ryujinx.Memory/Range/MultiRange.cs
```diff
@@ -319,5 +320,21 @@ public override int GetHashCode()
...
+
+        /// <summary>
+        /// Returns a string summary of the ranges contained in the MultiRange.
+        /// </summary>
+        /// <returns>A string summary of the ranges contained within</returns>
+        public override string ToString()
+        {
+            if (HasSingleRange)
+            {
+                return $"[{_singleRange.Address:x16}:{_singleRange.Size:x8}]";
```
gdkchan, line 332: I'd probably add a `ToString` method to `MemoryRange` and call it here (then we can also easily get a string out of it for debugging).

FWIW I had something like this on my virtual buffer cache PR:
https://github.com/Ryujinx/Ryujinx/blob/6451cb01a8361f30a1644badbcb41b5682df6487/Ryujinx.Memory/Range/MemoryRange.cs#L78
https://github.com/Ryujinx/Ryujinx/blob/6451cb01a8361f30a1644badbcb41b5682df6487/Ryujinx.Memory/Range/MultiRange.cs#L374

riperiperi: I'll use your implementation, address to end address might be clearer.

Bjorn29512:
Fixes #4698

src/Ryujinx.Memory/Range/MultiRange.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
```
gdkchan, line 3: Can be removed now I believe.

gdkchan (approved):
lgtm minus comment, thanks for the quick fix!

marysaka (approved):

GamerzHell9137:
Fixes UI textures in WVSC eBASEBALL: POWER PROS and possibly in the other games in the series

image
image

PR #4777, opened 2023-05-03
Reducing Memory Allocations 202305

jhorv:
**TL;DR:**  this PR reduces LOH memory allocations by 47% in one test scenario

This PR focuses on reducing memory allocations, especially by eliminating large object heap (LOH) allocations caused by copying memory blocks around. The major changes involve around:

- adds memory ownership to `WritableRegion`, so that the memory it wraps can be rented and released when the `WritableRegion` is disposed
- changes some `LayoutConverter` methods to return `IOwnedMemory<byte>` rented from `ByteMemoryPool`, instead of returning newly allocated byte arrays.
- changes `SurfaceFlinger.Parcel` to implement `IDisposable` and to rent memory from `ByteMemoryPool`
- adds `IVirtualMemoryManager.GetReadOnlySequence()`, enabling zero-copy reads of `IVirtualMemory`-managed memory.
    - The alternative (`IVirtualMemoryManager.GetSpan()`) allocates a new copy when the requested memory is discontiguous, but accessing the same memory through a `ReadOnlySequence<byte>` allows direct reads without copying.
- adds `SequenceReaderExtensions` extension methods to read values from a `SequenceReader<byte>`/`ReadOnlySequence<byte>` with minimal copying.
- changes `AudioRenderSystem` and related types (especially `StateUpdater`) to read input from `ReadOnlySequence<byte>` instead of `ReadOnlyMemory<byte>`, which was a major source of allocations/copies.
    - This change alone eliminates 2214 LOH allocations that occur within a span of ~50 profile-run seconds of the test scenario, which is 35% of all LOH allocations that occur during the entire run.

Test Scenario Stats:

|                              | master-250fc51 | This PR   | Delta       | Delta %              |
| ---------------------------- | -------------- | ------------- | ----------- | -------------------- |
| Realtime Run Time (seconds)  | 24             | 24            |             |                     |
| Profile Run Time (seconds)   | 66             | 46            |             |                     |
| SOH Allocations              | 8,903,210      | 7,577,590     | -1,325,620   | **-15%**                  |
| SOH Allocated Bytes          | 839,724,663    | 735,596,560   | -104,128,103 | **-12%**                  |
| LOH Allocations              | 6,417          | 3,416         | -3,001       | **-47%**                  |
| LOH Allocated Bytes          | 3,058,521,755  | -2,794,893,240 | 263,628,515 | **-9%**                   |
| Total GC                     | 65             | 59            | -6           | **-9%**                   |
| Total GC Pause Duration (ms) | 7,527          | 7,149         | -379         | **-5%**                   |

Full comparison numbers [available here](https://1drv.ms/x/s!AuvteI9ykts1krFIAaybDsc7HXe90Q?e=NqaJmB)


jhorv:
Major revision to this PR - LOH memory allocations in the test scenario are now reduced by 78%, with LOH allocated bytes reduced by 80%, from 3 gigabytes down to 620 MB.

Additional changes:
- cleaned up the commits so that they're each more focused and better explain their intent.
- adds a new set of changes to implement `ITexture.SetData(IMemoryOwner<byte>)` so that texture data can be passed in using pooled memory.
    - This is another major reduction in LOH allocations, even more than the audio rendering changes.

Updated Test Scenario Stats:

|                              | master-250fc51 | perf-202305    | Delta         | Delta %              |
| ---------------------------- | -------------- | -------------- | ------------- | -------------------- |
| Realtime Run Time (seconds)  | 24             | 24          |               |                     |
| Profile Run Time (seconds)   | 66             | 45          |               |                     |
| SOH Allocations              | 8,903,210      | 7,608,522   | -1,294,688     | **-15%**                  |
| SOH Allocated Bytes          | 839,724,663    | 734,169,940 | -105,554,723   | **-13%**                  |
| LOH Allocations              | 6,417          | 1,399.00       | -5,018         | **-78%**                  |
| LOH Allocated Bytes          | 3,058,521,755  | 620,677,596.00 | -2,437,844,159 | **-80%**                  |
| Total GC                     | 65             | 48          | -17            | **-26%**                  |
| Total GC Pause Duration (ms) | 7,527          | 6,302.43       | -1,225         | **-16%**                  |

Updated full comparison numbers [available here](https://1drv.ms/x/s!AuvteI9ykts1krIcp9TY_0qJQAWvZA?e=TqqFMM)

sachahjkl:
Legendary PR, congrats

riperiperi:
Note: These results may be skewed by video decoding, which I can see on the profile. Nothing allocates more bytes and uploads more texture data than video decoding, but also nothing is less performance critical as games generally don't do much else while they're doing this.

src/Ryujinx.Graphics.GAL/Multithreading/Resources/ThreadedTexture.cs
```diff
@@ -114,6 +115,12 @@ public void CopyTo(BufferRange range, int layer, int level, int stride)
             _renderer.QueueCommand();
         }
 
+        public void SetData(IMemoryOwner<byte> data)
```
gdkchan, line 118: It would be nice if we didn't need to add yet another overload for this. I don't have a good suggestion right now, though.

src/Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -419,6 +420,25 @@ private void WriteTo(IntPtr data, bool forceBgra = false)
...
+            }
+
+            unsafe
+            {
+                var dataSpan = data.Memory.Span;
+                fixed (byte* ptr = dataSpan)
+                {
+                    ReadFrom((IntPtr)ptr, dataSpan.Length);
+                }
+            }
```
gdkchan, line 439: Shouldn't it dispose `data` here?

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
```diff
@@ -1526,7 +1527,7 @@ public Result UnmapPhysicalMemory(ulong address, ulong size)
 
                     while (size > 0)
                     {
-                        ulong copySize = 0x100000; // Copy chunck size. Any value will do, moderate sizes are recommended.
+                        ulong copySize = 0x100000; // Copy chunk size. Any value will do, moderate sizes are recommended.
```
gdkchan, line 1530: The reason behind the copy in chunks here is that `GetSpan` might potentially allocate. When copying a very large region, it could try to allocate a very large array (maybe above the 2GB limit) and crash or cause other issues. If this is not a problem for sequences, then copying the whole size at once could be considered.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/Parcel.cs
```diff
@@ -217,5 +221,10 @@ public ReadOnlySpan<byte> Finish()
 
             return Raw.Slice(0, (int)(Header.PayloadSize + Header.ObjectsSize + Unsafe.SizeOf<ParcelHeader>()));
         }
+
+        public void Dispose()
```
gdkchan, line 225: We should ptobably get the habit of properly implementing the dispose pattern.

gdkchan (commented):
Just a quick look, this is not a extensive review.

src/Ryujinx.Common/Memory/BytesReadOnlySequenceSegment.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace Ryujinx.Common.Memory
+{
+    /// <summary>
+    /// A concrete implementation of <seealso cref="System.Buffers.ReadOnlySequence{byte}"/>,
+    /// with methods to help build a full sequence.
+    /// </summary>
+    public sealed class BytesReadOnlySequenceSegment : System.Buffers.ReadOnlySequenceSegment<byte>
```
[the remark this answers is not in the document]
jhorv: fixed in commit 31e02f7

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/Parcel.cs
jhorv: Fixed in commit 31e02f7

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
jhorv: I appreciate that you look at the surrounding code to understand and catch things like this!

- partially fixed in commit 31e02f7
- fully fixed in commit fae35ae, eliminating the manual chunking code entirely

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <remarks>
+        /// DO NOT use <paramref name="copyDestinationIfRequiredDoNotUse"/> after calling this method, as it will only
+        /// contain a value if the value couldn't be referenced directly because it spans multiple <see cref="ReadOnlyMemory{byte}"/> segments.
+        /// To discourage use, it is recommended to to call this method like the following:
+        /// <c>
+        ///     ref readonly MyStruct value = ref sequenceReader.GetRefOrRefToCopy<MyStruct>>(out _);
+        /// </c>
+        /// </remarks>
+        /// <exception cref="ArgumentOutOfRangeException">The <see cref="SequenceReader{Byte}"/> does not contain enough data to read a value of type <typeparamref name="T"/></exception>
+        public static ref readonly T GetRefOrRefToCopy<T>(this scoped ref SequenceReader<byte> reader, out T copyDestinationIfRequiredDoNotUse) where T : unmanaged
```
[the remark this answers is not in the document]
jhorv: It exists to provide a safe memory location that can be copied to, in the case where the value isn’t in contiguous memory and a copy must be made.

This method tries to return pure references addresses inside the `ReadOnlySequence<byte>` memory, but if the value to be read spans multiple segments, then it needs to be “re-assembled” into some contiguous location. This `out T copyDestinationIfRequiredDoNotUse` is that contiguous location. If used, the returned `ref readonly` points to it.

I spent a few days benchmarking variations on this. One approach didn’t have this “copy storage” parameter. It compiled and returned the right results in trivial scenarios, but I’m pretty confident it was doing a “bad thing” - it worked by effectively returning a ref to a method-local variable. That is normally a compiler error, but after some `MemoryMarshal` magic, the compiler couldn’t tell anymore that was happening.

src/Ryujinx.Graphics.GAL/Multithreading/Resources/ThreadedTexture.cs
jhorv: Agreed. There are still some uses of the `SpanOrArray<T>` version or else I would have replaced it. I intend to to continue looking at reducing/eliminating those, since they always call `SpanOrArray<T>.ToArray()`, but I wanted to limit the scope of this already large PR.

I also considered changing `TextureSetDataCommand` to keep a `TableRef<object>` instead of `TableRef<SpanOrArray<byte>>`, but opted to leave it alone and crate the new `TextureSetByteMemoryDataCommand` instead.

If/when we replace the `SpanOrArray<byte>` overloads, I’d remove the new command type.

piplup55:
this pr causes crashes when opening any game on linux
[Ryujinx_1.0.0-dirty_2023-05-05_19-28-58.log](https://github.com/Ryujinx/Ryujinx/files/11409189/Ryujinx_1.0.0-dirty_2023-05-05_19-28-58.log)

jhorv:
> ` [62020f4](/Ryujinx/Ryujinx/pull/4777/commits/62020f4ac5f6c04e76d4c74b1dad5d541f2fc98a)`

fixed in commit [62020f4](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/62020f4ac5f6c04e76d4c74b1dad5d541f2fc98a)

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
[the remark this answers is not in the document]
gdkchan: Ok, I get it now. It needs to return a reference to *something*, and if it can't get a reference from the memory that was passed because it's not contiguous,, it needs storage that is owned by the caller where it can copy the data. It's a bit fortunate that it needs to be implemented like this, but I don't have a better suggestion.

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
gdkchan: I was wondering how the segment will handle memory regions > 2GB. The previous implementation was able to handle it since the copy was done in chuncks. The new one may not be  able to if segments doesn't take that limitation into account (I guess you would need to split it into multiple 2GB segments in that case). I don't expect this case to be common in practice though, in fact it wouldn't even work with a lot of service implementations right now, as they use array or spans which doesn't support long lengths either.

jhorv: `ReadOnlySequence<T>` supports 64-bit lengths, but individual segments are limited to 2 GB. The `GetSpan()` and `GetReadOnlySequence()` implementations take an `int size`, so they won't be bumping into the 2 GB limit, but `KPageTableBase` can be dealing with ulong sizes so... I think the solution is to put back the manual copy-chunking code, but with a chunk size of `int.MaxValue`. I'll push that change shortly.

jhorv: fixed in commit #[3402e7e](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/3402e7e951f8a4c96994a421f928d9c648dff7a8)

src/Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
```diff
@@ -331,6 +333,76 @@ private void WriteImpl(ulong va, ReadOnlySpan<byte> data)
...
+                SignalMemoryTracking(va, (ulong)size, false);
+            }
+
+            if (IsContiguousAndMapped(va, size))
+            {
+                return new ReadOnlySequence<byte>(_backingMemory.GetMemory(GetPhysicalAddressInternal(va), size));
+            }
+            else
+            {
+                BytesReadOnlySequenceSegment first = null, last = null;
```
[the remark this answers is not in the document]
jhorv: Yes, fewer segments is better. I made an assumption that the algorithm used by `GetSpan()` to copy underlying non-contiguous segments into a contiguous one is already "good enough" at minimizing copy operations, so I used the same algorithm to enumerate contiguous chunks when building the list of `BytesReadOnlySequenceSegments`.

Please say if you'd me to dig into this further, otherwise I intend to leave it as-is to avoid additional complexity to this PR.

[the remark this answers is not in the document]
gdkchan: 'GetSpan' just copies one page at a time if the range is not contiguous. I think it's fine to look at this separately.

jhorv:
Minor revision, commit https://github.com/Ryujinx/Ryujinx/pull/4777/commits/1e9dd20ae4d8ed44ec281d58d271575964760547 completely replaces the `ITexture.SetData(SpanOrArray)` methods with versions that take an `IMemoryOwner<byte>`. 

Profiling TotK comparing this PR with [master-b3bf0535](https://github.com/Ryujinx/Ryujinx/commit/b3bf05356be755bcbf82611530ebd44c9b61b384) shows the following changes in memory consumption:

|                              | master-b3bf0535 | perf-202305 (1e9dd20) | Delta           | Delta %  |
| ---------------------------- | --------------- | --------------------- | --------------- | -------- |
| Profile Run Time (seconds)   | 114             | 117                   |                 |          |
| SOH Allocations              | 24,019,044      | 24,657,830            | 638,786         | **2.66%**    |
| SOH Allocated Bytes          | 3,810,403,812   | 2,248,817,666         | \-1,561,586,146 | **\-40.98%** |
| LOH Allocations              | 7,047           | 3,166                 | \-3,881         | **\-55.07%** |
| LOH Allocated Bytes          | 4,332,309,005   | 3,397,194,830         | \-935,114,175   | **\-21.58%** |
| Total Memory Allocated Bytes | 8,142,712,817   | 5,646,012,496         | \-2,496,700,321 | **\-30.66%** |
| Total GC                     | 236             | 139                   | \-97            | **\-41.10%** |
| Total GC Pause Duration (ms) | 31,308.42       | 24,790.98             | \-6,517.44      | **\-20.82%** |

Full comparison numbers [here](https://1drv.ms/x/s!AuvteI9ykts1krtVoxZtJjVkj32r9w?e=ePxKK4)

marco-carvalho:
looking forward to it being merged into the master branch as soon as possible

src/Ryujinx.Memory/AddressSpaceManager.cs
```diff
@@ -190,15 +243,15 @@ public unsafe WritableRegion GetWritableRegion(ulong va, int size, bool tracked
 
             if (IsContiguousAndMapped(va, size))
             {
-                return new WritableRegion(null, va, new NativeMemoryManager<byte>((byte*)GetHostAddress(va), size).Memory);
+                return new WritableRegion(null, va, GetHostMemoryContiguous(va, size));
             }
             else
             {
-                Memory<byte> memory = new byte[size];
+                IMemoryOwner<byte> memoryOwner = ByteMemoryPool.Shared.Rent(size);
```
lostromb, line 250: Just speaking from experience, pooled memory entices you with promises of huge savings, BUT if you mess up the disposal pattern or leak a handle to the array anywhere then you corrupt the entire state for every other user of pooled memory. I'm going to try and look at WritableRegion and other places more in depth to make sure everything is handled properly in this regard

macvip:
Cannot merge this PR with buffer mirrors #4899 together
When using this PR to play with TOTK, my M1 MAX (16-inch 2021) runs at a lower temperature (< 70℃).

makigumo:
This PR currently breaks Megaman 11 with a crash on startup.

```log
00:00:10.694 |E| HLE.OsThread.6 Application : Unhandled exception caught: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.ArgumentOutOfRangeException: The sequence is not long enough to read the desired value. (Parameter 'reader')
   at Ryujinx.Common.Extensions.SequenceReaderExtensions.GetRefOrRefToCopy[T](SequenceReader`1& reader, T& copyDestinationIfRequiredDoNotUse) in Ryujinx/src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs:line 84
   at Ryujinx.Audio.Renderer.Server.StateUpdater.CheckMixParametersValidity(MixContext mixContext, UInt32 mixBufferCount, UInt32 inputMixCount, SequenceReader`1 parameters) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/StateUpdater.cs:line 388
   at Ryujinx.Audio.Renderer.Server.StateUpdater.UpdateMixes(MixContext mixContext, UInt32 mixBufferCount, EffectContext effectContext, SplitterContext splitterContext) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/StateUpdater.cs:line 438
   at Ryujinx.Audio.Renderer.Server.AudioRenderSystem.Update(Memory`1 output, Memory`1 performanceOutput, ReadOnlySequence`1 input) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs:line 458
   at Ryujinx.HLE.HOS.Services.Audio.AudioRenderer.AudioRenderer.RequestUpdate(Memory`1 output, Memory`1 performanceOutput, ReadOnlySequence`1 input) in Ryujinx/src/Ryujinx.HLE/HOS/Services/Audio/AudioRenderer/AudioRenderer.cs:line 78
   at Ryujinx.HLE.HOS.Services.Audio.AudioRenderer.AudioRendererServer.RequestUpdate(ServiceCtx context) in Ryujinx/src/Ryujinx.HLE/HOS/Services/Audio/AudioRenderer/AudioRendererServer.cs:line 81
   at System.RuntimeMethodHandle.InvokeMethod(Object, Void**, Signature, Boolean)
   at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)
   at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at Ryujinx.HLE.HOS.Services.IpcService.CallCmifMethod(ServiceCtx context) in Ryujinx/src/Ryujinx.HLE/HOS/Services/IpcService.cs:line 124
   at Ryujinx.HLE.HOS.Services.ServerBase.Process(Int32 serverSessionHandle, UInt64 recvListAddr) in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 435
   at Ryujinx.HLE.HOS.Services.ServerBase.ServerLoop() in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 278
   at Ryujinx.HLE.HOS.Services.ServerBase.Main() in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 168
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart() in Ryujinx/src/Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs:line 1258
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.ArgumentOutOfRangeException: The sequence is not long enough to read the desired value. (Parameter 'reader')
   at Ryujinx.Common.Extensions.SequenceReaderExtensions.GetRefOrRefToCopy[T](SequenceReader`1& reader, T& copyDestinationIfRequiredDoNotUse) in Ryujinx/src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs:line 84
   at Ryujinx.Audio.Renderer.Server.StateUpdater.CheckMixParametersValidity(MixContext mixContext, UInt32 mixBufferCount, UInt32 inputMixCount, SequenceReader`1 parameters) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/StateUpdater.cs:line 388
   at Ryujinx.Audio.Renderer.Server.StateUpdater.UpdateMixes(MixContext mixContext, UInt32 mixBufferCount, EffectContext effectContext, SplitterContext splitterContext) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/StateUpdater.cs:line 438
   at Ryujinx.Audio.Renderer.Server.AudioRenderSystem.Update(Memory`1 output, Memory`1 performanceOutput, ReadOnlySequence`1 input) in Ryujinx/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs:line 458
   at Ryujinx.HLE.HOS.Services.Audio.AudioRenderer.AudioRenderer.RequestUpdate(Memory`1 output, Memory`1 performanceOutput, ReadOnlySequence`1 input) in Ryujinx/src/Ryujinx.HLE/HOS/Services/Audio/AudioRenderer/AudioRenderer.cs:line 78
   at Ryujinx.HLE.HOS.Services.Audio.AudioRenderer.AudioRendererServer.RequestUpdate(ServiceCtx context) in Ryujinx/src/Ryujinx.HLE/HOS/Services/Audio/AudioRenderer/AudioRendererServer.cs:line 81
   at System.RuntimeMethodHandle.InvokeMethod(Object, Void**, Signature, Boolean)
   at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)
   at System.Reflection.MethodInvoker.Invoke(Object, IntPtr*, BindingFlags)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at Ryujinx.HLE.HOS.Services.IpcService.CallCmifMethod(ServiceCtx context) in Ryujinx/src/Ryujinx.HLE/HOS/Services/IpcService.cs:line 124
   at Ryujinx.HLE.HOS.Services.ServerBase.Process(Int32 serverSessionHandle, UInt64 recvListAddr) in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 435
   at Ryujinx.HLE.HOS.Services.ServerBase.ServerLoop() in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 278
   at Ryujinx.HLE.HOS.Services.ServerBase.Main() in Ryujinx/src/Ryujinx.HLE/HOS/Services/ServerBase.cs:line 168
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart() in Ryujinx/src/Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs:line 1258
```

jhorv:
> This PR currently breaks Megaman 11 with a crash on startup.

Please attach a log file.

jhorv:
> This PR currently breaks Megaman 11 with a crash on startup.


Fixed with commit [8c1ecc2](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/8c1ecc29fafe963a8cd00a0cd6a7b76b8deecacb)

makigumo:
> > This PR currently breaks Megaman 11 with a crash on startup.
> 
> Fixed with commit [8c1ecc2](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/8c1ecc29fafe963a8cd00a0cd6a7b76b8deecacb)

Thanks, confirmed.

DuckSoft:
Amazing! Tried this PR with TOTK on Apple M1 Pro (16", 2021). Smoother fps and lower temperature!

NitishM2022:
> Amazing! Tried this PR with TOTK on Apple M1 Pro (16", 2021). Smoother fps and lower temperature!

Does this PR play better on TOTK than riperiperi buffer mirror pr?

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
```diff
@@ -132,7 +134,7 @@ public DescriptorSetUpdater(VulkanRenderer gd, PipelineBase pipeline)
 
         public void Initialize()
         {
-            Span<byte> dummyTextureData = stackalloc byte[4];
```
lostromb, line 135: This buffer gets finalized without being released. Either put it in a `using` block, or bring back `TextureView.SetData(ReadOnlySpan<byte> data)` and keep the existing stackalloc code as-is (I would prefer stackalloc as the more performant option, but this code rarely gets called anyway)

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
```diff
@@ -132,7 +134,7 @@ public DescriptorSetUpdater(VulkanRenderer gd, PipelineBase pipeline)
...
+            IMemoryOwner<byte> dummyTextureData = ByteMemoryPool.Shared.RentCleared(4);
```
lostromb, line 137: `dummyTextureData` gets finalized without being released. Either put it in a `using` block, or bring back `TextureView.SetData(ReadOnlySpan<byte> data)` and keep the existing stackalloc code as-is (I would prefer stackalloc as the more performant option, but this code rarely gets called anyway)

src/Ryujinx.Memory/AddressSpaceManager.cs
lostromb: I added internal assertions locally and ran a handful of games. The only leak it found was in `DescriptorSetUpdater` which I have pointed out. Otherwise this should be fine.

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
jhorv: fixed in commit [637cc5d](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/637cc5d57e02689d831ebf269c195eba25bfee08)

src/Ryujinx.Audio/Renderer/Server/Splitter/SplitterContext.cs
```diff
@@ -164,10 +166,16 @@ private void UpdateState(scoped ref SplitterInParameterHeader inputHeader, ref R
                     {
                         ref SplitterState splitter = ref GetState(parameter.Id);
 
-                        splitter.Update(this, ref parameter, input.Slice(Unsafe.SizeOf<SplitterInParameter>()));
+                        splitter.Update(this, in parameter, ref input);
                     }
 
-                    input = input.Slice(0x1C + (int)parameter.DestinationCount * 4);
+                    // What is this?
```
lostromb, line 172: Good question, what is this? Is it tested? It looks like you're incrementing an unmanaged pointer? 0xC is a magic number here so if you could replace that with `count * sizeof(int)` or something that would be more self-documenting

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
```diff
@@ -0,0 +1,155 @@
+﻿using System;
+using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Common.Extensions
+{
+    public static class SequenceReaderExtensions
```
lostromb, line 9: most of this class should be easy to implement unit tests for

src/Ryujinx.Audio/Renderer/Server/Splitter/SplitterContext.cs
jhorv: Yes, I tested while fixing the reported Mega Man 11 crash. With the magic bump it works. Without it, subsequent magic values aren't what they should be and it chokes on processing the remainder.

src/Ryujinx.Common/Utilities/StreamUtils.cs
```diff
@@ -18,6 +18,22 @@ public static byte[] StreamToBytes(Stream input)
             }
         }
 
+        public static IMemoryOwner<byte> StreamToOwnedMemory(Stream input)
+        {
+            long bytesExpected = input.Length;
```
lostromb, line 23: What if the stream is dynamic or throws an exception when `length` is touched?
In that case, we already pull `RecyclableMemoryStream` into this project; you could copy the input data to a temporary memory stream, and then call this function recursively on that stream.

src/Ryujinx.Common/Utilities/StreamUtils.cs
```diff
@@ -18,6 +18,22 @@ public static byte[] StreamToBytes(Stream input)
...
+
+            IMemoryOwner<byte> ownedMemory = ByteMemoryPool.Shared.Rent(bytesExpected);
+
+            int bytesRead = input.Read(ownedMemory.Memory.Span);
```
lostromb, line 27: `Stream.Read` is technically not guaranteed to return the amount of bytes you request; it's rare in practice but it's not handled very well here

src/Ryujinx.Common/Utilities/StreamUtils.cs
```diff
@@ -18,6 +18,22 @@ public static byte[] StreamToBytes(Stream input)
             }
         }
 
+        public static IMemoryOwner<byte> StreamToOwnedMemory(Stream input)
```
lostromb, line 21: would there be any benefit to making this function async?

src/Ryujinx.Graphics.GAL/Multithreading/Commands/Texture/TextureSetDataCommand.cs
```diff
@@ -19,7 +19,9 @@ public void Set(TableRef<ThreadedTexture> texture, TableRef<byte[]> data)
         public static void Run(ref TextureSetDataCommand command, ThreadedRenderer threaded, IRenderer renderer)
         {
             ThreadedTexture texture = command._texture.Get(threaded);
-            texture.Base.SetData(new ReadOnlySpan<byte>(command._data.Get(threaded)));
+            IMemoryOwner<byte> data = command._data.Get(threaded);
+            texture.Base.SetData(data);
+            data.Dispose();
```
lostromb, line 24: you could replace this manual `Dispose()` with a `using`, right?

src/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
```diff
@@ -198,7 +199,7 @@ private void FinishTransfer()
                     if (target != null)
                     {
                         target.SynchronizeMemory();
-                        target.SetData(data, 0, 0, new GAL.Rectangle<int>(_dstX, _dstY, _lineLengthIn / target.Info.FormatInfo.BytesPerPixel, _lineCount));
+                        target.SetData(ByteMemoryPool.Shared.RentCopy(data), 0, 0, new GAL.Rectangle<int>(_dstX, _dstY, _lineLengthIn / target.Info.FormatInfo.BytesPerPixel, _lineCount));
```
lostromb, line 202: isn't this more wasteful than before because now we're making 2 copies of the texture data and immediately deleting one of them? Is there a reason not to keep `SetData(ReadOnlySpan<byte> data)` if that's what it is doing under the hood anyways?

src/Ryujinx.Common/Utilities/StreamUtils.cs
gdkchan: Yep, I think we had a problem with this when using Deflate compression with `DeflareStream`. The `Read` method indeed does not return the number of bytes requested there (and that was something that change with .NET 6 iirc).

src/Ryujinx.Common/Utilities/StreamUtils.cs
jhorv: None of the current callsites are in async methods, and method `StreamUtils.StreamToBytesAsync()` is unused as it is.

src/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
jhorv: This isn't an additional copy that wasn't happening before. Yes, old line 201 was calling `Texture.SetData(ReadOnlySpan<byte>...)`, but that in turn calls `HostTexture.SetData(SpanOrArray<byte>...)`, where `HostTexture` is of type `ITexture`. `ITexture` never had a `SetData(ReadOnlySpan<byte>...)` overload, only overloads taking a `SpanOrArray<byte>`, and the usage of `SpanOrArray<byte>` always created copies.

In other words: a copy was being made before this PR, it was just hidden/obfuscated by implicit conversion of `ReadOnlySpan<byte>` to `SpanOrArray<byte>`. With this PR, since `ITexture.SetData()` overloads only accept `IMemoryOwner<byte>`, the copy is visible and it's done with pooled memory.

src/Ryujinx.Common/Utilities/StreamUtils.cs
jhorv: fixed in commit [dac4fc9](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/dac4fc9a4a6caa50c5b5d69f879454b6b4365fb2)

src/Ryujinx.Common/Utilities/StreamUtils.cs
jhorv: commit [dac4fc9](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/dac4fc9a4a6caa50c5b5d69f879454b6b4365fb2) handles dynamic input streams

src/Ryujinx.Graphics.GAL/Multithreading/Commands/Texture/TextureSetDataCommand.cs
jhorv: made as suggested in commit [737bfe2](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/737bfe27f41fcabcf6b75ea24d8f0bc9bfd4ef93)

src/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
lostromb: Does `SetData` take ownership of the `IMemoryOwner` then? That would be fine I guess

src/Ryujinx.Common/Utilities/StreamUtils.cs
```diff
@@ -10,11 +11,52 @@ public static class StreamUtils
...
+            }
+            else
+            {
+                // If input is (non-seekable) then copy twice: first into a RecyclableMemoryStream, then to a rented IMemoryOwner<byte>.
+
+                using RecyclableMemoryStream output = StreamToRecyclableMemoryStream(input);
+
+                output.Position = 0;
+
+                IMemoryOwner<byte> ownedMemory = ByteMemoryPool.Shared.Rent(output.Length);
```
lostromb, line 55: You could just do a tail call `return StreamToOwnedMemory(output)` instead of duplicating the stream read logic here

src/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
jhorv: Yes. To help make that clear I will add xmldocs to `ITexture.SetData()` and adjust the existing ones on the `Texture.SetData()` methods.

lostromb: Okay, I see it depends on the implementation. `TextureView` (GPU host objects) will read the span and dispose of the memory owner, while `ThreadedTexture` will stash the data in a transient buffer as its goes through the command queue. We could theoretically remove an extra copy of the span in the case of TextureView where it just gets read directly to GPU memory, but that would interfere with what I assume is the more common case of ThreadedTexture passing data asynchronously, in which case it makes sense for it to hold a `MemoryOwner` with its own copy of the data. This is probably fine.

lostromb (approved):

src/Ryujinx.Common/Utilities/StreamUtils.cs
jhorv: I know, and I'll do that if you really want but... the logic is pretty simple when we know the `Stream` is a `RecyclableMemoryStream` and `Read()` will get the desired length in 1 call. There's no need for the `while` loop and other complications.

Commit [5a929e0](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/5a929e0c1c7fc9c1de18e993d514ad0dcb75fcc3) adds `MemoryStreamToOwnedMemory(MemoryStream input)` and adjusts `StreamToOwnedMemory()` to use it in 2 places. Best of both worlds? Over-engineered? Yes and yes!  :D

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
jhorv: added unit tests in commit [e597394](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/e597394e609d8ef74a696867d3303dade189f6e6)

src/Ryujinx.Graphics.Gpu/Engine/InlineToMemory/InlineToMemoryClass.cs
jhorv: > but that would interfere with what I assume is the more common case of ThreadedTexture passing data asynchronously, in which case it makes sense for it to hold a `MemoryOwner` with its own copy of the data.

Correct.

Commit [b331e8a](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/b331e8ad9e757b86b2d4436173af1b82b6e9ed7c) adds xml docs to `ITexture.SetData()` method signatures and adjusts existing docs on `Texture.SetData()` methods indicating that the texture data will be disposed.

marysaka (approved):
seems fine to me :+1:

Bjorn29512:
Tested about 10+ titles :

Looked for any performance changes compared to master or any other oddities that can be reliably spotted in limited playtime for each of the following titles.

Key observations :
  -  No Regression/change was noted compared to master in any of the titles tested :
   
      - Bayonetta 3    
      - FE: Engage
      - Metroid Prime Remastered 
      - Pokemon Violet
      - SMT 5
      - Super Mario Odyssey
      - TloZ: Breath of The Wild
      - Xenoblade Chronicles 2
      - Xenoblade Chronicles 3
  
  (Attachments below for evidence of testing, there is nothing else meaningful to extrapolate.)
  
  
image
image
image
image
image
image
image
image
image
image
image

saldabain:
Looking forward to seeing this merged

saldabain (approved):

riperiperi (commented):
I haven't yet fully reviewed the PR, there are about 9 files remaining and most of them are AudioRenderer.

There is a dangerously high number of things this PR fundamentally changes in ways that could subtly break and need to be explicitly tested:

- Most GPU texture decoding/encoding/layout conversion
- Every graphics backend
  - with backend multithreading enabled/disabled
  - including some platform specific workarounds that might not be testable on nvidia windows
- Every audio backend
- AudioRenderer
- Kernel
- Software memory

I'd recommend not changing so much in one PR, it's really likely that something will break and people won't have tested every possible combination.

I'm still not sure what the benefit of using IMemoryOwner<> is right now. We're pretty much only dealing in ByteMemoryPoolBuffer, and using the interface will require a virtual method call.

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
```diff
@@ -1526,7 +1527,9 @@ public Result UnmapPhysicalMemory(ulong address, ulong size)
 
                     while (size > 0)
                     {
-                        ulong copySize = 0x100000; // Copy chunck size. Any value will do, moderate sizes are recommended.
+                        // Copy chunk size. Since moving to use ReadOnlySequence<byte>, which may already be segmented,
+                        // we'll use a segment's max size (Int32.MaxValue) here for the max to read and write at once.
+                        ulong copySize = int.MaxValue;
```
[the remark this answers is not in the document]
gdkchan: Isn't the comment explaining the reason? It is not possible to create a `ReadOnlySequence` that is larger than `int.MaxValue`, so it the copy was larger than that, it would not work.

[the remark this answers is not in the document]
riperiperi: I didn't see that size was ulong. It was mostly odd that it moved from a chunk size to the largest possible size.

src/Ryujinx.Audio/Renderer/Server/Splitter/SplitterContext.cs
riperiperi: You'll have to ask mary about this one, maybe it is simply unused space.

src/Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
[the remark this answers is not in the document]
riperiperi: I guess this change wasn't done then?

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    throw new ArgumentOutOfRangeException(nameof(reader), "The sequence is not long enough to read the desired value.");
+                }
+
+                reader.Advance(lengthRequired);
+
+                return ref valueSpan[0];
+            }
+        }
+
```
riperiperi, line 61: Double newline

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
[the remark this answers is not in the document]
gdkchan: See the discussion at https://github.com/Ryujinx/Ryujinx/pull/4777#discussion_r1185676822 for more details.
I agree that using `int.MaxValue` as the maximum chunk size is rather weird (especially since it's not a power of 2 or page aligned value).

gdkchan:
This approach seems a lot more fragile in general. If we don't return the rented storages we will end leaking memory. I think something like that is fine for scoped uses where we need to rent and return only once in a single method. In this case, there are multiple users of those texture methods which makes it much more error prone. Is the improvement that this brings large enough to justify this?

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
```diff
@@ -132,7 +134,7 @@ public DescriptorSetUpdater(VulkanRenderer gd, PipelineBase pipeline)
 
         public void Initialize()
         {
-            Span<byte> dummyTextureData = stackalloc byte[4];
+            using IMemoryOwner<byte> dummyTextureData = ByteMemoryPool.Shared.RentCleared(4);
```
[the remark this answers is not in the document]
jhorv: It is necessary since all `SetData()` overloads now take an `IMemoryOwner`, so stack-allocated memory is no longer usable here. The stack allocated array wasn't an advantage here anyway, since the old version of `SetData()` using `SpanOrArray` was guaranteed to create at least one copy of the array.

src/Ryujinx.Graphics.Texture/Astc/AstcDecoder.cs
```diff
@@ -1162,7 +1162,7 @@ static Span<int> ReadIntColorValues(int number, Span<int> colorValues, ref int c
 
                 Debug.Assert(bitLength >= 1);
 
-                int a = 0, b = 0, c = 0, d = 0;
+                int a, b = 0, c = 0, d = 0;
```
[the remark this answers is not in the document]
jhorv: Because it is immediately overwritten in the next line.

src/Ryujinx.Graphics.GAL/Multithreading/Commands/Texture/TextureSetDataSliceRegionCommand.cs
```diff
@@ -25,7 +25,8 @@ public void Set(TableRef<ThreadedTexture> texture, TableRef<byte[]> data, int la
         public static void Run(ref TextureSetDataSliceRegionCommand command, ThreadedRenderer threaded, IRenderer renderer)
         {
             ThreadedTexture texture = command._texture.Get(threaded);
-            texture.Base.SetData(new ReadOnlySpan<byte>(command._data.Get(threaded)), command._layer, command._level, command._region);
+            using IMemoryOwner<byte> data = command._data.Get(threaded);
```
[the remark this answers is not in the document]
jhorv: The command `Run()` method seems a logical place IMO, tying the memory's disposal to the command's completion. I considered other places, like in the `ITexture` implementations, but there are many of them and each has 3 `SetData()` versions. It felt like that'd be more complicated and harder to get right.

src/Ryujinx.Graphics.OpenGL/Image/TextureBuffer.cs
```diff
@@ -54,19 +54,22 @@ public void CopyTo(BufferRange range, int layer, int level, int stride)
             throw new NotImplementedException();
         }
 
-        public void SetData(SpanOrArray<byte> data)
+        /// <inheritdoc/>
+        public void SetData(IMemoryOwner<byte> data)
         {
-            var dataSpan = data.AsSpan();
+            var dataSpan = data.Memory.Span;
```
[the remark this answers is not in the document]
jhorv: This is intentional, the `SetData()` implementations aren't responsible for disposing of the memory. The `IGALCommand`s is responsible for that after calling an `ITexture`'s `SetData()`.

src/Ryujinx.Graphics.Vulkan/TextureView.cs
```diff
@@ -616,19 +617,22 @@ private ReadOnlySpan<byte> GetData(CommandBufferPool cbp, PersistentFlushBuffer
             return GetDataFromBuffer(result, size, result);
         }
 
-        public void SetData(SpanOrArray<byte> data)
+        /// <inheritdoc/>
+        public void SetData(IMemoryOwner<byte> data)
         {
-            SetData(data, 0, 0, Info.GetLayers(), Info.Levels, singleSlice: false);
+            SetData(data.Memory.Span, 0, 0, Info.GetLayers(), Info.Levels, singleSlice: false);
```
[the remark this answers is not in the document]
jhorv: If the backend multithreading is disabled, the unreturned memory pool memory will be GC'd like it is today.

src/Ryujinx.Graphics.OpenGL/Image/TextureView.cs
```diff
@@ -421,70 +421,83 @@ private void WriteTo(IntPtr data, bool forceBgra = false)
...
             if (Format == Format.S8UintD24Unorm)
             {
-                dataSpan = FormatConverter.ConvertS8D24ToD24S8(dataSpan);
+                var convertedData = FormatConverter.ConvertS8D24ToD24S8(data.Memory.Span);
+                data.Dispose();
+                data = convertedData;
             }
 
-            unsafe
+            using (data)
```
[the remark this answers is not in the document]
jhorv: The method is a little different, but I can't think of another way for it to accomplish what it's doing - either using then disposing the passed-in `data` as-is, or possibly converting it via `FormatConverter.ConvertS8D24ToD24S8`, disposing the original `data`, then using/disposing the copy. Please show if you think of a cleaner way to structure this.

src/Ryujinx.Graphics.GAL/Multithreading/Commands/Texture/TextureSetDataSliceRegionCommand.cs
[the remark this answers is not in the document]
jhorv: While working through your comments I found that some `SetData()` implementations _do_ call `Dispose()`. I'd forgotten... I think originally in this PR I called `Dispose()` in the `IGALCommand`s only, and only later added it to some `SetData()` implementations by a reviewer's suggestion, figuring they understood the pipeline better than I. I don't think that's ideal, and the best I can say is that it's at least not dangerous - the IMemoryOwner gracefully handles being disposed of more than once.

jhorv:
> If we don't return the rented storages we will end leaking memory. 

Not so - the memory will be collected by the GC. We don't want to use the memory pool incorrectly, but it isn't likely to be as catastrophic as you may think.

> Is the improvement that this brings large enough to justify this?

When I think about running Ryujinx on constrained or mobile devices, and seeing before/after results like [here](https://github.com/Ryujinx/Ryujinx/pull/4777#issuecomment-1552349331) showing things like 30% less memory allocated overall - a difference of about 2.5 GB in that sample - and those numbers are just consumption. They say nothing about the performance gained from eliminating needless array copies all over the place.

src/Ryujinx.Common/Extensions/SequenceReaderExtensions.cs
jhorv: fixed in commit [7de4aa2](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/7de4aa2b191e73235e14f7a77250e8fdb9149881)

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
```diff
@@ -2929,6 +2944,7 @@ public bool OutsideStackRegion(ulong address, ulong size)
         /// <exception cref="Ryujinx.Memory.InvalidMemoryRegionException">Throw for unhandled invalid or unmapped memory accesses</exception>
         protected abstract ReadOnlySpan<byte> GetSpan(ulong va, int size);
 
+
```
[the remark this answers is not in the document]
jhorv: fixed in commit [7de4aa2](https://github.com/Ryujinx/Ryujinx/pull/4777/commits/7de4aa2b191e73235e14f7a77250e8fdb9149881)

jhorv:
> This approach seems a lot more fragile in general. If we don't return the rented storages we will end leaking memory. I think something like that is fine for scoped uses where we need to rent and return only once in a single method. In this case, there are multiple users of those texture methods which makes it much more error prone. Is the improvement that this brings large enough to justify this?

As I commented earlier: IMO this isn't more fragile, because unreturned pool memory will still be GC'd.

Please let me know if you want this. If yes, I will split it up into smaller PRs, with adjustments to put Dispose() responsibility in the ITexture implementations.

jhorv:
Superseded by PRs #6251 #6604 #6623 #6691

Issue #4164, opened 2022-12-22
Implement customizable game speed limit that doesn't break VSync


jrouzies-fr:
#### Overview:
- Similar to Yuzu where you can unlock and increase the speed limit to over or lower than 100%

### Why would this feature be useful?
It is not possible to speed up games in Ryujinx without disabling VSync, which is very useful in Yuzu.
Disabling it obviously causes screen tearing.
Using tools like cheat engine sadly is not stable, results in regular stutters even if stable FPS.

My very specific case is running Pokemon SV at 133% speed to arrive at 40 FPS (on a forced 40Hz mode on my 120hz TV). This way I get faster gameplay (but nothing extreme), and way smoother experience than 30 FPS. Sadly Yuzu is not optimized enough to hold this with some mods, and Ryujinx is, but without VSync too much screen tearing.

MetrosexualGarbodor:
Ryujinx Vsync has nothing to do with screen tearing. It's the emulated console's vsync. If you're experiencing tearing you need to enable it on your graphics driver.

gdkchan:
Are you requesting a "speed hack" that would make the game run faster internally, or the option to have a custom frame rate cap? In some cases one might imply the other, but only when the game limits its own frame rate, rather than depending on the emulator to do that.

jrouzies-fr:
> Ryujinx Vsync has nothing to do with screen tearing. It's the emulated console's vsync. If you're experiencing tearing you need to enable it on your graphics driver.

Actually I noticed the following:
-Ryujinx in windowed mode, vsync disabled in options and limited FPS: no tearing, it syncs at whatever spead.
-Ryujinx in fullscreen mode: screen tearing, FPS still limited

Forcing VSync ON in Nvidia doesn't do anything.

jrouzies-fr:
> Are you requesting a "speed hack" that would make the game run faster internally, or the option to have a custom frame rate cap? In some cases one might imply the other, but only when the game limits its own frame rate, rather than depending on the emulator to do that.
> 
> I suppose doing what you want (raising the frame rate without affecting the game speed?) would require both.

For my use case, it is at the same time a speed hack, and a way to run my game at 40 FPS which is smoother (not speaking about game speed itself).

AlistairAA:
[...]

PR #4778, opened 2023-05-03
Add Custom FPS Limiter

makotech222:
Addresses issue https://github.com/Ryujinx/Ryujinx/issues/4164

image

image

image

Taking design cues from duckstation and pcsx2, I've added 3 states for the fps limit:
 * Normal
 * Fast Forward
 * Turbo

This allows user to define a custom emulation speed while playing the game. The names "Normal, Fast Forward, Turbo" are not special in any technical way; the user can configure them to be any speed. The naming convention is just inherited from other emulators to provide consistent user experience.

The speeds are stored as a Decimal, and are multiplied against the expected normal emulation speed of 60 fps (even if a game runs at a lower fps).

Two new hotkeys are added to toggle Fast Forward and Turbo. If nothing is toggled, then speed will be in normal state. If both are toggled on, then turbo will override fast forward.

Vsync functionality has changed. It is no longer used to limit emulation speed. A speed limit of Unlimited (-1) will replicate the functionality that was previously covered by disabling vsync, for purpose of emulation speed and simulating Switch functionality. Vsync is still used by the Window classes and the Vulkan renderer, but I'm not entirely sure those do anything functionally. Vsync label has been removed from the bottom status bar and replaced with the current SpeedState.

Future Work not in this PR:
* Add code path for avalonia gui

On a side note, absolute major kudos to ryujinx dev team; I was able to clone repo and build project in first try, and find the place where I needed to make changes within 2 minutes. Absolutely marvelously structured project, great job!


al81-ru:
Hi! What about games that run at 30 or 20 (ha!) fps?

al81-ru:
Hi! I would suggest leave vsynced mode as it is, as it does frameskipping when under target framerate. And instead limit unlocked mode

makotech222:
> Hi! I would suggest leave vsynced mode as it is, as it does frameskipping when under target framerate. And instead limit unlocked mode

I'm doubtful it does frameskipping, but if you set your speed to 60 fps, it will do the same thing that enabling vsync does.

al81-ru:
> I'm doubtful it does frameskipping, but if you set your speed to 60 fps, it will do the same thing that enabling vsync does.

Corrected, not frame skipping, but frame delaying. Compose() may not post frame if it is not ready, but will wait another _ticksPerFrame to try again

emmauss:
Vsync value is used in the village window to enable immediate present mode, which is necessary on supported GPUs to bypass/disable driver vsync. It's the equivalent of swap interval for opengl.

emmauss:
Also, those 3 states don't truly work as you would think in ryujinx. There is no fast forward or high speed mode. It's completely up to the game as to how and when to send frames, even when we tell the game that a frame is always ready. Some games adjust game play speed to the current frame rate, effectively running at 100% game speed, even when frame rate drops. Those options don't so what they are meant to do. Other games run their game loop around the swap rate. In those games, we can have fast forward and turbo modes.

makotech222:
Any example games of that behavior? In the games i tested, it worked perfectly.

al81-ru:
> Any example games of that behavior?

Luigi's Mansion 3 relies on TickSource's ticks, it maintains 30 fps with internal waitings

emmauss:
Celeste for instance, always run at full game speed, even when frame rate drops to like 20 and looks like a slide show. Game physics run at their own pace

makotech222:
So disabling vsync doesn't have any effect on the games? In that case, there's likely not much we can do outside of someone writing a mod it seems like? In any case, it seems like most games respond perfectly to it, so no reason to not implement it.

emmauss:
> So disabling vsync doesn't have any effect on the games? In that case, there's likely not much we can do outside of someone writing a mod it seems like? In any case, it seems like most games respond perfectly to it, so no reason to not implement it.

Both games behave differently with vsync off. Celeste will simply generate as many frames as possible, but game physics run at a steady rate.
How different is fast forward and turbo, from each other and vsync off currently?

gdkchan:
I think it makes more sense to add a slider where you can set any target frame rate that you want. There was a request for that on #4164 (I think?).

makotech222:
Yeah the GUI portion isn't done, but it will be user customizable limits. Currently, just set the numbers in the config file.

MutantAura:
Worth mentioning that the numbers themselves won't be fully accurate to what the user may think the setting does. As emmauss sort of explained games can do different things and while a 60FPS game will work as expected, a game targeting 30FPS won't see those values. Instead they'll see exactly half of the target.

Currently this PR acts as more of a "framerate target multiplier" which is fine, but isn't really what the requester had in mind.

makotech222:
@MutantAura That is correct, the terminology can probably be reworded to something like what duckstation does, which is a percentage of full speed (150%, 100%, etc)

emmauss:
> @MutantAura That is correct, the terminology can probably be reworded to something like what duckstation does, which is a percentage of full speed (150%, 100%, etc)

What work 150% speed be for a 30fps game?

emmauss:
If possible, could you drop in on our dev channel on discord, for much easier discussion.

riperiperi:
Just a few things:

- This is adjusting the refresh rate of the device, rather than a frame cap.
- Vsync off (and swapInterval 0 behaviour, which _is_ present on switch) should not be removed. They allow the framerate to be completely uncapped, with the vsync event coinciding with the frame presentation. This is useful for fully dynamic framerates in certain games, and benchmarking in others.

src/Ryujinx.HLE/Switch.cs
```diff
@@ -127,6 +142,49 @@ public void EnableCheats()
             FileSystem.ModLoader.EnableCheats(Processes.ActiveApplication.ProgramId, TamperMachine);
         }
 
+        public void SetSpeedState(SpeedState newState)
```
MutantAura, line 145: Left a general codestyle message on Discord.

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -749,6 +773,9 @@ public void LoadDefault()
             Graphics.GraphicsBackend.Value            = OperatingSystem.IsMacOS() ? GraphicsBackend.Vulkan : GraphicsBackend.OpenGl;
             Graphics.PreferredGpu.Value               = "";
             Graphics.ShadersDumpPath.Value            = "";
+            Graphics.NormalEmulationSpeed.Value       = 1.0m;
+            Graphics.FastForwardEmulationSpeed.Value           = 1.5m;
+            Graphics.TurboEmulationSpeed.Value                 = -1.0m;
```
MutantAura, line 778: Misaligned.

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -1409,6 +1464,9 @@ public ConfigurationLoadResult Load(ConfigurationFileFormat configurationFileFor
             Graphics.AntiAliasing.Value               = configurationFileFormat.AntiAliasing;
             Graphics.ScalingFilter.Value              = configurationFileFormat.ScalingFilter;
             Graphics.ScalingFilterLevel.Value         = configurationFileFormat.ScalingFilterLevel;
+            Graphics.NormalEmulationSpeed.Value       = configurationFileFormat.NormalEmulationSpeed;
+            Graphics.FastForwardEmulationSpeed.Value           = configurationFileFormat.FastForwardEmulationSpeed;
+            Graphics.TurboEmulationSpeed.Value                 = configurationFileFormat.TurboEmulationSpeed;
```
MutantAura, line 1469: Misaligned

src/Ryujinx/Ui/MainWindow.glade
```diff
@@ -698,15 +698,15 @@
...
                         <child>
-                          <object class="GtkLabel" id="_vSyncStatus">
+                          <object class="GtkLabel" id="_speedState">
                             <property name="visible">True</property>
                             <property name="can-focus">False</property>
                             <property name="halign">start</property>
                             <property name="margin-left">5</property>
                             <property name="margin-right">5</property>
-                            <property name="label" translatable="yes">VSync</property>
+                            <property name="label" translatable="yes">Normal</property>
```
MutantAura, line 709: Not sure I like the "Normal", "Fast-Forward", "Turbo" terminology too much although I've never used Duckstation or PCSX2. Personally I would simply leave it as "VSync", "VSync (XYZ%)", "Unlocked" which I think better describe to our current userbase what these do. Would also allow us to maintain the colouring, which I also quite like!

More thoughts on this from others would be good though.

MutantAura (commented):
Functionality wise this works as expected. Left a few comments.

Would be preferable to also implement Avalonia side before merge as it is the priority for all new features.

src/Ryujinx.HLE/Switch.cs
makotech222: Okay checked in some changes to address codestyle. take another look and let me know if you want me to do any more

src/Ryujinx/Ui/MainWindow.glade
makotech222: With your terminology, it implies the first and third options are not configurable, which wouldn't be as desirable. I'd welcome a better terminology, though, if anyone has any suggestions.

Sentox6: Speaking from a user perspective, I'm not sure they should be configurable. Personally, I'd like to see a Normal, Custom, and Unlocked setting, with only the second being, well, customisable. I do think this would better communicate the actual behaviour of the setting than the current VSync nomenclature.

If per-game settings are eventually implemented, this would become significantly more useful, of course.

makotech222: I disagree. As a user who loves this feature in other emulators, I customize all three and swap between them pretty commonly as I play. It's pretty easy to understand what it does.

Sentox6: If that's the desired functionality, then the user should be able to add and label any number of custom speeds as they see fit, rather than three arbitrarily-named settings.

makotech222:
Added change to include Custom values in the GTK UI:

image

I think I've implemented all that I wanted to. @MutantAura is Avalonia support going to be a requirement before merge? I'd like to take that on at a later date, since its not the official gui yet.

MutantAura:
> I think I've implemented all that I wanted to. @MutantAura is Avalonia support going to be a requirement before merge? I'd like to take that on at a later date, since its not the official gui yet.

On this front I'm not sure. Ideally we don't want them to diverge on features and if they do then Avalonia should be prioritized. 
Most of the legwork is already done here, if you don't have the time to finish that side then I can possibly add it in a follow up or push to the branch...

TSRBerry:
This needs a rebase, a workflow should have timed out here, but didn't.

makotech222:
Okay rebased it

makotech222:
Added avalonia GUI:

image

image

jcm93:
Just tried to test this out on macOS but it seems not to be working, is that just because it's not wired up with Avalonia (Avalonia being the only available GUI on Mac)? Looks cool, looking forward to testing it.

jcm93:
> I tried to test it on Ava (macOS), but it seems to have no effect yet.
> If you type in a custom value, there is an x shown while the field is active.
> Because of that you can't actually see, what you are typing.

I had a comment earlier that I deleted until I could do a proper followup.

I was testing on macOS and also thought that it wasn't working. It turns out that it is, but the changes only take effect after restarting the title. You should see the changes reflected after restarting the game.

I agree that the input is a mess, not sure if it's specific to macOS. The custom input asking for a decimal number is also odd, given that the dropdown has percent values. Should be uniform across both I would think.

siegmund-heiss-ich:
True.. I didn't think about it..
Nothing in `System` can be changed while a game is running.
Normal emulation speed works fine.
But i can't toggle the other two options.
I assigned the Hotkeys correctly but it didn't work for me.
I also tested it with resolution scaling to be sure I did nothing wrong.
Also if I disable vsync, the game is locked at `Normal Emulation Speed`.

makotech222:
Sorry guys, didn't test avalonia as much as I should have. All those issues should now be fixed! Please try testing again.

siegmund-heiss-ich:
Sadly all the described problems are still there on macOS.
I can't toggle any other speed while emulating a game.

Also the custom field should also has values in %:
Bildschirmfoto 2023-05-30 um 12 51 36

This `x` is quite annoying. It appears, when the input field is active:
Bildschirmfoto 2023-05-30 um 12 51 46

But great work! Keep it up!

siegmund-heiss-ich:
I deleted my last comment because I didn't test the latest version.
Sorry for that.

Now everything seems to work great on Avalonia.

My last request would be, to show values in % if you want to set a custom value.
At the Moment it looks like this:
Bildschirmfoto 2023-05-30 um 13 12 31

makotech222:
Okay, i've changed the custom value to be more like a percent to match dropdown options (multiplied decimal by 100)

MutantAura:
Functionality wise it's all working good on my end. I still have reservations about a few mainly UI related things though.

1. The naming system I still really dislike. They simply do not align with how Switch games work. Fast-Forward and Turbo make no sense in literally thousands of games. They won't fast forward, nor speed up at all. Plenty of first party titles also have dynamic framerates and so this entire aspect of naming feels stuck in the past and non-applicable to modern emulation. I would still really suggest naming things _what they do_. What is happening here is simply changing the FPS target for present interval, so simply "VSync (X%)" is way more accurate, correlates with what users already know, and is easy to see in screenshots for troubleshooting or sharing purposes.

2. This breaks the VSync toggle hotkey. F1/TAB depending on when the config file was created. Should probably be fixed to either cycle through the options or simply flit between what currently is called "normal" and "turbo". 

3. Keep the green/red colours for the widget text. At this point I think they're part of the program identity and to lose them would make me sad :( 

I would appreciate if anyone else could chime in on my first point as it's obviously subjective.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -409,6 +483,16 @@ public void LoadCurrentConfiguration()
...
+            var foundFastForwardSpeed = _emulationSpeeds.Values.Contains(config.Graphics.FastForwardEmulationSpeed) ? _emulationSpeeds.First(x => x.Value == config.Graphics.FastForwardEmulationSpeed).Key : null;
+            FastForwardSpeed = foundFastForwardSpeed != null ? AvailableFastForwardSpeeds.IndexOf(AvailableFastForwardSpeeds.First(x => (string)x.Content == foundFastForwardSpeed)) : AvailableFastForwardSpeeds.IndexOf(AvailableFastForwardSpeeds.First(x => (string)x.Content == _emulationSpeeds.First(x => x.Value == -2).Key));
+            CustomFastForwardSpeed = config.Graphics.FastForwardEmulationSpeed * 100.0m;
+            var foundTurboSpeed = _emulationSpeeds.Values.Contains(config.Graphics.TurboEmulationSpeed) ? _emulationSpeeds.First(x => x.Value == config.Graphics.TurboEmulationSpeed).Key : null;
+            TurboSpeed = foundTurboSpeed != null ? AvailableTurboSpeeds.IndexOf(AvailableTurboSpeeds.First(x => (string)x.Content == foundTurboSpeed)) : AvailableTurboSpeeds.IndexOf(AvailableTurboSpeeds.First(x => (string)x.Content == _emulationSpeeds.First(x => x.Value == -2).Key));
+            CustomTurboSpeed = config.Graphics.TurboEmulationSpeed * 100.0m;
```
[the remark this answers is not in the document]
MutantAura: Split these out for readability, may also wish to define constants for 100.0m and -2.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -393,17 +393,6 @@ public Brush ProgressBarForegroundColor
...
-        {
-            get => _vsyncColor;
-            set
-            {
-                _vsyncColor = value;
-
-                OnPropertyChanged();
-            }
-        }
-
```
MutantAura, line 406: I like the colours :(

MutantAura (commented):
Couple of design issues to resolve, mainly the abundant use of some magic decimals which are not instantly obvious. GUI layout and design should be shifted to glade and perhaps think about your hotkey design because at the moment there is some toggle vsync code still lying around not really doing much.

BrunoSilvestre00 (approved):

jcm93:
I was trying to rebase this for a custom build and it seems like the Avalonia combo boxes that bind to ``AvailableNormalSpeeds``, ``AvailableFastForwardSpeeds`` are broken, presumably by the recent Avalonia update, just a word of warning to anyone trying to make builds off of this PR.

makotech222:
Okay i've rebased again. Can we please get one more approval on this? I don't have much resources to continue maintaining this PR as things change over time. We can always iterate on the feature over time after it is merged as well.

BrunoSilvestre00 (approved):

TSRBerry:
I'm not sure what you are asking here.

1. The format workflow is still failing.
2. You didn't even address most of the review comments.

> We can always iterate on the feature over time after it is merged as well.

This can't be merged in the current state. Open review comments generally mean there at least needs to be some discussion and maybe also a few changes. We don't want to merge PRs if there are still concerns about them, especially if there weren't discussed yet.

> I don't have much resources to continue maintaining this PR as things change over time.

I understand the time issue of course and get that it's not easy to keep your PR up to date all the time. Especially, since we are really slow at reviewing things. That's why I like to rebase for people if I notice the PR isn't up to date and we have taken quite some time to review it already.
I'd do the same with your PR if all checks were succeeding and review comments addressed.

makotech222:
Sorry i was being terse, just tired at the moment.

Format workflow failing is news to me. I'll take a look at that when i have time.
Forgot those review comments were there from mutant also. Will try to look when i have time

jcm93:
The following are just some usability opinions from having used this PR a fair amount on a Switch title that features tech that involves specific FPS values.

Integrating with the VSync option should probably be done. In its current state it's fairly confusing to hit a button to change the target refresh rate to 200%, but still be at 60 Hz because VSync is still enabled. I agree that it would make the most sense to use the existing VSync toggle hotkey to switch between the options in order, instead of having specific hotkeys for "fast forward" and "turbo". Especially because the behavior is somewhat confusing (i.e. if you're on fast forward, and hit the turbo hotkey twice, are you back at fast forward? or at normal?) At a minimum, incorporating the VSync option and hotkey seems good.

I agree with MutantAura's comments about the naming, but I don't see a super easy solution. Perhaps the "fast forward" and "turbo" names could be kept in the Settings dialog, but in the Ryujinx status bar use the suggested "VSync %" and show the value? I'd personally prefer nameable presets by the user, but that's just my opinion. I also think it's clearer to use number refresh rate values instead of percentages, but again, just my opinion. Overall the PR seems to work quite well and I certainly find it very useful and I thank you a lot for implementing it!

jcm93:
Hey there,

I wanted to check and see if this had been abandoned or not. If you don't have enough time to see it across the finish line, I was going to offer to take it over. I don't want to step on your toes if you are going to finish though. Let me know!

jcm93:
OK, sounds good. Thanks for all the work!

StealthTranslations:
> Hey there,
> 
> I wanted to check and see if this had been abandoned or not. If you don't have enough time to see it across the finish line, I was going to offer to take it over. I don't want to step on your toes if you are going to finish though. Let me know!

Did you take over the project?

jcm93:
Yes, I should have tagged it, the new PR is at https://github.com/Ryujinx/Ryujinx/pull/5735.

PR #4785, opened 2023-05-03
Update SettingsWindow.cs

ekeskin:
fix saving if directory path directly pasted in to the text field instead of using FileChooser.


TSRBerry (approved):
Thank you, lgtm!

MutantAura (approved):
Must have missed this case. Cheers.

PR #4446, opened 2023-02-19
IPC - Refactor Bcat service to use new ipc

emmauss:

PR #4803, opened 2023-05-04
IPC - Refactor Bcat service to use new ipc - Revisit

emmauss:
This should fix an issue with #4446 .
ServiceCreator is returned for bcat services, instead of the previous BcatService.


AcK77:
It could be nice to know the change with #4446

AcK77 (approved):

gdkchan (approved):
Code looks good, and we got confirmation that the issues with the previous PR were fixed.

Bjorn29512:
Forgot to leave a comment before but i can confirm that Smash Ult and Pokemon BDSP don't crash anymore and work as intended, cant confirm for ACNH since i don't have it.

image
image

PR #4815, opened 2023-05-05
Use ToLowerInvariant when detecting GPU vendor.

blueskythlikesclouds:
Fixes GPU vendor detection on Turkish systems where "NVIDIA Corporation" for example would become "nvıdıa corporation" and fail the check.


riperiperi (approved):
Makes sense, this just keeps getting us. C# should have really defaulted to invariant culture for these methods.

AcK77 (approved):

PR #4817, opened 2023-05-05
GPU: Remove CPU region handle containers

riperiperi:
Another one for the "I don't know why I didn't do this earlier" pile.

This removes the "Cpu" prefixed region handle classes, which each mirror a region handle type from Ryujinx.Memory.

Originally, not all projects had a reference to Ryujinx.Memory, so these classes were introduced to bridge the gap. Someone else crossed that bridge since, so these classes don't have much of a purpose anymore.

This PR replaces all uses of CpuRegionHandle etc to their direct Ryujinx.Memory versions.

RegionHandle methods (specifically QueryModified) are about the hottest path there is in the entire emulator, so there is a nice boost from doing this.


riperiperi:
Before:
image

After:
image

gdkchan (approved):
Makes sense, the extra indirection also causes more work when we need to change something related to the memory tracking.

Foul-Tarnished:
I hope this will be merged soon, great work, no issues so far for me

marysaka (approved):

PR #4821, opened 2023-05-05
Shader: Use correct offset for storage constant buffer elimination

riperiperi:
This fixes a regression from #2237 where constant buffer access elimination would not work if the newly mapped storage slot didn't match up with the storage slot on hardware.

I've also changed it to work no matter what constant buffer the address is on.


gdkchan (approved):
lgtm, thanks for fixing this. There is more room for improvement here, since it already has to find the base address and offset before calling this replace method, so what it is doing is a bit redundant.

marysaka (approved):

Issue #2430, opened 2021-06-30
[Pokemon Sw/Sh] Daily events + Pokejobs broken


SinFlame:
### What's the issue you encountered?
RTC doesn't run ingame. This affects majority of time based events, berry picking, den resets and pokejobs.
Changing clock in Ryu settings doesn't fix it. Although weirdly, Weather changes happen. 

### How can the issue be reproduced?
Register for a pokejob in pokemon centre and wait a few hours. Check back on the job and it will still state 24 hours remaining.

### Log file
No log files given as it isn't crash related. 

### Environment?

- Ryujinx version: 1.0.6945</br>
- Game version: 1.3.2</br>
- System Specs:
  - OS: Microsoft Windows 10.0.19042 (X64)
  - CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz ; 8 logical
  - GPU: NVIDIA RTX 3060
  - RAM: 32GB
- Applied Mods : Yes

### Additional context?

Additional info about your environment:</br>

Leischii:
This effects all events that need time to pass. So in addition to the Pokejobs problem, it also effects the respawning of berries, raids and wandering Pokemon.
Noteworthy behaviour: Pokejobs work as long as you don't close the emulator. If you close it, the time resets to the initial time.
E.g. Starting 2h Pokejob, waiting 1:59h without closing the emulator, restarting the game --> Time left is 2h again

TheSuspect2306:
few month later, when will it be fixed, its really not fun :/

da-welsh:
Strangely enough, last night I updated from 1.0.7082 to 1.0.7093. 

On doing so, when I launched I noticed it was night time in game, which seemed odd. Mostly due to me never having played at night. 

However, it did cause me to check my pokejobs, and lo and behold, a job that was set for 1 hour - that had been active for weeks - was complete.  This learly was not because i had been running for an hour, i had just launched the emulator after an update. It did lead me to wonder if there was something in the updates that had messed with the clock. after checking the log, i wondered if it had anything to do with the timezone implementation? Im off UTC by 1 hour, so that could be something? or maybe it was a freak action that forced it to complete? 

Either way, **something** happened in those updates that caused a 1 hour pokejob to complete.

Begueiro:
I'm having a similar issue, on the latest Ryujinx release, 1.0.7107 (which I think came out today).

[...]

PR #4822, opened 2023-05-05
time: Update for 15.0.0 changes and fixes long standing issues

marysaka:
Last time we did an upgrade on the time service was during 9.x era, it was about time to take back that reverse again!

15.0.0 added a new structure on the shared memory to get steady clock raw timepoints with a granularity in nanoseconds.

This PR implements this new part.

I plan to write a follow up with a bit of refactoring of this ancient part of the emulator.

As always, reverse and work done by your truly.

This should also fix local and network clock returning 0 under usage with shared memory.

FIx https://github.com/Ryujinx/Ryujinx/issues/2430.

PS: As a reminder, if this change is reused anywhere else, work should be credited as Ryujinx and not my person.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Horizon.cs
+++ b/src/Ryujinx.HLE/HOS/Horizon.cs
@@ -200,6 +200,7 @@ namespace Ryujinx.HLE.HOS
 
             LibHacHorizonManager = device.Configuration.LibHacHorizonManager;
 
+            // We hardcode a clock source id to avoid it changing between each start.
             // TODO: use set:sys (and get external clock source id from settings)
             // TODO: use "time!standard_steady_clock_rtc_update_interval_minutes" and implement a worker thread to be accurate.
             UInt128 clockSourceId = new UInt128(0x36a0328702ce8bc1, 0x1608eaba02333284);
--- a/src/Ryujinx.HLE/HOS/Horizon.cs
+++ b/src/Ryujinx.HLE/HOS/Horizon.cs
@@ -222,22 +222,22 @@ namespace Ryujinx.HLE.HOS
                 internalOffset = internalOffset.AddSeconds(-3600L);
             }
 
-            internalOffset = new TimeSpanType(-internalOffset.NanoSeconds);
+            systemTime = new TimeSpanType(systemTime.NanoSeconds + internalOffset.NanoSeconds);
 
             // First init the standard steady clock
-            TimeServiceManager.Instance.SetupStandardSteadyClock(TickSource, clockSourceId, TimeSpanType.Zero, internalOffset, TimeSpanType.Zero, false);
+            TimeServiceManager.Instance.SetupStandardSteadyClock(TickSource, clockSourceId, TimeSpanType.Zero, TimeSpanType.Zero, TimeSpanType.Zero, false);
             TimeServiceManager.Instance.SetupStandardLocalSystemClock(TickSource, new SystemClockContext(), systemTime.ToSeconds());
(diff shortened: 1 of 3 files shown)


src/Ryujinx.HLE/HOS/Services/Time/TimeSharedMemory.cs
```diff
@@ -20,6 +20,7 @@ class TimeSharedMemory
         private const uint LocalSystemClockContextOffset    = 0x38;
         private const uint NetworkSystemClockContextOffset  = 0x80;
         private const uint AutomaticCorrectionEnabledOffset = 0xC8;
+        private const uint ContinuousAdjustmentTimePointOffset = 0xD0;
```
gdkchan, line 23: Those constants are not aligned. Personally I'd just remove  the alignment here to avoid needing to change it in the future, if more constants are added.

src/Ryujinx.HLE/HOS/Services/Time/TimeSharedMemory.cs
```diff
@@ -58,10 +51,40 @@ public void SetAutomaticCorrectionEnabled(bool isAutomaticCorrectionEnabled)
...
+        }
+
+        private void UpdateSteadyClock(ITickSource tickSource, UInt128 clockSourceId, TimeSpanType currentTimePoint)
+        {
+            ulong counter = tickSource.Counter;
+
+            ContinuousAdjustmentTimePoint adjustmentTimePoint = new ContinuousAdjustmentTimePoint
+            {
+                ClockOffset = counter,
+                ClockMultiplier = -1,
```
gdkchan, line 66: Just so that I understand this, why does it have a multiplier of -1 and divides the frequency by 2? Is that just something that the service does?

src/Ryujinx.HLE/HOS/Services/Time/TimeSharedMemory.cs
TSRBerry: Doesn't really matter, since formatting is addressed by #4670 anyway and would remove the alignment entirely here, right?

gdkchan: While that change is not merged, I'm still making formatting related suggestions on PRs.

TSRBerry:
Testing with [switch-examples](https://github.com/switchbrew/switch-examples): time
image

In game time and prepo logs for ACNH were also always correct! So seems like everything is working fine!

TSRBerry (approved):
Lgtm!

src/Ryujinx.HLE/HOS/Horizon.cs
```diff
@@ -202,7 +202,7 @@ public Horizon(Switch device)
 
             // TODO: use set:sys (and get external clock source id from settings)
             // TODO: use "time!standard_steady_clock_rtc_update_interval_minutes" and implement a worker thread to be accurate.
-            UInt128 clockSourceId = UInt128Utils.CreateRandom();
+            UInt128 clockSourceId = new UInt128(0x36a0328702ce8bc1, 0x1608eaba02333284);
```
gdkchan, line 205: It would be nice to add a comment about this being a randomly generated value, and the actual value not being important.

gdkchan (approved):
lgtm other than the comment. Nice to see the long standing Pokémon issue fixed.

src/Ryujinx.HLE/HOS/Horizon.cs
AcK77: I suggested the same privatly :)

AcK77 (approved):

PR #4825, opened 2023-05-06
AM: Stub some service calls

AcK77:
Some IPC I have stubbed during private testing and I don't want to deal with them anymore. Nothing more.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ICommonStateGetter.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ICommonStateGetter.cs
@@ -9,8 +9,10 @@ using System;
 
 namespace Ryujinx.HLE.HOS.Services.Am.AppletAE.AllSystemAppletProxiesService.SystemAppletProxy
 {
-    class ICommonStateGetter : IpcService
+    class ICommonStateGetter : DisposableIpcService
     {
+        private readonly ServiceCtx _context;
+
         private Apm.ManagerServer       _apmManagerServer;
         private Apm.SystemManagerServer _apmSystemManagerServer;
         private Lbl.LblControllerServer _lblControllerServer;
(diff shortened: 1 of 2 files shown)


marysaka (approved):

TSRBerry (approved):

src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ICommonStateGetter.cs
```diff
@@ -23,11 +23,16 @@ class ICommonStateGetter : IpcService
         private int  _messageEventHandle;
         private int  _displayResolutionChangedEventHandle;
 
+        private KEvent _acquiredSleepLockEvent;
+        private int _acquiredSleepLockEventHandle;
```
gdkchan, line 27: I think this handle should be closed when the service object is disposed (if it's non-zero), to prevent handle leak on the service.

AcK77: We could consider it if that will cause issues. A lot of other classes doesn't currently do that either so I think it's fine as-is.

gdkchan: FWIW we are currently *not* doing that in many service objects right now, which is wrong. But it usually does not cause issues, it's only a problem if the service object is destroyed and re-created multiple times (or if it generates a new handle many times in general).

TSRBerry: Although we aren't doing that right now, it feels wrong to ignore a potential issue here.

gdkchan (approved):

src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ICommonStateGetter.cs
AcK77: Done

PR #4827, opened 2023-05-06
Ryujinx.Ava: fixes for random hangs on exit

jhorv:
I've experienced a lot of instances recently where Ryujinx.Ava would hang on exit, sometimes for so long that I'd open up Task Manager and kill it. It was always with Release builds and attaching the debugger didn't show much. It happened so often, I didn't imagine I should have tried harder because today, try as I might, I've had a real hard time getting it to happen so that I could investigate.

**Hang #1**
I finally got it to happen though, and with a Debug build too. I found one thread stuck waiting, started by `NotificationHelper.SetNotificationManager()` with `Task.Run()`. The thread itself does a simple foreach over a `BlockingCollection<Notification>` but it doesn't have an exit condition. This PR adds one via a `CancellationTokenSource`.

This uses the ~~`WindowNotificationManager.DetachedFromLogicalTree`~~ `MainWindow.Closing` event to detect when it's time to shut down. I don't have much Avalonia experience, maybe someone who does can suggest a better way to detect when it's time to shut down.

**Hang #2**
After fixing the above, I found another location that appeared to hang up stopping was the `ManualResetEvent _terminationEvent` in `AudioRenderSystem`. It was determined this event was unnecessary, and so was removed completely.

With these changes, the # of hangs on exit are noticeably reduced. The ones that do occur all appear to be from known issues in the threaded renderer.

### DIFF
--- a/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
+++ b/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
@@ -389,7 +389,12 @@ namespace Ryujinx.Audio.Renderer.Server
 
             if (_executionMode == AudioRendererExecutionMode.Auto)
             {
-                _terminationEvent.WaitOne();
+                TimeSpan terminationWaitTimeout = TimeSpan.FromSeconds(5);
+
+                if (_terminationEvent.WaitOne(terminationWaitTimeout) == false)
+                {
+                    Logger.Warning?.PrintMsg(LogClass.AudioRenderer, $"AudioRenderSystem.Stop() timed out after {terminationWaitTimeout} while waiting for the termination signal. Execution will continue to avoid interfering with emulation shutdown. Please file a detailed bug report.");
+                }
             }
 
             Logger.Info?.Print(LogClass.AudioRenderer, $"Stopped renderer id {_sessionId}");
--- a/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
+++ b/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
@@ -677,7 +677,6 @@ namespace Ryujinx.Audio.Renderer.Server
                         _isDspRunningBehind = true;
                     }
                 }
-
             }
         }
 
--- a/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
+++ b/src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
@@ -31,7 +31,6 @@ namespace Ryujinx.Audio.Renderer.Server
         private AudioRendererRenderingDevice _renderingDevice;
         private AudioRendererExecutionMode _executionMode;
         private IWritableEvent _systemEvent;
-        private ManualResetEvent _terminationEvent;
         private MemoryPoolState _dspMemoryPoolState;
         private VoiceContext _voiceContext;
         private MixContext _mixContext;
@@ -83,7 +82,6 @@ namespace Ryujinx.Audio.Renderer.Server
         public AudioRenderSystem(AudioRendererManager manager, IWritableEvent systemEvent)
         {
             _manager = manager;
-            _terminationEvent = new ManualResetEvent(false);
             _dspMemoryPoolState = MemoryPoolState.Create(MemoryPoolState.LocationType.Dsp);
             _voiceContext = new VoiceContext();
             _mixContext = new MixContext();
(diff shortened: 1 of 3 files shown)


src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
```diff
@@ -389,7 +389,12 @@ public void Stop()
 
             if (_executionMode == AudioRendererExecutionMode.Auto)
             {
-                _terminationEvent.WaitOne();
+                TimeSpan terminationWaitTimeout = TimeSpan.FromSeconds(5);
+
+                if (_terminationEvent.WaitOne(terminationWaitTimeout) == false)
```
gdkchan, line 394: I'm not sure why this event is necessary. If it's to ensure this method will not return while there is some command still being processed, just taking the `_lock` lock already serves that purpose, since the `SendCommands` method takes this lock.

jhorv: I'm going to study how it could go unset, then I'll make another change to remove the event. Meanwhile, I'll make this PR a draft until the set of changes stabilizes.

jhorv: Removed `_terminationEvent` in commit #[8c44f7d](https://github.com/Ryujinx/Ryujinx/pull/4827/commits/8c44f7dd8f4487d3aa7c4c6c119c6eac2eb1026a)

jhorv:
Moving this out of Draft. This PR is an attempt at fixing 2 apparent causes of "hang on exit" when closing Ryujinx.Ava. Attaching the debugger when a hang occurred, I found that in many instances these 2 events were being waited on and seemingly nothing else:

- `NotificationHelper.SetNotificationManager()` was using a `ManualResetEvent _templateAppliedEvent` being waited on indefinitely. This was the majority of cases that I personally experienced.
- `AudioRenderSystem` had a `ManualResetEvent _terminationEvent` that could be waited on indefinitely in the `Stop()` method. I experienced this personally but far less frequently.

This PR:
- reworks `NotificationManger` to use the existing `AsyncWorkQueue<T>` to manage the threading/cancellation details.
- removes the `AudioRenderSystem` `_terminationEvent`, it having been determined to be redundant anyway.

Anecdotal evidence suggests these changes do reduce the frequency of "hang on exit" events. When a "hang on exit" has occurred with this PR, the debugger shows it likely to be caused by known issues when disposing of many created resources, many client threads, etc., which this PR will not try to address.

src/Ryujinx.Ava/UI/Helpers/NotificationHelper.cs
```diff
@@ -29,25 +28,29 @@ public static void SetNotificationManager(Window host)
...
-                foreach (var notification in _notifications.GetConsumingEnumerable())
+            Lazy<AsyncWorkQueue<Notification>> maybeAsyncWorkQueue = new Lazy<AsyncWorkQueue<Notification>>(
+                () => new AsyncWorkQueue<Notification>(notification =>
                 {
                     Dispatcher.UIThread.Post(() =>
                     {
                         _notificationManager.Show(notification);
                     });
+                }, "UI.NotificationThread", _notifications)
+                , LazyThreadSafetyMode.ExecutionAndPublication);
```
AcK77, line 39:
```suggestion
            var maybeAsyncWorkQueue = new Lazy<AsyncWorkQueue<Notification>>(
                () => new AsyncWorkQueue<Notification>(notification =>
                    {
                        Dispatcher.UIThread.Post(() =>
                        {
                            _notificationManager.Show(notification);
                        });
                    },
                    "UI.NotificationThread",
                    _notifications),
                LazyThreadSafetyMode.ExecutionAndPublication);
```

AcK77 (changes requested):

src/Ryujinx.Audio/Renderer/Server/AudioRenderSystem.cs
```diff
@@ -686,10 +677,7 @@ public void SendCommands()
                         _isDspRunningBehind = true;
                     }
                 }
-                else
-                {
-                    _terminationEvent.Set();
-                }
+
```
AcK77, line 680:
```suggestion
```

AcK77 (approved):
LGTM (with my last comment fixed)!

riperiperi (approved):
Looks fine, though it is weird that the PR has audren changes and does not mention why. I am not able to reproduce a freeze with this PR, though I don't use avalonia casually so not sure if I could get it to happen on master either.

jhorv:
> Looks fine, though it is weird that the PR has audren changes and does not mention why. I am not able to reproduce a freeze with this PR, though I don't use avalonia casually so not sure if I could get it to happen on master either.

Sorry for that, I added text to the top comment describing how this fixes 2 hang causes, the 2nd one being in `AudioRenderSystem`.

PR #4832, opened 2023-05-06
Fix the issue of unequal check for amiibo file date due to the lack o…

2435043xia:
Fix the issue of unequal check for amiibo file date due to the lack of sub-second units in the header, causing slow opening of the amiibo interface.


riperiperi (approved):
This does seem to be faster, though not by much as my internet is pretty speedy.

AcK77:
Could you change it in Ava project too please :

https://github.com/Ryujinx/Ryujinx/blob/0bc8151c7ecdacc1506305a8d60e7b3c7b13622d/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs#L373

2435043xia:
> Could you change it in Ava project too please :
> 
> https://github.com/Ryujinx/Ryujinx/blob/0bc8151c7ecdacc1506305a8d60e7b3c7b13622d/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs#L373

sure, it has already been supplemented

PR #4837, opened 2023-05-07
Update release workflow & Add jobs for macOS

### LINKED ISSUE #4836: [Feature Request] Automatic builds for macOS
### Overview

Currently, the bot which generates binaries appears to only do so for Windows and Linux ([example](https://github.com/Ryujinx/release-channel-master/releases/tag/1.1.763)):

image

It would be nice to provide builds for macOS users, too.

### Smaller details

N/A

### Nature of request

Infrastructure

So macOS users can try the latest builds without needing to download the .Net SDK and build the project manually.

TSRBerry:
This PR resolves #4836. This allows macOS users to get a release (or PR build) based on the current master.

Releases will also be able to use the updater, since @marysaka previously added support for that as well.

Additionally some changes to the release workflow have been made to take advantage of matrix strategies, which should help to shorten the time until people are able to update to the next version after a new push to master.

**Note**: We haven't upstreamed all macOS changes yet (see: #4062), so these releases won't work as well as the 1.1.0-macos1 build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,9 +18,14 @@ on:
       - '*.yml'
       - 'README.md'
 
+env:
+  POWERSHELL_TELEMETRY_OPTOUT: 1
+  DOTNET_CLI_TELEMETRY_OPTOUT: 1
+  RYUJINX_BASE_VERSION: "1.1.0"
+
 jobs:
   build:
-    name: ${{ matrix.os }} (${{ matrix.configuration }})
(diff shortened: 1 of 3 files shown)


marysaka (approved):

TSRBerry:
Note that we will need to change the required workflows which need to pass since the naming changed slightly.

image

(cc @marysaka, since you probably will be the one doing that :sweat_smile:)

riperiperi (approved):

PR #4842, opened 2023-05-07
Stop SDL from inhibiting sleep (#4828).

### LINKED ISSUE #4828: [Bug] Computer not going to sleep with no game running
### Description of the issue

Ryujinx prevents the computer from sleeping when no game is launched or when a game execution is suspended. It should only prevent sleep when a game is running.

#1850 is the PR that added the functionality to prevent sleep.

On windows you can check that Ryujinx is preventing the computer from going to sleep by opening a terminal with admin rights and the command : 
> powercfg /requests

The result : 
> DISPLAY :
> [PROCESS] \Device\HarddiskVolume9\Emulators\Ryujinx\bin\Ryujinx.exe
> 
> SYSTEM :
> None.
> 
> AWAYMODE :
> None.
> 
> EXECUTION :
> None.
> 
> PERFBOOST :
> None.
> 
> ACTIVELOCKSCREEN :
> None.

### Reproduction steps

- Start Ryujinx with no other program preventing sleep
- Wait enough time without touching the keyboard or mouse so the computer can go to sleep
- The computer does not go to sleep

### OS

Windows 11

### Ryujinx version

1.1.762

LoneBoco:
This fixes #4828.

The problem was that SDL was inhibiting the display from sleeping even when a game was not running.  I added a hint to disable that SDL feature.  The normal behavior in `DisplaySleep.cs` will still apply and prevent sleep while a game is currently running.

### DIFF
--- a/src/Ryujinx.SDL2.Common/SDL2Driver.cs
+++ b/src/Ryujinx.SDL2.Common/SDL2Driver.cs
@@ -63,6 +63,7 @@ namespace Ryujinx.SDL2.Common
                 SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
                 SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_SWITCH_HOME_LED, "0");
                 SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_JOY_CONS, "1");
+                SDL_SetHint(SDL_HINT_VIDEO_ALLOW_SCREENSAVER, "1");
 
 
                 // NOTE: As of SDL2 2.24.0, joycons are combined by default but the motion source only come from one of them.


EliEron:
Thank you for looking into this. When I came across #4828 I was a bit confused because @d3xMachina was correct, the display sleep PR (which I authored) was only meant to take effect when a game was running. 

I tried looking into it myself a bit but I had no idea it would end up being SDL. So nice catch 👍.

TSRBerry (approved):
Thank you, I had no idea this is default behavior for SDL!

d3xMachina:
I want to note that DisplaySleep is only implemented for Windows. It might be needed to implement it for Linux (example [here](https://github.com/libsdl-org/SDL/blob/26903f237f276f1c472d8c1a5378ff93fa7783d2/src/video/x11/SDL_x11events.c#L1770) from the SDL library) or to disable the hint for Linux. I don't have Linux so I can't check the behavior.

LoneBoco:
> I want to note that `DisplaySleep.cs` is only implemented for Windows. It might be needed to implement it for Linux (example [here](https://github.com/libsdl-org/SDL/blob/26903f237f276f1c472d8c1a5378ff93fa7783d2/src/video/x11/SDL_x11events.c#L1770) from the SDL library) or to disable the hint for Linux. I don't have Linux so I can't check the behavior.

You could probably just call `SDL2_DisableScreenSaver()` and `SDL2_EnableScreenSaver()` inside `DisplaySleep.cs` for other operating systems.  Just rely on however SDL does it.

Foul-Tarnished:
Does SDL inhibit sleep when emulation is paused ? (with F5)

There's no cpu/gpu usage, but ram/vram usage stays high obviously, so no idea what is the most desirable.
afaik Sleep should by default be able to work when ram is full.

LoneBoco:
> Does SDL inhibit sleep when emulation is paused ? (with F5)
> 
> There's no cpu/gpu usage, but ram/vram usage stays high obviously, so no idea what is the most desirable. afaik Sleep should by default be able to work when ram is full.

No.  When SDL is initialized, it, by default (on Windows), calls this:
`SetThreadExecutionState(ES_CONTINUOUS | ES_DISPLAY_REQUIRED);`

This prevents your monitor from sleeping as long as Ryujinx is running.  Even if no game is being played.

Ryujinx only registers itself as using the display (and system) when you start a game.  It unregisters itself when you stop the game.  Just pausing the game would keep sleep disabled.

marysaka (approved):

PR #4843, opened 2023-05-07
Vulkan: Batch vertex buffer updates

riperiperi:
This PR batches together vertex buffer updates to reduce the number of individual calls to update vertex buffers. It also avoids rebinding vertex buffers that have not changed, which is common when games are doing instanced rendering.

image

Improves performance slightly on BOTW on my system. Some areas bind more vertex buffers per draw than others.

Note that any improvement greatly depends on the game and the driver.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/VertexBufferUpdater.cs
+++ b/src/Ryujinx.Graphics.Vulkan/VertexBufferUpdater.cs
@@ -1,16 +1,18 @@
 ﻿using Silk.NET.Vulkan;
+using System;
+
+using VkBuffer = Silk.NET.Vulkan.Buffer;
 
 namespace Ryujinx.Graphics.Vulkan
 {
-    internal class VertexBufferUpdater
+    internal class VertexBufferUpdater : IDisposable
     {
-        private int MaxVertexBuffers = 16;
         private VulkanRenderer _gd;
 
         private uint _baseBinding;
         private uint _count;
 
-        private NativeArray<Buffer> _buffers;
+        private NativeArray<VkBuffer> _buffers;
         private NativeArray<ulong> _offsets;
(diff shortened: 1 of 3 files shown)


src/Ryujinx.Graphics.Vulkan/VertexBufferUpdater.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    internal class VertexBufferUpdater
+    {
+        private int MaxVertexBuffers = 16;
+        private VulkanRenderer _gd;
+
+        private uint _baseBinding;
+        private uint _count;
+
+        private NativeArray<Buffer> _buffers;
```
gdkchan, line 13: Those needs to be disposed at some point since they allocate unmanaged memory.

src/Ryujinx.Graphics.Vulkan/VertexBufferUpdater.cs
```diff
@@ -0,0 +1,74 @@
...
+        public void BindVertexBuffer(CommandBufferScoped cbs, uint binding, Buffer buffer, ulong offset, ulong size, ulong stride)
+        {
+            if (_count == 0)
+            {
+                _baseBinding = binding;
+            }
+            else if (_baseBinding + _count != binding)
+            {
+                Commit(cbs);
+                _baseBinding = binding;
```
gdkchan, line 37: Shouldn't that path also set `_count` to 0?

riperiperi: Commit sets `_count` to 0.

gdkchan (approved):
lgtm, thanks!

marysaka (approved):

PR #4845, opened 2023-05-07
Vulkan: Simplify MultiFenceHolder and managing them

riperiperi:
MultiFenceHolder is currently absolutely reliant on all users submitting to the _same command buffer pool_, which means they must all come from the same thread. This is because waitiers register the command buffer index, which is only valid for entries of the same pool. The locking and dictionary structures greatly slow down the process of binding and ref counting for _all types of buffer_, while also being unnecessary when these rules are followed.

This PR simplifies MultiFenceHolder in a few ways:
- `ReservedCommandBuffer.Waitables` is now a List for less overhead. Enforcing uniqueness is done by `MultiFenceHolder.AddFence` returning true only when the command buffer index is not yet registered on the waitable. Similarly, the call to Contains has been replaced with asking the waitable if it has the current fence, rather than checking the set on the command buffer.
- `MultiFenceHolder` now maintains an array of fences rather than a dictionary. Each entry in the array represents a command buffer in the pool. If the value in the array at that command buffer index is null, there is no registered fence yet. This removes all cost for trying to add and trying to remove fences, which was pretty considerable.
  - The WaitForFences methods have been updated to extract active fences from this array, and properly handle a fence already being disposed by the time it's iterated.
  - Everything is completely lockless, which is good for a method this hot.
- Added "Try" methods to `Auto<>` and `FenceHolder` to properly handle cases where we get a resource but it might already be disposed. These are a bit slower as they use a compare exchange to increment the value only if it's not 0, but they are at least correct if a bunch of threads are contending with each other.

There was an existing case where the rules about keeping `Get(cbs)` uses to the same pool are broken, which is the copy source when flushing data from a device local buffer. The destination is fine, because it is only used by the command buffer pool owned by the background thread. This will probably cause issues right now, but to avoid creating new ones I decided it's worth fixing.

I've changed this case to manage keeping the source buffer alive outside the copy method. Since it waits for the copy to complete anyways, it can just increment and decrement the reference count around the use manually.

---

Improves performance in Vulkan bottlenecked games such as BOTW. (at least on my system)

Testing various games would be helpful to ensure nothing ends up broken.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
+++ b/src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
@@ -82,7 +82,7 @@ namespace Ryujinx.Graphics.Vulkan
         /// </summary>
         /// <param name="cbIndex">Command buffer index of the command buffer that owns the fence</param>
         /// <param name="fence">Fence to be added</param>
-        /// <returns>True if the constant buffer's previous fence value was null</returns>
+        /// <returns>True if the command buffer's previous fence value was null</returns>
         public bool AddFence(int cbIndex, FenceHolder fence)
         {
             ref FenceHolder fenceRef = ref _fences[cbIndex];
@@ -106,7 +106,7 @@ namespace Ryujinx.Graphics.Vulkan
         }
 
         /// <summary>
-        /// Determines if a fence referenced on the given constant buffer.
+        /// Determines if a fence referenced on the given command buffer.
         /// </summary>
         /// <param name="cbIndex"></param>
         /// <returns></returns>
--- a/src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
+++ b/src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
@@ -170,15 +170,15 @@ namespace Ryujinx.Graphics.Vulkan
 
             for (int i = 0; i < count; i++)
             {
-                fences[fenceCount] = fenceHolders[i].Get();
-
-                if (fenceCount < i)
+                if (fenceHolders[i].TryGet(out Fence fence))
                 {
-                    fenceHolders[fenceCount] = fenceHolders[i];
-                }
+                    fences[fenceCount] = fence;
 
-                if (fences[i].Handle != 0)
-                {
+                    if (fenceCount < i)
+                    {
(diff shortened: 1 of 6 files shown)


src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
```diff
@@ -80,25 +82,37 @@ public bool IsBufferRangeInUse(int offset, int size)
...
-            {
-                _fences.Remove(fence);
-            }
+            _fences[cbIndex] = null;
+        }
+
+        /// <summary>
+        /// Determines if a fence referenced on the given command buffer.
+        /// </summary>
+        /// <param name="cbIndex"></param>
```
gdkchan, line 111: Missing description for parameter and return value.

src/Ryujinx.Graphics.Vulkan/MultiFenceHolder.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using Silk.NET.Vulkan;
+using System;
+using System.Buffers;
```
gdkchan, line 3: Not needed?

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

Bjorn29512:
Tested about 20 titles :

Looked for any performance changes compared to master or any visual oddities that can be reliably spotted in limited playtime for each of the following titles.

Key observations :

 - No Regression/change was noted compared to master in any of the titles tested :

   (Note for users: The changes are tangible and do infact improve performance in BOTW but it depends on your setup, gains are mostly noted on PCs with weaker CPUs.)

      - Bayonetta 3
      - FE: Three Hopes
      - FE: Engage
      - FE: Three Houses
      - Mario + Rabbids: Sparks of Hope
      - Metroid Dread 
      - Metroid Prime Remastered 
      - Pokemon Violet
      - Pokemon Sword
      - SMT 5
      - Super Mario Odyssey
      - TloZ: Link's Awakening
      - TloZ: Skyward Sword HD
      - TloZ: Breath of The Wild
      - Xenoblade Chronicles DE
      - Xenoblade Chronicles 2
      - Xenoblade Chronicles 3
      - Yo Kai Watch 4

PR #4848, opened 2023-05-07
Ava: Fix SystemTimeOffset calculation

AcK77:
During testing of #4822, Mary pointed out the way we calculate time offset is wrong in our Avalonia UI. This PR fixed that. The axaml file is autoformatted too.

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
@@ -398,8 +398,10 @@ namespace Ryujinx.Ava.UI.ViewModels
             Language = (int)config.System.Language.Value;
             TimeZone = config.System.TimeZone;
 
-            CurrentDate = DateTime.Now.Date;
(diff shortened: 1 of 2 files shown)


src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -397,10 +398,9 @@ public void LoadCurrentConfiguration()
             Language = (int)config.System.Language.Value;
             TimeZone = config.System.TimeZone;
 
-            DateTime dateTimeOffset = DateTime.Now.AddSeconds(config.System.SystemTimeOffset);
+            CurrentDate = DateTime.Now.Date;
```
gdkchan, line 401: Can't you put `DateTime.Now` on a local variable and use that to ensure that both the date and time are coming from the same point in time?

AcK77: Fixed ;)

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #4859, opened 2023-05-08
vulkan: Pass Vk instance to VulkanRenderer

marysaka:
This will allow in the future possible multiple driver selection without any need of LD preload. (useful when testing custom version of mesa for example)


AcK77 (approved):

TSRBerry (approved):
Tested Gtk and Ava and both work fine!

Code also lgtm!

PR #4861, opened 2023-05-08
Changed LastPlayed field from string to nullable DateTime

### LINKED ISSUE #2660: `System.FormatException` when stopping emulation and the system date format is too custom

SamusAranX:
This should fix #2660 and make #2661 obsolete by changing the type of `ApplicationData`/`ApplicationMetadata`'s `LastPlayed` fields from string to nullable DateTime.

I fixed most the problems that cropped up after the change and added a value converter for Avalonia as well as a new readonly property `ApplicationData.LastPlayedString` for the GTK UI.

The one big problem that remains is that after the change, Ryujinx will consider old metadata.json files invalid and reset everything but the title back to default. I'm working on a solution that allows for a graceful migration but I wanted to submit the PR now to get feedback regarding the rest.

### DIFF
--- a/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -417,22 +417,22 @@ namespace Ryujinx.Ui.App.Common
 
                         if (appMetadata.LastPlayedOld == default || appMetadata.LastPlayed.HasValue)
                         {
-                            // don't do the migration if last_played doesn't exist or last_played_utc already has a value
+                            // Don't do the migration if last_played doesn't exist or last_played_utc already has a value.
                             return;
                         }
 
-                        // migrate from string-based last_played to DateTime-based last_played_utc
+                        // Migrate from string-based last_played to DateTime-based last_played_utc.
                         if (DateTime.TryParse(appMetadata.LastPlayedOld, out DateTime lastPlayedOldParsed))
                         {
                             Logger.Info?.Print(LogClass.Application, $"last_played found: \"{appMetadata.LastPlayedOld}\", migrating to last_played_utc");
                             appMetadata.LastPlayed = lastPlayedOldParsed;
 
-                            // migration successful: deleting last_played from the metadata file
+                            // Migration successful: deleting last_played from the metadata file
                             appMetadata.LastPlayedOld = default;
                         }
                         else
                         {
-                            // migration failed: emitting warning but leaving the unparsable value in the metadata file so the user can fix it
+                            // Migration failed: emitting warning but leaving the unparsable value in the metadata file so the user can fix it.
                             Logger.Warning?.Print(LogClass.Application, $"Last played string \"{appMetadata.LastPlayedOld}\" is invalid for current system culture, skipping (did current culture change?)");
                         }
                     });
--- a/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -416,20 +416,6 @@ namespace Ryujinx.Ui.App.Common
                         appMetadata.Title = titleName;
                     });
 
-                    if (appMetadata.LastPlayed != "Never")
-                    {
-                        if (!DateTime.TryParse(appMetadata.LastPlayed, out _))
-                        {
-                            Logger.Warning?.Print(LogClass.Application, $"Last played datetime \"{appMetadata.LastPlayed}\" is invalid for current system culture, skipping (did current culture change?)");
-
-                            appMetadata.LastPlayed = "Never";
-                        }
-                        else
-                        {
-                            appMetadata.LastPlayed = appMetadata.LastPlayed[..^3];
-                        }
-                    }
-
                     ApplicationData data = new()
                     {
                         Favorite = appMetadata.Favorite,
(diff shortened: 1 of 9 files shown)


riperiperi (approved):
I think this is a good way of solving this problem. Worth noting that going back to an older build will wipe the last played value (for any testers), but I think this is expected as settings changes tend to do that too.

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -414,21 +414,28 @@ byte[] Read(long position, int size)
...
                         appMetadata.Title = titleName;
-                    });
 
-                    if (appMetadata.LastPlayed != "Never")
-                    {
-                        if (!DateTime.TryParse(appMetadata.LastPlayed, out _))
+                        if (appMetadata.LastPlayedOld == default || appMetadata.LastPlayed.HasValue)
                         {
-                            Logger.Warning?.Print(LogClass.Application, $"Last played datetime \"{appMetadata.LastPlayed}\" is invalid for current system culture, skipping (did current culture change?)");
+                            // don't do the migration if last_played doesn't exist or last_played_utc already has a value
```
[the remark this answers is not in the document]
SamusAranX: Thanks, just pushed a commit that updates the comments!

marysaka (approved):

TSRBerry (approved):
Nicely done, thank you very much!

Eli0t77:
Will there be a migration fix for never played games?
I currently get this warning in the log for 37 games (I recently moved my games to a new pc)
`|W| GUI.ApplicationLibraryThread Application LoadApplications: Last played string "Never" is invalid for current system culture, skipping (did current culture change?)`

ExE-Boss:
@Eli0t77
The old system used `"Never"` where the new system uses `null`, so it’s fine (it just probably should’ve been handled silently).

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -414,21 +414,28 @@ byte[] Read(long position, int size)
...
-                            appMetadata.LastPlayed = "Never";
+                        // Migrate from string-based last_played to DateTime-based last_played_utc.
+                        if (DateTime.TryParse(appMetadata.LastPlayedOld, out DateTime lastPlayedOldParsed))
+                        {
+                            Logger.Info?.Print(LogClass.Application, $"last_played found: \"{appMetadata.LastPlayedOld}\", migrating to last_played_utc");
+                            appMetadata.LastPlayed = lastPlayedOldParsed;
+
+                            // Migration successful: deleting last_played from the metadata file.
+                            appMetadata.LastPlayedOld = default;
                         }
```
ExE-Boss, line 432: This should’ve also had a case for: `appMetadata.LastPlayed == "Never"`, like what the old code did.

SamusAranX:
rhe "Never" case is already implicitly handled, it's just that I left in the old warning message.
As someone in the Discord remarked (don't remember the name, sorry, but they were right) the warning message shouldn't have remained at all. The warning message is 100% superfluous at this point and can be ignored.

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
SamusAranX: See my comment below.

PR #4869, opened 2023-05-09
Fix the restart after an update.

LoneBoco:
After running the updater for Ryujinx and clicking Yes to restart, I am greeted with an error message:

> The application was unable to start correctly (0xc0000142).
> Click OK to close the application.

I then have to manually start Ryujinx to finish the update.  This PR changes the way that Ryujinx is launched after an update to fix this error.

### DIFF
--- a/src/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/src/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -295,14 +295,7 @@ namespace Ryujinx.Modules
                 if (shouldRestart)
                 {
                     List<string> arguments = CommandLineState.Arguments.ToList();
-                    string ryuName = Path.GetFileName(Environment.ProcessPath);
                     string executableDirectory = AppDomain.CurrentDomain.BaseDirectory;
-                    string executablePath = Path.Combine(executableDirectory, ryuName);
-
-                    if (!Path.Exists(executablePath))
-                    {
-                        executablePath = Path.Combine(executableDirectory, OperatingSystem.IsWindows() ? "Ryujinx.exe" : "Ryujinx");
-                    }
 
                     // On macOS we perform the update at relaunch.
                     if (OperatingSystem.IsMacOS())
@@ -310,14 +303,20 @@ namespace Ryujinx.Modules
                         string baseBundlePath = Path.GetFullPath(Path.Combine(executableDirectory, "..", ".."));
                         string newBundlePath = Path.Combine(UpdateDir, "Ryujinx.app");
                         string updaterScriptPath = Path.Combine(newBundlePath, "Contents", "Resources", "updater.sh");
-                        string currentPid = Process.GetCurrentProcess().Id.ToString();
+                        string currentPid = Environment.ProcessId.ToString();
 
-                        executablePath = "/bin/bash";
                         arguments.InsertRange(0, new List<string> { updaterScriptPath, baseBundlePath, newBundlePath, currentPid });
-                        Process.Start(executablePath, arguments);
+                        Process.Start("/bin/bash", arguments);
                     }
                     else
                     {
+                        // Find the process name.
+                        string ryuName = Path.GetFileName(Environment.ProcessPath);
+                        if (!Path.Exists(Path.Combine(executableDirectory, ryuName)))
+                        {
+                            ryuName = OperatingSystem.IsWindows() ? "Ryujinx.exe" : "Ryujinx";
+                        }
+
                         ProcessStartInfo processStart = new(ryuName, string.Join(' ', arguments))
                         {
                             UseShellExecute = true,
@@ -327,7 +326,6 @@ namespace Ryujinx.Modules
                         Process.Start(processStart);
                     }
 
-                    Process.Start(executablePath, arguments);
                     Environment.Exit(0);
                 }
             }
--- a/src/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/src/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -314,6 +314,17 @@ namespace Ryujinx.Modules
 
                         executablePath = "/bin/bash";
                         arguments.InsertRange(0, new List<string> { updaterScriptPath, baseBundlePath, newBundlePath, currentPid });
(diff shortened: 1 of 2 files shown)


LoneBoco:
> Does that only happen with Gtk?

Do you mean if this only happens with the default UI and not with the new Ava UI?  If so, then no, it happens with both.

src/Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -314,6 +314,17 @@ private static async void UpdateRyujinx(Window parent, string downloadUrl)
...
+                        ProcessStartInfo processStart = new(ryuName, string.Join(' ', arguments))
+                        {
+                            UseShellExecute = true,
+                            WorkingDirectory = executableDirectory
+                        };
+
+                        Process.Start(processStart);
                     }
 
                     Process.Start(executablePath, arguments);
```
AcK77, line 330: This doesn't have to be removed ?

src/Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -314,6 +314,17 @@ private static async void UpdateRyujinx(Window parent, string downloadUrl)
 
                         executablePath = "/bin/bash";
                         arguments.InsertRange(0, new List<string> { updaterScriptPath, baseBundlePath, newBundlePath, currentPid });
+                        Process.Start(executablePath, arguments);
```
AcK77, line 317: `executablePath` don't seems to be used elsewhere now. Maybe you could remove the above condition and related code too and then do:

```suggestion
                        Process.Start("/bin/bash", arguments);
```

AcK77 (changes requested):

src/Ryujinx.Ava/Modules/Updater/Updater.cs
LoneBoco: You are right.  I accidentally left a line of code in and didn't notice it.

src/Ryujinx.Ava/Modules/Updater/Updater.cs
LoneBoco: I cleaned up the code a bit, removing some variables that aren't really being used anymore.

AcK77 (approved):
LGTM! I think this has to be tested in all OSs!

LoneBoco:
> LGTM! I think this has to be tested in all OSs!

@AcK77 I tested on Linux with KDE Neon (Ubuntu 22.04).  The auto-update worked fine for the Gtk version but not for the Ava version.  The issue is that Environment.ProcessPath returns the name of the renamed executable, so we then run the fail condition and just choose "Ryujinx" as the executable name.  With Ava, the actual executable name is Ryujinx.Ava.  This is a problem that existed in the original code.

Should I change the code to look for Ryujinx.Ava instead for the Ava build?  I assume at some point Ava will become the norm.  So the code could be left as-is and work in the future, or I can change it to look for Ryujinx.Ava and the code will need to be changed in the future if the executable name ever changes.

One thing I am going to do is strip the .ryuold off the executable name if it finds it.  That will be the immediate fix for the problem.  My above question is mainly about how to handle the fallback case where the exe name couldn't be found.

LoneBoco:
I am unable to test on MacOS.  I wonder if MacOS will have the same issue with the Avalonia code (it looks for Ryujinx.app).

src/Ryujinx/Modules/Updater/UpdateDialog.cs
```diff
@@ -47,9 +48,14 @@ private void YesButton_Clicked(object sender, EventArgs args)
...
-                string ryuExe  = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, ryuName);
 
-                Process.Start(ryuExe, CommandLineState.Arguments);
+                ProcessStartInfo processStart = new(ryuName, string.Join(' ', CommandLineState.Arguments))
+                {
+                    UseShellExecute = true,
+                    WorkingDirectory = ReleaseInformation.GetBaseApplicationDirectory()
+                };
+
+                Process.Start(processStart);
```
TSRBerry, line 58:
```suggestion
                ProcessStartInfo processStart = new(ryuName)
                {
                    UseShellExecute = true,
                    WorkingDirectory = ReleaseInformation.GetBaseApplicationDirectory()
                };

                foreach (string argument in CommandLineState.Arguments)
                {
                    processStart.ArgumentList.Add(argument);
                }

                Process.Start(processStart);
```

I think doing something like that would be safer.

src/Ryujinx.Ava/Modules/Updater/Updater.cs
```diff
@@ -295,28 +295,37 @@ private static async void UpdateRyujinx(Window parent, string downloadUrl)
...
+                            ryuName = OperatingSystem.IsWindows() ? "Ryujinx.exe" : "Ryujinx";
+                        }
+
+                        ProcessStartInfo processStart = new(ryuName, string.Join(' ', arguments))
+                        {
+                            UseShellExecute = true,
+                            WorkingDirectory = executableDirectory
+                        };
+
+                        Process.Start(processStart);
```
TSRBerry, line 326:
```suggestion
                        ProcessStartInfo processStart = new(ryuName)
                {
                    UseShellExecute = true,
                    WorkingDirectory = executableDirectory
                };

                foreach (string argument in arguments)
                {
                    processStart.ArgumentList.Add(argument);
                }

                Process.Start(processStart);
```
So we don't run into escaping issues.

TSRBerry (changes requested):
MacOS is unaffected by these changes, so no need to worry about that.

I added 2 comments about the way the arguments are passed to `processStart`, but other than that, lgtm!

src/Ryujinx/Modules/Updater/UpdateDialog.cs
LoneBoco: Done.

src/Ryujinx.Ava/Modules/Updater/Updater.cs
LoneBoco: Done.

TSRBerry (approved):
Thank you, lgtm!

PR #4874, opened 2023-05-10
Ensure background translation threads exited before disposing JIT

gdkchan:
This fixes a potential issue that could happen if all guest threads exit, but there is still a background translation in flight on one of the background translation threads. In this situation, it would crash because some collections have already been disposed. This seems extremely rare since I never got it myself, but got a report of it happening from @AcK77.


AcK77 (approved):
After some tests my issue seems to be fixed! Thanks!

riperiperi (approved):
Seems reasonable.

PR #4879, opened 2023-05-10
Fix missing domain service object dispose

### LINKED ISSUE #4877: [Bug] Animal Crossing New Horizons - Infinite loading into save (BCat related?)
### Description of the issue

When loading into the save file (after pressing "A" in the title screen) it loads indefinetily.
The title menu shows my island as it is in my save file.
Game still runs (as in isn't frozen) because it changes the "tips" that shows at the loading screen and has the icon animation.
Work correctly in Ryujinx 1.1.773

Thing is, I tried with a previous save (I have daily backups) and it worked.
Tried the latest save file that didn't load, opened it in previous ryujinx versions and in yuzu, then saving. Those "updated" saves didn't work either.

I can share the save files, but I can think of two changes between them:
- First villager at campsite. ()
- International Museum Day (I think it was) announcement by Isabelle, mentioning that event in May 18th (this could have to do with the BCat changes on 1.1.774). 
Since I saved after the message, I haven't test if it shows at 1.1.773.
This would also explain the fact that has no issue showing my island at the menu (being able to load the save file but not the BCat event that would trigger Isabelle announcement)
 

### Reproduction steps

- Use ryujinx 1.1.774 or 1.1.775
- Open Animal Crossing: New Horizons
- Press A at the menu
- (Infinite loading... or at least 10min)

### Log file

[Ryujinx_1.1.774_2023-05-10_23-18-00.log](https://github.com/Ryujinx/Ryujinx/files/11446159/Ryujinx_1.1.774_2023-05-10_23-18-00.log)

### OS

Windows 11 - 22621.1635

### Ryujinx version

1.1.774

### Game version

2.0.6

### CPU

Ryzen 2700 OC@4GHz

### GPU

NVIDIA RTX 2070 Super

### RAM

2x8GB DDR4 3433Mhz

### List of applied mods

Had the ``60 fps`` mod but also check at some point without it, same result.

### Additional context?

_No response_

gdkchan:
Fixes a regression caused by #4803 that caused Animal Crossing New Horizons to hang when loading a save, because one of the BCAT service objects was not being disposed, and so future access would return `TargetLocked` errors forever.

Fixes #4877.

### DIFF
--- a/src/Ryujinx.Horizon/Sdk/Sf/Cmif/ServerDomainManager.cs
+++ b/src/Ryujinx.Horizon/Sdk/Sf/Cmif/ServerDomainManager.cs
@@ -165,6 +165,12 @@ namespace Ryujinx.Horizon.Sdk.Sf.Cmif
 
                     entry.Owner = null;
                     obj = entry.Obj;
+
+                    if (obj.ServiceObject is IDisposable disposableObj)
+                    {
+                        disposableObj.Dispose();
+                    }
+
                     entry.Obj = null;
                     _entries.Remove(entry.Node);
                     entry.Node = null;


MetrosexualGarbodor:
Should close #4877

AcK77 (approved):

riperiperi (approved):

PR #4880, opened 2023-05-11
Vulkan: Partially workaround MoltenVK InvalidResource error

riperiperi:
There's a bug with MoltenVK where binding a storage buffer more than once with different stage flags causes resource usage to register incorrectly, which causes the command buffer to fail. There's more information here: https://github.com/KhronosGroup/MoltenVK/issues/1870

This PR partially works around this issue by setting the storage buffer usage flags to all active stages when running under MVK. This means that it will always register the buffer with all possible stages it can be used on.

This can still break if there are non-rasterizing draws, which generally happens with TFB games. Working around that is a lot more involved, and it's probably not worth it when we don't have TFB emulation anyways.

This should be removed when the issue is fixed in MoltenVK.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/PipelineLayoutFactory.cs
+++ b/src/Ryujinx.Graphics.Vulkan/PipelineLayoutFactory.cs
@@ -12,6 +12,28 @@ namespace Ryujinx.Graphics.Vulkan
             ShaderStageFlags.FragmentBit |
             ShaderStageFlags.ComputeBit;
 
+        private static ShaderStageFlags ActiveStages(uint stages, bool supportsGeometryShader)
+        {
+            ShaderStageFlags stageFlags = 0;
+
+            while (stages != 0)
+            {
+                int stage = BitOperations.TrailingZeroCount(stages);
+                stages &= ~(1u << stage);
+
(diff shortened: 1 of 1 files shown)


src/Ryujinx.Graphics.Vulkan/PipelineLayoutFactory.cs
```diff
@@ -12,6 +12,28 @@ static class PipelineLayoutFactory
...
+
+            while (stages != 0)
+            {
+                int stage = BitOperations.TrailingZeroCount(stages);
+                stages &= ~(1u << stage);
+
+                stageFlags |= stage switch
+                {
+                    1 => ShaderStageFlags.FragmentBit,
+                    2 => supportsGeometryShader ? ShaderStageFlags.GeometryBit : ShaderStageFlags.VertexBit,
```
gdkchan, line 27: Not sure if we need that here, it was necessary due to the way how geometry shader emulation was done in the macOS branch, but I can just add it if needed on the geometry shader PR (I don't plan to do it on vertex anymore, so I don't think I will even need to add it back).

marysaka (approved):

gdkchan (approved):
lgtm, thanks.

PR #4885, opened 2023-05-11
GPU: Fix shader cache assuming past shader data was mapped

riperiperi:
This fixes a potential issue where a shader lookup could match the address of a previous _different_ shader, but that shader is now partially unmapped. This would just crash with an invalid region exception.

To compare a shader in the address cache with one in memory, we get the memory at the location with the previous shader's size. However, it's possible it has been unmapped and then remapped with a smaller size. In this case, we should just get back the mapped portion of the shader, which will then fail the comparison immediately and get to compile/lookup for the new one.

This might fix a random crash in TOTK that was reported by Piplup. I don't know if it does, because I don't have the game yet.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/ShaderCache.cs
@@ -591,7 +591,7 @@ namespace Ryujinx.Graphics.Gpu.Shader
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


marysaka (approved):

AcK77 (approved):

piplup55:
can't reproduce the same crash i had on master after 20 minutes of trying

PR #4888, opened 2023-05-11
amadeus: Fix wrong channel mapping check and an old typo

marysaka:
This was always going to happens, as a result quadratic would break and move index after the channel count point, effectively breaking input/output indices.

Fixes TOTK audio but also possible newer game using quadratic.

I also fixed an index issue on reverb 3d early delay.

PS: As a reminder, if this change is reused anywhere else, work should be credited as Ryujinx and not my person.


AcK77 (approved):

riperiperi (approved):

IsaacMarovitz (approved):
Sounds good to me

PR #4890, opened 2023-05-11
Fix incorrect ASTC endpoint color when using LuminanceDelta mode

gdkchan:
This fixes a bug on the ASTC decoder that caused some ASTC texture to be corrupt and have white pixels where they shouldn't. Only affects GPUs that don't support ASTC natively (which is pretty much everything other than Apple Silicon and old Intel integrated GPUs).

Reference form the official decoder: https://github.com/ARM-software/astc-encoder/blob/0a3269f9de2b457dbeb7c5c6265e96c45b64647c/Source/astcenc_color_unquantize.cpp#L229

This fixes some texture corruption in The Legend of Zelda: Tears of the Kingdom.
Before:
image
After:
image


marysaka (approved):

AcK77 (approved):
Nice catch!

PR #4892, opened 2023-05-11
Enable explicit LOD for array textures with depth compare on SPIR-V

gdkchan:
This removes what I believe to be a leftover from the GLSL backend (before it started using the shadow LOD extension). It was forcing explicit LOD to be removed when using depth-comparison with array textures, since that is not supported on unextended GLSL. AFAIK that limitation does not exist on SPIR-V.

This fixes some glitches on The Legend of Zelda: Tears of the Kingdom when using the Vulkan backend (as it's the only backend that uses SPIR-V shaders right now).
Before:
image
After:
image
I recommend testing on AMD and making sure it doesn't break.

### DIFF
--- a/src/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
+++ b/src/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
@@ -1442,14 +1442,6 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Spirv
                 return GetZeroOperationResult(context, texOp, AggregateType.FP32, colorIsVector);
             }
 
-            // This combination is valid, but not available on GLSL.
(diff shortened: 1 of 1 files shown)


AcK77 (approved):

riperiperi (approved):
I do have an AMD device, but I'm not sure what games I could test to trigger this case.

jerome-cs (approved):

jewom (approved):

marysaka (approved):

roxas0zero:
Hello @gdkchan, the issue still exists for me.

Vulkan Backend
All Default Settings
Ryujinx Ver 1.1.792
Game Version 1.1.0
AMD RX 580 Driver Version 23.4.3

ryujinx_capture_2023-05-12_15-27-56

roxas0zero:
That fixed it, thank you.

PR #4896, opened 2023-05-11
GPU: Remove swizzle undefined matching and rework depth aliasing

riperiperi:
@gdkchan pointed out that UI textures in TOTK seemed to be setting their texture swizzle incorrectly (texture was RGB but was sampling A, swizzle for A was wrong), so I determined that SwizzleComponentMatches was the problem and set on eliminating it. This PR combines existing work to select the most recently modified texture (now used when selecting which aliased texture to use) with some additional changes to remove the swizzle check and support aliased view creation.

The original observation (#1538) was that we wanted to match depth textures for the purposes of aliasing with color textures, but they often had different swizzle from what was sampled (as it's generally the identity swizzle once rendered). At the time, I decided to allow swizzles to match if only the defined components matched, which fixed the issue in all known cases but could easily be broken by a game _expecting_ a given swizzle, such as a 1/0 value on a component.

This error case could also occur in textures that don't even depth alias, such as R11G11B10, as the rule was created to generally apply to all cases.

The solution is now to fail this exact match test, and allow the search for an R32 texture to create a swizzled view of a D32 texture (and other such cases). This allows the creation of a view that mismatches the requested format, which wasn't present before and was the reason for the swizzle matching approach.

The exact match and view creation rules now follow the same rules over what textures to select when there are multiple options (such as a "perfect" match and an "aliased" match at the same time). It now selects the most recently modified texture, which is done with a new sequence number in the GpuContext, because we don't have enough of those.

Reportedly fixes UI having weird coloured backgrounds in TOTK. This also fixes an issue in MK8D where returning from a race resulted in the character selection cubemaps being broken. May work around issues introduced by the "short texture cache" PR due to modification ordering, though they won't be truly fixed.

Should allow (#4365) to avoid copies in more cases, but need to test that. May reduce copies in other games too.

I tested a bunch of games #1538 originally affected and they seem to be fine. This change affects all games so it would be good to get some wide testing on it.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
@@ -291,22 +291,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <returns>The minimum compatibility level of two provided view compatibility results</returns>
         public static TextureViewCompatibility PropagateViewCompatibility(TextureViewCompatibility first, TextureViewCompatibility second)
         {
-            if (first == TextureViewCompatibility.Incompatible || second == TextureViewCompatibility.Incompatible)
-            {
-                return TextureViewCompatibility.Incompatible;
-            }
-            else if (first == TextureViewCompatibility.LayoutIncompatible || second == TextureViewCompatibility.LayoutIncompatible)
-            {
-                return TextureViewCompatibility.LayoutIncompatible;
-            }
-            else if (first == TextureViewCompatibility.CopyOnly || second == TextureViewCompatibility.CopyOnly)
-            {
-                return TextureViewCompatibility.CopyOnly;
-            }
-            else
-            {
-                return TextureViewCompatibility.Full;
-            }
+            return (TextureViewCompatibility)Math.Min((int)first, (int)second);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
```diff
@@ -628,15 +613,21 @@ public static bool ViewLayoutCompatible(TextureInfo lhs, TextureInfo rhs, int lh
...
-        public static TextureViewCompatibility ViewFormatCompatible(TextureInfo lhs, TextureInfo rhs, Capabilities caps)
+        public static TextureViewCompatibility ViewFormatCompatible(TextureInfo lhs, TextureInfo rhs, Capabilities caps, TextureSearchFlags flags)
         {
             FormatInfo lhsFormat = lhs.FormatInfo;
             FormatInfo rhsFormat = rhs.FormatInfo;
 
             if (lhsFormat.Format.IsDepthOrStencil() || rhsFormat.Format.IsDepthOrStencil())
             {
-                return lhsFormat.Format == rhsFormat.Format ? TextureViewCompatibility.Full : TextureViewCompatibility.Incompatible;
+                return FormatMatches(lhs, rhs, flags.HasFlag(TextureSearchFlags.ForSampler), flags.HasFlag(TextureSearchFlags.DepthAlias)) switch
```
gdkchan, line 625: It seems `TextureViewCompatibility` and `TextureMatchQuality` serves the same purpose. Why not delete `TextureMatchQuality` and replace its uses with `TextureViewCompatibility`?

src/Ryujinx.Graphics.Gpu/Image/TextureInfo.cs
```diff
@@ -300,8 +300,9 @@ public SizeInfo CalculateSizeInfo(int layerSize = 0)
         /// </summary>
         /// <param name="parent">The parent texture</param>
         /// <param name="firstLevel">The first level of the texture view</param>
+        /// <param name="parentFormat">True if the parent format should be inherited</param>
         /// <returns>The adjusted texture information with the new size</returns>
-        public TextureInfo CreateInfoForLevelView(Texture parent, int firstLevel)
+        public TextureInfo CreateInfoForLevelView(Texture parent, int firstLevel, bool parentFormat)
```
gdkchan, line 305: This is a minor nitpick, but would be nice if `CreateInfoForLevelView` and `InheritFormat` followed the same naming, since they are both methods that are supposed to create a slightly different `TextureInfo` for a texture view with slightly different properties (be it a different size due to it being a different mip level, or different format).

src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
riperiperi: `TextureMatchQuality` has less unique results than `TextureViewCompatibility`, using compatibility might imply those results could be returned for exact match checking when they're meant for views. Also, "Perfect" means an exact match, whereas "Full" means it is view compatible.

src/Ryujinx.Graphics.Gpu/Image/TextureInfo.cs
riperiperi: Renamed to CreateInfoWithFormat, I also fixed `parentFormat` not properly being used in `CreateInfoForLevelView`.

IsaacMarovitz (approved):
Fixes UI in TOTK for me :)

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

riperiperi:
Don't merge this right now, there seems to be an issue causing it to try making a D32 view of R32 textures.

riperiperi:
Issue appears to be fixed.

PR #4899, opened 2023-05-12
Vulkan: Buffer Mirrors for MacOS performance

riperiperi:
This PR brings MacOS ryu up to the level of performance seen with the `macos1` build, but obviously still missing geometry shaders, etc.

## Problem

Guest GPU methods such as the constant buffer updater and inline2memory are used to load arbitrary data into buffer memory at any time. This can be done between draws in a render pass with no additional barriers. When a game uses a constant buffer update that changes a buffer used in a shader, the data must be uploaded each time it's used. Occasionally, either the driver or games queue many updates that are used by the next draw in sequence. Even we update a buffer between draws sometimes - the support buffer that contains things like texture scale.

Vulkan doesn't have an inline data update method like this - you need to be outside a render pass and either perform a buffer copy, update buffer command or compute write. Interrupting a render pass isn't too bad on desktop - you just need to pay for barriers to avoid changing the data while a draw is in progress, and a little bit for mostly useless Vulkan calls. However, on mobile GPUs (which are typically tiled) there is huge cost to ending a render pass.

On tiled GPUs, render passes load texture data from shared memory into tile memory, then store that back into shared memory when the render pass ends. This avoids straining shared memory bandwidth, at the cost of a full copy in and out at the start and end of each render pass. Games written with tiled GPUs in mind can sometimes ignore the data in shared memory (clear/ignore) and/or have it live entirely in tile memory to avoid storing back, but we don't get that luxury as all our textures must exist in shared memory and we have no idea how they will be used in the future.

The result is that consecutive inline updates end and begin the render pass repeatedly, which results in expensive loads and stores potentially each draw. The bigger and more numerous the render targets, the more issue this causes as the data must be copied back and forth every time. This obviously depends on the game, as some games don't read cbu results from the shader, but when it happens you will definitely feel it.

## Solution

To avoid both barriers and render pass splits, we can instead "mirror" entire bindings by copying their current state into a temporary buffer and binding that instead. This trades inline updates on the GPU for data copies on the CPU and binding swaps. The buffer data we make copies of ends up being way smaller than the framebuffer copies, so a lot of memory bandwidth and time is saved in the majority of cases.

On Apple Silicon GPUs this can greatly increase performance depending on the game's behaviour.

## Limitations

- Buffers that have been written from the GPU side (storage, buffer copy etc) cannot be mirrored is that write is still in-flight. The copies are made from a mapping of the buffer in the host address space, so we need to make sure no pending changes are present that might affect the buffer before making a copy.
  - When this happens _after_ a mirror has been created, all bindings of this mirror must be replaced with the original buffer, as that data may be intended for an existing binding.
  - If another mirrorable write happens after a mirror has been created and bound, the binding is recreated with a new mirror.
- Buffers that are not mapped in the host address space cannot be mirrored. This mostly affects desktop GPUs as buffers are always mapped on systems with shared memory. Mirrors are also mostly disabled on desktop GPUs anyways, so this isn't really an issue.
- Mirrors are considered valid until the next command buffer. They may be freed after the command buffer ends, as their fence on the staging buffer will be met when it completes. As a result, mirrors are recreated whenever the pipeline command buffer is submitted.

## Misc changes
- GetRange has been replaced with GetRangeAligned. This will get a range of a buffer that is aligned to a page size rather than getting the whole buffer range. This is used when mirroring read-only storage buffers, as otherwise they can use bindings much larger than requested.
  - This was originally done to handle some games (Yokai Watch #, can't remember which) reading outside of their binding range. This limits this to only be valid within the page the end address is contained within, which should be a good assumption but should probably be tested anyways.
- Bindings sent to the backend now include a write flag. This may actually avoid some existing Vulkan only issues where buffer usage rules might not be respected, but is much more important for mirrors to function correctly.
- BitMapStruct has been added as a bitmap that operates on a struct. This improves memory locality as the entire bitmap can be embedded within another class or struct, the only downside is that the bit array must be of a predefined size.
- There are changes to the DescriptorSetUpdater to accelerate batch operations on uniform/storage set booleans. This may change by the time this PR is undrafted.
- The staging buffer has gained a method that allows reserving a range of memory for any purpose. This is currently used for placing mirrors on the staging buffer. They are considered valid until the next command buffer, obviously.
- The staging buffer size has been increased as it can now hold mirrors.


iRonJ:
I'm seeing a guest broke execution exception with this, kernel 16.0.2. This also is broken with current main, but not the older release
```
00:00:03.599 |E| HLE.GuestThread.40 Application : Unhandled exception caught: Ryujinx.HLE.Exceptions.GuestBrokeExecutionException: The guest program broke execution!
   at Ryujinx.HLE.HOS.Kernel.SupervisorCall.Syscall.Break(UInt64 reason) in /Users/ronj/Applications/Ryujinx/src/Ryujinx.HLE/HOS/Kernel/SupervisorCall/Syscall.cs:line 1878
   at Ryujinx.HLE.HOS.Kernel.SupervisorCall.SyscallDispatch.Break64(Syscall syscall, IExecutionContext context) in /Users/ronj/Applications/Ryujinx/src/Ryujinx.HLE/Ryujinx.Horizon.Kernel.Generators/Ryujinx.Horizon.Generators.Kernel.SyscallGenerator/SyscallDispatch.g.cs:line 1807
```
M1 MBP

iRonJ:
On an 13" M1 MBP, this works well for me. Get about ~7FPS more than the current main branch, still under 30fps but still playable.

jcm93:
Thank you for posting this! It seems to stutter slightly more severely than macos1, but fixes many issues, and once the stutters level off, 30fps is very stable. Appreciate it and look forward to the rest of the upstreaming.

CestLucas:
Thank you. When can I expect the merge to the main channel release?

Téléchargez Outlook pour iOS<https://aka.ms/o0ukef>
________________________________
De : jcm ***@***.***>
Envoyé : Friday, May 12, 2023 1:03:38 PM
À : Ryujinx/Ryujinx ***@***.***>
Cc : Lucas Gao ***@***.***>; Manual ***@***.***>
Objet : Re: [Ryujinx/Ryujinx] Vulkan: Buffer Mirrors for MacOS performance (PR #4899)


Thank you for posting this! It seems to stutter slightly more severely than macos1, but fixes many issues, and once the stutters level off, 30fps is very stable. Appreciate it and look forward to the rest of the upstreaming.

iRonJ:
> Thank you for posting this! It seems to stutter slightly more severely than macos1, but fixes many issues, and once the stutters level off, 30fps in TotK is very stable (on my M1 Max, at least). Appreciate it and look forward to the rest of the upstreaming.

I have an M1 Max and it seems to run at the same speed as my standard M1. I tried one of the 30FPS mods but no adjustment. If i turn off VSYNC it is a constant 30+ FPS, but turn it back on, and it locks to 20FPS... any ideas?

JampaUchoa:
Got one crash using this PR, could not isolate what caused it, but here is the log:

[Ryujinx_1.1.1+0_2023-05-12_17-56-21.log](https://github.com/Ryujinx/Ryujinx/files/11468776/Ryujinx_1.1.1%2B0_2023-05-12_17-56-21.log)

dagstuan:
This PR should probably be added to #4062?

morceaudebois:
Tried for a few minutes and it seems to run a tiny bit smoother on M1 Pro. The depths are still about 20fps though and the ground glitches a little in that area.

Will keep testing and report anything interesting :)

nmpt88:
Thank you for your efforts for Mac players, it has quite a significant efficiency boost on my M1 Pro, I wonder if you can fix the white spot issue that has been present on Mac, thanks again

MCChristor:
The oldest Version is the fastest for me in TOTK. 10-15fps more then the newest.

morceaudebois:
So I got rid of my shader cache (I had been using the same for like three weeks), and I can now run the game at a relatively stable 30fps pretty much everywhere, which finally makes the depths playable! 🎉

Apart from all the glitches, this is starting to be a pretty good experience!

meritozh:
@coyote0770 just out of curiosity, do you have this issue? https://github.com/Ryujinx/Ryujinx/issues/4949

coyote0770:
> @coyote0770 just out of curiosity, do you have this issue? #4949

The white glowing stuff, yes...sometimes...The hair thing wiggin out in the menus, yes, unless Link is wearing a headband for whatever reason. The shadows of Link on the ground are messed up like that still, yes, I never even noticed it until recently when my gf pointed it out. I did have the chuchus glitch out weird on previous ryujinx versions but it seems okay now.  Most of these things don't bother me as it doesn't affect gameplay. Im just happy to have it running and running smoothly on my base-tier mac :)

jcm93:
Looks like the latest rebase has significant regressions with culling and actor updates.

riperiperi:
GPUs don't work in "culling and actor updates", can you be more specific?

jcm93:
Sure, pardon the discord links, but here are some clips from the latest CI builds (7d970aa): https://cdn.discordapp.com/attachments/1045857828360949830/1110278475861987368/totk.mp4

https://cdn.discordapp.com/attachments/1045857828360949830/1110279509464338523/totk2.mp4

meritozh:
test in Fire Emblem: Three Houses. have one strange rendering issue:

https://github.com/Ryujinx/Ryujinx/assets/7543300/c67a92fe-003e-439d-8e3b-3f6942cef19d

confirmed it not exist in release [1.1.819](https://github.com/Ryujinx/Ryujinx/releases/tag/1.1.819)

cobbycobbycobcob:
Related to #5103 ?

CestLucas:
Another issue with the latest build (c23b27b) is that TOTK will not quit properly (have to force quit each time), does not happen with 1.1.826

CestLucas:
On the latest build Ryujinx does not close when quitting TOTK, have to rely on force quit

morceaudebois:
> On the latest build Ryujinx does not close when quitting TOTK, have to rely on force quit

I’ve actually had that problem since basically forever (only with TOTK though)

iRonJ:
just want to say i'm really in awe at the elegance of the shader fix and the geometry bug fix.

siegmund-heiss-ich:
Hey peri!
Found a reproducable crash (regression) on this pr.
On master it seems to work fine.
But if I start Mario Party Superstars with this pr, it crashes as soon as I press L+R to basically start the game.
The settings are all standard. (Hypervisor on, native res, and so on...)

These are the last lines of the log:

```
00:00:43.520 |E| GUI.RenderThread Application : Unhandled exception caught: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Ryujinx.Graphics.Vulkan.BufferHolder.TryGetMirror(CommandBufferScoped cbs, Int32& offset, Int32 size, Auto`1& buffer) in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs:line 335
   at Ryujinx.Graphics.Vulkan.PipelineBase.RecreatePipelineIfNeeded(PipelineBindPoint pbp) in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs:line 1627
   at Ryujinx.Graphics.Vulkan.PipelineBase.Draw(Int32 vertexCount, Int32 instanceCount, Int32 firstVertex, Int32 firstInstance) in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs:line 397
   at Ryujinx.Graphics.GAL.Multithreading.CommandHelper.<>c.<InitLookup>b__5_1[T](Span`1 memory, ThreadedRenderer threaded, IRenderer renderer) in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs:line 41
   at Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.RunLoop(Action gpuLoop) in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs:line 116
   at Ryujinx.Ava.AppHost.RenderLoop() in /Users/aloispetutschnig/gitProjects/Ryujinx/src/Ryujinx.Ava/AppHost.cs:line 883
   ```

**Comment by gdkchan** [comment]
It does seem to be an issue with this PR, but from what I recall this crash is random, which makes it harder to debug.

**Comment by siegmund-heiss-ich** [comment]
I checked 7 times now and it always happened at the same point.
1 second after L+R is pressed.

**Comment by iRonJ** [comment]
Seems like the latest push on this fixes the stop emulation function?

**Comment by iRonJ** [comment]
Subjectively, the last commit seems noticeably "smoother". frame rates seem about the same, but the responsive feels better if that makes any sense... maybe just all in my head though.

**Comment by areidt** [comment]
Newest change to this PR has reintroduced the striped gloom bug in the depths in TOTK.

**Comment by phush0** [comment]
What I see last change is rebase so probably regression in main

**Comment by akedidi** [comment]
Hi, it's seems that it's not possible anymore to get GUI experimental release because of timeout.
I would like to download last release for macos, is it possible to make it available again?

**Comment by iRonJ** [comment]
> You can no longer download this release from this website. What's going on? Just times out.

I’ve just been building it locally, but where are you downloading it from?

**Comment by morceaudebois** [comment]
Not sure what's happening but I uploaded the release here if you'd like to download it: https://drive.proton.me/urls/QGQTS4RDA0#WlUN4BDEFU4b

It's the one just before the current one (7eddd3b), but it seems to be the better working one according to @areidt.

**Comment by iRonJ** [comment]
Running the Sonoma beta on my 13" MBP M1 and it definitely feels like I'm getting 2-4 more FPS. When I entered full screen using the menu item, a message popped up saying I was now running in Game Mode. I happened to enter a Shrine in TOTK at the same time and the Shrine was butter smooth. 

I can't update my M1 Max to Sonoma but I'm expecting it would get a similar boost. Really feels like this HW could get to a consistent 30fps though -- fingers crossed

**Comment by phush0** [comment]
TOTK is running ~45 fps on M1 Max with this PR

**Comment by gssci** [comment]
Using this as my main way to play totk at the moment.
Running at around 1440 w/ dyn. resolution disabled at 40fps everywhere (undergrounds included) 
on my 8-core M2 Pro Mac Mini 2023, 16GB

**Comment by iRonJ** [comment]
Latest commit is very smooth, 13" M1 MBP Sonoma:
image

Interestingly, in the menu screen the processing times get a little weird:
image

**Comment by KindofNihilist** [comment]
> Latest commit is very smooth, 13" M1 MBP Sonoma: image
> 
> Interestingly, in the menu screen the processing times get a little weird: image
> 
> Also another weird thing, it says it's rendering in "composited" mode instead of "direct", seems like games should generally be direct but not sure about that for MVK...

What mods do you use?

**Comment by iRonJ** [comment]
> > Latest commit is very smooth, 13" M1 MBP Sonoma: image
> > Interestingly, in the menu screen the processing times get a little weird: image
> > Also another weird thing, it says it's rendering in "composited" mode instead of "direct", seems like games should generally be direct but not sure about that for MVK...
> 
> What mods do you use?

Dynamic FPS/60FPS only

**Comment by iRonJ** [comment]
newest build subjectively seems to have slightly more graphical glitches than the previous. Seems like the whole screen will glitch out every 3-5 seconds for a single frame in TOTK on M1 Max

**Comment by kavilee** [comment]
newest build TOTK crash sometimes

**Comment by nvcuong1312** [comment]
Screenshot 2023-06-13 at 18 37 30
I got some graphic glitches on newest build.

**Comment by riperiperi** [comment]
As far as I'm aware, there were always issues, If there is an increased number of them, can someone actually verify a difference between builds? (you can check past builds using the edit history of the comment)

**Comment by gssci** [comment]
The latest build is actually running with less issues and crashes for me, using mods. 
No crashes to report yet, but will post logs if they happen

**Comment by AMP-808** [comment]
Same here, I used to have intermittent pink flashes on screen, but that is all but gone now. I do still experience hangs in the Depths where the game will be autosaving with sound playing. These take ~30s-1min. before the resolve and the game resumes with no issue.

Hardware:
16" MBP - M1 Max - 32GB Memory

Software:
macOS 14 Sanoma - Gaming Mode Enabled

**Comment by kavilee** [comment]
> 

03:35:36.099 |I| HLE.OsThread.16 ServicePrepo ProcessPlayReport: 
PlayReport log:
 Kind: Normal
 Pid: 98
 UserId: 00000000000000010000000000000000
 Room: event_player_start
 Report: {
    "PlayTime": 187256,
    "ReleaseVersion": 1333,
    "PlayDataUniqueID": 1116685400,
    "EventInfo": "Npc_AncientAssistant003_C1_CT",
    "PlayerPosX": 4210.675,
    "PlayerPosY": 313.44983,
    "PlayerPosZ": -1361.402,
    "system_report_tag": 2175206328
}

03:37:51.783 |E| HLE.GuestThread.70 Application : Unhandled exception caught: Ryujinx.Memory.InvalidMemoryRegionException: va=0x5E00B3FF4E00FFFF, size=0x0000000000000001
   at Ryujinx.Cpu.AppleHv.HvMemoryManager.AssertValidAddressAndSize(UInt64 va, UInt64 size) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvMemoryManager.cs:line 125
   at Ryujinx.Cpu.AppleHv.HvMemoryManager.IsRangeMapped(UInt64 va, UInt64 size) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvMemoryManager.cs:line 427
   at Ryujinx.Memory.Tracking.MemoryTracking.VirtualMemoryEvent(UInt64 address, UInt64 size, Boolean write, Boolean precise, Nullable`1 exemptId) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Memory/Tracking/MemoryTracking.cs:line 281
   at Ryujinx.Cpu.AppleHv.HvExecutionContext.DataAbort(MemoryTracking tracking, UInt64 vcpu, UInt32 esr) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvExecutionContext.cs:line 227
   at Ryujinx.Cpu.AppleHv.HvExecutionContext.SynchronousException(HvMemoryManager memoryManager, HvVcpu& vcpu) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvExecutionContext.cs:line 169
   at Ryujinx.Cpu.AppleHv.HvExecutionContext.Execute(HvMemoryManager memoryManager, UInt64 address) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvExecutionContext.cs:line 154
   at Ryujinx.Cpu.AppleHv.HvCpuContext.Execute(IExecutionContext context, UInt64 address) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.Cpu/AppleHv/HvCpuContext.cs:line 30
   at Ryujinx.HLE.HOS.ArmProcessContext`1.Execute(IExecutionContext context, UInt64 codeAddress) in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.HLE/HOS/ArmProcessContext.cs:line 50
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart() in /home/runner/work/Ryujinx/Ryujinx/src/Ryujinx.HLE/HOS/Kernel/Threading/KThread.cs:line 1266



I found the latest crash log, my device :  macbook pro 14    M1 pro + 16GB RAM

**Review by borisleehk — APPROVED** [review]

**Comment by phush0** [comment]
let's goooo

**Review by borisleehk — APPROVED** [review]

**Comment by Shihta** [comment]
> I checked 7 times now and it always happened at the same point.
> 1 second after L+R is pressed.

I have the same issue... it always crash after press L+R In this starting screen
截圖 2023-06-22 下午9 47 17

my environment:
OS: macOS 12.5.1
Ryujinx version: 1cb076106226e76d41d7e47a8ece397989c1d25e
Game version: Mario Party Superstars 1.0.0
CPU: Apple M2
GPU: Apple M2
RAM: 24G

log: [Ryujinx_1.1.1+0_2023-06-22_21-44-36.log](https://github.com/Ryujinx/Ryujinx/files/11834672/Ryujinx_1.1.1%2B0_2023-06-22_21-44-36.log)

**Comment by lbibass** [comment]
After pulling this in to the most recent master, performance noticeably increased, especially after clearing my shader cache. I went from below 20 FPS in the depths on an M1 Max to above 30.

**Comment by Shihta** [comment]
@riperiperi I built a Debug version to run the Mario Party Superstars, and It finally correctly shows that crash's location is at: https://github.com/Ryujinx/Ryujinx/blob/1cb076106226e76d41d7e47a8ece397989c1d25e/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs#L364

then I try to print some variables:
```
00:01:06.000 |I| GUI.RenderThread Gpu : _pendingData.Length=4096, offset=3072, size=48576
```

I'm not sure how to correct this, but I hope this information will be useful to you.

Log: [Ryujinx_1.0.0-dirty_2023-06-24_15-32-48.log](https://github.com/Ryujinx/Ryujinx/files/11855944/Ryujinx_1.0.0-dirty_2023-06-24_15-32-48.log)

**Comment by ilyassemokhtari** [comment]
not to be ungrateful. but the 2 or 3 last version are far worse in terms of FPS, stutters and framedorps. that why I'm using the Version:8857f54. not sure why ?

**Comment by riperiperi** [comment]
The last two versions are literally just rebases on top of changes in master. If this is an actual concern, do objective testing.

**Comment by Skyfied** [comment]
Depths are unplayable, Walls everywhere
Screenshot 2023-06-25 at 12 21 28 PM
Screenshot 2023-06-25 at 12 21 40 PM
Screenshot 2023-06-25 at 12 22 03 PM

**Comment by onelightinthedark** [comment]
> Depths are unplayable, Walls everywhere
> Screenshot 2023-06-25 at 12 21 28 PM
> Screenshot 2023-06-25 at 12 21 40 PM
> Screenshot 2023-06-25 at 12 22 03 PM

depths work fine for me on 1.1.0+e8af78c, totk 1.1.2, m1 mbp, ventura 13.4. if anything, performance is better than it's ever been.

not to sound harsh, but if you're having issues, just posting screenshots doesn't help anyone sort the issue. what exact version of ryujinx are you using, and what settings? what game version? what hardware and os are you running? have you tried clearing your shader cache, disabling mods, etc.? posting logs would help too.

**Comment by riperiperi** [comment]
I've fixed the crash that happens on Mario Party Superstars. It's trying to mirror a vertex buffer that has had its handle replaced. This might cause an additional mirror to be created in some cases, though - maybe these should be nullified somehow when the handle is disposed.

**Comment by Shihta** [comment]
> I've fixed the crash that happens on Mario Party Superstars. It's trying to mirror a vertex buffer that has had its handle repurposed. This might cause an additional mirror to be created in some cases, though - maybe these should be nullified somehow when the handle is disposed.

Thank you! it can work normally now!

**Comment by tgcordell** [comment]
Just had a crash running this build (1.1.0+d7fdae4) I wanted to report. Overall, the build has been awesome. Excellent FPS everywhere. I was able to capture the Apple crash log that popped up, but it exceeds Pastebin's max size of 512kb per paste. Is there a preferred way to share it? I also assume theres a specific Ryujinx log that would be helpful, but this is my first crash/bug report so I'm not sure where/what that is. Let me know and Im happy to provide it.

Ryujinx Crash 07/01/2023 @ 12:16pm PST
Switch Firmware: 16.0.3
TOTK version: 1.1.0
Ryujinx: 1.1.0+d7fdae4
MacOS: 13.4.1
Laptop: MacBook Pro 14”
CPU: M1 Pro
Ram: 16GB
Backend: Vulkan
Mods/Cheats/etc: None installed
All other system configs: default

**Comment by zenaufa** [comment]
> I've fixed the crash that happens on Mario Party Superstars. It's trying to mirror a vertex buffer that has had its handle repurposed. This might cause an additional mirror to be created in some cases, though - maybe these should be nullified somehow when the handle is disposed.

Thank you so much for this build!  
I just installed this build, and it has a massive fps boost for all the games I've played.

**Comment by Junes00** [comment]
[Ryujinx_1.1.0+e8af78c_2023-07-04_15-27-29.log](https://github.com/Ryujinx/Ryujinx/files/11949878/Ryujinx_1.1.0%2Be8af78c_2023-07-04_15-27-29.log)
logs report of the last periperi build crashing after charging the shaders of TOTK on macbook air m1 13, on MacOs 12.6.7 (monterey)

**Comment by Shihta** [comment]
I use the latest version to run `Mario + Rabbids Sparks of Hope`, it shows a lot of `no space :(` on in the console
Whether I check `Use Hypervisor` or not, it always shows `no space :(`

some screenshot, use JIT:
截圖 2023-07-06 上午8 35 06

use hypervisor:
截圖 2023-07-06 上午8 36 30

my environment:
OS: macOS 12.5.1
Ryujinx version: https://github.com/riperiperi/Ryujinx/commit/8f0d456b3ec7163986174d58bf80fb1eddf935d2
Game version: Mario + Rabbids Sparks of Hope 1.5.2205169
CPU: Apple M2
GPU: Apple M2
RAM: 24G

Log: [Ryujinx_1.0.0-dirty_2023-07-05_23-07-51.log](https://github.com/Ryujinx/Ryujinx/files/11963073/Ryujinx_1.0.0-dirty_2023-07-05_23-07-51.log)

**Comment by iRonJ** [comment]
On Sonoma, the compile shaders step got significantly faster in this latest build.

**Comment by riperiperi** [comment]
"no space :(" is a debug log printed when the staging buffer doesn't have enough space to allocate a mirror. It'll fall back to not using mirrors if this happens. It's not fatal or a bug when this happens.

Personally I think any case where it is creating 128MB of mirrors during command execution is probably not going to benefit much from them, as the CPU will spend so much time copying data for the mirrors that you'll lose a lot of performance. I don't have sparks of hope so I don't know how it performs either way.

File: src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs [hunk 1]
```diff
@@ -130,13 +145,83 @@ public bool TryPushData(CommandBufferScoped cbs, Action endRenderPass, BufferHol
...
+
+            // Temporary reserved data cannot be fragmented.
+
+            if (GetContiguousFreeSize(alignment) < size)
+            {
+                FreeCompleted();
+
+                if (GetContiguousFreeSize(alignment) < size)
+                {
+                    Console.WriteLine("no space :(");
```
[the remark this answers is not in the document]
riperiperi: Now uses Logging. Currently Warning while I check a few things.

src/Ryujinx.Graphics.Vulkan/BufferRangeList.cs
```diff
@@ -0,0 +1,271 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    struct BufferRangeList
```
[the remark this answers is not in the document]
gdkchan: Maybe `FillData` can just be moved somewhere else? As far I can tell, this is the only method that makes this not just a range list.

Georgi-Canary:
Would we get a new build with the recent changes?

artrelated:
Would we get a new build with the recent changes?

Shihta:
[Ryujinx_1.1.1+0_2023-08-03_13-44-47.log](https://github.com/Ryujinx/Ryujinx/files/12247341/Ryujinx_1.1.1%2B0_2023-08-03_13-44-47.log)

Here is some information about the `Sparks of Hope`. It seems that it often requires a buffer close to 1MB.

riperiperi:
I feel like that's a bit excessive. I don't have the game so I can't check the reason it's doing that.

src/Ryujinx.Graphics.Vulkan/BufferRangeList.cs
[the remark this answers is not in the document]
riperiperi: I don't think any of the existing RangeList implementations really match this one, so I won't refactor it for now.

- BufferRangeList (probably going to be called BufferMirrorRangeList)
  - More like a sparse usage bitmap, on/off.
- RangeList (Ryujinx.Memory)
  - More like entities with properties and methods that occupy a range, rather than just on/off.
  - NonOverlappingRangeList - can be used to implement this, but is better suited to things like the memory tracking
  - BufferModifiedRangeList - used to track modified ranges, but splits so that different sync numbers for flushing data are respected.

src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
```diff
@@ -339,6 +424,86 @@ public Auto<DisposableBuffer> GetBuffer(CommandBuffer commandBuffer, int offset,
...
+            {
+                _mirrors.Clear();
+            };
+        }
+
+        public void ClearMirrors(CommandBufferScoped cbs, int offset, int size)
+        {
+            // Clear mirrors in the given range, and submit overlapping pending data.
+
+            // TODO: within range. for now it just clears everything.
```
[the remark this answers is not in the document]
riperiperi: Done, everything seems fine in a few games I tried.

src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
```diff
@@ -1,12 +1,27 @@
+using Ryujinx.Common;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace Ryujinx.Graphics.Vulkan
 {
+    struct StagingBufferReserved
```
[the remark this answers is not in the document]
riperiperi: I think it's fine here. Probably.

src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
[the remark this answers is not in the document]
riperiperi: This should change to Debug before merging. Don't want anyone thinking this is an error.

src/Ryujinx.Graphics.Vulkan/BitMapStruct.cs
```diff
@@ -0,0 +1,263 @@
...
+            {
+                _masks[startIndex] |= startMask & endMask;
+            }
+            else
+            {
+                _masks[startIndex] |= startMask;
+
+                for (int i = startIndex + 1; i < endIndex; i++)
+                {
+                    _masks[i] |= -1;
```
gdkchan, line 214:
```suggestion
                    _masks[i] |= -1L;
```

gdkchan: Doesn't really matter since the value will be sign extended, just a nitpick.

src/Ryujinx.Graphics.Vulkan/BufferMirrorRangeList.cs
```diff
@@ -0,0 +1,305 @@
...
+
+            public bool OverlapsWith(int offset, int size)
+            {
+                return Offset < offset + size && offset < Offset + Size;
+            }
+        }
+
+        private List<Range> _ranges;
+
+        public readonly List<Range> All()
```
gdkchan, line 34: Is it possible to make it return `IEnumerable<Range>`? It's a bit weird that it allows the list to be accessed directly since the caller could modify the list which could break assumptions this class makes internally.

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
```diff
@@ -353,13 +445,26 @@ public void UpdateAndBindDescriptorSets(CommandBufferScoped cbs, PipelineBindPoi
...
+            int offset = buffer.Offset;
+            bool mirrored = false;
+
+            if (mirrorable)
+            {
+                info.Buffer = buffer.Buffer?.GetMirrorable(cbs, ref offset, (int)info.Range, out mirrored).Value ?? default;
+            }
+            else
+            {
+                info.Buffer = buffer.Buffer?.Get(cbs, offset, (int)info.Range, buffer.Write).Value ?? default;
```
gdkchan, line 464: Instead of using `buffer.Write` that comes from `BufferRange`, is it not possible to pass `segment.Access.HasFlag(ResourceAccess.Write)` from `UpdateAndBind` loop? That way, introducing the `Write` flag on `BufferRange` is not necessary (and it would be more efficient, since the binding information only needs to be set once per shader, while those buffer ranges are updated frequently, so making them as slim as possible is a good thing).

src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
```diff
@@ -130,13 +146,83 @@ public bool TryPushData(CommandBufferScoped cbs, Action endRenderPass, BufferHol
...
+
+            // Temporary reserved data cannot be fragmented.
+
+            if (GetContiguousFreeSize(alignment) < size)
+            {
+                FreeCompleted();
+
+                if (GetContiguousFreeSize(alignment) < size)
+                {
+                    Logger.Warning?.PrintMsg(LogClass.Gpu, $"Staging buffer out of space to reserve data of size {size}.");
```
gdkchan, line 218: I guess this is the log you want to change to debug before merge?

riperiperi: Yes, it should become debug.

src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
riperiperi: This didn't exist when this branch was created. I'll see if I can simplify things with it.

riperiperi: `SetStorageBuffers` relies on `BufferRange`.Write to properly track write usage for invalidating converted index/vertex buffers with GetBuffer - the segment isn't available from there. This was an error in master where storage writes wouldn't actually invalidate converted buffers.

The segment could be used in `DescriptorSetUpdater` when doing `_gd.BufferManager.GetBuffer` instead of the `BufferRef` though.

gdkchan: I wonder if there is any point on `ResourceAccess` information then. I added it to allow the backend to know how buffers are used, but if that information is going to be passed on `BufferRange` too, it becomes redundant. I wanted to remove the `ShaderStage` parameter from `SetTextureAndSampler` too, but I ran into the same problem. Getting it from the pipeline layout requires making an assumption that the correct shader is already bound by the time those set methods are called, which is not a good assumption, or delaying the use of the information until before draw, where we can be sure the correct shader is set.

gdkchan: I think it's better to keep it as it is. I don't think there's any benefit in using both, I can just remove `ResourceAccess` later which will make creating the `ResourceLayout` a bit simpler. Nothing else needs it anyway.

src/Ryujinx.Graphics.Vulkan/BufferMirrorRangeList.cs
riperiperi: This actually isn't used anymore, but it might be good to keep around as `IEnumerable<Range>` for future.

src/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
```diff
@@ -372,15 +372,27 @@ public void ClearBuffer(MemoryManager memoryManager, ulong gpuVa, ulong size, ui
...
 
         /// <summary>
-        /// Gets a buffer sub-range starting at a given memory address.
+        /// Gets a buffer sub-range for a given GPU memory range.
+        /// </summary>
+        /// <param name="memoryManager">GPU memory manager where the buffer is mapped</param>
+        /// <param name="gpuVa">Start GPU virtual address of the buffer</param>
+        /// <param name="size">Size in bytes of the buffer</param>
+        /// <returns>The buffer sub-range for the given range</returns>
+        public BufferRange GetGpuBufferRange(MemoryManager memoryManager, ulong gpuVa, ulong size)
```
gdkchan, line 381: Is that method unused? Can't find anything tha calls it.

riperiperi: I don't know how this ended up here.

gdkchan (approved):
lgtm, thanks! Really impressive the performance improvement this brings to some titles.

tgcordell:
Crash just occurred running the latest build - 1.1.0+e3fcd35. 

Log attached. 

Ryujinx Crash 07/01/2023 @ 12:16pm PST
Switch Firmware: 16.0.3
[Ryujinx_1.1.0+e3fcd35_2023-08-12_15-56-22.log](https://github.com/Ryujinx/Ryujinx/files/12328204/Ryujinx_1.1.0%2Be3fcd35_2023-08-12_15-56-22.log)

TOTK version: 1.1.0
MacOS: 13.5
Laptop: MacBook Pro 14”
CPU: M1 Pro
Ram: 16GB
Backend: Vulkan
Mods/Cheats/etc: None installed
All other system configs: default

marysaka (approved):

benjaminmordaunt:
This seems to have introduced graphical issues in Pikmin 4 (MacBook Pro M1):

image

solarmystic:
As I commented on ryujinx pr-testing discord, this PR seems to cause intermittent frametime spikes in SMO like clockwork after being merged into mainline 1.1.988. There were none on the previous mainline 1.1.987 before this PR got merged. 

1.1.988 - 140 Average FPS/106 1% FPS/16 0.1% FPS (large frametime spike captured in screenshot below) 
SMOza11988M 12700KF (527 56, -140 6mV) Ryujinx 1X Auto MT, All cores HT OFF, ALT Sch 3466 TUNED

1.1.987 - 139 Average FPS/105 1% FPS/99 0.1% FPS (smooth frametimes with no spikes whatsoever)
SMOza11987M 12700KF (527 56, -140 6mV) Ryujinx 1X Auto MT, All cores HT OFF, ALT Sch 3466 TUNED

PR #4903, opened 2023-05-12
Fixed "ErrorOutOfDeviceMemory" crash.

bule2015:
"_api.AllocateMemory()" seems to fail sometimes and succeeds the second time.

Fixed crash in the following image.
EroorOutOfDeviceMemory

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/MemoryAllocatorBlockList.cs
+++ b/src/Ryujinx.Graphics.Vulkan/MemoryAllocatorBlockList.cs
@@ -207,7 +207,14 @@ namespace Ryujinx.Graphics.Vulkan
                 MemoryTypeIndex = (uint)MemoryTypeIndex
             };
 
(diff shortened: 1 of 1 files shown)


Nerboruto:
try to temporarily allocate system memory when it fails? some games exceed 2gb vram when played for a long time.
better a slowdown than a crash.

marysaka:
That doesn't change anything to the actual code logic, closing as suspect of spam account.

marysaka:
@bule2015

That "fix" in its form will just throw considering your code change.

This is not valid.

Because we had an influx of spams and fake builds around, I wrongly treated this as spam.

Anyway, I would prefer that you don't send me emails for that and contact us on discord.

bule2015:
I would be satisfied if this crash, which occurs within an hour, could be eliminated.

@Nerboruto
Allocating to system memory when a call fails is a good implementation that does not depend on a specific environment.
In my environment, it succeeds by calling it again immediately after the failure, but I am not sure why it does so.

PR #4910, opened 2023-05-12
macOS CI Adjustments

IsaacMarovitz:
- macOS builds were not being named correctly; as a result, the updater couldn't find them. This PR brings them in line with the convention for Ava builds.
- Removes publishing for x86 Mac builds on PR runner.
- Adjust capitalisation `MacOS` -> `macOS`, `universal` -> `Universal`

### DIFF
--- a/distribution/macos/create_macos_build.sh
+++ b/distribution/macos/create_macos_build.sh
@@ -22,7 +22,7 @@ EXTRA_ARGS=$8
 
 if [ "$VERSION" == "1.1.0" ];
 then
-  RELEASE_TAR_FILE_NAME=Ryujinx-$CONFIGURATION-$VERSION+$SOURCE_REVISION_ID-macos_universal.app.tar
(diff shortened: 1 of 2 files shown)


distribution/macos/create_macos_build.sh
```diff
@@ -20,13 +20,7 @@ SOURCE_REVISION_ID=$6
 CONFIGURATION=$7
 EXTRA_ARGS=$8
 
-if [ "$VERSION" == "1.1.0" ];
-then
-  RELEASE_TAR_FILE_NAME=Ryujinx-$CONFIGURATION-$VERSION+$SOURCE_REVISION_ID-macos_universal.app.tar
-else
-  RELEASE_TAR_FILE_NAME=Ryujinx-$VERSION-macos_universal.app.tar
-fi
-
```
TSRBerry, line 29: You can't remove that. PR builds should still follow the correct naming scheme.
You only need to change the `RELEASE_TAR_FILE_NAME` for the release builds (within the else case).

TSRBerry (changes requested):
Aside from the PR build naming, lgtm!

TSRBerry (approved):
Lgtm!

marysaka (approved):

PR #4911, opened 2023-05-12
Vulkan: Device map buffers written more than flushed

riperiperi:
This PR retunes the buffer migration system to map buffers that are written significantly more than they are read back on device, rather than on host.

NVIDIA GPUs seem to have a lot of trouble mapping a storage buffer for write in a ton of places throughout a frame, probably because it has to ensure coherency every time. Unfortunately we can't seem to disable that, so using host buffers always has a large hidden cost when using a storage buffer in a lot of places. TOTK does a lot of storage writes from _fragment_ of all things, in many different places.

This causes a small performance regression in BOTW, but a big performance boost in TOTK. This is due to reading from device memory being uncached. There may be a better solution to this problem.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
+++ b/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
@@ -56,6 +56,7 @@ namespace Ryujinx.Graphics.Vulkan
         private int _writeCount;
         private int _flushCount;
         private int _flushTemp;
+        private int _lastFlushWrite = -1;
 
         private ReaderWriterLock _flushLock;
         private FenceHolder _flushFence;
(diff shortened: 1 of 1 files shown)


Bjorn29512:
Tested like 5 games : 

(I would like to test this PR more extensively at some point but rn i just wanna play TOTK.)

Key Observations : 
    

   -  Massive Performance Gain in TOTK in GPU bottlenecked scenarios with Vulkan: 15 fps avg increase.
       
       Master: 
       image
       
       PR: 
       image
  
       PR 2x Scaling: 
       image

   - Significant Performance regression in BOTW: 8 frames of fall
     
     Master:
     image
     
     PR: 
     image

  - No change noticed in any of the 3 Xenoblade games tested

marysaka (approved):

gdkchan:
Seems unfortunate that the change is making BOTW slower. Is there no way to tune heuristics for TOTK and not regress BOTW?

Foul-Tarnished:
Vulkan now performs slightly better than OGL on a gtx 1060 6gb

gdkchan (approved):
Approving since the amount of people using this to play Tears of the Kingdom right now should far out weight those trying to play Breath of the Wild (and they have more options to play the latter). But we should probably look into a different way to do this without the performance regression in the future.

jacky400:
.Just tested Mario Spark of Hope with vulkan. It run a lot faster now.

Bjorn29512:
Tested a few more titles

> Tested like 5 games :
> 
> (I would like to test this PR more extensively at some point but rn i just wanna play TOTK.)
> 
> Key Observations :
> 
>     * Massive Performance Gain in TOTK in GPU bottlenecked scenarios with Vulkan: 15 fps avg increase.(even more depending upon the setup)
>       Master:
>       image
>       PR:
>       image
>       PR 2x Scaling:
>       image
> 
>     * Significant Performance regression in BOTW: 8 frames of fall
>       Master:
>       image
>       PR:
>       image
> 
>     * No change noticed in any of the 3 Xenoblade games tested


Tested a few more titles : 

- No change was noticed compared to master in the following
   
     - Advance Wars 1+2: Re-Boot Camp
     - Bayonetta 3 
     - Fire Emblem: Engage
     - Pokemon Violet
     - Super Mario Odyssey
     - Shin Megami Tensei 5

tommy-sorce:
Literally playing botw right now... I guess gods have punished me for something 😔 I should visit a church and pray for no more Breath of the wild performance regression.

Niwu34:
Tested a few games including Acnh, HwAoc, SMO, Mario Tennis, Metroid Dread and Monster Hunter rise and saw no big regressions!

Belerus:
how do I download this?

MetrosexualGarbodor:
@Belerus this is already on the main builds, just update Ryujinx.

PR #4927, opened 2023-05-13
[GUI] Fix always hide cursor mode not hiding the cursor until it was moved

TSRBerry:
This PR fixes a bug where the hide cursor mode wasn't applied correctly if the cursor wasn't moved.
The bug only affected Avalonia, but I found a few other small issues within Gtk that were addressed as well.

Thank you to @jojodunet for notifying me about this issue!

### DIFF
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -86,7 +86,7 @@ namespace Ryujinx.Ava
         private KeyboardHotkeyState         _prevHotkeyState;
 
         private long _lastCursorMoveTime;
-        private bool _isCursorInRenderer;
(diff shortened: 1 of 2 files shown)


siegmund-heiss-ich:
Just tested this PR. For me, the problem still occurs on macOS. When I enable fullscreen (Ctrl+Cmd+f or Fullscreen Button) I have to move the mouse once.

marysaka (approved):

AcK77 (approved):

jojodunet:
@TSRBerry Unfortunately, the argument is still broken. I need to move the cursor outside the window to avoid seeing it.

I am using the following command:

Ruyjinx.Ava.exe -f -hide-cursor Always Game.nsp

PR #4938, opened 2023-05-14
Linux: Use gamemode if it is available when using Ryujinx.sh.

subanz:
When using the Ryujinx.sh script to start the emulator check if gamemoderun exists and use it if it does.

Gamemode mode on Linux changes some system settings to make performance during gaming more consistent mainly by changing the CPU governor to performance.

https://github.com/FeralInteractive/gamemode


distribution/linux/Ryujinx.sh
```diff
@@ -11,4 +11,4 @@ if [ -f "$SCRIPT_DIR/Ryujinx.Headless.SDL2" ]; then
     RYUJINX_BIN="Ryujinx.Headless.SDL2"
 fi
 
-env DOTNET_EnableAlternateStackCheck=1 "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
+env DOTNET_EnableAlternateStackCheck=1 "$(which gamemoderun)" "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
```
TSRBerry, line 14: This will not work, since which returns `gamemoderun not found` if the binary is not present.

subanz: You are right, I was incorrectly assuming that it would return nothing because of the examples I found for its usage but it seems to return something along the lines of which: no "xyzprogram" in (/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl) when the program doesn't exist. I'll try to fix this, sorry I'm very much an amateur here.

makigumo:
I think the current approach to determine the availability of the `gamemoderun` command is not really working. The current attempt will fail on other implementations of `which` (bash, zsh, external vs. internal command), where a different output might be generated.

I think a better approach might be something like this.
```sh
GAMEMODERUN=""
if command -v gamemoderun >/dev/null 2>&1
then
    GAMEMODERUN="$(which gamemoderun)"
fi
env DOTNET_EnableAlternateStackCheck=1 "$GAMEMODERUN" "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
```

subanz:
> I think the current approach to determine the availability of the `gamemoderun` command is not really working. The current attempt will fail on other implementations of `which` (bash, zsh, external vs. internal command), where a different output might be generated.
> 
> I think a better approach might be something like this.
> 
> ```shell
> GAMEMODERUN=""
> if command -v gamemoderun >/dev/null 2>&1
> then
>     GAMEMODERUN="$(which gamemoderun)"
> fi
> env DOTNET_EnableAlternateStackCheck=1 "$GAMEMODERUN" "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
> ```

Thanks for the advice this looks like a better approach, 3rd? 4th? times the charm?

I'm going to edit it a little bit because at least on my system if "$GAMEMODERUN" is an empty string env thinks its a parameter and throws an error.

distribution/linux/Ryujinx.sh
mikakev1, line 1: 'command -v' is a bashism, for portability reason, you should prefer something like whereis, you're not supposed to know which shell will be called behind after this kind of shebang. 

If you want bash all the way, be sure to use bashism like [[ ]] for tests (unless numerical comparison), then it'll be convenient to put a modern shebang like #!/usr/bin/env bash

You can redirect this way, stdout and stderr altogether  : &>/dev/null, it works with fish and zsh aswell interactively.

You do not need to bother using 'env' to set a variable, you can just put :

DOTNET_EnableAlternateStackCheck=1 "${SCRIPT_DIR }/${RYUJINX_BIN}" "${@}"

subanz: I feel like we are circling back to the issue with which having different outputs depending on Linux versions but this time with whereis, or is whereis more uniform with its output? 
Sorry complete noob here and I'm starting to think simply adding gamemoderun to an sh script is a bit above my abilities at this rate hehe.

subanz: Unless mikakev1 has objections I'll change it to Midnoclose's version.
Dumb question I know but could $(command -v gamemoderun) ever return a path with spaces?

subanz: @Midnoclose,
I mentioned this in a earlier comment but calling env DOTNET_EnableAlternateStackCheck=1 "" "$SCRIPT_DIR/$RYUJINX_BIN" "$@" if "$(command -v gamemoderun)" returns an empty string causes env to take the empty string as an argument and throw an error. So unless you have objections I'll keep the version with the if statement.
@mikakev1 seems to have lost interest in this?

distribution/linux/Ryujinx.sh
```diff
@@ -11,4 +11,9 @@ if [ -f "$SCRIPT_DIR/Ryujinx.Headless.SDL2" ]; then
     RYUJINX_BIN="Ryujinx.Headless.SDL2"
 fi
 
-env DOTNET_EnableAlternateStackCheck=1 "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
+if command -v gamemoderun >/dev/null 2>&1
+then
+    env DOTNET_EnableAlternateStackCheck=1 gamemoderun "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
+else
+    env DOTNET_EnableAlternateStackCheck=1 "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
+fi
```
TSRBerry, line 19: How do you feel about something like this?

```suggestion
if command -v gamemoderun > /dev/null 2>&1; then
    COMMAND="$COMMAND gamemoderun"
fi

$COMMAND "$SCRIPT_DIR/$RYUJINX_BIN" "$@"
```

Additionally you'd add this line after line 4:
```sh
COMMAND="env DOTNET_EnableAlternateStackCheck=1"
```

(GitHub doesn't let me comment on unmodified lines, so I had to add that part to this comment as well.)

TSRBerry (commented):
I made a suggestion to combine the commands into one, so we don't have to check multiple places for later edits.

Otherwise lgtm!

distribution/linux/Ryujinx.sh
subanz: Updated based on your suggestion, looks good to me.

TSRBerry (approved):
Thank you!

AcK77 (approved):

PR #4952, opened 2023-05-14
Remove warning for unparsable old-style Last Played strings

SamusAranX:
A followup to #4861. This fixes #4944 and any issues like it, should any crop up in the future.

Issue #4920, opened 2023-05-12
[Bug] last played / time played


jdvannorman:
### Description of the issue

ryujinx ldn3.1.3 crashes when sorting by last played, time played or file size. 

### Reproduction steps

ryujinx ldn3.1.3 crashes when sorting by last played, time played or file size. 

### Log file

00:00:00.036 |I| Configuration LogValueChange: ResScale set to: 1
00:00:00.040 |I| Configuration LogValueChange: ResScaleCustom set to: 1
00:00:00.040 |I| Configuration LogValueChange: MaxAnisotropy set to: -1
00:00:00.040 |I| Configuration LogValueChange: AspectRatio set to: Fixed16x9
00:00:00.041 |I| Configuration LogValueChange: BackendThreading set to: Auto
00:00:00.042 |I| Configuration LogValueChange: GraphicsBackend set to: Vulkan
00:00:00.042 |I| Configuration LogValueChange: PreferredGpu set to: 0x10DE_0x2504
00:00:00.042 |I| Configuration LogValueChange: AntiAliasing set to: None
00:00:00.043 |I| Configuration LogValueChange: ScalingFilter set to: Bilinear
00:00:00.043 |I| Configuration LogValueChange: ScalingFilterLevel set to: 80
00:00:00.045 |I| Configuration LogValueChange: EnableDockedMode set to: True
00:00:00.051 |I| Configuration LogValueChange: EnableVsync set to: True
00:00:00.051 |I| Configuration LogValueChange: EnableShaderCache set to: True
00:00:00.051 |I| Configuration LogValueChange: EnableTextureRecompression set to: False
00:00:00.051 |I| Configuration LogValueChange: EnableMacroHLE set to: True
00:00:00.051 |I| Configuration LogValueChange: EnablePtc set to: True
00:00:00.051 |I| Configuration LogValueChange: EnableInternetAccess set to: True
00:00:00.051 |I| Configuration LogValueChange: EnableFsIntegrityChecks set to: True
00:00:00.051 |I| Configuration LogValueChange: FsGlobalAccessLogMode set to: 3
00:00:00.051 |I| Configuration LogValueChange: AudioBackend set to: OpenAl
00:00:00.052 |I| Configuration LogValueChange: AudioVolume set to: 1
00:00:00.052 |I| Configuration LogValueChange: MemoryManagerMode set to: HostMappedUnsafe
00:00:00.052 |I| Configuration LogValueChange: ExpandRam set to: False
[...]

PR #4956, opened 2023-05-15
Overhaul of string formatting/parsing/sorting logic for TimeSpans, DateTimes, and file sizes

SamusAranX:
This one kinda grew out of #4952 after I found other formatting and parsing issues, including a possible cause for the crash in #4920. I've closed #4952 in favor of this PR.
This is a big PR that touches a bunch of files but I believe it makes every touched part either more robust or simply better.

I'm gonna explain some of the changes in longer form than what commit messages allow:

# Value parsing and formatting

I noticed that the handling of `DateTime`s, `TimeSpan`s, and file sizes was kind of all over the place, so I added a new helper class called `ValueFormatUtils` (naming's not my strong suit) that can handle formatting and parsing.
This class is intended as a centralized place where those operations take place, as discrepancies in how formatting and parsing are done can cause and have caused problems in the past.
(see also: #4861, which solved one of those problems)

## DateTime

Building on #4861, I moved `DateTime` formatting and parsing to the new `ValueFormatUtils` class.
I removed `NullableDateTimeConverter` again because it wasn't actually needed.

## TimeSpan

Similar to what I did in #4861, I replaced the `double`-based `TimePlayed` property with one backed by a `TimeSpan` and added a migration path that upgrades the old "time_played" JSON field to a "timespan_played" field that contains a serialized `TimeSpan`.

I replaced the old `TimePlayed` format ("1d 09h 20m") with another one that's mostly stock `TimeSpan` formatting except for the number of days ("1d, 09:20:00"). The old format was visually very noisy and excluded seconds for no reason.
Additionally, the `SortHelper` code meant to re-parse the string back into a sortable value looked fragile, so I replaced it.

Time Played values of 0 seconds no longer display as "Never", but as 00:00:00.
"Never" makes no sense as an answer to "How long did I play that game for".

## File sizes

`ApplicationLibrary.LoadApplications()` contained this wild line of code:

```cs
double fileSize = new FileInfo(applicationPath).Length * 0.000000000931;
```

I couldn't find out where it came from, let alone what it was supposed to do, but in the name of making the code less fragile, I ripped it out and made `fileSize` a `long`.

The formatting of file sizes in particular was done in various ways in several places in the codebase (SaveModel.cs, SystemInfo.cs, ApplicationLibrary.cs), so I changed all those to use `ValueFormatUtils.FormatFileSize()` instead.
Conversely, `ValueFormatUtils.ParseFileSize()` can turn strings generated by `ValueFormatUtils.FormatFileSize()` back into file size numbers. That function is only used for sorting purposes as of this PR. More on sorting below.

I took some liberties regarding files size formatting here:

`ValueFormatUtils` can parse both base 2 and base 10 file size units (read: KiB/MiB/GiB or KB/MB/GB respectively) but it will format *exclusively* by using base 2 calculations and appending base 10 units.
This matches what Windows and other systems do, as per https://devblogs.microsoft.com/oldnewthing/20090611-00/?p=17933 and this screenshot:
A screenshot of the Avalonia Ryujinx main window and an explorer.exe properties window side by side, showing that Smash Bros. Ultimate's displayed file sizes match

Personally, I think it's a little confusing that Ryujinx uses base 2 units for display when, to my knowledge, no other program that's not aimed at nerds (affectionate) does. This change helps legibility. No one knows or cares what a GiB is. 

I also made `SystemInfo` print RAM values in GB instead of MiB.
A screenshot of Ryujinx's console output showing RAM values in gigabytes instead of mebibytes

# Sorting

Prior to this PR, the sorting for `TimePlayed`, `LastPlayed`, and `FileSize` was backwards.
To verify that, open a Ryujinx window and an Explorer window with more than a few files side by side and sort both by size so that the largest files come first.
In this situation, the triangle in the column header is supposed to point downwards.
This PR fixes the order of sort operands in all the places that sorting happens.

Also, as mentioned earlier, I believe the old `SortHelper` code might've been causing the crash mentioned in #4920. I was able to reproduce it and fixed it by making `SortHelper` a thin-ish wrapper around `ValueFormatUtils`.

If there are any changes I didn't mention here that need further explanation, please let me know.


src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Linq;
+
+namespace Ryujinx.Common.Utilities
+{
+    public static class ValueFormatUtils
+    {
+        private static readonly string[] FILE_SIZE_UNITS =
+        {
+            "B", "KB", "MB", "GB", "TB", "PB", "EB",
+            "KiB", "MiB", "GiB", "TiB", "PiB", "EiB",
```
AcK77, line 14:
```suggestion
            "KiB", "MiB", "GiB", "TiB", "PiB", "EiB"
```

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+        public enum FileSizeUnits
+        {
+            Auto = -1,
+            Bytes = 0,
+            Kilobytes = 1,
+            Megabytes = 2,
+            Gigabytes = 3,
+            Terabytes = 4,
+            Petabytes = 5,
+            Exabytes = 6,
```
AcK77, line 29:
```suggestion
            Exabytes = 6
```

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+            if (timeSpan.Value.TotalDays < 1)
+            {
+                // Game was played for less than a day
+                return timeSpan.Value.ToString("c", CultureInfo.InvariantCulture);
+            }
+
+            // Game was played for more than a day
+            TimeSpan onlyTime = timeSpan.Value.Subtract(TimeSpan.FromDays(timeSpan.Value.Days));
+            string onlyTimeString = onlyTime.ToString("c", CultureInfo.InvariantCulture);
+            return $"{timeSpan.Value.Days}d, {onlyTimeString}";
```
AcK77, line 55:
```suggestion

            return $"{timeSpan.Value.Days}d, {onlyTimeString}";
```

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        /// <summary>
+        /// This method parses a string into a <see cref="TimeSpan"/>.
+        /// </summary>
+        /// <param name="timeSpanString">A string representing a <see cref="TimeSpan"/>.</param>
+        /// <returns>A <see cref="TimeSpan"/> object. If the input string couldn't been parsed, <see cref="TimeSpan.Zero"/> is returned.</returns>
+        public static TimeSpan ParseTimeSpan(string timeSpanString)
+        {
+            var returnTimeSpan = TimeSpan.Zero;
```
AcK77, line 65:
```suggestion
            TimeSpan returnTimeSpan = TimeSpan.Zero;
```

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// <param name="culture">The <see cref="CultureInfo"/> that's used in formatting. Defaults to <see cref="CultureInfo.CurrentCulture"/>.</param>
+        /// <returns>A formatted string that can be displayed in the UI.</returns>
+        public static string FormatDateTime(DateTime? utcDateTime, CultureInfo culture = null)
+        {
+            culture ??= CultureInfo.CurrentCulture;
+
+            if (!utcDateTime.HasValue)
+            {
+                // TODO: maybe put localized string here instead of just "Never"
+                return "Never";
```
AcK77, line 99: You said:
> Time Played values of 0 seconds no longer display as "Never", but as 00:00:00.
> "Never" makes no sense as an answer to "How long did I play that game for".

Then why this is still existing ? `utcDateTime` should always got a value here?

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,194 @@
...
+                }
+
+                // If the unit index is one that resolves to a base 2 unit, subtract 6 to arrive at a usable power value.
+                if (i > 6)
+                {
+                    i -= 6;
+                }
+                
+                number *= Math.Pow(_base, i);
+                return Convert.ToInt64(number);
```
AcK77, line 188:
```suggestion

                return Convert.ToInt64(number);
```

src/Ryujinx/Ui/MainWindow.cs
```diff
@@ -271,7 +271,7 @@ private MainWindow(Builder builder) : base(builder.GetRawOwnedObject("_mainWin")
                 typeof(string),
                 typeof(string),
                 typeof(BlitStruct<ApplicationControlProperty>));
-
+
```
AcK77, line 274: Extra spaces can be removed

AcK77 (changes requested):

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: My comment was about the Time Played column, which uses TimeSpan objects internally.
FormatDateTime is for DateTime objects, which are used by the Last Played column, where a displayed value of "Never" makes sense.

IsaacMarovitz, line 99: Please localise this value in this PR

IsaacMarovitz (changes requested):

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: `LocaleManager` is in the Ryujinx.Ava namespace, but this class is also used by the old GTK UI. Wouldn't pulling in Avalonia stuff cause problems?

SamusAranX: Fixed the comment mistakes. The comments now correctly state what each method is used for.

SamusAranX: The last commit localizes "Never" in the Avalonia UI.
I found one issue that might be beyond the scope of this PR: Switching languages doesn't reload the game list, which means if you started with English, "Never" will stick around in the UI until you manually hit the reload button.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        /// <summary>
+        /// Formats a number of bytes to a human-readable string with adjustable units.
+        /// </summary>
+        /// <param name="size">The file size in bytes.</param>
+        /// <param name="forceUnit">Formats the passed size value as this unit, bypassing the automatic unit choice.</param>
+        /// <returns>A human-readable file size string.</returns>
+        public static string FormatFileSize(long size, FileSizeUnits forceUnit = FileSizeUnits.Auto)
+        {
+            double _base = 1024;
```
TSRBerry, line 129:
```suggestion
            double base = 1024;
```

Local variables are not prefixed with an underscore.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            return $"{sizeFormatted} {FILE_SIZE_UNITS[unitIndex]}";
+        }
+
+        /// <summary>
+        /// Parses a previously-formatted value and returns a number of bytes. Mainly used for sorting purposes.
+        /// </summary>
+        /// <param name="sizeString">A string representing a file size formatted with <see cref="FormatFileSize"/>.</param>
+        /// <returns>A <see cref="long"/> containing a number of bytes.</returns>
+        public static long ParseFileSize(string sizeString)
```
TSRBerry, line 163: Should you change the method above to use a `ulong` then you could change the return type here to `ulong` as well.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+        {
+            double _base = 1024;
```
TSRBerry, line 165:
```suggestion
            double base = 1024;
```

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            // Enumerating over the units backwards because otherwise, sizeString.EndsWith("B") would exit the loop in the first iteration.
+            for (int i = FILE_SIZE_UNITS.Length - 1; i >= 0; i--)
+            {
+                string unit = FILE_SIZE_UNITS[i];
```
TSRBerry, line 170: Could you change this to a foreach-loop? Since you already imported `System.Linq` you could use `Reverse()` here to iterate over the array backwards.

TSRBerry (approved):
Thank you so much for this PR! Your description is awesome and the changes look very nice!

I added a few comments where I think we could make a few adjustments, but nothing major.
I won't add any comments about styling issues since my styling PR isn't merged yet and I'll just address these in there as well.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: C# doesn't allow you to name a variable `base`, so I put an underscore on it.

src/Ryujinx.Ui.Common/App/ApplicationMetadata.cs
```diff
@@ -7,13 +7,29 @@ public class ApplicationMetadata
...
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+        public double TimePlayedOld { get; set; }
+
         [JsonPropertyName("last_played")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
         public string LastPlayedOld { get; set; }
+
+        public void UpdateTimePlayed()
+        {
+            if (LastPlayed.HasValue)
```
patrick-hovsepian, line 27: super nit: test for the negative and immediately return to help readability

src/Ryujinx.Ui.Common/App/ApplicationData.cs
```diff
@@ -24,29 +24,18 @@ public class ApplicationData
...
-
-                return LastPlayed.Value.ToLocalTime().ToString(CultureInfo.CurrentCulture);
-            }
-        }
+        
+        public string TimePlayedString => ValueFormatUtils.FormatTimeSpan(TimePlayed);
+        
+        public string LastPlayedString => ValueFormatUtils.FormatDateTime(LastPlayed);
+        
+        public string FileSizeString => ValueFormatUtils.FormatFileSize(FileSize);
```
patrick-hovsepian, line 38: worth caching these with an internal prop?

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+            Exabytes = 6
+        }
+
+        /// <summary>
+        /// This method formats a <see cref="TimeSpan"/> so it can be displayed in the UI.
+        /// Only used as a canonical way to format the TimePlayed value.
+        /// </summary>
+        /// <param name="timeSpan">The <see cref="TimeSpan"/> to be formatted.</param>
+        /// <returns>A formatted string that can be displayed in the UI.</returns>
+        public static string FormatTimeSpan(TimeSpan? timeSpan)
```
patrick-hovsepian, line 38: maybe rename if this only for gameTime played

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: I really rather wouldn't because I need the value of `i` later.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            return $"{timeSpan.Value.Days}d, {onlyTimeString}";
+        }
+
+        /// <summary>
+        /// This method parses a string into a <see cref="TimeSpan"/>.
+        /// </summary>
+        /// <param name="timeSpanString">A string representing a <see cref="TimeSpan"/>.</param>
+        /// <returns>A <see cref="TimeSpan"/> object. If the input string couldn't been parsed, <see cref="TimeSpan.Zero"/> is returned.</returns>
+        public static TimeSpan ParseTimeSpan(string timeSpanString)
```
patrick-hovsepian, line 64: Would it be better to just define a Formatter and pass that to TimeSpan.Parse?

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
gdkchan: You could just call it `sizeBase`, `numberBase`, `baseValue` or something like that.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+
+                // Apply an upper bound so that exabytes are the biggest unit used when formatting.
+                if (unitIndex > 6)
+                {
+                    unitIndex = 6;
+                }
+            }
+            else
+            {
+                unitIndex = (int)forceUnit;
```
patrick-hovsepian, line 149: just initialize to this so you can remove this else block entirely

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+                }
+
+                string numberString = sizeString.Split(" ")[0];
+                if (!double.TryParse(numberString, CultureInfo.InvariantCulture, out double number))
+                {
+                    break;
+                }
+
+                // If the unit index is one that resolves to a base 2 unit, subtract 6 to arrive at a usable power value.
+                if (i > 6)
```
patrick-hovsepian, line 183: im not a fan of the magic number and its tied to the order of the FILE_SIZE_UNIT array. maybe consider breaking it up into two separate arrays to better distinguish special cases?

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
...
+            return parsedDateTime;
+        }
+
+        /// <summary>
+        /// Formats a number of bytes to a human-readable string with adjustable units.
+        /// </summary>
+        /// <param name="size">The file size in bytes.</param>
+        /// <param name="forceUnit">Formats the passed size value as this unit, bypassing the automatic unit choice.</param>
+        /// <returns>A human-readable file size string.</returns>
+        public static string FormatFileSize(long size, FileSizeUnits forceUnit = FileSizeUnits.Auto)
```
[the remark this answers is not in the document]
SamusAranX: It's mostly because this function is used from classes that use `long` properties to represent bytes and it'd be a hassle to refactor all of them. There's one instance in `SystemInfo.ToGBString()` where I cast a `ulong` to `long` but that's it.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,196 @@
+﻿using System;
+using System.Buffers.Text;
+using System.Drawing;
```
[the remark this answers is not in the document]
SamusAranX: Could've sworn I had resharper give this file a once-over. welp

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: Renamed all `_base` variables to `SizeBase`.

src/Ryujinx.Ui.Common/App/ApplicationMetadata.cs
SamusAranX: Good point actually.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: Truth be told, I find .NET formatting strings, especially for dates, to be extremely tedious to work with. Personally, I prefer to use the bog standard `.Parse()` overloads and doing everything else myself.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: Done.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: I'm not sure how to do this at this moment. Splitting the array into two separate arrays would make the function you highlighted harder to understand, as I'd have to loop through both arrays in order.
I changed this bit in the last commit to use a properly named variable in place of the magic 6, I hope that makes this part more readable.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
TSRBerry: Ah sorry, I didn't see that. Just leave it like that!

src/Ryujinx.Ui.Common/App/ApplicationData.cs
SamusAranX: I'd have to keep track of every time the non-string properties these refer to are updated and then update the cached value. It's a good suggestion but since these are called at most once per game list entry, I don't think it's worth the effort atm.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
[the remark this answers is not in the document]
SamusAranX: Oh, I forgot: The main reason is that `FileInfo.Length` is a long. Honestly idk why it's a ulong in some places.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
patrick-hovsepian: fair enough

SamusAranX:
I put the methods in `ValueFormatUtils.cs` into formatter and parser `#regions` in the last commit.
IMO this makes the code more readable as the parsers are only ever used by the GTK UI.

SamusAranX:
The last commit integrates #5056's changes but moves metadata update logic to two methods on `ApplicationMetadata`.
Functionally, `ApplicationMetadata.UpdateTimePlayed()` (from an earlier commit in this PR) is now `UpdatePostGame()`.
`UpdatePreGame()` wraps the changes from #5056.

src/Ryujinx/Ui/Helper/SortHelper.cs
```diff
@@ -7,88 +8,31 @@ static class SortHelper
...
-            if (float.Parse(aValue) > float.Parse(bValue))
-            {
-                return -1;
-            }
-            else if (float.Parse(bValue) > float.Parse(aValue))
-            {
-                return 1;
-            }
-            else
+            if (aSize == bSize)
```
gdkchan, line 30: Can't this just return `aSize.CompareTo(bSize);` directly?

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,202 @@
...
+            return $"{sizeFormatted} {FileSizeUnitStrings[unitIndex]}";
+        }
+
+        #endregion
+
+        #region Value parsers, only used by the GTK UI via SortHelper.cs
+
+        /// <summary>
+        /// Parses a string generated by <see cref="FormatTimeSpan"/> and returns the original <see cref="TimeSpan"/>.
+        /// Only used to sort the GTK UI's game list.
```
gdkchan, line 117: I don't think you should be making mentions to the GTK UI Here. If this is something that is only needed by GTK, then it shouldn't be here, it should be in the GTK UI Project.

src/Ryujinx.Ui.Common/App/ApplicationMetadata.cs
```diff
@@ -7,13 +7,45 @@ public class ApplicationMetadata
...
+
+        /// <summary>
+        /// Updates <see cref="LastPlayed"/> and <see cref="TimePlayed"/>. Call this after a game ends.
+        /// </summary>
+        public void UpdatePostGame()
+        {
+            DateTime? prevLastPlayed = LastPlayed;
+            UpdatePreGame();
+
+            if(!prevLastPlayed.HasValue)
```
gdkchan, line 41:
```suggestion
            if (!prevLastPlayed.HasValue)
```

src/Ryujinx/Ui/Helper/SortHelper.cs
SamusAranX: You're right, I forgot that method even existed

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
```diff
@@ -0,0 +1,202 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+
+namespace Ryujinx.Common.Utilities
+{
+    public static class ValueFormatUtils
```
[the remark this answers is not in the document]
SamusAranX: SystemInfo.cs makes use of ValueFormatUtils and I can't reference the `Ryujinx.Ui.Common` namespace from there. `Ryujinx.Common.Utilities` is usable from more places, so I'd like to keep this file there.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: I'd like to keep the formatting and parsing functions together in the same file because they're inherently related to each other. Ripping one group of functions out and putting it into another namespace would just result in code duplication and worse maintainability.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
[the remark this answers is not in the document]
gdkchan: I think `SystemInfo` (at least as it exists today) should also be in `Ryujinx.Ui.Common`. It is not really useful for anything other than UI, so it should be in the UI project. `Ryujinx.Common` is for truly common code that can be used by any project. I'd suggest just moving both there.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
gdkchan: I don't mind keeping them together, but I think it should be moved to `Ryujinx.Ui.Common` (my other suggestion). Also I'm not sure what mentioning that it's only used in GTK here adds if this is supposed to be a generic parsing.

gdkchan:
> I also made SystemInfo print RAM values in GB instead of MiB.

This will probably mess with the bot. It's worth keeping in mind that the bot will need to be updated after this change (assuming this specific change will go in).

gdkchan:
> Personally, I think it's a little confusing that Ryujinx uses base 2 units for display when, to my knowledge, no other program that's not aimed at nerds (affectionate) does. This change helps legibility. No one knows or cares what a GiB is.

If we are changing it here, we should probably change it consistently across the UIs too in a follow up PR.

SamusAranX:
> This will probably mess with the bot.

I didn't consider that. I can try drafting a PR for the bot to help deal with that.

> we should probably change it consistently across the UIs too in a follow up PR

This PR changes display units in all places where file sizes are displayed. I did some surface-level code searches and couldn't find any remaining uses of base 2 units in user-facing contexts. A followup PR might not be needed.

gdkchan:
> This PR changes display units in all places where file sizes are displayed. I did some surface-level code searches and couldn't find any remaining uses of base 2 units in user-facing contexts. A followup PR might not be needed.

One easy example is some tool tips, for example, this is present in all locales: https://github.com/SamusAranX/Ryujinx/blob/formatting-fixes/src/Ryujinx.Ava/Assets/Locales/en_US.json#L610

There are more (not all of them are user facing, but we should be consistent on code naming too regardless).

SamusAranX:
Bot PR is up: https://github.com/Ryujinx/ryuko-ng/pull/58

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
SamusAranX: Removed the GTK-specific comments.

src/Ryujinx.Common/Utilities/ValueFormatUtils.cs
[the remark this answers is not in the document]
SamusAranX: The last commit moves ValueFormatUtils.cs and SystemInfo.cs (along with its subclasses) to `Ryujinx.Ui.Common`.

src/Ryujinx.Ui.Common/SystemInfo/SystemInfo.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using Ryujinx.Common.Logging;
+using Ryujinx.Ui.Common.Helper;
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics.X86;
 using System.Text;
 
-namespace Ryujinx.Common.SystemInfo
+namespace Ryujinx.Ui.Common.SystemInfo
```
AcK77, line 8: I don't think it's needed to move it from `Ryujinx.Common.SystemInfo`  to `Ryujinx.Ui.Common.SystemInfo` since `Ryujinx.Ui.Common` only exist because we currently have 2 UI, one day when we will drop GTK we could not keep it!
(same for the 2 other moved classes)

gdkchan: If anyone wants to remove `Ryujinx.Ui.Common` after GTK is gone, those classes can just be moved to `Ryujinx.Ava`. The reason why I don't think it should be in `Ryujinx.Common` is because `Ryujinx.Common` is for functionality that is truly useful for any project, not something that is only used by the UI. `SystemInfo` as implemented today is not useful for anything other than UI.

TSRBerry:
@SamusAranX I resolved the merge conflicts for you, I hope that's okay for you.

@Ryujinx/developers this PR has been waiting for a really long time now and we need it. Could someone of you please review it?

gdkchan:
If we're changing KiB/MiB/GiB back to KB/MB/GB, we should also replace all UI strings that currently uses MiB/GiB for consistency, and stick with it.

TSRBerry:
Could be done in a follow-up PR as well. Especially now that we can't be sure anymore that the author wants/intents to work on this any further.

gdkchan:
Shouldn't be too hard to change it, and it's not very good to leave the emulator in a consistent state. It doesn't need to be done by the author, any maintainer should be able to change the branch.

marysaka (approved):
lgtm apart from the MiB and GiB changes.

I will merge with that part reverted, if you want to change it back to MB, GB,.. please open an issue/PR to continue that discussion.

PR #4982, opened 2023-05-16
experiment with Task threads

AaronO:
An experiment mapping emulated threads to C# Tasks in lieu of OS threads aiming to reduce contention and CPU overhead.

Very much WIP


raven02:
Can have a rebase to test it ?

PR #4992, opened 2023-05-17
Start GPU performance counter at 0 instead of host GPU value

OpaqueReptile:
Instead of returning host GPU counter, offset the value to 0 on the first call of GetTimestamp(). 

This seems to fix Tears of the Kingdom being locked at 20fps.


src/Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -219,6 +220,13 @@ public ulong GetTimestamp()
         {
             ulong ticks = ConvertNanosecondsToTicks((ulong)PerformanceCounter.ElapsedNanoseconds);
 
+            // Initialize ticks counter to start at 0, instead of the host GPU counter
+            if (_firstTimestamp == 0)
```
gdkchan, line 224: Why not initialize it on the constructor instead of doing lazy initialization here?

gdkchan:
This is not a host GPU counter btw, it is just a CPU tick count.

src/Ryujinx.Graphics.Gpu/GpuContext.cs
OpaqueReptile: I wasn't exactly sure which behavior would be correct - if the first getTimestamp() needed to be zero or if there was some delay between the constructor and the first call of getTimestamp(). Moving the initialization as suggested still seems to work 👍

OpaqueReptile:
> This is not a host GPU counter btw, it is just a CPU tick count.

Good to know, I was just going off of the description of GetTimestamp(), wasn't sure where the value itself was coming from on the host.

siegmund-heiss-ich:
Tested it on macOS.
Works great.

riperiperi (approved):
Very weird that the game depends on this, but it does work. I wonder if this is reset somehow, or automatically set to 0 at the start of each process?

src/Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -219,6 +222,9 @@ public ulong GetTimestamp()
         {
             ulong ticks = ConvertNanosecondsToTicks((ulong)PerformanceCounter.ElapsedNanoseconds);
 
+            // Initialize ticks counter to start at 0, instead of host value
```
riperiperi, line 225:
```suggestion
            // Initialize ticks counter to start at 0, instead of host value.
```

src/Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -219,6 +222,9 @@ public ulong GetTimestamp()
         {
             ulong ticks = ConvertNanosecondsToTicks((ulong)PerformanceCounter.ElapsedNanoseconds);
```
gdkchan, line 223:
```suggestion
            ulong ticks = ConvertNanosecondsToTicks((ulong)PerformanceCounter.ElapsedNanoseconds) - _firstTimestamp;
```
Why not do that here directly?

gdkchan: Yes, you can keep it.

src/Ryujinx.Graphics.Gpu/GpuContext.cs
```diff
@@ -217,7 +220,8 @@ public long GetModifiedSequence()
         /// <returns>The current GPU timestamp</returns>
         public ulong GetTimestamp()
         {
-            ulong ticks = ConvertNanosecondsToTicks((ulong)PerformanceCounter.ElapsedNanoseconds);
+            // Initialize ticks counter to start at 0, instead of host value.
```
gdkchan, line 223: The "initialize" part of the comment doesn't make sense anymore since it's no longer "initialized" here.

gdkchan (approved):
lgtm other than comment, thanks.

DaRkL3AD3R:
This seems to have solved the problem I was describing in a reddit post where rebooting my PC fixed the game being stuck at 20 fps, but leaving it on for a while (several hours) would cause the game to be locked to 20 fps. Glad to know it wasn't a rogue process or something.

Teiby:
Well, I get 30 FPS but now there drops to 0 FPS. See: https://youtu.be/PB_fm3NbY28
No mods used. vsync disabled, RivaTuner to limit frames.

PR #4999, opened 2023-05-17
GPU: Avoid using garbage size for non-cb0 storage buffers

riperiperi:
In the depths area, Tears of the Kingdom uses a global memory access with address on constant buffer slot 6. This isn't standard and thus doesn't actually have a size 8 bytes after it, so we were reading back a garbage size that ended up very large (at least in version 1.1.0), and would synchronize a lot of data per frame.

This PR makes storage buffers created from addresses outside constant buffer slot 0 get their size as the number of bytes remaining in the GPU mapping starting at the given virtual address. This should bound the buffer to a reasonable size, and ideally stop it crossing into other memory.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -23,7 +23,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         public const int PrimitiveRestartStateIndex = 12;
         public const int RenderTargetStateIndex = 27;
 
-        public const ulong MaxUnknownStorageSize = 1048576;
+        private const ulong MaxUnknownStorageSize = 0x100000;
 
         private readonly GpuContext _context;
         private readonly GpuChannel _channel;
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -366,6 +366,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                     }
                     else
                     {
+                        // TODO: Use full mapped size and somehow speed up buffer sync.
                         size = (uint)_channel.MemoryManager.GetMappedSize(sbDescriptor.PackAddress(), MaxUnknownStorageSize);
                     }
 
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -23,6 +23,8 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
         public const int PrimitiveRestartStateIndex = 12;
         public const int RenderTargetStateIndex = 27;
 
+        public const ulong MaxUnknownStorageSize = 1048576;
+
         private readonly GpuContext _context;
         private readonly GpuChannel _channel;
         private readonly DeviceStateWithShadow<ThreedClassState> _state;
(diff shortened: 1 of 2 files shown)


riperiperi:
I've limited the maximum size of the buffer as a temporary measure. I've looked pretty deep into this and there isn't any way to limit the bounds of the buffer, as the game creates a huge 200MB mapping and then just randomly uses parts of it.

I'm going to leave it like this for now until we can find a better solution to this problem. We can improve performance when synchronizing full buffers significantly (and mapping size check), but I'm not sure how fast it can be compared to having a nice well behaved size.

Since only TOTK is affected by this, check the depths and make sure nothing goes horribly wrong there.

Bjorn29512:
Can confirm this fixes the significant performance drops after entering depths in TOTK compared to master.

Tested and reproduced multiple scenarios several times that would trigger the performance collapse:
 
 - Dropping down the chasm from surface.
 - Teleporting to the depth.
 - Loading the save to the depths. 
 - Teleporting within the depths.
 - Loading a save in depth while exploring depths.
 
None of the above scenarios were able to reproduce that issue and it seems very likely fixed, i will keep playing in depth for sometime and if anything breaks i will update and leave a comment.

Master: 

image

PR :

image

src/Ryujinx.Graphics.Gpu/Memory/MemoryManager.cs
```diff
@@ -637,6 +637,33 @@ public ulong TranslateFirstMapped(ulong va, ulong size)
...
+
+            ulong pte = GetPte(va);
+
+            while (pte != PteUnmapped && va < endVa)
+            {
+                va += PageSize - (va & PageMask);
+                pte = GetPte(va);
+            }
+
+            return va - startVa;
```
gdkchan, line 664: I think if `endVa` is not page alined, `va` might be greater than `endVa` which could cause it to return a size that is slightly higher than `maxSize`.

src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -23,6 +23,8 @@ class StateUpdater
         public const int PrimitiveRestartStateIndex = 12;
         public const int RenderTargetStateIndex = 27;
 
+        public const ulong MaxUnknownStorageSize = 1048576;
```
gdkchan, line 26: Personally I think power of 2 values are nicer to read as hex. In this case it would be 0x100000. Also this constant doesn't need to be public.

gdkchan (approved):
lgtm, thanks. Assuming a maximum size is not ideal since it will break if the accessed region is larger than the assumed size, but given that until a few days ago the storage buffer from non-cb0 case was not supported at all, and most games were fine regardless, I think chances that this would regress anything are very low.

LeoJrC:
I've made some tests (walking around dephs) and vram usage dropped signficantly using this build. I don't feel like the it's stabbing a knife in my vram (i could deactivate texture recompression without killing my PC) but my performance still goes down a lot. Dropping from a steady 50is fps to 15~20. Imma attach my log. 

My specs;
Ryzen 5 5600x
RX 570
16GB RAM
Windows 11

[Ryujinx_1.1.0+8d3eec5_2023-05-18_01-07-59.log](https://github.com/Ryujinx/Ryujinx/files/11504313/Ryujinx_1.1.0%2B8d3eec5_2023-05-18_01-07-59.log)

marysaka (approved):

src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -356,7 +358,19 @@ private void UpdateStorageBuffers()
 
                     SbDescriptor sbDescriptor = _channel.MemoryManager.Physical.Read<SbDescriptor>(sbDescAddress);
 
-                    _channel.BufferManager.SetGraphicsStorageBuffer(stage, sb.Slot, sbDescriptor.PackAddress(), (uint)sbDescriptor.Size, sb.Flags);
+                    uint size;
+                    if (sb.SbCbSlot == 0)
```
gdkchan, line 362: I didn't notice yesterday, but this should be done on `ComputeClass` as well. It would be nice to extract this logic into a separate method and call it from both, since they are supposed to do more or less the same thing, the main difference being that one calls `SetGraphicsStorageBuffer` while the other calls `SetComputeStorageBuffer`.

PR #5002, opened 2023-05-17
ava: Fix crash when extracting sections from NCA with no data section

marysaka:
Tested against Omega Strickers.


AcK77 (approved):

marysaka:
> Is there a reason we don't have this inside some shared project? Is it different for each GUI?

It should be on shared project but isn't.. I guess that's a job for a future person ™️

PR #5003, opened 2023-05-18
Fix Vulkan blit-like operations swizzle

gdkchan:
On Vulkan, there are a few cases where blit operations are "emulated" using regular draws. Those cases would take the texture component swizzle into account, which is incorrect as swizzle does not affect blits or rendering, they are only supposed to affect regular texture sampling from shaders. This fixes the issue by adding a new `SetTextureAndSamplerIdentitySwizzle` that is now used to bind a `ImageView` with identity (RGBA) swizzle.

Fixes colors flashing on Omega Strikers.
Before:

https://github.com/Ryujinx/Ryujinx/assets/5624669/5acd0beb-2288-4c61-a3d0-e593b63821a2

After:

https://github.com/Ryujinx/Ryujinx/assets/5624669/d10613e3-96b4-413b-89b1-d9d0d5cf54ac

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
+++ b/src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
@@ -228,7 +228,12 @@ namespace Ryujinx.Graphics.Vulkan
             SignalDirty(DirtyFlags.Storage);
         }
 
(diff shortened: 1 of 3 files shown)


marysaka (approved):
lgtm, the game now render correctly for me 👍

riperiperi (approved):
Nice catch.

PR #5012, opened 2023-05-18
DIRTY HACK: Fix z-fighting on water in Zelda: ToTK

cstamford:
As reported in this issue https://github.com/Ryujinx/Ryujinx-Games-List/issues/4598#issuecomment-1545901248, there is extreme z-fighting on water and less extreme, but still visible z-fighting on distant geometry. The issue may be hardware or driver specific as the low frequency of reports combined with the obvious severity of the issue makes me suspect that most people aren't impacted by it.

Here is my own footage, GTX 4090:

https://github.com/Ryujinx/Ryujinx/assets/6191957/fa9ec07d-d38d-4a0a-8d62-d23fd673fc90

I spent some time tracking down the issue. It is caused by a poorly set depth bias during the water pass. The depth bias was set to 1 which was causing the extreme z-fighting seen above. 

image

In this change I set it to a more appropriate value. I don't know if the depth bias is correctly set because the game sets it and therefore it's a game problem that is coincidentally only visible in Ryujinx or whether Ryujinx is translating game render commands to Vulkan commands incorrectly.

Video with improved depth bias:

https://github.com/Ryujinx/Ryujinx/assets/6191957/0a3d26ce-f007-459a-969f-01bdf6e71a68

I strongly suspect the z-fighting on distant geometry is caused by the depth bias setting as well, but it's minor enough that I don't consider it a priority to check.

I'm not a Ryujinx contributor or expert and I don't know the code well enough to make a proper fix without days of learning and effort. **Obviously you shouldn't merge this change - it's a game-specific hack**. I'm making a PR here to provide visibility, a temporary fix for impacted users, and hopefully inspire a real fix from the Ryujinx emulator experts in the future.

P.S. I tested a bit in various scenarios and it seemed to work correctly, but it's possible that something else using the same depth bias constant value will be negatively impacted by this change, so you might see some weird behaviour with things like shadows and other intersecting geometry.

P.P.S. If you need to discuss this with me in real time, hit me up on Discord at Sea#8008. I'm open to collaborating on or implementing a real fix with guidance from the relevant Ryujinx experts.

P.P.P.S. Your codebase is very nice to work in. Great work, and a more general thanks for the hard work and high quality software you've created here.


Foul-Tarnished:
I have extreme z-fighting for water at least on a gtx 1060

DeniStoj:
I do want to note there is also z-fighting which occurs in Tears of the Kingdom on some geometry such as some rock faces as well as the top of the mushroom/plant trees in the Depths. I'm wondering if this has fixed that as well.

moistman7:
This fixed that water for me at certain distances.  There was still z-fighting a various elevations. No change to the geometry.  Appreciate it though!

fatjesusonabike:
Seconding what moistman7 said. Fixes it only at certain distances and reduces the general severity of it, but it's still very present and noticeable on both water and geometry. 

Vulkan/3070ti-latest drivers

riperiperi:
My first suggestion is to stop comparing against Yuzu. The emulator should be compared against console and console only. The second is that Ryujinx does not decide draw order, so that can’t be responsible.

It’s possible the bias calculation on the backends based on factor, units and depth mode is incorrect. I’m not sure as I haven’t compared against console. It’s less likely that it’s not updating properly, check the state updater to see how it decides how to do that.

Jomanakk:
https://github.com/Ryujinx/Ryujinx/assets/107001229/add3a3c0-7975-4219-988e-f458259e6cbc

fatjesusonabike:
Here's another example, this time on OpenGL

https://github.com/Ryujinx/Ryujinx/assets/134035322/98d49958-933d-4b23-96d2-334fe460aa1d

yell0wsuit:
> My intent with the comparison was not to check accuracy but compare what state we're setting differently as a learning exercise to help understand what we set and how we set it

1. yuzu's license is GPL, which is not compatible with Ryujinx's MIT license. It means the Ryujinx team cannot use yuzu's code as a "learning exercise".
2. That emulator has been notoriously known for bugs due to hacky and inaccurate development. So again, comparing to what the console does and mimicing that is the best solution, which is the direction the Ryujinx team is following.

gdkchan:
> So this reverse engineered structure is correct and we're reading the same offset, how could we be setting different depth biases for the same scene - a value the hardware is meant to be provided to us?

For future reference, you can find all fields on 'ThreedClassState' on official documentation from NVIDIA https://github.com/NVIDIA/open-gpu-doc/blob/master/classes/3d/clb197.h A lot of registers there are missing from 'ThreedClassState', and some have different names since when it was written, this documentation was not public yet.

It could also be worth providing a save to allow others to reproduce the issue (I don't have a save where I can easily reproduce this right now, I remember seeing Z-fighting on the first jump from the island at the start of the game, but that's it).

FaeTalan:
I'm on a RTX 3080, using latest Ryujinx in vulkan, and I have this extreme z-fighting issue. It's incredibly distracting.

In this save file you just look at the horizon and see the horrible flickering artifacts all over, among the clouds, (like in those videos posted here).

[0.zip](https://github.com/Ryujinx/Ryujinx/files/11518676/0.zip)

gdkchan:
It could be worth checking if the shader uses FP16 operations. The emulator implements them with FP32 operations right now, which has more precision, but could lead to different results compared to what the Switch produces. An easy way to check might be removing `(operation.Inst == Instruction.PackHalf2x16 && PropagatePack(operation)) ||` on https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Graphics.Shader/Translation/Optimizations/Optimizer.cs#L89 which will force it to convert values back to FP16 after each operation, which should cause it to have values closer to what it would have if it was using FP16 operations.

moistman7:
I would love to test it.

fatjesusonabike:
Ready for some testing too !

moistman7:
Sorry dumb question. How do I download the new build?

noxifoxi:
> Sorry dumb question. How do I download the new build?

click on `Show all checks`, then `Details` and then `Summary`, the build artifacts are at the bottom of the summary page.

GamerzHell9137:
> Sorry dumb question. How do I download the new build?

1st comment under the original post, click on the OS version you need.

https://github.com/Ryujinx/Ryujinx/pull/5012#issuecomment-1553333324

moistman7:
The geometry is much better now.  Water seems to be affected by the z-axis not distance.  Here's a video showing some of the issues.  https://youtu.be/33lppNsXog0

gdkchan:
> for one, why are we halving the depth bias to start with?

Because the NVN (Switch graphics API) driver multiplies it by 2. The emulator halves it to recover the original value.
This is the NVN function that writes those commands into the command buffer on the game:
```cs
void CommandBufferSetPolygonOffsetClamp
               (float factor,float units,float clamp,nvn_command_buffer *cmdBuff)

{
  uint *cmdPtr;
  
  cmdPtr = cmdBuff->commadsCurrent;
  if (cmdBuff->commandsEndPtr < cmdPtr + 6) {
    CommandBufferAllocMore(cmdBuff,cmdPtr,(ulonglong)(cmdBuff->arraySize + 6) << 2);
    cmdPtr = cmdBuff->commadsCurrent;
  }
  cmdPtr[2] = 0x2001056f;
  *cmdPtr = 0x2001055b;
  *(float *)(cmdPtr + 1) = factor;
  cmdPtr[4] = 0x2001061f;
  *(float *)(cmdPtr + 3) = units + units;
  *(float *)(cmdPtr + 5) = clamp;
  cmdBuff->commadsCurrent = cmdPtr + 6;
  return;
}
```
(This was not extracted from TOTK, but the code should be the same).
You can see it doing `units + units`  which is the same as multiplying by 2. 0x56f, 0x55b and 0x61f are the offsets of the depth bias registers on the `ThreedClassState`, but in words, so you need to multiply it by 4 to get the byte offsets that would match the NVIDIA header.
If you want to understand more about how those commands are encoded, you can read: https://switchbrew.org/wiki/GPU

Deko3D (homebrew graphics API) does the same thing: https://github.com/devkitPro/deko3d/blob/master/source/maxwell/gpu_3d_state.cpp#L154
I'm not sure *why* it needs to do that though. Probably related to how it is implemented on hardware.

Serfrost:
I just wanted to mention that before your recent changes from 33 minutes ago, it resulted in Hateno houses getting a ton of flickering while walking around town. You may want to check this area when making depth changes just to make sure it works as expected. The rest of the map (at a distance) looked great though, so you're on the right track.

riperiperi:
Disabling GlobalToStorage will break all storage buffer accesses. The "optimization" passes aren't all to produce faster code, they're also to produce code that actually runs on the emulator. (global load/store is not currently supported by our buffer cache)

I don't think it will be a synchronization problem, otherwise it would randomly work.

Linear524:
I see this branch will not be merged to Master.
Maybe create some sort of mod to load it for ToTK and fix its water and geometry Z-Fighting issues ?

MetrosexualGarbodor:
Closing as this is not getting merged or worked on anymore.

PR #5014, opened 2023-05-18
Fix macOS Update Script

IsaacMarovitz:
`lsof` will return a non-zero exit code if the process with the given PID does not exist (i.e., already closed), causing the script to fail. We should wait till we're moving stuff around to do error checking.


53jk1 (approved):

AcK77 (approved):

TSRBerry (approved):
Sorry for the delay!

PR #5016, opened 2023-05-19
Avoid removing Reshade config files on application upgrade (Vulkan)

fabiograssiotto:
Reshade (https://reshade.me/) uses two configuration files on Ryunjinx main folder for the dll injection.
When upgrading the app these two files, ReShade.ini and ReShadePreset.ini should be removed from the list of files to be excluded.
This works for Vulkan for now, for OpenGL it would necessary to add the opengl.dll file that ReShade creates too.

### DIFF
--- a/src/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/src/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -742,12 +742,15 @@ namespace Ryujinx.Modules
 
             if (OperatingSystem.IsWindows())
             {
+                var reshadeFiles = Directory.EnumerateFiles(HomeDir, "ReShade*.*").ToList();
+                IEnumerable<string> filesModified = files.Except(reshadeFiles);
+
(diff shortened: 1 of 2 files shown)


AcK77:
You can sets files as hidden and they will be ignored by the updater, in that way we don’t need to change the code since there is no point of add something related to ReShade in the updater code. We do that, that would mean we have to add changes for everything which could add files/folders in the Ryujinx folder!

fabiograssiotto:
> You can sets files as hidden and they will be ignored by the updater, in that way we don’t need to change the code since there is no point of add something related to ReShade in the updater code. We do that, that would mean we have to add changes for everything which could add files/folders in the Ryujinx folder!
--

Yes, I guess it makes sense. Just did a quick test and ReShade is able to pick up the ini files when they are set as hidden.
Thank you for the quick comments.

fabiograssiotto:
Just closing since it will not be necessary.

Issue #4915, opened 2023-05-12
[Bug] TLOZ: Tears of the Kindom, Gloom effect not matching texture


IsmaelOlj:
### Description of the issue

Gloom's area of effect (the red stuff) doesn't match it's texture in the underground areas, on AMD cards, with vulkan.

### Reproduction steps

1-AMD card
2-Turn on vulkan
3-go to depths and stand on the red stuff, see that no effect will occur.

### Log file

[Ryujinx_1.1.794_2023-05-12_22-22-32.log](https://github.com/Ryujinx/Ryujinx/files/11467940/Ryujinx_1.1.794_2023-05-12_22-22-32.log)

### OS

Windows 11

### Ryujinx version

1.1.794

### Game version

1.0

### CPU

Ryzen 5 5600

### GPU

RX 570 8GB

### RAM

32GB

IsmaelOlj:
Also worth noting, this doesn't happen with OpenGL on, on that same setup.

TheShotte:
I'm not sure if it's the texture itself or what. Gloom doesn't seem to be doing any damage to link where it should be (gloom is purple on mini-map, and in world too for me), and I can step in it with no problems. However, I was exploring random parts of the Depths yesterday, and areas that did not have ANY gloom (minimap or otherwise) were killing me.

gdkchan:
Does the issue happen at native resolution (no resolution scaling)?

NotNite:
Can confirm this happens on native resolution, RX 6800 XT, Vulkan.

gdkchan:
Did you have the resolution set to native before starting the game, or did you change it with the game running?

NotNite:
Resolution was already set to native before game start, no changes to settings in gameplay. I'm in the process of starting the game after switching to OpenGL to test, I can get logs if needed.

TheShotte:
I can confirm with the resolution too. Set it to Native before I started the game, changes nothing. Fire Temple is completely undoable because you can't set foot in it.

riperiperi:
I guess everyone with this issue has an AMD gpu?

piplup55:
> I guess everyone with this issue has an AMD gpu? (windows or linux?)

it seems to be windows only, as i can't reproduce it on linux

IsmaelOlj:
Yes, I'm on windows too. Also can confirm changing resolution doesn't fix it. Only going OpenGL works, but that has serious performance issues so it's not the best solution.

[...]

PR #5026, opened 2023-05-19
SPIR-V: Only allow implicit LOD sampling on fragment

riperiperi:
Sometimes games were emitting implicit LOD texture sampling instructions, which are invalid on anything except fragment, as only fragment can calculate implicit derivatives. Compute does have an implicit derivatives mode with an NV extension, though I think more would need to be done if the switch does actually support this, as it needs to select the mode used for that (linear/quads).

On AMD windows, this caused texture sampling from compute to fail. This caused gloom to malfunction in TOTK. Other drivers likely ignored the "implicit" part and just sampled LOD 0.

Testing is welcome to make sure nothing broke, and that it works on all AMD gpus.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 4578;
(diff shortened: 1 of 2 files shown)


gdkchan (approved):
lgtm, nice catch!

gdkchan:
Closes #4915?

riperiperi:
Fixed an issue that was causing invalid shaders to be generated for other things.

Littl3Guy:
Can confirm this fixes the issue with a RX 6799xt.
I'll probably test again when the fix is built.

TheShotte:
Tested on AMD 6600XT 23.4.3 and can confirm the Gloom works as intended. Spectacular work.

Littl3Guy:
Also fixes #4915 for RX 570

marysaka (approved):

rabidcopy:
Just being a bit OCD, has anyone on AMD + Linux made sure gloom damage still works there "correctly" as it already did? Would test myself but misplaced my save for it and would have to ask around. I'd imagine it's just fine though. Either way, very nice work. I'll try to make sure it's fine on AMD + Linux later if no one else does.

PR #5027, opened 2023-05-20
Add support for VK_EXT_depth_clip_control.

cstamford:
This PR adds support for `VK_EXT_depth_clip_control` on Vulkan. It provides a significant improvement to z-fighting on distant geometry in Tears of the Kingdom, though it doesn't solve the problem entirely. Other games may see an improvement as well.

Before:

https://github.com/Ryujinx/Ryujinx/assets/6191957/fefdcd38-8fb6-49b4-b5b9-54f44c55ae2b

After:

https://github.com/Ryujinx/Ryujinx/assets/6191957/59c1a897-2428-47e7-b0e6-d45623623286

Testing note: I have confirmed via manual testing that the unhappy path works as expected and falls back to the old method when the extension isn't loaded. However, I don't have access to hardware that doesn't support `VK_EXT_depth_clip_control` so I can't try it for 100% realsies.

See also: https://github.com/Ryujinx/Ryujinx/pull/5012

Thanks to [riperiperi](https://github.com/riperiperi) for pointing me towards this.


src/Ryujinx.Graphics.Vulkan/PipelineState.cs
```diff
@@ -471,6 +478,17 @@ public void Initialize()
                     PScissors = pScissors
                 };
 
+                if (gd.Capabilities.SupportsDepthClipControl)
+                {
+                    var viewportDepthClipControlState = new PipelineViewportDepthClipControlCreateInfoEXT()
+                    {
+                        SType = StructureType.PipelineViewportDepthClipControlCreateInfoExt,
+                        NegativeOneToOne = DepthMode,
```
gdkchan, line 486:
```suggestion
                        NegativeOneToOne = DepthMode
```

gdkchan (commented):
Shader specialization will still create 2 versions of the shader if it is used with the 2 different depth modes. It might be worth to disable it when it's not needed. You could do so by not updating the value on cases where the shader doesn't need any special code depending on depth mode here https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Graphics.Gpu/Engine/Threed/SpecializationStateUpdater.cs#L162

Ultimately it's not a big deal though becaus games usually don't change depth mode. They usually pick one at the start and stick with it. So it's unlikely that it would create 2 versions of the shader anyway.

GamerzHell9137:
Mind bumping the shader cache version?

src/Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
```diff
@@ -242,7 +242,7 @@ public void PrepareForVertexReturn()
                 this.Store(StorageKind.Output, IoVariable.Position, null, Const(1), this.FPFusedMultiplyAdd(y, yScale, negativeOne));
             }
 
-            if (Config.Options.TargetApi == TargetApi.Vulkan && Config.GpuAccessor.QueryTransformDepthMinusOneToOne())
+            if (Config.Options.TargetApi == TargetApi.Vulkan && Config.GpuAccessor.QueryTransformDepthMinusOneToOne() && !Config.GpuAccessor.QueryHostSupportsDepthClipControl())
```
riperiperi, line 245: Maybe `Config.Options.TargetApi == TargetApi.Vulkan` is no longer needed here, as OpenGL will return `true` for `Config.GpuAccessor.QueryHostSupportsDepthClipControl()`.

fatjesusonabike:
Unintended side-effect of the change that I noticed while playing earlier today (this does not happen on the main build) : 

https://github.com/Ryujinx/Ryujinx/assets/134035322/f17a2400-68e5-4f9f-9985-0eaff4aa96ef

IsaacMarovitz:
Worth noting that I don't think MoltenVK supports `VK_EXT_depth_clip_control`

riperiperi:
That’s fine, it will just have to use the existing less precise mode.

src/Ryujinx.Graphics.Vulkan/PipelineState.cs
```diff
@@ -471,6 +478,17 @@ public void Initialize()
                     PScissors = pScissors
                 };
 
+                if (gd.Capabilities.SupportsDepthClipControl)
+                {
+                    var viewportDepthClipControlState = new PipelineViewportDepthClipControlCreateInfoEXT()
+                    {
+                        SType = StructureType.PipelineViewportDepthClipControlCreateInfoExt,
+                        NegativeOneToOne = DepthMode
```
marysaka, line 486: As per spec, [VkPipelineViewportDepthClipControlCreateInfoEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineViewportDepthClipControlCreateInfoEXT.html)::negativeOneToOne is only supported if [VkPhysicalDeviceDepthClipControlFeaturesEXT](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceDepthClipControlFeaturesEXT.html)::depthClipControl is true.

Your VulkanInitialization and VulkanRenderer changes should use that value to determine if `VK_EXT_depth_clip_control` is truly supported for our needs.

src/Ryujinx.Graphics.Vulkan/VulkanInitialization.cs
```diff
@@ -41,7 +41,8 @@ public unsafe static class VulkanInitialization
             "VK_EXT_subgroup_size_control",
             "VK_NV_geometry_shader_passthrough",
             "VK_NV_viewport_array2",
-            "VK_KHR_portability_subset" // As per spec, we should enable this if present.
+            "VK_KHR_portability_subset", // As per spec, we should enable this if present.
+            "VK_EXT_depth_clip_control"
```
marysaka, line 45:
```suggestion
            "VK_EXT_depth_clip_control",
            "VK_KHR_portability_subset" // As per spec, we should enable this if present.
```

I would prefer to keep VK_KHR_portability_subset as the last in the list if possible.

marysaka (changes requested):

marysaka:
> Do you prefer your PRs to be squashed into one commit?

We do squash at merge so don't worry

marysaka (approved):

riperiperi:
(note: doesn't have much to do with correctness of this PR, but more observations about depth bias)

Interestingly, there is still a difference between OpenGL and Vulkan with this PR on SMO. Look at the pedestrian crossings at the bottom left.

OpenGL (ignore framerate, debug build):
image

Master Vulkan:
image

PR Vulkan:
image

Bias on pedestrian crossing:
image

Bias on manhole covers:
image

OpenGL is an exact match for the switch, Vulkan has the pedestrian crossing and the area with the manhole covers fighting for attention. This game is using the -1 to 1 depth mode. "depthBias" correlates to "units", and "slopeScaledDepthBias" correlates to factor, where both values are the same on both backends.

The "intended" way this is meant to appear is pretty weird. The square with the manhole covers overlaps the pedestrian crossing when close to the ground, but further up the bias for the pedestrian crossing actually wins out and appears on top (from this height at least). It starts to z fight at a lower altitude and is well past doing that at the example height, is what i'm saying.

riperiperi:
SMO matches on vulkan when units are multiplied by 2. Maybe this is a werid interaction with MinusOneToOne depth mode and depth bias on Vulkan, which seems to be the same with the depth rescaling workaround and this PR.

src/Ryujinx.Graphics.Shader/Translation/EmitterContext.cs
riperiperi: If the next platform were Metal, then it wouldn't support depth clip control, and this workaround would have to be activated. Any case where we can't change the depth mode will need the workaround, so I think checking with this capability is enough.

NikkMann:
Works great to fix the z fighting but oddly seems to break vulkan reshade with ryujinx. It just crashes on game load.

Linear524:
Luigis Mansion 3 is also suffering from shadow Z-fighting issues...
Intro - wooden sign right to the road with dynamic shadow on it - flickering...

riperiperi:
This PR isn't a complaints board for existing issues. LM3 has other issues with shadows that stem from depth aliasing as red.

riperiperi (approved):
LGTM, nice work!

gdkchan:
Needs a rebase.
Also might be worth looking at the depth bias issue mentioned earlier? Although if it was already wrong before, then I guess there's no need to fix it on this PR.

MetrosexualGarbodor:
Needs another rebase.

marysaka:
> > Needs a rebase. Also might be worth looking at the depth bias issue mentioned earlier? Although if it was already wrong before, then I guess there's no need to fix it on this PR.
> 
> The depth weirdness is on my radar but I have a couple of other things I want to get to first!

Should I under this as being part of the next iteration or this PR?

PR #5028, opened 2023-05-20
Limit compute storage buffer size

gdkchan:
#4999 limits the size of storage buffers used on graphics shaders, when the base address is not sourced from constant buffer 0, since in those cases the size of the buffer is usually not supplied, it's just a raw pointer. This PR is extending that change to compute shaders.

This fixes a regression on DOOM Eternal where the game would crash because it was trying to create very large buffers. The game is still not playable however due to the other existing issues.

I also added a log on Vulkan for when buffer creation fails, since it was failing silently and causing pretty confusing exceptions later on.


riperiperi (approved):
Thanks for doing this, totally forgot about compute.

marysaka (approved):

PR #5031, opened 2023-05-20
Implement p2rc, p2ri, p2rr and r2p.cc shaders

makigumo:
Hi!
This is my attempt at implementing these shaders.
I'm basically new to this (Ryujinx, c#, shaders). So please feel free to dismiss this, if it's totally off the mark.

It does seem to satisfy these [tests](https://github.com/yuzu-emu/yuzu/pull/3799#issuecomment-619467465).
Maybe there are other tests to check the implementation against?

fixes https://github.com/Ryujinx/Ryujinx/issues/5046, fixes https://github.com/Ryujinx/Ryujinx/issues/5047

### DIFF
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmit.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmit.cs
@@ -187,27 +187,6 @@ namespace Ryujinx.Graphics.Shader.Instructions
             context.Config.GpuAccessor.Log("Shader instruction Longjmp is not implemented.");
         }
 
-        public static void P2rR(EmitterContext context)
-        {
-            InstP2rR op = context.GetOp<InstP2rR>();
-
-            context.Config.GpuAccessor.Log("Shader instruction P2rR is not implemented.");
-        }
-
-        public static void P2rI(EmitterContext context)
-        {
-            InstP2rI op = context.GetOp<InstP2rI>();
-
-            context.Config.GpuAccessor.Log("Shader instruction P2rI is not implemented.");
-        }
-
-        public static void P2rC(EmitterContext context)
-        {
-            InstP2rC op = context.GetOp<InstP2rC>();
-
-            context.Config.GpuAccessor.Log("Shader instruction P2rC is not implemented.");
-        }
-
         public static void Pexit(EmitterContext context)
         {
             InstPexit op = context.GetOp<InstPexit>();
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs
@@ -209,21 +209,15 @@ namespace Ryujinx.Graphics.Shader.Instructions
                 return context.ICompareNotEqual(context.BitwiseAnd(value, Const(1 << bit)), Const(0));
             }
 
-            if (ccpr)
+            int count = ccpr ? RegisterConsts.FlagsCount : RegisterConsts.PredsCount;
+            RegisterType type = ccpr ? RegisterType.Flag : RegisterType.Predicate;
+            int shift = (int)byteSel * 8;
+            
+            for (int bit = 0; bit < count; bit++)
             {
-                // TODO: Support Register to condition code flags copy.
-                context.Config.GpuAccessor.Log("R2P.CC not implemented.");
-            }
-            else
-            {
-                int shift = (int)byteSel * 8;
-
-                for (int bit = 0; bit < RegisterConsts.PredsCount; bit++)
-                {
-                    Operand pred = Register(bit, RegisterType.Predicate);
-                    Operand res = context.ConditionalSelect(Test(mask, bit), Test(value, bit + shift), pred);
-                    context.Copy(pred, res);
-                }
+                Operand flag = Register(bit, type);
+                Operand res = context.ConditionalSelect(Test(mask, bit), Test(value, bit + shift), flag);
+                context.Copy(flag, res);
             }
         }
 
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
@@ -1,10 +1,10 @@
 using Ryujinx.Graphics.Shader.Decoders;
 using Ryujinx.Graphics.Shader.IntermediateRepresentation;
 using Ryujinx.Graphics.Shader.Translation;
-
 using static Ryujinx.Graphics.Shader.Instructions.InstEmitAluHelper;
 using static Ryujinx.Graphics.Shader.Instructions.InstEmitHelper;
 using static Ryujinx.Graphics.Shader.IntermediateRepresentation.OperandHelper;
+using Register = Ryujinx.Graphics.Shader.Decoders.Register;
 
 namespace Ryujinx.Graphics.Shader.Instructions
 {
@@ -49,6 +49,74 @@ namespace Ryujinx.Graphics.Shader.Instructions
 
             context.Copy(Register(op.DestPred, RegisterType.Predicate), p0Res);
             context.Copy(Register(op.DestPredInv, RegisterType.Predicate), p1Res);
+        }
+        
+        public static void P2rC(EmitterContext context)
+        {
+            InstP2rC op = context.GetOp<InstP2rC>();
+
+            Operand srcA = GetSrcReg(context, op.SrcA);
+            Operand dest = GetSrcReg(context, op.Dest);
+            Operand mask = GetSrcCbuf(context, op.CbufSlot, op.CbufOffset);
+            
+            EmitP2r(context, srcA, dest, mask, op.ByteSel, op.Ccpr);
+        }
+        
+        public static void P2rI(EmitterContext context)
+        {
+            InstP2rI op = context.GetOp<InstP2rI>();
+
+            Operand srcA = GetSrcReg(context, op.SrcA);
+            Operand dest = GetSrcReg(context, op.Dest);
+            Operand mask = Const(op.Imm20 & 0xff);
+            
+            EmitP2r(context, srcA, dest, mask, op.ByteSel, op.Ccpr);
+        }
+
+        public static void P2rR(EmitterContext context)
+        {
+            InstP2rR op = context.GetOp<InstP2rR>();
+
+            Operand srcA = GetSrcReg(context, op.SrcA);
+            Operand dest = GetSrcReg(context, op.Dest);
+            Operand mask = GetSrcReg(context, op.SrcB);
+            
+            EmitP2r(context, srcA, dest, mask, op.ByteSel, op.Ccpr);
+        }
+        
+        private static void EmitP2r(
+            EmitterContext context,
+            Operand srcA,
+            Operand dest,
+            Operand mask,
+            ByteSel byteSel,
+            bool ccpr
+        )
+        {
+            Operand Test(Operand value, int bit)
+            {
+                return context.ICompareNotEqual(context.BitwiseAnd(value, Const(1 << bit)), Const(0));
+            }
+            
+            int count = ccpr ? RegisterConsts.FlagsCount : RegisterConsts.PredsCount;
+            int shift = (int)byteSel * 8;
+
+            Operand insert = Const(0);
+            for (int i = 0; i < count; i++)
+            {
+                Operand condition = ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
+                insert = context.ConditionalSelect(Test(context.ShiftRightU32(mask, Const(1)), i), context.BitwiseOr(insert, bit), insert);
+            }
+
+            Operand masked = context.BitwiseAnd(srcA, context.INegate(context.ShiftLeft(mask,Const( shift))));
+
+            Operand res = context.BitwiseOr(masked, insert);
+
+            context.Copy(dest, res);
         }
     }
 }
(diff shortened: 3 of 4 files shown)


src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,38 @@ public static void Psetp(EmitterContext context)
...
+
+        public static void P2rI(EmitterContext context)
+        {
+            InstP2rI op = context.GetOp<InstP2rI>();
+
+            int mask = op.Imm20;
+            int count = op.Ccpr ? RegisterConsts.FlagsCount : RegisterConsts.PredsCount;
+            int shift = (int)op.ByteSel * 8;
+
+            Operand insert = GetSrcImm(context, 0);
```
gdkchan, line 62:
```suggestion
            Operand insert = Const(0);
```

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,38 @@ public static void Psetp(EmitterContext context)
...
+                Operand condition = op.Ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+                
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
+                insert = context.BitwiseOr(insert, bit);
+            }
+
+            Operand srcA = context.BitwiseNot(Register(op.SrcA, RegisterType.Gpr), false);
+            Operand masked = context.BitwiseAnd(srcA, GetSrcImm(context, ~(mask << shift)));
```
gdkchan, line 79:
```suggestion
            Operand masked = context.BitwiseAnd(srcA, Const(~(mask << shift)));
```

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,38 @@ public static void Psetp(EmitterContext context)
...
+
+                Operand condition = op.Ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+                
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
+                insert = context.BitwiseOr(insert, bit);
+            }
+
+            Operand srcA = context.BitwiseNot(Register(op.SrcA, RegisterType.Gpr), false);
```
gdkchan, line 78: If you pass `false` it will just return the original value, so what's the point of calling `BitwiseNot` here?

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,38 @@ public static void Psetp(EmitterContext context)
             context.Copy(Register(op.DestPred, RegisterType.Predicate), p0Res);
             context.Copy(Register(op.DestPredInv, RegisterType.Predicate), p1Res);
         }
+
+        public static void P2rI(EmitterContext context)
+        {
+            InstP2rI op = context.GetOp<InstP2rI>();
+
+            int mask = op.Imm20;
```
gdkchan, line 58: Should probably do `& 0xff` on that constant? It doesn't make sense to modify more than 8 bits on the register due to the byte selector.

src/Ryujinx.Graphics.Shader/Instructions/InstEmit.cs
```diff
@@ -193,14 +193,7 @@ public static void P2rR(EmitterContext context)
...
-
-        public static void P2rI(EmitterContext context)
-        {
-            InstP2rI op = context.GetOp<InstP2rI>();
-
-            context.Config.GpuAccessor.Log("Shader instruction P2rI is not implemented.");
-        }
-
+        
         public static void P2rC(EmitterContext context)
```
gdkchan, line 197: It would be nice if you implemented the other variants as well. The only thing that changes is where the mask comes from (constant, register or constant buffer), so it should be easy to implement with your existing code.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs
```diff
@@ -211,8 +211,12 @@ Operand Test(Operand value, int bit)
 
             if (ccpr)
             {
-                // TODO: Support Register to condition code flags copy.
-                context.Config.GpuAccessor.Log("R2P.CC not implemented.");
+                for (int bit = 0; bit < RegisterConsts.FlagsCount; bit++)
+                {
+                    Operand flag = Register(bit, RegisterType.Flag);
+                    Operand res = context.ConditionalSelect(Test(mask, bit),  Test(value, bit), flag);
```
gdkchan, line 217:
```suggestion
                    Operand res = context.ConditionalSelect(Test(mask, bit), Test(value, bit), flag);
```

gdkchan: Also I believe you need to take `byteSel` into account here as well, just like the case below.

gdkchan (commented):
https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs#L25
You need to change the `CodeGenVersion` to the PR number (5031) to force a shader cache rebuild, since the shader code changed.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs
makigumo: Yes, you are right. As this makes both cases pretty similar, would it be ok to squish them like so?
```c#
            int count = ccpr ? RegisterConsts.FlagsCount : RegisterConsts.PredsCount;
            RegisterType type = ccpr ? RegisterType.Flag : RegisterType.Predicate;
            int shift = (int)byteSel * 8;
            
            for (int bit = 0; bit < count; bit++)
            {
                Operand flag = Register(bit, type);
                Operand res = context.ConditionalSelect(Test(mask, bit), Test(value, bit + shift), flag);
                context.Copy(flag, res);
            }
```

gdkchan: Yes, it is fine.

src/Ryujinx.Graphics.Shader/Instructions/InstEmit.cs
makigumo: I will try 👍

makigumo: Sorry, stupid question.
Can I do something like this to get the actual mask value?
```C#
            Operand srcB = GetSrcReg(context, op.SrcB);
            int mask = srcB.Value;
```
Or this?
```C#
            Operand cbuf = GetSrcCbuf(context, op.CbufSlot, op.CbufOffset);
            int mask = cbuf.Value;
```

gdkchan: No, you should pass the `Operand` directly. For register, `Value` in this case is not the actual register value, but the register number. And for constant buffers, it is the packed constant buffer number and offset. The `Operand` object is what represents the value. For constants, it is indeed the actual constant value.

makigumo: I implemented the other variants. Without a test case I'm a little unsure I got this right.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,73 @@ public static void Psetp(EmitterContext context)
...
+            {
+                Operand condition = ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
+                insert = context.ConditionalSelect(Test(context.ShiftRightU32(mask, Const(1)), i), context.BitwiseOr(insert, bit), insert);
+            }
+
+            Operand masked = context.BitwiseAnd(srcA, context.INegate(context.ShiftLeft(mask,Const( shift))));
```
gdkchan, line 115:
```suggestion
            Operand masked = context.BitwiseAnd(srcA, context.INegate(context.ShiftLeft(mask, Const(shift))));
```

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,73 @@ public static void Psetp(EmitterContext context)
...
+            EmitP2r(context, srcA, dest, mask, op.ByteSel, op.Ccpr);
+        }
+        
+        public static void P2rI(EmitterContext context)
+        {
+            InstP2rI op = context.GetOp<InstP2rI>();
+
+            Operand srcA = GetSrcReg(context, op.SrcA);
+            Operand dest = GetSrcReg(context, op.Dest);
+            Operand mask = Const(op.Imm20 & 0xff);
```
gdkchan, line 71: This is not wrong, but I'd prefer if you did `GetSrcImm(context, op.Imm20)` for consistency with the other instructions.
BTW, the `& 0xff` should be done on `EmitP2r` as it applies to all variants. You would need to do it at the IR level like so `mask = context.BitwiseAnd(mask, Const(0xff));`. The optimizer can constant fold this to a single constant for the immediate variant (where mask is always constant).

makigumo: Thank you, I changed it as suggested.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +50,74 @@ public static void Psetp(EmitterContext context)
...
+            {
+                Operand condition = ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
+                insert = context.ConditionalSelect(Test(context.ShiftRightU32(mask, Const(1)), i), context.BitwiseOr(insert, bit), insert);
+            }
+
+            Operand masked = context.BitwiseAnd(srcA, context.INegate(context.ShiftLeft(mask, Const(shift))));
```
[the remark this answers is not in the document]
gdkchan: You still need to mask insert before that if you follow my other suggestion to remove `ConditionalSelect`. `insert = context.BitwiseAnd(insert, mask);`.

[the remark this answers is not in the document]
gdkchan: Also you'd need to change `Const(1 << (i + shift))` to `Const(1 << (i))` since bitfield insert will take the least significant bits of `insert`.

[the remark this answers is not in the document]
makigumo: Tried the BitfieldInsert approach. Makes it look a lot cleaner.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +49,67 @@ public static void Psetp(EmitterContext context)
...
+            mask = context.BitwiseAnd(mask, Const(0xff));
+            
+            Operand insert = Const(0);
+            for (int i = 0; i < count; i++)
+            {
+                Operand condition = ccpr
+                    ? Register(i, RegisterType.Flag)
+                    : Register(i, RegisterType.Predicate);
+
+                Operand bit = context.ConditionalSelect(condition, Const(1 << (i)), Const(0));
```
gdkchan, line 105:
```suggestion
                Operand bit = context.ConditionalSelect(condition, Const(1 << (i + shift)), Const(0));
```

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
```diff
@@ -50,5 +49,67 @@ public static void Psetp(EmitterContext context)
...
+                insert = context.BitwiseOr(insert, bit);
+            }
+
+            insert = context.BitwiseAnd(insert, mask);
```
gdkchan, line 109:
```cs
Operand maskShifted = context.ShiftLeft(mask, Const(shift));
Operand masked = context.BitwiseAnd(srcA, context.BitwiseNot(maskShifted));
Operand res = context.BitwiseOr(masked, context.BitwiseAnd(insert, maskShifted));
```

gdkchan (commented):
Sorry, I made a mistake yesterday when I suggested the use of `BitfieldInsert`. I forgot that the bits that are not set in the `mask` should be sourced from `srcA` rather than the `insert` value. So the bitfield insert approach doesn't work for this case. I added suggestions with the code you can use instead to do the right thing. It should be good to go once this code is added.

src/Ryujinx.Graphics.Shader/Instructions/InstEmitPredicate.cs
makigumo: No problem. Thank you again, I changed the code as suggested.

gdkchan (approved):
lgtm, thanks. Would be nice to get test results on games with shaders using those instructions (did anything change on them?).

lostromb:
Yeah, as a tester it would be nice to know of any games that use these instructions (would they show warnings on the console the same way that VOTEVG warning currently does?)

ExE-Boss:
**The Legend of Zelda: Tears of the Kingdom**[^1] uses these instructions.

[^1]: Ryujinx/Ryujinx-Games-List#4598

riperiperi (approved):
Nice work!

Cell-Block73:
The release changelog that mentions this PR states that implementing these shaders has an unknow effect in The Legend of Zelda: Tears of the Kingdom. I'm just here to say that I did play the game a decent amount yesterday and noticed a lot more subtle and pleasant effects being properly rendered in such as sparkles around heart and stamina containers, fire ashes, water slashes while cooking, airborne dust in the depths, halo effects in depths light shrines, and a handful of others... Great work!

Issue #5047, opened 2023-05-22
[GPU] Missing shader instruction: R2P.CC


ExE-Boss:
### Shader instruction

Shader translator: R2P.CC not implemented.

### Required by

- <https://github.com/Ryujinx/Ryujinx-Games-List/issues/4598>

### Additional context?

<https://github.com/Ryujinx/Ryujinx/blob/fc26189fe1338ffcba12d83a922da9c706738902/src/Ryujinx.Graphics.Shader/Instructions/InstEmitMove.cs#L212-L216>

Issue #5046, opened 2023-05-22
[GPU] Missing shader instruction: P2rI


ExE-Boss:
### Shader instruction

Shader translator: Shader instruction P2rI is not implemented.

### Required by

- <https://github.com/Ryujinx/Ryujinx-Games-List/issues/4598>

### Additional context?
<https://github.com/Ryujinx/Ryujinx/blob/fc26189fe1338ffcba12d83a922da9c706738902/src/Ryujinx.Graphics.Shader/Instructions/InstEmit.cs#L197-L202>

PR #5048, opened 2023-05-22
Add issue template for missing shader instructions

ExE-Boss:
For issues like:
- <https://github.com/Ryujinx/Ryujinx/issues/2794>
- <https://github.com/Ryujinx/Ryujinx/issues/5046>
- <https://github.com/Ryujinx/Ryujinx/issues/5047>

### DIFF
--- a/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
+++ b/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
@@ -11,13 +11,6 @@ body:
     validations:
       required: true
   - type: textarea
-    id: name
-    attributes:
-      label: Instruction description
-      description: Include the description/explanation from _tbd_ in this field
-    validations:
-      required: true
-  - type: textarea
     id: required
     attributes:
       label: Required by
--- a/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
+++ b/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
@@ -13,8 +13,8 @@ body:
   - type: textarea
     id: name
     attributes:
-      label: Instruction name
-      description: Include the name from _tbd_ in the above code block
+      label: Instruction description
+      description: Include the description/explanation from _tbd_ in this field
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
+++ b/.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
@@ -1,0 +1,26 @@
+name: Missing Shader Instruction
+description: Shader Instruction is missing in Ryujinx.
+title: "[GPU]"
+labels: [gpu, not-implemented]
+body:
+  - type: textarea
+    id: instruction
+    attributes:
+      label: Shader instruction
+      description: What shader instruction is missing?
+    validations:
+      required: true
+  - type: textarea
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
```diff
@@ -0,0 +1,26 @@
...
+      description: What shader instruction is missing?
+    validations:
+      required: true
+  - type: textarea
+    id: name
+    attributes:
+      label: Instruction name
+      description: Include the name from _tbd_ in the above code block
+    validations:
+      required: true
```
TSRBerry, line 19: This is the same as the textarea above.

TSRBerry (changes requested):
Could you remove the duplicated textarea from the form?

Lgtm otherwise!

.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
AcK77: This is just a copy paste from https://github.com/Ryujinx/Ryujinx/blob/master/.github/ISSUE_TEMPLATE/missing_cpu_instruction.yml?plain=1 so the only missing thing is probably the `tbd` which could be the log file or if our GPU wizards know where to find a list of shaders instructions, it could be that too.

ExE-Boss: I changed it to “Instruction description” to match what `missing_service_call.yml` does: <https://github.com/Ryujinx/Ryujinx/blob/4ca78eded52f21089400cc28351b9353279b8171/.github/ISSUE_TEMPLATE/missing_service_call.yml?plain=1#L12-L18>

.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
```diff
@@ -0,0 +1,26 @@
...
+      description: What shader instruction is missing?
+    validations:
+      required: true
+  - type: textarea
+    id: name
+    attributes:
+      label: Instruction description
+      description: Include the description/explanation from _tbd_ in this field
+    validations:
+      required: true
```
TSRBerry, line 19: I still think this field should be removed, unless @riperiperi or @gdkchan think we could use it to add useful information about missing shader instructions, but in that case the description should also be changed.

TSRBerry (changes requested):
Pinged the gpu wizards, so let's wait for clarification.

.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
gdkchan: The emulator already prints the instruction name in the log when there is a missing instruction (unlike the CPU emulator that does not know the name of instructions that are not implemented). So one can just get it from the log. There is no public documentation with instruction descriptions, so there's nowhere to get it from, that part should be just removed.

.github/ISSUE_TEMPLATE/missing_shader_instruction.yml
marysaka: I think it should be removed imo yup

AcK77, line 19:
```suggestion
```

marysaka (approved):

PR #5049, opened 2023-05-22
Fix unnecessary dylib in Windows release

### LINKED ISSUE #4967: [Bug] unnessicary dylib in windows release
### Description of the issue

in the windows zip on the website, theres a file called libarmeilleure-jitsupport.dylib, which is a macos arm64 dynamic library, and is useless on windows.

### Reproduction steps

download windows zip
look inside
libarmeilleure-jitsupport.dylib

### Log file

N/A

### OS

Windows 10

### Ryujinx version

1.1.800

### Game version

N/A

### CPU

N/A

### GPU

N/A

### RAM

N/A

### List of applied mods

_No response_

### Additional context?

_No response_

marcantoinegodde:
This PR fixes #4967. The ARMeilleure config is now more specific when specifying `osx-arm64`.

NB: this is my first PR :)

### DIFF
--- a/src/ARMeilleure/ARMeilleure.csproj
+++ b/src/ARMeilleure/ARMeilleure.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <RuntimeIdentifiers>win10-x64;linux-x64;osx-x64;osx-arm64</RuntimeIdentifiers>
   </PropertyGroup>
 
   <ItemGroup>
@@ -11,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ContentWithTargetPath Include="Native\libs\libarmeilleure-jitsupport.dylib" Condition="'$(RuntimeIdentifier)' == '' OR '$(RuntimeIdentifier)' == 'osx-arm64'">
+    <ContentWithTargetPath Include="Native\libs\libarmeilleure-jitsupport.dylib" Condition="'$(RuntimeIdentifier)' == 'osx-arm64'">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <TargetPath>libarmeilleure-jitsupport.dylib</TargetPath>
     </ContentWithTargetPath>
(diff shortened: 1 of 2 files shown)


src/ARMeilleure/ARMeilleure.csproj
```diff
@@ -11,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ContentWithTargetPath Include="Native\libs\libarmeilleure-jitsupport.dylib" Condition="'$(RuntimeIdentifier)' == '' OR '$(RuntimeIdentifier)' == 'osx-arm64'">
+    <ContentWithTargetPath Include="Native\libs\libarmeilleure-jitsupport.dylib" Condition="'$(RuntimeIdentifier)' == 'osx-arm64'">
```
TSRBerry, line 15: This will prevent local builds from including the native library.

src/ARMeilleure/ARMeilleure.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <RuntimeIdentifiers>win10-x64;linux-x64;osx-x64;osx-arm64</RuntimeIdentifiers>
```
TSRBerry, line 6: This is not needed.

TSRBerry (changes requested):
Thank you for this PR! Unfortunately this fix prevents local builds from working correctly.
Could you take another look?

src/ARMeilleure/ARMeilleure.csproj
marcantoinegodde: What do you mean ? :) If I understood well, this lib is only required on osx-arm64 and it is included when I use the `create_macos_build.sh` script

marysaka: When you build locally `RuntimeIdentifier` will be empty.

Thought I think most of this is broken @TSRBerry as when I tried to fix that last week, RuntimeIdentifier was always empty when entering this part of the project... meaning that all our usage of ContentWithTargetPath might be broken.

However, I remember this working with .NET Core 3.1 (or even lower so that might be regression on msbuild side...

marcantoinegodde: Indeed, I faced the issue with RuntimeIdentifier. Yet, adding `<RuntimeIdentifiers>win10-x64;linux-x64;osx-x64;osx-arm64</RuntimeIdentifiers>` makes it working with me. I have different output if I specify `-r "osx-arm64"` or `-r "osx-x64"`. Am I right ?

marysaka: Will need to do some testing on that side later...

PR #5056, opened 2023-05-22
Update LastPlayed date on emulation end.

tarrouye:
This is a pretty simple change that just makes it so the last played date will be updated on emulation end when the time played stat is updated, rather than only when the application was launched. To me, this behavior better reflects the name of the stat, as the user last played the application right when they closed it. 

Apologies if I missed anything here, this is my first PR on this project. I did a cursory look through the source code to see if this value was used anywhere other than UI and as far as I could tell, it isn't. 

Thanks !


marysaka (approved):
lgtm, I don't know why this wasn't the default behavior yet :+1:

TSRBerry (approved):
Thank you!

PR #5060, opened 2023-05-22
Vulkan: Do not set storage flag for multisample textures if not supported

gdkchan:
We shoud not set storage usage flag for multisample textures if the host does not support it, since that's invalid.
This fixes a crash that started happening on macOS after #5057 on a few games that uses multisample textures, like Dark Souls Remastered.


riperiperi (approved):
Looks sensible, even if the latest MVK does have other issues we need to resolve. It's worth noting that mobile GPUs have a bunch of weird performance pitfalls, so it might be a good idea in future to have some way of recreating a texture with different usage flags (mutable format/storage) only when it's used in that way. Obviously, such a thing would only be enabled on an affected GPU.

marysaka (approved):

PR #5064, opened 2023-05-22
Improve macOS updater

siegmund-heiss-ich:
In the current version of the `updater.sh` I had an issue, where the update was downloaded, but the Ryujinx.app never got replaced at first try.

I had a look at the script and found the following issue:
Because `set -e` is defined, every command with an exit status other than 0 will cause the script to stop.
In my case, the `lsof`command exited with a status > 0, because there seemed to be still fds open.
The `sleep 1` after that had no effect, because the script exited already.

My approach is to check if the PID still exists in the process table additionally to check for open fds.
If the process is closed correctly and has no open fds, the script continues like it has before.
There will be a 1 second delay until there is a new try (up to 5 times).
If it fails, the script will exit with code 1.

Also the error handling was moved up again, because it makes no sense to display anything when there is no UI anymore.

I am happy for any suggestions to further improve this script.

### DIFF
--- a/distribution/macos/updater.sh
+++ b/distribution/macos/updater.sh
@@ -28,21 +28,25 @@ error_handler() {
 trap 'error_handler ${LINENO}' ERR
 
 # Wait for Ryujinx to exit
-# NOTE: in case no fds are open, lsof could be returning with a process still living.
-# If the process is still acitve, we wait for 1 second and check it again.
+# If the main process as well as child processes are still acitve, we wait for 1 second and check it again.
 # After the third time checking, this script exits with status 1
 
 attempt=0
+
 while true; do
-    if lsof -p $APP_PID +r 1 &>/dev/null || kill -0 "$APP_PID" 2>/dev/null; then
-        if [ $attempt -eq 2 ]; then
+    process_status=$(ps -p "$APP_PID" -o state=)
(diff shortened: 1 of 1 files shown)


noahgg12:
I have the same issue

distribution/macos/updater.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
-
```
noahgg12, line 4: issue

noahgg12 (commented):
help with the updater

siegmund-heiss-ich:
> There is still a case when the application could be running without any fds... we should probably still wait one full second after lsof run imo

Thx for the feedback! I also added a check for the PID.. This should work, right?
That way we only need to wait if anything is still active, otherwise we could proceed without waiting and any issues.

siegmund-heiss-ich:
I implemented a new approach. This looks much cleaner.
Now lsof is removed because if all processes have exited, it is not possible for file descriptors associated with those processes to remain open. 
Instead of that, the current status of the PID gets stored in a variable.
There is also a variable that will contain running child processes.
If there is no process left, both variables will be empty, and the update proceeds.

marysaka (approved):
seems fine, hopefully that will be enough to fix the issue.

TSRBerry (approved):

PR #5073, opened 2023-05-23
Update to MoltenVK 1.2.5

IsaacMarovitz:
Waiting on https://github.com/KhronosGroup/MoltenVK/issues/1915 (workaround included in this PR) and https://github.com/KhronosGroup/MoltenVK/issues/1919

Should fix MVK bugs noted here https://github.com/Ryujinx/Ryujinx/issues/4062


gdkchan:
FWIW we can simply disable the extension on MoltenVK as work around for this issue to unblock the update. But we will likely need extensive testing before any update is merged.

PR #5077, opened 2023-05-23
Updater: Keep reshade files and opengl hook intact

Kurochi51:
It's done in a very simple manner, I wasn't getting far with testing on my end, but at the very least it stopped appending the intended files with `.ryuold`. I made the check case insensitive because reshade itself doesn't seem to care, and it's feasable to imagine that config files shared by users wouldn't have the standard `ReShade.ini` naming.

### DIFF
--- a/src/Ryujinx/Modules/Updater/Updater.cs
+++ b/src/Ryujinx/Modules/Updater/Updater.cs
@@ -473,6 +473,8 @@ namespace Ryujinx.Modules
             {
                 foreach (string file in allFiles)
                 {
+                    if (file.Contains("ReShade", StringComparison.OrdinalIgnoreCase) || file.Contains("opengl32", StringComparison.OrdinalIgnoreCase))
+                        break;
                     try
                     {
                         File.Move(file, file + ".ryuold");


MutantAura:
As per https://github.com/Ryujinx/Ryujinx/pull/5016 you can already mark any files as hidden and they won't get deleted by the updater.

Kurochi51:
> As I mentioned before, I'd prefer a general fix for the issue, otherwise we will be getting this type of PR forever, where the updater is changed to blacklist files for every single use case.

A general fix as in, the updater handling its own files, and ignoring everything else? That would just be hardcoding the loose files shipped with ryujinx, no? If that's the case, I can do that.

gdkchan:
IMO the best solution is making the updater *replace* the old files with the new files in the update package. Everything else it can just leave untouched. The only problem with that is that it may leave behind files that are no longer needed on a newer version of the emulator. But I don't think this is a big deal, it is certainly better to leave unnecessary files than deleting files that the user may still want without the user consent.

IsaacMarovitz:
> IMO the best solution is making the updater _replace_ the old files with the new files in the update package. Everything else it can just leave untouched. The only problem with that is that it may leave behind files that are no longer needed on a newer version of the emulator. But I don't think this is a big deal, it is certainly better to leave unnecessary files than deleting files that the user may still want without the user consent.

Yeah I agree with gdk on this

PR #5093, opened 2023-05-24
Cache ApplicationMetadata in Memory

patrick-hovsepian:
`ApplicationMetadata` was being read from and written to disk on every call to `LoadAndSaveMetaData`. Although this change keeps the write logic, it delegates it to a background task and doesn't block on it's completion. It also reads the metadata from disk once (if it exists) and caches it in memory.

There was also another minor bug found where the game directories were being tested for reference equality, not content equality to force a game list refresh on change.

The final change was removing the explicit Thread that was created to update the game table and instead use the Task mechanism to run in the existing threadpool. 

Still new to this whole thing so please let me know if I can explain anything further or if there's anything more/less I need to do 😸


src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -506,26 +514,27 @@ public ApplicationMetadata LoadAndSaveMetaData(string titleId, Action<Applicatio
             string metadataFolder = Path.Combine(AppDataManager.GamesDirPath, titleId, "gui");
             string metadataFile   = Path.Combine(metadataFolder, "metadata.json");
 
-            ApplicationMetadata appMetadata;
+            bool isCached = _metadataCache.TryGetValue(metadataFile, out ApplicationMetadata appMetadata);
 
-            if (!File.Exists(metadataFile))
+            if (!isCached)
```
TSRBerry, line 519:
```suggestion
            if (!_metadataCache.TryGetValue(metadataFile, out ApplicationMetadata appMetadata))
```

Since you just need the value of `isCached` once, I'd inline it.

TSRBerry (commented):
What happens if the cached file gets modified? Would the cache be updated as well?

As far as I understand, the current code would just cache the first read, so changes would never be seen on a new load.

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
patrick-hovsepian: inlined. Also, it's reset on line 93 when a read from disk occurs.

There's no explicit cache file it's just reading the existing metadata entries as they're already written

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -506,26 +514,25 @@ public ApplicationMetadata LoadAndSaveMetaData(string titleId, Action<Applicatio
...
-            }
+                try
+                {
+                    appMetadata = JsonHelper.DeserializeFromFile(metadataFile, SerializerContext.ApplicationMetadata);
+                }
+                catch (Exception ex) when (ex is JsonException || ex is DirectoryNotFoundException)
+                {
+                    if (ex is DirectoryNotFoundException)
+                    {
+                        Directory.CreateDirectory(metadataFolder);
```
AcK77, line 527: So the metadata directory is created and ...? Because the JSON serialization is just done when metadata is modified so I don’t get it.

AcK77 (commented):
Does this do any noticeable speed differences between disk cache and memory cache?

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
patrick-hovsepian: It keeps parity with what was there before. 

Every time this method was called prior to this change it would perform at least one read from disk and then a write to disk. The structure seems small enough to just keep in memory and trade off the win for less IO and faster method execution time. It'd make a difference especially if users are running on mechanical drives.

Issue #4949, opened 2023-05-14
[Bug] Vertex explosions and bright white texture artifacts in ToTK macOS


cobbycobbycobcob:
### Description of the issue

Certain colours (dark grey) appear glowing white, and textures (specifically Link in the 'pause screen', Link's shadow, the chu-chu enemies) are distorted.
<img width="1249" alt="Screen Shot 2023-05-13 at 12 57 36 PM" src="https://github.com/Ryujinx/Ryujinx/assets/88045608/15816d16-5b7c-4a0e-965b-0affbf5347ea">
- Dark grey as glowing white

<img width="1274" alt="Screenshot 2023-05-14 at 10 34 18 AM" src="https://github.com/Ryujinx/Ryujinx/assets/88045608/b1f93158-934a-4445-8bab-8f0f16aa5f84">
- Distorted Link

<img width="1250" alt="Screenshot 2023-05-14 at 10 33 17 AM" src="https://github.com/Ryujinx/Ryujinx/assets/88045608/fa02480e-5238-4875-98b2-456591bdadff">
- Distorted shadow

### Reproduction steps

- Load base .xci
- Load v 1.1.0 .nsp
- Compile + package shaders (n = 13494)
- Play the game, notice the glitches

### Log file

[Ryujinx_1.1.799_2023-05-14_00-34-01.log](https://github.com/Ryujinx/Ryujinx/files/11472430/Ryujinx_1.1.799_2023-05-14_00-34-01.log)

### OS

Mac OS Ventura 13.3.1

### Ryujinx version

1.1.799

### Game version

1.1.0

### CPU

Apple M2

### GPU

Apple M2 (Vulkan)

### RAM

16 GB

d9j:
And certain enemies glitching out like Chuchu for example.. they look totally messed up

xiongpahao:
Same problems on my 14‘’ M1 Pro MacBook Pro.

Foul-Tarnished:
`Download and try various shaders (n = 13494)`
Sharing shaders is illegal, as they contain copyrighted material.

Also, they are compiled and optimized for each gpu, so not really a good idea to even just copy them between your *own* computers either way.

Pbird99:
Same problems here. It seems the bug with Link's hair in the menu and the shadow for it doesn't occur when wearing armour.

IsaacMarovitz:
[...]

PR #5094, opened 2023-05-25
Truncate vertex attribute format if it exceeds stride on MoltenVK

gdkchan:
Zelda Tears of the Kingdom has a few cases where it uses a 4 components format instead of a 3 components format, but everything (offset, buffer stride) are calculated for a 3 components format, and the vertex shader also only accesses 3. The fact that it uses 4 causes the vertex attribute to exceed the stride in some cases. This is invalid and causes weird bugs on Metal.

It seems MoltenVK already has some code to work around this issue:
https://github.com/KhronosGroup/MoltenVK/blob/b3ae0f48d0170570e46c7b3a8bcd8278bb71f40e/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm#L1311

However it does not take offset into account, only the format size, so it will only kick in for attributes with a offset of zero. This PR implements the same change on the emulator, but it also adds the offset to the size, so it applies to all cases where the attribute exceeds the stride, not just cases where the offset is 0 and there's no other attribute after.

Fixes vertex explosions on The Legend of Zelda: Tears of the Kingdom.
Before:
Captura de Tela 2023-05-24 às 22 25 10
Captura de Tela 2023-05-24 às 22 25 16
After:
Captura de Tela 2023-05-24 às 22 18 40
Captura de Tela 2023-05-24 às 22 18 49

Fixes vertex explosions reported on #4949.


vesper8:
The hero we all need but don't deserve <3

marysaka (approved):

noahgg12:
Are we merging this to the nightly builds?

JustinSUF:
hi im dumb

riperiperi (approved):
Nice catch, this makes sense.

gdkchan:
> Nice catch, this makes sense. Maybe the format size method should be in FormatExtensions?

They work with the Vulkan `VkFormat` enum, not the `Format` enum in the GAL project. So it wouldn't make sense to add it there, but I can move it to the Vulkan `FormatTable` class.

khalidahusain:
Hey, i'm tyring to download it, but I'm not able to using the MacOS link or any other.

PR #5080, opened 2023-05-24
Implement transform feedback emulation for hardware without native support

gdkchan:
This implements partial support for transform feedback emulation, for platforms that do not have native support for this feature (like macOS).

Test results
-
I tested some games that uses transform feedback locally to validate that the implementation is working. Below I show the results and also some observations regarding issues that I found.

**Pokémon Scarlet/Violet:**
Captura de Tela 2023-05-23 às 20 25 06
This test was done with buffer mirrors (#4899). ~~Without it, the performance is abysmal and for some reason, some objects are flickering (maybe related to the low performance)~~ (The performance issue was fixed on 5447548610a45c5305cc5affcc07aadbbaec5862). No `InvalidResource` error with and without mirrors.

**Pokémon Legends Arceus:**
Captura de Tela 2023-06-08 às 15 50 00
~~This was tested without buffer mirrors.~~ (Can also reach 30 fps without mirrors now). With buffer mirros, it can reach 30 fps, but there are a lot of `InvalidResource` errors which prevents the game from rendering properly, so I don't recommend right now.

**SNK Heroines: Tag Team Frenzy:**
Captura de Tela 2023-05-23 às 20 33 47
Tons of `InvalidResource` error with and without mirrors. But the characters does render when it is not erroring, which shows that transform feedback is working. Can reach 60 fps with and without mirrors.

**Xenoblade Chronicles Definitive Edition:**
Captura de Tela 2023-05-23 às 20 41 32
Title screen render fine (the grass uses transform feedback). In-game it starts having a lot of `InvalidResource` errors and does not render properly unfortunately:
Captura de Tela 2023-05-23 às 20 43 22
I did not test with mirrors.

**Pokkén Tournament:**
Captura de Tela 2023-05-23 às 20 47 11
Captura de Tela 2023-05-23 às 20 47 21
Works, no `InvalidResource` errors, can reach 60 fps. I did not test with mirrors.

**Metroid Prime Remastered:**
Captura de Tela 2023-05-23 às 21 16 43
What is uses transform feedback for (Samus shadows) is not really working since it does it from a geometry shader. But at least the game doesn't crash anymore.

**Donkey Kong Country: Tropical Freeze:**
Captura de Tela 2023-05-25 às 23 34 31
Tested without mirrors, looks fine and runs at 60 fps the entire time.

Note: All tests performed on 2020 MacBook Air.

What's still not supported:
-

So I mentioned before that this is *partial* support, because there are a few cases that are not yet supported:
[...]

PR #5110, opened 2023-05-26
Force reciprocal operation with value biased by constant to be precise on macOS

### LINKED ISSUE #4949: [Bug] Vertex explosions and bright white texture artifacts in ToTK macOS

gdkchan:
There are a few cases where shaders add a small constant value before doing a reciprocal (`1.0 / x`) operation, to avoid a division by zero. When generating instructions without the "precise" qualifier on SPIR-V, such operations would be allowed to undergo optimizations that may reduce its precision, and if the bias constant value is low enough (near zero), it might get rounded down to zero and become useless. In this scenario, a division by zero would occur and generate invalid values.

To prevent that scenario, this change forces the add operation on the pattern of `1.0 / (x + CONSTANT_BIAS)` to be precise, preventing the value from ever being 0 assuming `x` is not negative.

This fixes the The Legend of Zelda: Tears of the Kingdom having super bright characters on macOS in some cases.
Before:
Captura de Tela 2023-05-26 às 14 03 55
After:
Captura de Tela 2023-05-26 às 14 02 09
(Pay attention to Link clothes).

Note that the use of lower precision operations is a performance optimization. So using higher precision operation may impact performance. I tried to apply that only to the patterns that are strictly needed here to minimize impact, and I didn't notice any impact on my testing, but it would be nice to get general testing to ensure that:
- All instances of this specific bug were fixed.
- There was no noticeable performance impact. Make sure to compare it to the proper builds (that is, if you want to compare it with buffer mirrors build (#4899), you need to merge buffer mirrors with that change. Otherwise, you should compare with master).

Closes #4949.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 5102;
+        private const uint CodeGenVersion = 5110;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
--- a/src/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
+++ b/src/Ryujinx.Graphics.Shader/CodeGen/Spirv/Instructions.cs
@@ -2245,7 +2245,7 @@ namespace Ryujinx.Graphics.Shader.CodeGen.Spirv
             {
                 var result = emitF(context.TypeFP64(), context.GetFP64(src1), context.GetFP64(src2));
 
-                if (!context.Config.GpuAccessor.QueryHostReducedPrecision())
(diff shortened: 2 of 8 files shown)


marysaka (approved):
lgtm

riperiperi (approved):
I think this is a reasonable compromise til MSL/spirv-cross is better about this. Maybe it will apply in more cases than needed with the selector just matching any added constant, but people are reporting it running about the same and it might cover weirder cases (offset + bias on a value that's potentially negative?).

gdkchan:
This also seems to fix Quaxly hair on Pokémon Scarlet/Violet.
Before (old screenshot from @riperiperi):
image
After:
image
Tested with #5080.

siegmund-heiss-ich:
before:
peri

after:
gdk

Tested with #4899

PR #5112, opened 2023-05-26
Fix incorrect vertex attribute format change

gdkchan:
This fixes an oversight from #5094. The `Binding` value on the vertex attribute struct is not an index into the vertex descriptors array, it must match the actual vertex buffer `Binding` value. So I have changed the code and added a function to look for the binding return the correct vertex buffer index.

This fixes a regression on macOS where a few games would not render correctly due to it changing the format of some attributes to the wrong format. It only affect games that have gaps on the vertex buffer bindings, which is something uncommon. So far the only game I tested that was affected is Gun Gun Pixies.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/FormatCapabilities.cs
+++ b/src/Ryujinx.Graphics.Vulkan/FormatCapabilities.cs
@@ -65,6 +65,13 @@ namespace Ryujinx.Graphics.Vulkan
             return (formatFeatureFlags & flags) == flags;
         }
 
+        public bool BufferFormatSupports(FormatFeatureFlags flags, VkFormat format)
+        {
+            _api.GetPhysicalDeviceFormatProperties(_physicalDevice, format, out var fp);
+
+            return (fp.BufferFeatures & flags) == flags;
+        }
+
         public bool OptimalFormatSupports(FormatFeatureFlags flags, GAL.Format format)
         {
             var formatFeatureFlags = _optimalTable[(int)format];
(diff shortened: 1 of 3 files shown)


marysaka (approved):

riperiperi (approved):
~It doing an O(n*m) loop on the attributes and bindings now isn't great, but would need some kind of weird lookup to avoid. I'm also not sure how expensive calling `vkGetPhysicalDeviceFormatProperties` each time the format is changed is. But it is definitely wrong right now.~

Since the cost is only on pipeline create, it's negligible in comparison to what the driver does, and shouldn't run as often as I was thinking it did.

gdkchan:
> It doing an O(n*m) loop on the attributes and bindings now isn't great, but would need some kind of weird lookup to avoid. I'm also not sure how expensive calling `vkGetPhysicalDeviceFormatProperties` each time the format is changed is. But it is definitely wrong right now.

Yea, that cost is only paid when new pipelines needs to be created. Given all the work the driver has to do for that, I would guess the cost of those sizes is not much in comparison. But yea, would be nice to improve this, it would not be hard to add a dictionary to cache support for each `VkFormat`, but using an array would be more difficult since `VkFormat` values are not always sequential.

BTW since MoltenVK does the same thing, I think it would be easy to fix it on MoltenVK and delete the code on the emulator, but it would depend on them accepting the changes. Also MoltenVK also has O(n*m) cost for looking up the vertex buffer since it does the same thing.

PR #5121, opened 2023-05-28
Fix problem with console not closing when exiting app

bule2015:
I ran into a problem with the console not exiting when exiting the app. #2863

I checked and found that I am stuck in an infinite loop in "Ryujinx.Graphics.GAL.Multithreading.ThreadedRenderer.New<>()".
This is due to the _consumerPtr not being changed in RenderLoop() when the app exits.
So I changed it to skip the code that falls into an infinite loop when the app exits.

### DIFF
--- a/src/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
+++ b/src/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
@@ -170,15 +170,15 @@ namespace Ryujinx.Graphics.GAL.Multithreading
 
         internal ref T New<T>() where T : struct
         {
-            if (_running && !_disposed)
+            while (_producerPtr == (Volatile.Read(ref _consumerPtr) + QueueCount - 1) % QueueCount)
             {
-                while (_producerPtr == (Volatile.Read(ref _consumerPtr) + QueueCount - 1) % QueueCount)
-                {
-                    // If incrementing the producer pointer would overflow, we need to wait.
-                    // _consumerPtr can only move forward, so there's no race to worry about here.
+                // If incrementing the producer pointer would overflow, we need to wait.
+                // _consumerPtr can only move forward, so there's no race to worry about here.
 
-                    Thread.Sleep(1);
-                }
+                Thread.Sleep(1);
+
+                // Execute the commands remaining in _commandQueue to avoid getting stuck in an infinite loop here when the app exits.
+                DisposeCommandQueueRun();
             }
 
             int taken = _producerPtr;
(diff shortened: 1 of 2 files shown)


src/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
```diff
@@ -170,12 +170,15 @@ private TableRef<T> Ref<T>(T reference)
 
         internal ref T New<T>() where T : struct
         {
-            while (_producerPtr == (Volatile.Read(ref _consumerPtr) + QueueCount - 1) % QueueCount)
+            if (_running && !_disposed)
             {
-                // If incrementing the producer pointer would overflow, we need to wait.
-                // _consumerPtr can only move forward, so there's no race to worry about here.
+                while (_producerPtr == (Volatile.Read(ref _consumerPtr) + QueueCount - 1) % QueueCount)
```
TSRBerry, line 175: Why not add the the two conditions to the while loop?

I couldn't reproduce the infinite loop, but from my tests `_running` will be false and `_disposed` will be true when the dispose commands get created, so adding them to the while loop should work as well.

I'm currently not sure if that could happen, but doing the check in the loop as well would also cancel the loop if `_running` or `_disposed` would change during execution.

TSRBerry (commented):
I don't think I'm qualified to approve this one, but I have a question that I added as a comment.

porcino:
I can finally close Tears of the Kingdom without freezing with this PR

src/Ryujinx.Graphics.GAL/Multithreading/ThreadedRenderer.cs
bule2015: Thanks for your comments.
I think it is a good idea to add a condition to the while loop as you mentioned.

In the first PR, I was able to avoid the infinite loop and achieve my objective.
However, commands queued from various Dispose() were exiting without being executed.
So I improved the PR and changed it to execute commands that remain in the queue.
We would appreciate it if you could check to see if there are any problems.

riperiperi (changes requested):
This changes the thread these commands are running on, which is invalid and will probably crash OpenGL.

The root cause of this issue is that the ThreadedRenderer is being disposed, but something else is trying to dispose a bunch of GPU resources after that. The ThreadedRenderer shouldn't be used after dispose, so either its dispose needs to be delayed until after those resources are destroyed, or the resources need to be destroyed sooner.

bule2015:
I also tried repositioning Renderer.Dispose(), but I got stuck in an infinite loop in New<>().
It seems necessary to execute the commands that remain in the queue.

bule2015:
I'm glad the problem has been resolved. #5144
I tried it and it terminated successfully for both GTK and Avalonia.

PR #5124, opened 2023-05-28
chore: Update Avalonia to 0.10.21

marysaka:
As title say.


TSRBerry (approved):
Works fine on Linux x64!

AcK77 (approved):

PR #5129, opened 2023-05-28
[Logger] Add print with stacktrace method

TSRBerry:
This small PR adds a convenience method `PrintStack()` to log the current stacktrace in addition to the regular message.

I used this while I was debugging and think it'd be useful to upstream this as well.


gdkchan:
Can't you use one of the regular log methods and pass `Environment.GetStackTrace()` as the message?

TSRBerry:
I could, yeah! That's why I called it a "convenience" method! :D

AcK77 (approved):

marysaka (approved):

PR #5133, opened 2023-05-28
Show/Hide UI Hotkey fix on Avalonia

### LINKED ISSUE #3277: Hide UI with F4
Currently you can show the UI with F4 but there's no shortcut to hide it again.

It would be really nice to be able to toggle the UI with the same key.

patrick-hovsepian:
Address ` Hide UI use F4 hotkey, but if you enable it, there is no way to get the UI back by pressing F4.` from #3662 

Fixes #3277


src/Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml
```diff
@@ -113,7 +113,9 @@
                     InputGesture="Escape"
                     IsEnabled="{Binding IsGameRunning}"
                     ToolTip.Tip="{locale:Locale StopEmulationTooltip}" />
-                <MenuItem Command="{ReflectionBinding SimulateWakeUpMessage}" Header="{locale:Locale MenuBarOptionsSimulateWakeUpMessage}" />
+                <MenuItem 
+					Command="{ReflectionBinding SimulateWakeUpMessage}"
```
gdkchan, line 117: This is using Tabs rather than spaces.

patrick-hovsepian: Ah, just reverted for now until I get my settings figured out.

AcK77 (approved):

TSRBerry (changes requested):
Could you add that for Gtk as well?

MutantAura:
> Could you add that for Gtk as well?

This behavior is fully intentional on GTK. And I'm not entirely sure how to resolve the issue between them here. 
The reason it doesn't simply toggle already, is because we had users who had pressed F4 by mistake, and as such lost access to the UI entirely, with no way to return it outside restarting the program. 

If we just had to deal with Ava, I'd say this change is fine, along with making the hotkey un-assigned by default perhaps, so people can't accidentally trigger it. On GTK, this obviously isn't possible because users wouldn't be able to map it. There is just a straight incompatibility between them.

patrick-hovsepian:
> along with making the hotkey un-assigned by default perhaps, so people can't accidentally trigger it

Can that be done as a follow up or would you want to require it for this PR?

patrick-hovsepian:
default to unset

MutantAura:
> default to unset

Again, this isn't a solution because GTK exists and there is a menu option that can hide the UI. If a user on either clicked this without binding a show hotkey, it would be a one way trip unless the app was restarted.

MutantAura:
To clarify as I'm not sure why this was closed, I'm fine with this being changed to a toggle (albeit with the inherant downsides this setting intrinsically will always have), but unmapping it by default cannot happen until GTK is gone.

patrick-hovsepian:
> but unmapping it by default cannot happen until GTK is gone. 

So this PR would just stick around until then? If that's the case, I'll re-open it but I closed it because I thought you wanted changes made there too and I didn't see the point if it was going to be deprecated.

patrick-hovsepian:
Been some back and forth on discord

MutantAura (approved):

TSRBerry (approved):
Sorry for the delay, lgtm!

PR #5140, opened 2023-05-29
Avalonia UI: Fix letter "x" being cut off in the About dialog & some styling fixes

yell0wsuit:
Implement some fixes in the About dialog for Avalonia UI:
- Fix the letter "x" in the Ryujinx text logo being cut off
- Make the pronunciation under the text logo center-aligned
- Make font size bigger
  - Make the About dialog bigger as a result

Master: 
Screenshot 2023-05-29 104409

PR:
Screenshot 2023-05-29 104349


src/Ryujinx.Ava/Assets/Styles/Styles.xaml
```diff
@@ -317,7 +317,7 @@
         <x:Double x:Key="ControlContentThemeFontSize">13</x:Double>
         <x:Double x:Key="MenuItemHeight">26</x:Double>
         <x:Double x:Key="TabItemMinHeight">28</x:Double>
-        <x:Double x:Key="ContentDialogMaxWidth">600</x:Double>
+        <x:Double x:Key="ContentDialogMaxWidth">800</x:Double>
```
AcK77, line 320: I think we already discussed about that somewhere (another PR or Discord) but we don’t want the About window bigger.

src/Ryujinx.Ava/UI/Windows/AboutWindow.axaml
```diff
@@ -8,11 +8,11 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:ui="clr-namespace:FluentAvalonia.UI.Controls;assembly=FluentAvalonia"
     xmlns:viewModel="clr-namespace:Ryujinx.Ava.UI.ViewModels"
-    Width="550"
-    Height="260"
-    Margin="0,-12,0,0"
-    d:DesignHeight="260"
-    d:DesignWidth="550"
+    Width="680"
```
AcK77, line 11: Same comment as the Style file

src/Ryujinx.Ava/UI/Windows/AboutWindow.axaml
```diff
@@ -79,7 +86,6 @@
                     Click="Button_OnClick"
                     Tag="https://github.com/Ryujinx/Ryujinx/wiki/Changelog#ryujinx-changelog">
                     <TextBlock
-                        FontSize="10"
```
AcK77, line 82: The small text looks better than the same size everywhere. (Same comment for other FontSize removed fields)

AcK77 (changes requested):
It’s nice to see some things fixed, but I don’t think everything have to be changed.

gdkchan:
I do think it looks better like that since it's so tiny right now.

src/Ryujinx.Ava/Assets/Styles/Styles.xaml
yell0wsuit: Hmm. I found that in GTK UI the font size is bigger which aids in legibility. On the other hand, in Ava UI the font size is a bit small. But if you guys don’t want that, I will close this PR (and make a separate one that fix the letter “x”).

src/Ryujinx.Ava/UI/Windows/AboutWindow.axaml
yell0wsuit: Seem reasonable. I just noticed that smaller size for buttons is a bit better.

src/Ryujinx.Ava/Assets/Styles/Styles.xaml
yell0wsuit: The issue is that, if the window size is not increased, it will cause issue to the layout. And according to gdkchan the font size is too small, so the window size has to be larger to accommodate that. But of course if you don’t want that then I can revert the changes.

yell0wsuit:
Close this in favor of #5176

PR #5145, opened 2023-05-29
Logger: Add warning that WinTerminal doesn't support hiding console window

yell0wsuit:
Windows Terminal doesn't support hiding console window, and unfortunately this is by design. This small PR will write a warning in the log to inform users about that.

I haven't tested this on Windows 10 or Win 11 before 22H2 update.

### DIFF
--- a/src/Ryujinx.Ui.Common/Helper/ConsoleHelper.cs
+++ b/src/Ryujinx.Ui.Common/Helper/ConsoleHelper.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
+using System.Diagnostics;
 
 namespace Ryujinx.Ui.Common.Helper
 {
@@ -13,7 +14,12 @@ namespace Ryujinx.Ui.Common.Helper
         {
             if (OperatingSystem.IsWindows())
             {
-                SetConsoleWindowStateWindows(show);
+                if (Process.GetProcessesByName("WindowsTerminal").Length < 0) {
+                    SetConsoleWindowStateWindows(show);
+                }
+                else {
+                    Logger.Warning?.Print(LogClass.Application, "Windows Terminal doesn't support hiding console window");
+                }
             }
             else if (show == false)
             {


src/Ryujinx.Ui.Common/Helper/ConsoleHelper.cs
```diff
@@ -13,7 +14,12 @@ public static void SetConsoleWindowState(bool show)
         {
             if (OperatingSystem.IsWindows())
             {
-                SetConsoleWindowStateWindows(show);
+                if (Process.GetProcessesByName("WindowsTerminal").Length < 0) {
```
sampletext32, line 17: this check will never succeed, because GetProcessesByName returns an array, which length is never below zero.
Also braces formatting is incoherent

sampletext32: and btw, user can open windows terminal with some other process and it won't be ryujinx. this check will produce some unwanted behavior, so i doubt it is correct at all, to do it this way

yell0wsuit: Oh yeah I forgot, I should have checked for null instead because the result is an array like you said. Also I will come up with a check to see if both the Windows Terminal and the Ryujinx process are running.

yell0wsuit: Oh yeah I forgot, I should have checked for null instead because the result is an array like you said. Also I will come up with a check to see if both the Windows Terminal and the Ryujinx process are running.

src/Ryujinx.Ui.Common/Helper/ConsoleHelper.cs
```diff
@@ -19,6 +20,17 @@ public static void SetConsoleWindowState(bool show)
...
                 Logger.Warning?.Print(LogClass.Application, "OS doesn't support hiding console window");
             }
+            if (Process.GetCurrentProcess().ProcessName.Contains("Ryujinx"))
+            {
+                Process[] WinTermprocesses = Process.GetProcessesByName("WindowsTerminal");
+                foreach (Process process in WinTermprocesses)
+                {
+                    if (process.MainWindowTitle.Contains("Ryujinx Console"))
+                    {
+                        Logger.Warning?.Print(LogClass.Application, "Windows Terminal doesn't support hiding console window");
```
sampletext32, line 30: I think that a break or return is a nice addition :)

PR #5151, opened 2023-05-29
Add nuget package caching for workflows

TSRBerry:
A few hours ago v3.2.0 of [actions/setup-dotnet](https://github.com/actions/setup-dotnet/releases/tag/v3.2.0) was released which allows us to cache NuGet packages based on a lock file.

So in order to take advantage of that, the use of a solution-wide [lock file](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#locking-dependencies) was enabled and the new caching options were added to our workflows!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,6 +85,9 @@ jobs:
           sed -r --in-place 's/\%\%RYUJINX_TARGET_RELEASE_CHANNEL_REPO\%\%/${{ env.RYUJINX_TARGET_RELEASE_CHANNEL_REPO }}/g;' src/Ryujinx.Common/ReleaseInformation.cs
         shell: bash
 
+      - name: Restore
+        run: dotnet restore --locked-mode
+
       - name: Create output dir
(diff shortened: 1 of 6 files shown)


marysaka (approved):
I hope it will not cause more CI time spend up just decompressing dependencies but look good to me 👍

.github/workflows/release.yml
```diff
@@ -66,6 +66,8 @@ jobs:
       - uses: actions/setup-dotnet@v3
         with:
           global-json-file: global.json
+          cache: true
```
patrick-hovsepian, line 69: should it be a param/variable just in case?

marysaka: I do not believe it matter here. if any issue arise we will do a patch for it.

TSRBerry:
Although I thought a solution-wide file worked correctly, it seems the lock files are currently project specific.

Will mark this as draft until I figured out another way or setup-dotnet supports specifying multiple lock files.

Issue #4787, opened 2023-05-03
[Feature Request] Show profile list in the "home" for quick access


Gotxiko:
### Overview

Show profile list in the "Home" for quick access.

### Smaller details

**See mockup for example**
List on the left side (for example), a vertical list of the different profiles.
Show with the blue color which profile is selected.
Show a (+) button to create new profiles. (Open the create profile window)
![Ryujinx 1 1 747 03_05_2023 16_45_37](https://user-images.githubusercontent.com/416493/235954152-c6b74483-e3ba-44e4-8e51-81f819cbf32f.png)

### Nature of request

Have seen that many users don't know that the profile feature exists, myself included, and that they can use different profiles to have different saves per game.

### Why would this feature be useful?

This would allow making use of different profiles easier and have quick access to change profiles before launching a game.

PR #5153, opened 2023-05-29
Show Active User Profile

### LINKED ISSUE #5104: Show Which User Profile Is Selected
### Overview

It would be nice if we could somehow see in the tool bar or at the bottom of the emulator, which user profile is active

### Smaller details

I have no idea how to program this but I don't think it would be too complicated to implement

### Nature of request

I noticed this while booting a game after my girlfriend played on her profile, it took me 4-5 minutes to load the game with all the shaders, then I noticed I was on the wrong user profile so I didn't have my save, I had to shut the game, load my profile and reload the game for another 4-5 minutes, it's time consuming when this happens

For people like me who forgets to look which user profile is active before booting any games

patrick-hovsepian:
Adds the icon of the active profile to the Ava library
image

The existing size of the bar has been preserved. The icon is clickable and will open the existing User Profiles Manager. When the profile is changed, the icon will be updated on the dialog's dismissal.

Fixes #5104


src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
```diff
@@ -78,98 +98,98 @@
...
+			</ui:DropDownButton.Flyout>
+		</ui:DropDownButton>
+		<TextBlock
             Margin="10,0"
             HorizontalAlignment="Right"
             VerticalAlignment="Center"
             DockPanel.Dock="Right"
             Text="{locale:Locale CommonSort}" />
-    </DockPanel>
+	</DockPanel>
```
tarrouye, line 194: Formatting seems a bit fucked ?

src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
```diff
@@ -11,46 +11,65 @@
...
+		<helpers:BitmapArrayValueConverter x:Key="ByteImage" />
+	</UserControl.Resources>
+	<Design.DataContext>
+		<viewModels:MainWindowViewModel />
+	</Design.DataContext>
+	<DockPanel
         Margin="0,0,0,5"
         HorizontalAlignment="Stretch">
-        <Button
+		<StackPanel
```
tarrouye, line 23: This seems like one layer too many. 

Couldn't you just put all the needed properties in one layer ? E.g. 

```
<Border
    Name="UserIconBorder"
    Margin="5,2,0,2"
    VerticalAlignment="Center"
    Width="30"
    CornerRadius="15"
    ClipToBounds="True">
    <Image
        Name="UserProfileIcon"
        VerticalAlignment="Center"
        HorizontalAlignment="Stretch"
        PointerPressed="UserProfileIcon_Click"
        Source="{Binding UserProfileIcon, Converter={StaticResource ByteImage}}" />
</Border>
```

piplup55:
it would be nice if when you hovered over the profile icon it shows the name of the profile

yell0wsuit:
While you are at it, I think adding active profile at the status bar (at the bottom) is a nice addition too. Bonus point if it has an icon.

MutantAura:
I would be against that. The status bar is already cluttered as is. Adding even more static information there seems overkill. 

Why does the user need to be able to see which profile is selected in-game? What is the use for that. In the game list it provides a quick way to tell if you need to switch before booting, in-game is useless as you'd need to close to change anyway, at which point you'd check the game list.

The ideal scenario here is basically mimicking the switch behaviour. Show the icon somewhere on the game list and then if there is more than 1 profile ask the user on game boot which to use.

tarrouye:
On the one hand, it feels a little out of place here. Everything surrounding it is completely unrelated (just view settings)
On the other hand, it's useful information and I'm not sure where else it could be put. 
To @MutantAura's point, if the game would ask on boot which profile to use, this badge would actually be useless IMO. It's useful right now specifically because Ryujinx _doesn't_ ask which profile to use, and just uses the 'active' one.

src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
patrick-hovsepian: yeah, i think its tabs vs spaces. I ran the formatter per the editor config but it looked like it janked it up more

patrick-hovsepian:
> it would be nice if when you hovered over the profile icon it shows the name of the profile

I'm going to omit that from this PR. If this gets merged, that can be follow on work. 

> The ideal scenario here is basically mimicking the switch behaviour. Show the icon somewhere on the game list and then if there is more than 1 profile ask the user on game boot which to use.

IMO I hate that as a user experience but can understand keeping it as close to the switch experience as possible. I also think that if that's the desired behavior it can go in a follow up PR.

> While you are at it, I think adding active profile at the status bar (at the bottom) is a nice addition too. Bonus point if it has an icon.

I agree with the clutter. If this gets in, the show name on hover might be enough?

> if the Ryujinx would ask on application launch which profile to use, this badge would actually be useless IMO

Doesn't the switch do this in it's library? I think this would be useful regardless of prompting the user on app start which feels intrusive but that's my personal feeling on it.

src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
patrick-hovsepian: That might work but i had issues with the image causing the buttons to grow which was undesirable (per discord). I can try taking a level out but also left it in the panel in case other user info goes in there and/or it needs to get moved entirely.

tarrouye:
> IMO I hate that as a user experience but can understand keeping it as close to the switch experience as possible.
I agree, it doesn't fit well in the desktop/emu context IMO. If it were introduced, it'd better be under a toggle to use the active one as it does now 

> Doesn't the switch do this in it's library? I think this would be useful regardless of prompting the user on app start which feels intrusive but that's my personal feeling on it.

The switch lays out icons for ALL the profiles on the device. Tapping them opens the settings. and afaik there is no indication of 'active' or 'selected' because every application launched will prompt you to pick a profile.

src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
patrick-hovsepian: great, thanks

AcK77 (approved):

marysaka (approved):
Seems a nice change to me, thanks 👍

src/Ryujinx.Ava/UI/Views/Main/MainViewControls.axaml
patrick-hovsepian: going to leave as-is since it's approved.

MutantAura:
While I like the functionality of this change, the single nit-pick I have is that there is no visual way to know the icon is actually clickable/responsive. Elements like this should either react by changing the mouse cursor or have a "highlight" effect imo. At the moment while it does everything it needs to, it feels very weird to actually use.

gdkchan:
I don't like this change for a few reasons:
- Currently everything on the top menu bar is related to the game list (the list/grid switch, the names toggle and  the search box, they are all to interact with the game list/grid). This would be the first unrelated element there, it feels a bit out of place.
- The icon is tiny, making hard to know what it even is in the first place. And making it bigger would make the bar taller just to fit the icon, which is even worse.

So to me that is not the right place to put it. To suggest a better solution, it would be nice to know which problem you're trying to solve in the first place. It is already possible to change the user profile by going to Options > Manage User Profile. Which use case requires the profile to be changed so frequently that this is not good enough? Maybe users would be better served with a dialog which asks which profile the user wants before starting the game.

tarrouye:
@gdkchan 

> Which use case requires the profile to be changed so frequently that this is not good enough?

I think it is rather about knowing at a glance which profile will be used when you launch a game, than actually changing the profile frequently. At least for me as a user, that is the current problem that it solves. 

It doesn't make it any faster for me to swap profiles, of course, but it does let me easily see which profile I have active when I go to boot a game, without having to go into that whole menu. 

> Maybe users would be better served with a dialog which asks which profile the user wants before starting the game.

I don't mind it that much since the Switch works this way, but I could see it getting annoying if someone has a second profile for whatever reason but 90% of the time they use the same profile. Then, I'd probably want some toggle for the current behavior of launching with the active profile. And then we are right back at the start, where there is a need for some kind of way to see which profile is active before launching the game.

patrick-hovsepian:
The original intent was to display the active profile so user's know at a glance. There was an ask in discord to make it clickable and initiate the profile management flow. I agree, it's odd without any visual feedback. I tried to make it highlight on mouseover but couldn't get it to work with Avalonia. I can either remove the clickable portion of this PR or just outright close it if you feel strongly about it not being in the bar.

MutantAura:
Would a suitable compromise just be to add a setting for this? We have plenty of space for more UI options. 
image

I don't really agree that everything in the bar needs to relate the game-list either. That isn't widespread UX design language (see even something as immediate as Chrome using it's top bar for not only URL options but also extensions and whatever user account you're logged in as) and I think it's the most intuitive place for it to go. 

Either way a settings toggle, likely off by default, seems the best thing to do with any subjective UI features going forward.

patrick-hovsepian:
Agreed. If the team can align on next steps, I'll move forward with those suggestions.

patrick-hovsepian:
Relates to #4787

AcK77:
image

What about something like that? I've made a quick mockup.
System informations are on top bar, with the list / grid button at the middle, if we want we can add more informations later, more games layout buttons, etc...
Game list/grid controls are grouped on the bottom bar so you can refresh the list, sort it, set the icon size, etc...

I personally thinks it's better and more clear like that and everyone could be happy :)

sakuka47:
It's not bad, it's still airy, and the quick access icon is very practical.😁

tarrouye:
Anything that moves the search bar out of the top right is a :-1: IMO

I personally prefer what is implemented on this branch to either of those mockups.

piplup55:
i feel like system version should be in the bottom right still and should be able to be hidden in the settings,
i also feel like this pr has turned into the "game window rework", i'm not saying that's a bad thing just odd

gustavomeli (approved):

siegmund-heiss-ich:
Why don't we just merge this PR?
It looks fine.
I don't know why you would change something design wise.
It blends perfect with the design of MacOS and Windows. The search bar is where it should be (upper right corner). 
It's the same for Windows Explorer and MacOS Finder. 🤷‍♂️

horstdog:
Just started using Ryujinx and surprised to not see active user profile on the main page as described on this page. Any update to this?

PR #5154, opened 2023-05-30
Add Context Menu Option to Run Application

patrick-hovsepian:
image

Need to add locales but opening up in case there's wording change requests


src/Ryujinx.Ava/Assets/Locales/en_US.json
```diff
@@ -282,6 +282,7 @@
   "ControllerSettingsSaveProfileToolTip": "Save Profile",
   "MenuBarFileToolsTakeScreenshot": "Take Screenshot",
   "MenuBarFileToolsHideUi": "Hide UI",
+  "GameListContextMenuRunApplication": "Run Application",
```
yell0wsuit, line 292: You should add newly made string at the bottom to make it easier to localize

patrick-hovsepian: it's right above the prop for the toggle favorites which is it's neighbor item in the menu so i figured that was a better logical grouping. I'd add the others in the same place

AcK77: Order doesn’t matter since translations are done using crowdin

AcK77 (approved):
Lgtm

src/Ryujinx.Ava/Assets/Locales/en_US.json
yell0wsuit: Oh, didn't know about that. I thought adding at the bottom would make it easier to compare diff.

PR #5044, opened 2023-05-22
Remove barrier on Intel if control flow is potentially divergent

gdkchan:
Apparently, having a `barrier` after return on a shader causes Intel GPUs to hang. As a work around, this change removes the barrier from the shader in those cases.

Allows The Legend of Zelda: Tears of the Kingdom to run on Intel GPUs.
image
However, from my tests it still crashes a few minutes later with DeviceLost error (apparently for a different reason?), so it would be nice to get testing on other Intel GPUs to see if the issue affects all of them.

PR #5159, opened 2023-05-30
Implement soft float64 conversion on shaders when host has no support

gdkchan:
Some GPUs don't support 64-bit floating point (double) operations. This change a soft-float implementation for doing conversion between float and double. When double shader instructions are used and the host does not support it, a separate pass turn them into float operations, and inserts float <-> double conversions as required. There is room for improvement in the code generation, I plan to improve it in the future.

Most notably, this benefits macOS since MSL has no double support apparently. Double precision operations on shaders are not common, so the amount of games that benefit small. The most notable game is Rune Factory 4 Special.

Before:
Captura de Tela 2023-05-30 às 00 24 37
After:
Captura de Tela 2023-05-30 às 00 21 55
Contributes to #4062 (it is missing from the list).
Should also help some integrated Intel GPUs that don't support FP64.


src/Ryujinx.Graphics.Shader/Translation/EmitterContextInsts.cs
```diff
@@ -655,7 +655,6 @@ public static Operand PackHalf2x16(this EmitterContext context, Operand a, Opera
 
         public static void Return(this EmitterContext context)
         {
-            context.PrepareForReturn();
```
marysaka, line 658: Any reasons that was delete here?

marysaka (approved):
Seems fine to me apart from my comment.

Not quite sure why PrepareForReturn was moved to before calling Return.

riperiperi (approved):
Nice work, seems like a reasonable compromise. I don't think I have any games where FP64 shaders are used in a visible way to test.

mamonup:
Hope this gets merged soon. This fix together with #5044 fixes all the DEVICE_LOSS crashes on Intel Iris XE graphics in Tears of the Kingom. It's exhausting going through each file change and manually adding them to the src folder then building ryujinx. Tested on 1.1.869

PR #5144, opened 2023-05-29
GPU: Dispose Renderer after running deferred actions

riperiperi:
Deferred actions from disposing physical memory instances always dispose the resources in their caches. The renderer can't be disposed before these resources get disposed, otherwise the dispose actions will not actually run (leaking memory), and the ThreadedRenderer may get stuck trying to enqueue too many commands when there is nothing consuming them (user must close console manually).

This should fix most instances of the emulator freezing on close.

Should fix #2863, should obsolete #5121. Needs a bunch of testing obviously.

PR #5163, opened 2023-05-30
UI: Warn if user attempts to open more than 1 Ryujinx instances

yell0wsuit:
Users are not supposed to open more than 1 instance of Ryujinx at once. In addition, it also serves as a useful information in case there's *that* one Ryujinx instance stuck in the background which takes up GBs of precious RAM memory \*shrug\*. RPCS3 emulator has it, so why not?

I also added in the console window in case the GUI sometimes doesn't show up for some reasons.

Screenshots:

Ava UI:

image

GTK UI:

image

### DIFF
--- a/src/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/src/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -639,5 +639,6 @@
   "NetworkInterfaceDefault": "Default",
   "PackagingShaders": "Packaging Shaders",
   "AboutChangelogButton": "View Changelog on GitHub",
-  "AboutChangelogButtonTooltipMessage": "Click to open the changelog for this version in your default browser."
+  "AboutChangelogButtonTooltipMessage": "Click to open the changelog for this version in your default browser.",
+  "DialogMessageOnlyOneInstance": "Only 1 instance of Ryujinx is allowed. Please close other Ryujinx instances if necessary."
 }
--- a/src/Ryujinx.Ava/Program.cs
+++ b/src/Ryujinx.Ava/Program.cs
@@ -17,6 +17,7 @@ using System;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
+using System.Diagnostics;
 
 namespace Ryujinx.Ava
 {
@@ -125,6 +126,17 @@ namespace Ryujinx.Ava
             if (CommandLineState.LaunchPathArg != null)
             {
                 MainWindow.DeferLoadApplication(CommandLineState.LaunchPathArg, CommandLineState.StartFullscreenArg);
+            }
+
+            // Check for multiple instances of Ryujinx
+            Process currentProcess = Process.GetCurrentProcess();
+            foreach (Process process in Process.GetProcessesByName(currentProcess.ProcessName))
+            {
+                if (process.Id != currentProcess.Id)
+                {
+                   MainWindow.ShowMultipleInstancesDialog = true;
+                   Logger.Warning?.PrintMsg(LogClass.Application, "Only 1 instance of Ryujinx is allowed, please close other Ryujinx instances if necessary"); // Add warning in the console just in case the GUI sometimes doesn't show up.
+                }
             }
         }
 
--- a/src/Ryujinx.Ava/UI/Windows/MainWindow.axaml.cs
+++ b/src/Ryujinx.Ava/UI/Windows/MainWindow.axaml.cs
@@ -25,6 +25,7 @@ using System.ComponentModel;
 using System.IO;
 using System.Threading.Tasks;
 using InputManager = Ryujinx.Input.HLE.InputManager;
+using System.Diagnostics;
 
 namespace Ryujinx.Ava.UI.Windows
 {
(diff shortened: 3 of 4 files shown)


d3xMachina:
Wouldn't a global mutex be more appropriate to force a single instance ? Something like : 
```
using (Mutex mutex = new Mutex(false, "Global\\" + appGuid)) {
  if (!mutex.WaitOne(5000, false)) {
      // Ryujinx already runnning
      return;
  }
  // Run
}
```

d3xMachina:
With the getprocessbyname approach you have the following problems :
-  You can have race conditions. Two instances of the application could launch at nearly the same time, see each other, and both shut down. 
- If a different program have the same name and is running then Ryujinx will shutdown

Also, in case you have multiple users running on your computer you can remove the "Global\\" prefix in my example (or replace it with "Local\\") if you want to allow each user to run Ryujinx once. Otherwise it will be one instance for all users.

yell0wsuit:
I was able to reproduce the two problems you've pointed out. The first one is pretty easy, by pressing Enter on the exe file twice quickly. The second one, by creating two exes in separate folders, then open each one (or creating a bogus app called Ryujinx).
Marked this PR as draft until I figure out how to use the mutex class.

gdkchan:
I'm not sure what's the point of doing this, it's a completely artificial limitation since there's nothing stopping someone from running two instances of the emulator. Some people does this to play games in multiplayer on the same machine using LDN, connecting the two instances to each other. So that can be used to play multiplayer on games that don't support multiplayer on the same console (or avoiding the split screen on the ones that do). I don't think we want to make that scenario not work anymore.

jvruyu:
> I'm not sure what's the point of doing this, it's a completely artificial limitation since there's nothing stopping someone from running two instances of the emulator. Some people does this to play games in multiplayer on the same machine using LDN, connecting the two instances to each other. So that can be used to play multiplayer on games that don't support multiplayer on the same console (or avoiding the split screen on the ones that do). I don't think we want to make that scenario not work anymore.

I agree, I think a warning may be fine, in order not to accidentally have two instances of Ryujinx open at the same, maybe even running a game, which would kill performance on not-high end machines, but there's no point on straight up banning it. A "do not remember again" field would be fine, also.

yell0wsuit:
> play games in multiplayer on the same machine using LDN

Oh I forgot this scenario exists, but then again…

> a warning may be fine, in order not to accidentally have two instances of Ryujinx open at the same, maybe even running a game, which would kill performance on not-high end machines

is true too. Some people may have limited RAM memory, so a warning would be fine. I will change this into a warning instead of forcing users to kill the current instance in case they want to do something else, like gdk said.

jduncanator:
Yea, I agree with gdkchan on this one. Users can have as many instances of Ryujinx running as they like, we don't need to be putting in artificial limits, even if it is just a warning.

I'm not really sure what this PR achieves?

gdkchan:
> > play games in multiplayer on the same machine using LDN
> 
> Oh I forgot this scenario exists, but then again…
> 
> > a warning may be fine, in order not to accidentally have two instances of Ryujinx open at the same, maybe even running a game, which would kill performance on not-high end machines
> 
> is true too. Some people may have limited RAM memory, so a warning would be fine. I will change this into a warning instead of forcing users to kill the current instance in case they want to do something else, like gdk said.

I'm still not sure what the warning would add. It sounds like an annoyance for those that do use multiple emulator instances for the case I described before. I know there is one case where the emulator freezes instead of exiting right now, but that is being fixed on #5144. There may be other bugs that also cause the emulator to freeze when closing in addition to that one, but in those cases the window would show as "not responding" and would be pretty clear that the emulator is still open. I'm not aware of any case where the window closes but the emulator process stays alive. Also I can't recall of any report where a user ran out of memory due to having other emulator instances open. I don't want to sound negative and appreciate your contribution, but it looks like you're trying to solve a problem that doesn't exist here.

yell0wsuit:
Check above comments. I mean, for low RAM memory, running many instances can cause issue tho. And btw, like my original PR description, sometimes the GUI crashed but it still stuck in the background. A warning IMO is enough, maybe write to log instead of showing dialog window, no?

yell0wsuit:
I appreciate your input though, don't worry. If you find this change controversial, I will change this to log level warning instead (but let's not forget #3299 exists).

> the window closes but the emulator process stays alive

I do think this happens usually on lower end machines, specifically those using 4GB of RAM. But most of the time killing the "dead" process or restarting the PC works so not many people report this.

> it looks like you're trying to solve a problem that doesn't exist here.

I think that it may be helpful for the following scenario:
- Low-end users (with 4GB of RAM) want to play popular big N games.
- The emulator crashes, but its process sometimes stuck in the Task manager.
- But they didn't know this and open another Ryujinx instance.
- They feel the game(s) running slower than "usual".
- Upon checking Task Manager, they find that another Ryujinx instance stuck in the background and takes a whopping half of the RAM memory.

So my idea of this PR is to warn user about that, so they notice the problem sooner instead of experiencing the 4th scenario above.

> where a user ran out of memory due to having other emulator instances open

I think that it doesn't crash the whole OS, it just the OS will use pagefile as fallback. And as you know, low-end machines use HDD, and of course its performance is... atrocious.

Foul-Tarnished:
Isn't the issue the fact that Ryujinx can't open files when they are already in-use ?
Is this about some ryu files, or just the game save files ?


Maybe just make Ryu load the _next_ profile if another instance is opened

AcK77:
Limited instances of Ryujinx make no sense at all, if there is an issue because you opened too much instances at the same time, then it's better to fix the issue instead of limited things.
I'm sorry but this kind of solution will never be merged, so I close it as invalid.
Thanks for your contribution anyway!

PR #5167, opened 2023-05-30
macOS Headless Fixes

IsaacMarovitz:
- Make `use-hypervisor` nullable
   - There is a long-standing bug in CommandLineParser that makes it impossible to disable a bool argument if it is set to true by default and is not nullable
- Properly sign headless builds so they can use Hypervisor
- Bundle MoltenVK in headless builds when building for macOS
- Force Vulkan on macOS

Other Potential Fixes:
- CommandLineParser can't seem to read macOS path strings with escaped spaces in them correctly

### DIFF
--- a/src/Ryujinx.Headless.SDL2/Options.cs
+++ b/src/Ryujinx.Headless.SDL2/Options.cs
@@ -129,8 +129,8 @@ namespace Ryujinx.Headless.SDL2
         [Option("audio-volume", Required = false, Default = 1.0f, HelpText ="The audio level (0 to 1).")]
         public float AudioVolume { get; set; }
 
-        [Option("use-hypervisor", Required = false, Default = false, HelpText = "Uses Hypervisor over JIT if available.")]
-        public bool UseHypervisor { get; set; }
+        [Option("use-hypervisor", Required = false, Default = true, HelpText = "Uses Hypervisor over JIT if available.")]
+        public bool? UseHypervisor { get; set; }
 
         [Option("lan-interface-id", Required = false, Default = "0", HelpText = "GUID for the network interface used by LAN.")]
         public string MultiplayerLanInterfaceId { get; set; }
--- a/src/Ryujinx.Headless.SDL2/Options.cs
+++ b/src/Ryujinx.Headless.SDL2/Options.cs
@@ -129,7 +129,7 @@ namespace Ryujinx.Headless.SDL2
         [Option("audio-volume", Required = false, Default = 1.0f, HelpText ="The audio level (0 to 1).")]
         public float AudioVolume { get; set; }
 
-        [Option("use-hypervisor", Required = false, Default = true, HelpText = "Uses Hypervisor over JIT if available.")]
+        [Option("use-hypervisor", Required = false, Default = false, HelpText = "Uses Hypervisor over JIT if available.")]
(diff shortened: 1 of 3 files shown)


src/Ryujinx.Headless.SDL2/Options.cs
```diff
@@ -129,7 +129,7 @@ public class Options
         [Option("audio-volume", Required = false, Default = 1.0f, HelpText ="The audio level (0 to 1).")]
         public float AudioVolume { get; set; }
 
-        [Option("use-hypervisor", Required = false, Default = true, HelpText = "Uses Hypervisor over JIT if available.")]
+        [Option("use-hypervisor", Required = false, Default = false, HelpText = "Uses Hypervisor over JIT if available.")]
         public bool UseHypervisor { get; set; }
```
TSRBerry, line 133: You can deal with this by using a nullable bool here, instead of setting the default to false.

src/Ryujinx.Headless.SDL2/Program.cs
```diff
@@ -339,6 +339,12 @@ static void Load(Options option)
 
             GraphicsConfig.EnableShaderCache = true;
 
+            if (OperatingSystem.IsMacOS())
+            {
+                option.GraphicsBackend = GraphicsBackend.Vulkan;
+                Logger.Warning?.Print(LogClass.Application, "OpenGL is not supported on macOS, switching to Vulkan!");
+            }
+
```
TSRBerry, line 347: Only print this warning if the user attempted to launch it with the OpenGL backend.

TSRBerry (changes requested):
A few comments, but lgtm otherwise!

src/Ryujinx.Headless.SDL2/Ryujinx.Headless.SDL2.csproj
```diff
@@ -29,6 +34,7 @@
 
   <ItemGroup>
     <PackageReference Include="CommandLineParser" />
+    <PackageReference Include="Ryujinx.Graphics.Vulkan.Dependencies.MoltenVK" Condition="'$(RuntimeIdentifier)' != 'linux-x64' AND '$(RuntimeIdentifier)' != 'win10-x64'" />
```
TSRBerry, line 37:
```suggestion
    <PackageReference Include="Ryujinx.Graphics.Vulkan.Dependencies.MoltenVK" Condition="'$(RuntimeIdentifier)' == 'osx-x64' OR '$(RuntimeIdentifier)' == 'osx-arm64' OR $([MSBuild]::IsOSPlatform('OSX'))" />
```
How about something like this? Maybe this could also help with #5049? (cc @marysaka)

TSRBerry (commented):
Thank you! Just thought of something, so I added another comment.

src/Ryujinx.Headless.SDL2/Ryujinx.Headless.SDL2.csproj
marysaka: I would prefer to keep the solution the same as we have on other projects for now @TSRBerry.
I think we should unify this kind of checks in some helper file to avoid duplicating it everywhere.

marysaka (approved):
lgtm

TSRBerry (approved):
Lgtm, thanks!

PR #5169, opened 2023-05-30
Dont Error on Invalid Enum Values

IsaacMarovitz:
Stops Ryujinx from throwing out the whole config file if one enum has an invalid value (very annoying if you're working on a PR to add a new backend lol)

### DIFF
--- a/src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
+++ b/src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
@@ -23,7 +23,14 @@ namespace Ryujinx.Common.Utilities
                 return default;
             }
 
-            return Enum.Parse<TEnum>(enumValue);
+            try
+            {
+                return Enum.Parse<TEnum>(enumValue);
+            }
+            catch
+            {
+                return default;
+            }
         }
 
         public override void Write(Utf8JsonWriter writer, TEnum value, JsonSerializerOptions options)
--- a/src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
+++ b/src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
@@ -1,4 +1,5 @@
 ﻿#nullable enable
+using Ryujinx.Common.Logging;
 using System;
 using System.Text.Json;
 using System.Text.Json.Serialization;
@@ -24,6 +25,7 @@ namespace Ryujinx.Common.Utilities
                 return value;
             }
 
+            Logger.Warning?.Print(LogClass.Configuration, $"Failed to parse enum value \"{enumValue}\" for {typeof(TEnum)}, using default \"{default(TEnum)}\"");
             return default;
         }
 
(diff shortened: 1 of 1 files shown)


src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
```diff
@@ -18,17 +18,18 @@ public sealed class TypedStringEnumConverter<TEnum> : JsonConverter<TEnum> where
         public override TEnum Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
             var enumValue = reader.GetString();
-            if (string.IsNullOrEmpty(enumValue))
+
+            if (Enum.TryParse(enumValue, out TEnum value))
             {
-                return default;
+                return value;
             }
```
DuckSoft, line 26: Would it be advisable to log a warning when the enum value is unknown to Ryujinx, like `Logger.Warning?.Print(LogClass.Configuration, "Failed to parse enum value xxx for yyy, using default zzz")`?

marysaka: I like this idea, @IsaacMarovitz can you add this?

src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
```diff
@@ -18,17 +19,19 @@ public sealed class TypedStringEnumConverter<TEnum> : JsonConverter<TEnum> where
...
 
-            return Enum.Parse<TEnum>(enumValue);
```
myyk, line 26: This is maybe more a matter of style, but this code looked easier to read to me with the error case in the `if-statement` and the happy path at the end. (this also sets the pattern for future exception statements before the happy path at the bottom.

src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
```diff
@@ -18,17 +19,19 @@ public sealed class TypedStringEnumConverter<TEnum> : JsonConverter<TEnum> where
...
+            Logger.Warning?.Print(LogClass.Configuration, $"Failed to parse enum value \"{enumValue}\" for {typeof(TEnum)}, using default \"{default(TEnum)}\"");
```
patrick-hovsepian, line 28: super duper nits: just re-use `value` rather than `default(TEnum)`, same for return. also the quotes aren't really worth it IMO

or flip it to only have 1 return
```
if (!Enum.TryPars(...))
{Log}

return value;

src/Ryujinx.Common/Utilities/TypedStringEnumConverter.cs
patrick-hovsepian: exceptions are expensive and should be avoided.

TSRBerry (approved):
Lgtm!

---

Related to: #4707, since that also helps to resolve some config issues.

PR #5171, opened 2023-05-30
[Ava] Create a Backup of a Single Application

patrick-hovsepian:
Adds an context menu option to create  backup for a single application, **for the currently active profile**.
image

There can potentially be several save types for a given application. There's always one type that will exist however the `SaveDataType.Account` that's the only "required" sava data needed for this operation to be considered a success. 

Once completed, each save type will be zipped up and placed in a directory for the current day. At present, there is only 1 back for a given day. The last backup wins.

image

There will be a zip for each type of save data encountered but not one for all. There will always be the Account_backup entry though.  They're stored in the following path relative to Ryujinx `/backup/user/[userid]/[titleId]/[date]/[saveType]_backup.zip`


src/Ryujinx.Ava/UI/Controls/ApplicationContextMenu.axaml
```diff
@@ -77,4 +77,11 @@
             Header="{locale:Locale GameListContextMenuExtractDataLogo}"
             ToolTip.Tip="{locale:Locale GameListContextMenuExtractDataLogoToolTip}" />
     </MenuItem>
+	<MenuItem Header="Save Management">
+		<MenuItem
+            Click="BackupSaveData_Click"
+            Header="Backup Save Data"
+			IsEnabled="True"
+            ToolTip.Tip="Create a zip file that contains the saves for the currently active profile" />
```
patrick-hovsepian, line 85: need locales

tarrouye:
I don't think this backup structure makes sense. 
It is redundant to backup Device, BCAT, etc saves for each user, as they are the same for all profiles.

patrick-hovsepian:
> Device, BCAT, etc saves for each user, as they are the same for all profiles. 

I wasn't aware that BCAT and Device were the same for all profiles. Thanks, I'll fix that. The user specific data is absolutely different though so that warrants its own entry.

> I also find it weird to zip each folder separately instead of zipping the container.

Fair, but that leaf `yyy-MM-dd` directory is meant to be the container anyway

> Also, as a user, I would expect such an option to backup ALL the save data, but this only backs it up for the active profile. (I know the tooltip makes that clear, but I still don't like the approach)

Right, that's the point of that tip and I actually find your suggestion counter-intuitive. I'm active on Ryujinx with a single profile at a time so the user-specific option makes sense in this view. This PR is meant to be a stepping stone if you will for a better overall save management and can be used for an all data save in the future.

> Additionally, I don't think this should default to dumping the backup into Ryujinx's data folder. I would much prefer defaulting to a save file dialog that lets the user choose the output location of the backup. 

Not in scope for this PR, same reasoning as above comment.

> It also seems odd to add a submenu with only one item, at the very bottom. This new feature should live with the existing save options in that context menu IMO, as they are closely related.

I see imports going here too which is why I made it separate.

Issue #5172, opened 2023-05-31
[Bug] Opening >3 instances of Ryujinx at the same time crashes the emulator (due to being used by another process)


yell0wsuit:
### Description of the issue

Per discussion in PR #5163, users may open many instances to play multiplayer games (or something idk). But the problem is that opening more than 3 instances of Ryujinx at the same time will crash the 4th instance.
This is because the emulator tries to overwrite the oldest log, but that oldest log is being used by the previous instance. Also the maximum log file written is only 3, so there's that.

### Reproduction steps

Open 3 instances of Ryujinx, then open the 4th one.
Alternately, open Ryujinx quickly 3 times in a row will cause the issue.

### Log file

```
00:00:00.090 |E| Application : Unhandled exception caught: System.IO.IOException: The process cannot access the file 'E:\Nintendo emulators\ryujinx\Logs\Ryujinx_1.1.842_2023-05-31_11-17-37.log' because it is being used by another process.
   at System.IO.FileSystem.DeleteFile(String)
   at System.IO.FileInfo.Delete()
   at Ryujinx.Common.Logging.FileLogTarget..ctor(String path, String name, FileShare fileShare, FileMode fileMode) in D:\a\Ryujinx\Ryujinx\src\Ryujinx.Common\Logging\Targets\FileLogTarget.cs:line 29
   at Ryujinx.Common.Logging.FileLogTarget..ctor(String path, String name) in D:\a\Ryujinx\Ryujinx\src\Ryujinx.Common\Logging\Targets\FileLogTarget.cs:line 16
   at Ryujinx.Ui.Common.Configuration.LoggerModule.ReloadFileLogger(Object sender, ReactiveEventArgs`1 e) in D:\a\Ryujinx\Ryujinx\src\Ryujinx.Ui.Common\Configuration\LoggerModule.cs:line 82
   at Ryujinx.Common.ReactiveObject`1.set_Value(T value) in D:\a\Ryujinx\Ryujinx\src\Ryujinx.Common\ReactiveObject.cs:line 39
   at Ryujinx.Ui.Common.Configuration.ConfigurationState.Load(ConfigurationFileFormat configurationFileFormat, String configurationFilePath) in D:\a\Ryujinx\Ryujinx\src\Ryujinx.Ui.Common\Configuration\ConfigurationState.cs:line 1400
[...]

PR #5173, opened 2023-05-31
Log file: append the current process ID to make log file name unique

yell0wsuit:
Scenario: If a user opens Ryujinx 2+ times quickly in a row (in 1 second), the emulator will crash. This is because the emulator tries to read/overwrite the current log file, causing the 2nd instance crash due to "being used by another process".

Since process ID is always unique, so this simple PR will resolve the race condition outlined in #5172.

image


marysaka:
That's not a proper solution here.

That will still cause the issue as the race condition is on log file clean up.

A proper solution would be to catch the removal error and log it in the logs.

Closing as invalid.

yell0wsuit:
> A proper solution would be to catch the removal error and log it in the logs.

But what if they want to open 2 Ryujinx instances at the same time? Because the file name is the same (down to seconds), then the second Ryujinx instance will always crash. So how do you log it when the log file is being used?

PR #5180, opened 2023-05-31
Mark classes as "sealed"

marco-carvalho:
There are some indirect performance benefits related to this approach:

**Preventing virtual method calls:** When a class is sealed, it tells the JIT compiler that there is no need to check whether a method is overridden in a derived class. This can lead to a minor optimization because it can remove the need for virtual method table lookups and can potentially allow for inlining methods.

**Better predictability:** From a design point of view, a sealed class can't be extended, so it won't have performance costs that can come from poorly designed subclasses. This is more of a software design benefit rather than a runtime performance benefit.

**Less Memory Overhead:** Since a sealed class cannot be subclassed, it eliminate the potential for additional memory overhead that can come with subclassing (e.g., additional properties or methods).

edit - useful content:
https://www.youtube.com/watch?v=d76WWAD99Yo
https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852

edit2:

Benchmark

```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    private PublicClass _publicClass => new();
    private PublicSealedClass _publicSealedClass => new();
    private InternalClass _internalClass => new();
    private InternalSealedClass _internalSealedClass => new();
    private Class _class => new();
    private SealedClass _sealedClass => new();

    [Benchmark]
    public void PublicInt() => _publicClass.IntMethod();

    [Benchmark]
    public void PublicSealedInt() => _publicSealedClass.IntMethod();

    [Benchmark]
    public void Int() => _class.IntMethod();

    [Benchmark]
    public void SealedInt() => _sealedClass.IntMethod();

    [Benchmark]
    public void InternalInt() => _internalClass.IntMethod();

    [Benchmark]
    public void InternalSealedInt() => _internalSealedClass.IntMethod();
}

public class BaseClass
{
    public virtual int IntMethod() => 0;
}

public class PublicClass : BaseClass
{
    public override int IntMethod() => 1;
}

public sealed class PublicSealedClass : BaseClass
{
    public override int IntMethod() => 1;
}

class Class : BaseClass
{
    public override int IntMethod() => 2;
}

sealed class SealedClass : BaseClass
{
    public override int IntMethod() => 2;
}

internal class InternalClass : BaseClass
{
    public override int IntMethod() => 3;
}

internal sealed class InternalSealedClass : BaseClass
{
    public override int IntMethod() => 3;
}
```

Results:

```sh
|            Method |      Mean |     Error |    StdDev |    Median | Rank |   Gen0 | Allocated |
|------------------ |----------:|----------:|----------:|----------:|-----:|-------:|----------:|
|         PublicInt | 3.5945 ns | 0.2361 ns | 0.6660 ns | 3.3185 ns |    2 | 0.0076 |      24 B |
|   PublicSealedInt | 0.0006 ns | 0.0020 ns | 0.0019 ns | 0.0000 ns |    1 |      - |         - |
|               Int | 4.2589 ns | 0.0361 ns | 0.0320 ns | 4.2609 ns |    3 | 0.0077 |      24 B |
|         SealedInt | 0.0000 ns | 0.0000 ns | 0.0000 ns | 0.0000 ns |    1 |      - |         - |
|       InternalInt | 4.2403 ns | 0.0560 ns | 0.0468 ns | 4.2401 ns |    3 | 0.0077 |      24 B |
| InternalSealedInt | 0.0002 ns | 0.0005 ns | 0.0005 ns | 0.0000 ns |    1 |      - |         - |
```


src/ARMeilleure/CodeGen/Arm64/Assembler.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace ARMeilleure.CodeGen.Arm64
 {
-    class Assembler
+    sealed class Assembler
```
marysaka, line 9: I don't think this is needed for internal class to get this kind of optimization?

marco-carvalho: @marysaka I've updated the description of the PR showing that `sealed` optimize this case

lostromb:
I recall watching Nick's video on this and applying `sealed` on a 300K-line codebase that didn't use sealed classes beforehand... Didn't get a measurable change in performance, but I guess it didn't hurt either v0v

AcK77:
> Didn't get a measurable change in performance, but I guess it didn't hurt either v0v

Yeah I wonder the same, what's the real performance improvement in Ryujinx?
Because I'm not a huge fan to change all existing classes to `sealed`, one of your link above said:

> [...] so we’ve adopted a general policy that all non-public types that can be sealed should be

which isn't the case here.

marco-carvalho:
> Yeah I wonder the same, what's the real performance improvement in Ryujinx?

@AcK77 There is a benchmark in the description of the PR showing that `sealed` have performance gains

jduncanator:
I don't think the performance gains from this is worth the overhead in maintainability of this change, it's also an abuse of a keyword that has a very specific semantic meaning. None of these classes benefit from the semantics of being sealed.

If you can show a demonstrable improvement in performance at runtime of Ryujinx from this change, then I'm happy to revisit it, but without any evidence of real world improvements, I don't see the merit in making this change.

AcK77:
Sorry but this was discussed privately and the majority of devs core team thinks this change is too large and don't prove any improvement on Ryujinx directly, feel free to provide more real informations and that could be discussed.
I close it. Thanks for your contribution.

PR #5181, opened 2023-05-31
Share ResourceManager between vertex A and B shaders

gdkchan:
This fixes a regression from #4646 that caused Borderlands 2 to crash before the title screen while generating a shader. It had an issue while trying to reconcile the bindings between the vertex A and B stages when they are combined. It would have missing bindings which would cause crashes later on as it would not be able to find the constant buffer from the binding when generating the SPIR-V or GLSL code. This change fixes the issue by using the same `ResourceManager` instance for both stages. This should be fine since they share the same constant buffers (and as consequence, also shares the same storage buffer and textures/images, since the base address for the storage buffer and handles for the textures also comes from constant buffers).

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/GpuAccessorBase.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/GpuAccessorBase.cs
@@ -17,8 +17,6 @@ namespace Ryujinx.Graphics.Gpu.Shader
         private readonly ResourceCounts _resourceCounts;
         private readonly int _stageIndex;
 
-        private readonly int[] _constantBufferBindings;
-
         /// <summary>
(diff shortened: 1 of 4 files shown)


marysaka (approved):

riperiperi (approved):

x4080:
@gdkchan I just tested, and its not crashed anymore, but I noticed that the release version of mac os, when vsync off, in borderlands can reach almost 60fps on my m2 pro. But in this build, it cannot reach that (or crash when I tried to). Just wanna let you know - no pressure

gdkchan:
You can try #4899, it's the only missing change that affects performance IIRC.

x4080:
@gdkchan thanks again

PR #5184, opened 2023-05-31
test: Make tests runnable on system without 4KiB page size

marysaka:
This fixes running tests on Asahi Linux that uses 16KiB pages.

I disabled `CpuTestThumb::TestRandomTestCases` on system without 4KiB pages for now as it needs to be rewritten.
Also disable `Vshl` test on ARM64 as unicorn crash when trying to run it.

### DIFF
--- a/src/Ryujinx.Tests/Cpu/CpuTest.cs
+++ b/src/Ryujinx.Tests/Cpu/CpuTest.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Tests.Cpu
     public class CpuTest
     {
         protected static readonly ulong Size = MemoryBlock.GetPageSize();
-        protected static readonly ulong CodeBaseAddress = Size;
+        protected static ulong CodeBaseAddress = Size;
         protected static ulong DataBaseAddress = CodeBaseAddress + Size;
 
         private static bool Ignore_FpcrFz = false;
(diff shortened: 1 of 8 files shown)


src/Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -39,12 +39,24 @@ public class CpuTest
...
-            _memory = new MemoryManager(_ram, 1ul << 16);
+            _memory = new MemoryManager(_ram, 1ul << (pageBits + 4));
             _memory.IncrementReferenceCount();
-            _memory.Map(CodeBaseAddress, 0, Size * 2, MemoryMapFlags.Private);
+
+            // Some tests depends on hardcoded address that were computed for 4KiB.
+            // We change the layout on non 4KiB platform to keep compat here.
+            if (Size > 0x1000)
+            {
+                DataBaseAddress = 0;
```
gdkchan, line 52: Not sure why this is needed. Data and code can live on the same 16K page, since it is not going to actually execute that code, and both are RW.

src/Ryujinx.Tests/Cpu/CpuTest.cs
```diff
@@ -39,12 +39,24 @@ public class CpuTest
...
+            int pageBits = (int)MathF.Log2(Size);
 
             _ram = new MemoryBlock(Size * 2);
-            _memory = new MemoryManager(_ram, 1ul << 16);
+            _memory = new MemoryManager(_ram, 1ul << (pageBits + 4));
             _memory.IncrementReferenceCount();
-            _memory.Map(CodeBaseAddress, 0, Size * 2, MemoryMapFlags.Private);
+
+            // Some tests depends on hardcoded address that were computed for 4KiB.
+            // We change the layout on non 4KiB platform to keep compat here.
```
gdkchan, line 49:
```suggestion
            // We change the layout on non 4KiB platforms to keep compat here.
```

src/Ryujinx.Tests/Cpu/CpuTest32.cs
```diff
@@ -33,12 +33,24 @@ public class CpuTest32
         [SetUp]
         public void Setup()
         {
-            _currAddress = CodeBaseAddress;
+            int pageBits = (int)MathF.Log2(Size);
```
gdkchan, line 36: Same comments as the other file.

src/Ryujinx.Tests/Cpu/CpuTestThumb.cs
```diff
@@ -268,6 +268,12 @@ public void SubSpTest()
         [Test]
         public void TestRandomTestCases([ValueSource(nameof(RandomTestCases))] PrecomputedThumbTestCase test)
         {
+            if (Size != 0x1000)
+            {
+                // TODO: Change it to depends on DataBaseAddress instead.
```
gdkchan, line 273:
```suggestion
                // TODO: Change it to depend on DataBaseAddress instead.
```

src/Ryujinx.Tests/Cpu/CpuTest.cs
marysaka: I really would prefer mapping to be separated for unicorn here

AcK77 (approved):

src/Ryujinx.Tests/Cpu/CpuTest.cs
gdkchan: What about this question ^

marysaka: Because it does a clear separation of what is RW and what is RX

TSRBerry (approved):
Lgtm! Nice to see some cleanup of the CPU tests!

PR #5185, opened 2023-05-31
Vulkan: Include DepthMode in ProgramPipelineState

riperiperi:
Depth mode has been added to the pipeline state by #5027, but was not properly recorded for background compilation or shader caching, which would result in the pipeline compiled to mismatch on first draw, and have to compile another. This could cause additional stuttering.

This PR adds depth mode to the pipeline state and handles it in the converter. Since this already exists in the specialization state, it is always sourced from there when not available (but really, just always). In future it might be worth passing the read size of the struct to the method to decide to overwrite values based on that, for example if this state were to somehow become optional.


gdkchan (approved):
lgtm, nice catch.

marysaka (approved):

PR #5191, opened 2023-06-01
misc: Implement address space size workarounds

marysaka:
This adds code to support userland with less than 39 bits of address space available by testing reserving multiple sizes and reducing guess address space when needed.

This is required for ARM64 support when the kernel is configured to use 63..39 bits for kernel space. (meaning only 39 bits is available to userland)

### DIFF
--- a/src/Ryujinx.Cpu/AddressSpace.cs
+++ b/src/Ryujinx.Cpu/AddressSpace.cs
@@ -177,7 +177,7 @@ namespace Ryujinx.Cpu
             AddressSpaceSize = addressSpaceSize;
         }
 
-        public static bool TryCreate(out AddressSpace addressSpace, MemoryBlock backingMemory, ulong asSize, bool supports4KBPages)
+        public static bool TryCreate(MemoryBlock backingMemory, ulong asSize, bool supports4KBPages, out AddressSpace addressSpace)
         {
             addressSpace = null;
 
@@ -197,14 +197,10 @@ namespace Ryujinx.Cpu
 
                     break;
                 }
-                catch (OutOfMemoryException) { }
-                finally
+                catch (OutOfMemoryException)
                 {
-                    if (addressSpace == null)
-                    {
-                        baseMemory?.Dispose();
-                        mirrorMemory?.Dispose();
-                    }
+                    baseMemory?.Dispose();
+                    mirrorMemory?.Dispose();
                 }
             }
 
--- a/src/Ryujinx.Cpu/AddressSpace.cs
+++ b/src/Ryujinx.Cpu/AddressSpace.cs
@@ -183,8 +183,10 @@ namespace Ryujinx.Cpu
 
             MemoryAllocationFlags asFlags = MemoryAllocationFlags.Reserve | MemoryAllocationFlags.ViewCompatible;
 
+            ulong minAddressSpaceSize = Math.Min(asSize, 1UL << 36);
 
-            for (ulong addressSpaceSize = asSize; addressSpaceSize >= (1UL << 36); addressSpaceSize >>= 1)
+            // Attempt to create the address space with expected size or try to reduce it until it succeed.
+            for (ulong addressSpaceSize = asSize; addressSpaceSize >= minAddressSpaceSize; addressSpaceSize >>= 1)
             {
                 MemoryBlock baseMemory = null;
                 MemoryBlock mirrorMemory = null;
--- a/src/Ryujinx.Cpu/AddressSpace.cs
+++ b/src/Ryujinx.Cpu/AddressSpace.cs
@@ -5,7 +5,7 @@ using System;
 
 namespace Ryujinx.Cpu
 {
-    class AddressSpace : IDisposable
+    public class AddressSpace : IDisposable
     {
         private const ulong PageSize = 0x1000;
 
@@ -154,7 +154,9 @@ namespace Ryujinx.Cpu
         public MemoryBlock Base { get; }
         public MemoryBlock Mirror { get; }
 
-        public AddressSpace(MemoryBlock backingMemory, ulong asSize, bool supports4KBPages)
+        public ulong AddressSpaceSize { get; }
+
+        public AddressSpace(MemoryBlock backingMemory, MemoryBlock baseMemory, MemoryBlock mirrorMemory, ulong addressSpaceSize, bool supports4KBPages)
         {
             if (!supports4KBPages)
             {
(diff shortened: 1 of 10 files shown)


src/Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
```diff
@@ -22,6 +22,8 @@ public sealed class MemoryManagerHostMapped : MemoryManagerBase, IMemoryManager,
         public const int PageToPteShift = 5; // 32 pages (2 bits each) in one ulong page table entry.
         public const ulong BlockMappedMask = 0x5555555555555555; // First bit of each table entry set.
 
+        public static bool Supports4KBPageSize => MemoryBlock.GetPageSize() == PageSize;
```
gdkchan, line 25: It's rather weird to have this static property in addition to the other one (even more so because they have different names). I would prefer if the static one was removed and the code was inlined at the call site (since it''s something pretty simple anyway).

src/Ryujinx.HLE/HOS/ArmProcessContextFactory.cs
```diff
@@ -49,31 +50,49 @@ public IProcessContext Create(KernelContext context, ulong pid, ulong addressSpa
...
-                processContext = new ArmProcessContext<HvMemoryManager>(pid, cpuEngine, _gpu, memoryManager, for64Bit);
+                processContext = new ArmProcessContext<HvMemoryManager>(pid, cpuEngine, _gpu, memoryManager, addressSpaceSize, for64Bit);
             }
             else
             {
                 MemoryManagerMode mode = context.Device.Configuration.MemoryManagerMode;
 
                 if (!MemoryBlock.SupportsFlags(MemoryAllocationFlags.ViewCompatible))
                 {
+                    Logger.Error?.Print(LogClass.Cpu, "Host system doesn't support view, falling back to software page table");
```
gdkchan, line 61:
```suggestion
                    Logger.Error?.Print(LogClass.Cpu, "Host system doesn't support views, falling back to software page table");
```

gdkchan: Also this should be a warning, not an error.

src/Ryujinx.HLE/HOS/ArmProcessContextFactory.cs
```diff
@@ -49,31 +50,49 @@ public IProcessContext Create(KernelContext context, ulong pid, ulong addressSpa
...
 
                 var cpuEngine = new JitEngine(_tickSource);
 
+                AddressSpace addressSpace = null;
+
+                if (mode == MemoryManagerMode.HostMapped || mode == MemoryManagerMode.HostMappedUnsafe)
+                {
+                    if (!AddressSpace.TryCreate(out addressSpace, context.Memory, addressSpaceSize, MemoryManagerHostMapped.Supports4KBPageSize))
+                    {
+                        Logger.Error?.Print(LogClass.Cpu, "Address space creation failed, falling back to software page table");
```
gdkchan, line 74: Should also be a warning.

src/Ryujinx.HLE/HOS/ArmProcessContextFactory.cs
```diff
@@ -49,31 +50,49 @@ public IProcessContext Create(KernelContext context, ulong pid, ulong addressSpa
...
                         break;
 
                     case MemoryManagerMode.HostMapped:
                     case MemoryManagerMode.HostMappedUnsafe:
-                        bool unsafeMode = mode == MemoryManagerMode.HostMappedUnsafe;
-                        var memoryManagerHostMapped = new MemoryManagerHostMapped(context.Memory, addressSpaceSize, unsafeMode, invalidAccessHandler);
-                        processContext = new ArmProcessContext<MemoryManagerHostMapped>(pid, cpuEngine, _gpu, memoryManagerHostMapped, for64Bit);
+                        if (addressSpaceSize != addressSpace.AddressSpaceSize)
+                        {
+                            Logger.Warning?.Print(LogClass.Emulation, $"Allocated address space (0x{addressSpace.AddressSpaceSize:x8}) is smaller than guest application requirements (0x{addressSpaceSize:x8})");
```
gdkchan, line 91:
```suggestion
                            Logger.Warning?.Print(LogClass.Emulation, $"Allocated address space (0x{addressSpace.AddressSpaceSize:X}) is smaller than guest application requirements (0x{addressSpaceSize:X})");
```
That is better IMO.

src/Ryujinx.HLE/HOS/Kernel/Memory/KPageTableBase.cs
```diff
@@ -196,23 +201,30 @@ private class Region
...
+                    int addressSpaceWidth = (int)ulong.Log2(_reservedAddressSpaceSize);
+
+                    if (_reservedAddressSpaceSize < addrSpaceEnd)
+                    {
+                        addrSpaceEnd = 1UL << addressSpaceWidth;
+                    }
+
+                    aliasRegion.Size = 1UL << (addressSpaceWidth - 3);
                     heapRegion.Size = 0x180000000;
-                    stackRegion.Size = 0x80000000;
```
gdkchan, line 208: I'd prefer if the custom size "allocation" was separate from the hard-coded sizes that the official kernel uses, to have a clear separation of what is custom emulator behaviour and what the official kernel does.

gdkchan (commented):
> (meaning only 38 bits is available to userland)

I guess you mean 39 bits here?

I think this is "fine", although in the long term I'd prefer to use a solution that allows us to accurately emulate the guest address space. I was thinking in a memory manager mode that is a middle ground between software and host mapped.

src/Ryujinx.Cpu/AddressSpace.cs
```diff
@@ -163,17 +165,50 @@ public AddressSpace(MemoryBlock backingMemory, ulong asSize, bool supports4KBPag
...
 
             _backingMemory = backingMemory;
             _supports4KBPages = supports4KBPages;
 
+            Base = baseMemory;
+            Mirror = mirrorMemory;
+            AddressSpaceSize = addressSpaceSize;
+        }
+
+        public static bool TryCreate(out AddressSpace addressSpace, MemoryBlock backingMemory, ulong asSize, bool supports4KBPages)
```
TSRBerry, line 180:
```suggestion
        public static bool TryCreate(MemoryBlock backingMemory, ulong asSize, bool supports4KBPages, out AddressSpace addressSpace)
```
Try methods usually have the output parameter at the end.

src/Ryujinx.Cpu/AddressSpace.cs
```diff
@@ -163,17 +165,50 @@ public AddressSpace(MemoryBlock backingMemory, ulong asSize, bool supports4KBPag
...
+                    break;
+                }
+                catch (OutOfMemoryException) { }
+                finally
+                {
+                    if (addressSpace == null)
+                    {
+                        baseMemory?.Dispose();
+                        mirrorMemory?.Dispose();
+                    }
```
TSRBerry, line 207: Why are you doing this here instead of doing it in the catch-block above?

src/Ryujinx.Cpu/Jit/MemoryManagerHostMapped.cs
TSRBerry, line 25: Is there a reason this was added instead of using `Supports4KBPages`?

TSRBerry (commented):
I agree with gdkchan's comments and commented some questions as well.

TSRBerry (approved):
Lgtm! :D

src/Ryujinx.Cpu/AddressSpace.cs
```diff
@@ -163,17 +165,46 @@ public AddressSpace(MemoryBlock backingMemory, ulong asSize, bool supports4KBPag
...
+        public static bool TryCreate(MemoryBlock backingMemory, ulong asSize, bool supports4KBPages, out AddressSpace addressSpace)
+        {
+            addressSpace = null;
+
             MemoryAllocationFlags asFlags = MemoryAllocationFlags.Reserve | MemoryAllocationFlags.ViewCompatible;
 
-            Base = new MemoryBlock(asSize, asFlags);
-            Mirror = new MemoryBlock(asSize, asFlags);
+
+            for (ulong addressSpaceSize = asSize; addressSpaceSize >= (1UL << 36); addressSpaceSize >>= 1)
```
gdkchan, line 187: Might be worth adding a comment here explaining why this is done?

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

Issue #4006, opened 2022-12-03
[Feature request] Avalonia - Backup and restore ALL Saves Games


jojodunet:
## Feature Request
Export / Import (one shot) all the Saves Games for backup or restore in Avalonia Save Manager

#### Nature of Request:  
- Addition of #3476

### Why would this feature be useful?
- Feature integrated natively in Avalonia Save Manager
- Easily backup all the Saves Games
- Easily restore all the Saves Games from a backup

jojodunet:
#4113

patrick-hovsepian:
I'm working on this

PR #5194, opened 2023-06-01
[Save Management] Backup and Restore User Profile Saves

patrick-hovsepian:
Edit: #4006 

~Introduce an option to backup the save data for a given application, for the currently active profile.~
~User, Device, and BCAT, data is saved to a single zip file. The directory is `Ryujinx/backup/[title]/[user]`~

Each type of save data is placed in it's own folder. This is easily configurable if saving bcat/device data isn't desirable (and would speed up the operation for slower machines - maybe a future config option?)
image

Edit: 
Since initially opening, the implementation has been significantly changed to create a backup of _all save data for the given user profile at a location of the user's choosing_ 

~As stated in a comment, **I'd like to split up the export and import into two separate PRs and work items.** My first pass will address the export. **_If and when_** that get's merged I'll work on imports as a fast follow.~ I'll add imports alongside the backup. It'll make testing easier and hopefully incentivize trying to get this merged sooner? I'd also like to address the save corruption case after.

New implementation:
The functionality is attached the user profile save management window
image

Once the user chooses a directory, the outcome is a save tagged with their profile name at the directory they chose
image

The contents of the resulting zip. The folder names are the TitleIds (in decimal)
image

There's also a metadata tag file that's generated too. This is purely for user convenience. I do not envision using this in the import process. It's for user convenience only. The contents of that file
image


sakuka47:
Hi, there,

It makes coffee :)

May I make a few comments?  

1. Add a backup import option.
2. Add a backup import redirection option.
Automatically create the backup in the ryujinx backup folder and let the user choose a folder of their choice.

patrick-hovsepian:
Both are good suggestions that will be done as separate PRs. This PR is only scoped to creating the backup at a predetermined location.

patrick-hovsepian:
@tarrouye 
You seem to not understand incremental PRs. 

> Ryujinx already offers to open the location of the saves for you, it takes like 5 seconds at that point to zip the contents and put it anywhere you want

yeah, and that's more than the 2 clicks here.

>  (which is super easy to implement.. including providing a default location such as the backup folder you create here) 

You're more than welcome to implement it once this is merged.

>  I actually consider this a net negative change

Great, then you or the maintainers can just close it.

> Why include Device & BCAT saves

Because it's part of the save data. Might be redundant but that's kinda the point of backups right? The code itself is modular enough to remove these extremely easily if needed (not sure if you looked at the actual implementation or are just going off of the description screenshots)

> Why not include Ryujinx metadata in the backup ?

Because this is exclusively scoped to the game save data...

> Also I mentioned it on your last PR regarding this, but it really does not make sense to add a "Save Management" menu item with only one subitem, and it makes no sense to separate it from the existing 3 save-related functions in that menu.
If you plan to add more subitems in a future PR, then make it a submenu at that point. Don't make it a submenu when there is only a single item.

I'll just let you go back and read my comments from that other PR.

> Just my 2 cents :)

Appreciate actual feedback rather than a single emoji 😄

tarrouye:
> You seem to not understand incremental PRs.

I understand it. I don't think this one makes sense. Like I said, just my opinion, and it's not my project so whether it is closed or not is absolutely not up to me. 

No offense intended.

patrick-hovsepian:
The previous PR #5171 in question to preserve the discussion

@tarrouye You're one of the only ones with context of that PR and you still chose to speak in circles and not continue from my last comment to you.

Please only continue to comment if you have something new (and constructive) to add ❤️

gdkchan:
It would be nice to know which problem this change is trying to solve. I know a few issues that save backups could potentially help with:
- Save data loss due to the emulator randomly crashing, power outages or anything that could prevent save data to be written.
- Quickly packing up a save to send it to someone (for example, to allow us to reproduce some bug easily).
- Syncing saves between multiple devices.

For the first point, it sort of helps with it, but it's not really ideal since the user has to manually click on a menu to make a backup, and most people would likely forget to click it and lose the data (since you can't predict when it will happen). For the second one, it also doesn't help since it saves it inside a backup folder that the user has to open and locate the correct file there, it would probably be faster to just click on "Open * Save Directory" and manually zip the contents. And the third one would require something that can pack saves from all games with an import option, not just from a single game (and in any case, it would still not be a ideal solution for that scenario,, but it would be at least better than what we have now).

Just to be clear, I don't expect that change to solve all problems above, and I'm in favor of smaller incremental PRs, but I don't see the value this change on its own adds. I think if this change is exposing a new option, that option should do something useful.

About the new save management sub-menu, I think it is fine, but the existing save options should probably be moved inside it. Also minor nitpick, I think it would be better if it was located above "Cache Management" so that all management related sub-menus are close to each other.

patrick-hovsepian:
Genuinely, thank you for the feedback, I appreciate it. 

I'm not trying to be snarky but did y'all actually read the implementation? It's purposely written in a way to help achieve the other examples you mention. 

It automatically opens the directory where the backup was saved to which I think makes cases 2 and 3 simpler as you described above.

If the actual save logic isn't valid or has issues it's separate from the user experiences in which it's used. I appreciate you calling it out though, there is nothing today, so _anything_ is better.

gdkchan:
> It automatically opens the directory where the backup was saved to which I think makes cases 2 and 3 simpler as you described above.

It does make case 2 simpler, but for that it would be even simpler and more effective to show a dialog where the user can select where the file will be saved. For case 3 I'm not sure how it would help. It would be only helpful if you want to move a single save to another device. At this point you might as well just open the folder and copy the files.

> If the actual save logic isn't valid or has issues it's separate from the user experiences in which it's used. I appreciate you calling it out though, there is nothing today, so _anything_ is better.

I think one depends on the other. Depending on the user experience that we expect from this, the implementation could be different. And generally I try to understand which problem the change is trying to solve before looking at the code.

My main question right now is what's the point of the current folder structure for daily backups. I'm assuming that the intended use is that the user must manually click there and make a backup at least once per day for case 1 that I described on my previous message. For this case the problem I described still exists, people would simply forget to click on it and ultimately the option wouldn't add much.

If you plan to evolve this functionality in multiple PRs, I think it would be better to outline what you plan to do so that we can understand how that is all supposed to work once everything is done.

gdkchan:
Here's what I think would make the most sense for each case:
- For the save corruption, I think a good option was having the emulator automatically backup the save files somewhere when the emulation is stopped. If the emulator crashes or is interrupted abruptly it would not backup the save, so there's no risk of it putting a broken save there, though it would still need to keep more than one save since the user would likely only discover that the save is corrupted when the game is launched again.
- For quickly packing, I think a dialog that asks the user where to save the zip would be the best.
- For device transfer, I think an option to export all save data and a import option would be the best.

Both 1 and 3 could benefit from a backup structure like what you propose here, but I'd suggest some changes:
- I think the output location should be selected by the user. I'm not a fan of storing it on the emulator folder. By default it would not be set which means that the feature is disabled.
- I don't think it's necessary to keep one save per day. Probably the last 3 saves or so is enough. Keeping a lot of saves will waste disk space.

Also I think the folder structure should be something that makes sense for cases 1 and 3. And it should be something the user can import from the emulator too. I'm not sure about the user profile separation since the user may want to import it into a different user profile.

IsaacMarovitz:
This should let you save at a place of your choosing.

patrick-hovsepian:
> If you plan to evolve this functionality in multiple PRs, I think it would be better to outline what you plan to do so that we can understand how that is all supposed to work once everything is done. 

I should have been more clear about this. I did loosely lay out my plans in Discord but that's moot. My hopes was that a smaller PR would lead to a quicker review process to iterate on but I guess it was too small a slice. 

That aside, let me state the problem that this (now evolved) PR is attempting to solve for.
> For device transfer, I think an option to export all save data and a import option would be the best. 

That's the first case I'll solve for. However, I'd like to split up the export and import into two separate PRs and work items. My first pass will address the export and _if and when_ that get's merged I'll work on imports as a fast follow. I'd also like to address the save corruption case after.

I'll shortly push some changes that attempt to address exporting all of a profile's save data in addition to the bcat and device data.

patrick-hovsepian:
I'll amend my previous statement, I'll add import functionality here too.

patrick-hovsepian:
I'll re-open after implementing some of the suggestions on discord.

PR #5202, opened 2023-06-02
Armeilleure: Fix support for Windows on ARM64

marysaka:
Tested on Windows DevKit 2023 with #5184.

### DIFF
--- a/src/ARMeilleure/Translation/Cache/JitCache.cs
+++ b/src/ARMeilleure/Translation/Cache/JitCache.cs
@@ -6,10 +6,11 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 
 namespace ARMeilleure.Translation.Cache
 {
-    static class JitCache
+    static partial class JitCache
     {
         private const int PageSize = 4 * 1024;
         private const int PageMask = PageSize - 1;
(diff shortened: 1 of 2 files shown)


src/ARMeilleure/Translation/Cache/JitCache.cs
```diff
@@ -77,7 +82,14 @@ public static IntPtr Map(CompiledFunction func)
...
                     ReprotectAsExecutable(funcOffset, code.Length);
 
-                    _jitCacheInvalidator.Invalidate(funcPtr, (ulong)code.Length);
+                    if (OperatingSystem.IsWindows() && RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
+                    {
+                        FlushInstructionCache(Process.GetCurrentProcess().Handle, funcPtr, (UIntPtr)code.Length);
+                    }
+                    else
+                    {
+                        _jitCacheInvalidator.Invalidate(funcPtr, (ulong)code.Length);
```
gdkchan, line 91: Even if this is not called, we still instantiate the `JitCacheInvalidation`, and it will map the code to do the invalidation which is ultimately unused. I think it would be better to clean that up since it's a bit wasteful. You could do lazy instantiation of the JIT cache invalidator, and only do it for platforms other than Windows and macOS/iOS (plus remove the macOS checks on the `JitCacheInvalidation` class).

gdkchan (approved):
lgtm, thanks.

PR #5203, opened 2023-06-02
Allow BGRA storage images on Vulkan

### LINKED ISSUE #5193: [Bug] Colors are all wrong in persona 4 golden
### Description of the issue

The game looks like it has a colorblind filter on 
Screenshot 2023-06-01 at 3 28 15 PM

### Reproduction steps

Just launching the game is enough

### Log file

[Ryujinx_1.1.855_2023-06-01_15-20-06.log](https://github.com/Ryujinx/Ryujinx/files/11629332/Ryujinx_1.1.855_2023-06-01_15-20-06.log)

### OS

Mac os ventura 13.4

### Ryujinx version

1.1.855

### Game version

_No response_

### CPU

m2

### GPU

m2

### RAM

8

### List of applied mods

N/A

### Additional context?

_No response_

gdkchan:
There's a bug on the FXAA and SMAA filters, where if the presentation texture has a BGRA format, it will create a intermediary BGRA texture, but bind it as a RGBA format. This causes the R and B components to be inverted once the filter is applied. It seems there was an attempt to correct that by inverting the R and B components on the swizzle, but that does nothing on *storage image operations*, at least on Vulkan.

This change fixes the issue by simply allowing the BGRA format to be bound directly as a storage image. Most vendors supports this. It also makes the code simpler and more efficient as we don't need to create a new view with different format.

Fixes the issue on Persona 4 Golden and any other game presenting BGRA textures, with FXAA and SMAA filters.
Before:
image
After:
image
Fixes #5193.

### DIFF
--- a/src/Ryujinx.Graphics.GAL/Format.cs
+++ b/src/Ryujinx.Graphics.GAL/Format.cs
@@ -383,6 +383,7 @@ namespace Ryujinx.Graphics.GAL
                 case Format.R10G10B10A2Unorm:
                 case Format.R10G10B10A2Uint:
                 case Format.R11G11B10Float:
+                case Format.B8G8R8A8Unorm:
                     return true;
             }
 
--- a/src/Ryujinx.Graphics.Vulkan/Effects/FsrScalingFilter.cs
+++ b/src/Ryujinx.Graphics.Vulkan/Effects/FsrScalingFilter.cs
@@ -96,8 +96,6 @@ namespace Ryujinx.Graphics.Vulkan.Effects
             {
                 var originalInfo = view.Info;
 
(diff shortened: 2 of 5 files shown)


gdkchan:
This is a regression caused by #5003 (sort of). The fact that the "blit" operations took swizzle into account before that change compensated for the fact that the swizzle does nothing on the compute image store that the filter does. So once that was fixed, those filters broke for BGRA formats.

AcK77 (approved):

riperiperi (approved):
Seems reasonable. I don't have any affected games to test.

Linear524:
Persona 5 Royal main menu squares fixed with that :)

PR #5204, opened 2023-06-02
ava: Fix Input Touch

AcK77:
There is currently a regression caused by #4900 where the wrong input touch is released. Since `Avalonia.Input.MouseButton` first entry is `None` and the `Avalonia.Input.PointerUpdateKind` first entry is `Left` we have to change the index of `MouseButton` in the released event handler. This is now fixed.

### DIFF
--- a/src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
+++ b/src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
@@ -30,14 +30,14 @@ namespace Ryujinx.Ava.Input
             _window = window;
 
             _widget.PointerMoved        += Parent_PointerMovedEvent;
-            _widget.PointerPressed      += Parent_PointerPressEvent;
-            _widget.PointerReleased     += Parent_PointerReleaseEvent;
-            _widget.PointerWheelChanged += Parent_ScrollEvent;
+            _widget.PointerPressed      += Parent_PointerPressedEvent;
+            _widget.PointerReleased     += Parent_PointerReleasedEvent;
+            _widget.PointerWheelChanged += Parent_PointerWheelChanged;
             
             _window.PointerMoved        += Parent_PointerMovedEvent;
-            _window.PointerPressed      += Parent_PointerPressEvent;
-            _window.PointerReleased     += Parent_PointerReleaseEvent;
-            _window.PointerWheelChanged += Parent_ScrollEvent;
+            _window.PointerPressed      += Parent_PointerPressedEvent;
+            _window.PointerReleased     += Parent_PointerReleasedEvent;
+            _window.PointerWheelChanged += Parent_PointerWheelChanged;
 
             PressedButtons = new bool[(int)MouseButton.Count];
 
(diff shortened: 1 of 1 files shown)


src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
```diff
@@ -63,16 +63,16 @@ private void Resized(Rect rect)
...
 
-        private void Parent_PointerReleaseEvent(object o, PointerReleasedEventArgs args)
+        private void Parent_PointerReleasedEvent(object o, PointerReleasedEventArgs args)
         {
             if (args.InitialPressMouseButton != Avalonia.Input.MouseButton.None)
             {
-                int button = (int)args.InitialPressMouseButton;
+                int button = (int)args.InitialPressMouseButton - 1;
 
                 if (PressedButtons.Count() >= button)
```
gdkchan, line 77:
```suggestion
                if ((uint)PressedButtons.Length > (uint)button)
```
- It is better to avoid using LINQ if we can, as it can be slow. In this case we can access the array length property directly.
- `>=` is not the right check since the button value being *equal* to the array length would be also out of range. It must be *less* than the array length.
- Casting the value to `uint` and doing a unsigned check also catches cases where `button` value is negative, so I believe that you can remove the `if (args.InitialPressMouseButton != Avalonia.Input.MouseButton.None)` check if you do that.

gdkchan: This check is done in another place, it should be fixed there too.

AcK77: All fixed ;)

src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
```diff
@@ -63,29 +62,25 @@ private void Resized(Rect rect)
...
-            if (args.InitialPressMouseButton != Avalonia.Input.MouseButton.None)
-            {
-                int button = (int)args.InitialPressMouseButton;
+            uint button = (uint)args.InitialPressMouseButton - 1;
 
-                if (PressedButtons.Count() >= button)
-                {
-                    PressedButtons[button] = false;
-                }
+            if ((uint)PressedButtons.Length > button)
```
gdkchan, line 74:
```suggestion
            if ((uint)PressedButtons.Length > (uint)button)
```

src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
```diff
@@ -63,29 +62,25 @@ private void Resized(Rect rect)
...
         }
-
-        private void Parent_PointerPressEvent(object o, PointerPressedEventArgs args)
+        private void Parent_PointerPressedEvent(object o, PointerPressedEventArgs args)
         {
-            int button = (int)args.GetCurrentPoint(_widget).Properties.PointerUpdateKind;
+            uint button = (uint)args.GetCurrentPoint(_widget).Properties.PointerUpdateKind;
 
-            if (PressedButtons.Count() >= button)
+            if ((uint)PressedButtons.Length > button)
```
gdkchan, line 83:
```suggestion
            if ((uint)PressedButtons.Length > (uint)button)
```

gdkchan (approved):
lgtm other than comments.

src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
AcK77: `button` is already casted to `uint`

src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
AcK77: `button` is already casted to `uint`

src/Ryujinx.Ava/Input/AvaloniaMouseDriver.cs
gdkchan: Yea, sorry I didn't notice that the type was already `uint`.

marysaka (approved):

PR #5207, opened 2023-06-02
ava: Fix exit dialog while guest is running.

AcK77:
There is currently an issue while a game runs, the content dialog creation method check if `IsGameRunning` is true to show the popup. But the condition here is wrong (`window` is null) so it throw a NullException silently in `Dispatcher.UIThread`. This is now fixed by using the right casting.

### DIFF
--- a/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
+++ b/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
@@ -318,7 +318,7 @@ namespace Ryujinx.Ava.UI.Helpers
 
             Window parent = GetMainWindow();
 
-            if (parent.IsActive == true && (parent as MainWindow).ViewModel.IsGameRunning)
+            if (parent != null && parent.IsActive && (parent as MainWindow).ViewModel.IsGameRunning)
             {
                 contentDialogOverlayWindow = new()
                 {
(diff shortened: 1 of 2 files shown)


src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
```diff
@@ -318,7 +318,7 @@ public static async Task<ContentDialogResult> ShowAsync(ContentDialog contentDia
 
             Window parent = GetMainWindow();
 
-            if (parent is { IsActive: true } and MainWindow window && window.ViewModel.IsGameRunning)
+            if (parent != null && parent.IsActive && (parent as MainWindow).ViewModel.IsGameRunning)
```
gdkchan, line 321: I'm not sure if assuming that `parent` is `MainWindow` is right here. The previous check would check if the `parent` *is* of type `MainWindow`.

gdkchan: personally I would do
```cs
            if (parent is MainWindow mainWindow && mainWindow.IsActive && mainWindow.ViewModel.IsGameRunning)

gdkchan: I  spoke with Ac_K and apparently this suggestion doesn't work, and I don't want to debug that right now, so I will mark as resolved.

gdkchan (approved):
lgtm, thanks.

src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
AcK77: More info, there is an issue with the `is` operator but I don't get why... 
If we use it, `parent` is null (and then `mainWindow`). If we remove the `is` everything works as expected.

riperiperi (approved):

PR #5210, opened 2023-06-03
Replacing ZbcColorArray with Array4<uint>

marco-carvalho:


gdkchan:
Nice catch. This code can be improved further:
- Private fields names should start with underscore, so it would be `_element0` for example, not `element0`.
- This could use a switch expression which would be more compact than the `if`/`else if`/`else` chain.

gdkchan:
Actually, I'd just delete the `ZbcColorArray` type and replace it with `Array4<uint>` from `Ryujinx.Common.Memory`.

marco-carvalho:
I'm sorry, but I'm not really familiar with the code base for me to make those changes.

gdkchan:
You just need to replace `ZbcColorArray` with `Array4<uint>` and then delete the `ZbcColorArray` struct.

marco-carvalho:
@gdkchan I just did what you said

gdkchan:
Can you adjust the struct alignment? After that I will approve.

marco-carvalho:
@gdkchan Could you check again?

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

Littl3Guy:
This PR broke grass/tree/bush animations in Pocket Monsters S/V.

gdkchan:
> This PR broke grass/tree/bush animations in Pocket Monsters S/V.
> 
> Edit: The animations start, but don't loop for whatever reason.

It was not caused by this change, it was the one merged right before this one, #5220.

Littl3Guy:
Yup sorry about that, the changelog on github looks a little confusing:
image

 I thought the version number was at the bottom of the changelog, not the top.

gdkchan:
It's fine, you helped narrow it down, I didn't expect this change would cause it so I knew it was probably the one before or after it.

PR #5212, opened 2023-06-03
Blocks should be synchronized on read-only fields

marco-carvalho:
Let's take a look at the example WITHOUT `readonly`:

```cs
public class ExampleClass
{
    private object _lock = new object();

    public void DoSomething()
    {
        lock(_lock)
        {
            // ...
        }
    }
}
```

The `_lock` object is not `readonly`, which means it can be modified elsewhere in the class. This is potentially dangerous because if `_lock` is accidentally reassigned somewhere else in the `ExampleClass`, it would mean that the `lock` statement would be using a different lock object. This could lead to race conditions because threads could potentially enter the `lock` statement concurrently, which would defeat the whole purpose of using `lock`.

On the other hand, in the example WITH `readonly`:

```cs
public class ExampleClass
{
    private readonly object _lock = new object();

    public void DoSomething()
    {
        lock(_lock)
        {
            // ...
        }
    }
}
```

The `_lock` object is declared as `readonly`. This means once it is assigned, it cannot be changed. This ensures that the same lock object is always used for the `lock` statement, which properly synchronizes access to the block of code and ensures thread safety.

Making the `_lock` object `readonly` is a good practice because it protects the lock object from being accidentally reassigned, which could lead to bugs that are hard to detect and diagnose.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
+++ b/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
@@ -27,7 +27,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
 
         private TRef<bool> _cancelled = null;
         private Thread     _thread    = null;
-        private object     _lock      = new object();
+        private readonly object _lock = new object();
 
         public bool IsRunning
         {
--- a/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
+++ b/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
@@ -26,7 +26,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
         }
 
         private TRef<bool> _cancelled = null;
-        private Thread     _thread    = null;
+        private Thread _thread = null;
         private readonly object _lock = new object();
 
         public bool IsRunning
--- a/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
+++ b/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
@@ -27,7 +27,7 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
 
         private TRef<bool> _cancelled = null;
         private Thread _thread = null;
-        private readonly object _lock = new object();
+        private readonly object _lock = new();
 
         public bool IsRunning
         {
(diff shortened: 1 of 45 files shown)


src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
```diff
@@ -27,7 +27,7 @@ public SleepSubstepData(int sleepMilliseconds)
 
         private TRef<bool> _cancelled = null;
         private Thread     _thread    = null;
-        private object     _lock      = new object();
+        private readonly object _lock = new object();
```
gdkchan, line 30: This is misaligned now. I'd prefer if you remove alignment from all those fields.

src/Ryujinx.HLE/HOS/Services/Ssl/BuiltInCertificateManager.cs
```diff
@@ -46,7 +46,7 @@ public static BuiltInCertificateManager Instance
         private bool                _initialized;
```
gdkchan, line 46: Might as well fix this that is misaligned while you are at it.

src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/TimedAction.cs
marco-carvalho: committed

src/Ryujinx.HLE/HOS/Services/Ssl/BuiltInCertificateManager.cs
marco-carvalho: committed

gdkchan (approved):
lgtm, thanks.

src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ISelfController.cs
```diff
@@ -17,7 +17,7 @@ class ISelfController : IpcService
         private KEvent _accumulatedSuspendedTickChangedEvent;
         private int    _accumulatedSuspendedTickChangedEventHandle;
 
-        private object _fatalSectionLock = new object();
+        private readonly object _fatalSectionLock = new object();
```
riperiperi, line 20: The alignment of the next line mismatches now.

riperiperi (approved):
Makes sense.

src/Ryujinx.HLE/HOS/Services/Am/AppletAE/AllSystemAppletProxiesService/SystemAppletProxy/ISelfController.cs
marco-carvalho: could you check again, please?

riperiperi (approved):

AcK77 (changes requested):
Could be nice if you could change ` = new object();` with ` = new();`. Thanks.

marco-carvalho:
> Could be nice if you could change ` = new object();` with ` = new();`. Thanks.

@AcK77 could you check again, please?

TSRBerry:
image
You should be able to re-request a review by clicking this button.

PR #5216, opened 2023-06-04
Ava: Fix OpenGL on Linux again

### LINKED ISSUE #4600: [Bug] OpenGL crashes everytime on Avalonia under Linux
### Description of the issue

There is a regression with 1.1.563 with OpenGL that makes it crash everytime on Avalonia. I have currently only tested with Mario Kart 8 Deluxe and Super Mario Odyssey, this might happen with other games too but they are the only ones i currently have dumped..

### Reproduction steps

Set graphics backend to OpenGL and launch game

### Log file

[Ryujinx_1.1.563_2023-03-25_08-34-43.log](https://github.com/Ryujinx/Ryujinx/files/11068306/Ryujinx_1.1.563_2023-03-25_08-34-43.log)

### OS

Linux Ubuntu

### Ryujinx version

1.1.563

### Game version

_No response_

### CPU

Intel® Core™ i5-7300HQ 

### GPU

GeForce GTX 1060 Mobile

### RAM

8GB

### List of applied mods

_No response_

### Additional context?

_No response_

TSRBerry:
I have absolutely no clue why this works.

The error produced by X11 looks like that: `XError: X_Error result is BadMatch (invalid parameter attributes) (code: 8)`
(You can get the error by applying https://github.com/marysaka/SPB/pull/4 to SPB.)

As far as I understood, the issue was related to one of the `FramebufferFormat` parameters being incorrect.
I was about to test different values for `depthBits` (as this [answer](https://stackoverflow.com/a/17127200) suggested), but tried the empty constructor once just to see what would happen. To my surprise it just started to work, so here we are.

---

Fixes #4600

### DIFF
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -40,6 +40,7 @@ using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.Formats.Png;
 using SixLabors.ImageSharp.PixelFormats;
 using SixLabors.ImageSharp.Processing;
+using SPB.Graphics.Exceptions;
 using SPB.Graphics.Vulkan;
 using System;
 using System.Collections.Generic;
@@ -475,11 +476,34 @@ namespace Ryujinx.Ava
                 _windowsMultimediaTimerResolution = null;
             }
 
-            (_rendererHost.EmbeddedWindow as EmbeddedWindowOpenGL)?.MakeCurrent();
+            if (_rendererHost.EmbeddedWindow is EmbeddedWindowOpenGL openGlWindow)
+            {
+                // Try to bind the OpenGL context before calling the shutdown event
+                try
+                {
+                    openGlWindow.MakeCurrent();
+                }
+                catch (ContextException e)
+                {
+                    Logger.Warning?.Print(LogClass.Ui, $"Failed to bind OpenGL context: {e}");
+                }
 
-            Device.DisposeGpu();
+                Device.DisposeGpu();
 
-            (_rendererHost.EmbeddedWindow as EmbeddedWindowOpenGL)?.MakeCurrent(null);
+                // Unbind context and destroy everything
+                try
+                {
+                    openGlWindow.MakeCurrent(null);
+                }
+                catch (ContextException e)
+                {
+                    Logger.Warning?.Print(LogClass.Ui, $"Failed to unbind OpenGL context: {e}");
+                }
+            }
+            else
+            {
+                Device.DisposeGpu();
+            }
         }
 
         private void HideCursorState_Changed(object sender, ReactiveEventArgs<HideCursorMode> state)
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -478,27 +478,13 @@ namespace Ryujinx.Ava
 
             if (_rendererHost.EmbeddedWindow is EmbeddedWindowOpenGL openGlWindow)
             {
-                // Try to bind the OpenGL context before calling the shutdown event
-                try
-                {
-                    openGlWindow.MakeCurrent();
-                }
-                catch (ContextException e)
-                {
-                    Logger.Warning?.Print(LogClass.Ui, $"Failed to bind OpenGL context: {e}");
-                }
+                // Try to bind the OpenGL context before calling the shutdown event.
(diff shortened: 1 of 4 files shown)


src/Ryujinx.Ava/UI/Renderer/EmbeddedWindow.cs
```diff
@@ -123,7 +123,7 @@ private IPlatformHandle CreateLinux(IPlatformHandle control)
             }
             else
             {
-                X11Window = PlatformHelper.CreateOpenGLWindow(FramebufferFormat.Default, 0, 0, 100, 100) as GLXWindow;
```
marysaka, line 126: This is wrong, the right fix would be to apply the same logic as in EmbeddedWindowOpenGL [here](https://github.com/Ryujinx/Ryujinx/blob/68848000f77e587ed4cd99e84a234fcf79fd7ea3/src/Ryujinx.Ava/UI/Renderer/EmbeddedWindowOpenGL.cs#L53).

marysaka (changes requested):

src/Ryujinx.Ava/UI/Renderer/EmbeddedWindow.cs
TSRBerry: I see, thank you! I had no idea why it worked like that and was confused why this was needed twice as well.

marysaka: I really do not understand too!

all OpenGL related code should be in EmbeddedWindowOpenGL not EmbeddedWindow imo.

TSRBerry:
Marked as draft, because I just noticed I'm crashing when stopping emulation now.

I think I'm crashing with the same XError, but don't have my SPB fork active right now, so I don't know for sure.

https://github.com/Ryujinx/Ryujinx/blob/1874305040ed4e93cd77f771cdaae99f735f8375/src/Ryujinx.Ava/AppHost.cs#L478

I'll check this later again and hope to fix this completely somehow.

TSRBerry:
I basically copied the logic from GTK and added logging messages to prevent Ava from crashing when stopping emulation.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -475,11 +476,34 @@ public void DisposeGpu()
...
+            {
+                // Try to bind the OpenGL context before calling the shutdown event
+                try
+                {
+                    openGlWindow.MakeCurrent();
+                }
+                catch (ContextException e)
+                {
+                    Logger.Warning?.Print(LogClass.Ui, $"Failed to bind OpenGL context: {e}");
+                }
```
AcK77, line 489: Maybe the `try`/`catch` should be added in `EmbeddedWindowOpenGL.MakeCurrent()` directly, to avoid duplicated code and put everything at the same place ?

TSRBerry: It has to succeed initially when the window is set up.

riperiperi (commented):
Should probably be tested on AMD. Does this same thing cause issues on GTK? (since it creates the background context this way)

riperiperi:
It does look like both Ava and GTK create their main context with these parameters, so the change to Ava will probably work on AMD. It's weird that GTK was fine though, as the background context is created in a similar way.

marysaka:
> It does look like both Ava and GTK create their main context with these parameters, so the change to Ava will probably work on AMD. It's weird that GTK was fine though, as the background context is created in a similar way.

I think the main difference is that the window is provided by GTK (and is setup that way?)

So to me it's a mismatch between the surface requested when creating the window and the one requested for the OpenGL context.

I think asking for default framebuffer format on context creation might work also, but I do not want to have a divergence between GTK and Avalonia here that might cause issues.

src/Ryujinx.Ava/AppHost.cs
marysaka: > Maybe the `try`/`catch` should be added in `EmbeddedWindowOpenGL.MakeCurrent()` directly, to avoid duplicated code and put everything at the same place ?

Depending of the environment, those might fail here because for example the context is already current or already unbind.

In other cases we want the exception to be throw.

marysaka (approved):

AcK77 (approved):

riperiperi (approved):
Tested on AMD+SteamOS and looks fine. Nice work!

PR #5218, opened 2023-06-04
Avalonia: Adjust Grid Library alignment

tarrouye:
Personally I found the center alignment for rows that are not full to look off. I updated the flexbox constraints so empty rows would left-align and the items would always line up with the rows above. 

**This is a subjective change, so I'm happy to make it an option in the settings somewhere if people would prefer.** 

Some examples of the difference: 

| Before                                                                                           | After                                                                                            |
|--------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| image | image |
| image | image |

### DIFF
--- a/src/Ryujinx.Ava/UI/Controls/ApplicationGridView.axaml
+++ b/src/Ryujinx.Ava/UI/Controls/ApplicationGridView.axaml
@@ -32,10 +32,10 @@
             <ListBox.ItemsPanel>
                 <ItemsPanelTemplate>
                     <flex:FlexPanel
-                        HorizontalAlignment="Stretch"
+                        HorizontalAlignment="Center"
                         VerticalAlignment="Stretch"
                         AlignContent="FlexStart"
-                        JustifyContent="Center" />
+                        JustifyContent="FlexStart" />
                 </ItemsPanelTemplate>
             </ListBox.ItemsPanel>
             <ListBox.Styles>


MutantAura:
Personally I think almost any UI change like this going forward should really be behind a customisation page in the settings window. The issue is that at the moment our "user interface" tab is quite badly designed for its named purpose and would also need a re-think. For now I wonder if a simpler solution would be a right-click context menu on a blank space click with those two options: Center vs Flex.

tarrouye:
@MutantAura Yeah like I said, I'm happy to make it an option somewhere. Open to any suggestions about where to put it ! 

There are some other UI options that I personally would like to have in Avalonia, such as removing the border on the grid tiles, etc. Probably those would all go together. 

The 'User Interface' settings currently is a little odd like you mention. Theme is really the only thing there that feels like it applies to the UI. 

How would people feel about breaking the General & Game Directories parts into a 'General' tab, leaving the Theme & any future options under the 'User Interface' tab ? Or, alternatively, renaming the tab to 'General' and making 'User Interface' a subsection under it like how Theme currently is ?

AcK77 (approved):
IMO this kind of change doesn’t need a setting.
Since it’s a grid, it’s looks better to get things aligned as it’s done in this PR instead of our actual code.

gdkchan:
I don't mind either of them. I think it being left aligned (this change) makes a bit more sense since it actually looks like a grid at the bottom, it's a little bit weird to have the last row centered.

If that must be a setting, I'd put it at the top bar where all the game list related settings are. There are currently 2 buttons: One for list and another for grid. It could be changed to 3: list, grid left aligned, grid centered. Or you could add 2 more buttons for left and centered. I guess 99% of users won't care about this, so not sure if it's worth a setting just for that.

gdkchan (approved):

tarrouye:
Personally, I'd rather not add a setting for it until there are more niche UI settings available in the main settings menu. 
I don't really want to clutter the library menu bar with such a minor setting. 


I'll leave it up to y'all whether this gets merged as is, or if you want me to add a toggle somewhere.

AcK77:
> I guess 99% of users won't care about this, so not sure if it's worth a setting just for that.

I guess the same, so I merge it!

PR #5220, opened 2023-06-04
Texture: Fix layout conversion when gobs in z is used with depth = 1

riperiperi:
The size calculator methods deliberately reduce the gob size of textures if they are deemed too small for it. This is required to get correct sizes when iterating mip levels of a texture.

Rendering to a slice of a 3D texture can produce a 3D texture with depth 1, but a gob size matching a much larger texture. We _can't_ "correct" this gob size, as it is intended as a slice of a larger 3D texture. Ignoring it causes layout conversion to break on read and flush.

This caused an issue in Tears of the Kingdom where the compressed 3D texture used for the gloom would always break on OpenGL, and seemingly randomly break on Vulkan. In the first case, the data is forcibly flushed to decompress the BC4 texture on the CPU to upload it as 3D, which was broken due to the incorrect layout. In the second, the data may be randomly flushed if it falls out of the cache, but it will appear correct if it's able to form copy dependencies.

This change only allows gob sizes to be reduced once per mip level. For the purpose of aligned size, it can still be reduced infinitely as our texture cache isn't properly able to handle a view being _misaligned_.

The SizeCalculator has also been changed to reduce the size of rendered depth slices to only include the exact range a single depth slice will cover. (before, the size was way too small with gobs in z reduced to 1, and too large when using the correct value)

Gobs in Y logic remains untouched, we don't support Y slices of textures so it's fine as is.

This is probably worth testing in a few games as it also affects texture size and view logic. I've tested a few myself and they seem OK.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCompatibility.cs
@@ -541,7 +541,8 @@ namespace Ryujinx.Graphics.Gpu.Image
                     depth,
                     lhs.FormatInfo.BlockHeight,
                     lhs.GobBlocksInY,
-                    lhs.GobBlocksInZ);
+                    lhs.GobBlocksInZ,
+                    level);
 
                 return gobBlocksInY == rhs.GobBlocksInY &&
                        gobBlocksInZ == rhs.GobBlocksInZ;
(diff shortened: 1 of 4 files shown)


riperiperi:
Before (pending):

(insert image)

This happens all the time on OGL, and sometimes on VK.

After:
image

gdkchan:
Fixes character rendering on Spiritfarer.
master:
image
PR:
image

gdkchan (approved):
lgtm, thanks! Nice find. That makes sense, the only way for the case where the full depth is unknown to work is the GPU using the unmodified GOB size.

AcK77 (approved):

PR #5227, opened 2023-06-04
Duplicate casts should not be made

marco-carvalho:
Because the `is` operator performs a cast if the object is not null, using `is` to check type and then casting the same argument to that type, necessarily performs two casts.

### DIFF
--- a/src/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
+++ b/src/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
@@ -53,10 +53,8 @@ namespace ARMeilleure.Instructions
 
         public static void EmitVStoreOrLoadN(ArmEmitterContext context, int count, bool load)
         {
-            if (context.CurrOp is OpCode32SimdMemSingle)
+            if (context.CurrOp is OpCode32SimdMemSingle op)
             {
-                OpCode32SimdMemSingle op = (OpCode32SimdMemSingle)context.CurrOp;
-
                 int eBytes = 1 << op.Size;
 
                 Operand n = context.Copy(GetIntA32(context, op.Rn));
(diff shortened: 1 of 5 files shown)


TSRBerry:
Part of #4670

src/ARMeilleure/Instructions/InstEmitSimdMemory32.cs
```diff
@@ -82,73 +80,71 @@ public static void EmitVStoreOrLoadN(ArmEmitterContext context, int count, bool
...
                     }
                     else
                     {
-                        SetIntA32(context, op.Rn, context.Add(n, Const(count * eBytes)));
+                        SetIntA32(context, opSingle.Rn, context.Add(n, Const(count * eBytes)));
                     }
                 }
             }
-            else
+            else if (context.CurrOp is OpCode32SimdMemPair opPair)
```
AcK77, line 128: What if `context.CurrOp` isn't `OpCode32SimdMemSingle` or `OpCode32SimdMemPair` ?

marco-carvalho: I think it would throw an exception because the previous code looked like this:
```cs
            else
            {
                OpCode32SimdMemPair op = (OpCode32SimdMemPair)context.CurrOp;
```

So I assumed that it could only flow inside `else` if this check was `true`.

TSRBerry:
I don't think the benchmark code is valid here, since you are missing the assignment in the second test. This might lead to incorrect and misleading results.

marco-carvalho:
> I don't think the benchmark code is valid here, since you are missing the assignment in the second test. This might lead to incorrect and misleading results.

You're right, there is no performance improvement in this change.

PR #5229, opened 2023-06-05
Per-User Profile Metadata

tarrouye:
Logically, data such as whether an app is a Favorite or not, when it was last played, and how long it was played should all belong to each user profile, not Ryujinx as a whole. 

For time played, this better reflects the behavior of the Switch console itself.
As far as I know, the Switch doesn't have Favorites so there is no precedent for that, but IMO it makes more sense as a per-user setting. 

This PR:
- Adjusts this information to be tracked per user profile. 
- Handles the migration from the old system
  - On first boot, it will migrate the previous global data to the last opened user profile. (Thoughts welcome on a better migration strategy)
- Handles refreshing the application library when the user profile changes to reflect that user's metadata
  - Implemented for GTK and Avalonia

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1357,7 +1357,14 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
+            
+            if (prevUserId != AccountManager.LastOpenedUser.UserId)
+            {
+                // current user changed, so refresh application metadata
+                LoadApplications();
+            }
         }
 
         public void SimulateWakeUpMessage()
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1523,7 +1523,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         public void UpdateGameMetadata(string titleId)
         {
-            ApplicationLibrary.LoadAndSaveMetaData(titleId, appMetadata =>
+            ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), titleId, appMetadata =>
             {
                 if (appMetadata.LastPlayed.HasValue)
                 {
--- a/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -521,12 +521,20 @@ namespace Ryujinx.Ui.App.Common
 
             // Handle migration from old default to current user
             string legacyFile = Path.Combine(guiFolder, "metadata.json");
-            if (File.Exists(legacyFile) && !File.Exists(metadataFile))
+            if (File.Exists(legacyFile))
             {
-                File.Move(legacyFile, metadataFile);
-                File.Delete(legacyFile);
+                if (!File.Exists(metadataFile))
+                {
+                    File.Move(legacyFile, metadataFile);
+                }
+                
+                if (File.Exists(metadataFile))
+                {
+                    File.Delete(legacyFile);
+                }
             }
 
+            // Create metadata file if it doesn't exist yet
             ApplicationMetadata appMetadata;
 
             if (!File.Exists(metadataFile))
@@ -536,6 +544,7 @@ namespace Ryujinx.Ui.App.Common
                 JsonHelper.SerializeToFile(metadataFile, appMetadata, SerializerContext.ApplicationMetadata);
             }
 
+            // Read from the metadata file
             try
             {
                 appMetadata = JsonHelper.DeserializeFromFile(metadataFile, SerializerContext.ApplicationMetadata);
@@ -547,6 +556,7 @@ namespace Ryujinx.Ui.App.Common
                 appMetadata = new ApplicationMetadata();
             }
 
+            // Modify the metadata and save it back to the file
             if (modifyFunction != null)
             {
                 modifyFunction(appMetadata);
--- a/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
+++ b/src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
@@ -81,7 +81,7 @@ namespace Ryujinx.Ui.App.Common
             controlFile.Get.Read(out _, 0, outProperty, ReadOption.None).ThrowIfFailure();
         }
 
-        public void LoadApplications(List<string> appDirs, Language desiredTitleLanguage)
+        public void LoadApplications(UserId userId, List<string> appDirs, Language desiredTitleLanguage)
         {
             int numApplicationsFound  = 0;
             int numApplicationsLoaded = 0;
@@ -418,7 +418,7 @@ namespace Ryujinx.Ui.App.Common
                         continue;
                     }
 
-                    ApplicationMetadata appMetadata = LoadAndSaveMetaData(titleId, appMetadata =>
+                    ApplicationMetadata appMetadata = LoadAndSaveMetaData(userId, titleId, appMetadata =>
                     {
                         appMetadata.Title = titleName;
 
@@ -508,17 +508,29 @@ namespace Ryujinx.Ui.App.Common
             titleId   = controlNca?.Header.TitleId.ToString("x16");
         }
 
-        public ApplicationMetadata LoadAndSaveMetaData(string titleId, Action<ApplicationMetadata> modifyFunction = null)
+        public ApplicationMetadata LoadAndSaveMetaData(UserId userId, string titleId, Action<ApplicationMetadata> modifyFunction = null)
         {
-            string metadataFolder = Path.Combine(AppDataManager.GamesDirPath, titleId, "gui");
+            string guiFolder      = Path.Combine(AppDataManager.GamesDirPath, titleId, "gui");
+            string metadataFolder = Path.Combine(guiFolder, userId.ToString());
             string metadataFile   = Path.Combine(metadataFolder, "metadata.json");
+
+            if (!Directory.Exists(metadataFolder))
+            {
+                Directory.CreateDirectory(metadataFolder);
+            }
+
+            // Handle migration from old default to current user
+            string legacyFile = Path.Combine(guiFolder, "metadata.json");
+            if (File.Exists(legacyFile) && !File.Exists(metadataFile))
+            {
+                File.Move(legacyFile, metadataFile);
+                File.Delete(legacyFile);
+            }
 
             ApplicationMetadata appMetadata;
 
             if (!File.Exists(metadataFile))
             {
-                Directory.CreateDirectory(metadataFolder);
-
                 appMetadata = new ApplicationMetadata();
 
                 JsonHelper.SerializeToFile(metadataFile, appMetadata, SerializerContext.ApplicationMetadata);
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -682,7 +682,7 @@ namespace Ryujinx.Ava
 
             DiscordIntegrationModule.SwitchToPlayingState(Device.Processes.ActiveApplication.ProgramIdText, Device.Processes.ActiveApplication.Name);
 
-            _viewModel.ApplicationLibrary.LoadAndSaveMetaData(Device.Processes.ActiveApplication.ProgramIdText, appMetadata =>
+            _viewModel.ApplicationLibrary.LoadAndSaveMetaData(_accountManager.LastOpenedUser.UserId.ToLibHacFsUid(), Device.Processes.ActiveApplication.ProgramIdText, appMetadata =>
             {
                 appMetadata.LastPlayed = DateTime.UtcNow;
             });
--- a/src/Ryujinx.Ava/UI/Controls/ApplicationContextMenu.axaml.cs
+++ b/src/Ryujinx.Ava/UI/Controls/ApplicationContextMenu.axaml.cs
@@ -42,7 +42,7 @@ namespace Ryujinx.Ava.UI.Controls
             {
                 viewModel.SelectedApplication.Favorite = !viewModel.SelectedApplication.Favorite;
 
-                viewModel.ApplicationLibrary.LoadAndSaveMetaData(viewModel.SelectedApplication.TitleId, appMetadata =>
+                viewModel.ApplicationLibrary.LoadAndSaveMetaData(viewModel.AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), viewModel.SelectedApplication.TitleId, appMetadata =>
                 {
                     appMetadata.Favorite = viewModel.SelectedApplication.Favorite;
                 });
--- a/src/Ryujinx.Ava/UI/Models/SaveModel.cs
+++ b/src/Ryujinx.Ava/UI/Models/SaveModel.cs
@@ -74,7 +74,7 @@ namespace Ryujinx.Ava.UI.Models
             }
             else
             {
-                var appMetadata = MainWindow.MainWindowViewModel.ApplicationLibrary.LoadAndSaveMetaData(TitleIdString);
+                var appMetadata = MainWindow.MainWindowViewModel.ApplicationLibrary.LoadAndSaveMetaData(UserId, TitleIdString);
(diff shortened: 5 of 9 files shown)


src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,7 +1357,14 @@ public void ToggleFileType(string fileType)
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
+            
+            if (prevUserId != AccountManager.LastOpenedUser.UserId)
+            {
+                // current user changed, so refresh application metadata
+                LoadApplications();
```
patrick-hovsepian, line 1366: This is going to re-read application data from the cart files which seems inefficient

tarrouye: This is true. I've updated it to refresh just the metadata.

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -508,22 +508,43 @@ private void GetControlFsAndTitleId(PartitionFileSystem pfs, out IFileSystem con
...
+                if (!File.Exists(metadataFile))
+                {
+                    File.Move(legacyFile, metadataFile);
+                }
+                
+                if (File.Exists(metadataFile))
+                {
+                    File.Delete(legacyFile);
+                }
+            }
```
marysaka, line 535: Simplify this

```suggestion
            // Handle migration from old default to current user
            string legacyFile = Path.Combine(guiFolder, "metadata.json");
            if (File.Exists(legacyFile) && !File.Exists(metadataFile))
            {
                File.Move(legacyFile, metadataFile);
                File.Delete(legacyFile);
            }
```

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -508,22 +508,43 @@ private void GetControlFsAndTitleId(PartitionFileSystem pfs, out IFileSystem con
...
         {
-            string metadataFolder = Path.Combine(AppDataManager.GamesDirPath, titleId, "gui");
+            string guiFolder      = Path.Combine(AppDataManager.GamesDirPath, titleId, "gui");
+            string metadataFolder = Path.Combine(guiFolder, userId.ToString());
             string metadataFile   = Path.Combine(metadataFolder, "metadata.json");
 
+            if (!Directory.Exists(metadataFolder))
+            {
+                Directory.CreateDirectory(metadataFolder);
+            }
```
marysaka, line 520: As per [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-7.0):
`Any and all directories specified in path are created, unless they already exist or unless some part of path is invalid. If the directory already exists, this method does not create a new directory, but it returns a DirectoryInfo object for the existing directory.`

```suggestion
            Directory.CreateDirectory(metadataFolder);
```

marysaka (approved):
Some stuffs but otherwise seems okay to me.

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
tarrouye: Hi, it was like this originally, then I separated the checks in order to handle a specific edge-case (if user boots old version of Ryujinx, it would create legacy file for every game)

I can change it back if we are ok not handling that edge-case !

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
tarrouye: Thank you ! Not a .NET expert for sure 😅  I will take this part in mind.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,37 @@ public void ToggleFileType(string fileType)
...
         }
 
         public void SimulateWakeUpMessage()
         {
             AppHost.Device.System.SimulateWakeUpMessage();
         }
 
+        private async void RefreshApplicationsMetadata()
+        {
+            await Task.Run(() => {
```
patrick-hovsepian, line 1377: you don't need to await this, it's going to block until that operation completes so it's equivalent to running the foreach without wrapping it in the Task.Run.

If the idea is to do this in the background, change the await to a discard `_ = Task.Run(..`

I'd also argue that this should just be the foreach code and the callers can decide if they want to wait for it to complete or not. so, I'd move this up to where you call RefreshApplicationMetadata
`_ = Task.Run(() => RefreshApplicationMetadata());

that way, the user can continue to interact with the UI while the metadata is refreshed in the background.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,37 @@ public void ToggleFileType(string fileType)
...
+                {
+                    var metadata = ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), app.TitleId);
+                    app.Favorite = metadata.Favorite;
+                    app.LastPlayed = metadata.LastPlayed;
+                    app.TimePlayedNum = metadata.TimePlayed;
+                    app.TimePlayed = ApplicationLibrary.ConvertSecondsToFormattedString(metadata.TimePlayed);
+                }
+            });
+
+            RefreshView();
```
patrick-hovsepian, line 1388: This is going to re-paint the entire screen, can you try to see if you can avoid that?

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,37 @@ public void ToggleFileType(string fileType)
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
+            
+            if (prevUserId != AccountManager.LastOpenedUser.UserId)
+            {
+                // current user changed, so refresh application metadata
+                RefreshApplicationsMetadata();
```
patrick-hovsepian, line 1366: will this reflect the correct save data directory for the user? I thought that currently came from the control header struct which is populated by reading the rom data. It'd be nice to remove the dependency because you can find the dir with just uid and titleid but not sure if you want to tackle that here or in a different change.

tarrouye: This has nothing to do with the changes made here, so I'd probably want to handle that part in a different PR. Prior to this change, absolutely nothing happened here when changing user.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
tarrouye: The same happens when you toggle favorite on any application. I am just following the existing structure

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
tarrouye: Yeah, probably moving the Task.Run out to the caller makes sense.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
tarrouye: Also, the user save directories does not come from the control header. The directory is resolved at runtime using the LastOpenedUser. You can try it yourself, switching users does reflect the correct save data directory - on master and on this branch.

patrick-hovsepian: Seems like a bug, at best.

tarrouye: What seems like a bug ?

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
patrick-hovsepian: I understand it happens today, doesn't mean it cant be improved.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
patrick-hovsepian: I missed your last comment, if it's resolved at runtime then it's fine.

patrick-hovsepian: well the control structure would still reference a different user? is it ever reset in the application datA?

tarrouye: Again, this doesn't relate to the changes made here. What happened before is what happens now, in terms of the control structure. If you think there is a bug with how that part is handled, then probably you should open a separate issue for it :) 

I'm not sure which part of the ApplicationControlProperty you think is even affected by the selected user. I just took a brief look and to me it seems like none of it is.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
+            
+            if (prevUserId != AccountManager.LastOpenedUser.UserId)
+            {
+                // current user changed, so refresh application metadata
+                _ = Task.Run(() => RefreshApplicationsMetadata());
```
patrick-hovsepian, line 1366: you might be able to simplify further with `_ = Task.Run(RefreshApplicationMetadata)`

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
patrick-hovsepian: I was just asking 😄 if you traced it and can confirm it's all kosher then all good thanks for looking into it.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
...
+            }
         }
 
         public void SimulateWakeUpMessage()
         {
             AppHost.Device.System.SimulateWakeUpMessage();
         }
 
+        private async void RefreshApplicationsMetadata()
```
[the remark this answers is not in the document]
tarrouye: Thanks !

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
tarrouye: Neat; Thanks !

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
...
 
+        private void RefreshApplicationsMetadata()
+        {
+            foreach (var app in _applications)
+            {
+                var metadata = ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), app.TitleId);
+                app.Favorite = metadata.Favorite;
+                app.LastPlayed = metadata.LastPlayed;
+                app.TimePlayedNum = metadata.TimePlayed;
+                app.TimePlayed = ApplicationLibrary.ConvertSecondsToFormattedString(metadata.TimePlayed);
```
patrick-hovsepian, line 1383: if you wanted to change the `ApplicationData`  (I think that's the type of app) to a record type, you could use `with` syntax. (I'm not suggesting you change it but just sharing for awareness) 

```
app = app with {
Favorite = metadata.Favorite,
LastPlayed = metadata.LastPlayed, 
...
}
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
...
         {
             AppHost.Device.System.SimulateWakeUpMessage();
         }
 
+        private void RefreshApplicationsMetadata()
+        {
+            foreach (var app in _applications)
+            {
+                var metadata = ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), app.TitleId);
+                app.Favorite = metadata.Favorite;
```
patrick-hovsepian, line 1380: super nit: newline before this for readability

patrick-hovsepian (approved):
lgtm, thanks for digging through the control structure

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
tarrouye: Are there any implications of making it a record type ? I need to look at the docs 😅

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
```
AcK77, line 1361:
```suggestion

            await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
 
         public async void ManageProfiles()
         {
+            var prevUserId = AccountManager.LastOpenedUser.UserId;
```
AcK77, line 1360:
```suggestion
            var oldLastOpenedUserId = AccountManager.LastOpenedUser.UserId;
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
...
             await NavigationDialogHost.Show(AccountManager, ContentManager, VirtualFileSystem, LibHacHorizonManager.RyujinxClient);
+            
+            if (prevUserId != AccountManager.LastOpenedUser.UserId)
```
AcK77, line 1363:
```suggestion
            if (oldLastOpenedUserId != AccountManager.LastOpenedUser.UserId)
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1357,14 +1357,35 @@ public void ToggleFileType(string fileType)
...
         public void SimulateWakeUpMessage()
         {
             AppHost.Device.System.SimulateWakeUpMessage();
         }
 
+        private void RefreshApplicationsMetadata()
+        {
+            foreach (var app in _applications)
+            {
+                var metadata = ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), app.TitleId);
```
AcK77, line 1379:
```suggestion
                var metadata = ApplicationLibrary.LoadAndSaveMetaData(AccountManager.LastOpenedUser.UserId.ToLibHacFsUid(), app.TitleId);
                
```

src/Ryujinx.Ui.Common/App/ApplicationLibrary.cs
```diff
@@ -508,22 +508,32 @@ private void GetControlFsAndTitleId(PartitionFileSystem pfs, out IFileSystem con
...
+            Directory.CreateDirectory(metadataFolder);
+
+            // Handle migration from old default to current user
+            string legacyFile = Path.Combine(guiFolder, "metadata.json");
+            if (File.Exists(legacyFile) && !File.Exists(metadataFile))
+            {
+                File.Move(legacyFile, metadataFile);
+            }
+
+            // Create metadata file if it doesn't exist yet
```
AcK77, line 526:
```suggestion
            // Create metadata file if it doesn't exist yet.
```

AcK77 (changes requested):

PR #5230, opened 2023-06-05
Inheritance should not be redundant

marco-carvalho:
An inheritance list entry is redundant if:

- It is `Object` - all classes extend `Object` implicitly.
- It is `int` for an `enum`
- It is a base class of another listed inheritance.

Such redundant declarations should be removed because they needlessly clutter the code and can be confusing.

### DIFF
--- a/src/ARMeilleure/Decoders/OpCode32SimdSel.cs
+++ b/src/ARMeilleure/Decoders/OpCode32SimdSel.cs
@@ -13,7 +13,7 @@
         }
     }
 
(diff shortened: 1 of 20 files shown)


marysaka (approved):
Seems reasonable, lgtm

gdkchan (approved):
lgtm, thanks.

Thealexbarney:
You could argue that the `int`s _should_ be explicit if they're used for interop with stuff like data structures

PR #5231, opened 2023-06-05
Prefer a 'TryGetValue' call over a Dictionary indexer access guarded by a 'ContainsKey'

marco-carvalho:
Benchmark:
```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    private Dictionary<int, string> dict = new Dictionary<int, string>();
    private int size = 1000;
    private int keyToFind = 999;

    [GlobalSetup]
    public void Setup()
    {
        for (int i = 0; i < size; i++)
        {
            dict.Add(i, "aaa");
        }
    }

    [Benchmark]
    public void ContainsKeyPlusArrayLookup()
    {
        if (dict.ContainsKey(keyToFind))
        {
            var response = dict[keyToFind];
        }
    }

    [Benchmark]
    public void TryGetValue()
    {
        dict.TryGetValue(keyToFind, out var response);
    }
}
```

Results:
```sh
|                     Method |      Mean |     Error |    StdDev |    Median | Rank | Allocated |
|--------------------------- |----------:|----------:|----------:|----------:|-----:|----------:|
|                TryGetValue |  9.649 ns | 0.3920 ns | 1.0927 ns |  9.136 ns |    1 |         - |
| ContainsKeyPlusArrayLookup | 14.196 ns | 0.1173 ns | 0.1040 ns | 14.186 ns |    2 |         - |
```

### DIFF
--- a/src/Ryujinx.Input/Motion/CemuHook/Client.cs
+++ b/src/Ryujinx.Input/Motion/CemuHook/Client.cs
@@ -338,11 +338,11 @@ namespace Ryujinx.Input.Motion.CemuHook
                         {
                             int slot = inputData.Shared.Slot;
 
-                            if (_motionData.ContainsKey(clientId))
+                            if (_motionData.TryGetValue(clientId, out var motionDataItem))
                             {
-                                if (_motionData[clientId].ContainsKey(slot))
+                                if (motionDataItem.ContainsKey(slot))
                                 {
-                                    MotionInput previousData = _motionData[clientId][slot];
+                                    MotionInput previousData = motionDataItem[slot];
 
                                     previousData.Update(accelerometer, gyroscrope, timestamp, cemuHookConfig.Sensitivity, (float)cemuHookConfig.GyroDeadzone);
                                 }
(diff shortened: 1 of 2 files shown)


src/Ryujinx.Input/Motion/CemuHook/Client.cs
```diff
@@ -338,11 +338,11 @@ public void HandleResponse(byte[] data, int clientId)
...
                             int slot = inputData.Shared.Slot;
 
-                            if (_motionData.ContainsKey(clientId))
+                            if (_motionData.TryGetValue(clientId, out var motionDataItem))
                             {
-                                if (_motionData[clientId].ContainsKey(slot))
+                                if (motionDataItem.ContainsKey(slot))
                                 {
-                                    MotionInput previousData = _motionData[clientId][slot];
+                                    MotionInput previousData = motionDataItem[slot];
```
sampletext32, line 345: why not also use TryGetValue on motionDataItem? Isn't it a dict as well?

marysaka: It seems that they are using a tool to automatically replace some kind of lints here considering their other PRs.

So I guess the tool didn't caught that.

marysaka (approved):
lgtm

src/Ryujinx.Input/Motion/CemuHook/Client.cs
marco-carvalho: @sampletext32 could you check again and resolve it if this is fixed?

sampletext32: seems fine

AcK77 (approved):

riperiperi (approved):

PR #5241, opened 2023-06-05
Implement Load/Store Local/Shared and Atomic shared using new instructions

gdkchan:
This is the last PR in the series of PRs to migrate memory related shader operations to the new `Load` and `Store` IR instructions that also carry a `StorageKind`, which I implemented a while ago. This one does the local and shared memory types, which are the only ones remaining. Those are not really buffer backed, rather they use some private memory. Shared memory is only valid for compute shaders, and is shared by all invocations on the workgroup, but is not externally visible. Meanwhile, local memory is only visible within a single invocation,

There are 2 helper functions, one for storing small integer (8-bit and 16-bit) values on shared memory, and another for doing atomic operations on signed integers. The latter is only needed on GLSL since SPIR-V has separate instructions for signed and unsigned atomic operations, regardless of the variable type. Both were moved out of the backend and are now generated at the IR level, like what I did on the previous changes.

Unlike the old implementation, this one allows multiple local memories and shared memories to exist. The old one only allowed one. This can be useful if we combine shader stages in the future for example. It is also good for symmetry: We have the first operand of all `Load` and `Store` operations reserved for the buffer binding number, which must be constant. For local and shared memory, they represent an ID that identifies the memory used.

Advantages are the same of the previous PRs: More code sharing between GLSL and SPIR-V, makes new backends easier to implement, and simpler code (this code manages to remove almost 100 lines of code which is quite nice). It also makes the IR more flexible as we can have multiple separate memory storages now if we need them for different purposes.

This also fixes a regression from #4993, but not a user facing one: The `set` index value was not being set when generating GLSL code, which made GLSL shaders to not work properly on Vulkan. Currently there's no way for users to enable that, it must be enabled manually by changing the code.

After that, there's still work to be done on the shader refactoring:
- Move texture bindings information to `ShaderProperties` like buffers and memory.
- Move remaining values to `TranslatorContext` and delete `ShaderConfig`.
- Move instruction type to `Operation`, stop depending on `InstructionInfo` on GLSL generation and delete `InstructionInfo`.
- Ensure all remaining texture and packing operations can use vector outputs and delete `Index` field from `Operation`.
- Stop using `IGpuAccessor` on the backends and store all required information on `ShaderProperties`.

Testing is welcome.


src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGen.cs
```diff
@@ -68,7 +68,7 @@ private static string GetExpression(CodeGenContext context, AstOperation operati
 
                 string args = string.Empty;
 
-                if (atomic && operation.StorageKind == StorageKind.StorageBuffer)
+                if (atomic && (operation.StorageKind == StorageKind.StorageBuffer || operation.StorageKind == StorageKind.SharedMemory))
```
patrick-hovsepian, line 71: super nit: might be more readable with `is` and `or` if it works in this context

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
```diff
@@ -136,6 +136,19 @@ public static AggregateType GetNodeDestType(CodeGenContext context, IAstNode nod
...
+                        case StorageKind.LocalMemory:
+                        case StorageKind.SharedMemory:
+                            if (!(operation.GetSource(0) is AstOperand bindingId) || bindingId.Type != OperandType.Constant)
+                            {
+                                throw new InvalidOperationException($"First input of {operation.Inst} with {operation.StorageKind} storage must be a constant operand.");
+                            }
+
+                            MemoryDefinition memory = operation.StorageKind == StorageKind.LocalMemory
+                                ? context.Config.Properties.LocalMemories[bindingId.Value]
+                                : context.Config.Properties.SharedMemories[bindingId.Value];
```
patrick-hovsepian, line 148: worth making a helper function that takes a source index and returns a memory definition so it can be reused here and in `InstGenMemory`?

src/Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
```diff
@@ -287,19 +282,19 @@ private static Operand EmitLoadConstant(EmitterContext context, Operand slot, Op
...
-            int count = 1;
-
-            switch (size)
+            int count = size switch
             {
-                case LsSize2.B64: count = 2; break;
-                case LsSize2.B128: count = 4; break;
-            }
+                LsSize2.B64 => count = 2,
+                LsSize2.B128 => count = 4,
```
patrick-hovsepian, line 293: does this need the assignment? can't you just return the value?

src/Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
```diff
@@ -372,52 +363,54 @@ private static Operand EmitLoadConstant(EmitterContext context, Operand slot, Op
...
```
patrick-hovsepian, line 374: similar comment

src/Ryujinx.Graphics.Shader/StructuredIr/MemoryDefinition.cs
```diff
@@ -0,0 +1,18 @@
+using Ryujinx.Graphics.Shader.Translation;
+
+namespace Ryujinx.Graphics.Shader.StructuredIr
+{
+    readonly struct MemoryDefinition
+    {
+        public string Name { get; }
+        public AggregateType Type { get; }
+        public int ArrayLength { get; }
```
patrick-hovsepian, line 9: worth marking these as init?

src/Ryujinx.Graphics.Shader/Translation/HelperFunctionManager.cs
```diff
@@ -51,6 +67,67 @@ private Function GenerateFunction(HelperFunctionName functionName)
...
+        private static Function GenerateSharedAtomicSigned(int id, bool isMin)
+        {
+            EmitterContext context = new EmitterContext();
+
+            Operand wordOffset = Argument(0);
+            Operand value = Argument(1);
+
+            Operand result = GenerateSharedAtomicCasLoop(context, wordOffset, id, (memValue) =>
+            {
+                return isMin
```
patrick-hovsepian, line 91: worth making this a local function? There's an implication with the closures but I don't recall offhand

src/Ryujinx.Graphics.Shader/StructuredIr/MemoryDefinition.cs
patrick-hovsepian: it might imply the constructor too and you could default ArrayLength to 1

patrick-hovsepian (commented):
minor comments/observations

src/Ryujinx.Graphics.Shader/Instructions/InstEmitMemory.cs
gdkchan: Yea, that's what it is meant to do, I fixed it now, thanks.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGen.cs
gdkchan: I'm not a big fan of `is`, `or`, `and` etc expressions. It is easier to understand like this for me.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/OperandManager.cs
gdkchan: I plan to delete that code soon-ish, so I didn't worry too much about de-duplicating it.

src/Ryujinx.Graphics.Shader/Translation/HelperFunctionManager.cs
gdkchan: I could extract it into a separate function. Both are fine for me.

src/Ryujinx.Graphics.Shader/StructuredIr/MemoryDefinition.cs
gdkchan: What's the benefit of using `init` here? AFAIK with `init` the property can only be set on the constructor, but get-only properties also behave like this already, they can only be initialized on the constructor.

patrick-hovsepian: my bad, force of habit and I'm partial to object initializer syntax. I also thought by making it a record the compiler would generate the constructor for you

src/Ryujinx.Graphics.Shader/Translation/HelperFunctionManager.cs
patrick-hovsepian: I think it's fine. I feel like @marco-carvalho might be interested in benchmarking this vs a local function?

marysaka (approved):

Exhigh:
Needs a rebase

riperiperi (approved):
Looking good, sorry for the late review.

PR #5245, opened 2023-06-05
Updater: Stop old exe from being left behind if renamed by user.

Kurochi51:
Fix edge-case scenario introduced by [#5092 ](https://github.com/Ryujinx/Ryujinx/pull/5092)
I'm unsure if this check needs to be performed against `Ryujinx` too, for compatibility with Linux / MacOS

### DIFF
--- a/src/Ryujinx.Ava/Modules/Updater/Updater.cs
+++ b/src/Ryujinx.Ava/Modules/Updater/Updater.cs
@@ -748,6 +748,12 @@ namespace Ryujinx.Modules
                 var newFiles = Directory.EnumerateFiles(UpdatePublishDir, "*", SearchOption.TopDirectoryOnly).Select(Path.GetFileName);
                 var userFiles = oldFiles.Except(newFiles).Select(filename => Path.Combine(HomeDir, filename));
 
+                // Filter out renamed Ryujinx executables for proper disposal
+                if (Path.GetFileName(Environment.GetCommandLineArgs()[0]) != "Ryujinx.exe")
+                {
+                    oldFiles = oldFiles.Where(s => !s.Contains(Path.GetFileName(Environment.GetCommandLineArgs()[0])));
+                }
+
                 // Remove user files from the paths in files.
                 files = files.Except(userFiles);
             }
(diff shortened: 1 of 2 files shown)


gdkchan:
I don't think this is necessary, if the user renamed the exe, maybe it's because they want to keep multiple versions. In any case, it would "revert" back to the default name after the update, so not sure how that helps.

Kurochi51:
I'm concerned that the renamed exe would be left behind, unbeknown to the user. And since it's not an insignificant size (50 MB on Windows), I figured it's better to proper handle the situation?

gdkchan:
I think if the user manually renamed the executable, then said user probably already has the habit of opening the emulator folder and would notice something like that. I'm not sure why someone would do that, the only reason I can think is having multiple version, maybe they want to give a more meaningful name for each of them (but if you have multiple version, makes more sense to store them on separate folders).

Did someone hit this problem in practice already? It appears to be a rather unlikely scenario, and the problem that it would cause is not that bad either (a few MB of wasted disk space), I'm not sure if it's worth the special case in the updater code.

Kurochi51:
No, no one actually ran into this issue, it was just kinda half-jokingly mentioned and I realised the issue. Also the exe would break with a new build no? since dll dependencies will change. However if this would rather be treated as a .... user error situation, I don't mind closing it.

gdkchan:
Most of the time the native libraries that we ship with the emulator don't change. But yea, if one day we changed one of them and the versions are not compatible, it could break.

I think there are two possible scenarios right now:
- User always opens the emulator folder and run the exe from there. On this scenario the user would see that there's 2 executables right away.
- User uses a shortcut (probably the case for most people).

For the second scenario, there are two outcomes:
- With current approach, it would continue running the old version after the update (presumably the user would notice that, if auto-update is enabled it would keep asking to update forever, otherwise if the user updated manually, it should be easy to notice that the version didn't change).
- With the new approach, the shortcut would be broken and it would show some error trying to launch the emulator.

I don't know which one is preferable. Both of them would likely result in support queries, so I'd be in favor of the one that requires the least amount of code (so, not doing this change).

Kurochi51:
Fair enough. Alternatively a popup could be made to show up if you're trying to update from a renamed exe, while there's a `Ryujinx.exe` already in the folder, but at that point it sounds like a different PR. If that's desired, I can do it.

PR #4714, opened 2023-04-23
Generate scaling helper functions on IR

gdkchan:
This PR moves the generation of resolution scale related code out of the GLSL and SPIR-V backends, and instead creates helper functions in the IR code.

Advantages:
- Less code on the backend, so the chances of they behaving differently is lower. It also makes implementing new backends easier since there's less work to do.
- Some optimizations become simpler, like for example using vector outputs with `textureSize` (which was not possible before since the `textureSize` implementation on the backend would try to un-scale the size, and the method that does that only takes scalar inputs right now, not vectors.

Disadvantages:
- More IR code can make the shader translation process a bit slower since there are more things to process.
- The helper functions will become a little less readable since currently there's no variable names, and it can't use vectors right now, although for SPIR-V it doesn't matter too much since it already didn't have names before, and it was generating the scaling code inline.

Sample diff:
master:
```cs
ivec2 Helper_TexelFetchScale(ivec2 inputVec, int samplerIndex)
{
    float scale = s_render_scale[1 + samplerIndex];
    if (scale == 1.0)
    {
        return inputVec;
    }
    if (scale < 0.0) // If less than 0, try interpolate between texels by using the screen position.
    {
        return ivec2(vec2(inputVec) * (-scale) + mod(gl_FragCoord.xy, 0.0 - scale));
    }
    else
    {
        return ivec2(vec2(inputVec) * scale);
    }
}

    // 0x0001B0: 0xDB1000C162473300 Tlds
    temp_45 = texelFetch(fp_tex_tcb_C, ivec3(Helper_TexelFetchScale((ivec3(temp_43, temp_44, int(temp_22))).xy, 1), (ivec3(temp_43, temp_44, int(temp_22))).z), 0).xyzw;
    temp_46 = temp_45.x;
    temp_47 = temp_45.y;
    temp_48 = temp_45.z;
    temp_49 = temp_45.w;
```
PR:
```cs
int TexelFetchScale(int a0, int a1, int a2)
{
    int temp_7;
    precise float temp_8;
    bool temp_9;
    bool temp_10;
    precise float temp_11;
    precise float temp_12;
    precise float temp_13;
    precise float temp_14;
    precise float temp_15;
    precise float temp_16;
    precise float temp_17;
    precise float temp_18;
    int temp_19;
    precise float temp_20;
    precise float temp_21;
    int temp_22;
    temp_7 = 1 + a1;
    temp_8 = support_buffer.s_render_scale[temp_7];
[...]

PR #5266, opened 2023-06-08
Stop identifying shader textures with handle and cbuf, use binding instead

gdkchan:
This does a change similar to what I did to buffers, but with textures and images instead. Previously each `TextureOperation` had a constant buffer slot number, and word offset where the handle is stored in that comes (sometimes erroneously called just `Handle` or `HandleIndex`). This change instead uses the host binding number as a way to identify those textures in the IR. Like the buffer changes, the main advantage here is that this allow us to combine shader stages, and also allows us to reserve and use textures for emulation purposes, without messing with the bindings of the game textures.

Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Shader/Translation/ResourceManager.cs
+++ b/src/Ryujinx.Graphics.Shader/Translation/ResourceManager.cs
@@ -303,12 +303,11 @@ namespace Ryujinx.Graphics.Shader.Translation
                 }
 
                 var definition = new TextureDefinition(
+                    isImage ? 3 : 2,
                     binding,
                     $"{_stagePrefix}_{nameSuffix}",
                     meta.Type,
                     info.Format,
-                    info.CbufSlot,
-                    info.Handle,
                     meta.UsageFlags);
 
                 if (isImage)
@@ -347,9 +346,24 @@ namespace Ryujinx.Graphics.Shader.Translation
 
         public void SetUsageFlagsForTextureQuery(int binding, SamplerType type)
         {
-            if (Properties.Textures.TryGetValue(binding, out var definition))
+            TextureInfo selectedInfo = default;
+            TextureMeta selectedMeta = default;
+            bool found = false;
+
+            foreach ((TextureInfo info, TextureMeta meta) in _usedTextures)
             {
-                definition = definition.SetFlag(TextureUsageFlags.NeedsScaleValue);
+                if (meta.Binding == binding)
+                {
+                    selectedInfo = info;
+                    selectedMeta = meta;
+                    found = true;
+                    break;
+                }
+            }
+
+            if (found)
+            {
+                selectedMeta.UsageFlags |= TextureUsageFlags.NeedsScaleValue;
 
                 var dimensions = type.GetDimensions();
                 var isIndexed = type.HasFlag(SamplerType.Indexed);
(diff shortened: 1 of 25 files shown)


Exhigh:
Needs a rebase, commit [Implement transform feedback emulation for hardware without native support](https://github.com/Ryujinx/Ryujinx/commit/eb0bb36bbfc3a4f5f2ac1c8721e192239f899a1d) breaks build

gdkchan:
I will rebase after #5241 is merged.

Exhigh:
> I will rebase after #5241 is merged.

[Reaction](https://media.tenor.com/x0iX0V3Kw4QAAAAC/hyper-excited.gif)

src/Ryujinx.Graphics.Shader/Translation/ResourceManager.cs
```diff
@@ -140,6 +162,184 @@ public bool TryGetConstantBufferSlot(int binding, out int slot)
...
+            if (existingMeta.AccurateType)
+            {
+                meta.AccurateType = true;
+                meta.Type = existingMeta.Type;
+            }
+
+            return meta;
+        }
+
+        public void SetUsageFlagsForTextureQuery(int binding, SamplerType type)
```
riperiperi, line 322: The usage of this is a bit weird. `InsertCoordNormalization`, `InsertCoordGatherBias`, `InsertConstOffsets` don't care about the scale values, they only use the texture query result internally, so they shouldn't `NeedScaleValue`. The only thing that needs that is use of int coords and texture size instruction, which needs to be unscaled.

FAST RMX appears to be broken at 2x scale right now - the cars are black in the vehicle selection and ingame is very messy (though you'll run out of memory there anyways)

riperiperi: This actually appears to be broken right now, so maybe it was caused by a previous refactor. Still, the same logic applies.

gdkchan: I think the problem is the opposite. When coordinate normalisation is *disabled*, the coordinates used will be "integer" coordinates (they are actually float, but in the range 0..W  and 0..H rather than 0..1). Anyway, the game will pass coordinate values for unscaled texture, so unscaling the texture size is required to make this behave correctly. This has probably been broken by #4714 since it moves size unscaling out of the backend, and right now it happens before those `TextureSize` instructions are inserted. As for `InsertCoordGatherBias` and `InsertConstOffsets `, I think they indeed don't need the size unscaling.

gdkchan: I fixed it and removed the `SetUsageFlagsForTextureQuery` calls on `InsertCoordGatherBias` and `InsertConstOffsets`. It still sets `config.SetUsedFeature(FeatureFlags.IntegerSampling);` which I'm not sure if it's needed.

riperiperi: I'm not actually sure what the IntegerSampling feature does now, the flag isn't checked anywhere.

riperiperi (approved):
Looks good, tested in a bunch of games and seems fine after the coord normalization fix.

riperiperi:
Oh yeah, the shader host cache version needs to be incremented.

marysaka (approved):

Issue #5143, opened 2023-05-29
[Bug] ryujinx not working on radv on 1.1.805+


A-w-x:
### Description of the issue

glitches and crashes in multiple games

a few:
legend arceus
scarlet\violet
links awakening

### Reproduction steps

run games

### Log file

an example shot from scarlet
![Screenshot_20230529_135709](https://github.com/Ryujinx/Ryujinx/assets/80461138/04895f46-5887-43d2-b830-c7238a7ad6d5)

log from links awakening
[Ryujinx_1.1.805_2023-05-29_13-57-39.log](https://github.com/Ryujinx/Ryujinx/files/11591543/Ryujinx_1.1.805_2023-05-29_13-57-39.log)

### OS

archlinux

### Ryujinx version

1.1.805 - current

### Game version

any

### CPU

ryzen 6800U

### GPU

radeon 680m

### RAM

32gb

### List of applied mods

none

0x647262:
> glitches and crashes in multiple games

For crashes specifically, is there any information in your dmesg log (`journalctl --dmesg --grep amdgpu`)? I'm having similar issues and when a game outright crashes I get:

```
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu: [gfxhub] page fault (src_id:0 ring:24 vmid:2 pasid:32769, for process Ryujinx pid>
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:   in page starting at address 0x000080010b8f8000 from client 0x1b (UTCL2)
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu: GCVM_L2_PROTECTION_FAULT_STATUS:0x00201031
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:          Faulty UTCL2 client ID: TCP (0x8)
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:          MORE_FAULTS: 0x1
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:          WALKER_ERROR: 0x0
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:          PERMISSION_FAULTS: 0x3
May 29 22:18:01 onyx kernel: amdgpu 0000:0f:00.0: amdgpu:          MAPPING_ERROR: 0x0
[...]

PR #5267, opened 2023-06-08
Vulkan: Use aspect flags for identity views for bindings

riperiperi:
We still care about the DepthStencilMode for identity image views. Creates a separate view for identity and draw if the aspect flags end up being different for draw.

Fixes visual issues on Mesa (#5143, not sure if it solves all the crashes), and also validation errors elsewhere when asking for identity views of depth textures.


gdkchan (approved):
lgtm, thanks for fixing this! I should have noticed that the aspect flags were wrong, but I completely forgot about it.

AcK77 (approved):

gdkchan:
Should fix #5143? There was another one (for Windows), but I can't find it right now.

PR #5270, opened 2023-06-09
Fix regression in macOS updater

siegmund-heiss-ich:
This should fix a regression introduced by 1.1.869 where the macOS updater won't work anymore.

Because we use a `*.app` package, it isn't necessary to check if the user changed any files in this directory.
The updater for macos is designed, that it always replaces the whole package.


marysaka (approved):

TSRBerry (approved):
Thank you for fixing this, we missed that during review before!

PR #5271, opened 2023-06-09
Check if existing oldConfigPath is a Symlink

siegmund-heiss-ich:
In #4296 there was an unhandled exception introduced, which I want to address with this PR.
I recognized this issue thanks to a user in macos-support on discord.

If you used the orinal macos ryujinx release and installed later versions, you have a symlink in the old directory (~/.config/Ryujinx) and the ryujinx config folder in the new one (~/Library/Application Support).
If you somehow (classical skill issue) manage to delete the new config folder, ryujinx won't start anymore, because it tries to copy the old directory (which is a symlink now).

I included a check, if the existing old path is just a symlink.
If it is true, a new ryujinx config folder will be created, and the app starts.

Of course we could just remove the symlink actions, but it is to early at this point because the official macos-2 release of ryujinx isn't even released yet.


marysaka (approved):

TSRBerry (approved):
Thank you!

PR #5272, opened 2023-06-09
Add macOS Headless release workflow

siegmund-heiss-ich:
Added a release workflow for macOS headless build.

I think it's better to have a seperate script to do this.
For that I modified the existing `create_macos_build.sh`.

Because we don't have a package here, codesigning happens for every Mach-O file.
App bundle and the plist things are removed.


distribution/macos/create_macos_build_headless.sh
```diff
@@ -0,0 +1,107 @@
...
+X64_OUTPUT="$TEMP_DIRECTORY/publish_x64"
+UNIVERSAL_OUTPUT="$OUTPUT_DIRECTORY/publish"
+EXECUTABLE_SUB_PATH=Ryujinx.Headless.SDL2
+
+rm -rf "$TEMP_DIRECTORY"
+mkdir -p "$TEMP_DIRECTORY"
+
+DOTNET_COMMON_ARGS="-p:DebugType=embedded -p:Version=$VERSION -p:SourceRevisionId=$SOURCE_REVISION_ID --self-contained true $EXTRA_ARGS"
+
+dotnet restore
```
patrick-hovsepian, line 40: will the script exit on error? i.e. would it make sense to do `&&`

patrick-hovsepian (commented):
from a cursory look/understanding a separate script makes sense to me

distribution/macos/create_macos_build_headless.sh
siegmund-heiss-ich: Yes this script will exit on error. But that's how every `*.sh` script (that I have seen) works in ryujinx.
It seems there are some problems with signing. I just checked codesign but not rcodesign locally. I have to figure out what's happening there.

siegmund-heiss-ich: Everything should work correctly now.
Could you please explain what your suggestion is?
I don't exactly understand what you mean.
Thanks

patrick-hovsepian: Sure, apologies for being vague. What I meant was there's multiple steps here to generating the final dll. These include a restore, build, publish. If the restore fails, the other two are likely going to fail as well so it wouldn't be worth attempting those.

Say `dotnet restore` fails, if the script exits, all is good there's nothing more to do. If it doesn't and still invokes, `dotnet build -c ...` it's wasted cycles. My suggestion was to chain them with the `&&` operator which should only run the next command if the previous was successful so something like 

```
dotnet restore && \
dotnet build -c $CONFIGURATION src/Ryujinx.Headless.SDL2 && \
dotnet publish -c $CONFIGURATION -r osx-arm64 -o "$TEMP_DIRECTORY/publish_arm64" $DOTNET_COMMON_ARGS src/Ryujinx.Headless.SDL2 && \
dotnet publish -c $CONFIGURATION -r osx-x64 -o "$TEMP_DIRECTORY/publish_x64" $DOTNET_COMMON_ARGS src/Ryujinx.Headless.SDL2
```

if restore worked, it'll run build, if that worked, it'll publish the arm64 and finally the last version. However, if any intermediate step fails, say restore succeeds, but build fails, it'll not even run the remaining steps. (The `&& \` is to escape newlines)

siegmund-heiss-ich: Thank you for the explanation. But I don't think this is necessary because the script exits as soon as there is an error because of `set -e` at the start of the script, and the steps are executed one after another anyway.

patrick-hovsepian: ah thanks, I missed that bit

marysaka (approved):
lgtm

TSRBerry (approved):
Lgtm, thank you!

PR #5274, opened 2023-06-09
Using Span<T> instead of T[]

marco-carvalho:
`Span<T>` is a struct introduced in .NET Core which provides a way to work with contiguous blocks of memory in a safe and efficient way. This can provide significant performance improvements, particularly in scenarios where you would otherwise be forced to allocate and deallocate large arrays frequently. Here are some of the benefits of using `Span<T>`:

- **Performance Improvement**: `Span<T>` allows you to operate on portions of data without copying data or allocating additional arrays. This can drastically reduce unnecessary allocations and therefore, improve memory utilization and overall performance.

- **Safe Operations**: `Span<T>` provides a safe way to refer to contiguous regions of arbitrary memory, whether that memory is part of an array, a portion of an array, a stack-allocated span of memory, or even an unmanaged block of memory. This provides a safe way to work with memory without having to directly deal with pointers.

- **Enhanced Array Segmentation**: With an array (`T[]`), if you want to work with a segment of the array, you would need to make a new array and copy the segment into it. `Span<T>` allows you to "slice" the data and work with a portion of it, without creating a new array, which can be much more efficient.

- **Versatility**: `Span<T>` can point to managed memory (arrays), stack memory (span local variables), and native memory (from interop). So it's more versatile than just arrays.

- **Stack Allocation**: With `Span<T>`, you can create slices of data structures on the stack, avoiding heap allocations altogether in some cases. This can provide performance benefits because stack allocations are cheaper than heap allocations.

However, `Span<T>` has some limitations. For example, it can't be stored in class fields, array elements, or other places that might cause it to be stored beyond the scope in which it was created, because it's a stack-only type. This is to ensure safety and prevent memory leaks.

In scenarios where you need to store data for longer periods or across asynchronous boundaries, you might need to use `Memory<T>` or traditional arrays. Also, since `Span<T>` is a recent addition to the .NET ecosystem, older libraries might not support it and may require arrays instead.

Benchmark:

```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    [Benchmark]
    public void Array()
    {
        int[] array = new[] { 1, 2, 3 };
        for (int i = 0; i < array.Length; i++)
        {
            var x = array[i];
        }
    }

    [Benchmark]
    public void Span()
    {
        Span<int> span = new[] { 1, 2, 3 };
        for (int i = 0; i < span.Length; i++)
        {
            var x = span[i];
        }
    }
}
```

Results:
```sh
| Method |     Mean |     Error |    StdDev |   Median | Rank |   Gen0 | Allocated |
|------- |---------:|----------:|----------:|---------:|-----:|-------:|----------:|
|  Array | 5.108 ns | 0.1344 ns | 0.2423 ns | 5.003 ns |    1 | 0.0127 |      40 B |
|   Span | 5.043 ns | 0.1210 ns | 0.1132 ns | 5.004 ns |    1 | 0.0127 |      40 B |
```

### DIFF
--- a/src/ARMeilleure/Decoders/OpCodeTable.cs
+++ b/src/ARMeilleure/Decoders/OpCodeTable.cs
@@ -1405,7 +1405,7 @@ namespace ARMeilleure.Decoders
             int xMask = 0;
             int xBits = 0;
 
-            int[] xPos = new int[encoding.Length];
+            Span<int> xPos = new int[encoding.Length];
 
             int blacklisted = 0;
 
--- a/src/Ryujinx.Ava/UI/Models/SaveModel.cs
+++ b/src/Ryujinx.Ava/UI/Models/SaveModel.cs
@@ -41,7 +41,7 @@ namespace Ryujinx.Ava.UI.Models
         private string GetSizeString()
         {
             const int scale = 1024;
-            string[] orders = { "GiB", "MiB", "KiB" };
+            Span<string> orders = new []{ "GiB", "MiB", "KiB" };
             long max = (long)Math.Pow(scale, orders.Length);
 
             foreach (string order in orders)
--- a/src/Ryujinx.Common/SystemInfo/LinuxSystemInfo.cs
+++ b/src/Ryujinx.Common/SystemInfo/LinuxSystemInfo.cs
@@ -61,7 +61,7 @@ namespace Ryujinx.Common.SystemInfo
                 string line;
                 while ((line = file.ReadLine()) != null)
                 {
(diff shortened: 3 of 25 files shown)


gdkchan:
I don't think those improvements applies to most of the cases that have been changed here, since they are still backed by arrays. For methods that accept an array, I think changing to span is generally better as spans are more flexible than arrays. But in most of the cases changed here, a `ReadOnlySpan<T>` should be used rather than a `Span<T>`, since the values are not really modified.

PR #5277, opened 2023-06-10
Prevent crash when game is corrupted.

kekonn:
Shows an error if a game is corrupted instead of crashing to desktop.

This is my first PR, so definitely looking for constructive feedback on desired code style etc.

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1480,10 +1480,29 @@ namespace Ryujinx.Ava.UI.ViewModels
 
             async void Action()
             {
-                if (!await AppHost.LoadGuestApplication())
+                var loadSuccessful = false;
+                try
                 {
-                    AppHost.DisposeContext();
+                    loadSuccessful = await AppHost.LoadGuestApplication();
+                }
+                catch (HorizonResultException hex)
+                {
+                    // Apphost Cleanup
                     AppHost = null;
+                    SelectedIcon = null;
+                
+                    await Dispatcher.UIThread.InvokeAsync(async () =>
+                    {
+                        await ContentDialogHelper.CreateErrorDialog(
+                            string.Format(LocaleManager.Instance[LocaleKeys.DialogCorruptedGameError], hex.Message));
+                    });
+                    
+                    return;
+                }
+                
+                if (!loadSuccessful)
+                {
+                    PostAppCleanup();
 
                     return;
                 }
@@ -1505,7 +1524,7 @@ namespace Ryujinx.Ava.UI.ViewModels
                 Thread gameThread = new(InitializeGame) { Name = "GUI.WindowThread" };
                 gameThread.Start();
             }
-
+            
             Dispatcher.UIThread.Post(Action);
         }
 
@@ -1562,6 +1581,11 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         public void AppHost_AppExit(object sender, EventArgs e)
         {
+            PostAppCleanup();
+        }
+
+        private void PostAppCleanup()
+        {
             if (IsClosing)
             {
                 return;
@@ -1590,10 +1614,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
             SelectedIcon = null;
 
-            Dispatcher.UIThread.InvokeAsync(() =>
-            {
-                Title = $"Ryujinx {Program.Version}";
-            });
+            Dispatcher.UIThread.InvokeAsync(() => { Title = $"Ryujinx {Program.Version}"; });
         }
 
         public void ToggleFullscreen()
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1502,7 +1502,8 @@ namespace Ryujinx.Ava.UI.ViewModels
                 
                 if (!loadSuccessful)
                 {
-                    PostAppCleanup();
+                    AppHost.DisposeContext();
+                    AppHost = null;
 
                     return;
                 }
@@ -1580,11 +1581,6 @@ namespace Ryujinx.Ava.UI.ViewModels
         }
 
         public void AppHost_AppExit(object sender, EventArgs e)
-        {
-            PostAppCleanup();
-        }
-
-        private void PostAppCleanup()
         {
             if (IsClosing)
             {
--- a/src/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/src/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -650,5 +650,6 @@
   "NetworkInterfaceDefault": "Default",
   "PackagingShaders": "Packaging Shaders",
   "AboutChangelogButton": "View Changelog on GitHub",
-  "AboutChangelogButtonTooltipMessage": "Click to open the changelog for this version in your default browser."
+  "AboutChangelogButtonTooltipMessage": "Click to open the changelog for this version in your default browser.",
+  "DialogCorruptedGameError": "It appears this game is corrupted. Please dump the game again.\nInternal error code: {0}" 
 }


gdkchan:
Most of the corruption errors I have seen happens when trying to load RomFS data using `IStorage` interface (which happens after initialization), so I guess this would not be able to catch such errors.

Before something like that is attempted, I would prefer if we had discussion and implementation of the structure required to catch errors that may happens at various places in the emulator, and show a helpful message on the UI for those errors, instead of crashing.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1480,7 +1480,27 @@ public async void LoadApplication(string path, bool startFullscreen = false, str
 
             async void Action()
             {
-                if (!await AppHost.LoadGuestApplication())
+                var loadSuccessful = false;
+                try
+                {
+                    loadSuccessful = await AppHost.LoadGuestApplication();
+                }
+                catch (HorizonResultException hex)
```
patrick-hovsepian, line 1488: would this be better served if the exception was caught within the call to `AppHost.Load..` That way, the cleanup method stays consistent? IMO it'd be a nice stop gap to stop hard crashes until the team can align on an more robust error handling mechanism to gdkchan's point

patrick-hovsepian: it would also be good to log an error too IMO

kekonn: Calling CreateErrorDialog logs an error already. I had originally put in a separate error statement but during testing I noticed I got two error lines. The contents of the string.Format are already logged as an error.

kekonn:
> Before something like that is attempted, I would prefer if we had discussion and implementation of the structure required to catch errors that may happens at various places in the emulator, and show a helpful message on the UI for those errors, instead of crashing.

Sorry but I'm not sure what you mean by this? Isn't that exactly what I'm doing? Or do you mean to say we should look into preventing the exception instead of catching it?

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
kekonn: @patrick-hovsepian I just tried moving it to the AppHost, but the problem is that the cleanup has to be separate. Because if this error occurs and you call dispose on the apphost, you get another exception.

There might be a more intricate rework that makes it possible, but I feel right now, this is the most efficient way to solve this error hard crashing the emulator.

kekonn:
> Rebase is wrong

Could you be more specific please? Do you want me to undo the rebase commits?

AcK77:
> Could you be more specific please? Do you want me to undo the rebase commits?

Your PR include the merged PR on master, it should contains only your commits rebased on top of master.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
AcK77, line 1488:
```suggestion
                catch (HorizonResultException ex)
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1480,7 +1480,27 @@ public async void LoadApplication(string path, bool startFullscreen = false, str
 
             async void Action()
             {
-                if (!await AppHost.LoadGuestApplication())
+                var loadSuccessful = false;
+                try
+                {
+                    loadSuccessful = await AppHost.LoadGuestApplication();
```
AcK77, line 1486:
```suggestion
                    isApplicationLoaded = await AppHost.LoadGuestApplication();
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1480,7 +1480,27 @@ public async void LoadApplication(string path, bool startFullscreen = false, str
...
+                    await Dispatcher.UIThread.InvokeAsync(async () =>
+                    {
+                        await ContentDialogHelper.CreateErrorDialog(
+                            string.Format(LocaleManager.Instance[LocaleKeys.DialogCorruptedGameError], hex.Message));
+                    });
+                    
+                    return;
+                }
+                
+                if (!loadSuccessful)
```
AcK77, line 1503:
```suggestion
                if (!isApplicationLoaded)
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1505,7 +1525,7 @@ async void Action()
                 Thread gameThread = new(InitializeGame) { Name = "GUI.WindowThread" };
                 gameThread.Start();
             }
-
+
```
AcK77, line 1528: Extra empty spaces

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1590,10 +1610,7 @@ public void AppHost_AppExit(object sender, EventArgs e)
 
             SelectedIcon = null;
 
-            Dispatcher.UIThread.InvokeAsync(() =>
-            {
-                Title = $"Ryujinx {Program.Version}";
-            });
+            Dispatcher.UIThread.InvokeAsync(() => { Title = $"Ryujinx {Program.Version}"; });
```
AcK77, line 1613: I'd prefer to keep that unaligned. Thanks.

AcK77 (changes requested):

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1480,7 +1480,27 @@ public async void LoadApplication(string path, bool startFullscreen = false, str
 
             async void Action()
             {
-                if (!await AppHost.LoadGuestApplication())
+                var loadSuccessful = false;
```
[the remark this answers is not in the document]
kekonn: is the change from var to bool also intentional? Should I try to avoid using var?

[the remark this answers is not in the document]
AcK77: Yes it is, we usually use `var` when the type name is too long, or if the assignation have the type too. It’s just a personnal preference here.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
kekonn: I'm not sure what's going on here, but those lines don't contain any double newlines in my IDE. Is this the github preview that's messing up?

patrick-hovsepian: Try just deleting the line entirely and just pressing enter at the end of 1526

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1480,7 +1480,26 @@ public async void LoadApplication(string path, bool startFullscreen = false, str
...
+                    // Apphost Cleanup
+                    AppHost = null;
+                    SelectedIcon = null;
+                
+                    await Dispatcher.UIThread.InvokeAsync(async () =>
+                    {
+                        await ContentDialogHelper.CreateErrorDialog(string.Format(LocaleManager.Instance[LocaleKeys.DialogCorruptedGameError], ex.Message));
+                    });
+                    
+                    return;
```
patrick-hovsepian, line 1499: IMO, let this launch the error dialog then just drop down and invoke the existing cleanup. that way you can remove 1491/2 too

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
kekonn: Doesn't change anything.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
kekonn: The existing cleanup does not work. Calling AppHost.DisposeContext will generate another exception.

Issue #1028, opened 2020-03-24
hidbus IHidbusServer: 1 (GetBusHandle) not implemented 


B3astTech:
IHidbusServer GetBusHandle is not implemented.

## Service description:
[Switchbrew](https://switchbrew.org/wiki/HID_services#GetBusHandle)

```
Ryujinx.HLE.HOS.Services.Hid.IHidbusServer: 1
```

```
Takes an input u32 #NpadIdType, an u64 #BusType, an u64 AppletResourceUserId, returns an output u8 bool and a #BusHandle.

The bool indicates whether the #BusHandle is valid.

Official sw will assert when #BusHandle InternalIndex is >=0x11 (>=0x13 with [6.0.0+]). This same check is also done for all funcs using #BusHandle as input.

This cmd will throw an error when #BusType is invalid
```

## Required by:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/88
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/261
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/67

gdkchan:
Which game?

EmulationFanatic:
Is also required by https://github.com/Ryujinx/Ryujinx-Games-List/issues/67 with more updated versions.

PR #5284, opened 2023-06-11
hle: Stub IHidbusServer.GetBusHandle

Steveice10:
Partially resolves https://github.com/Ryujinx/Ryujinx/issues/1028

Reference: https://switchbrew.org/wiki/HID_services#GetBusHandle

Tested with NES Switch Online, now boots and is playable.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Hid/IHidbusServer.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Hid/IHidbusServer.cs
@@ -1,8 +1,29 @@
-﻿namespace Ryujinx.HLE.HOS.Services.Hid
+﻿using Ryujinx.Common;
+using Ryujinx.Common.Logging;
+
+namespace Ryujinx.HLE.HOS.Services.Hid
 {
     [Service("hidbus")]
     class IHidbusServer : IpcService
     {
         public IHidbusServer(ServiceCtx context) { }
+
+        [CommandCmif(1)]
+        // GetBusHandle(nn::hid::NpadIdType, nn::hidbus::BusType, nn::applet::AppletResourceUserId) -> (bool HasHandle, nn::hidbus::BusHandle)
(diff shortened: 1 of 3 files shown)


src/Ryujinx.HLE/HOS/Services/Hid/IHidbusServer.cs
```diff
@@ -1,8 +1,26 @@
...
+
+namespace Ryujinx.HLE.HOS.Services.Hid
 {
     [Service("hidbus")]
     class IHidbusServer : IpcService
     {
         public IHidbusServer(ServiceCtx context) { }
+
+        [CommandCmif(1)]
+        // GetBusHandle(nn::hid::NpadIdType, nn::hidbus::BusType, nn::applet::AppletResourceUserId) -> bool HasHandle, nn::hidbus::BusHandle
```
gdkchan, line 11:
```suggestion
        // GetBusHandle(nn::hid::NpadIdType, nn::hidbus::BusType, nn::applet::AppletResourceUserId) -> (bool HasHandle, nn::hidbus::BusHandle)
```
IIRC that's how we do multiple value returns in the other comments.
Also it seems that you write the bool, but you never write the output `BusHandle`.

Steveice10: The bool indicates whether the following structure is valid, since we write false the following data shouldn't matter. But I can add in some dummy structure data if you prefer.

Steveice10: Updated to fix style and include handle data in the stubbed output.

AcK77:
Kind of duplicated https://github.com/Ryujinx/Ryujinx/pull/3846

marysaka (approved):
Should be fine for now I guess

Issue #5136, opened 2023-05-28
Hide cursor option is still broken


TSRBerry:
@TSRBerry Unfortunately, the argument is still broken. 

The cursor disappears during the loading screen, but it reappears when the game is loaded.

Ava UI version 1.1.800

I am using the following command:

Ruyjinx.Ava.exe -f --hide-cursor Always Game.nsp

I need to move the cursor outside the window to avoid seeing it.

_Originally posted by @jojodunet in https://github.com/Ryujinx/Ryujinx/issues/4927#issuecomment-1552031910_

Exhigh:
> Ruyjinx.Ava.exe -f -hide-cursor Always Game.nsp

in the command it should be --hide-cursor instead of -hide-cursor

gdkchan:
Fixed by #6725.

PR #5288, opened 2023-06-11
[Ava] Hide Cursor and Window Main Loop Refactor

patrick-hovsepian:
This PR addresses the Ava UI not properly hiding the cursor. The check for whether the cursor is within the active window did not take into account the windows position as an offset.

This PR addresses that issues and also slightly refactors some of the logic for readability and modularity. There does appear to be frame updates that are made even when nothing changed. I'd like to go back and address that but will do so separately.

This may address #5136 and is related to #3662


src/Ryujinx.Ava/AppHost.cs
```diff
@@ -191,20 +191,22 @@ internal class AppHost
...
-                }
+            if (_rendererHost.EmbeddedWindow.TransformedBounds != null)
+            {
+                var point = e.GetCurrentPoint(window).Position;
+                var bounds = _rendererHost.EmbeddedWindow.TransformedBounds.Value.Clip;
+
+                _isCursorInRenderer = point.X >= bounds.X &&
+                                        point.X <= bounds.Width + bounds.X + window.Position.X &&
+                                        point.Y >= bounds.Y &&
+                                        point.Y <= bounds.Height + bounds.Y + window.Position.Y;
```
patrick-hovsepian, line 209: account for window position

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -994,9 +997,54 @@ private bool UpdateFrame()
...
+                    ConfigurationState.Instance.Graphics.AspectRatio.Value.ToFloat());
+
+            if (!hasTouch)
+            {
+                Device.Hid.Touchscreen.Update();
+            }
+
+#if DEBUG
+            Device.Hid.DebugPad.Update();
+#endif
```
patrick-hovsepian, line 1022: wrapped this in a pre-processor check

patrick-hovsepian (commented):
highlights

src/Ryujinx.Ava/AppHost.cs
AcK77: DebugPad is something related to the Switch hid service, it shouldn’t be wrapped since it’s not related to a Debug build.

patrick-hovsepian: thanks for the context. reverted that bit.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -852,6 +856,7 @@ private void MainLoop()
         {
             while (_isActive)
             {
+                // wait for an event to trigger this rather than poll.
```
TSRBerry, line 859: Why are these comments here?

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -994,9 +999,53 @@ private bool UpdateFrame()
                 return false;
             }
 
+            // Only call if needed
```
TSRBerry, line 1002: You changed nothing here, what was that comment added for?

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1031,103 +1080,66 @@ private bool UpdateFrame()
...
 
-                Dispatcher.UIThread.Post(() =>
-                {
-                    if (_keyboardInterface.GetKeyboardStateSnapshot().IsPressed(Key.Delete) && _viewModel.WindowState != WindowState.FullScreen)
-                    {
-                        Device.Processes.ActiveApplication.DiskCacheLoadState?.Cancel();
-                    }
-                });
-
+            void TriggerKeyboardHotkey()
```
TSRBerry, line 1085:
```suggestion
            void UpdateHotkeyState()
```

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1031,103 +1080,66 @@ private bool UpdateFrame()
...
+            {
                 KeyboardHotkeyState currentHotkeyState = GetHotkeyState();
-
-                if (currentHotkeyState != _prevHotkeyState)
+                if (currentHotkeyState == _prevHotkeyState)
```
TSRBerry, line 1088:
```suggestion
                
                if (currentHotkeyState == _prevHotkeyState)
```

TSRBerry (changes requested):
Most of the changes lgtm! Just left a few comments where I didn't understand your changes or where I think something could be better named.

Exhigh:
This commit seems to aggressively hide the cursor, like whilst moving the cursor it tries to hide the cursor, making the cursor go active and hidden constantly. ( This is on Avalonia btw, also the setting that I am on is to hide the cursor while on idle )

src/Ryujinx.Ava/AppHost.cs
patrick-hovsepian: Just thinking out loud - I'll clean this and the other comment up.

patrick-hovsepian:
@Exhigh this should be the same behavior as the master branch. Would you mind confirming if you see the same behavior there?

TSRBerry:
I can't reproduce the issue you mentioned @Exhigh. For me direct mouse access behaves as expected, so the mouse is hidden even when moving it and only is shown outside of the emulator window (and for the menu items).
Do you have a video of the behavior you are seeing?

TSRBerry (approved):
Lgtm, thank you!

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -191,21 +191,31 @@ internal class AppHost
...
-                }
+            if (_rendererHost.EmbeddedWindow.TransformedBounds is null)
+            {
+                return;
             }
+
+            if (_viewModel.WindowState == WindowState.FullScreen)
+            {
+                _isCursorInRenderer = true;
+                return;
```
AcK77, line 209:
```suggestion

                return;
```

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -191,21 +191,31 @@ internal class AppHost
...
+            }
+
+            var point = e.GetCurrentPoint(window).Position;
+            var bounds = _rendererHost.EmbeddedWindow.TransformedBounds.Value.Clip;
+
+            _isCursorInRenderer = point.X >= bounds.X &&
+                point.X <= bounds.Width + bounds.X + window.Position.X &&
+                point.Y >= bounds.Y &&
+                point.Y <= bounds.Height + bounds.Y + window.Position.Y;
```
AcK77, line 218:
```suggestion
            _isCursorInRenderer = point.X >= bounds.X && point.X <= bounds.Width + bounds.X + window.Position.X
                && point.Y >= bounds.Y && point.Y <= bounds.Height + bounds.Y + window.Position.Y;
```

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -997,6 +1007,48 @@ private bool UpdateFrame()
...
+
+            UpdateHotkeyState();
+
+            if (ScreenshotRequested)
+            {
+                ScreenshotRequested = false;
+                _renderer.Screenshot();
+            }
+
+            void UpdateCursorState()
```
AcK77, line 1051: I'm not a huge fan of getting the method included in `UpdateActiveFrame`

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1031,103 +1083,67 @@ private bool UpdateFrame()
...
 
-                Dispatcher.UIThread.Post(() =>
-                {
-                    if (_keyboardInterface.GetKeyboardStateSnapshot().IsPressed(Key.Delete) && _viewModel.WindowState != WindowState.FullScreen)
-                    {
-                        Device.Processes.ActiveApplication.DiskCacheLoadState?.Cancel();
-                    }
-                });
-
+            void UpdateHotkeyState()
```
AcK77, line 1088: I'm not a huge fan of getting the method included in `UpdateActiveFrame`

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -997,6 +1007,48 @@ private bool UpdateFrame()
             NpadManager.Update(ConfigurationState.Instance.Graphics.AspectRatio.Value.ToFloat());
 
             if (_viewModel.IsActive)
+            {
+                UpdateActiveFrame();
+            }
+
+            // Touchscreen.
```
AcK77, line 1014:
```suggestion
```
This can be removed, since we haven't any comment for other input, it make no sense to keep it.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1031,103 +1083,67 @@ private bool UpdateFrame()
...
+                        _newVolume = MathF.Round((Device.GetVolume() - VolumeDelta), 2);
+                        Device.SetVolume(_newVolume);
+
+                        _viewModel.Volume = Device.GetVolume();
+                        break;
+                    case KeyboardHotkeyState.None:
+                        (_keyboardInterface as AvaloniaKeyboard).Clear();
+                        break;
                 }
+                _prevHotkeyState = currentHotkeyState;
```
AcK77, line 1145:
```suggestion

                _prevHotkeyState = currentHotkeyState;
```

AcK77 (changes requested):

Exhigh:
> y

I could create a video, would take me some time though.

Exhigh:
> I can't reproduce the issue you mentioned @Exhigh. For me direct mouse access behaves as expected, so the mouse is hidden even when moving it and only is shown outside of the emulator window (and for the menu items). Do you have a video of the behavior you are seeing?

[Video of the previously stated behavior](https://we.tl/t-5glnurkRxE)

TSRBerry:
Seems to be a Windows only issue then.

src/Ryujinx.Ava/AppHost.cs
TSRBerry: It was part of it before as well, what's the issue here?

Exhigh:
> Seems to be a Windows only issue then.

well

src/Ryujinx.Ava/AppHost.cs
patrick-hovsepian: I think it's the use of the local function vs a private one in the class? I'll update it if you really want me to but i think there's a small performance boost (although not 100% sure on that) the other thing is it makes it clear that the function is only intended to be used within that calling scope not arbitrarily by other class methods.

Exhigh:
As of the most recent commit, the cursor hides itself and seems to re-appear after some duration and hides itself again, I am not sure if its an issue just on my end.

TSRBerry:
Doesn't happen on Linux at least, but that's all I can tell you.

patrick-hovsepian:
It definitely is a bit wonky on Windows for sure. I tested in Windows 10 and MacOS and didn't see the problem you mentioned but maybe document an issue with settings and someone can dig into it at some point?

jojodunet:
ava-ryujinx-Release-1.1.0+ca2bc33-win_x64.zip
Windows 10

Ruyjinx.Ava.exe -f --hide-cursor Always Game.nsp

The cursor disappears during the loading screen. It reappears when the game is launched. 
However, if I move the mouse afterwards, it disappears again.

jojodunet:
> As of the most recent commit, the cursor hides itself and seems to re-appear after some duration and hides itself again, I am not sure if its an issue just on my end.

@Exhigh I have the same issue, but it disappears, reappears, and disappears again at precise moments, as I explained above.

patrick-hovsepian:
Thanks for checking, are you able to confirm if you see the same thing on master?

jojodunet:
> Thanks for checking, are you able to confirm if you see the same thing on master?

@patrick-hovsepian It’s not the same on master. 

Here the issue on master actually :
https://github.com/Ryujinx/Ryujinx/pull/4927#issuecomment-1552031910

AcK77:
It could be nice to get this issues fixed before considering a merge. Thanks :)

TSRBerry:
If it's the same as master I don't think it's on them to fix it.

jojodunet:
Who can handle it and fix the issue on Windows?

TSRBerry:
Why was this closed?

patrick-hovsepian:
@TSRBerry Exhigh reached out offline about adding some changes which is #5415

PR #5290, opened 2023-06-12
Make LM skip instead of crashing for invalid messages

gdkchan:
On #4701, I fixed some `IndexOutOfRangeException` crashes on LM service caused by invalid sizes that would make it try to read more data than what is in the message buffer, but it would only apply to cases where it reads a string with variable length. This extends it to all reads. A new `TryRead` method was added to `SpanReader` that will return false if the value can't be read due to it not fitting in the span. In the cases where `TryRead` returns false, it returns immediately and prints whatever we have.

Fixes a crash that I found while trying to start a new match on Mortal Kombat 11 (base version).

### DIFF
--- a/src/Ryujinx.Common/Memory/SpanReader.cs
+++ b/src/Ryujinx.Common/Memory/SpanReader.cs
@@ -24,6 +24,24 @@ namespace Ryujinx.Common.Memory
             return value;
         }
 
+        public bool TryRead<T>(out T value) where T : unmanaged
+        {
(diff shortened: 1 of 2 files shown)


gdkchan:
> Lgtm!
> 
> Would be nice to figure out how hardware deals with invalid messages as well.

The LM service is stubbed on retail units, it would just do nothing.

TSRBerry:
Ah true. Sorry I forgot about that.

AcK77 (approved):

PR #5291, opened 2023-06-12
Add an alternative backend parallel compilation queue for macOS

### LINKED ISSUE #5282: [Bug] apple silicon running splatoon 3 crashing
### Description of the issue

.NET ThreadPool Worker [mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
.NET ThreadPool Worker output of type float4 is not compatible with a MTLPixelFormatRGBA32Uint color attachement..
.NET ThreadPool Worker Gpu : Background Compilation failed: Unexpected API error "ErrorInitializationFailed".
.NET ThreadPool Worker [mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
.NET ThreadPool Worker output of type float4 is not compatible with a MTLPixelFormatRGBA32Uint color attachement..
.NET ThreadPool Worker Gpu : Background Compilation failed: Unexpected API error "ErrorInitializationFailed".
.NET ThreadPool Worker [mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
GUI.RenderThread Gpu : Background Compilation failed: Unexpected API error "ErrorInitializationFailed".
.NET ThreadPool Worker output of type float4 is not compatible with a MTLPixelFormatRGBA16Uint color attachement..
HLE.OsThread.10 ServiceNv ZbcSetTable: Stubbed. 
HLE.OsThread.10 ServiceNv GetActiveSlotMask: Stubbed. 

### Reproduction steps

When running Splatoon 3 4.0.0 version.

### Log file

[Ryujinx_1.1.881_2023-06-11_04-00-42.log](https://github.com/Ryujinx/Ryujinx/files/11713232/Ryujinx_1.1.881_2023-06-11_04-00-42.log)

### OS

macOS 13.4 / NS 16.0.3

### Ryujinx version

1.1.881

### Game version

4.0.0

### CPU

M1P

### GPU

M1P

### RAM

16G

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
Due to the way how SPIRV-Cross works, some shaders with deeply nested if/else blocks can cause a stack overflow depending on the stack size. SPIRV-Cross is apparently using recursive function calls in some function where it visits blocks in the code (which is bad in general IMO). To work around this issue, this change basically implements a custom thread pool. Since we create the threads ourselves, we can specify a custom stack size that is large enough to make SPIRV-Cross happy.

The obvious downside with that change is that we need to create our own thread pool, which is most likely going to perform worse than .NET one (also the current implementation is pretty basic, it has 8 queues and adds compilation requests to them in a rotating way, maybe it should have a better method for balancing the load).

An alternative to this would passing some value to the backend that indicates the amount of nesting the shader has. The backend would use that value to decide if it should use `Task`s to compile shaders in parallel, or if it should just compile it on the render thread (which already has an increased stack size).

Suggestions are welcome. I didn't really want to upstream that since I think not using `Task`s here is a pretty significant downside.  But as we come very close to upstreaming all the changes, and there's a general desire to tell everyone to switch to more up-to-date builds instead of using macos1 for specific games, I felt it was worth PRing it, even if just to start discussion about a better way to solve this problem.

We will need tests to ensure shader compilation is not slower on macOS with this change.

Affected games
---
This fixes a stack overflow on Mortal Kombat 11. I also opened #5290 to fix a different crash that I found while testing it. For this game, I had to increase the stack size to 2MB (on macos1, I used 1MB as that was enough for Splatoon 3).
Captura de Tela 2023-06-12 às 00 44 35
Also fixes the stack overflow crash when starting Splatoon 3.
Captura de Tela 2023-06-12 às 00 47 00
Captura de Tela 2023-06-12 às 00 48 15
Captura de Tela 2023-06-12 às 00 48 46
Contributes to #4062, closes #5282.

### DIFF
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -134,7 +134,7 @@ namespace Ryujinx.Ava
             _inputManager           = inputManager;
             _accountManager         = accountManager;
             _userChannelPersistence = userChannelPersistence;
-            _renderingThread        = new Thread(RenderLoop, 1 * 1024 * 1024) { Name = "GUI.RenderThread" };
+            _renderingThread        = new Thread(RenderLoop, 2 * 1024 * 1024) { Name = "GUI.RenderThread" };
             _lastCursorMoveTime     = Stopwatch.GetTimestamp();
             _glLogLevel             = ConfigurationState.Instance.Logger.GraphicsDebugLevel;
             _topLevel               = topLevel;
--- a/src/Ryujinx.Graphics.Vulkan/ShaderCollection.cs
+++ b/src/Ryujinx.Graphics.Vulkan/ShaderCollection.cs
@@ -53,7 +53,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         private ProgramPipelineState _state;
         private DisposableRenderPass _dummyRenderPass;
-        private Task _compileTask;
+        private ShaderCompilationRequest _compileRequest;
         private bool _firstBackgroundUse;
 
         public ShaderCollection(
@@ -119,7 +119,7 @@ namespace Ryujinx.Graphics.Vulkan
             ClearSegments = BuildClearSegments(resourceLayout.Sets);
             BindingSegments = BuildBindingSegments(resourceLayout.SetUsages);
 
-            _compileTask = Task.CompletedTask;
+            _compileRequest = new ShaderCompilationRequest(Task.CompletedTask);
             _firstBackgroundUse = false;
         }
 
@@ -133,7 +133,9 @@ namespace Ryujinx.Graphics.Vulkan
         {
             _state = state;
 
-            _compileTask = BackgroundCompilation();
+            _compileRequest = gd.ShaderCompilationQueue != null
+                ? gd.ShaderCompilationQueue.Add(BackgroundCompilation)
+                : new ShaderCompilationRequest(BackgroundCompilationAsync());
             _firstBackgroundUse = !fromCache;
         }
 
@@ -252,10 +254,25 @@ namespace Ryujinx.Graphics.Vulkan
             return segments;
         }
 
-        private async Task BackgroundCompilation()
+        private async Task BackgroundCompilationAsync()
         {
             await Task.WhenAll(_shaders.Select(shader => shader.CompileTask));
 
+            BackgroundCompilationImpl();
+        }
+
+        private void BackgroundCompilation()
+        {
+            foreach (var shader in _shaders)
+            {
+                shader.CompileTask.Wait();
+            }
+
+            BackgroundCompilationImpl();
+        }
+
+        private void BackgroundCompilationImpl()
+        {
             if (_shaders.Any(shader => shader.CompileStatus == ProgramLinkStatus.Failure))
             {
                 LinkStatus = ProgramLinkStatus.Failure;
@@ -397,11 +414,11 @@ namespace Ryujinx.Graphics.Vulkan
                     }
                 }
 
-                if (!_compileTask.IsCompleted)
+                if (!_compileRequest.IsCompleted)
                 {
                     if (blocking)
                     {
-                        _compileTask.Wait();
+                        _compileRequest.Wait();
 
                         if (LinkStatus == ProgramLinkStatus.Failure)
                         {
--- a/src/Ryujinx.Graphics.Vulkan/ShaderCompilationQueue.cs
+++ b/src/Ryujinx.Graphics.Vulkan/ShaderCompilationQueue.cs
@@ -7,7 +7,7 @@ namespace Ryujinx.Graphics.Vulkan
     class ShaderCompilationQueue
     {
         private const int MaxParallelCompilations = 8;
-        private const int MaxThreadStackSize = 1 * 1024 * 1024; // MB
+        private const int MaxThreadStackSize = 2 * 1024 * 1024; // MB
 
         private struct Request
         {
--- a/src/Ryujinx.Graphics.Vulkan/ShaderCompilationQueue.cs
+++ b/src/Ryujinx.Graphics.Vulkan/ShaderCompilationQueue.cs
@@ -1,0 +1,131 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    class ShaderCompilationQueue
+    {
+        private const int MaxParallelCompilations = 8;
+        private const int MaxThreadStackSize = 1 * 1024 * 1024; // MB
+
+        private struct Request
+        {
+            public readonly ulong Id;
+            public readonly Action Callback;
+
+            public Request(ulong id, Action callback)
+            {
+                Id = id;
+                Callback = callback;
+            }
+        }
+
+        private readonly Thread[] _workerThreads;
+        private readonly CancellationTokenSource _cts;
+        private readonly BlockingCollection<Request>[] _queues;
+        private readonly ulong[] _finishedIds;
+        private ulong _currentId;
+        private int _currentQueueIndex;
+
+        public ShaderCompilationQueue()
+        {
+            _workerThreads = new Thread[MaxParallelCompilations];
+            _queues = new BlockingCollection<Request>[MaxParallelCompilations];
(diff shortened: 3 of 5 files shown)


Foul-Tarnished:
Does this new custom thread pool is only for MacOS, or it's for all platforms?
I hope it won't affect other platform, or at least be sure the perf hit is insignificant on modern pc

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -134,7 +134,7 @@ internal class AppHost
             _inputManager           = inputManager;
             _accountManager         = accountManager;
             _userChannelPersistence = userChannelPersistence;
-            _renderingThread        = new Thread(RenderLoop, 1 * 1024 * 1024) { Name = "GUI.RenderThread" };
+            _renderingThread        = new Thread(RenderLoop, 2 * 1024 * 1024) { Name = "GUI.RenderThread" };
```
marysaka, line 137: Do the same change in SDL2 headless and GTK for consistency.

gdkchan:
> Does this new custom thread pool is only for MacOS, or it's for all platforms? I hope it won't affect other platform, or at least be sure the perf hit is insignificant on modern pc

It is only used on macOS, nothing should change for other platforms.

src/Ryujinx.Ava/AppHost.cs
gdkchan: This should probably be in a common place, but I'm not sure where to put it. It will be annoying to change 4 files every time we want to change the stack size.

gdkchan:
> > It is only used on macOS, nothing should change for other platforms.
> 
> Maybe it could be nice to notice that somewhere in the code or at least add an OS check (even if it's useless) just in case?

It already has both the OS check and a comment explaining why it is done:
```cs
            if (OperatingSystem.IsMacOS())
            {
                MVKInitialization.Initialize();

                // Any device running on MacOS is using MoltenVK, even Intel and AMD vendors.
                IsMoltenVk = true;

                // The default thread stack size on MacOS is low, and can cause stack overflow
                // on SPIR-V Cross during shader compilation.
                // As a workaround, we use this custom queue which allows us to specify the stack
                // size of the threads used for compilation.
                ShaderCompilationQueue = new ShaderCompilationQueue();
            }
 ```

**Review by AcK77 — APPROVED** [review]
Oh yeah the check is already in the existing code, sorry! LGTM then (In cases there is no performance issue in the shader compilation)!

File: src/Ryujinx.Ava/AppHost.cs (hunk 1 above)
Comment (gdkchan) [reply src/Ryujinx.Ava/AppHost.cs:137]: Ideally I think ideally, the render thread should be created by the `ThreadedRenderer` and the GPU thread should be created on the `GpuContext`. But it's not that simple since the OpenGL context is tied to the thread it is created on, and there's also some coupling with the UI and a few places that needs to know if the current thread is the "render" thread. Not something that I will change here though, so I guess I will just increase the stack size of the other threads for now.

Comment (patrick-hovsepian) [inline src/Ryujinx.Ava/AppHost.cs:137]: For the sake of readability, it might be nice to make it a named constant - either the whole stack size or the multiplier

**Comment by patrick-hovsepian** [comment]
> An alternative to this would passing some value to the backend that indicates the amount of nesting the shader has. The backend would use that value to decide if it should use Tasks to compile shaders in parallel, or if it should just compile it on the render thread (which already has an increased stack size).

This sounds cleaner to me at face value. Granted, I have very little experience with gpu code myself but if it's something that I'd like to take a crack at. Just not sure if priority/timing is a concern as I'd probably fumble through it.

**Comment by riperiperi** [comment]
I really don't think that's cleaner, the level of nesting it would split at is arbitrary, and the information is not useful for most platforms. An alternate task queue changes the least amount of shared code.

**Comment by patrick-hovsepian** [comment]
It seems to me that using framework managed Tasks would be cleaner than hand rolling custom thread management. You both know the space way better than I do though 😅

**Comment by Shihta** [comment]
hope it's not a stupid question... if it's just because the stack size is too small, why not just increase it?

refer to https://github.com/dotnet/fsharp/issues/9231#issuecomment-658297597

`export COMPlus_DefaultStackSize=800000` and then start Ryujinx, Splatoon 3 seems to start as normal

**Comment by gdkchan** [comment]
> hope it's not a stupid question... if it's just because the stack size is too small, why not just increase it?
> 
> refer to [dotnet/fsharp#9231 (comment)](https://github.com/dotnet/fsharp/issues/9231#issuecomment-658297597)
> 
> `export COMPlus_DefaultStackSize=800000` and then start Ryujinx, Splatoon 3 seems to start as normal

I tried that on macOS some time ago and got a crash, maybe I did something wrong. Anyway, it is undesirable to tell users to set an environment variable before running the emulator, so we need a solution that doesn't require any user intervention.

**Comment by Shihta** [comment]
> > hope it's not a stupid question... if it's just because the stack size is too small, why not just increase it?
> > refer to [dotnet/fsharp#9231 (comment)](https://github.com/dotnet/fsharp/issues/9231#issuecomment-658297597)
> > `export COMPlus_DefaultStackSize=800000` and then start Ryujinx, Splatoon 3 seems to start as normal
> 
> I tried that on macOS some time ago and got a crash, maybe I did something wrong. Anyway, it is undesirable to tell users to set an environment variable before running the emulator, so we need a solution that doesn't require any user intervention.

environment variable can be set in `Ryujinx.app/Contents/Info.plist`, refer to [here](https://stackoverflow.com/questions/16184505/set-environment-variable-for-the-process-before-startup)
so we can add these:
```
    <key>LSEnvironment</key>
    <dict>
        <key>COMPlus_DefaultStackSize</key>
        <string>800000</string>
    </dict>
```
in this way, users can start Ryujinx in the same way (simply click on the icon)

gdkchan:
> > > hope it's not a stupid question... if it's just because the stack size is too small, why not just increase it?
> > > refer to [dotnet/fsharp#9231 (comment)](https://github.com/dotnet/fsharp/issues/9231#issuecomment-658297597)
> > > `export COMPlus_DefaultStackSize=800000` and then start Ryujinx, Splatoon 3 seems to start as normal
> > 
> > 
> > I tried that on macOS some time ago and got a crash, maybe I did something wrong. Anyway, it is undesirable to tell users to set an environment variable before running the emulator, so we need a solution that doesn't require any user intervention.
> 
> environment variable can be set in `Ryujinx.app/Contents/Info.plist`, refer to [here](https://stackoverflow.com/questions/16184505/set-environment-variable-for-the-process-before-startup) so we can add these:
> 
> ```
>     <key>LSEnvironment</key>
>     <dict>
>         <key>COMPlus_DefaultStackSize</key>
>         <string>800000</string>
>     </dict>
> ```
> 
> in this way, users can start Ryujinx in the same way (simply click on the icon)

Nice, I didn't know you could do that. Still, maybe increasing the stack size of all threads is a bit excessive (but it would have the advantage that we can match the default stack size of other platforms, which is usually higher than macOS).

> I am wondering if this is a reasonable alternative [master...sunshineinabox:Ryujinx:AltSplat](https://github.com/Ryujinx/Ryujinx/compare/master...sunshineinabox:Ryujinx:AltSplat)
> 
> would be better than changing stacksize for all threads.

Isn't this basically what this PR does. Although the implementation doesn't look correct, it creates new threads every time a new shader needs to be compiled which I guess is not going to be very efficient. Also it waits for the threads to exit right after which makes it kinda pointless. It should only need to wait when the shaders are actually used.

Shihta:
> > environment variable can be set in `Ryujinx.app/Contents/Info.plist`, refer to [here](https://stackoverflow.com/questions/16184505/set-environment-variable-for-the-process-before-startup) so we can add these:
> > ```
> >     <key>LSEnvironment</key>
> >     <dict>
> >         <key>COMPlus_DefaultStackSize</key>
> >         <string>800000</string>
> >     </dict>
> > ```
> > 
> >   
> > in this way, users can start Ryujinx in the same way (simply click on the icon)
> 
> Nice, I didn't know you could do that. Still, maybe increasing the stack size of all threads is a bit excessive (but it would have the advantage that we can match the default stack size of other platforms, which is usually higher than macOS).
> 

OK! I think I could look into how to add this to the CI process
I'll send PR when I'm done

gdkchan:
Superseded by #5349.

Issue #2116, opened 2021-03-16
Sound crackling and/or clipping occurs in some games - audio backend specific


EmulationFanatic:
### What's the issue you encountered?

In some games, one or both audio backends suffer from audio crackling/static, or clipping. This can happen either during nvdec cutscenes or regular gameplay. See below for a few examples of each scenario.

### How can the issue be reproduced?

**Examples where OpenAL exhibits issues but SoundIO is fine:**
- Shantae and the Pirate's Curse
- Vitamin Connection
- Dragon Marked for Death
- Yoshi's Crafted World
- Romancing SaGa 2

**Examples where SoundIO exhibits issues but OpenAL is fine:**
- Kamari Damacy Reroll
- Bayonetta
- Valkyria Chronicles 4
- Tokyo Mirage Sessions #FE Encore
- Darksiders Warmastered Edition

**Examples where both audio backends exhibit issues:**
- The Witcher 3 (during cutscenes)
- Need for Speed Hot Pursuit Remastered (everywhere on OpenAL, in-game on SoundIO)
- Kine

### Environment?

- Ryujinx version: current/recent versions
- Game version: N/A
- System Specs:
  - OS: Windows 10 20H2
  - CPU: Ryzen 7 5800X
  - GPU: NVIDIA GTX 1080
  - RAM: 32GB
- Applied Mods : No

### Additional context?

N/A

Shahil-Ayato:
@EmulationFanatic I believe this also happens on OpenAL with Fire emblem Three houses when you talk with the class. Unsure about SoundIO

EmulationFanatic:
@Shahil-Ayato there are dozens of games that exhibit these issues. I just listed a few confirmed current examples for testing & verification purposes.

Doctorwho1909:
![image](https://user-images.githubusercontent.com/64975148/111392022-86d01e00-8694-11eb-8627-27279a7d8d4e.png)
happens with this game also openal is all broken and soundio is fine

tvoracek:
Sphinx and the Cursed Mummy is also affected

Egyember:
this aloso happening in Legen of Zelda: Tears of the Kingdom.
(don't ask where I got it)

jzatarain:
This also happens with persona 5 tactica.
I changed the audio backend, and it fixed the issue.

Taobis:
[...]

PR #5292, opened 2023-06-12
Fix Arm32 double to int/uint conversion on Arm64

gdkchan:
This fixes the implementation of the VCVT Arm32 instruction, when doing conversion from double to int or uint (32-bit) on Arm64. The problem is that it was using the Arm64 scalar conversion instructions, but Arm64 only does double to long/ulong or float to int/uint conversion. Meanwhile, Arm32 does double to int/uint and float to int/uint. To fix the issue, this uses the instruction that puts the value on a GPR instead of SIMD register, since those instructions does allow to specify the destination type as well. A similar thing was already being done for the conversion instruction that rounds towards zero.

This fixes bad audio on Prinny Presents NIS Classics Volume 3: La Pucelle: Ragnarok / Rhapsody: A Musical Adventure on Arm64 platforms.

### DIFF
--- a/src/ARMeilleure/Instructions/InstEmitSimdCvt32.cs
+++ b/src/ARMeilleure/Instructions/InstEmitSimdCvt32.cs
@@ -165,7 +165,7 @@ namespace ARMeilleure.Instructions
                     {
                         Operand m = GetVecA32(op.Vm >> 1);
 
-                        Operand toConvert = InstEmitSimdHelper32Arm64.EmitExtractScalar(context, m, op.Vm, doubleSize);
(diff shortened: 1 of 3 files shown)


marysaka (approved):

riperiperi (approved):
Nice catch.

blisstik:
Any chance #2116 has a similar issue?

gdkchan:
> Any chance #2116 has a similar issue?

No, the change on this PR only affects 32-bit games on the Arm64 backend. This issue pre-dates the Arm64 backend.

Issue #5273, opened 2023-06-09
[Bug] Improper filename check?


pubct:
### Description of the issue

In `/Roaming/Ryujinx/sdcard/atmosphere/contents`, let me use an imaginary `ApplicationId`, say `001a0011`.

When,
`01a0011` (remove beginning),
`001a001` (remove end),
`e001a0011` (add beginning),
the mod doesn't load. Which are correctly what happened, as they don't match..

When `001a0011e` (add end) however, the mod loads. Which I think **shouldn't** happen.

### Reproduction steps

1. Right click game
2. Select `Open Atmosphere Mod Directory`
3. Throw in any mods
4. Go back one folder level
5. Change the folder name
6. Launch game

### Log file

 Not needed I believe

### OS

Windows 10

### Ryujinx version

1.1.881

AcK77:
Well, the folder you put in `content` should be the `ApplicationId` (and only that) which is unique by game, so it make sense if the mods isn’t detected  by Ryujinx as you have changed it by add/remove a char at the beginning or the end of the folder name.
Or I have misunderstood what you explained, in that case, add some examples please.

pubct:
I'm not at my PC at the moment. So, let me use an imaginary `ApplicationId`. Say `001a0011`.

When,
`01a0011` (remove beginning),
`001a001` (remove end),
`e001a0011` (add beginning),
the mod doesn't load. Which is correctly what should happen, as they don't match..

When `001a0011e` (add end) however, the mod loads. Which I think **shouldn't** happen.

TSRBerry:
This would have been nice to include in the initial description as I wasn't sure what you meant until you added that.

pubct:
Yeah. My bad. Will remember this for next time.

PR #5298, opened 2023-06-12
Mod Loader: Stop loading mods from folders that don't exactly match titleId

Kurochi51:
This addresses issue #5273.
Not quite sure if there was any reasoning behind the wildcard, I looked in the file but I couldn't see anything that would explicitly need it.

### DIFF
--- a/src/Ryujinx.HLE/HOS/ModLoader.cs
+++ b/src/Ryujinx.HLE/HOS/ModLoader.cs
@@ -154,7 +154,7 @@ namespace Ryujinx.HLE.HOS
         }
 
         private static DirectoryInfo FindTitleDir(DirectoryInfo contentsDir, string titleId)
-            => contentsDir.EnumerateDirectories($"{titleId}*", DirEnumOptions).FirstOrDefault();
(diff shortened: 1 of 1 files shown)


src/Ryujinx.HLE/HOS/ModLoader.cs
```diff
@@ -154,7 +154,7 @@ private static string EnsureBaseDirStructure(string modsBasePath)
         }
 
         private static DirectoryInfo FindTitleDir(DirectoryInfo contentsDir, string titleId)
-            => contentsDir.EnumerateDirectories($"{titleId}*", DirEnumOptions).FirstOrDefault();
+            => contentsDir.EnumerateDirectories($"{titleId}", DirEnumOptions).FirstOrDefault();
```
AcK77, line 157: `titleId` is already a string, so the string interpolation isn’t needed anymore!

Kurochi51: I figured it doesn't hurt, but sure thing :D

AcK77 (approved):

TSRBerry (approved):
Lgtm, thank you!

PR #5299, opened 2023-06-12
infra: Fix action version for triage workflow

TSRBerry:
For some reason this doesn't work with v1 alone, although other actions do.

### DIFF
--- a/.github/workflows/pr_triage.yml
+++ b/.github/workflows/pr_triage.yml
@@ -16,43 +16,43 @@ jobs:
           dot: true
 
       - name: Auto Assign [Audio]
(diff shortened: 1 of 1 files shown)


Kurochi51:
That's because `kentaro-m/auto-assign-action` doesn't have a v1, you can see them all on the marketplace.

marysaka (approved):

TSRBerry:
As far as I understand you should be able to only reference the major version of an action and it would pull the latest version in:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses

Issue #4863, opened 2023-05-08
[Bug] Intel iGPU Iris


DarrenWarrenV:
### Description of the issue

So I have a laptop with i5-1155G7 and an Iris iGPU. I checked the last good version where Tony Hawk's Pro Skater 1+2 is running without graphical glitches. I made just two screenshots from main menu, but further on - for example all the SKATE later are white, skater's shadowing is screwed, in Hangar II billboards are very dark. Basically it's a regression.

Last good version: 1.1.732
First bad version: 1.1.733

Changelog from 1.1.733:

1.1.733 - 2023-04-25

Fixed:

 *Refactor attribute handling on the shader generator.
-Refactors shader translator code to better handle "attributes".
-Fixes compilation failures for tessellation shaders on MoltenVK, which would cause some games to render nothing except the UI on macOS.
-Fixes an issue on older GPUs that don't support setting gl_Layer on the vertex shader should be fixed. This allows Dragon Quest 3 and possibly other games to render on older GPUs.

I have no idea which change did this, but from this version of emulator (onwards) it's the same.

[1.1.732](https://ibb.co/y8ttRgJ)

[1.1.733](https://ibb.co/bmV6fnb)

### Reproduction steps

Just boot up game to the main menu, you can also try any stage.

### Log file

Shit, I don't have any... sorry guys! I already packed my laptop as I'm switching to a bit more powerful one and I'm sending this one back.

### OS

Windows 11

### Ryujinx version

1.1.732 - 1.17XX

### Game version

1.0.3

### CPU

i5-1155G7

### GPU

Iris

### RAM

16GB

### List of applied mods

None

AcK77:
Could you check if it's still happen in latest version ?

Issue #5251, opened 2023-06-06
[Bug] Strange thing happening to 3d Models in pokemon Scarlet starting from 1.1733


AreoGit:
### Description of the issue

Starting from version 1.1733 all pokemon and characters look strange / washed out. 
The first screenshot shows how it should be and was before 1.1733. The second screenshot is how it is now up until the newest version. I can post more examples  if necessary.

![Screenshot 2023-06-06 202438](https://github.com/Ryujinx/Ryujinx/assets/1416943/946bf4c9-db88-4181-9333-caa9573f9bf4)
![Screenshot 2023-06-06 202621](https://github.com/Ryujinx/Ryujinx/assets/1416943/06095480-df2d-4889-92ec-4f20d35eef31)

As said before the issue is introduced in 1.1733. Version 1.1732 does not have this problem. It is version 1.03 of Pokemon Scarlet.
Clean install with wiping the folder in AppData does not resolve the problem so it was a change introduced in the code in 1.1733.

Machine specs:

i7-1255u Intel Irix XE (96EU) with 16GB of RAM.

No alterations made to stock Ryujinx settings (so Vulkan is used).

### Reproduction steps

Start Pokemon Scarlet with any Ryujinx version newer then 1.1733 on a Intel Irix XE gpu.

### Log file

00:00:00.038 |I| Configuration LogValueChange: ResScale set to: 1
00:00:00.042 |I| Configuration LogValueChange: ResScaleCustom set to: 1
00:00:00.042 |I| Configuration LogValueChange: MaxAnisotropy set to: -1
00:00:00.042 |I| Configuration LogValueChange: AspectRatio set to: Fixed16x9
00:00:00.043 |I| Configuration LogValueChange: BackendThreading set to: Auto
00:00:00.043 |I| Configuration LogValueChange: GraphicsBackend set to: Vulkan
00:00:00.044 |I| Configuration LogValueChange: PreferredGpu set to: 0x8086_0x46A8
00:00:00.044 |I| Configuration LogValueChange: AntiAliasing set to: None
00:00:00.045 |I| Configuration LogValueChange: ScalingFilter set to: Bilinear
00:00:00.045 |I| Configuration LogValueChange: ScalingFilterLevel set to: 80
00:00:00.046 |I| Configuration LogValueChange: EnableDockedMode set to: True
00:00:00.052 |I| Configuration LogValueChange: EnableVsync set to: True
[...]

PR #5311, opened 2023-06-14
Fix regression introduced by 1.1.1733 on Intel GPUs

Kurochi51:
Requires testing, and should address #5251 if actually the cause.
No idea if this is the actual issue, but when 1.1733 was released and `gl_FrontFacing` was moved out of the backend along with the attribute refactor, the workaround that was in place to fix a bug on Intel GPUs seems to have been altered. This should, mirror the behaviour, but alas I have no way to test it.


src/Ryujinx.Graphics.Shader/Instructions/InstEmitAttribute.cs
```diff
@@ -166,7 +166,7 @@ public static void Ipa(EmitterContext context)
                 {
                     // gl_FrontFacing sometimes has incorrect (flipped) values depending how it is accessed on Intel GPUs.
                     // This weird trick makes it behave.
-                    res = context.ICompareLess(context.INegate(context.IConvertS32ToFP32(res)), Const(0));
+                    res = context.ICompareLess(context.INegate(context.FP32ConvertToS32(res)), Const(0));
```
gdkchan, line 169: This is wrong since the value is not FP32. It should be `context.ConditionalSelect(res, ConstF(1f), ConstF(0f))`.

AreoGit:
The issue persists so this is unfortunately not the solution. I do get several pipeline errors in the log now (see attached file).

[Ryujinx_1.1.0+faddbdf_2023-06-15_09-14-57.log](https://github.com/Ryujinx/Ryujinx/files/11754322/Ryujinx_1.1.0%2Bfaddbdf_2023-06-15_09-14-57.log)

Kurochi51:
out of curiosity, does that warning persist in opengl? also can you try without a shader cache enabled on vulkan?

AreoGit:
The warning is only visible on Vulkan, also with shader cache disabled (see log). The issue remains the same.
OpenGL does not give this warning but is a complete mess graphically (see attached files). Max window or full screen also does not work on opengl.

[opengl_no_shader_cache.log](https://github.com/Ryujinx/Ryujinx/files/11757719/opengl_no_shader_cache.log)
[vulkan_no_shadercache.log](https://github.com/Ryujinx/Ryujinx/files/11757720/vulkan_no_shadercache.log)
Screenshot 2023-06-15 145317
Screenshot 2023-06-15 145230

Kurochi51:
I'm not quite sure what's happening, why the pipeline state would be incorrect, and I can't really tell what's what. Check to see if the latest change..... does anything when it's done building, otherwise I can't tell what's going on.

AreoGit:
@Kurochi51 you fixed this bug :). The warning is till there (not in 1.1.1732) but the graphic issue is resolved on Vulkan. So thank you for solving the graphic bug. Do you want me to register the warning or the OpenGL graphic issues as new bugs?

[Ryujinx_1.1.0+88d4b2e_2023-06-15_21-09-01.log](https://github.com/Ryujinx/Ryujinx/files/11761757/Ryujinx_1.1.0%2B88d4b2e_2023-06-15_21-09-01.log)

AreoGit (approved):
Fixes the graphical bug on Vulkan for Intel GPU

Kurochi51:
are the issues on OpenGL still there? the rainbow hair and fullscreen issue

AreoGit:
Yes and additionally on OpenGL also some distorted sound
Screenshot 2023-06-15 212015

MetrosexualGarbodor:
Wasn't it always like this on Intel OpenGL? Not that it matters much, since Intel OpenGL on Windows is basically unsupported

AreoGit:
Could be, I never tried it with OpenGL before.

Kurochi51:
oh, so these OpenGL issues are unrelated?

MetrosexualGarbodor:
@AreoGit can you please check? https://github.com/Ryujinx/release-channel-master/releases/tag/1.1.732

AreoGit:
I now verified with version 1.1.1732. This version has the same OpenGL issues so my guess is it is unrelated.

gdkchan:
Any change to the generated shader code requires a shader cache version change. It should be set to the PR number on https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs#L25 otherwise the shader cache will not be rebuilt.

Also this seems a bit different from the previous work-around which is a bit weird, but technically it's not *incorrect*.

Kurochi51:
I couldn't really say, since I don't quite understand from just reading the code what's exactly going on here, but it seems to work without breaking other stuff.

riperiperi (approved):
Seems reasonable. I'm not sure if the latest drivers even need this workaround, but keeping it around is good for compatibility (obviously only if works).

Kurochi51:
It worked for the submitter of the issue, so there's that 😅

gdkchan (approved):
Looks fine.

gdkchan:
I think it should also fix #4863, but I did not test it.

PR #5313, opened 2023-06-15
Fix multiple enumerations in loops

ACGNnsj:
The possible additional enumerations here might be costly, so I convert the IEnumerable objects to lists.


kekonn:
This is just my two cents, but arrays are even faster if you're worried about micro optimizations. And they take up less space in memory as well.

Also: 'possible' multiple enumerations don't cost anything. Only 'actual' multiple enumerations *can* have an overhead depending on the backing collection. In this case, it only makes sense to me to convert the `target` collection.

But again, just my two cents. I am definitely curious to hear what the reviewers will say about this.

patrick-hovsepian:
This actually introduces the problem you're trying to solve. You can force the enumeration by calling `ToList`, which you've done. The list is then enumerated again, when you run the `foreach`. The framework is really good about deferring evaluation when needed. The enumeration takes place when `foreach` is invoked.`List` implements `IEnumerable` https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=net-7.0

ACGNnsj:
My bad. In both cases, GetEnumerator won't be called unnecessarily.

patrick-hovsepian:
No worries! It's a good thought - I like where your head is at 😺 

IEnumerable semantics aren't immediately obvious without digging into a bit

PR #5319, opened 2023-06-16
Null pointers should not be dereferenced

marco-carvalho:


src/Ryujinx.Horizon/Sdk/Sf/Cmif/ServerDomainManager.cs
```diff
@@ -54,7 +54,10 @@ public void FreeEntry(Entry entry)
             {
                 lock (_freeList)
                 {
-                    DebugUtil.Assert(entry.Owner == null);
+                    if (entry != null)
```
gdkchan, line 57: This looks rather weird. Also it accesses `entry.Obj` right below, so if it was null, it would throw there anyway, so not sure what this check adds. Perhaps it should first assert that entry itself is not null, if it's going to use it and there's the possibility of it being null. Or maybe it should just throw `ArgumentNullException` at the start of the method if it is null. Though I think in general, we don't validate method parameters, they are assumed to be correct.

marco-carvalho: You were right, I added `entry.Obj` inside the `if` statement.

src/Ryujinx.Ava/UI/ViewModels/ControllerInputViewModel.cs
```diff
@@ -773,7 +773,10 @@ public async void SaveProfile()
                         config = (Configuration as InputConfiguration<GamepadInputId, ConfigStickInputId>).GetConfig();
                     }
 
-                    config.ControllerType = Controllers[_controller].Type;
+                    if (config != null)
```
AcK77, line 776: In our current case, if `IsKeyboard` is false then `IsController` is true, or the invert. So `config` can't be null here.
I'd prefer, at least, to change the above `else if (IsController)` into `else // if (IsController)` if we want something cleaner.
But the null check isn't really useful here.

src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
```diff
@@ -461,17 +461,23 @@ public void LoadShaders(GpuContext context, ParallelDiskCacheLoader loader)
             }
 
             int offset = Unsafe.SizeOf<TocHeader>() + programIndex * Unsafe.SizeOf<OffsetAndSize>();
-            if (offset + Unsafe.SizeOf<OffsetAndSize>() > tocFileStream.Length)
+            if (tocFileStream != null)
```
AcK77, line 464: Above this code there is `if (tocFileStream == null ...` if the condition is reached then `tocFileStream` is assigned. If there is something wrong in the method called to assign it, then I'm pretty sure it will throw something. I don't see why we should add two `if (tocFileStream != null)` here.

src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/InlineResponses.cs
```diff
@@ -32,7 +32,10 @@ private static uint WriteString(string text, BinaryWriter writer, uint maxSize,
             }
 
             writer.Write(bytes);
-            writer.Seek((int)maxSize - bytes.Length, SeekOrigin.Current);
+            if (bytes != null)
+            {
+                writer.Seek((int)maxSize - bytes.Length, SeekOrigin.Current);
+            }
```
AcK77, line 38:
```suggestion

            if (bytes != null)
            {
                writer.Seek((int)maxSize - bytes.Length, SeekOrigin.Current);
            }

```

but this condition do any sense. There is `writer.Write(bytes);` just above, if `bytes` where `null` it will reach a NRE here, so the check will change anything.

src/Ryujinx.HLE/HOS/Services/Nv/NvDrvServices/NvHostAsGpu/NvHostAsGpuDeviceFile.cs
```diff
@@ -111,7 +111,10 @@ private NvInternalResult BindChannel(ref BindChannelArguments arguments)
                 // TODO: Return invalid Fd error.
             }
```
AcK77, line 113:
```suggestion
            else
            {
                channelDeviceFile.Channel.BindMemory(_asContext.Gmm);
            }
```

Or maybe resolve the `TODO` should be enough.

src/Ryujinx.HLE/HOS/Services/Nv/NvDrvServices/NvHostAsGpu/NvHostAsGpuDeviceFile.cs
```diff
@@ -111,7 +111,10 @@ private NvInternalResult BindChannel(ref BindChannelArguments arguments)
...
 
-            channelDeviceFile.Channel.BindMemory(_asContext.Gmm);
+            if (channelDeviceFile != null)
+            {
+                channelDeviceFile.Channel.BindMemory(_asContext.Gmm);
+            }
```
AcK77, line 117: Not needed, read the above comment.

src/Ryujinx.Horizon/Sdk/Sf/Cmif/ServerDomainManager.cs
AcK77: I agree with gdkchan, the check here isn't needed, even more because it's a debug assert too (same for all other checks).

AcK77 (changes requested):
There are really some strange changes here. Some null checks could be useful, I'm sorry, but for some others I wonder if you readed the code before you did things.

src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/InlineResponses.cs
AcK77: > but this condition do any sense. There is writer.Write(bytes); just above, if bytes where null it will reach a NRE here, so the check will change anything.

This is not resolved!

AcK77:
Could you please fixes old opened PR like this one before opening new ones?
Thanks.

PR #5325, opened 2023-06-17
MemoryManagement: Change return types for Commit/Decommit to void

TSRBerry:
This PR makes the memory mangement behavior for `Commit()` and `Decommit()` between Unix and Windows consistent again.

This behavior has been inconsistent since I added `SystemException`s  should these methods fail to Unix only.
See: #4702

Now both implementations throw `SystemException`s and the return types were changed to `void` since we always return `true` if they don't fail.

### DIFF
--- a/src/Ryujinx.Memory/MemoryManagementUnix.cs
+++ b/src/Ryujinx.Memory/MemoryManagementUnix.cs
@@ -68,7 +68,7 @@ namespace Ryujinx.Memory
             return ptr;
         }
 
-        public static bool Commit(IntPtr address, ulong size, bool forJit)
+        public static void Commit(IntPtr address, ulong size, bool forJit)
         {
             MmapProts prot = MmapProts.PROT_READ | MmapProts.PROT_WRITE;
 
@@ -81,11 +81,9 @@ namespace Ryujinx.Memory
             {
                 throw new SystemException(Marshal.GetLastPInvokeErrorMessage());
             }
-
-            return true;
         }
 
-        public static bool Decommit(IntPtr address, ulong size)
+        public static void Decommit(IntPtr address, ulong size)
         {
             // Must be writable for madvise to work properly.
             if (mprotect(address, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE) != 0)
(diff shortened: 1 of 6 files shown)


src/Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -12,7 +10,7 @@ namespace Ryujinx.Memory
     [SupportedOSPlatform("macos")]
     static class MemoryManagementUnix
     {
-        private static readonly ConcurrentDictionary<IntPtr, ulong> _allocations = new ConcurrentDictionary<IntPtr, ulong>();
+        private static readonly ConcurrentDictionary<IntPtr, ulong> Allocations = new();
```
gdkchan, line 13: Is the naming  style for private static fields specified anywhere? And is it something that can be enforced with #4670? I'm never sure if they should be named `_likeThis` or `LikeThis`.

gdkchan (approved):
Makes sense. I had some access violation crashes while trying to fill guest heap memory in the past, likely caused by Commit failing on  Windows. It's better to just throw the exception up front than ignoring the error and crashing later.

src/Ryujinx.Memory/MemoryManagementUnix.cs
TSRBerry: There is no specification for this currently afaik. The .NET naming [guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields) exclude private static fields if I understand correctly.

image
Rider defaults to UpperCamelCase for them, so we should create an `.editorconfig` rule for these.

We can't really enforce the naming style yet, but we could if we write an analyzer for that. We might need to do that anyway since we have a lot of classes in `Ryujinx.HLE` starting with "I" even though they aren't interfaces, which is something we can't allow with dotnet naming rules currently.

AcK77: I’d prefer to keep `_name` for private field, even if it’s readonly or static.

AcK77 (approved):

src/Ryujinx.Memory/MemoryManagementUnix.cs
gdkchan: I'd prefer that we follow whatever s the standard for C# projects. Other than that I don't really have a preference for this.

PR #5328, opened 2023-06-17
[Save Manager] Backup and Restore User Profile Save Data

patrick-hovsepian:
This is a continuation of #5194 

This PR introduces a mechanism to backup and restore the save date for a single user profile. The user can choose the folder to save to and the zip file to restore from. The operation applies to the save data of the current active profile. By default, that includes BCAT and Device data too which is needed for things like AC Horizons.

There are two new buttons added to the User Save for each function
image

On completion, a notification is shown. 
Screenshot 2023-06-17 at 3 02 36 PM

There's also a loading bar to show progress and the back button is disabled while the operation takes place
image

On restore, the following prompt is shown to the user to make it explicit that it's a destructive operation. 
image

The generated backup zip has the following hierarchy `[titleid]/[save_type]/` There's also a `tag.json` file that's created for user friendly metadata about the save. This could also be used by external tools for compatibility purposes. 
image

I've tested with cross user on the same device, cross device, and cross os (windows to mac)


MutantAura:
Functionality wise it's looking good. Was able to backup a set of saves from one profile and restore them to another with no issues. Progress bar seems a little misaligned though?
image

sakuka47:
I've tested the Pr, it works well, but is it possible to select the game(s) you want to make a backup of?

patrick-hovsepian:
@sakuka47 There are no plans to do that in this PR and I'm not sure if this will even get merged. A feature request might be your best bet.

patrick-hovsepian:
lack of developer response.

PR #5331, opened 2023-06-19
GetHashCode should not reference mutable fields

marco-carvalho:
`GetHashCode` is used to file an object in a `Dictionary` or `Hashtable`. If `GetHashCode` uses non-`readonly` fields and those fields change after the object is stored, the object immediately becomes mis-filed in the `Hashtable`. Any subsequent test to see if the object is in the `Hashtable` will return a false negative.


AcK77 (approved):
Should be fine I guess.

marysaka (approved):

PR #5332, opened 2023-06-19
Ava UI: Make some settings methods async

MutantAura:
Creating the settings window was being blocked by a few synchronous methods, the main timesink being the creation of the Vulkan device list which took over a second on my system. By making this check, audio backend checks, timezone creation and network interface population asynchronous, the window is able to open _much_ faster.

https://github.com/Ryujinx/Ryujinx/assets/44103205/bddf51c4-ecb7-401f-ac06-c521c5af2359

The first time the window is opened is still somewhat JIT limited, but subsequent openings will be massively better.


src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
         {
-            IsOpenAlEnabled = OpenALHardwareDeviceDriver.IsSupported;
-            IsSoundIoEnabled = SoundIoHardwareDeviceDriver.IsSupported;
-            IsSDL2Enabled = SDL2HardwareDeviceDriver.IsSupported;
+            await Task.Run(() => 
+            {
+                IsOpenAlEnabled = OpenALHardwareDeviceDriver.IsSupported;
+                IsSoundIoEnabled = SoundIoHardwareDeviceDriver.IsSupported;
+                IsSDL2Enabled = SDL2HardwareDeviceDriver.IsSupported;
+            });
```
patrick-hovsepian, line 313: I'm not familiar with the sound drivers but are these really hotspots in the code? Does this need to happen async? Would it make sense to use a discard and bake in the `OnPropertyChanged` events into the Action?

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
+                    {
+                        _gpuIds.Add(device.Id);
+                        _gpuNames.Add($"{device.Name} {(device.IsDiscrete ? "(dGPU)" : "")}");
+                    }
+                }
+            });
 
             AvailableGpus.Clear();
-            AvailableGpus.AddRange(names.Select(x => new ComboBoxItem { Content = x }));
+            AvailableGpus.AddRange(_gpuNames.Select(x => new ComboBoxItem { Content = x }));
```
patrick-hovsepian, line 342: Does it make sense to pull the clear up before the if/else. That way you can add the combox as you're iterating the devices anyway and not have to loop through it twice

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
+
+            OnPropertyChanged(nameof(PreferredGpuIndex));
         }
 
-        public void LoadTimeZones()
+        public async void LoadTimeZones()
         {
             _timeZoneContentManager = new TimeZoneContentManager();
 
             _timeZoneContentManager.InitializeInstance(_virtualFileSystem, _contentManager, IntegrityCheckLevel.None);
```
patrick-hovsepian, line 351: curious if you're aware of why this can't be done in the constructor so it's not repeated on every invocation

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
             {
-                int hours = Math.DivRem(offset, 3600, out int seconds);
-                int minutes = Math.Abs(seconds) / 60;
+                foreach ((int offset, string location, string abbr) in _timeZoneContentManager.ParseTzOffsets())
+                {
+                    int hours = Math.DivRem(offset, 3600, out int seconds);
+                    int minutes = Math.Abs(seconds) / 60;
 
-                string abbr2 = abbr.StartsWith('+') || abbr.StartsWith('-') ? string.Empty : abbr;
+                    string abbr2 = abbr.StartsWith('+') || abbr.StartsWith('-') ? string.Empty : abbr;
```
patrick-hovsepian, line 360: are you able to reassign to abbr?

patrick-hovsepian (commented):
nice improvement! some minor thoughts

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
riperiperi: The SDL driver can take some time to initialize, I think the GTK UI already defers this.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
-                    names.Add($"{device.Name} {(device.IsDiscrete ? "(dGPU)" : "")}");
+                    IsVulkanAvailable = false;
+                    GraphicsBackendIndex = 1;
                 }
-            }
+                else
+                {
+                    foreach (var device in devices)
+                    {
+                        _gpuIds.Add(device.Id);
```
riperiperi, line 335: This list is now accessed from multiple threads, so it there should be some lock around any time it's used, such as in `LoadCurrentConfiguration` and `SaveSettings`.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
MutantAura: Either way as far as UI goes I prefer the opposite mentality of "Does this need to block?" rather than "Does this need to happen async?". Realistically not much outside of the drawing of the windows themselves actually needs to be synchronous on the UI thread. All of the methods here related to settings that are not on the first navpage so I do believe should not be blocking, no matter how small their contribution is.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
+                foreach ((int offset, string location, string abbr) in _timeZoneContentManager.ParseTzOffsets())
+                {
+                    int hours = Math.DivRem(offset, 3600, out int seconds);
+                    int minutes = Math.Abs(seconds) / 60;
 
-                string abbr2 = abbr.StartsWith('+') || abbr.StartsWith('-') ? string.Empty : abbr;
+                    string abbr2 = abbr.StartsWith('+') || abbr.StartsWith('-') ? string.Empty : abbr;
 
-                TimeZones.Add(new TimeZone($"UTC{hours:+0#;-0#;+00}:{minutes:D2}", location, abbr2));
+                    TimeZones.Add(new TimeZone($"UTC{hours:+0#;-0#;+00}:{minutes:D2}", location, abbr2));
```
riperiperi, line 362: `AvaloniaList` doesn't appear to be thread safe. Maybe this should be modified via an invoke on the UI thread. (actually, maybe all the collection modifications should? after we get the expensive parts out of the way)

You can do this with `Dispatcher.UIThread.Post`.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +303,85 @@ public SettingsViewModel()
...
             _networkInterfaces.Add(LocaleManager.Instance[LocaleKeys.NetworkInterfaceDefault], "0");
 
-            foreach (NetworkInterface networkInterface in NetworkInterface.GetAllNetworkInterfaces())
+            await Task.Run(() =>
             {
-                _networkInterfaces.Add(networkInterface.Name, networkInterface.Id);
-            }
+                foreach (NetworkInterface networkInterface in NetworkInterface.GetAllNetworkInterfaces())
+                {
+                    _networkInterfaces.Add(networkInterface.Name, networkInterface.Id);
```
riperiperi, line 380: Same here, not thread safe. Modify with `Dispatcher.UIThread.Post` is probably a good shout, but do `NetworkInterface.GetAllNetworkInterfaces()` first outside the invoke.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
riperiperi: Changing it to use `Dispatcher.UIThread.Post` as described below may be easier.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
patrick-hovsepian: Again, I wasn't aware of the underlying semantics so figured that played into it some. It looks like a simple boolean assignment which seems like a code smell to wrap in a task. If these were method calls instead, the intent would be much clearer IMO

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
MutantAura: Can you check these again. I believe everything used by multiple methods should now be thread safe?

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
MutantAura: Simplified this logic also allowed me to remove _gpuNames entirely.

IsaacMarovitz (approved):
lgtm

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +301,100 @@ public SettingsViewModel()
...
-            IsSoundIoEnabled = SoundIoHardwareDeviceDriver.IsSupported;
-            IsSDL2Enabled = SDL2HardwareDeviceDriver.IsSupported;
+            await Task.Run(() => 
+            {
+                IsOpenAlEnabled = OpenALHardwareDeviceDriver.IsSupported;
+                IsSoundIoEnabled = SoundIoHardwareDeviceDriver.IsSupported;
+                IsSDL2Enabled = SDL2HardwareDeviceDriver.IsSupported;
+            });
+
+            OnPropertyChanged(nameof(IsOpenAlEnabled));
```
riperiperi, line 313: These might need to run in `Dispatcher.UIThread.Post`.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +301,100 @@ public SettingsViewModel()
...
+                }
+            });
 
-            AvailableGpus.Clear();
-            AvailableGpus.AddRange(names.Select(x => new ComboBoxItem { Content = x }));
+            // GPU configuration needs to be loaded after the async method or it will always return 0.
+            PreferredGpuIndex = _gpuIds.Contains(ConfigurationState.Instance.Graphics.PreferredGpu) ? 
+                                _gpuIds.IndexOf(ConfigurationState.Instance.Graphics.PreferredGpu) : 0;
+
+            OnPropertyChanged(nameof(PreferredGpuIndex));
```
riperiperi, line 349: Same here.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +301,100 @@ public SettingsViewModel()
...
 
-                TimeZones.Add(new TimeZone($"UTC{hours:+0#;-0#;+00}:{minutes:D2}", location, abbr2));
+                        _validTzRegions.Add(location);
+                    });
+                }
+            });
 
-                _validTzRegions.Add(location);
-            }
+            OnPropertyChanged(nameof(TimeZone));
```
riperiperi, line 376: Same here.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -300,65 +301,100 @@ public SettingsViewModel()
...
+                foreach (NetworkInterface networkInterface in allInterfaces)
+                {
+                    Dispatcher.UIThread.Post(() =>
+                    {
+                        _networkInterfaces.Add(networkInterface.Name, networkInterface.Id);
+                    });
+                }
+            });
+
+            OnPropertyChanged(nameof(NetworkInterfaceList));
```
riperiperi, line 397: Same here.

riperiperi (commented):
Looks good, just one more nit about calls to `OnPropertyChanged`.

I'm not sure what is listening for these property changes, I'm just assuming it is something internal to Avalonia and it might not be thread safe.

riperiperi (approved):
This is considerably better. Nice work!

TSRBerry (changes requested):
```
Unhandled exception caught: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at Avalonia.Collections.AvaloniaList`1.get_Item(Int32 index)
   at Ryujinx.Ava.UI.ViewModels.SettingsViewModel.set_NetworkInterfaceIndex(Int32 value) in /src/Ryujinx/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs:line 271
   at Ryujinx.Ava.UI.ViewModels.SettingsViewModel.LoadCurrentConfiguration() in /src/Ryujinx/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs:line 478
   at Ryujinx.Ava.UI.ViewModels.SettingsViewModel..ctor() in /src/Ryujinx/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs:line 299
   at Ryujinx.Ava.UI.ViewModels.SettingsViewModel..ctor(VirtualFileSystem virtualFileSystem, ContentManager contentManager) in /src/Ryujinx/src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs:line 275
   at Ryujinx.Ava.UI.Windows.SettingsWindow..ctor(VirtualFileSystem virtualFileSystem, ContentManager contentManager) in /src/Ryujinx/src/Ryujinx.Ava/UI/Windows/SettingsWindow.axaml.cs:line 19
   at Ryujinx.Ava.UI.Views.Main.MainMenuBarView.OpenSettings(Object sender, RoutedEventArgs e) in /src/Ryujinx/src/Ryujinx.Ava/UI/Views/Main/MainMenuBarView.axaml.cs:line 128
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
   at Avalonia.Threading.SendOrPostCallbackDispatcherOperation.InvokeCore()
   at Avalonia.Threading.DispatcherOperation.Execute()
   at Avalonia.Threading.Dispatcher.ExecuteJob(DispatcherOperation job)
   at Avalonia.Threading.Dispatcher.ExecuteJobsCore(Boolean fromExplicitBackgroundProcessingCallback)
   at Avalonia.Threading.Dispatcher.OnOSTimer()
   at Avalonia.X11.X11PlatformThreading.RunLoop(CancellationToken cancellationToken)
   at Avalonia.Threading.DispatcherFrame.Run(IControlledDispatcherImpl impl)
   at Avalonia.Threading.Dispatcher.PushFrame(DispatcherFrame frame)
   at Avalonia.Threading.Dispatcher.MainLoop(CancellationToken cancellationToken)
   at Avalonia.Controls.ApplicationLifetimes.ClassicDesktopStyleApplicationLifetime.Start(String[] args)
   at Avalonia.ClassicDesktopStyleApplicationLifetimeExtensions.StartWithClassicDesktopLifetime(AppBuilder builder, String[] args, ShutdownMode shutdownMode)
   at Ryujinx.Ava.Program.Main(String[] args) in /src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 51
```

Opening the settings window again after closing it, leads to the crash above. Could you look into that?

TSRBerry (approved):
Lgtm, thank you! :D

PR #5338, opened 2023-06-20
Print Process ID On Startup

patrick-hovsepian:
This is useful for general debugging purposes including attaching to the right process in VS Code 😄

### DIFF
--- a/src/Ryujinx.Common/SystemInfo/SystemInfo.cs
+++ b/src/Ryujinx.Common/SystemInfo/SystemInfo.cs
@@ -24,6 +24,7 @@ namespace Ryujinx.Common.SystemInfo
 
         public void Print()
         {
+            Logger.Notice.Print(LogClass.Application, $"Process ID: {Environment.ProcessId}");
             Logger.Notice.Print(LogClass.Application, $"Operating System: {OsDescription}");
             Logger.Notice.Print(LogClass.Application, $"CPU: {CpuName}");
             Logger.Notice.Print(LogClass.Application, $"RAM: Total {ToMiBString(RamTotal)} ; Available {ToMiBString(RamAvailable)}");


AcK77 (approved):

gdkchan:
It's a bit weird to add that information there since it's technically not "system information". I never had troubles attaching a debugger since they usually show the application title too which makes spotting the emulator easy. The only cases where I needed the PID was when using gdb or lldb for native debugging, but it's pretty easy to get the PID from the task manager.

marysaka:
> It's a bit weird to add that information there since it's technically not "system information". I never had troubles attaching a debugger since they usually show the application title too which makes spotting the emulator easy. The only cases where I needed the PID was when using gdb or lldb for native debugging, but it's pretty easy to get the PID from the task manager.

It might be easy on Windows but from other OS the process is different. I wouldn't mind a change like especially when working on macOS or Linux.

gdkchan:
> > It's a bit weird to add that information there since it's technically not "system information". I never had troubles attaching a debugger since they usually show the application title too which makes spotting the emulator easy. The only cases where I needed the PID was when using gdb or lldb for native debugging, but it's pretty easy to get the PID from the task manager.
> 
> It might be easy on Windows but from other OS the process is different. I wouldn't mind a change like especially when working on macOS or Linux.

It's the same process on other OS. They all have a "task manager" that shows the process PID. On macOS it is called activity monitor.

I guess my main gripe with that is that it adds a feature that is only useful for local debugging to the logs. Maybe there is a better way to expose this. Maybe it can be removed from release builds and only be done when the emulator is built locally.

gdkchan (changes requested):
My suggestions:
- Only print that information for "dirty" builds. I don't think this is useful for release builds in general, it would be only adding useless noise to the logs we receive, as the PID is only useful when the process is still alive. And I think if someone is going to debug the emulator, they will use a local build not an official release.
- This is less important, but if there's a simple way to log it to console only, it could be nice as I don't think logging this information to the log file would ever be useful. But it's not very important and I don't think it's worth complicating the logging system just for that.

patrick-hovsepian:
Sure, even one extra line in the log can be noise. This is virtually free and has no bearing on the actual functionality of the emulator. The one thing we do agree on is that it technically isn't system information. If you're so opposed to it, just close it.

jduncanator:
I agree with @gdkchan on this one.  I am not opposed to including the PID in the log output, but I'd go one step further and only print it in Debug builds.

PR #5344, opened 2023-06-22
"Find" method should be used instead of the "FirstOrDefault" extension

marco-carvalho:
Both the `List.Find` method and `IEnumerable.FirstOrDefault` method can be used to find the first element that satisfies a given condition in a collection. However, `List.Find` can be faster than `IEnumerable.FirstOrDefault` for `List` objects. For small collections, the performance difference may be minor, but for large collections, it can make a noticeable difference. The same applies for `ImmutableList` and arrays too.

Benchmark:

```cs
private List<string> data;
private Random random = new Random();

[Params(1_000)]
public int N { get; set; }

[GlobalSetup]
public void Setup() =>
    data = Enumerable.Range(0, N).Select(x => Guid.NewGuid().ToString()).ToList();

[Benchmark(Baseline = true)]
public void FirstOrDefault()
{
    for (var i = 0; i < N; i++)
    {
        var value = data[random.Next(N - 1)];
        _ = data.FirstOrDefault(x => x == value);   // Enumerable.FirstOrDefault()
    }
}

[Benchmark]
public void Find()
{
    for (var i = 0; i < N; i++)
    {
        var value = data[random.Next(N - 1)];
        _ = data.Find(x => x == value);             // List<T>.Find()
    }
}
```

Results:

<html>
<body>

Method | Runtime | Mean | StdDev | Ratio | Allocated
-- | -- | -- | -- | -- | --
FirstOrDefault | .NET 7.0 | 5.373 ms | 0.1049 ms | 1.00 | 125 KB
Find | .NET 7.0 | 1.691 ms | 0.0334 ms | 0.32 | 85.94 KB
FirstOrDefault | .NET Framework 4.6.2 | 5.035 ms | 0.0421 ms | 1.00 | 125.38 KB
Find | .NET Framework 4.6.2 | 1.779 ms | 0.0107 ms | 0.35 | 86.2 KB

</body>
</html>

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
@@ -264,7 +264,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         private void SelectLastScannedAmiibo()
         {
-            AmiiboApi scanned = _amiiboList.FirstOrDefault(amiibo => amiibo.GetId() == LastScannedAmiiboId);
+            AmiiboApi scanned = _amiiboList.Find(amiibo => amiibo.GetId() == LastScannedAmiiboId);
 
             SeriesSelectedIndex = AmiiboSeries.IndexOf(scanned.AmiiboSeries);
             AmiiboSelectedIndex = AmiiboList.IndexOf(scanned);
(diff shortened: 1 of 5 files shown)


gdkchan (approved):
Looks good.

AcK77 (approved):

PR #5345, opened 2023-06-22
"Exists" method should be used instead of the "Any" extension

marco-carvalho:
Both the `List.Exists` method and `IEnumerable.Any` method can be used to find the first element that satisfies a predicate in a collection. However, `List.Exists` can be faster than `IEnumerable.Any` for `List` objects, as well as requires significantly less memory. For small collections, the performance difference may be negligible, but for large collections, it can be noticeable. The same applies to `ImmutableList` and arrays too.

Benchmark:

```cs
private List<int> data;
private readonly Random random = new Random();

[Params(1_000)]
public int N { get; set; }

[GlobalSetup]
public void Setup() =>
    data = Enumerable.Range(0, N).Select(x => 43).ToList();

[Benchmark(Baseline = true)]
public void Any()
{
    for (var i = 0; i < N; i++)
    {
        _ = data.Any(x => x % 2 == 0);          // Enumerable.Any
    }
}

[Benchmark]
public void Exists()
{
    for (var i = 0; i < N; i++)
    {
        _ = data.Exists(x => x % 2 == 0);       // List<T>.Exists
    }
}
```

Results:

<html>
<body>

Method | Runtime | Mean | StdDev | Ratio | Allocated
-- | -- | -- | -- | -- | --
Any | .NET 7.0 | 6.670 ms | 0.1413 ms | 1.00 | 40004 B
Exists | .NET 7.0 | 1.364 ms | 0.0270 ms | 0.20 | 1 B
Any | .NET Framework 4.6.2 | 5.380 ms | 0.0327 ms | 1.00 | 40128 B
Exists | .NET Framework 4.6.2 | 1.575 ms | 0.0348 ms | 0.29 | -

</body>
</html>

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureGroupHandle.cs
@@ -512,7 +512,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <returns>True if at least one of the handles is dirty</returns>
         private bool CheckDirty()
         {
-            return Handles.Any(handle => handle.Dirty);
+            return Array.Exists(Handles, handle => handle.Dirty);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


gdkchan (approved):
Looks good.

AcK77 (approved):

PR #5349, opened 2023-06-23
Set COMPlus_DefaultStackSize to 2M in macOS

Shihta:
Echo #5291, we can directly increase the stack size by setting the environment variable in `Ryujinx.app/Contents/Info.plist`.

Using this method, Mortal Kombat 11 still needs stack size of at least 2MB to run properly. Splatoon 3 doesn't need that much, but it's dominated by higher demand.

Update the copyright date by the way.

My environment:
OS: macOS 12.5.1
CPU: Apple M2
RAM: 24G

Some screenshop:
截圖 2023-06-23 下午9 21 07
截圖 2023-06-23 下午4 59 12

### DIFF
--- a/distribution/macos/Info.plist
+++ b/distribution/macos/Info.plist
@@ -39,10 +39,15 @@
     <key>CSResourcesFileMapped</key>
     <true/>
     <key>NSHumanReadableCopyright</key>
-    <string>Copyright © 2018 - 2022 Ryujinx Team and Contributors.</string>
+    <string>Copyright © 2018 - 2023 Ryujinx Team and Contributors.</string>
     <key>LSApplicationCategoryType</key>
     <string>public.app-category.games</string>
     <key>LSMinimumSystemVersion</key>
     <string>11.0</string>
+    <key>LSEnvironment</key>
+    <dict>
+        <key>COMPlus_DefaultStackSize</key>
+        <string>200000</string>
+    </dict>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


gdkchan:
Can you remove the custom thread stack size on Ryujinx.Ava? The `1 * 1024 * 1024` value. Now that the default size for all threads is changed, this is not needed anymore (and might be actually a bad thing since it will make the stack smaller).

Shihta:
> Can you remove the custom thread stack size on Ryujinx.Ava? The `1 * 1024 * 1024` value. Now that the default size for all threads is changed, this is not needed anymore (and might be actually a bad thing since it will make the stack smaller).
> https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Ava/AppHost.cs#L137

it's done.
I also verified the ava-ryujinx-Release-1.1.0+291a143-macos_universal generated by the CI process. It can run both titles normally.

distribution/macos/Info.plist
```diff
@@ -39,10 +39,15 @@
...
-    <string>Copyright © 2018 - 2022 Ryujinx Team and Contributors.</string>
+    <string>Copyright © 2018 - 2023 Ryujinx Team and Contributors.</string>
     <key>LSApplicationCategoryType</key>
     <string>public.app-category.games</string>
     <key>LSMinimumSystemVersion</key>
     <string>11.0</string>
+    <key>LSEnvironment</key>
+    <dict>
+        <key>COMPlus_DefaultStackSize</key>
+        <string>200000</string>
```
makigumo, line 50: It seems to me there is a `0` missing here (200K vs. 2M).
```suggestion
        <string>2000000</string>
```

gdkchan: This is a hex vaalue, so it is correct. 0x200000 = 2097152.

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
LGTM. People might have to manually set this environment variable when running via `dotnet run`, but linux has its own issues that need resolved via environment vars so needing it here isn't out of character.

PR #5350, opened 2023-06-23
Non-derived "private" classes and records should be "sealed"

marco-carvalho:
I know https://github.com/Ryujinx/Ryujinx/pull/5180 was closed because the devs core team thought the change was too large, but I think at least this small PR should be merged because classes and records with either `private` or `file` access modifiers aren’t visible outside of their assemblies or files, so if they’re not extended inside their scope, they should be made explicitly non-extensible with the addition of the `sealed` keyword.

Benchmark:

```cs
[Params(1_000_000)]
public int Iterations { get; set; }

private readonly UnsealedClass unsealedType = new UnsealedClass();
private readonly SealedClass sealedType = new SealedClass();

[Benchmark(Baseline = true)]
public void UnsealedType()
{
    for (int i = 0; i < Iterations; i++)
    {
        unsealedType.DoNothing();
    }
}

[Benchmark]
public void SealedType()
{
    for (int i = 0; i < Iterations; i++)
    {
        sealedType.DoNothing();
    }
}

private class BaseType
{
    public virtual void DoNothing() { }
}

private class UnsealedClass : BaseType
{
    public override void DoNothing() { }
}

private sealed class SealedClass : BaseType
{
    public override void DoNothing() { }
}
```

Results:

<html>
<body>

Method | Runtime | Mean | StdDev | Ratio
-- | -- | -- | -- | --
UnsealedType | .NET 7.0 | 1,074.5 us | 3.15 us | 1.00
SealedType | .NET 7.0 | 216.1 us | 1.19 us | 0.20

</body>
</html>

### DIFF
--- a/src/ARMeilleure/CodeGen/Arm64/PreAllocator.cs
+++ b/src/ARMeilleure/CodeGen/Arm64/PreAllocator.cs
@@ -10,7 +10,7 @@ namespace ARMeilleure.CodeGen.Arm64
 {
     static class PreAllocator
     {
-        private class ConstantDict
+        private sealed class ConstantDict
         {
             private readonly Dictionary<(ulong, OperandType), Operand> _constants;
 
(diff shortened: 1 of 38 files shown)


neon-sunset:
`sealed` only affects virtual calls on derived types.

PR #5353, opened 2023-06-23
Empty "case" clauses that fall through to the "default" should be omitted

marco-carvalho:
Empty `case` clauses that fall through to the default are useless. Whether or not such a `case` is present, the `default` clause will be invoked. Such `case`s simply clutter the code, and should be removed.

### DIFF
--- a/src/ARMeilleure/Instructions/SoftFloat.cs
+++ b/src/ARMeilleure/Instructions/SoftFloat.cs
@@ -229,6 +229,9 @@ namespace ARMeilleure.Instructions
             switch (context.Fpcr.GetRoundingMode())
             {
                 default:
+                    throw new ArgumentException($"Invalid rounding mode \"{context.Fpcr.GetRoundingMode()}\".");
+
+                case FPRoundingMode.ToNearest:
                     roundUp       = (error > 0.5d || (error == 0.5d && (intMant & 1u) == 1u));
                     overflowToInf = true;
                     break;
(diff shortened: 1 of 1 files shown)


gdkchan:
even though the additional cases are technically useless, they tell which rounding mode is being implemented. If you remove it, it will be hard to know at a first glance which mode the `default` case implements. I believe the `default` case should never be actually reached on those functions and is only there to avoid compiler warnings.

marco-carvalho:
@gdkchan I noticed in `X86GetRoundControl` that it throws an `ArgumentException`. Do you believe those cases that I changed should throw `ArgumentException` in the `default` case?
image

gdkchan:
Yes, it should be impossible for the rounding mode to be anything other than those 4 enum values.

marco-carvalho:
@gdkchan Could you check again, please?

src/ARMeilleure/Instructions/SoftFloat.cs
```diff
@@ -229,6 +229,8 @@ public static ushort FPRoundCv(double real, ExecutionContext context)
             switch (context.Fpcr.GetRoundingMode())
             {
                 default:
```
gdkchan, line 231: Can you move the default case to the end (after all other cases) for consistency with other methods?

src/ARMeilleure/Instructions/SoftFloat.cs
```diff
@@ -1433,7 +1439,7 @@ public static float FPRecipEstimateFpscr(float value, bool standardFpscr)
 
                 switch (fpcr.GetRoundingMode())
                 {
-                    default:
+                    default:                                  throw new ArgumentException($"Invalid rounding mode \"{fpcr.GetRoundingMode()}\".");
```
gdkchan, line 1442: Same comment as above. In addition, I would prefer if this was like the other methods, and the case, assignment and `break;` where all on separate lines, without the weird alignment.

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #5354, opened 2023-06-23
Prefer indexing instead of "Enumerable" methods on types implementing "IList"

marco-carvalho:
Benchmark:

```cs
private List<byte> data;
private Random random;

[Params(1_000_000)]
public int SampleSize;

[Params(1_000)]
public int LoopSize;

[GlobalSetup]
public void Setup()
{
    random = new Random(42);
    var bytes = new byte[SampleSize];
    random.NextBytes(bytes);
    data = bytes.ToList();
}

[Benchmark]
public void First()
{
    for (int i = 0; i < LoopSize; i++)
    {
        _ = data.First();
    }
}

[Benchmark]
public void First_Index()
{
    for (int i = 0; i < LoopSize; i++)
    {
        _ = data[0];
    }
}
```

Results:

<html>
<body>

Method | Runtime | Mean | StdDev
-- | -- | -- | --
First | .NET 7.0 | 7,790.2 ns | 165.70 ns
First_Index | .NET 7.0 | 398.5 ns | 5.36 ns

</body>
</html>

### DIFF
--- a/src/Ryujinx.Ui.LocaleGenerator/LocaleGenerator.cs
+++ b/src/Ryujinx.Ui.LocaleGenerator/LocaleGenerator.cs
@@ -14,7 +14,7 @@ namespace Ryujinx.Ui.LocaleGenerator
 
             context.RegisterSourceOutput(contents, (spc, content) =>
             {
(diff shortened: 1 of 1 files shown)


zclnma (approved):

AcK77 (approved):

riperiperi (approved):
This method isn't performance critical at all, but might as well.

PR #5359, opened 2023-06-24
[Ryujinx.Audio.Backends.OpenAL] Address dotnet-format issues

TSRBerry:
Base PR: #4670


src/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceSession.cs
```diff
@@ -32,23 +32,17 @@ public OpenALHardwareDeviceSession(OpenALHardwareDeviceDriver driver, IVirtualMe
...
-                    }
-                default:
-                    throw new NotImplementedException($"Unsupported sample format {RequestedSampleFormat}");
-            }
+                SampleFormat.PcmInt16 => RequestedChannelCount switch
+                {
+                    1 => ALFormat.Mono16,
+                    2 => ALFormat.Stereo16,
+                    6 => ALFormat.Multi51Chn16Ext,
+                    _ => throw new NotImplementedException($"Unsupported channel config {RequestedChannelCount}"),
```
AcK77, line 42:
```suggestion
                    _ => throw new NotImplementedException($"Unsupported channel config {RequestedChannelCount}")
```

src/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceSession.cs
```diff
@@ -32,23 +32,17 @@ public OpenALHardwareDeviceSession(OpenALHardwareDeviceDriver driver, IVirtualMe
...
+                },
+                _ => throw new NotImplementedException($"Unsupported sample format {RequestedSampleFormat}"),
```
AcK77, line 44:
```suggestion
                _ => throw new NotImplementedException($"Unsupported sample format {RequestedSampleFormat}")
```

AcK77 (changes requested):

src/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceSession.cs
marysaka: That's actually cleaner to have the comma and standard in most language (with some formatter)

marysaka (approved):

src/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceSession.cs
AcK77: It’s just a personal preference, I’m fine with both

TSRBerry: I removed the trailing commas everywhere at first, but the diff of that was very large because we are using them a lot actually. That's why I reverted that again.

TSRBerry:
Dropped the commit containing the workflow, since that was not supposed to be in here.

AcK77 (approved):

TSRBerry:
Just did a final check run for this one and all checks pass, so this is good to merge now!

TSRBerry:
I have to check these again, since dotnet format doesn't like the include option very much.
Hopefully it still doesn't want any changes.

src/Ryujinx.Audio.Backends.OpenAL/OpenALHardwareDeviceSession.cs
TSRBerry: Tailing commas might be enforceable soon: https://github.com/dotnet/roslyn/pull/54859

AcK77: Then tailing commas should be added everywhere it’s needed!

PR #5366, opened 2023-06-24
[Ryujinx.Graphics.GAL] Address dotnet-format issues

TSRBerry:
Base PR: #4670


src/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
```diff
@@ -160,5 +160,20 @@ bool IEquatable<TextureCreateInfo>.Equals(TextureCreateInfo other)
                    SwizzleB == other.SwizzleB &&
                    SwizzleA == other.SwizzleA;
         }
+
+        public override bool Equals(object obj)
+        {
+            return obj is TextureCreateInfo info && ((IEquatable<TextureCreateInfo>)this).Equals(info);
```
gdkchan, line 166: Why is this cast to `IEquatable<TextureCreateInfo>` needed? Doesn't it pick the right overload if you just call `Equals` directly?

src/Ryujinx.Graphics.GAL/Multithreading/CommandHelper.cs
```diff
@@ -17,8 +17,8 @@ static class CommandHelper
     {
         private delegate void CommandDelegate(Span<byte> memory, ThreadedRenderer threaded, IRenderer renderer);
 
-        private static int _totalCommands = (int)Enum.GetValues<CommandType>().Max() + 1;
-        private static CommandDelegate[] _lookup = new CommandDelegate[_totalCommands];
+        private static readonly int _totalCommands = (int)Enum.GetValues<CommandType>().Max() + 1;
```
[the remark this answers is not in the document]
TSRBerry: I commented about the default in another PR, but I need to search for it again.
There is no default defined for them. Rider defaults to UpperCamelCase for them, but AcK said, he'd prefer _lowerCamelCase instead.

src/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
TSRBerry: image
If I do that I get this complaint, but for some reason it's gone if I add this cast again.
image

I'd like to compare the IL for that, but I can't find the Equals method in there at all.

marysaka (approved):

src/Ryujinx.Graphics.GAL/TextureCreateInfo.cs
gdkchan: Maybe it works if you replace `bool IEquatable<TextureCreateInfo>.Equals(TextureCreateInfo other)` with `public bool Equals(TextureCreateInfo other)`? I'm not sure why the interface is explicitly implemented here. THere is similar code in a lot of other places, and none of them explicitly implements the interface.

TSRBerry: Yup that works, the warning is now also gone without the cast!

gdkchan (approved):
lgtm, thanks.

PR #5386, opened 2023-06-24
[Ryujinx.Memory] Address dotnet-format issues

TSRBerry:
Base PR: #4670


src/Ryujinx.Memory/MemoryBlock.cs
```diff
@@ -365,9 +364,11 @@ public WritableRegion GetWritableRegion(ulong offset, int size)
         /// <param name="pointer">Native pointer</param>
         /// <param name="offset">Offset to add</param>
         /// <returns>Native pointer with the added offset</returns>
-        private IntPtr PtrAddr(IntPtr pointer, ulong offset)
+        private static IntPtr PtrAddr(IntPtr pointer, ulong offset)
         {
+#pragma warning disable CA2020 // Prevent behavioral change caused by built-in operators of IntPtr/UIntPtr
             return (IntPtr)(pointer.ToInt64() + (long)offset);
```
AcK77, line 370:
```suggestion
            return new IntPtr(pointer.ToInt64() + (long)offset);
```

And remove the pragma ?

src/Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -186,17 +186,21 @@ public unsafe static IntPtr CreateSharedMemory(ulong size, bool reserve)
...
             }
 
-            return (IntPtr)fd;
+            return fd;
         }
 
         public static void DestroySharedMemory(IntPtr handle)
         {
+#pragma warning disable CA2020 // Prevent behavioral change caused by built-in operators of IntPtr/UIntPtr
             close((int)handle);
```
AcK77, line 195:
```suggestion
            close(handle.ToInt32());
```
And remove the pragma

src/Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -186,17 +186,21 @@ public unsafe static IntPtr CreateSharedMemory(ulong size, bool reserve)
...
+#pragma warning restore CA2020
         }
 
         public static IntPtr MapSharedMemory(IntPtr handle, ulong size)
         {
+#pragma warning disable CA2020 // Prevent behavioral change caused by built-in operators of IntPtr/UIntPtr
             return mmap(IntPtr.Zero, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE, MmapFlags.MAP_SHARED, (int)handle, 0);
```
AcK77, line 202:
```suggestion
            return mmap(IntPtr.Zero, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE, MmapFlags.MAP_SHARED, handle.ToInt32(), 0);
```
And remove the pragma

src/Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -206,12 +210,14 @@ public static void UnmapSharedMemory(IntPtr address, ulong size)
 
         public static void MapView(IntPtr sharedMemory, ulong srcOffset, IntPtr location, ulong size)
         {
+#pragma warning disable CA2020 // Prevent behavioral change caused by built-in operators of IntPtr/UIntPtr
             mmap(location, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE, MmapFlags.MAP_FIXED | MmapFlags.MAP_SHARED, (int)sharedMemory, (long)srcOffset);
```
AcK77, line 214:
```suggestion
            mmap(location, size, MmapProts.PROT_READ | MmapProts.PROT_WRITE, MmapFlags.MAP_FIXED | MmapFlags.MAP_SHARED, sharedMemory.ToInt32(), (long)srcOffset);
```
And remove the pragma

src/Ryujinx.Memory/MemoryManagerUnixHelper.cs
```diff
@@ -164,9 +164,11 @@ private static int MmapFlagsToSystemFlags(MmapFlags flags)
             return result;
         }
 
+#pragma warning disable IDE1006 // Naming rule violation
```
AcK77, line 167: Maybe we should rename the method instead of using a pragma

src/Ryujinx.Memory/Tracking/MultiRegionHandle.cs
```diff
@@ -406,6 +407,7 @@ public void RegisterPreciseAction(ulong address, ulong size, PreciseRegionSignal
 
         public void Dispose()
         {
+            GC.SuppressFinalize(this);
```
AcK77, line 410:
```suggestion
            GC.SuppressFinalize(this);

```

src/Ryujinx.Memory/Tracking/RegionHandle.cs
```diff
@@ -50,7 +49,7 @@ protected set
 
         internal IMultiRegionHandle Parent { get; set; }
 
-        private event Action _onDirty;
+        private event Action OnDirty;
```
AcK77, line 52: I'd prefer to keep private fields prefixed by `_` in cameCase

src/Ryujinx.Memory/Tracking/SmartMultiRegionHandle.cs
```diff
@@ -271,6 +271,7 @@ public void QueryModified(ulong address, ulong size, Action<ulong, ulong> modifi
 
         public void Dispose()
         {
+            GC.SuppressFinalize(this);
```
AcK77, line 274:
```suggestion
            GC.SuppressFinalize(this);

```

src/Ryujinx.Memory/WindowsShared/PlaceholderManager.cs
```diff
@@ -704,7 +706,9 @@ private void RestoreRangeProtection(ulong address, ulong size)
                 count = _protections.GetNodes(address, endAddress, ref overlaps);
             }
 
+#pragma warning disable IDE0059 // Remove unnecessary value assignment
             ulong startAddress = address;
```
AcK77, line 710: If it's not needed, maybe we should remove it.

src/Ryujinx.Memory/WindowsShared/WindowsFlags.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Ryujinx.Memory.WindowsShared
 {
     [Flags]
+    [SuppressMessage("Design", "CA1069: Enums values should not be duplicated")]
```
AcK77, line 7: As I said in another project, this should be removed.

AcK77 (changes requested):

src/Ryujinx.Memory/WindowsShared/WindowsFlags.cs
TSRBerry: If I remove that CI will fail. I either need to disable it in `.editorconfig` or replace this with a pragma.

src/Ryujinx.Memory/Tracking/RegionHandle.cs
TSRBerry: Events don't qualify as fields. C# defaults to PascalCase for these, see: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules#default-naming-styles

Should I still change it?

AcK77: It's a bit strange >< ... Let's other devs input on this.
I prefer `_camelCase` for `private` in any case when it's not a `const`.

src/Ryujinx.Memory/WindowsShared/WindowsFlags.cs
AcK77: Maybe we could fix the enum that way:

```
...
Decommit = 0x4000,
ReplacePlaceholder = Decommit,
...
```

TSRBerry: I can try that, but I think it leads to the same issue.

TSRBerry: Nvm, your suggestion seems to work just fine!

src/Ryujinx.Memory/Tracking/RegionHandle.cs
TSRBerry: @Ryujinx/developers need some input here.

marysaka: it follows property pattern right? so that kind of makes sense to me...

marysaka (approved):

src/Ryujinx.Memory/Tracking/RegionHandle.cs
TSRBerry: I'd also like to keep them as PascalCase tbh

gdkchan: I'd just follow standard naming rules for those which is PascalCase apparently?

marysaka: That makes sense as you can see `event`s as properties in a sense (and properties use PascalCase too)

src/Ryujinx.Memory/MemoryManagementUnix.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
+
```
gdkchan, line 5: I might be imagining things, but I think some of your other PRs removes the separation between `using` and `using static`.

src/Ryujinx.Memory/Tracking/MultiRegionHandle.cs
```diff
@@ -406,6 +407,8 @@ public void RegisterPreciseAction(ulong address, ulong size, PreciseRegionSignal
...
```
gdkchan, line 410: Might be wroth properly implementing the dispose pattern.

src/Ryujinx.Memory/MemoryManagementUnix.cs
TSRBerry: I thought so too, I'll check that.

src/Ryujinx.Memory/Tracking/MultiRegionHandle.cs
TSRBerry: What's missing for that? `IMultiRegionHandle` implements `IDisposable`

gdkchan: You can check https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#dispose-and-disposebool

gdkchan: I was thinking about the missing `Dispose(bool)` method, but since the `GC.SuppressFinalize(this);` goes into the main `Dispose` method it should be fine like that. We can change it to follow the recommended pattern on a separate PR.

gdkchan (approved):
lgtm, thanks.

src/Ryujinx.Memory/Tracking/SmartMultiRegionHandle.cs
AcK77, line 274:
```suggestion
            GC.SuppressFinalize(this);

```

src/Ryujinx.Memory/WindowsShared/WindowsFlags.cs
```diff
@@ -10,7 +10,7 @@ enum AllocationType : uint
         Commit = 0x1000,
         Reserve = 0x2000,
         Decommit = 0x4000,
-        ReplacePlaceholder = 0x4000,
+        ReplacePlaceholder = Decommit,
```
AcK77, line 13: I comment here since I can't at any other place, but it seems some enums here missing the tailing `,`

AcK77 (approved):
Little missing things, but I approve it with requested changes fixed!

PR #5387, opened 2023-06-24
[Ryujinx.SDL2.Common] Address dotnet-format issues

TSRBerry:
Base PR: #4670


src/Ryujinx.SDL2.Common/SDL2Driver.cs
```diff
@@ -80,8 +77,15 @@ public void Initialize()
                 }
 
                 // First ensure that we only enable joystick events (for connected/disconnected).
-                SDL_GameControllerEventState(SDL_DISABLE);
-                SDL_JoystickEventState(SDL_ENABLE);
+                if (SDL_GameControllerEventState(SDL_IGNORE) != SDL_IGNORE)
```
AcK77, line 80: This replace `SDL_GameControllerEventState(SDL_DISABLE);` so why it's now using `SDL_IGNORE` ?

marysaka (approved):
same comment as Ack otherwise lgtm

src/Ryujinx.SDL2.Common/SDL2Driver.cs
TSRBerry: I have no idea. I'll change it!

TSRBerry: Nvm, I figured it out.
https://wiki.libsdl.org/SDL2/SDL_GameControllerEventState
The only valid values for this function are: `SDL_IGNORE`, `SDL_ENABLE` and `SDL_QUERY`

AcK77 (approved):

PR #5394, opened 2023-06-24
[Spv.Generator] Address dotnet-format issues

TSRBerry:
Base PR: #4670


src/Spv.Generator/ConstantKey.cs
```diff
@@ -24,7 +24,7 @@ public bool Equals(ConstantKey other)
 
         public override bool Equals([NotNullWhen(true)] object obj)
         {
-            return obj is ConstantKey && Equals((ConstantKey)obj);
+            return obj is ConstantKey key && Equals(key);
```
AcK77, line 27: Mixing `is` and `&&` could cause this strange issue: https://github.com/Ryujinx/Ryujinx/pull/5207#discussion_r1214943768

src/Spv.Generator/DeterministicStringKey.cs
```diff
@@ -24,7 +24,7 @@ public bool Equals(DeterministicStringKey other)
 
         public override bool Equals([NotNullWhen(true)] object obj)
         {
-            return obj is DeterministicStringKey && Equals((DeterministicStringKey)obj);
+            return obj is DeterministicStringKey key && Equals(key);
```
AcK77, line 27: Mixing `is` and `&&` could cause this strange issue: https://github.com/Ryujinx/Ryujinx/pull/5207#discussion_r1214943768

src/Spv.Generator/Module.cs
```diff
@@ -262,104 +262,102 @@ public byte[] Generate()
             // Estimate the size needed for the generated code, to avoid expanding the MemoryStream.
             int sizeEstimate = 1024 + _functionsDefinitions.Count * 32;
 
-            using (MemoryStream stream = new MemoryStream(sizeEstimate))
-            {
-                BinaryWriter writer = new BinaryWriter(stream, System.Text.Encoding.ASCII);
+            using MemoryStream stream = new(sizeEstimate);
```
AcK77, line 265:
```suggestion
            using MemoryStream stream = new(sizeEstimate);

```

src/Spv.Generator/TypeDeclarationKey.cs
```diff
@@ -24,7 +24,7 @@ public bool Equals(TypeDeclarationKey other)
 
         public override bool Equals([NotNullWhen(true)] object obj)
         {
-            return obj is TypeDeclarationKey && Equals((TypeDeclarationKey)obj);
+            return obj is TypeDeclarationKey key && Equals(key);
```
AcK77, line 27: Mixing `is` and `&&` could cause this strange issue: https://github.com/Ryujinx/Ryujinx/pull/5207#discussion_r1214943768

src/Spv.Generator/spirv.cs
```diff
@@ -63,6 +65,7 @@ public enum SourceLanguage
             HLSL = 5,
         }
 
+        [SuppressMessage("Design", "CA1069: Enums values should not be duplicated")]
```
AcK77, line 68: I think we should remove that, in many places we could duplicate some enum values, so it just add extra useless code here. (Same for all other `SuppressMessage`

AcK77 (commented):
Some cleaned code is autogenerated one, maybe we should clean the generator aswell.

src/Spv.Generator/Operand.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Spv.Generator
 {
-    public interface Operand : IEquatable<Operand>
+    public interface IOperand : IEquatable<IOperand>
```
marysaka, line 6: Rename the file too.

marysaka (changes requested):

src/Spv.Generator/ConstantKey.cs
TSRBerry: Let's wait for gdkchans comment on the other PR

marysaka (approved):

src/Spv.Generator/DeterministicStringKey.cs
TSRBerry: The attribute for the parameter should just be `NotNull` instead of `NotNullWhen(true)`.

TSRBerry: Although Equals can be used with null as well, so the `NotNull` doesn't really make sense here anyway.

src/Spv.Generator/spirv.cs
TSRBerry: Instead of disabling the warning I'll solve them instead.
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1069#how-to-fix-violations

TSRBerry: List of other affected projects:

- Ryujinx.HLE
- Ryujinx.Graphics.Gpu
- Ryujinx.Graphics.GAL
- Ryujinx.Graphics.Shader
- Ryujinx.Memory
- Ryujinx.Ava

You can generate this list by executing this: `dotnet format analyzers --severity info -v d --report analyzers.json ; jq '.[] | select(.FileChanges[].DiagnosticId == "CA1069") | .FilePath' ./analyzers.json | uniq` and by adding `awk -F'/' '{print $<position of the project in the path>}'` in front of `uniq` you can get the list without any duplicates.

src/Spv.Generator/ConstantKey.cs
gdkchan: I have done this exact pattern in several other places and it has always been fine. I don't think it should be an issue here.

src/Spv.Generator/DeterministicStringKey.cs
gdkchan: Same as above.

src/Spv.Generator/GeneratorPool.cs
```diff
@@ -4,13 +4,15 @@ namespace Spv.Generator
 {
     public class GeneratorPool<T> where T : class, new()
     {
-        private List<T[]> _pool;
+        private readonly List<T[]> _pool;
         private int _chunkIndex = -1;
         private int _poolIndex = -1;
-        private int _initialSize;
-        private int _poolSizeIncrement;
+#pragma warning disable IDE0052 // Remove unread private member
```
gdkchan, line 10: If it's unused, just remove it.

AcK77 (approved):

PR #5399, opened 2023-06-25
misc: memory: Migrate from OutOfMemoryException to SystemException entirely

marysaka:
Fix a regression with address space allocation while providing more information about the context of the exception.


TSRBerry (approved):
Lgtm, thank you!

gdkchan (approved):

PR #5403, opened 2023-06-26
Ava: Swallow keyUp and keyDown events as well

TSRBerry:
Today I saw someone talking about the system bell annoying them while playing with keyboard input.
#4320 fixed this before, but it seems the issue came back.

Now we swallow all keyboard related events, so the system bell can't be triggered anymore.
This shouldn't break anything else, but I couldn't verify this, since I couldn't get keyboard input to work even before I added this fix. However keyboard input still works fine on Linux after adding that fix and I can confirm the bell won't play anymore on macOS.

### DIFF
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -65,10 +65,7 @@ namespace Ryujinx.Ava
     internal class AppHost
     {
         private const int CursorHideIdleTime = 5; // Hide Cursor seconds.
-        private const float MaxResolutionScale = 4.0f; // Max resolution hotkeys can scale to before wrapping.
         private const int TargetFps = 60;
-        private const float VolumeDelta = 0.05f;
-
         private static readonly Cursor _invisibleCursor = new(StandardCursorType.None);
         private readonly IntPtr _invisibleCursorWin;
         private readonly IntPtr _defaultCursorWin;
@@ -87,8 +84,8 @@ namespace Ryujinx.Ava
         public RendererHost RendererHost;
 
         private readonly GraphicsDebugLevel _glLogLevel;
-        private float _newVolume;
-        private KeyboardHotkeyState _prevHotkeyState;
+        // private float _newVolume;
+        // private KeyboardHotkeyState _prevHotkeyState;
 
         private long _lastCursorMoveTime;
         private bool _isCursorInRenderer = true;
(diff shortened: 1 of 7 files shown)


siegmund-heiss-ich:
Works so far for the sound.
Keyboard input still works.
But now the shortcuts (fullscreen or quit) don't work anymore.

TSRBerry:
Thank you for testing! I'll try again.

TSRBerry:
I was able to get my mac to recognize keyboard input properly and couldn't reproduce the issue with my fix.
But I noticed even without my fix you need to hold the hotkeys for a slightly longer time than on other platforms, so they get recognized.

@siegmund-heiss-ich so if you hold the hotkeys for a tiny bit longer are you able to use them successfully? From my testing everything else worked just fine.

siegmund-heiss-ich:
Aaah you are right!
But that's not how someone would work in reality.
We have shortcuts to work faster, not slower.
Tbh I am a bit concerned about this extremely bad responsiveness.
Maybe you (or me) can figure out, how it would work properly.

TSRBerry:
I feel like this might be an issue on the Avalonia side of things. Not sure how much we can do here.

gdkchan:
> I feel like this might be an issue on the Avalonia side of things. Not sure how much we can do here.

From what I recall, there are 2 different ways to handle hotkeys on Avalonia right now. One uses Avalonia Hotkey manager, the other simply reads currently pressed keys on the main render loop. So at least one of them is under our control. Which one is used depends on the hotkey.

siegmund-heiss-ich:
Most users will use controllers for emulation anyway. I would prefer to keep the annoying beep sound instead of introducing unusable hotkey latency.
Fun fact: If you quit the app with this changes, you get the beep sound because you have to press cmd+q that long.

TSRBerry:
As I said before this doesn't add to the latency. Current master behaves the same way, I just noticed it here.

TSRBerry:
In that case I'll need to check again.

IsaacMarovitz:
This is likely broken on Ava 11+ as the bell no longer sounds and such workarounds break input

TSRBerry:
Apparently we don't need any workaround for this and should probably remove the current one.

@siegmund-heiss-ich figured out that Ava11 solves this by running `defaults write org.ryujinx.Ryujinx ApplePressAndHoldEnabled -bool false`!

We could either do that for Ava10 as well and remove the workaround or we keep the old workaround for now and wait for Ava11 to remove it.

Hmm I guess before we forget to do that later, I'll adjust this PR to run the command and remove the workarounds.

PR #5409, opened 2023-06-28
Classes implementing "IEquatable<T>" should be sealed

marco-carvalho:
When a class implements the `IEquatable<T>` interface, it enters a contract that, in effect, states "I know how to compare two instances of type T or any type derived from T for equality.". However if that class is derived, it is very unlikely that the base class will know how to make a meaningful comparison. Therefore that implicit contract is now broken.


AcK77:
What does that mean? Even after reading your copy/pasted description I don’t get why those classes should be set as `sealed`!

PR #5411, opened 2023-06-28
Move support buffer update out of the backends

gdkchan:
Currently the logic for updating the support buffer data is implemented on the OpenGL and Vulkan. Both backends does pretty much the same thing, so this is essentially duplicated code. This duplication makes adding more data to the support buffer very annoying, as one needs to pass that data to the backend, which currently requires modifying the OpenGL, Vulkan backends and the GAL threading layer.

This change moves support buffer update to tthe GPU project, which solves the above issues, and has a nice diff of -198 lines.

Testing is welcome. It should be tested with resolution scale too, since resolution scale information is stored on the support buffer and is one of the things that could break.


Littl3Guy:
8 games tested while toggling resolution scale, no regressions found.

riperiperi (approved):
LGTM, a lot nicer this way. Seems to be fine with counter report rescaling and viewport inverse. Maybe could use some testing with BGRA on OpenGL since I don't have any affected games.

marysaka (approved):
lgtm

AcK77:
Needs rebase

Issue #4508, opened 2023-03-04
[Headless Build] Missing PP Effects arguments


jojodunet:
### Overview

I open this issue to track the request about adding PP Effects arguments to the Headless Build.

Initially requested here:
[Headless Build] Missing arguments & Potential issues https://github.com/Ryujinx/Ryujinx/issues/4190

#### Smaller details

Here the details:
Add Support for Post Processing Effects https://github.com/Ryujinx/Ryujinx/pull/3616#issuecomment-1360645066

#### Why would this feature be useful?

Add Post Processing Effects for the Headless Build.

PR #5339, opened 2023-06-21
Headless: Add support for fullscreen option

theundebruijn:
The headless SDL2 client was missing a way to launch in fullscreen mode.
This PR adds an optional argument similar to the one used in the GUI client's CLI.

In addition to getting headless in line with Ava/GTK, by adding this we support launching games in fullscreen mode from third party launchers (e.g. Steam Big Picture).

---

NOTE: This PR serves as a prerequisite for https://github.com/Ryujinx/Ryujinx/pull/5412

---

#### TODO's

_Implementation_ :
- [x] Implement `borderless fullscreen windowed` logic
- [x] Add `--fullscreen` option to Headless CLI argument list

_Testing_ :
- [x] Windows 11 (22H2) (Intel x64) — Nvidia (Vulkan + OpenGL)
- [x] MacOS 12.x (Intel x64) — AMD (Vulkan)
- [x] MacOS 13.x (ARM64) — Apple Silicon (Vulkan) (tyvm @siegmund-heiss-ich)
- [x] Linux (Fedora Workstation 38) (Intel x64) (Wayland) — AMD (Vulkan + OpenGL)
- [x] Linux (Fedora Workstation 38) (Intel x64) (X11) — AMD (Vulkan + OpenGL)

Issue #4505, opened 2023-03-03
[Bug] G-Sync doesn't work if Ryujinx boots a game with the V-Sync option enabled.


SixelAlexiS90:
### Description of the issue

If you boot a Ryujinx game with the "Enable VSync" option enabled via the emulator options, the games will always start with V-Sync enabled and this will not make the G-Sync works for some reasons.

Turning the V-Sync off via the bottom left button after a game is started will still not make the G-Sync work, and it won't work even for the next game boot since that button will only save the change for that current session until you close Ryujinx or stop the emulation.

### Reproduction steps

The G-Sync doesn't work if Ryujinx boots a game with the emulator V-Sync option enabled:
Options > Settings > System > Enable VSync (option enabled)

Once you start a game with that option enabled and you go fullscreen, the G-Sync will not work, even if you disable the V-Sync after the game boot via the bottom left button.

To make the G-Sync work you have to disable the "Enable Vsync" option, save the change and restart Ryujinx or stop the emulation and restart the game.
At the next boot the G-Sync will properly work (once you go in fullscreen), even if you enable the emulator V-Sync again via the bottom left button.

### Log file

This is the log with the G-Sync not working:

[Ryujinx_1.1.647_2023-03-03_15-50-03.log](https://github.com/Ryujinx/Ryujinx/files/10883093/Ryujinx_1.1.647_2023-03-03_15-50-03.log)

This is the log with the G-Sync properly working:

[Ryujinx_1.1.647_2023-03-03_15-51-43.log](https://github.com/Ryujinx/Ryujinx/files/10883107/Ryujinx_1.1.647_2023-03-03_15-51-43.log)

### OS

Windows 10

### Ryujinx version

1.1.647

### CPU

i7-4770K

### GPU

Nvidia RTX 2060

### RAM

16GB

### Additional context?

I've tested this on two different PCs with different G-Sync compatible monitors, one with W10 and 516.94 drivers and another with W11 and 531.18 drivers so it's easily reproducible and consistent.

For easily test this (if your monitor doesn't support a dynamic Hz counter) you can enable the G-Sync overlay that say if it's working or not via the Nvidia Control Panel:

Click on one of the "Display" options on the left, after that a "Display" option will pop on the top bar.
Click on that and enable the "G-SYNC Indicator".

This indication will enable a text on the top right that will show a "NORMAL" grey text if G-Sync doesn't work and a "G-SYNC" green text if it's properly working.

[...]

PR #5412, opened 2023-06-29
Headless: Add support for Scaling Filters, Anti-aliasing and Exclusive Fullscreen

theundebruijn:
This PR adds several features to the Headless SDL2 client:

_Missing feature(s) added:_
`Scaling Filters (Bilinear, Nearest, FSR)`, `Anti-aliasing (None, FXAA, SMAA (all levels))`

_New feature(s) added:_
`Support for an Exclusive Fullscreen mode`

All features mentioned come with CLI options. These, with their descriptions, are added to the Headless' `Options.cs`.
Here's an example of how to launch a game using them:
```
Ryujinx.Headless.SDL2(.exe) --display-id 0 --exclusive-fullscreen --exclusive-fullscreen-width 3840 --exclusive-fullscreen-height 2160 --anti-aliasing "SmaaUltra" --scaling-filter "Fsr" --scaling-filter-level 10 "<path to>.xci"
```
**_NOTE: to exit the app on macOS when using Exclusive Fullscreen use `cmd + option + shift + escape`_.**
**_NOTE: we really do need some global keyboard shortcuts for headless. Would be a good next PR._**

---

This PR builds on top of https://github.com/Ryujinx/Ryujinx/pull/5339
Closes: https://github.com/Ryujinx/Ryujinx/issues/4508

---

#### TODO's

_Implementation_ :
- [x] Implement `Scaling Filter` logic
- [x] Add `Scaling Filter` options to the Headless CLI
- [x] Implement `Anti-Aliasing` logic
- [x] Add `Anti-aliasing` options to the Headless CLI
- [x] Implement `Exclusive Fullscreen` logic (OpenGL + Vulkan)
- [x] Add `Exclusive Fullscreen` options to the Headless CLI

_Testing_ :
- [x] Windows 11 (22H2) (Intel x64) — Nvidia (Vulkan + OpenGL)
- [x] MacOS 12.x (Intel x64) — AMD (Vulkan)
- [x] MacOS 13.x (ARM64) — Apple Silicon (Vulkan) (tyvm @siegmund-heiss-ich)
- [x] Linux (Fedora Workstation 38) (Intel x64) (Wayland) — AMD (Vulkan + OpenGL)
- [x] Linux (Fedora Workstation 38) (Intel x64) (X11) — AMD (Vulkan + OpenGL)

---

### About Exclusive Fullscreen

Exclusive Fullscreen mode has been a feature of many games and applications as it can help to significantly reduce the overhead of the display stack. Resulting in (ideally) much lower application latency.

The second reason is to add 'native' support for VRR technology such as Nvidia's G-SYNC. Using VRR on supported systems results in, once again, lower latency, as it overrides the VSync functionality to draw frames as soon as they're ready.

To use VRR with the current Ryujinx we need to use the nvcp to set the present method to 'Prefer layered on DXGI Swapchain'. Which - while it impressively works - likely has performance implications, is rather unintuitive and doesn't give us control over what's happening.

### How

This (rather small) PR updates the Headless SDL2 implementation to use the `SDL_WINDOW_FULLSCREEN` flag to set the correct window mode. It also uses `Ryujinx.Common.SystemInterop` to force DPI awareness as that screws up the scaling in this mode.

It adds a couple extra options to the Headless CLI: `--exclusive-fullscreen, --exclusive-fullscreen-width` and `--exclusive-fullscreen-height` which are needed as we no longer inherit the desktop properties. An additional `--display-id` option has been added that allows the user to choose which display to use.

### Does it work?

Well, yes. It successfully sets an exclusive fullscreen mode and Ryujinx renders things as expected.
VRR (G-SYNC in my case) gets triggered automatically, without any nvcp overrides, and overall application latency seems to be _much_ reduced.

#### Windows latency reduction

To make sure I'm not experiencing massive placebo effects here's some quick test results using [PresentMon](https://github.com/GameTechDev/PresentMon):

Measured latency using Headless in `Hardware: Legacy Flip` mode (Exclusive Fullscreen):
Screenshot 2023-06-29 010730

Measured latency using the existing GTK client:
Screenshot 2023-06-29 010907

Measured latency using the existing Avalonia client:
Screenshot 2023-06-29 011851_

The Ava client is particularly slow (both in windowed and fullscreen mode). Indicative of why it can feel sluggish at times.

#### macOS (Apple Silicon) output

Metal Performance HUD - Headless (Exclusive Fullscreen) (macOS 13.4.1):
ryujinx_arm_headless-exclusivefullscreen

Metal Performance HUD - Avalonia (Fullscreen) (macOS 13.4.1):

**_NOTE: Please don't look at the FPS, that just happened to be fluctuating._**
ryujinx_arm_ava-fullscreen

@siegmund-heiss-ich was particularly inventive and tried measuring latency on macOS - in the absence of tools like PresentMon - by comparing the audio latency between physical button press and the sound output in Super Mario Odyssey. He found a noticeable reduction by switching to exclusive fullscreen.

Please keep in mind testing was limited to a couple manual scenarios ran between the two of us.

Bildschirmfoto_2023-07-06_um_18 32 02

#### VRR (G-SYNC) support

Finally, besides the potential latency improvements, here's a photo of the Nvidia driver indicating that G-SYNC is actually active:
IMG_5581


Foul-Tarnished:
Why does Avalonia do Composed : Copy with GPU GDI instead of hardware composed : independant flip ? 💀 
As far as I'm aware, and as a SpecialK user, this seems to be just way worse

src/Ryujinx.Headless.SDL2/WindowBase.cs
```diff
@@ -185,10 +220,16 @@ private void HandleWindowEvent(SDL_Event evnt)
...
                 {
                     case SDL_WindowEventID.SDL_WINDOWEVENT_SIZE_CHANGED:
-                        Width = evnt.window.data1;
-                        Height = evnt.window.data2;
-                        Renderer?.Window.SetSize(Width, Height);
-                        MouseDriver.SetClientSize(Width, Height);
+                        // Unlike on Windows, this event fires on macOS when triggering fullscreen mode.
+                        // And promptly crashes the process because `Renderer?.window.SetSize` is undefined.
+                        // As we don't need this to fire in either case we can test for fullscreen.
+                        if (!IsFullscreen || !IsExclusiveFullscreen)
```
siegmund-heiss-ich, line 226: Logic error (use &&)
With this changes fullscreen and also exclusive fullscreen works on macOS.

siegmund-heiss-ich:
Everything in this PR addressed works fine now on Apple Silicon.

src/Ryujinx.Headless.SDL2/WindowBase.cs
```diff
@@ -226,6 +246,53 @@ private string GetGpuVendorName()
...
+                Renderer?.Window.SetScalingFilter((Graphics.GAL.ScalingFilter)2);
+                Renderer?.Window.SetScalingFilterLevel(ScalingFilterLevel);
+            }
+            // This forces the default to be "None" (similar to GTK/Ava)
+            else
+            {
+                Renderer?.Window.SetScalingFilter((Graphics.GAL.ScalingFilter)0);
+            }
+        }
+
```
marysaka, line 295: Why is not using variants of the enum directly?

marysaka (changes requested):

src/Ryujinx.Headless.SDL2/WindowBase.cs
```diff
@@ -55,7 +56,14 @@ public static void QueueMainThreadAction(Action action)
         public IHostUiTheme HostUiTheme { get; }
         public int Width { get; private set; }
         public int Height { get; private set; }
+        public int DisplayId { get; set; }
         public bool IsFullscreen { get; set; }
+        public bool IsExclusiveFullscreen { get; set; }
+        public int ExclusiveFullscreenWidth { get; set; }
+        public int ExclusiveFullscreenHeight { get; set; }
+        public Ryujinx.Common.Configuration.AntiAliasing AntiAliasing { get; set; }
+        public Ryujinx.Common.Configuration.ScalingFilter ScalingFilter { get; set; }
```
marysaka, line 65: Can't use do a ``using Ryujinx.Common.Configuration;``  to import that namespace?

If you do then you can change this to:

```suggestion
        public AntiAliasing AntiAliasing { get; set; }
        public ScalingFilter ScalingFilter { get; set; }
```

marysaka (approved):
One comment otherwise seems fine :+1:

src/Ryujinx.Headless.SDL2/WindowBase.cs
AcK77: You can do:

`using AntiAliasing = Ryujinx.Common.Configuration.AntiAliasing;`

And same for `ScalingFilter`

AcK77 (approved):
I haven’t tested it but code-wise it LGTM!

src/Ryujinx.Headless.SDL2/Options.cs
```diff
@@ -14,9 +14,21 @@ public class Options
...
+        [Option("display-id", Required = false, Default = 0, HelpText = "Set the display to use - especially helpful for fullscreen mode. [0-n]")]
+        public int DisplayId { get; set; }
+
+        [Option("fullscreen", Required = false, Default = false, HelpText = "Launch the game in fullscreen mode.")]
         public bool IsFullscreen { get; set; }
 
+        [Option("exclusive-fullscreen", Required = false, Default = false, HelpText = "Launch the game in exclusive fullscreen mode.")]
+        public bool IsExclusiveFullscreen { get; set; }
+
+        [Option("exclusive-fullscreen-width", Required = false, Default = 1280, HelpText = "Set horizontal resolution for exclusive fullscreen mode.")]
```
TSRBerry, line 26:
```suggestion
        [Option("exclusive-fullscreen-width", Required = false, Default = 1920, HelpText = "Set horizontal resolution for exclusive fullscreen mode.")]
```
Lets default to the resolution in docked mode

src/Ryujinx.Headless.SDL2/Options.cs
```diff
@@ -14,9 +14,21 @@ public class Options
...
+        public int ExclusiveFullscreenWidth { get; set; }
+
+        [Option("exclusive-fullscreen-height", Required = false, Default = 720, HelpText = "Set vertical resolution for exclusive fullscreen mode.")]
```
TSRBerry, line 29:
```suggestion
        [Option("exclusive-fullscreen-height", Required = false, Default = 1080, HelpText = "Set vertical resolution for exclusive fullscreen mode.")]
```

TSRBerry (commented):
Sorry for the late review! Everything lgtm, I just have one change request.

TSRBerry (approved):
Lgtm, thank you! :D

(And sorry again for the late review :sweat_smile:)

SixelAlexiS90:
@theundebruijn Hi! I want to thank you for adding this features on Headless, especially for the Exclusive Fullscreen, it's a game changer.
I'm using this version now since Avalonia completely got rid of Exclusive Fullscreen support, which is a big shame, especially considering that Exclusive Fullscreen WAS supported on the old "GTK" UI.
Exclusive Fullscreen was enabled by default on OpenGL once you got into the fullscreen mode while with Vulkan it could be forced via a "trick", I talked about it back then:
https://github.com/Ryujinx/Ryujinx/issues/4505

With Avalonia the Exclusive Fullscreen isn't a thing at all, which is a downgrade.
I don't know why they are relying on _Prefer layered on DXGI Swapchain_ only since, as you said, it can introduce a performance hit and it isn't as reliable, for example in my case DXGI swapchian doesn't work on Ryujinx, talked about this issue here:
https://github.com/Ryujinx/Ryujinx/pull/6645#issuecomment-2186746292

At least now I can use Ryujinx with a proper Gsync support thanks to your implementation so thank you very much!
I still hope they will re-integrate Exclusive Fullscreen to Avalonia since it's the only emulator I've seen that doesn't use it, especially when we talk about Vulkan.

PR #5415, opened 2023-06-29
[Ava]: Fix Cursor States On Windows 

Exhigh:
With this PR, cursor states on windows should be finally fixed for good. The main reason for the issue persisting across commits was the fact that the cursor's position updates while the game was being rendered wasn't being updated with regards to the scaling factor of the system, which led to an empty area, where the cursor's position wasn't considered within the rendering area. Also, previously the cursor, was rendered on every draw of the frame, which when we hide the cursor leads to the cursor flickering. Which has been fixed as of this PR. Also, now the cursor states take precedence when direct mouse mode is enabled, while the cursor is in "Never" hide mode and direct mouse is enabled, the cursor is not hidden and the cursor now hides in both "Always" and "On-Idle" Cursor hiding states while direct mouse mode is turned on. I felt like this was the way it was meant to be implemented, however this can be changed depending on what the reviewers feel the cursor's state should be.

Testing needs to be done to ensure that this PR, fixes the issue on systems with different configurations ( Scaling Factor ) and on ( MacOS & Linux ) to ensure their functionality hasn't been broken.

Fixes Issue: [#5136](https://github.com/Ryujinx/Ryujinx/issues/5136)
P.S This PR Supersedes [#5288](https://github.com/Ryujinx/Ryujinx/pull/5288)

My Configuration: 
* Ryzen 7 5800H
* 16GB 
* RTX 3060
* Win 11 22H2

### DIFF
--- a/src/Ryujinx.Ava/AppHost.cs
+++ b/src/Ryujinx.Ava/AppHost.cs
@@ -93,8 +93,15 @@ namespace Ryujinx.Ava
         private long _latestCursorMoveTime;
         private bool _isCursorInRenderer = true;
 
-        private bool _isCursorVisible = !ConfigurationState.Instance.Hid.EnableMouse.Value;
-        private bool _changeCursorState = false;
+        private enum CursorStates
+        {
+            CursorIsHidden,
+            CursorIsVisible,
+            ForceChangeCursor
+        };
+
+        private CursorStates _cursorState = !ConfigurationState.Instance.Hid.EnableMouse.Value ? 
+                                             CursorStates.CursorIsVisible : CursorStates.CursorIsHidden;
 
         private bool _isStopped;
         private bool _isActive;
@@ -233,8 +240,7 @@ namespace Ryujinx.Ava
         private void TopLevel_PointerLeave(object sender, PointerEventArgs e)
         {
             _isCursorInRenderer = false;
-            _changeCursorState = true;
-            _isCursorVisible = true;
+            _cursorState = CursorStates.ForceChangeCursor;
         }
 
         private void UpdateScalingFilterLevel(object sender, ReactiveEventArgs<int> e)
@@ -261,8 +267,7 @@ namespace Ryujinx.Ava
                 }
             });
 
-            _isCursorVisible = true;
-                _changeCursorState = false;
+            _cursorState = CursorStates.CursorIsVisible;
         }
 
         private void HideCursor()
@@ -277,8 +282,7 @@ namespace Ryujinx.Ava
                 }
             });
 
-            _isCursorVisible = false;
-                _changeCursorState = false;
+            _cursorState = CursorStates.CursorIsHidden;
         }
 
         private void SetRendererWindowSize(Size size)
@@ -525,7 +529,7 @@ namespace Ryujinx.Ava
         private void HideCursorState_Changed(object sender, ReactiveEventArgs<HideCursorMode> state)
         {
             _latestCursorMoveTime = Stopwatch.GetTimestamp();
-            _changeCursorState = true;
+            _cursorState = CursorStates.ForceChangeCursor;
         }
 
         public async Task<bool> LoadGuestApplication()
@@ -1073,13 +1077,13 @@ namespace Ryujinx.Ava
                     switch (ConfigurationState.Instance.HideCursor.Value)
                     {
                         case HideCursorMode.Never:
-                            if (!_isCursorVisible || _changeCursorState)
-                              ShowCursor();
+                            if (_cursorState != CursorStates.CursorIsVisible)
+                                ShowCursor();
                             break;
                         case HideCursorMode.OnIdle:
                         case HideCursorMode.Always:
-                            if (_isCursorVisible || _changeCursorState)
-                                    HideCursor();
+                            if (_cursorState != CursorStates.CursorIsHidden)
+                                HideCursor();
                             break;
                     }
                 }
@@ -1088,31 +1092,31 @@ namespace Ryujinx.Ava
                     switch (ConfigurationState.Instance.HideCursor.Value)
                     {
                         case HideCursorMode.Never:
-                                if (!_isCursorVisible || _changeCursorState)
+                                if (_cursorState != CursorStates.CursorIsVisible)
                                     ShowCursor();
                                 break;
                         case HideCursorMode.OnIdle:
                             if (Stopwatch.GetTimestamp() - _latestCursorMoveTime >= CursorHideIdleTime * Stopwatch.Frequency)
                             {
-                                if (_isCursorVisible || _changeCursorState)
-                                        HideCursor();
+                                if (_cursorState != CursorStates.CursorIsHidden)
+                                    HideCursor();
                             }
                             else
                             {
-                                if (!_isCursorVisible || _changeCursorState)
-                                        ShowCursor();
+                                if (_cursorState != CursorStates.CursorIsVisible)
+                                    ShowCursor();
                             }
                             break;
                         case HideCursorMode.Always:
-                            if (_isCursorVisible || _changeCursorState)
-                                    HideCursor();
+                            if (_cursorState != CursorStates.CursorIsHidden)
+                                HideCursor();
                             break;
                     }
                 }
             }
             else
             {
-                if (!_isCursorVisible || _changeCursorState)
+                if (_cursorState != CursorStates.CursorIsVisible)
                     ShowCursor();
             }
         }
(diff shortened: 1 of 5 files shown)


TSRBerry:
@patrick-hovsepian are you okay with this? Since you opened a similar PR and closed it a few minutes ago, I'm wondering if that was done for this PR or if there were other reasons.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -87,9 +87,12 @@ internal class AppHost
         private float                       _newVolume;
         private KeyboardHotkeyState         _prevHotkeyState;
 
-        private long _lastCursorMoveTime;
+        private long _latestCursorMoveTime;
         private bool _isCursorInRenderer = true;
 
+        private bool _isCursorVisible = !ConfigurationState.Instance.Hid.EnableMouse.Value;
+        private bool _changeCursorState = false;
```
marysaka, line 94: I was thinking maybe to simplify that logic, you could create a private enum with two flags instead?

marysaka (changes requested):
Apologies I forgot to send my review yesterday :sweat: 

Seems mostly fine to me but quite complicated... we need people to test this to ensure nothing breaks :+1:

src/Ryujinx.Ava/AppHost.cs
Exhigh: Have a look at commit [#Feedback](https://github.com/Ryujinx/Ryujinx/pull/5415/commits/58344e6d8d642805186bb6e80497559d465ced92) and let me know if this is what was intended.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -250,6 +273,9 @@ private void HideCursor()
                     SetCursor(InvisibleCursorWin);
                 }
             });
+
+            _isCursorVisible = false;
+                _changeCursorState = false;
```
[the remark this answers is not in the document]
Exhigh: addressed in the latest commit

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -237,6 +257,9 @@ private void ShowCursor()
                     SetCursor(DefaultCursorWin);
                 }
             });
+
+            _isCursorVisible = true;
+                _changeCursorState = false;
```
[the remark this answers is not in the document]
Exhigh: addressed in the latest commit

src/Ryujinx.Ava/AppHost.cs
Exhigh: @marysaka

jojodunet:
@Exhigh Have the requested changes been addressed? (I still see the requested changes status)

Exhigh:
That's just Github being Github, i have made the changes, tho it needs to be reviewed / looked at.

jojodunet:
> That's just Github being Github, i have made the changes, tho it needs to be reviewed / looked at.

Tried Experimental GUI (Avalonia) [241c96a](https://nightly.link/Ryujinx/Ryujinx/actions/artifacts/806134244.zip)
Windows 10
NVidia GTX 1070M
Ryujinx.Ava.exe -f --hide-cursor Always game.nsp

The mouse cursor is still visible upon launching the game and on the game window.
I need to move the mouse (just a little) for it to disappear from the window.

Exhigh:
hmmm, i don't remember having that said issue, will look into it, but prior to this patch the cursor wouldn't hide itself right ?

jojodunet:
> hmmm, i don't remember having that said issue, will look into it, but prior to this patch the cursor wouldn't hide itself right ?

Yes exactly

Exhigh:
can you test where rather than moving your mouse on the screen after launching from the commandline, try pressing a mousebutton and see if it causes the cursor to hide itself.

jojodunet:
> can you test where rather than moving your mouse on the screen after launching from the commandline, try pressing a mousebutton and see if it causes the cursor to hide itself.

The mouse cursor still persists when I try to press both the right or left mouse buttons.

Exhigh:
okie, will look into the issue.

Exhigh:
> > can you test where rather than moving your mouse on the screen after launching from the commandline, try pressing a mousebutton and see if it causes the cursor to hide itself.
> 
> The mouse cursor still persists when I try to press either the right or left mouse button.

@jojodunet Could you test with the latest commit and let me know if it fixes the issue. Hopefully it should

jojodunet:
@Exhigh Could you please rebase ?

Exhigh:
Its already rebased, its waiting on the PR workflow approval.

Exhigh:
> @Exhigh Could you please rebase? I still see the old PR build from 4 days ago.

For testing as of now, you could download the win artifact from [Build Artifact](https://github.com/Ryujinx/Ryujinx/actions/runs/5596885945), linux debug build test seems to be bugging out, for now this should do.

jojodunet:
Tried Experimental GUI (Avalonia) [aec4e5a](https://github.com/Ryujinx/Ryujinx/suites/14404392925/artifacts/813004046)
Windows 10
NVidia GTX 1070M
Ryujinx.Ava.exe -f --hide-cursor Always game.nsp

I have tested it, and here's my observation:

> The mouse cursor is present during the game loading.
> The mouse cursor disappears on its own at the first frame of the game.

Maybe I'm asking for too much, but would it be possible not to see the cursor at all?

In my use case, it's not really an issue as I won't have a mouse, but I have also noticed the following. 
(Could this be an issue for those who don't want to see the cursor regardless of the use case?)"

> The mouse cursor reappears on the game image if you move the mouse (whether in fullscreen or windowed mode).
> In windowed mode, I had thought that the cursor should only be visible at the emulator's menu level.
> From time to time, when moving the mouse actively, the cursor disappears again.

Exhigh:
> The mouse cursor is present during the game loading.
> The mouse cursor disappears on its own at the first frame of the game.

This is by design, I feel like the cursor should be visible while the game is loading, inferring that the game hasn't started yet. I could change this behavior but would need feedback from the devs.

> The mouse cursor reappears on the game image if you move the mouse (whether in fullscreen or windowed mode).
> In windowed mode, I had thought that the cursor should only be visible at the emulator's menu level.
> From time to time, when moving the mouse actively, the cursor disappears again.

Could you explain this statement, what was the cursor hide state (i.e whether the cursor hide state was Always, OnIdle or Never) and also whether direct mouse input was enabled or disabled. Also if possible attach a video showing the incident.
If you don't mind sharing which game are you seeing this issue, i know that the game doesn't matter but just to make sure nothing weird is happening.

jojodunet:
All observations are done on my side with the same command line " Ryujinx.Ava.exe -f --hide-cursor Always game.nsp "

I'm not sure how to share a video, but I would be interested in receiving more feedback from other users' experiences.

I tested with different games. 
I tested with both Direct mouse input enabled and disabled.

> In my use case, it's not really an issue as I won't have a mouse, but I have also noticed the following. (Could this be an issue for those who don't want to see the cursor regardless of the use case?)"
> 
> > The mouse cursor reappears on the game image if you move the mouse (whether in fullscreen or windowed mode).
> > In windowed mode, I had thought that the cursor should only be visible at the emulator's menu level.
> > From time to time, when moving the mouse actively, the cursor disappears again.

The observation above only appears when I move the mouse. 

> > In windowed mode, I had thought that the cursor should only be visible at the emulator's menu level.
> 
> It does and it also does for the status bar so that the users can interact with the menu. what's the question for this behavior ?

Once the cursor has reappeared on the game's rendering image because (and only if) you move the mouse, that is no longer the case and the cursor is visible everywhere.

> > From time to time, when moving the mouse actively, the cursor disappears again.
> 
> Within the render region or how exactly and if possible explain the reproduction steps.

On the game's rendering image (render region).
Reproduction steps: 
Just move the mouse, and then you could see the cursor reappears and disappears randomly.

Exhigh:
If possible could you post a log file for the run and provide a screencapture of the events happening. As I am not able to reproduce the said issues. Also which win10 version are you running ?

jojodunet:
Tried Experimental GUI (Avalonia) [aec4e5a](https://github.com/Ryujinx/Ryujinx/suites/14404392925/artifacts/813004046)
Windows 11 Home 22H2
With a different mouse than the one used in the previous tests
Ryujinx.Ava.exe -f --hide-cursor Always game.nsp

The cursor is still present during game loading. It seems to work fine in fullscreen, but switching to windowed mode breaks the feature. Returning to fullscreen makes it work fine.

It is fully broken under Windows, as shared above.

marysaka (approved):

Exhigh:
Re-Opening A New PR Instead of Updating This PR, since its been stale for some time now.

Issue #5024, opened 2023-05-19
[Bug]: infinite saving in totk


piplup55:
### Description of the issue

the game will save infinitely, if there extra files in the folder

### Reproduction steps

1. grab [this save](https://github.com/Ryujinx/Ryujinx/files/11520558/0000000000000032.tar.gz)
2. copy the data inside the 0 folder to your save location
3. load the game and try and save

### Log file

this log is from my unsuccessful attempt to reproduce it: [Ryujinx_1.1.808_2023-05-19_21-46-08.log](https://github.com/Ryujinx/Ryujinx/files/11520564/Ryujinx_1.1.808_2023-05-19_21-46-08.log) fs logs added

### OS

Arch Linux

### Ryujinx version

1.1.808

### Game version

1.1.0

### CPU

AMD Ryzen 9 7900X3D

### GPU

AMD RX 6600 XT

### RAM

64gb

### List of applied mods

- 60fps
- Amiibo day bypass
- Disable Dynamic Resolution
- Disable FSR
- Resolution 720p

### Additional context?

there extra files in the folder images below, windows can't see the files but linux can
![image](https://github.com/Ryujinx/Ryujinx/assets/100526773/6904212f-ac1d-43a9-ba53-27d615ab2182)
![image](https://github.com/Ryujinx/Ryujinx/assets/100526773/cf5b63f0-856a-4092-81c1-400d5a20e06a)

the extra files are a save and can be loaded the extra files seem to be from the beginning of the game,
checking saves from different users with this issue on the discord, they all have the extra files the cause of the files are still unknown, i'm opening this to track the progress of the issue

LilyArlatto:
I'm having this exact problem and I've literally no idea why it started happening, I even downgraded to Ryujinx 1.1.806 (the last version I was playing the game on) and it's suddenly infinite saving.

piplup55:
if you have this issue you can follow these steps to fix it

1.    Right click game and select Open User Save directory.
2.   Left click an empty area and drag to select all files (Don't Ctrl A)
3.    Copy and paste them into a safe folder somewhere else (Desktop is fine)
4.    In the original save folder, click the up arrow twice at the top to go up two folders.
5.    Note the name of the currently highlighted folder (the one you were in) and then delete it.
6.    Create a new folder with the same name.
7.    Copy the backup you made in step 3. into the new folder.
8.    Launch the game and check your saves.

[...]

PR #5418, opened 2023-06-30
[Ryujinx.HLE] Log when `IFileSystem.Commit()` fails due to `PathAlreadyInUse`

jamestiotio:
When the filesystem resource is already in use by another process, the call to `IFileSystem.Commit()` might fail with a result code of `PathAlreadyInUse`. This commit adds a warning log when this occurs to inform the user. This is because the resource might not have been released by other processes on the user's machine. The actual issue resolution will be left to the user since the root cause might wildly vary depending on what processes are running.

Related Issue: #5024

Note that this commit technically does not actually resolve the issue above. It simply helps the user with debugging purposes.

Signed-off-by: James Raphael Tiovalen <jamestiotio@gmail.com>

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
@@ -2,6 +2,7 @@ using LibHac;
 using LibHac.Common;
 using LibHac.Fs;
 using Path = LibHac.FsSrv.Sf.Path;
+using Ryujinx.Common.Logging;
 
 namespace Ryujinx.HLE.HOS.Services.Fs.FileSystemProxy
 {
@@ -148,7 +149,15 @@ namespace Ryujinx.HLE.HOS.Services.Fs.FileSystemProxy
         // Commit()
         public ResultCode Commit(ServiceCtx context)
         {
-            return (ResultCode)_fileSystem.Get.Commit().Value;
+            Result result = _fileSystem.Get.Commit();
+            ResultCode resultCode = (ResultCode)result.Value;
+
+            if (result.IsFailure() && resultCode == ResultCode.PathAlreadyInUse)
+            {
+                Logger.Warning?.Print(LogClass.ServiceFs, "The filesystem resource is already in use by another process.");
+            }
+
+            return resultCode;
         }
 
         [CommandCmif(11)]


jamestiotio:
Note that I did consider implementing a more "permanent" solution to this issue. However, after thinking about it, the only way to truly free up filesystem resources locked by other processes on the system from Ryujinx's side is by killing those processes, which seems a little overkill and destructive. Killing random processes just because we want to access the files might also result in unwanted side effects such as file corruption. We also cannot and should not assume the user's intention of prioritizing Ryujinx's actions of committing filesystem changes since those other processes might be important. Thus, I ended up with this PR which just logs a simple warning. That way, the user can ultimately decide what to do to resolve the issue.

jamestiotio:
Oh also, while I have solved the original issue (#5024) via Thealexbarney/LibHac#276, I still recommend adding this logging since I found it significantly helpful during debugging.

src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
```diff
@@ -148,7 +149,13 @@ public ResultCode OpenDirectory(ServiceCtx context)
...
         public ResultCode Commit(ServiceCtx context)
         {
-            return (ResultCode)_fileSystem.Get.Commit().Value;
+            Result res = _fileSystem.Get.Commit();
+            ResultCode resCode = (ResultCode)res.Value;
+            if ((res.IsFailure()) && (resCode == ResultCode.PathAlreadyInUse))
+            {
+                Logger.Warning?.Print(LogClass.ServiceFs, "The filesystem resource is already in use by another process.");
+            }
+            return resCode;
```
marysaka, line 158: This should be spaced correctly.

Also use resultCode instead of resCode

marysaka (changes requested):
apart from comment nothing much to say.

I'm not sure if we want this in the emulator tbh, any thoughts @gdkchan?

Thealexbarney:
I should note that the proposed solution for fixing the issue won't really work

src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
```diff
@@ -148,7 +149,16 @@ public ResultCode OpenDirectory(ServiceCtx context)
         // Commit()
         public ResultCode Commit(ServiceCtx context)
         {
-            return (ResultCode)_fileSystem.Get.Commit().Value;
+            Result result = _fileSystem.Get.Commit();
+
+            ResultCode resultCode = (ResultCode)result.Value;
+
+            if ((result.IsFailure()) && (resultCode == ResultCode.PathAlreadyInUse))
```
AcK77, line 156:
```suggestion
            if (result.IsFailure() && resultCode == ResultCode.PathAlreadyInUse)
```

src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
```diff
@@ -148,7 +149,16 @@ public ResultCode OpenDirectory(ServiceCtx context)
         // Commit()
         public ResultCode Commit(ServiceCtx context)
         {
-            return (ResultCode)_fileSystem.Get.Commit().Value;
+            Result result = _fileSystem.Get.Commit();
+
```
AcK77, line 153: Remove extra empty line

AcK77 (changes requested):

AcK77:
Superseeded by #5672

PR #4758, opened 2023-05-02
UI: New Crowdin updates

AcK77:
> ~~• Chinese Simplified -> 100%~~
> ~~• Chinese Traditional -> 100%~~
> ~~• French -> 100%~~
> ~~• German -> 100%~~
> • Greek -> 94%
> • Hebrew -> 98%
> ~~• Italian -> 100%~~
> ~~• Japanese -> 100%~~
> ~~• Korean -> 100%~~
> • Polish -> 98%
> • Portuguese, Brazilian -> 94%
> ~~• Russian -> 100%~~
> ~~• Spanish -> 100%~~
> ~~• Turkish -> 100%~~
> • Ukrainian -> 88%

If you want to help to translate the project, https://crowdin.com/project/ryujinx/

PR #5419, opened 2023-06-30
update zh_CN locale, add locale util to sync locale json keys with en_US.json

beginor:
update zh_CN locale, add locale util to sync locale json keys with en_US.json


Ryan2603:
just walking-by but found this most ridiculous PR which removing their original translated languages line but to inverting all the english line/tag without sense...

beginor:
These locale files are changed by a script, because some locale files do not have the same keys as `en_US.json`. 

So I write a script to check and update locale files, and the logic is simple:

1. if a key exist in `en_US.json`, but doesn't exist in `xx_XX.json`,  add it.
2. if a key exists in `xx_XX.json`, but doesn't exist in `en_US`.json , remove it.

AcK77:
All translations have to be handled using https://crwd.in/ryujinx and are synced by it. I don’t see why anything have to be automated using a script.
There is already https://github.com/Ryujinx/Ryujinx/pull/4758 which is opened by crowdin too. I didn’t merge it because a lot of translations are missing. So what’s the point of this PR, you added a macOS script too which is out of scope.

beginor:
Sorry, I have not seen #4758 before. I just have compared `en_US.json` with other json files, and I find there are many keys changed in `en_US.json` , but there are not synced with other locale files, which means the keys in the locale files are outdated,  I just write a small script to keep all the keys in other locale files update with `en_US.json`, because I think it's not easy to find which key is missing by human in json files.

Maybe I should include the script in this PR. 

And I will update the script I made, do not delete the keys in other locale files that do not exits in `en_US.json`, keep them  in the it's locale file, but it's useless at runtime.

I think it's useful to sync keys in all locale files, because it's easy to find what needs to be translated.

But if you think sync keys in `en_US.json` with a script is useless, I will close this PR and only use it by myself.

MutantAura:
As Ac_K said, all translations are handled by CrowdIn and should be submitted via the link above. 
We will then merge those updates periodically. Thank you for your interest.

PR #5426, opened 2023-07-02
Assertion arguments should be passed in the correct order

marco-carvalho:
The standard assertions library methods expect the first argument to be the expected value (like a hard-coded value) and the second argument to be the actual value. Swap them and your test will still have the same outcome (succeed/fail when it should) but the error messages will be confusing.

I noticed that this pattern is already being applied in most of the tests, so this PR only aims to standardize what was out of standard.


AcK77:
Do you will change code and use the copy/pasted description over the whole project for each of SonarSource rules?

marco-carvalho:
> Do you will change code and use the copy/pasted description over the whole project for each of SonarSource rules?

Only the ones I think are useful. There is a problem with this?

PR #5439, opened 2023-07-06
Revert "sdl: set `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` to 0"

marysaka:
Reverts Ryujinx/Ryujinx#5433

`SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` seems to not have any effects on the official HID driver.

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/ControllerInputViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/ControllerInputViewModel.cs
@@ -597,6 +597,8 @@ namespace Ryujinx.Ava.UI.ViewModels
             }
             else if (activeDevice.Type == DeviceType.Controller)
             {
+                bool isNintendoStyle = Devices.ToList().Find(x => x.Id == activeDevice.Id).Name.Contains("Nintendo");
+
                 string id = activeDevice.Id.Split(" ")[0];
 
                 config = new StandardControllerInputConfig
@@ -631,10 +633,10 @@ namespace Ryujinx.Ava.UI.ViewModels
                     },
                     RightJoycon = new RightJoyconCommonConfig<ConfigGamepadInputId>
                     {
-                        ButtonA = ConfigGamepadInputId.B,
-                        ButtonB = ConfigGamepadInputId.A,
-                        ButtonX = ConfigGamepadInputId.Y,
-                        ButtonY = ConfigGamepadInputId.X,
+                        ButtonA = isNintendoStyle ? ConfigGamepadInputId.A : ConfigGamepadInputId.B,
+                        ButtonB = isNintendoStyle ? ConfigGamepadInputId.B : ConfigGamepadInputId.A,
+                        ButtonX = isNintendoStyle ? ConfigGamepadInputId.X : ConfigGamepadInputId.Y,
+                        ButtonY = isNintendoStyle ? ConfigGamepadInputId.Y : ConfigGamepadInputId.X,
                         ButtonPlus = ConfigGamepadInputId.Plus,
                         ButtonR = ConfigGamepadInputId.RightShoulder,
                         ButtonZr = ConfigGamepadInputId.RightTrigger,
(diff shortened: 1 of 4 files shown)


SuperSamus:
What's the motivation for the revert?

OldManKain:
> What's the motivation for the revert?

Official Nintendo Switch Pro/Joy-con controllers had their buttons swapped in input mapping so if you pressed A on the controller it would Register as B in the GUI.

SuperSamus:
It was intentional:

>- The mapping will be the same on all controllers, removing the need to adjust custom mappings depending on what controller is used

OldManKain:
> It was intentional:
> 
> > * The mapping will be the same on all controllers, removing the need to adjust custom mappings depending on what controller is used

What Ever you did changed nothing when it comes to other controllers, it just changed the mapping on Nintendo Offical Pro/joy-con, 8bitDo, and GuliKit controllers.    

I have a hardy assortment of controllers to test if needed, but I really do not understand what you are trying to do here.  

But If you press A on your controller it should register as A not B.

SuperSamus:
> I really do not understand what you are trying to do here.

Let's say someone has a Nintendo and an Xbox controller. They set the mappings with the Nintendo controller. Next time, they play with the Xbox controller.
However, the mappings are now inverted (because SDL was inverting Nintendo's mappings due to basing it on the labels). Buttons are pressed by muscle memory and not by looking at the controller every time you want to push a button, so this is bad.
The user needs to go in the settings and invert the mappings manually.
This is not good UX for someone who uses different controllers. (Even worse in case of multiple players.)

> But If you press A on your controller it should register as A not B.

It's true that visually it's confusing.

So, possible solutions (that involve setting `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` to 0 again):
- Invert A/B and X/Y only visually in the settings. So, if the player presses B on an Xbox controller, then switches to a Nintendo controller, now B will appear as A (but internally it's still B).
- evdev avoids this mess by giving the face buttons the names "South", "East", "North", and "West".
- Wait and see what SDL3 does (if it will do something).

OldManKain:
>Let's say someone has a Nintendo and an Xbox controller. They set the mappings with the Nintendo controller. Next time, they play with the Xbox controller.
However, the mappings are now inverted (because SDL was inverting Nintendo's mappings due to basing it on the labels). Buttons are pressed by muscle memory and not by looking at the controller every time you want to push a button, so this is bad.
The user needs to go in the settings and invert the mappings manually.
This is not good UX for someone who uses different controllers. (Even worse in case of multiple players.)

No clue how you are having this type of issue,  If I go from Switch, to Xbox, to PS4 controller the mappings are already correct if you talking about button placement,  but yes non Switch Controllers will show opposite in the GUI because it is a Switch emulator.

SuperSamus:
>  If I go from Switch, to Xbox, to PS4 controller the mappings are already correct if you are talking about feeling button placement

No, they don't.
https://github.com/Ryujinx/Ryujinx/assets/40663462/fc68beef-7106-4c38-9ec2-5e0032b78eed

OldManKain:
> 

You didn't press load in the top right after switching controllers.

SuperSamus:
> You didn't press load in the top right after switching controllers.

I know. That was what the PR was for: not having to do that.
Having to do that is not good UX.

OldManKain:
Well breaking functionality for Nintendo Controllers in the GUI just for something you want isn't really the right way to go about fixing this issue.   If you can come up with a better solution, we would be more than willing to implement it.

SuperSamus:
> Well breaking functionality for Nintendo Controllers in the GUI just for something you want isn't really the right way to go about fixing this issue

Considering it was merged in the first place, even though I explicitly pointed it out in https://github.com/Ryujinx/Ryujinx/pull/5433#issue-1789440242, I thought it was ok.

> If you can come up with a better solution, we would be more than willing to implement it.

https://github.com/Ryujinx/Ryujinx/pull/5439#issuecomment-1624218778

OldManKain:
It was misread, it happens.

PR #5440, opened 2023-07-06
MacOS: Allow barriers inside a render pass for non-Apple GPUs and don't treat as TBDR

gleng:
Barriers inside a render pass should only be disabled for Apple GPUs (which don't support them). [MoltenVK itself ](https://github.com/KhronosGroup/MoltenVK/pull/1540)supports them for other GPUs.

Also changed `isTBDR` to be by vendor rather than setting all GPUs using MoltenVK to `true`.


marysaka (approved):

riperiperi (approved):
I think this is sensible, I left it this way because I couldn't test those platforms and figured I'd stay on the safe side and keep things consistent. If it works, this is fine.

gdkchan (approved):
lgtm, thanks.

Issue #4973, opened 2023-05-15
[Feature Request] Smart support for Xbox One wireless controller


RetroFrito:
### Overview

As it is now, if the Xbox One wireless controller shuts off either due to sitting idle, or if the batteries die, the software of course issues a notification that the game requires at least one player, but any attempts to turn the controller back on fail. The controller light blinks a few times and then turns back off.

The only option is to shut down RyujinX completely, turn on the controller, and then relaunch the software. If the game hasn't been saved, this can be a problem.

### Smaller details

No smaller details, this pretty much covers it.

### Nature of request

Somehow alter the software to allow the controller to be turned back on and resume gameplay. Or if this is already present and requires a toggle, then make this toggle On by default.

### Why would this feature be useful?

Goes without saying, really.

BigEmperor26:
hold the power X button for 5-10 seconds and you power it off. It is something related to the drivers of the controller, not Ryujinx.

RetroFrito:
@BigEmperor26 Hello, thank you for your reply but I think there might be a misunderstanding. When you say to hold the power button for 5-10 seconds to power it off, I don't think that applies to the issue I described. The problem is not with being unable to power off the controller, but rather with being able to power ON the controller if it has powered off during use with RyujinX. There is something with the software that keeps the controller from being able to power back on while the application is active.
As far as it being a driver issue, I suppose that may well be but other games that I use this controller with allow the controller to be powered back on if it has turned off. When RyujinX is running, I cannot power it back on. If I close RyujinX, I am then able to turn the controller back on and launch RyujinX again, so this leads me to believe RyujinX is, at least to some extent, the culprit.

acidion-ad:
It would be great if someone could take a look at this, as it has become a real pain.

On top of what was already acurately described above, there is no work around, _while Ryujinx is still running_, to get the controller to turn back on, physically. Minimising the game and focusing elsewhere (ie. the Steam window for example) still does not return control of the controller. As soon as Ryujinx is closed, expected functionality is returned and the controller can be turned back on. 

While this issue is under way the log is being _utterly smashed_ with "Applet 'Controller' created" entries. This alone seems bad to me.

I will attach a clean log where I have started a game with the controller powered OFF. Waitied ~10 seconds and powered the controller ON. I am then waiting idle until the controller turns itself OFF, waiting ~10 seconds and attempting to power it back ON again without success, then exiting emulation, exiting Ryujinx. 

I hope this can help.

[...]

PR #5450, opened 2023-07-12
chore: Update Ryujinx.SDL2-CS to 2.28.0

marysaka:
Close https://github.com/Ryujinx/Ryujinx/issues/4973 & https://github.com/Ryujinx/Ryujinx/issues/3301


LoneBoco:
Instead of 2.28.0, could you consider 2.28.1 instead as it includes additional fixes for 3rd party Nintendo Switch Pro Controllers?

marysaka:
> Instead of 2.28.0, could you consider 2.28.1 instead as it includes additional fixes for 3rd party Nintendo Switch Pro Controllers?

I have heard about regression from that version and am waiting for 2.28.2 (that will hopefully be coming quite soon)

marysaka:
> Compared to 2.28.0, 2.28.1 is much better. 2.28.0 has faulty led lights which lead to users thinking their controllers are off. If there's a regression it isn't documented on the SDL repository.

I got multiple reports about 3rd party xbox wireless issues on 2.28.1

As we have full control over our dependencies however, I will address this by using current 2.28.x dev tree.

PR #5452, opened 2023-07-13
Fix some Vulkan validation errors

gdkchan:
This is mostly cherry picking fixes that I already had in other open PRs, fixing some Vulkan validation errors and asserts:
- Fix error when trying to clear the stencil component for a depth-only buffer, or the depth component for a stencil-only buffer. It now does not try to clear components that doesn't exist.
- Fix wrong types in barriers for depth-stencil image barriers.
- Fix ImageView being created with storage usage bit for a Image that does not have the storage bit set.
- Fix an assert caused by `Auto` reference count going below 0, due to `IndexBufferState` doing a incorrect decrement/increment count when swapping buffers.
- Safe blit logic has been inverted so that now using the safe blit is the default behaviour, and we only enable unsafe blit for vendors where we know it works (currently just NVIDIA and Intel). I think that makes more sense since the unsafe blit is undefined behaviour, we  should only use it when it has been tested.
- Blacklist for validation errors we want to ignore has been removed. I think it doesn't have any purpose anymore as some of those errors have been fixed already, and other introduced which are not on the list.


riperiperi (approved):
LGTM, nice to see some of these fixed.

marysaka (approved):

gdkchan:
> Would you mind adding following change to PR if it is correct?
> 
> [a2c6dd9](https://github.com/Ryujinx/Ryujinx/commit/a2c6dd9cd490cc7a66cddfd11f949e078cc60eec)
> 
> I did not want to create a PR for such a small change and was not sure if it was correct.
> 
> EDIT: NVM I saw the previous PR

The change does not look correct, this is a valid format, but requires an extension. I don't know why validation complains about it, maybe we should not check the format if the extension is not supported.

Issue #4939, opened 2023-05-14
[Bug] Ryujinx's popup window doesn't show in full screen mode on macOS


HubertZhang:
### Description of the issue

I'm using Ryujinx on M1 Max MacStudio. When using Ryujinx in full screen mode, popup window will not show and blocks the whole program.

### Reproduction steps

- Open "Mii edit applet"
- Enter fullscreen mode
- Choose Actions > Stop emulation
- Confirmation popup doesn't show and whole program stops responding.

### Log file

[Ryujinx_1.1.799_2023-05-14_14-00-14.log](https://github.com/Ryujinx/Ryujinx/files/11471249/Ryujinx_1.1.799_2023-05-14_14-00-14.log)

### OS

macOS 13.3.1

### Ryujinx version

1.1.799

### CPU

M1 Max

### GPU

M1 Max

### RAM

32GB

MutantAura:
Duplicate of https://github.com/Ryujinx/Ryujinx/issues/5186

PR #5456, opened 2023-07-16
Ava UI: Remove `IsActive` checks from dialog methods

### LINKED ISSUE #5186: [Bug] Avalonia dialogs not showing properly under Full-screen or loss of focus
### Description of the issue

When Ryujinx main window is maximized and a prompt for text pops, the window will not show at all, and the game will stall since I cannot interact with the window. 

Normally in a not maximized window, it shows properly like this:

Screen Shot 2023-06-01 at 6 14 45 午後

But when the game asks for input in full-screen, the software keyboard window will not show. You can see both windows after quitting full-screen with Mission Control (see all your open windows in all spaces), and entering App Exposé (to see all the open windows in the current app), but you cannot interact with either of them. You will have to force Ryujinx to quit.

Screen Shot 2023-06-01 at 6 15 49 午後

### Reproduction steps

Launch a game that can ask for input, go in full-screen, trigger the prompt.

### Log file

[Ryujinx_1.1.848_2023-06-01_16-23-32.log](https://github.com/Ryujinx/Ryujinx/files/11624108/Ryujinx_1.1.848_2023-06-01_16-23-32.log)

### OS

macOS 12.6.6

### Ryujinx version

1.1.848

### Game version

1.0.0

### CPU

Apple M1 Pro

### GPU

Apple M1 Pro

### RAM

16 GB

### List of applied mods

60FPS

### Additional context?

_No response_

MutantAura:
There is no guarantee that the main window, or any window for that matter will be focused and this should not impact the ability to spawn content dialogs. Should resolve people getting "softlocked" on macOS and Linux when the window is not active and a software keyboard is triggered.

Fixes #5186

### DIFF
--- a/src/Ryujinx.Ava/UI/Applet/AvaHostUiHandler.cs
+++ b/src/Ryujinx.Ava/UI/Applet/AvaHostUiHandler.cs
@@ -53,8 +53,6 @@ namespace Ryujinx.Ava.UI.Applet
 
                     bool opened = false;
 
-                    _parent.Activate();
-
                     UserResult response = await ContentDialogHelper.ShowDeferredContentDialog(_parent,
                        title,
                        message,
--- a/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
+++ b/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
@@ -371,7 +371,9 @@ namespace Ryujinx.Ava.UI.Helpers
                 }
                 else
                 {
-                    result = await contentDialog.ShowAsync();
+                    result = ContentDialogResult.None;
+
+                    Logger.Warning?.Print(LogClass.Ui, "Content dialog overlay failed to populate. Default value has been returned.");
                 }
 
                 return result;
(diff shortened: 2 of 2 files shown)


marysaka (approved):

siegmund-heiss-ich:
I wanted to test current master vs this pr, but something fixed the issue on macos already.
Maybe code cleanup? I don't want to dig into it and find out which commit it was, because it works anyways now.

So #4939 and the linked  issue in your pr description are already fixed.
You could also check the box in Ava issue tracker #3662.

Maybe the changes of this pr affect linux? Can't tell because I just use macOS native atm..

MutantAura:
I will test master again later. But while finding the cause, master still had this issue on my Mac so I'm not sure about it being fixed.

riperiperi (approved):
This makes sense, not sure why this check was there in the first place.

Littl3Guy:
There's a regression on Windows causing dialog boxes not to appear affecting Manage Title Updates, Manage DLC, Check for Updates, and About.

PR #5357, opened 2023-06-24
[ARMeilleure] Address dotnet-format issues

TSRBerry:
Base PR: #4670

PR #5457, opened 2023-07-16
[CPU] Hotfix missing ToNearest rounding mode cases

TSRBerry:
This PR fixes a regression from #5357 which led to some failing tests.


marysaka (approved):

riperiperi (approved):
Very odd that this was removed.

PR #5380, opened 2023-06-24
[Ryujinx.HLE] Address dotnet-format issues

TSRBerry:
Base PR: #4670 

---

Depends on #5362 
Depends on #5358

PR #5459, opened 2023-07-17
[Hotfix] sockets: Resolve empty port requests to 0 again

TSRBerry:
This PR fixes a regression from #5380 which was found by @demize.

`GetAddrInfoRequest()` can be called with an empty `service` parameter, which means that the requester doesn't care about the port. The logic introduced by #5380 didn't handle this case properly and would instead return `GaiError.Service`.

The fix here is to add another condition to the if-statement so empty strings aren't causing errors, because we still want to handle those properly.


marysaka (approved):

demize:
Thanks for the quick fix, Berry! I can confirm the issue is fixed with the build artifacts posted for this PR.

riperiperi (approved):

PR #5468, opened 2023-07-19
HLE: Fix corrupted Mii structs

TSRBerry:
This PR fixes a regression from #5380, which was found and debugged by @riperiperi.

Steps to reproduce the previous issue:
1. Start Mario Kart 8 Deluxe
2. Enter VS mode
3. Play one race with any character
4. Return to the course selection after the race finished
5. Crash


riperiperi (approved):
LGTM, swapping this for the better defined array type makes sense.

gdkchan (approved):

PR #5502, opened 2023-07-30
CPU (A64): Add Fmaxp & Fminp Scalar Inst.s, Fast & Slow Paths; with Tests.

LDj3SNuD:
Contributes to https://github.com/Ryujinx/Ryujinx-Games-List/issues/2828.

Now the game eventually progresses but would then seem to crash due to a problem with the `SuatomB` shader instruction:
```
00:02:11.443 |E| GPU.MainThread Application : Unhandled exception caught: System.NullReferenceException: Object reference not set to an instance of an object.
   at Ryujinx.Graphics.Shader.Translation.EmitterContextInsts.Copy(EmitterContext context, Operand d, Operand a) in \src\Ryujinx.Graphics.Shader\Translation\EmitterContextInsts.cs:line 217
   at Ryujinx.Graphics.Shader.Instructions.InstEmit.EmitSuatom(EmitterContext context, SuDim dimensions, SuatomOp atomicOp, SuatomSize size, Int32 imm, Int32 srcA, Int32 srcB, Int32 srcC, Int32 dest, Boolean byteAddress, Boolean isBindless, Boolean compareAndSwap) in \src\Ryujinx.Graphics.Shader\Instructions\InstEmitSurface.cs:line 290
   at Ryujinx.Graphics.Shader.Instructions.InstEmit.SuatomB(EmitterContext context) in \src\Ryujinx.Graphics.Shader\Instructions\InstEmitSurface.cs:line 18
```

Nop-ing the instruction, the game goes ingame:
image

PR #5504, opened 2023-07-30
(Graphics.Shader): Handle EmitSuatom constant dests and EmitSuld zero dest reg.

LDj3SNuD:
Contributes to https://github.com/Ryujinx/Ryujinx-Games-List/issues/2828.

Now the game goes ingame:
https://github.com/Ryujinx/Ryujinx/assets/35856442/5bb2c6ff-62a5-4e4f-a73a-45d24bbcd6b6

Reference: https://github.com/Ryujinx/Ryujinx/pull/5502#issuecomment-1657210852

### DIFF
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmitSurface.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmitSurface.cs
@@ -218,8 +218,6 @@ namespace Ryujinx.Graphics.Shader.Instructions
                 return context.Copy(Register(srcB++, RegisterType.Gpr));
             }
 
-            Operand d = dest != RegisterConsts.RegisterZeroIndex ? Register(dest, RegisterType.Gpr) : null;
-
             List<Operand> sourcesList = new();
 
             if (isBindless)
@@ -285,9 +283,13 @@ namespace Ryujinx.Graphics.Shader.Instructions
                 TextureOperation.DefaultCbufSlot,
                 imm);
 
-            Operand res = context.ImageAtomic(type, format, flags, binding, sources);
+            if (dest != RegisterConsts.RegisterZeroIndex)
+            {
+                Operand d = Register(dest, RegisterType.Gpr);
+                Operand res = context.ImageAtomic(type, format, flags, binding, sources);
 
-            context.Copy(d, res);
+                context.Copy(d, res);
+            }
         }
 
         private static void EmitSuld(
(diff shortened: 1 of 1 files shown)


gdkchan:
Can you fix `EmitSuld` too? It has the same issue. But in that case, `dests` array will be empty if `srcB` is equal to `RegisterZeroIndex`. There are places that accesses `dests[0]` so it is assuming that the size will be at least 1. Also I'm not sure what the effect of passing an empty array on the `Operation` would be in that case. For the `EmitSuld` case I think just doing `if (srcB == RegisterConsts.RegisterZeroIndex) return;` (properly formatted) would be fine since unlike SUATOM, SULD with a zero destination register should be a no-op.

LDj3SNuD:
surely, in a few hours

src/Ryujinx.Graphics.Shader/Instructions/InstEmitSurface.cs
```diff
@@ -285,9 +283,13 @@ Operand Rb()
                 TextureOperation.DefaultCbufSlot,
                 imm);
 
-            Operand res = context.ImageAtomic(type, format, flags, binding, sources);
+            if (dest != RegisterConsts.RegisterZeroIndex)
```
gdkchan, line 286: This is not completely correct. The `ImageAtomic` operation should be done regardless of the destination reg being RZ or not. That's because `ImageAtomic` also modifies the texture contents, it is an atomic operation. Even if you don't care about the result, it still has "side effects".

gdkchan: The most simple fix is just changing `Operand d = dest != RegisterConsts.RegisterZeroIndex ? Register(dest, RegisterType.Gpr) : null;` to `Operand d = Register(dest, RegisterType.Gpr);`. `Copy` method can deal with `d` being a constant.

LDj3SNuD:
okay, I will correct, I have no experience in GPU

LDj3SNuD:
you are right, I will correct as soon as I can

LDj3SNuD:
Let me know if this is okay; and if I need to increment any flags.

gdkchan (approved):
lgtm, thanks.

gdkchan:
> Let me know if this is okay; and if I need to increment any flags. I confirm that the game works with those changes.

It looks correct now (please test with Jurassic Park to be sure, with shader cache disabled or purged, otherwise it might be using the shaders from your previous test).

There is no need to increment the shader cache version since the bug makes it crash before the shader is generated, so the emulator wouldn't have a chance to save the broken shader in the cache.

LDj3SNuD:
Ok thanks, I always retest with the caches disabled or cleared.

riperiperi (approved):
LGTM

PR #5507, opened 2023-07-31
Add slightly better workaround for current workflow issues

TSRBerry:
This PR attempts to improve our current workflow issues by adding a slightly better workaround using my [unstable-commands](https://github.com/marketplace/actions/unstable-commands) action which I published yesterday.

I made this action primarily because we keep having issues with workflows that are either timing out or crashing for unknown reasons. Since we have trouble even replicating these issues on our own runners or local machines, the best thing we can do right now is to improve how we deal with them inside the workflows.

My action will retry the specified command if it runs into the timeout or one of the specified retry-codes was encountered. By default this action will only retry the command(s) up to 3 times before it gives up, we can of course adjust that if we think that could help.

In PR #4673 @gdkchan noticed that `dotnet format style` also segfaults sometimes, so I added my action to that command as well. I didn't add it to the whitespace step, since I don't think it could ever segfault.


marysaka (approved):

riperiperi (approved):
Seems to make sense. Would obviously prefer if format didn't randomly blow up. 🥴

PR #5509, opened 2023-07-31
Delete ShaderConfig and organize shader resources/definitions better

gdkchan:
This is just refactoring, it shouldn't really affect the code that is generated. The main goal of this change is getting rid of the `ShaderConfig` class. The reason is that it became some sort of "catch-all" class that would contain all kinds of state used by any phase of the shader generation process. On top of that, the translation process would also mutate the `ShaderConfig` state, which means that, for example, two calls to the `TranslatorContext.Translate` methods might not produce the same results, since the first call would mutate state that is used on the translation process.

This change addresses those issues by splitting `ShaderConfig` into other smaller and more scoped classes. For the most part they are also immutable, the main exception being `AttributeUsage`, but the methods that mutate state here are explicit and are not called as part of the translation process.  So (at least in theory) two or more calls to `Translate` should produce the same results now. The emulator doesn't currently depend on this to work, but in general this is how any method is expected to work.

Another notable change is that the `Rewriter` class, that had many methods responsible for doing transformations to the shader code IR was split. Now each transformation has its own separate class, each one implementing a `ITransformPass` interface that exposes two methods, `IsEnabled` that indicates if the pass should run at all, and `RunPass` that performs the transform in the code. The `IsEnabled` check means we can skip transforms that are not needed entirely, instead of checking each one in the inner loop.

Lastly a new `GpuGraphicsState` struct has been introduced, it groups a lot of the GPU state that was previously queried using individual methods on the `IGpuAccessor`. This simplifies the implementation and allows some code to be de-duplicated between the main and the shader cache GPU accessors.

This will be required for geometry shader emulation on macOS since those changes are currently done on top of this branch. It is not really a strict requirement, I mostly did it this way to reduce the work required for the rebase later on. But the transform pass change in particular is useful for geometry shader emulation, as I can have a separate transform for replacing vertex/geometry related IR instructions with something that works in a compute shader.

Testing is welcome.


devmachine1:
hi im new to ryujinx and github in general, I tried playing totk with your changes just for fun and for some reason the game didnt detect my shaders and it just overwrote them. But when I use the same shaders with the "normal" release (1.1.973), all goes as expected. Just wanted to ask if the behavior of your changes is as you wanted them to be

Littl3Guy:
Tested a handful of games, no regressions found.

gdkchan:
> hi im new to ryujinx and github in general, I tried playing totk with your changes just for fun and for some reason the game didnt detect my shaders and it just overwrote them. But when I use the same shaders with the "normal" release (1.1.973), all goes as expected. Just wanted to ask if the behavior of your changes is as you wanted them to be

Not really. I did test The Legend of Zelda: Tears of the Kingdom here, which shaders created on master, then switched to this branch, and it worked as expected, as in the shaders loaded.

src/Ryujinx.Graphics.Shader/Translation/ShaderDefinitions.cs
```diff
@@ -0,0 +1,315 @@
...
+            bool gpPassthrough,
+            int threadsPerInputPrimitive,
+            OutputTopology outputTopology,
+            int maxOutputVertices,
+            ImapPixelType[] imapTypes,
+            int omapTargets,
+            bool omapSampleMask,
+            bool omapDepth,
+            bool transformFeedbackEnabled,
+            ulong transformFeedkbackVecMap,
```
riperiperi, line 143: `Feedkback`?

riperiperi (commented):
Looks good. I'm going to test a bunch of stuff since it looks like pretty much everything was moved around.

riperiperi (approved):
Tested a bunch of games with/without res scale, looks good!

src/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
```diff
@@ -356,7 +407,54 @@ private static void SetUserAttributeUses(ShaderConfig config, InstName name, ulo
...
+                                    }
+                                    break;
+                                case AttributeConsts.InstanceId:
+                                    if (definitions.Stage == ShaderStage.Vertex)
+                                    {
+                                        context.SetUsedFeature(FeatureFlags.InstanceId);
+                                    }
+                                    break;
+                            }
+                        }
```
TSRBerry, line 457:
```suggestion
                        switch (attr, isStore)
                        {
                            case (AttributeConsts.Layer, true):
                                if (definitions.Stage != ShaderStage.Fragment)
                                {
                                    context.SetUsedFeature(FeatureFlags.RtLayer);
                                }
                                break;
                            case (AttributeConsts.ClipDistance0, true):
                            case (AttributeConsts.ClipDistance1, true):
                            case (AttributeConsts.ClipDistance2, true):
                            case (AttributeConsts.ClipDistance3, true):
                            case (AttributeConsts.ClipDistance4, true):
                            case (AttributeConsts.ClipDistance5, true):
                            case (AttributeConsts.ClipDistance6, true):
                            case (AttributeConsts.ClipDistance7, true):
                                if (definitions.Stage == ShaderStage.Vertex)
                                {
                                    context.SetClipDistanceWritten((attr - AttributeConsts.ClipDistance0) / 4);
                                }
                                break;
                            case (AttributeConsts.PositionX, false):
                            case (AttributeConsts.PositionY, false):
                                if (definitions.Stage == ShaderStage.Fragment)
                                {
                                    context.SetUsedFeature(FeatureFlags.FragCoordXY);
                                }
                                break;
                            case (AttributeConsts.InstanceId, false):
                                if (definitions.Stage == ShaderStage.Vertex)
                                {
                                    context.SetUsedFeature(FeatureFlags.InstanceId);
                                }
                                break;
                        }
```

Not sure if that's actually more readable, but what do you think? Maybe `definitions.Stage` could also be moved into the switch-statement?

src/Ryujinx.Graphics.Shader/StructuredIr/AstOptimizer.cs
```diff
@@ -15,7 +15,7 @@ public static void Optimize(StructuredProgramContext context)
 
             // When debug mode is enabled, we disable expression propagation
             // (this makes comparison with the disassembly easier).
-            if (!context.Config.Options.Flags.HasFlag(TranslationFlags.DebugMode))
+            if (!context.DebugMode)
```
TSRBerry, line 18: Is there a reason this wasn't coupled to the project debug configuration?

src/Ryujinx.Graphics.Shader/Translation/TransformFeedbackOutput.cs
```diff
@@ -0,0 +1,18 @@
...
+    readonly struct TransformFeedbackOutput
+    {
+        public readonly bool Valid;
+        public readonly int Buffer;
+        public readonly int Offset;
+        public readonly int Stride;
+
+        public TransformFeedbackOutput(int buffer, int offset, int stride)
+        {
+            Valid = true;
```
TSRBerry, line 12: Why does this struct have a field that's always true? I'm guessing there is a reason, but I don't understand this on my own.

TSRBerry (approved):
This was very hard to review for me. I usually go through these kind of PRs one commit at a time, but even then it was a little overwhelming for me. I guess I'm just not comfortable enough with this part of the codebase.

I found a few things where I wanted to ask some questions, but as far as I can tell everything looks good! 
I also tested a few games and didn't notice a difference, so it should be fine? :D

src/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
gdkchan: I would prefer to move `definitions.Stage` into the switch-statement instead of doing it for `isStore`.

TSRBerry: Yeah that could work as well! I'm not sure how you write that for `definitions.Stage != ShaderStage.Fragment` tho.

src/Ryujinx.Graphics.Shader/StructuredIr/AstOptimizer.cs
gdkchan: `DebugMode` makes it generate comments in the shader code with the original instruction name, encoding and offset in the shader. It could be enabled only for debug builds, but I usually test with release builds, so I find the way how it currently works useful. I think `DebugMode` flag also currently disables expression propagation (which causes other issues). We should probably just always disable it, it's not very useful.

TSRBerry: I see, thank you!

src/Ryujinx.Graphics.Shader/Translation/TransformFeedbackOutput.cs
gdkchan: I will be false if the struct is not actually initialized with this construct, for example by doing `TransformFeedbackOutput output = default;`, `TransformFeedbackOutput output = new();` or simply by not initializing it at all.

TSRBerry: Ohhh that makes sense!

src/Ryujinx.Graphics.Shader/Decoders/Decoder.cs
gdkchan: it would need 5 cases for Compute, Vertex, Geometry and the 2 Tessellation stages.

gdkchan: Actually it is only necessary for vertex, geometry and tessellation, setting gl_Layer from a compute shader is not valid at all.

TSRBerry: Hmm in that case the current version is probably more readable. Maybe it's better to leave it that way.

PR #5525, opened 2023-08-03
GPU: Enable VK_EXT_4444_formats

sunshineinabox:
Enables VK_EXT_4444_formats if GPU supports them resolves a validation error

`GUI.RenderLoop Gpu UserCallback: Validation Error: [ VUID-vkGetPhysicalDeviceFormatProperties-format-parameter ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x4addcb92 | vkGetPhysicalDeviceFormatProperties: value of format (1000340001) does not fall within the begin..end range of the core VkFormat enumeration tokens and is not an extension added token. The Vulkan spec states: format must be a valid VkFormat value (https://vulkan.lunarg.com/doc/view/1.3.250.1/windows/1.3-extensions/vkspec.html#VUID-vkGetPhysicalDeviceFormatProperties-format-parameter)`


gdkchan (approved):
lgtm, thanks.

riperiperi (approved):
Makes sense.

PR #5531, opened 2023-08-04
[Vulkan / Avalonia] Implement Color Space Passthrough option

jcm93:
### Context
Ryujinx outputs sRGB in conformance with the Switch hardware. This makes sense and is correct behavior, but people on wide gamut displays might want Ryujinx to be able to output wide gamut colors, in order to take advantage of their display's capabilities.

This PR implements the option for Ryujinx's Vulkan backend to attempt to not explicitly select a color space during output, and instead pass the color information through Vulkan to the rest of the display engine. This should result in Ryujinx displaying with RGB values stretched to the color space that the operating system and display are currently using.

Screenshot 2023-08-03 at 12 00 59 PM
Screenshot 2023-08-03 at 1 46 10 PM

### Description

Adds a "Enable Color Space Passthrough" checkbox option under the Graphics settings tab in Avalonia. In Vulkan's ``Window.cs``, inside the ``ChooseSwapSurfaceFormat`` function, a check for the new option status is implemented. If enabled, the function checks for ``ColorSpaceKHR.SpacePassThroughExt`` as an exposed color space and if it is available, returns it. 

``_vSyncEnabled`` is also renamed ``_swapchainIsDirty``, as it is now repurposed as a general flag indicating that the swapchain needs recreation when either vertical sync or the new color space option is toggled.

The relevant change to the configuration file and its version is implemented; as a first time contributor I'm not 100% sure I did this right but hopefully I did.

### Other Considerations

Currently, the option is only implemented in Avalonia and Vulkan. The reason for focusing on these two is that this PR primarily affects Mac users, who typically have wide gamut displays with this color space option exposed to Vulkan via MoltenVK. I do not have enough information to know how or if this option will work as expected on Windows or Linux.

Preliminary research suggests that this option will only function if the operating system and Vulkan driver together 'properly' allow surfaces to select their color space on an ad-hoc basis as macOS does. Unclear how Windows or Linux handles this in all cases.

I considered exposing this option *only* if the user is on macOS, but was not familiar enough with Avalonia's UI logic to do so. I also thought that it may not be harmful to leave it exposed for other systems where it will potentially work.

### DIFF
--- a/src/Ryujinx.Graphics.GAL/Multithreading/ThreadedWindow.cs
+++ b/src/Ryujinx.Graphics.GAL/Multithreading/ThreadedWindow.cs
@@ -38,5 +38,7 @@ namespace Ryujinx.Graphics.GAL.Multithreading
         public void SetScalingFilter(ScalingFilter type) { }
 
         public void SetScalingFilterLevel(float level) { }
+
+        public void SetColorSpacePassthrough(bool colorSpacePassthroughEnabled) { }
     }
 }
--- a/src/Ryujinx.Graphics.OpenGL/Window.cs
+++ b/src/Ryujinx.Graphics.OpenGL/Window.cs
@@ -307,6 +307,8 @@ namespace Ryujinx.Graphics.OpenGL
             _updateScalingFilter = true;
         }
 
+        public void SetColorSpacePassthrough(bool colorSpacePassthroughEnabled) { }
+
         private void UpdateEffect()
         {
             if (_updateEffect)
--- a/src/Ryujinx.Graphics.Vulkan/Window.cs
+++ b/src/Ryujinx.Graphics.Vulkan/Window.cs
@@ -28,7 +28,7 @@ namespace Ryujinx.Graphics.Vulkan
         private int _width;
         private int _height;
         private bool _vsyncEnabled;
-        private bool _vsyncModeChanged;
+        private bool _swapchainIsDirty;
         private VkFormat _format;
         private AntiAliasing _currentAntiAliasing;
         private bool _updateEffect;
@@ -38,6 +38,7 @@ namespace Ryujinx.Graphics.Vulkan
         private float _scalingFilterLevel;
         private bool _updateScalingFilter;
         private ScalingFilter _currentScalingFilter;
+        private bool _colorSpacePassthroughEnabled;
 
         public unsafe Window(VulkanRenderer gd, SurfaceKHR surface, PhysicalDevice physicalDevice, Device device)
         {
@@ -60,7 +61,7 @@ namespace Ryujinx.Graphics.Vulkan
         private void RecreateSwapchain()
         {
             var oldSwapchain = _swapchain;
-            _vsyncModeChanged = false;
+            _swapchainIsDirty = false;
 
             for (int i = 0; i < _swapchainImageViews.Length; i++)
             {
@@ -106,7 +107,7 @@ namespace Ryujinx.Graphics.Vulkan
                 imageCount = capabilities.MaxImageCount;
             }
 
-            var surfaceFormat = ChooseSwapSurfaceFormat(surfaceFormats);
+            var surfaceFormat = ChooseSwapSurfaceFormat(surfaceFormats, _colorSpacePassthroughEnabled);
 
             var extent = ChooseSwapExtent(capabilities);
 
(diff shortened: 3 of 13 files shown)


src/Ryujinx.Graphics.OpenGL/Window.cs
```diff
@@ -307,6 +307,11 @@ public void SetScalingFilter(ScalingFilter type)
             _updateScalingFilter = true;
         }
 
+        public void SetColorSpacePassthrough(bool colorSpacePassthroughEnabled)
+        {
+            throw new NotImplementedException();
```
siegmund-heiss-ich, line 312: I would just log a warning that this option couldn't be applied to OpenGL.
No need to throw an exception.
Or is it planned to add this option for this graphics backend too?

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -1391,6 +1400,15 @@ public ConfigurationLoadResult Load(ConfigurationFileFormat configurationFileFor
                 configurationFileUpdated = true;
             }
 
+            if (configurationFileFormat.Version < 48)
+            {
+                Ryujinx.Common.Logging.Logger.Warning?.Print(LogClass.Application, $"Outdated configuration version {configurationFileFormat.Version}, migrating to version 48.");
+
+                configurationFileFormat.EnableColorSpacePassthrough = true;
```
siegmund-heiss-ich, line 1407: I'd prefer that this option is false by default.
Otherwise new users wouldn't know that they don't play the game like originally intended.

siegmund-heiss-ich:
Nice work!
Looking forward for this PR to get merged.

Tested it on m1 pro mbp.
off
on

src/Ryujinx.Graphics.OpenGL/Window.cs
riperiperi: Yes, this should probably do nothing.

src/Ryujinx.Graphics.GAL/Multithreading/ThreadedWindow.cs
```diff
@@ -38,5 +38,7 @@ public void SetSize(int width, int height)
         public void SetScalingFilter(ScalingFilter type) { }
 
         public void SetScalingFilterLevel(float level) { }
+
+        public void SetColorSpacePassthrough(bool colorSpacePassthroughEnabled) { }
```
gdkchan, line 42: As a side note, we should implement those on the threaded renderer eventually. Doesn't need to be done on this PR since this is not being done for the other settings, and they are not currently called on the threaded renderer wrapper, but there's no reason for those to do nothing.

src/Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -38,6 +38,7 @@ class Window : WindowBase, IDisposable
         private float _scalingFilterLevel;
         private bool _updateScalingFilter;
         private ScalingFilter _currentScalingFilter;
+        private static bool _colorSpacePassthroughEnabled;
```
gdkchan, line 41: Why is that `static`?

src/Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -187,7 +188,14 @@ private static SurfaceFormatKHR ChooseSwapSurfaceFormat(SurfaceFormatKHR[] avail
 
             foreach (var format in availableFormats)
             {
-                if (format.Format == VkFormat.B8G8R8A8Unorm && format.ColorSpace == ColorSpaceKHR.PaceSrgbNonlinearKhr)
+                if (_colorSpacePassthroughEnabled)
```
gdkchan, line 191: I think we should prefer a sRGB color space with `B8G8R8A8Unorm` format if the passthrough option is enabled, but such color space is not available, like we would do if the setting was disabled. As it is now will it will just pick the first format if there's none with passthrough color space. Basically what I'm suggesting is having two loops, one that will look for a `B8G8R8A8Unorm` format with passthrough color space, and if none is found, it will have a second loop looking for a sRGB one, as it did before this change.

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
jcm93: Ah, woops. Yeah, I meant to have it default to false but looks like I left some testing values in. Will fix here and ``GraphicsConfig.cs``.

gdkchan: Yea, it should be disabled by default.

src/Ryujinx.Graphics.Vulkan/Window.cs
jcm93: The ``ChooseSwapSurfaceFormat`` function is declared static, so it wants an object reference if ``_colorSpacePassthroughEnabled`` isn't static. I just wasn't sure if it would be more correct to remove that declaration from the function, or just declare the new bool as static.

gdkchan: It should not be static. You should make `ChooseSwapSurfaceFormat` not static instead. Or keep the function static and pass `colorSpacePassthroughEnabled` as an argument to this function.

src/Ryujinx.Graphics.Vulkan/Window.cs
jcm93: Woops, good catch. Will fix.

src/Ryujinx.Graphics.OpenGL/Window.cs
jcm93: Changed to leave as stub.

src/Ryujinx.Graphics.Vulkan/Window.cs
jcm93: I left the function static and passed the argument to the function since that seemed more neat.

src/Ryujinx.Graphics.Vulkan/Window.cs
```diff
@@ -178,22 +179,40 @@ private unsafe Auto<DisposableImageView> CreateSwapchainImageView(Image swapchai
...
-                if (format.Format == VkFormat.B8G8R8A8Unorm && format.ColorSpace == ColorSpaceKHR.PaceSrgbNonlinearKhr)
+                foreach (var format in availableFormats)
                 {
-                    return format;
+                    if (format.Format == VkFormat.B8G8R8A8Unorm && format.ColorSpace == ColorSpaceKHR.SpacePassThroughExt)
+                    {
+                        formatToReturn = format;
+                        break;
+                    }
+                    else if (format.Format == VkFormat.B8G8R8A8Unorm && format.ColorSpace == ColorSpaceKHR.PaceSrgbNonlinearKhr)
```
gdkchan, line 198: doesn't that mean that if the passthrough color space is located after the sRGB one in the array, it will prefer the sRGB one?

The change I suggested was basically:
```cs
            if (colorSpacePassthroughEnabled)
            {
                foreach (var format in availableFormats)
                {
                    if (format.Format == VkFormat.B8G8R8A8Unorm && format.ColorSpace == ColorSpaceKHR.SpacePassThroughExt)
                    {
                        return format;
                    }
                }
            }
```
before the existing loop. Everything else could remain unchanged.

jcm93: No; sorry, I modified the control flow slightly because it seemed more readable. If passthrough is enabled and it finds the passthrough color space, it will set ``formatToReturn`` and exit the foreach early. Otherwise it'll set ``formatToReturn`` to the sRGB format if it's found and return it at the end.

gdkchan: Right, I missed the `break;`. I guess you did this to avoid the need to loop twice in the case where passthrough is enabled, but is not supported in the host.

gdkchan (approved):
lgtm, thanks.

siegmund-heiss-ich (approved):

marysaka (approved):

PR #5541, opened 2023-08-06
Simplify resolution scale updates

gdkchan:
Now that the support buffer update is not done on the backend anymore, we can update the support buffer array directly, instead of having the intermediary array for the scales that is set to the support buffer later. This makes the code a little bit simpler.

Testing is welcome, needs to be tested with resolution scaling of course.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
+++ b/src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
@@ -376,7 +376,7 @@ namespace Ryujinx.Graphics.Vulkan
             var program = _program;
             var bindingSegments = program.BindingSegments[setIndex];
 
-            if (bindingSegments.Length == 0 && setIndex != PipelineBase.UniformSetIndex)
+            if (bindingSegments.Length == 0)
             {
                 return;
             }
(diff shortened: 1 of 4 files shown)


src/Ryujinx.Graphics.Vulkan/DescriptorSetUpdater.cs
```diff
@@ -376,7 +376,7 @@ private void UpdateAndBind(CommandBufferScoped cbs, int setIndex, PipelineBindPo
             var program = _program;
             var bindingSegments = program.BindingSegments[setIndex];
 
-            if (bindingSegments.Length == 0 && setIndex != PipelineBase.UniformSetIndex)
+            if (bindingSegments.Length == 0)
```
gdkchan, line 379: The `setIndex != PipelineBase.UniformSetIndex` check is not needed anymore. It was needed before because the support buffer had special handling here, and had to be updated even if the guest shader had no bindings. Since it is a buffer like any other now, no special handling is needed. If `bindingSegments.Length == 0` is zero, it would be because there's no support buffer, but right now that doesn't happen in practice since the support buffer is always included. In the future, we could remove it if unused.

riperiperi (approved):
Cool, this was only introduced to reduce the # of commands sent over the multithreaded backend, but if it's all on one side that's not a problem.

marysaka (approved):

PR #5545, opened 2023-08-09
Replacing 'Assembly.GetExecutingAssembly()' with 'Type.Assembly'

marco-carvalho:
Using `Type.Assembly` to get the current assembly is nearly free in terms of performance; it’s a simple property access. On the other hand, `Assembly.GetExecutingAssembly()  can take up to 30 times as long because it walks up the call stack to find the assembly.

Note that `Assembly.GetExecutingAssembly()` is different than `Type.Assembly` because it dynamically returns the assembly that contains the startup object of the currently executed application. For example, if executed from an application it will return the application assembly, but if executed from a unit test project it could return the unit test assembly. `Type.Assembly` always returns the assembly that contains the specified type.

Benchmark:

```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    [Benchmark]
    public void AssemblyGetExecutingAssembly()
    {
        Assembly _ = Assembly.GetExecutingAssembly();
    }

    [Benchmark]
    public void TypeAssembly()
    {
        Assembly _ = typeof(Test).Assembly;
    }
}
```

Results:
```sh
|                       Method |      Mean |    Error |   StdDev | Rank | Allocated |
|----------------------------- |----------:|---------:|---------:|-----:|----------:|
|                 TypeAssembly |  11.08 ns | 0.046 ns | 0.043 ns |    1 |         - |
| AssemblyGetExecutingAssembly | 691.97 ns | 2.012 ns | 1.680 ns |    2 |         - |
```

### DIFF
--- a/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardRendererBase.cs
+++ b/src/Ryujinx.HLE/HOS/Applets/SoftwareKeyboard/SoftwareKeyboardRendererBase.cs
@@ -68,15 +68,15 @@ namespace Ryujinx.HLE.HOS.Applets.SoftwareKeyboard
             int ryujinxLogoSize = 32;
 
             string ryujinxIconPath = "Ryujinx.HLE.HOS.Applets.SoftwareKeyboard.Resources.Logo_Ryujinx.png";
-            _ryujinxLogo = LoadResource(Assembly.GetExecutingAssembly(), ryujinxIconPath, ryujinxLogoSize, ryujinxLogoSize);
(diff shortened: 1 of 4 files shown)


marco-carvalho:
IDK why the workflow failed.

TSRBerry:
It tells you why:
image

marco-carvalho:
But I didn't touched this `EnvironmentTest.cs`.
image

TSRBerry:
Might just be a new rule in this case.

marysaka (approved):

TSRBerry (approved):

PR #5547, opened 2023-08-09
[Hotfix] hid: Prevent out of bounds array access

TSRBerry:
After #5518 was merged, people reported a lot of crashes indicating an out of bounds access within hid.

This PR fixes this issue by returning from `SetSupportedPlayer()` before the array gets accessed should the index be too large.

With this change ARCropolis works properly again.

---

It seems none of my PRs can be merged without requiring a hotfix shortly after that. :upside_down_face:

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Hid/HidDevices/NpadDevices.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Hid/HidDevices/NpadDevices.cs
@@ -70,6 +70,11 @@ namespace Ryujinx.HLE.HOS.Services.Hid
 
         internal void SetSupportedPlayer(PlayerIndex player, bool supported = true)
         {
+            if ((int)player >= _supportedPlayers.Length)
+            {
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


src/Ryujinx.HLE/HOS/Services/Hid/HidDevices/NpadDevices.cs
```diff
@@ -70,6 +70,11 @@ internal void ClearSupportedPlayers()
 
         internal void SetSupportedPlayer(PlayerIndex player, bool supported = true)
         {
+            if ((int)player >= _supportedPlayers.Length)
```
gdkchan, line 73: Did you intend to cast it to `uint`? The current check won't protect in case the player index is a negative value.

TSRBerry: I just copied it from below, but protecting against a negative value also sounds like a good idea. Let me add that.

gdkchan (approved):
lgtm, thanks.

PR #5554, opened 2023-08-11
GPU: Track basic buffer copies that modify texture memory

### LINKED ISSUE #4550: [Regression] Icons in Dragon Quest Builders are random/broken
### Description of the issue

After #3754 the icons in Dragon Quest Builders either show as broken or exchanged with other icons from the game

1.1.566

image

1.1.565

image

### Reproduction steps

Just booting the game

### Log file

N/A

### OS

Windows 10

### Ryujinx version

1.1.665

### Game version

1.0.0

### CPU

Ryzen 3600

### GPU

NVIDIA RTX 3060 Ti

### RAM

16 GB

### List of applied mods

_No response_

### Additional context?

_No response_

riperiperi:
This branch changes the buffer copy fast path to notify memory tracking for all resources that aren't buffers. This fixes cases where games would copy buffer data directly into texture memory, which before would only work if the texture did not already exist. I imagine this happens when the guest driver is moving data between allocations or uploading it.

Since this only affects the fast path, cases where the source data has been modified from GPU (fast path copy destination doesn't count) will still fail to notify the texture, though I don't imagine games will do this. This should be resolved in future.

This fixes some texture issues with guest OpenGL games on switch, such as Dragon Quest Builders.

This may also be useful in future for games that move shader data around memory, if we end up using memory tracking for those.

Fixes #4550.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Memory/BufferCache.cs
@@ -344,7 +344,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
                 // Optimization: If the data being copied is already in memory, then copy it directly instead of flushing from GPU.
 
                 dstBuffer.ClearModified(dstAddress, size);
-                memoryManager.Physical.WriteUntracked(dstAddress, memoryManager.Physical.GetSpan(srcAddress, (int)size));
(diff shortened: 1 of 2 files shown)


GamerzHell9137:
- Fixes Dragon Quest Builders broken icons

Before:

image
image

After:

image
image

gdkchan (approved):
lgtm, thanks. Nice to see the "regression" fixed.

gdkchan:
Fixes #4550?

GamerzHell9137:
> Fixes #4550?

Yes

marysaka (approved):

PR #5558, opened 2023-08-12
Ava UI: Allow DPI switching on Windows

riperiperi:
ForceDpiAware.Windows has a side effect of forcing the application DPI to be the same as the primary monitor. This isn't good if you have multiple monitors with different DPI.

On Avalonia, I don't think there are any downsides to disabling this. When it's disabled, `ForceDpiAware.GetWindowScaleFactor` always returns 1.

### Before
image

### After
image-1

### DIFF
--- a/src/Ryujinx.Ava/Program.cs
+++ b/src/Ryujinx.Ava/Program.cs
@@ -100,8 +100,6 @@ namespace Ryujinx.Ava
 
             ReloadConfig();
 
-            ForceDpiAware.Windows();
-
             WindowScaleFactor = ForceDpiAware.GetWindowScaleFactor();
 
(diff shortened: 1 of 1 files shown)


TSRBerry (approved):

gdkchan (approved):
lgtm, does that mean we can eventually replace `WindowScaleFactor = ForceDpiAware.GetWindowScaleFactor();` with `WindowScaleFactor = 1f;` and delete `ForceDpiAware`?

riperiperi:
I think ForceDpiAware is still used by GTK. `ForceDpiAware.GetWindowScaleFactor` is still there because I'm not entirely sure what the point of it is.

PR #5560, opened 2023-08-12
"static readonly" constants should be "const" instead

marco-carvalho:
The value of a `static readonly` field is computed at runtime while the value of a `const` field is calculated at compile time, which improves performance.


gdkchan:
The fields that have been changed to const should have their name changed to `PascalCase` to match our naming conventions.

gdkchan (approved):
lgtm, thanks.

riperiperi (approved):

PR #5561, opened 2023-08-12
Sunset GTK + Update Actions

IsaacMarovitz:
**Changes:**
- Changes base version from `1.1.0` -> `1.2.0`
- Renames Avalonia build artefacts from `test-ava-ryujinx` -> `ryujinx`
- Renames GTK build artefacts from `ryujinx` -> `gtk-ryujinx`
- Points GTK to update to Avalonia

Blocked by #5741

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,9 +73,9 @@ jobs:
 
       - name: Set executable bit
         run: |
-          chmod +x ./publish_ava/Ryujinx ./publish_ava/Ryujinx.sh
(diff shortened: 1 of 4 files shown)


siegmund-heiss-ich:
Should the folders + project files  be renamed too?
I mean Ryujinx.Ava should be Ryujinx and so on.

gdkchan:
> Should the folders + project files be renamed too? I mean Ryujinx.Ava should be Ryujinx and so on.

I don't think so, I don't think the UI should be the "main" project.

IsaacMarovitz:
Need to check that this produces the desired base version and not 1.2.1000 or something

PR #5562, opened 2023-08-12
Prefer jagged arrays over multidimensional

marco-carvalho:
In a [multidimensional array](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/multidimensional-arrays), each element in each dimension has the same, fixed size as the other elements in that dimension. In a [jagged array](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/jagged-arrays), which is an array of arrays, each inner array can be of a different size. By only using the space that's needed for a given array, no space is wasted. This rule, CA1814, recommends switching to a jagged array to conserve memory.

Benchmark:
```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    private static readonly int[,] multiDimensionalArray = new int[,]
    {
        { 0b0001, 0b0010, 0b0100, 0b1000, 0b0011, 0b1001, 0b1010, 0b1100 },
        { 0b0111, 0b1011, 0b1101, 0b1110, 0b1111, 0b0000, 0b0000, 0b0000 },
    };

    private static readonly int[][] jaggedArray = new int[][]
    {
        new int[] { 0b0001, 0b0010, 0b0100, 0b1000, 0b0011, 0b1001, 0b1010, 0b1100 },
        new int[] { 0b0111, 0b1011, 0b1101, 0b1110, 0b1111, 0b0000, 0b0000, 0b0000 },
    };

    [Benchmark(Baseline = true)]
    public int MultiDimensionalArray()
    {
        int sum = 0;
        for (int i = 0; i < multiDimensionalArray.GetLength(0); i++)
        {
            for (int j = 0; j < multiDimensionalArray.GetLength(1); j++)
            {
                sum += multiDimensionalArray[i, j];
            }
        }
        return sum;
    }

    [Benchmark]
    public int JaggedArray()
    {
        int sum = 0;
        for (int i = 0; i < jaggedArray.Length; i++)
        {
            for (int j = 0; j < jaggedArray[i].Length; j++)
            {
                sum += jaggedArray[i][j];
            }
        }
        return sum;
    }
}
```

Results:
```sh
|                Method |     Mean |    Error |   StdDev | Ratio | Rank | Allocated | Alloc Ratio |
|---------------------- |---------:|---------:|---------:|------:|-----:|----------:|------------:|
|           JaggedArray | 19.61 ns | 0.160 ns | 0.134 ns |  0.48 |    1 |         - |          NA |
| MultiDimensionalArray | 40.51 ns | 0.265 ns | 0.221 ns |  1.00 |    2 |         - |          NA |
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -233,10 +233,6 @@ dotnet_naming_style.IPascalCase.required_suffix =
 dotnet_naming_style.IPascalCase.word_separator =
 dotnet_naming_style.IPascalCase.capitalization = pascal_case
 
-# Diagnostics
-
-dotnet_diagnostic.CA1814.severity = error
-
 [src/Ryujinx.HLE/HOS/Services/**.cs]
 # Disable "mark members as static" rule for services
 dotnet_diagnostic.CA1822.severity = none
--- a/.editorconfig
+++ b/.editorconfig
@@ -233,6 +233,10 @@ dotnet_naming_style.IPascalCase.required_suffix =
 dotnet_naming_style.IPascalCase.word_separator =
 dotnet_naming_style.IPascalCase.capitalization = pascal_case
 
+# Diagnostics
+
+dotnet_diagnostic.CA1814.severity = error
+
 [src/Ryujinx.HLE/HOS/Services/**.cs]
 # Disable "mark members as static" rule for services
 dotnet_diagnostic.CA1822.severity = none
--- a/src/Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
+++ b/src/Ryujinx.Graphics.Shader/Instructions/InstEmitTexture.cs
@@ -10,10 +10,10 @@ namespace Ryujinx.Graphics.Shader.Instructions
 {
     static partial class InstEmit
     {
-        private static readonly int[,] _maskLut = new int[,]
+        private static readonly int[][] _maskLut = new int[][]
         {
-            { 0b0001, 0b0010, 0b0100, 0b1000, 0b0011, 0b1001, 0b1010, 0b1100 },
-            { 0b0111, 0b1011, 0b1101, 0b1110, 0b1111, 0b0000, 0b0000, 0b0000 },
+            new int[] { 0b0001, 0b0010, 0b0100, 0b1000, 0b0011, 0b1001, 0b1010, 0b1100 },
+            new int[] { 0b0111, 0b1011, 0b1101, 0b1110, 0b1111, 0b0000, 0b0000, 0b0000 },
         };
 
         public const bool Sample1DAs2D = true;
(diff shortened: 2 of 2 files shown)


gdkchan:
> In a [jagged array](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/arrays/jagged-arrays), which is an array of arrays, each inner array can be of a different size. By only using the space that's needed for a given array, no space is wasted.

No space is being wasted here since all the inner arrays (at least the one you replaced) have the same size.

Performance wise, jagged arrays used to be faster than multi-dimensional arrays. I don't know if that is still the case (there was some improvement on that front in .NET 7 I think). Based on your benchmark, I guess it still is?

marco-carvalho:
> No space is being wasted here since all the inner arrays (at least the one you replaced) have the same size.

You're right that in this specific case, there's no difference in space since all the inner arrays have the same size. The change to a jagged array doesn't aim to reduce space but to provide flexibility and other potential benefits.

Regarding performance, the tests indeed showed that jagged arrays are still faster in the current use case, consistent with the general understanding that jagged arrays may offer better performance. But I encourage everyone to not take my word as granted, and ran those benchmarks locally. There are more info here: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1814

AcK77:
> Enabled by default in .NET 7 : No

Then why we should enable it as an error for only one place in our codebase which don’t do a really big difference performance wise?

marco-carvalho:
> Then why we should enable it as an error for only one place in our codebase which don’t do a really big difference performance wise?

I believe I enabled it as an error for the entire codebase. If this rule translate in a better performance, shouldn't the codebase intervene to prevent the developer to go on a slower path? While the performance difference might not be substantial, it aligns with continuous efforts to optimize code wherever possible. Even very small improvements can add up, especially in paths or in functions called frequently.

AcK77:
I mean, you enabled it for the entire codebase, but the non-default rule is only met at one place. IMO the continuous effort to optimize the codebase could be something when it’s something really significant in term of performances or readability, aswell as a default behavior. The project is still WIP and not at their final steps.

marco-carvalho:
I believe the proposed change aligns with best practices, and although the performance gain might not be substantial at this point, it sets a precedent for writing more efficient code. However, I recognize that the immediate impact may not be significant, as you pointed out.

Considering that the project is still a WIP, I see your point about focusing on more significant optimizations at this stage. I'm open to rolling back this change if it doesn't align with the collective vision for the project at this stage. My primary intention was to introduce an optimization pattern that might be beneficial in the long run.

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

PR #5551, opened 2023-08-11
Geometry shader emulation for macOS

gdkchan:
Overview
---

Some platforms, such as macOS, do not have geometry shader support at all. For those cases, we need to emulate the missing stage, either using a compute or separate vertex shader draw with rasterizer discard. This change implements geometry shader emulation using compute shader, which requires converting both the vertex and geometry stages to compute, 

Implementation
---

This implementation, unlike the one in macos1, uses compute shaders rather than vertex shaders. There are some advantages in doing that, such as being able to use subgroup operations (which are not supported on vertex shaders on macOS). It also has disadvantages, like not being able to easily fetch vertex buffer data.

**Vertex stage:**

Because the vertex shader runs before the geometry, we need to somehow run it before the geometry one runs, so it must also be done separately from the main draw, and this implementation will also do it on compute.

Compute shaders can't access vertex buffers, so instead they are bound as buffer textures. This has the advantage that *most* vertex formats are also supported as buffer texture formats. There are a few formats that are not supported, and in those cases, they are emulated using a format that is similar but has different component type (for example, some signed formats are emulated wth unsigned ones, and the sign extension is done on the shader). The implementation with buffer textures also happens to share the same limitations that the Vulkan spec has for vertex attributes: All offsets must be at least aligned to the format component size, or in case of packed formats, to the full packed size of the format. That is because for non-packed formats, the current implementation uses a single component. For example, RGB16 will be bound as R16, and the shader will add an offset for the actual component it needs to access (R, G or B). Since the format is R16, the vertex buffer offset must be aligned to 2 bytes.

The index buffer is also bound as a buffer texture. Like with vertex buffers, this has the advantage that we can easily control the format, based on the index type. For Ubyte index buffers, we set the format to R8Uint, for Ushort it is set to R16Uint, and for Uint it is set to R32Uint, there is no need for shader specialization.

The process for accessing a vertex input on the shader goes more or less like this:
- Read the vertex ID for the current instance from the index buffer, also calculate the vertex ID for the current instance.
- Read from the vertex info whether the respective vertex buffer is instanced.
- Read value from the buffer texture:
 - If the vertex buffer is instanced, use current instance vertex ID, otherwise use the one from the index buffer.
 - Add first vertex or first instance offsets (depending on the vertex buffer being instanced or not).
 - Multiply stride, if the format is not packed, add the component offset.
 - If the format is not packed, check if the accessed component actually exists on the format. If it doesn't, the value is forced to 0 (for R/G/B) or 1 (for A).
[...]

PR #5564, opened 2023-08-13
Implement scaled vertex format emulation

gdkchan:
This will be required for #5551 since it accesses vertex buffers using buffer textures, and scaled formats are not supported for buffer textures. This change is also useful on its own however, for GPUs that do not support scaled vertex formats, but I think all the GPUs we support right now supports them. In any case, I have added the required checks, so it will emulate them (using integer formats and converting to float in the shader) if not natively supported.


riperiperi (approved):
LGTM, only checking when the device also supports the int format is a good idea.

AcK77 (approved):

PR #5567, opened 2023-08-13
Ava UI: Fix double click anywhere

IsaacMarovitz:
This fixes an issue that allowed you to double-click anywhere within the list box and open the selected game. Now it will only open if the game's list box item was what was clicked. I'm not a big fan of this solution, but putting the double-click handler on the list box item itself resulted in some inconsistency in where you could click and have it register.

### DIFF
--- a/src/Ryujinx.Ava/UI/Controls/ApplicationGridView.axaml.cs
+++ b/src/Ryujinx.Ava/UI/Controls/ApplicationGridView.axaml.cs
@@ -1,10 +1,12 @@
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Interactivity;
+using Avalonia.LogicalTree;
 using Ryujinx.Ava.UI.Helpers;
 using Ryujinx.Ava.UI.ViewModels;
 using Ryujinx.Ui.App.Common;
 using System;
+using System.Linq;
 
 namespace Ryujinx.Ava.UI.Controls
 {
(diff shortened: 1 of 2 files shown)


src/Ryujinx.Ava/UI/Controls/ApplicationGridView.axaml.cs
```diff
@@ -42,10 +45,5 @@ public void GameList_SelectionChanged(object sender, SelectionChangedEventArgs a
                 (DataContext as MainWindowViewModel).GridSelectedApplication = listBox.SelectedItem as ApplicationData;
             }
         }
-
-        private void SearchBox_OnKeyUp(object sender, KeyEventArgs args)
-        {
-            (DataContext as MainWindowViewModel).SearchText = (sender as TextBox).Text;
-        }
```
marysaka, line 49: Why was that removed? Is it unused?

src/Ryujinx.Ava/UI/Controls/ApplicationListView.axaml.cs
```diff
@@ -42,10 +46,5 @@ public void GameList_SelectionChanged(object sender, SelectionChangedEventArgs a
                 (DataContext as MainWindowViewModel).ListSelectedApplication = listBox.SelectedItem as ApplicationData;
             }
         }
-
-        private void SearchBox_OnKeyUp(object sender, KeyEventArgs args)
-        {
-            (DataContext as MainWindowViewModel).SearchText = (sender as TextBox).Text;
-        }
```
marysaka, line 49: Why was that removed? Is it unused?

marysaka (approved):
Not really a fan of your solution but oh well...

AcK77:
Same comments as @marysaka here. Once you explain, I'll approve ;)

PR #5575, opened 2023-08-15
Make Vulkan memory allocator actually thread safe

gdkchan:
It seems there was some attempt to make `MemoryAllocator.Allocate` method thread safe, by locking the `MemoryAllocatorBlockList` object when it already exists, however this is not really enough to make it thread safe for several reasons:
- No suck locks exists when a new `MemoryAllocatorBlockList`. It is possible that another thread will access it right after it is added to the list, and then we have 2 threads calling the `Allocate` method which is what the first lock is supposed to prevent.
- Adding the new `MemoryAllocatorBlockList` to the `_blockLists` is also not thread safe.
- Block freeing is not thread safe, although I don't think this ever happens outside of the render thread, but even if it doesn't it will touch the internal `_blocks` list that can be accessed from other threads.

This change tries to address those issues by using a `ReaderWriterLock` to ensure any modification operation on the lists will be done on only one thread at once.

Following the advice from #5493 I used `ReadWriterLockSlim` instead of `ReadWriterLock`.

The motivation for this was the issue reported on https://github.com/Ryujinx/Ryujinx/pull/4899#issuecomment-1676082980, but if it's indeed a thread safety issue, it would be very hard to reproduce and confirm that the fix works.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
+++ b/src/Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
@@ -67,6 +67,7 @@ namespace Ryujinx.Graphics.Vulkan
             {
                 var newBl = new MemoryAllocatorBlockList(_api, _device, memoryTypeIndex, _blockAlignment, isBuffer);
                 _blockLists.Add(newBl);
+
                 return newBl.Allocate(size, alignment, map);
             }
             finally
--- a/src/Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
+++ b/src/Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
@@ -1,6 +1,7 @@
 ﻿using Silk.NET.Vulkan;
 using System;
 using System.Collections.Generic;
+using System.Threading;
 
 namespace Ryujinx.Graphics.Vulkan
 {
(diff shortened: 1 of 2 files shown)


marysaka (approved):

src/Ryujinx.Graphics.Vulkan/MemoryAllocator.cs
```diff
@@ -40,21 +43,36 @@ public MemoryAllocator(Vk api, VulkanPhysicalDevice physicalDevice, Device devic
...
+            _lock.EnterWriteLock();
 
-            var newBl = new MemoryAllocatorBlockList(_api, _device, memoryTypeIndex, _blockAlignment, isBuffer);
-            _blockLists.Add(newBl);
-            return newBl.Allocate(size, alignment, map);
+            try
+            {
+                var newBl = new MemoryAllocatorBlockList(_api, _device, memoryTypeIndex, _blockAlignment, isBuffer);
+                _blockLists.Add(newBl);
+                return newBl.Allocate(size, alignment, map);
```
AcK77, line 70:
```suggestion

                return newBl.Allocate(size, alignment, map);
```

AcK77 (approved):

PR #5578, opened 2023-08-16
GPU: Add Z16RUnormGUintBUintAUint format

riperiperi:
This format seems to be an alias for Z16Unorm used by OpenGL games, typically for shadowmaps.

Fixes issues with https://github.com/Ryujinx/Ryujinx-Games-List/issues/4384, https://github.com/Ryujinx/Ryujinx-Games-List/issues/2927.

May also fix:
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/4173
- https://github.com/Ryujinx/Ryujinx-Games-List/issues/2340

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/FormatTable.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/FormatTable.cs
@@ -179,6 +179,7 @@ namespace Ryujinx.Graphics.Gpu.Image
             R32G32B32A32Uint                 = R32G32B32A32      | RUint  | GUint  | BUint  | AUint,         // 0x49201
             R32G32B32A32Sint                 = R32G32B32A32      | RSint  | GSint  | BSint  | ASint,         // 0x36d81
             Z16Unorm                         = Z16               | RUnorm | GUnorm | BUnorm | AUnorm,        // 0x2493a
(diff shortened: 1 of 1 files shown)


gdkchan:
It might be better to ignore the B and A types for depth + stencil formats, and ignore R for depth-only formats. That would allow removing a few existing duplicate cases too.

gdkchan (approved):
lgtm, thanks for fixing this. As previous comment says, an approach that ignores component types that doesn't matter might  be more accurate and future proof as we won't need to implement similar combinations for other depth formats in the future, although I guess we already have all of them covered with this change.

GamerzHell9137:
Go Rally:

image
image

Asterix & Obelix XXL: Romastered:

image
image

F1ReB4LL:
Pyramid Quest:
ryujinx_pq1_11989_1
ryujinx_pq1_110-62a6c57_1

Monster Blast:
ryujinx_mb1_11989_1
ryujinx_mb1_110-62a6c57_1

Spencer:
ryujinx_sp1_11989_1
ryujinx_sp1_110-62a6c57_1

TSRBerry (approved):
Looking at the results of the commenters here, this should be fine even if I don't understand the formats myself! :D

AcK77 (approved):

PR #5580, opened 2023-08-16
mm: Migrate service in Horizon project

AcK77:
This PR migrate the `mm:u` service to the Horizon project, things were checked by some RE aswell, that's why some vars are renamed, the logic should be the same as before.

Tests are welcome.


src/Ryujinx.Horizon/MmNv/Ipc/Request.cs
```diff
@@ -0,0 +1,161 @@
...
+using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.Sdk.MmNv;
+using Ryujinx.Horizon.Sdk.Sf;
+using System.Collections.Generic;
+
+namespace Ryujinx.Horizon.MmNv.Ipc
+{
+    partial class Request : IRequest
+    {
+        private readonly object _sessionListLock = new();
```
gdkchan, line 11: Why not locking the list directly instead of having the separate lock object? Like
```cs
lock (_sessionListLock)
{
    // ...
}
```

src/Ryujinx.Horizon/MmNv/Ipc/Request.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            return null;
+        }
+
+        private uint Register(Module module, uint fgmPriority, bool isAutoClearEvent)
+        {
+            lock (_sessionListLock)
+            {
+                // Nintendo ignore the fgm priority as the other services were deprecated.
```
gdkchan, line 152:
```suggestion
                // Nintendo ignores the fgm priority as the other services were deprecated.
```

gdkchan (commented):
Just so that I know, why is the service called "MmNv" rather than just "Mm"?

AcK77:
> Just so that I know, why is the service called "MmNv" rather than just "Mm"?

I just followed what the symbols said:

```
nn::mmnv::IRequest
nn.capsrv.MmnvServer
...
```

But I can revert it to `mm` if you want.

gdkchan:
> > Just so that I know, why is the service called "MmNv" rather than just "Mm"?
> 
> I just followed what the symbols said:
> 
> ```
> nn::mmnv::IRequest
> nn.capsrv.MmnvServer
> ...
> ```
> 
> But I can revert it to `mm` if you want.

It's fine to call it MmNv if that's the official name.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

PR #5603, opened 2023-08-20
Fix some Vulkan validation errors (mostly related to barriers)

gdkchan:
Overview
---

This change aims to reduce the validation error spam that happens when using the Vulkan backend with validation and error logging enabled right now, to make debugging easier, and potentially fix issues on some platforms.

Each commit fixes a different error, so I will explain what each of them fixes individually, in commit order:

- 1st change fixes the following barrier error:
> vkCmdPipelineBarrier(): pImageMemoryBarriers[0].image Barrier for VkImage 0xee74570000000a8a[] does not match an image from the current VkFramebuffer 0x7803750000000ae7[]. The Vulkan spec states: If fname:vkCmdPipelineBarrier is called within a render pass instance using a VkRenderPass object, the image member of any image memory barrier included in this command must be an attachment used in the current subpass both as an input attachment, and as either a color or depth/stencil attachment

When image barrier are used with an active render pass, the image must be one of the render pass input and output attachments. We can't really ensure that since the image might have been rendered on a previous render pass. This change fixes the issue by replacing the image barrier usage with a more generic memory barrier, that does not trigger the error, but *could possibly have performance implications*. So that will need testing on different games/GPUs.

- 2nd change fixes a image creation error:
> vkCreateImage(): Format VK_FORMAT_BC3_UNORM_BLOCK is not supported for this combination of parameters and VkGetPhysicalDeviceImageFormatProperties returned back VK_ERROR_FORMAT_NOT_SUPPORTED. The Vulkan spec states: Each of the following values (as described in Image Creation Limits) must not be undefined : imageCreateMaxMipLevels, imageCreateMaxArrayLayers, imageCreateMaxExtent, and imageCreateSampleCounts

Actually I think I added this one to fix another validation error. But setting the storage image bit for compressed textures is not valid, so I made it only set that flag for images that are considered "image compatible".

- 3rd change fixes another barrier related error:
> vkCmdPipelineBarrier(): pMemoryBarriers[0].dstAccessMask (0x100) is not a subset of VkSubpassDependency dstAccessMask of subpass 0 of VkRenderPass 0xac95f0000000249a[]. Candidate VkSubpassDependency are pDependencies entries [0]. The Vulkan spec states: If fname:vkCmdPipelineBarrier is called within a render pass instance using a VkRenderPass object, the render pass must have been created with at least one subpass dependency that expresses a dependency from the current subpass to itself, does not include VK_DEPENDENCY_BY_REGION_BIT if this command does not, and has synchronization scopes and access scopes that are all supersets of the scopes defined in this command

- 4th one stops calls to `vkCmdSetScissors` with a scissor count of zeor:
> vkCmdSetScissor: parameter scissorCount must be greater than 0. The Vulkan spec states: scissorCount must be greater than 0

Should be pretty self explanatory. The same thing was already being done for viewports.

- 5th one fixes a semaphore validation error:
> vkAcquireNextImageKHR: Semaphore must not be currently signaled. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must be unsignaled

I think that was wrong in general since each swapchain image should have a separate semaphore, otherwise it could present the image before it is ready.

Read to write barriers (for clears) might specify color attachment in the destination access mask, but that flag was missing on the subpass dependency. So I updated the subpass dependency to use the same flag as both source and destination access, since the barriers work both ways now (read to write and write to read).

- 6th one fixes yet another barrier related error:
> vkCmdPipelineBarrier(): .srcStageMask (0x880) is not a subset of VkSubpassDependency srcAccessMask for any self-dependency of subpass 0 of VkRenderPass 0xd35cce0000006742[] for which dstAccessMask is also a subset. Candidate VkSubpassDependency are pDependencies entries [0].

It was possible it would try to insert a read to write barrier where one of the source stages is a compute shader. This is not valid because the subpass dependency does not include compute, and it is not possible to include compute as the subpass dependency can only have graphics stages. It was fixed by removing the compute bit, but that is not really ideal, so suggestions are welcome too.

- Last one does not actually fix a validation error, but that barrier was generally wrong for a long time.

`Pipeline.Barrier ` is supposed to be a generic barrier, but the only stage specified was the fragment shader. I changed it to include all shader stages except compute (which is not part of the subpass dependency). Maybe it should include compute too when not inside a render pass?

---

With this change, some simple 2D games can now run with no validation errors on Vulkan. After this, I plan to fix other validation errors that show on more complex games.

What to test
---

This makes changes to barriers, so it needs to be tested for performance or visual regressions on different games and GPU vendors. Xenoblade might be a good game to check for performance regression since from what I recall, it makes heavy use of barriers.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -18,6 +18,13 @@ namespace Ryujinx.Graphics.Vulkan
 {
     class PipelineBase : IDisposable
     {
+        private const PipelineStageFlags BarrierStageFlags =
+            PipelineStageFlags.VertexShaderBit |
+            PipelineStageFlags.TessellationControlShaderBit |
+            PipelineStageFlags.TessellationEvaluationShaderBit |
+            PipelineStageFlags.GeometryShaderBit |
+            PipelineStageFlags.FragmentShaderBit;
+
         public const int DescriptorSetLayouts = 4;
 
         public const int UniformSetIndex = 0;
@@ -151,8 +158,8 @@ namespace Ryujinx.Graphics.Vulkan
 
             Gd.Api.CmdPipelineBarrier(
                 CommandBuffer,
-                PipelineStageFlags.FragmentShaderBit,
-                PipelineStageFlags.FragmentShaderBit,
+                BarrierStageFlags,
+                BarrierStageFlags,
                 0,
                 1,
                 memoryBarrier,
--- a/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -18,13 +18,6 @@ namespace Ryujinx.Graphics.Vulkan
 {
     class PipelineBase : IDisposable
     {
-        private const PipelineStageFlags BarrierStageFlags =
-            PipelineStageFlags.VertexShaderBit |
-            PipelineStageFlags.TessellationControlShaderBit |
-            PipelineStageFlags.TessellationEvaluationShaderBit |
-            PipelineStageFlags.GeometryShaderBit |
-            PipelineStageFlags.FragmentShaderBit;
-
         public const int DescriptorSetLayouts = 4;
 
         public const int UniformSetIndex = 0;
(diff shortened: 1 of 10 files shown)


Littl3Guy:
12 games tested with an RDNA2 GPU, no performance or visual regressions found.

OldManKain:
Did not notice any regressions on XBDE, XB2, or XB3. All comparisons to mainline were within margin of error.

src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
```diff
@@ -18,6 +18,13 @@ namespace Ryujinx.Graphics.Vulkan
 {
     class PipelineBase : IDisposable
     {
+        private const PipelineStageFlags BarrierStageFlags =
+            PipelineStageFlags.VertexShaderBit |
+            PipelineStageFlags.TessellationControlShaderBit |
+            PipelineStageFlags.TessellationEvaluationShaderBit |
+            PipelineStageFlags.GeometryShaderBit |
```
marysaka, line 25: Geometry being optional this could cause validation errors on some platforms (like macOS)

This should be composed based on the device capabilities imo.

marysaka (changes requested):

src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
gdkchan: Yea, I forgot about that. I changed it to only set the geometry and tessellation stages if the host supports them.

marysaka (approved):

jcm93:
Copying information over from Discord for tracking, re: #5685 as well as #5694, I did some digging, and those two issues seem to be down to commit d6378d2. Changing ``insertReadToWriteBarrier`` to use the old longer call resolves those graphical issues.

IsaacMarovitz:
The issue here is specifically `InsertWriteToReadBarrier`, converting that back to the original function resolves the glitch.

PR #5612, opened 2023-08-23
Vulkan: Fix MoltenVK flickering

riperiperi:
#5576 changed where the position was declared, but forgot to add the Invariant declaration to position when the ReducedPrecision flag was enabled. This was causing weird graphical bugs in a bunch of games, mostly to do with mismatching depth between multiple draws of the same geometry.

Maybe the attempt to add it to Position in DeclareInputOrOutput can be removed now, assuming that path is never used.


gdkchan (approved):
Thanks for fixing this, I forgot it was a thing.

jcm93:
This change seems to result in a regression if using newer versions of MoltenVK. I haven't thoroughly tested many games yet, but in Pikmin 4, some types of textures appear to be highly borked. This is with a version of MoltenVK based off recent master:

Screenshot 2023-08-24 at 11 36 06 PM

PR #5623, opened 2023-08-29
Vulkan: Device Local and higher invocation count for buffer conversions

riperiperi:
Just some simple changes to the buffer conversion shaders. (stride conversion, D32S8 to D24S8)

The first change is using a device local buffer for converted vertex buffers, since they're only read/written on the GPU. These paths don't trigger on NVIDIA, but if you force them to use it demonstrates the full extent writing to host owned memory from compute absolutely destroys them. AMD GPUs are less heavily affected by this issue, but since the game in question was writing 230MB from compute, I imagine it should have some effect.

The second change is allowing the buffer conversion shaders to scale their work group count. While dividing the work between 32 invocations works OK for M1 macs, it's not so great for anything with more cores like AMD GPUs, which should be able to do a lot more parallel copies. Now, it scales by roughly one work group (32 invocations) per 100 elements.

Some stride change cases could be improved further by either limiting vertex buffer size somehow (reading the index buffer could help, but is always risky) or only updating regions that changed, rather than invalidating the whole thing.


riperiperi:
SMS with master, AMD vertex buffer stride conversion path enabled on NVIDIA:

https://github.com/Ryujinx/Ryujinx/assets/6294155/8083b12e-8602-487f-95fb-a505fb1441b3

Same but with changes:

https://github.com/Ryujinx/Ryujinx/assets/6294155/9596f440-1d87-404d-9369-7dece7bb0a72

Littl3Guy:
> SMS with master, AMD vertex buffer stride conversion path enabled on NVIDIA:
> Buffer.Stride.Convert.Master-1.mp4
> 
> Same but with changes:
> Changes-1.mp4

Can confirm this improves SMS performance on AMD as well, will test more games later.

gdkchan (approved):
lgtm, thanks.

marysaka (approved):
lgtm

PR #5626, opened 2023-08-30
Delete ResourceAccess

gdkchan:
I added it to allow us to know if a resource is written from the shader on the backend, however #4899 added a bool on `BufferRange` that serves the same purpose, making it redundant, so this PR removes the `ResourceAccess` enum and all of its uses.

I also set the `write` bool to true now on the `BufferRange`s passed from vertex/geometry as compute where the buffers are modified from the shader. I don't think this makes any difference since those buffers are never modified from CPU, but it's technically the right thing to do.


marysaka (approved):
lgtm

PR #5627, opened 2023-08-30
Fix numeric SWKB validation

IsaacMarovitz:
Pip pointed out that this was too restrictive. Numeric is used for situations like manual IP address and DNS setting, in which case numbers `0-9` AND `.` must be valid characters.

- Fixes SMOO IP address selection


TSRBerry (approved):
Lgtm, thank you!

marysaka (approved):
a bit overkill to use a regex but otherwise lgtm

PR #5631, opened 2023-08-31
pctl: Migrate service to Horizon

IsaacMarovitz:


src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+            {
+                if ((_permissionFlag & 0x40) == 0)
+                {
+                    // TODO: Get actual control flag from Application
+                    _parentalControlFlag = PctlFlagValue.FreeCommunication;
+                }
+
+                if ((_permissionFlag & 0x8040) == 0)
+                {
+                    // TODO: Service store TitleId and FreeCommunicationEnabled in another static object.
```
gdkchan, line 54: Comment needs some proofreading.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+
+            return result;
+        }
+
+        [CmifCommand(1001)]
+        public Result CheckFreeCommunicationPermission()
+        {
+            if (_parentalControlFlag == PctlFlagValue.FreeCommunication && _restrictionEnabled)
+            {
+                // TODO: It seems to checks if an entry exists in the FreeCommunicationApplicationList using the TitleId.
```
gdkchan, line 70: Comment needs some proofreading.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+        {
+            return Result.Success;
+        }
+
+        [CmifCommand(1018)] // 10.0.0+
+        public Result IsFreeCommunicationAvailable()
+        {
+            if (_parentalControlFlag == PctlFlagValue.FreeCommunication && _restrictionEnabled)
+            {
+                // TODO: It seems to checks if an entry exists in the FreeCommunicationApplicationList using the TitleId.
```
gdkchan, line 255: Comment needs some proofreading.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+        [CmifCommand(1062)] // 4.0.0+
+        public Result GetStereoVisionRestriction(out bool stereoVisionRestriction)
+        {
+            if ((_permissionFlag & 0x200) == 0)
+            {
+                stereoVisionRestriction = default;
+                return PctlResult.PermissionDenied;
+            }
+
+            stereoVisionRestriction = false;
```
gdkchan, line 485: Why not just
```cs
stereoVisionRestriction = _stereoVisionRestrictionConfigurable && _stereoVisionRestriction;
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+                return PctlResult.PermissionDenied;
+            }
+
+            if (!_featuresRestriction)
+            {
+                if (_stereoVisionRestrictionConfigurable)
+                {
+                    _stereoVisionRestriction = stereoVisionRestriction;
+
+                    // TODO: It signals an internal event of service. We have to determine where this event is used.
```
gdkchan, line 509:
```suggestion
                    // TODO: It signals an internal event of the service. We have to determine where this event is used.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+        [CmifCommand(1065)] // 5.0.0+
+        public Result IsStereoVisionPermitted(out bool isStereoVisionPermitted)
+        {
+            isStereoVisionPermitted = false;
+
+            Result result = IsStereoVisionPermittedImpl();
+
+            if (result == Result.Success)
+            {
+                isStereoVisionPermitted = true;
```
gdkchan, line 536: Why not just
```cs
isStereoVisionPermitted = result == Result.Success;
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1274 @@
...
+
+            return result;
+        }
+
+        private Result IsStereoVisionPermittedImpl()
+        {
+            /*
+                // TODO: Application Exemptions are read from file "appExemptions.dat" in the service savedata.
+                // Since we don't support the pctl savedata for now, this can be implemented later.
+                if (appExemption)
```
gdkchan, line 547: I don't think the commented code is necessary (or makes sense, the `appExemption` variable doesn't exist so one would need to guess what it is supposed to be).

gdkchan (commented):
It would be nice  to change the parameter names to something more meaningful in the class and interfaces, for the functions that are implemented at least.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+            }
+
+            Result result = PctlResult.InvalidPid;
+
+            if (_pid != 0)
+            {
+                if ((_permissionFlag & 0x40) == 0)
+                {
+                    // TODO: Get actual control flag from Application
+                    _parentalControlFlag = PctlFlagValue.FreeCommunication;
```
gdkchan, line 50: It seems that the previous implementation at least gets the correct value from the nacp. Would be nice to do the same here.

IsaacMarovitz: I'm not sure how to get the Application context in the new services implementation

AcK77: It have to be passed from HLE to `HorizonStatic.Options`. But yeah the original NACP value is needed here.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+                }
+
+                if ((_permissionFlag & 0x8040) == 0)
+                {
+                    // TODO: Services store TitleId and FreeCommunicationEnabled in a static object.
```
AcK77, line 55:
```suggestion
                    // TODO: Service store ApplicationId and FreeCommunicationEnabled in a static object.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+                    // When initialisation is complete, signal an event on that static object.
```
AcK77, line 56:
```suggestion
                    //       When initialisation is complete, signal an event on that static object.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+
+            return result;
+        }
+
+        [CmifCommand(1001)]
+        public Result CheckFreeCommunicationPermission()
+        {
+            if (_parentalControlFlag == PctlFlagValue.FreeCommunication && _restrictionEnabled)
+            {
+                // TODO: Checks if an entry exists in the FreeCommunicationApplicationList using the TitleId.
```
AcK77, line 71:
```suggestion
                // TODO: Checks if an entry exists in the FreeCommunicationApplicationList using the ApplicationId.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+                // Returns FreeCommunicationDisabled if entry doesn't exist.
```
AcK77, line 72:
```suggestion
                //       Returns FreeCommunicationDisabled if entry doesn't exist.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
+using Ryujinx.Common.Logging;
+using Ryujinx.Horizon.Pctl.Types;
+using Ryujinx.Horizon.Sdk.Pctl;
+using Ryujinx.Horizon.Sdk.Sf;
+using Ryujinx.Horizon.Sdk.Sf.Hipc;
+using System;
+using Result = Ryujinx.Horizon.Common.Result;
```
AcK77, line 7: Since LibHac isn't used here, I wonder why is it needed here.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+using TimeSpan = Ryujinx.Horizon.Pctl.Types.TimeSpan;
```
AcK77, line 8: I don't think `TimeSpan` type is part of `Pctl` in original service.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+
+        [CmifCommand(1002)]
+        public Result ConfirmLaunchApplicationPermission(ApplicationId applicationId, [Buffer(HipcBufferFlags.In | HipcBufferFlags.Pointer)] ReadOnlySpan<sbyte> arg1, bool arg2)
+        {
+            if (HorizonStatic.Options.IgnoreMissingServices)
+            {
+                return Result.Success;
+            }
+
+            throw new NotImplementedException();
```
AcK77, line 90: This make no sense. Ignore missing services option is already handled elsewhere so that should be removed.

(Same for similar IPCs)

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+            if (HorizonStatic.Options.IgnoreMissingServices)
+            {
+                arg0 = default;
+                return Result.Success;
+            }
+
+            throw new NotImplementedException();
+        }
+
+        [CmifCommand(1013)] // 4.0.0+
```
AcK77, line 206:
```suggestion
        [CmifCommand(1013)]
```

We do not put firmware version as comment anymore.

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+        {
+            return Result.Success;
+        }
+
+        [CmifCommand(1018)] // 10.0.0+
+        public Result IsFreeCommunicationAvailable()
+        {
+            if (_parentalControlFlag == PctlFlagValue.FreeCommunication && _restrictionEnabled)
+            {
+                // TODO: Checks if an entry exists in the FreeCommunicationApplicationList using the TitleId.
```
AcK77, line 256:
```suggestion
                // TODO: Checks if an entry exists in the FreeCommunicationApplicationList using the ApplicationId.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+                // Returns FreeCommunicationDisabled if entry doesn't exist.
```
AcK77, line 257:
```suggestion
                //       Returns FreeCommunicationDisabled if entry doesn't exist.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+            return Result.Success;
+        }
+
+        [CmifCommand(1031)]
+        public Result IsRestrictionEnabled(out bool restrictionEnabled)
+        {
+            if ((_permissionFlag & 0x140) == 0)
+            {
+                restrictionEnabled = default;
+                return PctlResult.PermissionDenied;
```
AcK77, line 273:
```suggestion

                return PctlResult.PermissionDenied;
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+                return PctlResult.StereoVisionRestrictionConfigurableDisabled;
+            }
+        }
+
+        [CmifCommand(1062)] // 4.0.0+
+        public Result GetStereoVisionRestriction(out bool stereoVisionRestriction)
+        {
+            if ((_permissionFlag & 0x200) == 0)
+            {
+                stereoVisionRestriction = default;
```
AcK77, line 482:
```suggestion
                stereoVisionRestriction = default;

```

src/Ryujinx.Horizon/Pctl/Ipc/PctlService.cs
```diff
@@ -0,0 +1,1259 @@
...
+
+            isStereoVisionPermitted = result == Result.Success;
+
+            return result;
+        }
+
+        private Result IsStereoVisionPermittedImpl()
+        {
+            // TODO: Application Exemptions are read from file "appExemptions.dat" in the service savedata.
+            // Since we don't support the pctl savedata for now, this can be implemented later.
```
AcK77, line 536:
```suggestion
            //       Since we don't support the pctl savedata for now, this can be implemented later.
```

src/Ryujinx.Horizon/Pctl/Ipc/PctlServiceFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private readonly int _permissionFlag;
+
+        public PctlServiceFactory(int permissionFlag)
+        {
+            _permissionFlag = permissionFlag;
+        }
+
+        [CmifCommand(0)]
+        public Result CreateService(out IPctlService pctlService, ulong arg1, [ClientProcessId] ulong pid)
```
AcK77, line 17: `arg1` is the pid placedholder, I don't think this is needed anymore.

src/Ryujinx.Horizon/Pctl/Ipc/PctlServiceFactory.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            pctlService = new PctlService(pid, true, _permissionFlag);
+
+            return Result.Success;
+        }
+
+        [CmifCommand(1)] // 4.0.0+
+        public Result CreateServiceWithoutInitialize(out IPctlService pctlService, ulong arg1, [ClientProcessId] ulong pid)
```
AcK77, line 25: `arg1` is the pid placedholder, I don't think this is needed anymore.

src/Ryujinx.Horizon/Pctl/PctlIpcServer.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            _sm = new SmApi();
+            _sm.Initialize().AbortOnFailure();
+
+            _serverManager = new ServerManager(allocator, _sm, MaxPortsCount, _managerOptions, TotalMaxSessionsCount);
+
+            _serverManager.RegisterObjectForServer(new PctlServiceFactory(0x303), ServiceName.Encode("pctl"), MaxSessionsCount);
+            _serverManager.RegisterObjectForServer(new PctlServiceFactory(0x838E), ServiceName.Encode("pctl:s"), SMaxSessionsCount);
+            _serverManager.RegisterObjectForServer(new PctlServiceFactory(0x83BE), ServiceName.Encode("pctl:a"), AMaxSessionsCount);
+            _serverManager.RegisterObjectForServer(new PctlServiceFactory(0x8040), ServiceName.Encode("pctl:r"), RMaxSessionsCount);
```
AcK77, line 38: Could be nice to get them sorted by service name.

src/Ryujinx.Horizon/Pctl/PctlIpcServer.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void Initialize()
+        {
+            HeapAllocator allocator = new();
+
+            _sm = new SmApi();
+            _sm.Initialize().AbortOnFailure();
+
+            _serverManager = new ServerManager(allocator, _sm, MaxPortsCount, _managerOptions, TotalMaxSessionsCount);
+
+            _serverManager.RegisterObjectForServer(new PctlServiceFactory(0x303), ServiceName.Encode("pctl"), MaxSessionsCount);
```
AcK77, line 35: 0x303 is the permission flag.
Maybe we should store them in an enum

```
enum PermissionFlag
{
    Normal = 0x303,
    Admin = 0x83BE,
    System = Admin,
    Recovery = 0x8040,
}

(Names are guessed)

Then the cast of `_permissionFlag` have to be changed in `ParentalControlServiceFactory` and `ParentalControlService`. Even when the Flag is checked.
```

src/Ryujinx.Horizon/Pctl/Types/AsyncData.cs
```diff
@@ -0,0 +1,7 @@
+using System.Runtime.InteropServices;
+
+namespace Ryujinx.Horizon.Pctl.Types
+{
+    [StructLayout(LayoutKind.Sequential, Size = 0x8, Pack = 0x4)]
+    struct AsyncData { }
+}
```
AcK77, line 7: This is useless and do any sense at all since it's empty...
Same for a lot of other `Types`, it have to reflect what the original service do.

src/Ryujinx.Horizon/Sdk/Pctl/IPctlService.cs
```diff
@@ -0,0 +1,113 @@
+using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.Pctl.Types;
+using Ryujinx.Horizon.Sdk.Sf;
+using System;
+using TimeSpan = Ryujinx.Horizon.Pctl.Types.TimeSpan;
+
+namespace Ryujinx.Horizon.Sdk.Pctl
+{
+    internal interface IPctlService : IServiceObject
```
AcK77, line 9: args have to be renamed to reflect the implementation when needed.

src/Ryujinx.Horizon/Sdk/Pctl/IPctlServiceFactory.cs
```diff
@@ -0,0 +1,11 @@
+using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.Sdk.Sf;
+
+namespace Ryujinx.Horizon.Sdk.Pctl
+{
+    internal interface IPctlServiceFactory : IServiceObject
+    {
+        Result CreateService(out IPctlService pctlService, ulong arg1, ulong pid);
+        Result CreateServiceWithoutInitialize(out IPctlService pctlService, ulong arg1, ulong pid);
```
AcK77, line 9: `arg1` is already a pid placeholder.

AcK77 (changes requested):

src/Ryujinx.Horizon/Pctl/Types/AsyncData.cs
gdkchan: It is necessary since the function receives those structs as argument. The `Size` field above on the `StructLayout` attribute will force the struct to have the correct size, even if it's empty. Fields can be added once someones reverse the service and figure out what the struct looks like. The structs without `StructLayout` attribute and a explicit size are probably wrong though.

AcK77: I mean, all those structs are used in unimplemented call, it's cleaner to remove the unimplemented call and the related types instead of fill everything with placeholders.

gdkchan: > I mean, all those structs are used in unimplemented IPCs, it's cleaner to remove the unimplemented IPCs and the related types instead of fill everything with placeholders.

IMO it's better to keep the empty implementation here. Then when the service is implemented, we just need to full it in, instead of adding the functions all over again, creating all the structs/enums again, etc which is more work.

AcK77: That will be a mess to maintain in case a new firmware which add new IPCs is released.
And the structs/enums aren't in the good namespaces for most of them.
In `HLE` we currently just have needed IPCs and Types, which is more cleaner.

IsaacMarovitz:
To be overridden by @AcK77 in the future.

PR #5640, opened 2023-09-02
Fix layer size for 3D textures with NPOT depth

gdkchan:
This fixes a regression caused by #5220 that caused grass/bushes/flowers/tree animations to stop running on Pokémon Scarlet/Violet. Textures with different layer sizes and more than one "slice" are not considered compatible. Because the layer size for render targets is supplied by the driver rather than being calculated by the emulator, the render texture would have an aligned layer size, while the sampled 3D texture would have an unaligned size. This change fixes it by making sure the layer size is always aligned.

I tested Spiritfarer which was one of the games fixed by the original change and it still seems fine, but I didn't test Zelda Tears of  the Kingdom.

Thanks to @Littl3Guy for narrowing down the build that caused this regression.
Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Texture/SizeCalculator.cs
+++ b/src/Ryujinx.Graphics.Texture/SizeCalculator.cs
@@ -38,6 +38,7 @@ namespace Ryujinx.Graphics.Texture
             bool is3D = depth > 1 || gobBlocksInZ > 1;
 
             int layerSize = 0;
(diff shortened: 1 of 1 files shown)


marysaka (approved):

riperiperi (approved):
Makes sense, nice catch! Still seems fine on TOTK, metroid prime and tony hawks 1+2.

PR #5663, opened 2023-09-07
Add VTimer as alternative interrupt method on Apple Hypervisor

gdkchan:
There are two ways (that I'm aware of, at least) to interrupt a VCPU on Apple Hypervisor. The most obvious one is the one that the emulator currently uses, calling `hv_vcpus_exit`. The second and slightly less efficient one is setting up a timer to interrupt the VCPU periodically.

The method that the emulator currently uses can miss interrupts in some cases. The most obvious case is when the thread is interrupted before it starts executing guest code again. Because it uses a VCPU pool, and it returns the VCPU to the pool when hadling guest interrupts (calling HLE kernel), there is no way to interrupt a VCPU at that time, so it just ignores the interrupt request. On top of that, I'm not sure how `hv_vcpus_exit` behaves if you call it before the guest code is running. For the emulator, it needs to interrupt even if `hv_vcpu_run` was not called yet, but it probably does nothing in this case.

So in order to address the case where the VCPU might not be interrupted, this implements the second approach mentioned before, setting up a virtual timer that will interrupt the VCPU periodically, which then gives it a chance to check if interrupts have been requested and service them. The `hv_vcpus_exit` approach is also still used, to allow the thread to be interrupted quickly if we can (and if we can't, the timer will interrupt it eventually, but right now it can take up to 16 ms, which is the timer interval currently being used).

This fixes softlocks and infinite loading screens on several games running on Hypervisor on macOS.

Here are a few examples.

**Persona 5 Scramble/Strikers:**
Captura de Tela 2023-09-07 às 11 38 57
Before that change, it would freeze before reaching the title screen, even after several attempts.

**Bravely Default II:**
Captura de Tela 2023-09-07 às 11 51 22
Captura de Tela 2023-09-07 às 11 52 07
Captura de Tela 2023-09-07 às 11 52 24
This one would usually freeze either before title screen or the introduction, I was never able to make it that far before this change.

**Life is Strange True Colors:**
Captura de Tela 2023-09-07 às 12 10 51
Captura de Tela 2023-09-07 às 12 12 15
Captura de Tela 2023-09-07 às 12 18 02
This one would get stuck on the loading screen before reaching the title screen on my previous test (and weirdly enough, the animation would continue playing forever).

---

Testing is welcome. If you have a game that freezes frequently with Hypervisor, you can give it a try. Keep in mind that this doesn't fix crashes, the nature of this problem is that it would cause some threads to never run, and other to run forever. So it's not something that would cause crashes.

Would be nice to check if there's no performance regression on the hypervisor since the timer approach is a little less efficient.

### DIFF
--- a/src/Ryujinx.Cpu/AppleHv/TimeApi.cs
+++ b/src/Ryujinx.Cpu/AppleHv/TimeApi.cs
@@ -1,4 +1,5 @@
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 
 namespace Ryujinx.Cpu.AppleHv
 {
@@ -8,6 +9,7 @@ namespace Ryujinx.Cpu.AppleHv
         public uint Denom;
     }
 
+    [SupportedOSPlatform("macos")]
     static partial class TimeApi
     {
         [LibraryImport("libc", SetLastError = true)]
--- a/src/Ryujinx.Cpu/AppleHv/TimeApi.cs
+++ b/src/Ryujinx.Cpu/AppleHv/TimeApi.cs
@@ -2,10 +2,10 @@ using System.Runtime.InteropServices;
 
 namespace Ryujinx.Cpu.AppleHv
 {
-    struct mach_timebase_info_t
+    struct MachTimebaseInfo
     {
-        public uint numer;
-        public uint denom;
+        public uint Numer;
+        public uint Denom;
     }
 
     static partial class TimeApi
@@ -14,6 +14,6 @@ namespace Ryujinx.Cpu.AppleHv
         public static partial ulong mach_absolute_time();
 
         [LibraryImport("libc", SetLastError = true)]
-        public static partial int mach_timebase_info(out mach_timebase_info_t info);
+        public static partial int mach_timebase_info(out MachTimebaseInfo info);
     }
 }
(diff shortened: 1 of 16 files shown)


src/Ryujinx.Cpu/AppleHv/TimeApi.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Ryujinx.Cpu.AppleHv
+{
+    struct MachTimebaseInfo
+    {
+        public uint Numer;
+        public uint Denom;
+    }
+
+    static partial class TimeApi
```
TSRBerry, line 11: Could you add the `SupportedOSPlatform` attribute here?

I haven't checked if it's present for the other `AppleHv` classes, but if it's not it would be nice to add them as well.

gdkchan: I added it. Had to also add it on classes consuming those APIs, otherwise the compiler would complain that the call site was reachable from all platforms.

marysaka (approved):

marysaka (approved):

AcK77 (approved):
Untested but code wise it's lgtm!

PR #5664, opened 2023-09-07
Strings should not be concatenated using '+' in a loop

marco-carvalho:
`StringBuilder` is more efficient than string concatenation, especially when the operator is repeated over and over as in loops.

In languages where strings are immutable, like Java and C#, each time you concatenate strings using the + operator, a new string object is created in memory to hold the result. This can lead to a significant overhead, both in terms of memory and processing time, especially when you're dealing with large strings or performing the operation multiple times. 

On the other hand, `StringBuilder` is designed to be mutable, meaning it can be modified after its creation without generating a new object in memory. When you append or insert a string to a `StringBuilder` object, it modifies the existing character array in memory, avoiding the need to create a new string object. Moreover, `StringBuilder` often pre-allocates more memory than actually needed for the string, leaving room for additional characters to be added without requiring a new memory allocation. This makes it particularly useful for scenarios where you're building a string dynamically and the final size is not known in advance.

Benchmark:
```cs
[SimpleJob]
[MemoryDiagnoser]
[RankColumn]
public class Test
{
    [Params(10, 100, 1000)]
    public int ArraySize;

    private string[] arrayOfStrings;

    [GlobalSetup]
    public void Setup()
    {
        arrayOfStrings = new string[ArraySize];
        for (int i = 0; i < ArraySize; ++i)
        {
            arrayOfStrings[i] = "a";
        }
    }

    [Benchmark(Baseline = true)]
    public string UsingPlusOperator()
    {
        string str = string.Empty;
        for (int i = 0; i < arrayOfStrings.Length; ++i)
        {
            str += arrayOfStrings[i];
        }
        return str;
    }

    [Benchmark]
    public string UsingStringBuilder()
    {
        StringBuilder bld = new StringBuilder();
        for (int i = 0; i < arrayOfStrings.Length; ++i)
        {
            bld.Append(arrayOfStrings[i]);
        }
        return bld.ToString();
    }
}
```

Results:
```sh
|             Method | ArraySize |          Mean |        Error |        StdDev |       Median | Ratio | RatioSD | Rank |     Gen0 | Allocated | Alloc Ratio |
|------------------- |---------- |--------------:|-------------:|--------------:|-------------:|------:|--------:|-----:|---------:|----------:|------------:|
| UsingStringBuilder |        10 |      70.90 ns |     3.347 ns |      9.548 ns |     68.65 ns |  0.43 |    0.08 |    1 |   0.0484 |     152 B |        0.45 |
|  UsingPlusOperator |        10 |     168.70 ns |     7.896 ns |     22.782 ns |    158.79 ns |  1.00 |    0.00 |    2 |   0.1068 |     336 B |        1.00 |
|                    |           |               |              |               |              |       |         |      |          |           |             |
| UsingStringBuilder |       100 |     742.38 ns |    45.495 ns |    129.799 ns |    715.96 ns |  0.29 |    0.06 |    1 |   0.2441 |     768 B |        0.06 |
|  UsingPlusOperator |       100 |   2,669.44 ns |   181.155 ns |    501.980 ns |  2,403.29 ns |  1.00 |    0.00 |    2 |   4.0092 |   12576 B |        1.00 |
|                    |           |               |              |               |              |       |         |      |          |           |             |
| UsingStringBuilder |      1000 |   5,487.48 ns |   278.948 ns |    818.106 ns |  5,262.09 ns |  0.05 |    0.01 |    1 |   1.4572 |    4576 B |       0.004 |
|  UsingPlusOperator |      1000 | 109,683.34 ns | 8,343.086 ns | 24,071.711 ns | 96,267.13 ns |  1.00 |    0.00 |    2 | 327.0264 | 1025976 B |       1.000 |
```

### DIFF
--- a/src/ARMeilleure/Diagnostics/Symbols.cs
+++ b/src/ARMeilleure/Diagnostics/Symbols.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text;
 
 namespace ARMeilleure.Diagnostics
 {
@@ -48,14 +49,18 @@ namespace ARMeilleure.Diagnostics
                         ulong diff = address - symbol.Start;
                         ulong rem = diff % symbol.ElementSize;
 
-                        result = symbol.Name + "_" + diff / symbol.ElementSize;
+                        StringBuilder resultBuilder = new StringBuilder();
+                        resultBuilder.Append(symbol.Name);
+                        resultBuilder.Append("_");
+                        resultBuilder.Append(diff / symbol.ElementSize);
 
                         if (rem != 0)
                         {
-                            result += "+" + rem;
+                            resultBuilder.Append("+");
+                            resultBuilder.Append(rem);
                         }
 
-                        _symbols.TryAdd(address, result);
+                        _symbols.TryAdd(address, resultBuilder.ToString());
 
                         return result;
                     }
--- a/src/ARMeilleure/Diagnostics/Symbols.cs
+++ b/src/ARMeilleure/Diagnostics/Symbols.cs
@@ -50,14 +50,11 @@ namespace ARMeilleure.Diagnostics
                         ulong rem = diff % symbol.ElementSize;
 
                         StringBuilder resultBuilder = new();
-                        resultBuilder.Append(symbol.Name);
-                        resultBuilder.Append("_");
-                        resultBuilder.Append(diff / symbol.ElementSize);
+                        resultBuilder.Append($"{symbol.Name}_{diff / symbol.ElementSize}");
 
                         if (rem != 0)
                         {
-                            resultBuilder.Append("+");
-                            resultBuilder.Append(rem);
+                            resultBuilder.Append($"+{rem}");
                         }
 
                         _symbols.TryAdd(address, resultBuilder.ToString());
--- a/src/ARMeilleure/Diagnostics/Symbols.cs
+++ b/src/ARMeilleure/Diagnostics/Symbols.cs
@@ -49,7 +49,7 @@ namespace ARMeilleure.Diagnostics
                         ulong diff = address - symbol.Start;
                         ulong rem = diff % symbol.ElementSize;
 
-                        StringBuilder resultBuilder = new StringBuilder();
+                        StringBuilder resultBuilder = new();
                         resultBuilder.Append(symbol.Name);
                         resultBuilder.Append("_");
                         resultBuilder.Append(diff / symbol.ElementSize);
--- a/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
@@ -341,8 +341,7 @@ namespace Ryujinx.Ava.UI.ViewModels
                         {
                             foreach (AmiiboApiUsage usageItem in item.AmiiboUsage)
                             {
-                                usageStringBuilder.Append(Environment.NewLine);
-                                usageStringBuilder.Append($"- {usageItem.Usage.Replace("/", Environment.NewLine + "-")}");
+                                usageStringBuilder.Append($"{Environment.NewLine}- {usageItem.Usage.Replace("/", Environment.NewLine + "-")}");
 
                                 writable = usageItem.Write;
                             }
@@ -351,8 +350,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
                     if (usageStringBuilder.Length == 0)
                     {
-                        usageStringBuilder.Append(LocaleManager.Instance[LocaleKeys.Unknown]);
-                        usageStringBuilder.Append(".");
+                        usageStringBuilder.Append($"{LocaleManager.Instance[LocaleKeys.Unknown]}.");
                     }
 
                     Usage = $"{LocaleManager.Instance[LocaleKeys.Usage]} {(writable ? $" ({LocaleManager.Instance[LocaleKeys.Writable]})" : "")} : {usageStringBuilder}";
--- a/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
@@ -327,7 +327,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
             string imageUrl = _amiiboList.Find(amiibo => amiibo.Equals(selected)).Image;
 
-            string usageString = "";
+            StringBuilder usageStringBuilder = new StringBuilder();
 
             for (int i = 0; i < _amiiboList.Count; i++)
             {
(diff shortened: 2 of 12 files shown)


marco-carvalho:
Plotted results:
image

gdkchan:
The new code is more verbose and less readable imo. I would prefer if it used string interpolation. I think that is faster than both and more readable.

src/ARMeilleure/Diagnostics/Symbols.cs
```diff
@@ -48,14 +49,18 @@ public static string Get(ulong address)
                         ulong diff = address - symbol.Start;
                         ulong rem = diff % symbol.ElementSize;
 
-                        result = symbol.Name + "_" + diff / symbol.ElementSize;
+                        StringBuilder resultBuilder = new();
+                        resultBuilder.Append(symbol.Name);
```
gdkchan, line 53: Would prefer to use string interpolation here (I don't mind keeping the `StringBuilder` approach, but I'd change it to `resultBuilder.Append($"{symbol.Name}_{diff / symbol.ElementSize}");`.

src/ARMeilleure/Diagnostics/Symbols.cs
```diff
@@ -48,14 +49,18 @@ public static string Get(ulong address)
...
+                        resultBuilder.Append("_");
+                        resultBuilder.Append(diff / symbol.ElementSize);
 
                         if (rem != 0)
                         {
-                            result += "+" + rem;
+                            resultBuilder.Append("+");
```
gdkchan, line 59: Would be better with string interpolation.

src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
```diff
@@ -341,20 +341,21 @@ private void SetAmiiboDetails()
                         {
                             foreach (AmiiboApiUsage usageItem in item.AmiiboUsage)
                             {
-                                usageString += Environment.NewLine +
-                                               $"- {usageItem.Usage.Replace("/", Environment.NewLine + "-")}";
+                                usageStringBuilder.Append(Environment.NewLine);
```
gdkchan, line 344: Would be better with string interpolation. Maybe `\n` rather than `Environment.NewLine` too.

src/Ryujinx.Ava/UI/ViewModels/AmiiboWindowViewModel.cs
```diff
@@ -341,20 +341,21 @@ private void SetAmiiboDetails()
...
                                 writable = usageItem.Write;
                             }
                         }
                     }
 
-                    if (usageString.Length == 0)
+                    if (usageStringBuilder.Length == 0)
                     {
-                        usageString = LocaleManager.Instance[LocaleKeys.Unknown] + ".";
+                        usageStringBuilder.Append(LocaleManager.Instance[LocaleKeys.Unknown]);
```
gdkchan, line 354: Would be better with string interpolation.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGen.cs
```diff
@@ -67,19 +67,20 @@ private static string GetExpression(CodeGenContext context, AstOperation operati
...
+                    builder.Append(GenerateLoadOrStore(context, operation, isStore: false));
 
                     AggregateType dstType = operation.Inst == Instruction.AtomicMaxS32 || operation.Inst == Instruction.AtomicMinS32
                         ? AggregateType.S32
                         : AggregateType.U32;
 
                     for (int argIndex = operation.SourcesCount - arity + 2; argIndex < operation.SourcesCount; argIndex++)
                     {
-                        args += ", " + GetSoureExpr(context, operation.GetSource(argIndex), dstType);
+                        builder.Append(", ");
```
gdkchan, line 82: Would be better with string interpolation.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
```diff
@@ -697,29 +698,33 @@ public static string GenerateLoadOrStore(CodeGenContext context, AstOperation op
...
                     // that the index expression must be *exactly* "gl_InvocationID",
                     // otherwise the compilation fails.
                     // TODO: Get rid of this and use expression propagation to make sure we generate the correct code from IR.
-                    varName += "[gl_InvocationID]";
+                    builder.Append("[gl_InvocationID]");
                 }
                 else
                 {
-                    varName += $"[{GetSoureExpr(context, src, AggregateType.S32)}]";
+                    builder.Append('[');
```
gdkchan, line 714: Would be better with string interpolation.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGenMemory.cs
```diff
@@ -697,29 +698,33 @@ public static string GenerateLoadOrStore(CodeGenContext context, AstOperation op
...
+                    builder.Append(GetSoureExpr(context, src, AggregateType.S32));
+                    builder.Append(']');
                 }
             }
 
             if (isStore)
             {
                 varType &= AggregateType.ElementTypeMask;
-                varName = $"{varName} = {GetSoureExpr(context, operation.GetSource(srcIndex), varType)}";
+                builder.Append(" = ");
```
gdkchan, line 723: Would be better with string interpolation.

gdkchan (commented):
The ones I didn't comment are fine to me as is.

En3Tho:
Is string usage in any of those files identified as bottleneck? I'm not too familiar with the code but generally if you're trying to optimize something that runs rarely or once and basically not a part of hot code then there is no real meaning in trying to squeeze more perf out of it because it simply won't matter. You would only complicate the code.

I would suggest actually using profiler and back up those changes with real usage. Maybe some of these changes are really needed. 

Also, if any of that string allocation code is actually used within the "main loop" then you would rather use StringBuilderPool or, even better, not allocate any strings at all because main loop should ideally not allocate at all to not get paused by gc.

marco-carvalho:
I initially made these changes based on general best practices for string manipulation in C#, but I see your point about the importance of focusing on actual performance bottlenecks. I haven't yet performed any profiling to identify whether the sections of code I modified are critical performance paths, so I appreciate the suggestion to back up these kinds of changes with real usage data.

Regarding the trade-off between performance and code complexity, my aim was to improve the former without significantly affecting the latter.

En3Tho:
@gdkchan Can you please give some tips on profiling Ryujinx? I just tried PefView and the app just crashed. I'm going to ry DotTrace next.

gdkchan:
Visual Studio has a profiler, you can use that. There's also Intel VTune Profiler, it works with .NET applications. I think Rider IDE has a profiler too, but I never used that one.
You can use `dotnet-trace` command line tool to collect a profile and then open it on PerfView. But that is one of the worst ways IMO, I don't recommend doing it.

src/Ryujinx.Graphics.Shader/CodeGen/Glsl/Instructions/InstGen.cs
```diff
@@ -2,7 +2,7 @@
 using Ryujinx.Graphics.Shader.StructuredIr;
 using Ryujinx.Graphics.Shader.Translation;
 using System;
-
+using System.Text;
```
marysaka, line 5:
```suggestion
using System.Text;

```

marysaka: Keep the line return.

marysaka (approved):
One nit, otherwise should be fine I guess.

Most of the codepath aren't really the most used so I don't expect this to make any difference at all, but it does looks cleaner in some cases.

TSRBerry (approved):

PR #5668, opened 2023-09-10
Fix shader GlobalToStorage pass when base address comes from local or shared memory

gdkchan:
This fixes a regression introduced on #5241 where the shader translator would fail to find the base address for global memory operations if it comes from a local or shared memory. The issue is that I updated the IR to place a "ID" value to identify the memory being accessed, but by doing that I moved the offset value to the operand at index 1, but did not update `GlobalToStorage` code to account for that change.

This fixes a few failures to find storage buffers on Splatoon 3 and probably other games, though I'm not sure which visible effects that may have (I didn't notice anything visibly broken on master at a first glance).


AcK77 (approved):

riperiperi (approved):
Nice catch.

PR #5670, opened 2023-09-10
Stub unsupported BSD socket options

gdkchan:
The Switch uses the FreeBSD sockets implementation, while the emulator uses .NET sockets abstraction, which is basically a wrapper around WinSock (later adapted to work on Linux/MacOS). Right now we pretend that any socket option that .NET sockets don't support are not supported at all, and return an error to the game. This is of course not correct and can cause problems since the game does not expect those operations to fail (since they are supported on the Switch).

This change improves the handling of unsupported socket options a little bit by not returning an error if the socket option is valid for FreeBSD, but not supported on .NET sockets.  It doesn't provide an actual implementation for them though, so if the application actually depends on the socket option being set to function, it will not work.

This allows Crysis Remastered to go in-game (it would go back to the main menu immediately before):
image

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
@@ -311,7 +311,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
                 if (!WinSockHelper.TryConvertSocketOption(option, level, out SocketOptionName optionName))
                 {
                     Logger.Warning?.Print(LogClass.ServiceBsd, $"Unsupported GetSockOpt Option: {option} Level: {level}");
-                    optionValue.Fill(0);
+                    optionValue.Clear();
 
                     return LinuxError.SUCCESS;
                 }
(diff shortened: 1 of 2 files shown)


src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
```diff
@@ -299,11 +299,21 @@ public LinuxError GetSocketOption(BsdSocketOption option, SocketOptionLevel leve
...
+                {
+                    Logger.Warning?.Print(LogClass.ServiceBsd, $"Invalid GetSockOpt Option: {option} Level: {level}");
+
+                    return result;
+                }
+
                 if (!WinSockHelper.TryConvertSocketOption(option, level, out SocketOptionName optionName))
                 {
                     Logger.Warning?.Print(LogClass.ServiceBsd, $"Unsupported GetSockOpt Option: {option} Level: {level}");
+                    optionValue.Clear();
```
TSRBerry, line 314: Why was this added?

TSRBerry (approved):
Lgtm! I just have one question that I commented.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedSocket.cs
gdkchan: I wanted to make sure that the returned value would not contain "garbage" if the option is not supported.

TSRBerry (approved):
(Don't mind me just fixed an issue with Mako)

marysaka (approved):
Not a huge fan of changing EOPNOTSUPP to SUCCESS but oh well...

PR #5671, opened 2023-09-11
Vulkan: Add RAII-based diagnostic labels to CommandBuffers

Wunkolo:
Adds the `CommandBufferLabelScope` class. A RAII-based utility-object to automatically decorate command buffers with diagnostic data to aid in debugging and profiling. Conditionally enabled when `Graphics Backend Log Level` is set to `All`(`GraphicsDebugLevel.All`). Some more types like `Extents2D` might need to implement a `ToString` override at some point but that's outside the scope of this PR.

Before:
qrenderdoc_FFgwQTx7Rl

After:
qrenderdoc_9G5AHpyn6D


jduncanator:
I'm a little cautious on the performance impact of this. Have you made any measurements? These are potentially quite hot paths, and after a quick look over the code, it doesn't look like the call to `CreateLabelScope` is something the JIT would be able to optimise away due to the non-const nature of `_enableDebugLabels`. Perhaps a `[Conditional]` attribute, and making this available only in dirty builds might be useful?

Wunkolo:
> Have you made any measurements?

None so far, this is specifically for a graphics debugging context but is ideally zero-cost outside of that particular context. A similar pattern would be used for object-naming(likely another PR I make after this one).

> it doesn't look like the call to `CreateLabelScope` is something the JIT would be able to optimise away due to the non-const nature of `_enableDebugLabels`

`_enableDebugLabels` has been removed. This was a local branch that was about a year old before `VulkanDebugMessenger` existed. Now the core implementation exists within `VulkanDebugMessenger` which is likely where a compile-time `[Conditional]` flag could be used to try and remove any overhead while this is disabled.

> A using usually sets up additional exception handlers/scopes (as it's essentially implemented as a try/finally block), which has non-negligible overhead.

This is the tricky bit. This utilizes `IDisposable` to automatically call begin/end functions when the object goes in and out of scope with `using`. The costs don't mean much in a developer-debugging context but I'm not sure how to statically hint to the compiler to optimize that stuff away in the case that it knows `CreateLabelScope` will always return `null` like in a user-facing build.

jduncanator:
> This is the tricky bit. This utilizes `IDisposable` to automatically call begin/end functions when the object goes in and out of scope with `using`. The costs don't mean much in a developer-debugging context but I'm not sure how to statically hint to the compiler to optimize that stuff away in the case that it knows `CreateLabelScope` will always return `null` like in a user-facing build.

This is where `[Conditional]` comes in handy, the compiler will completely omit the function call (at compile time, not runtime!), so the `using` would be removed as well.

Wunkolo:
> This is where [Conditional] comes in handy, the compiler will completely omit the function call (at compile time, not runtime!), so the using would be removed as well.

It does not seem possible to use conditional attributes on functions that either return a value or have an `out` parameter it seems. It only works on functions that return `void`.

marysaka:
I do not expect the perf cost to be that big here tbh. The only thing that could be "scary" is all the string formats but that should be fine.

PR #5673, opened 2023-09-12
Reduce the amount of descriptor pool allocations on Vulkan

gdkchan:
This was started by extracting some changes from #3001: I changed the descriptor pool creation there to use lower pool sizes. Before the change, it would create a pool with sizes for all used descriptor types (uniform/storage buffer, combined image sampler, storage buffers, uniform texel buffers and storage texel buffers). Since it creates a descriptor pool per set, and each set can only use some of those types, this change the descriptor pool to only have pool sizes for the descriptor types that the set uses.

The other change is a bug fix. `PipelineLayoutCacheEntry.GetNewDescriptorSetCollection` knows that the command buffer changed by checking `_dsLastCbIndex != commandBufferIndex`. When the command buffer changes, it sets the current cache entry index to 0, which allows it to re-use descriptor pools that have already been allocated. However, it is possible for a pipeline layout cache entry to be used consecutively with the same command buffer, even after it has been submitted. In other words, there are cases where it would create a lot of new descriptor pools even when it's valid to re-use existing ones, just because it doesn't know that the buffer has already been submitted since the last call. To fix this I added a `SubmissionCount` that is used in addition to the command buffer index to know if it's valid to re-use existing descriptor pools.

This fixes cases where descriptor pool creation would return VK_ERROR_UNKNOWN result on AMD, likely because it was creating hundreds of thousands of descriptor pools in some cases. The issue is very easy to reproduce on Puyo Puyo Tetris by just letting it run for a few minutes.

Some logic from `GetNewDescriptorSetCollection` was moved to `UpdateCommandBufferIndex` which is called less often, to avoid performance hit from the extra check.

Testing is welcome to ensure this doesn't break anything.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/CommandBufferPool.cs
+++ b/src/Ryujinx.Graphics.Vulkan/CommandBufferPool.cs
@@ -27,6 +27,7 @@ namespace Ryujinx.Graphics.Vulkan
         {
             public bool InUse;
             public bool InConsumption;
+            public int SubmissionCount;
             public CommandBuffer CommandBuffer;
             public FenceHolder Fence;
             public SemaphoreHolder Semaphore;
@@ -193,6 +194,11 @@ namespace Ryujinx.Graphics.Vulkan
             return _commandBuffers[cbIndex].Fence;
         }
 
+        public int GetSubmissionCount(int cbIndex)
+        {
+            return _commandBuffers[cbIndex].SubmissionCount;
+        }
+
         private int FreeConsumed(bool wait)
         {
             int freeEntry = 0;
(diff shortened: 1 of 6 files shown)


Littl3Guy:
Tested a bunch of games, no regressions found.

AlexandraDel:
Um-m I understand correctly that I need to download 
ryujinx-Release-1.1.0+081f5e1-win_x64
and just run the game. 
I think I'm doing something wrong, but I have no idea what else to do, it just goes dark.

I guess I need to do a local build like I did before by downloading NET.

image

gdkchan:
The PR that is supposed to fix this game is #3001, not this one.

AlexandraDel:
> The PR that is supposed to fix this game is #3001, not this one.
Oh, thank you. Right.
I just recently reported a small bug and thought it had something to do with it

https://github.com/Ryujinx/Ryujinx/pull/3001#issuecomment-1714517359

marysaka (approved):

AcK77 (approved):

PR #5681, opened 2023-09-14
Implement NGC service

### LINKED ISSUE #5678: ngc:u is not implemented
### Service call

ngc:u is not implemented

### Service description

From: https://switchbrew.org/wiki/NGC_services

> NGC (No Good Content) is a sysmodule designed for handling bad words. 

This sysmodule was added with firmware version `16.0.0`.

### Required by

[Baten Kaitos I & II HD Remaster (Asia)](https://github.com/Ryujinx/Ryujinx-Games-List/issues/4739)
[Baten Kaitos I & II HD Remaster (Europe/USA)](https://github.com/Ryujinx/Ryujinx-Games-List/issues/4745)

gdkchan:
This implements NGC (No Good Content) service, a service introduced on firmware version 16.0.0 to filter bad words from text. This functionality already existed, but was not part of the OS (as a service) before.

Overview
---

On 16.0.0, this service exposes 4 functions. Below there is a brief explanation of what each one does.

**GetContentVersion:** Get the version of the dictionaries with the bad words. This is just a 4 bytes `version.dat` file stored in the system archive, the version is a 32-bit integer stored in big endian.

**Check:** Check if a word contains any bad words. There is a `option` parameter that controls the check. If `SkipAtSignCheck` is set to `DoNotSkip`, it will set the bit on the mask if the word contains the `@` character. The `SystemRegionMask` is supposed to have a mask based on the current system configured region, and is not supposed to be set by the application directly. The other fields on the `option` has no effect on the check. The `regionMask` parameter is a mask, where each bit set indicates a different language region that should be checked. There are 16 dictionaries, one per region (one for Japan, one for North America, one for Italy, one for Brazil, etc). In addition to those, there is a "common" dictionary. The common dictionary is always checked against regardless of the `regionMask` value.

**Mask:** Replaces bad words with asterisks on the input text. Unlike `Check`, this function can process text composed of multiple words, not a single word, but there is a limit of 512 characters. Anything past the first 512 characters will not be processed, if the string is that large. The string should be UTF-8 encoded and null terminated. Like `Check`, `SkipAtSignCheck` controls if it will attempt to detect e-mail addresses on the input string, and mask them. The check is pretty crude and it can only detect e-mails with the characters `A-Z`, `a-z`, `0-9` and `-._`. `MaskMode` can be `Overwrite` (in which case it replaces each character in the bad word with an asterisk), or `ReplaceByOneCharacter` (in which case it replaces the whole word with an asterisk). `Flags` seems to be something new on firmware 16.0.0, and can be used to normalize the string (according to form KC rules on Unicode standard), which transforms characters with multiple possible encodings into a single, "canonical" encoding. There is another flag that can transform the input string from a "similar form", into its "canonical" form. This is used to replace similar characters (or character combinations) into their equivalent. For example, `vv` is *similar* to `w`, so it transforms `vv` into `w`. `|-|` is similar to `H`, `3` is similar to `E`, `4` is similar to `A`, etc.

**Reload:** This simply causes the system archive to be unmounted and then mounted again, presumably to refresh its contents. Unsure what the purpose of that is.

The matching related functions above makes use of the Aho-Corasick trie based string matching algorithm. There are 3 dictionaries *per language*, and also 3 common ones. The first (suffixed with `not_b` in the file name) is used for matching words that can match even if they are in the middle of other words, it does not require any separator. The second (suffixed with `b1`) is used to match words that must be either prefixed or suffixed (or both) by a separator, and is always used with the `b2` dictionary. When there is a match with `b1` dictionary, it surrounds the matched word with `\b` or `a` depending on whether the word is prefixed/suffixed by a separator or not (if there is a separator, it inserts `\b`, otherwise it inserts `a`). The word on `b2` dictionary will be either prefixed, suffixed or both with `\b` if they require a separator at that position to match.

There is an additional dictionary called `ac_similar_form` that is used to match "similar form" strings. It then replaces those with the "canonical" string stored in the `table_similar_form` file.

While the method used for matching the "canonicalized" string (when the `Flags` are not 0 on the `option` struct) is the exact same Aho-Corasick match function used for the regular match, the actual masking logic is different, and actually much more inefficient. While the regular implementation only requires a single pass over the text (per region/dictionary), the canonicalized string version does many of them. First it matches from 0 up to the string length until it can find a match ending at **N**, then it matches starting from 0 up to **N** to find where the match starts. Only then it can match the string. This makes masking string a LOT slower when the `Flags` are not zero, for no good reason. I did not try to optimize this, I Just implemented it the same way as the service does.

Implementation
---

The implementation should be complete, with the exception of the notification thread that was not implemented (it requires `OpenSystemDataUpdateEventNotifier` to be implemented on FS). AFAIK the only purpose of this thread is reloading the NgWord2 system archive if it changes (due to a partial system update perhaps?).

I tried to keep the implementation close to the original, but there are a few differences:
- No attempt was made to match how the service does memory allocation, it just uses regular, garbage collected objects in C#. As a result it never returns "allocation failed" errors.
- Where possible, .NET UTF-8 string manipulation functions where used, instead of trying to port whichever library the service uses for that. I beleive they should be equivalent but I'm not 100% sure.
- A few changes were made to ensure it will at least not crash if a very old firmware where the NgWord2 system archive doesn't exist is used.

What it fixes
---

Baten Kaitos Remaster requires this, with this change (this game uses it to check the user name, and replace any bad words with asterisks), it can go in-game without enabling ignore missing services.
image

Closes #5678.


Illynir:
**STAR OCEAN THE SECOND STORY R** Demo Seems to need this too. The game boots but when you start a new game the game makes an "exception error stopping point".

The demo works flawlessly with Ignore missing service or this PR.

src/Ryujinx.HLE/HOS/HorizonFsClient.cs
```diff
@@ -0,0 +1,117 @@
...
+using Ryujinx.Horizon;
+using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.Sdk.Fs;
+using System;
+using System.Collections.Concurrent;
+using System.IO;
+
+namespace Ryujinx.HLE.HOS
+{
+    class HorizonFsClient : IFsClient
```
AcK77, line 16: Maybe add a TODO comment since content manager have to be done through `Horizon` project, then the property `_system` will not needed anymore, so `_system.LibHacHorizonManager.FsClient.Fs` could be passed in the constructor argument, and then the whole class could be moved in `Horizon` project ?

(I hope I don't have wrong here)

src/Ryujinx.HLE/HOS/HorizonFsClient.cs
```diff
@@ -0,0 +1,117 @@
...
+            // TODO: Return correct result here, this is likely wrong.
+
+            return LibHac.Fs.ResultFs.TargetNotFound.Handle().ToHorizonResult();
+        }
+
+        public Result OpenFile(out FileHandle handle, string path, OpenMode openMode)
+        {
+            var result = _fsClient.OpenFile(out var libhacHandle, path.ToU8Span(), (LibHac.Fs.OpenMode)openMode);
+            handle = new(libhacHandle);
+            return result.ToHorizonResult();
```
AcK77, line 91:
```suggestion

            return result.ToHorizonResult();
```

src/Ryujinx.HLE/HOS/HorizonFsClient.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public Result QueryMountSystemDataCacheSize(out long size, ulong dataId)
+        {
+            // TODO.
+
+            size = 0;
+            return Result.Success;
```
AcK77, line 99:
```suggestion

            return Result.Success;
```

src/Ryujinx.Horizon/Ngc/NgcMain.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    class NgcMain : IService
+    {
+        public static void Main(ServiceTable serviceTable)
+        {
+            NgcIpcServer ipcServer = new();
+
+            ipcServer.Initialize(HorizonStatic.Options.FsClient);
+
+            // TODO: Notification thread, requires implementing OpenSystemDataUpdateEventNotifier on FS.
```
AcK77, line 11: Maybe you could explain a bit what's the notification thread does, because if it will be implemented later, we could forgot :)

src/Ryujinx.Horizon/Sdk/Ngc/Detail/AhoCorasick.cs
```diff
@@ -0,0 +1,249 @@
...
+                        storage[storageOffset] = (byte)c;
+                        int nextNodeId = nodePlainIndex;
+
+                        if (_wordMap.Has(nodePlainIndex))
+                        {
+                            sb.AppendLine(Encoding.UTF8.GetString(storage[..(storageOffset + 1)]));
+
+                            // Some basic validation to ensure we imported the dictionary properly.
+                            int wordLength = _wordLengths[_wordMap.Rank1((uint)nodePlainIndex) - 1];
+                            Debug.Assert(storageOffset + 1 == wordLength);
```
AcK77, line 217:
```suggestion

                            Debug.Assert(storageOffset + 1 == wordLength);
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/AhoCorasick.cs
```diff
@@ -0,0 +1,249 @@
...
+                                int startIndex = storageOffset + 1 - wordLength;
+
+                                multiWord += Encoding.UTF8.GetString(storage[lastMultiWordIndex..startIndex]) + " ";
+                                lastMultiWordIndex = startIndex;
+                            }
+
+                            if (lastMultiWordIndex != 0)
+                            {
+                                multiWord += Encoding.UTF8.GetString(storage[lastMultiWordIndex..(storageOffset + 1)]);
+                                sb.AppendLine(multiWord);
```
AcK77, line 239:
```suggestion

                                sb.AppendLine(multiWord);
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/BinaryReader.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            if ((uint)(_offset + byteLength) <= (uint)_data.Length)
+            {
+                value = MemoryMarshal.Cast<byte, T>(_data[_offset..])[0];
+                _offset += byteLength;
+                return true;
+            }
+
+            value = default;
+            return false;
```
AcK77, line 29:
```suggestion

            return false;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/BinaryReader.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public bool Read<T>(out T value) where T : unmanaged
+        {
+            int byteLength = Unsafe.SizeOf<T>();
+
+            if ((uint)(_offset + byteLength) <= (uint)_data.Length)
+            {
+                value = MemoryMarshal.Cast<byte, T>(_data[_offset..])[0];
+                _offset += byteLength;
+                return true;
```
AcK77, line 25:
```suggestion

                return true;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/BinaryReader.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public int AllocateAndReadArray<T>(ref T[] array, int length, int maxLengthExclusive) where T : unmanaged
+        {
+            return AllocateAndReadArray(ref array, Math.Min(length, maxLengthExclusive));
+        }
+
+        public int AllocateAndReadArray<T>(ref T[] array, int length) where T : unmanaged
+        {
+            array = new T[length];
+            return ReadArray(array);
```
AcK77, line 40:
```suggestion

            return ReadArray(array);
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/BinaryReader.cs
```diff
@@ -0,0 +1,58 @@
...
+        public int ReadArray<T>(T[] array) where T : unmanaged
+        {
+            if (array != null)
+            {
+                int byteLength = array.Length * Unsafe.SizeOf<T>();
+                byteLength = Math.Min(byteLength, _data.Length - _offset);
+
+                MemoryMarshal.Cast<byte, T>(_data.Slice(_offset, byteLength)).CopyTo(array);
+                _offset += byteLength;
+                return byteLength / Unsafe.SizeOf<T>();
```
AcK77, line 52:
```suggestion

                return byteLength / Unsafe.SizeOf<T>();
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/BinaryReader.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public int ReadArray<T>(T[] array) where T : unmanaged
+        {
+            if (array != null)
+            {
+                int byteLength = array.Length * Unsafe.SizeOf<T>();
+                byteLength = Math.Min(byteLength, _data.Length - _offset);
+
+                MemoryMarshal.Cast<byte, T>(_data.Slice(_offset, byteLength)).CopyTo(array);
+                _offset += byteLength;
```
AcK77, line 51:
```suggestion

                _offset += byteLength;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/CompressedArray.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private ReadOnlySpan<BitfieldRange> GetBitfieldRanges()
+        {
+            return MemoryMarshal.Cast<uint, BitfieldRange>(_bitfieldRanges);
+        }
+
+        public bool Import(ref BinaryReader reader)
+        {
+            if (!reader.Read(out int bitfieldRangesCount) ||
+                reader.AllocateAndReadArray(ref _bitfieldRanges, bitfieldRangesCount) != bitfieldRangesCount)
```
AcK77, line 86: I'd put that on one line.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/CompressedArray.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                return false;
+            }
+
+            if (!reader.Read(out int bitfieldsCount) ||
+                reader.AllocateAndReadArray(ref _bitfields, bitfieldsCount) != bitfieldsCount)
```
AcK77, line 92: I'd put that on one line.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ContentsReader.cs
```diff
@@ -0,0 +1,400 @@
...
+                    result = _fsClient.QueryMountSystemDataCacheSize(out long cacheSize, DataId);
+                    if (result.IsFailure)
+                    {
+                        return result;
+                    }
+
+                    if (_cacheSize < (ulong)cacheSize)
+                    {
+                        result = NgcResult.InvalidSize;
+                        return result;
```
AcK77, line 139:
```suggestion
                        return NgcResult.InvalidSize;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ContentsReader.cs
```diff
@@ -0,0 +1,400 @@
...
+            SimilarFormTable table = new();
+
+            if (!table.Import(ref reader))
+            {
+                // Official firmware doesn't return an error here and just assumes the import was successful.
+                return NgcResult.DataAccessError;
+            }
+
+            similarFormTable = table;
+            return Result.Success;
```
AcK77, line 280:
```suggestion

            return Result.Success;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ContentsReader.cs
```diff
@@ -0,0 +1,400 @@
...
+            AhoCorasick ac = new();
+
+            if (!ac.Import(ref reader))
+            {
+                // Official firmware doesn't return an error here and just assumes the import was successful.
+                return NgcResult.DataAccessError;
+            }
+
+            trie = ac;
+            return Result.Success;
```
AcK77, line 321:
```suggestion

            return Result.Success;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ContentsReader.cs
```diff
@@ -0,0 +1,400 @@
...
+
+            return Result.Success;
+        }
+
+        private static byte[] DecompressGZipCompressedStream(byte[] data)
+        {
+            using MemoryStream input = new(data);
+            using GZipStream gZipStream = new(input, CompressionMode.Decompress);
+            using MemoryStream output = new();
+            gZipStream.CopyTo(output);
```
AcK77, line 373:
```suggestion

            gZipStream.CopyTo(output);

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/MatchRangeList.cs
```diff
@@ -0,0 +1,112 @@
...
+            if (_count == _capacity)
+            {
+                int newCapacity = _count * 2;
+
+                if (newCapacity == 0)
+                {
+                    newCapacity = 1;
+                }
+
+                Array.Resize(ref _ranges, newCapacity);
```
AcK77, line 45:
```suggestion
                Array.Resize(ref _ranges, newCapacity);

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/MatchRangeList.cs
```diff
@@ -0,0 +1,112 @@
...
+                    if (prevStartOffset == currStartOffset)
+                    {
+                        if (prevEndOffset <= currEndOffset)
+                        {
+                            prevEndOffset = currEndOffset;
+                        }
+                    }
+                    else if (prevEndOffset <= currStartOffset)
+                    {
+                        output.Add(prevStartOffset, prevEndOffset);
```
AcK77, line 75:
```suggestion
                        output.Add(prevStartOffset, prevEndOffset);

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/MatchRangeList.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            int baseIndex = 0;
+            int range = _count;
+
+            while (range != 0)
+            {
+                MatchRange currRange = _ranges[baseIndex + (range / 2)];
+
+                if (currRange.StartOffset < startOffset ||
+                    (currRange.StartOffset == startOffset && currRange.EndOffset < endOffset))
```
AcK77, line 97:
```suggestion
                if (currRange.StartOffset < startOffset || (currRange.StartOffset == startOffset && currRange.EndOffset < endOffset))
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/MatchRangeListState.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public MatchRangeListState()
+        {
+            MatchRanges = new();
+        }
+
+        public static bool AddMatch(ReadOnlySpan<byte> text, int startOffset, int endOffset, int nodeId, ref MatchRangeListState state)
+        {
+            state.MatchRanges.Add(startOffset, endOffset);
+            return true;
```
AcK77, line 17:
```suggestion

            return true;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+
+            Span<byte> data = stackalloc byte[4];
+            result = _contentsReader.GetVersionData(data);
+            if (result.IsFailure)
+            {
+                return Result.Success;
+            }
+
+            version = BinaryPrimitives.ReadUInt32BigEndian(data);
+            return Result.Success;
```
AcK77, line 53:
```suggestion

            return Result.Success;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+            noSeparatorText = RemoveWordSeparators(noSeparatorText, convertedText, noSeparatorMap, notSeparatorTrie);
+
+            trieMatched = false;
+
+            partialWordsTrie.Match(noSeparatorText, MatchSimple, ref trieMatched);
+
+            if (trieMatched)
+            {
+                matched = true;
+                return Result.Success;
```
AcK77, line 556:
```suggestion

                return Result.Success;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+            canonicalText.CopyTo(convertedText[..canonicalText.Length]);
+            convertedText[canonicalText.Length] = 0;
+
+            return Result.Success;
+        }
+
+        private static bool MatchCheck(ReadOnlySpan<byte> text, int matchStartOffset, int matchEndOffset, int nodeId, ref MatchCheckState state)
+        {
+            state.CheckMask |= state.RegionMask != 0 ? state.RegionMask : state.Option.SystemRegionMask;
+            return true;
```
AcK77, line 625:
```suggestion

            return true;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+        }
+
+        private static bool MatchSingleWord(ReadOnlySpan<byte> text, int matchStartOffset, int matchEndOffset, int nodeId, ref MatchState state)
+        {
+            MatchCommon(ref state, matchStartOffset, matchEndOffset);
+            return true;
```
AcK77, line 631:
```suggestion

            return true;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+
+            if (convertedEndOffset < 0)
+            {
+                convertedEndOffset = state.NoSeparatorMap.Set.BitVector.BitLength;
+            }
+
+            int endOffsetBeforeSeparator = TrimEnd(state.ConvertedText, convertedEndOffset);
+
+            MatchCommon(ref state, convertedStartOffset, endOffsetBeforeSeparator);
+            return true;
```
AcK77, line 647:
```suggestion

            return true;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
+            else
+            {
+                delimitedText[delimitedTextOffset++] = (byte)'a';
+            }
+
+            // Copy the word to our temporary buffer used for the next match.
+
+            int matchLength = matchEndOffset - matchStartOffset;
+
+            text.Slice(matchStartOffset, matchLength).CopyTo(delimitedText.Slice(delimitedTextOffset, matchLength));
```
AcK77, line 687:
```suggestion
            text.Slice(matchStartOffset, matchLength).CopyTo(delimitedText.Slice(delimitedTextOffset, matchLength));

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilter.cs
```diff
@@ -0,0 +1,878 @@
...
```
AcK77, line 785:
```suggestion
            text.Slice(matchStartOffset, matchLength).CopyTo(delimitedText.Slice(delimitedTextOffset, matchLength));

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
```diff
@@ -0,0 +1,793 @@
...
+
+            return matchCount;
+        }
+
+        private static bool IsValidEmailAddressCharacter(byte character)
+        {
+            return char.IsAsciiLetterOrDigit((char)character) ||
+                character == '-' ||
+                character == '.' ||
+                character == '_';
```
AcK77, line 219:
```suggestion
            return char.IsAsciiLetterOrDigit((char)character) || character == '-' || character == '.' || character == '_';
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
```diff
@@ -0,0 +1,793 @@
...
+                for (int index = 0; index < input.Length; index++)
+                {
+                    bool isWordSeparator = false;
+
+                    for (int separatorIndex = 0; separatorIndex < _wordSeparators.Length; separatorIndex++)
+                    {
+                        ReadOnlySpan<byte> separator = _wordSeparators[separatorIndex];
+
+                        if (index + separator.Length < input.Length &&
+                            input.Slice(index, separator.Length).SequenceEqual(separator))
```
AcK77, line 546:
```suggestion
                        if (index + separator.Length < input.Length && input.Slice(index, separator.Length).SequenceEqual(separator))
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
```diff
@@ -0,0 +1,793 @@
...
+                        {
+                            map.Set.TurnOn(index, separator.Length);
```
AcK77, line 548:
```suggestion
                            map.Set.TurnOn(index, separator.Length);

```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
```diff
@@ -0,0 +1,793 @@
...
+        }
+
+        protected static int TrimEnd(ReadOnlySpan<byte> text, int offset)
+        {
+            for (int separatorIndex = 0; separatorIndex < _wordSeparators.Length; separatorIndex++)
+            {
+                ReadOnlySpan<byte> separator = _wordSeparators[separatorIndex];
+
+                if (offset >= separator.Length &&
+                    text.Slice(offset - separator.Length, separator.Length).SequenceEqual(separator))
```
AcK77, line 574:
```suggestion
                if (offset >= separator.Length && text.Slice(offset - separator.Length, separator.Length).SequenceEqual(separator))
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
```diff
@@ -0,0 +1,793 @@
...
+                index += characterByteLength;
+            }
+
+            return index;
+        }
+
+        protected static bool MatchSimple(ReadOnlySpan<byte> text, int matchStartOffset, int matchEndOffset, int nodeId, ref bool matched)
+        {
+            matched = true;
+            return false;
```
AcK77, line 790:
```suggestion

            return false;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/SparseSet.cs
```diff
@@ -0,0 +1,124 @@
...
+                return false;
+            }
+
+            return _sbv.Import(ref reader);
+        }
+
+        public bool Has(long index)
+        {
+            int plainIndex = Rank1(index);
+            return plainIndex != 0 && Select1Ex(plainIndex - 1) == index;
```
AcK77, line 37:
```suggestion

            return plainIndex != 0 && Select1Ex(plainIndex - 1) == index;
```

src/Ryujinx.Horizon/Sdk/Ngc/Detail/Utf8Text.cs
```diff
@@ -0,0 +1,103 @@
...
+        public static Utf8ParseResult Create(out Utf8Text utf8Text, ReadOnlySpan<byte> text)
+        {
+            try
+            {
+                utf8Text = new(text);
+            }
+            catch (ArgumentException)
+            {
+                utf8Text = default;
+                return Utf8ParseResult.InvalidCharacter;
```
AcK77, line 53:
```suggestion

                return Utf8ParseResult.InvalidCharacter;
```

AcK77 (changes requested):
Some nits here and there, but otherwise it's an awesome work as usual!

src/Ryujinx.HLE/HOS/HorizonFsClient.cs
gdkchan: I'm not sure if I get this. The reason this class needs to exist is because Libhac does not implement `MountSystemData`. I'm not sure if it will be implemented on Libhac in the future, or if we will need to implement it ourselves. It seems that Libhac also does not implement `OpenDataStorageByDataId` (which we would end up calling on `MountSystemData`). I'm not sure if that is something that will be implemented on Libhac in the future either, or we will need to implement it ourselves. Also I guess that this function will need to call NCM, but it is something to be concerned about when FS is implemented on Horizon project.

AcK77: Then it's fine to leave things as-is and discuss about everything when FS will be implemented on Horizon project.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/CompressedArray.cs
gdkchan: This line is already 145 characters long. If I put in a single line it will be almost 200 characters long, and impossible to read without scrolling horizontally.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/CompressedArray.cs
gdkchan: More or less the same comment as above.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/CompressedArray.cs
AcK77: I have no issue to read it with a single line on a 1920x1080 screen... But for what it cost, you can leave things like that.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/MatchRangeList.cs
gdkchan: If I put it on a single line, it will be too long. As a general rule, I don't like letting any line be longer than 120 characters, because then it requires scrolling horizontally to be fully readable.

src/Ryujinx.Horizon/Sdk/Ngc/Detail/ProfanityFilterBase.cs
gdkchan: Line will be too long if I change that.

marysaka (approved):

GamerzHell9137:
> **STAR OCEAN THE SECOND STORY R** Demo Seems to need this too. The game boots but when you start a new game the game makes an "exception error stopping point".
> 
> The demo works flawlessly with Ignore missing service or this PR.

Where does it trigger the ```ngc:u is not implemented``` ?

PR #5682, opened 2023-09-14
Fix gl_Layer to geometry shader change not writing gl_Layer

gdkchan:
This fixes a regression ~~from #5551~~ (actually I think the regression happened before that change, probably on the one that refactors `TranslatorContext`?) that caused the generated geometry shader used to assign `gl_Layer` on older GPUs to not actually assign it, which caused Pokémon Scarlet/Violet to render mostly black.

Only affects old GPUs that do not support `gl_Layer` on vertex shaders.


marysaka (approved):
oops

riperiperi (approved):

PR #5686, opened 2023-09-14
Remove more usages of reflection binding

emmauss:
Remove more usages of ReflectionBinding. Only the controller window still uses it.


marysaka (approved):

TSRBerry (approved):
Lgtm!

PR #4734, opened 2023-04-27
Add "Create Shortcut" To app context menu

NitroTears:
This adds a 'Create Shortcut' button to an app/games right-click context menu. The shortcut is created on the desktop and launches the game directly without needing to use the GUI (in fullscreen or not to be confirmed, windows launches fullscreen while linux does not). closes #3250

win2
mac
linux

[Here is a gif of it in action.](https://cdn.discordapp.com/attachments/410432399500115968/1094833530757972168/bingo-og.gif)

~~As mentioned on discord `#if OS_Windows` is used instead of `OperatingSystem.IsWindows()` due to the annoying COM reference that is neccessary for this to work on Windows. See the project file: 
https://github.com/NitroTears/Ryujinx/blob/b89637a4b0dc2038ef9926924b5e9679f357d113/Ryujinx.Common/Ryujinx.Common.csproj#L12-L22~~ No COM reference is used anymore.

~~This currently does not work on MacOS as there doesn't seem to be a way to do that without mucking around in AppleScript, let alone in .NET (Which to me is a little bit out of scope).~~
Thanks to @JoseEstradaC we now have MacOS Shortcut support! (I have tested this on my mac, and it works good, but I would like someone else to verify this as well)

PR #5688, opened 2023-09-15
desktop shortcut support

DanielSvoboda:
This pull request adds two new options to the submenu:

1. **Start Game**: This option allows users to start the game directly from the submenu.
2. **Create Shortcut**: This option allows users to create a shortcut for the game on the desktop.

Here you can see the buttons in the menu:
print2

However, there are two issues that need to be resolved:

1. **Image conversion to .ico**: The current conversion of the JPEG image to an .ico file is not producing a high-quality image. We need to improve the `GenerateStandardFaviconIco` function to ensure that the converted image maintains its quality.
image
2. **Setting desiredTitleLanguage in GetApplicationIcon**: Currently, the `GetApplicationIcon` function is not correctly setting the `desiredTitleLanguage`. As a result, the generated icon is in a different language than expected. We need to fix this to ensure that the icon is displayed in the correct language.
image

To add the IWshRuntimeLibrary to your project in Visual Studio, you can follow these steps:

Right-click on the project name within the Solution Explorer.
Select “Add Reference”.
Select the “COM” tab, find and select the “Windows Script Host Object Model” in the listbox.
Click “Select”, and then click "OK"

### DIFF
--- a/src/Ryujinx.Ava/Assets/Locales/de_DE.json
+++ b/src/Ryujinx.Ava/Assets/Locales/de_DE.json
@@ -44,6 +44,8 @@
   "GameListHeaderFileExtension": "Dateiformat",
   "GameListHeaderFileSize": "Dateigröße",
   "GameListHeaderPath": "Pfad",
+  "GameListContextMenuCreateDesktopShortcut": "Verknüpfung Erstellen",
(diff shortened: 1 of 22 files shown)


TSRBerry:
I'm sorry.

Duplicate of #4734

PR #5697, opened 2023-09-19
Fixes compiled bindings in cheat window

emmauss:
Compiled bindings can only access public members with getters.


TSRBerry (approved):
Sorry for missing this. Thank you for the quick fix!

AcK77 (approved):

PR #4998, opened 2023-05-17
Ava UI: Input Menu Refactor

IsaacMarovitz:
**Changes:**
- Removes all remaining ReflectionBindings from Avalonia 🎉 
- Localised Keys (finally!)
- Added platform-specific keys (Win key will show as Cmd on Mac etc...)
- Destroyed and simultaneously created some crimes against humanity 

image
image

PR #5716, opened 2023-09-22
Use compiled binding for controller view

emmauss:
Convert remaining usages of ReflectionBinding in the controller view to use compiled Bindings.
This should now make the Ryujinx.Ava project AOT friendly.


IsaacMarovitz:
Conflicts with #4998

PR #5719, opened 2023-09-23
GPU: Discard data when getting texture before full clear

riperiperi:
When binding a render target, we ask the texture cache for the texture, which synchronizes the data both from guest memory, and performs any deferred copies from other textures with a copy dependency. When clearing a texture, this same path is taken, but it's possible that none of that data that was synchronized or copied would ever be used, meaning it just wasted time.

For us to say we can fully ignore the existing data when clearing a texture, the following conditions need to be met:
- All aspects of the texture must be cleared. For color textures, that's a component mask of 15 (RGBA), and for depth textures, that's depth and stencil clears (including stencil mask) when the texture has depth or stencil aspects.
- Viewport scissor must be disabled or larger than the screen scissor state.
- The texture size must be less than or equal to the screen scissor. (this is passed as the sizeHint to the texture cache)

When all these conditions are met, loading texture data from guest memory and incoming copy dependencies can be entirely skipped.

This aims to unblock https://github.com/Ryujinx/Ryujinx/pull/4365, as it prevents the extra copies from happening in BOTW and TOTK for depth/color aliasing, as these R32/D32 textures are _not_ actually used for the same purpose, but just happen to overlap in memory.

### Benefit:
- When creating textures going directly into a clear, their data does not need to be loaded from guest memory. This saves layout conversion and upload time, which could reduce the stutter when these textures first show up, or need to recreate for some reason.
  - When overlapping incompatible textures already exist (and their data has not been overwritten), the emulator already knows that the data is garbage, so will not load it. The new clear rule has the benefit of not needing any overlapping textures to work, and never accidentally losing any data.
- This could prevent copies from format aliased textures. Most of these actually alias memory right now, but this isn't always optimal and can actually be broken in some cases (older opengl drivers, amd gpus). For cases where it's optimal or required, copying between aliased textures introduces cost that we'd like to skip when possible.
  - One such case is D32->R32 aliasing, which does affect all GPUs. This becomes important with https://github.com/Ryujinx/Ryujinx/pull/4365, which is required for some games to function properly.
  
This rule could become important for a future rewrite of how the texture cache works to better support games like Fast RMX which rapidly rearrange render targets in memory. Generally, the more cases we can "steal" a host texture from another texture whose data has become invalid, the better.

It's probably worth testing this in a variety of games. You probably won't perceive anything being any faster due to this mostly affecting initial texture creation without #4365, but the improvement is there. (games like MK8D/SMO do avoid setting texture data for render targets on creation)

### DIFF
--- a/src/Ryujinx.Graphics.GAL/Format.cs
+++ b/src/Ryujinx.Graphics.GAL/Format.cs
@@ -336,6 +336,45 @@ namespace Ryujinx.Graphics.GAL
         }
 
         /// <summary>
+        /// Checks if the texture format is a depth or depth-stencil format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
+        /// <returns>True if the format is a depth or depth-stencil format, false otherwise</returns>
+        public static bool HasDepth(this Format format)
+        {
+            switch (format)
+            {
+                case Format.D16Unorm:
+                case Format.D24UnormS8Uint:
+                case Format.S8UintD24Unorm:
+                case Format.D32Float:
+                case Format.D32FloatS8Uint:
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks if the texture format is a stencil or depth-stencil format.
+        /// </summary>
+        /// <param name="format">Texture format</param>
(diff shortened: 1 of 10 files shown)


gdkchan (approved):
lgtm, thanks!

marysaka (approved):

PR #5727, opened 2023-09-26
GPU: Don't create tracking handles for buffer textures

riperiperi:
Buffer texture memory is handled by the buffer cache - the texture shouldn't create any tracking handles as they aren't used. This change simply makes them create and iterate 0 tracking handles, while keeping the rest of the texture group around.

This prevents a possible issue where many buffer textures are created as views of overlapping buffer ranges, and virtual regions have many dependant textures that don't actually contribute anything to handle state.

Also included:
- Fix a longstanding issue where every texture returned from the interval tree lookup would resize the lookup array. (yes, every)
- Buffer textures now skip checking view compatibility with any other textures, since they are never compatible and don't remove each other due to overlap.
  - In MK1, more than 100 buffer textures would overlap at once, making this overlap check very expensive as well as pointless.

This PR is thanks to testing and profiling by @gdkchan. I don't have an affected game myself.

Should improve performance in Mortal Kombat 1, possibly certain UE4 games when FIFO raises to 100%. Needs testing in UE4 games and the 2 Mortal Kombat games.

### DIFF
--- a/src/ARMeilleure/Translation/IntervalTree.cs
+++ b/src/ARMeilleure/Translation/IntervalTree.cs
@@ -189,7 +189,7 @@ namespace ARMeilleure.Translation
             {
                 if (start.CompareTo(node.End) < 0)
                 {
-                    if (overlaps.Length >= overlapCount)
+                    if (overlaps.Length <= overlapCount)
                     {
                         Array.Resize(ref overlaps, overlapCount + ArrayGrowthSize);
                     }
--- a/src/Ryujinx.Common/Collections/IntervalTree.cs
+++ b/src/Ryujinx.Common/Collections/IntervalTree.cs
@@ -192,7 +192,7 @@ namespace Ryujinx.Common.Collections
                     {
                         if (start.CompareTo(overlap.End) < 0)
                         {
-                            if (overlaps.Length >= overlapCount)
+                            if (overlaps.Length <= overlapCount)
                             {
                                 Array.Resize(ref overlaps, overlapCount + ArrayGrowthSize);
                             }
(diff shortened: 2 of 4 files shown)


GamerzHell9137:
Fixes FPS during gameplay in R-TYPE FINAL 2 (Demo, most likely translates to full game). Pretty much goes from 0 to 60 FPS with the change. Game has still FIFO maxed at the opening sequence and other graphical issues not related to this PR.

Master: 
image-3

PR:
image-5

gdkchan:
Performance seems much better on Mortal Kombat 1 here with this change.
Master:
image
image
PR:
image
image
The frame rate is also more stable with this PR, on master it was dropping quite a bit.
I'm probably not the best person test it though as I only have an integrated GPU on this PC right now.

gdkchan (approved):
lgtm, thanks! I think it could be better to create a separate `FindOrCreateTexture` method just for buffer textures, given that most of the code there does nothing when the overlaps count is 0. But maybe the amount of code that can still be shared makes that not worth it.

I was pretty surprised with the resize bug, but I guess something like that is easy to go unnoticed, as it does not create correctness issues, and only becomes apparent on games creating textures non-stop.

marysaka (approved):

sakuka47:
Hi, great improvement in the MK1 game but the cinematics still bug.

Image tearing

Niwu34:
Tried Mortal Kombat 11:
Max fifo can still reach 99% (60fps all the time) but is lower in general

`Master:`
Screenshot 2023-09-26 110610

`PR:`
Screenshot 2023-09-26 111037

GamerzHell9137:
> Hi, great improvement in the MK1 game but the cinematics still bug.
> 
> Image tearing

Happens on Switch too, not emulator issue.

PR #5516, opened 2023-08-01
Ava UI: Refactor `async` usage

IsaacMarovitz:
Changes:
- Removes `async void` replacing with `async Task` where possible
- Makes `LoadApplications()` async

Significantly improves opening times, better than current Ava or GTK.

https://github.com/Ryujinx/Ryujinx/assets/42140194/c34b4514-71b0-4050-8d9f-3562714dc44c

Video Source: piplup

PR #5728, opened 2023-09-26
Ava: Fix regressions by rewriting CheckLaunchState

AcK77:
This fix a regression caused by #5516 when CheckLaunchState is assumed to be executed async.
Since the UI isn't created when launch state are check, the vmcount popup, the missing keys popup or the fullscreen game loading on startup just make the UI hang.
if we revert the async changes to use the dispatcher and rewrite the method, it works as expected.


gdkchan (approved):

gdkchan (approved):
lgtm, thanks.

marysaka (approved):

TSRBerry:
image
Just to confirm that the `ShowVmMaxMapCountDialog` works as well! :D

PR #5735, opened 2023-09-27
[Avalonia] Add custom refresh interval target

jcm93:
Firstly, the bulk of the original work for this was done by [@makotech222](https://github.com/makotech222). They ran out of time so I took it over and made some changes. Thanks to them for the initial work.

## Description
Replaces the VSync option in Ryujinx with the Present Interval State option. This includes the original functionality of VSync, which by default is unchanged, but also adds a third option for a specific refresh interval. In some titles, this specific interval acts as an FPS cap. In others, it will lead to odd behavior, or do nothing. The option to enable the custom refresh interval is behind a checkbox, or must be explicitly selected in a dropdown. When selected/enabled, the former VSync hotkey will include the Custom interval when cycling through present interval state options. There are also hotkey options to raise and lower the custom refresh interval.
Screenshot 2023-09-27 at 2 07 26 PM Screenshot 2023-09-27 at 2 07 32 PMScreenshot 2023-09-27 at 2 08 07 PM

## Motivation
This new option has the utility of being able to cap FPS arbitrarily for some games without the use of external tools e.g. RTSS, which aren't available on all platforms. It also will allow players in some games to speed up emulation without the speedup being completely unbounded as it was with VSync off formerly. This is similar to the "turbo" functionality found in other emulators.

## Discussion
There was some difficulty in settling on an appropriate name, UI, and implementation for this feature. I tried to create something balanced that gives users the an idea power and flexibility of what the feature can do, without misleading them into thinking they will be able to speed up all games or flawlessly cap FPS.

## Issues
Developing on macOS, I am unable to test GTK. The intent is to not provide this feature via GTK, given its imminent deprecation, and only preserve the legacy VSync behavior. I believe that's what I've done, but as I was unable to test I would very much welcome help if things are broken.

## Misc
This PR remains in draft state for moment. Needs:
- clean up button appearance in main status bar
- custom interval value does not sync when set in settings and then shown in main status bar, or vice versa.
- couple small remaining sync issues with status bar button


sakuka47:
I'm going to test this Pr, it can be very useful for mods.
Like those that limit to 45fps, it can allow you to put directly into ryujinx without going through a third party. I'm testing under pkm Scarlet and I'll give you feedback tomorrow.

NineKain:
This is something I've wanted for a long time, thanks so much!!

fann1993814:
Thanks for yours contribution.
This is very useful for me, because it let my MacBook temperature lowered.

I have a  feedback.

To make my MacBook temperature lowered, I would set the custom fps to 30 (50%).
However, some game base fps is 30, it will drop to 15 fps, so I have to set Origin Vsync(60) for those game (base is 30fps)
I need to switch the setting frequently for different games.

jcm93:
> Thanks for yours contribution. This is very useful for me, because it let my MacBook temperature lowered.
> 
> I have a feedback.
> 
> To make my MacBook temperature lowered, I would set the custom fps to 30 (50%). However, some game base fps is 30, it will drop to 15 fps, so I have to set Origin Vsync(60) for those game (base is 30fps) I need to switch the setting frequently for different games.

Thanks for the feedback. This is why I was reluctant to use "fps" to describe what this change does. Some games will even change the "base FPS" (in Ryujinx this is called the swap interval) mid-game, for example for cutscenes. I decided it was more appropriate to respect the game's choice of swap interval despite possible inconsistent behavior between titles. I am guessing that in the future, there will be better support in Ryujinx for title-specific settings.

jcm93:
Something I'm wondering about proposing: Does the project think it would be better to keep the option named "VSync", and then have the options be "On", "Off", and "Custom Refresh Interval"? I've been reflecting on how "Present Interval State" is somewhat wordy and overly verbose. Same functionality and similar tooltips otherwise; just thinking about naming.

NineKain:
Just my two cents, the option being called VSync is more readable and to the point:)

jcm93:
Updated, keeping the old "VSync" naming for the option, modifying tooltips, etc. The status bar UX and name is now identical when "Custom" is not enabled (text is either green "VSync", red "VSync", or blue "Custom" with gear). Screenshots in OP are updated.

jcm93:
Updated again to move the setting under "Hacks" with a couple UX tweaks accordingly.

NineKain:
Is this still alive?

jcm93:
As far as I'm concerned it is. I can rebase it at any time, it's just awaiting a request to do so and review from maintainers.

makotech222:
Requesting a rebase if you have the time. And please reviewers, give this an approval if you can.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1065,9 +1108,40 @@ private bool UpdateFrame()
...
-                        case KeyboardHotkeyState.ToggleVSync:
-                            Device.EnableDeviceVsync = !Device.EnableDeviceVsync;
-
+                        //todo default
+                        case KeyboardHotkeyState.TogglePresentIntervalState:
+                            PresentIntervalState oldState = Device.PresentIntervalState;
+                            PresentIntervalState newState;
+                            if (oldState == PresentIntervalState.Switch)
+                            {
+                                newState = PresentIntervalState.Unbounded;
```
MutantAura, line 1117: Is there any reason why we can't just update `Device.PresentStateInterval` here?
Creating two excess enums seems wasteful, and you just end up using it to update itself later anyway.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -1065,9 +1108,40 @@ private bool UpdateFrame()
...
+                                {
+                                    newState = PresentIntervalState.Switch;
+                                }
+                            }
+                            else
+                            {
+                                newState = PresentIntervalState.Switch;
+                            }
+                            UpdatePresentIntervalState(this, new ReactiveEventArgs<PresentIntervalState>(oldState, newState));
+                            _viewModel.ShowCustomPresentIntervalPicker =
```
MutantAura, line 1135: Why not just set this in the same condition that already checked for the custom interval above?

src/Ryujinx.Ava/AppHost.cs
jcm93: `Device` could be null here, and given the need to (potentially) change the present mode in the Vulkan renderer, it seemed cleaner to move the actual update into the `UpdatePresentIntervalState` function. We do the silliness with ReactiveEventArgs so that the same function can also be neatly used when clicking to toggle the state via the status bar UI.

src/Ryujinx.Ava/AppHost.cs
jcm93: This was just a style/readability choice, I figured it was cleaner to have the flow of 1. figure out what the new state is, then 2. do the updates that need to be done accordingly. Can certainly be changed.

jcm93:
As an aside, I think I'm going to rename `PresentIntervalState` internally to something clearer like `VSyncMode` to match the UI and be more readable. Will do that when I have time.

jcm93:
Done. I also combined the two pieces of code that do the toggle logic, one in `AppHost` and one in `MainWindowViewModel`, into one function, `VSyncToggle()` in `AppHost`, to clean up the code and remove redundancy. I've done testing to make sure nothing broke, but testing of this new commit would be good to make sure I didn't miss anything.

makotech222:
Latest changes are working well for me. Thanks!

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -1430,6 +1457,32 @@ public void Load(ConfigurationFileFormat configurationFileFormat, string configu
...
+                    Pause = configurationFileFormat.Hotkeys.Pause,
+                    ToggleMute = configurationFileFormat.Hotkeys.ToggleMute,
+                    ResScaleUp = configurationFileFormat.Hotkeys.ResScaleUp,
+                    ResScaleDown = configurationFileFormat.Hotkeys.ResScaleDown,
+                    VolumeUp = configurationFileFormat.Hotkeys.VolumeUp,
+                    VolumeDown = configurationFileFormat.Hotkeys.VolumeDown,
+                    CustomVSyncIntervalIncrement = Key.Unbound,
+                    CustomVSyncIntervalDecrement = Key.Unbound,
+                };
+                configurationFileFormat.CustomVSyncInterval = 90;
```
riperiperi, line 1481: For a new configuration file, it's initialized as 120. For an upgraded one (here) and SDL2 it's 120. It should probably be consistent.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -236,6 +240,69 @@ private void UpdateColorSpacePassthrough(object sender, ReactiveEventArgs<bool>
...
         }
 
+        public void UpdateVSyncMode(object sender, ReactiveEventArgs<VSyncMode> e)
+        {
+            if (Device != null)
+            {
+                Device.VSyncMode = e.NewValue;
+                Device.UpdateVSyncInterval();
+            }
+            //vulkan present mode may change in response, so recreate the swapchain
```
riperiperi, line 250: Other backends could change their presentation mode based on this too. The main thing is that when it's unbounded, we disable host vsync.

src/Ryujinx.Ava/AppHost.cs
```diff
@@ -236,6 +240,69 @@ private void UpdateColorSpacePassthrough(object sender, ReactiveEventArgs<bool>
...
+            _renderer.Window?.ChangeVSyncMode((Ryujinx.Graphics.GAL.VSyncMode)e.NewValue);
+
+            _viewModel.ShowCustomVSyncIntervalPicker = (e.NewValue == VSyncMode.Custom);
+
+            ConfigurationState.Instance.Graphics.VSyncMode.Value = e.NewValue;
+            ConfigurationState.Instance.ToFileFormat().SaveConfig(Program.ConfigurationPath);
```
riperiperi, line 256: None of the other status bar actions are saving config right now, so this is a bit inconsistent. Also, this seems to call twice when you click to switch vsync mode.

riperiperi (commented):
I like the functionality here a lot more than its initial iteration. Hiding the option entirely until this checkbox is here, then allowing granular updates from the status bar makes it a lot more compelling as a feature.

Sorry for the delayed review, had I known the implementation had changed so much I would have gotten back to it sooner.

The weirdest thing right now is how the GDK UI handles things. I think it should use the new configuration setting (and delete the old one), but still snap back to Vsync On/Off when the user toggles it or settings are changed.

src/Ryujinx.Ava/UI/ViewModels/SettingsViewModel.cs
```diff
@@ -137,7 +142,78 @@ public bool DirectoryChanged
...
+        }
+
+        public string CustomVSyncIntervalPercentageText
+        {
+            get
+            {
+                string text = CustomVSyncIntervalPercentageProxy.ToString() + "%";
+                return text;
+            }
+            set
```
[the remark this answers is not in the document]
jcm93: Appears not to; removed.

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -403,13 +410,85 @@ public Brush ProgressBarForegroundColor
...
+                OnPropertyChanged();
+            }
+        }
+
+        public int CustomVSyncIntervalPercentageProxy
+        {
+            get => _customVSyncIntervalPercentageProxy;
+            set
+            {
+                int newInterval = (int)(((decimal)value / 100) * 60);
```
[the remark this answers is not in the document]
jcm93: Resolved, used float instead.

src/Ryujinx.Common/Configuration/VSyncMode.cs
```diff
@@ -0,0 +1,15 @@
+using Ryujinx.Common.Utilities;
+using System;
+using System.Text.Json.Serialization;
+
+namespace Ryujinx.Common.Configuration
+{
+    [JsonConverter(typeof(TypedStringEnumConverter<VSyncMode>))]
+    [Flags]
+    public enum VSyncMode
```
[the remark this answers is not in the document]
jcm93: Seems to work fine with a vanilla enum; changed.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -81,14 +82,18 @@ private void UpdateSwapInterval(int swapInterval)
             _swapInterval = swapInterval;
 
             // If the swap interval is 0, Game VSync is disabled.
-            if (_swapInterval == 0)
-            {
-                _nextFrameEvent.Set();
-                _ticksPerFrame = 1;
-            }
-            else
+            if (_targetVSyncIntervalChanged)
```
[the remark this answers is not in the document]
jcm93: Adjusted this block to just use the old logic. Also got rid of the invocation thing, and instead directly compare the SurfaceFlinger's interval with the device's interval inside of `Compose()`, since that route seemed more in line with how other things like this work.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -371,14 +376,11 @@ public void Compose()
...
+                    if (_device.VSyncMode == VSyncMode.Unbounded)
                     {
-                        if (_swapInterval != 0)
-                        {
-                            UpdateSwapInterval(0);
-                        }
+                        UpdateSwapInterval(0);
                     }
-                    else if (item.SwapInterval != _swapInterval)
+                    else
```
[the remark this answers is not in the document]
jcm93: Resolved per above.

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
jcm93: Changed to 120. I figure larger default is better so initial discovery is possibly more apparent.

src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs
```diff
@@ -468,10 +468,25 @@ public class GraphicsSection
             public ReactiveObject<string> ShadersDumpPath { get; private set; }
 
             /// <summary>
-            /// Enables or disables Vertical Sync
+            /// Toggles VSync on or off. Deprecated, included for GTK compatibility.
```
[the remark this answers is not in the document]
jcm93: Agree, went ahead and removed the last `EnableVsync` stuff from GTK. Bit awkward due to use of both enums but should work fine.

src/Ryujinx.Ava/AppHost.cs
jcm93: Fixed the duplicate call.

To me, persisting to the config seems correct if we change the vsync mode from the status bar; it seems like the emulator should always just remember the most recent vsync setting, and it might be confusing if someone sets a custom rate, then goes into settings and sees a different one. I recognize that it's inconsistent though. Let me know if you want me to change it so it doesn't persist though.

This is a separate conversation but I'd be willing to PR consistency across the board. It's always seemed more like a bug to me that the emulator doesn't remember/persist volume set from the status bar, for example. From a cursory check, it looks like aspect ratio settings are also saved to the config so that's another existing inconsistency.

MutantAura: Personally I don't think any status bar options should persist. Not only should it be consistent with the other options (which is king really), the status bar to me is about trying stuff out and not having to go back into the settings menu if you don't like it, or more fundamentally, a bit of a bandaid for game specific settings. I don't want something I change when playing Breath of the Wild, to impact what happens when I boot Smash Bros.
If I set a 2x refresh on BoTW to hit 60, I don't want that to persist when I boot a true 60FPS title.

MutantAura: To add more detail to that, whenever we do implement game-specific settings I think it would be good to make the status bar update that game-specific setting. But while we don't have that, I would prefer it to just be for that session.

jcm93:
Forgot that I had not addressed persistence. Rebased, and made status bar/hotkey changes not persist in the config.

jcm93:
It's in a fit state to review and is still awaiting comment from maintainers. I'll rebase again later today since it looks like there are merge conflicts.

jcm93:
Rebased and squashed. Also elected to rename the VSync On/Off to "Switch" and "Unbounded" to reflect recent discussions bikeshedding the VSync option.

OrdinaryCloverOC:
tested the mod "ultimate s" for smash bros using 90fps and did nothing.

jcm93:
Please elaborate on what you mean by "did nothing;" if you would include exactly what options you set, what precisely you expected as a result, versus what did happen, that would be helpful. You may be seeking behavior different from the functionality this PR implements.

OrdinaryCloverOC:
sorry, set it to 90, and it was uncapped

jcm93:
> sorry, set it to 90, and it was uncapped, for all i know it's a bug in the mod since they're attempting ryujinx-specific fixes, i believe it implements uncapped fps.

Could you reproduce this, then share a screenshot of your game window including the status bar while the refresh rate is supposed to be capped at 90, but isn't?

Fl4sh9174:
This PR is fantastic! I really hope it gets merged; it will be very useful for FPS mods.

M0REKZ:
I need something like this, im connected to a 30hz screen and that breaks some 60 fps games by making them stutter

NineKain:
Hope this gets merged

PR #5741, opened 2023-09-28
Ryujinx.Ava: Fix menus on gamescope

marysaka:
gamescope seems to only support rendering one surface window at a time.

This works fine for our GTK frontend as everything is embedded in the same window.

Avalonia on the other hand will attempt to create a new window for every menu, making them not render properly.

Adding insult to the injury, gamescope seems to ignore the surface update for those Avalonia's menus and refuse to present them.

To workaround all of this, we switch OverlayPopups to true to ensure we embed all popups in the window.

Note that all our current dialogs (about us, warning about keys,...) will still show on top without the main window behind because of gamescope behavior.


AcK77:
This is really fixed in : https://github.com/AvaloniaUI/Avalonia/pull/14366
We have to wait Avalonia to upstream a new version and update it in Ryujinx aswell.
We can close this PR.

PR #5750, opened 2023-10-01
Implement textureSamples texture query shader instruction

gdkchan:
This implements support for querying the amount of samples that a multisample texture has on the shader. This was not supported before, and it would actually generate invalid code for that case. The handlng of the `TXQ` shader instruction was improved a bit to log unsupported query types (instead of assuming that everything is a dimension query as it did before).

I found this while testing the game Cocoon. Unfortunately the game still crashes while compiling shaders on AMD, so I couldn't get very far, and I don't know which effects this change may have on the game...


marysaka (approved):

GamerzHell9137:
Fixes water in COCOON

Before: 
image-3

After:
image-3

PR #3284, opened 2022-04-19
Fix broken motion controls when using SDL2.

Haronee:
Right now on Master, when using motion controls with SDL2 (the default option), multiple games suffer from completely broken motion controls making them unplayable when the option is used.
Some games such as Kirby and the Forgotten Land require motion controls to complete them at 100%.

This was due to a missing `-` in front of `gyroscope.Z` in the controller config file.
This PR fixes that.

I tested Mario Kart 8 Deluxe and The Legend of Zelda: Breath of the Wild and in both cases it fixed the issues reported by users and myself, this PR should also fix most of the other issues reported as they were all the same issue.

Needs testing on multiple games to make sure nothing broke and needs testing on different controllers such as DS4 and DS5 just to be sure. I tested on the Pro Controller and Joycon where it worked fine.

Some footage:

Master:

https://user-images.githubusercontent.com/53187465/164028119-969a32fa-f81d-4358-999f-1dd560e1f61f.mp4

PR:

https://user-images.githubusercontent.com/53187465/164027698-0f61811d-e50f-404d-bfe3-a2ef5adb8f6e.mp4

Thanks to @Thog for finding the issue!

Closes #2480

PR #5751, opened 2023-10-01
Added option to invert gyroscope's Z motion 

maxthetomas:
Horizontal gyroscope motion can be broken for some people. I've seen previous PR that's related to this issue: #3284. That did not fix it for me so I decided to add option to allow users to change that behavior.

Added `InvertZAzis` field to MotionConfigController and option to change it in UI:
image

It still prefers fix from #3284 and uses `-gyroscope.Z` by default.

### DIFF
--- a/src/Ryujinx.Common/Configuration/Hid/Controller/Motion/MotionConfigController.cs
+++ b/src/Ryujinx.Common/Configuration/Hid/Controller/Motion/MotionConfigController.cs
@@ -21,5 +21,10 @@ namespace Ryujinx.Common.Configuration.Hid.Controller.Motion
         /// Enable Motion Controls
         /// </summary>
         public bool EnableMotion { get; set; }
+
+        /// <summary>
+        /// Invert Z axis (Horizontal)
+        /// </summary>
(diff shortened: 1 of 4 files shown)


german77:
Normally flipping an axis requires changing a few other axis to match gyro with accelerometer on the kalman filter. Isn't this the case here?

I won't deny that only this axis could be wrong from the very start. But feels a bit weird to me.

emmauss:
If Z axis motion is broken for people, find out why it's broken for them, and if it's an issue with our motion system, then you open a **fix** for it, not a **hack**. There could be many reason why motion is broken and not a result of out motion system.

AcK77:
As @emmaus stated in the previous comment, we usually prefer a proper fix instead of a hack to fix something. I think the issue needs more investigation to understand why in some cases the Z axe is wrong.
Thanks for your contribution, we could reopen the PR if you’ll edit things with a proper fix.

steelplasma256:
How is the very basic option to invert an axis considered a "Hack". Rather it should be there from the start, If this had been merged then I wouldn't be having the same problem 8 months later because it wasn't fixed.

PR #5753, opened 2023-10-01
Decrement nvmap reference count on surface flinger prealloc

gdkchan:
The game Sifu was calling `SetPreallocatedBuffer` and replacing buffers that have been queued. Because it is replaced before consumption, the nvmap reference count is never decremented, and this later causes the nvmap `Free` call to set the `NotYetFreed` flag. This prevents the memory attributes from being restored from `Uncached` to `None`, which in turn causes the `UnmapPhysicalMemory` to return an error, and ultimately the game crashes with `GuestProgramBrokeExecution`. This is fixed here by making sure the problematic case also decrements the reference count. The crash was somewhat random though.


AcK77 (approved):

riperiperi (approved):
Nice catch.

PR #5757, opened 2023-10-03
Use unique temporary variables for function call parameters on SPIR-V

gdkchan:
Currently the SPIR-V backend will copy each function call parameter to a temporary variable. Right now, it reserves one set of variables per function (one for each input parameter), and reuses them across calls to the same function. For some reason, this now confuses the AMD compiler on their latest drivers, with causes rendering issues that didn't exist before.

This PR changes how it works, now all function calls gets their own unique set of temporary variables without any reuse, which seems to fix the issue.

Example on Pokémon Scarlet.
Before:
image
After:
image

Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 5750;
+        private const uint CodeGenVersion = 5757;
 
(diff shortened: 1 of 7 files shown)


Littl3Guy:
Also fixes Tears of the Kingdom.
Before:
Untitled2
After:
Untitled1

I'll test more games later.

goldenx86:
Reported this issue to AMD over the Vanguard program but they require more information about the codebase than I can provide.
Consider applying to Vanguard so you can talk to the driver developers directly, they are very receptive if you provide all the info they need.

gdkchan:
> Reported this issue to AMD over the Vanguard program but they require more information about the codebase than I can provide. Consider applying to Vanguard so you can talk to the driver developers directly, they are very receptive if you provide all the info they need.

Is that bug report public? If so you could post the link here.

goldenx86:
> Is that bug report public? If so you could post the link here.

No, "what happens in Vanguard stays in Vanguard". For the public channels, you can report to amdvlk.

marysaka (approved):

riperiperi (approved):
Odd thing to break, nice catch.

tgcordell:
FYI - had a crash on macOS w/ M1 and the log files show some things related to SPIR-V:
https://github.com/Ryujinx/Ryujinx/issues/5782

PR #5760, opened 2023-10-03
Avalonia: Make slider scrollable with mouse wheel

ahmadtantowi:
This will allow changing slider value with mouse pointer wheel, the new value will be based on tick frequency

### DIFF
--- a/src/Ryujinx.Ava/UI/Views/Input/MotionInputView.axaml
+++ b/src/Ryujinx.Ava/UI/Views/Input/MotionInputView.axaml
@@ -3,6 +3,7 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:controls="clr-namespace:Ryujinx.Ava.UI.Controls"
     xmlns:ui="clr-namespace:FluentAvalonia.UI.Controls;assembly=FluentAvalonia"
     xmlns:locale="clr-namespace:Ryujinx.Ava.Common.Locale"
     xmlns:viewModels="clr-namespace:Ryujinx.Ava.UI.ViewModels"
@@ -23,11 +24,11 @@
                     Margin="0"
                     HorizontalAlignment="Center"
                     Text="{locale:Locale ControllerSettingsMotionGyroSensitivity}" />
-                <Slider
+                <controls:SliderScroll
                     Margin="0,-5,0,-5"
                     Width="150"
                     MaxWidth="150"
-                    TickFrequency="0.01"
+                    TickFrequency="1"
                     IsSnapToTickEnabled="True"
                     SmallChange="0.01"
                     Maximum="100"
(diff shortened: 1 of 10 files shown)


src/Ryujinx.Ava/UI/Views/Input/MotionInputView.axaml
```diff
@@ -23,11 +24,11 @@
                     Margin="0"
                     HorizontalAlignment="Center"
                     Text="{locale:Locale ControllerSettingsMotionGyroSensitivity}" />
-                <Slider
+                <controls:SliderScroll
                     Margin="0,-5,0,-5"
                     Width="150"
                     MaxWidth="150"
-                    TickFrequency="0.01"
+                    TickFrequency="1"
```
TSRBerry, line 31: Could you revert that?

TSRBerry (approved):
Awesome, thank you!
Code lgtm, but I'd like the gyro controls to stay as precise as they are now.

src/Ryujinx.Ava/UI/Views/Input/MotionInputView.axaml
ahmadtantowi: Since the value is int, I think using previous value will make no changes. It does, but by scrolling a lot
Is it skill okay?

MutantAura:
Scrolling on sliders that are within the context of a general scrollviewer is also a little jank at the moment.
An easy example is the general stick deadzone and range. It will both adjust the deadzone AND scroll the window. Unsure if we can disable the general scrollviewer input event while hovering sliders or other relevant controls.

ahmadtantowi:
Have considered to exclude this in controller input view, but ended out to update all slider
Actually we can make scrollviewer to stop by holding shift key since there is no horizontal scrollview

src/Ryujinx.Ava/UI/Views/Input/MotionInputView.axaml
TSRBerry: What do you mean with "by scrolling a lot"?

ahmadtantowi: I mean scrolling the mouse wheel, need to scroll several times for the value to change

TSRBerry: I see. Let me check that in a bit and report back.

TSRBerry: Leave it as is, the current way works as expected! I thought it behaves differently, but since that doesn't seem to be the case there is no reason to revert that change.

TSRBerry:
I set the `PointerWheelEvent` as handled after we process it with the `SliderScroll` element, which prevents other scrolling elements from receiving this event as well.

You can learn more about that [here](https://docs.avaloniaui.net/docs/input/routed-events#adding-and-implementing-an-event-handler-for-a-routed-event) (You need to scroll down until `The Concept of Handled`, since I can't link to that directly).

ahmadtantowi:
Thanks for the fix and learning resource, didn't expect to miss those property.

TSRBerry:
Although it's only slightly related to this PR, I added the scroll event to the `VolumeStatus` button as well, so you can change the volume by scrolling while hovering over the button now!

marysaka (approved):

PR #5766, opened 2023-10-06
HLE: Fix Mii crc generation and minor issues

riperiperi:
Validating CRCs for data and device involves calculating the crc of all data including the crc being checked, which should then be 0.

The crc should be _generated_ on all data _before_ the crc in the struct. It shouldn't include the crcs themselves.

This fixes all generated miis (eg. default) having invalid crcs. This does not affect mii maker normally, as that generates its own StoreData crcs.

Also includes some miscellaneous fixes:
- Use Array<> instead of a byte storage field for CoreData, Nickname and Ver3StoreData
  - This was observed causing core data to become all 0s, and nickname to only display the first character in debug mode. This probably depends on if the struct is copied, I imagine it doesn't think the undefined space in the struct holds any data.
  - Ver3StoreData was not tested to have an issue, but it has been changed just in case.
  - This leaves me a bit worried for other places in the codebase that do this. I recommend searching for `MemoryMarshal.CreateSpan(ref ` to see similar examples.
- Flipped comparison that always errored on `ConvertCharInfoToCoreData`.

Does not fix MK8D crash. I have more info on it, though:

Loading a save file with a mii saved from Ryujinx _or_ the switch does work if the mii doesn't currently exist in the database (deleted in mii maker). In this case, it tries to GetIndex for the player mii, and all the data seems intact.

If the saved mii does exist in the database retrieved via the `Get` method (matching createid), then the GetIndex call instead has a mii that is all 0s, indicating that the mii was blanked out for some reason. It crashes later on, probably because it tries to draw a null mii for the menu.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Mii/DatabaseImpl.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Mii/DatabaseImpl.cs
@@ -290,7 +290,7 @@ namespace Ryujinx.HLE.HOS.Services.Mii
         {
             coreData = new CoreData();
 
-            if (charInfo.IsValid())
+            if (!charInfo.IsValid())
             {
                 return ResultCode.InvalidCharInfo;
             }
--- a/src/Ryujinx.HLE/HOS/Services/Mii/Types/Ver3StoreData.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Mii/Types/Ver3StoreData.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿using Ryujinx.Common.Memory;
+using Ryujinx.Common.Utilities;
+using System;
 using System.Runtime.InteropServices;
 
 namespace Ryujinx.HLE.HOS.Services.Mii.Types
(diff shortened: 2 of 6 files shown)


src/Ryujinx.HLE/HOS/Services/Mii/Types/Ver3StoreData.cs
```diff
@@ -8,9 +10,10 @@ struct Ver3StoreData
     {
         public const int Size = 0x60;
 
-        private byte _storage;
+        private Array64<byte> _storage;
+        private Array32<byte> _storage2;
 
-        public Span<byte> Storage => MemoryMarshal.CreateSpan(ref _storage, Size);
+        public Span<byte> Storage => SpanHelpers.CreateSpan(ref MemoryMarshal.GetReference(_storage.AsSpan()), Size);
```
gdkchan, line 16: I think adding a `Array96<T>` type is better than doing this.

src/Ryujinx.HLE/HOS/Services/Mii/DatabaseImpl.cs
```diff
@@ -290,7 +290,7 @@ public ResultCode ConvertCharInfoToCoreData(CharInfo charInfo, out CoreData core
         {
             coreData = new CoreData();
 
-            if (charInfo.IsValid())
+            if (!charInfo.IsValid())
```
marysaka, line 293: nice catch...

marysaka (approved):

gdkchan (approved):
lgtm, thanks!

PR #5768, opened 2023-10-06
Symbols.cs Get function return value fix

saldabain:
Fixes a regression from #5664 that causes the wrong value to be returned in the Get function for Symbols.cs when a result does not currently exist for an address.


gdkchan (approved):
lgtm, nice catch!

TSRBerry (approved):
Thank you!

PR #5780, opened 2023-10-08
Avalonia: Show aspect ratio popup options in status bar

ahmadtantowi:
Added button to show aspect ratio popup options in status bar, so we can easily see and select available aspect ratios.
Aspect ratio ddl


marysaka (approved):
lgtm thanks :+1:

TSRBerry (approved):
Awesome, thank you!

PR #5785, opened 2023-10-09
Replace ReaderWriterLock with ReaderWriterLockSlim

sunshineinabox:
Replaces Use of ReaderWriterLock with ReaderWriterLockSlim, ReaderWriterLock is now deprecated in favor of ReaderWriterLockSlim. In cases of many readers and few writer it is preferred and in other cases a simple lock should be used. UpgradeableReadLock also is a thread blocking state so was replaced with simple switching between reading and writing. 

Supersedes #5493

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
+++ b/src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
@@ -550,9 +550,11 @@ namespace Ryujinx.Graphics.Vulkan
         {
             bool wasInReadLock = _flushLock.IsReadLockHeld;
 
-            // Try to get into a write lock if not already.
+            // Try to get into a read lock if not already.
             if (!wasInReadLock)
+            {
                 _flushLock.EnterReadLock();
+            }
 
             try
             {
@@ -585,7 +587,9 @@ namespace Ryujinx.Graphics.Vulkan
                     {
                         _flushLock.ExitWriteLock();
                         if (wasInReadLock)
+                        {
                             _flushLock.EnterReadLock(); // Re-enter read lock if we were originally in it.
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


gdkchan (approved):
lgtm, thanks!

marysaka (approved):

src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
```diff
@@ -548,42 +548,66 @@ private void ClearFlushFence()
 
         private void WaitForFlushFence()
         {
-            // Assumes the _flushLock is held as reader, returns in same state.
+            bool wasInReadLock = _flushLock.IsReadLockHeld;
 
-            if (_flushFence != null)
+            // Try to get into a read lock if not already.
+            if (!wasInReadLock)
```
gdkchan, line 554: This function changed quite a bit. The old code assumed the read lock was being held when this function is called. Why the new one no longer makes that assumption (and acquires the read lock if its not already held)?

src/Ryujinx.Graphics.Vulkan/BufferHolder.cs
```diff
@@ -548,42 +548,66 @@ private void ClearFlushFence()
...
             {
-                // If storage has changed, make sure the fence has been reached so that the data is in place.
-
-                var cookie = _flushLock.UpgradeToWriterLock(Timeout.Infinite);
+                _flushLock.EnterReadLock();
+            }
 
+            try
+            {
                 if (_flushFence != null)
```
gdkchan, line 561: Could be better to change this to
```cs
if (_flushFence == null)
{
    return;
}
```
To avoid some nesting.

AcK77 (approved):

PR #5791, opened 2023-10-10
infra: Reformat README.md & add new generic Mako workflow

TSRBerry:
This PR includes a bunch of smaller changes and adds the new Mako workflow as previously announced in Ryujinx/Ryujinx-Mako#3.

The new Mako workflow:

- This workflow replaces the previous Ryujinx-Mako steps and should allow us to change Mako's behavior without changing any files here.
- It receives all the necessary information from the event that triggered the workflow, so we can execute the appropriate tasks for them. All of this is handled by the `exec-ryujinx-tasks` command.
  Currently only the `update-reviewers` task will be executed, since I want to make sure everything works before adding more stuff on top of it.
- Although there are a lot of events that trigger this workflow, most runs shouldn't take longer than ~30 seconds, so hopefully we don't run into any issues because of that.
- We should be able to use this workflow without any issues in our other projects as well.

Unrelated smaller adjustments:

- `.editorconfig` needed to be changed to account for the files where we indent with 2 spaces instead of 4.
- ~~The .NET SDK version in our current `global.json` is not able to compile Ryujinx anymore, so it was updated.~~
- Our `README.md` contained a bunch of small formatting/styling issues which I addressed.
  I also added a small section to link to our new documentation and fixed a few links, but didn't make any larger changes to the wording.
  Note: Most of these formatting changes don't affect the rendered markdown file.
  Currently we link to the MIT license twice for some reason, but I'm not sure if this is the right place to address that.

---

Depends on Ryujinx/Ryujinx-Mako#3

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,21 @@
-
 <h1 align="center">
   <br>
-  <a href="https://ryujinx.org/"><img src="https://i.imgur.com/WcCj6Rt.png" alt="Ryujinx" width="150"></a>
+  <a href="https://ryujinx.org/"><img src="https://raw.githubusercontent.com/Ryujinx/Ryujinx/master/distribution/misc/Logo.svg" alt="Ryujinx" width="150"></a>
   <br>
   <b>Ryujinx</b>
   <br>
   <sub><sup><b>(REE-YOU-JINX)</b></sup></sub>
   <br>
-
 </h1>
 
 <p align="center">
-       Ryujinx is an open-source Nintendo Switch emulator, created by gdkchan, written in C#.
-       This emulator aims at providing excellent accuracy and performance, a user-friendly interface and consistent builds.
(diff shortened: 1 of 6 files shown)


TSRBerry:
I didn't know we had so many files using the wrong encoding. Guess, I'll fix that as well.

README.md
```diff
@@ -34,87 +34,111 @@
     <img src="https://raw.githubusercontent.com/Ryujinx/Ryujinx-Website/master/public/assets/images/shell.png">
 </p>
 
-<h5 align="center">
+## Compatibility
 
-</h5>
+As of April 2023, Ryujinx has been tested on approximately 4,050 titles;
+over 4,000 boot past menus and into gameplay, with roughly 3,400 of those being considered playable.
```
AcK77, line 40:
```suggestion
As of October 2023, Ryujinx has been tested on approximately 4,200 titles;
over 4,150 boot past menus and into gameplay, with roughly 3,500 of those being considered playable.
```

marysaka (approved):

AcK77 (approved):

PR #5792, opened 2023-10-10
GPU: Add fallback when textureGatherOffsets is not supported

riperiperi:
This PR adds a fallback for GPUs or APIs that don't support an equivalent to the method `textureGatherOffsets`, where each of the 4 gathered texels has an individual offset. This is done by reusing the existing code to handle non-const offsets for texture instructions, though it has also been corrected as there were a few implementation issues.

MoltenVK reports support for this capability, and it didn't error when we initially released the MacOS build, but that has since changed. MVK still reports support, but spirv-cross has been fixed in a way that it _attempts_ to use this capability, and the metal compiler errors since it doesn't exist.

Some other fixes:
- textureGatherOffsets emulation has been changed significantly. It now uses 4 texture sample instructions (not gather), calculates a base texel (i=0 j=0) and adds the offsets onto it before converting into a tex coord. The final result is offset into a texel center, so it shouldn't be subject to interpolation, though this isn't perfect and could have some error with floating point formats with linear sampling. It is subject to texture wrap mode as it should be, which is why texelFetch was not used.
  - Maybe gather should be used here with component `w` (i=0, j=0), though this multiplies number of texels fetched by 4... The way it was doing this before _was_ wrong, but doing it right would avoid issues with texel center precision and interpolation enabled.
  - Also uses the base texture size, as texture gather only accesses base level without an AMD specific extension.
- textureGatherOffset (singular) fallback now performs textureGather with the offset applied to the coords, rather than the slower fallback where each texel is fetched individually.

Xenoblade Chronicles DE now renders on the latest MoltenVK. There's still a bunch of disappearing stuff since the occlusion queries are busted, though.

### DIFF
--- a/src/Ryujinx.Graphics.Shader/Translation/Transforms/TexturePass.cs
+++ b/src/Ryujinx.Graphics.Shader/Translation/Transforms/TexturePass.cs
@@ -449,38 +449,6 @@ namespace Ryujinx.Graphics.Shader.Translation.Transforms
 
                 int destIndex = 0;
 
-                Operand[] sourceTexels = new Operand[coordsCount];
-
-                for (int index = 0; index < coordsCount; index++)
-                {
-                    Operand source = sources[coordsIndex + index];
-
-                    Operand sourceUnscaled = Local();
-
-                    node.List.AddBefore(node, new Operation(
-                        Instruction.Multiply | Instruction.FP32,
-                        sourceUnscaled,
-                        source,
-                        GenerateI2f(node, texSizes[index])));
-
-                    Operand floatSourceTexel = Local();
-
-                    node.List.AddBefore(node, new Operation(
-                        Instruction.Subtract | Instruction.FP32,
-                        floatSourceTexel,
-                        sourceUnscaled,
-                        ConstF(0.5f)));
-
-                    Operand floorSourceTexel = Local();
-
-                    node.List.AddBefore(node, new Operation(
-                        Instruction.Floor | Instruction.FP32,
-                        floorSourceTexel,
-                        floatSourceTexel));
-
-                    sourceTexels[index] = GenerateF2i(node, floorSourceTexel);
-                }
-
                 for (int compIndex = 0; compIndex < 4; compIndex++)
                 {
                     if (((texOp.Index >> compIndex) & 1) == 0)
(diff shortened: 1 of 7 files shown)


riperiperi:
I pushed a commit that does this with 4 gathers to avoid possible issues due to texCoord precision errors causing accidental interpolation. Not sure how much slower it will be in xenoblade, but it does avoid the issue mentioned in the main message.

src/Ryujinx.Graphics.Shader/Translation/Transforms/TexturePass.cs
```diff
@@ -502,7 +507,8 @@ private static LinkedListNode<INode> InsertConstOffsets(LinkedListNode<INode> no
                 }
                 else
                 {
-                    Operand[] texSizes = InsertTextureLod(node, texOp, lodSources, bindlessHandle, coordsCount, stage);
+                    Operand[] texSizes = isGather ? InsertTextureBaseSize(node, texOp, bindlessHandle, coordsCount) :
```
gdkchan, line 510:
```suggestion
                    Operand[] texSizes = isGather
                        ? InsertTextureBaseSize(node, texOp, bindlessHandle, coordsCount)
                        : InsertTextureLod(node, texOp, lodSources, bindlessHandle, coordsCount, stage);
```

src/Ryujinx.Graphics.Shader/Translation/Transforms/TexturePass.cs
```diff
@@ -549,6 +555,43 @@ private static LinkedListNode<INode> InsertConstOffsets(LinkedListNode<INode> no
...
+
+            for (int index = 0; index < coordsCount; index++)
+            {
+                texSizes[index] = Local();
+
+                Operand[] texSizeSources;
+
+                if (bindlessHandle != null)
+                {
+                    texSizeSources = new Operand[] { bindlessHandle, ConstF(0.0f) };
```
gdkchan, line 574: nit: Should probably use `Const(0)` here, since LOD is an integer, not float parameter? Doesn't really matter since 0.0f and 0 are encoded as the same value, but it's nice to make the type clear on the constant.

gdkchan (approved):
lgtm, thanks. Even on current macOS build it might show some improvement, since apparently it would just ignore the offsets before and use a regular texture sample operation.

marysaka (approved):

tgcordell:
I just had a crash on MacOS with this merged, and I don't fully understand the log files, but theres a bunch of references to texture details in mine. Wanted to tag it here in case it's useful. https://github.com/Ryujinx/Ryujinx/issues/5821

PR #5793, opened 2023-10-11
editorconfig: Set default encoding to UTF-8

TSRBerry:
Moved this change here from #5791


gdkchan:
Diffs shows a "change" on the first line of every file. I'm guessing it inserted a BOM and it detects that as a change.
It would be nice to see an explanation of why we need this change, and what was the default encoding before this change.

TSRBerry:
This change is not exactly necessary, but dotnet format fails after specifying the default encoding in `.editorconfig`, so I decided to fix it.

Before we didn't have a default encoding, but most editors default to UTF-8 which is why most files aren't affected by this change. Apparently Visual Studio defaults to UTF-8-BOM, so that's probably why some files use that instead.

This PR really just sets the same encoding for all files, although in reality nothing really changes. The only bonus is that newly added files will always have the same encoding as the existing ones.

marysaka (approved):
:tada: no more BOM

PR #5801, opened 2023-10-14
Enable copy between MS and non-MS textures with different height

gdkchan:
This change allows Perky Little Things to render again. The game uses 8x multisampling on a 1920x1088 texture (really 1920x1080), later it tries to sample it as a 7680x2160 texture. While sampling multisampled textures as non-multisampled is already supported, this case was not working because the height doesn't match (1088 * 2 = 2076). This change enables copy compatibility for this specific case (as long the *aligned* height matches, at least).

image

According to Muffin it's a regression from #4662, but I'm not sure how that is possible.


riperiperi (approved):
Nice catch.

AcK77 (approved):

PR #5802, opened 2023-10-14
Horizon: Implement arp:r and arp:w services

AcK77:
This PR implement `arp:r` and `arp:w` service to make other services migration easier. `prepo` is updated to reflect the changes.
Everything is checked by RE from latest firmware (17.0.0) :

- `Issue` IPC aren't implemented for now (we don't use it).
- Instances have to be deregistered/deleted when an application isn't running anymore (exited or idle).
- Buffer support code in `ServiceUtil` ~~and `LmApi` were~~ is made by @gdkchan (Thanks to him).
- Current arp HOS isn't removed (Will be done soon).
- Some cleanup here and there while browsing the codebase.


src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public Reader(ApplicationInstanceManager applicationInstanceManager)
+        {
+            _applicationInstanceManager = applicationInstanceManager;
+        }
+
+        [CmifCommand(0)]
+        public Result GetApplicationLaunchProperty(out ApplicationLaunchProperty applicationLaunchProperty, ulong applicationInstanceId)
+        {
+            // NOTE: Returns InvalidArgument if applicationLaunchProperty is null, doesn't occur in our case.
```
gdkchan, line 23: I don't think those comments are necessary (I didn't include them on NGC for example).

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+                return ArpResult.InvalidInstanceId;
+            }
+
+            applicationLaunchProperty = (ApplicationLaunchProperty)_applicationInstanceManager.Entries[applicationInstanceId].LaunchProperty;
+
+            return Result.Success;
+        }
+
+        [CmifCommand(1)]
+        public Result GetApplicationControlProperty([Buffer(HipcBufferFlags.Out | HipcBufferFlags.MapAlias | HipcBufferFlags.FixedSize, 0x4000)] out ApplicationControlProperty applicationControlProperty, ulong applicationInstanceId)
```
gdkchan, line 39:
```suggestion
        public Result GetApplicationControlProperty([Buffer(HipcBufferFlags.Out | HipcBufferFlags.MapAlias, 0x4000)] out ApplicationControlProperty applicationControlProperty, ulong applicationInstanceId)
```
But no harm in keeping it either.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        [CmifCommand(0)]
+        public Result GetApplicationLaunchProperty(out ApplicationLaunchProperty applicationLaunchProperty, ulong applicationInstanceId)
+        {
+            // NOTE: Returns InvalidArgument if applicationLaunchProperty is null, doesn't occur in our case.
+            //       Returns InvalidPointer if applicationInstanceId is null, doesn't occur in our case.
+
+            if (_applicationInstanceManager.Entries[applicationInstanceId] == null)
+            {
+                applicationLaunchProperty = new ApplicationLaunchProperty();
```
gdkchan, line 28:
```suggestion
                applicationLaunchProperty = default;
```
For consistency with what is done on other places.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                return ArpResult.InvalidInstanceId;
+            }
+
+            applicationLaunchProperty = (ApplicationLaunchProperty)_applicationInstanceManager.Entries[applicationInstanceId].LaunchProperty;
```
gdkchan, line 33:
```suggestion
            applicationLaunchProperty = _applicationInstanceManager.Entries[applicationInstanceId].LaunchProperty.Value;
```

gdkchan: No need to check if `LaunchProperty.HasValue` here (and return an error if it doesn't)?

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
gdkchan: This and a few of the comments below also applies to other functions here, I won't repeat them.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            return Result.Success;
+        }
+
+        [CmifCommand(5)]
+        public Result ListApplicationInstanceId(out int counter, [Buffer(HipcBufferFlags.Out | HipcBufferFlags.MapAlias)] Span<ulong> applicationInstanceIdList)
+        {
+            // NOTE: Returns InvalidArgument if counter is null, doesn't occur in our case.
+
+            counter = 0;
```
gdkchan, line 114: I think it would be better to call it `count` rather than `counter`.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (pid == 0)
+            {
+                return ArpResult.InvalidPid;
+            }
+
+            for (int i = 0; i < _applicationInstanceManager.Entries.Length; i++)
+            {
+                if (_applicationInstanceManager.Entries[i] != null && _applicationInstanceManager.Entries[i].Pid == pid)
+                {
+                    applicationInstanceId = (ulong)i;
```
gdkchan, line 90: This seems odd to me, shouldn't it store the instance ID on the entry? From what I recall, the service uses an ever incrementing counter as instance ID, no?

src/Ryujinx.Horizon/Arp/Ipc/Updater.cs
```diff
@@ -0,0 +1,61 @@
...
+        public Updater(ApplicationInstanceManager applicationInstanceManager, ulong applicationInstanceId)
+        {
+            _applicationInstanceManager = applicationInstanceManager;
+            _applicationInstanceId = applicationInstanceId;
+        }
+
+        [CmifCommand(0)]
+        public Result Issue()
+        {
+            throw new NotImplementedException();
```
gdkchan, line 24: Shouldn't that be implemented?

src/Ryujinx.Horizon/Arp/Ipc/Writer.cs
```diff
@@ -0,0 +1,85 @@
...
+            return Result.Success;
+        }
+
+        [CmifCommand(3)]
+        public Result AcquireApplicationCertificateUpdater(out IUpdater updater, ulong applicationInstanceId)
+        {
+            // NOTE: Returns InvalidPointer if applicationInstanceId is null, doesn't occur in our case.
+            //       Returns AllocationFailed if it can't create the object, doesn't occur in our case.
+
+            updater = new Updater(_applicationInstanceManager, applicationInstanceId);
```
gdkchan, line 80: Are you sure this is correct? It seems to do the same thing that the function above does.

src/Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
```diff
@@ -210,6 +213,18 @@ private static Result ProcessPlayReport(PlayReportKind playReportKind, ReadOnlyS
                 builder.AppendLine($" ApplicationId: {applicationId}");
             }
 
+            using var arpApi = new ArpApi(_heapAllocator);
```
gdkchan, line 216: I think it would be better to create his on `PrepoIpcServer` or `PrepoMain`, and pass it on the constructor (instead of the `HeapAllocator`) to avoid instantiating it all the time.

src/Ryujinx.Horizon/Sdk/Arp/Detail/ApplicationInstanceManager.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Ryujinx.Horizon.Sdk.Arp.Detail
+{
+    class ApplicationInstanceManager : IDisposable
+    {
+        private int _disposalState;
+
+        public SystemEventType SystemEvent;
+        public int EventHandle;
+
+        public ApplicationInstance[] Entries = new ApplicationInstance[2];
```
gdkchan, line 14:
```suggestion
        public readonly ApplicationInstance[] Entries = new ApplicationInstance[2];
```

src/Ryujinx.Horizon/Sdk/Lm/LmApi.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Ryujinx.Horizon.Common;
+using Ryujinx.Horizon.Sdk.Sf.Cmif;
+using Ryujinx.Horizon.Sdk.Sf.Hipc;
+using System;
+
+namespace Ryujinx.Horizon.Sdk.Sm
+{
+    class LmApi : IDisposable
```
gdkchan, line 8: I think there is no need to include this here, I wrote this just as an example, I didn't actually intend for it to be used.

src/Ryujinx.Horizon/ServiceTable.cs
```diff
@@ -23,6 +25,9 @@ public class ServiceTable
 
         private readonly ManualResetEvent _servicesReadyEvent = new(false);
 
+        public IReader ArpReader { get; set; }
```
gdkchan, line 28:
```suggestion
        public IReader ArpReader { get; internal set; }
```

src/Ryujinx.Horizon/ServiceTable.cs
```diff
@@ -23,6 +25,9 @@ public class ServiceTable
...
+        public IWriter ArpWriter { get; set; }
```
gdkchan, line 29:
```suggestion
        public IWriter ArpWriter { get; internal set; }
```

src/Ryujinx.Horizon/Arp/Ipc/Registrar.cs
```diff
@@ -0,0 +1,52 @@
...
+                return ArpResult.DataAlreadyBound;
+            }
+
+            _applicationInstance.LaunchProperty = applicationLaunchProperty;
+
+            return Result.Success;
+        }
+
+        [CmifCommand(2)]
+        public Result SetApplicationControlProperty([Buffer(HipcBufferFlags.In | HipcBufferFlags.MapAlias | HipcBufferFlags.FixedSize, 0x4000)] ApplicationControlProperty applicationControlProperty)
```
gdkchan, line 40:
```suggestion
        public Result SetApplicationControlProperty([Buffer(HipcBufferFlags.In | HipcBufferFlags.MapAlias | HipcBufferFlags.FixedSize, 0x4000)] in ApplicationControlProperty applicationControlProperty)
```
Should avoid struct copies, this struct is pretty large.

src/Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
AcK77: It's not "really" possible, since `PrepoIpcServer` is based of `ServerManager` that include to change the ServerManager constructor to handle `ArpApi` which isn't the case for all services.

src/Ryujinx.Horizon/Sdk/Lm/LmApi.cs
AcK77: I think it could be better to keep it for the example of a buffer in an Api. For what it cost to keep it.

src/Ryujinx.Horizon/Arp/Ipc/Updater.cs
AcK77: `arp:*` call are never done by application, only services (as far as I know), we currently don't call it in any case, for now it could be not implemented and can be fixed later if needed.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
AcK77: Wdym by "on the entry"? It's not really what the service currently does, but the logic is the same here, I did that to simplify things in C#.

src/Ryujinx.Horizon/Arp/Ipc/Writer.cs
AcK77: Nice catch!

src/Ryujinx.Horizon/Sdk/Lm/LmApi.cs
gdkchan: It is not "usable" in the current state (as in you would need to write all the message headers yourself too). It also does not contain the initialization code, I didn't really put effort into making it great since I never expected it to be actually used. So I'd just remove it.

src/Ryujinx.Horizon/Prepo/Ipc/PrepoService.cs
gdkchan: You don't need to change `ServerManager`, you can change only `PrepoIpcServer` constructor. Just because it inherits from `ServerManager`, doesn't mean that the constructors must match.

src/Ryujinx.Horizon/Arp/Ipc/Reader.cs
gdkchan: With "entry" I meant on `ApplicationInstance`. It should have an `ApplicationInstanceId` field. Then any calls that receives an `ApplicationInstanceId` would check both entries, and use the one where the ID matches. If none of them matches, it would just return an error.

This difference effectively means that the service always returns an unique ID for applications, while your implementation will only return 0 or 1. Which means that different applications (that have been registered on the past, then unregistered) will have the same IDs. This is a pretty significant difference, and some code might depend on it. I think it is better to do things like the original service (and it would be also pretty simple to do).

src/Ryujinx.Horizon/Arp/Ipc/Updater.cs
gdkchan: Doesn't it just assign the `ApplicationInstanceId`? If so it should be pretty simple to implement.

gdkchan:
This currently crashes when I try to launch a game:
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NotImplementedException: arp:r
   at Ryujinx.HLE.HOS.Services.Sm.IUserInterface.GetService(ServiceCtx context) in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Services\Sm\IUserInterface.cs:line 113
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at Ryujinx.HLE.HOS.Services.IpcService.CallCmifMethod(ServiceCtx context) in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Services\IpcService.cs:line 126
   at Ryujinx.HLE.HOS.Services.ServerBase.Process(Int32 serverSessionHandle, UInt64 recvListAddr) in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Services\ServerBase.cs:line 343
   at Ryujinx.HLE.HOS.Services.ServerBase.ServerLoop() in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Services\ServerBase.cs:line 247
   at Ryujinx.HLE.HOS.Services.ServerBase.Main() in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Services\ServerBase.cs:line 168
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart() in C:\Projects\Ryujinx\src\Ryujinx.HLE\HOS\Kernel\Threading\KThread.cs:line 1256
```

gdkchan (approved):
lgtm, as discussed on Discord I'm happy to approve this as-is  to unblock the remaining work on the services. Tested it with a few games and it seems to work.

marysaka (approved):

PR #5807, opened 2023-10-15
Allow skipping draws with broken pipeline variants on Vulkan

### LINKED ISSUE #5790: Fire Emblem: Three Houses  Using 'Divine Pulse' ability cause 100% crash on MacOS Sonoma(14.0) M1 Max 64GB.
Using 'Divine Pulse' ability cause 100% crash on MacOS Sonoma(14.0) M1 Max 64GB.
Including ryujinx version 1.1.1036~1.1.1049.
When crash happend no ryujinx logs were dumped, only got MacOS error log.

Thread 60 Crashed:
0 libsystem_kernel.dylib       0x1843cb11c __pthread_kill + 8
1   libsystem_pthread.dylib   0x184402cc0 pthread_kill + 288
2   libsystem_c.dylib              0x184312a50 abort + 180
3   Ryujinx                       	       0x103500000 PROCAbort + 64
4   Ryujinx                       	       0x1034fff10 PROCEndProcess(void*, unsigned int, int) + 388
5   Ryujinx                       	       0x103143d94 UnwindManagedExceptionPass1(PAL_SEHException&, _CONTEXT*) + 876
6   Ryujinx                       	       0x103143e78 DispatchManagedException(PAL_SEHException&, bool) + 68
7   Ryujinx                       	       0x1030a6e9c IL_Throw(Object*) + 536
8   ???                           	       0x10b5fa914 ???
9   ???                           	       0x10c5ca518 ???
10  ???                           	       0x10c5c9148 ???
11  ???                           	       0x10c804a78 ???
12  ???                           	       0x10c804888 ???
13  ???                           	       0x10c5759b8 ???
14  ???                           	       0x10b434180 ???
15  ???                           	       0x10ab24030 ???
16  ???                           	       0x1074bc4bc ???
17  ???                           	       0x1074bc3a8 ???
18  ???                           	       0x1074bc1a0 ???
19  Ryujinx                       	       0x1031cdff4 CallDescrWorkerInternal + 132
20  Ryujinx                       	       0x1030438a0 DispatchCallSimple(unsigned long*, unsigned int, unsigned long,unsigned int) + 272
21  Ryujinx                       	       0x103057fbc ThreadNative::KickOffThread_Worker(void*) + 148
22  Ryujinx                       	       0x103011cdc ManagedThreadBase_DispatchOuter(ManagedThreadCallState*) + 260
23  Ryujinx                       	       0x103012248 ManagedThreadBase::KickOff(void (*)(void*), void*) + 32
24  Ryujinx                       	       0x103058094 ThreadNative::KickOffThread(void*) + 172
25  Ryujinx                       	       0x1035045ec CorUnix::CPalThread::ThreadEntry(void*) + 380
26  libsystem_pthread.dylib       	       0x184403034 _pthread_start + 136
27  libsystem_pthread.dylib       	       0x1843fde3c thread_start + 8

[crash_log.txt](https://github.com/Ryujinx/Ryujinx-Games-List/files/12837566/crash_log.txt)
[Ryujinx_1.1.1049_2023-10-08_17-37-01.log](https://github.com/Ryujinx/Ryujinx-Games-List/files/12840320/Ryujinx_1.1.1049_2023-10-08_17-37-01.log)

_Originally posted by @hakcumx in https://github.com/Ryujinx/Ryujinx-Games-List/issues/230#issuecomment-1751683452_

gdkchan:
Currently we skip draws when the background compilation of the shaders on Vulkan failed. This is detected with the result code from the pipeline creation call. If an is error returned, we assume that the host failed to compile the shader that the emulator gave. While this generally works, there is an odd situation on macOS with MoltenVK where *some* pipeline variants may compile successfully, and others may fail. This means that the emulator can still crash in some cases due to shader compilation failures.

This PR allows draws to be skipped when *any* pipeline variant creation fails, not just the background one. This is only done for graphics shaders since compute generally doesn't have pipeline variants (since the only pipeline state on compute is the shader itself). This also includes minor improvements to try preventing any performance regression from additional checks. The `!_program.IsLinked` has been moved to `CreatePipeline` to avoid checking it all the time. `RecreatePipelineIfNeeded` has been split in two, one for graphics and other for compute. Most of what it was doing is not really necessary for compute (like updating dynamic state, vertex/index/transform feedback buffers).

Fixes #5790, although it does not fix the underlying issue (the compilation error). The only error that the compiler produces here is "Compiler encountered an internal error.", which does not really give much indication of what is wrong. I was able to track down the specific pipeline state that makes it produce that error, it was `ColorBlendAttachmentState`, on the background pipeline it was disabled, on the new one it is enabled. There is a single color attachment with VkFormat B10G11R11UfloatPack32.
Captura de Tela 2023-10-15 às 00 20 09
Does not render 100% due to the skipped draw, but it is better than the game crashing.

Testing is welcome, also to ensure that nothing broke on platforms other than macOS.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -749,12 +749,15 @@ namespace Ryujinx.Graphics.Vulkan
             _vertexBufferUpdater.Commit(Cbs);
         }
 
+
+#pragma warning disable CA1822 // Mark member as static
         public void SetAlphaTest(bool enable, float reference, CompareOp op)
         {
             // This is currently handled using shader specialization, as Vulkan does not support alpha test.
             // In the future, we may want to use this to write the reference value into the support buffer,
             // to avoid creating one version of the shader per reference value used.
         }
+#pragma warning restore CA1822
 
         public void SetBlendState(AdvancedBlendDescriptor blend)
         {
--- a/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
+++ b/src/Ryujinx.Graphics.Vulkan/PipelineBase.cs
@@ -374,7 +374,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         public void Draw(int vertexCount, int instanceCount, int firstVertex, int firstInstance)
         {
-            if (!_program.IsLinked || vertexCount == 0)
+            if (vertexCount == 0)
             {
                 return;
             }
@@ -442,7 +442,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         public void DrawIndexed(int indexCount, int instanceCount, int firstIndex, int firstVertex, int firstInstance)
         {
-            if (!_program.IsLinked || indexCount == 0)
+            if (indexCount == 0)
             {
                 return;
             }
@@ -486,11 +486,6 @@ namespace Ryujinx.Graphics.Vulkan
 
         public void DrawIndexedIndirect(BufferRange indirectBuffer)
         {
-            if (!_program.IsLinked)
-            {
-                return;
-            }
-
             var buffer = Gd.BufferManager
                 .GetBuffer(CommandBuffer, indirectBuffer.Handle, indirectBuffer.Offset, indirectBuffer.Size, false)
                 .Get(Cbs, indirectBuffer.Offset, indirectBuffer.Size).Value;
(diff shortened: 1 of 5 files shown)


marysaka (approved):
Really not a big fan of this, but it will do for now I guess...

gdkchan:
> Really not a big fan of this, but it will do for now I guess...

I guess the title makes it sound like I'm introducing a hack on Vulkan to work around MoltenVK/Metal issues, but this just extends something that is already being done. We already skip draws when shader compilation fails on OpenGL, and do more or less the same on Vulkan, but only when the first, background pipeline creation fails. This is just extending it to work on all pipelines. I would prefer to fix the underlying issue, but I think there is some bug on their compiler since it just reports an "internal compiler error", it might be possible to work around it, but it would require either a lot of trial and error or reversing their compiler (maybe not that hard since it's LLVM based, but still).

psnluiz:
This fixed Fire Emblem Three Houses crashing for me.

riperiperi (approved):
Looks good to me. What was the expected extra cost of adding this? I can't seem to find anything that would cost too much more.

gdkchan:
> Looks good to me. What was the expected extra cost of adding this? I can't seem to find anything that would cost too much more.

There are a few extra checks needed for updating pipelines, but IIRC I tried to make up for it (for example by moving the shader `IsLinked` check).

PR #5818, opened 2023-10-19
Remove SoundIO

IsaacMarovitz:
**Why?**
This is likely to be controversial, but I have reasons:
- We should aim to have only one backend that works across every device (SDL2)
- Having two fallbacks is unnecessary
- SoundIO has poor compatibility, it's crashing, and lack of functionality can be a point of confusion for users
- Reducing to two backends (hopefully one eventually) will reduce technical debt and make it easier to isolate issues


TSRBerry:
Meh, if SoundIO was actively blocking something else I'd be in favor of removing it, but as of now that's not the case.

Aside from that I don't think that this qualifies as technical debt, since we don't have to work around it and it's not preventing us from building new things or adjusting how audio works right now.

@marysaka if you feel different about this please reopen this PR, but for now I'll close this.

gdkchan:
AFAIK there are some games that only sound properly with SoundIO still.

MetrosexualGarbodor:
I wouldn't say outright remove it, but disable it until the "No valid channel configuration" crash is fixed.

PR #5838, opened 2023-10-22
macOS: Use user-friendly macOS version string

jcm93:
### Description
Logs currently output the Darwin kernel version string. While correct, this is a version string that is never exposed to the user. It would be more useful to output the user-facing macOS version string (i.e. macOS 14.0 for Sonoma, versus Darwin 23.0). This will make debugging at a glance much easier, as most Mac developers do not hold the kernel versioning system in their heads simultaneously with macOS release versions.

It is fairly remote that the much more verbose version string will contain pertinent information that the common version string does not.

Before: Screenshot 2023-10-22 at 4 54 40 PM
After: 
Screenshot 2023-10-24 at 6 58 14 AM
(screenshot reflects changes from [639d641](https://github.com/Ryujinx/Ryujinx/commit/639d64115915f040d82f367959b9b7b2284cd68d))

### DIFF
--- a/src/Ryujinx.Common/SystemInfo/MacOSSystemInfo.cs
+++ b/src/Ryujinx.Common/SystemInfo/MacOSSystemInfo.cs
@@ -14,6 +14,8 @@ namespace Ryujinx.Common.SystemInfo
         {
             string cpuName = GetCpuidCpuName();
 
+            OsDescription = $"macOS {Environment.OSVersion.Version} ({RuntimeInformation.OSArchitecture})";
+
             if (cpuName == null && SysctlByName("machdep.cpu.brand_string", out cpuName) != 0)
             {
                 cpuName = "Unknown";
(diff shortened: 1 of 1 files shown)


marysaka (approved):

riperiperi (approved):
> as most Mac developers do not hold the kernel versioning system in their heads simultaneously with macOS release versions.

disappointing... 😔

jcm93:
Sorry for the late revision, I realized we also had convenient access to the build identifier string (i.e. 23A344 for current macOS) by querying ``sysctlbyname`` for ``"kern.osversion"``, so I added that to the output. This is the generally preferred convention these days for identifying specific builds with more granularity than the common version number. Will be useful for macOS betas, etc.

AcK77 (approved):

ultratiem:
> convenient access to the build identifier string (i.e. 23A344 for current macOS)

These are likely even more value at the coding level than the major build. Together they are ideal, especially over the kernel ID (TIL Darwin is at v23)

PR #5840, opened 2023-10-23
Fix the AOC manager using incorrect paths

Thealexbarney:
The content manager was trying to use "paths" of the form `xxxx.nca` instead of `/xxxx.nca` when opening a file from a file system


gdkchan (approved):
lgtm, thanks for the quick fix!

riperiperi (approved):

Issue #5827, opened 2023-10-21
[Bug] macOS - Application shortcut broken


nickoanastassiu:
### Description of the issue

The application shortcut that is generated is not working because the binary specified in the script is `Ryujinx.Ava` but the binary is actually called `Ryujinx`. Fixing this manually solves the issue.

A quick suggestion would also be to put the shortcuts in the Applications folder rather than on the Desktop.

### Reproduction steps

1. Create application shortcut
2. Try to open -> "crashes" on launch as the script cannot execute
3. Go to `Shortcut.app/Contents/MacOS/runner.sh` and fix binary name
4. Works as expected

### Log file

No logs necessary

### OS

macOS Sonoma 14.1 RC

### Ryujinx version

1.0.1059

Bentheminernz:
I have found a fix!!! In the runner.sh file which is located in the macos folder of the application, it has a path for `/Applications/Ryujinx.app/Contents/MacOS/Ryujinx.ava`. When I set it to `/Applications/Ryujinx.app/Contents/MacOS/Ryujinx` it fixes itself.

nickoanastassiu:
Yeah, that's exactly what I said.

> ... the binary specified in the script is `Ryujinx.Ava` but the binary is actually called `Ryujinx`. Fixing this manually solves the issue.

Bentheminernz:
Oops, also it runs so much quicker if you copy and paste the runner.sh code into an automaton application for some reason

NitroTears:
Alright, I'll work on this soon. I don't have a mac to test this on straight away, so I might need a hand in that later

nickoanastassiu:
Feel free to DM me anytime.

miche-egg564:
where do i edit the shortcut from

nickoanastassiu:
Locate the shortcut on your desktop. Right click and choose `Show package contents`. Go to `Contents/MacOS/runner.sh` and open it with TextEdit. Locate `Ryujinx.Ava` in the file and remove the `.Ava`

miche-egg564:
thanks

PR #5852, opened 2023-10-27
Create Desktop Shortcut fixes

### LINKED ISSUE #5839: [Bug] "Create application shortcut" feature creates partially broken shortcuts on Windows
### Description of the issue

This feature creates a shortcut target command where the ryujinx binary filepath is duplicated, eg:
```
E:\Emulators\Ryujinx\Ryujinx.exe E:\Emulators\Ryujinx\Ryujinx.exe "E:\Emulators\Ryujinx\Games\CHRONO CROSS THE RADICAL DREAMERS EDITION.nsp"
```
[These surprisingly do work](https://miro.medium.com/v2/resize:fit:1274/1*ON_d7DWgW8g8uu3EBntfNw.png) when run from the desktop but don't work when parsed by other applications such as Nvidia Game Experience, which I use to play these games remotely on my TV via Moonlight.
Here's an example of a shortcut I created myself, which does work:
```
E:\Emulators\Ryujinx\Ryujinx.exe --fullscreen "E:\Emulators\Ryujinx\Games\CHRONO CROSS THE RADICAL DREAMERS EDITION.nsp"
```

Obviously the fullscreen option is an addition of my own but I'd argue that if someone wants to create a shortcut for a game, they probably want it to go right into fullscreen too, so that'd be a nice addition, but now we're bordering on feature request [so I'll leave that up to you...](https://media.tenor.com/xijGlpXoHYIAAAAC/winkwinknudgenudge-wink.gif)

Thanks for the emu, works great otherwise!

### Reproduction steps

This is fairly self-explanatory. Right-click on a game in the UI and create a game shortcut then view the target of the .lnk file.

### Log file

N/A.

### OS

W10

### Ryujinx version

v1.1.1061

### Game version

ALL

### CPU

N/A

### GPU

N/A

### RAM

N/A

### List of applied mods

None

### Additional context?

_No response_

NitroTears:
This pull request aims to fix current issues with the Create Shortcut functionality, specifically this will close #5839 #5827.
This will also add the `--fullscreen` arg to the shortcuts so by default launching games with shortcuts automatically open Ryujinx in fullscreen mode.

I will need some testing with the MacOS issue (#5827), as I do not currently have a Mac to test with myself. Perhaps @nickoanastassiu could help with this?

### DIFF
--- a/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
+++ b/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
@@ -30,7 +30,7 @@ namespace Ryujinx.Ui.Common.Helper
             graphic.DrawImage(image, 0, 0, 128, 128);
             SaveBitmapAsIcon(bitmap, iconPath);
 
-            var shortcut = Shortcut.CreateShortcut(basePath, GetArgsString(basePath, applicationFilePath), iconPath, 0);
+            var shortcut = Shortcut.CreateShortcut(basePath, GetArgsString(applicationFilePath), iconPath, 0);
             shortcut.StringData.NameString = cleanedAppName;
             shortcut.WriteToFile(Path.Combine(desktopPath, cleanedAppName + ".lnk"));
         }
@@ -46,7 +46,7 @@ namespace Ryujinx.Ui.Common.Helper
             image.SaveAsPng(iconPath);
 
             using StreamWriter outputFile = new(Path.Combine(desktopPath, cleanedAppName + ".desktop"));
-            outputFile.Write(desktopFile, cleanedAppName, iconPath, GetArgsString(basePath, applicationFilePath));
+            outputFile.Write(desktopFile, cleanedAppName, iconPath, $"{basePath} {GetArgsString(applicationFilePath)}");
         }
 
         [SupportedOSPlatform("macos")]
@@ -69,7 +69,7 @@ namespace Ryujinx.Ui.Common.Helper
             using StreamWriter scriptFile = new(scriptPath);
 
             scriptFile.WriteLine("#!/bin/sh");
-            scriptFile.WriteLine(GetArgsString(basePath, appFilePath));
+            scriptFile.WriteLine($"{basePath} {GetArgsString(appFilePath)}");
 
             // Set execute permission
             FileInfo fileInfo = new(scriptPath);
@@ -125,7 +125,7 @@ namespace Ryujinx.Ui.Common.Helper
             throw new NotImplementedException("Shortcut support has not been implemented yet for this OS.");
         }
 
-        private static string GetArgsString(string basePath, string appFilePath)
+        private static string GetArgsString(string appFilePath)
         {
             // args are first defined as a list, for easier adjustments in the future
             var argsList = new List<string>();
@@ -137,7 +137,6 @@ namespace Ryujinx.Ui.Common.Helper
             }
 
             argsList.Add($"\"{appFilePath}\"");
-            argsList.Add($"--fullscreen");
 
             return String.Join(" ", argsList);
         }
--- a/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
+++ b/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
@@ -55,7 +55,7 @@ namespace Ryujinx.Ui.Common.Helper
             string basePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx");
             var plistFile = EmbeddedResources.ReadAllText("Ryujinx.Ui.Common/shortcut-template.plist");
             // Macos .App folder
-            string contentFolderPath = Path.Combine(desktopPath, cleanedAppName + ".app", "Contents");
+            string contentFolderPath = Path.Combine("/Applications", cleanedAppName + ".app", "Contents");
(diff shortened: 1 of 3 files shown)


src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
```diff
@@ -140,7 +137,7 @@ private static string GetArgsString(string basePath, string appFilePath)
             }
 
             argsList.Add($"\"{appFilePath}\"");
-
+            argsList.Add($"--fullscreen");
```
TSRBerry, line 140:
```suggestion
```

I prefer using Ryujinx settings instead of overriding them for shortcuts

src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
```diff
@@ -128,10 +128,7 @@ public static void CreateAppShortcut(string applicationFilePath, string applicat
         private static string GetArgsString(string basePath, string appFilePath)
         {
             // args are first defined as a list, for easier adjustments in the future
-            var argsList = new List<string>
-            {
-                basePath,
-            };
+            var argsList = new List<string>();
```
TSRBerry, line 131: How do you want to launch Ryujinx without adding the path to the binary?
```suggestion
            var argsList = new List<string>
            {
                basePath,
            };
```

TSRBerry (changes requested):
Sorry we missed this before. Lgtm except for the comments!

TSRBerry:
Fixes #5827

nickoanastassiu:
Does not fix #5827. Here's a before and after of the runner.sh file:

#### Before
`/Applications/Ryujinx.app/Contents/MacOS/Ryujinx.Ava "{Path to game file}"`
#### After
`"{Path to game file}" --fullscreen`

Removing the executable path does not fix it (the shell cannot simply open the file, it needs the executable to open it with and then the game path as an argument, just like on other platforms). The problem is that the name of the executable isn't `Ryujinx.Ava` but just `Ryujinx` (see attachment):
Screenshot 2023-10-27 at 15 10 37

This issue can be resolved by removing the `.Ava` part of the executable's name from the script. As a last resort although it's probably not recommended I guess the executable's name could just be hardcoded.

TSRBerry:
See my review comments.

src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
```diff
@@ -52,7 +52,7 @@ private static void CreateShortcutLinux(string applicationFilePath, byte[] iconD
         [SupportedOSPlatform("macos")]
         private static void CreateShortcutMacos(string appFilePath, byte[] iconData, string desktopPath, string cleanedAppName)
         {
-            string basePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, AppDomain.CurrentDomain.FriendlyName);
+            string basePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Ryujinx");
             var plistFile = EmbeddedResources.ReadAllText("Ryujinx.Ui.Common/shortcut-template.plist");
             // Macos .App folder
             string contentFolderPath = Path.Combine(desktopPath, cleanedAppName + ".app", "Contents");
```
nickoanastassiu, line 58: The changes proposed by TSRBerry close #5827.

I would also suggest putting the shortcut in the Applications folder on macOS.
```suggestion
            string contentFolderPath = Path.Combine("/Applications", cleanedAppName + ".app", "Contents");
```

TSRBerry: @marysaka is this desired?

NitroTears:
From what I can see here, I think we might need to move the logic in `GetArgsString` into the individual platform functions since it seems we might need different logic per platform. 

> Removing the executable path does not fix it (the shell cannot simply open the file, it needs the executable to open it with and then the game path as an argument, just like on other platforms).

To me this seems to conflict with the suggested fix at the top of the description in #5839. That is what I implemented, and while it seems it might OK for windows, not so much for mac. Let me know what you think.

TSRBerry:
@NitroTears how does this look for you?

gdkchan:
I didn't review the original PR so commenting now, there should be a separation on the context menu since the "Create Desktop Shortcut" option is not related to the cache management options before it.
image

nickoanastassiu:
Perhaps it would be smarter to bundle it in the first section?

Screenshot 2023-10-28 at 22 11 28

gdkchan:
Yea that would be better, although I guess it doesn't exist on GTK.

NitroTears:
I'll combine what @gdkchan and @nickoanastassiu said and move the create shortcut button to the top section (and create a seperator for it to be at the top in GTK, if that is alright). I'll also:

- Make the application folder the default shortcut location on Mac
- Add the changes by @TSRBerry 
- remove `--fullscreen` arg again until I know a definitive yay or nah for having in by default.

NitroTears:
I realise @TSRBerry did most of that already in a commit lol. So it's good to review / have a look at again now. Testing on Mac would be much appreciated.

nickoanastassiu:
I already tried this exact change on my machine locally yesterday, it should work.

marysaka (approved):

NitroTears:
Excellent, Thank you!

marysaka:
Forcing merge as all comments were addressed appropriately.

Issue #5798, opened 2023-10-12
[Bug][MacOS]After pause the emulation, Ryujinx still takes a lot of resources on Mac, causing heat and battery draining fast


jixinliu666:
### Description of the issue

Seems even after pause of the emulation, Ryujinx still use one-two CPU cores running some task
<img width="623" alt="image" src="https://github.com/Ryujinx/Ryujinx/assets/131228024/6d356c95-83d6-460f-a685-1dfcc88c9c3f">
<img width="733" alt="image" src="https://github.com/Ryujinx/Ryujinx/assets/131228024/6a130f03-d5bf-479c-a1b4-e37bccd19b45">

### Reproduction steps

Start any switch game in Mac, pause emulation,check macOS Activity Monitor, still seeing Ryujinx seems running a lot of 

### Log file

N/A

### OS

Mac OS Sonoma 

### Ryujinx version

1.1.1050

gdkchan:
Does the issue still happen on latest version?

jixinliu666:
> Does the issue still happen on latest version?

@gdkchan it improves a lot from 174% usage to 35%

jixinliu666:
@gdkchan I also noticed performance of ToTK has improved a little bit on Mac

ballo:
I can't even get the game to pause in my M1 MacBook Pro most of the time. The F-keys don't register and I have to use the menu. I'm going to make a ticket to auto-pause the game when it's not in the foreground (since I can't see that option anywhere)

PR #5855, opened 2023-10-29
[HLE] Remove ServerBase 1ms polling

riperiperi:
Added a KEvent for each ServerBase which signals whenever a session is added to the _sessions list, which allows it to rerun the ReplyAndReceive with the new session handle.

This greatly reduces the presence of ServerBase on CPU profiles, especially of games that aren't particularly busy. It should also increase responsiveness when adding session objects, as it doesn't take at most 1ms for them to start working.

It also reduces the load on KTimeManager, which could allow it to spin less often. I have noticed that a bunch of games still do 1ms waits (they actually request a bit less than 1ms), so they still end up spinning to the next millisecond. Maybe for waits like this, it could attempt to nudge the timepoints to snap to each other when they're close enough, and also snap to whole millisecond waits when close enough.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -39,6 +39,8 @@ namespace Ryujinx.HLE.HOS.Services
         private readonly KernelContext _context;
         private KProcess _selfProcess;
         private KThread _selfThread;
+        private KEvent _wakeEvent;
+        private int _wakeHandle = 0;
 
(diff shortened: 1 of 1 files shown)


gdkchan:
Does it fix #5798?

riperiperi:
It might, the guest threads should not be scheduling anything when the emulation is paused.

gdkchan (approved):
lgtm, thanks! This issue should be fixed with the new IPC, but since it may take a *while* to migrate all services, something like this makes sense, and the code changes look simple enough.

AcK77 (approved):

PR #5869, opened 2023-10-31
Better handle instruction aborts when hitting unmapped memory

shadowninja108:
At the advice of @gdkchan, adjusted ARMeilleure to better handle unmapped memory. It now actually performs a read when beginning decoding code. This will trip the invalid access handler when execution jumps directly into unmapped memory. This allows for a proper stack trace and register dump, instead of just an exception being thrown.

### DIFF
--- a/src/ARMeilleure/Decoders/Decoder.cs
+++ b/src/ARMeilleure/Decoders/Decoder.cs
@@ -38,7 +38,9 @@ namespace ARMeilleure.Decoders
                 {
                     block = new Block(blkAddress);
 
-                    if ((dMode != DecoderMode.MultipleBlocks && visited.Count >= 1) || opsCount > instructionLimit || (visited.Count > 0 && !memory.IsMapped(blkAddress)))
+                    if ((dMode != DecoderMode.MultipleBlocks && visited.Count >= 1) ||
+                        opsCount > instructionLimit ||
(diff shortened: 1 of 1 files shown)


TSRBerry (approved):

src/ARMeilleure/Decoders/Decoder.cs
```diff
@@ -38,7 +38,7 @@ Block GetBlock(ulong blkAddress)
                 {
                     block = new Block(blkAddress);
 
-                    if ((dMode != DecoderMode.MultipleBlocks && visited.Count >= 1) || opsCount > instructionLimit || !memory.IsMapped(blkAddress))
+                    if ((dMode != DecoderMode.MultipleBlocks && visited.Count >= 1) || opsCount > instructionLimit || (visited.Count > 0 && !memory.IsMapped(blkAddress)))
```
gdkchan, line 41:
```suggestion
                    if ((dMode != DecoderMode.MultipleBlocks && visited.Count >= 1) ||
                        opsCount > instructionLimit ||
                        (visited.Count > 0 && !memory.IsMapped(blkAddress)))
```

gdkchan:
We might want to handle data abort and instruction abort differently in the future, in which case we would need to do that in a different way. I did not think on what is the best way to handle this yet.

marysaka (approved):

PR #5884, opened 2023-11-02
Ava UI: Add accelerator keys for Options and Help

IsaacMarovitz:
Fixes comment here: https://github.com/Ryujinx/Ryujinx/issues/3662#issuecomment-1791106799


MutantAura (approved):
Do other languages need a similar change?

IsaacMarovitz:
> Do other languages need a similar change?

I think they would yes, but I'm not sure how it works for non-Latin languages.

marysaka (approved):

PR #5898, opened 2023-11-06
Force all exclusive memory accesses to be ordered on AppleHv

### LINKED ISSUE #5219: [Bug] Hypervisor not working with BotW on MacBook Pro
### Description of the issue

Can only use JIT with BotW, no hypervisor for some reason

### Reproduction steps

Run Zelda with Hypervisor enabled

### Log file

[Ryujinx_1.1.862_2023-06-04_22-47-48.log](https://github.com/Ryujinx/Ryujinx/files/11645945/Ryujinx_1.1.862_2023-06-04_22-47-48.log)

### OS

macOS

### Ryujinx version

1.1.862, 1.1.0-macos1

### Game version

1.6.0

### CPU

Apple M1 Pro

### GPU

Apple M1 Pro

### RAM

16GB

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
The LDXR/STXR (along with its variants) instructions can be used to perform exclusive memory load/store operations without any memory ordering guarantee in relation to previous or future memory operations. This means that the CPU is free to effectively re-order them, such as that the effects of future memory operations might be visible to other CPU cores before the effects of earlier ones. Whether such re-ordering occurs depends entirely on the CPU implementation, which means that the code can behave differently depending on the CPU.

Some games such as Zelda Breath of the Wild, use LDXR/STXR instructions in places where it should not be doing so, because a barrier is actually necessary but it is missing. This probably only works on the Switch due to "luck", or rather that the CPU implementation ensures proper ordering but the spec does not. This is certainly not the case on Apple Silicon and other Arm CPUs.

This PR fixes the issue by patching all LDXR/STXR (and friends) instructions to instead use instructions with implicit ordering guarantees, such as LDAXR/STLXR.

This fixes crashes and softlocks when running Zelda Breath of the Wild on Apple Silicon macOS with the hypervisor enabled. Apparently also fixes crashes on Splatoon 2 (couldn't really test that one a lot since I can't climb walls there, so it's not possible to finish the tutorial).

Suggestions are welcome. It doesn't really feel very good to replace all unordered exclusive instructions like that, after all they exist for a reason, the ordered variant is slower, but I don't know how much impact that will have in practice.

Testing is also welcome, I didn't really test this a lot.

Fixes #5219.


src/Ryujinx.Cpu/AppleHv/HvMemoryManager.cs
```diff
@@ -736,6 +721,24 @@ private static int GetPagesCount(ulong va, ulong size, out ulong startVa)
             return (int)(vaSpan / PageSize);
         }
 
+        /// <inheritdoc/>
+        public void Reprotect(ulong va, ulong size, MemoryPermission protection)
+        {
+            if (protection.HasFlag(MemoryPermission.Execute))
+            {
+                // Some applications uses unordered exclusive memory access instructions
```
riperiperi, line 729:
```suggestion
                // Some applications use unordered exclusive memory access instructions
```

IsaacMarovitz (approved):

riperiperi (approved):
Very nice find, this must have been a nightmare. I haven't crashed yet in Animal Crossing: New Horizons, so that may have also been fixed.

I think this is probably the best way to do it, if it is an issue with the CPU as a whole. We can't really trust that the instruction will be fine as is.

jjaksic:
This fixed my BotW with hypervisor on M1 (yay!). However, if hypervisor is on, loading of a saved game frequently hangs at around 80% (oddly, the UI remains responsive, just the loading progress stops and never completes). This happens at around 50% rate or higher, which is quite annoying, while it never happens if hypervisor is off.

Are there any other settings that might impact stability under hypervisor?

gssci:
> This fixed my BotW with hypervisor on M1 (yay!). However, if hypervisor is on, loading of a saved game frequently hangs at around 80% (oddly, the UI remains responsive, just the loading progress stops and never completes). This happens at around 50% rate or higher, which is quite annoying, while it never happens if hypervisor is off.
> 
> Are there any other settings that might impact stability under hypervisor?

Yes, multiple people have the same issue and @gdkchan is already looking into it ;) 
It seems that there's still odd behaviour during game execution, possibly some threads being terminated before they have a chance to start. Hopefully they will find a workaround, but it doesn't seem easily fixable because it's random and not easily reproducible.

jjaksic:
Yes, without knowing really anything, I'm guessing it's some sort of race condition.
The issue seems to happen only when the loading progress moves slowly. If it loads quickly, then it doesn't hang.
I just realized that having not enough free RAM makes the issue way worse. I have 32 GB, but I'm too lazy to close any work apps, so my RAM was all used up even before starting Ryujinx, so I just enabled texture recompression and hoped for the best :) Now out of curiosity I freed 14 GB by closing IntelliJ (and disabled texture recompression - not sure if it plays a role), and now it loads way faster and doesn't hang.

In summary, you can repro the issue quite consistently by having no RAM available :)

fann1993814:
I find `Grand Theft Auto: San Andreas – The Definitive Edition` cannot launch after the PR updating.
Toggle off `Use Hypervisor` can launch , but it will be crash easily.

siegmund-heiss-ich:
> I find `Grand Theft Auto: San Andreas – The Definitive Edition` cannot launch after the PR updating. Toggle off `Use Hypervisor` can launch , but it will be crash easily.

Can you provide a logfile of the crash? It may help to figure out a fix.

PR #5909, opened 2023-11-09
Do not set modified flag again if texture was not modified

### LINKED ISSUE #5908: [Bug][Regression] Digimon Story Cyber Sleuth: Complete Edition - 1.1.1066 causes graphics regression
### Description of the issue

In Digimon Story Cyber Sleuth: Complete Edition on OpenGL (may affect Vulkan as well, but unable to test), texture colors occasionally get garbled upon reloading the map - be it by getting into a battle, or changing areas, and can sometimes, but not always, be corrected by doing the same thing, often requiring changing areas 5-6 times before temporarily correcting itself, and sometimes garbling again immediately after the next transition again. Here is a screenshot of one example of this 'garbling':
image

This issue is a regression. I started having this issue after updating to 1.1.1061 to 1.1.1069 2 weeks ago, and have tested both 1.1.1065, where the issue does not happen anymore, and 1.1.1066, where the issue starts happening. The 'flushes', or at least some of them, seem to actually be necessary after all, as without them, this mess is what you get with this game. I will be using 1.1.1065 until this is fixed, but I thought I would let everyone know.

### Reproduction steps

1. Install Ryujinx 1.1.1066 or above
2. Run Digimon Story Cyber Sleuth: Complete Edition, with this save: [0000000000000005.zip](https://github.com/Ryujinx/Ryujinx/files/13302399/0000000000000005.zip), using OpenGL as the graphics backend.
3. Change areas by going north into the detective agency (pictured above in the issue description).
4. Change areas again by leaving the agency.
5. Enter the agency again.
6. Exit the agency again.
7. Keep on the left side of the screen and move south until you get a prompt to press A, then press A, which will take you to the DigiLab.
8. Move away from the starting point in the DigiLab, then move back until you get an A button prompt again; press A to leave the DigiLab.
9. Repeat steps 3 to 8 until the graphics get garbled. It will eventually happen, just keep at it.

### Log file

[Ryujinx_1.1.1072_2023-11-08_18-16-23.log](https://github.com/Ryujinx/Ryujinx/files/13302420/Ryujinx_1.1.1072_2023-11-08_18-16-23.log)

### OS

Windows 10 22H2 (64-bit)

### Ryujinx version

1.1.1066

### Game version

1.0.3

### CPU

i7-7700

### GPU

MSI NVIDIA GeForce GTX 1060 3GB

### RAM

16GB

### List of applied mods

_No response_

### Additional context?

Some cheats were added, but this issue started long after cheats were started being used, so they are most certainly unrelated.

gdkchan:
Basically aims to fix the same problem as #5755, but implemented using a different approach. Instead of skipping the flush, this avoids setting the modified flag if the texture has not been modified since the last time it was set. This fixes regressions caused by #5755 in some OpenGL games.

Testing is welcome.

Fixes #5908.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -104,7 +104,7 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// <summary>
         /// Indicates that the texture was modified since the last time it was flushed.
         /// </summary>
-        public bool ModifiedSinceLastFlush  { get; set; }
+        public bool ModifiedSinceLastFlush { get; set; }
 
         /// <summary>
         /// Increments when the host texture is swapped, or when the texture is removed from all pools.
--- a/src/Ryujinx.Graphics.Gpu/Image/Texture.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/Texture.cs
@@ -102,9 +102,9 @@ namespace Ryujinx.Graphics.Gpu.Image
         public bool AlwaysFlushOnOverlap { get; private set; }
 
         /// <summary>
-        /// Indicates that the texture was fully unmapped since the modified flag was set, and flushes should be ignored until it is modified again.
+        /// Indicates that the texture was modified since the last time it was flushed.
         /// </summary>
-        public bool FlushStale { get; private set; }
+        public bool ModifiedSinceLastFlush  { get; set; }
 
         /// <summary>
         /// Increments when the host texture is swapped, or when the texture is removed from all pools.
@@ -1417,7 +1417,6 @@ namespace Ryujinx.Graphics.Gpu.Image
         /// </summary>
         public void SignalModified()
         {
-            FlushStale = false;
             _scaledSetScore = Math.Max(0, _scaledSetScore - 1);
 
             if (_modifiedStale || Group.HasCopyDependencies)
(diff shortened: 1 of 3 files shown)


Littl3Guy:
Found a regression in 3D All Stars:

https://github.com/Ryujinx/Ryujinx/assets/115218507/c1c82894-16d1-43ad-a1e1-926087b504ee


I couldn't find any other regressions.

gdkchan:
I didn't realize that the `Modified` flag is used for something other than just flush, so this change is not really correct. I will try to think on something else, but everything I try breaks something...

IsaacMarovitz:
Fixes lens flare flicker in BOTW, and might fix Persona 5 too #5910

gdkchan:
I fixed the Super Mario 64 regression.

M1CR0H4CK3R:
> I didn't realize that the `Modified` flag is used for something other than just flush, so this change is not really correct. I will try to think on something else, but everything I try breaks something...

It may need something more in-depth, like a case-by-case basis for every instance of the Modified flag changing... which may lead to slight slowdowns every time the flag changes (like, 0.3 to 0.5 milliseconds, so not much, but it's still there) depending on how it's coded (if it's an async function, it may not give a performance drop, but may lead to bugs during edge cases). I know precious little about cpp, but I at least know that much.

marysaka (approved):

riperiperi (approved):
Something is still confusing about the 3d allstars issue, but since this works for now and fixes some regressions I’ll approve.

ThePiban:
The issue still occurs in Persona 5 Royal.
https://youtu.be/c2SPHMUOb0U 
(sorry for a yt vid I cant upload file larger than 10mb)

ThePiban:
> The issue still occurs in Persona 5 Royal. https://youtu.be/c2SPHMUOb0U (sorry for a yt vid I cant upload file larger than 10mb)

sorry I uploaded the vid with wrong version, here's the vid with the recent master version
https://www.youtube.com/watch?v=_b5tTAmyew8

PR #5912, opened 2023-11-11
chore: Update OpenTK to 4.8.1

marysaka:
OpenTK.OpenAL was renamed to OpenTK.Audio.OpenAL.


AcK77 (approved):

TSRBerry (approved):
I suppose I have some driver issues. OpenGL doesn't work on master for me as well, always freezes somewhere (both Ava and GTK) and segfaults.
The update doesn't make it worse, so I guess it's fine.

gdkchan (approved):
Code changes lgtm, but I did not test it.

PR #5913, opened 2023-11-11
misc: Default to Vulkan if available or running on macOS

MutantAura:
MessageBox asking users to choose on fresh installation has been removed in favour of Vulkan if any of two conditions are met:
- The `VulkanRenderer.GetPhysicalDevices()` method returns any valid devices. There are already checks within said method for version and featureset.
- The system is running macOS, where Vulkan was already the only target.

Due to the maturity of the Vulkan backend I think this is a better default experience than the current wall of text popup for new users.

### DIFF
--- a/src/Ryujinx/Program.cs
+++ b/src/Ryujinx/Program.cs
@@ -177,8 +177,6 @@ namespace Ryujinx
                     ? appDataConfigurationPath
                     : null;
 
-            bool showVulkanPrompt = false;
-
             if (ConfigurationPath == null)
             {
                 // No configuration, we load the default values and save it to disk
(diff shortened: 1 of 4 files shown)


src/Ryujinx/Program.cs
```diff
@@ -187,25 +185,17 @@ static void SetEnvironmentVariableNoCaching(string key, string value)
                 ConfigurationState.Instance.LoadDefault();
                 ConfigurationState.Instance.ToFileFormat().SaveConfig(ConfigurationPath);
```
gdkchan, line 187:
```suggestion
```

gdkchan:
Seems fine to me, most people probably doesn't know the difference, and might choose OpenGL when most of the time Vulkan is the best choice.

gdkchan (approved):
lgtm, thanks.

IsaacMarovitz (approved):

riperiperi (approved):
Looks sensible. UI common project importing Vulkan backend is a bit odd, though it is the best place for this check with all the other default initializatoon.

src/Ryujinx/Program.cs
```diff
@@ -177,35 +177,24 @@ static void SetEnvironmentVariableNoCaching(string key, string value)
...
-
-                showVulkanPrompt = true;
             }
             else
             {
                 if (ConfigurationFileFormat.TryLoad(ConfigurationPath, out ConfigurationFileFormat configurationFileFormat))
                 {
                     ConfigurationLoadResult result = ConfigurationState.Instance.Load(configurationFileFormat, ConfigurationPath);
```
gdkchan, line 192: The result can be removed as well (this `Load` method can be changed to return `void`, and `ConfigurationLoadResult` type can be deleted).

MutantAura: Remove the type and updated the method.

Issue #5918, opened 2023-11-12
MacBook Pro Os BigSur instantly crashing when running any game ( see crashlog )


clemensphilipps:
### Description of the issue

When i run a game i see for a second the loading bar and game image, then the app crashes. MacBook Pro, i7, Vulkan with Intel Iris Graphics

Model: MacBookPro11,2, BootROM 478.0.0.0.0, 4 processors, Quad-Core Intel Core i7, 2,2 GHz, 16 GB, SMC 2.18f15
Graphics: kHW_IntelIrisProItem, Intel Iris Pro, spdisplays_builtin
Memory Module: BANK 0/DIMM0, 8 GB, DDR3, 1600 MHz, 0x80AD, 0x484D54343147533641465238412D50422020
Memory Module: BANK 1/DIMM0, 8 GB, DDR3, 1600 MHz, 0x80AD, 0x484D54343147533641465238412D50422020
AirPort: spairport_wireless_card_type_airport_extreme (0x14E4, 0x134), Broadcom BCM43xx 1.0 (7.77.111.1 AirPortDriverBrcmNIC-1680.11)
Bluetooth: Version 8.0.5d7, 3 services, 19 devices, 1 incoming serial ports
Network Service: Wi-Fi, AirPort, en0
Serial ATA Device: APPLE SSD SM0256F, 251 GB
USB Device: USB 3.0 Bus
USB Device: Apple Internal Keyboard / Trackpad
USB Device: BRCM20702 Hub
USB Device: Bluetooth USB Host Controller
Thunderbolt Bus: MacBook Pro, Apple Inc., 17.1

Thread 81 Crashed:
0   libsystem_kernel.dylib        	0x00007fff2065790e __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff206865bd pthread_kill + 263
2   libsystem_c.dylib             	0x00007fff205db406 abort + 125
3   org.ryujinx.Ryujinx           	0x000000010f83d8a2 PROCAbort + 50
4   org.ryujinx.Ryujinx           	0x000000010f83d7d0 PROCEndProcess(void*, unsigned int, int) + 320
5   org.ryujinx.Ryujinx           	0x000000010f3ed637 UnwindManagedExceptionPass1(PAL_SEHException&, _CONTEXT*) + 1127
6   org.ryujinx.Ryujinx           	0x000000010f3ed683 DispatchManagedException(PAL_SEHException&, bool) + 67
[...]

PR #5925, opened 2023-11-14
Change minimum OS to macOS 12 in Info.plist

shinra-electric:
This should prevent the app from opening on macOS 11 and lower, informing the user that their OS is unsupported.

This should prevent (at least some) people from filing bug reports for unsupported OS versions. 

Important: I don't know for certain if macOS 12 is actually technically the lowest supported version. I have just seen it mentioned [here](https://github.com/Ryujinx/Ryujinx/issues/5918#issuecomment-1809349501) and in several other posts on Discord. This should be confirmed before the PR is merged. 

Testing: 
Try to open the app on macOS 11 or lower and confirm that the "Unable to open" dialog appears.

### DIFF
--- a/distribution/macos/Info.plist
+++ b/distribution/macos/Info.plist
@@ -43,7 +43,7 @@
     <key>LSApplicationCategoryType</key>
     <string>public.app-category.games</string>
     <key>LSMinimumSystemVersion</key>
-    <string>11.0</string>
+    <string>12.0</string>
     <key>UTExportedTypeDeclarations</key>
     <array>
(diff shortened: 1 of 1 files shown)


shinra-electric:
The app will have a 🚫 over it, and if the user tries to open it the dialog should look like this example: 

Screenshot 2023-11-14 at 13 10 42

marysaka (approved):

IsaacMarovitz (approved):

AcK77 (approved):

PR #5927, opened 2023-11-14
Disable DMA GPU copy for block linear to linear copies

### LINKED ISSUE #5889: [Bug] Fashion Dreamer
### Description of the issue

Three issues found:
1. When inside Item Creator mode, groups of outfits show the same texture, but they should show what the item really looks like.  Tried many different combo of settings, same result.

2. When starting a new game, after walking around a bit you will see some models face show some wireframe, or whiteness, darkness, some clothes also become low resolution.

3. Low resolution clothes when walking around (may take 10-30 mins to present itself).

Thanks in advance.  Keep up the good work guys!

### Reproduction steps
Issue1:
1. Go in-game.
2. Progress until you unlock item creator (takes about 10 mins maybe)
4. Go to options - Item Creator (where you craft stuff)
5. View the icons in large or normal mode (any mode) the preview shows wrong clothing item preview, none should look identical.

Issue2:
1. Go in-game.
2. Walk around to some NPCs, after a few mins you should see some with weird faces.

Issue3:
1. Go in-game.
2. Walk around a lot and keep changing outfits, primarily printed dresses.

### Log file

[Ryujinx_1.1.1069_2023-11-03_13-13-15.log](https://github.com/Ryujinx/Ryujinx/files/13246202/Ryujinx_1.1.1069_2023-11-03_13-13-15.log)

### OS

Windows 11

### Ryujinx version

1.1.1069

### Game version

1.1.0

### CPU

Ryzen 5 3600

### GPU

Nvidia GTX 3070

### RAM

64GB

### List of applied mods

None

### Additional context?

(Issue1) Item Creator Mode - Example, Clothing items should all look different.  When you enter them to craft, it shows the right item, but when you go back to this screen it always show wrong items.
image

(Issue2) Walking around - weird faces sometimes.
image
image

Expected (from issue2): Changing scenes, it refreshes to be good.
image

(Issue3) Textures on clothes can go low-res sometimes.
image

gdkchan:
We have a fast path where we can avoid some layout conversion operations by setting texture data directly to the destination texture on DMA copy. However, this can also trigger for block linear to linear textures. Typically, block linear to linear copies are used when the target texture will be read from CPU, and in those cases setting the data on the target texture is not worth it (since it will need to be flushed again later). This change disables the GPU path for this case, doing the copy on CPU instead.

Also has the advantage of avoiding flush related bugs, it seems to fix #5889 (this game triggers this case frequently).
Also includes #5926 since it affects that game too.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -283,7 +283,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                 // but only if we are doing a complete copy,
                 // and not for block linear to linear copies, since those are typically accessed from the CPU.
 
-                if (completeSource && completeDest && (!dstLinear || srcLinear == dstLinear))
+                if (completeSource && completeDest && (srcLinear || !dstLinear))
                 {
                     var target = memoryManager.Physical.TextureCache.FindTexture(
                         memoryManager,
--- a/src/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
@@ -279,7 +279,11 @@ namespace Ryujinx.Graphics.Gpu.Engine.Dma
                 bool completeSource = IsTextureCopyComplete(src, srcLinear, srcBpp, srcStride, xCount, yCount);
                 bool completeDest = IsTextureCopyComplete(dst, dstLinear, dstBpp, dstStride, xCount, yCount);
 
-                if (completeSource && completeDest)
+                // Try to set the texture data directly,
+                // but only if we are doing a complete copy,
+                // and not for block linear to linear copies, since those are typically accessed from the CPU.
+
+                if (completeSource && completeDest && (!dstLinear || srcLinear == dstLinear))
                 {
                     var target = memoryManager.Physical.TextureCache.FindTexture(
                         memoryManager,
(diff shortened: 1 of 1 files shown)


src/Ryujinx.Graphics.Gpu/Engine/Dma/DmaClass.cs
```diff
@@ -279,7 +279,11 @@ private void DmaCopy(int argument)
                 bool completeSource = IsTextureCopyComplete(src, srcLinear, srcBpp, srcStride, xCount, yCount);
                 bool completeDest = IsTextureCopyComplete(dst, dstLinear, dstBpp, dstStride, xCount, yCount);
 
-                if (completeSource && completeDest)
+                // Try to set the texture data directly,
+                // but only if we are doing a complete copy,
+                // and not for block linear to linear copies, since those are typically accessed from the CPU.
+
+                if (completeSource && completeDest && (srcLinear || !dstLinear))
```
riperiperi, line 286:
```suggestion
                if (completeSource && completeDest && !(dstLinear && !srcLinear))
```

This condition really confused me when I first saw it, which I thought was odd given the explanation of the PR which should have made it obvious.

I think the reason for that is that it uses an OR to exclude a single case using all others, rather than using an AND to identify the single case and negating it. You have to think about all the cases that the condition is true, rather than just the case where it isn't.

It does add an extra negation, but I think it's clearer.

gdkchan: For me, everything with an additional negation makes things more confusing, because I have to think "actually, this case is *not* allowed, so the allowed cases are all others except this". But I don't mind changing it.

riperiperi (approved):
I think it makes sense. I did look into making this case accelerated for Pokemon Sword (one pixel copy from depth buffer, at least I think it was block linear to linear) where it would copy the data on GPU then allow deferred readback later on, so the root cause would need to be fixed if we want to do that in the future.

It might be worth checking if the root cause of the issue is host import sync, since flushing a linear texture takes that path.

gdkchan:
> It might be worth checking if the root cause of the issue is host import sync, since flushing a linear texture takes that path.

Someone said Fashion Dreamer renders correctly on the LDN build, so it seems plausible (that the cause is something added after the last LDN release), but I didn't test it myself.

PR #5887, opened 2023-11-02
Migrate to .NET 8

zcsizmadia:
.NET 8 will be released in 2 weeks. This PR is the preparation for that.

1. .NET 8 is still in RC, as soon as it is released, `global.json` must be modified
2. .NET 8+ recommends to use the portable RIDs, becuase of this, `win10-x64` is changed to `win-x64`. See https://aka.ms/netsdk1083

~~The following new warnings needs some attention and might need some discussion as well.~~

`
C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Texture\Utils\RgbaColor32.cs(109,72): warning CA1857: The argument should be a constant for optimal performance (https
://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1857)[C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Texture\Ryujinx.Graphics.Texture.cspro
j]
C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Texture\Utils\RgbaColor32.cs(122,73): warning CA1857: The argument should be a constant for optimal performance (https
://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1857) [C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Texture\Ryujinx.Graphics.Texture.cspro
j]
C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Vulkan\PipelineBase.cs(192,49): warning CS9192: Argument 1 should be passed with 'ref' or 'in' keyword [C:\OpenSource\
Ryujinx\src\Ryujinx.Graphics.Vulkan\Ryujinx.Graphics.Vulkan.csproj]
C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Vulkan\VulkanException.cs(37,85): warning SYSLIB0051: 'Exception.Exception(SerializationInfo, StreamingContext)' is ob
solete: 'This API supports obsolete formatter-based serialization. It should not be called or extended by application code.' (https://aka.ms/dotnet-warnings/SYSL
IB0051) [C:\OpenSource\Ryujinx\src\Ryujinx.Graphics.Vulkan\Ryujinx.Graphics.Vulkan.csproj]
C:\OpenSource\Ryujinx\src\Ryujinx.Audio\Renderer\Utils\SpanIOHelper.cs(28,65): warning CS9191: The 'ref' modifier for argument 2 corresponding to 'in' parameter
[...]

PR #5928, opened 2023-11-14
Resolve .NET 8 Warnings

IsaacMarovitz:
Follow up to #5887 that resolves all new warnings.

I've kept the disable for `IDE0290` as devs unanimously agreed against this change.


src/Ryujinx.Graphics.Shader/Instructions/InstEmitHelper.cs
```diff
@@ -140,22 +140,22 @@ public static Operand[] GetHalfUnpacked(EmitterContext context, Operand src, Hal
         {
             return swizzle switch
             {
-                HalfSwizzle.F16 => new Operand[]
-                                    {
+                HalfSwizzle.F16 =>
+                                    [
```
gdkchan, line 144: Indentation is off here (commenting now because I might miss it later).

PR #5933, opened 2023-11-15
Fix missing texture flush for draw then DMA copy sequence without render target change

gdkchan:
Textures are currently marked as modified on the first use of that texture as render target, and when it is swapped out for another texture. This effectively creates a "live range" for the render target, and the texture is set as modified at the start and end of that range. It is however, possible that the render target will be modified and accessed multiple times within that range, and that case can happen with DMA copies (where we currently always flush the source). This was causing a bug on the game Fashion Dreamer because it would render different objects to the same texture, then use DMA engine to copy it to another texture, and then (presumably) access it on the CPU, where it seems to compress the texture using BC3 compression. The problem is that only the first use of the texture would set it as modified, so it would only flush the first "version" of that texture, which means it would copy the same object texture multiple times.

The change here is on the same vein of my previous changes to `SignalModifying`, which also caused problems for another game recently. This adds a new `RefreshModifiedTextures` method that can be called at any time to end the render target "live range", which will set the texture as modified again. If the render target is still used in a future, it will start the range again and mark it as modified once more. This is done on `UpdateRenderTargets` which has been modified to check if render targets are "bound", and call `SignalModifying` again to "bind" them if not. `RefreshModifiedTextures` is currently only called before DMA copy, since that solves all cases I have found so far, but there are other cases that probably should call it too, suggestions are welcome.

The change above triggers a different bug that was revealed on https://github.com/Ryujinx/Ryujinx/pull/5909#issuecomment-1803174725. Copy dependencies currently rely on the `Modified` flag to know if the GPU still has the most up to date version of the texture. The assumption was that CPU writes would trigger flush and clear the flag. While this is correct for *writes*, reads also trigger flush, but they do not modify the data from the CPU side. Which means that the GPU still has the latest version of the texture, and the copy should still happen, but since flush clears the `Modified` flag, that copy is lost. To fix this issue, I made the `Copy` method always do the copy on the relevant path, and changed the `DirtyAction` method (which has also been moved) to clear the deferred copy. I also audited all places that set `Modified` to false, and made it clear the deferred copy where it made sense (or in the `CreateCopyDependency` case, not set it at all if the texture was not modified).

This fixes the duplicate items bug on the game Fashion Dreamer (at least on the Item Creator menu, it's the only place where I was able to reproduce the bug so far):
Before:
image
After:
image

Testing is welcome. Changing anything texture cache related makes me uneasy since it has a certain tendency to break specific cases on specific games that are hard to know in advance, so let me know if anything breaks as a result of this change, especially since it touches copy dependency...


riperiperi (approved):
Nice catch, this was probably causing weird issues all over the place. Tested Super Mario 3D All-Stars, Mario Kart 8 Deluxe, Pikmin 4 and Yoshi's Wooly World and everything seems to be fine. I wasn't able to replicate the old bug where Pikmin 4's lighting would break at 2x after an hour, but maybe I just got lucky.

marysaka (approved):

PR #5936, opened 2023-11-16
Halve primitive ID when converting quads to triangles

gdkchan:
When the API or driver does not have native native support for quad primitives, we convert them to triangles. Because we convert each quad to 2 triangles, when we render quads, we have twice as many primitives than the game submitted. For this reason, the value for `gl_PrimitiveID` which is the index of the primitive on the shader is incorrect, and it is also double of what it should be.

To fix this issue, this change divides the ID by 2 if the primitive type is `Quads` or `QuadStrip`, the host does not support quads, and there is no geometry shader (as geometry shader must always generate new primitives, and they can't output quads).

I'm not aware of any visible improvements from this change on its own, **but when combined with #3001**, it fixes rendering on Castle Crashers Remastered.
Before:
image
After:
image


riperiperi (approved):
Looks sensible. Is this case not covered by `ShaderSpecializationState.MatchesGraphics`? (won't compile variants where quads is used and not)

gdkchan:
Superseded by #6721.

PR #5943, opened 2023-11-16
Enable copy dependency between RGBA8 and RGBA32 formats

gdkchan:
A few OpenGL games do a image store using a texture that has 1/4 of the width and 4x the amount of bytes per pixel (effectively storing 4 pixels in one invocation). Then later, the texture is accessed using the "true" format. This does not work right now because the formats are incompatible, one texture uses a RGBA32 format, and the other RGBA8. A while ago, I enabled copy between incompatible formats since the game Mario + Rabbids Sparks of Hope does something similar (but the formats are R8 and RGBA8 instead). This change enables copy dependency between RGBA32 and RGBA8 formats.

Fixing this reveals another bug, which is why I did not include this fix on the Mario + Rabbids targeted change back then. The format of the input texture that the shader samples for the image store was also wrong. It was using a RGBA8 format, which is the "default" format that the emulator uses right now when the format is invalid. I did not realize that the format was "invalid" back then because it was not logging any error.

The reason for the log not printing is this code:
```cs
if (gpuVa != 0 && (int)format > 0)
{
    Logger.Error?.Print(LogClass.Gpu, $"Invalid texture format 0x{format:X} (sRGB: {srgb}).");
}
```
The `(int)format > 0` check means it won't print if the most significant bit is set, and this was the case for this format. I don't remember why this check was added, I believe it was to prevent log spam on some games with invalid formats.

When the most significant bit of the format is set, the components should be tightly packed on the registers in the shader. To emulate this behaviour without needing to change the shader, I made it instead return compatible formats that only have a single component (or 2/4 components for 8 and 16 bytes per pixel formats) which should ensure values are tightly packed.

This finally fixes cicadas not rendering when you catch them on Yo-kai Watch 1.
Before:
image
After:
image
Also improves rendering on Wet Steps, ~~but this game still has other rendering issues~~. There was also another issue, when comparing the widths for textures with different formats, it did not take into account the fact that they might not be a multiple of each other, and one of them having a little padding for this reason. I have fixed that as well, and the game seems to be rendering correctly now.
Before:
image
image
After:
image
image
Other OpenGL games might also be positively impacted.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
@@ -430,7 +430,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (!FormatTable.TryGetTextureFormat(format, srgb, out FormatInfo formatInfo))
             {
-                if (gpuVa != 0 && (int)format > 0)
+                if (gpuVa != 0)
                 {
                     Logger.Error?.Print(LogClass.Gpu, $"Invalid texture format 0x{format:X} (sRGB: {srgb}).");
                 }
--- a/src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
@@ -430,7 +430,7 @@ namespace Ryujinx.Graphics.Gpu.Image
 
             if (!FormatTable.TryGetTextureFormat(format, srgb, out FormatInfo formatInfo))
             {
-                if (gpuVa != 0)
+                if (gpuVa != 0 && format != 0)
                 {
                     Logger.Error?.Print(LogClass.Gpu, $"Invalid texture format 0x{format:X} (sRGB: {srgb}).");
                 }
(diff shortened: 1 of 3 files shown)


marysaka (approved):

src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
```diff
@@ -430,7 +430,7 @@ private static TextureInfo GetInfo(in TextureDescriptor descriptor, out int laye
 
             if (!FormatTable.TryGetTextureFormat(format, srgb, out FormatInfo formatInfo))
             {
-                if (gpuVa != 0 && (int)format > 0)
+                if (gpuVa != 0)
```
riperiperi, line 433: I think the `(int)format > 0` check was added to stop cases where the format is `0` from spamming logs.

riperiperi (approved):
Nice.

src/Ryujinx.Graphics.Gpu/Image/TexturePool.cs
gdkchan: In that case it should have been `format != 0`. In any case, I added a `format != 0` check.

PR #5948, opened 2023-11-17
HLE: Add OS-specific precise sleep methods to reduce spinwaiting

riperiperi:
This PR adds methods that help the emulator wait for more precise amounts of time while also minimizing impact to CPU utilization wherever possible. This is done differently for platforms with libc nanosleep (linux, macos) and NtDelayExecution. The goal is to reduce threads spinning while also keeping a reasonable amount of scheduler timing accuracy, and even improving it in some cases.

## Problems with concurrent waits

#5855 stopped all ServerBase threads from waiting 1ms, which greatly reduced the amount of millisecond waits that bisected each other. Problem solved, right? Well, no, a game's own threads can still do this same thing, and they can wait for slightly less than a millisecond, which will cause us to spin all the way there.

## Nanosleep adjustment

While Linux and MacOS both have `nanosleep` for waiting a precise number of nanoseconds, they deliberately adjust the timeout and introduce their own inaccuracy that can add onto our own timing inaccuracies in a negative way.

### MacOS

image

MacOS on M1 has a curious behaviour here. At low nanosecond waits, you get extremely accurate sleep durations. At long waits such as 20ms, you get _3ms_ of additional error, which is a lot. Setting the thread priority to "above normal" _halves_ how much it adds onto the thread, so it's what I've used for all timings. Highest isn't any better, so above normal is where it stays.

It adds a linear bias to the timeout of around 0.14ms per millisecond of waiting. This bias is a lot more than found on linux, but there's a lot more precision in the wake time. When accounting for this, we can get surprisingly high accuracy for the majority of waits.

### Linux

image

Testing platform is SteamOS 3 on Steam Deck (arch linux). Below 0.5ms, we get a very precise nanosleep with a small constant penalty on the timeout. As soon as 0.5ms is exceeded, a larger timeout penalty is added (with a visible spike), which increases until the input is 1.5ms. Past this, it seems pretty constant.

## The hardware clock on Windows and timepoint alignment

On a typical x86 system nowadays, you can query the clock resolution and discover that there's a 0.5ms base clock that you can align to (at least, that's the minimum). Not only that, but all your waits seem to align to it automatically. This is both a boon and a detriment - a naive sleep or timeout might take extra time to align to the next clock tick... But if you know the clock phase, then you can make a very precise guess about how long you can wait before a timepoint, and nearly always wake before it.

This brought about the timer thread idea that just has one thread woken on every tick. From this thread, we can register the clock's alignment, and signal any potential waiters out of sleep. The thread has the highest priority, as missing a clock tick is undesirable, though its low utilization should avoid impacts on other important threads.

This can let us wait til at most the last 0.5ms before a timepoint, but there is still an issue. Think back to the section on concurrent waits - if they're out of phase, then even 20 1ms periodic waits can cause us to wait 0.05ms between each. We really want all of the waits on the same phase... aligned with the host clock, if possible, to prevent as much spinning as we can.

Alignment is only performed if the requested timeout aligns with the host clock (0.5ms). We don't want to align precise waits like 16.667ms, or 0.1ms, as doing that repeatedly could drastically change the expected rate that threads are resumed. There is a bias to align to a later clock tick, but it can also move the timepoint backwards if it's close enough (ahead by 1/6 of the clock frequency).

The hope is that when a timepoint is aligned, if the thread does minimal work it will stay aligned, and error will be minimized.

```
00:00:15.479 |W| HLE.GuestThread.71 KernelScheduler : Realigned timepoint -0.0194ms
00:00:15.479 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0766ms
00:00:15.480 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0395ms
00:00:15.481 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0347ms
00:00:15.482 |W| HLE.GuestThread.67 KernelScheduler : Realigned timepoint -0.049ms
00:00:15.482 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0882ms
00:00:15.482 |W| HLE.GuestThread.65 KernelScheduler : Realigned timepoint -0.1248ms
00:00:15.482 |W| HLE.GuestThread.63 KernelScheduler : Realigned timepoint 0.3177ms
00:00:15.483 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.042ms
00:00:15.484 |W| HLE.GuestThread.71 KernelScheduler : Realigned timepoint 0.1586ms
00:00:15.484 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0429ms
00:00:15.485 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0439ms
00:00:15.486 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0261ms
00:00:15.487 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0424ms
00:00:15.487 |W| HLE.GuestThread.63 KernelScheduler : Realigned timepoint -0.0503ms
00:00:15.488 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0569ms
00:00:15.489 |W| HLE.GuestThread.45 KernelScheduler : Realigned timepoint -0.0405ms
```

### Windows still spins in some cases

Sometimes games are not very friendly. Breath of the Wild in particular regularly sleeps for 0.25ms, which is not clock aligned and which we have to spin to wake it in time. 

## Accuracy

Existing accuracy, while using spin-waits, was not great. When a timepoint is more than a millisecond away, we waited 1ms on the `_waitEvent`. However, as explained, this actually aligns to the _next_ clock tick on windows (we also asked for 1ms before, not 0.5ms) and adds a bias on MacOS and Linux. This was already accounted for with the SurfaceFlinger spin, as it only sleeps until the last 2ms before the scheduled vsync event, but also meant that spent more time spinning the last 1-2ms.

Windows, Linux and MacOS are not real-time operating systems, and dotnet is not built for real-time applications. This means that we can easily be knocked off course by the OS scheduler, garbage collection, hangups when performing OS calls, thread interruptions. These typically result in errors ranging from 0.1ms to spikes of 2ms or 3ms. I think these should just be expected for now - maybe they can be improved by reducing overall garbage generation, and maybe using less contentious locks in the time manager.

## Results
Generally, this reduces active time of both the TimeManager and SurfaceFlinger threads. On Windows this depends on the game - most will improve, but a few will remain the same, such as BOTW. On Linux and MacOS, the improvement should be universal.

I'll post results in followup comments. I did have some screenshots, but things have changed since then and I want to make sure they're still accurate.


jcm93:
Here's a screenshot that illustrates the difference decently well for Breath of the Wild on macOS. The first half of the graph is before the changes, the second half is with them. Significant reduction in activity for the P cores.

Screenshot_2023-11-07_at_6 14 58_PM

riperiperi:
On Steam Deck, the difference is most noticable on the BOTW title screen. The game does start loading assets while the title screen is displayed, but after that is done, the CPU clocks still stay high and drain more battery than necessary.




Using nanosleep for the timemanager stops it from spinning 100% of the time, reducing the CPU clocks, temperature and power usage:

src/Ryujinx.Common/PreciseSleep/IPreciseSleepEvent.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// An event which works similarly to an AutoResetEvent, but is backed by a
+    /// more precise timer that allows waits of less than a millisecond.
+    /// </summary>
+    public interface IPreciseSleepEvent : IDisposable
+    {
+        /// <summary>
+        /// Adjust a timepoint to better fit the host clock.
+        /// When no adjustment is made, the inout timepoint will be returned.
```
gdkchan, line 13: Is `inout` supposed to be `input` here?

src/Ryujinx.Common/PreciseSleep/IPreciseSleepEvent.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <param name="timePoint">Timepoint to adjust</param>
+        /// <param name="timeoutNs">Requested timeout in nanoseconds</param>
+        /// <returns>Adjusted timepoint</returns>
+        long AdjustTimePoint(long timePoint, long timeoutNs);
+
+        /// <summary>
+        /// Sleep until a timepoint, or a signal is received.
+        /// Given no signal, may wake considerably before, or slightly after the timeout.
+        /// </summary>
+        /// <param name="timePoint"></param>
```
gdkchan, line 24: Missing description.

src/Ryujinx.Common/PreciseSleep/Nanosleep.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        /// Sleep for at most a given time period in nanoseconds.
+        /// Uses a stricter bias to wake before the requested duration.
+        /// </summary>
+        /// <remarks>
+        /// Due to OS scheduling behaviour, this timeframe may still be missed.
+        /// </remarks>
+        /// <param name="nanoseconds">Maximum allowed time for sleep</param>
+        public static void SleepBefore(long nanoseconds)
```
gdkchan, line 146: Maybe `SleepAtMost` is a better name?

src/Ryujinx.Common/PreciseSleep/NanosleepPool.cs
```diff
@@ -0,0 +1,223 @@
...
+                return false;
+            }
+
+            /// <summary>
+            /// Dispose the NanosleepThread, interrupting its worker loop.
+            /// </summary>
+            public void Dispose()
+            {
+                _running = false;
+                _newWaitEvent.Set();
```
gdkchan, line 120: I guess this will throw `ObjectDisposedException` if this is disposed twice. Could be nice to make that not happen (for example, by checking if `_running` is already false), since dispose methods should be callable multiple times IIRC.

src/Ryujinx.Common/PreciseSleep/NanosleepPool.cs
```diff
@@ -0,0 +1,223 @@
...
+        private readonly List<NanosleepThread> _active = new();
+        private readonly Stack<NanosleepThread> _free = new();
+        private readonly AutoResetEvent _signalTarget;
+
+        private long _signalId;
+
+        /// <summary>
+        /// Creates a new NanosleepPool with a target event to signal when a nanosleep completes.
+        /// </summary>
+        /// <param name="signalTarget"></param>
```
gdkchan, line 135: Missing description.

src/Ryujinx.Common/PreciseSleep/NanosleepPool.cs
```diff
@@ -0,0 +1,223 @@
...
+        public NanosleepPool(AutoResetEvent signalTarget)
+        {
+            _signalTarget = signalTarget;
+        }
+
+        /// <summary>
+        /// Signal the target event (if the source sleep has not been superceded)
```
gdkchan, line 142: Typo on `superceded`?

src/Ryujinx.Common/PreciseSleep/PreciseSleepHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Sleeps up to the closest point to the timepoint that the OS reasonably allows.
+        /// The provided event is used by the timer to wake the current thread, and should not be signalled from any other source.
+        /// </summary>
+        /// <param name="evt">Event used to wake this thread</param>
+        /// <param name="timePoint">Timepoint in host ticks</param>
+        public static void SleepUntilTimePoint(AutoResetEvent evt, long timePoint)
```
gdkchan, line 35: This is not very important, but maybe this should take a `EventWaitHandle` instead (which both `AutoResetEvent` and `ManualResetEvent` inherits from) to make the function a bit more flexible? (If we need to use it with a `ManualResetEvent` in the future for example).

src/Ryujinx.Common/PreciseSleep/NanosleepEvent.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace Ryujinx.Common.PreciseSleep
+{
+    /// <summary>
+    /// A precise sleep event for linux and macos that uses nanosleep for more precise timeouts.
+    /// </summary>
+    [SupportedOSPlatform("macos")]
+    [SupportedOSPlatform("linux")]
+    [SupportedOSPlatform("android")]
+    [SupportedOSPlatform("ios")]
+    public class NanosleepEvent : IPreciseSleepEvent
```
gdkchan, line 14: I think this and other classes doesn't need to be public, since they are supposed to be accessed using the `IPreciseSleepEvent` interface?

gdkchan (commented):
Nice work! I was mostly looking for typos etc, will try to review the logic itself and give it a try later.

MutantAura:
Some fairly large SoC power savings on my M2, probably scales across larger chips.
ToTK: 9.2W -> 6.9W (-25%)
BoTW: 9.3W -> 8.2W (-12%)
Red Dead Redemption: 9.8W -> 8.2W (-16%)

Compared to before the ServerBase polling change for my own interest:
ToTK: 10.9W -> 6.9W (-38%)

riperiperi:
> Any reason to use a dynamic bias? Maybe using EMA or even simple average with threshold to reset? or is there not enough variance to justify the cost of the calculation?

I posted the error given different input waiting times for both linux and MacOS. For both there is random variance, but it is greatly overshadowed by a clear pattern that adds additional waiting time the larger the requested timeout. To get the most accuracy, removing that component separately makes sense. For MacOS, this pattern is basically all of the error - there's not much variance for the same waiting time. On Linux there's a lot more variance, but there's definitely a linear increase.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -25,7 +27,7 @@ class SurfaceFlinger : IConsumerListener, IDisposable
 
         private readonly Stopwatch _chrono;
 
-        private readonly ManualResetEvent _event = new(false);
+        private readonly AutoResetEvent _event = new(false);
```
gdkchan, line 30: Was there a reason for changing this other than `SleepUntilTimePoint` not accepting a `EventWaitHandle` before?

src/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
```diff
@@ -103,13 +106,7 @@ private void WaitAndCheckScheduledObjects()
...
                         if (next.TimePoint > timePoint)
                         {
-                            long ms = Math.Min((next.TimePoint - timePoint) / PerformanceCounter.TicksPerMillisecond, int.MaxValue);
-
-                            if (ms > 0)
-                            {
-                                _waitEvent.WaitOne((int)ms);
-                            }
-                            else
+                            if (!_waitEvent.SleepUntil(next.TimePoint))
```
gdkchan, line 109: Why not implement the spin wait logic on the precise helper too (as part of the `SleepUntil` etc methods)? I think it would be a lot cleaner since this method would not need to concern itself with the wait logic, and would just call the sleep method. It could also help reduce code duplication between this and surface flinger which also spin waits.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
```diff
@@ -339,6 +342,10 @@ private void HandleComposition()
                     long diff = _ticksPerFrame - (_ticks + _chrono.ElapsedTicks - ticks);
                     if (diff > 0)
                     {
+                        PreciseSleepHelper.SleepUntilTimePoint(_event, PerformanceCounter.ElapsedTicks + diff);
```
gdkchan, line 345: Seems a litle weird that we use `PerformanceCounter.ElapsedTicks` here and `_chrono.ElapsedTicks` in others. Maybe se should remove `_chrono` and use `PerformanceCounter` everywhere?

src/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
riperiperi: The spin wait's a bit different between SurfaceFlinger and KTimeManager - on the latter it can be interrupted by an atomic variable flipping to 1, on SurfaceFlinger it just runs.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
riperiperi: Yes, the event gets signalled in `WindowsGranularTimer.SleepUntilTimePointWithoutExternalSignal`. It doesn't get cleared between waits, so if it were manual reset it would immeditately wake up every time.

src/Ryujinx.HLE/HOS/Services/SurfaceFlinger/SurfaceFlinger.cs
riperiperi: Agreed. I think `_chrono` exists for the possibility of pausing this timer, but it was never used. and none of the other timers pause right now anyways.

src/Ryujinx.HLE/HOS/Kernel/Common/KTimeManager.cs
gdkchan: It seems there are already 2 variants of the sleep method, one that can be signalled and another that can't (for surface flinger). So two difference spin waits could be added there too (although I guess in terms of code size it would be the same). Surface flinger could also use the variant that can be signalled, but never signal it, which would be less code, but I guess a bit more inefficient.

riperiperi: I've moved the spinwaiting out, but both spinwaits do have different behaviour. The KTimeManager spin is allowed to yield, wakes on signal and uses the `SpinWait` struct (and its tuning for wait durations and yields), whereas the SurfaceFlinger spin uses `Thread.SpinWait` and never yields or returns early.

gdkchan (approved):
lgtm, thanks! Nice to see this issue finally addressed, it will be great for devices running on battery.

amayra:
i do think that linux kernel support Real-time

PR #5949, opened 2023-11-17
Default to Avalonia on Windows

IsaacMarovitz:


MutantAura:
Exectuable name would need to be updated and/or NVAPI executable also needs updating to disable threaded opts.

AcK77:
I think it’s a bit too early for that since there are still some issue in Avalonia sadly, I know we now use this UI since a long time, but getting a PR opened for that since more than a month even with some long-term issues make no sense, this have to be discussed between developers before.
I close this PR in that way, it can be reopened later if needed.
Thanks.

MetrosexualGarbodor:
It's not too early at all, in fact it's long overdue; the only blocking issue Avalonia has affects the Steam Deck's gaming mode exclusively. Besides that, it's basically ahead of GTK in every way.

Several Windows users have been using Avalonia already as their default for months, and they can hardly find any issues with it. It's also been the default on macOS for over a year, and I've read practically zero complaints from macOS users about the UI, despite the fact that it had a lot more issues back when the first macOS build was pushed. 

GTK issues keep plaguing the Windows support channel, and external code contributors keep being mislead into writing code for a GUI that should've been phased out a while ago. I don't think Reshade not working on Avalonia (or whatever other minor issues remain) balances out either of these points alone. Besides, defaulting it would lead to whatever issues are left being better documented, and therefore most likely fixed sooner.

There's a frequent question of "when will Avalonia be merged into the main build?", despite the fact that it was merged so long ago; the majority of users don't even know about it, and one of the main complaints people have with Ryujinx is... The GUI. And understandably so, since GTK sometimes won't even let you install firmware or do something as simple as double-click a game to run it. Therefore, continuing to have such a broken interface as the default only hinders the development progress, the general user experience and, subsequently, the users' general perception of Ryujinx.

Please at least let Avalonia be the default on Windows. Thank you. Merry Christmas. Happy holidays. Good night.

PR #5951, opened 2023-11-17
Use new InlineArray attribute for inline array structs

gdkchan:
This starts using the `InlineArrayAttribute` that was introduced on .NET 8 to implement the inline array structs, instead of using fields with smaller inline arrays as types to ensure the struct has the correct size. This makes the code a bit cleaner. I also deleted the `ByteArrayN` structs and replaced them with `ArrayN<byte>` structs.

This required changing existing code for one of those reasons:
- The `IArray` struct was removed. Implementing it would require quite a bunch of code per struct, which would reduce the size savings from this change.
- No `AsSpan` method anymore, instead the struct can be implicit cast to span. The code has been adjusted to stop calling it and instead, cast to span when necessary.
- Compiler complaining that a ref to `this` was being returned from the structs.

I will comment on the lines of code that probably needs more attention on review.


src/Ryujinx.Graphics.Vulkan/BitMapStruct.cs
```diff
@@ -9,21 +9,21 @@ interface IBitMapListener
         void BitMapSignal(int index, int count);
     }
 
-    struct BitMapStruct<T> where T : IArray<long>
+    struct BitMapStruct
```
gdkchan, line 12: I removed the generic parameter because `IArray` interface is gone. It means we can't access a arbitrary length inline array anymore. Currently the size has been hardcoded to 2 (which is the only size it was using anyway). There are other possible, potentially better solutions:
- Use regular managed array, pass size on the constructor.
- Continue passing inline array as generic parameter, cast it to span and use the span for access.

TSRBerry:
Closes #4218

gdkchan:
It seems the analysers are not playing nicely with the `InlineArray` attribute. It is suggesting some member to be made `readonly` where it doesn't make sense to do so. I'm not sure what is the best thing to do here, I don't want to disable it on all places where it's broken since there's quite a lot of them, maybe it should be disabled globally until the analyser is fixed.

TSRBerry:
Related to https://github.com/dotnet/roslyn/issues/69143.

I'll try my luck with this PR again as soon as the issue mentioned above is solved.

PR #5957, opened 2023-11-19
Extend bindless elimination to see through Phis with the same results

### LINKED ISSUE #5946: [Bug] Super Mario RPG: Incorrect lighting on some character and environment geometry
### Description of the issue

When playing [Super Mario RPG](https://github.com/Ryujinx/Ryujinx-Games-List/issues/4811), some geometry appears improperly illuminated giving a dimmed and flat appearance.
This particularly affects player and NPC character models, and are highly noticeable on specific plants and bushes and MarioHouse on the Mario's Pad map.

These display issues occur regardless of graphical backend (OpenGL or Vulkan), and graphical enhancements such as resolution scale increases or enabling Macro HLE.

### Reproduction steps

Play through the game as normal, progressing through Bowser's Keep at the start of the game until you reach Mario's Pad where the issues become most notable. It is also noticeable on the Save screen and Map screen.
Lighting issues appear to vary from scene to scene and are not as noticeable during Bowser's Keep.

Switch hardware screenshots:
2023111618342300-DD36EAD444102ED69B0AF6882696E2E7
2023111618344300-DD36EAD444102ED69B0AF6882696E2E7
2023111618351000-DD36EAD444102ED69B0AF6882696E2E7
2023111618353100-DD36EAD444102ED69B0AF6882696E2E7
2023111609544200-DD36EAD444102ED69B0AF6882696E2E7
2023111609545000-DD36EAD444102ED69B0AF6882696E2E7

Ryujinx v1.1.1087 screenshots:
Screenshot (82)
Screenshot (84)
Screenshot (83)
Screenshot (85)
Screenshot (74)
Screenshot (75)

This provided save file will skip past the intro, starting at Mario's Pad: 
[SMRPG Save.zip](https://github.com/Ryujinx/Ryujinx/files/13390302/SMRPG.Save.zip)

### Log file

[Ryujinx_1.1.1087_2023-11-16_18-21-31.log](https://github.com/Ryujinx/Ryujinx/files/13390304/Ryujinx_1.1.1087_2023-11-16_18-21-31.log)

### OS

Windows 10 Pro version 22H2 build 19045.3570

### Ryujinx version

1.1.1087

### Game version

1.0.0

### CPU

12th Gen Intel(R) Core(TM) i7-12700K 3.60 GHz

### GPU

NVIDIA RTX 3070

### RAM

32GB DDR5 (2x16gb)

### List of applied mods

n/a

### Additional context?

For transparency, screenshots and logs were captured on 16/11/2023 a day ahead of the official release date.
These were captured using a personal dump of a physical UK copy that was delivered a day early.

gdkchan:
This extends bindless elimination to also work on cases where the handle is "materialized" on multiple places in the code. It was not able to reason about that before because a Phi was on the way, so it was not able to find the handle comining operation.

Basically, this code:
```glsl
handle = fp_c2.data[24].z | fp_c2.data[88].x;
if (someCond) {
    handle = fp_c2.data[24].z | fp_c2.data[88].x;
}
texture(handle, vec2(x, y));
```
Because `handle` has two assignment, and may have different values depending on the path it takes, it was not able to find the handle value. But the value is actually the same regardless of the path it takes, since constant buffers can't be modified inside the shader, and they both access the same values from the constant buffer, and perform the same operation.

To fix this I extended the code to also match this case and pick the first value.

Fixes some shadow issues on Super Mario RPG.
Before:
image
After:
image
Fixes #5946.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
+++ b/src/Ryujinx.Graphics.Gpu/Shader/DiskCache/DiskCacheHostStorage.cs
@@ -22,7 +22,7 @@ namespace Ryujinx.Graphics.Gpu.Shader.DiskCache
         private const ushort FileFormatVersionMajor = 1;
         private const ushort FileFormatVersionMinor = 2;
         private const uint FileFormatVersionPacked = ((uint)FileFormatVersionMajor << 16) | FileFormatVersionMinor;
-        private const uint CodeGenVersion = 5791;
+        private const uint CodeGenVersion = 5957;
 
         private const string SharedTocFileName = "shared.toc";
         private const string SharedDataFileName = "shared.data";
--- a/src/Ryujinx.Graphics.Shader/Translation/Optimizations/BindlessElimination.cs
+++ b/src/Ryujinx.Graphics.Shader/Translation/Optimizations/BindlessElimination.cs
@@ -55,7 +55,7 @@ namespace Ryujinx.Graphics.Shader.Translation.Optimizations
                         continue;
                     }
 
-                    if (bindlessHandle.AsgOp is not Operation handleCombineOp)
+                    if (!TryGetOperation(bindlessHandle.AsgOp, out Operation handleCombineOp))
                     {
                         continue;
                     }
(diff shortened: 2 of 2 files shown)


riperiperi (approved):
Nice catch, weird that it generates code like this.

PR #5967, opened 2023-11-20
Centralize TargetFramework management in Directory.Build.props

marco-carvalho:
This PR centralizes the `TargetFramework` setting for the projects using `Directory.Build.props`. It sets a default framework (`net8.0`), streamlining project consistency and maintenance. The remaining projects needing different frameworks (`netstandard2.0`) have explicit overrides in their project files.

Centralizing framework settings reduces discrepancies and simplifies updates, and should be easier to manage and update framework versions across multiple projects.


zcsizmadia:
@marco-carvalho I would suggest to consider the following pattern below as well. This should apply for `netstandard2.x` targets as well. One con for setting `TargetFramework` in `Directory.Build.props` is that some projects are still using `netstandard`.  With the below pattern, other target framework can be forced by running e.g. `dotnet build -p:DefaultTargetFramework=net7.0`

Additionally setting `LangVersion` and other commonly used props in `Directory.Build.props` make sense as well. 

In `Directory.Build.props`:
```
<PropertyGroup>
  <DefaultTargetFramework Condition="'$(DefaultTargetFramework)' == ''">net8.0</DefaultTargetFramework>
  <DefaultNetstandardTargetFramework Condition="'$(DefaultNetstandardTargetFramework)' == ''">netstandard2.0</DefaultNetstandardTargetFramework>
</PropertyGroup>
```

In a `.csproj` file:
```
<PropertyGroup>
  <TargetFramework>$(DefaultTargetFramework)</TargetFramework>
</PropertyGroup>
```

zcsizmadia:
> DefaultTargetFramework

se someone can go crazy overboard and set the `DefaultTargetFramework` in `Directory.Build.props` by regexing `global.json`. In that case when NET 9 comes out, the only change needed is changing the `global.json` file :)

marco-carvalho:
@zcsizmadia I updated with your feedback.

zcsizmadia:
What else might be usefule to define in `Directory.Build.props`? Possible candidates:

- `<DefaultRuntimeIdentifiers Condition="'$(DefaultRuntimeIdentifiers)' == ''">win-x64;osx-x64;linux-x64</DefaultRuntimeIdentifiers>`
- <LangVersion>latest</LangVersion> (maybe with a condition of TargetFrameworks is netstandard?)

marysaka (approved):

marco-carvalho:
I don't know if there is any benefit in adding `<LangVersion>latest</LangVersion>` to `Directory.Build.props`.
According to [this documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults), I understood that if you update the project to the latest version, the `LangVersion` will update accordingly.
image

zcsizmadia:
If you use the latest net8.0 framework, all the net8 projects will enable C# 12 lang features, however if you use netstandard as your  target framework, the used deault language will be only 7.3 even you have the latest NET SDK installed.

https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Ui.LocaleGenerator/Ryujinx.Ui.LocaleGenerator.csproj#L6 is the only neststandard project out of the 3 netstandard projects.

To be consistent IMO all projects should use the default language version, or all should use the same (latest in this case). 

The 3 non default language features in that project are, and those can easily be reverted to default lang features (and how to revert them).

1.  https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Ui.LocaleGenerator/LocaleGenerator.cs#L12
Use of static in LINQ lambda

2. https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Ui.LocaleGenerator/LocaleGenerator.cs#L14
Use of ! (null forgiving) operator

3. https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Ui.LocaleGenerator/LocaleGenerator.cs#L19
Use of new()

The reasoning behind using latest in netstandard project is that developers using the latest language switch to use those features regularly, so when you work on a netstandard project which is relatively old language features, it becomes annoying that you cannot use the cool new language features, which you are used to and very much simplifies your life :)

zcsizmadia:
Aaaaargh.. Of course the latest language now will enable some code analyzer/Intellisense messages, which makes the dotnet format step fail... 

Probably it is better not to open that can of worms and deal with that in a separate PR instead of now and just not set LangVersion for now. If someone really need  C# 12 language features later, it will be easy to add to the Build props then.

zcsizmadia:
Sorry for making the case for it and tell you lets not do it ;)

marco-carvalho:
@zcsizmadia  I managed to deal with the analyzers/style formatters, but the macOS perform checks are generating an `CS0535 does not implement interface member 'IServiceObject.GetCommandHandlers()'` error, so I'm gonna revert the branch 'til @marysaka approved.

Issue #4716, opened 2023-04-24
LDN upstreaming roadmap


TSRBerry:
It's been a few years since the LDN build was created and in order to open source it completely, we need to upstream all the changes in smaller pieces.

I'm mostly creating this issue to be more transparent about the progress we've been making and how much still needs to be done. It will also help to keep me accountable, so I keep working on LDN and don't prioritize other issues.

## Website

- [ ] Separate the website from the LDN server (status: **In progress**)
- [ ] Get the website ready and open source it (status: **To do**)

## Sockets service

- [ ] Proxy socket support (required for RyuLdn)

## LDN service

- [ ] Change the RyuLdn protocol to use UDP and increment the version (status: **In progress**)
    - (This is the blocking issue we are currently working on internally and is also the reason this is taking so long.)
- [ ] `IUserLocalCommunicationService` (ldn:u) implementation
    - This will also contain the `INetworkClient` interface and the `LdnDisabledClient` implementation

### Network clients

- [ ] ldn_mitm
    - Support for the [ldn_mitm](https://github.com/spacemeowx2/ldn_mitm) sysmodule by @spacemeowx2
- [ ] RyuLdn
    - Our own LDN network client implementation

---

There are also some refactorings and other small things which aren't listed here, but this is usually something we do while we get the upstream PRs ready.

PR #5989, opened 2023-11-26
bsdsockets: Add proxy socket support

TSRBerry:
This PR is part of our ongoing effort to upstream our LDN implementation (#4716).

The new `ManagedProxySocket` allows SOCKS5 compatible proxy servers to intercept and manipulate traffic from the guest.

I implemented many parts of this according to [RFC1928 - SOCKS5 Protocol](https://datatracker.ietf.org/doc/html/rfc1928), but since there are some things I didn't do, this isn't a compliant implementation. I added a TODO comment for everything that needs to be implemented to be SOCKS5 compliant.

I tested the TCP connect part of this implementation using @Raytwo's Cobalt mod which connects to the GitHub API for a version check and mitmproxy. Sadly mitmproxy does not support UDP for SOCKS5, so this part still needs to be tested.

There are a few other things I'm not sure about where feedback and/or testing would be nice:
- [ ] Does proxying UDP packets actually work?
- [ ] Does the TCP proxy work in server mode (using the `Bind` command)?
- [ ] For `LocalEndPoint` and `RemoteEndPoint` I could use the endpoints I receive from the proxy server like I do now, or use the actual endpoints from the sockets directly. Which one is better? (Both could work, don't consider RyuLDN for this question, since it doesn't matter for the way I want to add RyuLDN with this proxy implementation)
- [ ] Should I throw exceptions for proxy responses containing errors or should I convert them to `LinuxError`s somehow?
- [ ] Is the way the `Handle` property works fine?
- [ ] Should I provide an implementation for `RecvMMsg` and `SendMMsg`?

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
@@ -105,7 +105,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
             byte[] requestData = new byte[Marshal.SizeOf<TReq>()];
             byte[] responseData = new byte[Marshal.SizeOf<TResp>() + (_proxyAuth?.WrapperLength ?? 0)];
 
-            MemoryMarshal.Write(requestData, ref request);
+            MemoryMarshal.Write(requestData, request);
 
             int expectedSentBytes;
             int sentBytes;
@@ -353,7 +353,7 @@ namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
                 DestinationPort = (ushort)remoteEndPoint.Port,
             };
 
-            MemoryMarshal.Write(data, ref header);
+            MemoryMarshal.Write(data, header);
             buffer[..size].CopyTo(data.AsSpan()[Marshal.SizeOf<SocksUdpIpv4Header>()..]);
 
             try
--- a/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
@@ -1,0 +1,730 @@
+using Ryujinx.Common.Logging;
+using Ryujinx.Common.Memory;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Proxy;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Proxy.Auth;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Proxy.Packets;
+using Ryujinx.HLE.HOS.Services.Sockets.Bsd.Types;
+using System;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Security.Authentication;
+
+namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
+{
+    class ManagedProxySocket : ISocket
+    {
+        private static readonly IPEndPoint _endpointZero = new(IPAddress.Any, 0);
+        private readonly EndPoint _proxyEndpoint;
+
+        private IProxyAuth _proxyAuth;
+        private bool _ready;
+
+        private IPEndPoint _udpEndpoint;
+        private Socket _udpSocket;
+
+        public Socket Socket { get; private set; } = new(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+
+        public int RefCount { get; set; }
+
+        public AddressFamily AddressFamily { get; }
+        public SocketType SocketType { get; }
+        public ProtocolType ProtocolType { get; }
+
+        public bool Blocking {
+            get
+            {
+                return _udpEndpoint != null ? _udpSocket.Blocking : Socket.Blocking;
+            }
+
+            set
+            {
+                if (_udpEndpoint != null)
+                {
+                    _udpSocket.Blocking = value;
+                }
+                else
+                {
+                    Socket.Blocking = value;
+                }
+            }
+        }
+
+        public IntPtr Handle => _udpEndpoint != null ? _udpSocket.Handle : Socket.Handle;
+
+        public IPEndPoint RemoteEndPoint { get; private set; }
+        public IPEndPoint LocalEndPoint { get; private set; }
+
+        public ManagedProxySocket(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType, EndPoint proxyEndpoint)
+        {
+            AddressFamily = addressFamily;
+            SocketType = socketType;
+            ProtocolType = protocolType;
+            _proxyEndpoint = proxyEndpoint;
+            RefCount = 1;
+        }
+
+        private ManagedProxySocket(ManagedProxySocket oldSocket)
+        {
+            AddressFamily = oldSocket.AddressFamily;
+            SocketType = oldSocket.SocketType;
+            ProtocolType = oldSocket.ProtocolType;
+            LocalEndPoint = oldSocket.LocalEndPoint;
+            RemoteEndPoint = oldSocket.RemoteEndPoint;
+            _proxyEndpoint = oldSocket._proxyEndpoint;
+            Socket = oldSocket.Socket;
+            RefCount = 1;
+        }
+
+        #region Proxy methods
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void EnsureVersionIsValid(byte version)
+        {
+            if (version != ProxyConsts.Version)
+            {
+                throw new InvalidDataException($"Invalid proxy version: {version}");
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void EnsureSuccessReply(ReplyField replyField)
+        {
+            if (replyField != ReplyField.Succeeded)
+            {
+                throw new ProxyException(replyField);
+            }
+        }
+
+        private TResp SendAndReceive<TReq, TResp>(TReq request)
+            where TReq: unmanaged
+            where TResp: unmanaged
+        {
+            byte[] requestData = new byte[Marshal.SizeOf<TReq>()];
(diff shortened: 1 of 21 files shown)


riperiperi:
>  Does proxying UDP packets actually work?

Why is this an unknown? This is what nearly all LDN games require.

TSRBerry:
I just rebased this PR and replaced the old SOCKS5 implementation with [RyuSOCKS](https://www.nuget.org/packages/RyuSocks)!

I also adjusted the description accordingly and listed all the things that need to be done, before I mark this PR as ready to review.

I expect to finish this todo list within the next few weeks.

TSRBerry:
Seems like the stream wrapper in `RyuSocks.SocksClientStream` isn't working correctly yet.

I just repeated the mitmproxy test with Cobalt and successfully froze Ryujinx before the request even reached the proxy server.

Once I figure out how to fix that, I'll be able to mark this PR as ready to review again.

TSRBerry:
Finally ready for review! :D

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocketPollManager.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Collections.Generic;
+using System.Net.Sockets;
+
+namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Impl
+{
+    class ManagedProxySocketPollManager : IPollManager
+    {
+        private static ManagedProxySocketPollManager _instance;
+
+        public static ManagedProxySocketPollManager Instance
```
gdkchan, line 12: I'm not a fan of those static instances, but it seems the other poll managers do the same, so it should be fine for now.

TSRBerry: I suppose that's something we could tackle once we move this to `Ryujinx.Horizon` as well?

gdkchan: Yes.

gdkchan (approved):
lgtm, thanks. Should wait for @riperiperi review before it is merged, I did not work on LDN so I'm not the best person to review this.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Proxy/ProxyManager.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Sockets;
+using System.Runtime.CompilerServices;
+
+namespace Ryujinx.HLE.HOS.Services.Sockets.Bsd.Proxy
+{
+    public static class ProxyManager
+    {
+        private static readonly Dictionary<string, EndPoint> _proxyEndpoints = new();
```
riperiperi, line 11: Maybe this dictionary should be thread safe? I can imagine other threads adding Endpoints and maybe multiple threads making clients at once, though I don't think the BSD service can do that right now.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
```diff
@@ -0,0 +1,472 @@
...
+            {
+                Blocking = false;
+                shouldBlockAfterOperation = true;
+            }
+
+            try
+            {
+                EndPoint temp = new IPEndPoint(IPAddress.Any, 0);
+
+                receiveSize = ProxyClient.ReceiveFrom(proxyBuffer, WinSockHelper.ConvertBsdSocketFlags(flags), ref temp);
```
riperiperi, line 177: Is this missing a copy to `buffer`? Its data is never updated.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
```diff
@@ -0,0 +1,472 @@
...
+
+            try
+            {
+                receiveSize = ProxyClient.Receive(
+                    proxyBuffer,
+                    WinSockHelper.ConvertBsdSocketFlags(flags),
+                    out SocketError errorCode
+                );
+
+                proxyBuffer[..receiveSize].CopyTo(buffer);
```
riperiperi, line 131: Is the wrapper at the end of the buffer? If it's at the start, this should be `proxyBuffer[^recieveSize..]`.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
```diff
@@ -0,0 +1,472 @@
...
+            {
+                return WinSockHelper.ConvertError((WsaError)exception.ErrorCode);
+            }
+
+            return LinuxError.SUCCESS;
+        }
+
+        public LinuxError Connect(IPEndPoint remoteEndPoint)
+        {
+            ProxyClient.RequestCommand = ProxyCommand.Connect;
```
riperiperi, line 377: Why do these commands need to be set on a property if a method specific to them is called later? (like `ProxyClient.Connect`)

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
```diff
@@ -0,0 +1,472 @@
...
+            {
+                return WinSockHelper.ConvertError((WsaError)exception.ErrorCode);
+            }
+
+            return LinuxError.SUCCESS;
+        }
+
+        public void Disconnect()
+        {
+            ProxyClient.Disconnect();
```
riperiperi, line 450: Should this (and connection errors) reset `ProxyClient.RequestCommand`? I can't remember if games end up reusing sockets after they disconnect, but in case it's possible it might need reset to prevent errors on some of the methods.

riperiperi (commented):
Looks good, I have a few questions about certain parts of the code. I don't know if this is supposed to be testable in any form right now, I guess I can see if LAN and ldn_mitm still work properly.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
TSRBerry: Currently it will be at the start, but this is up to the implementation of the authentication method and the proxy command. The `Receive()` call already performs the unwrapping and copies everything to the start of the buffer.
I only use `[..receiveSize]` here so it doesn't attempt to copy more data than necessary.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
TSRBerry: Yes it is, thank you.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
TSRBerry: Yeah, this part feels a little weird.

I was looking for a way to generate these command specific methods to avoid using a property for this, but as far as I know this would prevent people from adding/using their own commands and authentication methods, since the NuGet package only ships as a DLL and my code can't be modified anymore at that point.

Since I want people to be able to add commands and authentication methods to RyuSocks easily, I currently work around this problem by providing methods similar to `System.Net.Sockets.Socket` in `SocksClient` while still allowing them to be used with any provided proxy command.

I don't like how it currently works and I'm looking to improve it. Maybe I should provide a more general method that handles the proxy commands and remove the more specific ones?

Although I'd love to have something that could allow me to ship the RyuSOCKS NuGet package as source code and generate the methods with a source generator - in such a way that people could add an attribute or something to their own commands, so they also get picked up and handled by the source generator.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocket.cs
TSRBerry: Oh yeah, that's definitely safer. I'll do that.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocketPollManager.cs
```diff
@@ -0,0 +1,211 @@
...
+                    return LinuxError.EINVAL;
+                }
+            }
+
+            try
+            {
+                int actualTimeoutMicroseconds = timeoutMilliseconds == -1 ? -1 : timeoutMilliseconds * 1000;
+                int totalEvents = eventDict[SelectMode.SelectRead].Count + eventDict[SelectMode.SelectWrite].Count + eventDict[SelectMode.SelectError].Count;
+                // TODO: Maybe check all events first, wait for the timeout and then check the failed ones again?
+                int timeoutMicrosecondsPerEvent = actualTimeoutMicroseconds == -1 ? -1 : actualTimeoutMicroseconds / totalEvents;
```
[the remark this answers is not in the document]
TSRBerry: The only way I can think of to fix this would be to provide a `Socket.Select()` method from `RyuSOCKS`, which then calls `Socket.Select()` with the underlying sockets itself?

If you think this solution could work I'd implement it for this PR. Otherwise or if you don't think it's important I'll open an issue for this at the `RyuSOCKS` repo and work on that some time after completing LDN upstreaming.

src/Ryujinx.HLE/HOS/Services/Sockets/Bsd/Impl/ManagedProxySocketPollManager.cs
```diff
@@ -0,0 +1,211 @@
...
+            {
+                ManagedProxySocket socket = ((ManagedProxySocket)evnt.FileDescriptor);
+
+                PollEventTypeMask outputEvents = evnt.Data.OutputEvents & ~evnt.Data.InputEvents;
+
+                if (eventDict[SelectMode.SelectError].Contains(socket))
+                {
+                    outputEvents |= PollEventTypeMask.Error;
+
+                    // TODO: Check ProxyClient.Connected and ProxyClient.IsBound when implemented.
```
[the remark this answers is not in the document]
TSRBerry: Forgot about that - it's implemented now!

PR #6014, opened 2023-12-06
Vulkan: Use templates for descriptor updates

riperiperi:
This PR changes the descriptor updater to use descriptor update templates unique to each program. A descriptor update template describes all of the descriptor fields that will be updated at the same time, and how their values will be laid out in memory. Each template defines one subset of descriptors that will be updated, you can't omit or add any without creating another template. This makes sense for us, since we always update all descriptors that are used for a program when updating a descriptor set (as we don't keep track of its existing state).

The descriptors are assumed to update in segment order, and be back to back in memory. There's a simple class for pushing data into a native array that can then be submitted as part of a descriptor update.

The old code for non-templated updates is still present as a VulkanConfiguration switch, though it should be supported by all Vulkan 1.1 conformant drivers. I originally left this in in case the change to how null buffer textures are handled breaks anything and needs templates to be disabled... I guess it's worth testing UE4 games across a variety of platforms to see if that difference causes any issues.

I'm not sure if we should just remove the non-template path if this pans out. It would simplify DescriptorSetUpdater a little and remove a bunch of methods from DescriptorSetCollection.

## Improves
AMD Mesa (Linux)

## Very small improvment (benchmarks imply, at least)
NVIDIA

## Unknown
AMD (Windows)
Intel (Mesa/Windows)
Apple (MacOS)

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
+++ b/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
@@ -77,7 +77,7 @@ namespace Ryujinx.Graphics.Vulkan
                     }
                     else
                     {
-                        // TODO: current normal impl likes to skip these if null?
+                        // NOTE: Current non-templated path skips these if null, we just send a 0 handle.
 
                         entries[entryI++] = new DescriptorUpdateTemplateEntry()
                         {
@@ -108,7 +108,7 @@ namespace Ryujinx.Graphics.Vulkan
                     }
                     else
                     {
-                        // TODO: current normal impl likes to skip these if null?
+                        // NOTE: Current non-templated path skips these if null, we just send a 0 handle.
 
                         entries[entryI++] = new DescriptorUpdateTemplateEntry()
                         {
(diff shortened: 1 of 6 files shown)


src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
```diff
@@ -0,0 +1,149 @@
...
+                            DescriptorCount = (uint)count,
+                            Offset = structureOffset,
+                            Stride = (nuint)Unsafe.SizeOf<DescriptorImageInfo>()
+                        };
+
+                        structureOffset += (nuint)(Unsafe.SizeOf<DescriptorImageInfo>() * count);
+                    }
+                    else
+                    {
+                        // NOTE: Current non-templated path skips these if null, we just send an empty handle.
```
gdkchan, line 76: I think if the old code is kept, it should be updated too to pass a null handle instead of not updating the entry (assuming that it is valid). If the old code is removed, the comment can be removed too.

riperiperi:
I profiled on NVIDIA, and there seems to be a _very small_ improvement. Like, 19.80% to 19.20% backend time. The majority of the cost there seems to be paid on bind. Push descriptors could work around this, though there would be a lot to fix up to get that working.

I have not yet tested on other platforms.

riperiperi:
Updated - only templates can be used now, new template updater that uses spans when possible.

gdkchan (approved):
lgtm, thanks! Nice to see performance improvements for AMD GPUs on Linux.

marysaka (approved):

PR #6023, opened 2023-12-09
[macOS] Correctly set filetypes in Info.plist

shinra-electric:
Currently the filetype association is not working: 

Screenshot 2023-12-09 at 14 55 54

This should hopefully fix it. 

Testing: 

- [ ] Check that Finder correctly associates `nca`, `nro`, `nso`, `nsp` and `xci` files with Ryujinx. 
An example on how it should look: 
Screenshot 2023-12-09 at 13 16 20

### DIFF
--- a/distribution/macos/Info.plist
+++ b/distribution/macos/Info.plist
@@ -10,14 +10,25 @@
     <string>Ryujinx</string>
     <key>CFBundleIconFile</key>
     <string>Ryujinx.icns</string>
-    <key>CFBundleTypeExtensions</key>
-    <array>
-        <string>nca</string>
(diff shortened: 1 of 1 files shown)


shinra-electric:
Seems to work as expected: 

Screenshot 2023-12-09 at 17 12 30

It would be preferable when a file is double-clicked that Ryujinx would: 
- Check if the file is in the games list
- If not, add it
- Run the game

But this functionality could come in a future PR.

shinra-electric:
I've found an example for implementing the open function in Qt: 
https://doc.qt.io/qt-6/qfileopenevent.html#macos-example

PR #6029, opened 2023-12-11
Input: Improve controller identification

riperiperi:
Controllers were identified before by a combination of their _global_ index in the list of controllers and their GUID. The problem is, disconnecting and reconnecting a controller can change its global index; the controller can appear at the end when it was originally at the start. This would give it another ID, and the controller would need to be reconfigured.

This happened to me a lot with a switch pro controller and a USB game controller, it was essentially random which appeared first, and sometimes I had to pull cables or reconfigure the controllers. Now, it consistently detects them.

This PR changes the controller identification to be a combination of an index of controllers with the same GUID (generally 0), and its GUID. It also reworks managing the list of controllers to properly consider instance IDs.

This also changes the NpadManager to attempt to reuse old controllers when refreshing input configuration, which can prevent input from going dead for seconds whenever an unrelated controller connects or disconnects (and the switch pro controller just entirely dying).

Testing with different controller types, OS and Avalonia is welcome. Remember that the target is connecting a ton of controllers, and pulling/reconnecting them.

### DIFF
--- a/src/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
+++ b/src/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
@@ -138,7 +138,6 @@ namespace Ryujinx.Input.SDL2
                     OnGamepadDisconnected?.Invoke(id);
                 }
 
-
                 lock (_lock)
                 {
                     _gamepadsIds.Clear();
--- a/src/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
+++ b/src/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
@@ -9,8 +9,18 @@ namespace Ryujinx.Input.SDL2
     {
         private readonly Dictionary<int, string> _gamepadsInstanceIdsMapping;
         private readonly List<string> _gamepadsIds;
+        private readonly object _lock = new object();
 
-        public ReadOnlySpan<string> GamepadsIds => _gamepadsIds.ToArray();
+        public ReadOnlySpan<string> GamepadsIds
+        {
+            get
+            {
+                lock (_lock)
+                {
+                    return _gamepadsIds.ToArray();
+                }
+            }
+        }
 
         public string DriverName => "SDL2";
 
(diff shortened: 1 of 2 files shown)


willianholtz:
For example: I have 2 controls that I use here, the Pro on the Switch and one on the PS4 (and I rarely use it on the PS5), however, when I remove the control, and open the emulator again without the control being plugged in, the controller configuration control disappears and I have to set again which control I want to use. It would be interesting for the emulator to automatically identify the already plugged-in control, depending on which one I'm using, and reconfigure it on its own.

Below is a log of your PR, with 1 and 2 controls:
[Ryujinx_1.1.0+4cb79e4_2023-12-13_14-39-08.log](https://github.com/Ryujinx/Ryujinx/files/13663966/Ryujinx_1.1.0%2B4cb79e4_2023-12-13_14-39-08.log)

[Ryujinx_1.1.0+4cb79e4_2023-12-13_14-46-44.log](https://github.com/Ryujinx/Ryujinx/files/13663967/Ryujinx_1.1.0%2B4cb79e4_2023-12-13_14-46-44.log)

riperiperi:
Is part of your message missing? You just started it with "for example" and I don't really know what the anecdote has to do with these changes.

The expected behaviour is as follows:
- A specific controller is bound to Player 1
- A specific 2nd controller is bound to Player 2
- When the 2nd controller is disconnected, the first controller stays bound to P1. Player 2 is disconnected.
- When the 1st controller is disconnected, the second controller stays bound to P2. Player 1 is disconnected.
- If the controllers are plugged in a different order, then they stay bound to the same players and don't randomly unbind
- When a 3rd unbound controller is plugged or unplugged, players 1 and 2 should not be interrupted.

Anything that would change what player a specific controller is bound to depending on user preference would require an entirely different implementation of Input, where controllers would be configured without a player number and they press L+R to assign them with the controller applet similar to the switch. This is not in the scope of this PR.

willianholtz:
> Is part of your message missing? You just started it with "for example" and I don't really know what the anecdote has to do with these changes.
> 
> The expected behaviour is as follows:
> 
> * A specific controller is bound to Player 1
> * A specific 2nd controller is bound to Player 2
> * When the 2nd controller is disconnected, the first controller stays bound to P1. Player 2 is disconnected.
> * When the 1st controller is disconnected, the second controller stays bound to P2. Player 1 is disconnected.
> * If the controllers are plugged in a different order, then they stay bound to the same players and don't randomly unbind
> * When a 3rd unbound controller is plugged or unplugged, players 1 and 2 should not be interrupted.
> 
> Anything that would change what player a specific controller is bound to depending on user preference would require an entirely different implementation of Input, where controllers would be configured without a player number and they press L+R to assign them with the controller applet similar to the switch. This is not in the scope of this PR.

Dude, I apologize... As I use Google translate, I think I missed the beginning of the subject...

Now, I don't even remember what I wrote...

But it was something like RPCS3 does, when I have two controllers connected, and I don't use it in any other controller slot (players), the emulator automatically recognizes that I removed the controller and moves to the second one connected (if it is not set to player 2 or other players logically)

src/Ryujinx.Input.SDL2/SDL2GamepadDriver.cs
```diff
@@ -110,7 +138,11 @@ protected virtual void Dispose(bool disposing)
                     OnGamepadDisconnected?.Invoke(id);
                 }
 
-                _gamepadsIds.Clear();
+
```
gdkchan, line 141: Double empty line.

gdkchan (approved):
lgtm, but that is not really my code. I had this problem a few times when playing with more than one controller, it was a bit annoying.

Issue #6040, opened 2023-12-16
[Bug] JoyCon Stick configuration settings for Invert X / Y and Rotate 90 Degrees map to Physical Stick instead of Logical Stick


amurgshere:
### Description of the issue

It is not possible to configure an Xbox controller to act as a sole Right JoyCon using the Left Physical Stick, this is because the Invert X / Y and Rotate options for both Left and Right JoyCons map relate to the physical stick and not the logical stick.

### Reproduction steps

I want to use an Xbox controller as a single Right JoyCon
1) Menu -> Options -> Settings -> Input
2) Select Input Device as Xbox Controller
3) Select Controller Type as "JoyCon Right"
4) Map Logical (JoyCon) Right Stick to Xbox Controller Physical Left Stick
5) Check both "Invert Stick X" and "Invert Stick Y"
6) Check "Rotate 90 Clockwise"
7) Test with a game that supports JoyCon for single player, e.g. Super Mario Party
## Note that Inversion and Rotation of stick does not occur

However, if you map the Logical (JoyCon) Right Stick to Physical Right Stick, the inversion and rotation does occur

It appears as if the Invert X / Y and Rotate Options only apply to the Physical Xbox Right stick for Right JoyCon and Physical Xbox Left stick for Left JoyCon, where as these settings should refer to the logical JoyCon stick instead.

### Log file

[Ryujinx_1.1.1+0_2023-12-16_23-42-09.log](https://github.com/Ryujinx/Ryujinx/files/13692825/Ryujinx_1.1.1%2B0_2023-12-16_23-42-09.log)

### OS

MacOS 14.1.2

### Ryujinx version

1.1.110

### Game version

1.1.0

### CPU

M1 MacbookAir

### GPU

8 Core Apple GPU

### RAM

16GB

### Additional context?

This issue occurs because the Invert X / Y and Rotate settings are applied when getting the Physical Input value of the stick rather than after their values have been mapped to the Logical Stick as per this code here:
* https://github.com/Ryujinx/Ryujinx/blob/9864675a0bb74e79bef5cd4658d96cc00553c144/src/Ryujinx.Input.SDL2/SDL2Gamepad.cs#L346-L364

Suggest this code be moved to after the stick mapping has occurred, the mapping appears to occur on these lines when referencing the _stickUserMapping array:

* https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Input.SDL2/SDL2Gamepad.cs#L299-L300

[...]

Issue #5538, opened 2023-08-06
[Bug]The stick rotation does not take effects in "joycon side" mode if stick is oppsite one


dragonation:
### Description of the issue

Config "left sitck" to "right stick" in joycon left mode and make it rotate 90 clockwise. the rotation is not working
The same, if configing "right stick" to "left stick" in joycon right mode with rotations , it does not working either.

I don't know whether is a feature or bug, but i think it does not make sense.

And not tested in traditional version, only tested it in ava version.

### Reproduction steps

1. Change controller type to "JoyCon Left",
2. Set Left Stick to the "Right"
3. Turn on Rotate 90 Clockwise
4. Play a game, and see the direction of joystick, for example: mario party?

### Log file

I think this does not need logs...

### OS

Windows 11

### Ryujinx version

1.1.974

NotVeryAnonymous:
Issue still occurs on 1.1.1279, macOS.

Issue #4028, opened 2022-12-05
Single Right Joy Con issues on macOS 13+


Candygoblen123:
### What's the issue you encountered?

Because of SDL's use of GameController framework, single Joy Cons appear to Ryujinx as a "normal controller" with the stick mapped to the left stick rotated 90 degrees, the face buttons mapped to their counterparts rotated 90 degrees, SL and SR mapped to L and R, and L/R and ZL/ZR mapped to nothing at all. 

Dual Joy Cons are not affected by this issue, and work just fine.

In Ryujinx's input settings, it is impossible to map L/R or ZL/ZR to anything, it is impossible to rotate the stick on the right Joy Con 90 degrees (it is possible to on a left Joy Con), and the face buttons need to be remapped manually. 

### How can the issue be reproduced?

Pair a single Joy Con to a Mac running macOS 13+, open Ryujinx's input settings, select one of the 2 provided single Joy Con controllers listed, and try and map something. 

### Log file

[Ryujinx_ 1.1.0-macos1_2022-12-05_14-55-38.log](https://github.com/Ryujinx/Ryujinx/files/10157671/Ryujinx_.1.1.0-macos1_2022-12-05_14-55-38.log)

### Environment?

- Ryujinx version: 1.1.0-macos1
- Game version: N/A, but I tested with Pokèmon: Let's Go, Eevee v1.0.2
- System Specs:
  - OS: macOS 13.0.1 (22A400)
  - CPU: M1
  - GPU: M1
  - RAM: 8GB
- Applied Mods : No

### Additional context?

It seems that this was fixed in SDL v2.26.0, `Added the hint SDL_HINT_JOYSTICK_HIDAPI_VERTICAL_JOY_CONS to control whether Nintendo Switch Joy-Con controllers will be in vertical mode when using the HIDAPI driver`.

HomeStarRunnerTron:
[...]

PR #6041, opened 2023-12-16
#6040 Fix issues with Invert X / Y and Rotate when mapping different physical stick to JoyCon logical stick

amurgshere:
Ensure the Invert X / Y and Rotate stick configuration is used based on the logical mapped stick instead of the physical stick. This change addresses Bug #6040 .

### DIFF
--- a/src/Ryujinx.Input.SDL2/SDL2Gamepad.cs
+++ b/src/Ryujinx.Input.SDL2/SDL2Gamepad.cs
@@ -313,6 +313,33 @@ namespace Ryujinx.Input.SDL2
             return value * ConvertRate;
         }
 
+        private JoyconConfigControllerStick<GamepadInputId, Common.Configuration.Hid.Controller.StickInputId> GetLogicalJoyStickConfig(StickInputId inputId) 
+        {
+            switch(inputId) 
+            {
+                case StickInputId.Left:
+                    if (_configuration.RightJoyconStick.Joystick == Common.Configuration.Hid.Controller.StickInputId.Left)
+                    {
+                        return _configuration.RightJoyconStick;
+                    }
+                    else
+                    { 
+                        return _configuration.LeftJoyconStick;
+                    }
+                case StickInputId.Right:
+                    if (_configuration.LeftJoyconStick.Joystick == Common.Configuration.Hid.Controller.StickInputId.Right)
+                    {
+                        return _configuration.LeftJoyconStick;
+                    } 
+                    else
+                    {
+                        return _configuration.RightJoyconStick;
+                    }
+            }
+
+            return null;
+        }
+
         public (float, float) GetStick(StickInputId inputId)
         {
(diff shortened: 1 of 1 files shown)


NotVeryAnonymous:
@AcK77 For what it's worth (coming from someone that doesn't know the codebase at all), I did brief testing on the macOS GUI test build for this, and it appears to work. I believe this also might fix or at least be related to #6469 #6040 #5538. Also, #4028 may be sorta related (at least the parts about rotating/inverting). 

I could play Mario Party on my Dualsense by doing right joy con, but using left stick and remapping the ABXY and SL/SR to make sense! Crucially for this setup, the rotate 90º and invert X/Y all worked! (Also tested using the right stick on a left joy con setting)

KaKen0:
The 90° Rotation for Player 2 with the Seperate JoyCon Selection still doesnt work, if anything else then the Auto Assigned Right Stick is Selected/Configured. And for the love of everything please make the "Number of Controller Inputs dont Match" a Checkeable thing to never show up again.

amurgshere:
> The 90° Rotation for Player 2 with the Seperate JoyCon Selection still doesnt work, if anything else then the Auto Assigned Right Stick is Selected/Configured. And for the love of everything please make the "Number of Controller Inputs dont Match" a Checkeable thing to never show up again.

Could you please reference another issue number with more detailed instructions or raise one and let me know for this as I would love to look into this further.

amurgshere:
@KaKen0 , I haven’t been able to reproduce your issue above, are you able to supply some more detailed steps to reproduce please so that I can reproduce and hopefully fix? Thank you.

src/Ryujinx/UI/ViewModels/Input/InputViewModel.cs
```diff
@@ -439,6 +438,24 @@ private static string GetShortGamepadId(string str)
...
         public void LoadDevices()
         {
+            string GetGamepadName(IGamepad gamepad, int controllerNumber)
+            {
+                return $"{GetShortGamepadName(gamepad.Name)} ({controllerNumber})";
+            }
+
+            string GetUniqueGamepadName(IGamepad gamepad, ref int controllerNumber)
+            {
+                var name = GetGamepadName(gamepad, controllerNumber);
```
amurgshere, line 448: NOTE: Could potentially remove the need for the additional GetGamepadName function and simply recursively call into GetUniqueGamepadName, but figured we wouldn't want a recursive call just in case...

amurgshere:
Thanks @KaKen0 ,

Ahh that explains it, these changes have not been merged into the mainline yet, so you would have to use a version from the build of this Pull Request for it to work, as the issue you are seeing is the exact same issue that my fix rectifies. The builds for this pull request are no longer available (404 Error) so you will need to wait until a new build for the Pull Request is complete then download the link from that build and try that.

PR #6043, opened 2023-12-17
Change macOS icon to match Big Sur style

### LINKED ISSUE #5939: Redesign icon for macOS[Feature Request]
### Overview

macOS has standardized icons starting from Bigsur. 
https://developer.apple.com/design/human-interface-guidelines/app-icons
Ryujinx should design new icons for macOS. The current icon look a bit abrupt compared with others.
image
Would it be better to change it to this : )
image

### Smaller details

None

### Nature of request

None

None

vanelizarov:
It looks cleaner when added to Dock:
SCR-20231217-eshi

Applications list:
image

and Launchpad:
image

Fix #5939


AcK77:
Tbh if we choose to change our current macOS icon,
I’d prefer to use https://discord.com/channels/410208534861447168/412730623405457408/1182129922240684043

image

Which looks better to me, but it’s a personal choice here. That needs to be discussed even if it’s not our priority!
In any cases, thanks for your work and your suggestion.

iMonZ:
> Tbh if we choose to change our current macOS icon, I’d prefer to use https://discord.com/channels/410208534861447168/412730623405457408/1182129922240684043
> 
> image
> 
> Which looks better to me, but it’s a personal choice here. That needs to be discussed even if it’s not our priority! In any cases, thanks for your work and your suggestion.

Wow this looks amazing!
@AcK77 When can we switch to your Logo?

AcK77:
After discussion between team members, we will not change the current macOS icon to the one suggested (or the one I had suggested too), further discussion is needed on our Discord.
In that way I close this PR. Thanks for your contribution.

PR #6068, opened 2023-12-28
Revert Apple hypervisor force ordered memory change

gdkchan:
This reverts the change made on #5898 since there has been a increasing number of reports of some games freezing very often after this change. I'm not sure if that was caused by this change as those games already had problems on hypervisor before that change (AFAIK?), but they do seem to happen much more frequently now.

It does mean that BOTW will be back to the previous state of crashing/freezing almost immediately, but it was not playable anyway due to constant freezes. #6057 should generally improve the situation for playing games with hypervisor disabled, which is the current solution for games that don't work properly with hypervisor.

Left the `Reprotect` code since it will be useful in the future.


AcK77 (approved):

gdkchan:
Would be nice to get a confirmation from someone with the issue that this does indeed fixes it. I tried to test with Pokémon Scarlet, but I can't get it to freeze in `master` after a short test.

fann1993814:
I test `Grand Theft Auto: San Andreas – The Definitive Edition` can launch.
In https://github.com/Ryujinx/Ryujinx/pull/5898 PR, the game cannot launch .

Shihta:
For Super Mario RPG, both master (tag: 1.1.1102) and master + this PR will suffer from freezing randomly in my environment, but I don't think it's happening very often.

I would play it for about 30 minutes at a time. for the master branch, maybe two or three times I would have a freeze issue.
after apply this PR, I've only played it twice, the first time it freeze, but the second time it doesn't.

However, I forget to keep the log...
I will try to remember to keep the log the next time I encounter it.

MutantAura:
Tested RDR1, Legends Arceus and Violet (three main games reported).

RDR1
Master: Consistent freeze within 5-10 mins of play.
PR: Couldn't get it to freeze.

PLA: Couldn't get a freeze in either.

Violet
Master: Freezes on average (across 3 runs) between 20 - 25mins
PR: Freezes on average (across 3 runs) between 40 - 50 mins

As both PKMN games I think have had random HV freezes since day 1 it's hard to compare, RDR1 however is night and day.

Shihta:
2 logs are as the follows:

[Ryujinx_1.1.1+0_2024-01-12_20-14-49.log](https://github.com/Ryujinx/Ryujinx/files/13927168/Ryujinx_1.1.1%2B0_2024-01-12_20-14-49.log)
[Ryujinx_1.1.1+0_2024-01-12_20-05-02.log](https://github.com/Ryujinx/Ryujinx/files/13927169/Ryujinx_1.1.1%2B0_2024-01-12_20-05-02.log)

marysaka (approved):

PR #6101, opened 2024-01-09
Fix vertex buffer size when switching between inline and state draw parameters

### LINKED ISSUE #5974: [Bug] Citizens of Space and Citizens Unite Earth x Space only show half a screen
### Description of the issue

during gameplay only half a screen is shown 

### Reproduction steps

start the game 

### Log file

ryujinx_capture_2023-11-22_16-05-39
ryujinx_capture_2023-11-22_16-07-27

### OS

Windows 10

### Ryujinx version

1.11092

### Game version

1.0

### CPU

Ryzen 5 5700X

### GPU

NVIDIA RTX 2060

### RAM

16gb

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
Some draw commands have passes the parameters (such as vertex count as first vertex) "inline", and does not modify 3D engine state. This currently causes a problem, where the vertex buffer sizes are calculated using the new inline vertex count, but when a draw using 3D engine state happens, it does not detect the count as being modified (since the inline draw did not modify it), and keeps using the previous vertex buffer size.

This PR fixes the issue by forcing a vertex buffer update when switching between inline and state parameter draws.

Fixes rendering on Citizens of Space (it was using the wrong vertex buffer size for a specific draw, and last vertex was 0 as a result).

Before:
image
After:
image

Fixes #5974.
Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -1,4 +1,4 @@
-﻿using Ryujinx.Common.Logging;
+using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Engine.Threed.Blender;
 using Ryujinx.Graphics.Gpu.Engine.Types;
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -1,4 +1,4 @@
-using Ryujinx.Common.Logging;
+﻿using Ryujinx.Common.Logging;
 using Ryujinx.Graphics.GAL;
 using Ryujinx.Graphics.Gpu.Engine.Threed.Blender;
 using Ryujinx.Graphics.Gpu.Engine.Types;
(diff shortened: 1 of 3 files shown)


src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -260,6 +264,8 @@ public void Update()
             if (_drawState.DrawIndirect != _prevDrawIndirect)
             {
                 _updateTracker.ForceDirty(VertexBufferStateIndex);
+
+                _prevDrawIndirect = _drawState.DrawIndirect;
```
gdkchan, line 268: I think it was an oversight that this was not being updated, so I fixed it too. Probably a good idea to make sure games using indirect draws did not break somehow as a result of this change.

ayurai: can you give examples of games that use indirect draws?

gdkchan: Monster Hunter Rise, Bayonetta 3, Pokémon Scarlet/Violet, Ghosts n Goblins Resurrection, Mario Golf Super Rush (adventure mode), Zombie Army 4

riperiperi (approved):
Nice catch. Tested a few games that use indirect draw and they seem to be fine.

marysaka (approved):

PR #6110, opened 2024-01-13
CI: Dependabot Groups

IsaacMarovitz:
Add groups for certain dependencies:
- Avalonia
- Silk.NET
- OpenTK
- SixLabors
- NUnit

Changes PR limit from 5 to 10.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,7 +13,6 @@ updates:
 
   - package-ecosystem: nuget
     directory: /
-    open-pull-requests-limit: 5
     schedule:
       interval: daily
     labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -22,3 +22,22 @@ updates:
...
+          - "OpenTK*"
+      SixLabors:
+        patterns:
+          - "SixLabors*"
+      NUnit:
+        patterns:
+          - "NUnit"
+      Ryujinx:
+        patterns:
+          - "Ryujinx*"
```
TSRBerry, line 43: Not sure if this is a good idea, since there are a lot of different dependencies in that namespace.

TSRBerry (commented):
Lgtm, I'm just not sure about the Ryujinx group. I'll check that again with my next review.

.github/dependabot.yml
IsaacMarovitz: Fair enough

AcK77 (approved):

PR #6122, opened 2024-01-13
CI: Update to NUnit 4

IsaacMarovitz:
Supersedes #6116

Screenshot 2024-01-29 at 13 31 54

Data was collected over 10 different runs on both NUnit 4 & 3. NUnit 4 did not show an appreciably higher failure rate than NUnit 3, with the only fails occurring on macOS runners.

All runners apart from those on macOS saw a substantial mean time decrease between 1-3 minutes. macOS saw a mean time **increase** of around 3-4 minutes.

Notably, the standard deviation of the Windows, and Linux Release runners is significantly smaller, showing less variability across runs than NUnit 3.


IsaacMarovitz:
@TSRBerry Why has this been marked `do-not-merge`? I've upgraded all the tests (All passing).

TSRBerry:
This was addressed in a Discord conversation: https://discord.com/channels/410208534861447168/410432399500115968/1195917650522538105

However let me quickly summarize:

- The previous dependabot PRs are also marked as `do-not-merge` because we have been seeing random segfaults every few workflow runs.
- We are unable to reproduce this locally unfortunately and still need to take a deeper look at that.
- Although newer versions seem to cause this issue less and less, we still need to track it down and fix it properly. Otherwise we'll need to manually restart workflows for a lot of PRs, which is less than ideal.
- The changes of this PR aren't the issue here, it's just about the random segfaults that we couldn't investigate yet.

Issues like these are the reason why dependency updates should usually be left to maintainers.

>  If you really want to update dependency stuff or do rewrites of existing stuff, it is usually better to ask devs/maintainers beforehand, because there might be reasons why it wasn't already updated or rewritten. It just helps everyone to not waste time.

> [P]lease ask about these [kind of PRs] beforehand, so we can tell you if there is something blocking that we need to deal with first.

PR #6128, opened 2024-01-14
Move most of signal handling to Ryujinx.Cpu project

gdkchan:
This moves most of the signal handling code to the `Ryujinx.Cpu` project. It helps changes like #6057, when we need to install the signal handlers for other execution engines (like other JITs), we still need to set up ARMeilleure JIT cache just to map the signal handling code, and also reference ARMeilleure just to register the tracking ranges. This improves the situation by moving that code out of there. With this change, LightningJit should no longer need to initialize ARMeilleure JIT cache just to add the signal handler.

This can also be useful in the future if we decide to install other signal handler (like for thread suspension for example).

Testing is welcome, I only tested on Windows and macOS.


src/ARMeilleure/Signal/WindowsPartialUnmapHandler.cs
```diff
@@ -137,17 +157,6 @@ private static Operand EmitThreadLocalMapIntGetValuePtr(EmitterContext context,
             return context.Add(structsPtr, context.SignExtend32(OperandType.I64, offset));
         }
 
-#pragma warning disable IDE0051 // Remove unused private member
-        private static void EmitThreadLocalMapIntRelease(EmitterContext context, IntPtr threadLocalMapPtr, Operand threadId, Operand index)
```
gdkchan, line 141: Would be nice to get confirmation from @riperiperi that this is really some leftover, and can be deleted.

riperiperi: I think this method was part of the thread local map, but it's not used on the native side so it's OK to remove.

riperiperi (approved):
LGTM, tested on linux and seems fine. Much cleaner split out like this.

PR #6131, opened 2024-01-14
Change shader cache init wait method

gdkchan:
We current have a static field event on `ARMeilleure.Translator`, which is initially unsignaled, and we signal it once shader cache load is finished. This prevents the game from running before the shader cache has been loaded, but there are many problems with it:
- It is not reset consistently across UI, only GTK seems to reset it, which means that Avalonia probably doesn't wait for shader cache load on the 2nd launch onwards.
- Does not play nicely with multiple execution engines (as it only exists on ARMeilleure). For example, I think it currently does not wait for shader cache load when using hypervisor on macOS (but it would be hard to notice since GPU does not process commands until shader cache is loaded, so most games would get stuck until it is loaded anyway).
- It is bad design in general, CPU emulation should not need to concern itself with this. Plus the way how it is exposed (public static field) is also bad, does not work well with multiple translator instances, etc.

To remedy the situation I introduced a `WaitUntilGpuReady` method on the GPU context that waits until shader is loaded, and it is called from `ArmProcessContext.Execute` (which is execution engine agnostic, so it is always called regardless of which JIT/hypervisor is being used).

While I'm still not 100% happy with this solution, I believe it is better than what we had before, and should solve the existing problems. Eventually we might need a strategy to handle shader cache load for multiple processes.

Testing is welcome.

### DIFF
--- a/src/ARMeilleure/Translation/Translator.cs
+++ b/src/ARMeilleure/Translation/Translator.cs
@@ -57,9 +57,6 @@ namespace ARMeilleure.Translation
         private Thread[] _backgroundTranslationThreads;
         private volatile int _threadCount;
 
-        // FIXME: Remove this once the init logic of the emulator will be redone.
-        public static readonly ManualResetEvent IsReadyForTranslation = new(false);
-
         public Translator(IJitMemoryAllocator allocator, IMemoryManager memory, bool for64Bits)
         {
             _allocator = allocator;
@@ -105,8 +102,6 @@ namespace ARMeilleure.Translation
         {
             if (Interlocked.Increment(ref _threadCount) == 1)
             {
-                IsReadyForTranslation.WaitOne();
-
                 if (_ptc.State == PtcState.Enabled)
                 {
                     Debug.Assert(Functions.Count == 0);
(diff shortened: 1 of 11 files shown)


MysticalOS:
Curious, is this why game still stutters with a fully cached game on macOS? In fact, I disabled shader caching cause it made slow game loads for no gains.

I couldn't find any reduction in stuttering at all. I still got initial hitch every time a new object is seen for first time that session even with a full shader cache on TOTK (and i tried both pre loaded and my own locally built caches)

 i ended up just disabling caching for faster game loading since I always had to run in circles hitching a few minutes til it's all smooth anyways.

So I'm happy to do thorough testing of this on my M3 max to see if it improves situation

gdkchan:
> Curious, is this why game still stutters with a fully cached game on macOS? In fact, I disabled shader caching cause it made slow game loads for no gains.

No, it would still wait until the shader cache is loaded before rendering anything, which I noted on:
> but it would be hard to notice since GPU does not process commands until shader cache is loaded, so most games would get stuck until it is loaded anyway

Stuttering on macOS might be because it doesn't cache pipelines (it just caches the shader and the first pipeline state, and loads them). It could also be that there's something wrong with the shader cache and it is not working.

riperiperi (approved):
Looks considerably cleaner.

Stutters on mac are because programs -> pipelines relationship is not 1:1. For example, sometimes a program will be used for a subset of render targets, then later used for the full set. The shader will be recompiled the second time on some GPUs, while desktop ones typically take a fast path and reuse some existing code.

PR #6132, opened 2024-01-14
Vulkan: Cache delegate for EndRenderPass

riperiperi:
This prevents a small allocation each time this method is called. This is a top 3 SOH allocation during gameplay in most games, and eliminating it is pretty free.

Not much change expected. Biggest offender for SOH allocations is actually scheduler.


gdkchan (approved):
Looks fine, thanks.

Issue #3682, opened 2022-09-08
[Feature Request] Use VkDriverID instead of vendor name in the status bar for Vulkan.


MutantAura:
#### Overview:

Changing the usual "GPU: 'Vendor Name'" section of the bottom status bar with "GPU: 'Driver name'". In practise this solely affects AMD, Intel and exotic renderers on Linux when using Vulkan.

#### Smaller Details: 

Specification: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceDriverProperties.html
Silk.NET: https://github.com/dotnet/Silk.NET/blob/main/src/Vulkan/Silk.NET.Vulkan/Structs/PhysicalDeviceDriverProperties.gen.cs

#### Nature of Request:  

- Change

### Why would this feature be useful?

The Vulkan API provides a lot of detail on what sort of device is being used for rendering. While It is certainly useful to see what vendor Ryujinx is using it is even more helpful to see what driver is being used. On Windows most users would notice no difference at all as most are using NVIDIA, AMD or Intels own proprietary drivers but on Linux there are a number of driver ID's that Vulkan can distinguish between. 

Vulkan currently has the following discrete driver identifications:
```
VK_DRIVER_ID_AMD_PROPRIETARY
VK_DRIVER_ID_AMD_OPEN_SOURCE
VK_DRIVER_ID_MESA_RADV
VK_DRIVER_ID_NVIDIA_PROPRIETARY
VK_DRIVER_ID_INTEL_PROPRIETARY_WINDOWS
VK_DRIVER_ID_INTEL_OPEN_SOURCE_MESA
VK_DRIVER_ID_IMAGINATION_PROPRIETARY
VK_DRIVER_ID_QUALCOMM_PROPRIETARYVK_DRIVER_ID_ARM_PROPRIETARY
VK_DRIVER_ID_GOOGLE_SWIFTSHADER
VK_DRIVER_ID_GGP_PROPRIETARY
VK_DRIVER_ID_BROADCOM_PROPRIETARY
VK_DRIVER_ID_MESA_LLVMPIPE
 VK_DRIVER_ID_MOLTENVK
VK_DRIVER_ID_COREAVI_PROPRIETARY
VK_DRIVER_ID_JUICE_PROPRIETARY
VK_DRIVER_ID_VERISILICON_PROPRIETARY
VK_DRIVER_ID_MESA_TURNIP
VK_DRIVER_ID_MESA_V3DV
[...]

PR #6146, opened 2024-01-18
Use driver name instead of vendor name in the status bar for Vulkan.

etkramer:
This PR handles #3682 by displaying the more specific driver name in the status bar rather than only the device vendor. OpenGL has no such API, so there we just continue to display the vendor name.

In many cases such as Windows/NV the name displayed for the vendor and the driver are the same. Seeking testers on Linux to confirm.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
+++ b/src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
@@ -25,7 +25,6 @@ namespace Ryujinx.Graphics.Vulkan
             PhysicalDeviceFeatures = api.GetPhysicalDeviceFeature(PhysicalDevice);
 
             api.GetPhysicalDeviceProperties(PhysicalDevice, out var physicalDeviceProperties);
-
             PhysicalDeviceProperties = physicalDeviceProperties;
 
             api.GetPhysicalDeviceMemoryProperties(PhysicalDevice, out PhysicalDeviceMemoryProperties);
--- a/src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
+++ b/src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
@@ -26,7 +26,7 @@ namespace Ryujinx.Graphics.Vulkan
             PhysicalDeviceFeatures = api.GetPhysicalDeviceFeature(PhysicalDevice);
 
             PhysicalDeviceProperties physicalDeviceProperties = default;
-            PhysicalDeviceVulkan12Properties physicalDeviceVulkan12Properties = new()
+            PhysicalDeviceVulkan12Properties physicalDeviceVulkan12Properties = new() // We need this to retrive driver-related details.
             {
                 SType = StructureType.PhysicalDeviceVulkan12Properties
             };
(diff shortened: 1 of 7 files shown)


src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -24,8 +25,27 @@ public VulkanPhysicalDevice(Vk api, PhysicalDevice physicalDevice)
             PhysicalDevice = physicalDevice;
             PhysicalDeviceFeatures = api.GetPhysicalDeviceFeature(PhysicalDevice);
 
-            api.GetPhysicalDeviceProperties(PhysicalDevice, out var physicalDeviceProperties);
+            PhysicalDeviceProperties physicalDeviceProperties = default;
+            PhysicalDeviceVulkan12Properties physicalDeviceVulkan12Properties = new() // We need this to retrive driver-related details.
```
marysaka, line 29: I would prefer this be queried with ``VK_KHR_driver_properties`` extension instead and follow back to the old behavior otherwise.

I have plans to make Ryujinx compatible with Vulkan 1.1 + extensions

marysaka (changes requested):

src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
etkramer: Should be resolved now

marysaka (approved):

MutantAura:
I'd agree it's worth preferring vendor name on macOS as the driver name won't give the information that this change gives on other platforms, in fact it does the exact opposite.

src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -58,6 +58,31 @@ public VulkanPhysicalDevice(Vk api, PhysicalDevice physicalDevice)
...
+            PhysicalDeviceProperties2 physicalDeviceProperties2 = new()
+            {
+                SType = StructureType.PhysicalDeviceProperties2,
+                PNext = &physicalDeviceDriverProperties
+            };
+
+            api.GetPhysicalDeviceProperties2(PhysicalDevice, &physicalDeviceProperties2);
+
+            res = physicalDeviceDriverProperties;
+            return true;
```
AcK77, line 83:
```suggestion

            return true;
```

src/Ryujinx.Graphics.Vulkan/VulkanPhysicalDevice.cs
```diff
@@ -58,6 +58,31 @@ public VulkanPhysicalDevice(Vk api, PhysicalDevice physicalDevice)
 
         public bool IsDeviceExtensionPresent(string extension) => DeviceExtensions.Contains(extension);
 
+        public unsafe bool TryGetPhysicalDeviceDriverPropertiesKHR(Vk api, out PhysicalDeviceDriverPropertiesKHR res)
+        {
+            if (!IsDeviceExtensionPresent("VK_KHR_driver_properties"))
+            {
+                res = default;
+                return false;
```
AcK77, line 66:
```suggestion

                return false;
```

AcK77 (changes requested):

gdkchan (approved):
lgtm, thanks.

AcK77 (approved):

IsaacMarovitz:
macOS issue was not addressed prior to merging...

PR #6154, opened 2024-01-20
Vulkan: Use push descriptors for uniform bindings when possible

riperiperi:
Followup to https://github.com/Ryujinx/Ryujinx/pull/6014. Uses push descriptors for uniform buffers when possible - when it's not possible it just falls back to the regular descriptor set with templated updates.

Push descriptors are also using templated updates, since it allows us to submit multiple bindings in one call. This is useful when all the bindings change, or we have to resubmit the bindings due to a pipeline layout change. There's a dictionary lookup for the template within the pipeline layout, but there's a fast path if it's the same updated bindings as last time.

## Reserved bindings
There can only be 32 push descriptors, but we tend to have 37 uniform buffer bindings when fragment and vertex are both using uniforms. The chosen solution to this problem is to select a subset of uniforms that are "reserved" and skip them in the push descriptor set. If any shader programs use these reserved bindings, they must use the full layout and normal template update for their uniform set.

This set of reserved bindings was manually selected by observation after checking a few NVN games, to maximize the number of programs that can use push descriptors in popular games. However, guest OGL games tend to use different bindings, so there is an alternate mode that's activated if binding 3 is used on the first shader using push descriptors (typically not used on NVN). This mode is set for the remainder of the life of the VulkanRenderer, so it prefers whatever 

The reason we want one set of bindings for an entire run is so that bindings can continue to live between program bindings without being reset. This is the case as long as the new program doesn't use a different pipeline layout, so we get this for free on similar shader programs so long as the push descriptor arrangement is the same between them.

## Improvements
All screenshots here have been taken with a modified version of Ryujinx that's able to measure how much % time the backend thread is active. You can find that here: https://github.com/riperiperi/Ryujinx/tree/backend-statistics-bad . Measuring the backend time _can_ interfere with performance unlike FIFO% measurement, and the way it's done is kind of messy, so no PR for that.

Improvements depend on the game and the system you're using. Every PC and game combination could have a different bottleneck limiting your framerate, and could have different reactions to using push descriptors vs full set update + binds.

### NVIDIA
The NVIDIA driver is already wonderfully fast, but using push descriptors still shaves off an extra 10% or more depending on game. Unfortunately, it was so fast before that the difference isn't visible in FPS in my testing games, but CPU time is saved and framerate stability could be improved due to there being more breathing room in intense situations.

**Before:**
image

**After:**
image-1

### AMD (mesa, steam deck)
The mesa driver was most impacted by using descriptor set templates, and similarly is most impacted by switching to use push descriptors. In many games, the backend performance is actually the bottleneck on mesa, so improvements are seen as real FPS gain, as well as the reduced CPU usage and improvements to stability.

**Before:**
Screenshot_20240120_104313

**After:**
Screenshot_20240120_104604

### AMD (windows, steam deck)
The Windows AMD driver has consistently been faster than mesa in the hardest to run games, but that lead seems to be changing. While push descriptors do improve performance on this driver, it's not as much as with mesa.

**Before:**
image

**After:**
image-1

### MoltenVK
image

MoltenVK, uh... CPU time here is essentially doubled between VK and Metal threads, so I'm hoping that the push descriptor model better matching how games actually update uniform buffer bindings can reduce that significantly. Not that we can test that right now, since it just breaks completely. ~~Might be related to mirrors,~~ (still happens when they are disabled and doesn't happen when forcing them on NV) might be something else.

## TODO
This PR is draft until the following things are done:
- More game testing to ensure nothing breaks
- More performance testing + examples (ACNH has big gains in backend time from ~40% to ~32%, but ARM JIT execution is usually the bottleneck)
- Identify and fix issue with MoltenVK, or disable push descriptors on MoltenVK.
- Why does renderdoc break when updating more than one binding at a time in a single push descriptor template update entry?
- Think about compute shaders. They don't use push descriptors right now, but I don't think there's as much reason to use it. Draws can be heavily batched with a minimal set of uniform bindings changing between each, while compute is usually one and done or multiple different passes that need different data.
- Clean up magic numbers + implicit limits? (64 max binding index limit on push descriptor templates, though this can't activate right now)

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/Constants.cs
+++ b/src/Ryujinx.Graphics.Vulkan/Constants.cs
@@ -16,6 +16,7 @@ namespace Ryujinx.Graphics.Vulkan
         public const int MaxStorageBufferBindings = MaxStorageBuffersPerStage * MaxShaderStages;
         public const int MaxTextureBindings = MaxTexturesPerStage * MaxShaderStages;
         public const int MaxImageBindings = MaxImagesPerStage * MaxShaderStages;
+        public const int MaxPushDescriptorBinding = 64;
 
         public const ulong SparseBufferAlignment = 0x10000;
     }
--- a/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
+++ b/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
@@ -20,7 +20,13 @@ namespace Ryujinx.Graphics.Vulkan
         public readonly DescriptorUpdateTemplate Template;
         public readonly int Size;
 
-        public unsafe DescriptorSetTemplate(VulkanRenderer gd, Device device, ResourceBindingSegment[] segments, PipelineLayoutCacheEntry plce, PipelineBindPoint pbp, int setIndex)
+        public unsafe DescriptorSetTemplate(
+            VulkanRenderer gd,
+            Device device,
+            ResourceBindingSegment[] segments,
+            PipelineLayoutCacheEntry plce,
+            PipelineBindPoint pbp,
+            int setIndex)
         {
             _gd = gd;
             _device = device;
@@ -144,7 +150,14 @@ namespace Ryujinx.Graphics.Vulkan
             Template = result;
         }
 
-        public unsafe DescriptorSetTemplate(VulkanRenderer gd, Device device, ResourceDescriptorCollection descriptors, long updateMask, PipelineLayoutCacheEntry plce, PipelineBindPoint pbp, int setIndex)
+        public unsafe DescriptorSetTemplate(
+            VulkanRenderer gd,
+            Device device,
+            ResourceDescriptorCollection descriptors,
+            long updateMask,
+            PipelineLayoutCacheEntry plce,
+            PipelineBindPoint pbp,
+            int setIndex)
         {
             _gd = gd;
             _device = device;
@@ -156,7 +169,7 @@ namespace Ryujinx.Graphics.Vulkan
             int entry = 0;
             nuint structureOffset = 0;
 
-            void addBinding(int binding, int count)
+            void AddBinding(int binding, int count)
             {
                 entries[entry++] = new DescriptorUpdateTemplateEntry()
                 {
@@ -183,7 +196,7 @@ namespace Ryujinx.Graphics.Vulkan
                     {
                         if (bindingCount > 0 && (RenderdocPushCountBug || startBinding + bindingCount != binding))
                         {
-                            addBinding(startBinding, bindingCount);
+                            AddBinding(startBinding, bindingCount);
 
                             bindingCount = 0;
                         }
@@ -200,7 +213,7 @@ namespace Ryujinx.Graphics.Vulkan
 
             if (bindingCount > 0)
             {
-                addBinding(startBinding, bindingCount);
+                AddBinding(startBinding, bindingCount);
             }
 
             Size = (int)structureOffset;
--- a/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
+++ b/src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
@@ -1,12 +1,19 @@
 using Ryujinx.Graphics.GAL;
 using Silk.NET.Vulkan;
 using System;
+using System.Numerics;
 using System.Runtime.CompilerServices;
 
 namespace Ryujinx.Graphics.Vulkan
 {
     class DescriptorSetTemplate : IDisposable
     {
+        /// <summary>
+        /// Renderdoc seems to crash when doing a templated uniform update with count > 1 on a push descriptor.
+        /// When this is true, consecutive buffers are always updated individually.
+        /// </summary>
+        private const bool RenderdocPushCountBug = true;
+
         private readonly VulkanRenderer _gd;
         private readonly Device _device;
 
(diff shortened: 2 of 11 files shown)


amayra:
can you explain to me more why mesa driver is bottleneck here ?

riperiperi:
> can you explain to me more why mesa driver is bottleneck here ?

Just more CPU time spent in Vulkan method calls, and bigger gains to be made switching to templated descriptors or push descriptors. I'm not familiar with mesa codebase so I don't know the reason. It may be related to the large descriptor sets we use for bindings, where only a small subset of the bindings are used by each shader.

riperiperi:
Push descriptors are now disabled on MoltenVK, and the binding number limit of 64 is now properly handled. The MVK issue seems to happen when either texture or storage buffer resources are bound to the vertex stage, though I'm not sure why.

src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
```diff
@@ -137,6 +144,86 @@ public unsafe DescriptorSetTemplate(VulkanRenderer gd, Device device, ResourceBi
...
+            _device = device;
+
+            // Create a template from the set usages. Assumes the descriptor set is updated in segment order then binding order.
+            int segmentCount = BitOperations.PopCount((ulong)updateMask);
+
+            DescriptorUpdateTemplateEntry* entries = stackalloc DescriptorUpdateTemplateEntry[segmentCount];
+            int entry = 0;
+            nuint structureOffset = 0;
+
+            void addBinding(int binding, int count)
```
gdkchan, line 159:
```suggestion
            void AddBinding(int binding, int count)
```
That's how other local methods are named.

src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
```diff
@@ -137,6 +144,86 @@ public unsafe DescriptorSetTemplate(VulkanRenderer gd, Device device, ResourceBi
             Template = result;
         }
 
+        public unsafe DescriptorSetTemplate(VulkanRenderer gd, Device device, ResourceDescriptorCollection descriptors, long updateMask, PipelineLayoutCacheEntry plce, PipelineBindPoint pbp, int setIndex)
```
gdkchan, line 147: Parameter list is quite long here, would be nice to put each one on a separate line.

src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -400,6 +418,25 @@ private void SetupContext(GraphicsDebugLevel logLevel)
             _initialized = true;
         }
 
+        public int[] GetPushDescriptorReservedBindings(bool isOgl)
```
gdkchan, line 421:
```suggestion
        internal int[] GetPushDescriptorReservedBindings(bool isOgl)
```

src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
```diff
@@ -400,6 +418,25 @@ private void SetupContext(GraphicsDebugLevel logLevel)
...
+        {
+            // The first call of this method determines what push descriptor layout is used for all shaders on this renderer.
+            // This is chosen to minimize shaders that can't fit on 32 entry push descriptor sets.
+            if (_pdReservedBindings == null)
+            {
+                if (Capabilities.MaxPushDescriptors == 32)
```
gdkchan, line 427: Is it possible for that to be less than 32, or the spec ensures it will be at least 32? If it can be less, we should probably disable push descriptors in that case. Also if the limit is say, 33 we would still need some reserved bindings. I think it can only return the empty array if the limit is greater than or equal 36 (18 * 2).

src/Ryujinx.Graphics.Vulkan/Constants.cs
```diff
@@ -16,6 +16,7 @@ static class Constants
         public const int MaxStorageBufferBindings = MaxStorageBuffersPerStage * MaxShaderStages;
         public const int MaxTextureBindings = MaxTexturesPerStage * MaxShaderStages;
         public const int MaxImageBindings = MaxImagesPerStage * MaxShaderStages;
+        public const int MaxPushDescriptorBinding = 64;
```
gdkchan, line 19: Where does that limit come from? Is it something that the spec dictates as a fixed limit?

gdkchan (commented):
Looking good, the only thing that is not ideal is the bindings reservation part, since it sort of introduces a dependency on the specific usage of the bindings that the emulator currently does (for the reservations), but I don't know a good way to improve this.

gdkchan:
> the only thing that is not ideal is the bindings reservation part, since it sort of introduces a dependency on the specific usage of the bindings that the emulator currently does (for the reservations), but I don't know a good way to improve this.

I guess one way is changing the `ResourceLayout` it passes to the backend, to only have 32 bindings. Then the backend check would just be "if we have 32 bindings or less, we can use push descriptors".

src/Ryujinx.Graphics.Vulkan/Constants.cs
riperiperi: This limit is from the bitmask passed from DescriptorSetUpdater to identify push descriptor templates. There's only room for 64 bindings since it's a ulong.

src/Ryujinx.Graphics.Vulkan/DescriptorSetTemplate.cs
```diff
@@ -1,12 +1,19 @@
...
 
 namespace Ryujinx.Graphics.Vulkan
 {
     class DescriptorSetTemplate : IDisposable
     {
+        /// <summary>
+        /// Renderdoc seems to crash when doing a templated uniform update with count > 1 on a push descriptor.
+        /// When this is true, consecutive buffers are always updated individually.
+        /// </summary>
+        private const bool RenderdocPushCountBug = true;
```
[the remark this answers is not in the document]
riperiperi: I'll leave it for now, but that could be useful for other reasons... such as disabling forced command buffer submission based on time, as recording with renderdoc tends to make vk calls a lot slower so command buffers are split very frequently.

riperiperi:
> > the only thing that is not ideal is the bindings reservation part, since it sort of introduces a dependency on the specific usage of the bindings that the emulator currently does (for the reservations), but I don't know a good way to improve this.
> 
> I guess one way is changing the `ResourceLayout` it passes to the backend, to only have 32 bindings. Then the backend check would just be "if we have 32 bindings or less, we can use push descriptors".

It's not really that simple, since for optimal usage of push descriptors the binding layout shouldn't change between programs. Otherwise, you lose any potential optimisation where the program changes, but most of the bindings remain the same (and in the same binding locations). Right now, they can persist as long as the programs have matching pipeline layouts, but changing which 32 bindings are in the set will make it different every time.

gdkchan:
> > > the only thing that is not ideal is the bindings reservation part, since it sort of introduces a dependency on the specific usage of the bindings that the emulator currently does (for the reservations), but I don't know a good way to improve this.
> > 
> > 
> > I guess one way is changing the `ResourceLayout` it passes to the backend, to only have 32 bindings. Then the backend check would just be "if we have 32 bindings or less, we can use push descriptors".
> 
> It's not really that simple, since for optimal usage of push descriptors the binding layout shouldn't change between programs. Otherwise, you lose any potential optimisation where the program changes, but most of the bindings remain the same (and in the same binding locations). Right now, they can persist as long as the programs have matching pipeline layouts, but changing which 32 bindings are in the set will make it different every time.

What I meant is basically doing the same thing that is done on the backend right now, but on `ResourceLayout` creation. So it would still be "if binding X, Y and Z are unused, switch to layout that uses 32 bindings". It would just change where this is done and make the backend logic simpler (and `ResourceLayout` creation more complex).

BTW since this depends on guest API, it might be worth testing a game that uses Vulkan such as Super Mario Galaxy, but since those are very rare its not very important.

src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
riperiperi: `CanUsePushDescriptors` in `ShaderCollection` handles rejection of shaders with too many uniform bindings, so that should reject it if the host has like 16 limit and we need 32. I've changed this method to add the reserved bindings whenever it's less than 37 (two stages and the support buffer), though it'll likely not enable it if it's less than 32 max, maybe sometimes it will with vertex-only programs.

gdkchan (approved):
lgtm, thanks. I tested a few games here and found no issues on my RTX 3060. It's unfortunate that we need the workaround on older NVIDIA GPUs, and a shame that it's not usable on MoltenVK right now, but the performance improvement on other platforms is always appreciated.

The reserved arrays  change (from backend to GPU project) change I suggested above would be nice to have, but it can always be done at a later point.

marysaka (approved):

PR #6160, opened 2024-01-21
Fix integer overflow on downsample surround to stereo

### LINKED ISSUE #5275: [Regression] Audio peaking and crakling with SDL2
### Description of the issue

After https://github.com/Ryujinx/Ryujinx/pull/4908 audio in a lot of games (with SDL2 set as the audio backend) crackles.
Really noticable on SMT V with certain attacks. Can't remember any other right now but i've seen it in 4-5 games so far.

!!! AUDIO WARNING, LOWER THE VOLUME !!!

1.1.797

https://github.com/Ryujinx/Ryujinx/assets/2002038/1e31b510-6365-406a-b52e-1e9faad66ba3

Everything before 797

https://github.com/Ryujinx/Ryujinx/assets/2002038/232ad57d-230d-4394-b72d-a54938af1f2e

### Reproduction steps

Just do special attacks in SMT V

### Log file

None

### OS

Windows 10

### Ryujinx version

1.1.797

### Game version

1.3.0

### CPU

Ryzen 3600

### GPU

RTX 3060 Ti

### RAM

16 GB

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
This fixes two related on the surround to stereo downsampling code:
- While the result of a multiplication between 2 16-bit values will always fit on a 32-bit integer, adding 4 of such values can overflow. To fix that, I have changed the type of the coefficients from int to long, which will not overflow.
- Even if we fix the overflow above, for PCM16 output we still need a 16-bit value, and the previous code was truncating the value. This was fixed by clamping the value.

Fixes #5275.

### DIFF
--- a/src/Ryujinx.Audio/Backends/CompatLayer/Downmixing.cs
+++ b/src/Ryujinx.Audio/Backends/CompatLayer/Downmixing.cs
@@ -31,7 +31,7 @@ namespace Ryujinx.Audio.Backends.CompatLayer
         private const int Minus6dBInQ15 = (int)(0.501f * RawQ15One);
         private const int Minus12dBInQ15 = (int)(0.251f * RawQ15One);
 
-        private static readonly int[] _defaultSurroundToStereoCoefficients = new int[4]
+        private static readonly long[] _defaultSurroundToStereoCoefficients = new long[4]
         {
             RawQ15One,
             Minus3dBInQ15,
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Seems fair, nice catch.

marysaka (approved):

PR #6161, opened 2024-01-21
Fix missing data for new copy dependency textures with mismatching size

gdkchan:
If two textures with different size overlap in memory, a copy dependency is created to ensure modifications to one textures flows into the other as needed. When the copy dependency is created, the assumption is that the existing texture has the latest data, so its data is copied to the new texture. This is a problem when the new texture is larger than the existing one, because the remaining width/height of the new texture will remain untiliased. To fix that, this PR forces the new texture to load its data from CPU if the width or height is greater than that of the existing texture.

Fixes text rendering on Shantae and the Pirate's Curse (regression caused by #4364).

Before:
image
After:
image

Testing is welcome, texture cache related changes are prone to exploding.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
+++ b/src/Ryujinx.Graphics.Gpu/Image/TextureCache.cs
@@ -790,8 +790,12 @@ namespace Ryujinx.Graphics.Gpu.Image
 
                         texture = new Texture(_context, _physicalMemory, info, sizeInfo, range.Value, scaleMode);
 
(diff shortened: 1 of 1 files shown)


riperiperi (approved):
Nice catch. Might be better to avoid the copy, but this works for now.

marysaka (approved):

tgcordell:
I saw a bug similar to this (corrupted text) on TOTK on Mac w/ Apple Silicon. It only occurred maybe 1 out of every 50 launches of the game (pure estimate here, no tracked data). When I encountered it, the solve I had was fully quitting out and re-launching Ryujinx.

Just FYI

PR #6162, opened 2024-01-21
A32: Fix Vrecps

merryhime:
- Use FMA to fix being off by 1ULP
- Use EmitSse41RecipStepSelectOpF to fix 0 * inf case
- Add EmitSse41ForceToDefaultNaN to fix NaN case
- Add tests


merryhime:
Huh. Not sure how I broke this between testing and pushing. Will rework.

PR #6167, opened 2024-01-22
Vulkan: Enumerate Query Pool properly

riperiperi:
Turns out that `ElementAt<T>` for `Queue<T>` runs the default implementation as it doesn't implement `IList<T>`, which enumerates elements of the queue up to the given index. This code was creating `count` enumerators and iterating way more queue items than it needed to at higher counts. The solution is just to use one enumerator and break out of the loop when we get the count that we need.

3.5% of backend time was being spent _just_ enumerating at the usual spot in SMO. Only affects games using occlusion queries.


marysaka (approved):
Nice catch

gdkchan (approved):
lgtm, nice catch.

PR #6168, opened 2024-01-23
Vulkan: Use staging buffer for temporary constants

riperiperi:
Helper shaders and post processing effects typically need some parameters to tell them what to do, which we pass via constant buffers that are created and destroyed each time.

This can vary in cost between different Vulkan drivers. It shows up on profiles on mesa and MoltenVK, so it's worth avoiding. Some games only do it once (BlitColor for present), others multiple times. It's also done for post processing filters and FSR upscaling, which creates two buffers.

For mirrors, I added the ability to reserve a range on the staging buffer for use as any type of binding. This PR allows these constant buffers to instead be temporarily reserved on the staging buffer similar to mirrors, skipping allocation and buffer management costs entirely. If there isn't enough space for the reservation, it creates a temporary buffer that will be disposed after use, similar to how it's done right now.

Two temporary allocations do remain:
- DrawTexture, because it doesn't have access to the command buffer scope.
- ~~Index buffer indirect conversion, because one of them is a storage buffer and thus is a little more complicated.~~
  - Added, but needs testing

There's a small cost in that the uniform buffer takes up more space due to alignment requirements. At worst that's 256 bytes (on a GTX 1070) but more modern GPUs should have a better time.

Worth testing across different games and post effects to make sure they still work.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
+++ b/src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
@@ -1,5 +1,6 @@
 using Ryujinx.Common;
 using Ryujinx.Common.Logging;
+using Ryujinx.Graphics.GAL;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -29,6 +30,9 @@ namespace Ryujinx.Graphics.Vulkan
 
         private readonly VulkanRenderer _gd;
         private readonly BufferHolder _buffer;
+        private readonly int _resourceAlignment;
+
+        public readonly BufferHandle Handle;
 
         private readonly struct PendingCopy
         {
@@ -48,9 +52,10 @@ namespace Ryujinx.Graphics.Vulkan
         public StagingBuffer(VulkanRenderer gd, BufferManager bufferManager)
         {
             _gd = gd;
-            _buffer = bufferManager.Create(gd, BufferSize);
+            Handle = bufferManager.CreateWithHandle(gd, BufferSize, out _buffer);
             _pendingCopies = new Queue<PendingCopy>();
             _freeSize = BufferSize;
+            _resourceAlignment = (int)gd.Capabilities.MinResourceAlignment;
         }
 
         public void PushData(CommandBufferPool cbp, CommandBufferScoped? cbs, Action endRenderPass, BufferHolder dst, int dstOffset, ReadOnlySpan<byte> data)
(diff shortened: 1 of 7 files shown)


src/Ryujinx.Graphics.Vulkan/StagingBuffer.cs
```diff
@@ -197,16 +202,21 @@ private int GetContiguousFreeSize(int alignment)
         /// Reserve a range on the staging buffer for the current command buffer and upload data to it.
         /// </summary>
         /// <param name="cbs">Command buffer to reserve the data on</param>
-        /// <param name="data">The data to upload</param>
-        /// <param name="alignment">The required alignment for the buffer offset</param>
+        /// <param name="size">The minimum size the reserved data requires</param>
+        /// <param name="alignment">The required alignment for the buffer offset. -1 uses the most permissive alignment</param>
         /// <returns>The reserved range of the staging buffer</returns>
-        public unsafe StagingBufferReserved? TryReserveData(CommandBufferScoped cbs, int size, int alignment)
+        public unsafe StagingBufferReserved? TryReserveData(CommandBufferScoped cbs, int size, int alignment = -1)
```
gdkchan, line 208: Probably doesn't matter too much, but would be more efficient to have an additional overload without the alignment, which could call this method and pass `_resourceAlignment` as the alignment. Would avoid the -1 check.

That would require doing the same for `ReserveOrCreate` and maybe other methods (it doesn't look like anything passes an alignment to those right now?), so I'll leave it up to you, it might not be worth it.

riperiperi: The alignment parameter isn't used for now, so I can remove it from everything outside StagingBuffer.

Shihta:
it causes "MARIO + RABBIDS SPARKS OF HOPE" crash...
log:
[crash.log](https://github.com/Ryujinx/Ryujinx/files/14040709/crash.log)
[Ryujinx_1.1.0+fd29e65_2024-01-24_23-54-44.log](https://github.com/Ryujinx/Ryujinx/files/14040711/Ryujinx_1.1.0%2Bfd29e65_2024-01-24_23-54-44.log)

riperiperi:
Should be fine now, give it a shot. Note that this game seems to be broken on later versions of MoltenVK regardless (the CI builds use an older version), will need to look into that.

gdkchan (approved):
lgtm, thanks. I didn't test many games, so it might benefit from additional testing.

Shihta:
> Should be fine now, give it a shot. Note that this game seems to be broken on later versions of MoltenVK regardless (the CI builds use an older version), will need to look into that.

I tested the new version, it can show the starting screen now, but still crash after enter the game.
here are 2 logs:

[crash1.txt](https://github.com/Ryujinx/Ryujinx/files/14053586/crash1.txt)
[Ryujinx_1.1.0+7361920_2024-01-25_22-38-05.log](https://github.com/Ryujinx/Ryujinx/files/14053588/Ryujinx_1.1.0%2B7361920_2024-01-25_22-38-05.log)

[crash2.txt](https://github.com/Ryujinx/Ryujinx/files/14053597/crash2.txt)
[Ryujinx_1.1.0+7361920_2024-01-25_22-43-12.log](https://github.com/Ryujinx/Ryujinx/files/14053598/Ryujinx_1.1.0%2B7361920_2024-01-25_22-43-12.log)

riperiperi:
Are you sure this issue has anything to do with this PR? The log shows a crash in an area that wasn't changed by the PR, does it happen on master?

Shihta:
It's hard to be sure...
the game didn't crash after switching to master(1.1.1122), and I could play the game for about 15 minutes.
then, switch back to this PR, and it doesn't seem to be broken either...

marysaka (approved):
lgtm

PR #6175, opened 2024-01-25
Ava UI: Fix Warn & Remove Custom Theming

IsaacMarovitz:


MetrosexualGarbodor:
>custom theming on Avalonia was not being used by end users

If we included a guide on how to install custom themes (or even bundled some with the emulator itself, as RPCS3 does), I would imagine they would get a lot more use. It also doesn't help that Avalonia is still only used on 1/3 of the platforms Ryujinx is on, or that it doesn't seem to be as popular as other UI frameworks.

Removing this is probably fine, but it might need to get readded soon after its removal. Therefore I would advise waiting until at least it's defaulted on Windows and Linux, **unless** the trimming issue is a blocker for something else.

AcK77:
This "feature" is here since we introduced Avalonia and it's not used at all since Avalonia don't provide any theme file. If a user want to change the theme, the only allowed things would be changing colors or font sizes, since modify the UI itself require too much work. I don't think it's useful to keep that in place, it's not like GTK.

AcK77 (approved):

MetrosexualGarbodor:
Can font size and color settings be introduced in place of this feature then? I'm sure a lot of people would engage with those.

gdkchan:
As mentioned, it is not "simple" to use it. There is no documentation with the format, no repository with themes, so there is basically no way for the end user to change the theme (without digging on the code to figure out how it is supposed to work on their own). I do think the correct course of action is either removing it or making it easier to use.

I don't think anyone has used the GTK theme feature, FWIW.

MetrosexualGarbodor:
I have seen users crashing on GTK with custom themes, so it does get use, but it doesn't seem to work very well (if at all). So I guess this is for the best.

gdkchan:
I think adding color presets and option to change the font and adjust its size would be better. Maybe also a option to change icons, but that is more complicated.

AcK77:
Options to change color presets could be nice but needs a lot of work, at least maybe a font size factor preset could be nice to have but that can be added later.

IsaacMarovitz:
I plan to implement custom accent colors

marysaka (approved):

PR #6178, opened 2024-01-26
Fs: Log when Commit fails due to PathAlreadyInUse

AcK77:
This fixes and superseed #5418, nothing more.
(See original PR for description)

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
+++ b/src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
@@ -150,12 +150,10 @@ namespace Ryujinx.HLE.HOS.Services.Fs.FileSystemProxy
         // Commit()
         public ResultCode Commit(ServiceCtx context)
         {
-            Result result = _fileSystem.Get.Commit();
-            ResultCode resultCode = (ResultCode)result.Value;
-
-            if (result.IsFailure() && resultCode == ResultCode.PathAlreadyInUse)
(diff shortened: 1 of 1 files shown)


src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
```diff
@@ -149,7 +150,15 @@ public ResultCode OpenDirectory(ServiceCtx context)
         // Commit()
         public ResultCode Commit(ServiceCtx context)
         {
-            return (ResultCode)_fileSystem.Get.Commit().Value;
+            Result result = _fileSystem.Get.Commit();
+            ResultCode resultCode = (ResultCode)result.Value;
+
+            if (result.IsFailure() && resultCode == ResultCode.PathAlreadyInUse)
```
gdkchan, line 156:
```suggestion
            if (resultCode == ResultCode.PathAlreadyInUse)
```
I don't think checking if the result is an error is necessary since you already compare with a specific error code anyway.

src/Ryujinx.HLE/HOS/Services/Fs/FileSystemProxy/IFileSystem.cs
```diff
@@ -149,7 +150,15 @@ public ResultCode OpenDirectory(ServiceCtx context)
...
+            {
+                Logger.Warning?.Print(LogClass.ServiceFs, "The filesystem is already in use by another process.");
```
gdkchan, line 158: Maybe it should say "file system"?

gdkchan (approved):
lgtm, thanks.

PR #6179, opened 2024-01-26
Ava UI: Mod Manager Fixes

IsaacMarovitz:
Changes:
- Fixes a crash when selecting an invalid mod directory
- Fixes a crash when trying to enumerate through a directory that Ryujinx doesn't have permissions to access
- Fixes deleting a mod not deleting at the parent directory, leaving invisible orphan files that would prevent re-importing of the same mod
- Better logging and user feedback

TODO: More robust testing of folders to see if they contain valid mods before copying them to the mod directory. This will still copy invalid folders to the mod directory if they have subdirectories, where they won't appear in the list.

Since copying is recursive, this is especially important. Selecting a top-level folder by accident could be very expensive.

### DIFF
--- a/src/Ryujinx.Ava/Assets/Locales/en_US.json
+++ b/src/Ryujinx.Ava/Assets/Locales/en_US.json
@@ -383,6 +383,7 @@
   "DialogControllerSettingsModifiedConfirmSubMessage": "Do you want to save?",
   "DialogLoadFileErrorMessage": "{0}. Errored File: {1}",
   "DialogModAlreadyExistsMessage": "Mod already exists",
+  "DialogModInvalidMessage": "The specified directory does not contain a mod!",
   "DialogDlcNoDlcErrorMessage": "The specified file does not contain a DLC for the selected title!",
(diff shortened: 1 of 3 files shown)


AcK77 (approved):

Littl3Guy:
An error makes it possible to duplicate the mod and have it load from both the mod source and the ryujinx mod folder:

https://github.com/Ryujinx/Ryujinx/assets/115218507/02deaf5c-d085-4650-832b-7f4197ba51d4

IsaacMarovitz:
@Littl3Guy I can see why this bug is occurring. To me, it feels pretty niche, and I'm not sure how best to guard against it. The issue is the format of mods is quite flexible, so validating them isn't easy.

Littl3Guy:
Yeah its pretty weird steps to trigger it so its probably safe to leave like that for now.

Littl3Guy:
I suppose it wouldn't be possible to grey out the "Select Folder" option unless a folder is selected?

IsaacMarovitz:
> I suppose it wouldn't be possible to grey out the "Select Folder" option unless a folder is selected?

That's up to Avalonia, not much to be done about it.

marysaka (approved):
lgtm

PR #6182, opened 2024-01-26
Vulkan: Add Render Pass / Framebuffer Cache

riperiperi:
Every time the set of render targets is changed in Ryujinx, we create a new render pass and framebuffer, leaving the old ones to be deleted when their command buffer is fully consumed. This means we're at the mercy of the driver for how much work is done creating and destroying these, which could lead to inconsistent performance across drivers.

This PR adds a render pass cache to each TextureView, and a framebuffer cache for each render pass. Render passes register themselves on each participating texture view when created. Whenever any texture view in a render pass is deleted, the render pass is removed from all participating texture views and queued for deletion. When the renderer wants a render pass, it asks the "base attachment" for a framebuffer + render pass from the cache. 

Similar benefits to https://github.com/Ryujinx/Ryujinx/pull/6168, driver specific creation/deletion costs for render passes and framebuffers are reduced by keeping them around for future use. In this case, there are a lot more render passes per frame than temporary buffers (100s), but most drivers tend to have it a bit cheaper. Again, AMD mesa drivers had more visible impact in a profile than NVIDIA did for me, though it does depend on the game as some are crazier than others.

TBDR GPU drivers put a lot more focus on render passes and framebuffers, since they're performance critical due to the copying of texture data in and out of tile memory. MoltenVK isn't one of those (it only pays in allocation), but it could help us in future for others. Eventually, render pass cache keys should also include usage flags for each attachment.

One weird thing is how render passes for the swapchain work. They're also using TextureView to own the render pass cache, but we don't have all the information that normal texture views have (and especially not storage). I've put a bunch of warnings on that constructor to try make it clearer what drawbacks it has, and to never construct that way for normal texture usage.


src/Ryujinx.Graphics.Vulkan/RenderPassHolder.cs
```diff
@@ -0,0 +1,180 @@
...
+            foreach (var texture in textures)
+            {
+                texture.AddRenderPass(key, this);
+            }
+
+            _textures = textures;
+            _key = key;
+        }
+
+        public Auto<DisposableFramebuffer> GetFramebuffer(VulkanRenderer gd, CommandBufferScoped cbs, FramebufferParams fb) //int width, int height, int layers
```
riperiperi, line 144: Remove comment

src/Ryujinx.Graphics.Vulkan/HashTableSlim.cs
```diff
@@ -86,6 +86,41 @@ public void Add(ref TKey key, TValue value)
...
+
+                    if (entry.Hash == hashCode && entry.Key.Equals(ref key))
+                    {
+                        if (bucket.Length == 1)
+                        {
+                            _hashTable[hashCode & TotalBucketsMask] = null;
+                        }
+                        else
+                        {
+                            var newBucket = new Entry[bucket.Length - 1];
```
riperiperi, line 108: Deletion always causing a new bucket to be allocated is a bit unfortunate. Might be better if we had a length separate from the array, though that can cause additional bounds checks when iterating.

src/Ryujinx.Graphics.Vulkan/TextureView.cs
```diff
@@ -939,6 +959,35 @@ public void SetStorage(BufferRange buffer)
             throw new NotImplementedException();
         }
 
+        public (Auto<DisposableRenderPass> renderPass, Auto<DisposableFramebuffer> framebuffer) GetFramebuffer(
+            VulkanRenderer gd,
+            Device device,
+            CommandBufferScoped cbs,
+            FramebufferParams fb)
+        {
+            // Searches for render pass matching params colors and ds
```
riperiperi, line 968: Make comment less ugly

src/Ryujinx.Graphics.Vulkan/HashTableSlim.cs
gdkchan: I think it might be possible to avoid the bounds check by creating a span with the right length from the array, but then you have the span creation cost, I'm not sure how much it costs. There are also some tricks to avoid bounds check using "Unsafe", but that is a last resort, I have used it here for example (since it's extremely hot): https://github.com/Ryujinx/Ryujinx/blob/master/src/Ryujinx.Graphics.Device/DeviceState.cs#L81.

src/Ryujinx.Graphics.Vulkan/PipelineHelperShader.cs
```diff
@@ -9,19 +9,19 @@ public PipelineHelperShader(VulkanRenderer gd, Device device) : base(gd, device)
         {
         }
 
-        public void SetRenderTarget(Auto<DisposableImageView> view, uint width, uint height, bool isDepthStencil, VkFormat format)
+        public void SetRenderTarget(TextureView view, uint width, uint height, bool isDepthStencil, VkFormat format)
```
gdkchan, line 12: Do we still need all the other parameters here when the entire `TextureView` is passed (which also has width, height, format etc)?

src/Ryujinx.Graphics.Vulkan/RenderPassCacheKey.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public override int GetHashCode()
+        {
+            int hc = _depthStencil?.GetHashCode() ?? 0;
+
+            if (_colors != null)
+            {
+                foreach (var color in _colors)
+                {
+                    hc = HashCode.Combine(hc, color?.GetHashCode() ?? 0);
```
gdkchan, line 25: Might be cheaper (and "cleaner") to create a `HashCode` instance called `hc` and then just do `hc.Add(color);`.
BTW it can already handle null values internally:
https://github.com/dotnet/runtime/blob/16cb41496d595e2568574cfe11c763d5e05136c9/src/libraries/System.Private.CoreLib/src/System/HashCode.cs#L304
https://github.com/dotnet/runtime/blob/16cb41496d595e2568574cfe11c763d5e05136c9/src/libraries/System.Private.CoreLib/src/System/HashCode.cs#L87
IDK if that is documented somewhere or just implementation detail.

src/Ryujinx.Graphics.Vulkan/PipelineHelperShader.cs
gdkchan: I see that you added a new `TextureView` constructor that does not set `_info` etc so they will just be 0. I think it would be desirable to set it to the correct dimension/format from the swapchain, to enable it to be used here, and prevent potential future misuse.

riperiperi: `width` and `height` are the framebuffer width and height, which could technically be different. Others are less important to pass and could be inferred from the texture view. I've pushed a change that passes this info to TextureView constructor now, though I haven't changed any of these arguments yet.

src/Ryujinx.Graphics.Vulkan/HashTableSlim.cs
riperiperi: I've used an approach that takes a span from the array, each bucket is now a struct with a Length and Entries. It'll be cheaper than bounds checks on cases with many items in one bucket, though I don't know the setup cost for creating the span. The main cost for iterating the buckets is probably in equality checks.

riperiperi:
Now ready for review. Sanity testing across various games/hardware is welcome.

riperiperi:
I've removed load/store op barriers that were introduced to this PR due to them being quite numerous and excessive. In future would like to omit barriers if they're unnecessary and batch together image barriers into a single call.

gdkchan:
This is causing some glitches on Zelda Tears of the Kingdom, GPU is a RTX 3060.

https://github.com/Ryujinx/Ryujinx/assets/5624669/08875572-22c0-4b42-aea3-722702490993

Does not happen on master.

gdkchan (approved):
lgtm, thanks! The regression on Tears of the Kingdom seems to be fixed on the last commit. I think it makes sense to do the barrier changes separately since they are not strictly related, and I look forward to seeing them fixed once and for all.

Littl3Guy:
Tested a bunch of games and couldn't find any bugs.

marysaka (approved):

PR #6190, opened 2024-01-28
Implement virtual buffer dependencies

### LINKED ISSUE #5747: Monster hunter rise sunbreak not opening past menu on macOS
### Description of the issue

the game starts off by loading saved data and checking for add ons but crashes every time on the latest Ryujinx. Base game is playable but not the sun break title update.

### Reproduction steps

downloaded Ryujinx and monster hunter rise, sun break update and dlc. applied all add ons and title update, but did not run Base game runs.

### Log file

[Ryujinx_1.1.1035_2023-09-29_08-45-21.log](https://github.com/Ryujinx/Ryujinx/files/12766264/Ryujinx_1.1.1035_2023-09-29_08-45-21.log)
[Ryujinx_1.1.1037_2023-09-29_08-46-05.log](https://github.com/Ryujinx/Ryujinx/files/12766265/Ryujinx_1.1.1037_2023-09-29_08-46-05.log)
[Ryujinx_1.1.1037_2023-09-29_08-51-46.log](https://github.com/Ryujinx/Ryujinx/files/12766266/Ryujinx_1.1.1037_2023-09-29_08-51-46.log)

### OS

macOS Sonoma

### Ryujinx version

1.1.1037

### Game version

_No response_

### CPU

M 1
### GPU

M1 8 core
### RAM

8 gb

### List of applied mods

none
### Additional context?

_No response_

### LINKED ISSUE #6181: [Bug] Ace Attorney Apollo Justice Trilogy 3D models flickering
### Description of the issue

In Apollo Justice Trilogy when a 3D model is on screen and in the middle of an animation it flickers in and out (as in disappearing). This happens during Dual Destinies gameplay, Spirit of Justice gameplay and in the 'animation studio' feature. It seems to alternate frames that it flashes out. I think it also happens with other switch emulators.

My log file is large so i isolated it to when the model was onscreen.

### Reproduction steps

1. Open up Apollo Justice Trilogy and select "Game Selection and Museum"
2. Choose either Dual Destinies (DD), Spirit of Justice (SOJ) or the Museum
3. If either of the former two either load up a new game, load an episode or view the costumes (the third is the fastest, the rotation animation flickers)
4. If in the museum, choose the animation studio option and view any of the 3D characters in DD or SOJ. 

### Log file

[Ryujinx_ 1.1.0-macos1_2024-01-27_02-59-45.txt](https://github.com/Ryujinx/Ryujinx/files/14066982/Ryujinx_.1.1.0-macos1_2024-01-27_02-59-45.txt)

### OS

Macos Monterey version  12.6

### Ryujinx version

1.1.0-macos1 (System ver 16.1.0)

### Game version

1.0.0

### CPU

Mac M1 (2020) cpu

### GPU

Mac M1 (2020) gpu

### RAM

16GB

### List of applied mods

None

### Additional context?

Drive link to my screen recording of it happening

https://drive.google.com/file/d/1ZEFIdy0y8YXg0DpN88wgWwaXJiyEFEgB/view?usp=sharing

gdkchan:
This is similar to #5427, but while the linked change required the host to support sparse buffer mapping on Vulkan, this one does not require anything special. If the host does support sparse mapping, nothing should change, it will continue using it. But if the feature is not supported, this change improves the implementation by creating "virtual" buffers, that are also backed by "physical" buffers, but instead of doing actual memory aliasing on the GPU, it creates copy dependencies. Every time a physical buffer is updated, the data is also copied to the virtual buffer. The virtual buffer keeps track of GPU modifications, and copies it back on use (`SynchronizeMemory` call).

This should benefit platforms where sparse buffer mapping is not supported on Vulkan (macOS), and OpenGL.

Fixes model flickering on Apollo Justice: Ace Attorney Trilogy on macOS (and presumably on OpenGL in all platforms).
Before:

https://github.com/Ryujinx/Ryujinx/assets/5624669/a753b511-f65a-48e9-beb6-d64f29052e59

After:

https://github.com/Ryujinx/Ryujinx/assets/5624669/6ebe4b36-892f-41be-a9c2-e5e03adda94a

Also fixes a crash that could happen sometimes before displaying 3D models, which is related to the flickering issue.

Allows Monster Hunter Rise Sunbreak to work:
Captura de Tela 2024-01-27 às 21 23 46
Captura de Tela 2024-02-04 às 13 56 34
(I only tested the demo, but in theory the full game should work too...)

Fixes #5747, fixes #6181.

Testing is welcome.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
+++ b/src/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
@@ -5,6 +5,7 @@ using Ryujinx.Memory.Tracking;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Ryujinx.Graphics.Gpu.Memory
 {
@@ -66,6 +67,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
         private readonly Action<ulong, ulong> _modifiedDelegate;
 
         private HashSet<MultiRangeBuffer> _virtualDependencies;
+        private ReaderWriterLockSlim _virtualDependenciesLock;
 
         private int _sequenceNumber;
 
@@ -154,6 +156,8 @@ namespace Ryujinx.Graphics.Gpu.Memory
             _externalFlushDelegate = new RegionSignal(ExternalFlush);
             _loadDelegate = new Action<ulong, ulong>(LoadRegion);
             _modifiedDelegate = new Action<ulong, ulong>(RegionModified);
+
+            _virtualDependenciesLock = new ReaderWriterLockSlim();
         }
 
         /// <summary>
@@ -214,6 +218,22 @@ namespace Ryujinx.Graphics.Gpu.Memory
         }
 
         /// <summary>
+        /// Copies modified data on all virtual buffers back to this buffer,
+        /// then performs guest to host memory synchronization of the buffer data.
+        /// </summary>
+        /// <remarks>
+        /// This causes the buffer data to be overwritten if a write was detected from the CPU,
+        /// since the last call to this method.
+        /// </remarks>
+        /// <param name="address">Start address of the range to synchronize</param>
+        /// <param name="size">Size in bytes of the range to synchronize</param>
+        public void SynchronizeMemoryWithVirtualCopyBack(ulong address, ulong size)
+        {
+            CopyFromDependantVirtualBuffers();
+            SynchronizeMemory(address, size);
+        }
+
+        /// <summary>
         /// Performs guest to host memory synchronization of the buffer data.
         /// </summary>
         /// <remarks>
@@ -222,14 +242,8 @@ namespace Ryujinx.Graphics.Gpu.Memory
         /// </remarks>
         /// <param name="address">Start address of the range to synchronize</param>
         /// <param name="size">Size in bytes of the range to synchronize</param>
-        /// <param name="copyBackVirtual">Whether virtual buffers that uses this buffer as backing memory should have its data copied back if modified</param>
-        public void SynchronizeMemory(ulong address, ulong size, bool copyBackVirtual = true)
+        public void SynchronizeMemory(ulong address, ulong size)
         {
-            if (copyBackVirtual)
-            {
-                CopyFromDependantVirtualBuffers();
-            }
-
             if (_useGranular)
             {
                 _memoryTrackingGranular.QueryModified(address, size, _modifiedDelegate, _context.SequenceNumber);
@@ -549,7 +563,7 @@ namespace Ryujinx.Graphics.Gpu.Memory
             using PinnedSpan<byte> data = _context.Renderer.GetBufferData(Handle, offset, (int)size);
 
             // TODO: When write tracking shaders, they will need to be aware of changes in overlapping buffers.
-            _physicalMemory.WriteUntracked(address, data.Get());
+            _physicalMemory.WriteUntracked(address, CopyFromDependantVirtualBuffers(data.Get(), address, size));
         }
 
         /// <summary>
@@ -638,7 +652,16 @@ namespace Ryujinx.Graphics.Gpu.Memory
         /// <param name="virtualBuffer">Dependant virtual buffer</param>
         public void AddVirtualDependency(MultiRangeBuffer virtualBuffer)
         {
-            (_virtualDependencies ??= new()).Add(virtualBuffer);
+            _virtualDependenciesLock.EnterWriteLock();
+
+            try
+            {
+                (_virtualDependencies ??= new()).Add(virtualBuffer);
+            }
+            finally
+            {
+                _virtualDependenciesLock.ExitWriteLock();
+            }
         }
 
         /// <summary>
@@ -647,14 +670,23 @@ namespace Ryujinx.Graphics.Gpu.Memory
         /// <param name="virtualBuffer">Dependant virtual buffer</param>
         public void RemoveVirtualDependency(MultiRangeBuffer virtualBuffer)
         {
-            if (_virtualDependencies != null)
-            {
-                _virtualDependencies.Remove(virtualBuffer);
+            _virtualDependenciesLock.EnterWriteLock();
 
-                if (_virtualDependencies.Count == 0)
+            try
+            {
+                if (_virtualDependencies != null)
                 {
-                    _virtualDependencies = null;
+                    _virtualDependencies.Remove(virtualBuffer);
+
+                    if (_virtualDependencies.Count == 0)
+                    {
+                        _virtualDependencies = null;
+                    }
                 }
+            }
+            finally
+            {
+                _virtualDependenciesLock.ExitWriteLock();
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
```diff
@@ -220,8 +222,14 @@ public bool FullyContains(ulong address, ulong size)
         /// </remarks>
         /// <param name="address">Start address of the range to synchronize</param>
         /// <param name="size">Size in bytes of the range to synchronize</param>
-        public void SynchronizeMemory(ulong address, ulong size)
+        /// <param name="copyBackVirtual">Whether virtual buffers that uses this buffer as backing memory should have its data copied back if modified</param>
+        public void SynchronizeMemory(ulong address, ulong size, bool copyBackVirtual = true)
```
gdkchan, line 226: `SyncAction` calls this frequently, which causes data from the virtual buffer to be copied back all the time. Maybe something should be done about this before merge. I think it's technically the right thing since we want the data to be in the physical buffer for the cases where a flush is needed (data accessed from CPU etc), and doing it on `ExternalFlush` only might be too late.

src/Ryujinx.Graphics.Gpu/Memory/VirtualRangeCache.cs
```diff
@@ -6,9 +6,9 @@
 namespace Ryujinx.Graphics.Gpu.Memory
 {
     /// <summary>
-    /// Virtual buffer cache.
+    /// Virtual range cache.
     /// </summary>
-    class VirtualBufferCache
+    class VirtualRangeCache
```
gdkchan, line 11: I renamed this since the name "virtual buffer cache" doesn't make sense, there's no buffers here. I guess it made sense in the first implementation way back where it still had a virtual buffer cache with actual buffers.

src/Ryujinx.Graphics.Gpu/Memory/VirtualRangeCache.cs
```diff
@@ -158,7 +157,7 @@ public bool TryGetOrAddRange(ulong gpuVa, ulong size, bool supportsSparse, out M
                 }
                 else
                 {
-                    found = true;
+                    found = overlap0.Range.Count == 1 || IsSparseAligned(overlap0.Range);
```
gdkchan, line 160: This is to prevent the following bug (example):
We have physical range `0x8000->0x9000; 0xE000->0xF000` for VA range `0x1000->0x3000` in the cache. It is not considered sparse aligned, so its transformed into `0x8000->0xA000`, and it creates a buffer for that range. Later it accesses VA range `0x2000->0x3000`, which returns physical range `0xE000->0xF000` (correct), but it assumes the physical buffer exists, but it doesn't because range was transformed to `0x8000->0xA000`. Fix is always forcing buffer creation if we have more than one range, and it is not sparse aligned.

I have not seen the bug trigger in practice yet when sparse buffer is supported, but when it's not supported, we had a report on https://github.com/Ryujinx/Ryujinx-Games-List/issues/4862#issuecomment-1912856454. (But this situation is no longer possible anyway as we now only do the transformation if the range is not sparse compatible, and no longer when host does not support sparse buffers).

src/Ryujinx.Graphics.Gpu/Memory/MultiRangeBuffer.cs
```diff
@@ -49,11 +106,108 @@ public BufferRange GetRange(MultiRange range)
...
+        /// <summary>
+        /// Adds a modified virtual memory range.
+        /// </summary>
+        /// <remarks>
+        /// This is only required when the host does not support sparse buffers, otherwise only physical buffers need to track modification.
+        /// </remarks>
+        /// <param name="range">Modified range</param>
+        public void AddModifiedRegion(MultiRange range)
+        {
+            _modifiedRanges ??= new(_context, null, null);
```
gdkchan, line 172: This is a bit weird since `BufferModifiedRangeList` is designed to do more than we need here, so I need to pass some arguments as null. But they are not used on the methods called here. Maybe a better solution would be making a simpler  `BufferModifiedRangeList` and a more complex one inheriting from it. Or maybe what I already have here is fine...

wjddbstl:
Thank you always. We have Ryujin's for Linux and Windows, but is there no way for Mac?

tgcordell:
> Thank you always. We have Ryujin's for Linux and Windows, but is there no way for Mac?

Expand the "Experimental GUI (Avalonia)" section from the bot's comment.

wjddbstl:
Thank you so much brother for letting me know. All my problems have been solved. gdkchan, who came up with this method, you are a genius. Thank you so much.

adrien-krebet:
I tested this build and found an issue still on Apollo Justice Trilogy (video footage). Though like your video the costume view works I'm not sure about your DD and SOJ or the DD/SOJ animation studio gameplay whether it'd be present for you.
https://github.com/Ryujinx/Ryujinx/issues/6181#issuecomment-1913579375
has the log in the reply below.

src/Ryujinx.Graphics.Gpu/Memory/Buffer.cs
riperiperi: I think doing it on ExternalFlush at all is too late, it's waiting on the sync for the original write - it would need to create new sync and register that on the modified range list to wait for the copy. Ideally it should copy here only when any dependant virtual buffers were modified, and flush ranges out of relevant virtual buffers rather than trying to get it all in one place first.

gdkchan: > I think doing it on ExternalFlush at all is too late

Maybe I should remove the `CopyFromDependantVirtualBuffers` call there then? I figured it could at least help with OpenGL where the wait is implicit (IIRC).

> and flush ranges out of relevant virtual buffers rather than trying to get it all in one place first.

I'm not sure how hard it would be to redirect the flush to the virtual buffer(s). The main point of copying from/to the "physical" buffers is avoiding all the complexity of doing all sync with virtual buffers (which was my first attempt at solving this problem). At least it would have to keep track of what is the latest virtual buffer to be modified, in case there are many, and flush them in modification order. I guess all those problems were already solved for textures.

gdkchan: BTW I don't think any of those games actually needs flush from a virtual buffer, they are usually used for read-only storage buffers, so might be a bit tricky to test the flush case.

gdkchan: I changed it to make flush source the data from the virtual buffer directly. Unfortunately it did not make much of a difference, at least on Monster Hunter Rise. It still does many copies from virtual buffers due to the indirect buffer. It appears to be modified from a shader, where presumably it uses a large storage buffer range with more than one sub-range, which requires a virtual buffer. But then the indirect buffer is small, so it can there's only one sub-range and it uses a physical buffer (which requires a copy back from the virtual buffer).

riperiperi (approved):
Cool, nice to have this covered when sparse buffers aren't supported. I do agree that using BufferModifiedRangeList outside of its intended context is weird, but it's not exactly a blocker if it's not hitting any of the cases where it attempts to use those null references.

marysaka (approved):

PR #6208, opened 2024-01-29
Vulkan: Better UI Vendor & Driver Reporting

IsaacMarovitz:
Report Vendor & Driver on the UI when using Vulkan. Fixes macOS just showing as `MoltenVK`.

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
+++ b/src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
@@ -637,7 +637,7 @@ namespace Ryujinx.Graphics.Vulkan
 
         public HardwareInfo GetHardwareInfo()
         {
-            return new HardwareInfo(GpuVendor, GpuRenderer, GpuDriver);
+            return new HardwareInfo(GpuVendor, GpuRenderer, $"{GpuVendor} ({GpuDriver})");
         }
 
         /// <summary>
--- a/src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
+++ b/src/Ryujinx.Graphics.Vulkan/VulkanRenderer.cs
@@ -637,7 +637,14 @@ namespace Ryujinx.Graphics.Vulkan
 
         public HardwareInfo GetHardwareInfo()
         {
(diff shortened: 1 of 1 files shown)


gdkchan:
On AMD, this shows the following:
image
So we can't depend on a simple equality check to verify that driver and vendor name are basically the same.
Here's some ideais:
- Have a `IsProprietaryDriver` function that returns true if the driver is the proprietary one, it can guess that from the driver name. We can build a list of strings returned by all proprietary drivers, maybe we can get it from vulkan.gpuinfo.org.
- Check if driver name starts with vendor name (ignore case comparison), somewhat simplified version of the above.

From a API design perspective, I think it would be better if the `HardwareInfo` structure always returned the driver name, and then the GUI would decide if it should display it or not. OpenGL could return null as the driver name (as there is no such a thing in OpenGL), and UI would need to deal with it (showing only vendor name).

I think I generally prefer the second option, and doing the driver name check on the UI.

gdkchan:
Code for what I proposed above (untested):
```cs
if (info.DriverName == null || info.DriverName.StartsWith(info.VendorName, StringComparison.OrdinalIgnoreCase))
{
    displayString = info.VendorName;
}
else
{
    displayString = $"{info.VendorName} ({info.DriverName})";
}
```
We can also special case macOS here if we want to keep the same behaviour as before (showing only "GPU: Apple" etc).

gdkchan:
Another alternative: Rename "GPU:" to "Driver:", only show driver name, except on macOS where it would be special case to show both driver and GPU, or just GPU.

MutantAura:
In general it may also be nice to have a quick look-up table of the available drivers matched with a set of "friendly" front-facing names. There are a few drivers that are nice to be aware of but have some pretty awful and long names, or some really weird capitalization.

Just a couple I'm aware of (more likely exist but may not be relevant at the moment):
- Turnip is called 'turnip Mesa driver`
- RADV is called 'radv'

In general it would likely be preferable to display:
`AMD (RADV)` instead of `AMD (radv)` 
`GPU: Qualcomm (Turnip)` rather than `GPU: Qualcomm (turnip Mesa driver)` 

Not sure what the cleanest way of doctoring these would be...

PR #6211, opened 2024-01-29
Avalonia: Fix dialog issues caused by 1.1.1105

### LINKED ISSUE #6207: [Bug] Avalonia UI is not interactive anymore
### Description of the issue

This issue was originally reported on Discord by `dorsalfins`.

This is a regression from 1.1.1105 due to the ContentDialogHelper changes I added.

If a dialog opens before the main window finished opening the UI won't be interactive at all, even if you manage to close the dialog.

### Reproduction steps

There are probably multiple ways to trigger this issue, but here is one example:

1. Make sure your `vm.max_map_count` value is lower than 524288.
2. Start Ryujinx
3. Optional: Try to close the dialog

### Log file

N/A

### OS

Arch Linux

### Ryujinx version

1.1.1143

### Game version

_No response_

### CPU

_No response_

### GPU

_No response_

### RAM

_No response_

### List of applied mods

_No response_

### Additional context?

I was not able to reproduce this issue myself, but I have an idea about the cause of this and will attempt to fix it shortly.

Thank you @piplup55 for investigating this and confirming that this issue was caused by `1.1.1105`.

TSRBerry:
This PR updates `CheckLaunchState()` to  be async, since we don't want to open multiple dialogs at the same time.

This allows people opening Ryujinx for the first time to actually interact with the window again.

Note: The mod manager is currently prone to the same issue due to the usage of `Dispatcher.UIThread.Post()` with an async lambda.

### DIFF
--- a/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
+++ b/src/Ryujinx.Ava/UI/Helpers/ContentDialogHelper.cs
@@ -388,6 +388,7 @@ namespace Ryujinx.Ava.UI.Helpers
             {
                 _contentDialogOverlayWindow.Content = null;
                 _contentDialogOverlayWindow.Close();
+                _contentDialogOverlayWindow = null;
             }
 
             return result;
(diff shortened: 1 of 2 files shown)


marysaka (approved):

AcK77 (approved):

Berbe:
Fixes: #6207?

PR #6213, opened 2024-01-29
Mako: Remove events that trigger from a forked repository

TSRBerry:
Events that trigger workflows from forked repositories don't get any secrets, so they are useless for Mako.


TSRBerry:
Merge this with `[skip ci]` in  the commit message.

PR #6214, opened 2024-01-29
Fix wrong shortcut path for macOS

siegmund-heiss-ich:
In the current solution the created shortcut gets stored in the `/Applications` folder, which is inconsistent to other platforms. It should be `~/Desktop`.

This PR fixes the output path for shortcut creation.

### DIFF
--- a/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
+++ b/src/Ryujinx.Ui.Common/Helper/ShortcutHelper.cs
@@ -56,7 +56,7 @@ namespace Ryujinx.Ui.Common.Helper
             var plistFile = EmbeddedResources.ReadAllText("Ryujinx.Ui.Common/shortcut-template.plist");
             var shortcutScript = EmbeddedResources.ReadAllText("Ryujinx.Ui.Common/shortcut-launch-script.sh");
             // Macos .App folder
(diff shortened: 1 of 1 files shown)


gdkchan:
I think it should be consistent with other macOS applications, not other platforms. Do macOS applications usually put shortcuts in the desktop?

jcm93:
I think the Apple canon thing is to place the .app bundle in /Applications always. 'Shortcuts' in Ryujinx are not genuinely such, they're just small .app bundle shell script shims. If you want an app shortcut on the desktop, I'm pretty sure the canonical way would be to use an alias that then links to the shim bundle in /Applications. Running an .app bundle from places that aren't /Applications are a frequent cause of sandboxing issues or library issues (for that matter, so are shell script app shims, but that's a different conversation).

siegmund-heiss-ich:
> I think it should be consistent with other macOS applications, not other platforms. Do macOS applications usually put shortcuts in the desktop?

The Applications folder would be correct for macOS afaik. 
But this message is displayed when you hover over the shortcut button:
Bildschirmfoto 2024-01-29 um 20 55 00

We could also just change the text of the info.

merryhime:
as a macOS user I would not expect app shortcuts on my desktop

siegmund-heiss-ich:
Anyways, what we are doing here technically isn't a shortcut.
MacOS users deserve to stay confused I guess.

PR #6234, opened 2024-01-31
Infra: Run Mac Actions on Apple Silicon

IsaacMarovitz:
This will run macOS actions on the newly released Apple Silicon runners. The reasoning is two-fold:
- Ryujinx is primarily targeting Apple Silicon not Intel Macs, so it makes sense for the tests to be run on this architecture
- Apple Silicon runners use significantly newer hardware, which should reduce the build time

TODO:
- Deal with Unicorn, going to leave how that should happen up to the maintainers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,12 +52,12 @@ jobs:
       - name: Change config filename
         run: sed -r --in-place 's/\%\%RYUJINX_CONFIG_FILE_NAME\%\%/PRConfig\.json/g;' src/Ryujinx.Common/ReleaseInformation.cs
         shell: bash
-        if: github.event_name == 'pull_request' && matrix.os != 'macOS-latest'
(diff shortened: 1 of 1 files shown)


TSRBerry:
If unicorn currently doesn't bundle a native library for macOS ARM we should PR that to them, since they are also using GitHub Actions for their packaging. I could do this as well if you can't or don't want to.

I'll mark your PR as draft in the mean time.

gdkchan:
I don't think we can run Unicorn on macOS Arm (from .NET at least) due to JIT issues. IIRC it does not leave the JIT protection n the right state for running .NET code (which is also JIT generated).

IsaacMarovitz:
I think separating CPU tests into their own project and skipping Unicorn-based tests on macOS would be a reasonable solution here.

TSRBerry:
Skipping CPU tests is not a good idea.

PR #6237, opened 2024-02-01
Ava UI: Add Open Mod Dir Buttons

IsaacMarovitz:
Adds back the old buttons... even though they really shouldn't be added back.


src/Ryujinx.Ava/UI/Windows/CheatWindow.axaml.cs
```diff
@@ -119,5 +124,32 @@ public void Save()
...
+                catch (Exception ex)
+                {
+                    Dispatcher.UIThread.InvokeAsync(async () =>
+                    {
+                        await ContentDialogHelper.CreateErrorDialog(LocaleManager.Instance.UpdateAndGetDynamicValue(
+                            LocaleKeys.DialogLoadFileErrorMessage,
+                            ex,
+                            cheatsPath));
+                    });
+                    return;
```
AcK77, line 148:
```suggestion

                    return;
```

AcK77 (approved):

src/Ryujinx.Ava/UI/Windows/CheatWindow.axaml.cs
```diff
@@ -119,5 +124,33 @@ public void Save()
 
             Close();
         }
+
+        public void OpenCheatFolder()
+        {
+            string modsBasePath = ModLoader.GetModsBasePath();
+            string cheatsPath = Path.Combine(modsBasePath, "contents", TitleId, "Cheats");
```
TSRBerry, line 131:
```suggestion
            string cheatsPath = Path.Combine(modsBasePath, "contents", TitleId, "cheats");
```

Why not restore the previous buttons? They were better imo.

TSRBerry:
I'd still prefer to have the old buttons back instead of this.

gdkchan (approved):
Nice.

PR #6238, opened 2024-02-01
Standardize logging locations across desktop platforms

### LINKED ISSUE #6252: [Bug] Logs is denied.
### Description of the issue

A couple of builds ago I had this problem, the app doesn't open because it says it doesn't have access to the logs.

### Reproduction steps

Just Open on linux.

### Log file

As it no longer generates logs, I have to send the log that comes out through the terminal:

```
[willian@archlinux ~]$ ryujinx
00:00:00.005 |I| Configuration LogValueChange: EnableFileLog set to: True
00:00:00.077 |E| Application : Unhandled exception caught: System.UnauthorizedAccessException: Access to the path '/opt/ryujinx/Logs' is denied.
 ---> System.IO.IOException: Permission denied
   at System.IO.FileSystem.CreateDirectory(String, UnixFileMode)
   at System.IO.FileSystem.CreateDirectory(String)
   at System.IO.DirectoryInfo.Create()
   at Ryujinx.Common.Logging.Targets.FileLogTarget.PrepareLogFile(String path) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Common/Logging/Targets/FileLogTarget.cs:line 29
   at Ryujinx.Ui.Common.Configuration.LoggerModule.ReloadFileLogger(Object sender, ReactiveEventArgs`1 e) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ui.Common/Configuration/LoggerModule.cs:line 88
   at Ryujinx.Common.ReactiveObject`1.set_Value(T value) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Common/ReactiveObject.cs:line 39
   at Ryujinx.Ui.Common.Configuration.ConfigurationState.Load(ConfigurationFileFormat configurationFileFormat, String configurationFilePath) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs:line 1433
   at Ryujinx.Ava.Program.ReloadConfig() in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 153
   at Ryujinx.Ava.Program.Initialize(String[] args) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 101
   at Ryujinx.Ava.Program.Main(String[] args) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 47
Unhandled exception. System.UnauthorizedAccessException: Access to the path '/opt/ryujinx/Logs' is denied.
 ---> System.IO.IOException: Permission denied
   at System.IO.FileSystem.CreateDirectory(String, UnixFileMode)
   at System.IO.FileSystem.CreateDirectory(String)
   at System.IO.DirectoryInfo.Create()
   at Ryujinx.Common.Logging.Targets.FileLogTarget.PrepareLogFile(String path) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Common/Logging/Targets/FileLogTarget.cs:line 29
   at Ryujinx.Ui.Common.Configuration.LoggerModule.ReloadFileLogger(Object sender, ReactiveEventArgs`1 e) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ui.Common/Configuration/LoggerModule.cs:line 88
   at Ryujinx.Common.ReactiveObject`1.set_Value(T value) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Common/ReactiveObject.cs:line 39
   at Ryujinx.Ui.Common.Configuration.ConfigurationState.Load(ConfigurationFileFormat configurationFileFormat, String configurationFilePath) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ui.Common/Configuration/ConfigurationState.cs:line 1433
   at Ryujinx.Ava.Program.ReloadConfig() in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 153
   at Ryujinx.Ava.Program.Initialize(String[] args) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 101
   at Ryujinx.Ava.Program.Main(String[] args) in /home/main-builder/pkgwork/src/Ryujinx/src/Ryujinx.Ava/Program.cs:line 47
/usr/bin/ryujinx, linha 20:  1137 Abortado                (imagem do núcleo gravada)$COMMAND "$SCRIPT_DIR/$RYUJINX_BIN" "$@"

```

### OS

Arch Linux

### Ryujinx version

Ryujinx-git 1.1.1157.r0.g24c8b0edc-1

### Game version

_No response_

### CPU

_No response_

### GPU

_No response_

### RAM

_No response_

### List of applied mods

_No response_

### Additional context?

_No response_

jcm93:
### Description
Unifies log directory finding and creation in `AppDataManager`, and ensures logs are created in the intended, correct places on each platform.

The preferred directory hierarchy is the following.

Windows: Prefer a 'Logs' folder in whatever directory Ryujinx was launched from; fall back to `%APPDATA%\Ryujinx\Logs`.

Mac: Prefer `~/Library/Logs/Ryujinx`. Fall back to `~/Library/Application Support/Ryujinx/Logs`.

Linux: Use `~/.config/Ryujinx/Logs`.

### Context

https://github.com/Ryujinx/Ryujinx/commit/70fcba39de34fc211d09da12783898161724b0dc had an unfortunate quality escape where logs on macOS were being written to the executable directory, which is within the .app bundle. This was invalidating the code signature of the app bundle, which can lead to various unintended side effects. This PR addresses that issue while standardizing log locations across Windows, Mac and Linux.

Fix #6252

### DIFF
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1353,8 +1353,6 @@ namespace Ryujinx.Ava.UI.ViewModels
             string logPath = AppDataManager.LogsDirPath;
             if (!string.IsNullOrEmpty(logPath))
             {
-                new DirectoryInfo(logPath).Create();
-
                 OpenHelper.OpenFolder(logPath);
             }
         }
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1350,7 +1350,7 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         public void OpenLogsFolder()
         {
-            string logPath = AppDataManager.LogsDirPath;
+            string logPath = AppDataManager.GetOrCreateLogsDir();
             if (!string.IsNullOrEmpty(logPath))
             {
                 OpenHelper.OpenFolder(logPath);
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1352,9 +1352,9 @@ namespace Ryujinx.Ava.UI.ViewModels
         {
             string logPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
 
-            if (LoggerModule.LogDirectoryPath != null)
+            if (AppDataManager.LogsDirPath != null)
             {
-                logPath = LoggerModule.LogDirectoryPath;
+                logPath = AppDataManager.LogsDirPath;
             }
 
             new DirectoryInfo(logPath).Create();
--- a/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
+++ b/src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
@@ -1350,16 +1350,13 @@ namespace Ryujinx.Ava.UI.ViewModels
 
         public void OpenLogsFolder()
         {
-            string logPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
-
-            if (AppDataManager.LogsDirPath != null)
+            string logPath = AppDataManager.LogsDirPath;
+            if (!string.IsNullOrEmpty(logPath))
             {
-                logPath = AppDataManager.LogsDirPath;
+                new DirectoryInfo(logPath).Create();
+
+                OpenHelper.OpenFolder(logPath);
             }
-
-            new DirectoryInfo(logPath).Create();
-
-            OpenHelper.OpenFolder(logPath);
         }
 
         public void ToggleDockMode()
--- a/src/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/src/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -118,6 +118,7 @@ namespace Ryujinx.Common.Configuration
             {
                 return LogsDirPath;
             }
+
             Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
             LogsDirPath = SetUpLogsDir();
 
--- a/src/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/src/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -127,7 +127,7 @@ namespace Ryujinx.Common.Configuration
                 {
                     Logger.Warning?.Print(LogClass.Application,
                         $"Logging directory could not be created '{logDir}'");
-                    logDir = "";
+                    return null;
                 }
             }
             else
@@ -148,11 +148,10 @@ namespace Ryujinx.Common.Configuration
                         logDir = "";
                     }
 
-                    if (logDir == "")
+                    if (string.IsNullOrEmpty(logDir))
                     {
                         //Should evaluate to ~/Library/Application Support/Ryujinx/Logs
-                        logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
-                            DefaultBaseDir, "Logs");
+                        logDir = Path.Combine(BaseDirPath, "Logs");
 
                         try
                         {
@@ -162,7 +161,7 @@ namespace Ryujinx.Common.Configuration
                         {
                             Logger.Warning?.Print(LogClass.Application,
                                 $"Logging directory could not be created '{logDir}'");
-                            logDir = "";
+                            return null;
                         }
                     }
                 }
@@ -181,7 +180,7 @@ namespace Ryujinx.Common.Configuration
                         logDir = "";
                     }
 
-                    if (logDir == "")
+                    if (string.IsNullOrEmpty(logDir))
                     {
                         //Should evaluate to C:\Users\user\AppData\Roaming\Ryujinx\Logs
                         logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
@@ -195,7 +194,7 @@ namespace Ryujinx.Common.Configuration
                         {
                             Logger.Warning?.Print(LogClass.Application,
                                 $"Logging directory could not be created '{logDir}'");
-                            logDir = "";
+                            return null;
                         }
                     }
                 }
@@ -213,7 +212,7 @@ namespace Ryujinx.Common.Configuration
                     {
                         Logger.Warning?.Print(LogClass.Application,
                             $"Logging directory could not be created '{logDir}'");
-                        logDir = "";
+                        return null;
                     }
                 }
             }
@@ -222,8 +221,8 @@ namespace Ryujinx.Common.Configuration
 
         private static void SetupBasePaths()
         {
-            LogsDirPath = SetUpLogsDir();
             Directory.CreateDirectory(BaseDirPath);
+            LogsDirPath = SetUpLogsDir();
             Directory.CreateDirectory(GamesDirPath = Path.Combine(BaseDirPath, GamesDir));
             Directory.CreateDirectory(ProfilesDirPath = Path.Combine(BaseDirPath, ProfilesDir));
             Directory.CreateDirectory(KeysDirPath = Path.Combine(BaseDirPath, KeysDir));
--- a/src/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/src/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -120,6 +120,7 @@ namespace Ryujinx.Common.Configuration
             }
             Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
             LogsDirPath = SetUpLogsDir();
+
             return LogsDirPath;
         }
 
@@ -137,6 +138,7 @@ namespace Ryujinx.Common.Configuration
                 catch
                 {
                     Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+
                     return null;
                 }
             }
@@ -144,7 +146,7 @@ namespace Ryujinx.Common.Configuration
             {
                 if (OperatingSystem.IsMacOS())
                 {
-                    // Should evaluate to ~/Library/Logs/Ryujinx/
+                    // NOTE: Should evaluate to "~/Library/Logs/Ryujinx/".
                     logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "Library", "Logs", DefaultBaseDir);
                     try
                     {
@@ -158,7 +160,7 @@ namespace Ryujinx.Common.Configuration
 
                     if (string.IsNullOrEmpty(logDir))
                     {
-                        // Should evaluate to ~/Library/Application Support/Ryujinx/Logs
+                        // NOTE: Should evaluate to "~/Library/Application Support/Ryujinx/Logs".
                         logDir = Path.Combine(BaseDirPath, "Logs");
 
                         try
@@ -168,13 +170,14 @@ namespace Ryujinx.Common.Configuration
                         catch
                         {
                             Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+
                             return null;
                         }
                     }
                 }
                 else if (OperatingSystem.IsWindows())
                 {
-                    // Should evaluate to a 'Logs' directory in whatever directory Ryujinx was launched from.
+                    // NOTE: Should evaluate to a "Logs" directory in whatever directory Ryujinx was launched from.
                     logDir = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
                     try
                     {
@@ -188,7 +191,7 @@ namespace Ryujinx.Common.Configuration
 
                     if (string.IsNullOrEmpty(logDir))
                     {
-                        // Should evaluate to C:\Users\user\AppData\Roaming\Ryujinx\Logs
+                        // NOTE: Should evaluate to "C:\Users\user\AppData\Roaming\Ryujinx\Logs".
                         logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir, "Logs");
 
                         try
@@ -198,13 +201,14 @@ namespace Ryujinx.Common.Configuration
                         catch
                         {
                             Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+
                             return null;
                         }
                     }
                 }
                 else if (OperatingSystem.IsLinux())
                 {
-                    // Should evaluate to ~/.config/Ryujinx/Logs
+                    // NOTE: Should evaluate to "~/.config/Ryujinx/Logs".
                     logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir, "Logs");
 
                     try
@@ -214,6 +218,7 @@ namespace Ryujinx.Common.Configuration
                     catch
                     {
                         Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+
                         return null;
                     }
                 }
--- a/src/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/src/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -48,8 +48,7 @@ namespace Ryujinx.Common.Configuration
 
         public static void Initialize(string baseDirPath)
         {
-            string appDataPath;
-            appDataPath = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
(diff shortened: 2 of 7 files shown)


gdkchan:
Why not `C:\Users\user\AppData\Roaming\Ryujinx\Logs` on Windows (as fallback) which is were the emulator config and everything else is stored?

jcm93:
Whoops, I misinterpreted something offthread and thought we wanted %LOCALAPPDATA% instead of %APPDATA%. I'll change it to %APPDATA%.

Steveice10:
This doesn’t appear to account for portable mode, which could make logs more annoying to get to.

jcm93:
Updated to respect portable mode.

jcm93:
There is some discussion still offthread about the appropriate location to log on Windows. Just wanted to put forward my two cents that I hope that discussion is non-blocking for this PR. Most of all, it is important to stop macOS from invalidating its own signature. An invalid code signature can stop Ryujinx from being able to write to its own Application Support folder, stop Ryujinx from being able to load its own libraries, stop it from launching entirely, etc. This also *must* be merged before https://github.com/Ryujinx/Ryujinx/pull/6241, assuming both are merged, since it's very probable that the copy and symlink removal operation there will fail if Ryujinx's code signature is invalid.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,8 +113,116 @@ public static void Initialize(string baseDirPath)
...
+                    {
+                        Logger.Warning?.Print(LogClass.Application,
+                            $"Logging directory could not be created '{logDir}'");
+                        logDir = "";
+                    }
+
+                    if (logDir == "")
+                    {
+                        //Should evaluate to ~/Library/Application Support/Ryujinx/Logs
+                        logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
```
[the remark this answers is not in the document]
jcm93: Not really, probably better to use BaseDirPath; changed.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -30,6 +30,8 @@ public enum LaunchMode
         public static string KeysDirPath { get; private set; }
         public static string KeysDirPathUser { get; }
 
+        public static string LogsDirPath { get; private set; }
```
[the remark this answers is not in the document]
jcm93: Removed. As a result, changed `MainWindowViewModel`'s `OpenLogsFolder()` to use `AppDataManager` as well, rather than LoggerModule.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,8 +113,116 @@ public static void Initialize(string baseDirPath)
...
+
+                    try
+                    {
+                        Directory.CreateDirectory(logDir);
+                    }
+                    catch
+                    {
+                        Logger.Warning?.Print(LogClass.Application,
+                            $"Logging directory could not be created '{logDir}'");
+                        logDir = "";
```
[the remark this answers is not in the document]
jcm93: Changed all final exception paths above to return null instead of set logDir to "".

jcm93:
I noticed while addressing feedback that if we pass an empty string or null to `FileLogTarget`'s `PrepareLogFile(string path)`, we have an uncaught exception. I assume we don't want to crash entirely if we can't open the log file, so I might address that in a bit.

jcm93:
Adjusted `FileLogTarget` to catch exception when creating a `DirectoryInfo` for the path. The catch is general; I'm not sure if we can realistically hit anything besides ArgumentException at this point, lemme know if you want to just catch that specifically.

Tested that Ryujinx will now start up properly if it gets a null/empty string when creating the log.

TSRBerry:
> I noticed while addressing feedback that if we pass an empty string or null to `FileLogTarget`'s `PrepareLogFile(string path)`, we have an uncaught exception here:
> 
> https://github.com/Ryujinx/Ryujinx/blob/25123232bd7186d8009332f67124a5991f2a978d/src/Ryujinx.Common/Logging/Targets/FileLogTarget.cs#L26
> 
> I assume we don't want to crash entirely if we can't open the log file, so I might address that in a bit.

It was fine before, since we used null checks before calling that function. But yes, a null check / catch statement is better anyway.

src/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -23,7 +23,17 @@ public FileLogTarget(string name, FileStream fileStream)
...
             // Ensure directory is present
-            DirectoryInfo logDir = new(path);
+            DirectoryInfo logDir;
+            try
+            {
+                logDir = new(path);
+            }
+            catch (Exception exception)
+            {
+                Logger.Warning?.Print(LogClass.Application, $"Logging directory path was invalid: {exception}");
```
TSRBerry, line 33: Catching `ArgumentException` specifically is better imo. I'd also log the invalid path, so we can actually see it in a log.

src/Ryujinx/Ui/MainWindow.cs
```diff
@@ -1378,9 +1378,9 @@ private void OpenLogsFolder_Pressed(object sender, EventArgs args)
         {
             string logPath = System.IO.Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
 
-            if (LoggerModule.LogDirectoryPath != null)
+            if (AppDataManager.LogsDirPath != null)
             {
-                logPath = LoggerModule.LogDirectoryPath;
+                logPath = AppDataManager.LogsDirPath;
             }
```
TSRBerry, line 1384: Same as for Avalonia. We only need to keep `AppDataManager.LogsDirPath` now, since never want to open a different directory here (if it isn't null because then we don't want to open anything).

jcm93:
Addressed, thanks.

TSRBerry (approved):
LGTM, thank you very much!
I'll just add a commit to address some styling issues (so we avoid stalling this for too long).

src/Ryujinx.Common/Logging/Targets/FileLogTarget.cs
```diff
@@ -23,7 +23,18 @@ public FileLogTarget(string name, FileStream fileStream)
...
             // Ensure directory is present
-            DirectoryInfo logDir = new(path);
+            DirectoryInfo logDir;
+            try
+            {
+                logDir = new DirectoryInfo(path);
+            }
+            catch (ArgumentException exception)
+            {
+                Logger.Warning?.Print(LogClass.Application, $"Logging directory path ({path}) was invalid: {exception}");
```
gdkchan, line 33:
```suggestion
                Logger.Warning?.Print(LogClass.Application, $"Logging directory path ('{path}') was invalid: {exception}");
```

src/Ryujinx.Ui.Common/Configuration/LoggerModule.cs
```diff
@@ -9,7 +9,6 @@ namespace Ryujinx.Ui.Common.Configuration
 {
     public static class LoggerModule
     {
-        public static string LogDirectoryPath { get; private set; }
```
gdkchan, line 12:
```suggestion
```

src/Ryujinx.Ava/UI/ViewModels/MainWindowViewModel.cs
```diff
@@ -1350,16 +1350,13 @@ public void OpenRyujinxFolder()
...
         {
-            string logPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Logs");
-
-            if (LoggerModule.LogDirectoryPath != null)
+            string logPath = AppDataManager.LogsDirPath;
+            if (!string.IsNullOrEmpty(logPath))
             {
-                logPath = LoggerModule.LogDirectoryPath;
-            }
+                new DirectoryInfo(logPath).Create();
```
gdkchan, line 1356: This doesn't need exception handling like the Logs folder creation code on Ui.Common? Maybe it should do that there?

gdkchan: BTW I just noticed it uses `Directory.CreateDirectory` there, so not sure the directory is created in a different way here.

jcm93: Hmm do we even need to create a directory here? `OpenHelper` takes the path and checks if the dir exists before opening it, `AppDataManager` should have already created the directory on launch, and I don't think there's any mechanism that would change the log directory while open.

gdkchan: I assume its for the case where the user deleted the folder manually for some reason.

jcm93: Alrighty, I'll just add a try/catch to OpenHelper and remove the create call here and GTK I think then.

src/Ryujinx.Ui.Common/Helper/OpenHelper.cs
```diff
@@ -19,6 +19,14 @@ public static partial class OpenHelper
 
         public static void OpenFolder(string path)
         {
+            try
```
gdkchan, line 22: I'd prefer if this did not try to create the directory since its supposed to be a general helper, and in other cases where it is used, this might not be desirable.
My suggestion was more along the lines of replacing `AppDataManager.LogsDirPath` accesss on `OpenLogsFolder` with something like `AppDataManager.GetOrCreateLogsPath()` that would create the path if needed, log messages on error, and return the path (which would be null/empty if it doesn't exist and was not possible to create). Such method could also be used in other places where it employs similar logic to create the path if it doesn't exist.

jcm93: Ah, my mistake, I wasn't paying proper attention to `OpenHelper`. Will address soon, thanks.

jcm93: Fixed.

gdkchan (approved):
lgtm, thanks.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+            {
+                logDir = Path.Combine(BaseDirPath, "Logs");
+                try
+                {
+                    Directory.CreateDirectory(logDir);
+                }
+                catch
+                {
+                    Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                    return null;
```
AcK77, line 140:
```suggestion

                    return null;
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                        logDir = Path.Combine(BaseDirPath, "Logs");
+
+                        try
+                        {
+                            Directory.CreateDirectory(logDir);
+                        }
+                        catch
+                        {
+                            Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                            return null;
```
AcK77, line 171:
```suggestion

                            return null;
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                    }
+                    catch
+                    {
+                        Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                        logDir = "";
+                    }
+
+                    if (string.IsNullOrEmpty(logDir))
+                    {
+                        // Should evaluate to ~/Library/Application Support/Ryujinx/Logs
```
AcK77, line 161:
```suggestion
                        // NOTE: Should evaluate to "~/Library/Application Support/Ryujinx/Logs".
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                        catch
+                        {
+                            Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                            return null;
+                        }
+                    }
+                }
+                else if (OperatingSystem.IsWindows())
+                {
+                    // Should evaluate to a 'Logs' directory in whatever directory Ryujinx was launched from.
```
AcK77, line 177:
```suggestion
                    // NOTE: Should evaluate to a "Logs" directory in whatever directory Ryujinx was launched from.
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                    }
+                    catch
+                    {
+                        Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                        logDir = "";
+                    }
+
+                    if (string.IsNullOrEmpty(logDir))
+                    {
+                        // Should evaluate to C:\Users\user\AppData\Roaming\Ryujinx\Logs
```
AcK77, line 191:
```suggestion
                        // NOTE: Should evaluate to "C:\Users\user\AppData\Roaming\Ryujinx\Logs".
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                        logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir, "Logs");
+
+                        try
+                        {
+                            Directory.CreateDirectory(logDir);
+                        }
+                        catch
+                        {
+                            Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                            return null;
```
AcK77, line 201:
```suggestion

                            return null;
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                        }
+                    }
+                }
+                else if (OperatingSystem.IsLinux())
+                {
+                    // Should evaluate to ~/.config/Ryujinx/Logs
```
AcK77, line 207:
```suggestion
                    // NOTE: Should evaluate to "~/.config/Ryujinx/Logs".
```

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,120 @@ public static void Initialize(string baseDirPath)
...
+                    logDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir, "Logs");
+
+                    try
+                    {
+                        Directory.CreateDirectory(logDir);
+                    }
+                    catch
+                    {
+                        Logger.Warning?.Print(LogClass.Application, $"Logging directory could not be created '{logDir}'");
+                        return null;
```
AcK77, line 217:
```suggestion

                        return null;
```

AcK77 (changes requested):

jcm93:
@AcK77 addressed.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,124 @@ public static void Initialize(string baseDirPath)
...
 
+        public static string GetOrCreateLogsDir()
+        {
+            if (Directory.Exists(LogsDirPath))
+            {
+                return LogsDirPath;
+            }
+            Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
+            LogsDirPath = SetUpLogsDir();
+            return LogsDirPath;
```
AcK77, line 123:
```suggestion

            Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
            LogsDirPath = SetUpLogsDir();

            return LogsDirPath;
```

AcK77 (commented):
I miss this one, sorry ^^'

src/Ryujinx.Common/Configuration/AppDataManager.cs
jcm93: fixed.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -118,9 +112,125 @@ public static void Initialize(string baseDirPath)
             SetupBasePaths();
         }
 
+        public static string GetOrCreateLogsDir()
+        {
+            if (Directory.Exists(LogsDirPath))
+            {
+                return LogsDirPath;
+            }
+            Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
```
AcK77, line 121:
```suggestion
            }

            Logger.Notice.Print(LogClass.Application, "Logging directory not found; attempting to create new logging directory.");
```

You missed the extra empty line after the `{`. (It's just for consistencies)

AcK77 (approved):

PR #6240, opened 2024-02-01
Vulkan: Improve texture barrier usage, timing and batching

riperiperi:
This PR aims to fix a few issues caused by barriers being in the wrong place, and generally being missing for render pass load and store ops. The main goal here is to fix rendering on Adreno GPUs using the turnip Mesa driver on linux.

## Batching
It's possible that every bound texture in a draw was previously drawn, that every framebuffer load was previously stored, so it makes sense that the barriers for them should be batched into less calls for more resources. The batching works by queuing the requested barrier with pipeline stage flags. The queued barriers are flushed when a command buffer ends, before a render pass begins, before a draw or before a compute dispatch. Flushing the `BarrierBatch` groups barriers by pipeline stage flags and submits all of them in batches with a maximum of 16 of each type.

The main focus for now is image memory barriers for guest bindings and framebuffer attachments, as these are the barriers that are the most important to get right.

Image memory barriers within a render pass have a different meaning from barriers outside one. A barrier inside a render pass can only be between images that are attachments in the subpass, and is scoped entirely within the render pass. Barriers outside render passes have no such restrictions and can be between any texture. Since we don't ever use image attachments, the first case is never useful to us - we only want barriers between resources on a global scale.

By batching barriers until just before a render pass begins, or immediately before the draw/dispatch, we minimize the times that the barrier is accidentally emitted inside the render pass and uses this alternate meaning. Image memory barriers are always submitted as such to the batch, however if they do have to be forcibly inserted by a draw inside a render pass, they will be converted to a generic memory barrier as before.

Even generic memory barriers are subject to the reduced synchronization scope when they are inside a render pass, however it's unclear if drivers will have issues with this - most issues before were likely caused by the ordering of barriers and render pass begin/end. I've currently left it converting to memory barrier, but it does have the opportunity to end the render pass if we discover that it is important somewhere.

## StoreOp -> LoadOp barrier
Vulkan render passes define a load and store op at the end. These are meaningless to a desktop GPU, but for a tile based GPU these loads and stores actually happen, as the data is copied to and from faster tile memory where the actual rasterization happens. It's very possible that you end a render pass that stores back image data, then you immediately attempt to load it back into tile memory before it's finished writing.

This barrier prevents textures from losing data from split render passes this way.

## StoreOp -> Texture Sample
We already have a barrier for texture data rendered to a framebuffer and then sampled later, but the current timing means that something very bad happens when render pass load and store aren't just pointless bureaucracy. The barrier is requested _immediately_, that is, as soon as the backend is told to change the bound texture.

As things are right now, this happens _before_ the active render targets are switched, meaning that we could start binding a texture before its render pass has ended, and insert a barrier before the render pass ends. This is fine if ending the render pass is a no-op, but if does actually store, we just inserted a barrier before the GPU did the work.

This is where the batching process actually helps us out. When a texture is bound, we queue the image barrier so that it happens later. When a render pass is about to begin, or immediately before a draw, we send out all the queued barriers in batch so that the image is ready for the draw to use. This means that they can happen outside the render pass, and work properly regardless of GPU.

## StoreOp -> Image
This is similar to the above case, but it wasn't actually done for some reason. This supercedes PR #4601 . Maybe this fixes something somewhere, since this might impact desktop GPUs as well.

## Other barrier types and future improvement
Image and buffer storage barriers tend to be explicitly inserted by the guest when necessary... however, these are usually within the scope. If a driver takes the synchronization scope rules too seriously, the barrier might need to be repeated after the render pass ends to properly await the barrier before running all future commands. This is something that needs to be seen though.

## Testing
This isn't expected to change Desktop GPUs much. It may fix some weird issues with newer GPUs, as their drivers tend to move more synchronization to be controlled by the barriers rather than by the driver. It shouldn't make things slower, even though it does emit barriers every time a render pass split happens.

More testing on Adreno would also be helpful, I don't have any hardware that's affected by barrier issues (GTX 1070 inserts basically everything implicitly).

### DIFF
--- a/src/Ryujinx.Graphics.Vulkan/BarrierBatch.cs
+++ b/src/Ryujinx.Graphics.Vulkan/BarrierBatch.cs
@@ -1,0 +1,214 @@
+﻿using Silk.NET.Vulkan;
+using System;
+using System.Collections.Generic;
+
+namespace Ryujinx.Graphics.Vulkan
+{
+    internal class BarrierBatch : IDisposable
+    {
+        private const int MaxBarriersPerCall = 16;
+
+        private readonly VulkanRenderer _gd;
+
+        private readonly NativeArray<MemoryBarrier> _memoryBarrierBatch = new(MaxBarriersPerCall);
+        private readonly NativeArray<BufferMemoryBarrier> _bufferBarrierBatch = new(MaxBarriersPerCall);
+        private readonly NativeArray<ImageMemoryBarrier> _imageBarrierBatch = new(MaxBarriersPerCall);
+
+        private readonly List<BarrierWithStageFlags<MemoryBarrier>> _memoryBarriers = new();
+        private readonly List<BarrierWithStageFlags<BufferMemoryBarrier>> _bufferBarriers = new();
+        private readonly List<BarrierWithStageFlags<ImageMemoryBarrier>> _imageBarriers = new();
+        private int _queuedBarrierCount;
+
+        public BarrierBatch(VulkanRenderer gd)
+        {
+            _gd = gd;
+        }
+
+        // New api:
+
+        private readonly record struct StageFlags : IEquatable<StageFlags>
+        {
+            public readonly PipelineStageFlags Source;
+            public readonly PipelineStageFlags Dest;
+
+            public StageFlags(PipelineStageFlags source, PipelineStageFlags dest)
+            {
+                Source = source;
+                Dest = dest;
+            }
+        }
+
+        private readonly struct BarrierWithStageFlags<T> where T : unmanaged
+        {
+            public readonly StageFlags Flags;
+            public readonly T Barrier;
+
+            public BarrierWithStageFlags(StageFlags flags, T barrier)
+            {
+                Flags = flags;
+                Barrier = barrier;
+            }
+
+            public BarrierWithStageFlags(PipelineStageFlags srcStageFlags, PipelineStageFlags dstStageFlags, T barrier)
+            {
+                Flags = new StageFlags(srcStageFlags, dstStageFlags);
+                Barrier = barrier;
+            }
+        }
+
+        private void QueueBarrier<T>(List<BarrierWithStageFlags<T>> list, T barrier, PipelineStageFlags srcStageFlags, PipelineStageFlags dstStageFlags) where T : unmanaged
+        {
+            list.Add(new BarrierWithStageFlags<T>(srcStageFlags, dstStageFlags, barrier));
+            _queuedBarrierCount++;
+        }
+
+        public void QueueBarrier(MemoryBarrier barrier, PipelineStageFlags srcStageFlags, PipelineStageFlags dstStageFlags)
+        {
+            QueueBarrier(_memoryBarriers, barrier, srcStageFlags, dstStageFlags);
+        }
+
+        public void QueueBarrier(BufferMemoryBarrier barrier, PipelineStageFlags srcStageFlags, PipelineStageFlags dstStageFlags)
+        {
+            QueueBarrier(_bufferBarriers, barrier, srcStageFlags, dstStageFlags);
+        }
+
+        public void QueueBarrier(ImageMemoryBarrier barrier, PipelineStageFlags srcStageFlags, PipelineStageFlags dstStageFlags)
+        {
+            QueueBarrier(_imageBarriers, barrier, srcStageFlags, dstStageFlags);
+        }
+
+        public unsafe void Flush(CommandBuffer cb, bool insideRenderPass, Action endRenderPass)
+        {
+            while (_queuedBarrierCount > 0)
+            {
+                int memoryCount = 0;
+                int bufferCount = 0;
+                int imageCount = 0;
+
+                bool hasBarrier = false;
+                StageFlags flags = default;
+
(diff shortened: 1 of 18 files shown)


BoofOof32:
Just coming back to confirm, there have been no noticeable regressions with the up-to-date PR build on my end

GamerzHell9137:
- Fixes Bayonetta Origins: Cereza and the Lost Demon water surfaces on Vulkan

Before: 
image

After:
image

riperiperi:
I've made a few additional changes that could help more games on both mobile and desktop GPUs.

- LoadOp barriers now also handle the read into write barrier. This was handled before only for render target clear, but it seems to be needed for more than that as Bayonetta Origins proves. This was likely not an issue on mobile as the write happens at the end, not during the draws... but it doesn't hurt being safe.
- Switching program iterates all texture/image bindings and emits barriers if necessary, as some bindings may not have changed going into the program (possibly from one frame ago) but have been re-enabled.

I'm going to leave the PR here as it should significantly improve things. Though there is room for improvement in the future, it requires example issues to fully understand what is required. Just keep in mind that the following issues could be causes for future issues, especially on adreno/mali GPUs:
- Mid-render-pass barriers
  - Barriers for image/buffer storage: do they need to be repeated outside the render pass?
  - Image barriers inside a render pass: What platforms (if any) would require render pass splits to get a global synchronization barrier?
- Sampling a texture and rendering it at the same time
  - This is meant to be undefined, but there are really specific circumstances in which games do this and expect some kind of atomic access. This is impossible in Vulkan without converting the draw to use image storage.
  - We don't want to emit barriers between each draw when this happens, as it seems it's possible it happens by accident, or because the sampled image attachment is actually masked out at the same time. It also wouldn't work for TBDR GPUs or overlapping triangles within the same draw.

gdkchan:
I gave it a try on Tears of the Kingdom, and I can see barrier related validation errors still, like for example:
```
Validation Error: [ VUID-vkCmdPipelineBarrier-pDependencies-02285 ] Object 0: handle = 0xbdff4e000000378f, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x3936bc0c | vkCmdPipelineBarrier(): .dstStageMask (0x2) is not a subset of VkSubpassDependency dstAccessMask for any self-dependency of subpass 0 of VkRenderPass 0xbdff4e000000378f[] for which srcAccessMask is also a subset. Candidate VkSubpassDependency are pDependencies entries [0]. The Vulkan spec states: If fname:vkCmdPipelineBarrier is called within a render pass instance using a VkRenderPass object, the render pass must have been created with at least one subpass dependency that expresses a dependency from the current subpass to itself, does not include VK_DEPENDENCY_BY_REGION_BIT if this command does not, and has synchronization scopes and access scopes that are all supersets of the scopes defined in this command (https://vulkan.lunarg.com/doc/view/1.3.250.1/windows/1.3-extensions/vkspec.html#VUID-vkCmdPipelineBarrier-pDependencies-02285)
```
Could be also worth taking a look on `PipelineBase.TextureBarrier`, where src and dst stage mask is just `FragmentShaderBit` which I beleive is incorrect? This is an existing issue though, and the barrier one might be too, I need to check.

riperiperi:
I've changed the location where bindings are iterated after a program change to be right before draws and compute dispatch. Queueing barriers for resources immediately could queue barriers for bindings that were meant for the previous draw, which were unnecessary. After setting the program, the bindings are updated to represent what is actually used. That's all we should care about.

src/Ryujinx.Graphics.Vulkan/BarrierBatch.cs
```diff
@@ -0,0 +1,228 @@
...
+                        {
+                            list.Clear();
+                        }
+                        else
+                        {
+                            int deleteCount = list.Count - firstMatch;
+
+                            list.RemoveRange(firstMatch, deleteCount);
+                        }
+                    }
```
marysaka, line 161: Probably easier to read:
```suggestion
                    if (firstMatch == 0)
                    {
                        list.Clear();
                    }
                    else if (firstMatch != -1)
                    {
                        int deleteCount = list.Count - firstMatch;

                        list.RemoveRange(firstMatch, deleteCount);
                    }
```

marysaka (approved):

src/Ryujinx.Graphics.Vulkan/BarrierBatch.cs
riperiperi: I'm not sure, there's more conditions but the wrapping `firstMatch != -1` means that "there is a match", and whether to clear or trim the list comes from that.

riperiperi: I'll go for it, since
```
                    if (firstMatch != -1)
                    {
                        if (firstMatch == 0)
                        {
```
does look kind of bad.

gdkchan (approved):
lgtm, thanks. Very nice to see the barriers issues finally fixed! The barrier validation errors I mentioned before is due to the barrier on `CommandBufferBarrier`, so its an existing issue, not sure how to fix it.

Needs a rebase now.

riperiperi:
I've rebased. The conflicts were just it being confused at the change to SetProgram arguments from the push descriptor stuff.

PR #5941, opened 2023-11-16
Fix macOS Path on .NET 8

IsaacMarovitz:
Post .NET 8 has changed `SpecialFolder.Personal` to point to `~/Documents` which resulted in Ryujinx creating/looking for its folder in a user's document directory, rather than actually in Application Support.

Fixes #5940

PR #6241, opened 2024-02-02
macOS: Stop storing user data in Documents for some users; fix symlinks

jcm93:
### Context

When .NET 8 was merged, there were some breaking changes to `Environment.SpecialFolder` values on macOS. One was immediately noticed, the change of `SpecialFolder.Personal` from `~/` to `~/Documents/`. This was addressed by https://github.com/Ryujinx/Ryujinx/pull/5941. Another important change was that of `SpecialFolder.ApplicationData` from `~/.config/` to `~/Library/Application Support`, correcting a longstanding discrepancy between macOS and Linux.

Now let's briefly go back in time to a commit dating from 2020 in `AppDataManager`, which got ahead of .NET, moving Mac configuration files from `~/.config/Ryujinx/` to `~/Library/Application Support/Ryujinx` to fix this discrepancy. The code block that does this is partially excerpted below:

```c#
if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
{
    if (Path.Exists(oldConfigPath) && !IsPathSymlink(oldConfigPath) && !Path.Exists(BaseDirPath))
    {
        FileSystemUtils.MoveDirectory(oldConfigPath, BaseDirPath);
        Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
    }
}
```

Here, on every startup, the existence of configuration files at the old config path, `~/.config/Ryujinx/`, is checked, and, if detected, they are moved to the new area, creating a symlink that links `~/.config/Ryujinx` to `~/Library/Application Support/Ryujinx` in the process.

Unfortunately, the key it used to describe the folder when checking for the existence of old configuration files was `SpecialFolder.ApplicationData`. Perhaps you can see where this is going.

When .NET 8 landed, `oldConfigPath` started evaluating to what should be *correct* directory, and the new, "correct" directory was an entirely incorrect one in `~/Documents`. The above code block, however, was still happily running, and moving all user files to the nonsensical folder `~/Documents/Library/Application Support/Ryujinx`, with a symlink being created at` ~/Library/Application Support/Ryujinx`.

Once https://github.com/Ryujinx/Ryujinx/pull/5941 was committed, Ryujinx began looking in the correct location for user data once more, and found it, but only because this directory was a symlink leading to ~/Documents.

This was actively noticed by both users and developers of Ryujinx, but was for some reason unaddressed. This is an editorial aside, but this represents a fairly large testing and process failure in my opinion. Not noticing this issue ahead of time is, of course, excusable, but noticing it and failing to fix it is less so. Because:

> [!IMPORTANT]
> Any person that updated and ran Ryujinx in the two days between 1084 and 1088 has had their saves and application data stored in this nonsensical folder ever since, vulnerable to wipes, iCloud or Dropbox shenanigans, or simply being cleaned out by a reasonable person wondering why `~/Documents/Library` exists.

### Now

This PR addresses all of the above.

* Config version is incremented, though no file format changes are made.
* The original code designed to migrate from `~/.config` to `~/Library/Application Support/` has been changed to run exactly once, on config update, and now uses a more explicit path to target the long-deprecated config path.
* On config update we will also run, exactly once, a check to see if the correct configuration directory, `~/Library/Application Support/Ryujinx`, is a symlink, and if so, we will remove it, and copy the files from the resolved path to the original path. 

### Testing

This has been tested locally on my Mac machine and verified to fix the symlink issue. It should be tested by other Mac users, especially those with the symlink resident in their Application Support folders.

There is some chance for this process to fail, if for some strange reason the Application Support directory isn't writable. However, I do not see a particularly good way around this externality. I would prefer not to run this code every time Ryujinx starts up. If this fails on a user's machine that has the symlink in their Application Support folder, they may just be stuck with it unless they manually resolve the situation. I added a log statement to `AppDataManager` to report if the user's data directory is a symlink, so that this circumstance will at least be apparent via support channels.

Please leave any feedback if anyone sees a better way to handle any aspect of this.

### DIFF
--- a/src/Ryujinx.Common/Configuration/AppDataManager.cs
+++ b/src/Ryujinx.Common/Configuration/AppDataManager.cs
@@ -95,18 +95,9 @@ namespace Ryujinx.Common.Configuration
 
             BaseDirPath = Path.GetFullPath(BaseDirPath); // convert relative paths
 
-            // NOTE: Moves the Ryujinx folder in `~/.config` to `~/Library/Application Support` if one is found
-            // and a Ryujinx folder does not already exist in Application Support.
-            // Also creates a symlink from `~/.config/Ryujinx` to `~/Library/Application Support/Ryujinx` to preserve backwards compatibility.
-            // This should be removed in the future.
-            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
+            if (IsPathSymlink(BaseDirPath))
             {
-                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
-                if (Path.Exists(oldConfigPath) && !IsPathSymlink(oldConfigPath) && !Path.Exists(BaseDirPath))
-                {
-                    FileSystemUtils.MoveDirectory(oldConfigPath, BaseDirPath);
-                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
-                }
+                Logger.Error?.Print(LogClass.Application, $"Application data directory is a symlink. This may be unintended.");
             }
 
             SetupBasePaths();
@@ -243,6 +234,62 @@ namespace Ryujinx.Common.Configuration
         {
             FileAttributes attributes = File.GetAttributes(path);
             return (attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint;
+        }
+
+        public static void FixMacOSConfigurationFolders()
+        {
+            if (OperatingSystem.IsMacOS())
+            {
+                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+                    ".config", DefaultBaseDir);
+                if (Path.Exists(oldConfigPath) && !IsPathSymlink(oldConfigPath) && !Path.Exists(BaseDirPath))
+                {
+                    FileSystemUtils.MoveDirectory(oldConfigPath, BaseDirPath);
+                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
+                }
+
+                string correctApplicationDataDirectoryPath =
+                    Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
+                if (IsPathSymlink(correctApplicationDataDirectoryPath))
+                {
+                    //copy the files somewhere temporarily
+                    string tempPath = Path.Combine(Path.GetTempPath(), DefaultBaseDir);
+                    try
+                    {
+                        FileSystemUtils.CopyDirectory(correctApplicationDataDirectoryPath, tempPath, true);
+                    }
+                    catch
+                    {
+                        Logger.Error?.Print(LogClass.Application,
+                            $"Critical error copying Ryujinx application data into the temp folder. Please manually move your application data to the correct folder.");
+                        return;
+                    }
+
+                    //delete the symlink
+                    try
+                    {
+                        //This will fail if this is an actual directory, so there is no way we can actually delete user data here.
+                        File.Delete(correctApplicationDataDirectoryPath);
+                    }
+                    catch
+                    {
+                        Logger.Error?.Print(LogClass.Application,
+                            $"Critical error deleting the Ryujinx application data folder symlink. Please manually move your application data to the correct location.");
+                        return;
+                    }
+
+                    //put the files back
+                    try
+                    {
+                        FileSystemUtils.CopyDirectory(tempPath, correctApplicationDataDirectoryPath, true);
+                    }
+                    catch
+                    {
+                        Logger.Error?.Print(LogClass.Application,
+                            $"Critical error copying Ryujinx application data into the correct location. Please manually move your application data to the correct folder.");
+                    }
+                }
+            }
         }
 
         public static string GetModsPath() => CustomModsPath ?? Directory.CreateDirectory(Path.Combine(BaseDirPath, DefaultModsDir)).FullName;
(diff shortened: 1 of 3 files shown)


TSRBerry:
> this represents a fairly large testing and process failure

I completely agree with this. :/

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -134,6 +118,62 @@ private static bool IsPathSymlink(string path)
             return (attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint;
         }
 
+        public static void FixMacOSConfigurationFolders()
+        {
+            if (OperatingSystem.IsMacOS())
+            {
```
TSRBerry, line 124:
```suggestion
        [SupportedOSPlatform("macos")]
        public static void FixMacOSConfigurationFolders()
        {
```
(Make sure to remove the if statement manually and not using the suggestion here.)

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -134,6 +118,62 @@ private static bool IsPathSymlink(string path)
...
+                    //copy the files somewhere temporarily
+                    string tempPath = Path.Combine(Path.GetTempPath(), DefaultBaseDir);
+                    try
+                    {
+                        FileSystemUtils.CopyDirectory(correctApplicationDataDirectoryPath, tempPath, true);
+                    }
+                    catch
+                    {
+                        Logger.Error?.Print(LogClass.Application,
+                            $"Critical error copying Ryujinx application data into the temp folder. Please manually move your application data to the correct folder.");
```
TSRBerry, line 146: Could you make sure to log the exception (at least the message) for these as well?

TSRBerry (changes requested):
Just minor changes, otherwise lgtm!

(I didn't comment on style issues, but will fix them once, you addressed my review comments if you don't mind.)

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -101,18 +94,9 @@ public static void Initialize(string baseDirPath)
...
-            if (OperatingSystem.IsMacOS() && Mode == LaunchMode.UserProfile)
+            if (IsPathSymlink(BaseDirPath))
             {
-                string oldConfigPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), DefaultBaseDir);
-                if (Path.Exists(oldConfigPath) && !IsPathSymlink(oldConfigPath) && !Path.Exists(BaseDirPath))
-                {
-                    FileSystemUtils.MoveDirectory(oldConfigPath, BaseDirPath);
-                    Directory.CreateSymbolicLink(oldConfigPath, BaseDirPath);
-                }
+                Logger.Error?.Print(LogClass.Application, $"Application data directory is a symlink. This may be unintended.");
```
TSRBerry, line 99:
```suggestion
                Logger.Warn?.Print(LogClass.Application, $"Application data directory is a symlink. This may be unintended.");
```
Feels more like a warning than an error.

jcm93:
Thanks, all addressed.

jcm93:
I mentioned it over in #6238, but #6238 should be merged before this. I think the probability of the delete/copy operation failing is higher if the bundle's code signature is invalid, which it is until we stop logging inside the bundle. I am not actually sure why more Mac users are not currently having issues launching Ryujinx; so maybe it's not as big a deal as the documentation implies, but better to be cautious and fix that first.

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -134,6 +119,63 @@ private static bool IsPathSymlink(string path)
...
+                string tempPath = Path.Combine(Path.GetTempPath(), DefaultBaseDir);
+
+                try
+                {
+                    FileSystemUtils.CopyDirectory(correctApplicationDataDirectoryPath, tempPath, true);
+                }
+                catch (Exception exception)
+                {
+                    Logger.Error?.Print(LogClass.Application,
+                        $"Critical error copying Ryujinx application data into the temp folder. Please manually move your application data to the correct folder. {exception}");
```
TSRBerry, line 149: Could you log which directory needs to be copied where?

src/Ryujinx.Common/Configuration/AppDataManager.cs
```diff
@@ -134,6 +119,63 @@ private static bool IsPathSymlink(string path)
...
+
+                // Put the files back
+                try
+                {
+                    FileSystemUtils.CopyDirectory(tempPath, correctApplicationDataDirectoryPath, true);
+                }
+                catch (Exception exception)
+                {
+                    Logger.Error?.Print(LogClass.Application,
+                        $"Critical error copying Ryujinx application data into the correct location. Please manually move your application data to the correct folder. {exception}");
```
TSRBerry, line 174: Could you log the temp and the correct directory as well?

TSRBerry (approved):
Lgtm! Just one minor thing i just noticed.

src/Ryujinx.Common/Configuration/AppDataManager.cs
jcm93: Resolving the symlink takes another try/catch, but added in new commit.

src/Ryujinx.Common/Configuration/AppDataManager.cs
jcm93: Done.

jcm93:
In addition to above, I also noticed that I hadn't set `configurationFileUpdated = true;` previously, whoops. Fixed.

marysaka (approved):
lgtm, just need a rebase :+1:

PR #6245, opened 2024-02-02
LightningJit: Reduce stack usage for Arm32 code

### LINKED ISSUE #6236: [Bug] Crash - Gothic 1 - and different moment of crash based on how I launch Ryujinx
### Description of the issue

I'm trying to run Gothic 1 and the game crashes. But what is interesting, it crashes in different ways based on how I launch Ryujinx.

1st launch method (GUI, like running an app in a regular user way): <kbd>⌘ Command</kbd> <kbd>Space</kbd> -> Ryujinx -> start Gothic

The game launches with no problems and I can select a new game in the game's menu. The loading begins and I can watch the intro movie, then the loading continues but before reaching 100% the game crashes together with Ryujinx. See the 1st log file.

2nd launch method (terminal): command line: `/Applications/Ryujinx.app/Contents/MacOS/Ryujinx` -> start Gothic

The game does not launch at all and the game crashes together with Ryujinx immediately. See the 2nd log file.

What's also interesting, both launch methods produce no crash when running a different game (Mario Wonder).

### Reproduction steps

You probably need M1 Pro Apple computer to observe the crashes, I couldn't reproduce on other machines. On the M1 Pro machine I can reproduce this behaviour every single time.
- Download Ryujinx
- put it into /Applications/
- run Ryujinx from macOS GUI and click open Ryujinx directory
- add keys
- add 17.0.0 firmware
- add Gothic 1 and its 1.0.2 patch
- configure Pro Controller
- close Ryujinx

1st

- run Ryujinx from macOS GUI
- run the game - it launches
- select new game
- skip the intro
- it loads for a second
- it crashes before the loading ends (1st log)

2nd

- run Ryujinx from terminal
- try to run the game - crashes immediately (2nd log)

### Log file

1st log:
[Ryujinx_1.1.1154_2024-02-01_17-34-27.log](https://github.com/Ryujinx/Ryujinx/files/14128362/Ryujinx_1.1.1154_2024-02-01_17-34-27.log)

2nd log:
[Ryujinx_1.1.1154_2024-02-01_17-36-09.log](https://github.com/Ryujinx/Ryujinx/files/14128366/Ryujinx_1.1.1154_2024-02-01_17-36-09.log)

2nd log - in terminal part:
```
00:00:02.855 |S| HLE.OsThread.47 ServiceAm SetOperationModeChangedNotification: Stubbed. {operationModeChangedNotification: True}
Stack overflow.
   at Ryujinx.Cpu.LightningJit.Translator.Execute(Ryujinx.Cpu.LightningJit.State.ExecutionContext, UInt64)
   at Ryujinx.Cpu.LightningJit.LightningJitCpuContext.Execute(Ryujinx.Cpu.IExecutionContext, UInt64)
   at Ryujinx.HLE.HOS.ArmProcessContext`1[[System.__Canon, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].Execute(Ryujinx.Cpu.IExecutionContext, UInt64)
   at Ryujinx.HLE.HOS.Kernel.Threading.KThread.ThreadStart()
   at System.Threading.Thread+StartHelper.RunWorker()
   at System.Threading.Thread+StartHelper.Run()
[1]    91275 abort      /Applications/Ryujinx.app/Contents/MacOS/Ryujinx
```

### OS

macOS Sonoma 14.3

### Ryujinx version

1.1.1154

### Game version

1.0.2

### CPU

M1 Pro

### GPU

M1 Pro

### RAM

16GB

### List of applied mods

_No response_

### Additional context?

_No response_

gdkchan:
When we do a managed call on JIT generated code, we need to place all caller saved registers somewhere, since the managed function might overwrite their values. Normally we store them on the stack, however for guest state, we can store them in the context struct, and avoid using the stack this way. This PR changes the "sync point"/interrupt check to use the guest context to store the guest register values, instead of the stack. Then, it changes the stack reservation size calculation to account for the cases where we store the guest states to context, in those cases we only need to reserve 16 bytes as we will only store 2 pointers, the context pointer and the guest address space base pointer.

This reduces the stack usage by a lot in some cases, and notably fixes #6236 which is caused by a stack overflow.
301745199-c4e271be-e9a2-42ae-8650-32091e942934
With the default macOS stack size (512KB), it crashes almost right away. With this change, it goes to menu but has another stack overflow when trying to go in-game. With the increased stack size (2MB) which is what we use, it can go in-game and I did not get any stack overflow, at least as far as I have tested.

Testing is welcome. It should be tested with 32-bit games on Arm platforms, it should not really affect anything else.

### DIFF
--- a/src/Ryujinx.Cpu/LightningJit/Arm32/RegisterAllocator.cs
+++ b/src/Ryujinx.Cpu/LightningJit/Arm32/RegisterAllocator.cs
@@ -106,6 +106,7 @@ namespace Ryujinx.Cpu.LightningJit.Arm32
                 if ((regMask & AbiConstants.ReservedRegsMask) == 0)
                 {
                     _gprMask |= regMask;
+                    UsedGprsMask |= regMask;
 
                     return firstCalleeSaved;
                 }
--- a/src/Ryujinx.Cpu/LightningJit/Arm32/Target/Arm64/Compiler.cs
+++ b/src/Ryujinx.Cpu/LightningJit/Arm32/Target/Arm64/Compiler.cs
@@ -607,7 +607,8 @@ namespace Ryujinx.Cpu.LightningJit.Arm32.Target.Arm64
                 name == InstName.Ldm ||
                 name == InstName.Ldmda ||
                 name == InstName.Ldmdb ||
-                name == InstName.Ldmib)
+                name == InstName.Ldmib ||
+                name == InstName.Pop)
             {
                 // Arm32 does not have a return instruction, instead returns are implemented
                 // either using BX LR (for leaf functions), or POP { ... PC }.
--- a/src/Ryujinx.Cpu/LightningJit/Arm32/Target/Arm64/Compiler.cs
+++ b/src/Ryujinx.Cpu/LightningJit/Arm32/Target/Arm64/Compiler.cs
@@ -305,12 +305,23 @@ namespace Ryujinx.Cpu.LightningJit.Arm32.Target.Arm64
                 ForceConditionalEnd(cgContext, ref lastCondition, lastConditionIp);
             }
 
+            int reservedStackSize = 0;
+
+            if (multiBlock.HasHostCall)
+            {
+                reservedStackSize = CalculateStackSizeForCallSpill(regAlloc.UsedGprsMask, regAlloc.UsedFpSimdMask, UsablePStateMask);
+            }
+            else if (multiBlock.HasHostCallSkipContext)
+            {
+                reservedStackSize = 2 * sizeof(ulong); // Context and page table pointers.
+            }
+
             RegisterSaveRestore rsr = new(
                 regAlloc.UsedGprsMask & AbiConstants.GprCalleeSavedRegsMask,
                 regAlloc.UsedFpSimdMask & AbiConstants.FpSimdCalleeSavedRegsMask,
                 OperandType.FP64,
-                multiBlock.HasHostCall,
-                multiBlock.HasHostCall ? CalculateStackSizeForCallSpill(regAlloc.UsedGprsMask, regAlloc.UsedFpSimdMask, UsablePStateMask) : 0);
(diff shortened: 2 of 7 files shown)


src/Ryujinx.Cpu/LightningJit/Arm32/RegisterAllocator.cs
```diff
@@ -106,6 +106,7 @@ private int AllocateTempRegisterWithPreferencing()
                 if ((regMask & AbiConstants.ReservedRegsMask) == 0)
                 {
                     _gprMask |= regMask;
+                    UsedGprsMask |= regMask;
```
gdkchan, line 109: This was an oversight, but does not cause issues. Not setting those registers as used means we would not save/restore them on the prologue/epilogue, but since all the functions set them to the same values and same registers on Arm32, it doesn't really matter. In any case, I think it's better to do the right thing here in case the code is changed in a way that can place different values on those registers.

Causes a little bit of code size increase and can have a very minor performance impact.

src/Ryujinx.Cpu/LightningJit/Arm32/Target/Arm64/Compiler.cs
```diff
@@ -596,7 +607,8 @@ private static void RewriteIndirectBranchInstructionWithTarget(in Context contex
                 name == InstName.Ldm ||
                 name == InstName.Ldmda ||
                 name == InstName.Ldmdb ||
-                name == InstName.Ldmib)
+                name == InstName.Ldmib ||
+                name == InstName.Pop)
```
gdkchan, line 611: I think this was an oversight too, but `POP` is only used for thumb code, and since almost no game uses it I think we just didn't have enough usage to hit an issue.

riperiperi (approved):
Appears to work fine, nice catch. If it's possible to avoid storing context as much in future by just leaving everything in the registers between translated functions, then it would be interesting to see how much faster that could be.

PR #6246, opened 2024-02-02
Ensure SM service won't listen to closed sessions

gdkchan:
If a port was closed, we should stop listening to it, otherwise kernel will return the `PortRemoteClosed` error non-stop and we won't be signalled when other services try to connect to SM. This is a bug that was happening if anything else tried to connect to SM after a process has exited, it started happening after #5802 because it uses SM to connect to HLE services from other HLE services. It caused everything that uses the JIT service (NSO Nintendo 64, etc) to lock up upon trying to load said service.

This should also fix the "loading an applet from another applet" scenario from locking up, for the people that was messing with that.

### DIFF
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -306,6 +306,7 @@ namespace Ryujinx.HLE.HOS.Services
         private void DestroySession(int serverSessionHandle)
         {
             _context.Syscall.CloseHandle(serverSessionHandle);
+
             if (RemoveSessionObj(serverSessionHandle, out var session))
             {
                 (session as IDisposable)?.Dispose();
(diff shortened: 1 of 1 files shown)


LDj3SNuD (approved):
Tested.

src/Ryujinx.HLE/HOS/Services/ServerBase.cs
```diff
@@ -299,6 +303,15 @@ private void ServerLoop()
             Dispose();
         }
 
+        private void DestroySession(int serverSessionHandle)
+        {
+            _context.Syscall.CloseHandle(serverSessionHandle);
```
LDj3SNuD, line 308: maybe newline

and maybe align `request.Type == IpcMessageType.CmifControlWithContext` @ line 376.

riperiperi (approved):

PR #6247, opened 2024-02-02
Remove component operand for texture gather with depth compare

gdkchan:
Yet another case of redundant operands being added to the list, and it causing problems later on. Texture gather with depth compare does not need a RGBA component being specified, because what is being sampled is a depth texture, so there is only the depth to select.

This fixes a crash translating shaders on the game Hotshot Racing on Vulkan.
image
Testing is welcome.


riperiperi (approved):

marysaka (approved):

legend800:
@gdkchan Was this PR intended to fix the striping issue? I still see it on all backends/configs.


This is how it should look:

gdkchan:
No, it is supposed to fix a crash on Vulkan, as noted in the PR description.

PR #6248, opened 2024-02-03
Limit remote closed session removal to SM service

gdkchan:
#6246 caused a regression on "multi-program" games (those that have multiple program NCAs inside and needs relaunch). This is working around the issue by only enabling the new behaviour for SM, while keeping what we did before the linked change for other services. This still works for games using the JIT service while not regressing multiple program applications.

Ultimately this code should be deleted and fully replaced by new IPC code, so I think this solution is good enough for now.


riperiperi (approved):
This appears to fix the error. 😌

TSRBerry (approved):

PR #6250, opened 2024-02-03
GPU: Clear fragment enable bit on draw

riperiperi:
In a few games, depth only draws seem to have random fragment shaders attached to them. In Pokemon Legends Arceus, this isn't too much of an issue as it has rasterize discard enabled, though it still binds resources for an unused fragment stage. However, in Assassin's Creed 2, an alpha testing shader is left bound during the majority of the occlusion culling pass, leading pretty much every fragment to fail and all geometry to be culled out.

This PR clears the fragment stage enable bit after each draw, as the commands set by the guest should re-enable it if it's being used. This means that stale fragment shader bindings are removed during depth-only draws, which can avoid incorrect results if the fragment shader could discard.

All guest GPU drivers seem to religiously set the enable bit for the fragment shader, other stages are better managed (they never stay accidentally enabled, and resetting them can break the guest OGL driver). I traced all the registers AC2 wrote when it did the incorrect draws, and this is the only conclusion I could come to.

It's worth testing this in a basically every game to make sure it's not breaking things. It's entirely possible that this flag is reset in another place, or under specific conditions. Guest OpenGL games are a good one to make sure still work.

### DIFF
--- a/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
+++ b/src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
@@ -319,6 +319,9 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
                 _context.Renderer.Pipeline.BeginTransformFeedback(_drawState.Topology);
                 _prevTfEnable = true;
             }
+
+            // Unset fragment enable bits for shader, as they need to be reset each draw.
+            _state.State.ShaderState[(int)ShaderStage.Fragment].Control &= ~(uint)1;
         }
 
         /// <summary>
@@ -1398,7 +1401,7 @@ namespace Ryujinx.Graphics.Gpu.Engine.Threed
 
             for (int index = 0; index < 6; index++)
             {
-                var shader = _state.State.ShaderState[index];
+                ref var shader = ref _state.State.ShaderState[index];
                 if (!shader.UnpackEnable() && index != 1)
                 {
                     continue;
--- a/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
+++ b/src/Ryujinx.HLE/HOS/Services/ServerBase.cs
@@ -287,7 +287,7 @@ namespace Ryujinx.HLE.HOS.Services
                             _wakeEvent.WritableEvent.Clear();
                         }
                     }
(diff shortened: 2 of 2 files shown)


riperiperi:
Before:
https://github.com/Ryujinx/Ryujinx/assets/6294155/27027490-5575-4cb4-8412-3076b8b3b541

After:
https://github.com/Ryujinx/Ryujinx/assets/6294155/c759884c-d5e7-4416-b5d2-06beeeb59f3e

Makes a much bigger difference in interior scenes, such as in Abstergo. Outdoors it manifests as parts of the world going missing. Game can still flicker due to occlusion query issues. OpenGL flashes alternating every frame, so it's only really playable on Vulkan.

GamerzHell9137:
Alan wake is same when you increase the resolution and disable vsync.

src/Ryujinx.Graphics.Gpu/Engine/Threed/StateUpdater.cs
```diff
@@ -319,6 +319,9 @@ public void Update()
                 _context.Renderer.Pipeline.BeginTransformFeedback(_drawState.Topology);
                 _prevTfEnable = true;
             }
+
+            // Unset fragment enable bits for shader, as they need to be reset each draw.
+            _state.State.ShaderState[(int)ShaderStage.Fragment].Control &= ~(uint)1;
```
gdkchan, line 324:
```suggestion
            _state.State.ShaderState[(int)ShaderStage.Fragment].Control &= ~1u;
```

gdkchan (commented):
Seems a bit odd that this is only done for fragment shaders. If doing it for other stages breaks OpenGL games, I guess they are using geometry or "vertex A" shader then? Regular vertex stage is assumed to be always enabled and shouldn't matter.

Also, if the game explicitly enables vertex and fragment shader on one draw, and on the next it doesn't touch the shader state at all, will this not keep the same vertex + fragment program, instead of removing fragment and using a vertex only shader? In other words, even if it clears the enable bit, it does not dirty the state on the state tracker, so this will only really have an effect if the game does happen to change some shader related state.

LukeWarnut:
Causes the tree textures in Kadingir Sanctum to break/flicker in DOOM 2016 v1.2. This does not occur in release 1.1.1156 or the current master build.
image

riperiperi:
This was drafted as it isn't correct. There's some other trigger for this that isn't obvious.

